PR #172, opened 2022-08-02
feat: bad bits implementation

Arqu:
First run against the bad bits list. It was a bit hard to know how the hashes were derived/utilized. Ended up following https://github.com/nftstorage/nft.storage/pull/1721/files as the source of it.

The nft.storage implementation doesn't do its own resolution so they just match on top level CIDs. 
I've added path, CID and resolved CID matching just to make sure we cover all the bases.

There's relevant discussion happening here https://github.com/ipfs/specs/pull/299/files but until this solidifies we just run the basics.


iroh-gateway/src/bad_bits.rs
```diff
@@ -0,0 +1,204 @@
...
+                    .collect();
+                bad_bits.write().await.update(new_denylist);
+                println!(
+                    "updated denylist: len={}",
+                    bad_bits.read().await.denylist.len()
+                );
+            } else {
+                error!("Failed to fetch denylist: {}", res.status());
+            }
+            tokio::time::sleep(Duration::from_secs(30)).await;
```
Arqu, line 77: eek need to bump this up to like 8hrs or so

iroh-gateway/Cargo.toml
```diff
@@ -51,6 +51,8 @@ toml = "0.5.9"
 http-serde = "1.1.0"
 config = "0.13.1"
 async-trait = "0.1.56"
+rust-crypto = "0.2.36"
```
dignifiedquire, line 54: this is outdated, please use individual crates from https://github.com/rustcrypto

iroh-gateway/src/bad_bits.rs
```diff
@@ -0,0 +1,206 @@
...
+impl Default for BadBits {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+pub fn bad_bits_update_handler(bad_bits: Arc<RwLock<BadBits>>) -> JoinHandle<()> {
+    tokio::spawn(async move {
+        loop {
+            let denylist_uri = "http://badbits.dwebops.pub/denylist.json";
```
dignifiedquire, line 62: configurable? or at least a constant

iroh-gateway/src/bad_bits.rs
```diff
@@ -0,0 +1,206 @@
...
+            let res = reqwest::get(denylist_uri).await.unwrap();
+            if res.status().is_success() {
+                let body = res.bytes().await.unwrap();
```
dignifiedquire, line 65: How large is this list currently? We should have at least a check here that it is not larger than X bytes to avoid run away memory allocations

iroh-gateway/src/bad_bits.rs
```diff
@@ -0,0 +1,206 @@
...
+                let body = serde_json::from_slice::<Vec<BadBitsAnchor>>(&body[..]).unwrap();
```
dignifiedquire, line 66: you can use `https://docs.rs/reqwest/latest/reqwest/struct.Response.html#method.json` directly, avoiding the manual call to `bytes()`

iroh-gateway/src/main.rs
```diff
@@ -66,16 +74,29 @@ async fn main() -> Result<()> {
...
+        rpc_addr,
+        gw_metrics,
+        &mut prom_registry,
+        Arc::clone(&bad_bits),
+    )
+    .await?;
 
+    let bad_bits_handle = match use_denylist {
+        true => bad_bits::bad_bits_update_handler(bad_bits),
+        false => tokio::spawn(async move {}),
```
dignifiedquire, line 98: just make the handle an `Option`

iroh-gateway/src/bad_bits.rs
```diff
@@ -0,0 +1,206 @@
...
+                let new_denylist: HashMap<String, bool> = body
+                    .into_iter()
+                    .map(|anchor| (anchor.anchor, true))
+                    .collect();
+                bad_bits.write().await.update(new_denylist);
```
dignifiedquire, line 71: instead of allocating a new hashmap, why not just update the old one?

iroh-gateway/src/bad_bits.rs
```diff
@@ -0,0 +1,206 @@
...
+
+pub fn bad_bits_update_handler(bad_bits: Arc<RwLock<BadBits>>) -> JoinHandle<()> {
+    tokio::spawn(async move {
+        loop {
+            let denylist_uri = "http://badbits.dwebops.pub/denylist.json";
+            let res = reqwest::get(denylist_uri).await.unwrap();
+            if res.status().is_success() {
+                let body = res.bytes().await.unwrap();
+                let body = serde_json::from_slice::<Vec<BadBitsAnchor>>(&body[..]).unwrap();
+                let new_denylist: HashMap<String, bool> = body
```
dignifiedquire, line 67: are these really `String`s?

iroh-gateway/src/bad_bits.rs
```diff
@@ -0,0 +1,206 @@
...
+                    .into_iter()
+                    .map(|anchor| (anchor.anchor, true))
```
dignifiedquire, line 69: What's the point of always storing `true`, why not use a `HashSet`?

iroh-gateway/src/bad_bits.rs
Arqu: 11.5k entries, slightly bellow 1mb. Added a 8mb cap.

iroh-gateway/src/bad_bits.rs
Arqu: Want to avoid locking too much not to introduce too much jitter on update so a simple swap should do much better.
Also while it probably won't happen things might also get taken off the list ie something was blocked due to legal reasons and then nullified. Instead of searching and filtering for each might as well just take the latest representative list.

iroh-gateway/src/bad_bits.rs
Arqu: 👍 I started out with a more fleshed out map because there's the notion of also assigning status codes/explanations for the block but not here yet.

iroh-gateway/src/bad_bits.rs
Arqu: Basically an array of bytes. Unsure if we really "gain" that much from it since it's all in a HashSet anyways now.

iroh-gateway/src/bad_bits.rs
```diff
@@ -0,0 +1,278 @@
...
+pub fn bad_bits_update_handler(bad_bits: Arc<RwLock<BadBits>>) -> JoinHandle<()> {
+    tokio::spawn(async move {
+        loop {
+            let denylist_uri = DEFAULT_DENY_LIST_URI;
+            let res = reqwest::get(denylist_uri).await.unwrap();
+            if res.status().is_success() {
+                let body = res.bytes().await.unwrap();
+                if body.len() > 1 << 26 {
+                    // 64MB
+                    error!("denylist too large");
```
dignifiedquire, line 81: maybe print the size?

iroh-gateway/src/bad_bits.rs
```diff
@@ -0,0 +1,278 @@
...
+                } else {
+                    let body = serde_json::from_slice::<Vec<BadBitsAnchor>>(&body[..]).unwrap();
+                    let new_denylist: HashSet<String> =
+                        body.into_iter().map(|anchor| anchor.anchor).collect();
+                    bad_bits.write().await.update(new_denylist);
+                    println!(
```
dignifiedquire, line 87: this shouldn’t be a println

iroh-gateway/src/bad_bits.rs
```diff
@@ -0,0 +1,278 @@
...
+                hasher.update(path);
+            } else {
+                hasher.update("/");
+                hasher.update(path);
+            }
+        } else {
+            hasher.update("/");
+        }
+        let x = hasher.finalize().to_vec();
+        let mut s = String::new();
```
dignifiedquire, line 58: why store the hex in memory, could just store the hash and deserialize that when receiving the deny list. this is pretty expensive given how often we check it

iroh-gateway/src/bad_bits.rs
```diff
@@ -69,6 +72,13 @@ impl Default for BadBits {
     }
 }
 
+pub fn decode_hex(s: &str) -> Result<Vec<u8>, std::num::ParseIntError> {
```
dignifiedquire, line 73: https://docs.rs/hex/

iroh-gateway/src/core.rs
```diff
@@ -185,13 +190,36 @@ async fn get_handler(
     service_worker_check(&request_headers, cpath.to_string(), &state)?;
     unsuported_header_check(&request_headers, &state)?;
 
+    // check if cid is in the denylist
+    if state.bad_bits.read().await.is_bad(cid, cpath) {
```
dignifiedquire, line 194: this adds overhead atm if not enabled, I would make bad_bits be an Option, so it is clear if it‘s enabled at all

iroh-gateway/src/bad_bits.rs
```diff
@@ -0,0 +1,279 @@
...
+    }
+}
+
+impl Default for BadBits {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+pub fn bad_bits_update_handler(bad_bits: Arc<RwLock<BadBits>>) -> JoinHandle<()> {
```
dignifiedquire, line 73: I woul call this sth like run_ or spawn_ to indicate that it spawns a task

dignifiedquire (approved):

PR #173, opened 2022-08-02
fix: fix builds

Arqu:
🤷 ~~guess this fixes it?~~ or not, worked locally...

### DIFF
--- a/iroh-gateway/src/config.rs
+++ b/iroh-gateway/src/config.rs
@@ -65,18 +65,20 @@ impl Config {
 
     /// Derive server addr for non memory addrs.
     pub fn server_rpc_addr(&self) -> Result<Option<GatewayServerAddr>> {
-        #![allow(clippy::unreachable)]
         self.rpc_client
             .gateway_addr
             .as_ref()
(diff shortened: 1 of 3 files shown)


dignifiedquire (approved):
missed that one 😅

iroh-gateway/src/config.rs
```diff
@@ -65,16 +65,18 @@ impl Config {
 
     /// Derive server addr for non memory addrs.
     pub fn server_rpc_addr(&self) -> Result<Option<GatewayServerAddr>> {
+        #![allow(clippy::unreachable)]
```
Arqu, line 68: this doesnt help :/

dignifiedquire: `#!` means apply to whole module, `#` means only next expression

dignifiedquire: the warning was for `unreachable_patterns` which is a compiler lint not a clippy lint ( so no `clippy:` prefix)

Arqu: 🙇

PR #175, opened 2022-08-02
feat: ctl ci

Arqu:


dignifiedquire:
what does this do?

Arqu:
pushes the `iroh-ctl` binary to s3 and allows us to further disseminate it. Intended to be shipped as part of the "quick setup" script.

dignifiedquire (approved):

PR #185, opened 2022-08-10
Remove resolver dependency from p2p

fabricedesre:
The dependency is causing issues when working on a 'all in one' daemon using memory channels
for rpc: we end up with a circular dependency: resolver -> p2p -> resolver.
Since the only use is the 'verify_hash()' function, this patch moves it to the
iroh_util crate and updates the call sites accordingly.


dignifiedquire (approved):
thanks

Issue #149, opened 2022-07-22
metrics: ability to compile without metrics


ramfox:

mxinden:
Would you mind expanding on the rational here @ramfox?

dignifiedquire:
this is mostly for mobile deployments, where we want the option to reduce build size and any form of tracking as much as possible

Arqu:
It's not a 100% match (need to do some voodoo if you want to compile without it) but I call it done for now and will revisit in the future if needed.

PR #193, opened 2022-08-16
fix: allow builds without metrics

Arqu:
Closes https://github.com/n0-computer/iroh/issues/149

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -309,6 +309,8 @@ jobs:
       - set-env-path
       - *restore-workspace
       - *restore-cache
+      - setup-protoc:
+          arch: "x86_64"
       - setup-sccache
       - restore-sccache-cache
       - run:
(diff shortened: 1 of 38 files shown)


dignifiedquire:
hmm, this is a lot of code 😥

Arqu:
> hmm, this is a lot of code 😥

It is, got any better ideas?

My alternative is to make a null metrics implementation and switch on that with features. Might or might not be better but should be more contained.

dignifiedquire:
> My alternative is to make a null metrics implementation and switch on that with features.

I think that might be nicer, I am worried about things like the resolver code having `cfg`s all over the place.

Arqu:
> > My alternative is to make a null metrics implementation and switch on that with features.
> 
> I think that might be nicer, I am worried about things like the resolver code having `cfg`s all over the place.

Yeah it's not pretty. And a bit tedious to annotate every call to every metrics call ever. I'll do another branch for it and see where that goes.

PR #198, opened 2022-08-19
feat(iroh-resolver): create a balanced tree dag

ramfox:
When adding content to iroh, create a balanced tree.

Our balanced tree implementation takes a stream of
chunks (`BytesMut`) and returns a stream of Bytes
of the encoded nodes of the dag, and their associated Cids.
`(Cid, Bytes)`

The implementation emits the node as soon as it has
been created.

All the leaves of the tree are encoded as unixfs `Raw`
nodes. All the stems & the root of the tree are encoded
as unixfs `File` nodes.

We've put effort into having our nodes be as similar to
kubo as possible, but we have intentionally differed in
a few ways:
- all of our "stem" nodes have `name: None`. In kubo, the
parents of leaf nodes have the go equivalent of
`name: Some("".to_string())`.
- the `tsize` of our links is the cumulative size of
the encoded tree associated with that link. The definition
of `tsize` seems to be unfixed, and can mean different things
depending on the implementation.

### DIFF
--- a/iroh-resolver/src/balanced_tree.rs
+++ b/iroh-resolver/src/balanced_tree.rs
@@ -1,0 +1,455 @@
+use std::collections::VecDeque;
+
+use anyhow::Result;
+use async_stream::try_stream;
+use bytes::{Bytes, BytesMut};
+use cid::Cid;
+use futures::{Stream, StreamExt};
+
+use crate::unixfs::{dag_pb, unixfs_pb, DataType, Node, UnixfsNode};
+use crate::unixfs_builder::encode_unixfs_pb;
+
+/// Default degree number for balanced tree, taken from unixfs specs
+/// https://github.com/ipfs/specs/blob/main/UNIXFS.md#layout
+pub const DEFAULT_DEGREE: usize = 174;
+
+pub enum TreeBuilder {
+    /// TreeBuilder that builds a "balanced tree" with a max degree size of
+    /// degree
+    Balanced { degree: usize },
+}
+
+impl TreeBuilder {
+    pub fn balanced_tree() -> Self {
+        Self::balanced_tree_with_degree(DEFAULT_DEGREE)
+    }
+
+    pub fn balanced_tree_with_degree(degree: usize) -> Self {
+        TreeBuilder::Balanced { degree }
+    }
+
+    pub fn stream_tree(
+        &self,
+        chunks: impl Stream<Item = std::io::Result<BytesMut>>,
+    ) -> impl Stream<Item = Result<(Cid, Bytes)>> {
+        match self {
+            TreeBuilder::Balanced { degree } => stream_balanced_tree(chunks, *degree),
+        }
+    }
+}
+
+fn stream_balanced_tree(
+    in_stream: impl Stream<Item = std::io::Result<BytesMut>>,
+    degree: usize,
+) -> impl Stream<Item = Result<(Cid, Bytes)>> {
+    try_stream! {
+        // degree = 8
+        // VecDeque![ vec![] ]
+        // ..
+        // VecDeque![ vec![0, 1, 2, 3, 4, 5, 6, 7] ]
+        // VecDeque![ vec![8], vec![p0] ]
+
+        // ..
+
+        // VecDeque![ vec![0, 1, 2, 3, 4, 5, 6, 7] vec![p0] ]
+        // VecDeque![ vec![], vec![p0, p1]]
+
+        // ..
+
+        // VecDeque![ vec![0, 1, 2, 3, 4, 5, 6, 7] vec![p0, p1, p2, p3, p4, p5, p6, p7], ]
+        // VecDeque![ vec![], vec![p0, p1, p2, p3, p4, p5, p6, p7], vec![] ]
+        // VecDeque![ vec![8], vec![p8], vec![pp0] ]
+        //
+        // A vecdeque of vecs, the first vec representing the lowest layer of stem nodes
+        // and the last vec representing the root node
+        // Since we emit leaf and stem nodes as we go, we only need to keep track of the
+        // most "recent" branch, storing the links to that node's children & yielding them
+        // when each node reaches `degree` number of links
+        let mut tree: VecDeque<Vec<(Cid, u64)>> = VecDeque::new();
+        tree.push_back(Vec::with_capacity(degree));
+
+        tokio::pin!(in_stream);
+
+        while let Some(chunk) = in_stream.next().await {
+            let chunk = chunk?;
+            let chunk = chunk.freeze();
+            let tree_len = tree.len();
+
+            // check if the leaf node of the tree is full
+            if tree[0].len() == degree {
+                // if so, iterate through nodes
+                for i in 0..tree_len {
+                    // if we encounter any nodes that are not full, break
+                    if tree[i].len() < degree {
+                        break;
+                    }
+
+                    // in this case we have a full set of links & we are
+                    // at the top of the tree. Time to make a new layer.
+                    if i == tree_len - 1 {
+                        tree.push_back(Vec::with_capacity(degree));
+                    }
+
+                    // create node, keeping the cid
+                    let links = std::mem::replace(&mut tree[i], Vec::with_capacity(degree));
+                    let (cid, bytes, len) = TreeNode::Stem(links).encode()?;
+                    yield (cid, bytes);
+
+                    // add cid to parent node
+                    tree[i+1].push((cid, len));
+                }
+                // at this point the tree will be able to recieve new links
+                // without "overflowing", aka the leaf node and stem nodes
+                // have fewer than `degree` number of links
+            }
+
+            // now that we know the tree is in a "healthy" state to
+            // recieve more links, add the link to the tree
+            let (cid, bytes, len) = TreeNode::Leaf(chunk).encode()?;
+            tree[0].push((cid, len));
+            yield (cid, bytes);
+            // at this point, the leaf node may have `degree` number of
+            // links, but no other stem node will
+        }
+
+        // our stream had 1 chunk that we have already yielded
+        if tree.len() == 1 && tree[0].len() == 1 {
+            return
+        }
+
+        // clean up, aka yield the rest of the stem nodes
+        // since all the stem nodes are able to recieve links
+        // we don't have to worry about "overflow"
+        while let Some(links) = tree.pop_front() {
+            let (cid, bytes, len) = TreeNode::Stem(links).encode()?;
+            yield (cid, bytes);
+
+            if let Some(front) = tree.front_mut() {
+                front.push((cid, len));
+            } else {
+                // final root, nothing to do
+            }
+        }
+    }
+}
+
+fn create_unixfs_node_from_links(links: Vec<(Cid, u64)>) -> Result<UnixfsNode> {
+    let links = links
+        .into_iter()
+        .map(|(cid, len)| dag_pb::PbLink {
+            hash: Some(cid.to_bytes()),
+            /// ALL "stem" nodes have `name: None`.
+            /// In kubo, nodes that have links to `leaf` nodes have `name: Some("".to_string())`
+            name: None,
+            /// tsize has no strict definition
+            /// Iroh's definiton of `tsize` is "the cumulative size of the encoded tree
+            /// pointed to by this link", so not just the size of the raw content, but including
+            /// all encoded dag nodes as well.
+            /// In the `go-merkledag` package, the `merkledag.proto` file, states that tsize
+            /// is the "cumulative size of the target object"
+            /// (https://github.com/ipfs/go-merkledag/blob/8335efd4765ed5a512baa7e522c3552d067cf966/pb/merkledag.proto#L29)
+            tsize: Some(len as u64),
+        })
+        .collect();
+
+    // PBNode.Data
+    let inner = unixfs_pb::Data {
+        r#type: DataType::File as i32,
+        ..Default::default()
+    };
+
+    // create PBNode
+    let outer = encode_unixfs_pb(&inner, links)?;
+
+    // create UnixfsNode
+    Ok(UnixfsNode::File(Node { inner, outer }))
+}
+
+// Leaf and Stem nodes are the two types of nodes that can exist in the tree
+// Leaf nodes encode to `UnixfsNode::Raw`
+// Stem nodes encode to `UnixfsNode::File`
+enum TreeNode {
+    Leaf(Bytes),
+    Stem(Vec<(Cid, u64)>),
+}
+
+impl TreeNode {
+    fn encode(self) -> Result<(Cid, Bytes, u64)> {
+        match self {
+            TreeNode::Leaf(bytes) => {
+                let node = UnixfsNode::Raw(bytes);
+                let (cid, bytes) = node.encode()?;
+                let len = bytes.len();
+                Ok((cid, bytes, len as u64))
+            }
+            TreeNode::Stem(links) => {
+                let mut cumulative_len: u64 = links.iter().map(|(_, len)| len).sum();
+                let node = create_unixfs_node_from_links(links)?;
+                let (cid, bytes) = node.encode()?;
+                cumulative_len += bytes.len() as u64;
+                Ok((cid, bytes, cumulative_len))
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use futures::StreamExt;
+    fn test_chunk_stream(num_chunks: usize) -> impl Stream<Item = std::io::Result<BytesMut>> {
+        futures::stream::iter((0..num_chunks).map(|n| Ok(BytesMut::from(&n.to_be_bytes()[..]))))
+    }
+
+    async fn build_expect_tree(num_chunks: usize, degree: usize) -> Vec<Vec<(Cid, Bytes)>> {
+        let chunks = test_chunk_stream(num_chunks);
+        tokio::pin!(chunks);
+        let mut tree = vec![vec![]];
+        let mut links = vec![vec![]];
+
+        if num_chunks / degree == 0 {
+            let chunk = chunks.next().await.unwrap().unwrap();
+            let leaf = TreeNode::Leaf(chunk.freeze());
+            let (cid, bytes, _) = leaf.encode().unwrap();
+            tree[0].push((cid, bytes));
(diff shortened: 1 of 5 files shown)


iroh-resolver/src/balanced_tree.rs
```diff
@@ -0,0 +1,380 @@
...
+    }
+
+    // create UnixfsNode
+    Ok(UnixfsNode::File(Node { inner, outer }))
+}
+
+// Leaf and Stem nodes are the two types of nodes that can exist in the tree
+// Leaf nodes encode to `UnixfsNode::Raw`
+// Stem nodes encode to `UnixfsNode::File`
+// Root nodes encode to `UnixfsNode::Directory`
```
ramfox, line 149: Okay, so in the original `file.encode()` the root node was a `UnixfsNode::Directory`. But we always talked about `UnixfsNode::File`s. Wanted to double check, what is the correct format for the root node?

I know that to preserve the file name we need to wrap the file in a directory, but that happens in `add_file`.

iroh-resolver/src/balanced_tree.rs
```diff
@@ -0,0 +1,380 @@
...
+}
+
+fn create_unixfs_node_from_links(links: Vec<(Cid, usize)>, is_root: bool) -> Result<UnixfsNode> {
+    let links = links
+        .into_iter()
+        .map(|(cid, len)| dag_pb::PbLink {
+            hash: Some(cid.to_bytes()),
+            // TODO: notes say that leaf chunks have "", does that mean that
+            // stem nodes should have None?
+            name: Some("".into()),
```
ramfox, line 124: the `unixfs` notes in question: https://www.notion.so/number-zero/UnixFs-742339892d9c47d5b79f4f942e661bbf#f74931c4d8b046c3807377cbe2b443f1

should I have `name` as `Some("".into())`, or `None`, or something else?

iroh-resolver/src/balanced_tree.rs
```diff
@@ -0,0 +1,380 @@
...
+    // Root(Vec<(Cid, usize)>),
+}
+
+impl TreeNode {
+    fn encode(self) -> Result<(Cid, Bytes)> {
+        match self {
+            TreeNode::Leaf(bytes) => {
+                // TODO: notes say that balanced in go-ipfs marks leaves as `File`, not `Raw`, should
+                // I change this?
+                let node = UnixfsNode::Raw(bytes);
```
ramfox, line 162: the `merkletree` section of the unixfs notes (https://www.notion.so/number-zero/UnixFs-742339892d9c47d5b79f4f942e661bbf#a2bcc0d96aea4688b27f65eee44a03aa) says that the balanced tree has leafs as `UnixfsNode::File`. Just wanted to double confirm that we are on purpose using `UnixfsNode::Raw`

ramfox: wait just answered my own question. unixfs spec says leaves should be marked as raw (https://github.com/ipfs/specs/blob/main/UNIXFS.md#chunking)

iroh-resolver/src/balanced_tree.rs
```diff
@@ -0,0 +1,388 @@
...
+
+// Leaf and Stem nodes are the two types of nodes that can exist in the tree
+// Leaf nodes encode to `UnixfsNode::Raw`
+// Stem nodes encode to `UnixfsNode::File`
+// Root nodes encode to `UnixfsNode::Directory`
+enum TreeNode {
+    Leaf(Bytes),
+    Stem(Vec<(Cid, u64)>),
+    // Root(Vec<(Cid, u64)>),
+}
```
ramfox, line 151: Should the root node of the tree be a `UnixfsNode::Directory`, or `UnixfsNode::File`?

The previous `file.encode()` had `UnixfsNode::Directory`, but we've only ever discussed `UnixfsNode::File`, so wanted to double-check. Will remove `TreeNode::Root` code if we don't  use `UnixfsNode::Directory`

ramfox:
tests failing because: "We cannot run this job using the selected resource class. Please check your configuration and try again."?

iroh-resolver/src/balanced_tree.rs
dignifiedquire: yes, intermediary nodes should have no name, aka `None`.

iroh-resolver/src/balanced_tree.rs
dignifiedquire: both are used in kubo, we are unifying to use Raw encoded leaves

iroh-resolver/src/balanced_tree.rs
dignifiedquire: trees are only used to encode files, so the root should always be a file.

iroh-resolver/src/balanced_tree.rs
```diff
@@ -0,0 +1,388 @@
...
+use futures::{Stream, StreamExt};
+
+use crate::unixfs::{dag_pb, unixfs_pb, DataType, Node, UnixfsNode};
+use crate::unixfs_builder::encode_unixfs_pb;
+
+// 3                          root
+// 2                                 nextparent0
+// 1             parent0              parent1    ... parent7
+// 0 [0, 1, 2, 3, 4, 5, 6, 7]      [0, ... ]
+//
```
dignifiedquire, line 16: we can probably remove our ascii art at some point 😅

iroh-resolver/src/balanced_tree.rs
```diff
@@ -0,0 +1,388 @@
...
+        // vec![ vec![0, 1, 2, 3, 4, 5, 6, 7] vec![p0] ]
+        // vec![ vec![], vec![p0, p1]]
+
+        // ..
+
+        // vec![ vec![0, 1, 2, 3, 4, 5, 6, 7] vec![p0, p1, p2, p3, p4, p5, p6, p7], ]
+        // vec![ vec![], vec![p0, p1, p2, p3, p4, p5, p6, p7], vec![] ]
+        // vec![ vec![8], vec![p8], vec![pp0] ]
+
+        let mut tree: VecDeque<Vec<(Cid, u64)>> = VecDeque::new();
```
dignifiedquire, line 39: Can you add a comment on what exactly we store in here? (mostly for ourselves in the future)

iroh-resolver/src/unixfs_builder.rs
```diff
@@ -286,21 +258,15 @@ impl FileBuilder {
...
-
         Ok(File {
             name,
-            nodes: Box::pin(nodes),
+            nodes: Box::pin(self.chunker.chunks(reader)),
         })
     }
 }
 
+// fn build_tree(
```
dignifiedquire, line 268: stray comment

iroh-resolver/src/balanced_tree.rs
```diff
@@ -0,0 +1,388 @@
...
+mod tests {
+    use super::*;
+
+    fn test_chunk_stream(num_chunks: usize) -> impl Stream<Item = std::io::Result<BytesMut>> {
+        async_stream::try_stream! {
+            for n in 0..num_chunks {
+                let bytes = BytesMut::from(&n.to_be_bytes()[..]);
+                yield bytes
+            }
+        }
```
dignifiedquire, line 190: there is an easier way to create sth like this

```rust
futures::stream::iter(
  (0..n).map(|n| BytesMut::from(&n.to_be_bytes()[..])
)
```

iroh-resolver/src/balanced_tree.rs
```diff
@@ -0,0 +1,388 @@
...
+
+        let got = stream_balanced_tree(test_chunk_stream(num_chunks), degrees);
+        tokio::pin!(got);
+        ensure_equal(expect, got).await;
+    }
+
+    #[tokio::test]
+    async fn balanced_tree_test_height_two_not_full() {
+        let num_chunks = 7;
+        let degrees = 3;
```
dignifiedquire, line 300: I would suggest to make the tests such that we can run over multiple different degrees, and num chunks in some generic way, so we can increase the coverage here

iroh-resolver/src/balanced_tree.rs
```diff
@@ -0,0 +1,388 @@
...
+
+fn create_unixfs_node_from_links(links: Vec<(Cid, u64)>, is_root: bool) -> Result<UnixfsNode> {
+    let links = links
+        .into_iter()
+        .map(|(cid, len)| dag_pb::PbLink {
+            hash: Some(cid.to_bytes()),
+            // TODO: notes say that leaf chunks have "", does that mean that
+            // stem nodes should have None?
+            name: Some("".into()),
+            tsize: Some(len as u64),
```
dignifiedquire, line 122: we should also check for the correct `tsize` in tests

iroh-resolver/src/balanced_tree.rs
```diff
@@ -0,0 +1,388 @@
...
+// 3                          root
+// 2                                 nextparent0
+// 1             parent0              parent1    ... parent7
+// 0 [0, 1, 2, 3, 4, 5, 6, 7]      [0, ... ]
+//
+
+pub fn stream_balanced_tree(
+    in_stream: impl Stream<Item = std::io::Result<BytesMut>>,
+    max_degrees: usize,
+) -> impl Stream<Item = Result<(Cid, Bytes)>> {
```
dignifiedquire, line 21: given we get `BytesMut` we should just pass those through, instead of freezing them

iroh-resolver/src/balanced_tree.rs
```diff
@@ -0,0 +1,388 @@
...
+        tokio::pin!(in_stream);
+
+        while let Some(chunk) = in_stream.next().await {
+            let chunk = chunk?;
+            let chunk = chunk.freeze();
+            let tree_len = tree.len();
+
+            // check if the leaf node of the tree is full
+            if tree[0].len() == max_degrees {
+                // if so, iterater through nodes
```
dignifiedquire, line 51: typo

Arqu:
> tests failing because: "We cannot run this job using the selected resource class. Please check your configuration and try again."?

Ignore mac tests for now, I have to figure it out seems like some permission issue with the runners.

iroh-resolver/src/balanced_tree.rs
ramfox: Hmmm, but we would really only "pass through" the `UnxifsNode::Raw` chunks as `BytesMut`, the `UnixfsNode::File` nodes would get encoded as `Bytes`. We could get the encoded `UnixfsNode::File` as `Bytes` and turn them into `BytesMut` to match the raw nodes?

Thoughts?

dignifiedquire: I guess leave as is for now, can optimize later

iroh-resolver/src/balanced_tree.rs
ramfox: I'm conceding to moving some of it, lol
Some I think will be helpful for anyone reading it over in the future

iroh-resolver/src/balanced_tree.rs
```diff
@@ -0,0 +1,451 @@
...
+use crate::unixfs::{dag_pb, unixfs_pb, DataType, Node, UnixfsNode};
+use crate::unixfs_builder::encode_unixfs_pb;
+
+/// Default degree number for balanced tree, taken from unixfs specs
+/// https://github.com/ipfs/specs/blob/main/UNIXFS.md#layout
+pub const DEFAULT_DEGREE: usize = 174;
+
+pub enum TreeBuilder {
+    // TreeBuilder that builds a "balanced tree" with a max degree size of
+    // degree
```
dignifiedquire, line 18: three slashes please

iroh-resolver/src/balanced_tree.rs
```diff
@@ -0,0 +1,451 @@
...
+        &self,
+        chunks: impl Stream<Item = std::io::Result<BytesMut>>,
+    ) -> impl Stream<Item = Result<(Cid, Bytes)>> {
+        match self {
+            TreeBuilder::Balanced { degree } => stream_balanced_tree(chunks, *degree),
+        }
+    }
+}
+
+pub fn stream_balanced_tree(
```
dignifiedquire, line 41: does this need to be pub?

iroh-resolver/src/balanced_tree.rs
```diff
@@ -0,0 +1,451 @@
...
+                for i in 0..tree_len {
+                    // if we encounter any nodes that are not full, break
+                    if tree[i].len() < degree {
+                        break;
+                    }
+
+                    // in this case we have a full set of links & we are
+                    // at the top of the tree. Time to make a new layer.
+                    if i == tree_len - 1 {
+                        tree.push_back(vec![]);
```
dignifiedquire, line 90: `with_capacity`?

iroh-resolver/src/balanced_tree.rs
```diff
@@ -0,0 +1,451 @@
...
+                    }
+
+                    // create node, keeping the cid
+                    let links = std::mem::replace(&mut tree[i], Vec::new());
```
dignifiedquire, line 94: `with_capacity`?

iroh-resolver/src/balanced_tree.rs
```diff
@@ -0,0 +1,451 @@
...
+    let inner = unixfs_pb::Data {
+        r#type: DataType::File as i32,
+        ..Default::default()
+    };
+
+    // create PBNode
+    let outer = encode_unixfs_pb(&inner, links)?;
+
+    if is_root {
+        return Ok(UnixfsNode::Directory(Node { inner, outer }));
```
dignifiedquire, line 156: If this is a directory, the links should have names, but I don't think that is the right type for the root here, unless I am missing sth

iroh-resolver/src/balanced_tree.rs
```diff
@@ -0,0 +1,451 @@
...
+                let (cid, bytes) = node.encode()?;
+                let len = bytes.len();
+                Ok((cid, bytes, len as u64))
+            }
+            TreeNode::Stem(links) => {
+                // keep track of `tsize`, aka the size of the encoded tree at the given link
+                let mut cumulative_len: u64 = links.iter().map(|(_, len)| len).sum();
+                let node = create_unixfs_node_from_links(links, false)?;
+                let (cid, bytes) = node.encode()?;
+                cumulative_len += bytes.len() as u64;
```
dignifiedquire, line 185: I am not sure if this length calculation is correct, @ribasushi thoughts?

iroh-resolver/src/balanced_tree.rs
dignifiedquire: I don't see any test with other degrees than 3?

iroh-resolver/src/unixfs_builder.rs
```diff
@@ -198,44 +197,14 @@ impl File {
...
-                    ..Default::default()
-                };
-                let outer = encode_unixfs_pb(&inner, links)?;
-                let node = UnixfsNode::Directory(Node { outer, inner });
-                let bytes = node.encode()?;
-                let cid = Cid::new_v1(Codec::DagPb as _, cid::multihash::Code::Sha2_256.digest(&bytes));
-                yield (cid, bytes)
-            }
-        }
+        TreeBuilder::balanced_tree().stream_tree(self.nodes)
```
dignifiedquire, line 200: lets store a `tree_builder`, which is initialized to balanced_tree by default on the builder and use that here

iroh-resolver/src/unixfs_builder.rs
```diff
@@ -259,6 +228,7 @@ impl Default for FileBuilder {
     }
 }
 
+// FileBuilder separates uses a reader or bytes to chunk the data into raw unixfs nodes
```
dignifiedquire, line 231: `///`

iroh-resolver/src/balanced_tree.rs
ramfox: Oof, this is dead code that should have been removed. Thank you for catching.

iroh-resolver/src/balanced_tree.rs
ramfox: yes please a million more eyes on this. 

https://github.com/ipfs/kubo/blob/b3faaad1310bcc32dc3dd24e1919e9edf51edba8/merkledag/pb/merkledag.proto#L27-L28

took this to mean: cumulative size of the tree that this link points to. Which would include the size of the encoded node that link points to + the tsize of all of that node's links.

ribasushi: @ramfox @dignifiedquire this value is effectively "freeform". You can omit it, or put whatever you feel like in there ( full DAG weight, deduplicated DAG weight, the payload-sum-size, whatever ). Aside from rendering "directory sizes" on gateways this value has no other function.

ramfox: okay @dignifiedquire what's iroh's opinion on what this value should be?

dignifiedquire: I think yours is fine for now, we might change our opinion later

iroh-resolver/src/balanced_tree.rs
dignifiedquire: @ramfox ping

ramfox: yea, thank you for re-pinging I somehow missed this during both passes...

iroh-resolver/src/balanced_tree.rs
```diff
@@ -0,0 +1,447 @@
...
+        }
+    }
+}
+
+fn create_unixfs_node_from_links(links: Vec<(Cid, u64)>) -> Result<UnixfsNode> {
+    let links = links
+        .into_iter()
+        .map(|(cid, len)| dag_pb::PbLink {
+            hash: Some(cid.to_bytes()),
+            name: None,
```
dignifiedquire, line 141: leafs should have some("") as name for compatibility

ramfox: This is only ever used to create "stem" nodes, the only "leaf" nodes we have here are `Raw`.

Let me know if I'm misunderstanding something.

dignifiedquire (dismissed):

iroh-resolver/src/balanced_tree.rs
ramfox: to wrap this discussion, we've decided to go with `name: None` in all stem nodes. We are choosing to stray from what kubo does here (which is to have `Some("".to_string())` for links to all leaf nodes).

Arqu (approved):
A-OK!

PR #203, opened 2022-08-19
feat: Make gossipsub configurable

fabricedesre:
I'm not sure what the consequences are to disable it, so maybe we don't want that at all? Eg. would that prevent 2 peers to discover each other and run bitswap with just mdns enabled?

cc @dignifiedquire


dignifiedquire (approved):
thanks

dignifiedquire:
> Eg. would that prevent 2 peers to discover each other and run bitswap with just mdns enabled?

no it would not, bitswap, dht and other discovery methods are independent from gossipsub

PR #204, opened 2022-08-22
feat(p2p): minimal bitswap providers

dignifiedquire:


ramfox:
This is work done on our previous bitswap implementation. Closing the PR, but leaving the branch up in case it becomes relevant again. We can always open the PR back up.

PR #205, opened 2022-08-23
Readme kaizen

mishmosh:
This PR includes some minor readme improvements:
- Grammar fixes
- More details for iroh cloud and mobile, pulled from prior literature


b5 (approved):
Looks good! Thanks so much for adding a little detail 😄

I pushed a commit that softens the "use at own risk" language (which is covered by the license) to instead focus on the fact that we haven't cut a release. Coupling this with a concrete time frame avoids the perpetual "it's your fault if this doesn't work" standard.

PR #209, opened 2022-08-27
refactor: global metrics

Arqu:


Cargo.toml
```diff
@@ -18,5 +18,5 @@ members = [
 ]
 
 [patch.crates-io]
-libp2p = { git = "https://github.com/dignifiedquire/rust-libp2p", branch = "feat-kad-count" }
+libp2p = { git = "https://github.com/arqu/rust-libp2p", branch = "feat-metrics-clone" }
```
Arqu, line 21: Did this because my branch (based of `feat-kad-count`) allows the libp2p metrics struct to be cloneable.

iroh-metrics/src/core.rs
```diff
@@ -0,0 +1,114 @@
...
+use crate::gateway;
+#[cfg(feature = "p2p")]
+use crate::p2p;
+#[cfg(feature = "resolver")]
+use crate::resolver;
+#[cfg(feature = "store")]
+use crate::store;
+
+lazy_static! {
+    pub(crate) static ref CORE: RwLock<Core> = RwLock::new(Core::default());
```
Arqu, line 16: Ok, let me explain this one. To make ergonomics simple I needed a global metrics instance with each of the relevant metrics registered on the prometheus registry just once. While not a favorite to have a singleton in rust, makes things work.

The reason this is actually feasible and shareable across threads is that the underlying structures are just references to the prometheus client counters/metrics which are all based on atomic units and data races don't really matter here. Only exception are the histograms but they are currently behind a lock (and might eventually move to a lock free design).

iroh-metrics/src/core.rs
```diff
@@ -0,0 +1,114 @@
...
+}
+
+pub(crate) struct Core {
+    enabled: bool,
```
Arqu, line 20: I was unsure about the best way to approach the enabled/disabled thing especially about doing it at compile time.

iroh-metrics/src/core.rs
```diff
@@ -0,0 +1,114 @@
...
+    pub(crate) fn set_enabled(&mut self, enabled: bool) {
+        self.enabled = enabled;
+    }
+
+    pub(crate) fn enabled(&self) -> bool {
+        self.enabled
+    }
+}
+
+pub trait MetricType {
```
Arqu, line 99: These traits are simply to differentiate calls and leverage compile time checks to avoid basic mistakes.

iroh-metrics/src/lib.rs
```diff
@@ -141,3 +153,97 @@ pub fn get_current_trace_id() -> TraceId {
...
+    let mut r = CORE.write().await;
+    r.set_enabled(true);
+}
+
+async fn disable() {
+    let mut r = CORE.write().await;
+    r.set_enabled(false);
+}
+
+pub async fn p2p_metrics() -> p2p::Metrics {
```
Arqu, line 246: p2p metrics are sourced from libp2p so without re-doing the whole thing (and assuming maintenance of it) we just use a pass through type. It also changes the metrics dynamics for p2p as libp2p works on an event based platform.

iroh-metrics/src/lib.rs
```diff
@@ -141,3 +153,97 @@ pub fn get_current_trace_id() -> TraceId {
...
+pub fn observe_sync<M>(c: Collector, m: M, v: f64)
+where
+    M: 'static + HistogramType + std::fmt::Display + Send,
+{
+    tokio::spawn(async move {
+        observe(c, m, v).await;
+    });
+}
+
+async fn print() {
```
Arqu, line 231: This is just debug stuff, will remove before merge.

iroh-metrics/src/lib.rs
```diff
@@ -93,12 +96,21 @@ pub async fn init_metrics(
                 }
             }
         }));
+    } else {
+        //TODO this just debug stuff
```
Arqu, line 100: Will drop before merge.

iroh-metrics/src/lib.rs
```diff
@@ -141,3 +153,97 @@ pub fn get_current_trace_id() -> TraceId {
...
+            Collector::Bitswap => r.bitswap_metrics().record(m, v),
+            #[cfg(feature = "store")]
+            Collector::Store => r.store_metrics().record(m, v),
+            _ => panic!("not enabled/implemented"),
+        };
+    }
+}
+
+#[allow(unused_variables, unreachable_patterns)]
+pub fn record_sync<M>(c: Collector, m: M, v: u64)
```
Arqu, line 191: Some of the code using metrics is not async so I made this as something to discuss, but if it were to live on, I'd rather have the task spawning in the call rather than having the user have to spawn tasks every time to shoot off a metric. Easier to refactor later if we try other runtimes.

iroh-metrics/src/lib.rs
```diff
@@ -141,3 +153,97 @@ pub fn get_current_trace_id() -> TraceId {
...
+where
+    M: 'static + MetricType + std::fmt::Display + Send,
+{
+    tokio::spawn(async move {
+        record(c, m, v).await;
+    });
+}
+
+#[allow(unused_variables, unreachable_patterns)]
+pub async fn observe<M>(c: Collector, m: M, v: f64)
```
Arqu, line 201: Also unsure if it makes sense to make this into macros?

iroh-metrics/src/bitswap.rs
```diff
@@ -65,6 +68,60 @@ impl Metrics {
     }
 }
 
+impl MetricsRecorder for Metrics {
```
dignifiedquire, line 71: what if the api was instead

```rust
BitswapMetrics::Requests.record(value);
```

this way you could avoid the `name` matching

iroh-metrics/src/core.rs
dignifiedquire: this seems like an interesting approach: https://docs.rs/libp2p/latest/libp2p/swarm/behaviour/toggle/struct.Toggle.html

iroh-metrics/src/bitswap.rs
dignifiedquire: alternatively, we could make the api more like this


```rust
// this is similar to how log! works

record!(Bitswap::Requests, 15);
record!(Store::GetRequests, 4.3); // with macros we can use different types here


// could have shortcuts generated if we want to
bitswap_requests!(15);
store_get_requests!(4.3);
```

iroh-metrics/src/core.rs
dignifiedquire: You can avoid the rwlock. If you change the `enabled` to be an `AtomicBool` you always only need an immutable reference to `Core`.

Arqu: LOVELY TY!

iroh-metrics/src/bitswap.rs
Arqu: K, think I got it in a solid place now and am happy with it.

iroh-metrics/src/bitswap.rs
```diff
@@ -65,6 +71,67 @@ impl Metrics {
...
+pub enum BitswapMetrics {
+    Requests,
+    Cancels,
+    BlockBytesOut,
+    BlockBytesIn,
+    Providers,
+}
+
+impl MetricType for BitswapMetrics {
+    fn name(&self) -> &str {
```
dignifiedquire, line 112: this should be `-> &'static str` to make sure this can be optimised well

dignifiedquire:
api is looking great

Cargo.toml
dignifiedquire: @Arqu can we change this/merge this? I would love to not have more branches. Alternatively we could just move to a specific fork on n0-computer maybe that we use

Arqu: yeah lets merge it

Arqu: for whatever reason, I cant use your branch as a base (cant find it) for a PR. Can you just locally smush it in with a cherry pick

Arqu: No longer needed

dignifiedquire (dismissed):

dignifiedquire (approved):

PR #212, opened 2022-08-29
feat: iroh-one

Arqu:
Based off of the awesome work @fabricedesre did on https://github.com/n0-computer/iroh/pull/188

Took it a little bit further in terms of making it mergeable for us and reducing the maintenance burden of the crate.

@fabricedesre Hope this is a nice middle ground that works for all.
- uds support remains, it's feature enabled now at compile time
- I've dropped the racing http/p2p work and we can do that as a separate piece with a custom `ContentLoader` as @dignifiedquire suggested
- made the cleanups I mentioned over there.
- Dropped the android hard coded bits (mainly the uds socket definition) however you can pass it in as an argument ~~(actually need to finish this - done)~~

LMK what y'all think


iroh-gateway/src/config.rs
```diff
@@ -167,6 +167,20 @@ impl Source for Config {
     }
 }
 
+impl crate::handlers::StateConfig for Config {
+    fn rpc_client(&self) -> iroh_rpc_client::Config {
```
dignifiedquire, line 171: it would be cleaner to return a reference and only `clone` on the caller side if needed

iroh-gateway/src/config.rs
```diff
@@ -167,6 +167,20 @@ impl Source for Config {
...
+        self.rpc_client.clone()
+    }
+
+    fn port(&self) -> u16 {
+        self.port
+    }
+
+    fn user_headers(&self) -> HeaderMap<HeaderValue> {
```
dignifiedquire, line 179: it would be cleaner to return a reference and only `clone` on the caller side if needed

iroh-one/README.md
```diff
@@ -0,0 +1,28 @@
+# Iroh One Gateway
+
+A rust implementation of an IPFS gateway.
```
dignifiedquire, line 3: this is not just a gateway

iroh-one/README.md
```diff
@@ -0,0 +1,28 @@
...
+- Run with `cargo run --release -- -h` for details
+- `-wcf` Writeable, Cache, Fetch (options to toggle write enable, caching mechanics and fetching from the network); currently exists but is not implemented
+- `-p` Port the gateway should listen on
+
+## ENV Variables
+
+- `IROH_INSTANCE_ID` - unique instance identifier, preferably some name than hard id (default: generated lower & snake case name)
+- `IROH_ENV` - indicates the service environment (default: `dev`)
+
+## Endpoints
```
dignifiedquire, line 20: can we point to the iroh-gateway readme for these? otherwise they will get out of sync soon

iroh-one/src/cli.rs
```diff
@@ -0,0 +1,61 @@
+/// CLI arguments support.
+use clap::Parser;
+use std::collections::HashMap;
+use std::path::PathBuf;
+
+#[derive(Parser, Debug, Clone)]
+#[clap(author, version, about, long_about = None)]
+pub struct Args {
+    //Gateway
```
dignifiedquire, line 9: should be `/// Gateway` if this is meant to be kept around

iroh-one/src/cli.rs
```diff
@@ -0,0 +1,61 @@
...
+    #[clap(short, long)]
+    port: Option<u16>,
```
dignifiedquire, line 11: maybe `gateway_port`?  there will be more ports on this sooner rather than later I expect

iroh-one/src/config.rs
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    // When running in ipfsd mode, the resolver will use memory channels to
+    // communicate with the p2p and store modules.
+    // The gateway itself is exposing a UDS rpc endpoint to be also usable
+    // as a single entry point for other system services if feature enabled.
+    pub fn default_ipfsd() -> RpcClientConfig {
+        #[cfg(feature = "uds-gateway")]
+        let path: PathBuf =
+            format!("{}", std::env::temp_dir().join("ipfsd.gateway").display()).into();
```
dignifiedquire, line 68: please use `tempdir`, it is more reliable

dignifiedquire: also, no need for the `format` call,

iroh-one/src/config.rs
```diff
@@ -0,0 +1,162 @@
...
+            store,
+            p2p,
+            rpc_client,
+            metrics: MetricsConfig::default(),
+            #[cfg(feature = "uds-gateway")]
+            uds_path,
+        }
+    }
+
+    // When running in ipfsd mode, the resolver will use memory channels to
```
dignifiedquire, line 61: doc comments should always be `///` (many instances in this pr ;) )

iroh-one/src/config.rs
```diff
@@ -0,0 +1,162 @@
...
+        self.p2p.rpc_client = self.rpc_client.clone();
+        self.store.rpc_client = self.rpc_client.clone();
+    }
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        #[cfg(feature = "uds-gateway")]
+        let uds_path: PathBuf =
+            format!("{}", std::env::temp_dir().join("ipfsd.gateway").display()).into();
```
dignifiedquire, line 92: same as above

iroh-one/src/config.rs
```diff
@@ -0,0 +1,162 @@
...
+            gateway_addr: Some(Addr::GrpcUds(path)),
+            #[cfg(not(feature = "uds-gateway"))]
+            gateway_addr: None,
+            p2p_addr: None,
+            store_addr: None,
+        }
+    }
+
+    // synchronize the rpc config across subsystems
+    pub fn sync_rpc_client(&mut self) {
```
dignifiedquire, line 81: what exactly is this used for?

iroh-one/src/rpc.rs
```diff
@@ -0,0 +1,25 @@
+use anyhow::Result;
+use async_trait::async_trait;
+use iroh_rpc_types::gateway::{Gateway as RpcGateway, GatewayServerAddr, VersionResponse};
+
+#[derive(Default)]
+pub struct Gateway {}
+
+#[async_trait]
+impl RpcGateway for Gateway {
```
dignifiedquire, line 9: Is this actually needed? given we also have a store and a p2p node wouldn't we proxy those too if we do this?

iroh-one/src/config.rs
fabricedesre: That won't do anything sensible on Android though. Maybe it's best to make that path part of the config.

Arqu: @fabricedesre just a note that this is the default and is overridden by `--gateway-uds-path` if set.

iroh-one/src/rpc.rs
Arqu: Don't think it's used, carry over from @fabricedesre work.

Arqu:
@fabricedesre just to clarify some of the stuff here. I don't think we can take on the `android` specific code you had, but I've aimed to give enough flexibility so that it's an easy setup. I assume that you don't have the luxury of cli arguments when starting things on android thus relying on defaults.

If that is the case I think the best way to support this is by having `iroh-one` being flexible and configurable enough that you can simply import it as a crate in your own rust project and simply copy over just the `main.rs` and slightly modify it to fit your needs. Think for this specific case, it would be enough to simply modify the config with the correct uds path before setting the rpc client.

fabricedesre:
@Arqu  I can actually use cli arguments in my case (it's not regular android either), but I dislike having a mix of configuration through cli args and in a config file. If I can use `--cfg` and have all the configuration in a config file that's ideal.

Arqu:
@fabricedesre confirming that `--cfg=path/to/your/config` works

Sample toml config
```
gateway_uds_path = 'tmp.sock'

[gateway]
writeable = false
fetch = false
cache = false
port = 9051
denylist = false

[gateway.headers]
access-control-allow-headers = 'content-type, content-disposition, last-modified, cache-control, accept-ranges, etag, service-worker, x-ipfs-gateway-prefix, x-trace-id, x-content-type-options, x-ipfs-path, x-ipfs-roots'
access-control-allow-methods = 'GET, PUT, POST, DELETE, HEAD, OPTIONS'
accept-ranges = 'none'
access-control-allow-origin = '*'
cache-control = 'no-cache, no-transform'

[store]
path = 'iroh-store-db'

[p2p.libp2p]
listening_multiaddr = '/ip4/0.0.0.0/tcp/4444'
bootstrap_peers = [
    '/dnsaddr/bootstrap.libp2p.io/p2p/QmNnooDu7bfjPFoTZYxMNLWUQJyrVwtbZg5gBMjTezGAJN',
    '/dnsaddr/bootstrap.libp2p.io/p2p/QmQCU2EcMqAqQPR2i9bChDtGNJchTbq5TbXJJ16u19uLTa',
    '/dnsaddr/bootstrap.libp2p.io/p2p/QmbLHAnMoJPWSCR5Zhtx6BHJX9KiKNN6tpvbUcqanj75Nb',
    '/dnsaddr/bootstrap.libp2p.io/p2p/QmcZf59bWwK5XFi76CZX8cbJ4BhTzzA3gU1ZjYZcYW3dwt',
    '/ip4/104.131.131.82/tcp/4001/p2p/QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ',
]
mdns = false
kademlia = true
autonat = true
relay_server = true
relay_client = true
gossipsub = true
max_conns_out = 256
max_conns_in = 256
max_conns_pending_out = 256
max_conns_pending_in = 256
max_conns_per_peer = 8
notify_handler_buffer_size = 256
connection_event_buffer_size = 256
dial_concurrency_factor = 16
```

Remember to build with `--feature=uds-gateway` to enable uds support.

If this is ok with you I'd like to move towards merging this?

fabricedesre (approved):
looks great, thanks!

PR #213, opened 2022-08-30
feat: recursive directories & directory link limits

ramfox:


iroh-resolver/src/unixfs_builder.rs
```diff
@@ -1,55 +1,172 @@
...
     unixfs::{dag_pb, unixfs_pb, DataType, Node, UnixfsNode},
 };
 
+// CID_LEN is the length of a cid
+const CID_LEN: usize = 64;
+
+// the PBNode.Data field may contain some data, this buffer is to ensure that
+// the contents of that data field + the size of the encoded protobuf links
+// will not be larger than the chunk size
+const BUFFER: usize = 1024;
```
ramfox, line 30: arbitrary buffer size.

https://github.com/ipfs/go-unixfs/blob/e0ef9c44e92cab793bdf321ac1531ceff418fd5f/io/directory.go#L21-L27
from `go-unixfs`, the comment on line 26 says that the "ProtoNode doesn't use the Data field so this estimate is pretty accurate)"

But as far as I can tell, in our implementation, when we encode the Protobuf Dag, we _are_ using the `data` field, so it seemed prudent to have some kind of buffer to represent the content that can exist in the field.

iroh-resolver/src/unixfs_builder.rs
```diff
@@ -1,55 +1,172 @@
...
+
+impl Default for DirectoryBuilder {
+    fn default() -> Self {
+        Self {
+            path: None,
+            name: None,
+            entries: Default::default(),
+            typ: DirectoryType::Basic,
+            estimated_size: 0,
+            chunker: Chunker::fixed_size(),
```
ramfox, line 73: Okay, so this also feels a bit odd.

the go impl just assumes chunks are 2MB.

Adding a `chunker` here to give some semblance of configurability. But it also feels odd because this chunker is potentially independant of any chunker being used to determine the chunk size of any files belonging in this directory. This might be fine. The reason we need to pay attention to chunk size here is to calculate whether or not we need a HAMT, not to ensure that all of our chunks are uniform. But it seemed odd to me so I thought I'd highlight it.

iroh-resolver/src/unixfs_builder.rs
```diff
@@ -1,55 +1,172 @@
...
+                    let f = file_builder.build().await?;
+                    builder.add_file(f);
+                } else if path.is_dir() {
+                    let mut dir_builder = DirectoryBuilder::new();
+                    dir_builder.with_path(path);
+                    let d = dir_builder.build().await?;
+                    builder.add_dir(d);
+                } else {
+                    anyhow::bail!("directory entry is neither file nor directory")
+                }
```
ramfox, line 146: reiterating my comments here:
this seems dumb, but when I poked around this seemed like that way to differentiate `FileTypes`

Also, not sure what I should be doing to add `symlinks`. Should I be adding a `struct SymlinkBuilder` & `struct Symlink`? I know we have `UnixfsNode` representations for them.

iroh-resolver/src/unixfs_builder.rs
```diff
@@ -1,55 +1,172 @@
...
+            }
+            self = builder;
+        }
+
+        if self.typ == DirectoryType::Hamt {
+            // Question: is it too late to be erroring here? should we be
+            // erroring when we add the entry? to catch this as soon as
+            // possible?
+            anyhow::bail!("too many links to fit into one chunk, must be encoded as a HAMT. However, HAMT creation has not yet been implemented.");
+        }
```
ramfox, line 156: This is more of a question about changing the current API for DirectoryBuilder methods.

Should we be returning `Result` for `add_file` & `add_dir`, and erroring when we get to a hamt sized directory?

iroh-resolver/src/unixfs_builder.rs
```diff
@@ -361,6 +480,39 @@ pub async fn add_file<S: Store>(path: &Path, rpc: Option<S>) -> Result<Cid> {
...
+/// Adds a directory.
+/// - storing the content using `rpc.store`
+/// - returns the root Cid
+/// - wraps into a UnixFs directory to preserve the filename
+pub async fn add_dir<S: Store>(path: &Path, rpc: Option<S>, _recursive: bool) -> Result<Cid> {
+    ensure!(path.is_dir(), "provided path was not a directory");
+
+    let mut dir = DirectoryBuilder::new();
+    dir.with_path(PathBuf::from(&path));
+    let dir = dir.build().await?;
```
ramfox, line 492: my only... question/concern with this is that we are potentially opening up a huge number of readers for each entry in the dir. We don't consume those readers until we `dir.encode()`

Not sure how much I should be worrying about this.

iroh-resolver/src/unixfs_builder.rs
dignifiedquire: @ribasushi please help

iroh-resolver/src/unixfs_builder.rs
dignifiedquire: a directory doesn’t use a chunker, lets chat about this

iroh-resolver/src/unixfs_builder.rs
```diff
@@ -34,12 +71,22 @@ impl DirectoryBuilder {
     }
 
     pub fn add_file(&mut self, file: File) -> &mut Self {
+        self.estimated_size += self.estimated_size + file.name().len() + CID_LEN;
```
dignifiedquire: this adds self.estimated_size twice

iroh-resolver/src/chunker.rs
```diff
@@ -73,6 +73,12 @@ impl Chunker {
             }
         }
     }
+
+    pub fn chunk_size(&self) -> usize {
```
dignifiedquire, line 77: most chunkerd will not have this, should avoid an explicit assumption in the api that this exists.

iroh-resolver/src/unixfs_builder.rs
```diff
@@ -1,63 +1,62 @@
...
-    files: Vec<File>,
+    entries: Vec<Entry>,
 }
 
 impl Directory {
     pub fn name(&self) -> &str {
         &self.name
     }
 
+    /// wrap an entry in an unnamed directory
```
dignifiedquire, line 50: doc comments should be full sentences with capitalization and punctuation

iroh-resolver/src/unixfs_builder.rs
ramfox: conclusion here: we will eventually support Symlinks, but not for this pr. Check for symlinks & error for now.

iroh-resolver/src/unixfs_builder.rs
```diff
@@ -102,92 +117,67 @@ impl Directory {
...
 
-        f.debug_struct("FileBuilder")
-            .field("name", &self.name)
-            .field("content", &content)
-            .field("chunker", &self.chunker)
-            .finish()
+impl PartialEq for Content {
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (Content::Reader(_), Content::Reader(_)) => true,
```
dignifiedquire, line 141: this should not return `true`, as this means that two different content readers will be considered equal

iroh-resolver/src/unixfs_builder.rs
```diff
@@ -102,92 +117,67 @@ impl Directory {
...
     tree_builder: TreeBuilder,
+    chunker: Chunker,
 }
 
 impl Debug for File {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.debug_struct("File")
             .field("name", &self.name)
-            .field("nodes", &"Stream<Item = Result<UnixfsNode>>")
+            .field("content", &self.content)
```
dignifiedquire, line 161: missing some fields here

iroh-resolver/src/unixfs_builder.rs
```diff
@@ -102,92 +117,67 @@ impl Directory {
...
-        let hash = cid::multihash::Code::Sha2_256.digest(root);
-        let codec = match self {
-            EncodedFile::Raw(_) => Codec::Raw,
-            EncodedFile::Chunked { .. } => Codec::Sha2256,
-        };
-
-        Cid::new_v1(codec as _, hash)
+impl File {
+    pub fn name(&self) -> String {
+        self.name.clone()
```
dignifiedquire, line 168: this should return `&str` and any needed cloning should happen at the call site

iroh-resolver/src/unixfs_builder.rs
```diff
@@ -197,74 +187,197 @@ impl File {
...
-    fn default() -> Self {
-        Self {
-            name: None,
-            content: None,
-            chunker: chunker::Chunker::fixed_size(),
-        }
+impl Debug for FileBuilder {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("FileBuilder")
+            .field("path", &self.path)
```
dignifiedquire, line 215: missing fields

iroh-resolver/src/unixfs_builder.rs
```diff
@@ -197,74 +187,197 @@ impl File {
...
+        self.typ = DirectoryType::Hamt;
+        self
+    }
+
+    pub fn allow_recursive(&mut self) -> &mut Self {
+        self.recursive = true;
+        self
+    }
+
+    pub fn with_name<N: Into<String>>(&mut self, name: N) -> &mut Self {
```
dignifiedquire, line 324: why the change to `with_name` instead of `name`?

iroh-resolver/src/unixfs_builder.rs
```diff
@@ -197,74 +187,197 @@ impl File {
...
+        self.name = Some(name.into());
+        self
+    }
+
+    pub fn add_dir(&mut self, dir: Directory) -> Result<&mut Self> {
+        if !self.recursive {
+            anyhow::bail!("recursive directories not allowed");
```
dignifiedquire, line 331: you can write this nicer as 

```rust
use anyhow::ensure;

ensure!(self.recursive, "recursive directories not allowed");
```

iroh-resolver/src/unixfs_builder.rs
```diff
@@ -197,74 +187,197 @@ impl File {
...
+        }
+        self.entries.push(Entry::Directory(dir));
+        Ok(self)
+    }
+
+    pub fn add_file(&mut self, file: File) -> &mut Self {
+        if self.typ == DirectoryType::Basic && self.entries.len() >= DIRECTORY_LINK_LIMIT {
+            self.typ = DirectoryType::Hamt
+        }
+        self.entries.push(Entry::File(file));
```
dignifiedquire, line 345: maybe have a shared `fn push_entry` to have the HAMT size check in a single place

iroh-resolver/src/unixfs_builder.rs
```diff
@@ -197,74 +187,197 @@ impl File {
...
+        self
+    }
+
+    pub fn build(self) -> Result<Directory> {
+        let DirectoryBuilder {
+            name, entries, typ, ..
+        } = self;
+
+        if typ == DirectoryType::Hamt {
+            anyhow::bail!("too many links to fit into one chunk, must be encoded as a HAMT. However, HAMT creation has not yet been implemented.");
```
dignifiedquire, line 355: `ensure!`

iroh-resolver/src/unixfs_builder.rs
```diff
@@ -325,45 +450,75 @@ pub async fn add_file<S: Store>(path: &Path, rpc: Option<S>) -> Result<Cid> {
...
+    let mut dir = DirectoryBuilder::new();
+    dir.with_name(
+        path.file_name()
+            .and_then(|s| s.to_str())
+            .unwrap_or_default(),
+    );
+    if recursive {
+        dir.allow_recursive();
+    }
+    let entries = std::fs::read_dir(path.clone())?;
```
dignifiedquire, line 465: this should use tokios read_dir to make sure we do not block

iroh-resolver/src/unixfs_builder.rs
ramfox: Ahhh, thank you, I meant to ask you about this!!! 

Wasn't sure if the readers should always return `false`, or if there was some way to figure out how to force checking for an equivalency.

iroh-resolver/src/unixfs_builder.rs
ramfox: We don't have to do this. I just personally find it confusing, expecting a method called `name` to return the content in the name field.

I've also seen `named` as a way to describe "this method is adding a name".

If you have a personal preference for `name`, or it's just the "rust-y" way to do it, I'm happy to change it back!

iroh-resolver/src/unixfs_builder.rs
ramfox: TYTYTY. I was wondering about this. Need to get it through my head that we are doing async read/write!!!

iroh-resolver/src/unixfs_builder.rs
dignifiedquire: just return false, we can't really compare them

iroh-resolver/src/unixfs_builder.rs
dignifiedquire: I like consistency, so either all the builder methods should be prefixed with `with_` or `add_`, or we should drop it. up to you

ramfox: dropping it, if we are going for uniform.

dignifiedquire: sorry, I didn't mean to remove the `add_`

dignifiedquire (approved):

PR #215, opened 2022-08-31
fix: auto-raise FD limits where possible

Arqu:


iroh-util/src/lib.rs
```diff
@@ -18,6 +18,7 @@ use dirs::home_dir;
 use tracing::debug;
 
 const IROH_DIR: &str = ".iroh";
+const DEFAULT_NOFILE_LIMIT: u64 = 1 << 16;
```
dignifiedquire, line 21: an we please use regular numbers not shifts, they are just harder to read, eg 1024 * 1024 is much easier to read

iroh-util/src/lib.rs
```diff
@@ -152,6 +153,22 @@ pub fn verify_hash(cid: &Cid, bytes: &[u8]) -> Option<bool> {
...
 }
 
+/// If supported sets a preffered limit for file descriptors.
+#[cfg(unix)]
+pub fn increase_fd_limit() -> std::io::Result<u64> {
+    let (_, hard) = rlimit::Resource::NOFILE.get()?;
+    let target = std::cmp::min(hard, DEFAULT_NOFILE_LIMIT);
+    rlimit::Resource::NOFILE.set(target, hard)?;
+    let (soft, _) = rlimit::Resource::NOFILE.get()?;
+    if soft < DEFAULT_NOFILE_LIMIT >> 5 {
```
dignifiedquire, line 163: same as above

dignifiedquire (dismissed):

iroh-one/src/main.rs
```diff
@@ -35,6 +36,14 @@ async fn main() -> Result<()> {
     )
     .unwrap();
 
+    #[cfg(unix)]
+    {
+        match iroh_util::increase_fd_limit() {
```
ralphtheninja, line 41: @dignifiedquire I'm learning rust now and curious about code duplication here. Since there are three identical pieces of calls to `iroh_util::increase_fd_limit()` with `match` and error checks. Would it be frowned upon to move this error check into `iroh_util::increase_fd_limit()` instead? (also the `use tracing::{debug, error}` can be moved).

dignifiedquire: potentially, I m a bit torn on it here as the usages are different binaries. but agree with the sentiment

Arqu: I mostly did it because error handling might be different in different use cases even though they are the same now 🤷

mxinden:
:pray:

PR #218, opened 2022-09-01
refactor: use OS-specific config & data dirs, default iroh-store dir

### LINKED ISSUE #142: Improve default paths
use https://github.com/xdg-rs/dirs to follow the various os specific recommendations

b5:
switch to using https://dirs.dev/ standard application directories instead of $HOME/.iroh

iroh-store now uses a default directory for writing rocksDB files instead of explicitly requiring a store path. Defaults to `/iroh/store` within the OS application data directory.

closes #142


iroh-store/src/config.rs
```diff
@@ -17,6 +17,14 @@ pub const CONFIG_FILE_NAME: &str = "store.config.toml";
 /// For example, `IROH_STORE_PATH=/path/to/config` would set the value of the `Config.path` field
 pub const ENV_PREFIX: &str = "IROH_STORE";
 
+/// the path to data directory. If arg_path is `None`, the default iroh_data_path()/store is used
+/// iroh_data_path() returns an operating system-specific directory
+/// falls back to an empty path in the unlikely scenario where iroh_data_path() is `None`
+/// likely b/c iroh is not running on a Linux/Windows/MacOS system
+pub fn config_data_path(arg_path: Option<PathBuf>) -> PathBuf {
+    arg_path.unwrap_or_else(|| iroh_data_path("store").unwrap_or_else(|| PathBuf::from("")))
+}
```
b5, line 26: unclear on if this should live here, or exist at all. the nested `unwrap_or_else` also feels smelly

iroh-util/src/lib.rs
```diff
@@ -44,15 +44,43 @@ pub async fn block_until_sigint() {
     ctrlc_oneshot.await.unwrap();
 }
 
-/// Path to the iroh home directory.
-pub fn iroh_home_root() -> Option<PathBuf> {
-    let home = home_dir()?;
-    Some(Path::new(&home).join(IROH_DIR))
+/// Returns the path to the user's iroh config directory.
+///
+/// The returned value depends on the operating system and is either a `Some`, containing a value from the following table, or a `None`.
```
b5, line 49: do our comments use a ruler? on other projects I've line-wrapped at 80 characters of width.

iroh-store/src/main.rs
```diff
@@ -16,10 +15,10 @@ async fn main() -> anyhow::Result<()> {
     let version = env!("CARGO_PKG_VERSION");
     println!("Starting iroh-store, version {version}");
 
-    let sources = vec![iroh_home_path(CONFIG_FILE_NAME), args.cfg.clone()];
+    let sources = vec![iroh_config_path(CONFIG_FILE_NAME), args.cfg.clone()];
     let config = make_config(
         // default
-        Config::new_grpc(args.path.clone().unwrap_or_else(|| PathBuf::from(""))),
+        Config::new_grpc(config_data_path(args.path.clone())),
```
b5, line 21: testing locally on a mac rockDB `create_if_missing` option creates all nested folders, so feeding this `iroh/store` `PathBuf` worked without any additional changes

iroh-util/src/lib.rs
dignifiedquire: yes, cargo fmt checks line length and I think breaks them at 100

iroh-util/src/lib.rs
```diff
@@ -14,10 +14,10 @@ use cid::{
     Cid,
 };
 use config::{Config, ConfigError, Environment, File, Map, Source, Value, ValueKind};
-use dirs::home_dir;
```
dignifiedquire, line 17: can we remove this dep now?

iroh-util/src/lib.rs
```diff
@@ -44,15 +44,43 @@ pub async fn block_until_sigint() {
...
+/// | Windows  | `{FOLDERID_RoamingAppData}/iroh`              | C:\Users\Alice\AppData\Roaming\iroh           |
+pub fn iroh_data_root() -> Option<PathBuf> {
+    let path = dirs_next::data_dir()?;
+    Some(path.join(&IROH_DIR))
 }
 
-/// Path that leads to a file in the iroh home directory.
-pub fn iroh_home_path(file_name: &str) -> Option<PathBuf> {
-    let path = iroh_home_root()?.join(file_name);
+// Path that leads to a file in the iroh data directory.
```
dignifiedquire, line 81: tthree slashes

iroh-store/src/config.rs
dignifiedquire: we should error out and not fall back to an empty path, that is very surprising

iroh-util/src/lib.rs
```diff
@@ -44,15 +44,43 @@ pub async fn block_until_sigint() {
...
+/// Returns the path to the user's iroh data directory.
+///
+/// The returned value depends on the operating system and is either a `Some`, containing a value from the following table, or a `None`.
+///
+/// | Platform | Value                                         | Example                                  |
+/// | -------- | --------------------------------------------- | ---------------------------------------- |
+/// | Linux    | `$XDG_DATA_HOME`/iroh or `$HOME`/.local/share/iroh | /home/alice/.local/share/iroh                 |
+/// | macOS    | `$HOME`/Library/Application Support/iroh      | /Users/Alice/Library/Application Support/iroh |
+/// | Windows  | `{FOLDERID_RoamingAppData}/iroh`              | C:\Users\Alice\AppData\Roaming\iroh           |
+pub fn iroh_data_root() -> Option<PathBuf> {
```
dignifiedquire, line 76: Lets return `Result`s for these, and error when they are not available. All uses cases atm are to error if we do't have a path

iroh-store/src/config.rs
Arqu: It errors out lower in the stack because it cant use an "" path, but yeah should be handled higher up.

iroh-util/src/lib.rs
b5: in2it. it'll make for a more explicit division from embedded systems

iroh-store/src/config.rs
b5: ok, changes made. latest refactor handles this by erroring out when no path is present. error check happens before DB creation.

iroh-store/src/config.rs
```diff
@@ -17,6 +17,15 @@ pub const CONFIG_FILE_NAME: &str = "store.config.toml";
 /// For example, `IROH_STORE_PATH=/path/to/config` would set the value of the `Config.path` field
 pub const ENV_PREFIX: &str = "IROH_STORE";
 
+/// the path to data directory. If arg_path is `None`, the default iroh_data_path()/store is used
+/// iroh_data_path() returns an operating system-specific directory
+pub fn config_data_path(arg_path: Option<PathBuf>) -> Result<PathBuf> {
+    match arg_path {
+        Some(p) => Ok(p),
+        None => iroh_data_path("store"),
+    }
```
b5, line 26: I'm unclear on weather or not there's a more succinct way to express this

fabricedesre: I think `arg_path.map_or(iroh_data_path("store"), |v| Ok(v))` is equivalent - but not easier to read!

b5: oooh I hadn't found `map_or`. Thanks for the tip @fabricedesre! I agree with you that the  `match` is easier to read, and will leave it as is unless I'm overruled by dig 😄

dignifiedquire: seems fine and readable

dignifiedquire (approved):
lgtm

Arqu (dismissed):
Looks good from the ops side too. All of the deployment stuff is hardwired to avoid sudden config changes like this.

Arqu (approved):
👍

Issue #217, opened 2022-09-01
gateway: pass in custom resolver


Arqu:
@fabricedesre:
> @Arqu & @dignifiedquire I'm looking at adding back http gateway racing to the resolving code but I'm not sure how to do it without touching iroh-gateway itself since the resolver is created in iroh-gateway/src/client.src
> One way would be to turn the client's State in iroh-gateway/src/core.rs to a trait instead, and pass it in make_state() like we do for the configuration.

Arqu:
Think we can simplify that. The resolver is created when the Client is instantiated which is on `Core::new`. Maybe we can simply make an `Optional<Resolver>` arg on the `make_with_state` function and simply set a custom resolver on the client if present? @dignifiedquire any input on this. Might be worthwhile to convert the Core to follow a builder pattern so you can do `Core::builder().with_state(state).with_resolver(resolver)....`

fabricedesre:
Fixed by #220

PR #220, opened 2022-09-01
Allow to pass custom resolvers

fabricedesre:
The main change is to switch `iroh-gateway/src/client.rs#Client` to be:
`pub struct Client<T: ContentLoader>` and to sprinkle the generic and
other needed bounds where needed.

`iroh-one/src/content_loader.rs` showcases a custom content loader. For now it's
similar to the regular one without the gateway racing implementation.

### DIFF
--- a/iroh-gateway/src/client.rs
+++ b/iroh-gateway/src/client.rs
@@ -11,7 +11,7 @@ use iroh_metrics::{
     observe, record,
 };
 use iroh_resolver::resolver::{
-    CidOrDomain, Metadata, Out, OutMetrics, OutPrettyReader, Resolver, Source,
+    CidOrDomain, ContentLoader, Metadata, Out, OutMetrics, OutPrettyReader, Resolver, Source,
 };
 use tokio::io::AsyncReadExt;
 use tokio_util::io::ReaderStream;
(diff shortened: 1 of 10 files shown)


fabricedesre:
This is to fix https://github.com/n0-computer/iroh/issues/217

fabricedesre:
note: all tests are passing locally on a Mac

iroh-gateway/src/utils.sh
```diff
@@ -0,0 +1,226 @@
+#!/bin/bash
```
dignifiedquire, line 1: did you mean to add this?

fabricedesre: No, the shell scripts are a leftover of a `git add src/` ...

fabricedesre: fixed in latest version

fabricedesre:
Rebased to current tree. ping @dignifiedquire & @Arqu

iroh-gateway/src/client.rs
```diff
@@ -69,8 +66,8 @@ impl http_body::Body for PrettyStreamBody {
     }
 }
 
-impl Client {
-    pub fn new(rpc_client: &iroh_rpc_client::Client) -> Self {
+impl<T: ContentLoader + Clone + std::marker::Unpin> Client<T> {
```
dignifiedquire, line 69: is `Clone` needed, I thought `ContentLoader` already required that

dignifiedquire (approved):
not the prettiest change, but makes sense from my side

iroh-gateway/src/client.rs
fabricedesre: You're right, fixed in latest update.

Arqu (approved):
Agree it's not the most ergonomic setup but functionally I'm fine with it and these don't move too much around.

Arqu:
@fabricedesre just noticed the branch is out of date. Can you rebase and we can merge.

PR #221, opened 2022-09-01
Add iroh-one to CI steps

fabricedesre:
- build with the 'uds-gateway' feature to catch build errors.
- push the iroh-one binary to S3.

Ping @Arqu

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -377,7 +377,7 @@ jobs:
           profile-name: default
       - run:
           name: build aarch64 release
-          command: cargo build --release -v
+          command: cargo build --release -v --features=uds-gateway
       - push-to-s3:
           path: ./target/release/
           os: linux
@@ -405,7 +405,7 @@ jobs:
       - run:
           name: build x86_64_darwin
           no_output_timeout: 20m
-          command: cargo build --release -v
+          command: cargo build --release -v --features=uds-gateway
       - push-to-s3:
           path: ./target/release/
           os: darwin
@@ -423,7 +423,7 @@ jobs:
       - run:
           name: build aarch64_darwin
           no_output_timeout: 20m
-          command: cargo build --release -v --target aarch64-apple-darwin
+          command: cargo build --release -v --target aarch64-apple-darwin --features=uds-gateway
       - push-to-s3:
           path: ./target/aarch64-apple-darwin/release/
           os: darwin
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -348,7 +350,7 @@ jobs:
           profile-name: default
       - run:
           name: build x86_64 release
-          command: cargo build --release -v
+          command: cargo build --release -v --features=uds-gateway
```
Arqu, line 353: Need to replicate this for `build_release_aarch64` & `build_release_osx`

Arqu (changes requested):
Looks good, lets just make sure other architectures are supported too.

.circleci/config.yml
fabricedesre: Done in 0ec0ec0c97362647768b980f547597cbdf84542b

Arqu (changes requested):
I'm sorry but can we take off the `--features=uds-gateway` from the builds. Don't think I want to have those uploaded as the "default" builds. Those who want to include specific features should generally build their own and if it becomes widely used we can move that into a config value rather than compile time thing.

fabricedesre:
> I'm sorry, been thinking about this, but can we take off the `--features=uds-gateway` from the builds. Don't think I want to have those uploaded as the "default" builds. Those who want to include specific features should generally build their own and if it becomes widely used we can move that into a config value rather than compile time thing.

I understand your argument, but that also means that we have no CI coverage for build breaking with that feature enabled. Or should we build both with and without the feature, and only upload the featureless one?

Arqu:
> > I'm sorry, been thinking about this, but can we take off the `--features=uds-gateway` from the builds. Don't think I want to have those uploaded as the "default" builds. Those who want to include specific features should generally build their own and if it becomes widely used we can move that into a config value rather than compile time thing.
> 
> I understand your argument, but that also means that we have no CI coverage for build breaking with that feature enabled. Or should we build both with and without the feature, and only upload the featureless one?

Right, that makes sense. So what I would do instead is introduce a new CI job and run `cargo check --all-features` and `cargo check --no-default-features` in it as a prequel to building releases and make it depend on `cargo fetch` and `clippy` and `rustfmt`. That way we get build coverage. Potentially extend the linux tests to run tests in the same fashion.

Arqu:
This is now covered by CI. I'd be happy to revisit adding iroh one to the official builds.

dignifiedquire:
Closing  this PR. @fabricedesre @Arqu whoever gets to it first, lets get iroh-one added to the binary list.

PR #225, opened 2022-09-02
Switch to current Rustc stable version (1.63)

fabricedesre:
'-A clippy::derive-partial-eq-without-eq' was added to clippy to silence this lint which is not helpful for us
and was also triggering errors in prost generated code which we could not really fix otherwise.


Arqu (approved):
👍 thanks

PR #231, opened 2022-09-11
Always delete an UDS file before binding it

fabricedesre:
The RAII struct `UdsGuard` works well when the binary terminates normally but this is unfortunately not always the case:
- crashes.
- on Android, daemons restarted by the init system.
- power loss on mobile devices.

To avoid these failures, this patch forces the deletion before trying to bind the socket. This means that we don't really need the `UdsGuard` anymore -  let me know if I should also remove it. cc @dignifiedquire

### DIFF
--- a/iroh-rpc-types/src/macros.rs
+++ b/iroh-rpc-types/src/macros.rs
@@ -54,6 +54,14 @@ macro_rules! proxy {
                             }
                         }
 
+                        // We still need to remove the file before binding, because the above
+                        // guard won't run in some situations:
+                        // - crash
+                        // - Android start/stop
+                        // - device power loss
+                        // ...
+                        let _ = std::fs::remove_file(&path);
+
                         let uds = UnixListener::bind(&path)
                             .with_context(|| format!("failed to bind to {}", path.display()))?;
                         let _guard = UdsGuard(path.clone().into());


dignifiedquire:
I don’t think unconditionally deleting these is a good idea, because now you break an already running system if you run the binary twice, instead of erroring out.

dignifiedquire:
Thoughts @fabricedesre ? This is tricky, any idea how other applications typically do this?

fabricedesre:
> Thoughts @fabricedesre ? This is tricky, any idea how other applications typically do this?

I've seen 2 approaches:
- use a file containing the PID of the running instance. In theory this can be racy, in practice that works well. In iroh we could provide that to all programs with a helper in iroh-utils.
- scan /proc/$pids/cmdline or equivalent to find an already running instance. It's quite tedious and can be slow, and OS specific.

fabricedesre:
> Cleaning them up in all cases including program crashes is the hard part in my humble experience.

Having the pid in the lock file allows you to ignore / override it after a crash because you can check that the pid is actually one of a running process. Of course pid reuse is a thing, but well...

I can take that bug if that helps.

dignifiedquire:
On linux at least `fcntl` works pretty well I believe, as the OS releases the lock when the process dies

dignifiedquire:
> I can take that bug if that helps.

Yes, that would be great

dignifiedquire:
@fabricedesre can we close this, or is this still needed?

fabricedesre:
I forgot about it because I kept it in my fork, so feel free to close. It likely less of an issue now with the lock files.

PR #234, opened 2022-09-15
gitignore .vscode files

faassen:
Ignore any local .vscode config files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 .env
 Cargo.lock
 foo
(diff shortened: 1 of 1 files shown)


dignifiedquire:
I prefer folks put these into their own gitignore : https://tekin.co.uk/2020/03/maintain-a-global-git-ignore-file

faassen:
I've now created my own `.gitignore_global` file, thanks for the tip!

PR #235, opened 2022-09-16
Remove invalid If-Modified-Since header

fabricedesre:
"0" is not a valid value (see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Last-Modified) and for ipfs/ fetches the etag header and the immutable cache control value give the proper caching behavior from browsers.


fabricedesre:
cc @Arqu

Arqu:
Lgtm
Thanks

PR #236, opened 2022-09-16
Enable the CompressionLayer middleware

fabricedesre:
Verified with Firefox's devtools.


fabricedesre:
cc @Arqu

dignifiedquire (dismissed):
Seems fine to me, but want to hear @Arqu s comments before merging

Arqu (approved):
We should benchmark it for some scenarios, but am 99% sure its a net positive.

PR #239, opened 2022-09-20
(re-)add Cargo.lock

flokli:
iroh produces binaries.

As per
https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html, this means it should have its Cargo.lock file committed to the repository.


dignifiedquire:
Thanks, but we've decided to skip the lock file for now until the code is more stable, as they mostly generate merge conflicts at the moment 😅. Do you have a specific need for them?

fabricedesre:
> Thanks, but we've decided to skip the lock file for now until the code is more stable, as they mostly generate merge conflicts at the moment 😅. Do you have a specific need for them?

In the Capyloon fork we commit `Cargo.lock` and all crates vendored, because the Android build system disallow network access during builds. We need a fork anyway for other purposes so it's just a secondary reason for us.

flokli:
> Thanks, but we've decided to skip the lock file for now until the code is more stable, as they mostly generate merge conflicts at the moment sweat_smile. Do you have a specific need for them?

Sorry for not replying sooner. Like @fabricedesre, I'm also packaging this in a stricter build environment which needs these lock files to function. I wanted to give some of your binaries a try. Of course I can manually create them, but having them in the repo would ensure everyone is using the same versions. For reproducibility reasons, I'd really like to see these being included, but it's ultimately your choice. Do you think it's still too much churn?

flokli:
I had someone else opening a nixpkgs PR packaging iroh, and they also had to manually provide a Cargo.lock file: https://github.com/NixOS/nixpkgs/pull/198678#pullrequestreview-1161173069

With the 0.1.0 release now out, would you reconsider shipping a lockfile in the repo? It'd really make it easier for others to build iroh, and you'll also have more guarantees about what third-party libraries it is built with.

dignifiedquire:
We will be adding one eventually, see https://github.com/n0-computer/iroh/issues/548 but not right now. So I am going to close this PR for now. But we have the issue of tracking to add it once we are ready for it.

PR #240, opened 2022-09-21
feat: introduce an Api trait and its implementation

faassen:
The implementation is a thin wrapper over the RPC client.

This is exposed as a public API and will also be used to implement the CLI in iroh-ctl.

The idea is that the trait can be reimplemented to help test the iroh-ctl later, and it defines clearly what the public API is.

We smooth over the underlying client API somewhat: make `Cid` and `PeerId` into reference arguments, and use `&[]` slices instead of vectors for arguments as well.

In order to implement `ClientApi` I had to expose `P2pClient` and `StoreClient` in `iroh-rpc-client`.

### DIFF
--- a/iroh/Cargo.toml
+++ b/iroh/Cargo.toml
@@ -16,3 +16,6 @@ async-trait = "0.1.53"
 tokio = { version = "1" }
 bytes = "1.1.0"
 libp2p = "0.48"
+tracing = "0.1.34"
+futures = "0.3.21"
+async-stream = "0.3.3"
--- a/iroh/Cargo.toml
+++ b/iroh/Cargo.toml
@@ -7,4 +7,12 @@ readme = "README.md"
 description = "coming soon"
 
 [dependencies]
-
+cid = "0.8.5"
+iroh-resolver = { path = "../iroh-resolver" }
+iroh-rpc-types = { path = "../iroh-rpc-types" }
+iroh-rpc-client = { path = "../iroh-rpc-client" }
+anyhow = "1"
+async-trait = "0.1.53"
+tokio = { version = "1" }
+bytes = "1.1.0"
+libp2p = "0.48"
--- a/iroh/src/api.rs
+++ b/iroh/src/api.rs
@@ -5,6 +5,7 @@ use anyhow::Result;
 use async_trait::async_trait;
 use bytes::Bytes;
 use cid::Cid;
+use iroh_resolver::resolver::Path as IpfsPath;
 use libp2p::gossipsub::MessageId;
 use libp2p::{Multiaddr, PeerId};
 
@@ -29,8 +30,8 @@ pub trait Accessors<P: P2p, S: Store> {
 
 #[async_trait]
 pub trait GetAdd {
-    async fn get(&self, cid: &Cid, path: &Path) -> Result<()>;
-    async fn add(&self, path: &Path) -> Result<Cid>;
+    async fn get(&self, ipfs_path: &IpfsPath, output_path: Option<&Path>) -> Result<()>;
+    async fn add(&self, path: &Path, recursive: bool, no_wrap: bool) -> Result<Cid>;
 }
 
 impl<T: Accessors<P, S> + GetAdd, P: P2p, S: Store> Api<P, S> for T {}
--- a/iroh/src/api.rs
+++ b/iroh/src/api.rs
@@ -1,0 +1,104 @@
+use std::collections::{HashMap, HashSet};
+use std::path::Path;
+
+use anyhow::Result;
+use async_trait::async_trait;
+use bytes::Bytes;
+use cid::Cid;
+use libp2p::gossipsub::MessageId;
+use libp2p::{Multiaddr, PeerId};
+
+pub struct Id {
+    pub peer_id: PeerId,
+    pub listen_addrs: Vec<Multiaddr>,
+    pub local_addrs: Vec<Multiaddr>,
+}
+
+#[derive(Debug)]
+pub enum Ping {
+    PeerId(PeerId),
+    Multiaddr(Multiaddr),
+}
+
+pub trait Api<P: P2p, S: Store>: Main + Accessors<P, S> + GetAdd {}
+
+#[async_trait]
+pub trait Main {
+    async fn version(&self) -> Result<String>;
+}
+
+#[async_trait]
+pub trait Accessors<P: P2p, S: Store> {
+    fn p2p(&self) -> Result<P>;
+    fn store(&self) -> Result<S>;
+}
+
+#[async_trait]
+pub trait GetAdd {
+    // XXX get and add are centered around the filesystem.
+    // We can imagine an underlying version that produces a stream of
+    // Out as well.
+    async fn get(&self, cid: &Cid, path: &Path) -> Result<()>;
+    async fn add(&self, path: &Path) -> Result<Cid>;
+}
+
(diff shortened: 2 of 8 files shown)


iroh/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+pub mod api;
+mod clientapi;
+
+pub use crate::clientapi::ClientApi as Api;
+pub use bytes::Bytes;
+pub use cid::Cid;
+pub use libp2p::gossipsub::MessageId;
+pub use libp2p::{Multiaddr, PeerId};
```
fabricedesre, line 8: Making the client api depend on libp2p is problematic because of how much dependencies libp2p comes with. Something leaner and more self contained would be better.

faassen: I've brought this up to @dignifiedquire

ramfox (changes requested):
very fun!

Some general notes:
- There are a few places with stray comments, I don't think I pointed them all out, but just something to be aware of as you are going back through the code
- speaking of comments, since this will be a public API, we are going to need documentation that can be generated for the docs.rs site for each public method/trait/enum/struct. I'd appreciate it if you could add `///` comments to any pieces you feel comfortable with. I'll help with this.

iroh/Cargo.toml
```diff
@@ -7,4 +7,12 @@ readme = "README.md"
...
-
+cid = "0.8.5"
+iroh-resolver = { path = "../iroh-resolver" }
+iroh-rpc-types = { path = "../iroh-rpc-types" }
+iroh-rpc-client = { path = "../iroh-rpc-client" }
+anyhow = "1"
+async-trait = "0.1.53"
+tokio = { version = "1" }
+bytes = "1.1.0"
+libp2p = "0.48"
```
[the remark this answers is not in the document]
faassen: In each case here I used the procedure of looking at another crate with the same dependency and copying that over here.

That's not a policy but I think it's enough for this PR.

iroh/src/api.rs
```diff
@@ -0,0 +1,104 @@
...
+#[async_trait]
+pub trait P2pId: std::marker::Sync {
+    async fn p2p_version(&self) -> Result<String>;
+    async fn local_peer_id(&self) -> Result<PeerId>;
+    async fn addrs_listen(&self) -> Result<Vec<Multiaddr>>;
+    async fn addrs_local(&self) -> Result<Vec<Multiaddr>>;
+    // can be implemented on the trait itself as it combines others
+    async fn id(&self) -> Result<Id>;
+    // async fn addrs gets a map right now
+    async fn peers(&self) -> Result<HashMap<PeerId, Vec<Multiaddr>>>;
```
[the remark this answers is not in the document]
faassen: I don't understand this comment. Do you mean it changed from `addrs` to `peers`? I didn't change anything as this API is new, but it may definitely be inconsistent with the existing CLI somewhere. I saw both a `peers` command and an `addr` command in the CLI code if I recall correctly.

I think in general this whole API  needs a rethink the light of the discussion we had yesterday about the complex relationship between all this stuff.

iroh/src/api.rs
```diff
@@ -0,0 +1,104 @@
...
+    PeerId(PeerId),
+    Multiaddr(Multiaddr),
+}
+
+pub trait Api<P: P2p, S: Store>: Main + Accessors<P, S> + GetAdd {}
+
+#[async_trait]
+pub trait Main {
+    async fn version(&self) -> Result<String>;
+}
```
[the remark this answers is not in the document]
faassen: Yes, I think that makes the most sense. I'll just remove it for now. Version makes more sense for the underlying nodes like p2p.

faassen:
Thanks for the feedback! async clouded my brain making me over-async.

Concerning adding comments to produce public API documentation. I'm a bit wary of doing this early as I think we're still going to shake out a lot of details when we implement the CLI. I'd prefer to work on the CLI proposal first, let the API reflect that, and then do a comment pass in a future PR.

faassen:
I've tried to integrate the `add` and `get`  code but bounced off both of them due to some async ``Send`` marker trait stuff.

The code in the CLI currently assumes that client is owned, not a reference like in my CLI code. But even once I fiddled with lifetimes and made client owned, I still get the problem.

So I think we need to look into this together.

faassen:
I've tried to integrate `block_put` but I don't think I can. I'm pretty sure that `block_put` needs to take either byte data or perhaps some kind of file handle as the CLI command in kubo works with stdin. But `BlockCommands::Put` is based around `PathBuf`, and that doesn't seem right. We're not adding a file at all, just a block directly, right?

faassen:
I try to look for opportunities to merge sooner rather than later. There are three issues left that I can determine:

* `add` and `get` are tricky; I've now prepared the API but I don't know how to integrate the code. I propose we do a later `add/get` PR where we figure this out.

* `block_put` as mentioned needs further discussion.

* As mentioned earlier, I want to hold off on API docs until more has been shaken out, in a future PR

None of these look like they have to block the merge, given the early state of this API in general.

iroh/src/api.rs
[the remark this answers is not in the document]
ramfox: yes I did mean from `addrs` to `peers`, my bad. Yea I agree about the API rethink.

faassen:
@dignifiedquire could you take a look at this? 

In particular @ramfox and I would like to know whether it's okay to use `#[async_trait(?Send)]` for the `GetAdd` trait. Without `?Send` it it's not possible to call the code in `getadd` without getting errors involving missing `Send` markers for `dyn std::future::Future<Output = std::result::Result<iroh_resolver::unixfs_builder::Directory, anyhow::Error>>`, `dyn libp2p::futures::Stream<Item = std::result::Result<(cid::CidGeneric<64_usize>, bytes::Bytes), anyhow::Error>>` and `(dyn tokio::io::AsyncRead + 'static)`, which puzzled us for a while.

The status of the various discussions:

* @ramfox and I agreed that API docs are going to land in a future PR once things shakes out

* the whole addresses API may need a rethink, also driven by what we want in the CLI. It's good enough to land for now as it mirrors the CLI.

iroh/src/api.rs
```diff
@@ -0,0 +1,93 @@
...
+
+pub trait Api<P: P2p, S: Store>: Accessors<P, S> + GetAdd {}
+
+pub trait Accessors<P: P2p, S: Store> {
+    fn p2p(&self) -> Result<P>;
+    fn store(&self) -> Result<S>;
+}
+
+#[async_trait(?Send)]
```
faassen, line 31: @dignifiedquire here's the `?Send`

iroh/src/api.rs
```diff
@@ -0,0 +1,93 @@
...
+impl<T: Accessors<P, S> + GetAdd, P: P2p, S: Store> Api<P, S> for T {}
+
+#[async_trait]
+pub trait P2pConnectDisconnect {
+    async fn connect(&self, peer_id: &PeerId, addrs: &[Multiaddr]) -> Result<()>;
+    async fn disconnect(&self, peer_id: &PeerId) -> Result<()>;
+}
+
+#[async_trait]
+pub trait P2pId: std::marker::Sync {
```
dignifiedquire, line 46: `Send` and `Sync` are auto imported, no need to add `std::marker::` infront of them

iroh/src/api.rs
dignifiedquire: If we require `GetAdd` to be `Send`, we shouldn't use `?Send` here

faassen:
We're going back to the drawing board. Closing this PR for the time being as I cook up something new.

PR #244, opened 2022-09-24
Update opentelemetry to 0.18

fabricedesre:
opentelemetry was pulling in tonic 0.6.2 while iroh uses the latest 0.8. With this update we don't have a 2 versions of tonic in the build anymore.


dignifiedquire:
thanks, that’s great

PR #252, opened 2022-09-28
fix: use the uds-gateway feature only to control the http uds endpoint

fabricedesre:
The current code also turns off the grpc control endpoint, but we need to keep it available in all configurations.

The fix in iroh-p2p/Cargo.toml makes it possible to build iroh-one or iroh-p2p without building the whole workspace.


fabricedesre:
cc @Arqu - This is something we overlooked in the initial iroh-one merge, but it's causing issue for the upcoming iroh-one rpc work.

Arqu (approved):
ty

PR #253, opened 2022-09-29
feat: multi-service rpc for iroh-one

fabricedesre:
This patch enables iroh-one rpc control enpoint to serve the same
calls as the independent gateway, store and p2p programs. To that end, 
the following changes are made:
- decouple iroh-one rpc from iroh-gateway rpc.
- create a iroh-one specific grpc definition, with 3 services.
- upddate the rpc macros to support multiple services.

### DIFF
--- a/iroh-gateway/src/bad_bits.rs
+++ b/iroh-gateway/src/bad_bits.rs
@@ -191,14 +191,14 @@ mod tests {
 
         let rpc_addr = "grpc://0.0.0.0:0".parse().unwrap();
         let content_loader = RpcClient::new(config.rpc_client.clone()).await.unwrap();
-        let handler = crate::core::Core::new(
-            Arc::new(config),
-            rpc_addr,
-            Arc::new(Some(RwLock::new(bbits))),
-            content_loader,
-        )
-        .await
-        .unwrap();
+        tokio::spawn(async move {
+            if let Err(err) = crate::rpc::new(rpc_addr, crate::rpc::Gateway::default()).await {
+                tracing::error!("Failed to run gateway rpc handler: {}", err);
+            }
+        });
+        let handler = crate::core::Core::new(Arc::new(config), Arc::new(Some(RwLock::new(bbits))), content_loader)
+            .await
+            .unwrap();
         let server = handler.server();
         let addr = server.local_addr();
         let core_task = tokio::spawn(async move {
--- a/iroh-gateway/src/bad_bits.rs
+++ b/iroh-gateway/src/bad_bits.rs
@@ -196,9 +196,13 @@ mod tests {
                 tracing::error!("Failed to run gateway rpc handler: {}", err);
             }
         });
-        let handler = crate::core::Core::new(Arc::new(config), Arc::new(Some(RwLock::new(bbits))), content_loader)
-            .await
-            .unwrap();
+        let handler = crate::core::Core::new(
+            Arc::new(config),
+            Arc::new(Some(RwLock::new(bbits))),
+            content_loader,
+        )
+        .await
+        .unwrap();
         let server = handler.server();
         let addr = server.local_addr();
         let core_task = tokio::spawn(async move {
--- a/iroh-gateway/src/rpc.rs
+++ b/iroh-gateway/src/rpc.rs
@@ -1,6 +1,6 @@
 use anyhow::Result;
 use async_trait::async_trait;
-use iroh_rpc_types::gateway::{Gateway as RpcGateway, GatewayServerAddr, VersionResponse};
+use iroh_rpc_types::gateway::{GatewayGateway as RpcGateway, GatewayServerAddr, VersionResponse};
 
 #[derive(Default)]
 pub struct Gateway {}
--- a/iroh-one/src/rpc.rs
+++ b/iroh-one/src/rpc.rs
@@ -1,0 +1,101 @@
+use crate::config::Config;
+use anyhow::Result;
+use async_trait::async_trait;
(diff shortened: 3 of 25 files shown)


fabricedesre:
@dignifiedquire I changed or added a bunch of paste!() modifiers (from :lower to :snake, and :camel) to match better the code generated by `prost`.

I tried to minimize manually written code as much as possible in `iroh-one/src/rpc.rs` since we just relay calls to the store and p2p services that are already instantiated.

iroh-gateway/src/bad_bits.rs
```diff
@@ -191,9 +191,13 @@ mod tests {
 
         let rpc_addr = "grpc://0.0.0.0:0".parse().unwrap();
         let content_loader = RpcClient::new(config.rpc_client.clone()).await.unwrap();
+        tokio::spawn(async move {
+            if let Err(err) = crate::rpc::new(rpc_addr, crate::rpc::Gateway::default()).await {
```
Arqu, line 195: Slightly unsure how this actually communicates with the GW itself. I guess there's no commands that it uses that need to talk to GW yet.

iroh-gateway/src/rpc.rs
```diff
@@ -1,6 +1,6 @@
 use anyhow::Result;
 use async_trait::async_trait;
-use iroh_rpc_types::gateway::{Gateway as RpcGateway, GatewayServerAddr, VersionResponse};
+use iroh_rpc_types::gateway::{GatewayGateway as RpcGateway, GatewayServerAddr, VersionResponse};
```
Arqu, line 3: Can we fix these not to be `GatewayGateway` or `P2pP2p` etc ?

iroh-one/src/rpc.rs
```diff
@@ -0,0 +1,98 @@
...
+    store::{StoreClientAddr, StoreStore},
+};
+use paste::paste;
+use std::mem::transmute;
+
+/// This macro generates the implementation for a service trait in iroh-one. This works
+/// by relaying the calls to the "real" client's backend.
+/// Use of `transmute` here is safe because we know the structs are the same, but protobuf
+/// imports are duplicated in each crate instead of being shared.
+macro_rules! relay {
```
Arqu, line 20: :face_with_spiral_eyes: @dignifiedquire I need you to vet this :D

iroh-rpc-types/proto/gateway_one.proto
```diff
@@ -0,0 +1,40 @@
...
+}
+
+service Store {
+  rpc Put(store.PutRequest) returns (google.protobuf.Empty) {}
+  rpc Get(store.GetRequest) returns (store.GetResponse) {}
+  rpc Has(store.HasRequest) returns (store.HasResponse) {}
+  rpc GetLinks(store.GetLinksRequest) returns(store.GetLinksResponse) {}
+}
+
+service P2p {
```
Arqu, line 22: Is it necessary to copy this all over. Thought the imports should be fine.

Arqu (commented):
It's a bit messy, though a step in the right direction. Would love @dignifiedquire's take on this, especially on the RPC macros part.

iroh-gateway/src/rpc.rs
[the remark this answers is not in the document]
fabricedesre: That's because we now have a Package/Service setup. I know it's weird for the existing cases where the package and the service have the same name.

fabricedesre: Maybe, but it all comes from the protobuf grpc structure.

ramfox:
Hey @fabricedesre! We've been putting in some real thought about how we want to proceed with allowing for communication between an `iroh-one` process and a `iroh` cli, and I wanted to gather feedback.

Rather than having a specific RPC API that opens up only the gateway to outside communication, we would rather keep a single RPC API & refactor the RPC servers to allow for passing RPC messages on multiple types of communication channels.

So for the `iroh-one` case, you would have a gateway, p2p, and store service each with an RPC client that can communicate over mem to each other, and an RPC server that can respond to both mem requests & gRPC requests for interacting with each other & the `iroh` cli.

This would double down on our reliance on the gRPC APIs internally, but would still keep the APIs private.

Question, though: would this setup still suit your use case?

fabricedesre:
I'm not sure I understood everything you wrote @ramfox :)

The current situation in iroh-one is:
- store and p2p are accessible each over a memory rpc channel.
- the gateway uses these memory channels to coordinate the resolver with store & p2p.
- a grpc (tcp or uds) server exposes the gateway service only. Store & p2p are not accessible from the outside.

What this PR does is:
- no change to the internal store and p2p setup. The gateway still uses memory channels to communicate internally.
- the grpc endpoint, instead of just exposing the Gateway service, also exposes a Store and a P2P service (in gateway_one.proto). When a request is received for one of these services, we either answer directly for the Gateway one, or proxy the call over a memory channel if it's a Store or p2p call.

I'm not sure how it would look like from a client point of view. Each service should be usable independently even if they are all served over the same grpc endpoint, but I don't know if the tonic code generation allows that.

fabricedesre:
I favored option 2 because that looked cleaner from the `iroh-one` perspective, and I envisioned the api to be smart enough to know what to do when being configured with a single grpc access point vs. several ones. I don't know if there's an overhead to run a single grpc endpoint with 3 services vs. 3 endpoints with a single service each.

So... feel free to implement your way, that's not a blocker for me. I feel it's more work though.

dignifiedquire:
Given the upcoming changes in RPC we won't be merging this as is, so I am going to close this for now. But we are planning to fix the problem this PR addresses, promised :)

Issue #255, opened 2022-09-30
iroh-one fails to start


rklaehn:
When starting iroh-one, it bails out with an error message: 

```
iroh % cargo run --bin iroh-one                       
   Compiling iroh-bitswap v0.1.0 (/Users/rklaehn/projects_git/iroh/iroh-bitswap)
   Compiling iroh-store v0.1.0 (/Users/rklaehn/projects_git/iroh/iroh-store)
   Compiling iroh-gateway v0.1.0 (/Users/rklaehn/projects_git/iroh/iroh-gateway)
   Compiling iroh-p2p v0.1.0 (/Users/rklaehn/projects_git/iroh/iroh-p2p)
   Compiling iroh-one v0.1.0 (/Users/rklaehn/projects_git/iroh/iroh-one)
    Finished dev [unoptimized + debuginfo] target(s) in 16.12s
     Running `target/debug/iroh-one`
Error: IO error: No such file or directory: While mkdir if missing: No such file or directory
```

Looking into it a bit, it seems the default config that is being used when starting without config is lacking some things. When manually setting the three parameters `config.store.path`, `config.p2p.key_store_path` and `config.gateway.rpc_client.gateway_addr`, it works.

It might be that there is something wrong with how the default args are being constructed.

Tried this both on m1 mac and amd64 linux. Exactly the same behaviour.

Arqu:
Hmmm, so the only required bit should be the store path which you should be able to pass in with `cargo run --bin iroh-one -- --store-path=some_path`. Also think that a recent update (think b5 did it) should have sane defaults so this should have worked out of the box. 
Either case it should at least be documented in a readme or something.

Arqu:
Reopening just to wrap up the readme later.

rklaehn:
Yes, when passing just the store path it is working now.

I think it would be nice to print a nicer error message when it is run without any parameters, but that can be done separately.

```
Error: IO error: No such file or directory: While mkdir if missing: No such file or directory
```

b5:
This will be fixed by #309

PR #256, opened 2022-09-30
fix: properly setup the default config for iroh-one

fabricedesre:
- key_store_path was an empty path, instead of the data root as used by iroh-p2p.
- the gateway_addr was still bound to the uds-gateway feature, which was a mistake.

This fixes https://github.com/n0-computer/iroh/issues/255


Arqu (approved):
👍 ty

ddimaria:
I'm getting the same issue.  The code removed in this PR got added back in this commit: https://github.com/n0-computer/iroh/commit/563aac0bfc42c662b69c418c2f1004d1c5f82e12#diff-56f5dd42bbb949a1b430cfd259f01d6a4a62d4e030af7adba7a4fb08788168f0

Was that intentional?

PR #275, opened 2022-10-03
iroh API and CTL take 2

faassen:
This PR is a new take on the API, the CLI and how to test it. This PR focuses on getting the structures right rather than making sure we actually implement everything we want.

It contains:

## an implementation of the Iroh API

Compared to the previous approach, I used more big traits (@ramfox had the right intuition there). These are mockable using `mockall`. Traits work better than just plain mocking a struct for a variety of reasons I'll go into upon request.

## The CLI layered over the API

The CLI makes an instance of the Iroh API and then uses that, instead of dealing with implementation details itself.

The CLI is still very basic as there are a ton of questions about precise behavior to tackle.

## A testable CLI

 A testing system based on the the trycmd library and the "fixture" feature flag that is only set during development (it can't be just `#[cfg(test)]` by itself as we need to compile a CLI to execute)

You can register mock implementations of the API that can then be enabled for particular tests using an environment variable.

## Note

The p2p peers command it implements is probably not correct and may change entirely, but it helped me tracer bullet through the system.


iroh-ctl/run.rs
```diff
@@ -0,0 +1,27 @@
...
+    let api = Api::new(&client).await?;
+
+    run_cli_command(&api, cli).await
+}
+
+#[cfg(test)]
+async fn fake_cli(cli: Cli) -> Result<()> {
+    let fake = crate::fake::FakeApi::default();
+    run_cli_command(&api, cli).await
+}
```
ramfox, line 27: Is this being used? If so what is it used for?

ramfox (dismissed):
Some questions and documentation suggestions!

iroh-ctl/run.rs
```diff
@@ -0,0 +1,27 @@
+async fn main_cli(cli: Cli) -> Result<()> {
```
dignifiedquire, line 1: why is this not in the src dir?

iroh/src/api.rs
```diff
@@ -0,0 +1,71 @@
...
+    type P: P2p;
+    type S: Store;
+
+    fn p2p(&self) -> Result<Self::P>;
+    fn store(&self) -> Result<Self::S>;
+    async fn get<'a>(&self, ipfs_path: &IpfsPath, output: Option<&'a Path>) -> Result<()>;
+    async fn add(&self, path: &Path, recursive: bool, no_wrap: bool) -> Result<Cid>;
+    async fn check(&self) -> iroh_rpc_client::StatusTable;
+    // This won't work
+    async fn watch(&self) -> Pin<Box<dyn futures::Stream<Item = iroh_rpc_client::StatusTable>>>;
```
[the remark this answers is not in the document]
faassen: Thank you, applied!

iroh-ctl/Cargo.toml
```diff
@@ -23,8 +32,15 @@ iroh-util = { path = "../iroh-util" }
...
+iroh = { path = "../iroh"}
 libp2p = { version = "0.48", default-features = false }
 cid = "0.8.5"
 multiaddr = "0.14.0"
 bytes = "1.1.0"
 async-stream = "0.3.3"
+
+[dev-dependencies]
+trycmd = "0.13.7"
+# this is a workaround based on for enabling features only during test:
```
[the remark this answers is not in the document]
faassen: That language indeed got broken. Fixed!

iroh/src/store.rs
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    async fn block_get(&self, cid: &Cid) -> Result<Option<Bytes>> {
+        self.client.get(*cid).await
+    }
+
+    async fn block_put(&self, _data: &Bytes) -> Result<Cid> {
+        // this awaits ramfox's work in the resolver
+        // would be nice if that work only relied on the store and not
+        // on the full client
```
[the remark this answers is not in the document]
faassen: Thanks for the comments! Is there anything you'd want me to do in this PR? Since we're not aiming to support this in the release I think we can defer this work until after that, correct?

The comment about not relying on the full client was because previously some code involving put relied on the rpc client as a whole instead of just the store bit. I'll just remove that comment though; the 'todo' is enough.

iroh-ctl/run.rs
faassen: This is a stray file I don't know how I managed to add. I'll remove the whole file!

iroh-ctl/run.rs
faassen: I don't know how I managed to do it, but that file isn't supposed to exist. I'll remove it.

faassen:
I've processed the PR comments, thanks everyone! Most embarassing was a completely stray `run.rs` sitting around top level doing nothing. :)

I managed to make `#[cfg_attr]` work for me now. I've added a `mock` feature to `iroh` and the mocks are only built if that feature is set. The `iroh-ctl` dev dependencies then require this feature in `iroh`.

iroh-ctl/src/fixture.rs
```diff
@@ -0,0 +1,44 @@
...
+
+fn fixture_peer_ids() -> MockApi {
+    let mut api = MockApi::default();
+    api.expect_p2p().returning(|| {
+        let mut mock_p2p = MockP2p::default();
+
+        mock_p2p.expect_peer_ids().returning(|| {
+            let peer_id0 = PeerId::from_bytes(&[
+                0, 32, 15, 231, 162, 148, 52, 155, 40, 187, 217, 170, 125, 185, 68, 142, 156, 196,
+                145, 178, 64, 74, 19, 27, 9, 171, 111, 35, 88, 236, 103, 150, 96, 66,
```
dignifiedquire, line 18: Please use the `FromStr` for peer ids, that makes these much more readable

iroh-ctl/src/fixture.rs
```diff
@@ -0,0 +1,44 @@
...
+    registry.insert("peer_ids".to_string(), fixture_peer_ids);
+    registry
+}
+
+pub fn get_fixture_api() -> MockApi {
+    let registry = register_fixtures();
+    let fixture_name = env::var("IROH_CTL_FIXTURE").expect("IROH_CTL_FIXTURE must be set");
+    let fixture = registry
+        .get(&fixture_name)
+        .unwrap_or_else(|| panic!("unknown fixture: {}", fixture_name));
```
dignifiedquire, line 42: if you want to panic on an error, use `exepct`, instead of `unwrap` with `panic`

iroh-ctl/Cargo.toml
```diff
@@ -23,8 +32,16 @@ iroh-util = { path = "../iroh-util" }
...
 multiaddr = "0.14.0"
 bytes = "1.1.0"
 async-stream = "0.3.3"
+
+[dev-dependencies]
+trycmd = "0.13.7"
+iroh = { path = "../iroh", features = ["mock"] }
+# this is a workaround for enabling features only during testing: 
+# https://github.com/DesmondWillowbrook/cargo-optional-features-for-testing-and-examples
+iroh-ctl = { path = ".", features = ["fixture"] }
```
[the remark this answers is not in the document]
faassen: Yes, though I'm a bit worried `testing` is so similar to `test` that it's going to be confusing.

iroh-ctl/src/fixture.rs
faassen: I had that before, but clippy went and suggested this:

https://rust-lang.github.io/rust-clippy/master/#expect_fun_call

iroh-ctl/src/run.rs
```diff
@@ -0,0 +1,140 @@
...
+    )]
+    Get {
+        path: resolver::Path,
+        #[clap(long, short)]
+        output: Option<PathBuf>,
+    },
+}
+
+#[cfg(not(feature = "fixture"))]
+pub async fn run_cli(cli: Cli) -> Result<()> {
```
dignifiedquire, line 72: these methods that take `Cli` should be implemented as 

```rust
impl Cli {
  pub async fn run(self) -> Result<() { .. }
}
```

iroh-ctl/tests/cmd/peer_ids.trycmd
```diff
@@ -0,0 +1,6 @@
+```
+$ iroh-ctl p2p peers
+1AXRDqR8jTkwzGqyu3qknicAC5X578zTMxhAi2brppK2bB
+1Ag5LvC3vwsQicTy1dxNdd1xiNxLUS4Aic4NtdHBqAHD3j
+
```
dignifiedquire, line 5: why is there an empty line?

iroh-ctl/Cargo.toml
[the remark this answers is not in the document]
faassen: I've just renamed things.

iroh-ctl/tests/cmd/peer_ids.trycmd
faassen: I was confused about that too. It's unfortunately a trycmd thing. It requires it and uses that in its own examples too. I think it's so it can separate multiple commands from each other in the same docstrings, but it's definitely ugly.

iroh-ctl/src/fixture.rs
faassen: Yeah, I struggled with this before not realizing there was a trait. I've went and used `parse` now.

iroh-ctl/src/main.rs
```diff
@@ -1,251 +1,14 @@
...
-    }
+async fn main() -> Result<()> {
+    let cli = iroh_ctl::run::Cli::parse();
+    // `run_cli` exists in two versions:
+    // When using the `fixture` feature, the
+    // version of `run_cli` designed for testing purposes using mocked test
+    // fixtures is invoked.
+    // Without the `fixture` feature, the version of
+    // `run_cli` that interacts with the real Iroh API is used.
+    iroh_ctl::run::run_cli(cli).await
```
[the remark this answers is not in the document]
faassen: This has now simply become a method call following you other suggestion.

iroh-ctl/src/run.rs
faassen: I've done that. It suggests further changes in p2p, store, gateway CLI handling code as well to turn functions into methods. I propose we don't do this for now as we'll be cutting away a lot of this code anyway in the near future, and I'll rearrange it then if still necessary.

iroh-ctl/tests/cmd/peer_ids.trycmd
dignifiedquire: I see, thanks

iroh-ctl/src/fixture.rs
dignifiedquire: ah I see, makes sense because of the formatting

iroh-ctl/Cargo.toml
```diff
@@ -2,10 +2,19 @@
...
 license = "Apache-2.0/MIT"
 repository = "https://github.com/n0-computer/iroh"
 description = "Client for interacting with running iroh processes."
+default-run = "iroh-ctl"
+
+[features]
+testing = []
+
+[[bin]]
+bench = false
```
dignifiedquire, line 15: this shouldn't be necessary

iroh-ctl/src/fixture.rs
```diff
@@ -0,0 +1,42 @@
...
+        });
+        Ok(mock_p2p)
+    });
+    api
+}
+
+fn register_fixtures() -> FixtureRegistry {
+    let mut registry = FixtureRegistry::new();
+    registry.insert("peer_ids".to_string(), fixture_peer_ids);
+    registry
```
dignifiedquire, line 32: fyi you can also write

```rust
[("peer_ids".to_string(), fixture_peer_ids)].into_iter().collect()
```

to populate a hashmap

iroh/Cargo.toml
```diff
@@ -6,5 +6,22 @@ license = "Apache-2.0/MIT"
...
+iroh-rpc-client = { path = "../iroh-rpc-client" }
+anyhow = "1"
+async-trait = "0.1.53"
+tokio = { version = "1" }
+bytes = "1.1.0"
+libp2p = "0.49"
+tracing = "0.1.34"
+futures = "0.3.21"
+async-stream = "0.3.3"
+mockall = "0.11.2"
```
dignifiedquire, line 25: shouldn't this be optional and only enabled when the feature `testing` is enabled?

iroh/src/getadd.rs
```diff
@@ -0,0 +1,132 @@
...
+                yield (path, OutType::Dir);
+            } else {
+                let reader = out.pretty(resolver.clone(), Default::default())?;
+                yield (path, OutType::Reader(reader));
+            }
+        }
+    }
+}
+
+// make_output_path adjusts the full path to replace the root with any given output path
```
dignifiedquire, line 73: Please do not repeat the name of the function, please, please, please. Also should be a full sentence with capitalization and use `///` for doc comments

iroh/src/getadd.rs
```diff
@@ -0,0 +1,132 @@
...
+
+use anyhow::{Context, Result};
+use cid::Cid;
+use futures::Stream;
+use futures::StreamExt;
+use iroh_resolver::resolver::Path as IpfsPath;
+use iroh_resolver::{resolver, unixfs_builder};
+use iroh_rpc_client::Client;
+
+pub async fn get(client: &Client, ipfs_path: &IpfsPath, output: Option<&Path>) -> Result<()> {
```
dignifiedquire, line 11: these should all be methods on `Client`

iroh-ctl/Cargo.toml
faassen: `bench=false` or the `[[bin]]` section? I had added the `[[bin]]` section so I could run the right command, but I realize now I did that in the prototype within `iroh`. It'll be necessary again once I move it into `iroh-one`.

I'll drop the `bench = false` bit; that's cargo culted and I copied it from somewhere.

iroh-ctl/src/fixture.rs
faassen: Unfortunately I'm getting

```
32   |         .collect()
     |          ^^^^^^^ value of type `HashMap<std::string::String, fn() -> MockApi>` cannot be built from `std::iter::Iterator<Item=(std::string::String, fn() -> MockApi {fixture_peer_ids})>`
```
I had to add a `as GetFixture` to fix that.

iroh-ctl/Cargo.toml
dignifiedquire: I meant just the bench false

iroh/Cargo.toml
faassen: Good point. I learned another cargo trick with optional packages, hopefully it looks good now.

iroh/src/getadd.rs
faassen: This wasn't code I wrote but I will adjust it.

iroh/src/getadd.rs
faassen: Where you say make these methods of Client, do you mean we should move this code away from `iroh` into the rpc implementation, or do you want me to define a trait to that effect? I didn't write this code originally; I moved it and make it take client references instead of taking ownership, so now they look like methods.

faassen: I could make them methods on `ClientApi` too.

faassen: I've now reshuffled things, so when you're ready please take another look!

iroh-ctl/src/run.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    #[cfg(feature = "testing")]
+    pub async fn run(&self) -> Result<()> {
+        let api = get_fixture_api();
+        self.cli_command(&api).await
+    }
+
+    // extracted this into a public function so that we don't get a lot of
+    // rust analyzer unused code errors, which we do if we inline this code inside
+    // of run
```
dignifiedquire, line 85: That's not how it is supposed to work. Why is it unused? if it is just used in testing, it should be disabled using the feature detection in the regular case

dignifiedquire: also why is this not just inline with the above method?

faassen: I started out having it just inlined.

Rust analyzer declares the version of `run` with the annotation `#[cfg(not(feature = "testing))"`
as unused. This is because the `testing` is enabled in dev mode due to the line in `Cargo.toml` I need to have in order to have `iroh-ctl` compile in testing mode. 

When I use `cargo run --bin iroh-ctl p2p peers` wants to talk to iroh over RPC, so this logic does get picked up.

The version of `run` with `#[cfg(feature = "testing))]"`  is not declared unused by Rust analyzer as we're in development mode in the editor and therefore `testing` is enabled.

I extracted the function `run_impl` as I wanted to have Rust analyzer help me instead of marking it all as unused. Now only the small invocation of `run_impl` is marked as unused. 

Without  `pub` for `run_impl`, the function itself becomes marked unused. This gives me the following build warnings:

```
  --> iroh-ctl/src/run.rs:33:8
   |
33 |     fn make_overrides_map(&self) -> HashMap<String, String> {
   |        ^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: associated function `run_impl` is never used
  --> iroh-ctl/src/run.rs:86:14
   |
86 |     async fn run_impl(&self) -> Result<()> {
```

I've tried to add some comments to explain this better now.

faassen: I considered alternatives. This approach you gave me:

https://github.com/bazelbuild/rules_rust/discussions/913

I don't know how to control `rustc_flags` in Cargo, or how to have a special bin that's ideally called `iroh-ctl` (or `iroh` in the future) that's compiled in `test` mode. See also for the open issue:

https://github.com/rust-lang/cargo/issues/1982

I also looked at:

https://docs.rs/test-binary/latest/test_binary/

This requires a lot of project setup and assumes I build the test binary in the test. I also don't know how to get `trycmd` to use these binaries as they're in sub-projects.

dignifiedquire (dismissed):

dignifiedquire (approved):

PR #266, opened 2022-10-03
update to libp2p@0.49

dignifiedquire:

PR #278, opened 2022-10-03
Revert "update to libp2p@0.49"

dignifiedquire:
Reverts n0-computer/iroh#266

Unfortunately we have to go back due to a regression in rust-libp2p : https://github.com/libp2p/rust-libp2p/issues/2971


Arqu (approved):
:(

PR #279, opened 2022-10-03
Revert "Revert "update to libp2p@0.49""

dignifiedquire:
Reverts n0-computer/iroh#278

Turns out we just needed one more feature


Arqu (approved):
🚀

Issue #177, opened 2022-08-08
when starting check for other instances


Arqu:
UX wise people might experience odd behavior if they try to run multiple instances of `iroh` or already have `kubo` running. Should make sure we are either compatible with that setup or error out gracefully.

b5:
we're hammering out that issue in https://github.com/n0-computer/iroh/issues/274#issuecomment-1267166971

b5:
Also related: https://github.com/n0-computer/iroh/pull/231

PR #293, opened 2022-10-05
feat: provide a mechanism to check if an iroh program is already running

fabricedesre:
This addresses https://github.com/n0-computer/iroh/issues/177 (and https://github.com/n0-computer/iroh/issues/231)

cc @dignifiedquire : I implemented a non opinionated way to acquire a lock and check its state. The example use in iroh-one works as expected so we can generalize to other binaries if the api is fine.

### DIFF
--- a/iroh-util/src/lock.rs
+++ b/iroh-util/src/lock.rs
@@ -1,0 +1,110 @@
+use anyhow::Result;
+use file_guard::{FileGuard, Lock};
+use std::fs::File;
+use std::path::PathBuf;
+use std::rc::Rc;
+
+/// Manages a lock file used to track if an iroh program
+/// is already running.
+/// The lock is released either when the object is dropped
+/// or when the program stops.
+pub struct ProgramLock {
+    path: PathBuf,
+    lock: Option<FileGuard<Rc<File>>>,
+}
+
+impl ProgramLock {
+    /// Create a new lock for the given program. This does not yet acquire the lock.
+    pub fn new(prog_name: &str) -> Result<Self> {
+        let path = crate::iroh_data_path(&format!("{}.lock", prog_name))?;
+        Ok(Self { path, lock: None })
+    }
+
+    /// Check if the current program is locked or not.
+    pub fn is_locked(&self) -> bool {
+        if !self.path.exists() {
+            return false;
+        }
+
+        // Even if we manage to lock the file this won't last since the drop implementation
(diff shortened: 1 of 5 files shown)


iroh-util/src/lock.rs
```diff
@@ -0,0 +1,112 @@
+use anyhow::Result;
+use file_lock::{FileLock, FileOptions};
```
dignifiedquire, line 2: `file_lock` is basically a `C` file, any chance we can find a library that is actually written in rust 😓

dignifiedquire: maybe: https://github.com/brunoczim/fslock it looks a little more cross platform as well

dignifiedquire:
General approach looks good to me, just need to make sure the locking works cross platform in some reasonable manners.

fabricedesre:
> General approach looks good to me, just need to make sure the locking works cross platform in some reasonable manners.

Do we care about windows? I tested on Mac & Linux so far.

fabricedesre:
@dignifiedquire  I switched to https://docs.rs/file-guard/0.1.0/file_guard/ that has the benefits of being cross platform, pure Rust, and to work in our case :) (fslock is not a good fit because its locks are per-handle and not per-process )

iroh-util/src/lock.rs
```diff
@@ -0,0 +1,110 @@
...
+    use super::*;
+
+    fn create_test_lock(name: &str) -> ProgramLock {
+        ProgramLock {
+            path: PathBuf::new().join(name),
+            lock: None,
+        }
+    }
+
+    #[test]
```
dignifiedquire, line 59: This should be probably guarded by `cfg(unix)`, same for the `nix` dependency

dignifiedquire:
> @dignifiedquire I switched to https://docs.rs/file-guard/0.1.0/file_guard/ that has the benefits of being cross platform, pure Rust, and to work in our case :) (fslock is not a good fit because its locks are per-handle and not per-process )

Thanks, looks great.

iroh-util/src/lock.rs
fabricedesre: done in latest commit.

b5 (approved):
We've had a proper review from dig, and I've checked that all comments are addressed. Looks good!

b5 (approved):

PR #294, opened 2022-10-06
fix(unixfs) Fix unixfs roundtrip bug

rklaehn:
There was an issue when the tree was at least 3 levels deep, and there were two empty/expended link lists above the current position. Fixes https://github.com/n0-computer/beetle/issues/174, so the two ignored tests are now enabled.

### DIFF
--- a/iroh-resolver/src/unixfs.rs
+++ b/iroh-resolver/src/unixfs.rs
@@ -522,19 +522,22 @@ fn load_next_node<T: ContentLoader + 'static>(
     loader: T,
 ) -> bool {
     // Load next node
-    if current_links.is_empty() {
-        // no links left we are done
-        return true;
-    }
-    if current_links.last().unwrap().is_empty() {
-        // remove emtpy
-        current_links.pop();
-    }
 
-    let links = current_links.last_mut().unwrap();
-    if links.is_empty() {
-        return true;
-    }
+    // find non empty links
+    let links = loop {
+        if let Some(last_mut) = current_links.last_mut() {
+            if last_mut.is_empty() {
+                // ignore empty links
+                current_links.pop();
+            } else {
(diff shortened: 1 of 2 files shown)


iroh-resolver/src/unixfs.rs
```diff
@@ -522,19 +522,22 @@ fn load_next_node<T: ContentLoader + 'static>(
...
+                current_links.pop();
+            } else {
+                // found non empty links
+                break last_mut;
+            }
+        } else {
+            // no links left we are done
+            return true;
+        }
+    };
```
faassen, line 540: I found this logic difficult to reason about and tried to understand why it's the way it is:

* There's a potentially infinite loop
* if there is no last item (i.e. vec is empty), we return true from the function
* if there is a last item
    * then if it's empty we pop from the current_links vec and proceed to the next iteration
        * the loop should eventually halt when current_links becomes empty   
    * if it's not empty, we break out of the loop and those are going to be the new `links`
  
I'm considering alternatives, but you can't break with a value out of a while or for loop, so that doesn't really help, unless you make it a function so you can `return`, but that's less potentially efficient as it requires an additional check here to determine the early return. Reverse iterating through current_links maybe possible but this code is mutating `links` so also doesn't really apply. So I'm starting to see better why it is the way it is.

rklaehn: I also tried a while let() before, but it wasn't better. It is not an infinite loop since you are emptying the links.

I am always a bit put off by these very mutable and imperative loops you end up writing for async rust state machines. Guess I still am a bit of a FP purist sometimes. But the intention of this was to make the smallest possible fix. We might revisit this at some point.

Thanks for the comments.

dignifiedquire (approved):
nice find

PR #296, opened 2022-10-06
Iroh ctl into iroh one

faassen:
This PR is a mostly mechanical move of what was in `iroh-ctl` into `iroh-one`, and removing `iroh-ctl`.

I've introduced a `start` command in `iroh-one` and moved everything related to the existing `iroh-one` CTL under there. I've then added everything under the existing CLI. `iroh start` doesn't, as far as I know, present enough RPC surface at present to allow future invocations of `iroh` to connect to it, so that's not part of this PR.

Consolidation of stuff appears possible - in particular `config_ctl.rs` and `config_one.rs` appear to have overlaps. I haven't dealt with this in the interest of first completing the move of the code itself.


faassen:
There was a change in direction and we're keeping `iroh-ctl` as a separate crate.

PR #299, opened 2022-10-06
fix: pass links to store when adding data

### LINKED ISSUE #287: Make sure links get persisted in the database
Currently, links are not being persisted in the store. Even when adding an object like an unixfs branch node, there are 0 links being passed via RPC.

Make sure links are properly passed via RPC so the store has the information it needs to traverse the DAG for caching/gc.

rklaehn:
Note that this may lead to decreased performance, since we are doing more work than before. But the store needs this information.

The way links are parsed can be optimized, see #297 . And I am not quite sure if finding links should not be done on the store side. But that is another discussion to be had.

Fixes #287

### DIFF
--- a/iroh-resolver/src/unixfs_builder.rs
+++ b/iroh-resolver/src/unixfs_builder.rs
@@ -675,9 +675,12 @@ mod tests {
         stream: impl Stream<Item = Result<Block>>,
     ) -> Result<(Cid, Resolver<Arc<fnv::FnvHashMap<Cid, Bytes>>>)> {
         tokio::pin!(stream);
-        let items: Vec<_> = stream.try_collect().await?;
-        let root_block = items.last().context("no root")?.clone();
-        let store: fnv::FnvHashMap<Cid, Bytes> = items
+        let blocks: Vec<_> = stream.try_collect().await?;
+        for block in &blocks {
+            block.validate()?;
+        }
+        let root_block = blocks.last().context("no root")?.clone();
+        let store: fnv::FnvHashMap<Cid, Bytes> = blocks
             .into_iter()
             .map(|block| {
                 let (cid, bytes, _) = block.into_parts();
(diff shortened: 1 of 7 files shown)


iroh-resolver/src/unixfs_builder.rs
```diff
@@ -458,7 +459,8 @@ pub async fn add_file<S: Store>(store: Option<S>, path: &Path, wrap: bool) -> Re
     while let Some(part) = parts.next().await {
         let (cid, bytes) = part?;
         if let Some(ref store) = store {
-            store.put(cid, bytes, vec![]).await?;
+            let links = parse_links(&cid, &bytes)?;
+            store.put(cid, bytes, links).await?;
```
dignifiedquire, line 463: we already know the links, when we encode things. instead of parsing them out, we should expose them from the underlying generator and use them here

rklaehn: OK

rklaehn: Digging into this now. This will be a bit of a bigger operation then... Got lots of places where the links need to be funneled though....

rklaehn:
I am now passing through all the links, so at all places where there was a (Cid, Bytes) before there is a (Cid, Bytes, Vec<Cid>> now. But I don't like it very much. It is a bit verbose and error prone. So will try making a struct and see how that goes...

iroh-resolver/src/unixfs_builder.rs
```diff
@@ -671,14 +672,20 @@ mod tests {
...
     ) -> Result<(Cid, Resolver<Arc<fnv::FnvHashMap<Cid, Bytes>>>)> {
         tokio::pin!(stream);
         let items: Vec<_> = stream.try_collect().await?;
-        let (root, _) = items.last().context("no root")?.clone();
-        let store: fnv::FnvHashMap<Cid, Bytes> = items.into_iter().collect();
+        let root_block = items.last().context("no root")?.clone();
+        let store: fnv::FnvHashMap<Cid, Bytes> = items
+            .into_iter()
+            .map(|block| {
+                let (cid, bytes, _) = block.into_parts();
```
dignifiedquire, line 683: Probably should have a test that checks that the links are actually stored in the store

rklaehn: Will do. Just want to see how I like the Block struct. I think it is good...

dignifiedquire:
Lets get this in and make a new issue for adding adding tests

dignifiedquire (dismissed):

rklaehn:
Made a followup issue for the proposed tests: #315

dignifiedquire (approved):
nice catch

PR #300, opened 2022-10-06
feat: add a size method for Directory

faassen:
Adds a `size` method so we can get the size in bytes of a `Directory`.

We can use this for the progress bar in future code, as we need the total amount of bytes there. We can then use the length of the bytes for each chunk to update the progress bar.

There's one big question: what to do for reader based files? When do they occur in the real world when we use add or the API? Right now they have a file size of 0, as the alternative is to actually consume the reader to determine the size. Or is there some way to require another trait so we know the file size of the async read?

### DIFF
--- a/iroh-resolver/src/unixfs_builder.rs
+++ b/iroh-resolver/src/unixfs_builder.rs
@@ -56,6 +56,11 @@ impl Directory {
         }
     }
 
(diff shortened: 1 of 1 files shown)


ramfox:
let's close this for now and tackle this at a later date when we have a better answer for getting a size for readers

PR #304, opened 2022-10-07
Gateway: Fetch recursive DAG as CAR

Frando:
Hi!
I was looking on how to use iroh from other codebases, and a bit of a precursor to a writable gateway, this PR adds functionality to fetch a DAG recursively into a CAR file. This was already stubbed in the gateway but not implemented.

The implementation works similar to the existing recursive loaders. However to save the cost of reserializing, I added a new `OutRaw` struct in the resolver for this (because the `Out` struct in the resolver currently only keeps the deserialized representation for UnixFS nodes). Currently, the nodes are still fully deserialized anyways for link scraping but that would be fixed with #297.

What's still missing is, I think, a configurable upper limit on DAG sizes that can be requested from the Gateway to avoid DoS attacks. Also misses tests, I can add some if the PR seems on the right track to you.

### DIFF
--- a/iroh-resolver/src/resolver.rs
+++ b/iroh-resolver/src/resolver.rs
@@ -155,6 +155,42 @@ impl FromStr for Path {
 }
 
 #[derive(Debug, Clone)]
+pub struct OutRaw {
+    metadata: Metadata,
+    content: Bytes,
+    cid: Cid,
+}
+
+impl OutRaw {
+    pub fn from_loaded(cid: Cid, loaded: LoadedCid) -> Self {
+        let metadata = Metadata {
+            path: Path::from_cid(cid),
+            size: None,
+            typ: OutType::Raw,
+            unixfs_type: None,
+            source: loaded.source,
+            resolved_path: vec![cid],
+        };
+        Self {
+            metadata,
+            content: loaded.data,
+            cid,
+        }
+    }
+    pub fn metadata(&self) -> &Metadata {
+        &self.metadata
+    }
+
+    pub fn cid(&self) -> &Cid {
+        &self.cid
+    }
+
+    pub fn content(&self) -> &Bytes {
+        &self.content
+    }
+}
+
+#[derive(Debug, Clone)]
 pub struct Out {
     metadata: Metadata,
     pub(crate) content: OutContent,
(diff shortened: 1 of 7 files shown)


iroh-resolver/src/resolver.rs
```diff
@@ -581,6 +617,42 @@ impl<T: ContentLoader> Resolver<T> {
         }
     }
 
+    /// Resolve a path recursively and yield the raw bytes plus metadata.
+    #[tracing::instrument(skip(self))]
+    pub fn resolve_recursive_raw(&self, root: Path) -> impl Stream<Item = Result<OutRaw>> {
```
dignifiedquire, line 622: For security reasons (very easy to DOS a gateway with this currently), could you add a recursion limit please, maybe just as a parameter to the function and then passing a reasonable constant from the gateway?

iroh-resolver/src/resolver.rs
```diff
@@ -581,6 +617,42 @@ impl<T: ContentLoader> Resolver<T> {
...
+        let mut cids = VecDeque::new();
+        let this = self.clone();
+        async_stream::try_stream! {
+            let root_block = this.resolve_root(&root).await.map(|(cid, loaded)| OutRaw::from_loaded(cid, loaded));
```
dignifiedquire, line 626: I think this is the same code as `resolve_recursive` effectively, other than mapping things differently, maybe this could be abstracted by passing a closer to a shared function?

Arqu (commented):
Gateway code looks good to me. Once digs comments are resolved, I'm happy to have this merged.

iroh-resolver/src/resolver.rs
Frando: I added a simple recursion limit in the latest commit.

iroh-resolver/src/resolver.rs
Frando: Sounds good. I implemented an abstraction to be shared between the two `resolve_recursive` functions.  `resolve_recursive_with_path` doesn't yet use it because there's additional logic around keeping the path. Could add it to the generic one, if always maintaining the path isn't considered too expensive (don't think it would matter).

Frando:
I added a test as well. @Arqu because there's not any integration tests in the gateway yet maybe review if the testing approach is correct.

iroh-resolver/src/resolver.rs
Frando: @dignifiedquire I've looked into having the remaining `resolve_recursive_with_path` also use the `resolve_recursive_mapped` function from the latest commit in this PR. I made it work but with the cost of a couple of additional clones of the path even if it's not being used. See this commit (not on this PR): https://github.com/Frando/iroh/commit/91c04401c27ea593cd7cdd24a44def2186f7ce0d
I wasn't sure if it's worth it.

Arqu:
Lovely, thank you. The testing story is a bit grim currently besides some smaller bits. There's a pending issue I have to tackle to get tests going but will have to wait a little bit longer.

Issue #282, opened 2022-10-04
Restrict commands we don't support yet in the CLI


faassen:
We have a bunch of half-baked commands that don't work fully yet and that we won't support yet in the first release. I think the simplest way to go about this is to remove them for now, given that they're easy to add back when we get to it.

b5:
Fully agreed. Let's remove anything that isn't intended to be polished up for a release. We have git history if we'd like to resurrect

Issue #310, opened 2022-10-10
cli booleans avoid negatives


faassen:
b5 wrote on #270 

> Instead of --no-wrap I'd prefer --wrap, which defaults to true. 
> IMHO using negation in boolean flag labels is confusing:
> --wrap=false is clearer than --no-wrap

I agree and would like to make it work this way, but Clap doesn't do this out of the box, so I've created a separate issue to figure this out.

What works is this (where the flag defaults to true):

```rust
/// Track metrics
#[clap(long, action = clap::ArgAction::Set, default_value_t=true)]
metrics: bool
```

Note that this does *not* support `--metrics` by itself, only with explicit `=true` and `=false`. I think that's acceptable for flags that default to `true`.

Documentation generated by Clap seems reasonable:

```
     --metrics <METRICS>  Track metrics [default: true] [possible values: true, false]
```

For flags that default to `false` we can use Clap's normal flag mechanism:

```rust
/// Verbose mode
#[clap(long, short = 'v')]
verbose: bool,
```

Documentation is like this:

```
 --verbose            Verbose mode
```

Note that this is a true Clap flag: it's only used to enable something, and you can't use an explicit `true` or `false`.

I wanted to highlight this difference, though I think it's acceptable. If we want to stick with true Clap flags for the "flag is turned on by default" use case, we  *have* to take the `--no-` flags route instead.

## Relevant Clap issues & PRs

The general explicit flag story:

https://github.com/clap-rs/clap/issues/1649

Automatic support for `--no` flags. Isn't implemented.

https://github.com/clap-rs/clap/issues/815

This misdirected me for a bit but has to do with environment variable parsing, so it not related:

https://github.com/clap-rs/clap/pull/2664

b5:
hmmm, now that I think about it, I'm not a massive fan of mixing "true clap flags" and "boolean clap flags". I'd prefer only "true clap flags", and you're saying the only way to do that is to go the `--no` route?

faassen:
You either have a boolean option, ehich can default to true, and they always need a value, and have no single letter flag (as value is required)

Or clap flag which takes no value, defaults to false, and can have a single letter version.

faassen:
I've gone for explicit `--no-wrap` as opposed to `--wrap=false`, with no single letter shortcut, because:

* you're not a massive fan of mixing

* true clap flags are easier to express with Clap

* `--wrap=true` would never be used anyway, so we really only need one flag `--no-wrap`

[...]

Issue #267, opened 2022-10-03
iroh --version


faassen:
Have an `iroh version` command that does something sensible.

Currently it reports the version of the `iroh-ctl` crate. It makes sense that we report on this *somewhere* given that it can be useful information debugging, but it's not quite what `ipfs version` seems to do.

`ipfs version` shows the version of the kubo release itself. We could show the version of the iroh release. Should we base this off some Rust crate? Which one?

`ipfs version` also can be given a flag to get the commit hash, though in my case it just displays the release version, not a commit hash.

`ipfs version` also can be given a flag to get the repo version. What is a "repo version" and do we want that?

It supports an `all` option which does this:

```
$ ipfs version --all
Kubo version: 0.15.0
Repo version: 12
System version: amd64/linux
Golang version: go1.18.5
```

It reveals a system version and golang version too.

Then there's a `deps` subcommand which shows all the go dependencies, which I suggest we definitely *not* implement in Iroh.

A difference we have is that we have `iroh cloud` and `iroh one` and may want to see this too.

Then there's some overlap with `iroh id`, which also reports on version information:

```
	"AgentVersion": "kubo/0.15.0/desktop",
	"ProtocolVersion": "ipfs/0.1.0",
	"Protocols": [
		"/ipfs/bitswap",
		"/ipfs/bitswap/1.0.0",
		"/ipfs/bitswap/1.1.0",
		"/ipfs/bitswap/1.2.0",
		"/ipfs/id/1.0.0",
		"/ipfs/id/push/1.0.0",
		"/ipfs/lan/kad/1.0.0",
		"/ipfs/ping/1.0.0",
		"/libp2p/autonat/1.0.0",
		"/libp2p/circuit/relay/0.1.0",
		"/libp2p/circuit/relay/0.2.0/stop",
		"/libp2p/dcutr",
		"/p2p/id/delta/1.0.0",
		"/x/"
	]
```

Is there really a need for a top-level version command? Would it make sense to roll all of this into a `ipfs id` command instead?

faassen:
Conclusion from the meeting: we have a `-v` or `--version` flag which reports the version of the binary. For iroh-one this is the version of iroh-one. For a future `iroh-ctl` command we report that.

faassen:
I realized that `-v` is often also used for `--verbose`. Do we anticipate a verbose mode in the future? @dignifiedquire wants to avoid the situation where `-v` is verbose but `--version` is version. @b5, what do you think?

[...]

PR #311, opened 2022-10-10
CLI and API starting point

faassen:
The CLI and API as proposed only has a few commands, and the original CLI and API had a lot more. In this PR I've stripped out all parts of the API and CLI we're not using yet, to have a clean starting point.

I have changed the `no-metrics` flag to the `metrics` option, as discussed in #310 

The individual commands don't conform to the proposed new designs yet. I also haven't implemented any `start` or `stop` command. But before I make more changes, this is a good point to do a PR, as it's still simple to review what's going on.

The CLI test that exercises a fixture has been changed to be `iroh p2p lookup` as well, though it doesn't yet define the correct answer.

This fixes https://github.com/n0-computer/iroh/issues/282

It implements https://github.com/n0-computer/iroh/issues/267 too, though by just following what Clap does by default anyway (see the comment about the `iroh-ctl` name we get back though)


iroh-ctl/Cargo.toml
```diff
@@ -6,14 +6,14 @@ authors = ["Kasey Huizinga <klhuizinga@gmail.com>", "Martijn Faassen <faassen@st
...
 description = "Client for interacting with running iroh processes."
-default-run = "iroh-ctl"
+default-run = "iroh"
 
 [features]
 testing = []
 
 [[bin]]
 path = "src/main.rs"
-name = "iroh-ctl"
```
dignifiedquire, line 16: lets skip this in this PR, or do the full renaming please

faassen: if I promise I will do the full renaming can I keep the name = "iroh" (for the binary) in there for now? Otherwise I have to move my trycmd test into the wrong direction back to iroh-ctl, and I already wrote a few more on followup work

faassen: Example of what I'll need to change back:

```
$ iroh get /ipfs/QmP8jTG1m9GSDJLCbeWhVSVgEzCPPwXRdCRuJtQ5Tz9Kc9
Saving file(s) to QmP8jTG1m9GSDJLCbeWhVSVgEzCPPwXRdCRuJtQ5Tz9Kc9

```

Note that trycmd has an extremely annoying behavior of silently skipping a trycmd test if the binary is missing. :(

faassen: (discussed on chat, it's okay for now)

dignifiedquire (approved):

PR #316, opened 2022-10-10
CLI improvements: more in API, iroh-api rename, iroh get basics

faassen:
This PR contains two changes:

## move RPC client construction into `iroh` (read: `iroh-api`)

I've moved creation of the RpcClient away from the CLI into `iroh` (to be `iroh-api`). This means that the API gains a bunch of dependencies and the CLI loses them.

Doing this made me realize that the api should own the rpc client, so changed from a reference (with a lifetime) to ownership.

## `iroh get` start of implementation

Start of implementation of `iroh get` with a few mocked CLI tests

## `iroh` -> `iroh-api` crate

The plan is to have an `iroh` crate that contains the CLI and an `iroh-api` crate that contains the Rust API. In this PR I've done half of that renaming: `iroh` is renamed to `iroh-api`. I haven't done the second half, renaming `iroh-ctl` to `iroh`, as that makes this PR far too difficult to read.

### DIFF
--- a/iroh-p2p/src/behaviour.rs
+++ b/iroh-p2p/src/behaviour.rs
@@ -10,7 +10,7 @@ use libp2p::core::identity::Keypair;
 use libp2p::core::PeerId;
 use libp2p::gossipsub::{Gossipsub, GossipsubConfig, MessageAuthenticity};
 use libp2p::identify::{Identify, IdentifyConfig};
-use libp2p::kad::store::MemoryStore;
+use libp2p::kad::store::{MemoryStore, MemoryStoreConfig};
 use libp2p::kad::{Kademlia, KademliaConfig};
 use libp2p::mdns::TokioMdns as Mdns;
 use libp2p::multiaddr::Protocol;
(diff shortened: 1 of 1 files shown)


iroh-ctl/tests/cmd/get.trycmd
```diff
@@ -0,0 +1,16 @@
+We can get a CID:
```
rklaehn, line 1: Nice test.

iroh-api/src/api.rs
```diff
@@ -37,23 +40,46 @@ pub trait Api {
...
 }
 
-impl<'a> Iroh<'a> {
-    pub fn new(client: &'a Client) -> Self {
+impl Iroh {
+    pub async fn new(
+        config_path: Option<&Path>,
+        overrides_map: HashMap<String, String>,
+    ) -> Result<Self> {
+        let cfg_path = iroh_config_path(CONFIG_FILE_NAME)?;
```
rklaehn, line 48: This is not new code, but just moved around, correct?

rklaehn (dismissed):

iroh-api/src/api.rs
faassen: Correct, from iroh-ctl

iroh-api/src/p2p.rs
```diff
@@ -34,7 +34,9 @@ pub trait P2p: Sync {
 }
 
 #[async_trait]
-impl<'a> P2p for ClientP2p<'a> {
+impl P2p for ClientP2p {
+    /// XXX really should be an API that intos a peer id, and then also accepts
```
ramfox, line 38: Can we use this format for todos:
```rust
// TODO(faassen): should be an API that...
```

That way we can all be uniform & know who to ask for clarity if someone else is working on implementing that todo.

I'm going to add this preference to the contributor/onboarding documentation

ramfox: You don't need to change it for this PR, just going forward.

ramfox (approved):

iroh-api/src/p2p.rs
faassen: :+1:

PR #317, opened 2022-10-10
fix(iroh-one): use default store path when no --store-path flag present

### LINKED ISSUE #255: iroh-one fails to start

### LINKED ISSUE #309: Default store-path for `iroh start`
Currently `--store-path` is required to start `iroh`. The path given is created in the context of the current working directory.

b5:

> We should pick a default store path, creating it automatically if it doesn't exist. Default path should be inside a store directory in the 
> [application files location](https://iroh.computer/docs/data-locations/), so $HOME/.local/share/iroh/store on linux.

b5:
use store exported path in iroh-one.

closes #255, closes #309. cc @fabricedesre

### DIFF
--- a/iroh-one/src/config.rs
+++ b/iroh-one/src/config.rs
@@ -6,6 +6,7 @@ use iroh_metrics::config::Config as MetricsConfig;
 use iroh_p2p::Libp2pConfig;
 use iroh_rpc_client::Config as RpcClientConfig;
 use iroh_rpc_types::Addr;
+use iroh_store::config::config_data_path;
 use iroh_util::insert_into_config_map;
 use serde::{Deserialize, Serialize};
 use std::path::PathBuf;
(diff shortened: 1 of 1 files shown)


Arqu (approved):
👍 👍

Issue #277, opened 2022-10-03
iroh-gateway spec compatibility


Arqu:
Originally we took the `go-ipfs` (now `kubo`) implementation of the gateway as the baseline of what's expected from a gateway. Since then [this spec](https://github.com/ipfs/specs/blob/main/http-gateways/PATH_GATEWAY.md#path-gateway-specification) came out and we should aim to comply with it as much as possible (which will also lead us to sand down some of our rough edges). 

From the spec so far we're missing:
- [x] `HEAD` support
- [x] Proper ETag and Cache control handling
- [x] Support for `Range` header/reqs
- [x] Review `Accept-Headers`
- [x] Double check `Service-Worker` handling
- [x] Align/simplify request url `?params`
- [x] Audit response headers
- [x] Adjust responses and response codes to be in line with spec

Issues we already have that cross over into it or are appropriate improvements:
- [x] https://github.com/n0-computer/iroh/issues/230
- [x] https://github.com/n0-computer/iroh/issues/103
- [x] https://github.com/n0-computer/iroh/issues/96
- [x] https://github.com/n0-computer/iroh/issues/197
- [x] https://github.com/n0-computer/iroh/issues/98
- [x] https://github.com/n0-computer/iroh/issues/130
- [x] https://github.com/n0-computer/iroh/issues/92
- [ ] https://github.com/n0-computer/iroh/issues/97
- [x] https://github.com/n0-computer/iroh/issues/89

Punting due to other priorities:
- [x] https://github.com/n0-computer/iroh/issues/93
- [x] https://github.com/n0-computer/iroh/issues/99

Potential extensions:
- https://github.com/ipfs/specs/pull/288

Arqu:
range reqs https://github.com/n0-computer/iroh/pull/330

Arqu:
Missing 404 handling / templates for completion.

Arqu:
closing in favor of https://github.com/n0-computer/iroh/issues/465

Issue #230, opened 2022-09-07
Clarify expected cache headers


fabricedesre:
From discord:

```
[10:46 AM] fabrice:
why is "no-cache" used for the default cache header value? 
[10:59 AM] Arqu:
Lots of those bits are still not fully fleshed out
[10:59 AM] Arqu:
Technically we cache everything if the store is present
[11:00 AM] Arqu:
But we dont handle the etags and cache headers fully so it doesnt make sense we advertise anything else
[11:00 AM] Arqu:
The blessing with content addressed storage is you can cache a cid indefinitely technically as the contents wont change
[11:01 AM] Arqu:
And you never cache ipns type reqs
```

I'm not sure if it's worth exposing cache headers as configurable. As @Arqu says, behavior needs to different between `/ipfs` and `/ipns`. For `raw` and `car` response, we should return `Cache-Control: immutable`.

Thoughts?

fabricedesre:
The IPFS gateway spec section for the `Cache-Control` response header is at https://github.com/ipfs/specs/blob/main/http-gateways/PATH_GATEWAY.md#cache-control-response-header

There are no differences in https://github.com/ipfs/specs/blob/main/http-gateways/TRUSTLESS_GATEWAY.md

Arqu:
Spec is already being followed, in the next update the misleading flags will be removed. The only exception is that for the IPNS based `Last-Modified` date we follow what kubo does instead of taking the spec recommendation to either do dns ttl based dates or not at all.

Issue #197, opened 2022-08-18
better mime type detection


Arqu:
Currently we do very simple extension/path based mime type detection. Usually that is not enough and we rely on clients and browsers to figure it out on their own.
We should do deeper inspection of the data to try and guess the correct mime type.
Kubo reference: https://github.com/ipfs/kubo/blob/9c6094e651bd30a11ee7a501bf9aa7eb5cc34deb/core/corehttp/gateway_handler_unixfs_file.go#L48

fabricedesre:
I don't know how good it is, but https://docs.rs/mime_classifier/0.0.1/mime_classifier/ could be used for that.

Issue #103, opened 2022-05-31
gateway(paths): use resolver::Path instead of stringified version where possible


Arqu:
maybe store the `Path` directly instead of the stringified version?

_Originally posted by @dignifiedquire in https://github.com/n0-computer/iroh/pull/87#discussion_r886107177_

PR #319, opened 2022-10-11
gateway: spec parity

Arqu:
Tackles https://github.com/n0-computer/iroh/issues/277

Closes: 
- https://github.com/n0-computer/iroh/issues/230
- https://github.com/n0-computer/iroh/issues/103
- https://github.com/n0-computer/iroh/issues/197
- https://github.com/n0-computer/iroh/issues/98
- https://github.com/n0-computer/iroh/issues/89

Various other QoL fixes, most of the `ETag` support (missing dir etags), `HEAD` request handling (missing a small bit to be done), ensures headers, params etc are in line with spec.

Remainder of work will continue to be tracked on the tracker issue.

### DIFF
--- a/iroh-gateway/src/client.rs
+++ b/iroh-gateway/src/client.rs
@@ -16,6 +16,7 @@ use iroh_resolver::resolver::{
     CidOrDomain, ContentLoader, Metadata, Out, OutMetrics, OutPrettyReader, OutType, Resolver,
     Source,
 };
+use mime::Mime;
 use tokio::io::{AsyncBufReadExt, AsyncReadExt, AsyncWrite};
 use tokio_util::io::ReaderStream;
 use tracing::{info, warn};
@@ -31,7 +32,7 @@ pub struct Client<T: ContentLoader> {
 pub struct PrettyStreamBody<T: ContentLoader>(
     ReaderStream<tokio::io::BufReader<OutPrettyReader<T>>>,
     Option<u64>,
-    Vec<u8>,
+    Option<Mime>,
 );
 
 #[allow(clippy::large_enum_variant)]
@@ -42,7 +43,7 @@ pub enum FileResult<T: ContentLoader> {
 }
 
 impl<T: ContentLoader> PrettyStreamBody<T> {
-    pub fn get_sample(&self) -> Vec<u8> {
+    pub fn get_mime(&self) -> Option<Mime> {
         self.2.clone()
     }
 }
@@ -111,14 +112,11 @@ impl<T: ContentLoader + std::marker::Unpin> Client<T> {
                 .map_err(|e| e.to_string())?;
 
             let mut buf_reader = tokio::io::BufReader::with_capacity(1024 * 1024, reader);
-            let body_sample = buf_reader
-                .fill_buf()
-                .await
-                .map_err(|e| e.to_string())?
-                .to_vec();
+            let body_sample = buf_reader.fill_buf().await.map_err(|e| e.to_string())?;
+            let mime = sniff_content_type(body_sample);
             let stream = ReaderStream::new(buf_reader);
 
-            let body = PrettyStreamBody(stream, metadata.size, body_sample);
+            let body = PrettyStreamBody(stream, metadata.size, Some(mime));
 
             if metadata.typ == OutType::Raw {
                 return Ok((FileResult::Raw(body), metadata));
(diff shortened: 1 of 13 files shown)


iroh-gateway/src/handlers.rs
```diff
@@ -156,6 +156,9 @@ pub async fn get_handler<T: ContentLoader + std::marker::Unpin>(
         .map_err(|e| error(StatusCode::BAD_REQUEST, &e, &state))?;
     let resolved_cid = resolved_path.root();
 
+    let return_if_head = handle_only_if_cached(&request_headers, &state, &resolved_cid).await?;
```
dignifiedquire: this is never used?

iroh-resolver/src/resolver.rs
```diff
@@ -62,15 +62,25 @@ impl Path {
         &self.tail
     }
 
+    pub fn is_dir(&self) -> bool {
+        self.tail.len() > 0 && self.tail.last().unwrap().eq("")
```
dignifiedquire: That's not guranteed to be true, that's at best and indication that this is a dir

iroh-gateway/src/headers.rs
```diff
@@ -10,16 +10,34 @@ pub fn add_user_headers(headers: &mut HeaderMap, user_headers: HeaderMap) {
...
+    let mut content_type = String::new();
+    if let Some(ct) = guess.first() {
+        content_type = ct.to_string();
+    } else {
         // todo(arqu): deeper content type checking
-        // todo(arqu): if mime type starts with text/html; strip encoding to let browser detect
-        headers.insert(
-            CONTENT_TYPE,
-            HeaderValue::from_str(content_type.as_ref()).unwrap(),
+        let classifier = mime_classifier::MimeClassifier::new();
```
dignifiedquire: could you store & reuse this classifier?

iroh-gateway/src/headers.rs
```diff
@@ -10,16 +10,34 @@ pub fn add_user_headers(headers: &mut HeaderMap, user_headers: HeaderMap) {
...
+        let computed_type = classifier.classify(
+            context,
+            no_sniff_flag,
+            apache_bug_flag,
+            &supplied_type,
+            body_sample,
         );
     }
+
+    if content_type.starts_with("text/html") {
```
dignifiedquire, line 26: what case is this?

iroh-gateway/src/client.rs
```diff
@@ -98,8 +108,15 @@ impl<T: ContentLoader + std::marker::Unpin> Client<T> {
...
                 .map_err(|e| e.to_string())?;
 
-            let stream = ReaderStream::new(reader);
-            let body = PrettyStreamBody(stream, metadata.size);
+            let mut buf_reader = tokio::io::BufReader::with_capacity(1024 * 1024, reader);
+            let body_sample = buf_reader
+                .fill_buf()
+                .await
+                .map_err(|e| e.to_string())?
+                .to_vec();
```
dignifiedquire, line 116: Why not avoid the storing in a `Vec`, by doing the mime check  here, and store the result of it

iroh-resolver/src/resolver.rs
Arqu: Technically it's intended use is purely string/browser path based manipulation, so it is a bit misleading, especially in the resolver crate. I don't se a sensible way to have info from the resolver if its a UnixFS::Dir before we get to the actual call we want to make when just parsing the path. 
The benefit is that this is only called/used when `FileResult::Directory` is returned by the client. I could potentially rename it or just separate out this logic into the handlers not to let others lean on it?

iroh-gateway/src/response.rs
```diff
@@ -96,37 +89,25 @@ pub fn get_response_format(
...
-                return Err("invalid format".to_string());
-            }
+    }
+
+    match ResponseFormat::try_from_headers(request_headers) {
+        Ok(format) => {
+            return Ok(format);
+        }
+        Err(_) => {
+            return Ok(ResponseFormat::Fs(String::new()));
```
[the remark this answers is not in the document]
Arqu: Thought I was hallucinating. Couldn't find the code at all :D Guess you reviewed commit by commit. Clippy would have yelled at me anyways :D

iroh-gateway/src/headers.rs
Arqu: It's neither large nor stateful really, we save nothing and I have to bother with a fork to add clone to it. Also having a global one would mean we have one of these cloned/instantiated per req. This way it's only when the req needs it.

iroh-gateway/src/client.rs
Arqu: and here I am shuffling 1MB around for no reason 👍

iroh-resolver/src/resolver.rs
dignifiedquire: maybe call it has_trailing_slash? then it is up to the caller to give it meaning

PR #320, opened 2022-10-11
fix(p2p): Increase the size of the kad mem store

rklaehn:
This allows adding large files without getting an error. It is a stopgap until we have a persistent kad store.

Before:
```
❯ time cargo run --release -p iroh-ctl -- add --no-wrap test1g  
    Finished release [optimized] target(s) in 0.17s
     Running `target/release/iroh add --no-wrap test1g`
Error: status: Internal, message: "the store cannot contain any more provider records", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Tue, 11 Oct 2022 10:39:21 GMT", "content-length": "0"} }
cargo run --release -p iroh-ctl -- add --no-wrap test1g  0.71s user 0.33s system 7% cpu 13.072 total
```

After:
```
❯ time cargo run --release -p iroh-ctl -- add --no-wrap test1g
    Finished release [optimized] target(s) in 0.18s
     Running `target/release/iroh add --no-wrap test1g`
/ipfs/bafybeiepocidrgivy3uqfh54klmtmfjolx7mceouvt2cggrpkghwsdxzni
cargo run --release -p iroh-ctl -- add --no-wrap test1g  4.75s user 2.29s system 4% cpu 2:53.84 total
```

### DIFF
--- a/iroh-p2p/src/behaviour.rs
+++ b/iroh-p2p/src/behaviour.rs
@@ -10,7 +10,7 @@ use libp2p::core::identity::Keypair;
 use libp2p::core::PeerId;
 use libp2p::gossipsub::{Gossipsub, GossipsubConfig, MessageAuthenticity};
 use libp2p::identify::{Identify, IdentifyConfig};
-use libp2p::kad::store::MemoryStore;
+use libp2p::kad::store::{MemoryStore, MemoryStoreConfig};
 use libp2p::kad::{Kademlia, KademliaConfig};
 use libp2p::mdns::TokioMdns as Mdns;
 use libp2p::multiaddr::Protocol;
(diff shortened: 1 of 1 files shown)


dignifiedquire:
this will still fail if you do it many times asfaik. so this is fine to improve things for the moment but we very much need to store provider records on disk to avoid this limitation entirely

rklaehn:
> this will still fail if you do it many times asfaik. so this is fine to improve things for the moment but we very much need to store provider records on disk to avoid this limitation entirely

Yes, that would be the very next thing I would do. Just wanted to do something quickly so you can at least add a 1gb file without error.

dignifiedquire (approved):

PR #328, opened 2022-10-12
initial directory listing support

### LINKED ISSUE #92: gateway(unixfs): improve dir listing
Dir listing is very plain. We should list size where possible, breadcrumbs to parent, potentially resolve paths to list items.
Improve overall template styling, add execution time.

b5:
- now with actual HTML
- breadcrumb headers
- icons for 64 file extensions (matching kubo)
- human readable file sizes (via a new `iroh-util` func, cc @faassen for CLI usage)
- sizes are in SI units (MiB, KiB, GiB), too nerdy?
- all data prep handled on the rust side, with no helper functions in template code
- use `include_str!` for easier asset maintenance

re-interprets the kubo template with iroh colors (I don't think we should be screwing with end-user expectations _too_ much for this first release), but help links point to iroh to avoid bombing kubo maintainers with iroh issues:

Screen_Shot_2022-10-12_at_4 24 08_PM

punted until later:
- root directory size
- load time indicator
- proper tests to ensure directory template doesn't break everything

closes #92


iroh-gateway/src/templates.rs
```diff
@@ -17,3 +89,616 @@ pub const NOT_FOUND: &str = "
...
+
+.table-responsive {
+	display:block;
+	width:100%;
+	overflow-x:auto;
+	-webkit-overflow-scrolling:touch
+}
+"##;
+
+pub const ICONS_STLESHEET_TEMPLATE: &str = r##"
```
fabricedesre, line 300: nit: typo, should be ICONS_STYLESHEET_TEMPLATE. Also, I would rather use `include_str!()` and keep these files out of the rust code so you can edit them with better tools.

b5: OMG thank you for `include_str!`. I'm coming from go (before the language tackled this problem) and just assumed embedding separate files would be hard. This is a delight

iroh-gateway/src/templates.rs
```diff
@@ -1,19 +1,93 @@
...
-</div>
-";
+pub const DIR_LIST_TEMPLATE: &str = include_str!("../assets/dir_list.html");
+pub const NOT_FOUND_TEMPLATE: &str = include_str!("../assets/404.html");
+pub const STYLESHEET: &str = include_str!("../assets/style.css");
+pub const ICONS_STYLESHEET: &str = include_str!("../assets/icons.css");
+
+lazy_static! {
+	static ref KNOWN_ICONS: HashMap<&'static str, ()> = HashMap::from([
+		(".aac", ()),
```
dignifiedquire, line 15: `HashSet`s are cool :)

b5: zomg these are amazing. ty

iroh-util/src/human.rs
```diff
@@ -0,0 +1,17 @@
+use humansize::{format_size, BINARY};
+
+/// Format byte count as a human-readable size string eg: 1_000_000u64 -> "1 MB"
```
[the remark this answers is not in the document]
b5: I ended up switching back to decimal units. SI Units are super irritating in practice

iroh-gateway/src/handlers.rs
```diff
@@ -42,15 +44,15 @@ use crate::{
...
     response::{get_response_format, GatewayResponse, ResponseFormat},
+    templates::{icon_class_name, ICONS_STYLESHEET, STYLESHEET},
 };
 
 /// Trait describing what needs to be accessed on the configuration
 /// from the shared state.
 pub trait StateConfig: std::fmt::Debug + Sync + Send {
     fn rpc_client(&self) -> &iroh_rpc_client::Config;
-
+    fn public_url_base(&self) -> String;
```
dignifiedquire, line 54: Should return `&str` to avoid allocations when not needed, like the other methods on this trait.

iroh-gateway/src/handlers.rs
```diff
@@ -550,16 +572,57 @@ async fn serve_fs_dir<T: ContentLoader + std::marker::Unpin>(
...
+    let mut breadcrumbs: Vec<HashMap<&str, String>> = Vec::new();
+    root_path
+        .trim_matches('/')
+        .split('/')
+        .fold(&mut breadcrumbs, |accum, path_el| {
+            let mut el = HashMap::new();
+            let path = match accum.last() {
+                Some(prev) => {
+                    let base = prev.get("path");
+                    format!("/{}/{}", base.unwrap_or(&"".to_string()), encode(path_el))
```
dignifiedquire, line 585: this will not do what you want, if `base` is empty this will produce `//<path_el>`.

```rust
match prev.get("path") {
  Some(base) => format!("/{}/{}", base, encode(path_el)),
  None => format!("/{}", encode(path_el)),
}
```
should be closer to what you want

dignifiedquire: actually looking at the whole structure this can be simplified to 

```rust
match accum.last().and_then(|prev| prev.get("path")) {
    Some(base) => format!("/{}/{}", base, encode(path_el)),
    None => format!("/{}", encode(path_el)),
}
```

iroh-gateway/src/templates.rs
```diff
@@ -1,19 +1,28 @@
...
-    </p>
-</div>
-";
+pub const DIR_LIST_TEMPLATE: &str = include_str!("../assets/dir_list.html");
+pub const NOT_FOUND_TEMPLATE: &str = include_str!("../assets/404.html");
+pub const STYLESHEET: &str = include_str!("../assets/style.css");
+pub const ICONS_STYLESHEET: &str = include_str!("../assets/icons.css");
+
+lazy_static! {
+    static ref KNOWN_ICONS: HashSet<&'static str> = HashSet::from([
```
dignifiedquire, line 10: there is a nicer version that avoids lazy_static: https://docs.rs/phf for this use case (also faaster ;))

iroh-gateway/src/handlers.rs
```diff
@@ -550,16 +572,59 @@ async fn serve_fs_dir<T: ContentLoader + std::marker::Unpin>(
     if !root_path.ends_with('/') {
         root_path.push('/');
     }
+
+    let mut breadcrumbs: Vec<HashMap<&str, String>> = Vec::new();
+    root_path
```
dignifiedquire, line 577: can I get a todo at least for writing a test for this?

dignifiedquire (approved):

dignifiedquire (approved):

PR #329, opened 2022-10-12
add persistent kad store

rklaehn:
Very much WIP...


rklaehn:
It will be a while until something like this becomes really useful, given the constraints of the current rust libp2p kad RecordStore trait. So closing this for the time being.

Issue #130, opened 2022-06-20
feat(gateway): add support for range requests


dignifiedquire:
This will allow proper seeking and streaming of videos and other files

We should be able to use https://github.com/scottlamb/http-serve to do the heavy lifting.

PR #330, opened 2022-10-12
feat: range request support

Arqu:
Closes https://github.com/n0-computer/iroh/issues/130

### DIFF
--- a/iroh-gateway/src/headers.rs
+++ b/iroh-gateway/src/headers.rs
@@ -137,7 +137,8 @@ pub fn set_etag_headers(headers: &mut HeaderMap, etag: String) {
 pub fn add_etag_range(headers: &mut HeaderMap, range: Range<u64>) {
     if headers.contains_key(ETAG) {
         let etag = headers.get(ETAG).unwrap().to_str().unwrap();
-        let etag = format!("{}.{}-{}", etag, range.start, range.end - 1);
+        let etag = etag.trim_end_matches('"');
+        let etag = format!("{}.{}-{}\"", etag, range.start, range.end - 1);
         headers.insert(ETAG, HeaderValue::from_str(&etag).unwrap());
     }
 }
--- a/iroh-gateway/src/headers.rs
+++ b/iroh-gateway/src/headers.rs
@@ -3,7 +3,7 @@ use ::time::OffsetDateTime;
 use axum::http::header::*;
 use iroh_resolver::resolver::{CidOrDomain, Metadata, PathType};
 use mime::Mime;
-use std::{fmt::Write, time};
+use std::{fmt::Write, ops::Range, time};
 
 #[tracing::instrument()]
 pub fn add_user_headers(headers: &mut HeaderMap, user_headers: HeaderMap) {
@@ -63,10 +63,39 @@ pub fn add_content_disposition_headers(
 
 #[tracing::instrument()]
 pub fn set_content_disposition_headers(headers: &mut HeaderMap, filename: &str, disposition: &str) {
+    // TODO: handle non-ascii filenames https://github.com/ipfs/specs/blob/main/http-gateways/PATH_GATEWAY.md#content-disposition-response-header
     headers.insert(
         CONTENT_DISPOSITION,
         HeaderValue::from_str(&format!("{}; filename={}", disposition, filename)).unwrap(),
     );
+}
+
+#[tracing::instrument()]
+pub fn add_content_range_headers(headers: &mut HeaderMap, range: Range<u64>, size: Option<u64>) {
+    let content_range = if let Some(size) = size {
+        format!("bytes {}-{}/{}", range.start, range.end - 1, size)
+    } else {
+        format!("bytes {}-{}/{}", range.start, range.end - 1, "*")
+    };
+    headers.insert(
+        CONTENT_RANGE,
+        HeaderValue::from_str(&content_range).unwrap(),
+    );
+}
+
+pub fn parse_range_header(range: &HeaderValue) -> Option<Range<u64>> {
+    let range = range.to_str().ok()?;
+    let mut parts = range.splitn(2, '=');
+    if parts.next() != Some("bytes") {
+        return None;
+    }
+    let mut range = parts.next()?.splitn(2, '-');
+    let start = range.next()?.parse().ok()?;
+    let end = range.next()?.parse().ok()?;
+    if start >= end || end == 0 {
+        return None;
+    }
+    Some(Range { start, end })
 }
 
 #[tracing::instrument()]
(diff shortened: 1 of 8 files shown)


iroh-gateway/src/headers.rs
```diff
@@ -74,12 +74,41 @@ pub fn add_content_disposition_headers(
...
     headers.insert(
         CONTENT_DISPOSITION,
         HeaderValue::from_str(&format!("{}; filename={}", disposition, filename)).unwrap(),
     );
 }
 
+#[tracing::instrument()]
+pub fn add_content_range_headers(headers: &mut HeaderMap, range: Range<u64>, size: Option<u64>) {
+    let content_range = if let Some(size) = size {
+        format!("bytes {}-{}/{}", range.start, range.end - 1, size)
```
dignifiedquire, line 87: if `range.end`  is `0` this will panic

iroh-gateway/src/headers.rs
```diff
@@ -74,12 +74,41 @@ pub fn add_content_disposition_headers(
...
+    } else {
+        format!("bytes {}-{}/{}", range.start, range.end - 1, "*")
+    };
+    headers.insert(
+        CONTENT_RANGE,
+        HeaderValue::from_str(&content_range).unwrap(),
+    );
+}
+
+pub fn parse_range_header(range: &HeaderValue) -> Option<Range<u64>> {
```
dignifiedquire, line 97: should have tests for parsing and generating these headers

iroh-resolver/src/resolver.rs
```diff
@@ -85,6 +85,22 @@ impl Path {
     }
 }
 
+#[derive(Debug, Clone, Copy)]
+pub enum ResponseClip {
```
dignifiedquire, line 89: docs please

iroh-resolver/src/resolver.rs
```diff
@@ -480,6 +516,41 @@ impl<T: ContentLoader + Unpin + 'static> AsyncRead for OutPrettyReader<T> {
...
+            | OutPrettyReader::DagJson(bytes_reader)
+            | OutPrettyReader::Raw(bytes_reader) => {
+                let pos_current = bytes_reader.pos;
+                let data_len = bytes_reader.bytes.len();
+                match position {
+                    std::io::SeekFrom::Start(pos) => {
+                        let i = std::cmp::min(data_len, pos as usize);
+                        bytes_reader.pos = i;
+                    }
+                    std::io::SeekFrom::End(_pos) => {
```
dignifiedquire, line 533: I believe `End` also supports negative seek, eg `-5` which should be easily supported in this case

iroh-resolver/src/resolver.rs
```diff
@@ -480,6 +516,41 @@ impl<T: ContentLoader + Unpin + 'static> AsyncRead for OutPrettyReader<T> {
...
+                        // This technically means we can seek past the end which we don't support
+                        // TODO: support rolling buffer reads, ie use overflow remainder to read from the start again
+                        todo!()
+                    }
+                    std::io::SeekFrom::Current(pos) => {
```
dignifiedquire, line 538: `pos` can be negative, just casting it to `usize` is not enough

iroh-resolver/src/resolver.rs
```diff
@@ -1417,9 +1496,13 @@ mod tests {
             if let OutContent::Unixfs(node) = ipld_hello_txt.content {
                 assert_eq!(
                     read_to_string(
```
dignifiedquire, line 1498: Should have tests with clipping and seeking

iroh-resolver/src/unixfs.rs
```diff
@@ -492,16 +498,57 @@ impl<T: ContentLoader + Unpin + 'static> AsyncRead for UnixfsContentReader<T> {
...
+                        *pos = (*pos as i64 + offset) as usize;
+                    }
+                }
+            }
+        }
+        Ok(())
+    }
+
+    fn poll_complete(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<std::io::Result<u64>> {
+        Poll::Ready(Ok(0))
```
dignifiedquire, line 532: wrong return value

iroh-gateway/src/headers.rs
Arqu: I can add another guard, but this path is not invoked unless range exists and is passed in. And that is checked against in `parse_range_header`

dignifiedquire: then maybe add a comment about that

iroh-resolver/src/resolver.rs
```diff
@@ -480,6 +516,41 @@ impl<T: ContentLoader + Unpin + 'static> AsyncRead for OutPrettyReader<T> {
...
+                    }
+                }
+                Ok(())
+            }
+            OutPrettyReader::Unixfs(r) => Pin::new(&mut *r).start_seek(position),
+        }
+    }
+
+    fn poll_complete(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<std::io::Result<u64>> {
+        Poll::Ready(Ok(0))
```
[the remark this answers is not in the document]
Arqu: Tbh I don't fully understand this part around poll_complete and I just pieced it together from tokio's BufReader. 
The result if not `(0)` is `self.as_mut().get_pin_mut().poll_complete(cx)?` which to me in their case seems to resolve to the same.

iroh-resolver/src/unixfs.rs
Arqu: Same as above, would love some input on what it should be. I'm assuming the delta from the initial position which in a simplified version returns the seek offset?

iroh-resolver/src/resolver.rs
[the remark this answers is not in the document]
dignifiedquire: Even it does resolve to the same, that's what it relies on for the seeking, it should proxy there.

iroh-resolver/src/unixfs.rs
dignifiedquire: Yes, I think this is the delta, so we need to store it and return it. Later when we do seeking with loading, this will also delegate to an internal future.

Arqu: So I took some time to understand this and as far as I get it it's not the delta but rather the final position it seeked to. Because otherwise the consumer would need to figure out what the delta means especially in circular buffer and negative moves etc.

Arqu:
@dignifiedquire can you take another look at the unixfs seeking, it's funkier now. 
Also added tests and cleaned it up.

iroh-resolver/src/resolver.rs
```diff
@@ -1678,6 +1700,129 @@ mod tests {
...
+
+            // seek and clip
+            let cr = seek_and_clip(&node, resolver.clone(), 1..3).await;
+            assert_eq!(read_to_string(cr).await, "el");
+        } else {
+            panic!("invalid result: {:?}", ipld_hello_txt);
+        }
+    }
+
+    #[tokio::test]
```
dignifiedquire, line 1759: Either in this PR or at least have an issue for seeking across dag-cbor, dag-json and raw files as well please.

iroh-resolver/src/unixfs.rs
dignifiedquire: Yes new offset from the start is what should be returned 👍

iroh-resolver/src/resolver.rs
Arqu: https://github.com/n0-computer/iroh/issues/346

Arqu: Gotta do other stuff before release :/

dignifiedquire:
merge conflict 😅

dignifiedquire (dismissed):
Looks good, please also open an issue to add some additional tests on the seeking math, that is easy to get wrong on edge cases and I don't trust my own review of them. Some nice proptests should help for confidence.

Issue #269, opened 2022-10-03
iroh get command


faassen:
## Discussion proposal

`iroh-ctl get <ipfs-path> [<path>]`

- `ipfs-path` can be a true IPFS path, or a CID
- `path` is a filesystem path and is optional. If it’s not supplied, path is set to the `CID`.

Download file or directory from IPFS into `path`. If `path` already exists and is a file then it’s overwritten with the new downloaded file. If `path` already exists and is a directory, the command fails with an error. If `path` already exists, is a file and the downloaded data is a directory, that’s an error

### Differences with Kubo:

- `ipfs get` has an `--output` flag, instead we have a second optional `path` argument.
- `ipfs get` downloads into `--output` if that output path already exists and is a directory. We consider that an error. The motivation is that it’s while it mirrors the behavior of Unix `cp`, we can’t fully mirror `cp` because we don’t always have the source file name or directory, and it risks creating an accidental nested directory structure when you repeatedly download.
- `ipfs get` silently overwrites the directory if no `--output` was given. Instead of silently overwriting, we choose to present the user with an error so that the behavior of the omitted `path` is identical to when you supply one.
- We don’t support compression of downloaded content; you can do that with a separate tool.

## Test cases

- [x] get with CID -> ok
- [x] get with something that looks like a CID but isn't -> error
- [x] get with IPFS path that points to a CID -> ok
- [x] get with IPFS path that doesn't point to a CID ->  error
- [x] get wrapped file
- [x] get unwrapped file
- [ ] get unwrapped file, overwrites previous
- [x] get directory without explicit path
- [x] get directory with explicit path
- [x] get directory without explicit path, previous file/directory already there -> error
- [x] get directory with explicit path, previous file/directory already there -> error
- [ ] get a file with an ipfs path: `iroh get cid/a/b` where `b` is a file
- [ ] get a directory with an ipfs path `iroh get cid/a/b` where `b` is a directory
- [ ] fail to get a file/directory with an ipfs path `iroh get cid/a/b` where b doesn't exist

## Questions

- [ ] get with CID that doesn't exist -> not sure how to test this; I tried a made up CID with Kubo and it's just hanging looking for it. Can we determine a CID doesn't exist?

b5:
Looks great to me. I'd like to advocate for one flag that will help users see just how terrible network fetching is:

```
--force-fetch  ignore local cache & fetch all data from the network
```

[...]

PR #331, opened 2022-10-12
feat: CLI get command improvements

faassen:
This PR focuses on creating tests for the `iroh get` CLI command. The Iroh API is also improved along the way.

This required more work in the underlying API to make it more testable. What's tested here is everything *except* the actual IPFS behavior -- the focus is on the behavior of the CLI and API and its interactions with the filesystem.

 ## trycmd tests

In `iroh/tests/cmd` you can see now quite a few `.trycmd` files. These describes the command-line interaction. Stuff behind `$ is a command, there's a special `? failed` indicator if the command is considered to have failed, and the output of the command is shown.

New are the `.out` and `.in` directories with the same names as the `.trycmd` files. These describe the filesystem before the command runs (may be missing), and the filesystem after the command has run. This way we can describe the effects of the `iroh get` command - directories and files are supposed to be created. We can also test failure scenarios where we refuse to overwrite a directory that already exists. #269 tracks various test cases.

@b5 is probably most interested in this bit

## `get_stream`

The `get` high level CLI method has been removed from the mockable `Api` trait (read on to see where it went). Instead, a more low level but still useful API method `get_stream` has been added to the `Api` trait. This gets a stream of relative paths and `OutType`, describing the directories and files you can create. 

This is a refactored version of the code @ramfox originally wrote and was in `getadd.rs` previously. The big difference is that it returns relative paths and doesn't calculate final destination paths -- that's up to the user of the API.

## No `async_trait` macro for `Api` trait

The interactions between the `Api` trait, `mockall` macro and `async_trait` were getting so hairy I couldn't figure out how to express things anymore once I wanted to add `get_stream`. For my sanity and also to learn better how this really works underneath, I've rewritten the `Api` trait to describe itself explicitly in terms of `(Local)BoxFuture` and `(Local)_BoxStream`. It's more verbose but functionally equivalent and I could express what I wanted. 

## `ApiExt` trait

The `ApiExt` trait is a trait that implements the high level `get` command. It puts everything together: it handles various error conditions, accesses the stream and then writes the stream to the filesystem. It's basically `iroh get`.

The `ApiExt` trait is solely intended to contain default trait methods, and is automatically available when the `Api` contract is fulfilled (if you `use` it).

I factored out `save_get_stream` from `getadd.rs` to be solely concerned with turning a stream into files and directories on the files system. That makes it possible to test its behavior in isolation.

## test fixture

The `get` test fixture now mocks `get_stream` and returns a fake stream made from a `Vec`. This defines the actual stuff that the CLI writes to disk.

## relative_path

I now depend on the [`relative-path` crate](https://crates.io/crates/relative-path) because what the get stream returns are clearly relative paths, and I want to force the user to do something with them for being able to actually write stuff. I want to expand the use of relative paths in the CLI where it makes sense in the future, as the use is now only minimal. It should help make it really explicit where the CLI is going to read and write.


faassen:
Hm, `get_cid_directory_overwrite_explicit_failure.trycmd` is failing in CI but it passes locally if I run the tests for the `iroh` CLI crate specifically or when I run them for all crates.

faassen:
One thing I realized is that the fixture stream I created may not actually reflect a real world stream; I'm unclear how it works when directories are wrapped or not. I'm going to experiment a bit with real-world streams.

faassen:
Okay, after some manual testing I see I definitely broke the correct behavior. Withdrawing this from review until I fix it.

faassen:
Okay, I have resolved the underlying issues; it turns out that the stream was not returning what I expected. I've adjusted it so it does.

I've also written tests for the single unwrapped file scenario.

iroh/tests/cmd/get_ipfs_path_success.out/QmP8jTG1m9GSDJLCbeWhVSVgEzCPPwXRdCRuJtQ5Tz9Kc9/b
```diff
@@ -0,0 +1 @@
+hello
```
ramfox, line 1: https://github.com/n0-computer/iroh/blob/d3fb9b4677170ea70595e607078a6ce1687d760c/iroh/src/fixture.rs#L31-L45

`fixture_get` sends a stream of 1) the wrapped directory, 2) a directory called "a" 3) a file called "b"
(If "b" is supposed to be nested below "a", then the path to "b" should be `RelativePathBuf::from_path("a/b")`)

Which means that there should be 2 entries below the `get_ipfs_path_success.out/QmP8jTG1m9GSDJLCbeWhVSVgEzCPPwXRdCRuJtQ5Tz9Kc9` folder, an empty folder "a", and the file "b" that reads "hello".

When I run `TRYCMD=dump cargo test --test cli_tests`
The dump output does show the directory `get_ipfs_path_success.out/QmP8jTG1m9GSDJLCbeWhVSVgEzCPPwXRdCRuJtQ5Tz9Kc9/a`, which leads me to believe that it should be creating it. But it isn't flagging as an error.  Unless this ignores empty directories for some reason?

iroh-api/src/api_ext.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// High level get, equivalent of CLI `iroh get`
+    async fn get<'a>(
+        &self,
+        ipfs_path: &IpfsPath,
+        output_path: Option<&'a Path>,
+    ) -> Result<PathBuf> {
+        let cid = ipfs_path
+            .cid()
+            .ok_or_else(|| anyhow!("IPFS path does not refer to a CID"))?;
+        let root_path = get_root_path(cid, output_path);
```
ramfox, line 21: Ah, okay, we are actually missing a step here that was covered by `make_output_path`

You can see for yourself if you take these steps:
- add a directory "a" with a file inside "b" with content "hello", let's say the cid that results is `bafyfoo`
- run `iroh get bafyfoo/a/b new_b`
- the expected result is to get a new file `./new_b` with content "hello", instead you get `./new_file/a/b`
- you can see the difference if you compare the output from `main` & this branch running the same `get` command

iroh/tests/cmd/get_ipfs_path_success.out/QmP8jTG1m9GSDJLCbeWhVSVgEzCPPwXRdCRuJtQ5Tz9Kc9/b
ramfox: This just has me worried that we aren't actually testing the behavior that happens on the actual command line that we want to test.

ramfox (changes requested):

iroh-api/src/api_ext.rs
ramfox: this pr shows the correct behaviour: https://github.com/n0-computer/iroh/pull/339

faassen: Ah, I had missed that `iroh get cid/a/b` is a thing at all! I thought you had to address it all with a cid. Wasn't in the list of test cases in the `iroh get` issue for that reason. I will look at your work and integrate it.

iroh/tests/cmd/get_ipfs_path_success.out/QmP8jTG1m9GSDJLCbeWhVSVgEzCPPwXRdCRuJtQ5Tz9Kc9/b
faassen: trycmd is definitely finicky sometimes, will investigate empty folder behavior. May adjust the test to deal with nested streams - a separate unit test for saving the stream can then confirm creation.

faassen: Aha, I was slightly confused about this as the empty directories do exist for me in the `out` directories, but I think this has to do with `git` just not adding empty directories and me missing that. So for me everything looks right but in the PR the directories are missing. This indeed means `trycmd` seems to tolerate missing (empty?) directories. I will adjust the tests to avoid empty directories first and we'll look into `trycmd` later.

faassen: Okay, `trycmd` seems to not only tolerate an `.out` with a directory entirely missing, but even a *file* entirely missing. The docs say: 

"When present, each file in this directory will be compared to generated or modified files."

so it only compares things that are both there, and doesn't warn that it's missing. Basically it only compares what's present in `.out` and nothing else. That's very surprising and I created an issue for it: https://github.com/assert-rs/trycmd/issues/149

faassen: Update: what I wrote here isn't exactly right. If a file is present in `.out` and not in the output, `trycmd` does trip up. It's the other way around that's tricky: it doesn't detect *extra* stuff is generated and is missing from `.out`.

faassen: I've now made sure I don't generate empty directories and ensured there's a file in all of them to compare. I'd still like a trycmd strict mode that fails, but it'll do for now.

iroh-api/src/api_ext.rs
faassen: I've now modified `get_stream` so it is aware of sub paths and skips anything unrelated. This appears to do the right thing when I test it by hand. 

I'm not very happy that I can't have this behavior under test, but I do think that `get_stream` should deliver only what someone asked for and not more; after that it is just a matter of writing it out. Ideally I think this behavior belongs in the resolver; after all `resolve_recursive_with_paths` knows everything it has to know already and it could potentially be doing optimizations that the API layer can't.

faassen:
@ramfox can you have another look to see whether it does the right thing? I went for a slightly different approach than you did to handle the ipfs tail paths, but it appears to work.

iroh-api/src/api.rs
```diff
@@ -82,45 +93,66 @@ impl Api for Iroh {
...
+        let results = resolver.resolve_recursive_with_paths(ipfs_path.clone());
+        let sub_path = ipfs_path.to_relative_string();
+        async_stream::try_stream! {
+            tokio::pin!(results);
+            while let Some(res) = results.next().await {
+                let (relative_ipfs_path, out) = res?;
+                let relative_path = RelativePathBuf::from_path(&relative_ipfs_path.to_relative_string())?;
+                // TODO this focusing in on sub-paths should really be handled in the resolver,
+                // because it'll be testable there, and it allows potential optimizations we
+                // can't do here
```
ramfox, line 111: can you write this TODO in this fashion:
`TODO(faassen)`

Also, I don't know if I agree with this. I would expect the path that comes out of the resolver be the path that it resolved the content from. I can see changing the output of the resolver from `PathBuf` to a full `IpfsPath`

iroh-resolver/src/resolver.rs
```diff
@@ -119,18 +119,20 @@ impl Path {
...
-            }
-            s.push_str(&format!("/{}", part)[..]);
-        }
-        if self.has_trailing_slash() {
-            s.push('/');
+    // Arqu says that empty path segments in the *middle* shouldn't occur,
+    // though they can occur at the end, which `join` handles.
+    // TODO: it would make sense to return a `RelativePathBuf` here at some
+    // point in the future so we don't deal with bare strings anymore and
+    // we're forced to handle various cases more explicitly.
```
ramfox, line 126: Can you write the TODO as `TODO(faassen)`

I think, let's also take out the reference to asmir here, let's not get into a finger-pointing situation if an empty path segment does occur, lol

ramfox (changes requested):
just come request for changes around comments!

iroh-api/src/api.rs
faassen: I attributed the todo and checked in a more elaborate argument. :)

ramfox (approved):

PR #333, opened 2022-10-13
use xtasks for release builds and man page generation

b5:
This PR adds support for manual pages generated from `clap` documentation, and the [cargo xtask pattern](https://github.com/matklad/cargo-xtask) for coding up low-overhead automation in Rust.


xtask/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "xtask"
+version = "0.1.0"
+edition = "2021"
```
dignifiedquire, line 4: this should be marked as private to avoid accidentally trying to publish this crate

xtask/src/main.rs
```diff
@@ -0,0 +1,105 @@
...
+
+fn main() {
+    if let Err(e) = try_main() {
+        eprintln!("{}", e);
+        std::process::exit(-1);
+    }
+}
+
+fn try_main() -> Result<(), DynError> {
+    let task = env::args().nth(1);
```
dignifiedquire, line 18: why not use clap instead?

xtask/src/main.rs
```diff
@@ -0,0 +1,105 @@
...
+fn dist_manpage() -> Result<(), DynError> {
+    let app = iroh::run::Cli::command();
+
+    // TODO(b5): This only generates the man page for the iroh command
+    // need to recursively handle all subcommands, generating man files
+    // for them as well
+
+    let man = clap_mangen::Man::new(app);
+    let mut buffer: Vec<u8> = Default::default();
+    man.render(&mut buffer)?;
```
dignifiedquire, line 88: instead of writing to memory you can create and open the file, create a bufwriter aroubd it and the write to it

dignifiedquire:
please remove .ds_store :)

xtask/src/main.rs
```diff
@@ -6,32 +6,33 @@ use std::{
     process::{Command, Stdio},
 };
 
+fn cli() -> clap::Command {
+    clap::Command::new("xtasks")
```
dignifiedquire, line 10: please use the struct form for command definitions, it‘s considerably easier to maintain and reason about

xtask/src/main.rs
```diff
@@ -6,32 +6,33 @@ use std::{
...
-    let task = env::args().nth(1);
-    match task.as_deref() {
-        Some("dist") => dist()?,
-        Some("man") => dist_manpage()?,
-        _ => print_help(),
+fn run_subcommand(matches: clap::ArgMatches) -> Result<()> {
+    match matches.subcommand() {
+        Some(("dist", _)) => dist()?,
+        Some(("man", _)) => dist_manpage()?,
+        _ => unreachable!(), // If all subcommands are defined above, anything else is unreachabe!()
```
dignifiedquire, line 31: None is very possible if no subcommand is given.

ramfox (dismissed):

xtask/src/main.rs
b5: for those of you who are well versed in rust compiler magic, sure. I personally find this form more readable.

That said, I'm sure I'll come to learn this compiler magic in due time, and will defer to you on all things Rust. Will change to the struct form

xtask/src/main.rs
b5: nah, that's caught by the combination of `cli().get_matches();` and the root command `arg_required_else_help`

dignifiedquire: great :)

xtask/src/main.rs
dignifiedquire: I am looking forward to once your brain has embraced the strength and statics of types 💙

dignifiedquire (approved):
Last note, maybe add something to our development docs and/or the readme about this?

b5:
yeah, I plan to put a little more work into the actual things that are automated (like generating _all_ release binaries) before listing this tool in `DEVELOPERS.md`. Also want to talk to @Arqu about how this may-or-may-not intersect with CI tasks

Issue #318, opened 2022-10-11
Write database consistency tests


rklaehn:
RocksDB is a very low level key value store. Since it does not enforce complex invariants like a relational DB, we need to be sure we do this correctly.

Write a property based test that does operations on the database and checks invariants. Such a test can also serve as good and up to date documentation about what invariants there are in the first place.

Operations on the DB should be done using the public interface, then an internal consistency check (yet to be written) should be called after a sufficiently large number of operations.

b5:
@rklaehn, do you think this is closable?

PR #336, opened 2022-10-13
Map id to cid instead of multihash

### LINKED ISSUE #335: Properly handle multiple cids with the same hash
Currently the store assumes a 1:1 mapping of hashes and cids. Ids in the store correspond to multihashes, not cids.

Here is a scenario where this can cause problems: You can have 2 cids with the exact same hash but different codecs. E.g. a block with a raw codec can never have any links, whereas a block with e.g. a dag-cbor codec can.

So you can add a block with codec=raw hash=1234 which contains dag-pb with links. This will assign an id, and record that there are no links for this id in the graph table (because it is the raw codec).

Then once you add a block with codec=dag-pb hash=1234, you will assume that you already have that hash and not update the graph. Alternatively you could update the graph, which would mean that the store now thinks that a cid with a raw codec has links.

I think the links table needs to be done for cids, not hashes. Which means that the store must be changed so that ids correspond to cids, and there is just possibily* a deduplication based on hashes.

(*) not sure if it is worth it, since this is probably going to be rare

rklaehn:
Fixes #335

### DIFF
--- a/iroh-store/src/store.rs
+++ b/iroh-store/src/store.rs
@@ -16,8 +16,8 @@ use iroh_metrics::{
 use iroh_rpc_client::Client as RpcClient;
 use multihash::Multihash;
 use rocksdb::{
-    BlockBasedOptions, Cache, ColumnFamily, DBPinnableSlice, Direction, IteratorMode, Options, WriteBatch,
-    DB as RocksDb,
+    BlockBasedOptions, Cache, ColumnFamily, DBPinnableSlice, Direction, IteratorMode, Options,
+    WriteBatch, DB as RocksDb,
 };
 use smallvec::SmallVec;
 use tokio::task;
@@ -641,8 +641,8 @@ mod tests {
         assert_eq!(store.get_links(&raw_cid).await?.unwrap().len(), 0);
         assert_eq!(store.get_links(&cbor_cid).await?.unwrap().len(), 2);
 
-        let ids = store.get_ids_for_hash(&hash)?.collect::<Vec<_>>();
-        assert_eq!(ids.len(), 2);
+        let ids = store.get_ids_for_hash(&hash)?;
+        assert_eq!(ids.count(), 2);
         Ok(())
     }
 }
--- a/iroh-store/src/store.rs
+++ b/iroh-store/src/store.rs
@@ -71,6 +71,15 @@ fn default_blob_opts() -> Options {
     opts
 }
 
+fn id_key(cid: &Cid) -> Vec<u8> {
+    let hash = cid.hash().to_bytes();
+    let code = cid.codec();
+    let mut key = Vec::with_capacity(8 + hash.len());
+    key.extend_from_slice(&code.to_be_bytes());
+    key.extend_from_slice(&hash);
+    key
+}
+
 impl Store {
     /// Creates a new database.
     #[tracing::instrument]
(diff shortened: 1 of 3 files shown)


dignifiedquire:
this really problematic, the dht only works on multihashes for provider records, so we must be able to find content based on only multihashes

rklaehn:
> this is really problematic, the dht only works on multihashes for provider records, so we must be able to find content based on only multihashes

Doesn't bitswap work on the level of cids? Where exactly do you need to *look up* content by multihash currently?

dignifiedquire:
- bitswap works with both depending on the protocol version
- when we provide records they are going to be by multihash, not by cid
- when making queries to the dht we always convert to a multihash

rklaehn:
> * bitswap works with both depending on the protocol version

Ah, I did not know that.

> * hey are going to be by multihash, not by cid
> * when making queries to the dht we always convert to a multihash

That just requires going from cid to multihash, not vice versa, so no big deal.


So as far as I can see there are two solutions:
- change the graph table so that the key is (code, hash_id) and the value is [(code, hash_id)]
  I would prefer to avoid that since it makes graph traversal less efficient
- keep ids as 1:1 with cids but provide a way to look up blobs by hash nonetheless. Key for id could be <hash><code>, then just look for entries with a prefix of <hash>.

I guess it depends on what we expect to be the more frequent use case - lookup by cid or by hash.

dignifiedquire:
@rklaehn yes

rklaehn:
The gist of the above is that ids are still mapped to cids, but the key for looking up ids from cids is

`<hash><code>`.

So to look up ids by hash you can just do a prefix search by hash and get all codes for which we have the hash. Just pick one that has associated data. Typically, in 99.9% of all cases, there should be exactly one?

Should we go with this? The other alternative would be to have id correspond to hash, and store (id, code) in both graph key and graph content. But I would really like to avoid that since it will make all graph traversal code slower and more cumbersome, for something that is not needed 99.9% of the time.

I think this case definitely needs to be solved one way or another before v1, because it allows you to get the store in a weird state (it is confused about links) from the outside.

iroh-store/src/store.rs
```diff
@@ -71,6 +73,13 @@ fn default_blob_opts() -> Options {
     opts
 }
 
+fn id_key(cid: &Cid) -> SmallVec<[u8; 64]> {
```
dignifiedquire, line 76: Can you add a doc comment describing the format please

iroh-store/src/store.rs
```diff
@@ -217,6 +226,40 @@ impl Store {
         Ok(())
     }
 
+    #[tracing::instrument(skip(self))]
+    pub async fn get_blob_by_hash(&self, hash: &Multihash) -> Result<Option<DBPinnableSlice<'_>>> {
```
dignifiedquire, line 230: can you add a test for these two new methods as well please?

iroh-store/src/store.rs
```diff
@@ -291,6 +334,38 @@ impl Store {
         }
     }
 
+    fn get_ids_for_hash(
+        &self,
+        hash: &Multihash,
+    ) -> Result<impl Iterator<Item = Result<(u64, u64)>> + '_> {
```
dignifiedquire, line 340: It is very unfortunate to return a tuple of `u64`s, maybe use a struct to avoid mixing them up? Or we could introduce a new type for `Id`.

dignifiedquire (commented):
I think this is a good way to go 👍

iroh-store/src/store.rs
rklaehn: A newtype for Id will be quite annoying since it is used all over the place. But I will make one for Code so you can not accidentally use it as an id.

rklaehn:
Addressed your PR comments.

These tests are pretty much just smoke tests. I think we need more tests, but that is unrelated to this particular issue. See https://github.com/n0-computer/iroh/issues/318

dignifiedquire (approved):

Issue #147, opened 2022-07-21
Full UnixFS DAG construction support


b5:
Tracking issue to follow our support for creating UnixFS files, directories with iroh

* [x] Create files of "any" size
* [x] Create files within a directory
* [x] Proper file size accounting #200 #254 
* [x] Recursive subdirectories `kubo add -r $DIR`
* [x] Can create symlinks
* [ ] ~~Exposed as a library-level primitive~~
* [ ] ~~Integration test that confirms kubo can open a multi-block file in doubly-nested directory created with iroh~~
* [x] adding works via `irohctl`

PR #337, opened 2022-10-14
feat(unixfs): add symlink support

### LINKED ISSUE #306: unixfs symlink support
We need to recognize when a directory entry or path refers to a symlink, and need to create a `UnixfsNode::Symlink` in response.

Currently, (at least on a mac), if a directory entry is a symlink, `iroh add` follows the symlink & adds it as a `UnixfsNode::File` or `UnixfsNode::Dir`.

ramfox:
closes #306 #147 

----

this PR touches code that is also touched by #331 , so we should probably wait for that to merge before merging this.

Still would love feedback before then tho!


iroh-resolver/src/unixfs_builder.rs
```diff
@@ -518,7 +633,10 @@ async fn make_dir_from_path<P: Into<PathBuf>>(path: P, recursive: bool) -> Resul
     let mut directory_reader = tokio::fs::read_dir(path.clone()).await?;
     while let Some(entry) = directory_reader.next_entry().await? {
         let path = entry.path();
-        if path.is_file() {
+        if path.is_symlink() {
+            let s = SymlinkBuilder::new(path).build().await?;
```
dignifiedquire, line 637: missing target?

ramfox: Ah, so In `FileBuilder`, if you only give it a path, it will derive the `name` & a `reader` for you from that `path`. I matched that pattern here, if you only provide a `path` it will derive the `target` for you. Or you can add it explicitly using the `target` method.

iroh/src/fixture.rs
```diff
@@ -78,6 +79,33 @@ fn fixture_get_unwrapped_file() -> MockApi {
...
+    api.expect_get_stream().returning(|_ipfs_path| {
+        futures::stream::iter(vec![Ok((
+            RelativePathBuf::from_path("").unwrap(),
+            OutType::Symlink(PathBuf::from("target/path/foo.txt")),
+        ))])
+        .boxed_local()
+    });
+    api
+}
+
```
ramfox, line 108: @faassen looking for clarify here

iroh-resolver/src/unixfs_builder.rs
dignifiedquire: ah I see, makes sense

dignifiedquire:
ci is very sad

iroh-resolver/src/unixfs_builder.rs
```diff
@@ -954,6 +954,21 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test]
+    async fn symlink_from_disk_test() -> Result<()> {
+        let temp_dir = std::env::temp_dir();
```
dignifiedquire, line 959: while this works, I would recommend using https://docs.rs/tempfile it is more reliable and takes care of cleanup

PR #338, opened 2022-10-14
feat: race a http gateway with p2p in iroh-one's content loader

fabricedesre:
This adds to iroh-one a new optional configuration item: resolver_gateway
If resolver_gateway is set, the resolver will race the p2p node with fetching a raw block from the gateway.

A new source is also introduced to classify blocks coming from a gateway out from those fetched from p2p or store. Hits from the gateway are considered as not cached for telemetry purposes.

ping @b5


Arqu (commented):
Looks good, can we get the `rustfmt` in order and I can merge then.

fabricedesre:
> Looks good, can we get the `rustfmt` in order and I can merge then.

fixed in latest push.

PR #340, opened 2022-10-14
feat(cli): start and stop commands

### LINKED ISSUE #290: iroh start command
Command for starting a long-running iroh process that participates in an IPFS network.

```
$ iroh start --help
iroh-start
start a long running iroh process

USAGE:
    iroh start

OPTIONS:
    -h, --help    Print help information

Iroh start kicks off a long running process that manages p2p connections, serves
any configured APIs, and proxies access for other iroh processes to shared 
resources like store and network connections. While iroh start is running iroh 
will initiate connections to other peers in the network to both provide and 
fetch content, all of which can be governed through configuration.

When iroh start is running, it acquires a lock on shared resources. Running an
iroh command from another terminal will be executed as a remote procedure call
on the 'iroh start' process. To check if a command will run through 'iroh start'
use `iroh status`.

Shutdown

To stop iroh start, send a SIGINT signal to it (e.g. by pressing 'Ctrl-C')
or send a SIGTERM signal to it (e.g. with 'kill'). It may take a while for the
daemon to shutdown gracefully, but it can be killed forcibly by sending a
second signal.
```

This is analagous to `kubo daemon`.

## Differences from Kubo

Kubo has 17 flags on `kubo daemon`, none of which we plan to support in the v0.1.0 release.

### LINKED ISSUE #308: iroh stop command
now that we have more clarity on what `start` does, and that it's just an alias for kicking off a background process, it makes sense to also provide a stop command. Proposed help text:

```
> iroh stop --help
iroh-stop
Stop iroh daemon(s)

USAGE:
    iroh stop

OPTIONS:
    -h, --help    Print help information


stop checks for a running iroh daemon and halts any daemon process it finds.
stop sends SIGINT to each process and waits for the process to exit gracefully.
If the process does not exit gracefully on it's own within 30 seconds stop sends
SIGKILL to the deamon process and exits.

If stop finds no running processes it does nothing and exits with status code 0.

Processes are identified via their process ID as recorded on the local program
lock. If stop cannot find lock files on the local filesystem (possibly because
iroh is configured to use a network-backed deamon), it will exit with an error.
```

b5:
when it's finished: closes #290, closes #308

### DIFF
--- a/iroh/Cargo.toml
+++ b/iroh/Cargo.toml
@@ -23,7 +23,8 @@ iroh-metrics = { path = "../iroh-metrics", default-features = false, features = 
 iroh-api = { path = "../iroh-api"}
 relative-path = { version = "1.7.2", optional = true }
 iroh-util = { path = "../iroh-util"}
-fork = "0.1"
+handlebars = "4"
+serde_json = "1.0.78"
 which = "4.3.0"
 sysinfo = "0.26.4"
 
--- a/iroh/Cargo.toml
+++ b/iroh/Cargo.toml
@@ -23,8 +23,7 @@ iroh-metrics = { path = "../iroh-metrics", default-features = false, features = 
 iroh-api = { path = "../iroh-api"}
 relative-path = { version = "1.7.2", optional = true }
 iroh-util = { path = "../iroh-util"}
-handlebars = "4"
-serde_json = "1.0.78"
+localops = { path = "../localops"}
 which = "4.3.0"
 sysinfo = "0.26.4"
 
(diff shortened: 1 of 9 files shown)


iroh/Cargo.toml
```diff
@@ -21,6 +21,10 @@ tonic = "0.8"
 git-version = "0.3.5"
 iroh-metrics = { path = "../iroh-metrics", default-features = false, features = ["rpc-grpc"] }
 iroh-api = { path = "../iroh-api"}
+iroh-util = { path = "../iroh-util"}
+fork = "0.1"
```
dignifiedquire, line 25: pretty sure this is entirely incompatible with windows :(

b5:
Closing in favour of #360

PR #343, opened 2022-10-14
feat: CLI tests for iroh add, and API changes

faassen:
This contains work on the add command, with a few CLI tests and some 
API tweaks.

## add layering

add_file and add_dir in the core Api that's mockable, and the slightly higher level `add` is above it in `ApiExt`.

## recursive simplication

Test for the `-r` flag in the CLI and just let `add_dir` in the resolver add a directory without worrying. This is tested.

## a few add CLI tests

A few basic add CLI tests.

## CLI test cleanup

I've went through the `cli_tests.rs` module step by step cleaning and reorganizing the tests so it's more easy to maintain them. Unfortunately `trycmd` gets too finicky if I try to use its globbing feature (as input and output directories seem shared sometimes). So now I maintain an explicit list in the same order as they show up alphabetically. It's too verbose but it's the best we can do for now.


iroh-api/src/api_ext.rs
```diff
@@ -29,6 +29,16 @@ pub trait ApiExt: Api {
         save_get_stream(&root_path, blocks).await?;
         Ok(root_path)
     }
+
+    async fn add(&self, path: &Path, wrap: bool) -> Result<Cid> {
```
rklaehn, line 33: So the code to figure out if it is a file or a dir is moved to ApiExt, and the main fns in api require you to be explicit? I like it.

iroh-resolver/src/unixfs_builder.rs
```diff
@@ -471,15 +463,10 @@ pub async fn add_file<S: Store>(store: Option<S>, path: &Path, wrap: bool) -> Re
 /// - storing the content using `rpc.store`
 /// - returns the root Cid
 /// - optionally wraps into a UnixFs directory to preserve the directory name
-pub async fn add_dir<S: Store>(
-    store: Option<S>,
-    path: &Path,
-    wrap: bool,
-    recursive: bool,
```
rklaehn, line 478: Why is the recursive flag not needed anymore? Do we always use recursive?

iroh/src/fixture.rs
```diff
@@ -51,6 +53,26 @@ fn fixture_get() -> MockApi {
     api
 }
 
+fn fixture_add_file() -> MockApi {
+    let mut api = MockApi::default();
+    api.expect_add_file().returning(|_ipfs_path, _| {
+        Box::pin(future::ready(
+            Cid::from_str("QmYbcW4tXLXHWw753boCK8Y7uxLu5abXjyYizhLznq9PUR").map_err(|e| e.into()),
```
rklaehn, line 60: So this returns the same cid whatever you add. Is the intention to make it more realistic, or just write another mock for that?

rklaehn (approved):
Looks good to me. Nice tests for the actual command output.

You have to be careful not to get bogged down by these kinds of tests though. We don't want to spend days making sure that every byte coming out of stdout is the same when refactoring.

But the way it is done here seems to be quick to update.

iroh-api/src/api_ext.rs
faassen: Yes, and the check for the `recursive` flag before copying a directory was removed from the API entirely, because the use case for writing code like this is very different from the reason it's in the CLI (which is to prevent accidents in interactive mode).

iroh-resolver/src/unixfs_builder.rs
faassen: The use case for the `recursive` flag in my mind is to prevent people from accidentally adding a directory in the CLI. That use case doesn't exist on the API level, so I've removed it from the API. The implementation is now on the CLI level which is much simpler than maintaining this in the UnixFS builder.

iroh/src/fixture.rs
faassen: Just write another mock if that becomes a use case. In general I don't think we should strive for realism in these mocks, just fulfilling the contract in the easiest way possible for a particular test case.

Unfortunately this means for `add` there isn't all that much to mock yet. Perhaps that will change somewhat once the `add` API starts to return more stuff related to progress, though I suspect tests for that will have to be normal Rust unit tests, not `trycmd` tests.

rklaehn (approved):

PR #347, opened 2022-10-14
chore: add dev-install xtask

b5:
fixes `private` config in the `xtask` crate (should be `publish = false`).

Adds new `cargo xtask dev-install` that'll copy all binaries to `$HOME/.cargo/bin/`, which is useful while I'm working on `iroh start`.

Optionally run `cargo xtask dev-install --build` to run a build of all binaries first

### DIFF
--- a/xtask/Cargo.toml
+++ b/xtask/Cargo.toml
@@ -2,7 +2,7 @@
 name = "xtask"
 version = "0.1.0"
 edition = "2021"
-private = true
+publish = false
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
@@ -10,4 +10,6 @@ private = true
 anyhow = "1.0"
 clap = { version = "4.0.9", features = ["derive"] }
 clap_mangen = "0.2.2"
+dirs-next = "2.0.0"
 iroh = { path = "../iroh" }
+which = "4.3.0"
--- a/xtask/Cargo.toml
+++ b/xtask/Cargo.toml
@@ -12,4 +12,3 @@ clap = { version = "4.0.9", features = ["derive"] }
 clap_mangen = "0.2.2"
 dirs-next = "2.0.0"
 iroh = { path = "../iroh" }
-which = "4.3.0"
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -59,13 +59,13 @@ fn dev_install(build: bool) -> Result<()> {
     if build {
         dist().unwrap();
     }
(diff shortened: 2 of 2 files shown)


xtask/Cargo.toml
```diff
@@ -2,12 +2,14 @@
 name = "xtask"
 version = "0.1.0"
 edition = "2021"
-private = true
+publish = false
```
dignifiedquire, line 5: good catch

xtask/src/main.rs
```diff
@@ -43,13 +49,28 @@ fn dist() -> Result<()> {
...
 
     Ok(())
 }
 
-fn dist_binary() -> Result<()> {
+fn dev_install(build: bool) -> Result<()> {
+    if build {
+        dist().unwrap();
+    }
+    let bins = vec!["iroh", "iroh-one", "iroh-gateway", "iroh-p2p", "iroh-store"];
```
dignifiedquire, line 62: could be an array (just remove the `vec!`)

xtask/src/main.rs
```diff
@@ -43,13 +49,28 @@ fn dist() -> Result<()> {
...
+    let home = dirs_next::home_dir().unwrap();
+    for bin in bins {
+        let from = project_root().join(format!("target/release/{}", bin));
+        let to = home.join(format!(".cargo/bin/{}", bin));
+        println!("copying {} to {}", bin, to.to_str().unwrap());
```
dignifiedquire, line 67: no unwrap please, `to.display()` is the way to go

xtask/src/main.rs
```diff
@@ -43,13 +49,28 @@ fn dist() -> Result<()> {
...
+        fs::copy(from, to).unwrap();
```
dignifiedquire, line 68: no `unwrap`, use the `?`

xtask/src/main.rs
```diff
@@ -64,12 +85,7 @@ fn dist_binary() -> Result<()> {
 
     fs::copy(&dst, dist_dir().join("iroh"))?;
 
-    if Command::new("strip")
-        .arg("--version")
-        .stdout(Stdio::null())
-        .status()
-        .is_ok()
-    {
+    if which::which("strip").is_ok() {
```
dignifiedquire, line 88: I missed it, do we really need `strip`?

b5: nah probably not. I'll remove

Arqu (approved):
Thx, this is neat

PR #348, opened 2022-10-14
docs: add full help text for `lookup`, `connect`, `get`, `p2p`, and `iroh` commands

ramfox:


iroh/src/p2p.rs
```diff
@@ -5,16 +5,52 @@ use clap::{Args, Subcommand};
...
+)]
 pub struct P2p {
     #[clap(subcommand)]
     command: P2pCommands,
 }
 
 #[derive(Subcommand, Debug, Clone)]
 pub enum P2pCommands {
+    #[clap(about = "Connect to a peer")]
+    #[clap(after_help = "
```
b5, line 21: I'd prefer we move these into constants defined in a separate file, like `help.rs`. Makes it harder to follow the code with these large breaks

b5 (approved):

PR #359, opened 2022-10-17
fix: make trace_id visible only if in use

Arqu:


iroh-gateway/src/error.rs
```diff
@@ -33,10 +33,14 @@ impl GatewayError {
 impl IntoResponse for GatewayError {
     fn into_response(self) -> Response {
         let mut headers = HeaderMap::new();
-        headers.insert(
-            &HEADER_X_TRACE_ID,
-            HeaderValue::from_str(&self.trace_id).unwrap(),
-        );
+        let has_trace_id =
+            !self.trace_id.is_empty() && self.trace_id != "00000000000000000000000000000000";
```
rklaehn, line 37: So why is this a string? We don't want to expose the TraceId type from OpenTelemetry?

It seems that TraceId has a public API to convert it to and from [u8; 16]. Maybe use that for traceid? Then we don't have to do string comparisons here.

iroh-gateway/src/error.rs
```diff
@@ -45,12 +49,19 @@ impl IntoResponse for GatewayError {
                 rb.body(BoxBody::default()).unwrap()
             }
             _ => {
-                let body = axum::Json(json!({
-                    "code": self.status_code.as_u16(),
-                    "success": false,
-                    "message": self.message,
-                    "trace_id": self.trace_id,
-                }));
+                let body = match has_trace_id {
```
rklaehn, line 52: Why a match instead of a if? Does clippy allow that?

rklaehn (approved):
Looks fine. Just added some questions.

iroh-gateway/src/error.rs
Arqu: Now that I look at it it is a bit odd for a boolean flag (was something else before though) and idk I just got used to the `match` magic :)

But yeah clippy is happy with it. 📎

iroh-gateway/src/error.rs
Arqu: Only reason was not to pull opentelemetry directly into iroh-gateway cause at some point we wanted to have metrics be a compile time toggle. I guess no hard reason right now. And the dep can be bound to the feature anyways.

I'll fix it up before merge.

Arqu: Also we still have the dependency anyways so duh...
Nice catch ty.

iroh-gateway/src/error.rs
rklaehn: It's not wrong, and also not difficult to read, just unusual. So 🤷‍♂️

PR #360, opened 2022-10-18
iroh start & stop commands

### LINKED ISSUE #177: when starting check for other instances

### LINKED ISSUE #290: iroh start command

### LINKED ISSUE #308: iroh stop command

b5:
supersedes #340.  closes #290. closes #308, closes #177

re-re-re worked start & stop commands that effectively do this on start:

```
#/bin/bash
nohup iroh-store > /dev/null &
nohup iroh-p2p > /dev/null &
nohup iroh-gateway > /dev/null &
```
all daemons now use lockfiles, and all lockfiles contain the PID of the owning process. `iroh stop` uses this PID to send `SIGINT` to each process. `iroh stop` also makes a best effort to remove stray lockfiles.

TODO:
* [x] docs strings
* [ ] windows support (in a follow-up PR)
* [ ] smoke test on linux
* [ ] tests, really of any kind
* [ ] confirm this isn't breaking android builds

cc @fabricedesre, this modifies lockfiles, writing the PID to the lockfile itself. I'd like to confirm you're ok with these changes. I'd also like to propose we exit with a non-zero status when a new process can't acquire the lock.


iroh-gateway/src/main.rs
```diff
@@ -10,12 +10,21 @@ use iroh_gateway::{
...
 use iroh_util::{iroh_config_path, make_config};
 use tokio::sync::RwLock;
 use tracing::{debug, error};
 
 #[tokio::main(flavor = "multi_thread")]
 async fn main() -> Result<()> {
+    let mut lock = ProgramLock::new("iroh-gateway")?;
+    if lock.is_locked() {
+        println!("iroh-gateway is already running, stopping.");
+        return Ok(());
```
b5, line 23: I think this should exit with a non-zero status. and should print in a nice pretty yellow

iroh-util/src/lock.rs
```diff
@@ -37,21 +40,62 @@ impl ProgramLock {
...
         file_guard::lock(file, Lock::Exclusive, 0, 1)
             .map(|lock| self.lock = Some(lock))
             .map_err(|err| err.into())
     }
 }
 
+/// Attempt to remove a stray lock file that wasn't cleaned up, returns true
+/// if a lock is successfully deleted, and will only attempt to delete if the
+/// lock is not currently held
+pub fn try_cleanup_dead_lock(prog_name: &str) -> Result<bool> {
```
b5, line 56: This needs a test, and I need to review it when I'm not braindead, please scrutinize.

iroh-util/src/lock.rs
```diff
@@ -37,21 +40,62 @@ impl ProgramLock {
 
     /// Try to acquire a lock for this program.
     pub fn acquire(&mut self) -> Result<()> {
-        let file = Rc::new(File::create(&self.path)?);
+        let mut file = File::create(&self.path)?;
+        file.write_u32::<LittleEndian>(process::id())?;
```
b5, line 44: I've considered just writing the PID in plaintext so humans can read it too, which is a double edged sword. My experience in the past with making things like this unreadable as a user is, well, constant irritation. but on the upside it's slightly tougher for users to go poking around with these files in a way that can mess things up.

iroh-util/src/lock.rs
```diff
@@ -28,7 +31,7 @@ impl ProgramLock {
 
         // Even if we manage to lock the file this won't last since the drop implementation
         // of FileGuard releases the underlying lock.
-        if let Ok(file) = File::create(&self.path) {
+        if let Ok(file) = File::open(&self.path) {
```
b5, line 34: this was creating a bug where calling `is_locked` was overwriting the lock itself. I'm unclear on if that would actually break the lock, but it certainly overwrites the data within the file

localops/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+anyhow = "1.0"
+
+[target.'cfg(unix)'.dependencies]
+nix = { version = "0.25", features = ["signal", "process"]}
+
+[target.'cfg(macos)'.dependencies]
+nix = { version = "0.25", features = ["signal", "process"]}
```
b5, line 15: am I holding this right? I have no idea what is even going on

fabricedesre:
> cc @fabricedesre, this modifies lockfiles, writing the PID to the lockfile itself. I'd like to confirm you're ok with these changes. I'd also like to propose we exit with a non-zero status when a new process can't acquire the lock.

Sure, that works for me. I'm not a fan of the binary pid format, but not a hill I want to die on either. Exiting with a non-zero status totally makes sense.

b5:
@faassen I'd appreciate you smoke testing this on linux, iirc you have a fedora distro lying around, right? I'll also be testing on ubuntu

iroh-util/src/exitcodes.rs
```diff
@@ -0,0 +1,23 @@
+//! iroh exit codes
```
b5, line 1: could also move this into an errors package. I've just crammed it here for now, and don't think it's a big deal to move later once we start getting more rigorous about error categorization

faassen:
> @faassen I'd appreciate you smoke testing this on linux, iirc you have a fedora distro lying around, right?

Yes, Fedora is my main platform so will do some testing tomorrow. In fact this command will be handy when I do manual testing with the CLI.

iroh-gateway/src/main.rs
dignifiedquire: - `eprintln` to print to stderr
- https://doc.rust-lang.org/std/process/fn.exit.html for exiting

iroh-util/src/lock.rs
dignifiedquire: I would recommend just using `.to_string()` (and `parse` for reading). To write human readable pids, this is good for at least us when we debug things :)

localops/Cargo.toml
dignifiedquire: `unix` includes `macos` I believe, there is `linux` which does not

ramfox:
Finding two bugs:
1) If you start up one of the binaries individually, the respective lock is not getting cleaned up on shutdown.
2) This ignores `cfg` input, so you can't currently have more than one process running on a single machine. Which puts a big damper on local testing. (previously, you could just pass in a path to a custom config file & have the different binaries use different configuration to ensure no port/rpc conflicts & have multiple iroh setups running at the same time)

The second problem is a way bigger conversation: do we want folks to be able to run more than one setup on their machine, if so how do we want them to do it, where would non-default locks live, etc

localops/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "localops"
```
dignifiedquire, line 2: If this is not private, should be `iroh-localops` for now

dignifiedquire: can we for now just put this functionality into `iroh-util`?

localops/src/process.rs
```diff
@@ -0,0 +1,66 @@
...
+        "deamonizing processes is not supported on your operating system"
+    ))
+}
+
+#[cfg(any(target_os = "macos", target_os = "linux"))]
+fn daemonize_process(bin_path: PathBuf) -> Result<()> {
+    // ¯\_(ツ)_/¯
+    let status = Command::new("bash")
+        .arg("-c")
+        // TODO(b5): might be nice to capture output in a log file at some point?
```
dignifiedquire, line 26: Yes, please make an issue, this is unusable to debug anything atm, so this should take some config of where to put the log file

iroh/src/services.rs
```diff
@@ -1,10 +1,109 @@
...
+        print!("starting {}...\t", &daemon_name.bold());
+        localops::process::daemonize(bin_path)?;
+        println!("{}", "success".green());
+        // TODO - confirm communication with RPC API
+    }
+
+    Ok(())
+}
+
+// TODO(b5) - in an ideal world the lock files would contain PIDs of daemon processes
```
dignifiedquire, line 75: isn't this how it works now?

iroh/src/services.rs
```diff
@@ -1,10 +1,109 @@
...
+    });
+
+    // construct a new set from the intersection of missing & expected services
+    let missing_services: HashSet<&str> = services
+        .into_iter()
+        .filter(|&service| missing_services.contains(service))
+        .collect();
+
+    if missing_services.is_empty() {
+        return Err(anyhow!(
```
dignifiedquire, line 49: This should not be an error

localops/src/process.rs
```diff
@@ -0,0 +1,66 @@
...
+
+#[cfg(not(any(target_os = "macos", target_os = "linux", target_os = "windows")))]
+fn stop_process(pid: u32) -> Result<()> {
+    Err(anyhow!(
+        "stopping processes is not supported on your operating system"
+    ))
+}
+
+#[cfg(any(target_os = "macos", target_os = "linux"))]
+fn stop_process(pid: u32) -> Result<()> {
```
dignifiedquire, line 58: please make all apis that use `u32`s currently, use a new type `Pid`, that looks sth like this

```rust
#[cfg(unix)]
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub struct Pid(nix::Pid);
#[cfg(not(unix))]
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub struct Pid; // TODO: fill in for each plattform when supported

#[cfg(unix)]
impl From nix::Pid for Pid {
 // .. 
}

impl std::fmt::Display for Pid {
 // ..
}
```

iroh-gateway/src/main.rs
```diff
@@ -10,12 +10,21 @@ use iroh_gateway::{
...
 use iroh_util::{iroh_config_path, make_config};
 use tokio::sync::RwLock;
 use tracing::{debug, error};
 
 #[tokio::main(flavor = "multi_thread")]
 async fn main() -> Result<()> {
+    let mut lock = ProgramLock::new("iroh-gateway")?;
+    if lock.is_locked() {
+        println!("iroh-gateway is already running, stopping.");
+        ::std::process::exit(iroh_util::exitcodes::LOCKED);
```
dignifiedquire, line 23: no `::` in front needed

iroh-gateway/src/main.rs
```diff
@@ -10,12 +10,21 @@ use iroh_gateway::{
...
+use iroh_util::lock::ProgramLock;
 use iroh_util::{iroh_config_path, make_config};
 use tokio::sync::RwLock;
 use tracing::{debug, error};
 
 #[tokio::main(flavor = "multi_thread")]
 async fn main() -> Result<()> {
+    let mut lock = ProgramLock::new("iroh-gateway")?;
+    if lock.is_locked() {
+        println!("iroh-gateway is already running, stopping.");
```
dignifiedquire, line 22: should be `eprintln`

iroh-p2p/src/main.rs
```diff
@@ -2,12 +2,21 @@ use anyhow::{anyhow, Context, Result};
...
 use iroh_util::{iroh_config_path, make_config};
 use tokio::task;
 use tracing::{debug, error};
 
 /// Starts daemon process
 fn main() -> Result<()> {
+    let mut lock = ProgramLock::new("iroh-p2p")?;
+    if lock.is_locked() {
+        println!("iroh-one is already running, stopping.");
+        ::std::process::exit(iroh_util::exitcodes::LOCKED);
```
dignifiedquire, line 15: as above

iroh-rpc-client/src/status.rs
```diff
@@ -164,6 +164,8 @@ pub struct StatusTable {
     pub gateway: StatusRow,
     pub p2p: StatusRow,
     pub store: StatusRow,
+
+    iter: u8,
```
dignifiedquire, line 168: Use a `usize`, faster and we are not really strapped for size here

iroh-rpc-client/src/status.rs
```diff
@@ -196,12 +199,32 @@ impl StatusTable {
     }
 }
 
+impl Iterator for StatusTable {
```
dignifiedquire, line 202: This is a bit strange. The usual pattern is to create a new type that holds a reference to `StatusTable` as the iterator

iroh-store/src/main.rs
```diff
@@ -5,11 +5,20 @@ use iroh_store::{
...
+use iroh_util::lock::ProgramLock;
 use iroh_util::{block_until_sigint, iroh_config_path, make_config};
 use tracing::{debug, error, info};
 
 #[tokio::main(flavor = "multi_thread")]
 async fn main() -> anyhow::Result<()> {
+    let mut lock = ProgramLock::new("iroh-store")?;
+    if lock.is_locked() {
+        println!("iroh-store is already running, stopping.");
+        ::std::process::exit(iroh_util::exitcodes::LOCKED);
```
dignifiedquire, line 17: as above

iroh-util/src/lock.rs
dignifiedquire: Should use https://doc.rust-lang.org/std/fs/struct.OpenOptions.html to specify to only create if not exists probably

iroh-util/src/lock.rs
```diff
@@ -37,21 +41,64 @@ impl ProgramLock {
...
+        path: crate::iroh_data_path(&format!("{}.lock", prog_name))?,
+        lock: None,
+    };
+    if lock.is_locked() {
+        info!("lock {} is currently active, cannot remove", prog_name);
+        return Ok(false);
+    }
+    match std::fs::remove_file(lock.path) {
+        Err(e) => {
+            info!("error removing {} lockfile: {}", prog_name, e);
```
dignifiedquire, line 68: why does this return `Ok`?

iroh-util/src/lock.rs
b5: agreed, change should be in already

localops/Cargo.toml
b5: I'd heavily prefer `iroh-localops`, even just for the amount of conditional compilation that needs to go into this crate. Most of this work is about vetting dependencies for cross-platform compatibility

localops/src/process.rs
b5: how about we default to `$daemon-name.log` in the application cache directory for now?

iroh-rpc-client/src/status.rs
b5: in2it, will change

iroh-util/src/exitcodes.rs
```diff
@@ -0,0 +1,23 @@
...
+//!
+//! # Example:
+//! ```
+//! extern crate exitcode;
+//!
+//! ::std::process::exit(exitcode::OK);
+//! ```
+
+/// Alias for the numeric type that holds system exit codes.
+pub type ExitCode = i32;
```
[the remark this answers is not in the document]
b5: From that page:

> ExitCode is intended to be consumed only by the standard library (via [Termination::report()](https://doc.rust-lang.org/std/process/trait.Termination.html#tymethod.report)), and intentionally does not provide accessors like PartialEq, Eq, or Hash. Instead the standard library provides the canonical SUCCESS and FAILURE exit codes as well as From<u8> for ExitCode for constructing other arbitrary exit codes.

the definition is irritating:

```rust
pub struct ExitCode(imp::ExitCode);
```

this means you can't define a nonstandard exit code as a constant. I'd argue that `std` authors are pointing us in the right direction: these are _iroh_ exit codes, and should have their own type.

I'm going to rename the type to `IrohExitCode` to further disambiguate, and await your clarification on why this isn't the right move 😄

localops/Cargo.toml
dignifiedquire: how does that change when putting it in utils? we already have a lot of crates :/

localops/src/process.rs
dignifiedquire: aren‘t there any defaults for log files?

iroh-util/src/exitcodes.rs
```diff
@@ -0,0 +1,16 @@
+//! iroh exit codes
+//!
+//! Exit code constants intended to be passed to
+//! `std::process::exit()`
```
faassen, line 4: Just a note: in the `iroh` CLI, clap does the exiting itself. I know it does a non-zero exit code when things are wrong, and since we don't have more information about errors at that level at the moment that's good enough for now.

iroh-util/src/lock.rs
```diff
@@ -28,52 +35,143 @@ impl ProgramLock {
...
+            Err(anyhow!("removing dead lockfile: {}", e))
```
faassen, line 77: I think seeing something like "Removing dead lockfile" as an error message may be a bit confusing, depending on how it's rendered. I suggest "could not remove dead lockfile: "

iroh-util/src/lock.rs
```diff
@@ -28,52 +35,143 @@ impl ProgramLock {
...
-            path: PathBuf::new().join(name),
+            path: PathBuf::from(name),
             lock: None,
         }
     }
 
+    #[test]
+    fn test_corrupt_lock() {
+        let path = PathBuf::from("lock.lock");
+        let mut f = File::create(&path).unwrap();
```
faassen, line 135: I'm confused where this stores the file it creates. Does this get created in cwd of the test runner?

I've used the `tempdir` crate https://crates.io/crates/tempdir to ensure I have a safe temporary directory for writing files during testing that gets cleaned up after the test completes.

I see that this pattern is already established in other tests so I guess it hasn't hurt things yet.

iroh-util/src/lock.rs
```diff
@@ -28,52 +35,143 @@ impl ProgramLock {
...
+    }
+
     #[test]
     fn test_locks() {
         use nix::unistd::{fork, ForkResult::*};
         use std::io::{Read, Write};
         use std::time::Duration;
 
-        // Start we no lock file.
+        // Start with no lock file.
```
faassen, line 155: I think with a tempdir this line wouldn't be needed as you'd always start with a clean directory before the test, as long as you make sure you only create files in that directory.

iroh/src/doc.rs
```diff
@@ -9,31 +9,58 @@ of the protocol.
...
+like add or get content, and `iroh start` is the fastest way to get iroh up &
+running locally. Once running, stop iroh with `iroh stop`.
+
+Use the start, stop, and status commands to monitor iroh on your local machine,
+and control it's uptime.
+
+Daemons provide 'services' like storage, peer-2-peer networking, and a gateway 
+to bridge IPFS to HTTP. Services work together to fullfill API requests. In this
+v0.1.0 release iroh is hard-coded to start three deamons: iroh-gateway, 
+iroh-p2p, and iroh-store. Each daemon provides gateay, p2p, and store services,
```
faassen, line 24: gateay -> gateway

iroh/src/doc.rs
```diff
@@ -9,31 +9,58 @@ of the protocol.
...
+respectively. To learn more about each service, see:
+  https://iroh.computer/docs/services
+
+Iroh start is a simple command, under the hood it just kicks off daemons to run 
+in the background. On startup each deamon writes it's process identifier (PID)
```
faassen, line 29: "it's" -> "its"

iroh/src/doc.rs
```diff
@@ -9,31 +9,58 @@ of the protocol.
...
-If iroh is online, status also reports the service configuration of the
-long running process, including the health of the configured subsystem(s).
-Possible configurations fall into two buckets:
-
-  one:     Iroh is running with all services bundled into one single process,
-           this setup is common in desktop enviornments.
-
-  cloud:   Iroh is running with services split into separate processes, which
-           are speaking to each other via remote procedure calls.
+Service status can be in one of four states:
```
faassen, line 53: I'd recommend "one of these states" instead of "four", because inevitably the amount of states is going to change in the future and the count will be off and someone won't update.

iroh/src/doc.rs
```diff
@@ -9,31 +9,58 @@ of the protocol.
...
+
+  Down:         The service is not currently running, or is 
+                not configured to connect to the proper port.
+                
+  Serving:      The service is running & healthy
+  
+  Not Serving:  The service is running, but unhealthy.
```
faassen, line 60: "unhealthy" is a bit ambiguous here - does that mean it's responding but not always, or incorrectly, or not responding at all? "but not responding" if it's the latter.

iroh/src/run.rs
```diff
@@ -62,6 +54,20 @@ enum Commands {
         /// filesystem path to write to. Optional and defaults to $CID
         output: Option<PathBuf>,
     },
+    #[clap(about = "Start iroh services locally")]
+    #[clap(after_help = doc::START_LONG_DESCRIPTION )]
+    Start {},
+    // status checks the health of the different processes
```
faassen, line 60: I see it was already two `//` before it got moved, but if we want this to be in the CLI help we'll need `///`.

iroh/src/services.rs
```diff
@@ -1,10 +1,124 @@
...
+
+// TODO(b5) - should check for configuration mismatch between iroh CLI configuration
+async fn start_services(api: &impl Api, services: HashSet<&str>) -> Result<()> {
+    // check for any running iroh services
+    let table = api.check().await;
+
+    let mut missing_services = HashSet::new();
+    let missing_services = table
+        .iter()
+        .fold(&mut missing_services, |accum, status_row| {
```
faassen, line 28: This may be lack of Rust experience, but this use of `fold` surprised me as I'd have expected the accumulator to be treated as immutable. Wouldn't a simple `for` loop be more easy to read for what's really repeated mutation of `missing_services`?

faassen (changes requested):

localops/src/process.rs
b5: this PR now writes to log files

localops/src/process.rs
b5: I've marked this as a TODO for later

b5:
yet another heads-up @fabricedesre, getting a cross-platform OS-backed file lock that also contains content proved a bridge too far. This PR keeps similar functionality, but uses a plain text file with the PID of the process with the lock. Programs that honor the lock use the [`sysinfo` crate](https://crates.io/crates/sysinfo) to check if the listed process is active, and if so, consider the lock held. I've kept the tests semantically the same to confirm.

I went with `sysinfo` because it has supports numerous operating systems, including Android. Lmk if this works.

fabricedesre:
Thanks for the heads up @b5 ! Nothing to worry on my side anyway since Android has its own init system that tracks the process lifecycle. If that conflicts with iroh's locking code I'll just remove it from iroh-one in my fork.

iroh-util/src/lock.rs
```diff
@@ -1,79 +1,300 @@
...
+pub enum LockError {
+    // lock present when one is not expected
+    #[error("Locked")]
+    Locked(PathBuf),
+    #[error("No lock file at {0}")]
+    NoLock(PathBuf),
+    /// Failure to parse contents of lock file
+    #[error("Corrupt lock file contents at {0}")]
+    CorruptLock(PathBuf),
+    #[error("Cannot detrmine status of process holding this lock")]
```
ramfox, line 231: detrmine -> determine

ramfox (changes requested):
mac bug:
`iroh start` says "starting" and "success", but doesn't seem to create any lock files or launch any processes.

the individual binaries create and remove their lock files appropriately

if I start the individual binaries up, I can use `iroh stop` to shut them down

`iroh stop` also seems to have 4 output lines, but should only be reporting about 3 processes.

iroh-localops/src/process.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+    Ok(())
+}
+
+#[cfg(target_os = "windows")]
+fn daemonize_process(bin_path: PathBuf, log_path: PathBuf) -> Result<()> {
+    Err(anyhow!("daemonizing processes on windows is not supported"))
+}
+
+pub fn stop(pid: i32) -> Result<()> {
```
dignifiedquire, line 69: for later: you don't really need this indirection, you can just `cfg` the public method

iroh-rpc-client/src/status.rs
```diff
@@ -196,6 +203,30 @@ impl StatusTable {
...
+        let current = match self.iter {
+            0 => Some(self.table.store.to_owned()),
+            1 => Some(self.table.p2p.to_owned()),
+            2 => Some(self.table.gateway.to_owned()),
+            _ => None,
+        };
+
+        self.iter += 1;
+        if current == None {
+            self.iter = 0;
```
dignifiedquire, line 224: is it on purpose that this iterator is infinite?

dignifiedquire: actually it doesn't seem to be, I am confused about this logic here. what are you trying to do?

iroh-util/src/lock.rs
```diff
@@ -1,79 +1,304 @@
...
-use std::rc::Rc;
+use sysinfo::{Pid, ProcessExt, ProcessStatus, System, SystemExt};
+use thiserror::Error;
+use tracing::warn;
 
-/// Manages a lock file used to track if an iroh program
-/// is already running.
-/// The lock is released either when the object is dropped
-/// or when the program stops.
+pub fn acquire_or_exit(lock: &mut ProgramLock) -> &mut ProgramLock {
```
dignifiedquire, line 11: this should be a method on `ProgramLock`

iroh-util/src/lock.rs
```diff
@@ -1,79 +1,304 @@
...
+    match lock.is_locked() {
+        Ok(false) => {
+            if let Err(e) = lock.acquire() {
+                eprintln!("error locking {}: {}", lock.program_name(), e);
+                std::process::exit(crate::exitcodes::ERROR);
```
dignifiedquire, line 16: please import things, and don't use the full paths, eg `process::exit(exitcodes::Error)` would be the main convention here

iroh-rpc-client/src/status.rs
b5: I mean yes, but now I realize that that doesn't make sense b/c why would you re-use iterators in Rust. I'll remove this reset entirely

iroh-util/src/lock.rs
dignifiedquire: I am confused, why does this exist here and down below again?

iroh-util/src/lock.rs
```diff
@@ -1,79 +1,304 @@
...
+        &self.path
+    }
+
+    pub fn program_name(&self) -> &str {
+        self.path
+            .file_name()
+            .unwrap_or_else(|| std::ffi::OsStr::new(""))
+            .to_str()
+            .unwrap()
+            .split('.')
```
dignifiedquire, line 83: Why do you need to split on `.` here?

iroh-util/src/lock.rs
b5: tried a few variations, and couldn't get the runtime to hang onto the reference at the call site:

what I wanted:
```rust
fn main() {
  let lock = ProgramLock::new("iroh-gateway")?.acquire_or_exit();
```
which caused the compiler to yell about an unused variable. changing `lock` -> `_lock` means the lock is `Drop`d right away, an makes the lock not work at all

iroh-util/src/lock.rs
b5: `.lock` extension in the file name. There's a forthcoming method in rust nightly that'll make this much cleaner

iroh-util/src/lock.rs
```diff
@@ -1,79 +1,304 @@
...
+    let this_pid = sysinfo::get_current_pid().unwrap();
+    if pid == this_pid {
+        return Ok(true);
+    }
+
+    // TODO(b5) - docs say we shouldn't be allocating on each call like this:
+    // https://docs.rs/sysinfo/0.26.5/sysinfo/index.html#usage
+    // I'm suspicious this pattern might be alright. Seems the underlying lib
+    // doesn't do much hydrating, but System may be a large memory allocation
+    let mut s = System::new();
```
dignifiedquire, line 173: `process_is_running` should just be a method on `ProgrammLock`, which can lazily store a `System`

iroh-util/src/lock.rs
```diff
@@ -1,79 +1,304 @@
...
+                ProcessStatus::Waking => Ok(true),
+
+                ProcessStatus::Stop => Ok(false),
+                ProcessStatus::Zombie => Ok(false),
+                ProcessStatus::Tracing => Ok(false),
+                ProcessStatus::Dead => Ok(false),
+                ProcessStatus::Wakekill => Ok(false),
+                ProcessStatus::Parked => Ok(false),
+                ProcessStatus::LockBlocked => Ok(false),
+                ProcessStatus::Unknown(_s) => Ok(false),
```
dignifiedquire, line 194: maybe this would be a little clearer

```rust
use ProcessStatus::*;
let is_running = matches!(process.status(), Idle | Run | Sleep | Waking);
```

iroh-util/src/lock.rs
dignifiedquire: `tempdir` is deprecated, but please use `tempfile::Tempdir` for all testing directory uses

iroh-util/src/lock.rs
```diff
@@ -1,79 +1,304 @@
...
+
     #[test]
     fn test_locks() {
         use nix::unistd::{fork, ForkResult::*};
         use std::io::{Read, Write};
         use std::time::Duration;
 
-        // Start we no lock file.
+        // Start with no lock file.
         let _ = std::fs::remove_file("test1.lock");
```
dignifiedquire, line 283: here as well, please use a temp directory

iroh/src/services.rs
```diff
@@ -0,0 +1,341 @@
...
+use iroh_api::{Api, ServiceStatus, StatusRow, StatusTable};
+use iroh_util::lock::{LockError, ProgramLock};
+
+const SERVICE_START_TIMEOUT_SECONDS: u64 = 15;
+
+/// start any of {iroh-gateway,iroh-store,iroh-p2p} that aren't currently
+/// running.
+pub async fn start(api: &impl Api) -> Result<()> {
+    // start_services(api, HashSet::from(["store"])).await
+    start_services(api, HashSet::from(["store", "p2p", "gateway"])).await
```
dignifiedquire, line 20: these should really not be strings but an enum

iroh/src/services.rs
```diff
@@ -0,0 +1,341 @@
...
+                iroh_api::ServiceStatus::Down(_reason) => {
+                    accum.insert(status_row.name());
+                    // TODO(b5) - warn user that a service is down & exit
+                }
+            }
+            accum
+        });
+
+    // construct a new set from the intersection of missing & expected services
+    let missing_services: HashSet<&str> = services
```
dignifiedquire, line 50: `HashSet::difference` is your friend

iroh/src/services.rs
```diff
@@ -0,0 +1,341 @@
...
+
+    if missing_services.is_empty() {
+        println!(
+            "{}",
+            "All iroh daemons are already running. all systems nominal.".green()
+        );
+        return Ok(());
+    }
+
+    for &service in missing_services.iter() {
```
dignifiedquire, line 63: `for service in &missing_services`

iroh/src/services.rs
```diff
@@ -0,0 +1,341 @@
...
+            );
+        }
+    }
+
+    Ok(())
+}
+
+/// stop all services by sending SIGINT to any active daemons identified
+/// by lockfiles
+pub async fn stop(api: &impl Api) -> Result<()> {
```
dignifiedquire, line 100: shouldn't this take a configuration of services as well?

iroh-util/src/lock.rs
b5: this is more complicated than it seems. I'll make an issue & address in a follow up

iroh-util/src/lock.rs
b5: 🤦 forgot to remove this

ramfox (approved):

iroh-util/src/lock.rs
dignifiedquire: > lock -> _lock means the lock is Dropd right away,

That is not true, this is how drop guards work, they are kept for the current scope. It is only dropped immediately if it is `_`.

dignifiedquire (approved):

PR #366, opened 2022-10-18
fix: ensure that the root data path exists before creating lock files

fabricedesre:
WIthout that fix, if `$HOME/.local/share/iroh` doesn't exist yet the lock creation fails.

ping @b5 since that may conflict with your work on start/stop.


b5:
considering it's a one-line fix, want me to incorporate this into #360?

fabricedesre:
> considering it's a one-line fix, want me to incorporate this into #360?

yes, wfm.

b5:
superseded by #360

Issue #353, opened 2022-10-17
Sync provider records from store to p2p


rklaehn:
This is a bit tricky since we don't know if storage or p2p will start up first. It will also introduce some coupling between the two services. But another solution is not quickly possible due to how rust-libp2p works. It requires all records to be available in an infallible iterator.

rklaehn:
So pull all records from store to p2p on store or p2p startup?

Store would have to notify p2p somehow that it has started so p2p can start pulling... And we would have to extend the rpc interface.

PR #367, opened 2022-10-19
Provide records from store

rklaehn:
- add store rpc to get all cids that the store has data for
- add task in p2p to get these records in regular intervals
- publish them on the dht by just calling StartProviding

Questions:
- What are the implications of repeatedly calling StartProviding for something we are already providing

Implements https://github.com/n0-computer/iroh/issues/353

### DIFF
--- a/iroh-rpc-client/src/store.rs
+++ b/iroh-rpc-client/src/store.rs
@@ -5,6 +5,7 @@ use bytes::Bytes;
 use cid::Cid;
 #[cfg(feature = "grpc")]
 use futures::Stream;
+use futures::TryStreamExt;
 #[cfg(feature = "grpc")]
 use iroh_rpc_types::store::store_client::StoreClient as GrpcStoreClient;
 use iroh_rpc_types::store::{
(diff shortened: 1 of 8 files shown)


dignifiedquire:
Maybe we could have a `status` call that returns information like uptime/start time, that way it is both useful and should allow you to do the detection properly?

iroh-rpc-client/src/store.rs
```diff
@@ -83,4 +84,15 @@ impl StoreClient {
         let size = self.backend.get_size(req).await?.size;
         Ok(size)
     }
+
+    #[tracing::instrument(skip(self))]
+    pub async fn get_block_cids(&self) -> Result<impl Stream<Item = Result<Cid>>> {
```
dignifiedquire, line 89: Heads up, with tarpc we don't have streaming builtin so we will need to do something more manual here

rklaehn: Have we decided already that we do tarpc? I guess I could do my own paging by exposing the id to the interface.

rklaehn:
So we now detect when the store goes down and up again and sync immediately in that case.

I think with that change it is good to go at least in terms of what it does. Could use more tests, but I would do that (mocking the other services or whatever for tests) in another issue. Also, I think so much will change that I would not overdo it with the tests here.

rklaehn:
We are going to close this for now. Merging this would mean publishing every single cid in our store to the DHT, which is probably not something we want to do.

We will reconsider this once we have data on which CIDs to publish.

PR #371, opened 2022-10-19
[WIP] Integrate tarpc instead of grpc

dignifiedquire:
Experiment to see how tarpc behaves when used instead of grpc

- Tarpc does not provide load balancing out of the box but we shouldn't actually need it currently.
- The current config is very likely wrong.
- The current deadlines are set to 60s
- There is no streaming in tarpc, right now we actually don't need it due to the changes in bitswap, but we will need to build sth manual for that if we want to use this
- UDS support is not done yet, but it should be straightforward to add, there is integrated support being added to master@tarpc currently


iroh-rpc-client/src/macros.rs
```diff
@@ -2,88 +2,79 @@
...
-            pub(crate) const SERVICE_NAME: &str = stringify!([<$label:lower>].[<$label>]);
-
-            /// The display name that we expect to see in the StatusTable.
-            #[cfg(feature = "grpc")]
-            pub(crate) const NAME: &str = stringify!([<$label:lower>]);
-
             #[derive(Debug, Clone)]
             pub struct [<$label Client>] {
-                backend: [<$label ClientBackend>],
+                backend: std::sync::Arc<tokio::sync::RwLock<[<$label BackendState>]>>,
```
Arqu, line 7: This might give some contention.

iroh-rpc-client/src/store.rs
```diff
@@ -1,86 +1,64 @@
...
-#[cfg(feature = "grpc")]
-use tonic_health::proto::health_client::HealthClient;
-
-#[cfg(feature = "grpc")]
-use crate::status::{self, StatusRow};
+use tarpc::context::Context;
 
 impl_client!(Store);
 
+const DEFAULT_DEADLINE: Duration = Duration::from_secs(60);
```
Arqu, line 10: should bump above 2 min (the gw timeout)

iroh-rpc-client/src/macros.rs
dignifiedquire: I haven't found a different way to get lazy connection working, which we need. It should be fine though, as the fast path only ever takes a read lock, so once connected, there should be no contention

dignifiedquire:
Closing this PR, as we are likely not going to use tarpc.

PR #374, opened 2022-10-20
fix: move bitswap msg counters out of encode/decode

Arqu:


rklaehn (commented):
So what was wrong before?

Arqu:
> So what was wrong before?

The counters were in the encode/decode functions which skewed results as we encode a lot more than we actually send.

rklaehn:
> The counters were in the encode/decode functions which skewed results as we encode a lot more than we actually send.

But that is also valuable information... Anyway, looks good.

iroh-bitswap/src/lib.rs
```diff
@@ -287,6 +288,7 @@ impl<S: Store> Bitswap<S> {
 
     fn receive_message(&self, peer: PeerId, message: BitswapMessage) {
         inc!(BitswapMetrics::MessagesReceived);
+        record!(BitswapMetrics::MessageBytesIn, message.encoded_len() as u64);
```
dignifiedquire, line 291: this encodes the message again :/

iroh-bitswap/src/network.rs
```diff
@@ -123,6 +123,10 @@ impl Network {
             for i in 1..=retries {
                 debug!("send:{}: try {}/{}", peer, i, retries);
                 let (s, r) = oneshot::channel();
+                record!(
```
dignifiedquire, line 126: this might fail, so we don't actually know if these are the bytes we sent

Arqu: I'm aware, but would rather over then under count here ie we streamed half a response and channel broke.

iroh-bitswap/src/lib.rs
Arqu: Ik :( Want me to extend BitswapMessage with a pre calculated len?

iroh-bitswap/src/network.rs
dignifiedquire: fair

iroh-bitswap/src/lib.rs
dignifiedquire: maybe we could cache it?

Arqu:
> > The counters were in the encode/decode functions which skewed results as we encode a lot more than we actually send.
> 
> But that is also valuable information... Anyway, looks good.

https://github.com/n0-computer/iroh/issues/378

dignifiedquire (approved):
lets get it in and worry about improving that encoding issue later

PR #380, opened 2022-10-20
fix: bitswap sessions shutdown

dignifiedquire:


iroh-bitswap/src/client/provider_query_manager.rs
```diff
@@ -75,7 +75,7 @@ impl ProviderQueryManager {
         &self,
         cid: &Cid,
     ) -> Result<async_channel::Receiver<Result<PeerId>>> {
-        let (s, r) = async_channel::bounded(8);
+        let (s, r) = async_channel::bounded(10);
```
Arqu, line 78: Why does 10 help more, should it be configurable or at least extracted as a const?

dignifiedquire:
Confirmed on the teststream that this results on a much more stable session count.

iroh-bitswap/src/lib.rs
```diff
@@ -512,7 +512,7 @@ impl<S: Store> NetworkBehaviour for Bitswap<S> {
     ) -> Poll<NetworkBehaviourAction<Self::OutEvent, Self::ConnectionHandler>> {
         inc!(BitswapMetrics::NetworkBehaviourActionPollTick);
         // limit work
-        for _ in 0..100 {
+        for _ in 0..50 {
```
Arqu, line 515: let's extract these real quick as const and I can play with magic numbers eventually.

Arqu (dismissed):
LGTM, just some nits

iroh-bitswap/src/client/provider_query_manager.rs
dignifiedquire: I will make it a constant eventually, it's just to match the number of providers we are fetching, they now fit all into the channel.

iroh-bitswap/src/lib.rs
dignifiedquire: can I do this in another PR? I would like CI to finish today..

iroh-resolver/src/resolver.rs
```diff
@@ -693,13 +681,15 @@ impl LoaderContext {
...
+        let count = Arc::strong_count(&self.inner);
+        debug!("session {} dropping loader context {}", self.id, count);
+        if count == 1 {
             if let Err(err) = self
                 .inner
                 .try_lock()
                 .expect("last reference, no lock")
                 .closer
-                .try_send(self.id)
+                .send_blocking(self.id)
```
Arqu, line 692: I assume `drop` is called in a separate task? Blocking is scary with tokio, on connectivity issues this might block other tasks.

Arqu (approved):
Lgtm, have that single q

iroh-bitswap/src/lib.rs
Arqu: Tbh not that important, we got lot's of magic numbers I want to eventually extract and then play with configs and do regression testing against benchmarks... eventually.

iroh-resolver/src/resolver.rs
dignifiedquire: nope it actually blocks 😓  because we can't do any a sync work in drop, or tokio panics

PR #386, opened 2022-10-20
fix: windows symlink support

ramfox:


.cargo/config
```diff
@@ -1,2 +1,5 @@
 [alias]
-xtask = "run --package xtask --"
\ No newline at end of file
+xtask = "run --package xtask --"
+
+[target.x86_64-pc-windows-gnu]
+linker = "x86_64-w64-mingw32-gcc"
```
b5, line 5: can you enlighten me?

b5 (approved):
this is great

.cargo/config
rklaehn: This is so you can cross compile for windows on mac and I don't have to go looking for a windows box.

b5:
CI is giving us headaches. This is clearly moving the ball forward on windows. I'm going to merge without waiting for it to finish

PR #388, opened 2022-10-21
perf: Use the libipld built in way to scrape references

rklaehn:
...without having to create an AST. Should be much faster.

This has been working all the time for dag-cbor since we needed that at actyx, but I have also made the dag-pb case a bit faster in https://github.com/ipld/libipld/pull/155

### DIFF
--- a/iroh-resolver/src/resolver.rs
+++ b/iroh-resolver/src/resolver.rs
@@ -82,8 +82,12 @@ impl Block {
             return Err(InvalidMultihash(mh.to_bytes()).into());
         }
         // check that the links are complete
-        let links = parse_links(&self.cid, &self.data)?;
-        anyhow::ensure!(links == self.links, "links do not match");
+        let expected_links = parse_links(&self.cid, &self.data)?;
+        let mut actual_links = self.links.clone();
+        actual_links.sort();
+        // TODO: why do the actual links need to be deduplicated?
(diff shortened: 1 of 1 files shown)


iroh-resolver/src/resolver.rs
```diff
@@ -1400,18 +1400,16 @@ impl<T: ContentLoader> Resolver<T> {
 /// Extract links from the given content.
 pub fn parse_links(cid: &Cid, bytes: &[u8]) -> Result<Vec<Cid>> {
     let codec = Codec::try_from(cid.codec()).context("unknown codec")?;
+    let mut cids = BTreeSet::new();
```
rklaehn, line 1403: You might argue that a hashset would be faster here, and you would not be wrong. I guess it is my determinism fetish...

dignifiedquire (dismissed):
nice 💙

rklaehn:
Note that with this I am not entirely convinced that sending the cids over the RPC is worth it. The store can scrape the links. Unless we foresee a scenario where you would want to send different links than the block contains, but that would be weird...

rklaehn:
Oh well, another error. Made an issue for this since it is unrelated to this: https://github.com/n0-computer/iroh/issues/390

dignifiedquire (approved):

PR #389, opened 2022-10-21
Add dumb entry point

rklaehn:
This is not really a PR but more a request for discussion with some code attached.

It is **very** dumb. But it would save us from having to publish 4 or 5 binaries.

Instead of `iroh-gateway ...`, you do `iroh gateway ...` etc. Otherwise everything is the same.

Of course you have one fat binary that has stuff you might not need. But I doubt people will care for just trying it out,
and our time might be better used fixing bugs and working on perf than building and publishing so many binaries, at least for the 0.1.0 release...

Here are the sizes:
```
iroh on  rklaehn/iroh-entry-point [$?] via 🦀 v1.63.0 
❯ ls -l target/release/iroh*                            
-rwxr-xr-x  1 rklaehn  staff  42750722 Oct 21 15:29 target/release/iroh // < this would be the only thing we need to publish
-rwxr-xr-x  1 rklaehn  staff  15891990 Oct 21 15:28 target/release/iroh-cli
-rwxr-xr-x  1 rklaehn  staff  19373786 Oct 21 15:28 target/release/iroh-gateway
-rwxr-xr-x  1 rklaehn  staff  24792278 Oct 21 15:29 target/release/iroh-p2p
-rwxr-xr-x  1 rklaehn  staff  20634744 Oct 21 15:28 target/release/iroh-store
```

And the help output:
```
Usage: iroh [COMMAND]

Commands:
  gateway  IPFS gateway
  p2p      Implementation of the p2p part of iroh
  store    Implementation of the storage part of iroh
  cli      A next generation IPFS implementation: https://iroh.computer
  help     Print this message or the help of the given subcommand(s)

Options:
  -h, --help     Print help information
  -V, --version  Print version information
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ members = [
   "iroh-store",
   "iroh-share",
   "iroh-util",
+  "iroh-entry-point",
   "stores/*",
   "examples",
   "xtask"
--- a/iroh-entry-point/Cargo.toml
+++ b/iroh-entry-point/Cargo.toml
@@ -1,0 +1,24 @@
+[package]
+name = "iroh-entry-point"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+anyhow = "1.0.66"
+clap = { version = "4.0.9", features = ["derive"] }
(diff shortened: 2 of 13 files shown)


b5:
after some lovely post-release sleep, we've come up with a solid direction for this.

PR #394, opened 2022-10-21
fix broken macos github action

b5:
found the source of the issue: the source is process forking on macOS after High Sierra. Solution: https://github.com/rails/rails/issues/38560#issuecomment-1159834445

Asmir, feel free to just merge this one upon approval


Arqu:
wat... 
this is a fun one...

Issue #381, opened 2022-10-20
`iroh p2p peers` - list of peers you are connected to


ramfox:
useful command, especially if we can differentiate between connections that the user has manually triggered & connections found via bootstrapping & in the "wild"

b5:
CLOSED. YUS.

b5:
linking to the PR would be nice: https://github.com/n0-computer/iroh/pull/397

it landed in v0.1.0

PR #397, opened 2022-10-22
`iroh p2p peers`: command to list peers

b5:
```
$ iroh p2p peers --help
List connected peers

Usage: iroh p2p peers

Options:
  -h, --help     Print help information
  -V, --version  Print version information


'p2p peers' lists the set of peer addresses this node is currently connected to.
The address format is a multiaddress, or 'multiaddr' for short. For example:

  /ip4/104.131.131.82/tcp/4001/p2p/QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ

The last element after the '/' is the Peer Identifier or 'PeerID'. Either the
PeerID or the entire multiaddr can be given to the 'p2p lookup' command
for additional details about the peer. For example:

  > iroh p2p lookup QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ
  > iroh p2p lookup /ip4/104.131.131.82/tcp/4001/p2p/QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ

For more info on multiaddrs see https://iroh.computer/docs/concepts#multiaddr.
```

builds on #402

### DIFF
--- a/iroh-api/src/p2p.rs
+++ b/iroh-api/src/p2p.rs
@@ -1,4 +1,5 @@
 use crate::error::map_service_error;
+use std::collections::HashMap;
 use anyhow::Result;
 use async_trait::async_trait;
 use iroh_rpc_client::{Lookup, P2pClient};
@@ -28,6 +29,7 @@ pub trait P2p: Sync {
     async fn lookup_local(&self) -> Result<Lookup>;
     async fn lookup(&self, addr: &PeerIdOrAddr) -> Result<Lookup>;
     async fn connect(&self, addr: &PeerIdOrAddr) -> Result<()>;
+    async fn peers(&self) -> Result<HashMap<PeerId, Vec<Multiaddr>>>;
 }
 
 #[async_trait]
@@ -68,6 +70,10 @@ impl P2p for ClientP2p {
             }
         }
         .map_err(|e| map_service_error("p2p", e))
+    }
+
+    async fn peers(&self) -> Result<HashMap<PeerId, Vec<Multiaddr>>> {
+        self.client.get_peers().await
     }
 }
 
--- a/iroh-api/src/p2p.rs
+++ b/iroh-api/src/p2p.rs
@@ -1,11 +1,11 @@
 use crate::error::map_service_error;
-use std::collections::HashMap;
 use anyhow::Result;
 use async_trait::async_trait;
 use iroh_rpc_client::{Lookup, P2pClient};
 use libp2p::{multiaddr::Protocol, Multiaddr, PeerId};
 #[cfg(feature = "testing")]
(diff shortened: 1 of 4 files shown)


iroh/src/p2p.rs
```diff
@@ -88,3 +95,12 @@ observed addresses: {:?}
         l.peer_id, l.listen_addrs, l.protocols, l.protocol_version, l.observed_addrs
     );
 }
+
+fn display_peers(peers: HashMap<PeerId, Vec<Multiaddr>>) {
+    // let mut pid_str: String;
+    for (peer_id, addrs) in peers {
+        if let Some(addr) = addrs.first() {
```
b5, line 102: This is just grabbing the first address, which I'm doubtful is actually the address we're connected to. We should make sure the addresses listed here are actually connected

ramfox: We need a rule of thumb for this in `lookup` as well: some entries return 20+ addrs, and that's... not necessary.

For the lookup of the currect peer, we can be more comfortable just displaying the first 5 (or whatever number we choose), because according to the docs they rank the `external_addrs` that we can be dialed at, but like you mention, we don't have that luxury here, or for the other `lookup` commands.

But also, we may want to pair this down. Maybe you use `p2p peers` to just get a list of the peer ids (not peer_ids and addrs), and use `p2p lookup PEER_ID` to investigate possible addresses?

iroh-api/src/error.rs
```diff
@@ -0,0 +1,27 @@
+use std::io;
+use thiserror::Error as ThisError;
+// use std::error::Error;
```
ramfox, line 3: have you gotten the error talk from dig yet? don't want to get in trouble for approving `thiserror` usage on top of `anyhow` usage 🤣

ramfox (dismissed):
approving this, but left 2 comments you can take or leave!

iroh-api/src/p2p.rs
```diff
@@ -69,6 +71,13 @@ impl P2p for ClientP2p {
         }
         .map_err(|e| map_service_error("p2p", e))
     }
+
+    async fn peers(&self) -> Result<HashMap<PeerId, Vec<Multiaddr>>> {
```
rklaehn, line 75: I prefer BTreeMap, less random...

rklaehn (approved):

iroh-api/src/p2p.rs
b5: interesting. I'll make a note in a follow-up issue

ramfox:
closes #381

PR #403, opened 2022-10-24
race on gateway

b5:
tried to make this as not-terrible as possible, by requiring an array of gateways, and randomly choosing a gateway when more than one is supplied.

### DIFF
--- a/iroh-gateway/src/config.rs
+++ b/iroh-gateway/src/config.rs
@@ -213,7 +213,7 @@ mod tests {
         expect.insert("port".to_string(), Value::new(None, default.port as i64));
         expect.insert(
             "use_denylist".to_string(),
-            Value::new(None, default.denylist),
+            Value::new(None, default.use_denylist),
         );
         expect.insert(
             "headers".to_string(),
--- a/iroh-gateway/src/config.rs
+++ b/iroh-gateway/src/config.rs
@@ -27,7 +27,9 @@ pub struct Config {
     /// default port to listen on
     pub port: u16,
     /// flag to toggle whether the gateway should use denylist on requests
-    pub denylist: bool,
+    pub use_denylist: bool,
+    /// URL of gateways to be used by the racing resolver.
+    pub http_resolvers: Option<Vec<String>>,
     /// rpc addresses for the gateway & addresses for the rpc client to dial
     pub rpc_client: RpcClientConfig,
     /// metrics configuration
@@ -46,8 +48,9 @@ impl Config {
             headers: HeaderMap::new(),
             port,
             rpc_client,
+            http_resolvers: None,
             metrics: MetricsConfig::default(),
-            denylist: false,
+            use_denylist: false,
         }
     }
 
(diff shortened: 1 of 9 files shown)


iroh-gateway/src/config.rs
```diff
@@ -27,7 +27,9 @@ pub struct Config {
     /// default port to listen on
     pub port: u16,
     /// flag to toggle whether the gateway should use denylist on requests
-    pub denylist: bool,
+    pub use_denylist: bool,
```
b5, line 30: heads up @Arqu, I've added a `use_` prefix to avoid using up the `denylist` configuration key on a boolean flag.

iroh-resolver/src/racing.rs
```diff
@@ -1,36 +1,44 @@
...
     pub fn try_raw_gateway(&self) -> Result<&String> {
-        self.gateway
-            .as_ref()
-            .ok_or_else(|| anyhow!("no gateway configured to fetch raw CIDs"))
+        match self.http_resolvers.len() {
+            0 => Err(anyhow!("no gateway configured to fetch raw CIDs")),
+            1 => self
+                .http_resolvers
+                .first()
+                .ok_or_else(|| anyhow!("no gateway configured to fetch raw CIDs")),
```
rklaehn, line 35: How can this ever happen? We have already checked that there is an element.

Can we just call rand all the time? It's not going to make a difference.

iroh-resolver/src/racing.rs
```diff
@@ -42,7 +50,10 @@ impl RacingLoader {
     async fn fetch_http(&self, cid: &Cid) -> Result<(Bytes, String), anyhow::Error> {
         let gateway = self.try_raw_gateway()?;
         let cid_str = multibase::encode(multibase::Base::Base32Lower, cid.to_bytes().as_slice());
-        let gateway_url = format!("https://{}.ipfs.{}?format=raw", cid_str, gateway);
+        let mut gateway_url = format!("https://{}.ipfs.{}?format=raw", cid_str, gateway);
```
rklaehn, line 53:
```rust
let gateway_url = if gateway.starts_with("https://") || gateway.starts_with("http://") {
  format!("{}/ipfs/{}?format=raw", gateway, cid_str);
} else {
  format!("https://{}.ipfs.{}?format=raw", cid_str, gateway);
}
```
?

rklaehn (dismissed):

iroh-resolver/src/racing.rs
b5: oh that's much cleaner. ty!

iroh-resolver/src/racing.rs
b5: > How can this ever happen? We have already checked that there is an element.
yeah this is dumb. Will fix.

> Can we just call rand all the time? It's not going to make a difference.
I was unclear on the overhead of the allocation, and didn't have time to research. I'll trust that you know this is a super low cost &. I'll just call it all the time 😄

b5:
cc @fabricedesre, this PR moves the racing gateway functionality into the "main" gateway, and changes the configuration location.

iroh-resolver/src/racing.rs
rklaehn: almost everything in rust is an expression. So coming from FP I like htat very much...

rklaehn (dismissed):

PR #404, opened 2022-10-24
feat: connection pooling

Arqu:


iroh-rpc-client/src/client.rs
```diff
@@ -70,10 +123,11 @@ impl Client {
             .ok_or_else(|| anyhow!("missing rpc gateway connnection"))
     }
 
-    pub fn try_store(&self) -> Result<&StoreClient> {
-        self.store
-            .as_ref()
-            .ok_or_else(|| anyhow!("missing rpc store connnection"))
+    pub fn try_store(&self) -> Result<StoreClient> {
+        if let Some(store_client) = self.store.get() {
```
rklaehn, line 127: Same here:
```rust
        self.gateway
            .as_ref()
            .context("missing rpc gateway connnection")
```

rklaehn (approved):
Looks good. See 2 tiny nits in the comments.

I am hesitant to add so much complexity for inter process communication between processes on the same machine. I think in the longer term we have to redesign the API so that this is not needed. But the code looks good, so it might be a good bandaid for now.

rklaehn (approved):
Still looks fine as far as the code is concerned

iroh-rpc-client/src/client.rs
```diff
@@ -10,8 +13,56 @@ use crate::store::StoreClient;
...
 pub struct Client {
     pub gateway: Option<GatewayClient>,
-    pub p2p: Option<P2pClient>,
-    pub store: Option<StoreClient>,
+    p2p: P2pLBClient,
+    store: StoreLBClient,
+}
+
+#[derive(Debug, Clone)]
+pub struct StoreLBClient {
```
dignifiedquire, line 21: I can haz doc comment of what this does?

iroh-rpc-client/src/client.rs
```diff
@@ -10,8 +13,56 @@ use crate::store::StoreClient;
...
+    clients: Vec<StoreClient>,
+    pos: Arc<AtomicUsize>,
+}
+
+impl StoreLBClient {
+    pub fn get(&self) -> Option<StoreClient> {
```
dignifiedquire, line 27: maybe add a comment that this does round robin

dignifiedquire (dismissed):

b5 (approved):

PR #405, opened 2022-10-24
Port iroh-store to libmdbx

vorot93:


rklaehn:
Thanks a lot. We will definitely take a look at this after ipfs-camp.

Independent of performance, I am quite curious what this does to our compile times. rocksdb takes some time to compile...

rklaehn:
This would be a pretty big change, so I would first have to be convinced that this is _massively_ superior to what we currently have (rocksdb). We are not yet at a stage where we can do big changes for small gains.

I have been playing around with mdbx in a separate repo, and saw performance degrade after some time. I would need to understand more about mdbx perf characteristics before deciding on this.

https://github.com/rklaehn/mdbxtest/tree/master/src

I get that mdbx has the advantage that it directly updates the db instead of having a WAL. But I am not sure if having a WAL is a big disadvantage in our use case.

b5:
Apologies @vorot93, but even if the perf is better, we can't accept this PR in the end. The reasons have almost nothing to do with technology. 

We plan to bundle iroh into a SaaS product at some point in the future. Because of that, the number 0 team needs iroh to be able to survive a corporate vendor audit, and the history of Github censoring libmdbx will almost definitely raise concerns, regardless of their legitimacy. We need those conversations to go well, because they generate the revenue we need to sustain this project long term.

Do I think there's actually any problem with libmdbx? I highly doubt it. I haven't had a chance to audit the code myself, but we routinely rely on other people's code with even less attention.

So in short, this sucks. I'm sorry to have to reject code on anything other than the merits of the work, and in this case the merits of the work seem very good indeed, which makes a decision like this all the more painful.

I really appreciate the time you've put into this contribution & others @vorot93. I still very much hope we can make iroh work for your use case, but I understand how my choice to prioritize an easier sales story may turn you off of such an idea. Please let me know if there's anything I can do to help, or any detail you'd like us to elaborate on.

PR #406, opened 2022-10-24
feat: provide only root, and only at the end of add

rklaehn:
We are still waiting for the start_provide of the root to complete


rklaehn:
Added progress for the DHT update. And it is really easy to switch from "provide just root" to "provide all"

b5 (approved):
I've hijacked this PR to add a few important additional things:
1. Add now requires p2p service is running. We need to get p2p load-on-start is for and for root CIDs only with this change. An early check confirms the service is up before proceeding.
2. That check warns the user with the specific service that's missing & a hint to run `iroh start $SERVICE`
3. I've added an `--all` flag to iroh start, based on issues asmir was having with running quick benchmarks, and p2p not being part of the default set of services

PR #408, opened 2022-10-25
fix(resolver): ipld, double-fetching, and trailing slash fixes

### LINKED ISSUE #207: Couldn't determine unixfs type
We still get some of these.
- `bafkreibwy56nhaix5vdyykmfcua5mmn26h5jza3ssr2b3krk6jasa2orfe`
- `bafkreiby6vd24facatofpvuxukwolotd2nnovquosoljk35fcstpvet7fy`
- `bafybeieb3ugnsjiauca33by6c27wf5uvbi3ec27svy62kvizf7e3tpctda`
- `bafkreicauoyhtm2os2kat7m7alfxlfdcyktth46a4lq77grlrtqtkhvbzm`
- `bafkreidofrgcvxcn4pgf7xmm65r7p2umovpyzd4jhplw4ud2bjyxqvl3ji`
- `bafkreidn2izx2c7s237ew6j66s6cmce7xadafk6qhb2rrye47tj5mvpoaa`
- `bafkreiczzdczgjgqwl56hao6zxy22wvrtwne62ver3q6uhsh2ppmwvgef4`
- `bafkreigfqffhqdcvnnwkrmidovtlckimkpuysstbxzv7j2b37ljf5ww5uu`
- `bafkreifw3yms46iqm5hhplawo4rwzm7rzrv7s3ftjcz7i37wlknraj3b4q`
- `bafkreifyhpsoa4wnc4ecwe4gzkwmfp5ymt3fiemi7qhsgn2cs26whftene/`

Fair warning I don't know what most of this CIDs are/contain its a random sample.

### LINKED ISSUE #383: allow gateway downloads of cbor files
cbor cid files are erroring with "couldn't determine unixfs type". Instead of erroring, need to have the code investigate the type of content & try to resolve using a different method.

Ties into #207 

example CID `bafyreicnwbboevx6g6fykitf4nebz2kqgkqz35qvlnlcgfulhrris66m6i`

cli cbor examiner, https://github.com/nullus157/cbor-diag-rs/ let's you view cbor content like so:
Screen_Shot_2022-10-20_at_21 56 28

paths that should be resolvable:

https://ipfs.io/ipfs/bafyreicnwbboevx6g6fykitf4nebz2kqgkqz35qvlnlcgfulhrris66m6i/metadata.json
https://ipfs.io/ipfs/bafyreicnwbboevx6g6fykitf4nebz2kqgkqz35qvlnlcgfulhrris66m6i/image
https://ipfs.io/ipfs/bafyreicnwbboevx6g6fykitf4nebz2kqgkqz35qvlnlcgfulhrris66m6i/image/cat.png

ramfox:
closes #207 
closes #383

### DIFF
--- a/iroh-resolver/src/resolver.rs
+++ b/iroh-resolver/src/resolver.rs
@@ -1263,9 +1263,9 @@ impl<T: ContentLoader> Resolver<T> {
         // first iteration is the link list itself
         let _ = links_iter
             .next()
-            .ok_or(anyhow!("expected DagPb links to exist"));
+            .ok_or_else(|| anyhow!("expected DagPb links to exist"));
 
-        while let Some(dagpb_link) = links_iter.next() {
+        for dagpb_link in links_iter {
             match dagpb_link
                 .clone()
                 .take("Name")
--- a/iroh-resolver/src/resolver.rs
+++ b/iroh-resolver/src/resolver.rs
@@ -1103,6 +1103,12 @@ impl<T: ContentLoader> Resolver<T> {
             let mut resolved_path = vec![cid];
 
             for part in tail {
+                // TODO(ramfox): we currently add an empty string to the path.tail vec
+                // when the original path ends in a "/". This is causing errors because
+                // we attempt to index into the data using an empty string, which fails
+                if part.is_empty() {
+                    continue;
+                }
                 self.inner_resolve(&mut current, &mut resolved_path, part, &mut ctx)
                     .await?;
             }
@@ -1200,6 +1206,12 @@ impl<T: ContentLoader> Resolver<T> {
         let mut codec = codec;
 
         for part in path {
+            // TODO(ramfox): we currently add an empty string to the path.tail vec
+            // when the original path ends in a "/". This is causing errors because
+            // we attempt to index into the data using an empty string, which fails
+            if part.is_empty() {
+                continue;
+            }
             if let libipld::Ipld::Link(c) = current {
                 (codec, current) = self.load_ipld_link(c, ctx).await?;
             }
--- a/iroh-resolver/src/resolver.rs
+++ b/iroh-resolver/src/resolver.rs
@@ -7,7 +7,7 @@ use std::sync::Arc;
 use std::task::{Context, Poll};
 use std::time::Instant;
 
-use anyhow::{anyhow, bail, ensure, Context as _, Result};
+use anyhow::{anyhow, bail, Context as _, Result};
 use async_trait::async_trait;
 use bytes::Bytes;
 use cid::multihash::{Code, MultihashDigest};
@@ -766,7 +766,6 @@ impl ContentLoader for Client {
 
     async fn load_cid(&self, cid: &Cid, ctx: &LoaderContext) -> Result<LoadedCid> {
         trace!("{:?} loading {}", ctx.id(), cid);
-
         // TODO: better strategy
 
         let cid = *cid;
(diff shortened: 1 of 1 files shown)


iroh-resolver/src/resolver.rs
```diff
@@ -1090,6 +1103,12 @@ impl<T: ContentLoader> Resolver<T> {
             let mut resolved_path = vec![cid];
 
             for part in tail {
+                // TODO(ramfox): we currently add an empty string to the path.tail vec
+                // when the original path ends in a "/". This is causing errors because
+                // we attempt to index into the data using an empty string, which fails
```
ramfox, line 1108: @Arqu would love some feedback on this

i originally took out these lines:
https://github.com/n0-computer/iroh/blob/96c614839127a0b881ac100e3102abca70cb3a87/iroh-resolver/src/resolver.rs#L253-L255

But it looks like this was added for gateway compliance. Just want to confirm, because it's makin' the resolve code smelly :)

iroh-resolver/src/resolver.rs
```diff
@@ -1119,229 +1138,151 @@ impl<T: ContentLoader> Resolver<T> {
...
+        let loaded_cid = self.load_cid(&cid, ctx).await?;
+
+        let ipld: Ipld = codec
+            .decode(&loaded_cid.data)
+            .map_err(|e| anyhow!("invalid {:?}: {:?}", codec, e))?;
+        Ok((codec, ipld))
+    }
+
+    #[tracing::instrument(skip(self, name))]
+    fn get_dagpb_link<I: Into<String>>(&self, ipld: Ipld, name: I) -> Result<Ipld> {
```
ramfox, line 1256: this... seems like a lot of trouble. Should probably encode to some DagPb struct and add methods there that make this less gross. But wasn't sure about the consequences of leaving "ipld land". I'd like to posit that this is fine for now, but should be changed asap.

iroh-resolver/src/resolver.rs
Arqu: Ok so the reason it's there is purely aesthetically (and compliance with spec) to redirect to a path with a trailing `/` if the result is a unixfs dir. For all intents and purposes you should basically use tail.filter(|s| !s.is_empty()) internally. It's just a way to pass info around it's a dir without having resolved the node yet.

Arqu (approved):
Top level LGTM, left a comment for the trailing `/` bit.

iroh-resolver/src/resolver.rs
```diff
@@ -890,6 +889,17 @@ impl<T: ContentLoader> Resolver<T> {
...
                 if let Some((current_output_path, current_out)) = blocks.pop_front() {
                     let current = current_out?;
+                    // if we have a unixfs::file, we have already followed these links when we
+                    // called this.resolve on the block, no need to double fetch the file
+                    // TODO(ramfox): it would be good to go through the possible
+                    // types that we are willing to support & try and understand if anything  other
+                    // than UnixfsNode::Hamt & UnixfsNode::Directory need to be resolved
+                    // recursively. EG, if we have a dag cbor block, are we expected to resolve all
+                    // the links the way we do with a unixfs dir
+                    if current.metadata().unixfs_type == Some(UnixfsType::File){
```
dignifiedquire, line 899: what about the other `resolve_recursive` calls?

iroh-resolver/src/resolver.rs
```diff
@@ -1090,6 +1103,12 @@ impl<T: ContentLoader> Resolver<T> {
             let mut resolved_path = vec![cid];
 
             for part in tail {
+                // TODO(ramfox): we currently add an empty string to the path.tail vec
+                // when the original path ends in a "/". This is causing errors because
+                // we attempt to index into the data using an empty string, which fails
+                if part.is_empty() {
+                    continue;
```
dignifiedquire, line 1110: why `continue` and not `break`?

iroh-resolver/src/resolver.rs
dignifiedquire: libipld is not the nicest to use

iroh-resolver/src/resolver.rs
Arqu: Might not be the end and skip only empty parts. Should never happen, but...

rklaehn:
So I can confirm that this fixes a bug on main that I discovered when looking into weird behaviour with getting to validate https://github.com/n0-computer/iroh/pull/411

iroh-resolver/src/resolver.rs
ramfox: cause 3am & asmir's suggestions for handling this are better than what I wrote :)

PR #410, opened 2022-10-25
Chunky add

rklaehn:
I did two things:
- add a bulk put option into the store api and chunked the store interactions
  (so we can take advantage of the fact that DBs like to process stuff in large chunks, and we reduce the overhead of all the indirection and wrapping we have in our code)
- parallelize the hashing of the content

I was able to remove the parallel requests, so I think this does not even benefit from the grpc connection pool all that much.
Ideally I think all apis should be stream based. I tried a long time to find a way through the maze that is the proxy! macro, but I gave up...

Results:
fast_add


rklaehn:
Splitting this up into 2 PRs, since it is 2 rather separate things and at least one of them is pretty simple...

https://github.com/n0-computer/iroh/pull/411

PR #411, opened 2022-10-25
perf: add parallelism to hashing

rklaehn:
The most expensive thing when adding a file is hashing the content.

This adds parallelism to this hashing by splitting the tree building into two stages. One where the leafs are encoded, and one where the actual tree is being built.

parallel_hash

This makes sure that iroh-cli now uses all available cores, and therefore increases perf quite a bit. But we are now limited by the DB.

### DIFF
--- a/iroh-resolver/src/balanced_tree.rs
+++ b/iroh-resolver/src/balanced_tree.rs
@@ -4,7 +4,7 @@ use anyhow::Result;
 use async_stream::try_stream;
 use bytes::{Bytes, BytesMut};
 use cid::Cid;
-use futures::{Stream, StreamExt};
+use futures::{Stream, StreamExt, TryFutureExt, TryStreamExt};
 
 use crate::resolver::Block;
 use crate::unixfs::{dag_pb, unixfs_pb, DataType, Node, UnixfsNode};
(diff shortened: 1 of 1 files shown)


iroh-resolver/src/balanced_tree.rs
```diff
@@ -77,11 +77,18 @@ fn stream_balanced_tree(
         let mut tree: VecDeque<Vec<(Cid, LinkInfo)>> = VecDeque::new();
         tree.push_back(Vec::with_capacity(degree));
 
+        let hash_par: usize = 8;
+
+        let in_stream = in_stream.err_into::<anyhow::Error>().map(|chunk| {
+            tokio::task::spawn_blocking(|| {
+                chunk.and_then(|chunk| TreeNode::Leaf(chunk.freeze()).encode())
+            }).err_into::<anyhow::Error>()
+        }).buffered(hash_par).map(|x| x.and_then(|x| x));
```
rklaehn, line 86: Not sure if there is a nicer way to do this than `.map(|x| x.and_then(|x| x))`.

b5:
I'm not qualified to review async code, but in smoke testing I did an `iroh add -> iroh get` loop on a 350MB file, and the result that came back was missing data from the end of the file, roughly ~75MB worth. I tested it on main without issue. Will play some more to see if the problem is repeatable

1. Any idea what might cause such a thing?
2. We should probably make sure there's a solid roundtrip test if we don't already have one.
3. All of this is totally worth it, the speedup is very nice

iroh-resolver/src/balanced_tree.rs
dignifiedquire: can't you use flatten?

rklaehn:
> I'm not qualified to review async code, but in smoke testing I did an `iroh add -> iroh get` loop on a 350MB file, and the result that came back was missing data from the end of the file, roughly ~75MB worth. I tested it on main without issue. Will play some more to see if the problem is repeatable

Thanks for the test. Trying things out is often more useful than staring at the code. I will check this, and also try to figure out how to test this.

iroh-resolver/src/balanced_tree.rs
rklaehn: Now that you mention it... But apparently not.

image

rklaehn:
> I'm not qualified to review async code, but in smoke testing I did an `iroh add -> iroh get` loop on a 350MB file, and the result that came back was missing data from the end of the file, roughly ~75MB worth. I tested it on main without issue. Will play some more to see if the problem is repeatable
> 
> 1. Any idea what might cause such a thing?
> 2. We should probably make sure there's a solid roundtrip test if we don't already have one.
> 3. All of this is totally worth it, the speedup is very nice

Can you retry once #408 is merged?

b5:
hrm, still seeing the issue on get when running this branch. hash results are the same compared to main, but this might be an issue with the content not actually being in the blocks?

b5 (commented):
I've tested this a bunch locally, and am good with it. I'd like @dignifiedquire to sign off

dignifiedquire (approved):

PR #412, opened 2022-10-25
perf: impl put_many and reuse column families

rklaehn:
This requires getting rid of a lot of async in the store. Async only in the public interface.

I want to do some more reorg here, but in the interest of getting something on main soon I will refrain from it...

### DIFF
--- a/iroh-resolver/src/unixfs_builder.rs
+++ b/iroh-resolver/src/unixfs_builder.rs
@@ -10,7 +10,8 @@ use async_recursion::async_recursion;
 use async_trait::async_trait;
 use bytes::Bytes;
 use cid::Cid;
-use futures::{stream::LocalBoxStream, Stream, StreamExt};
+use futures::stream::TryStreamExt;
+use futures::{future, stream::LocalBoxStream, Stream, StreamExt};
 use iroh_rpc_client::Client;
 use prost::Message;
 use tokio::io::AsyncRead;
@@ -31,7 +32,7 @@ use crate::{
 const DIRECTORY_LINK_LIMIT: usize = 6000;
 
 /// How many chunks to buffer up when adding content.
-const ADD_PAR: usize = 24;
+const _ADD_PAR: usize = 24;
 
 #[derive(Debug, PartialEq)]
 enum DirectoryType {
@@ -496,6 +497,7 @@ pub(crate) fn encode_unixfs_pb(
 pub trait Store: 'static + Send + Sync + Clone {
     async fn has(&self, &cid: Cid) -> Result<bool>;
     async fn put(&self, cid: Cid, blob: Bytes, links: Vec<Cid>) -> Result<()>;
+    async fn put_many(&self, blocks: Vec<Block>) -> Result<()>;
 }
 
 #[async_trait]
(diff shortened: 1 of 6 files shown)


iroh-resolver/src/unixfs_builder.rs
```diff
@@ -617,20 +640,54 @@ pub enum AddEvent {
...
+            let block = block?;
+            let block_size = block.data().len() as u64;
+            let cid = *block.cid();
+            if chunk_size + block_size > MAX_CHUNK_SIZE {
+                tracing::info!("adding chunk of {} bytes", chunk_size);
+                store.put_many(chunk.clone()).await?;
+                chunk.clear();
+                chunk_size = 0;
+            } else {
+                chunk_size += block_size;
```
dignifiedquire, line 663: aren’t you forgetting to store anything in the chunk?

rklaehn: It would seem so... Maybe that is why it is that fast...

iroh-resolver/src/unixfs_builder.rs
```diff
@@ -617,20 +640,56 @@ pub enum AddEvent {
...
+    const MAX_CHUNK_SIZE: u64 = 1024 * 1024 * 16;
+    let t = stream! {
+        while let Some(block) = blocks.next().await {
+            let block = block?;
+            let block_size = block.data().len() as u64;
+            let cid = *block.cid();
+            let raw_data_size = block.raw_data_size();
+            if chunk_size + block_size > MAX_CHUNK_SIZE {
+                tracing::info!("adding chunk of {} bytes", chunk_size);
+                store.put_many(chunk.clone()).await?;
```
dignifiedquire, line 660: if you do `std::mem::take` you can avoid cloning, and clearing

iroh-resolver/src/unixfs_builder.rs
```diff
@@ -617,20 +640,56 @@ pub enum AddEvent {
...
+                chunk.clear();
+                chunk_size = 0;
+            } else {
+                chunk.push(block);
```
dignifiedquire, line 664: the push still needs to happen in the other case, or it is lost

iroh-store/src/store.rs
```diff
@@ -191,12 +199,76 @@ impl Store {
...
+            if let Some(_blob) = self.inner.content.get_pinned_cf(cf.blobs, &id_bytes)? {
+                return Ok(true);
+            }
+        }
+        Ok(false)
+    }
+
+    #[tracing::instrument(skip(self))]
+    pub async fn get(&self, cid: &Cid) -> Result<Option<DBPinnableSlice<'_>>> {
+        self.get0(cid, &self.cfs()?)
```
dignifiedquire, line 241: what about using `spawn_blocking?

rklaehn:
Main:
image

This branch:
image

dignifiedquire (approved):

iroh-store/src/store.rs
dignifiedquire: ok

dignifiedquire:
@rklaehn did you switch up things? or does this make things slower?

Arqu:
> @rklaehn did you switch up things? or does this make things slower?

If Im reading the above correct, it's 1 sec faster but uses more CPU cores (hence user time higher)

b5 (approved):

PR #415, opened 2022-10-25
feat: offline adding

b5:
I have no idea what this does for our provide story if a user adds content offline then tries to take it online, but this makes our perf measurable.

I still think we should provide the root to the DHT by default. It's usually what users are trying to do, and while it's slower, it's the de-facto use case

### DIFF
--- a/iroh/src/run.rs
+++ b/iroh/src/run.rs
@@ -1,7 +1,7 @@
 use std::collections::{HashMap, HashSet};
 use std::path::{Path, PathBuf};
 
-use anyhow::{anyhow, Context, Result};
+use anyhow::{Context, Result};
 use clap::{Parser, Subcommand};
 use console::style;
 use crossterm::style::Stylize;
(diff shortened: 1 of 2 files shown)


iroh/src/run.rs
```diff
@@ -175,13 +186,35 @@ async fn add(api: &impl Api, path: &Path, no_wrap: bool, recursive: bool) -> Res
...
     }
 
+    let mut steps = 3;
     // we require p2p for adding right now because we don't have a mechanism for
     // hydrating only the root CID to the p2p node for providing if a CID were
     // ingested offline. Offline adding should happen, but this is the current
     // path of least confusion
-    require_services(api, HashSet::from(["store", "p2p"])).await?;
+    let svc_status = require_services(api, HashSet::from(["store"])).await?;
+    match (provide, svc_status.p2p.status()) {
```
rklaehn, line 195: Not for this PR, but we have to make sure that we refuse to speak to services with different build version

iroh/src/run.rs
```diff
@@ -175,13 +186,35 @@ async fn add(api: &impl Api, path: &Path, no_wrap: bool, recursive: bool) -> Res
...
+        (true, ServiceStatus::Down(_status)) => {
+            return Err(anyhow!("Add provides content to the IPFS network by default, but the p2p service is not running.\n{}",
+            "hint: try using the --offline flag, or run 'iroh start p2p'".yellow()
+            ));
+        }
+        (true, ServiceStatus::Unknown)
+        | (true, ServiceStatus::NotServing)
+        | (true, ServiceStatus::ServiceUnknown) => {
+            return Err(anyhow!("Add provides content to the IPFS network by default, but the p2p service is not running.\n{}",
```
rklaehn, line 204: There is anyhow::bail! which does this and maybe looks a bit nicer.

https://docs.rs/anyhow/latest/anyhow/macro.bail.html

iroh/src/services.rs
```diff
@@ -252,8 +252,12 @@ where
     Ok(())
 }
 
-/// require a set of services is up
-pub async fn require_services(api: &impl Api, services: HashSet<&str>) -> Result<()> {
+/// require a set of services is up. returns the underlying status table of all
+/// services for additional scrutiny
+pub async fn require_services(
+    api: &impl Api,
+    services: HashSet<&str>,
```
rklaehn, line 259: I would prefer BTreeSet in such cases because it is more deterministic. But in this case it does not really matter.

rklaehn (approved):
Looks good, except for the `return Err(anyhow!`.

iroh/src/run.rs
```diff
@@ -161,7 +166,13 @@ impl Cli {
     }
 }
 
-async fn add(api: &impl Api, path: &Path, no_wrap: bool, recursive: bool) -> Result<()> {
+async fn add(
+    api: &impl Api,
+    path: &Path,
+    no_wrap: bool,
+    recursive: bool,
+    provide: bool,
```
dignifiedquire, line 174: the next person touching this needs to create a config struct

iroh/src/services.rs
b5: yeah totally. I'll come through & refactor all uses of `HashSet` to `BTreeSet` in a follow up

PR #416, opened 2022-10-25
feat: integrate indexer resolution

dignifiedquire:


iroh-resolver/src/content_loader.rs
```diff
@@ -0,0 +1,258 @@
...
+                    if let Ok(providers) = indexer.find_providers(*cid).await {
+                        providers.into_iter().map(|p| p.id).collect()
+                    } else {
+                        Default::default()
+                    }
+                } else {
+                    Default::default()
+                };
+
+                let data = p2p.fetch_bitswap(ctx.into(), *cid, providers).await?;
```
willscott, line 148: `find_providers` here is returning all providers - you can filter on `Metadata` to find only the providers that are saying they work with Bitswap - if you don't filter you'll also get fil-markets providers that only speak graphsync

the `Metadata` is base64 encoding of a uvarint encoding of (0x900) - https://github.com/multiformats/multicodec/blob/master/table.csv#L138

dignifiedquire: fixed :)

iroh-resolver/src/content_loader.rs
```diff
@@ -0,0 +1,258 @@
...
+pub struct FullLoaderConfig {
+    pub indexer: Option<Url>,
+    pub http_gateways: Vec<GatewayUrl>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum GatewayUrl {
+    Full(Url),
+    Subdomain(String),
+}
```
b5, line 64: nice cleanup

iroh-gateway/src/bad_bits.rs
```diff
@@ -111,6 +111,9 @@ mod tests {
     use super::*;
     use hex_literal::hex;
     use http::StatusCode;
+    use iroh_resolver::content_loader::FullLoader;
+    use iroh_resolver::content_loader::FullLoaderConfig;
+    use iroh_resolver::content_loader::GatewayUrl;
```
b5, line 116: group these imports?

iroh-resolver/src/racing.rs
```diff
@@ -1,5 +1,6 @@
-use crate::resolver::{
```
b5, line 1: do we need the racing loader at all anymore?

dignifiedquire: no, it is fully integrated now

iroh-resolver/src/content_loader.rs
```diff
@@ -0,0 +1,258 @@
...
+            }
+            Either::Right((gateway, bitswap_future)) => {
+                if let Ok(Some(loaded)) = gateway {
+                    loaded
+                } else {
+                    bitswap_future
+                        .await?
+                        .ok_or_else(|| anyhow!("failed to find {}", cid))?
+                }
+            }
```
b5, line 248: we should be emitting metrics on which one of these wins

iroh-gateway/src/bad_bits.rs
dignifiedquire: fixed

iroh-resolver/src/racing.rs
b5: nice. I vote we remove in this PR

b5:
ok, I've run locally, seems to not explode. Best content to experiment with seems to be stuff coming from dag.house, like the example CID: bafybeigvgzoolc3drupxhlevdp2ugqcrbcsqfmcek2zxiw5wctk3xjpjwy

With that said, it's quite hard to know when the indexer is in fact returning usable providers. Would love to get this instrumented in a follow-up PR.

b5:
merged with #476

PR #418, opened 2022-10-25
feat: bitswap client mode

dignifiedquire:
When setting `bitswap_server: false`, there is no serving of blocks happening, only client requests.

### DIFF
--- a/iroh-bitswap/src/lib.rs
+++ b/iroh-bitswap/src/lib.rs
@@ -70,7 +70,7 @@ pub struct Bitswap<S: Store> {
     /// Set to true when dialing should be disabled because we have reached the conn limit.
     pause_dialing: bool,
     client: Client<S>,
-    server: Server<S>,
+    server: Option<Server<S>>,
     incoming_messages: mpsc::Sender<(PeerId, BitswapMessage)>,
     peers_connected: mpsc::Sender<PeerId>,
     peers_disconnected: mpsc::Sender<PeerId>,
@@ -101,16 +101,26 @@ impl PeerState {
 #[derive(Debug)]
 pub struct Config {
     pub client: ClientConfig,
-    pub server: ServerConfig,
+    /// If no server config is set, the server is disabled.
+    pub server: Option<ServerConfig>,
     pub protocol: ProtocolConfig,
     pub idle_timeout: Duration,
+}
+
+impl Config {
+    pub fn default_client_mode() -> Self {
+        Config {
+            server: None,
+            ..Default::default()
+        }
+    }
 }
 
 impl Default for Config {
     fn default() -> Self {
         Config {
             client: ClientConfig::default(),
-            server: ServerConfig::default(),
+            server: Some(ServerConfig::default()),
             protocol: ProtocolConfig::default(),
             idle_timeout: Duration::from_secs(30),
         }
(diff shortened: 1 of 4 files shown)


iroh-bitswap/src/lib.rs
```diff
@@ -216,16 +237,22 @@ impl<S: Store> Bitswap<S> {
 
     pub async fn stop(self) -> Result<()> {
         self.network.stop();
-        let (a, b) = futures::future::join(self.client.stop(), self.server.stop()).await;
-        a?;
-        b?;
+        if let Some(server) = self.server {
+            let (a, b) = futures::future::join(self.client.stop(), server.stop()).await;
+            a?;
+            b?;
```
matthiasbeyer, line 243: How about [try_join](https://docs.rs/futures/latest/futures/future/fn.try_join.html) here?

```suggestion
            futures::future::try_join(self.client.stop(), server.stop()).await?;
```

iroh-p2p/src/behaviour.rs
```diff
@@ -88,9 +88,14 @@ impl NodeBehaviour {
...
 
-        let bitswap = if config.bitswap {
+        let bitswap = if config.bitswap_client || config.bitswap_server {
             info!("init bitswap");
-            let bs_config = BitswapConfig::default();
+            // NOTE: server only mode is not implemented yet
+            let bs_config = if config.bitswap_server {
+                BitswapConfig::default()
+            } else {
+                BitswapConfig::default_client_mode()
```
matthiasbeyer, line 97: This is just a nitpick:

```suggestion
            let bs_config = config.bitswap_server
                .then(BitswapConfig::default)
                .or_else(ButswapConfig::default_client_mode);
```

Nice that rust is this expressive, but in this particular instance I am not fond of one way or another :laughing: just wanted to show off in this case, hope you don't mind :laughing:

matthiasbeyer (changes requested):

iroh-p2p/src/behaviour.rs
b5: tried to implement as suggested.  Ran into an issue:
`then` and `or_else` return `Option<T>`, which requires unwrapping, making it less clear than the `if / else` variant.

dignifiedquire:
@b5 ready for review

iroh-bitswap/src/lib.rs
```diff
@@ -127,16 +137,15 @@ pub trait Store: Debug + Clone + Send + Sync + 'static {
...
-        let server = Server::new(network.clone(), store.clone(), config.server).await;
-        let client = Client::new(
-            network.clone(),
-            store,
-            server.received_blocks_cb(),
-            config.client,
-        )
-        .await;
-
-        // EVIL: must eventually not be unbounded !!!
```
b5, line 139: just to confirm, these channels look bounded to me, so it's right to remove this comment, but nothing has changed in this patch in terms of channel bounding, right?

dignifiedquire: correct, it's just a cleanup, I had forgotten to remove the comment when I changed the channel size

iroh-p2p/src/config.rs
```diff
@@ -42,8 +42,10 @@ pub struct Libp2pConfig {
     pub bootstrap_peers: Vec<Multiaddr>,
     /// Mdns discovery enabled.
     pub mdns: bool,
-    /// Bitswap discovery enabled.
-    pub bitswap: bool,
+    /// Bitswap server mode enabled.
+    pub bitswap_server: bool,
+    /// Bitswap client mode enabled.
+    pub bitswap_client: bool,
```
b5, line 48: I'll annotate that these are default true on the docs site. I'd ask for a `default: true` in the comment, but there's too high a risk it'll de-syncronize from the `default` method over time

b5 (approved):
I haven't had a chance to properly smoke test, but doing a _real) smoke test would be much easier if we had #526 worked out, so I'll wait until after then to fully test. lgtm

PR #419, opened 2022-10-25
fix(p2p): cleanup sessions when they are canceled

dignifiedquire:


iroh-p2p/src/node.rs
```diff
@@ -221,6 +221,27 @@ impl<KeyStorage: Storage> Node<KeyStorage> {
...
+        for (session_id, workers) in &mut self.bitswap_sessions {
+            // Check if the workers are still active
+            workers.retain(|(_, worker)| !worker.is_finished());
+
+            if workers.is_empty() {
+                to_remove.push(*session_id);
+            }
+
+            // Only do a small chunk of cleanup on each iteration
+            if to_remove.len() >= 10 {
```
Arqu, line 235: can we just have a comment that this is another magic number. I really want to pull all of these out after release.

PR #421, opened 2022-10-25
Cleanup of store code

rklaehn:
This separates the store code into a LocalStore where all the things happen, that can properly store column family handles etc. and a Store that is just a forwarder for which RpcStore is implemented.

I think this is much nicer. We might want to use spawn_local at the async -> synchronous boundary though.

This is based on https://github.com/n0-computer/iroh/pull/412 , so once that is merged the diff would get smaller.


Arqu (approved):
lgtm

PR #428, opened 2022-10-26
fix: put store behind rwlock to brute force ensure consistency

rklaehn:
...until we have time to figure out the correct way to do this within RocksDb itself


ramfox:
This was a potential option when we had a time limit before the release.  Check out this issue to track progress on fixing this: https://github.com/n0-computer/iroh/issues/461

PR #430, opened 2022-10-26
switch macos data & config dir to ~/.config/iroh

b5:
Alright I've had it.

I'm working on installer scripts for iroh, and on os X "/Users/$USER/Library/Application Support/iroh" is just not a directory I can remember. The experience is shit. Who puts a SPACE in the name of a critical directory!?

But in more concrete terms, our users will need to interact with this directory if we can't automatically load binaries onto their $PATH. having every application under the sun pollute ~/.$APP_NAME isn't sustainable, so I vote we use the default linux location. Checking my own local machine I have the following applications already populating `.config`:

```
/Users/b5/.config
├── configstore
│   └── update-notifier-gatsby-cli.json
├── gatsby
│   ├── config.json
│   └── sites
│       └── fb8c3aad24651584365e9bad287f6c4e
│           ├── developproxy.json
│           ├── developstatusserver.json
│           ├── metadata.json
│           └── telemetryserver.json
├── gscatter
│   └── uid
├── htop
│   └── htoprc
├── iterm2
│   └── AppSupport -> /Users/b5/Library/Application Support/iTerm2
└── xm1
    ├── texmaker.ini
    └── texmakerapp.ini
```

I'm cool to have this one shot down, just want to raise this issue before our release, when it'll take way more work to change.

### DIFF
--- a/iroh-util/src/lib.rs
+++ b/iroh-util/src/lib.rs
@@ -57,12 +57,21 @@ pub async fn block_until_sigint() {
 /// | Platform | Value                                 | Example                          |
 /// | -------- | ------------------------------------- | -------------------------------- |
 /// | Linux    | `$XDG_CONFIG_HOME` or `$HOME`/.config/iroh | /home/alice/.config/iroh              |
-/// | macOS    | `$HOME`/Library/Application Support/iroh   | /Users/Alice/Library/Application Support/iroh |
+/// | macOS    | `$HOME`/.config/iroh   | /Users/Alice/.config/iroh |
 /// | Windows  | `{FOLDERID_RoamingAppData}`/iroh           | C:\Users\Alice\AppData\Roaming\iroh   |
+#[cfg(not(target_os = "macos"))]
 pub fn iroh_config_root() -> Result<PathBuf> {
     let cfg = dirs_next::config_dir()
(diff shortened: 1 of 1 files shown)


fabricedesre:
I like what iterm2 does, easy access through the symlink, while still obeying platform standards.

b5:
update: we don't have perms from the OS to run binaries from `~/Library/`, so we can't de-facto install binaries to `~/Library/Application Support/iroh/bin`, but we _can_ run binaries placed at `~/.config/iroh/bin`.

I think we should make the switch.

b5:
> I like what iterm2 does, easy access through the symlink, while still obeying platform standards.

Yeah this is a nice pattern. In that case, in that case if we're running an installer script, where do we put the binaries? (Ideally we'll be all set up with package managers as well, but that will take time)

b5:
one option would be to put the binaries at `~/.config/iroh/bin`, and an `AppData` symlink as a sibling to the bin dir

b5:
After discussing this with a number of folks, I think the right thing to do is to follow @fabricedesre's suggestion. Let's leave the directories where they are, and add a symlink to `~/.config/iroh/AppData` as part of the `https://sh.iroh.computer/install_laptop.sh` script

PR #433, opened 2022-10-27
default start all services, but accept config

b5:
this PR changes the default for `iroh start` back to {gateway,store,p2p}, but now iroh CLI reads a new `cli.config.toml` file that accepts a single configuration parameter. We can use this to get back the "only gateway & store" behaviour with an installer script that makes `cli.config.toml`.

```
# when running on a laptop, only turn on the gateway & store by default
start_default_services = [
  "gateway",
  "store",
]
```

I've kept `iroh start --all`, because it's still useful to have a CLI flag that overrides what's started in any case


Arqu (approved):
Code LGTM. Tested locally, works like a charm.

PR #440, opened 2022-10-28
Docker MVP

### LINKED ISSUE #439: docker mvp
I'd like to add a /docker/ folder with Dockerfile and docker-compose examples

I see in the docs that its mentioned, but don't see any issues.

Is anyone working on this?

If not, any advice and/or encouragement for me?

gotjoshua:
Curious what more is needed to make this feel MVP...

closes #439


b5:
Thanks for submitting @gotjoshua, someone will review in the next 36 hours

gotjoshua:
> someone will review in the next 36 hours

is anyone on that?

docker-compose.yml
```diff
@@ -0,0 +1,51 @@
+version: '2.4'
```
b5, line 1: please move this file into the docker directory.

docker-compose.yml
```diff
@@ -0,0 +1,51 @@
...
+
+networks:
+  # haven is a traefik network that is setup elsewhere (and beyond the scope of this example)
+  haven:
+    external: true
+    name: haven
```
b5, line 7: can we do this example without using `treafik`? I'm getting the following error when trying to run this locally without it:

```
network haven declared as external, but could not be found
```

docker/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+EXPOSE 9050
+
+COPY ./setup.sh /setup.sh
+# RUN  wget https://sh.iroh.computer/setup.sh && chmod 700 setup.sh
+RUN chmod 700 setup.sh
+RUN ./setup.sh init
+
+ENV PATH="${PATH}:~/.iroh/bin"
+
+CMD /bin/sh -c "./setup.sh start && tail -n 100 -f ~/.iroh/log/*"
```
b5, line 13: while this does technically "put iroh on docker", this isn't the right way to go about it. Instead we should have a dockerfile for each service (`iroh-store`, `iroh-p2p`, `iroh-gateway`), and those dockerfiles should use [multi-stage builds](https://docs.docker.com/build/building/multi-stage/) to build production rust binaries & run them on minimal distribution environments.

Given how involved the task is, it might make sense to have someone on the iroh team put together those Dockerfiles, and maybe rebase this PR to build a `docker-compose.yml` example on top of those

docker/setup.sh
```diff
@@ -0,0 +1,178 @@
+#!/bin/sh
```
b5, line 1: This is our user-facing installer script, and shouldn't be used for our docker infrastructure

b5 (changes requested):
Thanks so much for taking the time to put this together @gotjoshua. I'm sorry to say I think we'd need to do a lot of work to get this into the right shape for our docker story. 

If you're still interested in contributing even given this will take way more work, let me know. I've opened #440 to track what I think should happen first for a proper docker story. Either way I'd move to close this PR without merging & we can talk next steps on issues

gotjoshua:
> opened #440 to track

This should point to #473 


I also agree that using the setup.sh to create the images doesn't feel right... It was my quickest way to get some thing running in docker.

I'm fine to close this as an unmerged MVP.

docker-compose.yml
gotjoshua: One can
1. add traefik config into this compose file 
2. Or just comment that stuff out and bind the port(s)
3. Or actually setup traefik

On a host machine with multiple services traefik is a great tool for configurable reverse proxy routing.

b5:
Sounds good. Thanks again for your work @gotjoshua! I hope you keep this code around for yourself should you find it useful 😄

Issue #471, opened 2022-11-05
windows tests to pass again


b5:
windows support has fallen behind, and we're hoping to get back to parity in time for our next release. I've been tinkering with iroh on a local test box and can at least compile & run all binaries (nice), but the test suite is a mess. I figure it's easiest to work through all the crates in the workspace one-by-one.

- [ ] **iroh** 13 pass; 5 failed
- [x] **iroh-api** 5 passed; 0 failed 
- [x] **iroh-bitswap** 52 passed; 0 failed;
- [x] **iroh-car** 2 passed; 0 failed;
- [ ] **iroh-gateway**
- [ ] **iroh-localops**
- [ ] **iroh-metrics**
- [ ] **iroh-one**
- [ ] **iroh-p2p**
- [ ] **iroh-resolver**
- [ ] **iroh-rpc-client**
- [ ] **iroh-rpc-types**
- [ ] **iroh-store**
- [ ] **iroh-share**
- [ ] **iroh-util**
- [ ] **stores/***
- [ ] **examples**
- [ ] **xtask**

## Crate Specific issues
* `iroh` has ~5 tests failing with a lot of `Files left in unexpected state`
* `iroh-api`

PR #451, opened 2022-11-02
refactor mocking in iroh-api & iroh

dignifiedquire:
- Avoid using a trait to mock the `Api` struct
- Split `fs` operations from fetching
- General cleanup

PR #472, opened 2022-11-05
fix Iroh -> API, `iroh-util` crate tests pass on windows

b5:
#451 intro'd a bug at the top of the iroh crate, should be fixed.

This also cleans up the `iroh-util` crate for #471


b5:
Merging w/o approval to get `main` back in shape

PR #474, opened 2022-11-07
Clean up dependencies

flub:
This contains two commits:

- Removes unused dependencies.
- Sorts dependencies in Cargo.tom files.

This is meant to be committed without squashing, see the commit messages for details.

If the 2nd commit is too controversial I'm happy to remove it.


Arqu (commented):
LGTM but lets get the GHA ubuntu CI to pass too.

b5 (commented):
Thank you for sorting imports! huge pet peve of mine. Can this PR also add a note to `DEVELOPERS.md` that requests any dependency change keep deps sorted. Seems obvious, but it's important to have something to point to so contributors aren't frustrated with unspoken rules.

I think we should add a new section on dependencies with the following language:

# Dependencies
Any crate added to iroh will need to use a license compatible with ours. Any PR that introduces a new crate will require additional review time to audit the crate being introduced, including rationale on why you chose _this_ crate, and what alternatives you considered will speed up the review process.

Crate lists in `Cargo.toml` files must be kept  alphabetically sorted.

flub:
@b5 great, added that

PR #476, opened 2022-11-08
feat: indexer resolution

b5:
supersedes #416. Using a separate PR to keep things comparable. Updating this branch changed a lifetime parameter that has me somewhat concerned.


iroh-api/src/api.rs
```diff
@@ -66,17 +89,17 @@ impl Api {
     }
 
     /// High level get, equivalent of CLI `iroh get`.
-    pub fn get(
-        &self,
+    pub fn get<'a>(
+        &'a self,
         ipfs_path: &IpfsPath,
-    ) -> Result<LocalBoxStream<'static, Result<(RelativePathBuf, OutType)>>> {
+    ) -> Result<LocalBoxStream<'a, Result<(RelativePathBuf, OutType)>>> {
```
b5, line 95: on rebase I needed to switch the lifetime param of the return value from `'static` to a new generic. I'm not totally clear on why this is, and think it's a bug. Given that we clone the resolver below I don't see why the lifetime of the return value should need to match that of `self`.

flub: the problem is on (new) line 122 & 128: you're cloning resolver from `self` again, which pulls `self` into the stream future.  But you already clone `resolver` out of `self` on (new) line 102 so you don't need to modify those two lines at all: using `resolver.clone()` instead of `self.resolver.clone()` there solves the lifetime issue.

iroh-api/src/api.rs
```diff
@@ -96,13 +119,13 @@ impl Api {
                 if out.is_dir() {
                     yield (relative_path, OutType::Dir);
                 } else if out.is_symlink() {
-                    let mut reader = out.pretty(resolver.clone(), Default::default(), iroh_resolver::resolver::ResponseClip::NoClip)?;
+                    let mut reader = out.pretty(self.resolver.clone(), Default::default(), iroh_resolver::resolver::ResponseClip::NoClip)?;
```
flub, line 122:
```suggestion
                    let mut reader = out.pretty(resolver.clone(), Default::default(), iroh_resolver::resolver::ResponseClip::NoClip)?;
```

iroh-api/src/api.rs
```diff
@@ -96,13 +119,13 @@ impl Api {
...
                     let mut target = String::new();
                     reader.read_to_string(&mut target).await?;
                     let target = PathBuf::from(target);
                     yield (relative_path, OutType::Symlink(target));
                 } else {
-                    let reader = out.pretty(resolver.clone(), Default::default(), iroh_resolver::resolver::ResponseClip::NoClip)?;
+                    let reader = out.pretty(self.resolver.clone(), Default::default(), iroh_resolver::resolver::ResponseClip::NoClip)?;
```
flub, line 128:
```suggestion
                    let reader = out.pretty(resolver.clone(), Default::default(), iroh_resolver::resolver::ResponseClip::NoClip)?;
```

iroh-resolver/src/indexer.rs
```diff
@@ -0,0 +1,198 @@
...
+
+    use super::Transport;
+
+    pub fn serialize<S>(t: &Transport, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut bytes = unsigned_varint::encode::u32_buffer();
+        unsigned_varint::encode::u32(*t as u32, &mut bytes);
+        serializer.serialize_str(&base64::encode(&bytes))
```
flub, line 139: probably want to make stable clippy happy here and remove the `&`

flub (dismissed):

iroh-gateway/src/core.rs
```diff
@@ -135,7 +136,20 @@ mod tests {
     ) -> (SocketAddr, RpcClient, tokio::task::JoinHandle<()>) {
         let rpc_addr = "grpc://0.0.0.0:0".parse().unwrap();
         let rpc_client = RpcClient::new(config.rpc_client().clone()).await.unwrap();
-        let core = Core::new(config, rpc_addr, Arc::new(None), rpc_client.clone())
+        let loader_config = FullLoaderConfig {
+            http_gateways: config
+                .http_resolvers
+                .as_ref()
```
rklaehn, line 142:
```rust
.iter()
.flatten()
```

iroh-gateway/src/main.rs
```diff
@@ -47,10 +47,25 @@ async fn main() -> Result<()> {
...
 
-    let content_loader = RacingLoader::new(
+    let content_loader = FullLoader::new(
         RpcClient::new(config.rpc_client.clone()).await?,
-        config.http_resolvers.clone().unwrap_or_default(),
-    );
+        FullLoaderConfig {
+            http_gateways: config
+                .http_resolvers
+                .clone()
```
rklaehn, line 55: .iter().flatten()

iroh-one/src/main.rs
```diff
@@ -76,10 +76,20 @@ async fn main() -> Result<()> {
...
-    let content_loader = RacingLoader::new(
+    let content_loader = FullLoader::new(
         RpcClient::new(config.rpc_client.clone()).await?,
-        config.gateway.http_resolvers.clone().unwrap_or_default(),
-    );
+        FullLoaderConfig {
+            http_gateways: config
+                .gateway
+                .http_resolvers
+                .clone()
```
rklaehn, line 85: .iter().flatten()

rklaehn (approved):
It seems to do what is advertised.

I find it weird, so let's do it for now and try to get rid of it at some point. Not this code - the whole concept of a p2p system using centralised indexer nodes and gateways...

See the comments for some minor improvements.

fabricedesre:
> It is weird for a p2p system to do such things: call gateways and indexer nodes via http. Might as well put the data on s3. But it seems to do what it advertises...

Probably not the place to discuss this, but anyway: I don't see IPFS primarily as a p2p system, but as a content based addressing stack. The beauty of that is that you can rely on whatever transport and should use the most efficient one in all situations, which p2p often is not...

flub (approved):

PR #479, opened 2022-11-09
Remove unused mut

matthiasbeyer:
Found this `mut` is unused.

### DIFF
--- a/iroh-metrics/src/core.rs
+++ b/iroh-metrics/src/core.rs
@@ -36,7 +36,7 @@ pub(crate) struct Core {
 
 impl Default for Core {
     fn default() -> Self {
-        let mut reg = Registry::default();
+        let reg = Registry::default();
         Core {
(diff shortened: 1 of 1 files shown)


flub:
Hum, CI would indicate this `mut` was used... why did you think it wasn't?

matthiasbeyer:
Clippy reported it as unused for me.

flub:
Probably because your clippy does not enable the right crate features?  It seems to depend on what crate features are enabled as to whether this is used as `mut` or not.  If you want you could figure out what the right `cfg` incantation is to switch between it being mut and not mut and update this PR?

matthiasbeyer:
Yeah well so it looks like that any feature would require that `mut` to be there. TBH I think it is too much hassle to fix this warning. There is no harm in leaving it as is!

Feel free to close this PR!

flub:
ok, thanks for checking this out anyway

PR #483, opened 2022-11-10
ref(benches): Describe more ways of running benchmarks

flub:
You don't need to install cargo-criterion, it only gives you a nicer
experience.

Also fix up a comment.

### DIFF
--- a/iroh-p2p/benches/lru_cache.rs
+++ b/iroh-p2p/benches/lru_cache.rs
@@ -5,7 +5,13 @@
 //!
 //! # Running the benchmarks
 //!
-//! Install `cargo-criterion`:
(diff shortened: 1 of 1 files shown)


dignifiedquire (dismissed):

iroh-p2p/benches/lru_cache.rs
```diff
@@ -5,7 +5,13 @@
 //!
 //! # Running the benchmarks
 //!
-//! Install `cargo-criterion`:
+//! You can run the benchmarks using either by directly using cargo:
```
flub, line 8:
```suggestion
//! You can run the benchmarks either by directly using cargo:
```

Arqu (approved):
👍

PR #485, opened 2022-11-11
feat: iroh as a library code spike

b5:
this is a draft PR to illustrate a possible API for using iroh as a library. It factors the main parts of the iroh-one binary into a reusable `core` module, and includes an example that import the `iroh-one` crate as a library for use within another application.

I'm hoping we can mature a PR like this into something that's actually usable in the short term, with the caveat that anyone using this is signing up for dealing with a _lot_ of breaking changes as we work on the high level iroh API. But we've had so much interest in running iroh as a library I think we need to get moving on this ASAP.

I'm relatively happy with the high-level API articulated in the example:
```rust
use anyhow::Result;
use iroh_one::{config::Config, core::Core};

#[tokio::main(flavor = "multi_thread")]
    let mut cfg = Config::default();
    let mut iroh = Core::new(&mut cfg)?;
    iroh.start().await?;

    let cli = iroh.api().await?;
    let table = cli.check().await;
    println!("{:#?}", table);

    iroh.stop()
}
```

It currently doesn't work, because the examples compile with tests, which uses mock-apis in our case 😦


rklaehn:
I would like to know what is "short term" in this context.

Do people really want a full node inside their program, or could we just write something very lightweight that just pulls from a gateway for now? Importing basically the entire iroh codebase including grpc seems like a big burden in terms of size and compile time.

matthiasbeyer:
So you asked for initial feedback on [mastodon](https://octodon.social/@b5/109322950810105069), so here it is.

There's no API surface yet from what I see, so there's little I can actually review. FWIW, the `Api` struct only has the absolute minimum in functionality, that is adding some bytes to the IPFS store (or not even that, only adding `Path`s to the IPFS store, which is even less).

On the API side: I don't see why there needs to be a dedicated `start` method. Just `Core::start(config).await` would totally suffice from a users standpoint.
There's no option to specify where the IPFS storage is on disk yet, but I'm sure you're adding it. From what I see everything happens inmemory right now.

I see that components of iroh are started as tokio tasks. I like that! I feared that you'd have them started as seperate processes, but you did the right thing here starting them as tasks!

I don't particularly like that you're opening a `TempDir` in there. Not sure what the `uds-gateway` is, but opening a directory somewhere on the users machine should be made explicit to the user of the library, by letting them provide the path to said directory!
Also, you're starting a task in there that you cannot stop because you drop the handle!

Also, why a `stop()` function? Nothing in there seems to be able to fail or to be `async`, so why not implement it as a `Drop` implementation? Maybe there's a reason for that I cannot see right now.

---

I didn't write the above as code comments, because I am not sure whether these more general things would be beneficial if annotated on code.

I'd love to keep the feedback loop tight, so I am staying subscribed to this PR! :tada: :heart:

b5:
@rklaehn:

> I would like to know what is "short term" in this context.

Great question. Speaking practically, we have a number of folks who want to get started playing with iroh as a library for a hackathon that starts in 4 days. Having _something_ in place would mean we can gather feedback on what functionality folks need form this use case. This would come with the warning that the API is not yet in a release, and may be scrapped entirely.

With that said, this PR & early feedback around expectations indicates to me it's too early to be inviting users to build on this, so I'm thinking I'll tell the hackathon organizers that we're not quite ready to have folks building on this at the moment.

> Do people really want a full node inside their program, or could we just write something very lightweight that just pulls from a gateway for now? Importing basically the entire iroh codebase including grpc seems like a big burden in terms of size and compile time.

Yes, lots of projects want this. We ran an entire IPFS node in-process at qri, and numerous projects want this functionality. I totally agree that comes at a steep price in terms of build times at the moment, it's a thing we an improve upon iteratively.

@matthiasbeyer: thanks so much for your feedback!

> There's no API surface yet from what I see, so there's little I can actually review.

I hear you. I should have been more clear that I was seeking feedback on the server / api client arrangement, which you gave! appreciate it. If you have specific IPFS functionality that you'd like to see us prioritize, please post about it here: https://github.com/n0-computer/iroh/discussions/385

>  I don't particularly like that you're opening a TempDir in there. Not sure what the uds-gateway is, but opening a directory somewhere on the users machine should be made explicit to the user of the library, by letting them provide the path to said directory!

Agreed. the uds-gateway feature provides a unix domain socket. Issue opened: #487


> Also, you're starting a task in there that you cannot stop because you drop the handle!

lol well then, will fix :)

> Also, why a stop() function? Nothing in there seems to be able to fail or to be async, so why not implement it as a Drop implementation? Maybe there's a reason for that I cannot see right now.

I've inherited this from my days as a go programmer. I'll need to check with more experienced rustaceans on our team, but it'd be great to put all the `stop` stuff in `drop`

dignifiedquire:
> Also, why a stop() function?

We should have an `async stop` function, proper shutdown requires awaiting multiple pieces, and as there is on async drop yet, it is the only way to go for now unfortunately.

iroh-one/src/core.rs
```diff
@@ -0,0 +1,155 @@
...
+#[cfg(feature = "uds-gateway")]
+use iroh_one::uds;
+use iroh_resolver::racing::RacingLoader;
+use iroh_rpc_types::Addr;
+#[cfg(feature = "uds-gateway")]
+use tempdir::TempDir;
+use tokio::sync::RwLock;
+use tokio::task;
+
+pub struct Core<'a> {
```
dignifiedquire, line 18: `Core` is a really unfortunate and non-descriptive name, just `Iroh` or `One` (given this lives inside of `iroh-one` atm) would make more sense to me

matthiasbeyer:
Another thing I just noticed: You're using the `anyhow` crate here, but you're writing a library. `anyhow` is not [meant to be used for libraries](https://github.com/dtolnay/anyhow#comparison-to-thiserror) because it hides error type information. As a user of your library, I cannot catch and handle errors this way, only if I resort to some downcasting pseudo magic, and that only works if you're not using `String` errors internally (which would be used if you `anyhow::bail!()` somewhere in your code).

As a user of this library, I would like to be able to handle errors programatically, but a `anyhow::Error` in your API makes that _really, really_ hard (or even impossible) and error prone for me to get right!

dignifiedquire:
@matthiasbeyer unfortunately due to the current internal structures of the iroh crates we won't be moving away from `anyhow` any time soon. There are just too many different errors at the moment, which is just impractical to capture with something like `thiserror`.

matthiasbeyer:
Would you accept patches to clean the whole stack up? Piece by piece of course, not all at once?

dignifiedquire:
sure

dignifiedquire:
@b5 what's the plan & status here?

b5:
Plan was to have @flub take it over & push forward the iroh-as-a-lib project for E2E tests

flub:
yep, I'm working on this at the moment

dignifiedquire:
Closing in favor of #565

PR #486, opened 2022-11-11
chore(xtask): Make dev-install a little friendlier

flub:
Add a warning in case there's no release build to install.

Ideally it'd also warn if the build is out of date, but that's
for another day.


b5 (approved):
Lovely addition

PR #488, opened 2022-11-11
ref(api): Switch argument from PathBuf to &Path

flub:
Changes the file_paths argument of make_config to take &Path
references.  It is not using this allocation and instead only using it
a &Path.  Many callers also had to explicitly allocate this.

---

I started by simplifying the user-facing API by turning it into a
PathBuf, but after checking how this was used that's just a pointless
allocation.  So instead this does the slightly more complex thing.


ramfox (approved):
ty! appreciate the attention to detail.

PR #489, opened 2022-11-11
test(iroh-p2p): test interactions between multiple nodes

### LINKED ISSUE #455: p2p unit tests
we need to mock p2p connections. irritating, but required.

Let's write a unit test for each rpc method. make sure it's control flow is correct.

- [x] ExternalAddrs
- [x] LocalPeerId
- [ ] ~~BitswapRequest~~ (move to bitswap specific testing work)
- [ ] ~~BitswapNotifyNewBlocks~~ (move to bitswap specific testing work)
- [ ] ~~BitswapStopSession~~ (move to bitswap specific testing work)
- [x] ProviderRequest
- [x] StartProviding
- [x] StopProviding
- [x] NetListeningAddrs
- [x] NetPeers
- [x] NetConnectByPeerId
- [x] NetConnect
- [ ] ~~NetDisconnect~~ (not implemented)
- [x] Gossipsub
- [x] FindPeerOnDHT
- [x] LookupPeerInfo
- [x] ListenForIdentify
- [ ] CancelListenForIdentify
- [x] AddressesOfPeer
- [ ] ~~Shutdown~~ (Not implemented)

ramfox:
covers these `RpcMessages`

- ExternalAddrs
- LocalPeerId
- ProviderRequest
- StartProviding
- StopProviding
- NetListeningAddrs
- NetPeers
- NetConnectByPeerId
- NetConnect
- Gossipsub messages
- FindPeerOnDht
- LookupPeerInfo
- ListenForIdentify
- AddressesOfPeer

The missing tests are for:
- P2pClient::fetch_bitswap
- P2pClient::notify_new_blocks_bitswap
- P2pClient::stop_session_bitswap
- P2pClient::disconnect -> this isn't actually implemented
- P2pClient::shutdown -> this doesn't seem to do anything

The bitswap commands are a bit trickier to fake. I would like to either move on or tackle them in a separate PR.

closes #455

### DIFF
--- a/iroh-p2p/src/node.rs
+++ b/iroh-p2p/src/node.rs
@@ -1059,19 +1059,19 @@ mod tests {
     }
 
     struct TestRunnerBuilder {
-        // optional listening address for this node, when None, the swarm will connect to a random
-        // tcp port
+        /// optional listening address for this node, when None, the swarm will connect to a random
+        /// tcp port
         addr: Option<Multiaddr>,
-        // listening addresses for the p2p client, when None, the client will communicate over
-        // a memory rpc channel
+        /// listening addresses for the p2p client, when None, the client will communicate over
+        /// a memory rpc channel
         rpc_addrs: Option<(P2pServerAddr, P2pClientAddr)>,
-        // when true, allow bootstrapping to the network, otherwise don't provide any addresses
-        // from which to bootstrap
+        /// when true, allow bootstrapping to the network, otherwise don't provide any addresses
+        /// from which to bootstrap
         bootstrap: bool,
-        // optional seed to use when building a peer_id, when None will use a previously derived
-        // peer_id 12D3KooWFma2D63TG9ToSiRsjFkoNm2tTihScTBAEdXxinYk5rwE
+        /// optional seed to use when building a peer_id, when None will use a previously derived
+        /// peer_id 12D3KooWFma2D63TG9ToSiRsjFkoNm2tTihScTBAEdXxinYk5rwE
         seed: Option<ChaCha8Rng>,
-        // optional keys to tell the node to provide on the dht
+        /// optional keys to tell the node to provide on the dht
         keys: Option<Vec<Key>>,
     }
 
@@ -1175,7 +1175,10 @@ mod tests {
 
             let client = client.try_p2p()?;
 
-            let addr = get_addr(client.clone()).await?;
+            let addr =
+                tokio::time::timeout(Duration::from_millis(500), get_addr_loop(client.clone()))
+                    .await
+                    .context("timed out before getting a listening address for the node")??;
             let mut dial_addr = addr.clone();
             dial_addr.push(Protocol::P2p(peer_id.into()));
             Ok(TestRunner {
@@ -1189,33 +1192,27 @@ mod tests {
         }
     }
 
-    async fn get_addr(client: P2pClient) -> Result<Multiaddr> {
+    async fn get_addr_loop(client: P2pClient) -> Result<Multiaddr> {
         loop {
-            tokio::select! {
-                _ = tokio::time::sleep(Duration::from_millis(500)) => {
-                    return Err(anyhow!("timed out before we could get a listening address for the node"));
-                },
-                l = client.listeners() => {
-                    if let Some(a) = l?.get(0) {
-                        return Ok(a.clone());
-                    }
-                }
+            let l = client.listeners().await?;
+            if let Some(a) = l.get(0) {
+                return Ok(a.clone());
             }
         }
     }
 
     struct TestRunner {
-        // task for the running p2p node
+        /// task for the running p2p node
         task: JoinHandle<()>,
-        // rpc client, can use it to communicate with the p2p node
+        /// rpc client, can use it to communicate with the p2p node
         client: P2pClient,
-        // the node's peer_id
+        /// the node's peer_id
         peer_id: PeerId,
-        // a channel to recieve network events read by the node
+        /// a channel to recieve network events read by the node
         network_events: Receiver<NetworkEvent>,
-        // listening address for this node
+        /// listening address for this node
         addr: Multiaddr,
-        // multiaddr that is a combination of the listening addr and peer_id
+        /// multiaddr that is a combination of the listening addr and peer_id
         dial_addr: Multiaddr,
     }
 
@@ -1516,24 +1513,15 @@ mod tests {
 
         // when `start_providing` waits for the record to make it to the dht
         // we can remove this polling
-        loop {
-            tokio::select! {
-                _ = tokio::time::sleep(Duration::from_millis(500)) => {
-                    return Err(anyhow!("timed out before we were able to fetch providers off the dht"));
-                },
-                s = test_runner_a.client.fetch_providers_dht(&cid) => {
-                    let stream = s?;
-                    let providers: Vec<_> = stream.try_collect().await.unwrap();
-                    if providers.len() == 0 {
-                        continue;
-                    }
+        let providers = tokio::time::timeout(
+            Duration::from_millis(2500),
+            poll_for_providers(test_runner_a.client.clone(), &cid),
+        )
+        .await
+        .context("timed out before finding providers for the given cid")??;
 
-                    assert!(providers.len() == 1);
-                    assert!(providers.get(0).unwrap().contains(&test_runner_c.peer_id));
-                    break;
-                }
-            }
-        }
+        assert!(providers.len() == 1);
+        assert!(providers.get(0).unwrap().contains(&test_runner_c.peer_id));
 
         // c stop providing
         test_runner_c.client.stop_providing(&cid).await?;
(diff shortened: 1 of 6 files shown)


iroh-p2p/src/node.rs
```diff
@@ -1095,7 +1209,288 @@ mod tests {
...
+            .client
+            .connect(test_runner_b.peer_id, addrs.clone())
+            .await?;
+        test_runner_c
+            .client
+            .connect(test_runner_b.peer_id, addrs.clone())
+            .await?;
+
+        // give time for dht records to update
+        tokio::time::sleep(Duration::from_millis(500)).await;
```
flub, line 1458: Can there be something better than a sleep?  Tests with sleeps almost always become flaky at some point sadly.

iroh-p2p/src/node.rs
```diff
@@ -1095,7 +1209,288 @@ mod tests {
...
+    }
+
+    #[tokio::test]
+    async fn test_dht() -> Result<()> {
+        // set up three nodes
+        // two connect to one
+        // try to connect via id
+        let addr_a: Multiaddr = "/ip4/127.0.0.1/tcp/5009".parse().unwrap();
+        let addr_b: Multiaddr = "/ip4/127.0.0.1/tcp/5010".parse().unwrap();
+        let addr_c: Multiaddr = "/ip4/127.0.0.1/tcp/5011".parse().unwrap();
```
flub, line 1415: Would be really nice if there was a way not to hardcode ports but let the peers allocate their addresses dynamically so it works regardless of what else is using ports (maybe another run of these same tests!).

iroh-p2p/src/node.rs
```diff
@@ -1095,7 +1209,288 @@ mod tests {
...
+
+        // give time for dht records to update
+        tokio::time::sleep(Duration::from_millis(500)).await;
+
+        // c start providing
+        test_runner_c.client.start_providing(&cid).await?;
+
+        // when `start_providing` waits for the record to make it to the dht
+        // we can remove this `sleep` call
+        tokio::time::sleep(Duration::from_millis(500)).await;
```
flub, line 1465: This could be improved right now by polling repeatedly until either the stuff is found on the DHT or it times out, with a large timeout that certainly won't be too short regardless of how slow the machine is on which the tests run.

flub (commented):
Overall I think this is a fairly reasonable approach.  The main thing sticking out to me is that all the ports are assigned locally.  Is something like `/ip4/127.0.0.1/tcp/*` as multiaddr possible where the `*` selects a random port and then you can query for the actual addr once the socket exists?

iroh-p2p/src/node.rs
ramfox: ya, turns out there is a "relatively simple" way to listen for `PeerConnection` events, that I was actually already using in a different test that will work here!

iroh-p2p/src/node.rs
```diff
@@ -1095,7 +1266,290 @@ mod tests {
...
+        };
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_dht() -> Result<()> {
+        // set up three nodes
+        // two connect to one
+        // try to connect via id
+        let cid: Cid = "bafkreieq5jui4j25lacwomsqgjeswwl3y5zcdrresptwgmfylxo2depppq"
```
flub, line 1453: not a code comment but rather me learning ipfs...
I assume this CID refers to some block or file or something stored on IPFS somewhere.  How does this code ever get that cid?  And what does start_providing and stop_providing do with it?

iroh-p2p/src/node.rs
```diff
@@ -1103,7 +1122,13 @@ mod tests {
                 }
             };
             let mut network_config = Config::default_with_rpc(rpc_client_addr.clone());
-            network_config.libp2p.listening_multiaddr = self.addr;
+
+            if let Some(addr) = self.addr {
+                network_config.libp2p.listening_multiaddr = addr;
+            } else {
+                network_config.libp2p.listening_multiaddr = "/ip4/0.0.0.0/tcp/0".parse().unwrap();
```
flub, line 1129: Before you had some listening on 127.0.0.1 instead of all interfaces.  You could still force having to give an addr to `::new()` but describe that you can use `0` as port` and how to retrieve the final addr.  That way those tests are not needlessly exposing themselves entirely to the internet (or local coffee shop wifi where we happen to be running some tests ;))

iroh-p2p/src/node.rs
```diff
@@ -1147,20 +1172,51 @@ mod tests {
...
+                _ = tokio::time::sleep(Duration::from_millis(500)) => {
+                    return Err(anyhow!("timed out before we could get a listening address for the node"));
+                },
+                l = client.listeners() => {
+                    if let Some(a) = l?.get(0) {
+                        return Ok(a.clone());
+                    }
+                }
+            }
+        }
```
flub, line 1204: Doing something with a timeout is often simpler to write with
```
let l = tokio::time::timeout(Duration::from_millis(500), client.listeners()).context("waiting for node listening addr")?;
```

The other thing here is that the timeout applies to each select call, while maybe it should be global.  I'm struggling to find a nice way to do this simply, maybe `futures::stream::unfold` isn't too bad for this.  But it's also possible, and might be easier to follow, to wrap the loop into a separate `async fn` and use that future in the timeout.

iroh-p2p/src/node.rs
```diff
@@ -1095,7 +1266,290 @@ mod tests {
...
+            }
+        };
+
+        // c start providing
+        test_runner_c.client.start_providing(&cid).await?;
+
+        // when `start_providing` waits for the record to make it to the dht
+        // we can remove this polling
+        loop {
+            tokio::select! {
```
flub, line 1520: similar comment on the select and looping i guess

flub (commented):
Nice!  I was wondering if I was asking something impossible with the port allocation but that looks pretty reasonable.

iroh-p2p/src/node.rs
ramfox: good point!

ramfox: Although I do like the API of not having to pass in an address, but rather have the default address be `/ip4/127.0.0.1/tcp/0` instead of `/ip4/0.0.0.0/tcp/0`

iroh-p2p/src/node.rs
ramfox: So the CID "Content Identifier" refers to some block of content. When you provide content on the dht, the `kad` behaviour refers to the identifiers as `Key`s rather than `CID`s, so I'm assuming that means you can provide things other than block content over the dht, though I've never wired it up that way.

`start_providing` makes the CID "available" on the dht. Basically, you store it in the `RecordStore`, and you periodically send messages to the peers you have that are "closest" to the key saying "hey I have this". These records have timeouts, so they are only valid for a certain period of time. I believe the TTL defaults to 24hrs. (https://docs.rs/libp2p/latest/libp2p/kad/struct.Kademlia.html#method.start_providing for more details)

`stop_providing` removes the record from your store. This happens locally & doesn't push any sort of "stop providing" records, it just won't re-provide the record during the next interval. (https://docs.rs/libp2p/latest/libp2p/kad/struct.Kademlia.html#method.stop_providing)

As for how we obtain those CIDs, we currently only call `start_providing` after we have created a DAG from some content that was added to iroh. We create the DAG and call `start_providing` with the root hash of the DAG. This happen in `iroh` after we call `add`

We don't call `stop_providing` anywhere.

The kad record store is currently a mem store, so these records get cleared every time you re-start your node. We have plans to re-provide any root cids from the `iroh-store` on start-up for 0.2.0, but currently each time you restart your node you would need to "re-add" the content from the command line to trigger any providing. Details here: https://github.com/n0-computer/iroh/issues/353

flub (approved):

iroh-p2p/src/node.rs
```diff
@@ -1044,36 +1058,190 @@ mod tests {
         Ok(())
     }
 
+    struct TestRunnerBuilder {
+        /// optional listening address for this node, when None, the swarm will connect to a random
+        /// tcp port
```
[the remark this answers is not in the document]
ramfox: We should definitely follow style guidelines, I'm just ignorant. thank you!!!

iroh-p2p/src/node.rs
```diff
@@ -1044,36 +1058,190 @@ mod tests {
...
+        addr: Option<Multiaddr>,
+        /// listening addresses for the p2p client, when None, the client will communicate over
+        /// a memory rpc channel
+        rpc_addrs: Option<(P2pServerAddr, P2pClientAddr)>,
+        /// when true, allow bootstrapping to the network, otherwise don't provide any addresses
+        /// from which to bootstrap
+        bootstrap: bool,
```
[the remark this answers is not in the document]
ramfox: If you are curious about bootstrap nodes: https://github.com/n0-computer/iroh/blob/579c8edb13088190ceca0be4cbf6d7f5b563932d/iroh-p2p/src/config.rs#L22-L31

iroh-p2p/src/node.rs
```diff
@@ -1044,36 +1058,190 @@ mod tests {
         Ok(())
     }
 
+    struct TestRunnerBuilder {
```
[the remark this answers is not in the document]
flub: Never mind then.  I think it's good practice to derive those two when possible, but if it doesn't work for good reason then it's fine not to.

b5 (approved):
_stamped!_

flub (approved):

PR #490, opened 2022-11-12
refactor: remove anyhow from iroh-util

b5:
I've created this to help solidify our error conventions when removing anyhow so @matthiasbeyer & core maintainers can agree on a set of idioms for any potential replacing work.

Overall I'm super happy with the way this clears up the API for the utils crate.


b5:
cc @matthiasbeyer, I'd welcome a review 😄

matthiasbeyer (changes requested):
Thanks a lot for keeping me in the loop :heart: :rocket: 

I have one patch for this PR, please pull from https://github.com/matthiasbeyer/iroh/tree/b5/replace_anyhow_iroh_utils (commit https://github.com/matthiasbeyer/iroh/commit/c703541fd4b9b9b2aa059c9dbec388b5e9941969). You probably want to squash this into [deeeeda](https://github.com/n0-computer/iroh/pull/490/commits/deeeeda418e3ba2bf4aca5f49a69656b39aed1e4)!

Besides that, this looks like a good starting point! :heart:

matthiasbeyer:
FWIW, that's not the patch I wrote! I'm okay with it this time, as the `ok_or` argument does not allocate.

b5:
> FWIW, that's not the patch I wrote! I'm okay with it this time, as the ok_or argument does not allocate.

Clippy complains about unnecessary closures without this change

iroh-util/src/lib.rs
```diff
@@ -229,6 +231,22 @@ pub fn increase_fd_limit() -> std::io::Result<u64> {
     Ok(soft)
 }
 
+/// Alias for a `Result` with the error type set to `UtilError`
+pub type Result<T> = result::Result<T, UtilError>;
+
+#[derive(Error, Debug)]
+pub enum UtilError {
```
dignifiedquire, line 238: the convention is to call it just Error, so that in imports it will be iroh_utils::Error

iroh-util/src/lock.rs
```diff
@@ -1,16 +1,13 @@
-use crate::exitcodes;
-use anyhow::{anyhow, Result as AnyhowResult};
-use std::fs::File;
-use std::io::prelude::*;
-use std::io::ErrorKind;
-use std::io::Write;
-use std::path::PathBuf;
-use std::process;
+use std::{fs::File, io, io::ErrorKind, io::Read, io::Write, path::PathBuf, process, result};
```
flub, line 1: out of curiosity, do we explicitly follow the `module` import granulariy style (as can be configured in rust-analzer)?

iroh-util/src/lib.rs
```diff
@@ -60,8 +61,9 @@ pub async fn block_until_sigint() {
 /// | macOS    | `$HOME`/Library/Application Support/iroh   | /Users/Alice/Library/Application Support/iroh |
 /// | Windows  | `{FOLDERID_RoamingAppData}`/iroh           | C:\Users\Alice\AppData\Roaming\iroh   |
 pub fn iroh_config_root() -> Result<PathBuf> {
-    let cfg = dirs_next::config_dir()
-        .ok_or_else(|| anyhow!("operating environment provides no directory for configuration"))?;
+    let cfg = dirs_next::config_dir().ok_or(UtilError::NoDir {
+        purpose: "configuration data",
+    })?;
```
flub, line 66: I'm not really sure I like this style of using `UtilError` much.  This is now an error type that's a grab bag of errors that can happen for functions in this module.  It does help you handling the error from e.g. this function itself since you have to guess (I mean, the typesystem doesn't help you) which error variants you should handle for this function.  So it's not really any better then using `anyhow::Error`.

As for error ergonomics, I'd expect the error to preserve the underlying error.  It's all good saying there's no directory, but was that because the directory entry did not exist or because there was no read permission on the direntry?

To me a pretty reasonable way of improving this error message is to do something like `dirs_next::config_dir().context("no directory for configuration")?;`.  For something that can raise a single error variant only there's fairly low value in starting to provide custom error types, it would only wrap the single underlying error.  What you want to do is enrich it with the usage information, which is what anyhow's `Context` was made for.

iroh-util/src/lock.rs
b5: no explicit style at the moment. I've been winging this, and would love to coalesce around a style choice

iroh-util/src/lib.rs
b5: > This is now an error type that's a grab bag of errors that can happen for functions in this module. 

I disagree that it's a grab bag. It's an enumeration of the types of errors this module can emit. If it looks like a grab bag, it's probably because the module API is poorly constructed.

It does help you handling the error from e.g. this function itself since you have to guess (I mean, the typesystem doesn't help you) which error variants you should handle for this function.


>  As for error ergonomics, I'd expect the error to preserve the underlying error.

Yes totally. I'll fix this. It's wrong to swallow the underlying error. I also thing your point generalizes, we should always be propagating the underlying error.

>  For something that can raise a single error variant only there's fairly low value in starting to provide custom error types, it would only wrap the single underlying error. 

Up in application developer space, sure. But the motivation for this change is to make Iroh easier to consume as a library. Aside from not forcing our consumers to import `anythow`, the benefit of this change making a more explicit contract about the kinds of errors an iroh library will emit.

I like anyhow for application developers, as a library developer I find it lazy, and as a user of libraries that return `anyhow::Result` I've already encountered situations where `anyhow` is just too vague to be useful. I needed to make this conversion with `LockError` prior to our v0.1.0 release to be able to handle problems with program locks (which are common), and [provide feedback on missing services](https://github.com/n0-computer/iroh/blob/dc6f85925f9ad4681d0a647a1803180bedde28ee/iroh/src/main.rs#L41-L54).

But all that said, this is an important decision to make as a team. This overhead of enumerating error types will slow down the dev process, and I think there's a case to be made for doing this slowly over time with APIs as we stabilize them. The counterpoint is similar to our windows support story, not doing the work as we go sets us up for a massive amount of catch up work later, and may leave us exposed to having to undo decisions & revisit problems

dignifiedquire:
@b5 what's the plan & status on this?

b5:
I'd like to back out the `UtilError`, but keep the lock error refinements. Will handle in the next 48 hours

b5:
ok changes made. @dignifiedquire this is ready for review

flub (approved):

dignifiedquire (approved):

Issue #491, opened 2022-11-12
remove anyhow in all lib crates


b5:
Things we need to establish a convention for what we doing about `anyhow::Result` return types?

I've opened #490 to explore idioms, and in this one I'm experimenting with defining a `Result` alias in each module that replaces the `anyhow::Result` type.

Crates we can skip:
- iroh

Crates that need replacing:
- [ ] `iroh-api`: ~13 errors
- [ ] `iroh-bitswap`
- [ ] `iroh-car`
- [ ] `iroh-gatway`
- [ ] `iroh-localops`
- [ ] `iroh-metrics`
- [ ] `iroh-one`
- [ ] `iroh-p2p`
- [ ] `iroh-resolver`
- [ ] `iroh-rpc-client`
- [ ] `iroh-rpc-types`
- [ ] `iroh-store`
- [ ] `iroh-util` #490

matthiasbeyer:
Btw you can tick off `iroh-car`, it does not use `anyhow`! :wink:

matthiasbeyer:
In my efforts in #492 I noticed that all the crates are _so much_ integrated into eachother that it is not possible to replace anyhow one-by-one. Thus, my patches would transform all crates in one go ... I guess I'll submit a completely new PR once I am done and then we will see whether we can still split it up (I'll commit as cleanly as possible) or whether a one-step-for-everything transformation is feasible :thinking:

flub:
I'm not sure doing this blanket for all errors makes a lot of sense.  Especially if they're being replaced with module-wide catchall error types.

To me the main motivation to switch away from anyhow to a dedicated error type is if you notice you're having to use anyhow's downcasting to handle a particular error.  From a library's public API point of view that's kind of painful since every time you switch error type you make a backwards-incompatible change.  However the only way to avoid this is to give (almost) each function it's dedicated error type that is future-extensible.  That's also not a very practical solution.

matthiasbeyer:
> To me the main motivation to switch away from anyhow to a dedicated error type is if you notice you're having to use anyhow's downcasting to handle a particular error. From a library's public API point of view that's kind of painful since every time you switch error type you make a backwards-incompatible change.

This extends also to the following issue: If the library uses a Stringly typed error (`bail!("Something happened")`), then downcasting does not even help, you would also need to do string-matching on the user side of things...
And I guess we all know that stringly typed APIs are hell :laughing: 

[...]

PR #492, opened 2022-11-13
iroh-rpc-types: Rewrite error handling story

matthiasbeyer:
Part of #491 

This removes `anyhow` from the crate and replaces it with `thiserror`.

There's one TODO left, because there's a `String` somewhere used as error type, which is not nice. But for the first step, this might suffice.

### DIFF
--- a/iroh-rpc-client/src/network.rs
+++ b/iroh-rpc-client/src/network.rs
@@ -161,12 +161,12 @@ impl P2pClient {
                 peer_id: peer_id.to_bytes(),
                 addrs: addrs.iter().map(|a| a.to_vec()).collect(),
             };
-            self.backend.peer_connect(req).await
+            self.backend.peer_connect(req).await.map_err(anyhow::Error::from)
         } else {
             let req = ConnectByPeerIdRequest {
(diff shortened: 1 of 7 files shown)


matthiasbeyer:
FWIW, I also have a patch for `iroh-rpc-client`, but as it depends on this PR, I will only file it after this is merged.

matthiasbeyer:
This is obviously not ready yet, but I published it so nobody does double the work... :wink:

matthiasbeyer:
Hah, I begin to see why you said that it won't be that easy... because there's this macro hell in iroh-rpc-types :laughing: 

I will get there though.

b5:
if this ends up being too nighmarish, I'd suggest we leave it. We're actively trying to replace gRPC (and this macro insanity) in the next 4-6 weeks, we might be able to get away with leaving error handling until after such a transition

matthiasbeyer:
Why not the other way round? Get a good error story and then remove the grpc stuff,... this way you end up with the macro nightmare removed _and_ a good interface in terms of error handling! :laughing:

b5:
superseded by #495

PR #494, opened 2022-11-14
feat: initial arm64 dockerfile work

b5:


xtask/src/main.rs
```diff
@@ -130,3 +137,31 @@ fn project_root() -> PathBuf {
...
+        images = vec![String::from("iroh-one"), String::from("iroh-p2p")];
+    }
+
+    for image in images {
+        let status = Command::new("docker")
+            .current_dir(project_root())
+            .args([
+                "build",
+                "-t",
+                format!("{}:distroless", image).as_str(),
```
b5, line 153: this tags images in the form iroh-one:distroless... that's probably wrong

docker/Dockerfile.iroh-gateway
```diff
@@ -0,0 +1,51 @@
...
+# has the side effect of updating the crates.io index & installing rust toolchain
+# called in a separate step for nicer caching. the command itself will fail,
+# b/c empty-library is not a dependency, so we override with an exit code 0
+RUN cargo install empty-library; exit 0
+
+WORKDIR /iroh
+
+COPY ../ .
+
+RUN cargo build --bin iroh-gateway --release
```
Arqu, line 32: Can we build with `cargo build --profile ci --bin iroh-gateway`. This uses the release profile we want to maintain for binaries (additional perf optimizations & slimmer binaries) and also makes future changes simple `.toml` changes instead of needing to touch all scripts.

docker/Dockerfile.iroh-gateway
```diff
@@ -0,0 +1,51 @@
...
+################################################################################
+## Final image
+################################################################################
+FROM gcr.io/distroless/cc
+
+WORKDIR /iroh
+
+# Copy our build, changing owndership to distroless-provided "nonroot" user,
+# (65532:65532)
+COPY --from=builder --chown=65532:65532 /iroh/target/release/iroh-gateway ./
```
Arqu, line 43: I would love to validate the file descriptor limits get raised/set correctly.

docker/Dockerfile.iroh-gateway
b5: how about we do one-better & create a `docker` profile that matches `ci` so they can change independently?

docker/Dockerfile.iroh-one
```diff
@@ -0,0 +1,57 @@
...
+# has the side effect of updating the crates.io index & installing rust toolchain
+# called in a separate step for nicer caching. the command itself will fail,
+# b/c empty-library is not a dependency, so we override with an exit code 0
+RUN cargo install empty-library; exit 0
+
+WORKDIR /iroh
+
+COPY ../ .
+
+RUN cargo build --bin iroh-one --release
```
Arqu, line 38: Same as above

docker/Dockerfile.iroh-one
```diff
@@ -0,0 +1,57 @@
...
+
+# Copy our build, changing owndership to distroless-provided "nonroot" user,
+# (65532:65532)
+COPY --from=builder --chown=65532:65532 /iroh/target/release/iroh-one ./
+
+# Use nonroot (unprivileged) user
+USER nonroot
+
+# expose gateway port
+EXPOSE 9050
```
Arqu, line 55: Unsure what the current story on `iroh-one` <> `rpc api` is but might want to open 4400-4403

docker/Dockerfile.iroh-p2p
```diff
@@ -0,0 +1,51 @@
...
+# has the side effect of updating the crates.io index & installing rust toolchain
+# called in a separate step for nicer caching. the command itself will fail,
+# b/c empty-library is not a dependency, so we override with an exit code 0
+RUN cargo install empty-library; exit 0
+
+WORKDIR /iroh
+
+COPY ../ .
+
+RUN cargo build --bin iroh-p2p --release
```
Arqu, line 32: Same

docker/Dockerfile.iroh-p2p
```diff
@@ -0,0 +1,51 @@
...
+
+# Copy our build, changing owndership to distroless-provided "nonroot" user,
+# (65532:65532)
+COPY --from=builder --chown=65532:65532 /iroh/target/release/iroh-p2p ./
+
+# Use nonroot (unprivileged) user
+USER nonroot
+
+# expose the default RPC port
+EXPOSE 4401
```
Arqu, line 49: This is just the RPC port, we also have `4444` for all the p2p traffic as a default.

xtask/src/main.rs
Arqu: yeah, for starters let's do `:commit_hash` and then builders can also tag with `latest`

docker/Dockerfile.iroh-gateway
Arqu: sgtm, I can push that update then

b5: pretty sure I can handle in this PR, it's just a change to the workspace `Cargo.toml`, right?

docker/Dockerfile.iroh-gateway
b5: I believe these are set at runtime, no?

Arqu: Yes, but docker is weird. It sets up to max ~65k at runtime but if not root the "hard limit" specified on the system level might be low (usually 1024). Since it's a non-root user, we might hit that easily. The proper thing should be setting the FD limit on the image directly as the system level config. Then the runtime can self manage.

Arqu: NVM just confirmed, this should be handled by the host system running docker, either by configuring the docker runner directly or supplying `--ulimit=12321312` when running stuff. Either way think this is a solid point for docs.

Arqu (dismissed):
LGTM, just left a comment to document the fd limit stuff.

PR #495, opened 2022-11-14
Replace anyhow

### LINKED ISSUE #491: remove anyhow in all lib crates

matthiasbeyer:
Replaces #492 
Closes #491 

---

As I found in #492, it is not really possible (or at least not easy) to replace error types one-by-one in this project, because all the crates are so deeply integrated with eachother.

---

This is my ongoing effort in replacing anyhow with thiserror in the whole codebase. Of course this is not perfect yet, there are way too many error variants in many cases, there might even be error variants that are duplicated but could be merged...
There might also be things that could be optimized, error variants that could be more expressive, carry more context, be more detailed.

This is an effort to make it work, not to make it pretty. Making it pretty would be the next step.

---

Please note that this branch will be rebased as soon as master moves.

Also: Consider this WIP until CI is green!


b5:
Thanks so so much for your understanding @matthiasbeyer. the code may not have changed, but our approach has. Appreciated!

PR #497, opened 2022-11-15
refactor(ops): CI rework

Arqu:
**DO NOT MERGE EVEN AFTER APPROVAL/REVIEW**

Move completely over to Github Actions. The reasoning is mainly to consolidate and to shift to self hosted runners. CircleCI self hosted runners don't allow for forks to run builds.
We've also enforces mandatory approval for all external PRs. Moving forward please review changes to CI code carefully.

Results:
- own infra
- much speed
- less checks to wait for
- less maintenance / CI code

Things that will be done in a follow up PR:
- add more runners, specifically switch over windows builds to self hosted and swap mac from intel based to m1 based platforms


.github/workflows/ci.yml
```diff
@@ -138,6 +74,13 @@ jobs:
       with:
         command: check
         args: --workspace --all-features --bins --tests --examples --benches
+    
+    - name: clippy
+      uses: actions-rs/cargo@v1
+      if: matrix.os == 'ubuntu-latest'
+      with:
+        command: clippy
+        args: --workspace --all-features --tests --benches --examples --bins --all-targets -- -D warnings
```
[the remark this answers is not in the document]
flub: to be fair, we can do this as a followup since that simplifies things in a lot of places like readme's etc as well

.github/workflows/ci.yml
```diff
@@ -150,47 +93,85 @@ jobs:
...
+            release-arch: x86_64
+            runner: [self-hosted, macOS, X64]
+          # - os: windows-latest
+          #   release-os: windows
+          #   release-arch: amd64
+          #   runner: [windows-latest]
+    env:
+      RUST_BACKTRACE: full
+      # Force not building debuginfo to save space on disk.
+      RUSTFLAGS: "-C debuginfo=0"
```
flub, line 143: For releases we probably want to build debuginfo and use split debuginfo and make it available in the release as well.  But maybe that's also something that is better done later.

.github/workflows/ci.yml
```diff
@@ -21,111 +19,49 @@ jobs:
...
-            sccache-path: "%LOCALAPPDATA%\\sccache"
             release-os: windows
             release-arch: amd64
+            runner: [windows-latest]
     env:
       RUST_BACKTRACE: full
       # Force not building debuginfo to save space on disk.
       RUSTFLAGS: "-C debuginfo=0"
-      RUSTC_WRAPPER: sccache
+      # Force all builds to be non-incremental - better for cache
```
flub, line 37: I think you forgot to add the `CARGO_INCREMENTAL` var here.

.github/workflows/ci.yml
```diff
@@ -138,6 +74,13 @@ jobs:
       with:
         command: check
         args: --workspace --all-features --bins --tests --examples --benches
+    
+    - name: clippy
+      uses: actions-rs/cargo@v1
+      if: matrix.os == 'ubuntu-latest'
```
flub, line 80: Why not run clippy on all platforms?  We've already seen windows-only clippy errors, because of optionally compiled code around unix sockets.

.github/workflows/ci.yml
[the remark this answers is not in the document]
Arqu: made an issue for the readme will cover it in CI right away.

.github/workflows/ci.yml
Arqu: good point.

b5 (approved):
🤠

PR #498, opened 2022-11-15
chore(deps): Unify the anyhow version dependency

flub:
Since iroh is a library, or will be, we should not have too strict a
version requirement, so simply requiring version 1.


b5 (approved):
yessss tyvm

PR #500, opened 2022-11-16
feat: support lookup names in RPC addrs

b5:
support addrs like `grpc://example:4444`, which is required for DevOps infrastructure that provides hostname resolution as part of service discovery.

This patch uses the heuristic of being able to resolve to disambiguate between `grpc://foo` the lookup address and `grpc://foo` a path to a unix domain socket file location. In practice this shouldn't be too big a deal, but might get tricky if folks start naming UDS paths after domains 😵 .

I need this to make docker compose examples work, which is a good indication this'll be needed by all modern DevOps deployments

### DIFF
--- a/iroh-rpc-types/src/addr.rs
+++ b/iroh-rpc-types/src/addr.rs
@@ -1,6 +1,6 @@
 use std::{
     fmt::{Debug, Display},
-    net::SocketAddr,
+    net::{SocketAddr, ToSocketAddrs},
     str::FromStr,
 };
 
@@ -12,6 +12,8 @@ use tokio::sync::mpsc::{Receiver, Sender};
 pub enum Addr<T = ()> {
     #[cfg(feature = "grpc")]
     GrpcHttp2(SocketAddr),
+    #[cfg(feature = "grpc")]
+    GrpcHttp2Lookup(String),
     #[cfg(all(feature = "grpc", unix))]
     GrpcUds(std::path::PathBuf),
     #[cfg(feature = "mem")]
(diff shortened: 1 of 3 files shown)


iroh-rpc-types/src/macros.rs
```diff
@@ -19,6 +19,11 @@ macro_rules! proxy_serve {
                         Ok(())
                     }
 
+                    #[cfg(feature = "grpc")]
+                    $crate::Addr::GrpcHttp2Lookup(name) => {
+                        anyhow::bail!("cannot serve on lookup address: {}", name);
```
Arqu, line 24: Aren't we moving away from anyhow?

Arqu (approved):
LGTM, just a nit on the anyhow but that can be done in a separate PR once we start.

iroh-rpc-types/src/addr.rs
```diff
@@ -114,6 +127,18 @@ mod tests {
         assert_eq!(addr.to_string(), "grpc://198.168.2.1:1234");
     }
 
+    // TODO(b5): only running this on unix b/c windows doesn't have localhost
+    // enabled by default
```
flub, line 131: you could start by doing the localhost lookup locally and exit the test early if it doesn't resolve.  that way the test can run everywhere and will do something useful if localhost exists.

iroh-rpc-types/src/macros.rs
flub: The conclusion was yes we are, but not yet right now - only sometime before 1.0.

flub (approved):

PR #502, opened 2022-11-16
chore: Disable uds-gateway feature on windows

flub:
Perhaps it is even desireable to trigger a compilation failure when
trying to enable the feature on windows.  The downside is that the
`--all-features` argument to cargo would no longer work on windows and
this makes the development experience a lot more painful.  For now
this seems a reasonable compromise.


Arqu (approved):
LGTM, lets just get CI to pass

Arqu (approved):
:shipit:

PR #505, opened 2022-11-16
ref: Fix todo by using HashSet::difference

flub:
This is probably less efficient though, for this size of hashset.  But
it is a higher level abstraction so probably is a bit easier to read.


b5 (approved):
🙌  thanks!

PR #506, opened 2022-11-17
fix(cli): Respect the --cfg flag

flub:
Use the --cfg flag to load the configuration file for `iroh`.

This does not affect the config files for any of the managed
processes like iroh-p2p etc.

This flag was not used before.

Also optimises the make_config call a little by taking a slice
rather than a Vec.  And assert and document the priority order.

### DIFF
--- a/iroh-util/Cargo.toml
+++ b/iroh-util/Cargo.toml
@@ -23,5 +23,8 @@ thiserror = "1.0"
 toml = "0.5.9"
 tracing = "0.1.34"
 
+[dev-dependencies]
+testdir = "0.6.0"
+
 [target.'cfg(unix)'.dev-dependencies]
 nix = "0.25"
--- a/iroh/src/run.rs
+++ b/iroh/src/run.rs
@@ -85,12 +85,12 @@ enum Commands {
 impl Cli {
     pub async fn run(&self) -> Result<()> {
         let config_path = iroh_config_path(CONFIG_FILE_NAME)?;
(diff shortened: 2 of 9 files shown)


iroh/src/run.rs
```diff
@@ -85,8 +85,7 @@ enum Commands {
 impl Cli {
     pub async fn run(&self) -> Result<()> {
         let config_path = iroh_config_path(CONFIG_FILE_NAME)?;
-        // TODO(b5): allow suppliying some sort of config flag. maybe --config-cli?
-        let sources = vec![Some(config_path.as_path())];
+        let sources = vec![self.cfg.as_deref(), Some(config_path.as_path())];
```
flub, line 88: huh, all the other commands line interfaces seem to order this the other way around.

flub:
Added some test to figure out priority and documented it.  Fixed the source order appropriately.

Also optimised a vec into a slice, because it seemed a bit odd to have it as a vec.

I'd say this is ready for review now.

iroh-util/Cargo.toml
```diff
@@ -23,5 +23,8 @@ thiserror = "1.0"
 toml = "0.5.9"
 tracing = "0.1.34"
 
+[dev-dependencies]
+testdir = "0.6.0"
```
b5, line 27: 🔥

b5 (approved):
looks great!

b5:
macOS failure is unrelated. merging!

PR #507, opened 2022-11-17
fix(iroh-store): `put_many` bug

### LINKED ISSUE #423: Intermittent add/get roundtrip failure with real store
Branch for investigation: https://github.com/n0-computer/iroh/tree/rklaehn/get-bug-investigatin

ramfox:
closes #423

### DIFF
--- a/iroh-store/src/store.rs
+++ b/iroh-store/src/store.rs
@@ -1,6 +1,6 @@
-use std::collections::HashSet;
 use std::{sync::Arc, thread::available_parallelism};
 
+use ahash::AHashSet;
 use anyhow::{anyhow, bail, Context, Result};
 use bytes::Bytes;
 use cid::Cid;
@@ -367,9 +367,9 @@ impl<'a> WriteStore<'a> {
         let mut total_blob_size = 0;
 
         let mut batch = WriteBatch::default();
-        let mut cid_tracker: HashSet<Cid> = std::collections::HashSet::default();
+        let mut cid_tracker: AHashSet<Cid> = AHashSet::default();
         for (cid, blob, links) in blocks.into_iter() {
-            if self.has(&cid)? || cid_tracker.contains(&cid) {
+            if cid_tracker.contains(&cid) || self.has(&cid)? {
                 continue;
             }
 
--- a/iroh-store/src/store.rs
+++ b/iroh-store/src/store.rs
@@ -1,3 +1,4 @@
+use std::collections::HashSet;
 use std::{sync::Arc, thread::available_parallelism};
 
 use anyhow::{anyhow, bail, Context, Result};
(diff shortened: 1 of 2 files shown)


iroh-store/src/store.rs
```diff
@@ -366,11 +367,14 @@ impl<'a> WriteStore<'a> {
         let mut total_blob_size = 0;
 
         let mut batch = WriteBatch::default();
+        let mut cid_tracker: HashSet<Cid> = std::collections::HashSet::default();
         for (cid, blob, links) in blocks.into_iter() {
-            if self.has(&cid)? {
-                return Ok(());
+            if self.has(&cid)? || cid_tracker.contains(&cid) {
+                continue;
```
rklaehn, line 373: Yes, obvious in retrospect, like so many bugs...

iroh-store/src/store.rs
```diff
@@ -366,11 +367,14 @@ impl<'a> WriteStore<'a> {
         let mut total_blob_size = 0;
 
         let mut batch = WriteBatch::default();
+        let mut cid_tracker: HashSet<Cid> = std::collections::HashSet::default();
```
rklaehn, line 370: I would use an AHashSet ( https://docs.rs/ahash/latest/ahash/ ) since it is faster.

Dig is quite fond of it, and it is already in our dependencies anyway.

iroh-store/src/store.rs
```diff
@@ -366,11 +367,14 @@ impl<'a> WriteStore<'a> {
...
         for (cid, blob, links) in blocks.into_iter() {
-            if self.has(&cid)? {
-                return Ok(());
+            if self.has(&cid)? || cid_tracker.contains(&cid) {
```
rklaehn, line 372: Not that it matters that much, but do the cid_tracker membership test first. It is much faster than having to go all the way to the DB.

iroh-store/src/store.rs
```diff
@@ -911,4 +915,32 @@ mod tests {
         assert_eq!(Vec::<String>::new(), store.consistency_check()?);
         Ok(())
     }
+
+    #[tokio::test]
+    async fn test_put_many_repeat_ids() -> anyhow::Result<()> {
```
rklaehn, line 920: :+1:

rklaehn (approved):
See the comments for some minor improvements

iroh-store/src/store.rs
ramfox: love these little pieces of advice!

iroh-store/src/store.rs
ramfox: 🔥

iroh-store/src/store.rs
```diff
@@ -367,9 +367,9 @@ impl<'a> WriteStore<'a> {
         let mut total_blob_size = 0;
 
         let mut batch = WriteBatch::default();
-        let mut cid_tracker: HashSet<Cid> = std::collections::HashSet::default();
+        let mut cid_tracker: AHashSet<Cid> = AHashSet::default();
```
rklaehn, line 370: If we had a vec we could use https://docs.rs/ahash/latest/ahash/struct.AHashSet.html#method.with_capacity here to make sure the hashset does not need to rehash.

But with the iterator we could just use the size hint, which is probably not worth it.

rklaehn (approved):

iroh-store/src/store.rs
dignifiedquire: `HashSet`s have `with_capacity`: https://docs.rs/ahash/latest/ahash/struct.AHashSet.html#method.with_capacity

PR #508, opened 2022-11-17
feat(cli): Environment variables to override directories

flub:
Adding the IROH_CONFIG_DIR, IROH_DATA_DIR and IROH_CACHE_DIR
environment variables to override the OS-dependent locations.


iroh-util/src/lib.rs
```diff
@@ -52,14 +53,19 @@ pub async fn block_until_sigint() {
...
+/// table.
 ///
 /// | Platform | Value                                 | Example                          |
 /// | -------- | ------------------------------------- | -------------------------------- |
 /// | Linux    | `$XDG_CONFIG_HOME` or `$HOME`/.config/iroh | /home/alice/.config/iroh              |
 /// | macOS    | `$HOME`/Library/Application Support/iroh   | /Users/Alice/Library/Application Support/iroh |
 /// | Windows  | `{FOLDERID_RoamingAppData}`/iroh           | C:\Users\Alice\AppData\Roaming\iroh   |
 pub fn iroh_config_root() -> Result<PathBuf> {
+    if let Some(val) = env::var_os("IROH_CONFIG_DIR") {
+        return Ok(PathBuf::from(val));
```
b5, line 67: what happens when val isn't a valid filepath? Should this func check & error out?

iroh-util/src/lib.rs
```diff
@@ -240,17 +256,70 @@ pub fn increase_fd_limit() -> std::io::Result<u64> {
...
     use serde::Deserialize;
     use testdir::testdir;
 
     use super::*;
 
+    /// Helper struct to restore an environment variable once dropped.
+    struct EnvDropGuard {
+        name: OsString,
+        value: Option<OsString>,
+    }
```
b5, line 270: delightful

iroh-util/src/lib.rs
```diff
@@ -240,17 +256,70 @@ pub fn increase_fd_limit() -> std::io::Result<u64> {
...
+                None => {
+                    env::remove_var(self.name.as_os_str());
+                }
+            }
+        }
+    }
+
     #[test]
     fn test_iroh_config_path() {
         let got = iroh_config_path("foo.bar").unwrap();
```
b5, line 298: we should rename this test. how about `test_iroh_directory_paths`?

iroh-util/src/lib.rs
```diff
@@ -289,4 +358,6 @@ mod tests {
         .unwrap();
         assert_eq!(cfg.item, "one");
     }
+    #[test]
+    fn test_env_override() {}
```
b5, line 362: remove, covered above 😄

b5 (dismissed):
LGTM

iroh-util/src/lib.rs
```diff
@@ -240,16 +256,69 @@ pub fn increase_fd_limit() -> std::io::Result<u64> {
...
 #[cfg(test)]
 mod tests {
+    use std::ffi::{OsStr, OsString};
+
     use serde::Deserialize;
 
     use super::*;
 
+    /// Helper struct to restore an environment variable once dropped.
+    struct EnvDropGuard {
```
dignifiedquire, line 266: there is this nice crate by @vmx https://crates.io/crates/temp-env so one doesn't have to reimplement this every time :)

dignifiedquire:
Why are these not scoped to the libraries? I understand the desire to override all at once, but I was expecting to be able to override these for each library by using eg `IROH_GATEWAY_CFG_PATH`.

b5:
> I was expecting to be able to override these for each library by using eg IROH_GATEWAY_CFG_PATH

oh really, why would we want to do that? And can't that be accomplished with process isolation? for example:
```
$ IROH_CONFIG_DIR=/gateway && iroh-gateway
```

At the library level I'd expect we'd want some sort of constructor argument override

dignifiedquire:
Two cases I see (1) when the library is used in the binary, and your deployment wants to set different directories (2) when the libraries are all combined into (iroh)-one. Not saying it is a big use case, but I find it surprising that this overrides it everywhere tbh, given that most/all other config is fully scoped currently.

b5:
Where are we at on this? @dignifiedquire do we want to merge & add binary-specific env in a follow up?

flub:
> Why are these not scoped to the libraries? I understand the desire to override all at once, but I was expecting to be able to override these for each library by using eg `IROH_GATEWAY_CFG_PATH`.

If we'd want this I'd do it additional.  Though I also the individual services was already covered by using their `--cfg` options.  Also I started wanting this when invoking `iroh start` etc, so controlling all the services at once and then it seemed natural to give it a single env var for this.

We could complete this by adding the `IROH_$SVC_CONFIG_PATH` files as well I guess, do you think we should add this right now?

dignifiedquire (approved):
this is a good first pass, so lets get this

PR #509, opened 2022-11-17
fix: Switch to tempfile for temporary test directories

flub:
This removes the testdir crate for a temprorary testing directory and
uses the tempfile crate instead.  It seems a transitive dependency for
testdir does not work on mac, which is kind of required for us.


ramfox (approved):
LGTM!

PR #510, opened 2022-11-18
chore: add docker buildx xtask

b5:
much easier to do multi-arch builds with buildx, at the cost of taking eons & needing every resource the computer has. This is mainly here to document the technique for later inclusion in CI, and for my own quality of life

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -44,6 +44,18 @@ enum Commands {
         /// Set of services to publish. Any of {iroh-store,iroh-p2p,iroh-gateway,iroh-one}
         images: Vec<String>,
     },
+    #[command(
+        about = "build & push multi-platform docker images. requires image push credentials."
+    )]
+    DockerBuildx {
+        /// Publish all images, overrides images args
+        #[clap(short, long)]
+        all: bool,
+        /// Set of services to publish. Any of {iroh-store,iroh-p2p,iroh-gateway,iroh-one}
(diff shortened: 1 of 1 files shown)


Arqu:
> much easier to do multi-arch builds with buildx, at the cost of taking eons & needing every resource the computer has. This is mainly here to document the technique for later inclusion in CI, and for my own quality of life

Imma go cry in my own sad CI corner :D

xtask/src/main.rs
```diff
@@ -196,6 +213,65 @@ fn build_docker(all: bool, build_images: Vec<String>, progress: String) -> Resul
...
+                ".",
+            ])
+            .status()?;
+
+        if !status.success() {
+            Err(anyhow::anyhow!("docker buildx failed"))?;
+        }
+    }
+
+    // let status = Command::new("docker")
```
Arqu, line 259: Can we drop dead code?

Arqu (approved):
LGTM, `buildx` is definitely the way to go.

xtask/src/main.rs
b5: normally yes, but in this case I'd like to have it in commit history, and build upon it in a future iteration. If buildx is the way to go, I'll need to put more time into this command soon

Arqu:
`buildx` is for building docker images for different platforms from your host. Code incidentally also compiles for the dockerized platform, but you kinda get the same thing.

PR #516, opened 2022-11-18
tests: disable flakey test_dht

Arqu:
Really flakey, on new CI runners (m1 mac) it fails consistently.

### DIFF
--- a/iroh-p2p/src/node.rs
+++ b/iroh-p2p/src/node.rs
@@ -1445,6 +1445,7 @@ mod tests {
         Ok(())
     }
 
+    #[ignore = "flakey"]
     #[tokio::test]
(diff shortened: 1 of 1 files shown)


b5 (approved):
let's make sure it turns non-windows CI green, then good to go

iroh-p2p/src/node.rs
```diff
@@ -1445,6 +1445,7 @@ mod tests {
         Ok(())
     }
 
+    #[ignore = "flakey"]
```
dignifiedquire, line 1448: is there an issue to track this, why is there no todo to review this test and fix it?

b5: Sure thing: https://github.com/n0-computer/iroh/issues/518

PR #522, opened 2022-11-18
fix: iroh start service set construction

b5:
~~I'm going to collect up a bunch of fixes into one PR prior to releasing, adding tests wherever I can.~~
nm, that's a terrible idea. Will split into multiple PRs. This one is just for the service set now


iroh/tests/cmd/start_status_stop.trycmd
```diff
@@ -0,0 +1,14 @@
+```
+$ iroh start gateway
```
b5, line 2: this has to test individual services because we use a `HashSet` to determine sets, and order isn't deterministic, which makes testing the output nondeterministic ☹️

iroh/src/services.rs
```diff
@@ -82,7 +82,7 @@ async fn start_services(api: &Api, services: HashSet<&str>) -> Result<()> {
         });
 
     let missing_services: HashSet<&str> = services
-        .difference(missing_services)
+        .intersection(missing_services)
```
b5, line 85: this was breaking a front-of-house command and had no test. Has a test now, and a fix

iroh/tests/cmd/start_status_stop.trycmd
dignifiedquire: you are looking for a BTreeSet, or just use a vec it’s fine

dignifiedquire (approved):

iroh/src/services.rs
flub: oops

Issue #512, opened 2022-11-18
Wire up CORS config to a CORS middleware


fabricedesre:
We can use the support from https://docs.rs/tower-http/0.3.4/tower_http/cors/index.html easily. Most of the work will be to parse the config to map it to the middleware structs.

PR #523, opened 2022-11-18
Add a CORS middleware

fabricedesre:
Fixes https://github.com/n0-computer/iroh/issues/512

This uses the headers from the gateway config. `access-control-allow-headers` and `access-control-expose-headers` are usually the same so we allow omitting the second, less known one. Without it, headers specified in `access-control-allow-headers` are not visible in `fetch()` responses even if they are sent back.


fabricedesre:
cc @Arqu

Arqu (approved):
Thank you!

PR #524, opened 2022-11-18
fix: docker-compose RPC ports should only listen on loopback

### LINKED ISSUE #520: make sure docker compose example doesn't expose writeable APIs
We don't want default docker compose files to expose internal or writable APIs, sets a bad example for others building off this compose file as an example setup. See: https://github.com/n0-computer/iroh/issues/473#issuecomment-1320445696

Instead, example docker compose file should listen on loopback. Example of the technique:
https://github.com/ipfs/kubo/blob/039badf3b24bba4b3a6ca80e2b5163afa4e4b7cb/docker-compose.yaml#L1

b5:
closes #520

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -14,18 +14,29 @@ services:
       - "IROH_GATEWAY__RPC_CLIENT__P2P_ADDR=grpc://iroh-p2p:4401"
       - "IROH_GATEWAY__RPC_CLIENT__STORE_ADDR=grpc://iroh-store:4402"
     ports:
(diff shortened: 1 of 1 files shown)


docker/docker-compose.yaml
```diff
@@ -14,18 +14,29 @@ services:
...
   iroh-p2p:
     container_name: "iroh-p2p"
     image: "n0computer/iroh-p2p:latest"
     environment:
       - "IROH_P2P__RPC_CLIENT__STORE_ADDR=grpc://iroh-store:4402"
     ports:
+      # libp2p connection port. peers will dial your node here
       - "4401:4401"
-      - "4444:4444"
+      # RPC port, iroh CLI uses this to control your p2p service
```
Arqu, line 31: Think it's the inverse here. 4401 is the RPC port. 4444 is the libp2p port.

Arqu (changes requested):
Think we got ports mixed up there. Other than that LGTM

PR #525, opened 2022-11-18
Verify content fetched from racing http gateways

fabricedesre:
While doing some local network fetching, I encountered this error: `curl http://localhost:9050/ipfs/CID` failed to return the expected content but returned that:

```html
<html>
<head><title>504 Gateway Time-out</title></head>
<body>
<center><h1>504 Gateway Time-out</h1></center>
<hr><center>openresty</center>
</body>
</html>
```
This is a combination of 2 errors:
- Failure to verify the CID hash since that's not the expected content!
- Failure to consider non 2xx response from gateways as errors.

This patch fixes both, and we now get the timeout from the local gateway as expected:
```shell
curl http://localhost:9050/ipfs/bafkreibimdxekc6ebparbv3pp2hrcfzib7m42ysvzis3djdlumflidnkky
{"code":504,"message":"request timed out","success":false}
```

### DIFF
--- a/iroh-resolver/src/content_loader.rs
+++ b/iroh-resolver/src/content_loader.rs
@@ -161,11 +161,21 @@ impl FullLoader {
     async fn fetch_gateway(&self, cid: &Cid) -> Result<Option<LoadedCid>> {
         match self.next_gateway().await {
             Some(url) => {
-                let data = reqwest::get(url.as_url(cid)).await?.bytes().await?;
-                Ok(Some(LoadedCid {
-                    data,
-                    source: Source::Http(url.as_string()),
-                }))
+                let response = reqwest::get(url.as_url(cid)).await?;
+                // Filter out non http 200 responses.
+                if !response.status().is_success() {
+                    return Ok(None);
+                }
+                let data = response.bytes().await?;
+                // Make sure the content is not tampered with.
+                if iroh_util::verify_hash(cid, &data) == Some(true) {
+                    Ok(Some(LoadedCid {
+                        data,
+                        source: Source::Http(url.as_string()),
+                    }))
+                } else {
+                    Ok(None)
+                }
             }
             None => Ok(None),
         }
--- a/iroh-resolver/src/content_loader.rs
+++ b/iroh-resolver/src/content_loader.rs
@@ -164,7 +164,7 @@ impl FullLoader {
                 let response = reqwest::get(url.as_url(cid)).await?;
                 // Filter out non http 200 responses.
                 if !response.status().is_success() {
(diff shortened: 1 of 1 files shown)


fabricedesre:
cc @b5 since I think you touched that last.

dignifiedquire:
The content_loader in there should be replaced with a configured version of the `FullLoader` in `iroh-resolver`. It has this fix already and integrates racing gateways and other transports including indexers.

- https://github.com/n0-computer/iroh/blob/main/iroh-resolver/src/content_loader.rs#L45

--- 

At least I thought I did..seems I missed adding this when I worked on that code. This fix really needs to get into the `FullLoader`...

iroh-resolver/src/content_loader.rs
```diff
@@ -161,11 +161,21 @@ impl FullLoader {
...
             Some(url) => {
-                let data = reqwest::get(url.as_url(cid)).await?.bytes().await?;
-                Ok(Some(LoadedCid {
-                    data,
-                    source: Source::Http(url.as_string()),
-                }))
+                let response = reqwest::get(url.as_url(cid)).await?;
+                // Filter out non http 200 responses.
+                if !response.status().is_success() {
+                    return Ok(None);
```
dignifiedquire, line 167: this should return an error

iroh-resolver/src/content_loader.rs
```diff
@@ -161,11 +161,21 @@ impl FullLoader {
...
+                }
+                let data = response.bytes().await?;
+                // Make sure the content is not tampered with.
+                if iroh_util::verify_hash(cid, &data) == Some(true) {
+                    Ok(Some(LoadedCid {
+                        data,
+                        source: Source::Http(url.as_string()),
+                    }))
+                } else {
+                    Ok(None)
```
dignifiedquire, line 177: this should return an error

dignifiedquire:
Sorry for my earlier confused content, this is important and very unfortunate that I missed this. Thank you for catching this.

iroh-resolver/src/content_loader.rs
fabricedesre: ok - but why don't we also return an error when there is no gateway?

dignifiedquire: that’s what the None is for in me head, because we didn’t fail to retrieve, we just didn’t have one configured

dignifiedquire (approved):

iroh-resolver/src/content_loader.rs
fabricedesre: I swiched to errors in the latest commit.

PR #527, opened 2022-11-19
fix: convert v0 -> v1 before base32 conversion

### LINKED ISSUE #511: panic when loading from the gateway on main branch
### Steps to reproduce:

1. spin up the gateway
2. hit `http://localhost:9050/ipfs/QmYwAPJzv5CZsnA625s3Xf2nemtYgPpHdWEz79ojWnPbdG/contact` in a browser
3. gateway process logs will spit out:
```
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidCidV0Base', iroh-resolver/src/content_loader.rs:87:64
```

b5:
fixes #511

turns out you can't go straight from v0 -> base32 v1, which makes sense, but is irritating. Better than the alternative of silently converting into cidv1 I guess.

Anyway, fixes the gateway

### DIFF
--- a/iroh-resolver/src/content_loader.rs
+++ b/iroh-resolver/src/content_loader.rs
@@ -83,22 +83,19 @@ impl GatewayUrl {
         }
     }
 
-    pub fn as_url(&self, cid: &Cid) -> Url {
-        let cid_str = cid
-            .into_v1()
-            .unwrap()
-            .to_string_of_base(Base::Base32Lower)
-            .unwrap();
-        match self {
+    pub fn as_url(&self, cid: &Cid) -> Result<Url> {
+        let cid_str = cid.into_v1()?.to_string_of_base(Base::Base32Lower)?;
+        let url = match self {
             GatewayUrl::Full(raw) => {
                 let mut url = raw.join(&cid_str).unwrap();
                 url.set_query(Some("format=raw"));
                 url
             }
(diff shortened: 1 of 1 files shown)


iroh-resolver/src/content_loader.rs
```diff
@@ -84,7 +84,11 @@ impl GatewayUrl {
     }
 
     pub fn as_url(&self, cid: &Cid) -> Url {
-        let cid_str = cid.to_string_of_base(Base::Base32Lower).unwrap();
+        let cid_str = cid
+            .into_v1()
+            .unwrap()
+            .to_string_of_base(Base::Base32Lower)
+            .unwrap();
```
dignifiedquire, line 91: are you sure this can never error?

iroh-resolver/src/content_loader.rs
```diff
@@ -84,7 +84,11 @@ impl GatewayUrl {
     }
 
     pub fn as_url(&self, cid: &Cid) -> Url {
-        let cid_str = cid.to_string_of_base(Base::Base32Lower).unwrap();
+        let cid_str = cid
+            .into_v1()
+            .unwrap()
```
dignifiedquire, line 89: are you sure this can never error?

b5: Nope...

dignifiedquire: then though shall handle the error :)

dignifiedquire (approved):

PR #528, opened 2022-11-19
Rename the uds-gateway feature to http-uds-gateway

fabricedesre:
I noticed that `iroh-one` could not start without a config file, because the gateway rpc_addr was not set. This was due to some confusion about the role of the `uds-gateway` feature in `iroh-one/src/config.rs`.

I can't blame people for being confused since it's not a very good name. This patch tries to improve the situation by:
- renaming the feature to the more explicit `http-uds-gateway`.
- use the same rpc addr by default in `iroh-one` as in `iroh-gateway`.

With these changes `iroh-one` can start without parameters as expected, and the usage of the configuration file is unchanged.


iroh-one/Cargo.toml
```diff
@@ -40,6 +40,6 @@ http = "0.2"
 default = ["rpc-mem", "rpc-grpc"]
 rpc-grpc = ["iroh-rpc-types/grpc", "iroh-rpc-client/grpc", "iroh-metrics/rpc-grpc"]
 rpc-mem = ["iroh-rpc-types/mem", "iroh-rpc-client/mem"]
-# uds-gateway controls whether http over uds is enabled. This is independent of the
+# http-uds-gateway controls whether http over uds is enabled. This is independent of the
 # rpc-grpc control endpoint.
```
flub, line 44: I'd move this entirely to the readme, it is relevant for users of the crate plus having comments here makes the dependencies harder to sort automatically.

Speaking of which, `http-ud-gateway` sorts differently than `uds-gateway` so should move up (this sorting was added to DEVELOPERS.md not too long ago).

flub (commented):
lgtm.  agree this would make things slightly easier.

Arqu:
Can we rebase, lgmt

fabricedesre:
> Can we rebase, lgmt

done!

PR #529, opened 2022-11-19
[WIP] feat(p2p): dial MDNS peers

dignifiedquire:
- [x] upgrade libp2p to include the latest fixes to tokio-mdns
- [x] dial mdns peers when we discover them


iroh-p2p/src/swarm.rs
```diff
@@ -104,15 +104,19 @@ pub(crate) async fn build_swarm(
...
-        }))
         .build();
 
     Ok(swarm)
 }
+
+struct Tokio;
+impl Executor for Tokio {
+    fn exec(&self, fut: std::pin::Pin<Box<dyn futures::Future<Output = ()> + Send>>) {
+        tokio::task::spawn(fut);
```
Arqu, line 120: Is this so we can futz around with different executors / configs?

Arqu (approved):
👍

iroh-p2p/src/swarm.rs
dignifiedquire: nope, it is because the rust-libp2p api changed

PR #531, opened 2022-11-21
chore(docs): Recommend --all-targets

### LINKED ISSUE #514: chore(readme): update readme to use `--all-targets`
to be fair, we can do this as a followup since that simplifies things in a lot of places like readme's etc as well

_Originally posted by @flub in https://github.com/n0-computer/iroh/pull/497#discussion_r1026243118_

flub:
Document using --all-targets instead of the --tests --bins etc.  This
is more comprehensive and shorter than trying to list all the targets
by hand.

Closes #514

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -89,7 +89,7 @@ Setting up a [git hook][git-hook] to run these commands can save you many headac
 
 ```shell
 #!/bin/sh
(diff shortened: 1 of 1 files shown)


Arqu (changes requested):

Arqu (approved):
👍

PR #532, opened 2022-11-21
chore(ci): Test more features and MSRV

flub:
Adding explicit check and clippy steps for the default features,
because --all-features enables the "testing" feature which in turn
disables a bunch of production code.  Preferably we figure out how to
not make --all-features skip prod code like this, but this is a stop
gap in the mean time.

Adding an explicit MSRV check and removing the rust-toolchain file.
This means local dev will default to using the latest rust compiler
and we'll not be surprised by failing CI that worked locally.  The
explicit MSRV check makes it so that we at least notice when we break
this.


flub:
@Arqu  the msrv job needs a fair bit of copying of protoc etc.  I wonder if we instead should run it as one of the steps in `build_and_test` job instead?

flub:
@Arqu any idea how to fix windows here?  everything else seems to work

Arqu:
> @Arqu the msrv job needs a fair bit of copying of protoc etc. I wonder if we instead should run it as one of the steps in `build_and_test` job instead?

Yes, lets absorb it into the build and test, we do better with sequential builds in a single job instead of lots of smaller independent ones in parallel. 

> @Arqu any idea how to fix windows here? everything else seems to work
AFAIK windows doesn't do too well with env vars. Maybe try absorbing the `MSRV` into the build matrix as a single element and use `${{ matrix.msrv }}` to pipe it in. Should probably work out + leaves us the option to simply extend the matrix later.

Arqu (dismissed):
LGTM

Arqu:
I will need to update GH Actions required checks with the new flag

flub:
> > AFAIK windows doesn't do too well with env vars. Maybe try absorbing the `MSRV` into the build matrix as a single element and use `${{ matrix.msrv }}` to pipe it in. Should probably work out + leaves us the option to simply extend the matrix later.

Ah, that's a neat trick and works!

This PR is now ready.

The only related question I still have is how caching is handled in current CI?

Arqu:
> > > AFAIK windows doesn't do too well with env vars. Maybe try absorbing the `MSRV` into the build matrix as a single element and use `${{ matrix.msrv }}` to pipe it in. Should probably work out + leaves us the option to simply extend the matrix later.
> 
> Ah, that's a neat trick and works!
> 
> This PR is now ready.
> 
> The only related question I still have is how caching is handled in current CI?

No caching, currently timewise more is spent pulling and pushing the cache than it's saving.

Arqu:
If you have better ideas than shoving into the matrix I'd welcome it, otherwise we can merge for now and I'll bump the settings.

Arqu:
actually can you try with `$env:MSRV` as the env var `if: matrix.os == "windows-latest"`

flub:
> I will need to update GH Actions required checks with the new flag. Ideally this would have been an env var or similar so we can simply update without changing the GHA config every time but will do for now

Not sure I follow, the current approach of extending build_and_test does not change the required checks does it?

flub:
> actually can you try with `$env:MSRV` as the env var `if: matrix.os == "windows-latest"`

sure, i can give this a go

Arqu:
It does, because the matrix is part of the check signature. So "new ones" appear and the old ones don't exist any longer

flub:
ah, i see.  the environment var with different syntax for windows does duplicate a lot of stuff though :(

flub:
> ah, i see. the environment var with different syntax for windows does duplicate a lot of stuff though :(

but it works

Arqu:
> > ah, i see. the environment var with different syntax for windows does duplicate a lot of stuff though :(
> 
> but it works

Lets do that then.

.github/workflows/ci.yml
```diff
@@ -16,6 +16,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-latest]
         rust: [stable]
+        rust-msrv: ["1.63"]
```
dignifiedquire, line 19: we should document this in `Cargo.toml`s and in the `README.md`s

flub: Oh, I'd forgotten this was a Cargo.toml field now.

I've not added it to the READMEs yet, mostly because I'm not sure I like having to keep modifying so many files every time we bump it.  Especially if the Cargo.toml already contains it as metadata (which should be able to moved to the workspace Cargo.toml once the MSRV is 1.64, which in turn will probably happen when we add quic-rpc because that uses GATs)

Arqu (approved):
LGTM, thanks for all the back and forth on it.

PR #537, opened 2022-11-21
fix(`iroh-p2p`): implement full local lookup

### LINKED ISSUE #534: running iroh p2p lookup gives `Protocols (0)` every time
currently running `lookup` on main is missing a bunch of data that should be there:

Screen_Shot_2022-11-21_at_14 28 03

This is a regression. PR that fixes this should add a test

ramfox:
We were previously only returning the peer id, external addrs, and listening addrs. We now return the same content as performing a lookup on a remote peer.

closes #534


iroh-p2p/src/behaviour/peer_manager.rs
```diff
@@ -186,8 +192,26 @@ impl NetworkBehaviour for PeerManager {
...
+        // https://github.com/libp2p/rust-libp2p/issues/3124
+        // When that is resolved, we can hopefully remove this responsibility from the `peer_manager`,
+        // where it, frankly, doesn't belong.
+        if self.supported_protocols.is_empty() {
+            self.supported_protocols = params
+                .supported_protocols()
+                .map(|p| String::from_utf8_lossy(&p).to_string())
+                .collect();
+        }
+
```
ramfox, line 214: @dignifiedquire would specifically like your eyes on this.

I gave some rationale as to why this is how I decided to proceed, but I'm aware this is an inelegant solution to an inelegant problem.

dignifiedquire: seems fine to me, we can just update this code once libp2p provides a better method

dignifiedquire (approved):

PR #539, opened 2022-11-22
[WIP] feat: Basic IPNS support

ppodolsky:
I've found planned refactorings in GetRecords https://github.com/libp2p/rust-libp2p/pull/2712 (CC: @dignifiedquire) to make it streaming. In this PR I've supported DhtGet queries to iroh-p2p and gateway. Proposed changes take into account further possible changes in libp2p::GetRecords. 

All implemented APIs return `Stream<PeerRecord>` but streaming is not used yet, I'm just doing usual Quorum query, wait for response and send all received records through stream. It works, but there is also a lot of unwraps() and todo!() that I'll eliminate if general approach would be reviewed and accepted here.


dignifiedquire:
thank you, haven’t reviewed in detail, but high level structure looks good to me 👍

ppodolsky:
Alright, then I'll add error processing, logs and tag you afterwards

RangerMauve:
Have you done any benchmarking / testing to see the speed? :o I'd be very curious to hear how it compares to Kubo's implementation.

ppodolsky:
Did you mean bare performance? As far as I know, the bottleneck is inside networking and mostly in quorumed query.

In Kubo IPNS requests are ofter timeouted (60s) if records from DHT are outdated. In result, we have terrible timings and it renders IPNS almost unusable for several cases.

There are several proposals how to fix it, the most reliable approach will use aggressive caching + heuristics like using record from the source of truth as fast-track + stream of PeerRecords from Kademlia event loop that may be aborted after achieving enough records to make a good decision about IPNS value (here "enough" and "good" needed to be precisely defined later).

Kubo thread with the corresponding issue is [here](https://github.com/ipfs/kubo/issues/3860), it may be useful

.gitignore
```diff
@@ -2,3 +2,4 @@
 /iroh_gateway/test_files/*
 .env
 Cargo.lock
+.idea
```
dignifiedquire, line 5: please remove, you can always use a personal gitignore for this: https://tekin.co.uk/2020/03/maintain-a-global-git-ignore-file

iroh-p2p/src/dht_records.rs
```diff
@@ -0,0 +1,44 @@
...
+}
+
+pub struct DhtRecords {
+    current_queries: AHashMap<QueryId, DhtGetQuery>,
+}
+
+impl DhtRecords {
+    pub fn new() -> DhtRecords {
+        DhtRecords {
+            current_queries: AHashMap::new(),
```
dignifiedquire, line 22: could be just `derive(Default)`

iroh-resolver/src/content_loader.rs
```diff
@@ -39,6 +46,10 @@ impl<T: ContentLoader> ContentLoader for Arc<T> {
     async fn has_cid(&self, cid: &Cid) -> Result<bool> {
         self.as_ref().has_cid(cid).await
     }
+
+    async fn load_records_from_dht(&self, cid: &Cid) -> Result<iroh_rpc_types::p2p::PeerRecord> {
```
dignifiedquire, line 50: Should be singular `record`, given it only returns a single record?

iroh-resolver/src/content_loader.rs
```diff
@@ -133,6 +144,32 @@ impl FullLoader {
...
+                PeerId::from_multihash(*cid.hash())
+                    .unwrap()
+                    .to_bytes()
+                    .into_iter(),
+            )
+            .collect::<Vec<_>>();
+        let key = Key::from(key_value);
+        match self.client.try_p2p() {
+            Ok(p2p) => Ok(p2p
+                .dht_get(&key, Quorum::N(NonZeroUsize::new(16).unwrap()))
```
dignifiedquire, line 161: this should be a constant, with some docs on why we use this number

dignifiedquire:
I believe record verification is still missing: https://github.com/ipfs/specs/blob/main/ipns/IPNS.md#record-verification which we definitely need

iroh-resolver/src/content_loader.rs
ppodolsky: Let's even move this parameter to config?

I saw fights in discussion threads about why 16 have been chosen and multiple people had tried to play with the number. Tbh, I didn't find precise explanation how Kubo authors had derived this number and why is it not overkilling. Afaik even in Sybil attack the worst damage will be providing outdated records merely. Intuitively, 16 is too high but I won't bet anything on this claim.

So, I'm not sure how to explain why we use this number

dignifiedquire: yes, config sounds good 👍

.gitignore
ppodolsky: Done

iroh-p2p/src/dht_records.rs
ppodolsky: Done

iroh-resolver/src/content_loader.rs
ppodolsky: Done

RangerMauve:
Is this record owner optimization being considered as part of this implementation? https://github.com/ipfs/specs/issues/274

Winterhuman:
Yeah, they mention it at https://github.com/n0-computer/iroh/pull/539#issuecomment-1324082959 as "heuristics like using record from the source of truth as fast-track"

ppodolsky:
> Is this record owner optimization being considered as part of this implementation? [ipfs/specs#274](https://github.com/ipfs/specs/issues/274)

Yes. I'm going to finish https://github.com/n0-computer/iroh/pull/546 firstly and then will bring this PR to the end.

lidel:
Hi, just a heads-up that if you are implementing things, and `signatureV2` is present, focus on validating values from `data` CBOR in V2 records, so it is less painful to refactor when we switch to V2-only and stop duplicating values.

Wrote initial cleanup plan in  https://github.com/ipfs/specs/issues/376.

PR #540, opened 2022-11-22
feat(gateway): Add about page

flub:
Add an about page with a bit of package information.

---

I just found this useful locally as I learn how things work.  Maybe
others like it too.  If not, never mind.

### DIFF
--- a/iroh-gateway/Cargo.toml
+++ b/iroh-gateway/Cargo.toml
@@ -5,7 +5,7 @@ edition = "2021"
 license = "Apache-2.0/MIT"
 readme = "README.md"
 description = "IPFS gateway"
-repository = "https://github.com/dignifiedquire/iroh"
+repository = "https://github.com/n0-computer/iroh"
 rust-version = "1.63"
 
 [dependencies]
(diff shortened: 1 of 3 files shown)


Arqu (changes requested):
I would be fine with including this. However a common convetion for this is to use `z` routes. Like `/healthz` we can move this from an `about` page into an `infoz` and we can add more info later to it. 
I assume anybody hosting this as a gateway would either want to turn it off or customize the about page.

b5:
the `z` suffix is a google-ism that means "this route should not be publicly accessible". Let's _not_ use `z` suffixes unless we mean to enforce that rule (and we really can't, so let's just not use 'em). I'd vote for just `/info`.

Arqu:
> the `z` suffix is a google-ism that means "this route should not be publicly accessible". Let's _not_ use `z` suffixes unless we mean to enforce that rule (and we really can't, so let's just not use 'em). I'd vote for just `/info`.

Agree, I'm fine with just info but also it being a non-public route should be fine and is also probably expected for 3rd party users if the host it as a service.

dignifiedquire:
why not `/about` ?

Arqu:
> why not `/about` ? 

I don't think I've seen many/any about pages with this kind of info. 
Also people potentially hosting the gw would definitely prefer a different about.

flub:
I'm happy to settle on `/info`.  I also added git revision, PTAL.

Arqu (approved):
👍

dignifiedquire (approved):

PR #543, opened 2022-11-22
feat: Support `eth` domains

ppodolsky:
This PR does two things:
- Moved DNS resolver to dedicated struct and reused it instead of creation every time
- Added `resolver.cloudflare-eth.com` resolver for `.eth` domains as in Kubo for making possible to resolve Etherium domains. Ip addresses of resolver are hardcoded for avoiding excessive round-trip.

### DIFF
--- a/iroh-resolver/src/dns_resolver.rs
+++ b/iroh-resolver/src/dns_resolver.rs
@@ -1,0 +1,114 @@
+use crate::resolver::Path;
+use anyhow::Result;
+use std::collections::HashMap;
+use std::net::{IpAddr, Ipv4Addr};
+use trust_dns_resolver::config::{NameServerConfigGroup, ResolverConfig, ResolverOpts};
+use trust_dns_resolver::{AsyncResolver, TokioAsyncResolver};
+
+#[derive(Debug)]
+pub struct DnsResolver {
+    default_resolver: TokioAsyncResolver,
+    domain_resolvers: HashMap<String, TokioAsyncResolver>,
+}
+
+impl DnsResolver {
+    pub fn new() -> DnsResolver {
+        let mut domain_resolvers = HashMap::new();
+
+        let eth_resolver_config = ResolverConfig::from_parts(
+            None,
+            vec![],
+            NameServerConfigGroup::from_ips_https(
+                &[
+                    IpAddr::V4(Ipv4Addr::new(104, 18, 165, 219)),
+                    IpAddr::V4(Ipv4Addr::new(104, 18, 166, 219)),
+                ],
+                443,
+                "resolver.cloudflare-eth.com".to_string(),
+                true,
+            ),
+        );
+        let eth_resolver =
+            AsyncResolver::tokio(eth_resolver_config, ResolverOpts::default()).unwrap();
+
+        domain_resolvers.insert("eth".to_string(), eth_resolver);
+        DnsResolver {
+            default_resolver: AsyncResolver::tokio(
+                ResolverConfig::default(),
+                ResolverOpts::default(),
+            )
+            .unwrap(),
+            domain_resolvers,
+        }
+    }
+
+    #[tracing::instrument]
+    pub async fn resolve_dnslink(&self, url: &str) -> Result<Vec<Path>> {
+        let url = format!("_dnslink.{}.", url);
+        let records = self.resolve_txt_record(&url).await?;
+        let records = records
+            .into_iter()
+            .filter(|r| r.starts_with("dnslink="))
+            .map(|r| {
+                let p = r.trim_start_matches("dnslink=").trim();
+                p.parse()
+            })
+            .collect::<Result<_>>()?;
+        Ok(records)
+    }
+
+    pub async fn resolve_txt_record(&self, url: &str) -> Result<Vec<String>> {
+        // Construct a new Resolver with default configuration options
+        let tld = url.split('.').filter(|s| !s.is_empty()).last();
+        let resolver = tld
+            .and_then(|tld| self.domain_resolvers.get(tld))
+            .unwrap_or(&self.default_resolver);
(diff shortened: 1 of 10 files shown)


iroh-resolver/src/dns_resolver.rs
```diff
@@ -0,0 +1,156 @@
+use crate::resolver::Path;
+use anyhow::Result;
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use std::net::{IpAddr, Ipv4Addr};
+use trust_dns_resolver::config::{NameServerConfigGroup, ResolverConfig, ResolverOpts};
+use trust_dns_resolver::{AsyncResolver, TokioAsyncResolver};
+
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+pub struct DnsResolverConfig {
```
dignifiedquire, line 10: This should be just `Config`, reads nicer and is more in line with conventions we use. That way imports look like `dns_resolver::Config`.

iroh-resolver/src/dns_resolver.rs
```diff
@@ -0,0 +1,156 @@
...
+        DnsResolverConfig {
+            tld_resolvers: None,
+        }
+    }
+}
+
+impl Default for DnsResolverConfig {
+    fn default() -> Self {
+        DnsResolverConfig {
+            tld_resolvers: Some(HashMap::from_iter(vec![(
```
dignifiedquire, line 26: Could you add a comment pointing to the docs of the eth resolver from cloudflare please?

iroh-resolver/src/dns_resolver.rs
ppodolsky: Done

iroh-resolver/src/dns_resolver.rs
ppodolsky: Added link to https://eth.link, it is the most relevant documentation I've managed to find

iroh-resolver/src/dns_resolver.rs
```diff
@@ -0,0 +1,157 @@
+use crate::resolver::Path;
+use anyhow::Result;
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use std::net::{IpAddr, Ipv4Addr};
+use trust_dns_resolver::config::{NameServerConfigGroup, ResolverConfig, ResolverOpts};
+use trust_dns_resolver::{AsyncResolver, TokioAsyncResolver};
```
dignifiedquire, line 7: can you split imports into `std`, `other crates` and `crate::` please?

iroh-resolver/src/resolver.rs
```diff
@@ -23,6 +23,8 @@ use tokio::sync::Mutex;
 use tokio::task::JoinHandle;
 use tracing::{debug, trace, warn};
 
+use crate::dns_resolver::{Config, DnsResolver};
```
dignifiedquire, line 26: nit: please group together with the other `crate` imports below

iroh-resolver/src/dns_resolver.rs
```diff
@@ -0,0 +1,157 @@
...
+    pub async fn resolve_txt_record(&self, url: &str) -> Result<Vec<String>> {
+        let tld = url.split('.').filter(|s| !s.is_empty()).last();
+        let resolver = tld
+            .and_then(|tld| {
+                self.tld_resolvers
+                    .as_ref()
+                    .and_then(|tld_resolvers| tld_resolvers.get(tld))
+            })
+            .unwrap_or(&self.default_resolver);
+        let txt_response = resolver.txt_lookup(url).await?;
```
dignifiedquire, line 103: question: Currently if a non default resolver fails, we don't retry with the default resolver. Does that make the most sense, or should we try to use the default resolver if we fail with the specific one?

dignifiedquire (commented):
a couple of nits, but other than that this is looking good

iroh-resolver/src/dns_resolver.rs
ppodolsky: Wow, I relied upon rustfmt. May it be worth to add `rustfmt.toml` with `group_imports = "StdExternalCrate"` in the root? It will require launching `fmt` through `cargo +nightly fmt` but relieve from manual playing with order.

iroh-resolver/src/dns_resolver.rs
dignifiedquire: thanks

iroh-resolver/src/dns_resolver.rs
dignifiedquire: Yeah I am always torn about that. It works pretty well if you set it up manually splitting them by lines (then rustfmt keeps them sorted at least)

dignifiedquire: looks like it is stable now: https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#group_imports

dignifiedquire: made a tracking issue for now https://github.com/n0-computer/iroh/issues/550

ppodolsky: Done

iroh-resolver/src/resolver.rs
ppodolsky: Done

iroh-resolver/src/dns_resolver.rs
ppodolsky: Well, it depends on expected use-cases. I considered TLD-resolvers as ones that could deal with domains that default resolver couldn't deal, otherwise there is no even need to introduce this particular TLD-resolver.

dignifiedquire: makes sense 👍

dignifiedquire (approved):

PR #544, opened 2022-11-23
WIP Try quic rpc

rklaehn:
Very WIP. In fact, so WIP that it might not ever be merged... It might be the best way to proceed is to start from scratch.


rklaehn:
Replaced by https://github.com/n0-computer/iroh/pull/580

Issue #504, opened 2022-11-16
Gateway Subdomain Support


b5:
we currently don't support the [subdomain gateway spec](https://github.com/ipfs/specs/blob/main/http-gateways/SUBDOMAIN_GATEWAY.md)

Per the spec:

> Subdomain Gateway is an extension of [PATH_GATEWAY.md](https://github.com/ipfs/specs/blob/main/http-gateways/PATH_GATEWAY.md) that enables website hosting isolated per CID/name, while remaining compatible with web browsers relative pathing and security model of the web. Below should be read as a delta on top of that spec.

We do plan to support subdomain gateways, but the _earliest_ the core team will _start_ into work on this is Q1 2023. If you need this please react with a  👍 to this issue so we can assess how important this is for others.

Anyone willing to roll up their sleeves & submit a PR, we'll absolutely allocate time for review & PR planning, please respond in this issue before diving in 😄.

PR #546, opened 2022-11-23
feat: Better support of Gateway Subdomain spec

ppodolsky:
Added basic subdomain support by uplifting parameters extraction from `Host` header required for handlers.
DnsLink part of spec is not supported in this PR.

Related to #504 

**Spec:**

- [x] HEAD, GET methods
- [x] DNSLink support
- [x] Redirection from Path methods to Subdomain methods (behaviour is configured by option)
- [x] X-Forwarded-Host support
- [x] X-Forwarded-Proto support

Current PR won't cover [URI router](https://github.com/ipfs/specs/blob/main/http-gateways/SUBDOMAIN_GATEWAY.md#uri-router), I believe it is better to implement in a separate PR.


iroh-gateway/src/handlers.rs
```diff
@@ -49,6 +54,15 @@ use crate::{
     templates::{icon_class_name, ICONS_STYLESHEET, STYLESHEET},
 };
 
+static SUBDOMAINS_REGEX: Lazy<Regex> = Lazy::new(|| {
+    Regex::new(r"(?P<cid>[A-Za-z\d]+)\.(?P<scheme>ip(?:ns|fs))\.(?P<hostname>.*)").unwrap()
```
dignifiedquire, line 58: Is this a custom or based on somewhere?

dignifiedquire: in either case, should have definitely some tests just making sure this captures the right things on different inputs

ppodolsky: Added several tests

ppodolsky:
Can't figure out if we should do anything here with the note from [spec](https://github.com/ipfs/specs/blob/main/http-gateways/SUBDOMAIN_GATEWAY.md):

> Note: Base36 must be used to ensure CIDv1 with ED25519 fits in a single DNS label (63 characters).

Decode and then encode IPNS name? Or what?

iroh-gateway/src/core.rs
```diff
@@ -298,87 +426,167 @@ mod tests {
...
+        );
 
-        assert_eq!(http::StatusCode::OK, res.status());
-        assert!(res.headers().get("content-length").is_some());
-        assert_eq!(res.headers().get("content-length").unwrap(), "5");
+        // IPNS
+        let res = do_request(
+            "GET",
+            &format!("localhost:{}", test_setup.gateway_addr.port()),
+            "/ipns/k51qzi5uqu5dlvj2baxnqndepeb86cbk3ng7n3i46uzyxzyqj2xjonzllnv0v8",
```
lidel, line 521: Would be nice to add separate test for legacy notation "peerid in Base58"  that was not a valid CID (e.g., `12D3KooWJHxkQKX8C5KAyqEPhn2ssT2in4TExyG9SXxi519tycL9`)
See explainer in https://github.com/multiformats/cid-utils-website/issues/48

This is important feature of subdomain gateway: when a request is made to `example.com/ipns/id`, subdomain set up on `example.com` will return HTTP 301 redirect to proper subdomain. 
Before redirect is made, `id` is converted to `cidv1-libp2p-key-base36`.

iroh-gateway/src/core.rs
```diff
@@ -298,87 +426,167 @@ mod tests {
...
+                test_setup.gateway_addr.port()
+            ),
+            res.headers().get("Location").unwrap().to_str().unwrap()
+        );
+
+        // IPNS + DNSLink
+        let res = do_request(
+            "GET",
+            &format!("localhost:{}", test_setup.gateway_addr.port()),
+            "/ipns/ipfs-subdomain.ipfs-domain.eth",
```
lidel, line 557: nit: change to `en.wikipedia-on-ipfs.org`? this is our usual test case for handling `-`  (i suggest avoiding using  .eth in tests unrelated to ENS interop).

iroh-gateway/src/core.rs
```diff
@@ -298,87 +426,167 @@ mod tests {
...
+        assert_eq!(http::StatusCode::MOVED_PERMANENTLY, res.status());
+        assert_eq!(
+            format!(
+                "http://ipfs--subdomain-ipfs--domain-eth.ipns.localhost:{}/",
+                test_setup.gateway_addr.port()
+            ),
+            res.headers().get("Location").unwrap().to_str().unwrap()
+        );
+
+        // X-Forwarded-Proto
```
lidel, line 571: nit: missing the same test for `x-forwarded-host`

iroh-gateway/src/regex.rs
```diff
@@ -0,0 +1,28 @@
+use once_cell::sync::Lazy;
+use regex::Regex;
+
+pub static SUBDOMAIN_REGEX: Lazy<Regex> = Lazy::new(|| {
+    Regex::new(r"^(?P<cid_or_domain>[A-Za-z\d\-.]+)\.(?P<scheme>ip(?:ns|fs))\.(?P<hostname>[^.]+)$")
+        .unwrap()
```
lidel, line 6: Just to make sure there are no regressions in the future, add tests for edge cases where hostname is the same as scheme, e.g.:

`bafy.ipfs.ipfs.com` 
`bafy.ipns.ipfs.com`
`bafy.ipns.ipns.com`
`bafy.ipns.ipfs.com`

iroh-gateway/src/core.rs
ppodolsky: Done

I had to add nasty `from_peer_id` function for dealing with discrepancy between `multihash:0.16` and `multihash:0.17`. This func may be safely deleted after updateing to multihash:0.17

cc @dignifiedquire Would you mind to release multiaddr 0.17 for bringing new multihash crate to rust-libp2p?

iroh-gateway/src/core.rs
ppodolsky: Done

iroh-gateway/src/regex.rs
ppodolsky: Done

iroh-gateway/src/core.rs
ppodolsky: Done.

Btw, while I was implementing code around `-` processing with regular expressions I had faced that it is hard or even impossible to properly process cases like `aaaa-b-aaaa.com` (a single symbol between 2 dashes) because captured groups were overlapping there (`a-b` and `b-a` with `b` belonging to both groups) and it is forbidden in `Regex` crate.

I mean, it may be useful to consider in E2E and other tests this case too cause it belongs to a separate equivalence class.

iroh-gateway/src/core.rs
dignifiedquire: @vmx what't the status here?

vmx: @dignifiedquire I'm not maintaining rust-multiaddr, that's a question for @mxinden :)

Arqu (commented):
In regards to the gateway changes, this looks great. Regex is a bit scary looking.
Only thing I think we should do before the merge is resolve the multihash dependency.

iroh-resolver/src/resolver.rs
```diff
@@ -47,6 +47,25 @@ pub struct Block {
     links: Vec<Cid>,
 }
 
+// ToDo: Replace to PeerId::from_str after updating multihash crate to 0.17
```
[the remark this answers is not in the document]
ppodolsky: I need a moment to fix it, but we should decide who will release multiaddr 0.17 and bring it through rust-libp2p and possibly other deps

ppodolsky:
> In regards to the gateway changes, this looks great. Regex is a bit scary looking. Only thing I think we should do before the merge is resolve the multihash dependency.

No need to scare, I replace regex with function :)

iroh-gateway/src/core.rs
lidel: fwiw Kubo also started with regex, realized it comes with footguns, and replaced with a plain function :)

ppodolsky: And it doesn't bring extra crate. You are right, dunno why I've decided regex would be better. It might be useful for complicated cases or expected future changes in format, but replacing dashes are definitely not this case.

iroh-gateway/src/core.rs
mxinden: Kicked off release process here. Please approve. https://github.com/multiformats/rust-multiaddr/pull/66

ppodolsky:
Almost done, but should not merge until I set proper version of libp2p in Cargo.toml

ppodolsky:
We may create an issue for cleaning things out after libp2p release and merge this PR as is. If it is ok, I'll prepare proper commits tomorrow

dignifiedquire:
> We may create an issue for cleaning things out after libp2p release and merge this PR as is. If it is ok, I'll prepare proper commits tomorrow

Yeah I think this makes the most sense, appreciated

ppodolsky:
> > We may create an issue for cleaning things out after libp2p release and merge this PR as is. If it is ok, I'll prepare proper commits tomorrow
> 
> Yeah I think this makes the most sense, appreciated

Done

Cargo.toml
```diff
@@ -45,3 +45,6 @@ incremental = false
 
 [workspace.metadata.release]
 consolidate-commits = true
+
+[workspace.dependencies]
+libp2p = { version = "0.50" }
```
dignifiedquire, line 50: can you remove this from this PR here please, see discussion on #595

ppodolsky: Done

iroh-gateway/src/client.rs
```diff
@@ -231,13 +231,24 @@ impl<T: ContentLoader> Client<T> {
...
 #[derive(Debug, Clone)]
-pub struct Request {
+pub struct IpfsRequest {
     pub format: ResponseFormat,
     pub cid: CidOrDomain,
     pub resolved_path: iroh_resolver::resolver::Path,
     pub query_file_name: String,
     pub download: bool,
     pub query_params: GetParams,
+    pub is_subdomain_mode: bool,
```
Arqu, line 241: Nitpick, can we make this `subdomain_mode`

iroh-gateway/src/handlers.rs
```diff
@@ -46,27 +53,56 @@ use crate::{
     templates::{icon_class_name, ICONS_STYLESHEET, STYLESHEET},
 };
 
+enum RequestPreprocessingResult {
+    ResponseImmediately(GatewayResponse),
```
Arqu, line 57: Can we rename this to `RespondImmediately` is a little bit more readable.

Arqu (changes requested):
Just some very minor comments.
Need to get CI to pass, otherwise I'm happy to merge.

iroh-gateway/src/handlers.rs
ppodolsky: Done

iroh-gateway/src/client.rs
ppodolsky: Done

iroh-gateway/src/constants.rs
```diff
@@ -1,6 +1,8 @@
 use axum::http::{header::HeaderName, HeaderValue};
 
 // Headers
+pub static HEADER_X_FORWARDED_HOST: HeaderName = HeaderName::from_static("x-forwarded-host");
```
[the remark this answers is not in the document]
ppodolsky: Let's merge this one firstly and then I'll rebase and refine left one

iroh-gateway/src/config.rs
```diff
@@ -48,6 +52,9 @@ pub struct Config {
     /// set of user provided headers to attach to all responses
     #[serde(with = "http_serde::header_map")]
     pub headers: HeaderMap,
+    /// Redirects to subdomains for path requests
+    #[serde(default = "set_true")]
+    pub redirect_to_subdomain: bool,
```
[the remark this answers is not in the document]
ppodolsky: Done

ppodolsky:
Argh, something happened during rebase, let me check what is wrong with the test

ppodolsky:
Fixed tests

Arqu (approved):
LGTM!

Issue #542, opened 2022-11-22
Providers are not found in local network setup


fabricedesre:
Configuration: running `iroh-one` on 2 machines on the same local network, with no p2p bootstrap peers. Discovery happens just fine over mDNS.

Use the branch at https://github.com/capyloon/iroh/commit/a7c3b15f0bdc2e1a15092aab31f3afe9cc1f7b07#diff-4d226c1c8b9372b073d09fa48265e9826c6886a744b6b892eb2c2d800af9d18a to push  some content to machine1with:

```shell
#!/bin/bash

curl 'http://machine1:9050/ipfs/' \
  -v \
  -H 'Accept: */*' \
  -H 'Accept-Language: en-US' \
  -H 'Cache-Control: no-cache' \
  -H 'Connection: keep-alive' \
  -H 'Content-Type: text/plain;charset=UTF-8' \
  -H 'Origin: null' \
  -H 'Pragma: no-cache' \
  --data-raw 'Hello Capyloon!'
```
The CID bafkreia5so3urvs3o5mvguua7othrzg5ssz6djdwqtywrmc6m5l7h7wlnm is getting provided. Check that the CID is in store with `curl http://machine1:9050/ipfs/bafkreia5so3urvs3o5mvguua7othrzg5ssz6djdwqtywrmc6m5l7h7wlnm` 

Then from machine2, wait for both peer counts to reflect the mDNS discovery, and try to fetch the same CID locally: `curl http://localhost:9050/ipfs/bafkreia5so3urvs3o5mvguua7othrzg5ssz6djdwqtywrmc6m5l7h7wlnm`

I see that log in machine2:

```
  2022-11-22T18:56:33.182383Z  INFO iroh_p2p::node: bitswap find providers bafkreia5so3urvs3o5mvguua7othrzg5ssz6djdwqtywrmc6m5l7h7wlnm
    at iroh-p2p/src/node.rs:465
    in iroh_p2p::node::handle_node_event with event: Bitswap(FindProviders { key: Cid(bafkreia5so3urvs3o5mvguua7othrzg5ssz6djdwqtywrmc6m5l7h7wlnm), response: Sender { chan: Tx { inner: Chan { tx: Tx { block_tail: 0x7fd1fc0155a0, tail_position: 0 }, semaphore: Semaphore { semaphore: Semaphore { permits: 10 }, bound: 10 }, rx_waker: AtomicWaker, tx_count: 1, rx_fields: "..." } } }, limit: 10 })
[...]

PR #555, opened 2022-11-24
fix(iroh-one): Wire up mem addresses to each other

flub:
This copies the mem-addresses in the config to the bits of config that
are used by the individual services before starting them up.  This
should help them talk to each other a little better.

---

Tested this locally and seems to work, I think?

### DIFF
--- a/iroh-one/src/main.rs
+++ b/iroh-one/src/main.rs
@@ -48,16 +48,16 @@ async fn main() -> Result<()> {
 
     let (store_rpc, p2p_rpc) = {
         let (store_recv, store_sender) = Addr::new_mem();
-        config.rpc_client.store_addr = Some(store_sender);
-        let store_rpc = iroh_one::mem_store::start(store_recv, config.clone().store).await?;
-
         let (p2p_recv, p2p_sender) = Addr::new_mem();
+        config.rpc_client.store_addr = Some(store_sender);
         config.rpc_client.p2p_addr = Some(p2p_sender);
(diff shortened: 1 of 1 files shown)


dignifiedquire:
> Tested this locally and seems to work, I think?

What did you test? 

This might be one of the reasons for https://github.com/n0-computer/iroh/issues/542 because bitswap/p2p can't provide anything if it is not connected to the store.

flub:
I tried a simple get via the http interface after clearing all local state

dignifiedquire:
did you check if the content got stored? resolving content only requires the gateway to be connected to the store, not p2p

fabricedesre:
> > Tested this locally and seems to work, I think?
> 
> What did you test?
> 
> This might be one of the reasons for #542 because bitswap/p2p can't provide anything if it is not connected to the store.

You're spot on! I tried this branch on top of my http-put one and things work as expected! Woohoo!

dignifiedquire (approved):

flub:
> resolving content only requires the gateway to be connected to the store, not p2p

If the store was deleted first I assumed it needs to come back from the p2p node first to make it into the store?  So having deleted the store I'd be more concerned about whether the p2p node can fetch things and serve it without it making it into the store.  Is that possible or will it always go via the store?

dignifiedquire:
@flub I don’t get it 😅 But merging given @fabricedesre confirmation it fixes a bug for him. 

We should note that we get a test for this..

flub:
> We should note that we get a test for this..

yes, I thought that as well.  I think this should be a lot more doable once we have the e2e story progressed a bit more.

PR #565, opened 2022-11-30
iroh-embed MVP

flub:
This took long enough, but a minimalistic version of an iroh-embed with a working example.

The examples are refactored a bit to become real crates: reason is that it needs to compile without the `testing` feature of `iroh-api`.  The examples feature from cargo enables that unfortunately.  I think for now this is fine, once the dev profile does no longer swap out the real Api struct they can become more normal examples.

I'm particularly unhappy with the current way of configuring this - it's too easy to make mistakes.  There are a bunch of notes in the code still that need to be addressed.

# Todo for this PR

- [x] Default configuration should be isolated from system-installed iroh.
- [x] Configuration without iroh-one's `Config` object.
- [x] Explore splitting monolithic `Iroh` struct into separate `Store` and `P2p` nodes.

# Out of scope for this PR

But needed in followup PRs.

- HTTP & UDS Gateway.
- Allow hooking up prometheus metrics.
- Graceful shutdown instead of aborting tasks.
- An in-memory store.


iroh-embed/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+name = "iroh-embed"
+version = "0.1.1"
+edition = "2021"
+license = "Apache-2.0/MIT"
+description = "Embedable iroh library"
+rust-version = "1.63"
+
+[dependencies]
+anyhow = "1"
+iroh-api = { path = "../iroh-api" }
```
flub, line 11: Other workspace crates add a version to this dependency.  Why?

dignifiedquire: so we can publish to crates.io, otherwise when this is published it can not resolve the path dep

iroh-embed/src/lib.rs
```diff
@@ -0,0 +1,83 @@
+use anyhow::Result;
+use iroh_api::Api;
+use iroh_one::mem_p2p;
+use iroh_one::mem_store;
+use iroh_rpc_types::Addr;
+use tokio::task::JoinHandle;
+
+// TODO: This struct with all fields being pub is not suitable for an API as it is not
```
dignifiedquire, line 8: isn’t there non_exhaustive on structs now?

iroh-embed/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+    /// Returns a reference to the iroh API.
+    ///
+    /// This API gives you some high level functionality using the built-in p2p and store
+    /// nodes as well as a few lower-level and p2p-specific functions.
+    pub fn api(&self) -> &Api {
+        &self.api
+    }
+
+    /// Gracefully stop the iroh system.
+    pub fn stop(&self) {
```
dignifiedquire, line 72: this should be async

iroh-embed/src/lib.rs
flub: yes, though if you use that you can not create the struct outside of the crate that defines it.  So you'd have to switch to a builder of some sorts anyway.

iroh-embed/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+    /// tokio_test::block_on(async {
+    /// use iroh_embed::{Iroh, P2pService, RocksStoreService};
+    /// use testdir::testdir;
+    ///
+    /// let dir = testdir!();
+    /// let store = RocksStoreService::new(dir.join("store")).await.unwrap();
+    /// let p2p = P2pService::new(Default::default(), dir, store.addr()).await.unwrap();
+    /// let _iroh = Iroh::new(store, p2p, None, None).await.unwrap();
+    /// # })
+    /// ```
```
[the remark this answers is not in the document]
dignifiedquire: we should really figure out whyt the issue is there oO

iroh-embed/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+    // TODO: on this level we should use better API than parsing the resolvers and indexers
+    // from strings.
+    // TODO: the store will also want to support an in-memory version.
+    pub async fn new(
+        store: RocksStoreService,
+        p2p: P2pService,
+        http_resolvers: Option<Vec<String>>,
+        indexer_endpoint: Option<String>,
+    ) -> Result<Self> {
```
dignifiedquire, line 66: this is already a lot of options, I suspect we we will want (a) a version of this that just uses defaults and (b) a builder style configuration for this

iroh-embed/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+    /// Creates a new running iroh system.
+    ///
+    /// To create an iroh system first a store and p2p service must be created and must be
+    /// wired up to communicate together.  This means the [`P2pService`] must be created
+    /// with the [`StoreClientAddr`] from the store used.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// tokio_test::block_on(async {
```
matthiasbeyer:
```suggestion
    /// # tokio_test::block_on(async {
```

flub: not entirely sure what is going on, but i suspect this is a review of an earlier commit?  but github doesn't seem to say so?

iroh-embed/src/lib.rs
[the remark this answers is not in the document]
flub: I give up though, i really have no idea what is going on.  The example is fast, but running this under doctests becomes slower by the line you add, even when adding `use` lines only, without any reasonable explanation I can come up with.

iroh-embed/src/lib.rs
```diff
@@ -63,6 +96,12 @@ pub struct IrohBuilder {
...
 
+impl Default for IrohBuilder {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 impl IrohBuilder {
     /// Creates a new [`IrohBuilder`].
     pub fn new() -> Self {
```
matthiasbeyer, line 107: This can be removed in favour of `IrohBuilder::default()`!

flub: Is it good style not to provide new in this case?  I'm used to always providing both constructors, clippy's https://rust-lang.github.io/rust-clippy/master/index.html#new_without_default kind of also suggests it is the usual style.

dignifiedquire: having `new` just call `default` is pretty nice as an api imo

matthiasbeyer: Ah! I always use only either `new` with arguments or `default` if there are none. I messed up the way clippy suggests in my head...

Feel free to ignore my comment! :laughing:

iroh-embed/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+    ///
+    /// IPFS HTTP gateways are used to look up providers of CIDs, in addition to looking
+    /// them up in the Distrubuted Hash Table (DHT).
+    pub fn push_http_gateway(mut self, http_gateway: Url) -> Self {
+        self.http_resolvers.push(http_gateway.to_string());
+        self
+    }
+
+    /// Removes all IPFS HTTP gateways.
+    pub fn clear_http_gateways(mut self) -> Self {
```
dignifiedquire, line 153: This seems odd, why is this needed?

iroh-embed/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+        self.p2p = Some(p2p);
+        self
+    }
+
+    /// Adds IPFS HTTP gateways as resolvers to the iroh system.
+    ///
+    /// IPFS HTTP gateways can be used to resolve IPFS content in addition to retrieving the
+    /// content from the IPFS peer-to-peer system.
+    // TODO: Allow using this without a p2p node?
+    pub fn with_http_resolvers(mut self, http_resolvers: impl Iterator<Item = Url>) -> Self {
```
dignifiedquire, line 138: nit: There is different naming conventions, but I associate `with_` always with an alternative `new` constructor, rather than an additive method.

iroh-api/src/api.rs
```diff
@@ -44,9 +51,13 @@ impl fmt::Debug for OutType {
...
 #[cfg_attr(feature = "testing", allow(dead_code), automock)]
 impl Api {
+    /// Creates a new instance from the iroh configuration.
+    ///
+    /// This loads configuration from an optional configuration file and environment
+    /// variables.
     // The lifetime is needed for mocking.
     #[allow(clippy::needless_lifetimes)]
-    pub async fn new<'a>(
+    pub async fn from_iroh_env<'a>(
```
dignifiedquire, line 60: maybe just from_env, given this is the iroh api afterall

iroh-embed/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "iroh-embed"
+version = "0.1.1"
+edition = "2021"
+license = "Apache-2.0/MIT"
+description = "Embedable iroh library"
+rust-version = "1.63"
```
dignifiedquire, line 7: missing some filelds

iroh-embed/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+    /// nodes as well as a few lower-level and p2p-specific functions.
+    pub fn api(&self) -> &Api {
+        &self.api
+    }
+
+    /// Gracefully stop the iroh system.
+    ///
+    /// TODO: Graceful is a lie right now.
+    /// TODO: Will probably become async.
+    /// TODO: Maybe should consume self.
```
dignifiedquire, line 220: two of these seem to be done at least :)

examples/embed/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+        "/ip4/0.0.0.0/udp/0/quic-v1".parse().unwrap(),
+    ];
+    let p2p = P2pService::new(p2p_config, dir, store.addr()).await?;
+
+    // Note by default this is configured with an indexer, but not with http resolvers.
+    let iroh = IrohBuilder::new().store(store).p2p(p2p).build().await?;
+
+    let quick_start: IpfsPath =
+        "/ipfs/QmQPeNsJPyVWPFDVHb77w8G42Fvo15z4bG2X8D2GhfbSXc/quick-start".parse()?;
+    let mut stream = iroh.api().get(&quick_start)?;
```
dignifiedquire, line 26: would be nice to print out what it is actually doing, eg `Fetching quick start <cid> ... ` or sth like that

iroh-embed/README.md
```diff
@@ -0,0 +1,21 @@
+# iroh embed
+
+This provides an API to embed
```
dignifiedquire, line 3: can you add the various shields that we have in the other readmes please?

iroh-embed/src/lib.rs
```diff
@@ -0,0 +1,259 @@
...
+pub use iroh_p2p::Libp2pConfig;
+pub use iroh_unixfs::indexer::IndexerUrl;
+pub use reqwest::Url;
+
+/// An address allowing internal communication with the iroh store service.
+///
+/// This is only needed to wire up the several services together and should not be used to
+/// communicate directly to the service.
+///
+/// Can be created from [`store::RocksStoreService::addr`].
```
dignifiedquire, line 45: seems you mixed this up, this is the p2p addr and below you refer to p2p on the store addr

iroh-embed/src/lib.rs
```diff
@@ -0,0 +1,259 @@
...
+    pub fn new() -> Self {
+        Self {
+            store: None,
+            p2p: None,
+            http_resolvers: vec![],
+            indexer: Some(IndexerUrl::default()),
+        }
+    }
+
+    /// Adds a store service.
```
dignifiedquire, line 114: Maybe `Sets the store service`, given that there can only be one?

iroh-embed/src/lib.rs
```diff
@@ -0,0 +1,259 @@
...
+    ///
+    /// Every [`Iroh`] system needs a store so this can not be skipped.
+    pub fn store(mut self, store: RocksStoreService) -> Self {
+        self.store = Some(store);
+        self
+    }
+
+    /// Adds a p2p service.
+    ///
```
dignifiedquire, line 123: same as above

iroh-embed/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+    /// data for the CID.
+    ///
+    /// By default this uses the [`iroh_resolver::indexer::CID_CONTACT`] indexer.
+    pub fn with_indexer(mut self, indexer: IndexerUrl) -> Self {
+        self.indexer = Some(indexer);
+        self
+    }
+
+    /// Do not use any indexer.
+    pub fn clear_indexer(mut self) -> Self {
```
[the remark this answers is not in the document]
dignifiedquire: lets try it out

iroh-embed/src/lib.rs
```diff
@@ -0,0 +1,259 @@
...
+        };
+        let api = Api::new(api_config).await?;
+
+        Ok(Iroh { store, p2p, api })
+    }
+}
+
+/// The full iroh system.
+///
+/// This must be constructed using [`IrohBuilder`].  Creating an iroh system will start
```
dignifiedquire, line 197: I would use wording like `This is constructed using..` rather than `must be`

iroh-embed/src/p2p.rs
```diff
@@ -0,0 +1,128 @@
...
+use iroh_rpc_types::store::StoreAddr;
+use iroh_rpc_types::Addr;
+use tokio::task::JoinHandle;
+
+// TODO:
+//
+// - Need to allow configuring in memory keystore
+// - make Lib2p2Config non_exhaustive and provide a builder
+
+/// The iroh peer-to-peer (p2p) service.
```
dignifiedquire, line 18: general thought: I wonder if we should call this `networking` rather than `p2p` (including `iroh-p2p`)

iroh-embed/src/p2p.rs
```diff
@@ -0,0 +1,128 @@
...
+    addr: P2pAddr,
+}
+
+impl P2pService {
+    /// Starts a new iroh peer-to-peer service.
+    ///
+    /// This implicitly starts a task on the tokio runtime to manage the storage node.
+    ///
+    /// The `key_store_path` is the directory where the cryptographic identity of this p2p
+    /// node is stored.  This uses the ssh key files format.
```
dignifiedquire, line 33: maybe add a note that an identity is generated the first time, and reused if one is found?

iroh-embed/src/store.rs
```diff
@@ -0,0 +1,104 @@
...
+        let addr = Addr::new_mem();
+        let config = StoreConfig {
+            path,
+            rpc_client: RpcClientConfig {
+                gateway_addr: None,
+                p2p_addr: None,
+                store_addr: Some(addr.clone()),
+                channels: Some(1),
+            },
+            metrics: Default::default(),
```
dignifiedquire, line 37: can you add some documentation about the state of metrics?

dignifiedquire (commented):
some nits, other than that this is looking great

iroh-embed/README.md
flub: sure, i have a personal dislike for shields though ;)

iroh-embed/src/p2p.rs
flub: Hum, I'm not convinced `networking` is a good name - it is very general.  I don't have anything against the current naming, it helps you understand what it does.

iroh-embed/src/store.rs
flub: currently i'm trying to deny their existence :)  I'm not sure it makes sense to describe in the public docs that they exist but are not yet exposed.

i created https://github.com/n0-computer/iroh/issues/597 to follow up on this though - i agree that we should provide ways to use them.

dignifiedquire: 👍

iroh-embed/src/p2p.rs
dignifiedquire: fair, just a thought

iroh-embed/src/lib.rs
dignifiedquire: @flub I think you missed this one

dignifiedquire (approved):

PR #566, opened 2022-11-30
refactor: remove `ContentLoader` and `unixfs` from `iroh-resolver` package

### LINKED ISSUE #460: refactor: move `unixfs` out of the `iroh-resolver` package
We need to separate out the `unixfs` definition, `unixfs_builder`,  and `hamt` work from the rest of the `iroh-resolver` code.

This is a bit hairy as there is some cross-contamination of dependencies (`iroh-resolver::resolver` depends on some `iroh-resolver::unixfs` bits, and vice versa)

ramfox:
closes #460 

Moves `unxifs` content into it's own crate, removing any references to `Resolver` in favor of using a `ContentLoader`.

`ContentLoader` has been moved into it's own crate as well, called `iroh-content`

Basic types that are used by both `unixfs` & `resolver` (as well as a few other crates), have also been moved into `iroh-content`.

The roundtrip tests that were formerly in `unixfs` have been moved to the `iroh-resolver` packages.

Because this involved adjusting code imports, this looks like it touches ALOT. However, the main bulk of the changes are in slimming down `iroh-resolver` & splitting those files into `iroh-unixfs` and `iroh-content`.

EDIT:

moved `iroh-content`/`iroh-content-loader` into `iroh-unixfs`


iroh-unixfs/src/hamt.rs
```diff
@@ -74,54 +71,48 @@ impl Hamt {
         padding.len()
     }
 
-    pub fn children<'a, 'b: 'a, C: ContentLoader>(
+    #[allow(clippy::needless_lifetimes)]
```
ramfox, line 74: @dignifiedquire @rklaehn help! can I chat about this with someone more knowledgable than me tomorrow?
When I remove the lifetimes it tells me I need them, when I add them it tells me its needless.

iroh-content/src/content.rs
```diff
@@ -0,0 +1,286 @@
...
+        anyhow::ensure!(expected_links == actual_links, "links do not match");
+        Ok(())
+    }
+
+    pub fn into_parts(self) -> (Cid, Bytes, Vec<Cid>) {
+        (self.cid, self.data, self.links)
+    }
+}
+
+/// Represents an ipfs path.
```
dignifiedquire, line 92: should be in iroh-resolver

iroh-unixfs/src/hamt.rs
```diff
@@ -74,54 +71,48 @@ impl Hamt {
...
-                Ok(InnerNode::Leaf {
-                    link: link.clone(),
-                    value: node,
-                })
-            }
-            _ => bail!("unexpected node: {:?}", out.content.typ()),
+        let cid = link.cid;
+        let loaded_cid = loader.load_cid(&cid, &ctx).await?;
+        let node = UnixfsNode::decode(&cid, loaded_cid.data)?;
+        //
```
dignifiedquire, line 93: need to handle Raw

ramfox: double checked the code path, `UnixfsNode::decode` handles "decoding" from `Raw`: 
https://github.com/n0-computer/iroh/blob/5702c3e4666557e70993ebcf90b52cf1e04f0c0b/iroh-resolver/src/unixfs.rs#L187-L189

let me know if you meant something different

iroh-unixfs/src/hamt.rs
ramfox: for context on our convo: clippy actually gave good instructions on how to handle this, I just didn't understand what it was trying to say. 

conclusion: adding `'_` to the return value: `impl Stream<...> + '_` gave a clue to the compiler that it needed to do some lifetime eliding.

ramfox:
@dignifiedquire I think the most "controversial" thing here may be the naming & separation of the crate `iroh-content`

It contains two main things:
1) basic types (that are related to loading content) that are used by multiple other crates: Block, Codec, ResponseClip
2) things _directly_ related to loading content: ContentLoader, LoadedContext, LoadedCid, Indexer

I could see a different potential configuration, where we group the first bits into an `iroh-types` crate and the second bits into `iroh-content-loader`. I compromised here with `iroh-content`... which technically encompasses both.

Cargo.toml
```diff
@@ -14,8 +14,10 @@ members = [
   "iroh-rpc-client",
   "iroh-rpc-types",
   "iroh-share",
+  "iroh-unixfs",
   "iroh-store",
   "iroh-util",
+  "iroh-content",
```
dignifiedquire, line 20: are we happy with this name?😅

ramfox: my thoughts here: https://github.com/n0-computer/iroh/pull/566#issuecomment-1339796959

examples/src/bin/importer.rs
```diff
@@ -55,7 +55,7 @@ async fn main() -> Result<()> {
                 if iroh_util::verify_hash(&cid, &data) == Some(false) {
                     bail!("invalid hash {:?}", cid);
                 }
-                let links = iroh_resolver::parse_links(&cid, &data).unwrap_or_default();
+                let links = iroh_content::util::parse_links(&cid, &data).unwrap_or_default();
```
dignifiedquire, line 58: just reexport `parse_links` at the top level, no need for the `util` mod to be exposed

iroh-content/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[package]
+name = "iroh-content"
+version = "0.1.3"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
dignifiedquire, line 6: remove please

iroh-resolver/Cargo.toml
```diff
@@ -56,14 +58,7 @@ tempfile = "3.3.0"
 tokio = { version = "1", features = ["rt", "macros", "rt-multi-thread", "fs"] }
 ruzstd = "0.3"
 
-[build-dependencies]
-prost-build = "0.11.1"
```
dignifiedquire, line 60: I suspect there are more dependencies you can remove, e.g. `prost`

iroh-resolver/src/resolver.rs
```diff
@@ -351,14 +271,17 @@ impl Out {
...
+            // OutContent::Unixfs(node) => node.links().map(|l| l.map(|l| (l.name, l.cid))).collect(),
+            OutContent::Unixfs(_) => todo!(),
             _ => {
                 let links = self.content.links();
                 links.map(|l| l.into_iter().map(|l| (None, l)).collect())
             }
         }
     }
 
+    // TODO(ramfox): figure out circular dependencies
```
dignifiedquire, line 284: what circular dep is still here?

iroh-unixfs/Cargo.toml
```diff
@@ -0,0 +1,65 @@
...
+edition = "2021"
+authors = ["dignifiedquire <me@dignifiedquire.com>"]
+license = "Apache-2.0/MIT"
+repository = "https://github.com/n0-computer/iroh"
+description = "Implementation of unixfs for iroh"
+rust-version = "1.63"
+
+exclude = [
+  "tests/**",
+  "fixtures/**",
```
dignifiedquire, line 13: not needed, as the `fixtures` are still in the resolver crate, right?

iroh-unixfs/Cargo.toml
```diff
@@ -0,0 +1,65 @@
...
+criterion = { version = "0.4.0", features = ["async_tokio"] }
+iroh-car = { version = "0.1.3", path = "../iroh-car" }
+iroh-rpc-types = { version = "0.1.3", path = "../iroh-rpc-types", default-features = false }
+iroh-store = { version = "0.1.3", path = "../iroh-store", default-features = false }
+proptest = "1.0.0"
+rand = "0.8.5"
+rand_chacha = "0.3.1"
+tempfile = "3.3.0"
+tokio = { version = "1", features = ["rt", "macros", "rt-multi-thread", "fs"] }
+ruzstd = "0.3"
```
dignifiedquire, line 58: I don't think these dependencies are all needed, eg ruzstd is only used in iroh-resolver I think

iroh-content/src/lib.rs
```diff
@@ -0,0 +1,11 @@
+pub mod codecs;
+pub mod content_loader;
+pub mod indexer;
+pub mod types;
+pub mod util;
```
dignifiedquire, line 5: no need to make all these `pub` if you have top level reexports

iroh-resolver/src/resolver.rs
ramfox: none, left over note to myself

thanks for catching

iroh-unixfs/Cargo.toml
ramfox: ahh yes, thank you, I meant to do a cargo clean up!

iroh-unixfs/Cargo.toml
ramfox: hmmm, neither of these actually exists in this crate.

We do have a `benches` folder, should we be excluding that? Not sure what is "typically" published.

iroh-content/src/lib.rs
ramfox: good point. did some re-arranging.

iroh-unixfs/Cargo.toml
dignifiedquire: no not needed to be excluded, we only have this because there are some larger fixtures

Cargo.toml
ramfox: @dignifiedquire what do you think about `iroh-content-loader` instead of `iroh-content`. The bulk of the code deals with how to load and interpret content.

iroh-resolver/src/resolver.rs
```diff
@@ -351,7 +271,9 @@ impl Out {
     /// is unixfs
     pub fn named_links(&self) -> Result<Vec<(Option<&str>, Cid)>> {
         match &self.content {
-            OutContent::Unixfs(node) => node.links().map(|l| l.map(|l| (l.name, l.cid))).collect(),
+            // TODO(ramfox): add back in when we figure out circular dependencies
+            // OutContent::Unixfs(node) => node.links().map(|l| l.map(|l| (l.name, l.cid))).collect(),
+            OutContent::Unixfs(_) => todo!(),
```
dignifiedquire, line 276: what's the issue here?

iroh-unixfs/src/hamt.rs
```diff
@@ -74,55 +71,47 @@ impl Hamt {
...
+        // TODO(ramfox): should this then attempt to resolve as ipld?
```
dignifiedquire, line 93: what's the question here?

Cargo.toml
ramfox: moving `iroh-content-loader` contents into `iroh-unixfs`

PR #575, opened 2022-12-03
(chore) Remove dead code

fabricedesre:
That should never have been in...


dignifiedquire (approved):

dignifiedquire:
looks like there is some warnings still

fabricedesre:
> looks like there is some warnings still

fixed in latest commit

dignifiedquire (approved):

PR #576, opened 2022-12-06
ref: Enable backtrace support in anyhow

flub:
This enables backtrace support in anyhow, allowing backtraces to be
enabled when setting RUST_BACKTRACE=1.  This is useful when trying to
find the source of an error.


flub:
As you can see this messes up trycmd's outputs.  :(

dignifiedquire (approved):
lets add it, and if it becomes problematic we can always revert it

PR #578, opened 2022-12-07
ref: Require Debug implementation on structs

flub:
I heard you all like Debug.

The crux of this is adding `-Wmissing_debug_implementations` to the cargo config.  The rest is just filling in missing Debug impls.

This instructs the compiler to emit warnings for structs which do not
have a Debug implementation.


flub:
According to https://stackoverflow.com/questions/67568003/how-can-i-have-a-shared-clippy-configuration-for-all-the-crates-in-a-workspace the alternative to attributes is to add it to the `.cargo/config.toml` file.

dignifiedquire:
> According to https://stackoverflow.com/questions/67568003/how-can-i-have-a-shared-clippy-configuration-for-all-the-crates-in-a-workspace the alternative to attributes is to add it to the `.cargo/config.toml` file.

Lets try this, otherwise the next crate will be missing it 😅

flub:
> > According to https://stackoverflow.com/questions/67568003/how-can-i-have-a-shared-clippy-configuration-for-all-the-crates-in-a-workspace the alternative to attributes is to add it to the `.cargo/config.toml` file.
> 
> Lets try this, otherwise the next crate will be missing it sweat_smile

seems like that was a good call, works pretty well.

dignifiedquire (approved):

PR #580, opened 2022-12-08
Switch to quic-rpc

rklaehn:
This switches to quic-rpc with the hyper transport. Performance will be a bit better than before, but not astronomically so.

We should be able to get rid of the load balancers. But they are still in there. We can remove them in a subsequent PR.

Issues we should do immediately after merging this:
https://github.com/n0-computer/iroh/issues/581
https://github.com/n0-computer/iroh/issues/582
https://github.com/n0-computer/iroh/issues/583
https://github.com/n0-computer/iroh/issues/584

Another comment: a lot of the boilerplate in the types can be eliminated by the macros that are already in quic-rpc. I decided to go with the no macro approach for now because a) it is nice to see how it works under the hood b) I don't mind the boilerplate that much, compared to a somewhat intransparent macro.

I think in the long term we might want to use a macro do define the request/response enums and the traits for each case, but I think having the big ugly switch statement without a macro is good so we can customize the treatment for each request.


.github/workflows/ci.yml
```diff
@@ -38,7 +38,7 @@ jobs:
       # Force not building debuginfo to save space on disk.
       RUSTFLAGS: "-C debuginfo=0"
       RUSTV: ${{ matrix.rust }}
-      MSRV: "1.63"
+      MSRV: "1.65"
```
dignifiedquire, line 41: what's the reason for this?

.github/workflows/ci.yml
```diff
@@ -124,12 +124,12 @@ jobs:
     - name: tests
       timeout-minutes: 30
       run: |
-        cargo test --workspace --all-features --all-targets
+        cargo test --workspace --all-features --lib --bins --tests --examples
```
dignifiedquire, line 127: if we do this change, we still need `cargo check` to make sure things compile on all targets

iroh-qrpc-types/src/p2p.rs
```diff
@@ -0,0 +1,482 @@
...
+use crate::RpcResult;
+
+pub type P2pAddr = super::addr::Addr<P2pService>;
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Key(pub Bytes);
+
+// rpc Version(google.protobuf.Empty) returns (VersionResponse) {}
+#[derive(Serialize, Deserialize, Debug)]
+pub struct VersionRequest;
```
dignifiedquire, line 21: any chance we can avoid having types for empty requests?

iroh-qrpc-types/src/p2p.rs
```diff
@@ -0,0 +1,482 @@
...
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct VersionResponse {
+    pub version: String,
```
dignifiedquire, line 25: Could this be done as `type VersionResponse = String` instead, or do we have to use new types always?

iroh-qrpc-types/src/p2p.rs
```diff
@@ -0,0 +1,482 @@
...
+    type Response = RpcResult<LocalPeerIdResponse>;
+}
+
+impl RpcMsg<P2pService> for ExternalAddrsRequest {
+    type Response = RpcResult<ExternalAddrsResponse>;
+}
+
+impl RpcMsg<P2pService> for ListenersRequest {
+    type Response = RpcResult<ListenersResponse>;
+}
```
dignifiedquire, line 482: is the above covered by the macros? this is pretty painful to look at

iroh-resolver/src/unixfs_builder.rs
```diff
@@ -738,7 +738,7 @@ fn add_blocks_to_store_chunked<S: Store>(
 ) -> impl Stream<Item = Result<AddEvent>> {
     let mut chunk = Vec::new();
     let mut chunk_size = 0u64;
-    const MAX_CHUNK_SIZE: u64 = 1024 * 1024 * 16;
+    const MAX_CHUNK_SIZE: u64 = 1024 * 1024;
```
dignifiedquire, line 741: why?

iroh-store/src/rpc.rs
```diff
@@ -18,107 +16,131 @@ impl iroh_rpc_types::NamedService for Store {
...
 
-#[async_trait]
-impl RpcStore for Store {
+#[derive(Clone)]
+pub struct RpcStore(Store);
+
+impl RpcStore {
     #[tracing::instrument(skip(self))]
-    async fn version(&self, _: ()) -> Result<VersionResponse> {
+    async fn version(self, _: VersionRequest) -> VersionResponse {
```
dignifiedquire, line 24: just remove the argument?

iroh-store/src/rpc.rs
```diff
@@ -18,107 +16,131 @@ impl iroh_rpc_types::NamedService for Store {
     const NAME: &'static str = "store";
 }
 
-#[async_trait]
-impl RpcStore for Store {
+#[derive(Clone)]
+pub struct RpcStore(Store);
```
dignifiedquire, line 20: why do we need a new type and don't just implement this directly on the Store?

iroh-store/src/rpc.rs
```diff
@@ -18,107 +16,131 @@ impl iroh_rpc_types::NamedService for Store {
...
+async fn handle_session(server: StoreServer, store: Store) -> Result<()> {
+    use StoreRequest::*;
+    let s = server.clone();
+    let store = RpcStore(store);
+    loop {
+        let store = store.clone();
+        let s = s.clone();
+        let (req, chan) = s.accept_one().await?;
+        let store = store.clone();
+        #[rustfmt::skip]
```
dignifiedquire, line 116: why?

.github/workflows/ci.yml
rklaehn: I think this is already done. I only replaced --all-targets with the above for the tests.

```
    - name: check default features
      run: |
        cargo check --workspace --all-targets

    - name: check all features
      run: |
        cargo check --workspace --all-features --all-targets
```

iroh-qrpc-types/src/p2p.rs
rklaehn: Not easily. The purpose of the type is to distinguish between different requests. So you want two different enum cases.

Now, in principle you could write an enum that just has the request case without a value. But the current scheme works well with derive_more to automatically build you the two conversions that you need (From from the specific request to the request enum).

Also, I don't mind it that much.

iroh-qrpc-types/src/p2p.rs
rklaehn: For the responses - unlike for the requests - you don't have to have a newtype.

The response enum would just need to have a case for String, with a From conversion from String to the response enum.

I use this to avoid having to make a newtype struct for each unit response. I guess it is a matter of taste. What I did is to only use naked results for the unit response.

iroh-qrpc-types/src/p2p.rs
rklaehn: Yes, the macro written by @Frando allows you to declare these.

But I don't mind it that much, compared to a giant macro machine that is kinda write only.

It really sucks if you go into a foreign code base, use go to definition, and then just get pointed into a single line macro invocation all the time. (That was basically me when I tried to add client streams to the existing RPC stuff before ipfs camp)

iroh-resolver/src/unixfs_builder.rs
rklaehn: The http2 transport is reusing using the http2 framing. That way you don't have to add a framing layer on top, and they are specifically made to allow user level framing.

But - the default max frame size is pretty small. It is no problem to increase it, but that means a larger per connection buffer. Not the end of the world in the RPC case, but something you definitely want to avoid in a P2p case where you have lots of connections.

So, the current default is 1024 * 1024 plus margin. I just adjusted it here because once this is merged I want to get rid of the PutMany thing anyway, and replace it with a call where the blocks are streamed from the client, as I wanted to do in the first place.

We will have to think about how big this should be. I think bitswap allows a maximum block size of 4MB, but unixfs won't produce such large blocks.

iroh-store/src/rpc.rs
rklaehn: Because I don't want the long lines to be broken up.

Note that the giant switch statement is also something that the macro can do for you, if you don't like it.

But I think that having it in the open is not that bad. E.g. you can add custom timeouts to individual RPC calls by just wrapping the call, whereas with a macro you would need to come up with some macro arguments etc.

iroh-store/src/rpc.rs
rklaehn: This is the line that would not compile if you were to change this:
```rust
...
Version(req) => s.rpc(req, chan, store, RpcStore::version).await,
...
```

iroh-store/src/rpc.rs
rklaehn: Also, as you have probably seen from the code so far, I don't mind newtypes that much... :-)

iroh-resolver/src/unixfs_builder.rs
rklaehn: And yes, I guess quic-rpc needs to make this configurable. But I think we can leave that for later.

iroh-qrpc-types/README.md
```diff
@@ -0,0 +1,24 @@
+# iroh-rpc-types
+
+This crate defines types for use by the
+[iroh-rpc-client](https://github.com/n0-computer/iroh/tree/main/iroh-rpc-client),
+which is used for [iroh](https://github.com/n0-computer/iroh) services to
+communicate internally via RPC. The protocol used is [gRPC](https://grpc.io/).
+This crate defines the gRPC types in the form of [Protocol
+Buffers](https://developers.google.com/protocol-buffers). It uses the [Tonic
+framework](https://github.com/hyperium/tonic) to expose the gRPC types to Rust.
```
flub, line 9: blurb probably needs updating

iroh-qrpc-types/src/addr.rs
```diff
@@ -0,0 +1,117 @@
...
+/// An address. This can be either a memory address, already containing the channel, or a network
+/// address which will have to be opened.
+#[derive(SerializeDisplay, DeserializeFromStr)]
+pub enum Addr<S: Service> {
+    Http2(SocketAddr),
+    Http2Lookup(String),
+    Mem(
+        mem::ServerChannel<S::Req, S::Res>,
+        mem::ClientChannel<S::Res, S::Req>,
+    ),
```
flub, line 19: note entirely sure how much i like that these are always together.  it means you can't guarantee that someone will only use the client-side i think?

iroh-qrpc-types/src/p2p.rs
flub: It's pretty boring to write, but boring is good.  This way makes it easy to jump around to the right places.

iroh-gateway/src/bad_bits.rs
```diff
@@ -188,7 +188,7 @@ mod tests {
         );
         config.set_default_headers();
 
-        let rpc_addr = "grpc://0.0.0.0:0".parse().unwrap();
+        let rpc_addr = "http://0.0.0.0:0".parse().unwrap();
```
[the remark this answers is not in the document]
rklaehn: Yes, I had qrpc before. Changed it back to http because qrpc looked too much like grpc.

Any suggestions?

iroh-gateway/src/rpc.rs
```diff
@@ -1,24 +1,43 @@
...
 
-pub async fn new(addr: GatewayServerAddr, gateway: Gateway) -> Result<()> {
-    iroh_rpc_types::gateway::serve(addr, gateway).await
+pub async fn new(addr: GatewayAddr, gw: Gateway) -> Result<()> {
+    let mut stream = create_server_stream::<GatewayService>(addr).await?;
+    while let Some(server) = stream.next().await {
+        let s = server?;
+        if let Ok((req, chan)) = s.accept_one().await {
+            tracing::info!("accepted connection");
+            let s = s.clone();
```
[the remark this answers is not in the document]
rklaehn: Yes, sorry. This is part of the code that is going to change a bit.

iroh-qrpc-client/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "iroh-qrpc-client"
+version = "0.2.0"
+edition = "2021"
+authors = ["Friedel Ziegelmayer <me@dignifiedquire.com>", "ramfox"]
+license = "Apache-2.0/MIT"
+repository = "https://github.com/n0-computer/iroh"
+description = "RPC type client for iroh"
+rust-version = "1.63"
```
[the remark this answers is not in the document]
rklaehn: Updated it.

iroh-qrpc-client/src/network.rs
```diff
@@ -0,0 +1,300 @@
...
+
+    #[tracing::instrument(skip(self))]
+    pub async fn fetch_providers_dht(
+        &self,
+        key: &Cid,
+    ) -> Result<impl Stream<Item = Result<HashSet<PeerId>>>> {
+        let key = Key(key.hash().to_bytes().into());
+        let res = self
+            .client
+            .server_streaming(FetchProvidersDhtRequest { key })
```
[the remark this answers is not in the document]
rklaehn: Well, it would not compile. Which is what you would want. There is exactly one kind of interaction allowed for one request type and service.

iroh-qrpc-types/README.md
rklaehn: Right. I updated the one for client, but not this one.

iroh-qrpc-client/src/network.rs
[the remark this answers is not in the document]
flub: ah great.  sorry i missed that over here

iroh-qrpc-types/src/addr.rs
rklaehn: The way it was used before, there was a way to flip from client to server and back, which just did not work for mem channels. At least this way it is consistent.

As long as somebody has access to the crate you can not stop him from creating a server or client channel given an address anyway...

iroh-gateway/src/rpc.rs
```diff
@@ -1,24 +1,43 @@
...
+            let gw = gw.clone();
+            tokio::spawn(async move {
+                use GatewayRequest::*;
+                let target = gw.clone();
```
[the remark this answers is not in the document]
rklaehn: Some clones are not as needless as they look. But in any case this has changed now.

iroh-qrpc-types/src/p2p.rs
rklaehn: This is the part which the macro also does fine, if you want things a bit more compact. Where I am really not so sure is the dispatcher. There I really think the verbosity is preferable to hiding details.

iroh-gateway/src/bad_bits.rs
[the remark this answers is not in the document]
rklaehn: Yes, I also came up with `irpc` independenty. It can be `internal` or `iroh` and also sounds somewhat like IPC.

iroh-resolver/src/unixfs_builder.rs
rklaehn: Made an issue: https://github.com/n0-computer/quic-rpc/issues/11

iroh-metrics/Cargo.toml
```diff
@@ -53,8 +52,7 @@ features = [
 optional = true
 
 [features]
-default = ["rpc-grpc"]
-rpc-grpc = ["tonic"]
+default = []
```
flub, line 55: What's the reason to leave these as empty btw?  Does it do something special?

iroh-qrpc-client/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "iroh-qrpc-client"
+version = "0.2.0"
+edition = "2021"
+authors = ["Friedel Ziegelmayer <me@dignifiedquire.com>", "ramfox"]
```
flub, line 5: we probably should have some kind of `team@` or similar?  At sentry we used `oss@`.

iroh-qrpc-client/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+license = "Apache-2.0/MIT"
+repository = "https://github.com/n0-computer/iroh"
+description = "RPC type client for iroh"
+rust-version = "1.65"
+
+[dependencies]
+
+anyhow = "1"
```
flub, line 13:
```suggestion
anyhow = { version = "1", features = ["backtrace"] }
```

iroh-qrpc-types/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+version = "0.2.0"
+edition = "2021"
+authors = ["Friedel Ziegelmayer <me@dignifiedquire.com>", "ramfox"]
+license = "Apache-2.0/MIT"
+repository = "https://github.com/n0-computer/iroh"
+description = "RPC type definitions for iroh"
+rust-version = "1.65"
+
+[dependencies]
+anyhow = "1"
```
flub, line 12:
```suggestion
anyhow = { version = "1", features = ["backtrace"] }
```

iroh-rpc-client/Cargo.toml
```diff
@@ -2,41 +2,30 @@
...
-async-trait = "0.1.56"
 bytes = "1.1.0"
-cid = "0.9"
+cid = "0.9.0"
 config = "0.13.1"
 futures = "0.3.21"
-iroh-metrics = { version = "0.1.3", path = "../iroh-metrics", default-features = false }
-iroh-rpc-types = { version = "0.1.3", path = "../iroh-rpc-types", default-features = false }
-iroh-util = { version = "0.1.3", path = "../iroh-util" }
+iroh-metrics = { path = "../iroh-metrics", default-features = false }
```
flub, line 19: Will need `version` to be able to be published I think?
(also other entries)

iroh-rpc-client/Cargo.toml
```diff
@@ -2,41 +2,30 @@
...
 serde = { version = "1.0", features = ["derive"] }
 tokio = { version = "1", features = ["sync"] }
 toml = "0.5.9"
-tonic = { version = "0.8", optional = true }
-tonic-health = { version = "0.7", optional = true }
-tower = { version = "0.4.13", optional = true }
 tracing = "0.1.34"
-
+iroh-rpc-types = { path = "../iroh-rpc-types" }
+hyper = "0.14.23"
```
flub, line 28: nit: keep this alphabetical

iroh-rpc-client/src/config.rs
```diff
@@ -40,12 +40,12 @@ impl Source for Config {
...
 
 impl Config {
-    pub fn default_grpc() -> Self {
+    pub fn default_network() -> Self {
         Self {
-            gateway_addr: Some("grpc://0.0.0.0:4400".parse().unwrap()),
-            p2p_addr: Some("grpc://0.0.0.0:4401".parse().unwrap()),
-            store_addr: Some("grpc://0.0.0.0:4402".parse().unwrap()),
-            channels: Some(16),
+            gateway_addr: Some("irpc://127.0.0.1:4400".parse().unwrap()),
```
flub, line 45: good choice :)

iroh-rpc-client/src/gateway.rs
```diff
@@ -1,25 +1,53 @@
...
+        StatusRow {
+            name: "gateway",
+            number: 1,
+            status,
+        }
+    }
+
+    #[tracing::instrument(skip(self))]
+    pub async fn watch(&self) -> impl Stream<Item = StatusRow> {
+        futures::stream::pending()
```
flub, line 51: Is this supposed to be a TODO?

iroh-rpc-client/src/network.rs
```diff
@@ -1,57 +1,51 @@
...
+
     #[tracing::instrument(skip(self))]
     pub async fn version(&self) -> Result<String> {
-        let res = self.backend.version(()).await?;
+        let res = self.client.rpc(p2p::VersionRequest).await?;
         Ok(res.version)
     }
 
-    #[tracing::instrument(skip(self))]
+    // #[tracing::instrument(skip(self))]
```
flub, line 33: delete or keep?

iroh-gateway/src/config.rs
```diff
@@ -70,21 +70,8 @@ impl Config {
...
-                match addr {
-                    #[cfg(feature = "rpc-grpc")]
-                    Addr::GrpcHttp2(addr) => Ok(Addr::GrpcHttp2(*addr)),
-                    #[cfg(feature = "rpc-mem")]
-                    Addr::Mem(_) => bail!("can not derive rpc_addr for mem addr"),
-                    _ => bail!("invalid rpc_addr"),
-                }
-            })
-            .transpose()
+    pub fn server_rpc_addr(&self) -> Result<Option<GatewayAddr>> {
```
dignifiedquire, line 73: do we still need this? I added this just to work around the Mem issues before

iroh-rpc-client/src/status.rs
```diff
@@ -1,81 +1,10 @@
...
-use futures::Stream;
-use tonic::transport::channel::Channel;
-use tonic_health::proto::{
-    health_check_response::ServingStatus, health_client::HealthClient, HealthCheckRequest,
-    HealthCheckResponse,
-};
 
 // TODO: make configurable
-const WAIT: std::time::Duration = std::time::Duration::from_millis(1000);
+const _WAIT: std::time::Duration = std::time::Duration::from_millis(1000);
```
flub, line 5: If this isn't used anymore maybe delete it?

iroh-p2p/src/config.rs
```diff
@@ -212,21 +209,8 @@ impl Config {
...
-                match addr {
-                    #[cfg(feature = "rpc-grpc")]
-                    Addr::GrpcHttp2(addr) => Ok(Addr::GrpcHttp2(*addr)),
-                    #[cfg(feature = "rpc-mem")]
-                    Addr::Mem(_) => bail!("can not derive rpc_addr for mem addr"),
-                    _ => bail!("invalid rpc_addr"),
-                }
-            })
-            .transpose()
+    pub fn server_rpc_addr(&self) -> Result<Option<P2pAddr>> {
```
dignifiedquire, line 212: same as above

iroh-rpc-types/src/addr.rs
```diff
@@ -1,101 +1,96 @@
...
-                        }
-                        // attempt to resolve the address, if it can be resolved,
-                        // it's considered a lookup address
-                        if let Ok(mut addr_iter) = part.to_socket_addrs() {
-                            if addr_iter.next().is_some() {
-                                return Ok(Addr::GrpcHttp2Lookup(String::from(part)));
-                            }
-                        }
-                    }
+        let mut parts = s.split("://");
```
flub, line 85: `s.split_once("://")`?

flub: or `splitn` if you'd like to keep the changes smaller

iroh-rpc-types/src/p2p.rs
```diff
@@ -1,34 +1,482 @@
...
+use std::collections::BTreeMap;
+
+use crate::RpcResult;
+
+pub type P2pAddr = super::addr::Addr<P2pService>;
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Key(pub Bytes);
+
+// rpc Version(google.protobuf.Empty) returns (VersionResponse) {}
```
flub, line 19: delete?  as well as others in this file

iroh-p2p/src/node.rs
```diff
@@ -1074,35 +1076,35 @@ mod tests {
...
+        let server_addr = "irpc://0.0.0.0:4401".parse().unwrap();
+        let client_addr = "irpc://0.0.0.0:4401".parse().unwrap();
+        let t = fetch_providers(
             "/ip4/0.0.0.0/tcp/5001".parse().unwrap(),
             server_addr,
             client_addr,
         )
-        .await?;
+        .await;
+        if let Err(e) = t {
```
dignifiedquire, line 1093: why not just `unwrap`?

iroh-qrpc-client/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "iroh-qrpc-client"
```
dignifiedquire, line 2: can we please stick with the old name?

flub (approved):

iroh-rpc-client/src/client.rs
```diff
@@ -180,140 +188,3 @@ impl Client {
         }
     }
 }
-
-// TODO: write tests for mem transport
-#[cfg(all(test, feature = "grpc"))]
-mod tests {
```
dignifiedquire, line 186: this seems like a nice test, why through it all out?

iroh-rpc-client/src/config.rs
```diff
@@ -40,12 +40,12 @@ impl Source for Config {
...
+            p2p_addr: Some("irpc://127.0.0.1:4401".parse().unwrap()),
+            store_addr: Some("irpc://127.0.0.1:4402".parse().unwrap()),
+            channels: Some(1),
```
dignifiedquire, line 48: why 1 instead of 16? maybe add a comment as to why this is the choosen default

iroh-qrpc-client/Cargo.toml
rklaehn: These are dead files that for some reason I missed.

iroh-metrics/Cargo.toml
rklaehn: No, just made this empty instead of fully removing it. We got those all over the place. Want me to remove them all in this PR?

iroh-qrpc-client/Cargo.toml
rklaehn: Our team is small enough that we might as well add the individual names...

iroh-rpc-client/src/network.rs
rklaehn: Not sure why it is commented out. Keep.

iroh-rpc-client/src/status.rs
rklaehn: I think it is related to the status stuff that was entirely removed.

iroh-p2p/src/rpc.rs
```diff
@@ -188,36 +165,57 @@ impl RpcP2p for P2p {
...
+        async move {
+            let stream = self.fetch_provider_dht0(req).await?;
+            Ok(stream.map(|x| x.map_err(RpcError::from)))
+        }
+        .try_flatten_stream()
+        .boxed()
+    }
+
+    #[tracing::instrument(skip(self, req))]
+    async fn fetch_provider_dht0(
```
dignifiedquire, line 181: why are there two methods?

iroh-rpc-client/Cargo.toml
dignifiedquire: yep

iroh-rpc-client/README.md
```diff
@@ -1,17 +1,11 @@
...
 [iroh](https://github.com/n0-computer/iroh) services internally communicate via
-RPC, using the [gRPC protocol](https://grpc.io/) and [protocol
-buffers](https://developers.google.com/protocol-buffers). This crate provides
-an RPC client that can be used to talk to an [iroh
-gateway](https://github.com/n0-computer/iroh/tree/main/iroh-gateway), an [iroh
-p2p node](https://github.com/n0-computer/iroh/tree/main/iroh-p2p), and the
-[iroh store](https://github.com/n0-computer/iroh/tree/main/iroh-store). 
+RPC, using the [quic-rpc](https://github.com/n0-computer/quic-rpc) RPC system.
+
+TLDR: bincode encoded messages sent as http2 frames.
```
dignifiedquire, line 6: maybe "currently bincode..."

iroh-p2p/src/rpc.rs
rklaehn: One is boilerplate to adapt the function signature to what server_streaming expects. Added a comment.

iroh-rpc-types/src/addr.rs
```diff
@@ -1,101 +1,96 @@
...
         }
     }
 }
 
-impl<T> Addr<T> {
-    pub fn new_mem() -> (Addr<Receiver<T>>, Addr<Sender<T>>) {
-        let (s, r) = tokio::sync::mpsc::channel(256);
+impl<S: Service> Addr<S> {
+    pub fn new_mem() -> Self {
+        let (server, client) = mem::connection(1);
```
dignifiedquire, line 34: buffer of 1 seems to be very small

iroh-rpc-types/src/p2p.rs
```diff
@@ -1,34 +1,482 @@
...
+pub struct LookupRequest {
+    pub peer_id: PeerId,
+    pub addr: Option<Multiaddr>,
+}
+
+// rpc LookupLocal(google.protobuf.Empty) returns (PeerInfo) {}
+#[derive(Serialize, Deserialize, Debug)]
+pub struct LookupLocalRequest;
+
+// message PeerInfo {
```
dignifiedquire, line 199: can probably delete these protobuf comments before merging

iroh-rpc-client/src/network.rs
```diff
@@ -248,44 +232,61 @@ impl P2pClient {
...
+        StatusRow {
+            name: "p2p",
+            number: 2,
+            status,
+        }
+    }
+
+    #[tracing::instrument(skip(self))]
+    pub async fn watch(&self) -> impl Stream<Item = StatusRow> {
+        futures::stream::pending()
```
dignifiedquire, line 288: sad face

iroh-rpc-client/src/gateway.rs
```diff
@@ -1,25 +1,53 @@
...
+
+    #[tracing::instrument(skip(self))]
+    pub async fn check(&self) -> StatusRow {
+        let status: ServiceStatus = self
+            .version()
+            .await
+            .map(|_| ServiceStatus::Serving)
+            .unwrap_or_else(|_e| ServiceStatus::Unknown);
+        StatusRow {
+            name: "gateway",
```
dignifiedquire, line 43: I can has `const` back?

iroh-rpc-client/src/status.rs
```diff
@@ -1,81 +1,10 @@
...
-                },
-                Err(status) => yield StatusRow::new(display_name, 1, ServiceStatus::Down(status)),
-            }
-            /// wait before attempting to start a watch stream again
-            tokio::time::sleep(WAIT).await;
-        };
-    }
-}
-
-impl std::convert::From<HealthCheckResponse> for ServiceStatus {
```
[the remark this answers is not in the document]
rklaehn: Nope.

iroh-rpc-client/src/network.rs
rklaehn: My proposal was to deal with this after the merge. See https://github.com/n0-computer/iroh/issues/584

iroh-rpc-client/src/client.rs
rklaehn: This is a test for the status stuff, which is from tonic-health. It basically does not exist in this PR, so there is nothing to test.

iroh-metrics/Cargo.toml
flub: Personally I would clean them up, but I don't mind.  I was mostly curious if I missed something about what their presence achieved.

flub (approved):

iroh-rpc-client/src/status.rs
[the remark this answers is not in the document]
dignifiedquire: ack

iroh-rpc-client/src/client.rs
dignifiedquire: fair

PR #585, opened 2022-12-08
refactor(`iroh-api`): add takes a `UnixfsEntry` not a `Path`

### LINKED ISSUE #458: iroh API methods should not write or read to disk
An API call currently calls `save_get_stream` which writes to disk:

https://github.com/n0-computer/iroh/blob/bd68e487ed0d0e07806f36eea4f4d8b5edcdef85/iroh-api/src/api_ext.rs#L13-L33

This shouldn't happen. Instead we should surface a helper method that makes it easy to write to disk, but the result should be streamed back as an API response that the user has more options to deal with.

Our API reads & writes to disk in many contexts. We should audit & remove these for _all_ of the iroh API

ramfox:
closes #458 

- [x] re-export builder structs from `iroh-unixfs::builder`
- [x] allow building from a path in `DirectoryBuilder`
- [x] make `builder::Entry` public
- [x] refactor `add` to take an `Entry`
- [x] remove unneeded `add_file`, `add_symlink`, `add_dir` helper functions
- [x] adjust imports/tests/benches

### DIFF
--- a/iroh-api/benches/add.rs
+++ b/iroh-api/benches/add.rs
@@ -1,15 +1,17 @@
 use criterion::{black_box, criterion_group, criterion_main, BenchmarkId, Criterion};
 use futures::TryStreamExt;
 use iroh_metrics::config::Config as MetricsConfig;
-use iroh_rpc_client::Client;
-use iroh_rpc_client::Config as RpcClientConfig;
+use iroh_resolver::resolver::Resolver;
+use iroh_rpc_client::{Client, Config as RpcClientConfig};
 use iroh_rpc_types::Addr;
 use iroh_store::{Config as StoreConfig, Store};
 use iroh_unixfs::{
-    builder::Config as UnixfsConfig,
     chunker::{ChunkerConfig, DEFAULT_CHUNKS_SIZE},
+    content_loader::{FullLoader, FullLoaderConfig},
 };
 use tokio::runtime::Runtime;
+
+use iroh_api::{Api, UnixfsConfig, UnixfsEntry};
 
 pub fn add_benchmark(c: &mut Criterion) {
     let mut group = c.benchmark_group("unixfs_add_file");
@@ -44,7 +46,7 @@ pub fn add_benchmark(c: &mut Criterion) {
                     rpc_client: rpc_client.clone(),
                     metrics: MetricsConfig::default(),
                 };
-                let (_task, rpc) = executor.block_on(async {
+                let (_task, client, resolver) = executor.block_on(async {
                     let store = Store::create(config).await.unwrap();
                     let task = executor.spawn(async move {
                         iroh_store::rpc::new(server_addr, store).await.unwrap()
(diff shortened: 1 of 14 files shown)


iroh-api/benches/add.rs
```diff
@@ -42,30 +45,40 @@ pub fn add_benchmark(c: &mut Criterion) {
...
+                        },
+                    )
+                    .unwrap();
+                    let resolver = Resolver::new(content_loader);
+
+                    (task, client, resolver)
                 });
                 b.to_async(&executor).iter(|| {
-                    let rpc = rpc.clone();
+                    let api = Api::from_client_and_resolver(client.clone(), resolver.clone());
```
ramfox, line 70: I tried this first:
```rust
let (_task, api) = executor.block_on(async {
   ...
   let api = Api::from_config(api_config).await.unwrap();
   (task, api)
});
b.to_async(&executor).iter(|| {
   let api = api.clone();
});
```
But the complier said "no clone found in this scope", even those I derived `Clone` on `Api`, and tried implementing it manually.

Any thoughts on why I couldn't do this?

dignifiedquire: you might need to do this

```rust
```

dignifiedquire: also make sure you are not running into the mocks

iroh-api/benches/add.rs
```diff
@@ -1,16 +1,18 @@
...
 use iroh_unixfs::{
-    builder::Config as UnixfsConfig,
     chunker::{ChunkerConfig, DEFAULT_CHUNKS_SIZE},
+    content_loader::{FullLoader, FullLoaderConfig},
 };
 use tokio::runtime::Runtime;
 
+use iroh_api::{Api, UnixfsConfig, UnixfsEntry};
+
 pub fn add_benchmark(c: &mut Criterion) {
```
dignifiedquire, line 16: pretty sure this doesn't need to be `pub`

iroh-api/src/block_store.rs
```diff
@@ -0,0 +1,118 @@
...
+    async fn put_many(&self, blocks: Vec<Block>) -> Result<()> {
+        let mut this = self.lock().await;
+        for block in blocks {
+            this.insert(*block.cid(), block.data().clone());
+        }
+        Ok(())
+    }
+}
+
+use async_stream::stream;
```
dignifiedquire, line 57: please move to the top

iroh-api/src/block_store.rs
```diff
@@ -0,0 +1,118 @@
...
+                cid,
+                raw_data_size,
+            ));
+        }
+        // make sure to also send the last chunk!
+        store.put_many(chunk).await?;
+    }
+}
+
+fn _add_blocks_to_store_single<S: Store>(
```
dignifiedquire, line 89: do we need to keep this around?

iroh-unixfs/src/builder.rs
```diff
@@ -411,6 +452,8 @@ impl Default for DirectoryBuilder {
             name: None,
             entries: Default::default(),
             typ: DirectoryType::Basic,
+            chunker: None,
```
dignifiedquire, line 455: should be the same as the default for `File`

iroh-api/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 mod api;
+mod block_store;
```
dignifiedquire, line 2: call this just `store` please

iroh-api/benches/add.rs
ramfox: yea it was a mocks issue, I'm side stepping this by not cloning the `Api` struct directly, and cloning the `Resolver` and `Client` instead... should be refactored when we fix the mocks situation.

iroh-api/src/block_store.rs
ramfox: I think it was just being used to compare add speeds. It's a copy-paste from `iroh-unixfs::builder`. I'm comfortable removing it.

iroh-unixfs/src/builder.rs
ramfox: ahh, yea, it should also be able to take a `degree` as well, just like `FileBuilder`

thanks

dignifiedquire (approved):

PR #586, opened 2022-12-12
[wip] iroh-memesync

### LINKED ISSUE #563: implement first version of memesync

dignifiedquire:
Opening this mostly for visibility, but it does transfer some bytes around

Closes n0-computer/beetle#59


dignifiedquire:
@rklaehn trying to stream responses, but this results in a strange error (empty response stream). See https://github.com/n0-computer/iroh/pull/586/commits/9c88a29cf307a62c7432f351d9fe9844eb259e69 for the code I am using. 
The failing test is just the one in `iroh-share`.

iroh-memesync/src/message.rs
```diff
@@ -0,0 +1,222 @@
...
+    pub fn id(&self) -> QueryId {
+        match self {
+            Message::Request(r) => r.id,
+            Message::Response(r) => r.id,
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, Hash)]
+pub struct QueryId(u64);
```
hannahhoward, line 38: make it a UUID... you'll be happy you did in the future :P

- some rando browsing this code

iroh-memesync/src/message.rs
```diff
@@ -0,0 +1,222 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum Message {
+    Response(Response),
+    Request(Request),
+}
+
+impl Message {
+    /// Decoding from a given `BytesMut`.
+    pub fn from_bytes(bytes: BytesMut) -> Result<Self, Error> {
+        let msg = bincode::deserialize(&bytes)?;
```
hannahhoward, line 19: CBOR? Just for cross language compatibility maybe?

hannahhoward (commented):
I'm sure you don't expect anyone reviewing this but it's fun to follow a long your work. Generally looks awesome.

iroh-memesync/src/message.rs
dignifiedquire: I am still very undecided on the format, not a big fan of cbor, so for now this is a placeholder encoding

iroh-memesync/src/message.rs
dignifiedquire: what would be the practical benefit? this only needs to be unique for the sender-receiver pair, and even if it is not, not much bad happens. uuids are considerably larger though

hannahhoward: we had uints as graphsync ids originally. 
we eventual made it a UUID, largely to remove a bunch of cruft code matching up IDs with layered protocols like go-data-transfer and filecoin retrieval. Plus I find tracking things with just a single id quite pleasant (as opposed to always appending peer ids)
yes they are larger, it's certainly a tradeoff, maybe not worth making now.
the nice thing about uuids is you can share them between layered or seperate protocols.

PR #588, opened 2022-12-12
feat: Refine headers and block channel size

ppodolsky:
Fix may address #369 but extra tests are required. It helped for my case.
Also, I've added some headers for CORS. Opened to discussion
I'm not sure every header is needed but all headers I have taken from kubo.

### DIFF
--- a/iroh-bitswap/src/client.rs
+++ b/iroh-bitswap/src/client.rs
@@ -84,7 +84,7 @@ impl<S: Store> Client<S> {
         let (notify, mut default_receiver): (
             async_broadcast::Sender<Block>,
             async_broadcast::Receiver<Block>,
-        ) = async_broadcast::broadcast(4098);
+        ) = async_broadcast::broadcast(4096);
         // ensure no blocking is generated
 
         // TODO: track task
--- a/iroh-bitswap/src/client.rs
+++ b/iroh-bitswap/src/client.rs
@@ -84,7 +84,7 @@ impl<S: Store> Client<S> {
         let (notify, mut default_receiver): (
             async_broadcast::Sender<Block>,
             async_broadcast::Receiver<Block>,
-        ) = async_broadcast::broadcast(64);
+        ) = async_broadcast::broadcast(4098);
         // ensure no blocking is generated
 
         // TODO: track task
--- a/iroh-gateway/src/config.rs
+++ b/iroh-gateway/src/config.rs
@@ -215,7 +215,7 @@ mod tests {
     #[test]
     fn test_default_headers() {
         let headers = default_headers();
-        assert_eq!(headers.len(), 3);
+        assert_eq!(headers.len(), 5);
         let h = headers.get(&ACCESS_CONTROL_ALLOW_ORIGIN).unwrap();
         assert_eq!(h, "*");
     }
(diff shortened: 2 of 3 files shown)


iroh-bitswap/src/client.rs
```diff
@@ -84,7 +84,7 @@ impl<S: Store> Client<S> {
         let (notify, mut default_receiver): (
             async_broadcast::Sender<Block>,
             async_broadcast::Receiver<Block>,
-        ) = async_broadcast::broadcast(64);
+        ) = async_broadcast::broadcast(4098);
```
dignifiedquire, line 87: please use 4096 (mostly for the insane folks like me that like powers of two)

iroh-gateway/src/config.rs
```diff
@@ -109,11 +111,26 @@ fn default_headers() -> HeaderMap {
...
+            CONTENT_LENGTH,
+            CONTENT_RANGE,
+            HEADER_X_IPFS_PATH.clone(),
+            HEADER_X_IPFS_ROOTS.clone(),
+            HEADER_X_CHUNKED_OUTPUT.clone(),
+            HEADER_X_STREAM_OUTPUT.clone(),
+        ]
+        .into_iter()
+        .collect::<AccessControlExposeHeaders>(),
+    );
```
fabricedesre, line 133: Note that the CORS headers can be set in the config file, which is a more flexible solution than hardcoding them here. I agree sane defaults are good, but what qualifies as sane with CORS? (lol)

iroh-bitswap/src/client.rs
ppodolsky: Damn I did to mean degree of 2 but my fast math unit has failed

iroh-gateway/src/config.rs
ppodolsky: I will have university degree in this CORS crap soon, access-headers chase me in nightmares

There are two suggestions to define sanity. The first one is passing tests for base web-cases. Having nice defaults for then is not just important but vital. I have experience in distributing IPFS-based apps armed with instruction how to install it and people ALWAYS had been complaining about configuration part. Going to terminal to copy and paste some configuration lines are too hard for most people.

For example, without headers from this PR I wasn't able to access server response from inside service workers. IMO, it is essential because caching inside service workers are important for dWeb-apps, without caching app will degrade severely. 

Another option is just copy headers from kubo. They walked this path already and I believe CORS part may be taken from kubo without brainstorming because deep and sane analysis for CORS will fail anyway lol.

dignifiedquire: I agree having some sane defaults is good, will defer to @Arqu and @b5 for the actual values.

Arqu (approved):
I don't have any strong opinions on this. CORS is a mess as always. Since this feeds into the subdomain PR let's get it in and move on.

Arqu (changes requested):
Actually need tests to pass first.
Think these are the same ones failing on the subdomain PR currently.

dignifiedquire:
@ppodolsky could you make this green please? would love to get this in

ppodolsky:
> @ppodolsky could you make this green please? would love to get this in

Done

Issue #598, opened 2022-12-14
Stream returned from resolve_recursive_with_paths hangs for large non-hamt directory


rklaehn:
When creating a large directory and then traversing it again, the roundtrip test hangs as soon as the directory has more than 2048 entries.

This seems to be related to some bounded channel in the resolver:

```rust
        let (session_closer_s, session_closer_r) = async_channel::bounded(2048);
```

Setting this value to a larger value makes the test pass. Not sure what is happening? Something just accumulates in this channel is not pulled out, and when it is full things hang?

Arqu:
Oh my, we should clean that up! That might be part of bitswap issues too.

rklaehn:
So - this is easy enough to "fix". I tracked it down to `impl Drop for LoaderContext {`. There is a send_blocking in there. Blocking in Drop is never a good idea.

But - what is all this stuff, and why is the channel not drained? What are these sessions, and what is bad about them not being cleaned up? Just seems like a complex machinery...

dignifiedquire:
use a multithreaded executor and the issue goes away…thanks tokio (I haven’t found a way to make this work with the single threaded executor) and tokio defaults to single threaded for tokio::test.

dignifiedquire:
the machinery is needed because we must send an rpc message on drop, and tokio doesn’t allow us to execute any async code in drop sad face

dignifiedquire:
because we must not loose these messages, otherwise we leak memory

rklaehn:
So what exactly is being cleaned up here? A bitswap session?

rklaehn:
OK, I still don't like this very much. But it seems that it is not a bug.

PR #599, opened 2022-12-14
test: add large directory roundtrip test

rklaehn:
This was originally intended for the HAMT writing, but uncovered a bug unrelated to HAMTs. At >= 2048 items the stream just hangs. See issue https://github.com/n0-computer/iroh/issues/598

### DIFF
--- a/iroh-resolver/tests/roundtrip.rs
+++ b/iroh-resolver/tests/roundtrip.rs
@@ -1,7 +1,7 @@
 use anyhow::Result;
 use async_recursion::async_recursion;
 use bytes::Bytes;
-use futures::{Stream, TryStreamExt};
+use futures::{Stream, StreamExt, TryStreamExt};
 use iroh_metrics::resolver::OutMetrics;
 use iroh_unixfs::{
     balanced_tree::DEFAULT_DEGREE,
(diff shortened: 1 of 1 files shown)


rklaehn:
@dignifiedquire so you are right, using a multi-threaded executor makes the test work.

But still I really, really, _really_ don't like send_blocking in Drop.

dignifiedquire:
> But still I really, really, _really_ don't like send_blocking in Drop.

I would love alternatives, but we must guarantee the rpc message send on `Drop`, if we don't do that things break down badly.

rklaehn:
> I would love alternatives, but we must guarantee the rpc message send on Drop, if we don't do that things break down badly.

How can you rely on a RPC call for some essential cleanup task? You would think that there would have to be some additional fallback cleanup in case the RPC fails.

dignifiedquire:
> How can you rely on a RPC call for some essential cleanup task? Is there an alternative cleanup mechanism in case the RPC fails?

The more load the system is under, the more important it is to get this message through to clean up as soon as possible. There are timeout based cleanups happening as well, but they are just to a last resort. I ran tests with `try_send` and the system degrades much faster.

rklaehn:
Added this test to #603

PR #602, opened 2022-12-15
add back `iroh status -w` feature using new rpc streams

### LINKED ISSUE #584: Refine status monitoring
It used to be based on tonic-health, which of course no longer makes sense. We need to do something basic ourselves.

Currently it is just hammering the version endpoint, but I guess that needs some refinement.

Basically the tonic health thing allows a service to announce itself that it is currently non-functional, that is the biggest difference.

https://github.com/grpc/grpc/blob/master/doc/health-checking.md

ramfox:
closes #584

### DIFF
--- a/iroh-api/src/api.rs
+++ b/iroh-api/src/api.rs
@@ -7,7 +7,7 @@ use crate::IpfsPath;
 use crate::P2pApi;
 use anyhow::{ensure, Context, Result};
 use cid::Cid;
-use futures::stream::{BoxStream, LocalBoxStream};
+use futures::stream::BoxStream;
 use futures::{StreamExt, TryStreamExt};
 use iroh_resolver::resolver::Resolver;
 use iroh_rpc_client::{Client, ClientStatus};
@@ -171,8 +171,8 @@ impl Api {
         self.client.check().await
     }
 
-    pub async fn watch(&self) -> LocalBoxStream<'static, ClientStatus> {
-        self.client.clone().watch().await.boxed_local()
+    pub async fn watch(&self) -> BoxStream<'static, ClientStatus> {
+        self.client.clone().watch().await.boxed()
     }
 
     /// The `add_stream` method encodes the entry into a DAG and adds
--- a/iroh-api/src/api.rs
+++ b/iroh-api/src/api.rs
@@ -10,8 +10,7 @@ use cid::Cid;
 use futures::stream::BoxStream;
 use futures::{StreamExt, TryStreamExt};
 use iroh_resolver::resolver::Resolver;
-use iroh_rpc_client::Client;
-use iroh_rpc_client::StatusTable;
+use iroh_rpc_client::{Client, ClientStatus};
 use iroh_unixfs::{
     builder::Entry as UnixfsEntry,
     content_loader::{FullLoader, FullLoaderConfig},
@@ -168,11 +167,11 @@ impl Api {
         Ok(stream.boxed())
     }
 
-    pub async fn check(&self) -> StatusTable {
+    pub async fn check(&self) -> ClientStatus {
         self.client.check().await
     }
 
-    pub async fn watch(&self) -> BoxStream<'static, StatusTable> {
+    pub async fn watch(&self) -> BoxStream<'static, ClientStatus> {
         self.client.clone().watch().await.boxed()
     }
 
--- a/iroh-api/src/api.rs
+++ b/iroh-api/src/api.rs
@@ -7,7 +7,7 @@ use crate::IpfsPath;
 use crate::P2pApi;
 use anyhow::{ensure, Context, Result};
 use cid::Cid;
-use futures::stream::BoxStream;
+use futures::stream::{BoxStream, LocalBoxStream};
 use futures::{StreamExt, TryStreamExt};
 use iroh_resolver::resolver::Resolver;
 use iroh_rpc_client::{Client, ClientStatus};
@@ -171,8 +171,8 @@ impl Api {
         self.client.check().await
     }
 
-    pub async fn watch(&self) -> BoxStream<'static, ClientStatus> {
-        self.client.clone().watch().await.boxed()
+    pub async fn watch(&self) -> LocalBoxStream<'static, ClientStatus> {
+        self.client.clone().watch().await.boxed_local()
     }
 
     /// The `add_stream` method encodes the entry into a DAG and adds
--- a/iroh-gateway/src/metrics.rs
+++ b/iroh-gateway/src/metrics.rs
@@ -1,8 +1,10 @@
 use iroh_metrics::config::Config as MetricsConfig;
 
+use crate::VERSION;
+
 pub fn metrics_config_with_compile_time_info(cfg: MetricsConfig) -> MetricsConfig {
     // compile time configuration
-    cfg.with_service_name(env!("CARGO_PKG_NAME").to_string())
+    cfg.with_service_name(VERSION.to_string())
         .with_build(
             git_version::git_version!(
                 prefix = "git:",
--- a/iroh-gateway/src/metrics.rs
+++ b/iroh-gateway/src/metrics.rs
@@ -4,7 +4,7 @@ use crate::VERSION;
 
 pub fn metrics_config_with_compile_time_info(cfg: MetricsConfig) -> MetricsConfig {
     // compile time configuration
-    cfg.with_service_name(VERSION.to_string())
+    cfg.with_service_name(env!("CARGO_PKG_NAME").to_string())
         .with_build(
             git_version::git_version!(
                 prefix = "git:",
@@ -13,5 +13,5 @@ pub fn metrics_config_with_compile_time_info(cfg: MetricsConfig) -> MetricsConfi
             )
             .to_string(),
         )
-        .with_version(env!("CARGO_PKG_VERSION").to_string())
+        .with_version(VERSION.to_string())
 }
--- a/iroh-gateway/src/rpc.rs
+++ b/iroh-gateway/src/rpc.rs
@@ -1,9 +1,11 @@
 use std::result;
 
 use anyhow::Result;
+use futures::stream::{BoxStream, StreamExt};
 use iroh_rpc_client::{create_server, GatewayServer, ServerError, ServerSocket};
 use iroh_rpc_types::gateway::{
-    GatewayAddr, GatewayRequest, GatewayService, VersionRequest, VersionResponse,
+    GatewayAddr, GatewayRequest, GatewayService, VersionRequest, VersionResponse, WatchRequest,
+    WatchResponse,
 };
 use tracing::info;
 
@@ -11,6 +13,17 @@ use tracing::info;
 pub struct Gateway {}
 
 impl Gateway {
+    #[tracing::instrument(skip(self))]
+    fn watch(self, _: WatchRequest) -> BoxStream<'static, WatchResponse> {
+        async_stream::stream! {
+            loop {
+                yield WatchResponse { version: env!("CARGO_PKG_VERSION").to_string() };
+                tokio::time::sleep(std::time::Duration::from_millis(1000)).await;
+            }
+        }
+        .boxed()
+    }
+
     #[tracing::instrument(skip(self))]
     async fn version(self, _: VersionRequest) -> VersionResponse {
         let version = env!("CARGO_PKG_VERSION").to_string();
@@ -31,6 +44,7 @@ async fn dispatch(
 ) -> result::Result<(), ServerError> {
     use GatewayRequest::*;
     match req {
+        Watch(req) => s.server_streaming(req, chan, target, Gateway::watch).await,
         Version(req) => s.rpc(req, chan, target, Gateway::version).await,
     }
 }
--- a/iroh-gateway/src/rpc.rs
+++ b/iroh-gateway/src/rpc.rs
@@ -1,33 +1,36 @@
-use std::result;
+use std::{result, time::Duration};
 
 use anyhow::Result;
-use futures::stream::{BoxStream, StreamExt};
+use futures::stream::Stream;
 use iroh_rpc_client::{create_server, GatewayServer, ServerError, ServerSocket};
-use iroh_rpc_types::gateway::{
-    GatewayAddr, GatewayRequest, GatewayService, VersionRequest, VersionResponse, WatchRequest,
-    WatchResponse,
+use iroh_rpc_types::{
+    gateway::{GatewayAddr, GatewayRequest, GatewayService},
+    VersionRequest, VersionResponse, WatchRequest, WatchResponse,
 };
 use tracing::info;
+
+const VERSION: &str = env!("CARGO_PKG_VERSION");
+const WAIT: Duration = Duration::from_secs(1);
 
 #[derive(Default, Debug, Clone)]
 pub struct Gateway {}
 
 impl Gateway {
     #[tracing::instrument(skip(self))]
-    fn watch(self, _: WatchRequest) -> BoxStream<'static, WatchResponse> {
+    fn watch(self, _: WatchRequest) -> impl Stream<Item = WatchResponse> {
         async_stream::stream! {
             loop {
-                yield WatchResponse { version: env!("CARGO_PKG_VERSION").to_string() };
-                tokio::time::sleep(std::time::Duration::from_millis(1000)).await;
+                yield WatchResponse { version: VERSION.to_string() };
+                tokio::time::sleep(WAIT).await;
             }
         }
-        .boxed()
     }
 
     #[tracing::instrument(skip(self))]
     async fn version(self, _: VersionRequest) -> VersionResponse {
-        let version = env!("CARGO_PKG_VERSION").to_string();
-        VersionResponse { version }
+        VersionResponse {
+            version: VERSION.to_string(),
+        }
     }
 }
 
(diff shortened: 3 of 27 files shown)


iroh-gateway/src/rpc.rs
```diff
@@ -1,16 +1,29 @@
...
+    fn watch(self, _: WatchRequest) -> BoxStream<'static, WatchResponse> {
+        async_stream::stream! {
+            loop {
+                yield WatchResponse { version: env!("CARGO_PKG_VERSION").to_string() };
+                tokio::time::sleep(std::time::Duration::from_millis(1000)).await;
+            }
+        }
+        .boxed()
+    }
+
```
ramfox, line 26: Okay, so this loop seems silly, is there a way to say "just never close"?

iroh-gateway/src/rpc.rs
```diff
@@ -1,16 +1,29 @@
...
+    WatchResponse,
 };
 use tracing::info;
 
 #[derive(Default, Debug, Clone)]
 pub struct Gateway {}
 
 impl Gateway {
+    #[tracing::instrument(skip(self))]
+    fn watch(self, _: WatchRequest) -> BoxStream<'static, WatchResponse> {
```
rklaehn, line 17: This can be just `impl Stream<Item = WatchResponse>`. It's not a trait, so you don't have to box it.

iroh-gateway/src/rpc.rs
```diff
@@ -1,16 +1,29 @@
...
+        async_stream::stream! {
+            loop {
+                yield WatchResponse { version: env!("CARGO_PKG_VERSION").to_string() };
```
rklaehn, line 20: Can we DRY the computation of the version? Make a fn version_string() or something?

iroh-gateway/src/rpc.rs
rklaehn: No, I think this is fine. If the client closes the connection, the stream will no longer be polled and will be dropped.

iroh-gateway/src/rpc.rs
```diff
@@ -1,16 +1,29 @@
...
+                tokio::time::sleep(std::time::Duration::from_millis(1000)).await;
+            }
+        }
+        .boxed()
```
rklaehn, line 24: no box, see above.

iroh-p2p/src/rpc.rs
```diff
@@ -32,6 +32,17 @@ impl P2p {
         Self { sender }
     }
 
+    #[tracing::instrument(skip(self))]
+    fn watch(self, _: WatchRequest) -> BoxStream<'static, WatchResponse> {
```
rklaehn, line 36: no box, see above

iroh-p2p/src/rpc.rs
```diff
@@ -32,6 +32,17 @@ impl P2p {
...
+        async_stream::stream! {
+            loop {
+                yield WatchResponse { version: env!("CARGO_PKG_VERSION").to_string() };
```
rklaehn, line 39: DRY version, see above

iroh-rpc-client/src/client.rs
```diff
@@ -181,9 +181,9 @@ impl Client {
             }
 
             let mut stream = futures::stream::select_all(streams);
-            while let Some(status) = stream.next().await {
-                status_table.update(status).unwrap();
-                yield status_table.clone();
+            while let Some(s) = stream.next().await {
+                status.update(s).unwrap();
```
rklaehn, line 185: So this will panic if the service is unknown, right? Could the ClientStatus also simply be a map from service name to status, then you would not have to hardcode the service names?

iroh-rpc-client/src/gateway.rs
```diff
@@ -33,22 +35,38 @@ impl GatewayClient {
...
-        futures::stream::pending()
+    pub async fn watch(&self) -> impl Stream<Item = ServiceStatus> {
+        let client = self.client.clone();
+        stream! {
+            loop {
+                let res = client.server_streaming(WatchRequest).await;
+                match res {
+                    Ok(mut res) => {
+                        while let Some(v) = res.next().await {
+                            let (status, version) = v.map_or((StatusType::Down, String::new()), |v| (StatusType::Serving, v.version));
```
rklaehn, line 60: If you get an error here, there might be something wrong with the connection. It is probably better to just reconnect in that case. Otherwise you might just loop forever with a broken connection.

So just make it `while let Some(Ok(v)) = ...`.

iroh-rpc-client/src/network.rs
```diff
@@ -270,23 +271,39 @@ impl P2pClient {
...
-        StatusRow {
+    pub async fn check(&self) -> ServiceStatus {
+        let (status, version) = match self.version().await {
+            Ok(version) => (StatusType::Serving, version),
+            Err(_) => (StatusType::Down, String::new()),
+        };
+        ServiceStatus {
             name: "p2p",
-            number: 2,
+            number: 1,
```
rklaehn, line 281: Do we need this number at all? It seems it was meant to be the column or row for the service. But we are doing that with the name now, right?

iroh-rpc-client/src/network.rs
```diff
@@ -270,23 +271,39 @@ impl P2pClient {
...
-        // todo
-        futures::stream::pending()
+    pub async fn watch(&self) -> impl Stream<Item = ServiceStatus> {
+        let client = self.client.clone();
+        stream! {
+            loop {
+                let res = client.server_streaming(WatchRequest).await;
+                match res {
+                    Ok(mut res) => {
+                        while let Some(v) = res.next().await {
```
rklaehn, line 295: while let Some(Ok(...)) =

iroh-rpc-client/src/status.rs
```diff
@@ -2,32 +2,37 @@ use crate::{gateway, network, store};
...
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub enum ServiceStatus {
-    ///  Indicates rpc server is in an unknown state
+pub enum StatusType {
+    /// Indicates service status is unknown
     Unknown,
     /// Indicates service is serving data
     Serving,
-    /// Indicates service is not serving data, but the rpc server is not down
-    NotServing,
```
rklaehn, line 11: Not for this PR, but with this we have the ability for the service to actively signal that it is not feeling well.

iroh-rpc-types/src/gateway.rs
```diff
@@ -1,10 +1,23 @@
...
+use crate::RpcResult;
+
 /// Gateway address
 pub type GatewayAddr = crate::addr::Addr<GatewayService>;
 
+#[derive(Serialize, Deserialize, Debug)]
+pub struct WatchRequest;
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct WatchResponse {
```
rklaehn, line 17: Just FYI while the request has to be unique, the response does not have to. So in theory you could have the WatchRequest return VersionResponses.

Probably not a good idea though, since we might want to add additional info to the WatchResponse.

But one thing we could do: VersionRequest and WatchRequest / Response are common to all 3 services, and we want them to contain the same info for all 3 services. So we could move Watch/Version Request/Response to the top level of iroh-rpc-types. I think that would be a nice bit of DRY.

iroh-rpc-types/src/gateway.rs
```diff
@@ -34,3 +50,11 @@ impl Service for GatewayService {
 impl RpcMsg<GatewayService> for VersionRequest {
     type Response = VersionResponse;
 }
+
+impl Msg<GatewayService> for WatchRequest {
```
rklaehn, line 54: Even if we move the WatchRequest / VersionRequest one level up, these would have to be declared separately for each service.

iroh-rpc-types/src/p2p.rs
```diff
@@ -281,6 +291,14 @@ impl Service for P2pService {
     type Res = P2pResponse;
 }
 
+impl Msg<P2pService> for WatchRequest {
+    type Response = WatchResponse;
+
+    type Update = Self;
```
rklaehn, line 297: You will never get an update, but you have to put something here. A rust never type would be nice. But it is not stable yet...

https://doc.rust-lang.org/std/primitive.never.html

iroh-rpc-client/src/client.rs
rklaehn: No, actually this is fine. It is better for the code using this to have the three services we care about as explicit fields.

Just maybe make this an expect so we have a clue what is going on if this panics.

rklaehn:
Great work picking up the new stuff so quickly.

iroh-gateway/src/rpc.rs
dignifiedquire: also should be a `OnceCell` or sth like that, this only needs to be pulled once per runtime

iroh-gateway/src/rpc.rs
```diff
@@ -1,16 +1,29 @@
...
 #[derive(Default, Debug, Clone)]
 pub struct Gateway {}
 
 impl Gateway {
+    #[tracing::instrument(skip(self))]
+    fn watch(self, _: WatchRequest) -> BoxStream<'static, WatchResponse> {
+        async_stream::stream! {
+            loop {
+                yield WatchResponse { version: env!("CARGO_PKG_VERSION").to_string() };
+                tokio::time::sleep(std::time::Duration::from_millis(1000)).await;
```
dignifiedquire, line 21: `Duration::from_secs(1)` reads a little easier :) also would be nice to put into a constant

iroh-rpc-client/src/network.rs
dignifiedquire: that would break the loop on an error, which is not what happens here

iroh-rpc-client/src/status.rs
```diff
@@ -39,33 +44,38 @@ impl StatusRow {
         self.number
     }
 
-    pub fn status(&self) -> ServiceStatus {
+    pub fn status(&self) -> StatusType {
         self.status.clone()
     }
+
+    pub fn version(&self) -> String {
```
dignifiedquire, line 51: return a `&str` to avoid `cloning` when the caller doesn't need it

iroh-rpc-client/src/store.rs
```diff
@@ -66,22 +68,38 @@ impl StoreClient {
...
             status,
+            version,
         }
     }
 
     #[tracing::instrument(skip(self))]
-    pub async fn watch(&self) -> impl Stream<Item = StatusRow> {
-        // todo
-        futures::stream::pending()
+    pub async fn watch(&self) -> impl Stream<Item = ServiceStatus> {
```
dignifiedquire, line 85: these seem to be the same, maybe you could extract this into a generic function that is called for these?

iroh-rpc-types/src/p2p.rs
dignifiedquire: @rklaehn you could expose a custom `Never` type from  quick-rpc for these cases, which is just `struct Never;` for now and can be changed to the real one later

iroh-rpc-client/src/network.rs
ramfox: It's a holdover, preemptive work for when we have more than one of `Service`. It's unnecessary right now & I'm planning to do a round to remove it from the response, but keep it in the `iroh status` output.

iroh-rpc-client/src/status.rs
ramfox: Trying to find the balance of "keep it in for the future" and "take it out because it's not used"

I think this enum variant is a candidate for "keep it in for the future"

iroh-rpc-client/src/gateway.rs
ramfox: oh good point, thank you

iroh-gateway/src/rpc.rs
ramfox: any objection to making `VERSION` a constant?

`const VERSION: &str = env!(CARGO_PKG_VERSION);`

Gets set once at compile time.

dignifiedquire: I belive that should work 👍

iroh-rpc-client/src/store.rs
ramfox: attempted to do this and ran into some limitations. brought it up to @rklaehn and he is brainstorming.

iroh-rpc-types/src/p2p.rs
ramfox: can this be addressed in a follow up?

dignifiedquire: yes

iroh-rpc-client/src/store.rs
dignifiedquire: ok 👌

iroh-gateway/src/rpc.rs
```diff
@@ -1,20 +1,36 @@
...
 use iroh_rpc_client::{create_server, GatewayServer, ServerError, ServerSocket};
-use iroh_rpc_types::gateway::{
-    GatewayAddr, GatewayRequest, GatewayService, VersionRequest, VersionResponse,
+use iroh_rpc_types::{
+    gateway::{GatewayAddr, GatewayRequest, GatewayService},
+    VersionRequest, VersionResponse, WatchRequest, WatchResponse,
 };
 use tracing::info;
 
+const VERSION: &str = env!("CARGO_PKG_VERSION");
```
dignifiedquire, line 12: just realized, this will be different than the version we use in metrics, can you make an issue to update this to match please?

iroh-rpc-client/src/store.rs
```diff
@@ -66,22 +68,37 @@ impl StoreClient {
...
+                        while let Some(v) = res.next().await {
+                            let (status, version) = v.map_or((StatusType::Down, String::new()), |v| (StatusType::Serving, v.version));
+                            yield ServiceStatus::new("store", status, version);
+                        }
+                    },
+                    Err(_) => {
+                        yield ServiceStatus::new("store", StatusType::Down, "");
+                    }
+                }
+                tokio::time::sleep(std::time::Duration::from_millis(1000)).await;
```
dignifiedquire, line 100: nit: from_secs & const

iroh/src/services.rs
```diff
@@ -209,6 +205,7 @@ pub async fn status(api: &Api, watch: bool) -> Result<()> {
                 .queue(style::Print("\n"))?
                 .flush()?;
         }
+        println!("iterations {}", iterations);
         Ok(())
```
dignifiedquire, line 209: is this printed for the user to see?

ramfox: oof, thanks for catching this

iroh-rpc-client/src/store.rs
ramfox: issue filed: https://github.com/n0-computer/iroh/issues/614

iroh-gateway/src/rpc.rs
ramfox: I moved the definition to `lib.rs`, and referenced it in both the `metrics.rs` and `rpc.rs` files for each of `iroh-p2p`, `iroh-gateway`, and `iroh-store`

iroh-rpc-client/src/network.rs
ramfox: removed `number` field

iroh-rpc-client/src/client.rs
ramfox: this `unwrap` has been refactored away

iroh-gateway/src/metrics.rs
```diff
@@ -1,8 +1,10 @@
 use iroh_metrics::config::Config as MetricsConfig;
 
+use crate::VERSION;
+
 pub fn metrics_config_with_compile_time_info(cfg: MetricsConfig) -> MetricsConfig {
     // compile time configuration
-    cfg.with_service_name(env!("CARGO_PKG_NAME").to_string())
```
dignifiedquire, line 5: name and version are sth different 😅

iroh-gateway/src/metrics.rs
```diff
@@ -1,8 +1,10 @@
...
+    cfg.with_service_name(VERSION.to_string())
         .with_build(
             git_version::git_version!(
```
dignifiedquire, line 9: I mesnt the git version for unification

iroh-gateway/src/metrics.rs
ramfox: lmao of course that's what you meant, I just replaced the wrong line 🤦‍♀️ 
well, this is why we have review

ramfox: Okay, so I'm going to replace the `VERSION` to be the git version, to be used in the metrics files and the rpc `version` and `watch` calls

ramfox: hmmm.... the git version looks very bad on `iroh status -w` output

I think this may need further discussion. I'm going to file an issue like you originally asked.

dignifiedquire:
@ramfox you have merge conflicts

iroh-api/src/api.rs
```diff
@@ -168,12 +167,12 @@ impl Api {
...
     }
 
-    pub async fn check(&self) -> StatusTable {
+    pub async fn check(&self) -> ClientStatus {
         self.client.check().await
     }
 
-    pub async fn watch(&self) -> BoxStream<'static, StatusTable> {
-        self.client.clone().watch().await.boxed()
+    pub async fn watch(&self) -> LocalBoxStream<'static, ClientStatus> {
```
dignifiedquire, line 174: lets make this `BoxStream`, as we made all the APIs`Send`

PR #605, opened 2022-12-15
fix: only use name if name is set

### LINKED ISSUE #604: Ipfs add broken on main
I was doing some checks on #603 to see if a hamt directory would appear on the gateway. It does not. But it seems that directory adding on main is also broken.

I created a directory and tried to add it on main.

```
iroh on  main [$?] via 🦀 v1.65.0 
❯ cargo run -p iroh -- add --recursive --offline testdir_many_files/nohamt1 
    Finished dev [unoptimized + debuginfo] target(s) in 0.21s
     Running `target/debug/iroh add --recursive --offline testdir_many_files/nohamt1`
[1/2] Calculating size...
[2/2] Importing content 0 B...
/ipfs/bafybeieopdpzqhyyxeuuipsdxnxlgo5q5ontpwe2j7xdojql3zk6du3jzy
```

Requesting this from the gateway does not give a directory as expected:
image

Requesting this via iroh get also gives an error:
```
iroh on  main [$?] via 🦀 v1.65.0 
❯ cargo run -p iroh -- get /ipfs/bafybeieopdpzqhyyxeuuipsdxnxlgo5q5ontpwe2j7xdojql3zk6du3jzy
    Finished dev [unoptimized + debuginfo] target(s) in 0.21s
     Running `target/debug/iroh get /ipfs/bafybeieopdpzqhyyxeuuipsdxnxlgo5q5ontpwe2j7xdojql3zk6du3jzy`
Error: path contains non-relative component
```

rklaehn:
otherwise we end up with an empty string instead of the dir name.

Fixes #604


iroh-unixfs/src/builder.rs
```diff
@@ -522,15 +522,16 @@ impl DirectoryBuilder {
...
-            return Ok(dir);
-        }
-
-        Ok(Directory { name, entries })
+            if let Some(name) = name {
+                dir.name = name;
+            }
+            dir
+        } else {
+            let name = name.unwrap_or_default();
```
rklaehn, line 532: This fixes the issue, but nevertheless this is all a bit weird. Feels like this should be two different builders, one that builds from a path and one that builds from its contents.

```
enum Builder {
  Path { ... }
  Parts { ... }
}
```

Otherwise, e.g. if entries are set and path is set, the entries are being ignored.

PR #606, opened 2022-12-15
feat: Refactor unixfs

ppodolsky:
`ResponseClip` has been deleted but it is better to discuss. I did it because its semantics was not obvious, it was used for clipping both entire response and partial responses from `Node` with non-trivial transitions between them.

Worth to note, now `Range` header with value `start-end` translates to rust struct `Range` with values `start` and `end + 1`. `Range` in rust non-inclusive for right bound, it is more convinient.

### DIFF
--- a/iroh-gateway/src/headers.rs
+++ b/iroh-gateway/src/headers.rs
@@ -458,7 +458,7 @@ mod tests {
             ),
             Some(ResponseFormat::Raw),
         );
-        let wetag = format!("W/{}", etag);
+        let wetag = format!("W/{etag}");
         let other_etag = get_etag(
             &CidOrDomain::Cid(
                 Cid::try_from("bafkreigh2akiscaildcqabsyg3dfr6chu3fgpregiymsck7e7aqa4aaaaa")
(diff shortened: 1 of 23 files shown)


iroh-gateway/src/headers.rs
```diff
@@ -103,11 +103,14 @@ pub fn parse_range_header(range: &HeaderValue) -> Option<Range<u64>> {
     if start >= end || end == 0 {
         return None;
     }
-    Some(Range { start, end })
+    Some(Range {
```
Arqu, line 106: Ugh, this is a rough bug on my end.

iroh-resolver/src/resolver.rs
```diff
@@ -519,15 +518,14 @@ impl<T: ContentLoader + Unpin + 'static> AsyncRead for OutPrettyReader<T> {
             | OutPrettyReader::DagJson(bytes_reader)
             | OutPrettyReader::Raw(bytes_reader) => {
                 let pos_current = bytes_reader.pos;
-                let res = poll_read_buf_at_pos(
+                let bytes_read = read_data_to_buf(
                     &mut bytes_reader.pos,
-                    ResponseClip::Clip(bytes_reader.bytes.len()),
-                    &bytes_reader.bytes,
+                    Some(bytes_reader.bytes.len()),
+                    &bytes_reader.bytes[pos_current..],
```
Arqu, line 524: 👍

Arqu (approved):
Thank you for the deep work here. VERY subtle bugs you caught out. 
And magnificent improvements on the seeking.

PR #607, opened 2022-12-15
chore(ccov): disable annotations

Arqu:


flub:
Believe it or not, this has motivated me to add tests which turned out to be a useful endeavour...  But yeah, also annoyingly noisy.

Arqu:
Yeah, it can be useful, but it's main feature is bug me until I give in...
Either way, it should be more of a culture thing to keep an eye on the codecov reports and checks.

PR #608, opened 2022-12-15
fix(ci): make sure correct rust version is set

Arqu:


Arqu:
Irrelevant, just looking at something. 1.66 seems to break builds now on clippy.

PR #610, opened 2022-12-15
(chore): de-dupe axum

fabricedesre:
0.6 is pulled by opentelemetry-otlp -> tonic


Arqu:
Can you rebase, there are fixes to address some of the current tests (rust 1.66 made clippy fail).
There will probably be more issues due to the version change.

fabricedesre:
Tests are fixed locally - let's see if CI agrees!

Arqu (approved):
Not the prettiest with `subdomain_router2` etc but given the changed requirements for routing on Axum, makes sense. Tyvm!

PR #612, opened 2022-12-16
chore: Add weekly CI running beta clippy

flub:
This adds a weekly CI job which run clippy with the rust beta version.
This gives a heads up for failures which we can fix ahead of the next
stable release, instead of all PRs being blocked on release day.

---

I've previously used this without the issue creation, I'm not sure how
well the issue creation part works.

Problem is that without github will email the failure to **the last
committer** of the action file.  Which is a bit suboptimal.

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -1,0 +1,52 @@
+name: Weekly CI
+
+on:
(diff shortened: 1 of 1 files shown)


Arqu (approved):
LGTM

PR #623, opened 2022-12-19
fix: actually resolve and check paths

Arqu:
We never really completed resolving a path at the top and verifying some checks against it. 
Also threaded it through so we don't re-resolve.


dignifiedquire (approved):
tests? 😝

Arqu:
Agree this needs a test. To move forward on the 404 handling I filed an issue and put it on the board.

Issue #518, opened 2022-11-18
Fix flakey `test_dht` function


b5:
currently failing consistently on our M1 Mac CI runner:

https://github.com/n0-computer/iroh/blob/cd9fa2354e12cbeb6c4391007dd6170151108dfa/iroh-p2p/src/node.rs#L1448

ramfox:
- even when we ensure the nodes are connected and have been added to kademlia, they don't seem to make it into any kbuckets, need to investigate how/when these get updated

PR #625, opened 2022-12-20
fix(iroh-p2p): ensure `providers` stream closes

### LINKED ISSUE #455: p2p unit tests

ramfox:
Also implements final p2p unit test to close #455
Bug fix fixes the flakey dht tests #518

Some notes:
- Before we could get to fixing the flakiness, the test was hanging due to a bug in the `Providers` query manager. To summarize the issue: we weren't reacting to `GetProvidersOk::FinishedWithNoAdditionalRecord` events, which indicate no more records were found. Because this event was never handled, we never removed the `Query` from the `providers` Set, and so the stream of providers never closed, causing the `test_dht` test to hang.

- To test reliability, I looped the `test_dht` 10 times for my initial commits, but removed the loop it once things stabilized.

- I did a whole pass where the node emitted an event whenever the kbuckets were updated, and used that event to ensure everything was "in place" before attempting to `StartProviding` and `fetch_providers_dht`. However... this made things less reliable! So I removed it. It also would have potentially added a lot of "noise" to the `network_event` stream.

- The only thing that actually stabilized the test (when attempting multiple loops)... was increasing the timeout by 1 second. I'm not sure how to feel about this.

- In order to test the `CancelListenForIdentify`, I had to add a `NetworkEvent` that would emit after this cancellation happens. There was no other way I could think of that would allow us to inspect the node itself without breaking the borrow checker. This event would only ever be emitted due to a user action (manually attempting to lookup a peer), so I'm not concerned about it polluting the event space.

### DIFF
--- a/iroh-p2p/src/node.rs
+++ b/iroh-p2p/src/node.rs
@@ -1676,7 +1676,7 @@ mod tests {
             // when `start_providing` waits for the record to make it to the dht
             // we can remove this polling
             let providers = tokio::time::timeout(
-                Duration::from_millis(5000),
+                Duration::from_secs(7),
                 poll_for_providers(test_runner_a.client.clone(), &cid),
             )
             .await
--- a/iroh-p2p/src/node.rs
+++ b/iroh-p2p/src/node.rs
@@ -50,6 +50,8 @@ pub enum NetworkEvent {
     PeerConnected(PeerId),
     PeerDisconnected(PeerId),
     Gossipsub(GossipsubEvent),
+    CancelLookupQuery(PeerId),
+    RoutingTableUpdate(PeerId),
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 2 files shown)


dignifiedquire (approved):
looks good, nice catch

dignifiedquire (approved):

Issue #97, opened 2022-05-31
gateway(template): improve 404 handling


Arqu:
Serve pretty 404 where possible.

PR #628, opened 2022-12-20
feat: html error handling

Arqu:
closes https://github.com/n0-computer/iroh/issues/97

### DIFF
--- a/iroh-gateway/src/handlers.rs
+++ b/iroh-gateway/src/handlers.rs
@@ -129,7 +129,7 @@ pub fn get_app_routes<T: ContentLoader + Unpin>(state: &Arc<State<T>>) -> Router
                 .layer(Extension(Arc::clone(state)))
                 .layer(middleware::from_fn(request_middleware))
                 .layer(CompressionLayer::new())
-                .layer(HandleErrorLayer::new(middleware_error_handler::<T>))
+                .layer(HandleErrorLayer::new(middleware_error_handler))
                 .load_shed()
                 .concurrency_limit(2048 * 1024)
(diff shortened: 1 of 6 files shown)


iroh-gateway/src/handlers.rs
```diff
@@ -180,12 +180,18 @@ async fn request_preprocessing<T: ContentLoader + Unpin>(
         }
     }
 
-    let path_metadata = state
-        .client
-        .retrieve_path_metadata(path.clone())
-        .await
-        .map_err(|e| GatewayError::new(StatusCode::INTERNAL_SERVER_ERROR, &e))?;
-    // TODO: handle 404 or error
+    let path_metadata = match state.client.retrieve_path_metadata(path.clone()).await {
```
Arqu, line 183: @b5 this is the crucial bit - the earliest moment we ping store/p2p. I've updated error responses from `fetch_gateway` and `fetch_bitswap` so that it now catches them and returns a more sane response (`failed to find {cid}` || `offline`). 
We should do similar for `fetch_store` to make it less jarring to the end users.

dignifiedquire (approved):
tests :P ?

iroh-gateway/src/handlers.rs
```diff
@@ -180,12 +180,18 @@ async fn request_preprocessing<T: ContentLoader + Unpin>(
...
+        Ok(metadata) => metadata,
+        Err(e) => {
+            if e == "offline" {
```
dignifiedquire, line 186: 😭

dignifiedquire (approved):

PR #634, opened 2022-12-21
fix(rpc): dial down log level for accept errors

rklaehn:
An accept error can happen e.g. when a client starts an RPC request and then immediately cancels it because it has raced it against a local op.

Not that we should do this that often, but it totally happens.

Example when reading a large directory that is present locally:

```
  2022-12-21T12:07:25.283756Z ERROR iroh_p2p::rpc: p2p rpc accept error: RecvError(A(NetworkError(hyper::Error(Body, Error { kind: Io(Kind(ConnectionReset)) }))))
    at iroh-p2p/src/rpc.rs:582

  2022-12-21T12:07:25.283775Z ERROR iroh_p2p::rpc: p2p rpc accept error: RecvError(A(NetworkError(hyper::Error(Body, Error { kind: Io(Kind(ConnectionReset)) }))))
    at iroh-p2p/src/rpc.rs:582
```

### DIFF
--- a/iroh-gateway/src/rpc.rs
+++ b/iroh-gateway/src/rpc.rs
@@ -60,7 +60,7 @@ pub async fn new(addr: GatewayAddr, gw: Gateway) -> Result<()> {
                 tokio::spawn(dispatch(server.clone(), req, chan, gw.clone()));
             }
             Err(cause) => {
-                tracing::error!("gateway rpc accept error: {}", cause);
+                tracing::debug!("gateway rpc accept error: {}", cause);
             }
         }
     }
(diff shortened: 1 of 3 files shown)


rklaehn:
Not quite happy with this in the long term. Ideally we would have a single place somewhere where we distinguish between "noteworthy" hyper errors and other hyper errors, and log them at different error level.

Unfortunately that is not that easy. We would have to first check the RpcServerErrors, and then look at the inner hyper errors as well. They have all kinds of predicates defined on hyper::Error:

https://docs.rs/hyper/latest/hyper/struct.Error.html

dignifiedquire (approved):

PR #649, opened 2022-12-23
fix: Remove limits checking in BitswapHandler

ppodolsky:
I was debugging the case when only two nodes are launched, iroh and kubo, the first one trying to download a lot of data from the second one. It just didn't worked and stucked at first blocks.

Seems that bitswap handler implementation has been copied from gossipsub and uses the same inbound substream limiting logic that is possibly not correct for BitSwap.

`inbound_substream` counter never decreases and hit limits very soon, then iroh disconnects from kubo, reconnects and goes into erratic state (worth to debug too).


ppodolsky:
@dignifiedquire is doing a more appropriate fix in https://github.com/n0-computer/iroh/pull/650

PR #654, opened 2022-12-28
Small set of fixes for refining logs

ppodolsky:
Removed prints and decreased logging level for several messages that seems to be normal, not warning


iroh-bitswap/src/lib.rs
```diff
@@ -302,7 +302,7 @@ impl<S: Store> Bitswap<S> {
         record!(BitswapMetrics::MessageBytesIn, message.encoded_len() as u64);
         // TODO: Handle backpressure properly
         if let Err(err) = self.incoming_messages.try_send((peer, message)) {
-            warn!(
+            debug!(
```
dignifiedquire, line 305: this should stay a `warn` if this happens it means we are under too much load.

dignifiedquire (approved):
looks good, just that one log statement

iroh-bitswap/src/lib.rs
ppodolsky: Done

dignifiedquire: > I see this message every time without doing anything special

yes because as soon as you announce you are using bitswap, the network hammers you :'(

ppodolsky:
Let it merge :)

dignifiedquire (approved):

PR #655, opened 2022-12-31
test: iroh-api 'get' bench

b5:
adds a criterion test for getting a `{1KiB, 1MiB, 10MiB }` file that pipes to a throwaway writer just to see what peak throughput for get looks like.

I'm not sure I'm doing this properly, would appreciate a thorough review to confirm I'm actually benchmarking the right thing. 

Results I get on my machine (bottom three tests):
Screen Shot 2022-12-30 at 11 36 50 PM


iroh-api/benches/get.rs
```diff
@@ -0,0 +1,122 @@
...
+        10 * 1024 * 1024, // 10 MiB (triggers chunking)
+    ]
+    .iter()
+    {
+        let value = vec![8u8; *file_size];
+        let dir = tempfile::tempdir().unwrap();
+        let path = dir.path().join(format!("{}.raw", file_size));
+        std::fs::write(&path, value).unwrap();
+
+        group.throughput(criterion::Throughput::Bytes(*file_size as u64));
```
flub, line 30: These `as u64` casts are a bit weird.  I'd probably do the cast on the `vec![]` call above with `file_size.try_into().unwrap()` and have the values u64 to start with.

iroh-api/benches/get.rs
```diff
@@ -0,0 +1,122 @@
...
+use tokio::runtime::Runtime;
+
+fn get_benchmark(c: &mut Criterion) {
+    let mut group = c.benchmark_group("unixfs_get_file");
+    for file_size in [
+        1024,             //  1 KiB
+        1024 * 1024,      //  1 MiB (triggers chunking)
+        10 * 1024 * 1024, // 10 MiB (triggers chunking)
+    ]
+    .iter()
```
flub, line 23: I don't think you get much value out of `.iter()`?

iroh-api/benches/get.rs
```diff
@@ -0,0 +1,122 @@
...
+    {
+        let value = vec![8u8; *file_size];
+        let dir = tempfile::tempdir().unwrap();
+        let path = dir.path().join(format!("{}.raw", file_size));
+        std::fs::write(&path, value).unwrap();
```
flub, line 28: I guess 10MiB isn't that much to put in memory... Though I'd at least explicitly drop `value` after this.

iroh-api/benches/get.rs
```diff
@@ -0,0 +1,122 @@
...
+                        let (cid, _) = prog.unwrap();
+                        cids.push(cid);
+                    }
+                    *cids.last().unwrap()
+                });
+
+                b.to_async(&executor).iter(|| {
+                    let api = Api::from_client_and_resolver(client.clone(), resolver.clone());
+                    let root = IpfsPath::from_cid(root);
+                    let mut dev_null = tokio::io::sink();
```
flub, line 94: You are timing the creation and dropping of the `Api` here and other here.  You probably want to use something like `.iter_batched()` with a setup function.  At the end of the `async move` block you then probably should return this all as a tuple so it gets dropped outside of the bench function.

iroh-api/benches/get.rs
dignifiedquire: in practice I always casted here, for throughput, as you want `usize` in all other cases, and `try_into().unwrap()` is just very gross..

iroh-api/benches/get.rs
```diff
@@ -0,0 +1,122 @@
...
+                                }
+                                OutType::Reader(mut reader) => {
+                                    tokio::io::copy(&mut reader, &mut dev_null).await.unwrap();
+                                }
+                                OutType::Symlink(_) => {
+                                    todo!("symlink support");
+                                }
+                            }
+                        }
+                        black_box(())
```
dignifiedquire, line 112: You should call this on `dev_null`, calling it on `()` does nothing.

dignifiedquire:
@b5 this is a useful bench to have, would be great if you could finish this up so we can merge it.

iroh-api/benches/get.rs
flub: Heh, I love `.try_into().unwrap()` because it panics instead of silently doing things that are almost certainly not intended.  Maybe I just need to learn the overflow rules properly sometime...

dignifiedquire: Well casting as `u64` from `usize` is always safe (for now) because Rust is only supported on 64bit and less systems.

flub: So, I'm only discussing this out of technical curiosity.  Not because I think it's very important or a blocking issue.

Having said that, I think if `<usize> as <u64>` would be "always safe" there would be a `From<usize>` on `u64`.

dignifiedquire: I suspect it is not there because that would mean adding a 128bit target would require a breaking change in the `From` api.

PR #669, opened 2023-01-03
feat(rpc, api): add `gossipsub` methods and expose the p2p `network_events` to the `Api`

### LINKED ISSUE #664: refactor: `iroh-share` uses `iroh-embed`
`iroh-share` currently imports the p2p and store directly, it should instead but using `iroh-embed` to spin itself up.

We also need to add gossipsub to the API, for use in `iroh-share`

ramfox:
closes n0-computer/beetle#16 

Gets `iroh-api` ready for use in `iroh-share` (via `iroh-embed`):
- `p2p.network_events` - exposes stream of `NetworkEvent`s
- `p2p.gossipsub_subscribe`
- `p2p.gossipsub_publish`
- `p2p.gossipsub_add_peers`
- `p2p.peer_id` - the local peer id
- `p2p.addrs` - the addresses of the local node
- `resolver` - exposes the resolver
- `store` - exposes the store methods


iroh-embed/src/p2p.rs
```diff
@@ -59,6 +90,16 @@ impl P2pService {
...
+    /// Returns a Receiver of NetworkEvents that have been emitted by this p2p service.
+    ///
+    /// This can be used to examine NetworkEvents that have occured
+    pub async fn network_events(&mut self) -> Receiver<NetworkEvent> {
+        let (s, r) = channel(512);
+        let mut events = self.events.lock().await;
+        events.push(s);
+        r
+    }
+
```
ramfox, line 102: @flub 
needed a way to expose the `NetworkEvents` `Receiver` to `iroh-share`.
This was one way I thought of doing it.

The other would be to add an rpc & api `network_events`method that would return a stream of `NetworkEvents`. 

After implementing the above, I'm leaning toward refactoring it to go the rpc/api route, but  was wondering if you had any opinions before I re-wrote it tomorrow.

flub: Channels are great, though the p2p `Node::run` (unsurprisingly) takes `&mut self` which means you have to get out the events channel out before invoking run which is rather meh.

Channels also mean you can only use the in-memory version of the p2p node.  That is of course the current idea of iroh-embed, though I think I agree using the irpc is nicer and more future proof (and internally they're just flume channels for iroh-embed so no real efficiency penalty).

So that's a verbose way of coming the the same conclusion, changing it to irpc would probably be nice :)

iroh-api/src/p2p.rs
```diff
@@ -56,6 +65,25 @@ impl P2p {
...
+
+    pub async fn subscribe(&self, topic: String) -> Result<bool> {
+        let topic = TopicHash::from_raw(topic);
+        self.client.gossipsub_subscribe(topic).await
+    }
+
+    pub async fn publish(&self, topic: String, data: Bytes) -> Result<MessageId> {
+        let topic = TopicHash::from_raw(topic);
+        self.client.gossipsub_publish(topic, data).await
+    }
```
ramfox, line 86: Thought I'd leave these as `subscribe`, and `publish` rather than `gossipsub_subscribe` and `gossipsub_publish`, to keep the API a little slicker looking. But is it too confusing/vague?

iroh-rpc-types/src/p2p.rs
```diff
@@ -144,6 +144,69 @@ pub struct LookupResponse {
...
+        #[serde(with = "TopicHashDef")]
+        topic: TopicHash,
+    },
+    Message {
+        from: PeerId,
+        id: MessageId,
+        #[serde(with = "GossipsubMessageDef")]
+        message: GossipsubMessage,
+    },
+}
```
ramfox, line 185: To keep moving quickly, I just moved the definition of `NetworkEvent` & `GossipsubMessage` into `iroh-p2p-types`, since these are now being used by the rpc client when sending over a `NetworkEvent` stream & would cause a circular dependency if these definitions were not moved.

We can keep this as is and just re-export those structs in `iroh-p2p` (what this PR does currently).

Or, the other way I was considering, was to write mirrored structs here, `NetworkEventRPC` `GossipsubMessageRPC`, and then import them into `iroh-p2p` & have the `iroh-p2p` rpc methods be responsible for translating `NetworkEvent` and `GossipsubMessage` structs into their `iroh-rpc-types` counterparts before sending the actual rpc responses.

But I'm not sure which is more "right" from a rust-style perspective.

flub: I think the current approach of moving them here is very reasonable.  The alternative you describe seems to only involve more conversions without much obvious benefit.

iroh-api/src/p2p.rs
flub: In any case I think every function here should have a looong doc comment describing what it does and providing some examples.  :)

I'd probably be tempted to leave the `gossipsub` part it, maybe there will be other things to subscribe to in the future?  If leaving it out there's only one canonical way to subscribe a p2p node to topics.  But that's speaking as someone who has no experience using these things.  I don't really know the domain well enough to make a call on this kind of API naming.

iroh-api/src/p2p.rs
```diff
@@ -56,9 +85,69 @@ impl P2p {
...
+            .await
+            .map_err(|e| map_service_error("p2p", e))
+    }
+
+    /// Publish a message on a Gossipsub Topic.
+    ///
+    /// This allows you to publish a message on a given topic to anyone in your network that is
+    /// subscribed to that topic.
+    ///
+    /// Read the [`gossipsub_subscribe`][`crate::p2p::P2p::gossipsub_subscribe`] documentation for how to subscribe and receive
```
ramfox, line 128: this is failing:
```
warning: public documentation for `gossipsub_publish` links to private item `crate::p2p::P2p::gossipsub_subscribe`
   --> iroh-api/src/p2p.rs:123:5
    |
123 | /     /// Publish a message on a Gossipsub Topic.
124 | |     ///
125 | |     /// This allows you to publish a message on a given topic to anyone in your network that is
126 | |     /// subscribed to that topic.
127 | |     ///
128 | |     /// Read the [`gossipsub_subscribe`][`crate::p2p::P2p::gossipsub_subscribe`] documentation for how to subscribe and receive
129 | |     /// Gossipsub messages.
    | |___________________________^
    |
    = note: `#[warn(rustdoc::private_intra_doc_links)]` on by default
    = note: the link appears in this line:

            Read the [`gossipsub_subscribe`][`crate::p2p::P2p::gossipsub_subscribe`] documentation for how to subscribe and receive
                                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: this link resolves only because you passed `--document-private-items`, but will break without
```
I'm guessing it's because the `p2p` module is technically private, even though the `P2p` struct itself is public?

Any thoughts on how to fix? 

Make the `p2p` mod public? Just silence the warning? Or is there a different way to reference this that makes it clear to rustdoc that it's fine? Using [`P2p::gossipsub_subscribe`] also triggered this warning.

flub: Oh no!  The mocks strike again!  The CI uses `--all-features` for this... which then breaks due to the mocks which seem to have different visibility then the real things.  Invoking `cargo doc` without `--all-features` makes it work as expected (and you can use the simpler ``[`P2p::gossipsub_subscribe`]`` form).

Unless I manage to get rid of the mocks before you're awake I suggest to just write it without linking for now to get around this: `` `P2p::gossipsub_subscribe` ``

iroh-api/src/p2p.rs
```diff
@@ -56,9 +85,69 @@ impl P2p {
             .await
             .map_err(|e| map_service_error("p2p", e))
     }
+
+    /// Get a stream of [`NetworkEvent`].
+    pub async fn network_events(&self) -> Result<BoxStream<'static, Result<NetworkEvent>>> {
```
flub, line 90: I'd be tempted to wrap this return type in a newtype for the API purposes.

iroh-api/src/p2p.rs
```diff
@@ -56,9 +85,69 @@ impl P2p {
...
+        Ok(stream.boxed())
+    }
+
+    /// Subscribe to a Gossipsub Topic
+    ///
+    /// Gossipsub is a pub/sub protocol. This will subscribe you to a Gossipsub
+    /// topic. All Gossipsub messages for topics that you are subscribed to can
+    /// be read off the the `NetworkEvent` stream, as
+    /// `NetworkEvent::Gossipsub(GossipsubEvent::Message)`. You can access this
+    /// stream using `P2p::network_events`.
```
flub, line 105: This makes me wonder if the API would be better if it had a separate stream for just the gossipsub topic you request.  We could even return that stream right here.

iroh-api/src/p2p.rs
```diff
@@ -56,9 +85,69 @@ impl P2p {
...
+    ///
+    /// There is currently no `gossipsub_unsubscribe` exposed to `iroh-api` crate
+    /// yet.
+    ///
+    /// Learn more about the Gossipsub protocol in the `libp2p-gossipsub`
+    /// [documentation](https://docs.rs/libp2p-gossipsub/latest/libp2p_gossipsub/).
+    //
+    // TODO(ramfox): write `gossipsub_messages(topic: String)` method that returns
+    // a stream of only the gossipsub messages on that topic
+    pub async fn gossipsub_subscribe(&self, topic: String) -> Result<bool> {
```
flub, line 115: What does the bool indicate?

iroh-api/src/p2p.rs
flub: Try merging master and see if this is fixed, I hope it is!

iroh-api/src/p2p.rs
```diff
@@ -56,9 +85,69 @@ impl P2p {
...
+    }
+
+    /// Publish a message on a Gossipsub Topic.
+    ///
+    /// This allows you to publish a message on a given topic to anyone in your network that is
+    /// subscribed to that topic.
+    ///
+    /// Read the [`gossipsub_subscribe`][`crate::p2p::P2p::gossipsub_subscribe`] documentation for how to subscribe and receive
+    /// Gossipsub messages.
+    pub async fn gossipsub_publish(&self, topic: String, data: Bytes) -> Result<MessageId> {
```
flub, line 130: API nitpicking: when dealing with publishing I wonder if a sink is a more natural API?  I can't really decide myself, probably not worth it for now.

iroh-api/src/store.rs
```diff
@@ -19,20 +20,43 @@ pub trait Store: 'static + Send + Sync + Clone {
     async fn put_many(&self, blocks: Vec<Block>) -> Result<()>;
 }
 
+#[derive(Debug, Clone)]
```
flub, line 23: I'd love doc comments on this struct and the pub functions :)

iroh-api/src/store.rs
```diff
@@ -19,20 +20,43 @@ pub trait Store: 'static + Send + Sync + Clone {
...
+    pub async fn put(&self, cid: Cid, blob: Bytes, links: Vec<Cid>) -> Result<()> {
+        self.client.put(cid, blob, links).await
+    }
+
+    pub async fn put_many(&self, blocks: Vec<Block>) -> Result<()> {
+        self.client
+            .put_many(blocks.into_iter().map(|x| x.into_parts()).collect())
+            .await
+    }
+}
```
flub, line 46: I'm a bit confused why we have both pub methods here as well as the trait?  I guess I'd be happy removing the trait?  I'm not sure what it gives us?

iroh-p2p/src/rpc.rs
```diff
@@ -394,6 +395,33 @@ impl P2p {
...
+    }
+
+    #[tracing::instrument(skip(self))]
+    async fn network_events_0(self) -> impl Stream<Item = Box<NetworkEventsResponse>> {
+        let (s, r) = oneshot::channel();
+        self.sender
+            .send(RpcMessage::NetworkEvents(s))
+            .await
+            // TODO(ramfox): refactor to return result
+            .expect("expect sender to exist");
```
flub, line 415: `s/sender/receiver/` probably.  Though I'm not sure if you really can use `.expect`.  You don't really have any guarantee the receiver is still alive.

iroh-p2p/src/rpc.rs
```diff
@@ -394,6 +395,33 @@ impl P2p {
...
+
+        let mut r = r.await.expect("expect sender and receiver to still exist");
```
flub, line 417: Here too `.expect()` is probably too harsh.  If the remote server closes the connection before there's a response, is the sender not dropped?

iroh-p2p/src/rpc.rs
```diff
@@ -394,6 +391,54 @@ impl P2p {
...
+    #[tracing::instrument(skip(self))]
+    fn gossipsub_subscribe(self, req: GossipsubSubscribeRequest) -> GossipsubEventStream {
+        async move {
+            self.gossipsub_subscribe_0(req)
+                .await
+                .expect("FIX THIS - NEEDS TO BE ABLE TO RETURN A RESULT")
+        }
+        .flatten_stream()
+        .boxed()
+    }
```
ramfox, line 403: @rklaehn I'm coming into conflict with this method in two ways.

For context, a `GossipsubEventStream` is `BoxStream<'static, Box<GossipsubSubscribeResponse>>`

First question/problem: I can't figure out how to handle the possible error from `gossipsub_subscribe_0` When I try to use `?` instead of `expect` here, I keep running into 
```
cannot use the `?` operator in an async block that returns `Pin<Box<dyn Stream<Item = Box<iroh_rpc_types::p2p::GossipsubSubscribeResponse>> + std::marker::Send>>`
    |                    this function should return `Result` or `Option` to accept `?
```
And when I looked up how to solve this, I didn't understand the syntax that the solutions suggested. Would love some help!

The second conflict I'm running into is with my "preferred" return type. My "dream" scenario is that the p2p RpcClient's `gossipsub_subscribe` can have a return type of:
`Result<impl Stream<Item=GossipsubEvent>>`

But, it seems like I'm forced to return a `Result<GossipsubEvent>` Item, based on the expectations of `server_streaming`. This isn't critical, but It would be nice to streamline the API in this way.

rklaehn: A server streaming interaction returns a stream, not a result of stream. So if you have a situation where you have a possible error on creation of the stream, but then no errors once the stream is created, you have no choice but to have the stream item be a `Result`. Otherwise the try_flatten_stream thing won't work.

So the result must be `impl Stream<Item=Result<GossipsubEvent>>`, or we must extend quic-rpc to support a result of stream. Since it is all just decorators and most of the critical work is in the transports, this should be possible. Just a bit of work, so the question is how strongly you feel about this.

Is it really the case that once you have subscribed there is nothing whatsoever that can go wrong? That is usually quite rare.

PR #671, opened 2023-01-03
refactor: Split configs in server and service

flub:
We used to use the same config for the service and server (aka the binary).  This is confusing when creating configs to use with e.g. iroh-one, iroh-embed or iroh-share because some fields are not used.

This splits off the config structs to avoid this problem, services now have a Config and binaries a ServerConfig.  This allows creating the services standalone without all the baggage a server needs.

While this isn't many fields yet, this will get worse as we add more features (this is split off from another PR where this seemed useful).

# Todo

- [x] iroh-store
- [x] iroh-p2p
- [x] iroh-gateway

### DIFF
--- a/iroh-gateway/src/config.rs
+++ b/iroh-gateway/src/config.rs
@@ -21,9 +21,17 @@ pub const CONFIG_FILE_NAME: &str = "gateway.config.toml";
 pub const ENV_PREFIX: &str = "IROH_GATEWAY";
 pub const DEFAULT_PORT: u16 = 9050;
 
-/// Configuration for [`iroh-gateway`].
+/// The configuration for the gateway server.
+///
+/// This is the configuration which the gateway server binary needs to run.  This is a
+/// superset from the configuration needed by the gateway service, which can also run
+/// integrated into another binary like iroh-one, iroh-share or iroh-embed.
+// TODO: I'd prefer to include [`Config`] under the `gateway` field liek iroh-one does.  But
+// that's a backwards incompatible change.
+// TODO: If this was moved into main.rs it would not need to be public.  Our binary is build
+// as being outside of the lib crate, hence here it needs to be public.
 #[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
-pub struct Config {
+pub struct ServerConfig {
     /// Pretty URL to redirect to
     #[serde(default = "String::new")]
     pub public_url_base: String,
@@ -55,7 +63,7 @@ pub struct Config {
     pub redirect_to_subdomain: bool,
 }
 
-impl Config {
+impl ServerConfig {
     pub fn new(port: u16, rpc_client: RpcClientConfig) -> Self {
         Self {
             public_url_base: String::new(),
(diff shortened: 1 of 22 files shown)


iroh-store/src/config.rs
```diff
@@ -23,15 +23,77 @@ pub fn config_data_path(arg_path: Option<PathBuf>) -> Result<PathBuf> {
...
-/// The configuration for the store.
+/// The configuration for the store server.
+///
+/// This is the configuration which the store server binary needs to run.  This is a
+/// superset from the configuration needed by the store service, which can also run
+/// integrated into another binary like in iroh-one, iroh-share and iroh-embed.
+///
+/// This config can be deserialised from a config file.
+// TODO: I'd prefer to include [`Config`] under the `store` field like iroh-one does.  But
+// that's a backwards incompatible change.
```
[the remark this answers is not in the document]
flub: For the p2p and gateway services this duplication becomes painful if not.

iroh-share/src/p2p_node.rs
```diff
@@ -173,10 +172,6 @@ impl P2pNode {
         let store_config = iroh_store::Config {
             path: db_path.to_path_buf(),
             rpc_client: rpc_store_client_config,
-            metrics: iroh_metrics::config::Config {
-                tracing: false, // disable tracing by default
```
flub, line 177: This is proof that the current state of things is weird.  This did nothing IIUC.

iroh-gateway/src/config.rs
```diff
@@ -21,9 +21,17 @@ pub const CONFIG_FILE_NAME: &str = "gateway.config.toml";
 pub const ENV_PREFIX: &str = "IROH_GATEWAY";
 pub const DEFAULT_PORT: u16 = 9050;
 
-/// Configuration for [`iroh-gateway`].
+/// The configuration for the gateway server.
+///
+/// This is the configuration which the gateway server binary needs to run.  This is a
+/// superset from the configuration needed by the gateway service, which can also run
+/// integrated into another binary like iroh-one, iroh-share or iroh-embed.
+// TODO: I'd prefer to include [`Config`] under the `gateway` field liek iroh-one does.  But
```
dignifiedquire, line 29: I think this should be done before merging, to avoid the duplication, don’t be afraid to break apis

Arqu: Yeah, feel free to break this, there's nobody really depending on it yet.

iroh-store/src/config.rs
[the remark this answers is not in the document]
Arqu: You should break it, there's no real dependency on them yet. Better break them now than later.

Arqu (dismissed):
From a prue code perspective this is 👍 
However this "works" for the metrics piece more by coincidence than design at this point.
The reason each had their own metrics directly planted was that different services used to run a different feature set of the metrics crate, which would initialize the metrics just for that service.
This kinda nukes that notion completely. But it was not doing that anyways since the `Cargo.toml` refactors because it's now from a global workspace and I believe it just builds all the features regardless. 

Functionally this is fine, it just wastes a little bit more resources on the metrics side. Not a ton though. I'll spend some time thinking about this.

flub:
> From a prue code perspective this is +1 However this "works" for the metrics piece more by coincidence than design at this point. The reason each had their own metrics directly planted was that different services used to run a different feature set of the metrics crate, which would initialize the metrics just for that service. This kinda nukes that notion completely. 

> But it was not doing that anyways since the `Cargo.toml` refactors because it's now from a global workspace and I believe it just builds all the features regardless.

Hum, this should not be the case.  If you look at `iroh-store/Cargo.toml` you'll find `iroh-metrics = { workspace = true, features = ["store"] }`.  So if you **only** build the iroh-store crate the only feature built for iroh-metrics is the `store` feature.  If you however also build other crates at the same time I believe the iroh-metrics crate will be built with the combination of all the features requested.  This behaves exactly the same as before the `workspace.dependencies` refactor.

> Functionally this is fine, it just wastes a little bit more resources on the metrics side. Not a ton though. I'll spend some time thinking about this.

dignifiedquire (approved):

PR #672, opened 2023-01-04
ref(store): Remove unused RPC Client

flub:
The store does not use the RPC client at all.  Remove it.

## yolo, let's see what breaks

CI doens't break!  It must be good.  Teach me why not :)

### DIFF
--- a/iroh-store/src/store.rs
+++ b/iroh-store/src/store.rs
@@ -9,7 +9,6 @@ use iroh_metrics::{
     inc, observe, record,
     store::{StoreHistograms, StoreMetrics},
 };
-use iroh_rpc_client::Client as RpcClient;
 use multihash::Multihash;
 use rocksdb::{
     BlockBasedOptions, Cache, ColumnFamily, DBPinnableSlice, Direction, IteratorMode, Options,
@@ -31,7 +30,6 @@ struct InnerStore {
     content: RocksDb,
     next_id: RwLock<u64>,
     _cache: Cache,
-    _rpc_client: RpcClient,
 }
 
 impl fmt::Debug for InnerStore {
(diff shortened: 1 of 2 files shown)


Arqu:
> The store does not use the RPC client at all. Remove it.
> ## yolo, let's see what breaks
> 
> CI doens't break! It must be good. Teach me why not :)

Is this used on the CLI to get status or whatever from the service? Ie the outside API?
Maybe I just have no clue what I'm talking about though...

flub:
> > The store does not use the RPC client at all. Remove it.
> > ## yolo, let's see what breaks
> > CI doens't break! It must be good. Teach me why not :)
> 
> Is this used on the CLI to get status or whatever from the service? Ie the outside API? Maybe I just have no clue what I'm talking about though...

I don't think so, that's done by the rpc server which is still started.

dignifiedquire:
It was done only in preparation, as the store will need this eventually, to make sure that the architecture is not built such that it assumes the store doesn't need the rpc client.

flub:
> It was done only in preparation, as the store will need this eventually, to make sure that the architecture is not built such that it assumes the store doesn't need the rpc client.

So can I convince people this doesn't buy us anything?  We have plenty of examples to build the RpcClient.  A client that's unused in a struct doesn't make us do anything we shouldn't: it is Send and Sync anyway and since it's unused it does not get passed around anywhere.  Once there's a use for it I don't think there's going to be any major hurdle to add it.

Personally I find it difficult to reason about things when the types don't match what's unused. The type system should help show what goes where.

iroh-store/src/store.rs
```diff
@@ -9,7 +9,6 @@ use iroh_metrics::{
     inc, observe, record,
     store::{StoreHistograms, StoreMetrics},
 };
-use iroh_rpc_client::Client as RpcClient;
```
dignifiedquire, line 12: you can probably remove the dep as well?

dignifiedquire:
> So can I convince people this doesn't buy us anything? 

you can, fine to remove for now imo

iroh-store/src/store.rs
flub: no, it is still used by the config for server in `main.rs`.

dignifiedquire (approved):

PR #673, opened 2023-01-04
feat: switch store from rocksdb to redb

dignifiedquire:
Opening this for discussion, not sure if we want to ever merge this. You can find a description of `redb` at https://github.com/cberner/redb.

Some pros & cons 
- Redb is written in Rust and performs quite well according to benchmarks. 
- Redb is much less mature than Rocksdb, and unclear how well it will scale in the TB+ scale.
- Redb is considerably simpler & easier to work with, and has proper transactions, allowing us to remove the hacky in memory locks
- Redb compiles a lot faster (binary size is smaller as well, but I haven't measured by how much)


flub (commented):
The code looks fine I guess.

Do we have any benchmarks for the database that can be compared?

Otherwise my main concern would the lack of storage backwards compatibility.  Though in the worse case we need to link two versions of the crate and manually move things over, that's pretty awkward to operate though.

dignifiedquire:
> Do we have any benchmarks for the database that can be compared?

We have some in the `iroh-store/benches`, which currently suggest ~20% reduced perf in some cases, but I am not very confident in them. Need to look at some other benchmarks as well

rklaehn:
Since redb does not guarantee any data format compat, at the very least we would have to check in cargo.lock and basically tell people to do a full blown import/export to/from car or whatever whenever the data format changes.

Does this require nightly? I get this error:

```
error[E0658]: use of unstable library feature 'map_first_last'
```

dignifiedquire:
I was running with stable, so just a pretty new one probably

rklaehn:
Code looks reasonable, and the build times are quite good. I have got to play around with the thing in isolation to say more.

Why does it have an (optional) dependency on a python interpreter?

https://github.com/cberner/redb/blob/master/Cargo.toml#L23

dignifiedquire:
That’s a library to build python apis, I supect because they expose a python api

iroh-store/src/store.rs
```diff
@@ -100,236 +62,114 @@ struct CodeAndId {
...
+                let txn = db.begin_write()?;
+                {
+                    txn.open_table(CF_BLOBS_V0)?;
+                }
+                {
+                    txn.open_table(CF_METADATA_V0)?;
+                }
+                {
+                    txn.open_table(CF_GRAPH_V0)?;
+                }
```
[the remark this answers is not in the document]
dignifiedquire: mainly to make sure everything is dropped, but in this case it is not needed I believe

dignifiedquire:
> Since redb does not guarantee any data format compat, at the very least we would have to check in cargo.lock and basically tell people to do a full blown import/export to/from car or whatever whenever the data format changes.


- We can just use a fixed version eg `=1.2.3`, no need for `Cargo.lock`.
- The author said they are quite close to a stable file format, but in any case, as long as there is no actual changes in the format, we can upgrade and test manually when a new version is released. And if the format changes we can provide a migration using import/export functionality if needed.

iroh-store/src/store.rs
```diff
@@ -39,46 +36,11 @@ impl fmt::Debug for InnerStore {
...
             .field("content", &self.content)
             .field("next_id", &self.next_id)
-            .field("_cache", &"rocksdb::db_options::Cache")
             .field("_rpc_client", &self._rpc_client)
             .finish()
     }
 }
 
-/// Creates the default rocksdb options
-fn default_options() -> (Options, Cache) {
```
rklaehn, line 49: Nice to get rid of all these cryptic/cargo cult options.

rklaehn:
Not saying that this is a showstopper in general, but on this branch the add performance is totally... bad.

For adding a large file I get about 40 mb/s, whereas with rocksdb I get more like 400 mb/s. 

image

dignifiedquire:
we should figure it if we can improve that, that is very bad :/

rklaehn:
It is due to the transactions committing immediately fully to disk. If you set the transaction mode to "eventual" you get back the expected performance.

As I said on discord, you can't treat a database system that does not have a WAL the same way as one that works with a WAL and expect similar performance.

For the ipfs add thing you would just do transactions with mode eventual and then one with mode immediate at the very end, that should take care of it.

iroh-store/src/store.rs
```diff
@@ -314,11 +303,9 @@ impl Store {
...
     }
 
-    #[tracing::instrument(skip(self))]
-    fn next_id(&self) -> u64 {
-        let mut id = self.inner.next_id.write().unwrap();
-        let next_id = *id;
-        *id = id.checked_add(1).expect("store is full");
+    #[tracing::instrument(skip(self, metadata_table))]
+    fn next_id(&self, metadata_table: &Table<u64, &[u8]>) -> u64 {
+        let next_id = metadata_table.iter().unwrap().next_back().map(|(id, _)| id.value()).unwrap_or_default() + 1;
```
dignifiedquire, line 308: - are you sure these are sorted? this is why I was using `max` when reading this out
- this seems potentially expensive, did you check that?

ramfox:

PR #674, opened 2023-01-05
test: Remove the mocks

flub:
This removes the use of mockall which is causing a lot of trouble with
the cargo features intricacies it brings with it.  A lot of the
tooling struggles with these different versions of the structs which
also behave differently in surprising ways.

The tests affected by this will be converted into end-to-end tests as part of
is now unused.  This is possibly only temprorary and that might come
back when the end-to-end tests are added.


dignifiedquire (approved):
can you file an issue to bring these tests back in some form please?

dignifiedquire:
nvm, I see you already did 😅

PR #677, opened 2023-01-05
feat: introduce iroh-memstore crate

flub:
This adds a new iroh-memstore crate which provides an in-memory version of the store.

The current implementation is extremely rough and ugly.  I'd like to improve things but let's get something written down first.

Probably works?

## Early feedback

Things I'm interested about hearing:

- A new crate?  I considered adding to the existing iroh-store but decided not to.  It seems unlikely you'd want to run the `iroh-store` binary with this?  So far I like my choice here but happy to have dissenting opinions.
- Single threaded: I'm not making any attempt to handle requests concurrently.  And I wasn't planning on improving this later really.  That seems reasonable for something in-memory, any disagreements?

Of course also anything else you'd like to have an opinion on.


iroh-memstore/src/lib.rs
```diff
@@ -0,0 +1,210 @@
...
+}
+
+#[derive(Debug, Default, Clone)]
+struct MemStore {
+    inner: Arc<Mutex<InnerMemStore>>,
+}
+
+#[derive(Debug, Default)]
+struct InnerMemStore {
+    data: AHashMap<Cid, Bytes>,
```
rklaehn, line 75: So no id/cid mapping? My hunch is that this is worth it even for the mem store, but it can wait for later.

rklaehn (commented):
Looks pretty straightforward. How do you enable it?

flub:
> Looks pretty straightforward. How do you enable it?

That would be via iroh-embed only I think (and transitively iroh-share), there is already a `RocksStore` and I guess it would become a `MemStore` there.

I don't think there's much value in putting this in the `iroh-store` binary?  Or would this be useful in general?  The main motivation I can think of is for testing, but it is already possible to control where the store goes with some environment variables or config files.

dignifiedquire:
- Yes, new crate is good (at least for now)
- I think it would be cool if I could run this as a standalone binary, like `iroh-memstore`, but not entirely sure how much that is actually needed.

iroh-memstore/src/lib.rs
rklaehn: In the normal store we use u64 ids all over the place. Cids are only used in one place, to look up the corresponding id. The reason for that is that cids are quite large and unwieldy, whereas u64s are, well u64s.

For sure this is not needed in the mem store at this time, but eventually it will be. Cids in memory are something like 100 bytes.

iroh-memstore/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
+    async fn handle_get_links(
+        &mut self,
+        request: GetLinksRequest,
+        mut sink: StoreSink,
+        _source: StoreSource,
+    ) -> Result<()> {
+        let block = self.blocks.get(&request.cid);
+        let links = block.map(|b| b.links.clone());
+
+        let response = Ok(GetLinksResponse { links });
```
rklaehn, line 229: So what's to stop you from sending a completely inadequate response here? Or send multiple responses in a RPC pattern?

This basically completely ignores all the safety guarantees that the RpcServer gives you. E.g. you redefine the response type of a request, before it would have given you a compile error. Now it gives you nothing.

It might not come as a surprise to you, but I liked the version with the RpcServer better.

iroh-memstore/src/lib.rs
flub: I can only see a space saving if links can also be an id instead of full Cid.  But that means all links must exist in the store and seems a bit unlikely.  Or am I still getting this wrong?

iroh-memstore/src/lib.rs
flub: certainly valid criticism.  i need to think about that.

rklaehn: I am not entirely happy with the complexity of the stuff in the giant switch statement. But I also don't like the alternative of having that switch statement generated by a macro. And it does serve a purpose. The stuff in iroh-rpc-types is a protocol defining the return type and interaction pattern for each message. Do a change there, and the compiler will scream at you (albeit with occasionally cryptic errors) until you have updated all client *and* server usages to reflect the change.

ramfox:

PR #678, opened 2023-01-09
fix: Embed CSS files into templates

ppodolsky:
There are two changes:
- Small bugfix for the bug I did in gateway code. Redirecting to index.html does not require trailing slash
- Accessing files thought subdomain paths makes impossible to use paths like /style.css for loading resources. I mitigated it by embedding CSS into templates. Original files are left here intentionally for easier development in the future.


rklaehn (commented):
So this removes the assets from ../assets/style.css and ../assets/icons.css and includes them in the dir_list and 404 page.

Not sure if this is an improvement. It simplifies the routes for sure, but now you have to "recompile" the css and icons to html whenever you want to do a change.

ppodolsky:
> So this removes the assets from ../assets/style.css and ../assets/icons.css and includes them in the dir_list and 404 page.
> 
> Not sure if this is an improvement. It simplifies the routes for sure, but now you have to "recompile" the css and icons to html whenever you want to do a change.

It is not an improvement but a required fix. If you are requesting file through subdomain handler by url `https://bafy<...>.ipfs.localhost:8080/style.css`, request is routed to bafy<...>/style.css, not to custom handler in gateway.

dignifiedquire:
How does Kubo do this currently?

ppodolsky:
Kubo embeds all assets.
I though about putting assets to a separate subdomain but it would lead to all kinds of CORS issues on loading.

fabricedesre:
> Kubo embeds all assets. I thought about putting assets to a separate subdomain but it would lead to all kinds of CORS issues on loading.

In general css & images loads are not subject to SOP restrictions so you should not need CORS. But it should also be possible to serve them from roo_url/static/xxx and set the CORS headers on all of /static/

ppodolsky:
You mean second-level domain by root_url? We need to embed JS snippet for detecting root url then, is it ok?

fabricedesre:
I think the base url is part of the config, so you could use it to generate the css url without needing a script.

ppodolsky:
@fabricedesre Thanks for idea, done. However there is still small regression: by default, public_url_base is empty and users will have to set it manually for development environment

ppodolsky:
I've also fixed a bug here with `index.html`.
Before, it merely pushed index.html to the end of path, without reloading of `path_metadata` and other attributes. It led to inconsistency and erratic behaviour. Now `path_metadata` is reloaded.

ppodolsky:
Let's review and merge?

PR #679, opened 2023-01-09
feat: Adding raw blocks as leafs in `DirectoryBuilder`

ppodolsky:
In my case I need to manually craft directories, namely adding pre-existent nodes as children for new directory. Here are changes to support adding blocks in `DirectoryBuilder`


iroh-unixfs/src/builder.rs
```diff
@@ -559,6 +598,10 @@ impl DirectoryBuilder {
 
         Ok(if let Some(path) = path {
             let mut dir = make_dir_from_path(path, chunker.clone(), degree).await?;
+            match &mut dir {
+                Directory::Basic(basic) => basic.entries.extend(entries),
+                Directory::Hamt(_) => unimplemented!(),
```
dignifiedquire, line 603: `Hamt`s are implemented, so this shouldn't be the case

ppodolsky: Ok, I need to figure out how to add entries to existent HamtNode. Probably worth to rebuild it entirely

rklaehn: The way this builder works is a bit weird. If path is set, entries are ignored. So maybe you can just put in an assertion that either path is None or entries is empty and be done with it?

I tried to refactor all of this at some point, but was not entirely happy with the results.

ppodolsky: Afaik there are no stoppers for working both, `path` and `entries`. It is just implemented this way for some reasons. I may try to refactor this part

ppodolsky: @rklaehn Done with refactoring. I removed path variable in builder and added add_path method that walks over path, converts elements to `Entry` and adds them to `self.entries`. It makes API more consistent and allows to add multiple paths.

ppodolsky:
I have refactored code here as you suggested @rklaehn. Take a look plz when you'll have a time

ramfox:

PR #684, opened 2023-01-11
feat: Add support for an in-memory store to iroh-embed

flub:
This allows using the in-memory store provided by the iroh-memstore crate.

Possibly the stores should be behind features, but I don't think this is worth it yet.

Note this also doesn't convert the `P2pService` to use the `IrohService` trait.  It could, I'm a bit undecided on whether to do so or not.  So please let me know your preferences.

Also note that I didn't implement `Drop` for the `MemStoreService`.  I'm now regretting doing this originally.  Tokio tasks also keep running if you drop the handle.

Finally I'm still on the fence about the naming of the `*Service::new` methods, I much prefer the naming in `iroh-memstore` which uses `::spawn` instead of `::new` to signal this is spawning a task.  I'd be happy to rename those, but possibly as a separate API.

----

This PR depends on the `flub/memstore` branch and targets it.  That PR is only in draft, though I think there's nothing wrong with having this PR being approved before.  I'll just sit around a bit before being merged.

### DIFF
--- a/iroh-embed/src/lib.rs
+++ b/iroh-embed/src/lib.rs
@@ -29,8 +29,6 @@
 //!
 //! An example is available in the repository under `examples/embed`.
 
-use std::fmt::{self, Debug};
-
 use anyhow::{bail, Result};
 use async_trait::async_trait;
 use iroh_rpc_client::Config as RpcClientConfig;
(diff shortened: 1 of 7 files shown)


iroh-embed/src/lib.rs
```diff
@@ -86,14 +92,50 @@ pub use store::RocksStoreService;
...
+///                     .unwrap();
+/// # })
+/// ```
+// Note: The doc tests above are disabled because they run extremely slow.  However the same
+// tests run just fine as normal tests.  We should really figure out what is going on
+// sometime.
+
 pub struct IrohBuilder {
-    store: Option<RocksStoreService>,
+    store: Option<Box<dyn IrohService<StoreService>>>,
```
dignifiedquire, line 122: Why `Box` over generic?

flub: Hum, because I never thought of a builder as a generic struct and imagined someone calling repeatedly `.store()` with different stores...  But I guess that's a bit of a useless usecase.  Generic would get rid of the weird `fn stop(self: Box<Self>)` thing I had to do so that'd be a plus.  I'll change it and see what it looks like.

flub:
PTAL

PR #697, opened 2023-02-01
fix: Do not stop if there is no p2p

ppodolsky:


rklaehn (approved):
This seems pretty straightforward

PR #777, opened 2023-02-21
feat: Make get work on IPv6 network

flub:


src/get.rs
```diff
@@ -53,7 +53,11 @@ async fn setup(opts: Options) -> Result<quinn::Connection> {
 
     let tls_client_config = tls::make_client_config(&keypair, opts.peer_id, opts.keylog)?;
     let mut client_config = quinn::ClientConfig::new(Arc::new(tls_client_config));
-    let mut endpoint = quinn::Endpoint::client("0.0.0.0:0".parse().unwrap())?;
+    let bind_addr = match opts.addr.is_ipv6() {
```
dignifiedquire, line 56: shouldn't we do this on the provide side as well?

dignifiedquire: Actually nvm, this is only needed here.

dignifiedquire (approved):
lgtm, lets wait for @b5 to confirm this works for him

src/get.rs
flub: the provider or server endpoint already binds to whatever address you give on the commandline, so we already do.  it's not like we need to call a different api for quinn, it's just that we need to have the socket bound to the same family as the server endpoint.

PR #778, opened 2023-02-21
ref: Improve safety of BytesMut buffer usage

flub:
The implementation uses a BytesMut buffer to efficiently read protocol
messages from the network connections.  The way read_lp was combining
both reading the length-prefixed message and decoding it resulted in
the caller having to remember to advance the BytesMut buffer before
using it again.  That is error prone.

This reduces the function of the read_lp() function to only read the
raw message from the network.  It returns a new Bytes object and the
passed in BytesMut is in a safe state with all internal indexes
correctly advanced.

The caller is now responsible for calling postcard::from_bytes using
the returned Bytes, read_lp() can not both create the Bytes and return
a parsed struct from it due to lifetimes.  Overall this isn't any more
complicated to use and results in less error-prone code.

### DIFF
--- a/src/protocol.rs
+++ b/src/protocol.rs
@@ -4,7 +4,7 @@ use std::io;
 use std::str::FromStr;
 
 use abao::decode::AsyncSliceDecoder;
-use anyhow::{ensure, Result};
+use anyhow::{ensure, Context, Result};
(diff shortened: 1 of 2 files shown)


rklaehn (approved):
Looks reasonable.

Issue #720, opened 2023-02-06
feat: daemonize `sendme provide` to allow for adding multiple collections


ramfox:
## Sendme as a long-running process

We want to be able to add additional collections to `sendme`. To do this, we need to adjust `sendme` to be a long running process that can communicate with a CLI. All the data should still be ephemeral.

We need some sensible RPC process (something that can be doable in a week) to communicate between the process and the CLI.

### Commands

1) `sendme start`

- starts the server
- takes over the terminal
- output address & peer ID
- allow options for:
    - providing a specific port
    - providing a path to an existing private key
- (in the future we can add an option that will immediately provide a collection, but for now, let’s just keep it as two separate commands)

2) `sendme provide`

- same functionality as the current `sendme provide`, where you pass a path to a file or directory
- allow option for providing a given auth token (like in the current `sendme provide`)
- prints out progress in the terminal that made the request (can re-use current terminal output)
    - prints out files w/ file sizes
    - prints out hash of collection
    - prints out “all in one” ticket
- (In the code, we currently allow for giving the `Collection` a name, but let’s ignore that for now & just keep the `Collection` name as “collection” (which is how it is currently implemented))
- error if `sendme` is not running

3) `sendme list`

- print out all collection hashes in the db
- print out files w/ file sizes under each collection hash
- error if `sendme` is not running

4) `sendme` shuts down via `Ctrl/Cmd-c`

PR #779, opened 2023-02-21
Daemonize iroh

rklaehn:
Adds a `ProviderService` with methods list, provide and version / watch for monitoring the service state.

Implements #720


ramfox:
Not sure what I'm going wrong here, but my processes won't connect when I run `iroh start` and then `iroh add [PATH]` in another terminal.

Sometimes `iroh add` hangs and sometimes it errors with `Error: iroh server is not running`.

Looks like they are expecting the same RPC ports, so I'm not sure what the issue is.

rklaehn:
@ramfox thanks for trying it out.

src/provider.rs
```diff
@@ -34,30 +40,49 @@ use crate::blobs::{Blob, Collection};
...
 
 const MAX_CONNECTIONS: u32 = 1024;
 const MAX_STREAMS: u64 = 10;
+const HEALTH_POLL_WAIT: Duration = Duration::from_secs(1);
 
 /// Database containing content-addressed data (blobs or collections).
-#[derive(Debug, Clone)]
-pub struct Database(Arc<HashMap<Hash, BlobOrCollection>>);
+#[derive(Debug, Clone, Default)]
+pub struct Database(Arc<Mutex<HashMap<Hash, BlobOrCollection>>>);
```
flub, line 56: I'm having a deja-vu.  I'm sure I once wondered if this could be an `RwLock` but I can't remember what the conclusion of that conversation was.

src/provider.rs
```diff
@@ -34,30 +40,49 @@ use crate::blobs::{Blob, Collection};
...
+            .lock()
+            .unwrap()
             .iter()
             .filter_map(|(k, v)| match v {
                 BlobOrCollection::Blob(data) => Some((k, data)),
                 BlobOrCollection::Collection(_) => None,
             })
-            .map(|(k, data)| (k, &data.path, data.size))
+            .map(|(k, data)| (*k, data.path.clone(), data.size))
+            .collect::<Vec<_>>();
```
flub, line 82: This`.collect()` is pretty unfortunate, but I guess the lock forces our hand.

src/provider.rs
```diff
@@ -313,6 +375,81 @@ impl Future for Provider {
...
+impl Handler {
+    fn list(self, _msg: ListRequest) -> impl Stream<Item = ListResponse> + Send + 'static {
+        let items = self
+            .0
+            .blobs()
+            .into_iter()
+            .map(|(hash, path, size)| ListResponse { hash, path, size });
+        futures::stream::iter(items)
+    }
+    async fn provide(self, msg: ProvideRequest) -> anyhow::Result<ProvideResponse> {
```
flub, line 389: Maybe call this the same as on the cli where it's called `add` (and the request type).  I think that's more cohesive.

src/provider.rs
```diff
@@ -313,6 +375,81 @@ impl Future for Provider {
...
+        self.0.union_with(db);
+        Ok(ProvideResponse { hash })
+    }
+    async fn version(self, _: VersionRequest) -> VersionResponse {
+        VersionResponse {
+            version: env!("CARGO_PKG_VERSION").to_string(),
+        }
+    }
+    fn watch(self, _: WatchRequest) -> impl Stream<Item = WatchResponse> {
+        futures::stream::unfold((), |()| async move {
```
flub, line 417: til, nice function

flub (commented):
Generally looks fine I think.  Two high level things though:

- I don't think the `start` subcommand adds much over the `provide` subcommand.  I think we could do this inside `provide` and give it an new optional flag to enable rpc.
- The `RPC_PORT` should be configurable, probably with another command line flag.
- For integration tests we'll want to configure `RPC_PORT = 0` so we don't have to deal with conflicting ports.  We'll then have to find out where it is listening, so I guess we should print that as well together with the other bits that are being printed on startup.

(I have an incredible deja-vu in writing some of this, apologies if I missed earlier responses or reasoning)

src/provider.rs
rklaehn: RwLock would be fine. I just did the simplest thing possible here. Even used the stdlib mutex instead of my usual parking_lot.

src/provider.rs
rklaehn: If you don't want to .collect you would have to use an immutable collection for the db that has zero cost snapshots. Like https://crates.io/crates/im . But I think we can keep it like this.

In the long term this will have to be a database anyway, maybe redb...

src/provider.rs
rklaehn: Upgraded this to a RwLock

src/main.rs
```diff
@@ -348,6 +434,46 @@ async fn provide_interactive(
...
+    if let Some(ref encoded) = auth_token {
+        let auth_token = AuthToken::from_str(encoded)?;
+        builder = builder.auth_token(auth_token);
+    }
+    let provider = builder.spawn()?;
+
+    println!("Starting iroh process");
+    println!("PeerID:      {}", provider.peer_id());
+    println!("Auth token:  {}", provider.auth_token());
+    println!("Listen addr: {}", provider.listen_addr());
```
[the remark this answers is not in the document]
rklaehn: Why? It seems to do the exact same thing. Except that it is much more boilerplatey because of the format! and async.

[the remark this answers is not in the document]
rklaehn: Ah, it writes to stderr...

rklaehn:
@b5 I think from a technical POV this is ready to go. We just need to settle on an API. Currently it is

start - starts the server
add - connect to running server and add a path
provide - the old provide (for backwards compat, mostly for the tests by @Arqu )

get / get-ticket etc. is unchanged

---

How about this: we have an option to provide an initial path for start. Then start is functionally equivalent to provide and can be a drop in for @Arqu 's tests. It is also useful then for people that just want to immediately start sharing something.

`add` we can then rename to `provide`. Or not?

src/main.rs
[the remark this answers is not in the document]
flub: No idea, all I know is it's there and was introduced in https://github.com/n0-computer/iroh/commit/55ff6ed6aa02ff35ef6133ce484bb1e7e48ee901  The commit message doesn't say much.  It was changed in https://github.com/n0-computer/iroh/commit/ea94e982516f816650582d95e1a6a3fdf6f0319d which also doesn't say much.

ramfox:
> How about this: we have an option to provide an initial path for start. Then start is functionally equivalent to provide and can be a drop in for @Arqu 's tests. It is also useful then for people that just want to immediately start sharing something.
> 
> `add` we can then rename to `provide`. Or not?

This sounds in line with the original issue with the addition of allowing `start` to maintain the original `provide` flow. So those changes sound 🔥  to me!

flub:
> `add` we can then rename to `provide`. Or not?

I think this is going a bit far.  Our lib-facing APIs already use `provide` for what the `iroh provide` command does today.  Changing this now is going to be a bit painful.

b5:
Yeah at this point I'd say rough in whatever y'all think works best, and I'll take a pass for "product" in a follow up.

src/get.rs
```diff
@@ -45,13 +45,18 @@ impl Default for Options {
...
 }
 
-/// Setup a QUIC connection to the provided address.
-async fn setup(opts: Options) -> Result<quinn::Connection> {
+/// Create a quinn client endpoint
+pub fn make_client_endpoint(
+    peer_id: Option<PeerId>,
+    alpn_protocols: Vec<Vec<u8>>,
+    keylog: bool,
+    ipv6: bool,
```
flub, line 53: nitpicking/bikeshedding: I just noticed you made this `pub` which means I care a little more about the API.  And one boolean argument in a signature makes me squint.  Two makes me hesitate.

One alternative would be to keep accepting the SocketAddr to be connected to and only use it to determine the address family.  Could be a tad confusing though.

Another would be to introduce a custom `enum AddressFamily{IpV4, IpV6}`.

Another would be basically going the builder way is some form, e.g. `IrohEndpointBuilder::new().with_peer_id(peer_id).with_alpn(vec!["a", "b"]).with_keylog(false).with_ipv4().build()`

src/provider.rs
```diff
@@ -908,3 +1055,23 @@ mod tests {
...
     }
 }
+
+/// Create a [`quinn::ServerConfig`] with the given keypair and limits.
+pub fn make_server_config(
+    keypair: &Keypair,
+    max_streams: u64,
+    max_connections: u32,
+    alpn_protocols: Vec<Vec<u8>>,
+) -> anyhow::Result<quinn::ServerConfig> {
```
flub, line 1065: I'm wondering whether to ponder a builder here too.  But I'm not sure.

flub:
> @flub since you are at this point the main consumer of the library: leave it as it is? Or keep `start` and `add` but still add the option to provide immediately on start?

Library-wise I don't think you're changing anything?  With "leave it as is" you mean as currently in the PR?

I kind of liked `provide` as starting the server-side (exactly like it does today) and `add` to add extra files to the provider.  This matches the internal module naming as well with `iroh::provider` and `iroh::get`.  I guess long-term `iroh provide` would change once it morphs into something more p2p but predicting the future is futile I think.  Naming is already hard enough.

I don't see much of a motivation to change the `iroh::provider::Provider` and `iroh::get` APIs because we're introducing this feature.  Although I do also see the logic you where going for with using `iroh provide some_file` to add another file to the database, I think `iroh start` is very generic and `iroh provide [some_file]` + `iroh add some_file` seems adequate right now.

src/get.rs
rklaehn: I will just pass in the bind addr then.

src/provider.rs
rklaehn: It's a pretty straightforward fn. Just because it has 2 numeric args?

src/get.rs
```diff
@@ -45,13 +45,18 @@ impl Default for Options {
...
+) -> Result<quinn::Endpoint> {
     let keypair = Keypair::generate();
 
-    let tls_client_config = tls::make_client_config(&keypair, opts.peer_id, opts.keylog)?;
+    let tls_client_config = tls::make_client_config(&keypair, peer_id, alpn_protocols, keylog)?;
     let mut client_config = quinn::ClientConfig::new(Arc::new(tls_client_config));
-    let bind_addr = match opts.addr.is_ipv6() {
+    let bind_addr = match ipv6 {
         true => "[::]:0".parse().unwrap(),
         false => "0.0.0.0:0".parse().unwrap(),
```
dignifiedquire, line 61: no need to parse them, there is `IpV{4|6}Addr::UNSPECIFIED` constants

src/get.rs
```diff
@@ -45,13 +45,18 @@ impl Default for Options {
     }
 }
 
-/// Setup a QUIC connection to the provided address.
-async fn setup(opts: Options) -> Result<quinn::Connection> {
+/// Create a quinn client endpoint
+pub fn make_client_endpoint(
+    peer_id: Option<PeerId>,
+    alpn_protocols: Vec<Vec<u8>>,
```
dignifiedquire, line 51: could be `&[&'static u8]]` probably

src/main.rs
```diff
@@ -41,6 +53,25 @@ enum Commands {
         /// If this path is provided and it exists, the private key is read from this file and used, if it does not exist the private key will be persisted to this location.
         #[clap(long)]
         key: Option<PathBuf>,
+        /// Optional rpc port, defaults to 4919
+        #[clap(long)]
+        rpc_port: Option<u16>,
```
dignifiedquire, line 58: should be a "global" option, like `keylog`?

src/main.rs
```diff
@@ -41,6 +53,25 @@ enum Commands {
...
+    },
+    /// List hashes
+    #[clap(about = "List hashes")]
```
dignifiedquire, line 61: I belive the `about` is not needed, it will be parsed by default from the doc comment

src/main.rs
```diff
@@ -41,6 +53,25 @@ enum Commands {
         /// If this path is provided and it exists, the private key is read from this file and used, if it does not exist the private key will be persisted to this location.
         #[clap(long)]
         key: Option<PathBuf>,
+        /// Optional rpc port, defaults to 4919
+        #[clap(long)]
+        rpc_port: Option<u16>,
+    },
+    /// List hashes
```
dignifiedquire, line 60: Is this listing hashes or collections?

src/main.rs
```diff
@@ -41,6 +53,25 @@ enum Commands {
...
+        rpc_port: Option<u16>,
+    },
+    /// Add some data to the database.
+    #[clap(about = "Add data from the given path")]
+    Add {
+        /// The path to the file or folder to add.
+        path: PathBuf,
+        /// Optional rpc port, defaults to 4919
+        #[clap(long)]
+        rpc_port: Option<u16>,
```
dignifiedquire, line 74: if you use `default_value_t` to specify the default it should automatically document it

src/main.rs
```diff
@@ -177,11 +208,36 @@ impl FromStr for Blake3Cid {
     }
 }
 
+async fn make_rpc_client(
+    rpc_port: Option<u16>,
+) -> anyhow::Result<RpcClient<ProviderService, QuinnConnection<ProviderResponse, ProviderRequest>>>
+{
+    let rpc_port = rpc_port.unwrap_or(RPC_PORT);
+    let endpoint = iroh::get::make_client_endpoint(None, vec![RPC_ALPN.to_vec()], false, false)?;
+    let addr: SocketAddr = SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::LOCALHOST, rpc_port));
```
dignifiedquire, line 217: a little simpler

```rust
 let addr = SocketAddr::new(Ipv4Addr::LOCALHOST.into(), rpc_port);
```

src/main.rs
```diff
@@ -177,11 +208,36 @@ impl FromStr for Blake3Cid {
...
     "{msg}\n{spinner:.green} [{elapsed_precise}] [{wide_bar:.cyan/blue}] {bytes}/{total_bytes} ({eta})";
 
-#[tokio::main(flavor = "multi_thread")]
-async fn main() -> Result<()> {
+fn main() -> Result<()> {
+    let rt = tokio::runtime::Builder::new_multi_thread()
+        .enable_all()
+        .build()?;
+    rt.block_on(main_impl())?;
+    rt.shutdown_timeout(Duration::from_millis(500));
```
dignifiedquire, line 236: why is this construction needed?

src/main.rs
```diff
@@ -246,70 +302,95 @@ async fn main() -> Result<()> {
...
+                // println!("Collection: {}\n", Blake3Cid::new(hash));
+                // let mut total_size = 0;
+                // for (_, path, size) in db.blobs() {
+                //     total_size += size;
+                //     println!("- {}: {}", path.display(), HumanBytes(size));
+                // }
+                // println!("Total: {}", HumanBytes(total_size));
+                // println!();
+                // return all the things that need to live even after this task
+                // println!("All-in-one ticket: {}", provider.ticket(hash));
```
dignifiedquire, line 341: what about all these commented out lines?

src/main.rs
```diff
@@ -246,70 +302,95 @@ async fn main() -> Result<()> {
...
+                anyhow::Ok(tmp_path)
+            });
+
+            tokio::signal::ctrl_c().await?;
+            println!("Shutting down provider...");
+            provider.shutdown();
+            provider.await?;
+            fut.abort();
+            drop(fut);
```
dignifiedquire, line 350: why is this needed?

src/main.rs
rklaehn: It only applies to some ops, so no. I guess there might be some clap way to DRY this, but I think it is fine to keep it as a separate option.

src/get.rs
rklaehn: You would have to turn it into a Vec<Vec<u8>> to actually set it because rustls::ClientConfig takes a Vec<Vec<u8>>

src/main.rs
rklaehn: All the things have `about =`.

src/main.rs
rklaehn: Added a comment. Without this tokio::io::copy(stdin(), ...) hangs forever unless you press enter...

src/main.rs
rklaehn: the future holds the temp file. When it gets dropped, the temp file gets deleted. Goddamn temp file... Added a comment.

src/main.rs
```diff
@@ -1,18 +1,30 @@
...
+    ProviderService, VersionRequest,
+};
+use quic_rpc::transport::quinn::{QuinnConnection, QuinnServerEndpoint};
+use quic_rpc::{RpcClient, ServiceEndpoint};
 use tokio::io::AsyncWriteExt;
 use tracing_subscriber::{prelude::*, EnvFilter};
 
 use iroh::{get, provider, Hash, Keypair, PeerId};
 
+const RPC_PORT: u16 = 0x1337;
```
dignifiedquire, line 25: Maybe `DEFAULT_RPC_PORT`?

src/main.rs
```diff
@@ -177,11 +208,44 @@ impl FromStr for Blake3Cid {
     }
 }
 
+async fn make_rpc_client(
+    rpc_port: u16,
+) -> anyhow::Result<RpcClient<ProviderService, QuinnConnection<ProviderResponse, ProviderRequest>>>
+{
+    let endpoint = iroh::get::make_client_endpoint(None, vec![RPC_ALPN.to_vec()], false, false)?;
+    let addr = SocketAddr::new(Ipv4Addr::LOCALHOST.into(), rpc_port);
+    let server_name = "localhost".to_string();
```
dignifiedquire, line 217: Does this name have any actual meaning?

src/main.rs
```diff
@@ -177,11 +208,44 @@ impl FromStr for Blake3Cid {
...
+}
+
 const PROGRESS_STYLE: &str =
     "{msg}\n{spinner:.green} [{elapsed_precise}] [{wide_bar:.cyan/blue}] {bytes}/{total_bytes} ({eta})";
 
-#[tokio::main(flavor = "multi_thread")]
-async fn main() -> Result<()> {
+async fn close<C>(client: RpcClient<ProviderService, C>) -> anyhow::Result<()> {
+    drop(client);
+    tokio::time::sleep(Duration::from_millis(100)).await;
```
dignifiedquire, line 232: Could you add a comment what exactly is going on, why do we need to randomly wait here?

src/main.rs
```diff
@@ -246,70 +310,98 @@ async fn main() -> Result<()> {
             addr,
             auth_token,
             key,
+            rpc_port,
         } => {
-            tokio::select! {
```
dignifiedquire, line 250: why can't we select biased over this anymore, with just the ctrl_c having priority?

src/provider.rs
dignifiedquire: Came here for the same issue, it's pretty odd to read at the call site, there are just two random number constants, with no clarity what they mean. Could either make this a builder (or just a simple options struct) or at least name the numbers at the call sites as constants.

src/util.rs
```diff
@@ -118,6 +119,26 @@ impl MaxSize for Hash {
     const POSTCARD_MAX_SIZE: usize = 32;
 }
 
+/// A serializable error type for use in RPC responses.
+#[derive(Serialize, Deserialize, Debug)]
+pub struct RpcError(serde_error::Error);
+
+impl std::error::Error for RpcError {}
```
dignifiedquire, line 126: Pretty sure we already depend on `thiserror`, so that seems a little cleaner to use

src/provider.rs
```diff
@@ -171,14 +229,43 @@ impl Builder {
         auth_token: AuthToken,
         events: broadcast::Sender<Event>,
         cancel_token: CancellationToken,
+        rpc: E,
+        internal_rpc: impl ServiceEndpoint<ProviderService>,
```
dignifiedquire, line 233: I probably missed this, but why do we need `rpc` and `internal_rpc`?

src/main.rs
rklaehn: Yes.

src/main.rs
rklaehn: The provider is now useful independent of the path you might or might not want to add. So you want to start it up as soon as possible, and then add the path or the stuff from stdin.

Basically instead of
- build the collection
- start the server

it is now
- start the server
- build the collection
- add it to the already running server

This is necessary because when you wait for stdin you might wait for a long time / forever, but we still want iroh to be immediately available for rpc calls.

src/provider.rs
rklaehn: We don't strictly speaking need it. There is a feature in quic-rpc that allows you to combine two rpc mechanisms. E.g. in memory for internal use and quinn for inter process.

But I tried it out, and it leads to very long types (quic-rpc currently does not support boxing, so you have to drag the types around). Also, you want to process errors for internal rpc calls differently than errors from external calls. An accept error in the internal rpc means that the controller has been dropped, which means that you probably want to exit. An error in an accept from an external source is something you just want to log and otherwise ignore.

src/provider.rs
rklaehn: I guess I don't feel the pain as much because of vscode

image

Will put it in constants.

src/provider.rs
flub: > (for some reason this did not make it to docs.rs because it is disabled by default... How do I change that?)

`#[cfg(any(feature = "the_feature", doc))]` if my memory is right

flub (commented):
Are we concerned at all that now we always open a localhost port on which we accept commands without any authentication?

src/main.rs
```diff
@@ -241,70 +310,98 @@ async fn main() -> Result<()> {
...
+                    item.size
+                );
             }
+            close(client).await
+        }
+        Commands::Add { path, rpc_port } => {
+            let client = make_rpc_client(rpc_port).await?;
+            let response = client.rpc(ProvideRequest { path: path.clone() }).await??;
+            println!(
+                "path {} added. Hash {}",
```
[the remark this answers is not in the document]
rklaehn: I added this back.

src/main.rs
```diff
@@ -40,6 +52,25 @@ enum Commands {
         /// If this path is provided and it exists, the private key is read from this file and used, if it does not exist the private key will be persisted to this location.
         #[clap(long)]
         key: Option<PathBuf>,
+        /// Optional rpc port, defaults to 4919
+        #[clap(long, default_value_t = DEFAULT_RPC_PORT)]
+        rpc_port: u16,
```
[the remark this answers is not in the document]
rklaehn: Not sure. I guess we could have a special value to disable it, but default should be on

src/main.rs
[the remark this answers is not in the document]
rklaehn: Added it back.

src/main.rs
```diff
@@ -172,11 +203,49 @@ impl FromStr for Blake3Cid {
...
-#[tokio::main(flavor = "multi_thread")]
-async fn main() -> Result<()> {
+/// drop a RpcClient and wait for a short time.
+///
+/// The waiting is not strictly necessary, but if the program terminates immediately
+/// after closing the client, the close message is not being sent to the server,
+/// and the server will hold on to the connection until it times out.
+async fn close<C>(client: RpcClient<ProviderService, C>) -> anyhow::Result<()> {
+    drop(client);
+    tokio::time::sleep(Duration::from_millis(100)).await;
```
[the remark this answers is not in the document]
rklaehn: Yes, this would have to be a tokio task though because it is called from drop. Then I guess it would be just another background task that gets some grace period to execute.

Let's fix it in another PR though. This works.

rklaehn:
> Are we concerned at all that now we always open a localhost port on which we accept commands without any authentication?

Currently not. But if this turns out to be a problem we can use a named pipe. But hopefully not in this PR.

Note that the RPC is optional and disabled when you use ~sendme~ iroh as a library...

src/main.rs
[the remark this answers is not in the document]
rklaehn: I figured that you want it on by default, and if you use it as a library it is optional anyway.

src/main.rs
rklaehn: Fixed it. The text is now mostly the same as before, except that the peer id and stuff gets reported first...

src/main.rs
dignifiedquire: 👍

dignifiedquire (approved):

PR #781, opened 2023-02-22
test: cli integration test for provide-get loop

### LINKED ISSUE #723: tests: add integration tests for the binary

ramfox:
closes #723 

depends on #789


tests/cli.rs
```diff
@@ -0,0 +1,225 @@
...
+use std::process::{Child, ChildStderr, Command, Stdio};
+
+use anyhow::{Context, Result};
+use tempfile::tempdir;
+
+#[test]
+fn cli_provide_one_file() -> Result<()> {
+    // provide a path to a file, do not pipe from stdin, do not pipe to stdout
+    test_provide_get_loop(&PathBuf::from("transfer").join("foo.bin"), false, false)
+}
```
ramfox, line 15: one other potential change: we can use random data saved to temp files rather than keeping these fixtures around.

tests/cli.rs
```diff
@@ -0,0 +1,225 @@
...
+fn match_provide_stderr(stderr: BufReader<ChildStderr>, use_stdin: bool) -> Result<String> {
+    let mut lines = stderr.lines();
+
+    let mut res = vec![
+        r"Reading \S*",
+        r"Collection: [\da-z]{59}",
+        r"",
+        r"PeerID: [_\w\d-]*",
+        r"Auth token: [\w\d]*",
+    ];
```
ramfox, line 181: this can be made even more "general"
right now we iterate through each line & ensure the lines appear in order.
We could instead push all those lines to a single string and have each regex expect to exist _somewhere_ in the output, rather than in this specific order.

dignifiedquire: I think having an order is great

tests/cli.rs
```diff
@@ -0,0 +1,225 @@
...
+    test_provide_get_loop(&PathBuf::from("transfer"), false, false)
+}
+
+#[test]
+fn cli_provide_from_stdin_to_stdout() -> Result<()> {
+    // provide a file, pipe content to the provider's stdin, pipe content to the getter's stdout
+    test_provide_get_loop(&PathBuf::from("transfer").join("foo.bin"), true, true)
+}
+
+// Test the provide and get loop for success, stderr output, and file contents.
```
flub, line 29: Not that this matters much, but might as well use doc comments.  Often can be nice to look at `cargo doc --document-private-items`

src/main.rs
```diff
@@ -327,7 +328,9 @@ async fn provide_interactive(
 
     let (db, hash) = provider::create_collection(sources).await?;
 
-    println!("Collection: {}\n", Blake3Cid::new(hash));
+    out_writer
+        .println(format!("Collection: {}\n", Blake3Cid::new(hash)))
+        .await;
     for (_, path, size) in db.blobs() {
         println!("- {}: {} bytes", path.display(), size);
```
dignifiedquire, line 335: what about this print?

tests/cli.rs
```diff
@@ -0,0 +1,225 @@
...
+//
+// Can optionally pipe the given `path` content to the provider from stdin & can optionally save the output to an `out` path.
+//
+// Runs the provider as a child process that stays alive until the getter has completed. Then
+// checks the output of the "provide" and "get" processes against expected regex output. Finally,
+// test the content fetched from the "get" process is the same as the "provided" content.
+fn test_provide_get_loop(path: &Path, use_stdin: bool, use_stdout: bool) -> Result<()> {
```
dignifiedquire, line 36: it would be much more readable to make the `bool`s an enum like this

```rust
enum Output {
  Path,
  Stdout,
}

enum Input {
  Path,
  Stdin
}

```

tests/cli.rs
```diff
@@ -0,0 +1,225 @@
...
+    let provider = if use_stdin {
+        let f = File::open(&path)?;
+        let stdin = Stdio::from(f);
+        Command::new(iroh)
+            .stderr(Stdio::piped())
+            .stdout(Stdio::null())
+            .stdin(stdin)
+            .arg("provide")
+            .arg("--addr")
+            .arg("127.0.0.1:0")
```
dignifiedquire, line 62: make this a constant/variable

tests/cli.rs
```diff
@@ -0,0 +1,225 @@
...
+        matches(
+            lines
+                .next()
+                .context("Unexpected end of 'get' output")?
+                .trim(),
+            re,
+        )?;
+    }
+    Ok(())
+}
```
dignifiedquire, line 166: maybe make a helper to iterate over the lines, so that writing assertions is a bit cleaner

```rust
assert_matches_lines!(
    stderr,
    r"Fetching: [\da-z]{59}",
    r"\[1/3\] Connecting ...",
    r"\[2/3\] Requesting ...",
    r"\[3/3\] Downloading collection...",
    r"\d* file\(s\) with total transfer size [\d.]* ?[KMGT]?i?B",
    r"Done in \d* seconds?",    
);
```

tests/cli.rs
```diff
@@ -0,0 +1,225 @@
...
+        None => {
+            let expect_content = std::fs::read(path)?;
+            assert_eq!(expect_content, get_output.stdout);
+            Ok(())
+        }
+        Some(out) => compare_files(path, out),
+    }
+}
+
+// Wrapping the [`Child`] process here allows us to impl the `Drop` trait ensuring the provide
```
flub, line 109: `///`

tests/cli.rs
```diff
@@ -0,0 +1,225 @@
...
+        let file_name = expect_path.file_name().unwrap();
+        let expect = std::fs::read(expect_path)?;
+        let got = std::fs::read(got_dir_path.join(file_name))?;
+        assert_eq!(expect, got);
+    }
+
+    Ok(())
+}
+
+// Looks for regex matches on stderr output for the getter.
```
flub, line 141: `///`

tests/cli.rs
flub: in order is fine i guess. otherwise it would just make it more complex

dignifiedquire:
Do you need the checked in text fixtures, or could they be generated on the fly using some fixed randomness?

tests/cli.rs
```diff
@@ -0,0 +1,225 @@
...
+    if use_stdin {
+        res = res[1..].to_vec();
+    }
+
+    for re in res {
+        matches(next_line(&mut lines)?.trim(), re)?;
+    }
+
+    // get all-in-one ticket
+    let re = r"All-in-one ticket: ([_a-zA-Z\d-]*)";
```
flub, line 193: This is more general for the regexes, you seem to write fairly strict ones that will need adjusting if we'd change the representation.  You could take anything but whitespace to end of line.  Not saying this needs to be changes, more wondering if it's deliberate in being strict.

src/main.rs
```diff
@@ -309,6 +309,7 @@ async fn provide_interactive(
                     paths.push(el.path().into());
                 }
             }
+            paths.sort();
```
flub, line 312: Is this still needed?  It's a little unfortunate to have to do this, would be nice if it could be avoided.

flub (commented):
I don't hate it!  Generally look ok I think.

tests/cli.rs
ramfox: ty this was bothering me & I didn't understand the clearest way to fix it! 🔥

ramfox:
> Do you need the checked in text fixtures, or could they be generated on the fly using some fixed randomness?

definitely can be generated on the fly, can do that in this round.

tests/cli.rs
ramfox: Oh, ya. Part of the "test the cli" convo came out of us not realizing we were using the wrong hashes in the output. So I added strictness here to cover that original intention.

src/main.rs
ramfox: I debated with myself, glad you brought it up. Is it okay that the same content on two different machines (folders structured the same way) will create different hashes?

tests/cli.rs
ramfox: 🤦‍♀️ i meant to do doc comments... ty

src/main.rs
dignifiedquire: hmm, seems to be we should sort these, as adding the same files on two machines should result in the same hash

src/main.rs
ramfox: !!!!

added to the correct output & tested.

tests/cli.rs
ramfox: Okay, so to "make it work" for both the `get` and `provide` use cases, I needed to be able to ensure we could check for a specific line a certain number of times, which is where this odd `r"expr"; 1` syntax comes from. 

It's more clear in `match_provide_stderr` below how it's intended to be used.

I'd love some input on:
1) if this is too weird
2) if there is a better way to do the macro. This is basically my first time writing one, so I'm sure there are things I missed.
3) if I should not use a macro and return it to the previous version.

src/main.rs
flub: Hum, we need a way of hashing that works natively on sets: no matter which order we add files we'd like the same hash.  :smile:  not that that would help since we hash the metadata.

tests/cli.rs
```diff
@@ -0,0 +1,295 @@
...
+    let rx = regex::Regex::new(re)?;
+    let line = next_line(&mut remaining_lines)?;
+    if !rx.is_match(&line) {
+        anyhow::bail!(match_err_msg(&line, re))
+    }
+    let caps = rx
+        .captures(&line)
+        .context("expected match on 'All-in-one' ticket")?;
+    Ok(caps
```
dignifiedquire, line 239: if you make the macro return all captures from the regexps, you could add these to the macro invocation, which would read much nicer.

tests/cli.rs
```diff
@@ -0,0 +1,295 @@
...
+
+use anyhow::{Context, Result};
+use rand::RngCore;
+use tempfile::tempdir;
+
+const ADDR: &str = "127.0.0.1:0";
+
+fn make_rand_file(size: usize, path: &Path) -> Result<()> {
+    let mut content = vec![0u8; size];
+    rand::thread_rng().fill_bytes(&mut content);
```
dignifiedquire, line 16: please use `rand::rngs::StdRng::seed_from_u64(1)` to have the same randomness in every test run, otherwise if there is any issue it will be very hard to reproduce

tests/cli.rs
```diff
@@ -0,0 +1,295 @@
...
+            let expect_content = std::fs::read(path)?;
+            assert_eq!(expect_content, get_output.stdout);
+            Ok(())
+        }
+        Some(out) => compare_files(path, out),
+    }
+}
+
+// Wrapping the [`Child`] process here allows us to impl the `Drop` trait ensuring the provide
+// process is killed when it goes out of scope.
```
dignifiedquire, line 158: still only two slashes

tests/cli.rs
dignifiedquire: seems fine

src/main.rs
```diff
@@ -315,11 +316,15 @@ async fn provide_interactive(
 
     let (db, hash) = provider::create_collection(sources).await?;
 
-    println!("Collection: {}\n", Blake3Cid::new(hash));
+    out_writer
```
flub, line 319: Maybe not?  https://github.com/n0-computer/iroh/pull/790/files#r1115540394

ramfox: lmao you did this to hurt me

😆

tests/cli.rs
ramfox: i'm never going to learn this lesson

tests/cli.rs
ramfox: there is a need for multiple matches in `match_provider_output`

src/main.rs
```diff
@@ -280,15 +280,12 @@ async fn provide_interactive(
...
-    let out_writer = OutWriter::new();
     let keypair = get_keypair(key).await?;
 
     let mut tmp_path = None;
 
     let sources = if let Some(path) = path {
-        out_writer
-            .println(format!("Reading {}", path.display()))
-            .await;
+        println!("Reading {}", path.display());
```
ramfox, line 288: Moved "Reading" to `println`. However, leaving this to print to `out_writer` may have been intentional since it is technically "progress".

Easy to switch back, so just say the word.

tests/cli.rs
ramfox: I adjusted the macro accordingly! It definitely reads easier.

However, all of these matches (except for the all-in-one ticket) only require `is_match`, not `captures`. 

According to the `regex`, you should favor `is_match` over `captures` when you can, since `is_match` is less computationally heavy.

This is being used in testing, tho, so this is probably fine because it's probably better to lean on the side of legibility. But in case it matters, I thought I'd bring it up.

src/main.rs
dignifiedquire: works for me

tests/cli.rs
```diff
@@ -0,0 +1,295 @@
...
+        let entries = std::fs::read_dir(&path)?;
+        entries.count()
+    } else {
+        1
+    };
+
+    let iroh = env!("CARGO_BIN_EXE_iroh");
+
+    // spawn a provider & optionally provide from stdin
+    let provider = if input == Input::Stdin {
```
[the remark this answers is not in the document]
dignifiedquire: still not a `match`, do you not want it, or did you forget?

tests/cli.rs
dignifiedquire: Yeah, should be totally fine, given that these are not hot spots. Thanks for bringing it up though 👍

dignifiedquire (approved):
nice work, it's a pretty simple and clean end result :)

src/main.rs
flub: hey, i only wanted to be able to read "8534363 bytes".  originally i copied you `out_writer` approach until called out on it :sweat_smile:

tests/cli.rs
```diff
@@ -0,0 +1,280 @@
...
+fn make_rand_file(size: usize, path: &Path) -> Result<()> {
+    let mut content = vec![0u8; size];
+    rand::rngs::StdRng::seed_from_u64(1).fill_bytes(&mut content);
+    std::fs::write(path, content)?;
+    Ok(())
+}
+
+#[test]
+fn cli_provide_one_file() -> Result<()> {
+    let dir = tempdir()?;
```
flub, line 23: So I'm still curious if this was a real bug in testdir that needs fixing?

flub (approved):

tests/cli.rs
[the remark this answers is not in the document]
ramfox: nah i meant to change it, I just only changed it in one place and not both places that are using this same format. ty for catching it.

tests/cli.rs
ramfox: hi! just re-ran the tests to be sure: https://github.com/n0-computer/iroh/actions/runs/4264055450/jobs/7421605558

Yes, this is a bug:

"thread 'cli_provide_folder' panicked at 'Failed to create testdir: Failed to remove obsolete testdir-current symlink"

The windows machines aren't able to remove the `testdir-current` symlink in these cases, even though it works just fine when `testdir` is used as part of the code testing the library.

flub: thanks!

flub (approved):

PR #790, opened 2023-02-23
feat: Show more numbers in human readable form

flub:
This shows sizes in KiB, MiB etc.  Also adds a few totals and transfer
summaries.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -315,20 +315,14 @@ async fn provide_interactive(
 
     let (db, hash) = provider::create_collection(sources).await?;
 
-    out_writer
-        .println(format!("Collection: {}\n", Blake3Cid::new(hash)))
-        .await;
+    println!("Collection: {}\n", Blake3Cid::new(hash));
     let mut total_size = 0;
     for (_, path, size) in db.blobs() {
         total_size += size;
-        out_writer
-            .println(format!("- {}: {}", path.display(), HumanBytes(size)))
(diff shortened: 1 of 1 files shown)


dignifiedquire (approved):
so human

src/main.rs
```diff
@@ -315,11 +315,20 @@ async fn provide_interactive(
 
     let (db, hash) = provider::create_collection(sources).await?;
 
-    println!("Collection: {}\n", Blake3Cid::new(hash));
+    out_writer
```
dignifiedquire, line 318: are you sure we want to push all this to stderr? seems to me that this is stdout content for the provider

flub: Ugh, good catch.  I'm not sure at all.  I guess this is "data" and not progress information so should go to stdout?  But like the peer ID, auth token and ticket are also written using to OutWriter, and this seems to fall in that category to me and why I changed this.  Are they wrong too?

Also see, https://github.com/n0-computer/iroh/pull/781/files#diff-42cb6807ad74b3e201c5a7ca98b911c5fa08380e942be6e4ac5807f8377f87fcR319-R327

dignifiedquire: I think they are wrong too, yes

flub: Ok, updated them all to direct println! calls.  PTAL

dignifiedquire (approved):

PR #791, opened 2023-02-23
fix(ci): format output as table

Arqu:


dignifiedquire (approved):
beautiful

dignifiedquire:
One note, maybe still hide it in a collapsed thing by default? given its length

Arqu:
Thats the thing, the collapsed summary doesn't render tables properly....

PR #792, opened 2023-02-24
docs: fix typo

Nazeh:
I assume it is a typo.

> This removes dependencies that are only relevant when using `iroh` as
a ~library~ cli.


dignifiedquire (approved):
thanks

PR #797, opened 2023-02-27
No out writer

rklaehn:
I think I just don't get why OutWriter exists, so I thought I might as well do a PR to replace it, and you tell me why this is wrong...

I would rename out_println! to something else, maybe progress! Did so.

I checked the code, and it seems that tokio::io::stderr() is meant to be used as a singleton. It calls std::io::stderr() and wraps it into a `Blocking` helper and something else for windows compat, `SplitByUtf8BoundaryIfWindows`, neither of which does anything obviously very expensive.

I don't think it even allocates on creation.

Here is Blocking::new from the tokio guts:
```rust
cfg_io_blocking! {
    impl<T> Blocking<T> {
        #[cfg_attr(feature = "fs", allow(dead_code))]
        pub(crate) fn new(inner: T) -> Blocking<T> {
            Blocking {
                inner: Some(inner),
                state: State::Idle(Some(Buf::with_capacity(0))), // no allocation b/c cap 0
                need_flush: false,
            }
        }
    }
}
``` 

And here is tokio::io::stderr():

https://docs.rs/tokio/latest/src/tokio/io/stderr.rs.html#71

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -9,7 +9,6 @@ use indicatif::{
 use iroh::protocol::AuthToken;
 use iroh::provider::Ticket;
 use tokio::io::AsyncWriteExt;
-use tokio::sync::Mutex;
 use tracing_subscriber::{prelude::*, EnvFilter};
 
 use iroh::{get, provider, Hash, Keypair, PeerId};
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -82,26 +81,15 @@ enum Commands {
...
-    }
+macro_rules! progress {
+    // Match a format string followed by any number of arguments
+    ($fmt:expr $(, $args:expr)*) => {{
+        // Use the `format!` macro to format the string with the arguments
+        let mut message = format!($fmt $(, $args)*);
+        // Print the formatted string to the console with a newline
+        message.push('\n');
+        tokio::io::stderr().write_all(message.as_ref()).await.unwrap();
+    }};
```
flub, line 92: So why even use our own macro over `println!`?  Should we not call `println!` and `eprintln!` macros at all in tokio code?  My gut feeling says this only matters when we're piping entire files in and out of stdio/stdout and not for printing progress.  We already call `println!` in main anyway.

The benefit of directly using `eprintln!` is that we don't even have to worry about this garbled output problem as the stdlib takes care of it.

I admit wrapping `eprintln!` into `progress!` is nice sematically as it's more obvious what we're printing.

flub (approved):
Good call, I also don't get the point of `OutWriter` right now.  But let's maybe wait for feedback from @dignifiedquire who introduced this afaik.

src/main.rs
rklaehn: I think the difference is that `eprintln!` can block, and this would cause havoc if there is only a single thread. I mean, this is the reason tokio::io::stderr() exists I guess...

dignifiedquire (approved):
Works for me, in the first version of the outwriter there was more complex logic deciding between stderr and stdout at run time, which is why I introduced it in the first place, now it is gone, happy to simplify it.

PR #798, opened 2023-02-27
feat: use chunk groups feature

rklaehn:
caution: this is not compatible with existing iroh/sendme

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "abao"
-version = "0.1.2"
+version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


dignifiedquire:
what's missing here, given this is a draft?

rklaehn:
It was a draft because the crate was not published yet, and the tests were failing. Let's see, maybe they work now...

dignifiedquire:
@rklaehn do we want to ship this? did you compare the perf numbers?

rklaehn:
Well, perf is basically identical, a tad faster. The main benefit is that the outboard is smaller, so we can share larger files with outboard in memory.

As far as I am concerned it is ready to go - provided that we're clear that this might change in the future, together with the protocol.

dignifiedquire:
@flub any concerns merging this before DC release?

flub (approved):

PR #800, opened 2023-02-28
adjust ci test run order

ramfox:
any thoughts on this?

I feel like I'm missing information whenever the tests fail on nightly before seeing if it passed on stable.


flub:
You need `fail-fast: false` for what you want I think: https://github.com/n0-computer/beetle/blob/main/.github/workflows/ci.yml#L15

I'd be in favour of adding that.

PR #799, opened 2023-02-27
add more context for `cli_provide` test errors

ramfox:
- adds context in `assert_matches_line` about which line failed to be read
- this... shouldn't really make a difference, but just in case: make sure the `getter` command is dropped before testing the output
- test the success of the `get` command before testing the progress output

Hopefully with these combined, we can get more information out of the flakey failing linux tests seen here:
https://github.com/n0-computer/iroh/actions/runs/4282332291/jobs/7456474101

-----

Edit:
flakey tests were fixed by #804 

This is now just a PR to improve the error reporting

PR #802, opened 2023-02-28
fix: Do not send duplicate NotFound responses

flub:
The code was accidentally sending this response twice.  send_blob()
already sends NotFound, and it seems like it belongs more there.


flub:
Test failures are due to flaky tests that will be fixed in either #799 or #804.

dignifiedquire (approved):

PR #803, opened 2023-02-28
ref: micro-optimise by reserving buffer size

flub:
Before going into the loop we may as well make sure to reserve enough
space for the entire buffer.  Otherwise we could be allocating
multiple times.

This does not change the total allocated size.

### DIFF
--- a/src/protocol.rs
+++ b/src/protocol.rs
@@ -93,6 +93,7 @@ pub(crate) async fn read_lp(
     };
     let mut reader = reader.take(size);
     let size = usize::try_from(size).context("frame larger than usize")?;
+    buffer.reserve(size);
     loop {
         let r = reader.read_buf(buffer).await?;
         if r == 0 {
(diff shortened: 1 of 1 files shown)


flub:
Test failures are due to flaky tests that will be fixed in either #799 or #804.

src/protocol.rs
```diff
@@ -93,6 +93,7 @@ pub(crate) async fn read_lp(
     };
     let mut reader = reader.take(size);
     let size = usize::try_from(size).context("frame larger than usize")?;
+    buffer.reserve(size);
```
dignifiedquire, line 96: This should be limited, I can now easily attack your machine by putting a crazy value here, and causing an OOM

flub: My thinking was that I wasn't at all changing how much is being allocated, this attack already existed.  To be fair I also faultily assumed MAX_MESSAGE_SIZE was already being respected while it obviously wasn't.

So fair shout, MAX_MESSAGE_SIZE wasn't being checked yet so may as well improve this.

dignifiedquire (approved):

PR #810, opened 2023-03-03
ci: Add cargo-deny

flub:
This is a minimal cargo deny config which should not give us too many
issues.  Let's see how this works in practice.

I just accepted all the currently used licenses, did not vet them.


deny.toml
```diff
@@ -0,0 +1,21 @@
+[bans]
+multiple-versions = "allow"
```
flub, line 2: The other option is to list them something like this: https://github.com/deltachat/deltachat-core-rust/blob/c38ae31f3198fd56b01d7323cd8c7075c812db26/deny.toml#L14-L36

I don't think this is really worth it unless we really made an effort to remove dups.  Which I don't think we've done yet.

dignifiedquire: uff, yeah no need for that complexity, just allow multiple versions for now

dignifiedquire (approved):
looks reasonable

PR #812, opened 2023-03-06
Implements finding default route interface for linux

flub:


Cargo.toml
```diff
@@ -34,6 +34,7 @@ hostname = "0.3.1"
 indicatif = { version = "0.17", features = ["tokio"], optional = true }
 libc = "0.2.139"
 multibase = { version = "0.9.1", optional = true }
+netlink-packet-route = "0.15.0"
```
dignifiedquire, line 37: can you make these new dependencies linux only as well?

flub: good call, done

dignifiedquire (approved):

PR #817, opened 2023-03-08
feat: Brutally simple local network discovery

flub:
This adds an extremely simple form of local network discovery on an IPv4 network.  It is not meant to be easy to use but can be combined to announce a service on a particular socket address.


dignifiedquire:
@flub can we close this?

PR #820, opened 2023-03-09
feat: Set multiple addrs in the ticket

flub:
If the provider address is bound to a wildcard address put a list of addresses in the ticket.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -290,9 +290,13 @@ async fn main_impl() -> Result<()> {
             let Ticket {
                 hash,
                 peer,
-                addr,
+                addrs,
                 token,
             } = ticket;
+            let addr = addrs
+                .get(0)
+                .copied()
+                .context("missing SocketAddr in ticket")?;
             let opts = get::Options {
                 addr,
                 peer_id: Some(peer),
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -290,9 +290,13 @@ async fn main_impl() -> Result<()> {
...
                 hash,
                 peer,
-                addr,
+                addrs,
                 token,
             } = ticket;
+            let addr = addrs
+                .get(0)
+                .copied()
+                .context("missing SocketAddr in ticket")?;
```
flub, line 299: this is bascially ignoring this problem for now in our own implementation.  i don't feel like spending time on this yet.  I'll come back to this once working on trying to connect to all addresses at the same time.

src/provider.rs
```diff
@@ -1020,8 +1040,30 @@ impl FromStr for Ticket {
     }
 }
 
+/// Create a [`quinn::ServerConfig`] with the given keypair and limits.
```
dignifiedquire, line 1043: why is this public now?

flub: I don't know, I think @rklaehn wrote this.  I only moved it since it was below the tests module for some reason (so i missed there were already tests and then the compiler complained that i can't have two tests modules)

Arqu:
IP validation/integration test is blocked on this for CI.

src/main.rs
dignifiedquire: please file an issue, this will break easily. in most setups if we don’t try all addrs

dignifiedquire (approved):

flub:
DO NOT PRESS THE `Delete branch` button!

PR #821, opened 2023-03-09
Allow disabling rpc

rklaehn:
We might later want to allow rpc port `0` for "choose a free one", but I think for now this is fine. For a rpc port you really want it to be known beforehand.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -37,6 +37,42 @@ struct Cli {
     keylog: bool,
 }
 
+#[derive(Debug, Clone)]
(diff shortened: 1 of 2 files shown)


flub:
Hum, not sure I can quite get on board with `0` meaning disabled.  That's rather unintuitive.

rklaehn:
How about this?

```
iroh provide --rpc-port=disabled
```

flub (approved):
much nicer, thanks!

flub (approved):

PR #822, opened 2023-03-09
feat(net): implement local address detection

dignifiedquire:
Based on the work done in `x-hp`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,6 +8,8 @@ pub mod protocol;
 pub mod provider;
 pub mod rpc_protocol;
 
+pub mod net;
+
 mod tls;
 mod util;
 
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -8,6 +8,8 @@ pub mod protocol;
 pub mod provider;
 pub mod rpc_protocol;
 
+pub mod net;
```
dignifiedquire, line 11: Just for now, probably want to make this private once it is actually being used.

dignifiedquire:
Opened an issue for the cargo-deny warning: https://github.com/shellrow/default-net/issues/19

flub (dismissed):

flub (approved):

Issue #806, opened 2023-03-01
Add further RPC commands


rklaehn:
After we got iroh2 as a long runnning process (see https://github.com/n0-computer/iroh/issues/720 ), we need more ways to interact with it.

- get peer id / addresses
- shutdown
- ...

This involves defining new RPC commands and implementing them.

We also might want to revisit some of the existing commands, such as list. As it exists now it is not that useful.

PR #823, opened 2023-03-09
Id and shutdown

rklaehn:
Implements https://github.com/n0-computer/iroh/issues/806

What about events? Should we expose those as well?

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -13,7 +13,7 @@ use iroh::protocol::AuthToken;
 use iroh::provider::{Database, Provider, Ticket};
 use iroh::rpc_protocol::{
     ListRequest, ProvideRequest, ProvideResponse, ProvideResponseEntry, ProviderRequest,
-    ProviderResponse, ProviderService, VersionRequest,
+    ProviderResponse, ProviderService, ShutdownRequest, VersionRequest,
 };
 use quic_rpc::transport::quinn::{QuinnConnection, QuinnServerEndpoint};
 use quic_rpc::{RpcClient, ServiceEndpoint};
@@ -61,6 +61,17 @@ enum Commands {
     /// List hashes
     #[clap(about = "List hashes")]
     List {
+        /// Optional rpc port, defaults to 4919
+        #[clap(long, default_value_t = DEFAULT_RPC_PORT)]
+        rpc_port: u16,
+    },
+    /// Shutdown
+    #[clap(about = "Shutdown provider")]
+    Shutdown {
+        /// Shutdown mode.
+        /// Hard shutdown will immediately terminate the process, soft shutdown will wait for all connections to close.
+        #[clap(long, default_value_t = false)]
+        hard: bool,
         /// Optional rpc port, defaults to 4919
         #[clap(long, default_value_t = DEFAULT_RPC_PORT)]
         rpc_port: u16,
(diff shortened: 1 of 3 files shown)


src/provider.rs
```diff
@@ -448,6 +468,24 @@ impl Handler {
             version: env!("CARGO_PKG_VERSION").to_string(),
         }
     }
+    async fn id(self, _: IdRequest) -> IdResponse {
+        IdResponse {
+            peer_id: Box::new(self.peer_id),
```
flub, line 473: what's the reason for this boxing?

src/provider.rs
```diff
@@ -409,14 +418,25 @@ impl Future for Provider {
...
+    pub db: Database,
+    /// cancellation token
+    pub shutdown: CancellationToken,
+    /// peer id
+    pub peer_id: PeerId,
+    /// auth token
+    pub auth_token: AuthToken,
+    /// listen address
+    pub listen_addr: SocketAddr,
+}
```
flub, line 433: I guess the main difference with `Provider` is that you can clone this?  And the main thing that can't be cloned in the `Provider` is the `JoinHandle` afaik.  Maybe this is fine, maybe we can build `Provider` out of the `JoinHandle` & `ProviderHandle`?  And if going that way maybe we can have `Provider::handle()` as a public API at some point.

I guess for now the question of my rambling is: can we make these two more the same and use `ProviderHandles` inside the `Provider` to avoid duplication?

flub (commented):
I'd be happy to approve as-is as well, but curious what you make of my comments nevertheless

src/main.rs
```diff
@@ -65,6 +62,24 @@ enum Commands {
         #[clap(long, default_value_t = DEFAULT_RPC_PORT)]
         rpc_port: u16,
     },
+    /// Shutdown
+    #[clap(about = "Shutdown provider")]
+    Shutdown {
+        /// Shutdown mode.
+        /// Hard shutdown will immediately terminate the process, soft shutdown will wait for all connections to close.
```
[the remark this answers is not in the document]
rklaehn: I think it is ok to call it a hard shutdown despite the option being called force. A "forced shutdown" does sound weird.

src/provider.rs
```diff
@@ -398,6 +402,11 @@ impl Provider {
     pub fn shutdown(&self) {
         self.cancel_token.cancel();
     }
+
+    /// Returns a token that can be used to cancel the provider.
+    pub fn cancel_token(&self) -> CancellationToken {
```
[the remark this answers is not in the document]
rklaehn: I guess we could wrap it or use a oneshot.

src/provider.rs
rklaehn: Otherwise you get an enum variant size mismatch warning, because these things are large.

src/provider.rs
rklaehn: Yes, I was almost at the point where I wanted to make Provider cloneable. I did not do it because I saw that provider implements Future.

We could make Provider cloneable. It is already not really "the provider" (that is some tokio task) but some kind of handle that allows you to get info for the provider.

WDYT?

rklaehn: How about we see if we can make Provider a proper handle in a subsequent PR?

rklaehn:
https://github.com/n0-computer/iroh/issues/827 to be done in a subsequent PR

src/provider.rs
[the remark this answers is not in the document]
flub: let's just leave it for now and hope we don't regret it

src/provider.rs
flub: oh, the whole `IdResponse` becomes too large.  May want to explain that in the rpc_protocol.rs file with a comment.  Does it make sense to put everything in one "inner" box instead?

flub (approved):

PR #825, opened 2023-03-09
Fix local addrs via java on android

dignifiedquire:
This is very, very sad, but it appears netlink is not allowed starting with Android API 30, so we might as well just do this.. Tested on a linux box to work so far.

API References: https://developer.android.com/reference/java/net/NetworkInterface
J4Rs: https://github.com/astonbitecode/j4rs


flub (approved):

flub:
FWIW tailscale invokes the `ip route` command on android to find the default route.  That might mean `ip addr` could also be available and we could parse that too.

flub (changes requested):
This seems overkill now.  We can use netlink on android.

PR #826, opened 2023-03-10
begin impl Server side of derp, starting with the server side of the client connection

ramfox:
Implement structures that manage the connection from the server to the
client: the `ClientConnManager`, the `ClientConnWriter`, `ClientConnReader`.
When the server "accepts" a connection (`accept` is not yet
implemented), it separates the write and read portions of the
connection. It should create a `ClientConnManager` to interact with the
connection. The `ClientConnManager` runs the `ClientConnWriter` and `ClientConnReader` in a loop.
The `ClientConnWriter` accepts notifications & relayed packets from the server & sends them to the client.
The `ClientConnReader` to reads any notifications or packets it wants
the server to relay from the client, & sends them to be processed in the
server.

### DIFF
--- a/src/hp/derp.rs
+++ b/src/hp/derp.rs
@@ -10,6 +10,8 @@
 //! Based on tailscale/derp/derp.go
 
 mod client;
+mod client_conn;
+mod conn;
 pub mod http;
 mod map;
 mod server;
--- a/src/hp/derp.rs
+++ b/src/hp/derp.rs
@@ -9,7 +9,7 @@
 //! between very aggressive NATs, firewalls, no IPv6, etc? Well, DERP.
 //! Based on tailscale/derp/derp.go
 
-mod client;
+pub(crate) mod client;
 mod client_conn;
 mod conn;
 pub mod http;
--- a/src/hp/derp.rs
+++ b/src/hp/derp.rs
@@ -174,7 +174,7 @@ async fn write_frame_header(
     Ok(())
 }
 
-/// AsyncWrites a complete frame & flushes it.
+/// AsyncWrites a complete frame. Does not flush.
 async fn write_frame(
     mut writer: impl AsyncWrite + Unpin,
     frame_type: FrameType,
@@ -188,8 +188,31 @@ async fn write_frame(
     for b in bytes {
         writer.write_all(b).await?;
     }
-    writer.flush().await?;
     Ok(())
+}
+
+/// AsyncWrites a complete frame, errors if it is unable to write within the given `timeout`.
+/// Ignores the timeout if `timeout.is_zero()`
+///
+/// Does not flush.
+async fn write_frame_timeout(
+    writer: impl AsyncWrite + Unpin,
+    frame_type: FrameType,
+    bytes: Vec<&[u8]>,
+    timeout: Duration,
+) -> Result<()> {
+    if timeout.is_zero() {
+        return write_frame(writer, frame_type, bytes).await;
+    }
+    tokio::select! {
+        biased;
+        res = write_frame(writer, frame_type, bytes) => {
+            res
+        }
+        _ = tokio::time::sleep(timeout) => {
+            bail!("could not write {} within the deadline {:?}", frame_type, timeout);
+        }
+    }
 }
 
 #[cfg(test)]
@@ -207,6 +230,7 @@ mod tests {
 
         let expect_buf = b"hello world!";
         write_frame(&mut writer, FRAME_HEALTH, vec![expect_buf]).await?;
+        writer.flush().await;
         println!("{:?}", reader);
         let mut got_buf = BytesMut::new();
         let (frame_type, frame_len) = read_frame(&mut reader, 1024, &mut got_buf).await?;
--- a/src/hp/derp.rs
+++ b/src/hp/derp.rs
@@ -22,7 +22,7 @@ pub use self::server::Server;
 
 use std::time::Duration;
 
-use anyhow::{bail, Result};
+use anyhow::{ensure, Result};
 use bytes::BytesMut;
 use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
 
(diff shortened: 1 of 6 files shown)


ramfox:
SO FAR:
Implement structures that manage the connection from the server to the
client: the `ClientConnManager`, the `ClientConnWriter`, `ClientConnReader`.
When the server "accepts" a connection (`accept` is not yet
implemented), it separates the write and read portions of the
connection. It should create a `ClientConnManager` to interact with the
connection. The `ClientConnManager` runs the `ClientConnWriter` and `ClientConnReader` in a loop.
The `ClientConnWriter` accepts notifications & relayed packets from the server & sends them to the client.
The `ClientConnReader` to reads any notifications or packets it wants
the server to relay from the client, & sends them to be processed in the
server.
The server should hold onto the `ClientConnManager`, and use that as
a handle to communicate with the `ClientConnWriter` to forward any
relayed packets, or to shutdown the connection with the client if
something goes wrong.

TODO:
- test the `ClientConnReader` side of the connection
- Use the `Server` to accept connections & create a `ClientConnManager` for each and register it with the `ClientSet`
- server must know the list of mesh peers
- write struct `ClientSet`, that handles single clients & duplicate clients, 
- The server needs a loop that listens for 
   - updates to a client
   - packets it needs to relay to a client
   - clients that need to be shutdown
   - send notifications to clients about peers they have been communicating with who have dropped out of the network
   - other things I may be missing
 larger questions here about the right way to handle the `ClientSet`, using a mutex or can we do these updates sequentially & not have to worry about lock contension.

src/hp/derp.rs
```diff
@@ -186,10 +188,33 @@ async fn write_frame(
...
 
+/// AsyncWrites a complete frame, errors if it is unable to write within the given `timeout`.
+/// Ignores the timeout if `timeout.is_zero()`
+///
+/// Does not flush.
+async fn write_frame_timeout(
+    writer: impl AsyncWrite + Unpin,
+    frame_type: FrameType,
+    bytes: Vec<&[u8]>,
+    timeout: Duration,
```
dignifiedquire, line 202: should be an `Option` instead of checking for `zero`

src/hp/derp.rs
```diff
@@ -186,10 +188,33 @@ async fn write_frame(
...
+) -> Result<()> {
+    if timeout.is_zero() {
+        return write_frame(writer, frame_type, bytes).await;
+    }
+    tokio::select! {
```
dignifiedquire, line 207: [`tokio::time::timeout`](https://docs.rs/tokio/latest/tokio/time/fn.timeout.html) is a lot simpler to use than a manual `select`

src/hp/derp/client.rs
```diff
@@ -591,3 +556,105 @@ impl RateLimiter {
...
+    if packet.len() > MAX_PACKET_SIZE {
+        bail!("packet too big: {}", packet.len());
+    }
+    let frame_len = PUBLIC_KEY_LENGTH + packet.len();
+    if let Some(rate_limiter) = rate_limiter {
+        match rate_limiter.check_n(frame_len) {
+            Ok(_) => {}
+            Err(_) => {
+                tracing::warn!("dropping send: rate limit reached");
+                return Ok(());
```
dignifiedquire, line 575: `if rate_limiter.check_n(frame_len).is_err() { .. }` would be a bit more straightforward to read

src/hp/derp/client.rs
```diff
@@ -591,3 +556,105 @@ impl RateLimiter {
...
+
+pub async fn send_note_preferred<W: AsyncWrite + Unpin>(
+    mut writer: W,
+    preferred: bool,
+) -> Result<()> {
+    let byte = {
+        if preferred {
+            [0x00]
+        } else {
+            [0x01]
```
dignifiedquire, line 635: constants?

src/hp/derp/client_conn.rs
```diff
@@ -0,0 +1,840 @@
...
+    fn handle_frame_note_preferred(&mut self, data: &[u8]) -> Result<()> {
+        if data.len() != 1 {
+            bail!("FRAME_NOTE_PREFERRED content is an unexpected size");
+        }
+        self.set_preferred(data[0] != 0)
+    }
+
+    async fn handle_frame_watch_conns(&mut self, data: &[u8]) -> Result<()> {
+        if !data.is_empty() {
+            bail!("FRAME_WATCH_CONNS content is an unexpected size");
```
dignifiedquire, line 601: `ensure!`

src/hp/derp/client_conn.rs
```diff
@@ -0,0 +1,840 @@
...
+    async fn handle_frame_ping(&mut self, data: &[u8]) -> Result<()> {
+        if data.len() != 8 {
+            bail!("FRAME_PING unexpected length {}", data.len());
+        }
+        // TODO:stats
+        // c.s.gotPing.Add(1)
+
+        // TODO: add rate limiter
+
+        let data = <[u8; 8]>::try_from(data)?;
```
dignifiedquire, line 623: you already checked above, either `unwrap` here, or skip the check above

src/hp/derp/client_conn.rs
```diff
@@ -0,0 +1,840 @@
...
+    /// Read off the `reader`, erroring and returning at the first error
+    async fn read(&mut self) -> Result<()> {
+        let mut buf = BytesMut::new();
+        loop {
+            // TODO: return Ok(()) if EOF error
+            let (frame_type, _frame_len) =
+                read_frame(&mut self.reader, MAX_FRAME_SIZE, &mut buf).await?;
+            // TODO: "note client activity", meaning we update the server that the client with this
+            // public key was the last one to receive data
+            match frame_type {
```
dignifiedquire, line 540: These methods below should take just ownership over the read content. That avoids clones, and fixes a correctness issue here potentially. 

```rust
let (frame_type, frame_len) =  read_frame(&mut self.reader, MAX_FRAME_SIZE, &mut buf).await?;
// assuming frame_len is how much we read

let frame_bytes = buf.split_to(frame_len);
// frame_bytes can be now passed as owned to the next pieces
// `buf` can be used again
```

src/hp/derp/client.rs
ramfox: 🔥

PR #830, opened 2023-03-13
Implement holepunching - Initial implementation 

dignifiedquire:
Just opening this as a draft for tracking purpose


RangerMauve:
Are you going to be testing this on mobile by any chance?

src/hp/magicsock/conn.rs
```diff
@@ -1138,23 +1139,37 @@ impl Actor {
...
+                        addr,
+                        dst_ip,
+                        ecn: None,
+                    };
+                    NetworkReadResult::Ok {
+                        source: NetworkSource::Derp,
+                        bytes: part,
+                        meta,
+                    }
+                }
```
ramfox, line 1183: hi! If we are splitting the packets up in `magicsock::Conn`, then I think we should be combining them in `magicsock::Conn` as well. 

We are currently (technically) changing the derp protocol by pushing the batching into `derp` itself. If we instead do that batching inside `magicsock::Conn` in `send_derp`, & pass `derp_client.send` the batched bytes, then we aren't breaking the protocol & are also treating the packets the same on the way in and on the way out.

If we do decide that we would rather break from the protocol & should be batching the packets inside derp itself, that I think we should do the "unbatching" inside derp as well, and pass out a vec of bytes, instead of a single blob that `magicsock` is responsible for unbatching.

But to be clear, I'm advocating we do the batching and unbatching inside `magicsock::send_derp` & `magicsock::process_derp_read_result` instead of changing `derp` code at all.

rklaehn: I am happy to move stuff around. I just did this in the way that required the minimum amount of changes.

When you say breaking the derp protocol, you mean that we break the internal layering between the socket part and the derp part, right? Because in terms of the bytes we send over the wire I don't think this breaks anything. We are still sending `<peerid><bytes>`, just that <bytes> now is the concatenation of multiple udp transmits.

src/hp/magicsock/conn.rs
```diff
@@ -0,0 +1,3379 @@
...
+        // Split up transmits by destination, as the rest of the code assumes single dest.
+        let groups = TransmitIter::new(transmits);
+        for group in groups {
+            match self.network_sender.try_reserve() {
+                Err(mpsc::error::TrySendError::Full(_)) => {
+                    // TODO: add counter?
+                    self.network_send_wakers
+                        .lock()
+                        .unwrap()
+                        .replace(cx.waker().clone());
```
rklaehn, line 610: Probably does not make a big difference, but I think you only have to store the waker if you return pending.

If this returns with n>0 we will return Ready.

src/hp/magicsock/conn.rs
```diff
@@ -0,0 +1,3379 @@
...
+                io::ErrorKind::NotConnected,
+                "connection closed",
+            )));
+        }
+
+        let mut num_msgs = 0;
+        for (buf_out, meta_out) in bufs.iter_mut().zip(metas.iter_mut()) {
+            match self.network_recv_ch.try_recv() {
+                Err(flume::TryRecvError::Empty) => {
+                    self.network_recv_wakers
```
rklaehn, line 652: Same here. If we return ready we don't have to store the waker.

src/bin/derper.rs
```diff
@@ -0,0 +1,846 @@
...
+    server.run(cli.addr, cli.http_port).await?;
+
+    // Shutdown all tasks
+    tasks.abort_all();
+
+    Ok(())
+}
+
+const NO_CONTENT_CHALLENGE_HEADER: &str = "X-Tailscale-Challenge";
+const NO_CONTENT_RESPONSE_HEADER: &str = "X-Tailscale-Response";
```
flub, line 352: doesn't look like this is terribly important, but we should probably change these headers.

src/hp/magicsock/conn.rs
```diff
@@ -0,0 +1,3379 @@
...
+    #[debug("on_derp_active: Option<Box<..>>")]
+    pub(super) on_derp_active: Option<Box<dyn Fn() + Send + Sync + 'static>>,
+    /// A callback that provides a `cfg::NetInfo` when discovered network conditions change.
+    #[debug("on_net_info: Option<Box<..>>")]
+    on_net_info: Option<Box<dyn Fn(cfg::NetInfo) + Send + Sync + 'static>>,
+
+    /// Used for receiving DERP messages.
+    network_recv_ch: flume::Receiver<NetworkReadResult>,
+    /// Stores wakers, to be called when derp_recv_ch receives new data.
+    network_recv_wakers: std::sync::Mutex<Option<Waker>>,
```
rklaehn, line 175: Not a super fan of the explicit waker management. Alternatives are:

- store the sender and receiver in a mutex and then directly call the poll fns on them which allow passing in the context, see https://github.com/n0-computer/quic-rpc/blob/09628eb5a9a7fa7dd51db168723355d4ee420bfd/src/transport/interprocess.rs#L22
- make a thing that encapsulates the waker mgmt.

Now, this code is not wrong or anything. I just usually have a bad feeling with manual waker management code since it is so easy to get subtly wrong.

src/bin/derper.rs
```diff
@@ -0,0 +1,846 @@
...
+                    .iter()
+                    .map(|(k, v)| (k.parse().unwrap(), v.parse().unwrap())),
+            )
+        } else {
+            Default::default()
+        };
+        s.client_conn_handler(headers)
+    });
+    let tls_config = if serve_tls {
+        let contact = "d@iroh.computer".to_string(); // TODO: configurable.
```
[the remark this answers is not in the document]
dignifiedquire: yes 😅 this should be done, I am not the contact person

src/hp/magicsock/conn.rs
dignifiedquire: For now I would like to leave it as is, as the last time I tried to simplify it I failed and added more bugs. I would love to avoid this, but I think for the moment it is needed.

ramfox (approved):
let's merge this bad boy and make incremental improvements after

PR #832, opened 2023-03-13
Basic persistence

rklaehn:
Persists
- outboards in files named by their hash, in $IROH_DATA_DIR/outboard
- collections in files named by their hash, in $IROH_DATA_DIR/collections
- metadata in $IROH_DATA_DIR/paths.bin
- keypair in $IROH_DATA_DIR/keypair unless explicitly passed via key

Persistence is manual for now, gets triggered on control-c.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ data-encoding = { version = "2.3.3", optional = true }
 default-net = "0.13.0"
 der = { version = "0.6", features = ["alloc", "derive"] }
 derive_more = "0.99.17"
+dirs-next = "2.0.0"
 ed25519-dalek = { version = "1.0.1", features = ["serde"] }
 futures = "0.3.25"
 indicatif = { version = "0.17", features = ["tokio"], optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,10 +56,12 @@ proptest = "1.0.0"
 rand = "0.7"
 testdir = "0.7.2"
 regex = { version = "1.7.1", features = ["std"] }
+nix = "0.26.2"
 
 [features]
 default = ["cli"]
 cli = ["clap", "console", "indicatif", "data-encoding", "multibase"]
+test = []
 
 [[bin]]
 name = "iroh"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ derive_more = "0.99.17"
 dirs-next = "2.0.0"
 ed25519-dalek = { version = "1.0.1", features = ["serde"] }
 futures = "0.3.25"
+hex = "0.4.3"
 indicatif = { version = "0.17", features = ["tokio"], optional = true }
 multibase = { version = "0.9.1", optional = true }
 portable-atomic = "1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,7 +52,6 @@ x509-parser = "0.14"
 zeroize = "1.5"
 
 [dev-dependencies]
-hex = "0.4.3"
 proptest = "1.0.0"
 rand = "0.7"
 testdir = "0.7.2"
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,4 +1,3 @@
-use std::io;
 use std::net::{Ipv4Addr, SocketAddrV4};
 use std::time::Duration;
 use std::{fmt, net::SocketAddr, path::PathBuf, str::FromStr};
@@ -11,7 +10,7 @@ use indicatif::{
     HumanBytes, HumanDuration, ProgressBar, ProgressDrawTarget, ProgressState, ProgressStyle,
 };
 use iroh::protocol::AuthToken;
-use iroh::provider::{Database, Provider, Snapshot, Ticket};
+use iroh::provider::{Database, Provider, Ticket};
 use iroh::rpc_protocol::*;
 use iroh::rpc_protocol::{
     ListRequest, ProvideRequest, ProvideResponse, ProvideResponseEntry, ProviderRequest,
@@ -298,13 +297,7 @@ async fn main_impl() -> Result<()> {
             let use_data_root = persistent.unwrap_or_default();
             let iroh_data_root = iroh_data_root()?;
             let db = if use_data_root {
-                tokio::task::block_in_place(|| {
-                    tracing::info!("Loading snapshot from {}...", iroh_data_root.display());
-                    let snapshot = Snapshot::load(&iroh_data_root)?;
-                    tracing::info!("Loading database from snapshot");
-                    let db = Database::from_snapshot(snapshot)?;
-                    anyhow::Ok(db)
-                })?
+                Database::load(&iroh_data_root).await?
             } else {
                 Database::default()
             };
@@ -361,11 +354,7 @@ async fn main_impl() -> Result<()> {
             }
             // persist the db to disk. this is blocking code.
             if use_data_root {
-                tokio::task::block_in_place(|| {
-                    let snapshot = db.snapshot();
-                    snapshot.persist(iroh_data_root)?;
-                    io::Result::Ok(())
-                })?;
+                db.save(&iroh_data_root).await?;
             }
             // the future holds a reference to the temp file, so we need to
             // keep it for as long as the provider is running. The drop(fut)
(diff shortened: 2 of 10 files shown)


src/provider.rs
```diff
@@ -192,16 +223,18 @@ impl Database {
...
-                    hash,
-                    BlobOrCollection::Blob(Data {
-                        outboard: outboard.clone(),
-                        path,
-                        size,
-                    }),
-                );
+        for (hash, size, path) in paths {
+            if let Some(path) = path {
+                if let Some(outboard) = outboards.get(&hash) {
```
dignifiedquire, line 243: could be written as 
```rust
if let (Some(path), Some(outboard)) = (path, outboard) {
  // ..
}
```
to avoid the double nesting

src/provider.rs
```diff
@@ -57,7 +59,248 @@ const HEALTH_POLL_WAIT: Duration = Duration::from_secs(1);
 #[derive(Debug, Clone, Default)]
 pub struct Database(Arc<RwLock<HashMap<Hash, BlobOrCollection>>>);
 
+/// A snapshot of the database.
+///
+/// `E` can be `Infallible` if we take a snapshot from an in memory database,
+/// or `io::Error` if we read a database from disk.
+pub struct Snapshot<E> {
```
dignifiedquire, line 66: Feels like this could be its own file, together with `Database`

src/main.rs
```diff
@@ -397,17 +332,19 @@ async fn main_impl() -> Result<()> {
...
                     println!("Shutting down provider...");
-                    cancel_token.cancel();
+                    provider2.shutdown();
                 }
                 res = provider => {
                     res?;
                 }
             }
+            let snapshot = db.snapshot();
+            snapshot.persist(iroh_data_root)?;
```
dignifiedquire, line 347: these blocking calls should probably be wrapped into `spawn_blocking` or `block_in_place`

src/provider.rs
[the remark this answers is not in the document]
rklaehn: It's just so you can have no error at all when generating a snapshot from a db.

Cargo.toml
```diff
@@ -23,8 +23,10 @@ data-encoding = { version = "2.3.3", optional = true }
 default-net = "0.13.0"
 der = { version = "0.6", features = ["alloc", "derive"] }
 derive_more = "0.99.17"
+dirs-next = "2.0.0"
 ed25519-dalek = { version = "1.0.1", features = ["serde"] }
 futures = "0.3.25"
+hex = "0.4.3"
```
dignifiedquire, line 29: remove from dev-deps if added here

src/main.rs
```diff
@@ -85,11 +93,16 @@ enum Commands {
         #[clap(long)]
         auth_token: Option<String>,
         /// If this path is provided and it exists, the private key is read from this file and used, if it does not exist the private key will be persisted to this location.
+        ///
+        /// If this path is not provided and persistent is true, the private key will be persisted to the iroh data root.
```
dignifiedquire, line 97: This was just a placeholder until we got the config working, now that this adds proper config, we should probably just remove this option and always persist it to iroh data root, like we did in the previous version of iroh.

src/main.rs
```diff
@@ -85,11 +93,16 @@ enum Commands {
...
         #[clap(long)]
         key: Option<PathBuf>,
         /// Optional rpc port, defaults to 4919. Set to 0 to disable RPC.
         #[clap(long, default_value_t = ProviderRpcPort::Enabled(DEFAULT_RPC_PORT))]
         rpc_port: ProviderRpcPort,
+        /// If true, the provider will read and write from the iroh data root to persist data.
+        #[clap(long, default_value = "true")]
+        persistent: Option<bool>,
```
dignifiedquire, line 105: Is this really only needed for tests? This complicates the behaviour quite a bit.

src/main_util.rs
```diff
@@ -0,0 +1,183 @@
...
+/// If the `IROH_DATA_DIR` environment variable is set it will be used unconditionally.
+/// Otherwise the returned value depends on the operating system according to the following
+/// table.
+///
+/// | Platform | Value                                         | Example                                  |
+/// | -------- | --------------------------------------------- | ---------------------------------------- |
+/// | Linux    | `$XDG_DATA_HOME`/iroh or `$HOME`/.local/share/iroh | /home/alice/.local/share/iroh                 |
+/// | macOS    | `$HOME`/Library/Application Support/iroh      | /Users/Alice/Library/Application Support/iroh |
+/// | Windows  | `{FOLDERID_RoamingAppData}/iroh`              | C:\Users\Alice\AppData\Roaming\iroh           |
+#[allow(dead_code)]
```
dignifiedquire, line 53: this shouldn't be dead code?

tests/cli.rs
```diff
@@ -122,7 +123,8 @@ fn test_provide_get_loop(path: &Path, input: Input, output: Output) -> Result<()
             .arg("--addr")
             .arg(ADDR)
             .arg("--rpc-port")
-            .arg("0")
+            .arg("disabled")
+            .arg("--persistent=false")
```
dignifiedquire, line 127: what about adding a test for the actual persistence being enabled?

src/main.rs
rklaehn: I think it is nice to have the option to use iroh without touching some common dir.

I remember running `IPFS_PATH=$HOME/.public_ipfs ipfs` and similar things to have ipfs leave my ~/.ipfs alone.

I had briefly considered to split it up so that provide is always just a oneshot thing without persistence, and there is a new `start` subcommand that always uses persistence.

src/main.rs
rklaehn: So key is always read from iroh home? Fine by me, but that will complicate one shot usage, see below.

src/main_util.rs
rklaehn: It isn't anymore.

tests/cli.rs
rklaehn: will do.

src/main.rs
dignifiedquire: But you can do that with `IROH_PATH` as well now, can't you? So `IRO_PATH=/tmp/foo` iroh would do the same temporary thing

rklaehn: Sure. The question is if we want to provide an easy way to use iroh provide in a "oneshot" way without any persistence.

ramfox: So the question of persistence here actually centers around the interaction between `key`, identity, and network information. Because we will likely be persisting peer and network information, it would be beneficial for the user to use that information when they provide or fetch, regardless of whether they want to keep the collection they are providing or fetching. 

I think that the story of persisting during fetch and provide will actually come into play when we do garbage collecting / naming / tagging rather than the act of loading up the database itself. But I agree that we will definitely want this "easy path" option.

So I would vote for:
- remove `key` and only use a new identify/persistence directory when the `IROH_PATH` is different
- eventually, have some flag or mechanism for "not persisting" the provided or fetched thing, but for now keep the default of 'always persist'.

tests/cli.rs
```diff
@@ -47,6 +47,70 @@ fn cli_provide_from_stdin_to_stdout() -> Result<()> {
     test_provide_get_loop(&path, Input::Stdin, Output::Stdout)
 }
 
+#[cfg(unix)]
+#[tokio::test]
+async fn cli_provide_persistence() -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
dignifiedquire: There is an tokio compatible version of `Command` as well as the `fs` stuff, so that should make things cleaner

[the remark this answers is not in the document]
rklaehn: Yeah, the only reason this is async in the first place is that I did not want to put Snapshot in the public API. Made it all sync now.

src/provider/database.rs
```diff
@@ -185,32 +185,48 @@ where
 }
 
 impl Database {
+    /// Load a database from disk for testing. Synchonous.
```
dignifiedquire, line 188: typo

dignifiedquire (approved):

dignifiedquire (dismissed):

dignifiedquire (approved):

PR #845, opened 2023-03-14
Allow nested directories while still keeping the flat collection / blob hierarchy

rklaehn:


src/provider/mod.rs
```diff
@@ -405,21 +405,23 @@ impl RpcHandler {
...
-            vec![path.into()]
-        } else {
-            anyhow::bail!("path must be either a Directory or a File");
-        };
+        let root = msg.path;
+        anyhow::ensure!(
+            root.is_dir() || root.is_file(),
+            "path must be either a Directory or a File"
+        );
+        let files = read_dir_recursive(&root)?;
```
dignifiedquire, line 413: lets make this async? or at least call `spawn_blocking`

dignifiedquire:
Please add a test adding and retrieving a nested directory.

src/provider/mod.rs
rklaehn: I guess I should.

But I am getting a bit confused with all these branches building on persistence. I think we should merge that first.

rklaehn: Stupid of me to not just use WalkDir. Have changed it.

rklaehn:
@flub one thing this changes it that it enforces that a collection has only one item per name. I assume that is fine, but just want to make sure this does not cause any trouble in the deltachat use case.

src/main.rs
```diff
@@ -557,11 +557,11 @@ async fn get_interactive(
         let pb = &pb;
         async move {
             let name = if name.is_empty() {
-                hash.to_string()
+                PathBuf::new().join(hash.to_string())
```
flub, line 560: This is a rather strange way to spell `PathBuf::from(hash.to_string())` no?

src/util.rs
```diff
@@ -197,6 +198,32 @@ pub(crate) fn validate_bao(
...
+            anyhow::ensure!(
+                !c.contains('/') && !c.contains('\\'),
+                "invalid path component {:?}",
+                c
+            );
+            Ok(c)
+        })
+        .collect::<anyhow::Result<Vec<_>>>()?;
+    Ok(parts.join("/"))
+}
```
flub, line 225: There are now two disparate utilities to handle this, this one and `pathbuf_from_name`.  I wonder if it's better to add a `BlobNameWithPathComponents` (or whatever bikeshedding produces a better name) which does this cannoicalisation on `::new()` and has a `to_pathbuf()` method.

flub:
> @flub one thing this changes it that it enforces that a collection has only one item per name. I assume that is fine, but just want to make sure this does not cause any trouble in the deltachat use case.

yes, that's fine.  all blob names come from file names in a flat directory in deltachat.  I do prefix all names with `blob/` however, so file names are `blob/foo`, `blob/bar` etc.  all this canonicalisation is on the application side of things and not the lib right?

dignifiedquire (approved):

Cargo.toml
```diff
@@ -48,6 +48,7 @@ tokio-util = { version = "0.7", features = ["io-util", "io"] }
 tracing = "0.1"
 tracing-futures = "0.2.5"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
+walkdir = "2.3.2"
```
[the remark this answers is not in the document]
rklaehn: Right. Just used cargo add tbh.

src/provider/mod.rs
```diff
@@ -983,20 +1002,29 @@ async fn create_collection_inner(
...
+    let outboards = stream::iter(data_sources)
+        .map(|data| {
+            let (path, name) = match data {
+                DataSource::File(path) => (path, None),
+                DataSource::NamedFile { path, name } => (path, Some(name)),
+            };
+            tokio::task::spawn_blocking(move || compute_outboard(path, name))
+        })
+        // allow at most num_cpus tasks at a time, otherwise we might get too many open files
+        .buffer_unordered(num_cpus::get());
```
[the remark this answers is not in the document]
rklaehn: Yes, this assumes that the task is fully CPU bound, which is probably not the case.

But for all systems where num_cpus < max open files, this conveniently limits the number of open files, so I guess it will have to do for now. I will add a todo.

rklaehn:
> > @flub one thing this changes it that it enforces that a collection has only one item per name. I assume that is fine, but just want to make sure this does not cause any trouble in the deltachat use case.
> 
> yes, that's fine. all blob names come from file names in a flat directory in deltachat. I do prefix all names with `blob/` however, so file names are `blob/foo`, `blob/bar` etc. all this canonicalisation is on the application side of things and not the lib right?

Yes. The only place where / is interpreted as a directory separator is in main. We might want to add this to the lib in the future, since the way it is now the lib does not help you in creating a directory tree. But for now this is completely separate.

Arqu (approved):

flub (approved):

PR #846, opened 2023-03-15
impl: `Client` and `Clients`, how the server manages the different connected clients

ramfox:
Decided I needed to super simplify this, for now. This impl no longer attempts to handle muliple connections with the same key or track if a client is connected to multiple derp servers. 

Both situations are apparently not that common, but will be important to implement in the future in order to capture all situations.

This is the last large piece before turning to the actual server.

### DIFF
--- a/src/hp/derp/clients.rs
+++ b/src/hp/derp/clients.rs
@@ -1,0 +1,345 @@
+//! based on tailscale/derp/derp_server.go
+//!
+//! The "Server" side of the client. Uses the `ClientConnManager`.
+use crate::hp::key::node::PublicKey;
+use std::collections::HashMap;
+
+use tokio::sync::mpsc;
+
+use super::{
+    client_conn::{ClientConnManager, Packet, PeerConnState},
+    conn::Conn,
+};
+
+/// Number of times we try to send to a client connection before dropping the data;
+const RETRIES: usize = 3;
+
+/// Represents a connection to a client.
+///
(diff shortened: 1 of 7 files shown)


src/hp/derp/clients.rs
```diff
@@ -0,0 +1,345 @@
...
+{
+    pub fn new(conn: ClientConnManager<C>) -> Self {
+        Self {
+            preferred: None,
+            conn,
+        }
+    }
+
+    pub async fn shutdown(self) {
+        self.conn.shutdown().await;
```
ramfox, line 61: Ok, so I originally had these shutdowns happen in a spawned task & left to run on its own, without blocking.

I thought it might be better to have those shutdown without blocking, since these will be called in the main server loop, meaning we could be waiting for the connection to shutdown instead of picking up the next task of work in the loop.

However, if we do that, we won't ever have a way to know if we've actually cleaned up all of our spawned tasks... so I made them async again, and we wait until the shutdown finishes before moving on.

Not sure if that's the correct trade-off, though, so would like some feedback.

src/hp/derp/clients.rs
```diff
@@ -0,0 +1,345 @@
+//! based on tailscale/derp/derp_server.go
```
dignifiedquire, line 1: wrong comment?

dignifiedquire (approved):

PR #847, opened 2023-03-15
remove DataSource::NamedFile

rklaehn:
This is used nowhere in the code base except in tests, and it complicates some things quite a bit.

Can somebody tell me why this is a bad idea? Is DataSource::NamedFile used in deltachat?


dignifiedquire:
I believe @flub introduced this and uses it in deltachat, yes

PR #852, opened 2023-03-15
Add rpc command to validate the store contents

rklaehn:
Needed becuause I broke my store. I guess other people will do as well. It is easy to do with the files not being readonly.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -11,6 +11,7 @@ use indicatif::{
 };
 use iroh::protocol::AuthToken;
 use iroh::provider::{Database, Provider, Snapshot, Ticket};
+use iroh::rpc_protocol::*;
 use iroh::rpc_protocol::{
     ListRequest, ProvideRequest, ProvideResponse, ProvideResponseEntry, ProviderRequest,
     ProviderResponse, ProviderService, VersionRequest,
@@ -99,6 +100,24 @@ enum Commands {
     /// List hashes
     #[clap(about = "List hashes")]
     List {
+        /// Optional rpc port, defaults to 4919
+        #[clap(long, default_value_t = DEFAULT_RPC_PORT)]
+        rpc_port: u16,
+    },
+    /// Shutdown
+    #[clap(about = "Shutdown provider")]
+    Shutdown {
+        /// Shutdown mode.
+        /// Hard shutdown will immediately terminate the process, soft shutdown will wait for all connections to close.
+        #[clap(long, default_value_t = false)]
+        force: bool,
+        /// Optional rpc port, defaults to 4919
+        #[clap(long, default_value_t = DEFAULT_RPC_PORT)]
+        rpc_port: u16,
+    },
+    /// Identity
+    #[clap(about = "Identify provider")]
+    Id {
         /// Optional rpc port, defaults to 4919
         #[clap(long, default_value_t = DEFAULT_RPC_PORT)]
         rpc_port: u16,
@@ -275,7 +294,15 @@ async fn main_impl() -> Result<()> {
             tracing::info!("Loading database from snapshot");
             let db = Database::from_snapshot(snapshot)?;
 
-            let provider = provide(db, addr, auth_token, key, cli.keylog, rpc_port.into()).await?;
+            let provider = provide(
+                db.clone(),
+                addr,
+                auth_token,
+                key,
+                cli.keylog,
+                rpc_port.into(),
+            )
+            .await?;
             let controller = provider.controller();
             let mut ticket = provider.ticket(Hash::from([0u8; 32]));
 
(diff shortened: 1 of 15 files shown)


src/main.rs
```diff
@@ -392,6 +399,23 @@ async fn main_impl() -> Result<()> {
...
+            while let Some(item) = response.next().await {
+                let item = item?;
+                println!(
+                    "{} {} {} {}",
+                    item.path
+                        .map(|x| x.display().to_string())
+                        .unwrap_or("Collection".to_owned()),
+                    Blake3Cid(item.hash),
+                    HumanBytes(item.size),
+                    item.error.unwrap_or("OK".to_owned()),
```
dignifiedquire, line 414: Would be nice to have like a checkmark and a cross for each state

src/provider/mod.rs
```diff
@@ -404,6 +425,23 @@ impl RpcHandler {
...
     }
+
+    /// Invoke validate on the database and stream out the result
+    fn validate(
+        self,
+        _msg: ValidateRequest,
+    ) -> impl Stream<Item = ValidateResponse> + Send + 'static {
+        self.inner
+            .db
+            .validate(4)
```
dignifiedquire, line 436: constant, to make clear what this number means

src/util.rs
```diff
@@ -152,6 +154,46 @@ impl From<anyhow::Error> for RpcError {
...
+) -> result::Result<(), BaoValidationError> {
+    let hash = blake3::Hash::from(hash);
+    let outboard_reader = io::Cursor::new(outboard);
+    let mut decoder = abao::decode::Decoder::new_outboard(data_reader, outboard_reader, &hash);
+    // todo: expose chunk group size in abao, so people can allocate good sized buffers
+    let mut buffer = [0u8; 1024 * 16 + 4096];
+    loop {
+        match decoder.read(&mut buffer) {
+            Ok(0) => break,
+            Ok(_) => {}
```
dignifiedquire, line 187: shouldn't you make sure the whole thing gets read, by summing the read amount?

src/util.rs
```diff
@@ -152,6 +154,46 @@ impl From<anyhow::Error> for RpcError {
...
+pub(crate) fn validate_bao(
+    hash: Hash,
+    data_reader: impl Read + Seek,
+    outboard: Bytes,
+) -> result::Result<(), BaoValidationError> {
+    let hash = blake3::Hash::from(hash);
+    let outboard_reader = io::Cursor::new(outboard);
+    let mut decoder = abao::decode::Decoder::new_outboard(data_reader, outboard_reader, &hash);
+    // todo: expose chunk group size in abao, so people can allocate good sized buffers
+    let mut buffer = [0u8; 1024 * 16 + 4096];
```
dignifiedquire, line 183: this is quite large for a stack array, probably makes more sense to use a Vec or BytesMut

src/main.rs
```diff
@@ -392,6 +399,23 @@ async fn main_impl() -> Result<()> {
             }
             Ok(())
         }
+        Commands::Validate { rpc_port } => {
+            let client = make_rpc_client(rpc_port).await?;
+            let mut response = client.server_streaming(ValidateRequest).await?;
+            while let Some(item) = response.next().await {
```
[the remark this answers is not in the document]
rklaehn: Yes, I think this needs the same treatment as the progress branch. But maybe in the next PR?

src/provider/mod.rs
rklaehn: Was meant to replace this with `num_cpus::get()`.

src/util.rs
rklaehn: It is fine if less than the buffer is being read. The validation happens in the decoder. We just want to stop when read returns 0, which means EOF.

src/main.rs
rklaehn: You got it...

image

rklaehn:
Other than showing progress, this is ready to go.

I would do showing progress together with https://github.com/n0-computer/iroh/pull/853 since it is essentially the same thing.

src/provider/mod.rs
dignifiedquire: not replaced yet

src/util.rs
```diff
@@ -152,6 +154,46 @@ impl From<anyhow::Error> for RpcError {
 /// A serializable result type for use in RPC responses.
 pub type RpcResult<T> = result::Result<T, RpcError>;
 
+#[derive(Debug, Display, Error)]
+pub(crate) enum BaoValidationError {
+    /// Generic io error. We were unable to read the data.
+    IoError(io::Error),
+    // /// The hash of the data does not match the hash of the outboard.
+    // HashMismatch,
```
dignifiedquire, line 162: Why are these commented out/not yet used?

src/provider/database.rs
```diff
@@ -271,6 +275,58 @@ impl Database {
...
     }
 
+    /// Validate the entire database, including collections.
+    ///
+    /// This works by taking a snapshot of the database, and then validating. So anything you add after this call will not be validated.
+    pub(crate) fn validate(
+        &self,
+        parallelism: usize,
+    ) -> impl Stream<Item = (Hash, u64, Option<PathBuf>, Option<BaoValidationError>)> {
+        let mut data = self
```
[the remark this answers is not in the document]
rklaehn: I think it is fine. The DB is mostly bytes::Bytes, and cloning those is basically cloning an Arc.

But will add a comment.

src/util.rs
rklaehn: I want the validation errors to be a bit more informative in the future. E.g. it should say at what offset a hash mismatch occurs. Currently it just passes through the errors from the bao validation.

But I can remove the commented out stuff if you want.

dignifiedquire (approved):

PR #853, opened 2023-03-15
Progress when adding files

rklaehn:
So you don't stare at nothing for hours when adding a big dir like some ml models.


rklaehn:
The second attempt of this has been merged, so this is stale.

PR #854, opened 2023-03-16
impl derp `ServerActor`

ramfox:
impl the ServerActor, that does all the work of coordinating traffic on the server, including sending packets, forwarding packets, adding & removing clients, adding & removing packet forwarders, adding "watchers" (verified peers in the mesh that need to be kept informed about the topology of the network),  and shutting down the server.

### DIFF
--- a/src/hp/derp/clients.rs
+++ b/src/hp/derp/clients.rs
@@ -4,9 +4,9 @@
 use crate::hp::key::node::PublicKey;
 use std::collections::{HashMap, HashSet};
 
+use futures::future::join_all;
 use tokio::io::{AsyncRead, AsyncWrite};
 use tokio::sync::mpsc;
-use tokio::task::JoinSet;
 
 use super::{
     client_conn::{ClientBuilder, ClientConnManager},
@@ -159,11 +159,11 @@ where
     }
 
     pub async fn shutdown(&mut self) {
-        let mut set = JoinSet::new();
+        let mut handles = Vec::new();
         for (_, client) in self.inner.drain() {
-            set.spawn(async move { client.shutdown_await().await });
+            handles.push(tokio::spawn(async move { client.shutdown_await().await }));
         }
-        while let Some(_) = set.join_next().await {}
+        join_all(handles).await;
     }
 
     pub fn close_conn(&mut self, key: &PublicKey) {
@@ -265,7 +265,6 @@ where
         tracing::warn!("Could not find client for {key:?}, dropping packet");
     }
 
-    // TODO: send Ok or bail
     fn process_result(
         &mut self,
         key: &PublicKey,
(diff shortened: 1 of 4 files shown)


src/hp/derp/server.rs
```diff
@@ -368,26 +305,192 @@ pub(crate) struct ServerInfo {
...
+            client_mesh: HashMap::default(),
+            watchers: HashSet::default(),
+        }
+    }
+
+    pub(crate) async fn run(mut self, done: CancellationToken) -> Result<()> {
+        loop {
+            tokio::select! {
+                biased;
+                _ = done.cancelled() => {
```
rklaehn, line 352: This is a hard shutdown. It's good to have it, but if you also want a soft shutdown you can just make another normal message. That way you process all incoming msgs up to the shutdown msg.

src/hp/derp/server.rs
```diff
@@ -368,26 +305,192 @@ pub(crate) struct ServerInfo {
 use super::client_conn::ClientBuilder;
 // TODO: need a types module
 use super::client_conn::Packet;
+use super::client_conn::PeerConnState;
+
+use tokio::sync::mpsc;
+use tokio_util::sync::CancellationToken;
+
+pub(crate) struct ServerActor<C, R, W, P>
+where
```
rklaehn, line 314: Not sure if it is worth it here, but you can DRY these bounds using a "module" trait.

```rust
trait ServerTypes {
  type C: Conn,
  type R: AsyncRead + Unpin + Send + Sync + 'static,
  type W: AsyncWrite + Unpin + Send + Sync + 'static,
  type P: PacketForwarder
}

pub(crate) struct ServerActor<ST: ServerTypes>
```

Not sure if it is worth it here, but if you get more types and more constraints it would definitely be worth it.

src/hp/derp/clients.rs
```diff
@@ -134,9 +158,42 @@ where
         }
     }
 
-    pub fn shutdown(&mut self) {
+    pub async fn shutdown(&mut self) {
+        let mut set = JoinSet::new();
         for (_, client) in self.inner.drain() {
-            client.shutdown();
+            set.spawn(async move { client.shutdown_await().await });
```
dignifiedquire, line 164: sth like https://docs.rs/futures/latest/futures/future/fn.join_all.html seems more fitting here

src/hp/derp/clients.rs
```diff
@@ -159,48 +216,63 @@ where
...
         if let Some(client) = self.inner.get(key) {
             let res = client.send_mesh_updates(updates);
-            self.process_result(key, res);
+            let _ = self.process_result(key, res);
         };
         tracing::warn!("Could not find client for {key:?}, dropping packet");
     }
 
-    fn process_result(&mut self, key: &PublicKey, res: Result<(), SendError>) {
+    // TODO: send Ok or bail
```
dignifiedquire, line 268: seems fiexed?

PR #857, opened 2023-03-16
feat: create a CollectionBuilder

dignifiedquire:


src/provider.rs
```diff
@@ -903,95 +893,135 @@ fn compute_outboard(
...
+    async fn build_into_with_entries(
+        self,
+        db: &mut HashMap<Hash, BlobOrCollection>,
+    ) -> Result<(Vec<ProvideResponseEntry>, Hash)> {
+        // +1 is for the collection itself
+        let mut blobs = Vec::with_capacity(self.sources.len());
+        let mut total_blobs_size: u64 = 0;
+
+        // compute outboards in parallel, using tokio's blocking thread pool
+        let outboards = self.sources.into_iter().map(|data| {
```
rklaehn, line 951: Not related to the builder refactor, but this code can fail if you open too many files concurrently.

I changed this into a stream and then used `.buffer_unordered(num_cpus.get())` in some branch to fix this.

src/provider.rs
```diff
@@ -903,95 +893,135 @@ fn compute_outboard(
...
+            let name = name.unwrap_or_else(|| {
+                path.file_name()
+                    .and_then(|s| s.to_str())
+                    .unwrap_or_default()
+                    .to_string()
+            });
+            blobs.push(Blob { name, hash });
+        }
+
+        let c = Collection {
```
rklaehn, line 1002: I think it might be nice to canonicalize the collection here by calling sort on the blobs.

Not that we want to get overly religious regarding canonicalisation, but it helps a bit with tests, and if it is basically free why not do it.

dignifiedquire:
closing for now

dignifiedquire:
I did too, we can revive it at some point, but the async/sync setup we currrently have made this rebase quite problematic on latest `main`.

PR #861, opened 2023-03-19
[WIP] feat: compress blobs on transfer

### LINKED ISSUE #739: Use lz4 compression
When sending files that are not already somehow encoded, there would be a large gain 

There is a fully rust compression library called lz4_flex. https://crates.io/crates/lz4_flex

We could use this to reduce the amount of data that needs to be sent through en/decryption and through quic. Even for uncompressible data lz4 is very fast, so you probably would not even notice it WRT perf compared to the quic overhead.

The downside is the additional complexity. We should try this out only after having implemented/used larger bao block size. See n0-computer/iroh#741

Edit: zstd provides much better compression rates but is much more expensive. Also, there is no pure rust impl of zstd, unfortunately...

dignifiedquire:
Currently broken, likely due to a bug somewhere in the asnycread/write impls of async-compression or abao

Closes #739

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -40,6 +40,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "alloc-no-stdlib"
+version = "2.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cc7bb162ec39d46ab1ca8c77bf72e890535becd1751bb45f64c597edb4c8c6b3"
+
+[[package]]
+name = "alloc-stdlib"
+version = "0.2.2"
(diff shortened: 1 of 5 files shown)


dignifiedquire:
Looks like our usage is simply not possible currently: https://github.com/Nemo157/async-compression/issues/141

flub:
you could write a custom inner writer that doesn't pass through the closing...

rklaehn:
Can somebody enlighten me why we want to compress individual files instead of the entire stream?

rklaehn:
If we do this we should do the version where the entire request is compressed as one, seee #864

dignifiedquire:
meh, closing this one

PR #864, opened 2023-03-20
Compression 2 - compress the whole request stream

rklaehn:
As an alternative approach to https://github.com/n0-computer/iroh/pull/861


rklaehn:
If we ever need this we still have the branch, but I am going to close the PR since we are probably not going to do this any time soon.

PR #866, opened 2023-03-21
impl derp Server

### LINKED ISSUE #833: Derp server impl
- [x] struct `ClientSet`
- [x] struct `Clients`
- [x] serve loop
- [x] accept (and accept server/client handshake)
- [x] mesh peers (watchers)
- [x] shutting down clients
- [x] test relay: spin up two clients, manually connect them to the server, send packets!

ramfox:
closes #833

### DIFF
--- a/src/hp/derp/server.rs
+++ b/src/hp/derp/server.rs
@@ -1,19 +1,27 @@
 //! based on tailscale/derp/derp_server.go
 use std::collections::{HashMap, HashSet};
-use std::net::SocketAddr;
 use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
 use std::time::Duration;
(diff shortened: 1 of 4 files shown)


src/hp/derp/server.rs
```diff
@@ -453,14 +508,104 @@ where
...
+    if a.is_none() {
+        false
+    } else if b.is_none() {
+        false
+    } else {
+        let mesh_a = a.unwrap();
+        let mesh_b = b.unwrap();
+        mesh_a == mesh_b
+    }
+}
```
dignifiedquire, line 599: can be written more rusty as

```rust
it let (Some(a), Some(b)) = (a, b) {
  return a == b;
}

false
```

PR #867, opened 2023-03-21
Remove collection name

rklaehn:
And a bunch of other things

- make collection fields private so that creation has to be via a constructor
- make docs creation use stable toolchain


Arqu (approved):
LGTM, thanks for tagging in, I used the opportunity to fix up netsim ahead of this not to fail on parsing.

PR #868, opened 2023-03-21
Canonicalize collection

rklaehn:
This makes a few things easier in testing. Had issues with collection order in the directory hierarchy PR.

Did this as a separate PR so we can discuss it in isolation, but I don't see much downside.

This does not mean that we suddenly get religious about canonicalization of hashes, btw...

### DIFF
--- a/src/blobs.rs
+++ b/src/blobs.rs
@@ -8,17 +8,34 @@ use crate::util::Hash;
 #[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
 pub struct Collection {
     /// Links to the blobs in this collection
-    pub(crate) blobs: Vec<Blob>,
+    blobs: Vec<Blob>,
     /// The total size of the raw_data referred to by all links
-    pub(crate) total_blobs_size: u64,
+    total_blobs_size: u64,
 }
 
 impl Collection {
+    pub(crate) fn new(blobs: Vec<Blob>, total_blobs_size: u64) -> Self {
+        Self {
+            blobs,
+            total_blobs_size,
+        }
+    }
+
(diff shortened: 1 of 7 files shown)


src/blobs.rs
```diff
@@ -15,6 +15,11 @@ pub struct Collection {
 
 impl Collection {
     pub(crate) fn new(blobs: Vec<Blob>, total_blobs_size: u64) -> Self {
+        let mut blobs = blobs;
+        let n = blobs.len();
+        blobs.sort_by(|a, b| a.name.cmp(&b.name));
+        blobs.dedup_by(|a, b| a.name == b.name);
+        debug_assert_eq!(n, blobs.len());
```
dignifiedquire, line 22: what if I have two blobs of the same name?

rklaehn: I don't want to support that. Do you?

rklaehn: OK, I guess it would not too bad. It would make this basically a `BTreeMap<String, BTreeSet<Hash>>`. But still, do you have a specific use case in mind?

dignifiedquire: no, I just don’t want inconsistent panics 😅 if we don’t want duplicates I would expect this to error with a matching error message and not panic

rklaehn: Ah, ok. That is reasonable. I think it is ensured upward of this code that names are distinct, hence the assert! and not returning anyhow::Result.

rklaehn: Alright, implemented this as a runtime check in the deep dirs branch...

rklaehn:
Implemented in #845

PR #871, opened 2023-03-22
Implementation of bin/derper

dignifiedquire:
cc @ramfox


dignifiedquire:
basics are there, good enough to merge into the branch, so we can hook things up once they are there

PR #873, opened 2023-03-22
Progress for adding files

rklaehn:
Use an indicativ multiprogress bar.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,7 @@ ssh-key = { version = "0.5.1", features = ["ed25519", "std", "rand_core"] }
 tempfile = "3.4"
 thiserror = "1"
 tokio = { version = "1", features = ["full"] }
+tokio-stream = "0.1.12"
 tokio-util = { version = "0.7", features = ["io-util", "io"] }
 tracing = "0.1"
 tracing-futures = "0.2.5"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,6 @@ rust-version = "1.63"
 [dependencies]
 abao = { version = "0.2.0", features = ["group_size_16k", "tokio_io"], default-features = false }
 anyhow = { version = "1", features = ["backtrace"] }
-async-stream = "0.3.4"
 base64 = "0.21.0"
 blake3 = "1.3.3"
 bytes = "1"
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -44,6 +44,7 @@ ssh-key = { version = "0.5.1", features = ["ed25519", "std", "rand_core"] }
 tempfile = "3.4"
 thiserror = "1"
 tokio = { version = "1", features = ["full"] }
+tokio-stream = "0.1.12"
```
flub, line 47: is `0.1` enough?

src/main.rs
```diff
@@ -201,6 +202,125 @@ async fn make_rpc_client(
...
+    pub size: u64,
+}
+
+async fn aggregate_add_response(
+    stream: impl Stream<
+            Item = std::result::Result<
+                ProvideProgress,
+                impl std::error::Error + Send + Sync + 'static,
+            >,
+        > + Unpin,
```
flub, line 262: nested impl traits!  could consider moving this to a where clause

src/main.rs
```diff
@@ -201,6 +202,125 @@ async fn make_rpc_client(
...
+) -> anyhow::Result<(Hash, Vec<ProvideResponseEntry>)> {
+    let mut stream = stream;
```
flub, line 264: you can't declare this mut in the signature?

src/util.rs
```diff
@@ -242,3 +244,67 @@ mod tests {
         assert_eq!(canonicalize_path("foo/bar").unwrap(), "foo/bar");
     }
 }
+
+pub(crate) struct ProgressReader<R, F: Fn(ProgressReaderUpdate)> {
```
flub, line 248: i assume you looked at `crate::progress::ProgressEmitter` and didn't find it suitable?

flub (approved):

src/main.rs
rklaehn: I can, but I don't want to. I find that the mut is entirely an implementation detail and does not belong in the signature.

flub: putting `mut` in the "signature" has no effect on the caller at all.  It's a weirdness of Rust syntax where putting `mut` there or not has no effect on the signature at all, it is just something local to the function body.

So you're saying you so strongly disagree with Rust's syntax choices that you prefer to do manual work arounds :smile:  While I agree it's a bit odd, I don't tend to agree with fighting stuff like that.  Once you know how this works I think it's perfectly fine.

PR #874, opened 2023-03-22
Validate progress

rklaehn:
Also give validate some progress niceness...

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,3 +1,4 @@
+use std::collections::BTreeMap;
 use std::net::{Ipv4Addr, SocketAddrV4};
 use std::time::Duration;
 use std::{fmt, net::SocketAddr, path::PathBuf, str::FromStr};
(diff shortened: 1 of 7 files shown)


src/main.rs
```diff
@@ -247,6 +249,111 @@ impl ProgressBarState {
...
+                pb.set_style(ProgressStyle::default_bar().template("{msg}").unwrap());
+                pb.finish_with_message(msg);
+            }
+            //     let mp = self.mp.clone();
+            //     let pb = pb.clone();
+            //     tokio::spawn(async move {
+            //         tokio::time::sleep(Duration::from_millis(500)).await;
+            //         mp.remove(&pb);
+            //     });
+            // }
```
flub, line 350: ?

src/main.rs
```diff
@@ -247,6 +249,111 @@ impl ProgressBarState {
...
+        }
+    }
+
+    fn all_done(self) {}
```
flub, line 354: ?

flub (dismissed):

src/main.rs
rklaehn: Ah well, I tried keeping the successful ones for a second, but it flickered...

src/main.rs
rklaehn: Turned out that I did not want to do anything on all_done...

flub (approved):

PR #881, opened 2023-03-23
fix: netsim bin paths

Arqu:


.github/workflows/netsim.yml
```diff
@@ -71,13 +71,6 @@ jobs:
         fi
         echo "'$NETSIM_CONFIG'"
 
-    - name: Run tests
-      run: |
-        cd chuck/netsim
-        sudo kill -9 $(pgrep ovs)
-        sudo mn --clean
-        sudo python3 main.py sims/standard
-
```
flub, line 80: not clear if this was intentional from the PR title

flub (approved):

.github/workflows/netsim.yml
Arqu: Sorry, yeah just accidental cleanup while at it.

PR #884, opened 2023-03-24
ci: replace actions-rs with dtolnay/rust-toolchain

### LINKED ISSUE #878: ci: remove actions-rs from all jobs
`actions-rs` is unmaintained and we've had issues with it already. Should just convert all instances to use `rustup` and `cargo` directly.

bingcicle:
closes #878 

dtolnay's [rust-toolchain](https://github.com/dtolnay/rust-toolchain) is a nicer, more ergonomic way of installing and using the rust toolchain in the CI - projects within [the issue discussing the unmaintained actions-rs](https://github.com/actions-rs/toolchain/issues/216) already have a few projects switch over to using it.

I ran the CI manually on my own repo, and you can check the results [here](https://github.com/bingcicle/iroh/actions/runs/4509634140)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,23 +28,16 @@ jobs:
         submodules: recursive
 
     - name: Install ${{ matrix.rust }}
-      uses: actions-rs/toolchain@v1
+      uses: dtolnay/rust-toolchain@master
       with:
         toolchain: ${{ matrix.rust }}
-        override: true
-
+      
     - name: check
-      uses: actions-rs/cargo@v1
-      with:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -203,11 +185,8 @@ jobs:
         submodules: recursive
 
     - name: Install nightly
```
bingcicle, line 187: this naming doesn't seem right(?), but I left it as I found it. Let me know if this needs changing.

Arqu (approved):
Much appreciated. Thank you.

PR #888, opened 2023-03-27
feat: Add run_ticket to dial all addresses stored in a Ticket

flub:
This adds a new iroh::get::run_ticket call which will dial all the
addresses in a ticket concurrently, up to a specified concurrency
limit.  The first established connection is returned.

This allows much better fetching of a ticket, without having to know
how it is constructed and speeds up connecting.

When dialing an attempt is made to first try addresses which are
considered to be on the same subnet as any local address.

This doesn't yet update main.rs to use this.  Probably best as a
separate PR.

### DIFF
--- a/src/get.rs
+++ b/src/get.rs
@@ -172,19 +172,12 @@ async fn dial_ticket(
 ) -> Result<quinn::Connection> {
     // Sort the interfaces to make sure local ones are at the front of the list.
     let interfaces = default_net::get_interfaces();
-    let mut addrs: Vec<SocketAddr> = ticket
+    let (mut addrs, other_addrs) = ticket
         .addrs
         .iter()
-        .filter(|addr| is_same_subnet(addr, &interfaces))
-        .copied()
-        .collect();
-    let other_addrs: Vec<SocketAddr> = ticket
-        .addrs
-        .iter()
-        .filter(|addr| !addrs.contains(addr))
-        .copied()
-        .collect();
+        .partition::<Vec<_>, _>(|addr| is_same_subnet(addr, &interfaces));
     addrs.extend(other_addrs);
+
     let cancel_token = CancellationToken::new();
     let (results_tx, mut results_rx) = mpsc::unbounded_channel();
     tokio::spawn(dial_all_task(
--- a/src/get.rs
+++ b/src/get.rs
@@ -20,11 +20,9 @@ use abao::decode::AsyncSliceDecoder;
 use anyhow::{anyhow, bail, Context, Result};
 use bytes::BytesMut;
 use default_net::Interface;
-use futures::Future;
+use futures::{Future, StreamExt};
 use postcard::experimental::max_size::MaxSize;
 use tokio::io::{AsyncRead, AsyncReadExt, ReadBuf};
-use tokio::sync::{mpsc, OwnedSemaphorePermit, Semaphore};
-use tokio_util::sync::CancellationToken;
 use tracing::{debug, error};
 
 pub use crate::util::Hash;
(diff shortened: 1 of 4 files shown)


src/get.rs
```diff
@@ -129,14 +134,171 @@ impl AsyncRead for DataStream {
...
+    .await
+}
+
+async fn dial_ticket(
+    ticket: &Ticket,
+    keylog: bool,
+    max_concurrent: usize,
+) -> Result<quinn::Connection> {
+    // Sort the interfaces to make sure local ones are at the front of the list.
+    let interfaces = default_net::get_interfaces();
```
dignifiedquire, line 174: Do you have any idea how fast this?

src/get.rs
```diff
@@ -129,14 +134,171 @@ impl AsyncRead for DataStream {
...
+        .iter()
+        .filter(|addr| is_same_subnet(addr, &interfaces))
+        .copied()
+        .collect();
+    let other_addrs: Vec<SocketAddr> = ticket
+        .addrs
+        .iter()
+        .filter(|addr| !addrs.contains(addr))
+        .copied()
+        .collect();
```
dignifiedquire, line 186: you could use https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.partition instead of iterating twice

src/get.rs
```diff
@@ -129,14 +134,171 @@ impl AsyncRead for DataStream {
...
+    addrs.extend(other_addrs);
+    let cancel_token = CancellationToken::new();
+    let (results_tx, mut results_rx) = mpsc::unbounded_channel();
```
dignifiedquire, line 189: no, unbounded channels please

src/get.rs
```diff
@@ -129,14 +134,171 @@ impl AsyncRead for DataStream {
...
+        }
+    }
+    false
+}
+
+/// Task which will dial all addresses concurrently.
+///
+/// Results are sent back to `results_tx`.  Dialing can be interrupted by cancelling the
+/// `cancel_token`.
+async fn dial_all_task(
```
dignifiedquire, line 236: why manually implement this, and not sth like https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.buffer_unordered

src/get.rs
flub: 300-500 microseconds average on my machine, 2-3 miliseconds is the worst I've seen.  afaik this is just some syscalls for all platforms?

src/get.rs
flub: huh, that is a terribly named method.  but it is a joy to use!  TIL, I was indeed surprised how much work this was.

flub:
cargo-deny is fixed in #890

dignifiedquire (dismissed):

dignifiedquire (approved):

src/get.rs
```diff
@@ -129,14 +132,99 @@ impl AsyncRead for DataStream {
     }
 }
 
+/// Gets a collection and all its blobs using a [`Ticket`].
+pub async fn run_ticket<A, B, C, FutA, FutB, FutC>(
+    ticket: &Ticket,
+    keylog: bool,
```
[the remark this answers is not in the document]
flub: Heh, true doc comments for these run functions have some way to go.  I don't really like the functions in the first place, they would be better as builders instead of having all these options.  But an earlier attempt at doing that failed as there was some weirdness I forgot.

Anyway: the keylog boolean enables the rustls [KeyLog](https://docs.rs/rustls/latest/rustls/struct.KeyLogFile.html) functionality.  If you set it to `true` and run the application with `SSLKEYLOGFILE` environment variable set to a filename the TLS pre-master key will be written into that file and can be used by e.g. wireshark to inspect the captured QUIC traffic.

PR #889, opened 2023-03-28
adjust `derp::Client` and start impl on `derp::http::Client`

ramfox:
I'd like to merge this and do the next leg of work in another PR.

The only two parts left are:
- some mechanism for keeping track of pings when they go out & the associated pong when it comes back in. This should be a small amount of work
- the actual connection process, this should be much more work.

### DIFF
--- a/src/hp/derp/client.rs
+++ b/src/hp/derp/client.rs
@@ -1,11 +1,13 @@
 //! based on tailscale/derp/derp_client.go
 use std::net::SocketAddr;
+use std::sync::Arc;
 use std::time::Duration;
 
 use anyhow::{bail, ensure, Context, Result};
 use bytes::{Bytes, BytesMut};
 use tokio::io::{AsyncRead, AsyncWrite, AsyncWriteExt};
 use tokio::sync::mpsc;
+use tokio::sync::Mutex;
 use tokio::task::JoinHandle;
 
 use super::PER_CLIENT_SEND_QUEUE_DEPTH;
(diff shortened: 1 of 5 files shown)


src/hp/derp/http/client.rs
```diff
@@ -1,39 +1,92 @@
...
+struct InnerClient {
+    secret_key: key::node::SecretKey,
+    get_region:
+        Option<Box<dyn Fn() -> BoxFuture<'static, Option<DerpRegion>> + Send + Sync + 'static>>,
+    can_ack_pings: bool,
+    is_preferred: Mutex<bool>,
+    derp_client: Mutex<Option<DerpClient<tokio::net::tcp::OwnedReadHalf>>>,
+    is_closed: AtomicBool,
+    address_family_selector:
+        Option<Box<dyn Fn() -> BoxFuture<'static, bool> + Send + Sync + 'static>>,
```
ramfox, line 58: the `get_region` and `address_family_selector` fns are pretty ugly to look at, but I think maybe it's just fine?

src/hp/derp/client.rs
```diff
@@ -40,25 +70,30 @@ pub struct Client {
...
 
     /// Sends a packet to the node identified by `dstkey`
     ///
     /// Errors if the packet is larger than [`MAX_PACKET_SIZE`]
     // TODO: the rate limiter is only on this method, is it because it's the only method that
     // theoretically sends a bunch of data, or is it an oversight? For example, the `forward_packet` method does not have a rate limiter, but _does_ have a timeout.
-    pub async fn send(&mut self, dstkey: PublicKey, packet: Bytes) -> Result<()> {
-        self.writer_channel
+    pub async fn send(&self, dstkey: PublicKey, packet: Vec<u8>) -> Result<()> {
+        let mut buf = BytesMut::new();
```
dignifiedquire, line 94: unused?

src/hp/derp/http/client.rs
dignifiedquire: yeah, this is the way to do it unfortunately

PR #891, opened 2023-03-28
ref: Clean up public API regarding bound/local/listening addresses

flub:
This primarily renames .available_addresses() to
.listen_addresses() and .listen_addr() to .local_address().  this
makes .local_address() consistent with e.g. quinn's API and avoids the
confusion between available on the machine vs uses by the provider.

### DIFF
--- a/src/net.rs
+++ b/src/net.rs
@@ -2,6 +2,8 @@
 
 use std::net::{IpAddr, Ipv6Addr, SocketAddr};
 
+use anyhow::{ensure, Result};
+
 const IFF_UP: u32 = 0x1;
 const IFF_LOOPBACK: u32 = 0x8;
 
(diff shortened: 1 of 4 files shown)


src/net.rs
```diff
@@ -147,7 +149,7 @@ const fn is_unicast_link_local(addr: Ipv6Addr) -> bool {
 }
 
 /// Given a listen/bind address, finds all the local addresses for that address family.
-pub fn find_local_addresses(listen_addr: SocketAddr) -> Vec<SocketAddr> {
+pub(crate) fn find_local_addresses(listen_addr: SocketAddr) -> Result<Vec<SocketAddr>> {
```
flub, line 152: I don't think this should be pub, and probably should even be in the provider module.  Though that would make the diff harder to read.

dignifiedquire (approved):

PR #892, opened 2023-03-28
feat(ticket): Ensure a ticket always has at least one address

flub:
This makes it impossible to have a ticket with an empty address list.
It makes the fields of the ticket private and also makes creating the
ticket fallible.  Both are probably good things for the public API and
the latter e.g. allows us to add further checks like ensure the hash
really exists.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -489,24 +489,19 @@ async fn main_impl() -> Result<()> {
             }
         }
         Commands::GetTicket { out, ticket } => {
-            let Ticket {
-                hash,
-                peer,
-                addrs,
-                token,
-            } = ticket;
-            let addr = addrs
+            let addr = ticket
+                .addrs()
                 .get(0)
                 .copied()
                 .context("missing SocketAddr in ticket")?;
             let opts = get::Options {
                 addr,
-                peer_id: Some(peer),
+                peer_id: Some(ticket.peer()),
                 keylog: cli.keylog,
             };
             tokio::select! {
                 biased;
-                res = get_interactive(hash, opts, token, out) => {
+                res = get_interactive(ticket.hash(), opts, ticket.token(), out) => {
                     res
                 }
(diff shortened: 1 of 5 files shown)


src/provider/mod.rs
```diff
@@ -1158,26 +1153,61 @@ async fn write_response<W: AsyncWrite + Unpin>(
...
+    /// This will never be empty.
+    addrs: Vec<SocketAddr>,
     /// The authentication token with permission to retrieve the hash.
-    pub token: AuthToken,
+    token: AuthToken,
 }
 
 impl Ticket {
+    fn new(hash: Hash, peer: PeerId, addrs: Vec<SocketAddr>, token: AuthToken) -> Result<Self> {
+        ensure!(!addrs.is_empty(), "Invalid address list in ticket");
```
dignifiedquire, line 1169: Maybe more sth like, missing addresses?

dignifiedquire (dismissed):

src/main.rs
```diff
@@ -543,33 +538,35 @@ async fn main_impl() -> Result<()> {
...
+                        (path_buf, Some(path))
+                    };
+                    // tell the provider to add the data
+                    let stream = controller.server_streaming(ProvideRequest { path }).await?;
+                    let (hash, entries) = aggregate_add_response(stream).await?;
+                    print_add_response(hash, entries);
+                    let ticket = provider.ticket(hash)?;
+                    println!("All-in-one ticket: {ticket}");
+                    anyhow::Ok(tmp_path)
+                })
```
flub, line 568: this block is only indented, otherwise unchanged

flub:
PTAL: I additionally moved the Ticket to it's own module.  This ensures that the provider can not construct a ticket without upholding the invariant.  Also the provider module is pretty huge by now, so splitting it up does no harm.

PR #893, opened 2023-03-28
feat(get-ticket): Contact provider on all listening addrs

flub:
This tries to contact the provider on all addresses inside the ticket
when using the get-ticket command.  The first connection will be used.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -477,11 +477,14 @@ async fn main_impl() -> Result<()> {
             }
             let token = AuthToken::from_str(&auth_token)
                 .context("Wrong format for authentication token")?;
+            let get = GetInteractive::Hash {
+                hash: *hash.as_hash(),
+                opts,
+                token,
+            };
             tokio::select! {
                 biased;
-                res = get_interactive(*hash.as_hash(), opts, token, out) => {
-                    res
-                }
+                res = get_interactive(get, out) => res,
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -854,7 +861,14 @@ async fn get_interactive(
             Ok(reader)
         }
     };
-    let stats = get::run(hash, token, opts, on_connected, on_collection, on_blob).await?;
+    let stats = match get {
+        GetInteractive::Ticket { ticket, keylog } => {
+            get::run_ticket(&ticket, keylog, 16, on_connected, on_collection, on_blob).await?
```
dignifiedquire, line 866: I can haz constant or comment on that 16?

src/main.rs
```diff
@@ -753,13 +743,30 @@ async fn get_keypair(key: Option<PathBuf>) -> Result<Keypair> {
...
+    Ticket {
+        ticket: Ticket,
+        keylog: bool,
+    },
+    Hash {
+        hash: Hash,
+        opts: get::Options,
+        token: AuthToken,
+    },
+}
```
flub, line 757: This was the simplest way to add this functionality, it's not particularly elegant or well-named though.  But it works fine.  If you are disgusted however and have better ideas I'm happy to try different things.

dignifiedquire: seems fie

dignifiedquire (approved):

PR #894, opened 2023-03-29
Use the new bao-tree crate instead of abao

rklaehn:
What this gives us:

- async wherever you want (it is pretty easy to implement in bao-tree)
  note: I still think that we should not do async everywhere, but we can if we want!
- verified encoding, so validation errors get caught on the server side
- configurable block size, currently set to 4 (= 1024 * 2^4)

and most notably:

- querying for multiple ranges in a single request

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -110,7 +110,7 @@ mod tests {
         tokio::fs::write(&path, content).await?;
         // hash of the transfer file
         let data = tokio::fs::read(&path).await?;
-        let (_, expect_hash) = abao::encode::outboard(&data);
+        let expect_hash = blake3::hash(&data);
         let expect_name = filename.to_string();
 
         let (db, hash) =
@@ -207,7 +207,7 @@ mod tests {
             let name = name.into();
             let path = dir.join(name.clone());
             // get expected hash of file
-            let (_, hash) = abao::encode::outboard(&data);
+            let hash = blake3::hash(&data);
             let hash = Hash::from(hash);
 
             tokio::fs::write(&path, data).await?;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,6 +16,13 @@ mod util;
 pub use tls::{Keypair, PeerId, PeerIdError, PublicKey, SecretKey, Signature};
 pub use util::Hash;
 
+use bao_tree::BlockSize;
+
+pub(crate) const IROH_BLOCK_SIZE: BlockSize = match BlockSize::new(4) {
+    Some(bs) => bs,
+    None => panic!(),
+};
+
 #[cfg(test)]
 mod tests {
     use std::{
--- a/src/protocol.rs
+++ b/src/protocol.rs
@@ -4,7 +4,6 @@ use std::io;
 use std::str::FromStr;
 
 use anyhow::{bail, ensure, Context, Result};
-use bao_tree::ChunkNum;
 use bytes::{Bytes, BytesMut};
 use postcard::experimental::max_size::MaxSize;
 use quinn::VarInt;
@@ -119,7 +118,7 @@ pub(crate) async fn read_bao_encoded<R: AsyncRead + Unpin>(
 ) -> Result<Vec<u8>> {
     let mut decoder = bao_tree::r#async::AsyncResponseDecoder::new(
         hash.into(),
-        RangeSet2::from(ChunkNum(0)..),
+        RangeSet2::all(),
         IROH_BLOCK_SIZE,
         reader,
     );
--- a/src/protocol.rs
+++ b/src/protocol.rs
@@ -4,6 +4,7 @@ use std::io;
 use std::str::FromStr;
 
 use anyhow::{bail, ensure, Context, Result};
+use bao_tree::tokio_io::AsyncResponseDecoder;
 use bytes::{Bytes, BytesMut};
 use postcard::experimental::max_size::MaxSize;
 use quinn::VarInt;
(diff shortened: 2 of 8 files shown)


src/lib.rs
```diff
@@ -16,6 +16,13 @@ mod util;
 pub use tls::{Keypair, PeerId, PeerIdError, PublicKey, SecretKey, Signature};
 pub use util::Hash;
 
+use bao_tree::BlockSize;
+
+pub(crate) const IROH_BLOCK_SIZE: BlockSize = match BlockSize::new(4) {
```
rklaehn, line 21: unwrap is not const... 🤷

src/protocol.rs
```diff
@@ -113,7 +116,12 @@ pub(crate) async fn read_bao_encoded<R: AsyncRead + Unpin>(
     reader: R,
     hash: Hash,
 ) -> Result<Vec<u8>> {
-    let mut decoder = AsyncSliceDecoder::new(reader, &hash.into(), 0, u64::MAX);
+    let mut decoder = bao_tree::r#async::AsyncResponseDecoder::new(
```
dignifiedquire, line 119: this is an unfortunate module name

src/util.rs
```diff
@@ -177,33 +185,42 @@ impl From<io::Error> for BaoValidationError {
...
-    let mut buffer = vec![0u8; 1024 * 16 + 4096];
-    loop {
-        match decoder.read(&mut buffer) {
-            Ok(0) => break,
-            Ok(_) => {}
-            Err(err) => {
-                // todo: figure out exactly what went wrong
-                return Err(BaoValidationError::IoError(err));
+    let outboard =
+        bao_tree::outboard::PreOrderMemOutboard::new(hash, IROH_BLOCK_SIZE, outboard.to_vec())
```
dignifiedquire, line 196: that `to_vec` call is sad

src/provider/mod.rs
```diff
@@ -890,15 +890,10 @@ async fn send_blob<W: AsyncWrite + Unpin + Send + 'static>(
             // 'static lifetime.
             writer = tokio::task::spawn_blocking(move || {
                 let file_reader = std::fs::File::open(&path)?;
-                let outboard_reader = std::io::Cursor::new(outboard);
                 let mut wrapper = SyncIoBridge::new(&mut writer);
```
dignifiedquire, line 893: Can we not get rid of the `SyncIoBridge` now if we have async support?

src/protocol.rs
rklaehn: yes, was going to ask you about this. Not entirely happy with it. Maybe `tokio_io`?

src/util.rs
rklaehn: This entire section is sad, but it is going to be changed.

src/protocol.rs
dignifiedquire: Yeah `tokio_io` or `future_io` seems like a reasonable compromise. or sth like

```rust
mod io {
  mod sync {
  }
  mod tokio {
  }
}
```

rklaehn: For now it is io and tokio_io. But this looks nice as well...

src/provider/mod.rs
rklaehn: It si gone now.

src/util.rs
rklaehn: Got rid of the to_vec(). But it is really annoying having to write code twice, once for owned and once for borrowed...

dignifiedquire: Cow?

dignifiedquire: it is back

dignifiedquire:
This does not yet change anything in the protocol, correct?

src/provider/mod.rs
```diff
@@ -885,23 +884,17 @@ async fn send_blob<W: AsyncWrite + Unpin + Send + 'static>(
...
-                    0,
-                    size,
-                );
-                let _copied = std::io::copy(&mut slice_extractor, &mut wrapper)?;
-                std::io::Result::Ok(writer)
-            })
-            .await??;
+
+            let outboard =
+                PreOrderMemOutboard::new(name.into(), IROH_BLOCK_SIZE, outboard.to_vec());
```
dignifiedquire, line 889: is there a way to avoid this `to_vec` call?

src/util.rs
```diff
@@ -164,10 +168,14 @@ pub type RpcResult<T> = result::Result<T, RpcError>;
 pub(crate) enum BaoValidationError {
     /// Generic io error. We were unable to read the data.
     IoError(io::Error),
-    // /// The hash of the data does not match the hash of the outboard.
-    // HashMismatch,
-    // /// The size of the data does not match the size of the outboard.
-    // SizeMismatch,
+    /// The data failed to validate
+    EncodeError(EncodeError),
```
dignifiedquire, line 172: just use `#[from]` instead of the manual impl?

src/util.rs
```diff
@@ -177,33 +185,42 @@ impl From<io::Error> for BaoValidationError {
...
-            Ok(0) => break,
-            Ok(_) => {}
-            Err(err) => {
-                // todo: figure out exactly what went wrong
-                return Err(BaoValidationError::IoError(err));
+    let outboard =
+        bao_tree::outboard::PreOrderMemOutboard::new(hash, IROH_BLOCK_SIZE, outboard.to_vec())
+            .flip();
+    // little util that discards data but prints progress every 1MB
+    struct DevNull<F>(u64, F);
```
dignifiedquire, line 199: would be nice to pull this outside this function at least

rklaehn:
> This does not yet change anything in the protocol, correct?

No, there is all this fancy range set stuff, but we always request `RangeSet2::all()`...

src/provider/mod.rs
rklaehn: Ah, yes. I even did this on one of the branches that build on this. Just need to cherry pick it.

There are outboard versions that work with a &[u8] (and can't write), just need to use those.

src/util.rs
rklaehn: I can do this, but it is just used here, so I don't mind it being inside.

src/util.rs
```diff
@@ -224,6 +193,25 @@ pub(crate) fn validate_bao<F: Fn(u64)>(
...
 }
 
+// little util that discards data but prints progress every 1MB
+struct DevNull<F>(u64, F);
+
+impl<F: Fn(u64)> Write for DevNull<F> {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        let prev = self.0;
+        let curr = prev + buf.len() as u64;
+        if prev % 1000000 != curr % 1000000 {
```
dignifiedquire, line 203: constant? also we usually use MiB, so maybe use that?

dignifiedquire (approved):

PR #900, opened 2023-03-30
feat: metrics collection

Arqu:
This sets us up for collecting metrics throughout the codebase. 
Currently the metrics collection `Iroh` is added and meant to be for the general collection of metrics. Not married to the name. Didn't have a better idea and wanted to split it out based on subsystems (the incoming `derper` specifically).

Follow ups on this:
- make it a compile time feature so it can be completely off (this doesn't send anything outside and even the default endpoint is off by default)
- sprinkle metrics in the current codebase


dignifiedquire:
>  make it a compile time feature so it can be completely off

Lets do this as part of this PR please, trying to keep the minimum deps to an actual minimum.

Cargo.toml
```diff
@@ -53,6 +53,10 @@ walkdir = "2"
 webpki = "0.22"
 x509-parser = "0.14"
 zeroize = "1.5"
+lazy_static = "1.4.0"
```
dignifiedquire, line 56: please use https://docs.rs/once-cell instead, it has a lower compile time & dependency overhead

src/metrics/core.rs
```diff
@@ -0,0 +1,121 @@
...
+        let mut buf = vec![];
+        encode(&mut buf, self.registry())?;
+        Ok(buf)
+    }
+
+    pub(crate) fn set_enabled(&self, enabled: bool) {
+        self.enabled.swap(enabled, Ordering::Relaxed);
+    }
+
+    pub(crate) fn enabled(&self) -> bool {
```
dignifiedquire, line 47: `is_enabled` reads a bit nicer and is more convention

src/metrics/core.rs
```diff
@@ -0,0 +1,121 @@
...
+        self.enabled.load(Ordering::Relaxed)
+    }
+}
+
+/// defines the metric trait
```
dignifiedquire, line 52: comments should be capitalized, to read like a sentence

src/metrics/core.rs
```diff
@@ -0,0 +1,121 @@
...
+pub trait MetricType {
+    /// returns the name of the metric
+    fn name(&self) -> &'static str;
+}
+
+/// defines the histogram trait
```
dignifiedquire, line 58: that's a noop comment ;)

src/metrics/core.rs
```diff
@@ -0,0 +1,121 @@
...
+pub trait HistogramType {
+    /// returns the name of the metric
+    fn name(&self) -> &'static str;
+}
+
+/// definition of the metrics collection interfaces
```
dignifiedquire, line 64: noop

src/metrics/core.rs
```diff
@@ -0,0 +1,121 @@
...
+}
+
+/// interface to observe metrics
+pub trait MObserver {
+    /// observes a value for the metric
+    fn observe(&self, value: f64);
+}
+
+/// defines the generic record function
+#[allow(unused_variables, unreachable_patterns)]
```
dignifiedquire, line 89: why?

src/metrics/core.rs
```diff
@@ -0,0 +1,121 @@
...
+pub fn record<M>(c: Collector, m: M, v: u64)
+where
+    M: MetricType + std::fmt::Display,
+{
+    if CORE.enabled() {
+        match c {
+            Collector::Iroh => CORE.iroh_metrics().record(m, v),
+            _ => panic!("not enabled/implemented"),
```
dignifiedquire, line 97: `unimplemented!` is more idiomatic

src/metrics/core.rs
```diff
@@ -0,0 +1,121 @@
...
+        match c {
+            Collector::Iroh => CORE.iroh_metrics().observe(m, v),
+            _ => panic!("not enabled/implemented"),
+        };
+    }
+}
+
+/// list of all collectors
+#[derive(Debug, PartialEq, Eq)]
+pub enum Collector {
```
dignifiedquire, line 118: Best to make this enum non_exhaustive, (that way your clippy issue above goes away as well

src/metrics/macros.rs
```diff
@@ -0,0 +1,134 @@
+#[allow(unused_imports)]
```
dignifiedquire, line 1: why?

src/metrics/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+//!
+//! Enables and manages a global registry of metrics.
+//! Divided up into modules, each module has its own metrics.
+//! Starting the metrics service will expose the metrics on a OpenMetrics http endpoint.
+//!
+//! To enable metrics collection, call `init_metrics()` before starting the service.
+//! To record a metric, use the `record!` macro with the metric and the value to record.
+//! To increment a metric by 1, use the `inc!` macro with the metric.
+//! To observe a metric, use the `observe!` macro with the metric and the value to observe.
+//! To expose the metrics, start the metrics service with `start_metrics_service()`.
```
dignifiedquire, line 11: These instructions would make for a nice example in code :)

src/metrics/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+pub fn init_metrics() {
+    CORE.set_enabled(true);
+}
+
+/// Spawn a server to serve the OpenMetrics endpoint.
+pub async fn start_metrics_server(addr: Option<SocketAddr>) {
+    if let Some(metrics_addr) = addr {
+        self::service::run(metrics_addr).await
+    } else {
+        tracing::info!("Metrics server not started, no address provided");
```
dignifiedquire, line 35: maybe `warn`?

src/metrics/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+/// Expose iroh metrics
+pub mod iroh;
+mod service;
+
+/// Enables metrics collection, otherwise all inc!, record! & observe! calls are no-ops
+pub fn init_metrics() {
+    CORE.set_enabled(true);
+}
+
+/// Spawn a server to serve the OpenMetrics endpoint.
```
dignifiedquire, line 30: this doesn't `spawn`? it just starts it

src/metrics/service.rs
```diff
@@ -0,0 +1,46 @@
...
+            let handler = make_handler();
+            Ok::<_, io::Error>(service_fn(handler))
+        }))
+        .await
+        .unwrap();
+}
+
+/// This function returns a HTTP handler (i.e. another function)
+fn make_handler(
+) -> impl Fn(Request<Body>) -> Pin<Box<dyn Future<Output = io::Result<Response<Body>>> + Send>> {
```
dignifiedquire, line 28: very strange formatting

src/metrics/service.rs
```diff
@@ -0,0 +1,46 @@
...
+                .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))
+                .map(|r| {
+                    let body = Body::from(r);
+                    Response::builder()
+                        .header(
+                            hyper::header::CONTENT_TYPE,
+                            "application/openmetrics-text; version=1.0.0; charset=utf-8",
+                        )
+                        .body(body)
+                        .unwrap()
```
dignifiedquire, line 42: `expect`

Cargo.toml
flub: And is in std on nightly IIUC!  Or at least a subset.

src/main.rs
```diff
@@ -42,6 +44,9 @@ struct Cli {
     /// Log SSL pre-master key to file in SSLKEYLOGFILE environment variable.
     #[clap(long)]
     keylog: bool,
+    /// metrics server address
```
flub, line 47: "Bind address on which to serve Prometheus metrics" or something else more descriptive?

src/main.rs
```diff
@@ -460,7 +465,13 @@ async fn main_impl() -> Result<()> {
 
     let cli = Cli::parse();
 
-    match cli.command {
+    let metrics_addr = cli.metrics_addr;
+    init_metrics();
+    let metrics_fut = tokio::spawn(async move {
+        iroh::metrics::start_metrics_server(metrics_addr).await;
+    });
```
flub, line 472: This is initialised and spawned regardless of whether the option is set?

src/metrics/mod.rs
flub: Or take `addr: SocketAddr` instead of the `Option` and you don't have to handle this case.

src/metrics/service.rs
flub, line 1: hyper is quite an awkward thing to use manually... don't feel like using axum?

dignifiedquire: We are using `hyper` directly in the hole punching effort as well, so I'd prefer that for the moment, over pulling in more deps.

src/metrics/service.rs
Arqu: 🤷

src/metrics/service.rs
Arqu: Yeah I intentionally went with hyper just to reduce bloat in deps. I've used axum for almost every web service so far, but this endpoint specifically is very much internal use and doesn't really need any bells and whistles.

src/metrics/mod.rs
```diff
@@ -0,0 +1,46 @@
+//! Metrics collection
```
dignifiedquire, line 1: we have switched to the new module layout, so this file should be named `src/metrics.rs`

src/metrics/mod.rs
```diff
@@ -0,0 +1,46 @@
...
+mod macros;
+
+/// Expose core types and traits
+pub mod core;
+/// Expose iroh metrics
+pub mod iroh;
+mod service;
+
+/// Enables metrics collection, otherwise all inc!, record! & observe! calls are no-ops
+pub fn init_metrics() {
```
dignifiedquire, line 39: is there an actual use case for doing this and not starting the server?

src/metrics/service.rs
```diff
@@ -0,0 +1,45 @@
...
+    info!("Starting metrics server on {metrics_addr}");
+    Server::bind(&metrics_addr)
+        .serve(make_service_fn(move |_conn| async move {
+            let handler = make_handler();
+            Ok::<_, io::Error>(service_fn(handler))
+        }))
+        .await
+}
+
+/// This function returns a HTTP handler (i.e. another function)
```
dignifiedquire, line 25: to handle what?

src/metrics/iroh.rs
```diff
@@ -0,0 +1,16 @@
...
+use crate::{
+    metrics::core::Collector,
+    metrics::core::{HistogramType, MRecorder, MetricType, MetricsRecorder},
+};
+
+make_metrics! {
+    Iroh,
+    RequestsTotal: Counter: "Total number of requests received",
+    BytesSent: Counter: "Number of bytes streamed",
+    BytesReceived: Counter: "Number of bytes received"
```
dignifiedquire, line 15: are these metrics actually recorded?

src/metrics/mod.rs
Arqu: Right away, no. But probably very soon. A quick example would be to have a separate output mechanic we add in like just dump metrics to a file or directly into some local store or push to prometheus. All of which don't require the endpoint itself.

src/metrics/iroh.rs
Arqu: They are not used, mostly set as an example and I've used them for testing stuff out. I expect to refactor these as soon as we start collecting real metrics.

src/main.rs
```diff
@@ -440,6 +447,25 @@ fn print_add_response(hash: Hash, entries: Vec<ProvideResponseEntry>) {
 const PROGRESS_STYLE: &str =
     "{msg}\n{spinner:.green} [{elapsed_precise}] [{wide_bar:.cyan/blue}] {bytes}/{total_bytes} ({eta})";
 
+#[cfg(feature = "metrics")]
+fn init_metrics_collection(
+    metrics_addr: Option<SocketAddr>,
+) -> Option<tokio::task::JoinHandle<()>> {
+    init_metrics();
```
flub, line 454: This is still called unconditionally, IIUC this means even if `metrics_addr` is `None` it will track the metrics, but not start the server to export them.  Should this not be inside the `if let Some` just below so that all macros become no-ops?

src/metrics/core.rs
```diff
@@ -0,0 +1,125 @@
...
+        &self.registry
+    }
+
+    pub(crate) fn iroh_metrics(&self) -> &iroh::Metrics {
+        &self.iroh_metrics
+    }
+
+    pub(crate) fn encode(&self) -> Result<Vec<u8>, std::io::Error> {
+        let mut buf = vec![];
+        encode(&mut buf, self.registry())?;
```
flub, line 38: Rust is trying so hard to nerdsnipe me into wanting to overthink the allocations of this `Vec` here.  I'm trying to resist.  This is fine.

src/metrics/core.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+/// Defines the histogram trait which provides a common interface for all  based metrics
+pub trait HistogramType {
+    /// Returns the name of the metric
+    fn name(&self) -> &'static str;
+}
+
+/// Definition of the base metrics collection interfaces.
+/// Instances imlementing the MetricsRecorder are expected to have a defined mapping between
```
flub, line 64:
```suggestion
/// Definition of the base metrics collection interfaces.
///
/// Instances imlementing the MetricsRecorder are expected to have a defined mapping between
```

Leaving a blank line after a summary line renders much nicer in the indexes and search results in `cargo doc` output.  (Same in other doc comments)

src/metrics/core.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+/// Interface to observe metrics
+/// Helps expose the observe interface when using metrics as a library
+pub trait MObserver {
+    /// Observes a value for the metric
+    fn observe(&self, value: f64);
+}
+
+// used as an internal wrapper to record metrics only if the core is enabled
```
flub, line 91:
```suggestion
/// Internal wrapper to record metrics only if the core is enabled.
```

Doc comments everywhere!  `cargo doc --no-deps --document-private-items` is a great tool to use when working on a codebase.

flub (approved):

src/main.rs
Arqu: Same as an earlier comment. Currently there's only the endpoint to expose metrics, but we'll want other methods of dumping data soon and collection should be independent of serving.

flub (approved):

PR #905, opened 2023-04-03
ci: Test minimal crate versions

flub:
This check the dependencies as a library with the minimal crate versions to ensure we still compile with those.


flub:
@rklaehn the error from this comes from quic-rpc and needs to be fixed there.  Probably a good idea to add this CI check to that repo as well (and aboa, bao-tree)

.github/workflows/ci.yml
```diff
@@ -237,6 +237,17 @@ jobs:
...
+  minimal_crate_versions:
+    name: Minimal Crate Versions (library use only)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - uses: dtolnay/rust-toolchain@nightly
+      - name: Cargo check (minimal crate versions)
+        run: |
+          rm Cargo.lock
+          cargo +nightly check -Z minimal-versions --workspace --all-targets --no-default-features
```
dignifiedquire, line 249: We should also do this check for features enabled.

flub:
never mind

PR #906, opened 2023-04-03
abao debugging

flub:
test with `cargo test --package iroh --lib --all-features -- tests::many_files --exact --nocapture`


rklaehn:
Bug is fixed in #720. Abao was not at fault.

Issue #915, opened 2023-04-03
Bug when transferring many files


rklaehn:
Reproducer: just set the number of files in the many_files test to 40000 or so.

See https://github.com/n0-computer/iroh/pull/906

rklaehn:
I ran this test on the bao-tree branch, because bao-tree has better errors.

It seems that the hash mismatch happens very late in the collection:
```
running 1 test
NUM_FILES: 17000
5128 1316906
[src/lib.rs:271] provider.local_address() = 127.0.0.1:50450
on_connected
7cf4390b5e51ee6185b64b5205d594d95ee0bde2a5d4c177f80ab66f0771319a
7cf4390b5e51ee6185b64b5205d594d95ee0bde2a5d4c177f80ab66f0771319a
sent collection of size 17000
sent encoded collection 1322034
Error: leaf hash mismatch 1228800
```

PR #920, opened 2023-04-03
Fix many files bug

rklaehn:
I made the test fail by making the file name really long.

Made the file name repeat the number multiple times so a radix tree does not optimize the long name away, just in case we do radix trees in the future.


rklaehn:
Fixes https://github.com/n0-computer/iroh/issues/915

flub (approved):

flub (approved):

PR #929, opened 2023-04-03
fix clippy warnings in `hp::derp` & `hp::key` module

ramfox:


src/hp/key/node.rs
```diff
@@ -30,6 +30,12 @@ impl Hash for PublicKey {
     }
 }
 
+impl std::cmp::PartialEq for PublicKey {
```
dignifiedquire, line 33: why this manual impl?

dignifiedquire (approved):

src/hp/key/node.rs
ramfox: this was the clippy rule we were getting warned about:
https://rust-lang.github.io/rust-clippy/master/index.html#derive_hash_xor_eq

PR #931, opened 2023-04-04
fix: fetch PR details on issue comment

Arqu:


.github/workflows/netsim.yml
```diff
@@ -114,7 +120,7 @@ jobs:
       shell: bash  
       run: |  
         echo "LAST_COMMIT_SHA=$(git rev-parse --short ${GITHUB_SHA})" >> ${GITHUB_ENV}
-        echo "HEAD_REF=${{ github.head_ref }}" >> ${GITHUB_ENV}
+        echo "HEAD_REF=${{ github.ref }}" >> ${GITHUB_ENV}
```
Arqu, line 123: We don't really use this field, but it was empty, and this fixes it for now.

flub (approved):

PR #932, opened 2023-04-04
feat(loging): Improve logging output of provider and get

flub:
This makes the logging output a lot better, driven by real-life
debugging.  It introduces a few more spans and downgrades the
per-frame stuff to trace.


b5:
can confirm output of provide is _much_ nicer with this branch, especially with large directory trees

dignifiedquire (approved):

PR #935, opened 2023-04-05
ref(provider): Remove useless sort of data sources

flub:
This sort is entirely ignored, the outboards are unsorted and stable
sorting is done once the outboard are computed.


flub:
Going to replace this with a much larger refactor addressing a bunch of other issues.

PR #938, opened 2023-04-05
perf debug: remove things that could have caused a perf regression

rklaehn:
1. validated blob sending


rklaehn:
I think we have determined that the perf is mostly as expected, except for a minor regression involving validate on send. For the latter, there is an issue https://github.com/n0-computer/iroh/issues/941

PR #939, opened 2023-04-06
ref: Change how a collection is created

flub:
- If a file in a collection did not exist we would get a panic due to
  an unwrap, killing the entire task.  Now we correctly error showing
  which file was not found.

- We did a bunch of sync and async file operations while in the async
  functions.  This is now all moved to a single large spawn_blocking
  call avoiding the many small spawn_blocking calls the async
  operations give us as well as the potential blocking of the sync
  file operations.

- We did not check the maximum size of the serialised collection.

A few other smaller cleanups:

- The CollectionOrBlob::Blob variant is now directly a struct, the
  indirection of Data was not used anywhere to any effect and only
  distracted.

- The collection creation code is moved to it's own module now,
  provider is still huge.

### DIFF
--- a/src/provider/collection.rs
+++ b/src/provider/collection.rs
@@ -1,0 +1,213 @@
+//! Tools to build a collection to be added to a provider database.
+//!
+//! To create a collection one needs to create the [`Collection`] struct itself from all the
+//! blobs and treat this as a blob itself.  Then all blobs, including the "collection blob"
+//! are inserted in a hashmap.
+
+use std::collections::HashMap;
+use std::io::{BufReader, Cursor};
+use std::path::{Path, PathBuf};
+
+use anyhow::{bail, ensure, Context, Result};
+use bao_tree::outboard::PostOrderMemOutboard;
+use bytes::Bytes;
+use futures::{stream, StreamExt};
+use tracing::{trace, trace_span};
+
+use crate::blobs::{Blob, Collection};
+use crate::protocol::MAX_MESSAGE_SIZE;
+use crate::rpc_protocol::ProvideProgress;
+use crate::util::{Progress, ProgressReader, ProgressReaderUpdate};
+use crate::{Hash, IROH_BLOCK_SIZE};
+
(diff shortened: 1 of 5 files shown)


src/provider/collection.rs
```diff
@@ -0,0 +1,213 @@
...
+/// Synchronously compute the outboard of a file, and return hash and outboard.
+///
+/// It is assumed that the file is not modified while this is running.
+///
+/// If it is modified while or after this is running, the outboard will be
+/// invalid, so any attempt to compute a slice from it will fail.
+///
+/// If the size of the file is changed while this is running, an error will be
+/// returned.
+fn compute_outboard(
```
flub, line 173: This function is moved verbatim from it's previous location, no changes here.

rklaehn:
Hm, on my incremental branch I went the other way. I got this:

```rust
#[derive(Debug, Clone, PartialEq, Eq, From)]
pub(crate) enum BlobOrCollection {
    Blob(BlobData),
    Collection(CollectionData),
}
```

so I guess I am in more merge hell than before. But I guess I will probably have to start from scratch anyway and copy over changes...

rklaehn (approved):

flub:
> Hm, on my incremental branch I went the other way. I got this:
> 
> ```rust
> #[derive(Debug, Clone, PartialEq, Eq, From)]
> pub(crate) enum BlobOrCollection {
>     Blob(BlobData),
>     Collection(CollectionData),
> }
> ```

That's also a reasonable way to go, but having just the one as simply `Data` was too weird to handle.

rklaehn:
> > Hm, on my incremental branch I went the other way. I got this:
> > ```rust
> > #[derive(Debug, Clone, PartialEq, Eq, From)]
> > pub(crate) enum BlobOrCollection {
> >     Blob(BlobData),
> >     Collection(CollectionData),
> > }
> > ```
> 
> That's also a reasonable way to go, but having just the one as simply `Data` was too weird to handle.

Yeah, the inconsistency annoyed me as well...

PR #940, opened 2023-04-06
[WIP] refactors

dignifiedquire:


src/hp/magicsock/conn.rs
```diff
@@ -179,525 +189,267 @@ pub struct Inner {
...
+        derp_node_src: Option<key::node::PublicKey>,
+    ) -> bool {
+        let (s, r) = flume::bounded(1);
+        self.derp_sender.send(DerpMessage::HandleDiscoMessage {
+            msg,
+            src,
+            derp_node_src,
+            s,
+        });
+        r.recv().expect("dropped sender")
```
ramfox, line 357: this `r.recv()` blocks the executor. The `DerpHandler::run` loop is unable to receive the `DerpMessage::HandleDiscoMessage`, and we never process the `msg`, and never send anything back on `s`. This doesn't fix everything, but we are able to connect and pass data between `m1` & `m2` when this is fixed

dignifiedquire: fixed

PR #942, opened 2023-04-06
Remove FoundCollection - mostly this is about the backwards compat discussion - not want to merge this since it out of date

rklaehn:
We now only have Found and NotFound.

This also increases the ALPN because we are no longer compatible with the previous protocol.

### DIFF
--- a/src/tls.rs
+++ b/src/tls.rs
@@ -19,7 +19,7 @@ use ssh_key::LineEnding;
 
 use crate::util;
 
-pub(crate) const P2P_ALPN: [u8; 9] = *b"n0/iroh/1";
+pub(crate) const P2P_ALPN: [u8; 9] = *b"n0/iroh/2";
 
 /// A keypair.
 #[derive(Debug)]
--- a/src/tls.rs
+++ b/src/tls.rs
@@ -19,7 +19,7 @@ use ssh_key::LineEnding;
 
 use crate::util;
 
-pub(crate) const P2P_ALPN: [u8; 9] = *b"n0/iroh/2";
+pub(crate) const P2P_ALPN: [u8; 9] = *b"n0/iroh/1";
 
 /// A keypair.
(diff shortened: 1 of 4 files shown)


src/tls.rs
```diff
@@ -19,7 +19,7 @@ use ssh_key::LineEnding;
 
 use crate::util;
 
-pub(crate) const P2P_ALPN: [u8; 9] = *b"n0/iroh/1";
+pub(crate) const P2P_ALPN: [u8; 9] = *b"n0/iroh/2";
```
flub, line 22: We said we wouldn't change the ALPN, instead we should bump the version in `protocol::Handshake::version`.

flub (commented):
So this obviously raises the question of what to do with backwards compatibility right now.  Should we keep supporting the previous version?  Since we released iroh with it maybe we should?  Deltachat will be unhappy if we can no longer handle the version they use.

rklaehn:
I think it was very clear when we considered including iroh in deltachat that this would be a tool for transferring data from one device to another device, where both devices would be running the same version of deltachat.

If we have to retain backwards compatibility forever, it will slow us down a lot. I would say that then the inclusion would have been a mistake.

b5:
I agree we can’t afford to do backwards compat while we iterate, we can rely on minor crate versions and tell our downstream dependents to *not* upgrade with every minor release.

This is a fundamental tension we should interpret a pressure on our team to get the fundamentals right as fast as we can, and err on the side of releasing a 1.0 earlier than we’re currently thinking.

We need to clearly communicate with DeltaChat when we have shipped new code that might actually help them, not just on every release. And like the first time we should take time to work with them to land the upgrade.

rklaehn:
So we are going to remove FoundCollection as well as Found/NotFound in https://github.com/n0-computer/iroh/pull/930

With that I think we can close this.

PR #944, opened 2023-04-10
WIP new database implementation and format

rklaehn:
- no more distinction between blobs and collections, since that is fundamentally flawed
- multiple paths per hash
- inline storage (we need this for the collections)
- lazy outboard cache
- rudimentary db file versioning

Note that this builds on https://github.com/n0-computer/iroh/pull/930 , but the issue itself is very much independent of that. We could also add this to https://github.com/n0-computer/iroh/pull/942 if we wanted to take things slow.

BlobOrCollection is never going to work, since you can add the same data as both blob and collection, and then the current db will break.

E.g. 

- create a collection
- create a second collection with the collection data of the first collection added as a file/blob

Depending on which one you add first, the other just won't work.

### DIFF
--- a/src/provider/database.rs
+++ b/src/provider/database.rs
@@ -6,50 +6,63 @@ use crate::{
 };
 use anyhow::{Context, Result};
 use bytes::Bytes;
-use futures::StreamExt;
+use futures::{Future, FutureExt, StreamExt};
 use serde::{Deserialize, Serialize};
 use std::{
     collections::{BTreeMap, BTreeSet, HashMap},
-    fmt, io,
+    fmt,
+    io::{self, Read},
     path::{Path, PathBuf},
+    pin::Pin,
     result,
     sync::{Arc, RwLock},
 };
-use tokio::sync::mpsc;
+use tokio::{io::AsyncReadExt, sync::mpsc};
+use tokio_util::either::Either;
 
 /// Data for a (possibly very large) blob
 ///
 /// This can be used for both outboards and content. Be careful, this could
 /// point to a file that is a terabyte in size. So loading it into memory
 /// could be a problem.
-#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
-enum BlobData {
+///
+/// Blob data can either be inline (tiny data, stored in the db) or in a file.
+/// The file can be in the default location or a custom location.
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Clone)]
+pub enum BlobData {
     /// Data is stored inline in the db and exists in memory.
     Inline(Bytes),
     /// Data is stored in a file can be optionally cached in memory.
-    File {
-        /// Path to the file. We assume that this file does not change.
-        ///
-        /// None means the file is in the default location.
-        custom_path: Option<PathBuf>,
-        /// Possibly cached file data in memory.
-        cached: Option<Bytes>,
-    },
+    File(Option<PathBuf>),
 }
 
 impl BlobData {
-    fn path(&self) -> Option<Option<&PathBuf>> {
+    fn path(&self) -> BlobDataPath {
         match self {
-            Self::Inline(_) => None,
-            Self::File { custom_path, .. } => Some(custom_path.as_ref()),
+            Self::Inline(_) => BlobDataPath::Inline,
+            Self::File(custom_path) => custom_path
+                .as_ref()
+                .map(BlobDataPath::Custom)
+                .unwrap_or(BlobDataPath::Default),
         }
     }
 }
 
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
+enum BlobDataPath<'a> {
+    Inline,
+    Default,
+    Custom(&'a PathBuf),
+}
+
 /// Data for a (possibly very large) mutable blob
+///
+/// This is identical to BlobData, except that there is no default location
+/// for incomplete files. There can be multiple incomplete files for the same
+/// hash.
 #[derive(Debug, Serialize, Deserialize)]
 enum MutableBlobData {
-    Inline(Vec<u8>),
+    Inline(Bytes),
     File(PathBuf),
 }
 
@@ -62,7 +75,8 @@ impl MutableBlobData {
     }
 }
 
-enum DataReader {
+///
+pub enum DataReader {
     Memory(std::io::Cursor<Bytes>),
     File(std::fs::File),
 }
@@ -70,7 +84,7 @@ enum DataReader {
 impl std::io::Read for DataReader {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
         match self {
-            Self::Memory(data) => data.read(buf),
+            Self::Memory(data) => std::io::Read::read(data, buf),
             Self::File(file) => file.read(buf),
         }
     }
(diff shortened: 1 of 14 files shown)


src/provider/database.rs
```diff
@@ -6,21 +6,880 @@ use crate::{
...
+        Ok(match self {
+            Self::Inline(data) => data.len() as u64,
+            Self::File(custom_path) => {
+                let path = custom_path.as_ref().cloned().unwrap_or_else(default_path);
+                std::fs::metadata(path)?.len()
+            }
+        })
+    }
+
+    async fn len_async(&self, default_path: impl Fn() -> PathBuf) -> io::Result<u64> {
```
flub, line 59: right, let's hate async a bit more i guess

src/provider/database.rs
```diff
@@ -6,21 +6,880 @@ use crate::{
...
+        Ok(match self {
+            Self::Inline(data) => data.len() as u64,
+            Self::File(custom_path) => {
+                let path = custom_path.as_ref().cloned().unwrap_or_else(default_path);
+                tokio::fs::metadata(path).await?.len()
+            }
+        })
+    }
+
+    fn read_as_bytes(&self, default_path: impl Fn() -> PathBuf) -> io::Result<Bytes> {
```
flub, line 69: I'm used to seeing this as just `.bytes()` but maybe that's not a universal convention just something of the people I worked with.

src/provider/database.rs
```diff
@@ -6,21 +6,880 @@ use crate::{
...
+
+/// A bao file consists of outboard and content.
+///
+/// Either of them can stored inline or in a file in the file system.
+///
+/// Under some circumstances, we can have multiple content files. We will
+/// however not keep track of multiple outboard files.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+struct BaoFile {
+    outboard: BlobData,
```
flub, line 291: It is slightly odd to have the outboard stored in a `BlobData` as it already implied the content.  Maybe we could call that type `RawData` instead, but I appreciate that's more bikeshedding.

src/provider/database.rs
```diff
@@ -6,21 +6,880 @@ use crate::{
...
+    content: Vec<BlobData>,
```
flub, line 292: I'm a little unhappy this leaves the possibility for the raw data in memory multiple times.  Though it's behind a `Bytes` which has an `Arc` built-in, but still.  Maybe this `Vec` is better pushed down to a type distinguishing the in-mem storage from the file storage.

src/provider/database.rs
```diff
@@ -6,21 +6,880 @@ use crate::{
...
+/// They are related to each other.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+struct IncompleteBaoFile {
+    outboard: MutableBlobData,
+    content: MutableBlobData,
+}
+
+/// The data we retain for each hash
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+struct HashData {
```
flub, line 306: Is this not better as an enum?  Or am I misunderstanding this type?

flub (commented):
oh, this is big.  i've not wrapped my head around the structure yet and only got partway though.  so the comments aren't overly interesting yet.

src/provider/database.rs
```diff
@@ -6,21 +6,880 @@ use crate::{
...
+/// This can be used for both outboards and content. Be careful, this could
+/// point to a file that is a terabyte in size. So loading it into memory
+/// could be a problem.
+///
+/// Blob data can either be inline (tiny data, stored in the db) or in a file.
+/// The file can be in the default location or a custom location.
+#[derive(Clone, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
+pub enum BlobData {
+    /// Data is stored inline in the db and exists in memory.
+    Inline(Bytes),
```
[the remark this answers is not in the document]
rklaehn: The reason for the name is that it is stored inline in the database file, unlike the other case which gets its own file.

src/provider/database.rs
rklaehn: You mean I should call this len and call the other one len_sync?

src/provider/database.rs
rklaehn: This actually does read from disk. A method that is called something like `.bytes()` for me implies that it is a very cheap conversion or just passing out something that already exists.

src/provider/database.rs
rklaehn: Yeah, BlobData is a good name in isolation because it is after all a blob, but in the context of iroh it is not such a good name because blob is a specific concept in iroh.

Shrug. Naming is hard.

src/provider/database.rs
rklaehn: I think I am deduping this somewhere, but you are right that this could be hidden in its own type. I just want to keep these things simple since they also define the serialisation format.

src/provider/database.rs
rklaehn: I wanted to leave the option to have both. But maybe that is not necessary. Why would you hold on to incomplete data if you have complete data.

src/provider/database.rs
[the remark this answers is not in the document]
flub: oh, the database is actually on disk now.  didn't get that far :sweat_smile:

src/provider/database.rs
flub: sorry, no I'm not suggesting any change.  Only lamenting we need both.

src/provider/database.rs
flub: It's mostly bikeshedding and I don't really mind, but `.bytes()` does not imply it's cheap, at least to me but it's all relative.  `.as_bytes()` would suggest it's cheap.  Maybe that's the reason I noticed this, because the current name includes that as a substring.

src/provider/database.rs
flub: Hum, I don't think keeping incomplete data on disk after it's been completed would match user expectations.  You'd wonder where all your disk space is gone.  Bittorrent client also don't do this.  You know that as long as your client has this location and it's unmodified it can serve it, so it's up to you to manage it.  I think this works fine for iroh as well.

src/provider/database.rs
```diff
@@ -6,21 +6,880 @@ use crate::{
...
+
+    fn load(home_dir: PathBuf, validate: ValidateMode) -> io::Result<Self> {
+        tracing::debug!("loading database from {}", home_dir.display());
+        let content_dir = home_dir.join("data");
+        // make sure the content dir exists
+        std::fs::create_dir_all(&content_dir)?;
+        let content_dir = Arc::new(content_dir);
+        let db_file = Database::db_path(&home_dir);
+        let db_bytes = std::fs::read(db_file)?;
+        const HEADER: [u8; 8] = iroh_header(0);
```
dignifiedquire, line 385: version should be a constant

src/provider/database.rs
```diff
@@ -6,21 +6,880 @@ use crate::{
...
+/// could be a problem.
+///
+/// Blob data can either be inline (tiny data, stored in the db) or in a file.
+/// The file can be in the default location or a custom location.
+#[derive(Clone, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
+pub enum BlobData {
+    /// Data is stored inline in the db and exists in memory.
+    Inline(Bytes),
+    /// Data is stored in a file can be optionally cached in memory.
+    File(Option<PathBuf>),
```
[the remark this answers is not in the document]
rklaehn: It means that it is in the default location, which is the iroh data directory.

rklaehn:
Gonna start from scratch and copy things over - this is not salvageable given some changed assumptions and a lot of changes to x-hp...

rklaehn:
I will start from scratch. Too many things have changed

- we want to abstract over the storage format
- too many conflicts with x-hp

rklaehn:
No chance to merge this to x-hp. Also some premises have changed. So I am going to close this and start from scratch - and copy over things.

PR #948, opened 2023-04-12
feat: initial support for c bindings

b5:
I'm unclear on if this is allowed in our primary repo, but I personally think the c bindings should be here. Thoughts welcome.


irohc/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "irohc"
+version = "0.1.0"
+edition = "2021"
+
+[profile.release]
+strip = "symbols"
+
+[lib]
+name="irohc"
```
dignifiedquire, line 10: libiroh would be a more typical name

irohc/readme.md
```diff
@@ -0,0 +1,8 @@
+# C bindings for Iroh
+
+Running `cargo build --release` will produce a dynamic libaray. `irohc.h` is a handcrafted header file for that library. Should work on all platforms, I've only tested it on MacOS so far.
+
+For builds targeting older versions of MacOS, build with with:  `MACOSX_DEPLOYMENT_TARGET=10.7 && cargo build --target x86_64-apple-darwin --release`.
+
```
dignifiedquire, line 6: I would recommend to use https://docs.rs/safer-ffi/latest/safer_ffi/ it refuces safety and maintenance issues considerably

iroh-ffi/.gitignore
```diff
@@ -0,0 +1,4 @@
+/target
+/Cargo.lock
```
dignifiedquire, line 2: can we make this part of the workspace please, then you shouldn't need these two lines

iroh-ffi/.gitignore
```diff
@@ -0,0 +1,4 @@
...
+main
```
dignifiedquire, line 3: what's up with this?

iroh-ffi/.gitignore
```diff
@@ -0,0 +1,4 @@
...
+iroh.h
```
dignifiedquire, line 4: shouldn't the header file be checked in?

iroh-ffi/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "iroh_ffi"
+version = "0.1.0"
+edition = "2021"
+publish = false
+rust-version = "1.64"
```
dignifiedquire, line 6: missing a license

iroh-ffi/src/lib.rs
```diff
@@ -0,0 +1,279 @@
+use std::{path::PathBuf, str::FromStr};
+
+use ::safer_ffi::prelude::*;
```
dignifiedquire, line 3: shouldn't need the `::` prefix

iroh-ffi/src/lib.rs
```diff
@@ -0,0 +1,279 @@
...
+            out_path,
+        ))
+        .unwrap();
+        0
+    });
+    if result.is_err() {
+        eprintln!("error: rust panicked");
+        return 1;
+    }
+    result.unwrap()
```
dignifiedquire, line 55: no need to `unwrap` here

```rust
match result {
  Ok(res) => res,
  Err(err) => {
    eprintln!("rust panicked: {:?}", err);
    1
  }
}
```

iroh-ffi/src/lib.rs
```diff
@@ -0,0 +1,279 @@
...
+    peer: char_p::Ref<'_>,
+    peer_addr: char_p::Ref<'_>,
+    out_path: char_p::Ref<'_>,
+) -> u32 {
+    let result = std::panic::catch_unwind(|| {
+        let hash = hash.to_str().parse::<Hash>().unwrap();
+        let peer = peer.to_str().parse::<PeerId>().unwrap();
+        let peer_addr = peer_addr.to_str().parse().unwrap();
+        let out_path = PathBuf::from_str(out_path.to_str()).unwrap();
+        let rt = Runtime::new().unwrap();
```
dignifiedquire, line 34: this is a lot of `unwrap`, please use `anyhow::Result` to not just run into panics unnecessarily

iroh-ffi/src/util.rs
```diff
@@ -0,0 +1,90 @@
+//! Utility functions and types.
+use std::{fmt, str::FromStr};
+
+use anyhow::Result;
+use iroh::Hash;
+
+#[repr(transparent)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Blake3Cid(pub Hash);
```
dignifiedquire, line 9: doesn't this stuff already exist in the `iroh` code base? might as well export it, rather than duplicate this code

dignifiedquire:
closing in favor of #1321

PR #950, opened 2023-04-18
feat: Expose `pub mod util` module

matheus23:
Useful for experiment Appa.


matheus23:
Ignore this. `Hash` is already exposed.

PR #959, opened 2023-05-02
feat(provider): Allow binding to multiple addresses

flub:
This allows binding the provider to multiple addresses, the common
case is to bind to both IPv4 and IPv6 sockets.  But any suitable
combination is possible.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -131,7 +131,9 @@ mod tests {
 
         let (db, hash) =
             provider::create_collection(vec![provider::DataSource::File(path)]).await?;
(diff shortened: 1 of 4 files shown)


dignifiedquire:
do you have a specific use case for this?

flub:
Essentially the IPv4/IPv6 one as in deltachat I ended up skipping IPv4 because it was adding a bunch of user code just to work around the fact iroh didn't support it nicely.

dignifiedquire:
Unless you need this right now, I would like this to be merged/adapted into the holepunching world (which changes a lot in this direction already)

flub:
fair enough, i'll leave it as a branch then

PR #960, opened 2023-05-03
fix(database): Handle finding beetle data directory

flub:
This fixes very obscure errors when a beetle (formerly named iroh)
data directory is found.

- Error messages are improved to indicate what is failing.  Making use
  of anyhow's error chaining.

- The paths.bin file is used as a marker to decide if a database is
  present or needs to be created.

The end result is that this happily creates the new iroh database at
the same location of a previous beetle store, leaving the beetle files
in place.


dignifiedquire (approved):
looks reasonable to me, probably needs a look from the db chief @rklaehn

src/main.rs
```diff
@@ -884,9 +891,9 @@ async fn get_interactive(get: GetInteractive, out: Option<PathBuf>) -> Result<()
                 // Rename temp file, to target name
                 let filepath2 = filepath.clone();
                 if let Some(parent) = filepath2.parent() {
-                    tokio::fs::create_dir_all(parent)
-                        .await
-                        .context("Unable to create directory {parent}")?;
+                    tokio::fs::create_dir_all(parent).await.with_context(|| {
```
rklaehn, line 894: with_context is so that the formatting is lazy?

rklaehn:
Looks reasonable.

rklaehn (dismissed):

src/main.rs
flub: The previous version did not do any formatting, just literally printed `{parent}` in the output.

`.with_context` is if you need to do any anything that's not a static `Context` with a `Display`.  `format!` is just the most common version of this.

dignifiedquire (approved):

src/main.rs
rklaehn: oh, ok...

rklaehn (approved):

PR #969, opened 2023-05-05
Actually add the custom request handler

rklaehn:
This builds on the refactor to make multiple requests possible, and actually does add the custom request.

### DIFF
--- a/src/protocol/mod.rs
+++ b/src/protocol/mod.rs
@@ -8,6 +8,7 @@ use bytes::{Bytes, BytesMut};
 use derive_more::From;
 use postcard::experimental::max_size::MaxSize;
 use quinn::VarInt;
+use range_collections::RangeSet2;
 use serde::{Deserialize, Serialize};
 use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
 mod range_spec;
@@ -41,7 +42,14 @@ impl Handshake {
 pub enum Request {
     /// A get request for a blob or collection
     Get(GetRequest),
+    /// A get request that allows the receiver to create a collection
+    CustomGet(Bytes),
 }
+
+/// Currently all requests are get requests. But that won't always be the case.
+///
+/// Hence this type alias that will at some point be replaced by a proper enum.
+pub type AnyGetRequest = Request;
 
 /// A request
 #[derive(Deserialize, Serialize, Debug, PartialEq, Eq, Clone)]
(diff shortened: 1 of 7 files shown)


src/protocol/mod.rs
```diff
@@ -41,8 +42,15 @@ impl Handshake {
 pub enum Request {
     /// A get request for a blob or collection
     Get(GetRequest),
+    /// A get request that allows the receiver to create a collection
+    CustomGet(Bytes),
```
flub, line 46: Is it bikeshedding time?  I vote `Query`.  But never mind, I vote not slowing down because of bikeshedding.

src/protocol/mod.rs
```diff
@@ -67,6 +75,14 @@ impl GetRequest {
             ranges: RangeSpecSeq::all(),
         }
     }
+
+    /// Request just a single blob
+    pub fn just(hash: Hash) -> Self {
```
flub, line 80: I feel a bit stronger about bikeshedding here.  Not so much I care about my suggestion but I dislike this one a lot.  So I suggest `single`, it matches `all` fairly well.

src/provider/mod.rs
```diff
@@ -75,13 +75,40 @@ pub const DEFAULT_BIND_ADDR: ([u8; 4], u16) = ([127, 0, 0, 1], 4433);
...
+
+/// A custom get request handler that allows the user to make up a get request
+/// on the fly.
+pub trait CustomGetHandler: Send + Sync + Clone + 'static {
+    /// Handle the custom request, given an opaque data blob from the requester.
+    fn handle(
+        &self,
+        request: Bytes,
+        db: Database,
+    ) -> BoxFuture<'static, anyhow::Result<GetRequest>>;
```
flub, line 100: did you consider a version where this is responsible for directly feeding you the collection and blob data?

With this design you have to add the collection to the db, construct the request and then magically know when the response has been sent and remove the stuff from the db again if that's what you need.

flub (dismissed):

src/provider/mod.rs
rklaehn: Yes, I think we need something where the temp collection gets removed automatically after the request gets processed.

But I think @dignifiedquire said that he would prefer this to be merged more or less as-is so appa can depend on it, and we make it nicer later after we did hole punching stuff.

src/protocol/mod.rs
rklaehn: Query is a bit generic, since we are going to have a Query request for the discovery part. E.g. what children of this hash do you actually have...

My rationale for the naming is that this is still fundamentally a get request, just a named one, not a content-addressed one.

src/protocol/mod.rs
rklaehn: Yeah, will use single.

flub (approved):

PR #974, opened 2023-05-07
Implement custom collection parser

rklaehn:
This is basically the provider side of what has already been done on the get side.

Make the core independent of the notion of a collection. This allows the following things:

- people can use existing dag-cbor data as collections, provided that it contains only blake3 hashes.
- experiment with things like "a collection is just an array of links" / "metadata is in the first child"
- allow users such as @fabricedesre to customise the notion of a collection without having to fork the entire repo

This is written in "callback style". I wonder if we should rewrite the entire provider in fsm style as well to be super flexible, but I would prefer to do that in a separate PR.


rklaehn:
@fabricedesre can you check if this would allow you to add metadata such as a mime type to collections without having to fork it?

fabricedesre:
> @fabricedesre can you check if this would allow you to add metadata such as a mime type to collections without having to fork it?

I think I understand the parsing side (eg. moving from `let collection = postcard::from_bytes::<Vec<Hash>>(&data)?;` to something like `let collection = postcard::from_bytes::<Vec<(Hash, MimeType)>>(&data)?;`) but I could not see clearly what changes are needed when creating the collection.

rklaehn:
> > @fabricedesre can you check if this would allow you to add metadata such as a mime type to collections without having to fork it?
> 
> I think I understand the parsing side (eg. moving from `let collection = postcard::from_bytes::<Vec<Hash>>(&data)?;` to something like `let collection = postcard::from_bytes::<Vec<(Hash, MimeType)>>(&data)?;`) but I could not see clearly what changes are needed when creating the collection.

Right. There is something missing. There is a database refactor still to be done to fully decouple us from the notion of a collection. It is already open since before ipfs thing, but put on the back burner due to the current hole punching focus.

What you would do is to define a custom collection and add it *as a blob*, like the fn I added to enable my test. However this requires you to write the collection to a file first, since blobs are always files atm.

E.g. like this:

```rust
struct Entry {
  hash: Hash,
  name: String,
  mine_type: String,
}
```

Then to add a collection, add the files as blobs and then add the collection itself as a blob.

```rust
fn add_collection(entries: Vec<Entry>, db: Database) -> anyhow::Result<()> {
   let collection_bytes = postcard::to_stdvec(&entries)?;   
   // wriite to file and get path
   db.put_blob(path)?; // this is currently cfg(test), but could be made pub
   Ok(())
}
```

We will do the db refactor first thing once we come out of the hole punching sprint...

src/provider/mod.rs
```diff
@@ -112,6 +116,150 @@ impl CustomGetHandler for () {
...
 
+/// A custom collection parser that allows the user to define what a collection is.
+///
+/// A collection can be anything that contains an ordered sequence of blake3 hashes.
+/// Some collections store links with a fixed size and therefore allow efficient
+/// skipping. Others store links with a variable size and therefore only allow
+/// sequential access.
+///
+/// This API tries to accomodate both use cases. For collections that do not allow
+/// efficient random access, the `skip` method can be implemented by just repeatedly
```
flub, line 127:
```suggestion
/// efficient random access, the [`CollectionIter::skip`] method can be implemented by just repeatedly
```

src/provider/mod.rs
```diff
@@ -112,6 +116,150 @@ impl CustomGetHandler for () {
...
+/// calling `next`.
+///
+/// For collections that do allow efficient random access, the `skip` method can be
```
flub, line 130:
```suggestion
/// For collections that do allow efficient random access, the [`CollectionIter::skip`] method can be
```

src/provider/mod.rs
```diff
@@ -112,6 +116,150 @@ impl CustomGetHandler for () {
...
+    type NextFuture: Future<Output = (Self, anyhow::Result<Option<Hash>>)> + Send + 'static;
+    /// Future returned by [`CollectionIter::skip`].
+    ///
+    /// For an initial implementation where you don't care that much about allocations,
+    /// this can be just `BoxFuture`. This would be the default if defaults were allowed.
+    type SkipFuture: Future<Output = (Self, anyhow::Result<()>)> + Send + 'static;
+    /// Get the next hash in the collection.
+    fn next(self) -> Self::NextFuture;
+    /// Skip a number of hashes in the collection.
+    fn skip(self, n: u64) -> Self::SkipFuture;
```
flub, line 163: Maybe we could already implement the inefficient version by using `.next()` repeatedly as default implementation?

src/provider/mod.rs
```diff
@@ -112,6 +116,150 @@ impl CustomGetHandler for () {
...
+/// For collections that do allow efficient random access, the `skip` method can be
+/// used to move some internal offset.
+pub trait CollectionParser: Send + Default + 'static {
+    /// The iterator that allows to iterate over the collection.
+    type CollectionIter: CollectionIter;
+    /// Future returned by [`CollectionParser::parse`].
+    ///
+    /// For an initial implementation where you don't care that much about allocations,
+    /// this can be just `BoxFuture`. This would be the default if defaults were allowed.
+    type NewFuture: Future<Output = anyhow::Result<Self::CollectionIter>> + Send + 'static;
```
flub, line 139: I guess the naming oracle was on strike when `NewFuture` was chosen...  Maybe `ParseFuture` is slightly less generic.

src/provider/mod.rs
```diff
@@ -112,6 +116,150 @@ impl CustomGetHandler for () {
...
+    type SkipFuture = future::Ready<(Self, anyhow::Result<()>)>;
+    fn next(self) -> Self::NextFuture {
+        unreachable!()
+    }
+    fn skip(self, _n: u64) -> Self::SkipFuture {
+        unreachable!()
+    }
+}
+
+/// Define CustomCollection for (). This is for when you don't want to use support collections at all.
```
flub, line 177:
```suggestion
/// A CustomCollection for ().
///
/// This is useful for when you don't want to support collections at all.
```

src/provider/mod.rs
```diff
@@ -112,6 +116,150 @@ impl CustomGetHandler for () {
...
+    /// For an initial implementation where you don't care that much about allocations,
+    /// this can be just `BoxFuture`. This would be the default if defaults were allowed.
+    type NewFuture: Future<Output = anyhow::Result<Self::CollectionIter>> + Send + 'static;
+    /// Parse a collection with this parser
+    fn parse<R: AsyncSliceReader + Send + 'static>(format: u64, reader: R) -> Self::NewFuture;
+}
+
+/// An iterator over a collection. Allows to get the next hash or skip a number of hashes.
+///
+/// This does not implement `Iterator` or `Stream` because of the extra `skip` method.
```
flub, line 146:
```suggestion
/// An iterator over a collection.
///
/// Allows to get the next hash or skip a number of hashes.  Does not
/// implement `Iterator` or `Stream` because of the extra `skip` method.
```

src/provider/mod.rs
```diff
@@ -112,6 +116,150 @@ impl CustomGetHandler for () {
...
+    type NewFuture = future::Ready<anyhow::Result<Never>>;
+
+    fn parse<R: AsyncSliceReader + Send + 'static>(_format: u64, _reader: R) -> Self::NewFuture {
+        future::err(anyhow::anyhow!("collections not supported"))
+    }
+}
+
+/// Parser for the current iroh default collections
+#[derive(Debug, Clone, Copy, Default)]
+pub struct DefaultCollectionParser;
```
flub, line 190: Bikeshedding, but I'd be tempted to call this `IrohCollectionParser`.  Maybe also `IrohCollectionIter` though I guess array could be re-used otherwise as well.

src/provider/mod.rs
```diff
@@ -112,6 +116,150 @@ impl CustomGetHandler for () {
...
+
+/// Iterator for the current iroh default collections
+#[derive(Debug, Clone)]
+pub struct ArrayCollectionIter {
+    hashes: Box<[Hash]>,
+    offset: usize,
+}
```
flub, line 197: If we really want to handle any size collection then this wouldn't hold an array of hashes but would rather be a `&[Hash]` and it could be referring to a mmap or so for actual data.  Is it possible to make this or are the lifetimes just not doable?

src/provider/mod.rs
rklaehn: Unfortunately defaults are not allowed for types. Associated type defaults are unstable or something. So the only way to do this would be to make skip always return a BoxFuture<...>.

src/provider/mod.rs
rklaehn: Lol. The fn was called new at some point. It was more the refactoring consistency fairy that was on strike...

src/provider/mod.rs
rklaehn: The ArrayCollectionIter is intended to be reused. But here I guess you are right.

src/provider/mod.rs
flub: oh, TIL.  never mind then

src/lib.rs
```diff
@@ -551,6 +555,52 @@ mod tests {
         Ok((collection, items, stats))
     }
 
+    // helper to aggregate a get response and return all relevant data
```
flub, line 558: let's nitpick: Always write doc comments in proper style: 1 sentence summary with capitals and punctuation in imperative style, optionally followed by a blank line and more explanation.  That way it renders nicely with `cargo doc --document-private-items`.

src/provider/mod.rs
flub: still feel like changing this?

src/provider/mod.rs
flub: so you don't feel like making `IrohCollectionParser` deal with arbitrarily large collections out of the box?  I'm also fine with creating a follow up issue if the answer is "yes, but rather not in this PR"

flub (dismissed):

src/provider/mod.rs
rklaehn: The answer is "yes, but rather not in this PR" indeed...

rklaehn:
Two remarks:

1. there is an entire can of worms with custom collections involving the format int. Currently we just don't transfer it via the protocol at all, but if you wanted to write a node that supports different kinds of collections you would have to.

I vote we postpone dealing with this, since we are not promising stable protocol for this release.

2. with this the Collection (in iroh-bytes/blobs/) becomes less and less important. I did a quick test to entirely get rid of it and only specify the concrete collection type in the iroh crate. Probably doable, but also a major refactoring that we should probably do later and not in this PR.

iroh-bytes/src/provider.rs
```diff
@@ -211,6 +213,133 @@ pub trait CustomGetHandler: Send + Sync + Debug + 'static {
...
+///
+/// For collections that do allow efficient random access, the [`LinkStream::skip`] method can be
+/// used to move some internal offset.
+pub trait CollectionParser: Send + Debug + Clone + 'static {
+    /// Parse a collection with this parser
+    fn parse<'a, R: AsyncSliceReader + 'a>(
+        &'a self,
+        format: u64,
+        reader: R,
+    ) -> LocalBoxFuture<'a, anyhow::Result<(Box<dyn LinkStream>, CollectionStats)>>;
```
flub, line 235: Does this mean that at parse time you need to know the `CollectionStats`?  What if you have to iterate through the entire collection to know how many links are in it?  What if you don't have the metadata to figure out the total blob size?

iroh-bytes/src/provider.rs
```diff
@@ -211,6 +213,133 @@ pub trait CustomGetHandler: Send + Sync + Debug + 'static {
...
+    fn parse<'a, R: AsyncSliceReader + 'a>(
+        &'a self,
+        _format: u64,
+        _reader: R,
+    ) -> LocalBoxFuture<'a, anyhow::Result<(Box<dyn LinkStream>, CollectionStats)>> {
+        future::err(anyhow::anyhow!("collections not supported")).boxed_local()
+    }
+}
+
+/// Parser for the current iroh default collections
```
flub, line 275: maybe call out that this allocates all the hashes in an array on the heap?

iroh/tests/provide.rs
```diff
@@ -600,6 +629,66 @@ fn readme_path() -> PathBuf {
...
+    fn parse<'a, R: AsyncSliceReader + 'a>(
+        &'a self,
+        _format: u64,
+        mut reader: R,
+    ) -> LocalBoxFuture<'_, anyhow::Result<(Box<dyn LinkStream>, CollectionStats)>> {
+        async move {
+            let data = reader.read_to_end().await?;
+            let collection = postcard::from_bytes::<Vec<Hash>>(&data)?;
+            let iter: Box<dyn LinkStream> = Box::new(ArrayLinkStream::new(collection.into()));
+            Ok((iter, Default::default()))
```
flub, line 646: oh, i see.  you solved this stats problem by just lying about it...  Maybe allow for `None` in the struct in this case instead of using `0` as "i don't know"

flub (approved):
If I understand it right I'm not too fond on the CollectionStats thing.  Maybe allow them to take some kind of "i don't know" value?

PR #981, opened 2023-05-08
feat: remove AuthToken

dignifiedquire:
This simplifies the protocol, and now that we have the request extension, it is possible to reimplement this in userspace. 

When we upgrade deltachat to the version without this, it should be easy enough to factor out a small library that users can easily use if they want this functionality.

### DIFF
--- a/src/protocol/mod.rs
+++ b/src/protocol/mod.rs
@@ -1,7 +1,5 @@
 //! Protocol for communication between provider and client.
-use std::fmt::Display;
 use std::io;
-use std::str::FromStr;
 
 use anyhow::{bail, ensure, Context, Result};
 use bytes::{Bytes, BytesMut};
@@ -14,7 +12,7 @@ use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
 mod range_spec;
 pub use range_spec::{NonEmptyRequestRangeSpecIter, RangeSpec, RangeSpecSeq};
 
-use crate::util::{self, Hash};
+use crate::util::Hash;
 
 /// Maximum message size is limited to 100MiB for now.
 pub(crate) const MAX_MESSAGE_SIZE: usize = 1024 * 1024 * 100;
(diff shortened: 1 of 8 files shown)


src/protocol/mod.rs
```diff
@@ -25,15 +23,11 @@ pub const VERSION: u64 = 2;
 #[derive(Deserialize, Serialize, Debug, PartialEq, Eq, Clone, MaxSize)]
 pub(crate) struct Handshake {
```
rklaehn, line 24: So handshake now is just the version number on the wire. Nice.

src/protocol/mod.rs
```diff
@@ -131,55 +125,6 @@ pub(crate) async fn read_lp(
     Ok(Some(buffer.split_to(size).freeze()))
 }
 
-/// A token used to authenticate a handshake.
-///
-/// The token has a printable representation which can be serialised using [`Display`] and
-/// deserialised using [`FromStr`].
-#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, Deserialize, Serialize, MaxSize)]
-pub struct AuthToken {
```
rklaehn, line 139: I guess we can dig this out again when needing the auth token in the delta chat integration...

rklaehn (approved):

flub:
So I'm -0 on this.

I don't think we should leave authentication up to the users of the library but need to support several authentication mechanisms:

- Bi-directional using two PeerIDs.
- Bi-directional using a PeerID and AuthToken (what is being removed here).
- Uni-directional using the provider PeerID only (what this PR is implementing).
- Public, aka no authentication.

Since the current implementation is only one arbitrary one out of this list I don't object to arbitrarily changing it here as it makes our life a little easier in the short term.  But we should probably plan to implement this properly sometime.

dignifiedquire:
> * Bi-directional using two PeerIDs

This is already implemented in `main`, and will be the default when switching over to the hole punching world.

flub:
> > * Bi-directional using two PeerIDs
> 
> This is already implemented in `main`, and will be the default when switching over to the hole punching world.

Great.

> At least that is how I understand it, could you expand what exactly you mean with bi-directional?

Both peers authenticate the other peer while connecting.

dignifiedquire:
To reiterate some details on why this is being removed: 
Authentication needs to happen in two different places, on a connection level and on a per request level. The current mechanism does both, but badly, which is why it is being removed.

To support the use case in Delta Chat we will implement the request level one before cutting the next release. The connection level one is more complex and will come later.

Tracking issues: 

- https://github.com/n0-computer/iroh/issues/983
- https://github.com/n0-computer/iroh/issues/987

cc @b5

Issue #979, opened 2023-05-08
Tool to detect network state (public ips, relays, NAT traversal etc.)


rklaehn:
Write a tool similar to the old `dat doctor`.

It should have two main functions:

1. acquire and show basic information about your node
  - peer id etc.
  - information about available networks
  - reachable derp servers
  - public addresses

2. troubleshoot one particular connection
  - show a command that has to be entered on the remote peer
  - listen for incoming connections from said remote peer
  - find out and show info about all possible connections between the two nodes
  
![image](https://user-images.githubusercontent.com/248257/236790596-53eb85b3-0531-4f23-97d1-423d3e4a3872.png)

![image](https://user-images.githubusercontent.com/248257/236790684-f1caf0de-62f0-41a1-998d-78ca3af8ca24.png)

dignifiedquire:
@rklaehn can this be closed, or did you plan for more here?

PR #986, opened 2023-05-09
Add iroh doctor utility

rklaehn:
Closes https://github.com/n0-computer/iroh/issues/979

The intent of the utility is to be able to do two things

- find out global network status, public ip and properties of local LAN/WLAN
- connect to one particular other peer

Example usage:

# Report

```
> drderp report
getting report using derp map DerpMap {
    regions: {
        1: DerpRegion {
            region_id: 1,
            nodes: [
                DerpNode {
                    name: "default-1",
                    region_id: 1,
                    host_name: "derp.iroh.computer",
                    stun_only: false,
                    stun_port: 3478,
                    stun_test_ip: None,
                    ipv4: None,
                    ipv6: None,
                    derp_port: 0,
                },
            ],
            avoid: false,
            region_code: "default",
        },
    },
}
Report {
    udp: true,
    ipv6: false,
    ipv4: true,
    ipv6_can_send: false,
    ipv4_can_send: false,
    os_has_ipv6: true,
    icmpv4: false,
    mapping_varies_by_dest_ip: None,
    hair_pinning: Some(
        false,
    ),
    upnp: None,
    pmp: None,
    pcp: None,
    preferred_derp: 1,
    region_latency: {
        1: 123.08325ms,
    },
    region_v4_latency: {
        1: 123.08325ms,
    },
    region_v6_latency: {},
    global_v4: Some(
        82.76.245.176:64363,
    ),
    global_v6: None,
    captive_portal: None,
}
```

# Connect

```
> drderp connect
run 'drderp connect 537a4f6d07d4edec7fff226c1ac63a9f9655ed57a10e534dc633b9ad6b119a2d' in another terminal or on another machine
(do as you are told)
sending 1 bytes
done in 2.597788667 s
speed 0.3849427833384262 bytes/s
sending 16 bytes
done in 0.745654084 s
speed 21.457670980851223 bytes/s
sending 256 bytes
done in 0.7459795 s
speed 343.1729692303877 bytes/s
sending 4096 bytes
done in 0.623054958 s
speed 6574.058913114371 bytes/s
sending 65536 bytes
done in 2.202245291 s
speed 29758.719552189974 bytes/s
test done
```


src/bin/drderp.rs
```diff
@@ -0,0 +1,295 @@
...
+            Err(cause) => {
+                eprintln!("error accepting bidi stream {}", cause);
+                break Err(cause.into());
+            }
+        };
+    }
+}
+
+pub fn configure_derp_map() -> DerpMap {
+    // Use google stun server for now
```
dignifiedquire, line 144: lol, this is not google anymore 😅

rklaehn:
I turned it into a subcommand, as @dignifiedquire suggested:

```
❯ iroh dr-derp accept                                                                               
Run

iroh dr-derp connect 5b33f30b2b53bce0cb1c8853539fe373b95028c5783cbb62e420310219f0155e --remote-endpoint 82.76.245.176:57541 --remote-endpoint 192.168.1.135:57541

in another terminal or on another machine to connect by key and addr.
Omit the --remote-endpoint args to connect just by key.

Accepted connection. Performing test.

send: 193.63 MiB/s
recv: 193.15 MiB/s
echo: 93.73 MiB/s
```

src/main.rs
```diff
@@ -104,6 +104,13 @@ impl FromStr for ProviderRpcPort {
 #[derive(Subcommand, Debug, Clone)]
 #[allow(clippy::large_enum_variant)]
 enum Commands {
+    /// Diagnostic commands for the derp relay protocol.
+    DrDerp {
```
dignifiedquire, line 108: I believe we settled on `doctor` for now as the subcommand

rklaehn:
Renamed to doctor and added gui on the connect side, so the connect side also has something to look at.

src/hp/doctor.rs
```diff
@@ -71,38 +65,98 @@ pub enum Commands {
...
+fn update_pb(
+    pb: ProgressBar,
+    total_bytes: u64,
+    mut updates: sync::mpsc::Receiver<u64>,
+) -> tokio::task::JoinHandle<()> {
+    pb.set_position(0);
+    pb.set_length(total_bytes);
+    tokio::spawn(async move {
+        loop {
+            while let Some(position) = updates.recv().await {
```
dignifiedquire, line 87: double loop, why?

src/hp/doctor.rs
```diff
@@ -0,0 +1,609 @@
+use std::{
```
dignifiedquire, line 1: Can you add a short description of what this does roughly?

src/hp/doctor.rs
```diff
@@ -0,0 +1,609 @@
...
+#[allow(clippy::large_enum_variant)]
+pub enum Commands {
+    Report {
+        #[clap(long, default_value = "derp.iroh.computer")]
+        host_name: String,
+        #[clap(long, default_value_t = 3478)]
+        stun_port: u16,
+    },
+    Accept {
+        /// Our own private key, in hex. If not specified, a random key will be generated.
```
dignifiedquire, line 34: Does this allow for reusing the iroh key? would potentially be useful to fallback to the stored one.

src/hp/doctor.rs
```diff
@@ -0,0 +1,609 @@
...
+        #[clap(long)]
+        private_key: Option<String>,
+
+        #[clap(long, default_value_t = 1024 * 1024 * 16)]
+        size: u64,
+
+        #[clap(long, default_value_t = -1)]
```
dignifiedquire, line 41: comment on what negative numbers mean

src/hp/doctor.rs
rklaehn: https://discord.com/channels/949724860232392765/951848227139891341/1106493287013040128

src/hp/doctor.rs
rklaehn: I added an option local to use the key from the iroh dir. But I don't think it should be the default.

Maybe it should be the default for accept, but not connect? You don't want two nodes with the same peer id to try to connect, so I think random is a good default.

dignifiedquire (approved):

PR #988, opened 2023-05-10
Remove dumping packet content in Debug...

rklaehn:
...for DerpReadResult and ReceivedMessage::ReceivedPacket

This prevents seeing a giant wall of useless text when running with debug tracing. If we actually need the package contents we can do something more fancy like a hex dump of the first n chars, but I think this is already useful.


rklaehn:
We will wait some time for a new derive_more release https://github.com/JelteF/derive_more/issues/259 that includes custom debug.

PR #991, opened 2023-05-10
feat: print local endpoints on provide

dignifiedquire:


flub (approved):

dignifiedquire:
> This manual actor stuff is so verbose... But I need this as well.

do you want macros? this is how you get macros :P

rklaehn:
> > This manual actor stuff is so verbose... But I need this as well.
> 
> do you want macros? this is how you get macros :P

Something like quic-rpc using the flume transport would work as well, except that it is wasteful because it always allocates a stream backchannel even for unidirectional msgs and msgs with a single response.

But it is fine for now.

flub:
absolutely nothing wrong with boring easy to understand code

PR #995, opened 2023-05-11
[WIP] add metrics 

dignifiedquire:


Arqu:
LGTM, couldn't ask for more :D

rklaehn:
I think it would be neat to just show these counters in drderp. Then you will be able to see which counters increase and figure out what's going on without having to interpret them...

PR #996, opened 2023-05-11
Batched derp v1

rklaehn:
Combine all udp transmits that we are offered from quinn into as few derp packets as possible instead of doing a 1:1 translation.

This means fewer, larger packets to be sent to the derper, and also less handling overhead for the derper.

Encoding is currently just a u16be length prefix. This should be enough for UDP packets that are limited to 64k and should usually be smaller than the MTU.


rklaehn:
Tried this with Dr.Derp. It seems to help a lot, at least with a local derper:

image

src/hp/derp/client.rs
```diff
@@ -715,3 +720,133 @@ pub(crate) fn parse_recv_frame(frame: &[u8]) -> Result<(PublicKey, &[u8])> {
...
+        if !self.buffer.is_empty() {
+            Some(self.buffer.split().freeze())
+        } else {
+            None
+        }
+    }
+}
+
+/// Splits a packet into its component items.
+
```
dignifiedquire, line 771: stray newline

src/hp/derp/client.rs
```diff
@@ -715,3 +727,132 @@ pub(crate) fn parse_recv_frame(frame: &[u8]) -> Result<(PublicKey, &[u8])> {
...
+    fn next(&mut self) -> Option<Self::Item> {
+        use bytes::BufMut;
+        while let Some(next_bytes) = self.iter.peek() {
+            let next_bytes = next_bytes.as_ref();
+            let next_length: u16 = next_bytes.len().try_into().expect("items < 64k size");
+            if self.buffer.len() + next_length as usize + 2 > MAX_PACKET_SIZE {
+                break;
+            }
+            let length_prefix = next_length.to_be_bytes();
```
dignifiedquire, line 764: don't we use `LE` in other places of the protocol?

rklaehn: if you want LE you get LE.

src/hp/derp/client.rs
```diff
@@ -715,3 +727,130 @@ pub(crate) fn parse_recv_frame(frame: &[u8]) -> Result<(PublicKey, &[u8])> {
...
+    fn next(&mut self) -> Option<Self::Item> {
+        use bytes::BufMut;
+        while let Some(next_bytes) = self.iter.peek() {
+            let next_bytes = next_bytes.as_ref();
+            let next_length: u16 = next_bytes.len().try_into().expect("items < 64k size");
+            if self.buffer.len() + next_length as usize + 2 > MAX_PACKET_SIZE {
+                break;
+            }
+            self.buffer.put_u16_le(next_length);
+            self.buffer.put_slice(next_bytes);
```
dignifiedquire, line 765: this means we have to do a full copy for every packet, why can't we just write directly to the io instead?

rklaehn: The way to write io seems to be write_frame. It takes a &[&[u8]]. If I understand it correctly it will concatenate all the slices. Since the number of transmits per packet is not known before, you would have to allocate something in any case.

I guess you could rearrange it so that thing is smaller than the buffer, but at that point I doubt it would make a difference.

If you wanted to you could just have this format:

```
2 byte number of transmits
2 * n bytes sizes for each transmit
rest concatenated transmits
```

But I can't see how you can do this entirely without allocations given that you don't have direct access to the AsyncWrite.

rklaehn:
Here are some more saved copies: https://github.com/n0-computer/iroh/pull/997

src/hp/derp/client.rs
rklaehn: OKOK, you could do it like this:

```rust
    let mut i = 0;
    for (size, count) in PacketizeIter2::new(transmits.iter()) {
        write_frame_header(&mut writer, FrameType::SendPacket, size).await?;
        for transmit in &transmits[i..i + count] {
            let transmit = transmit.as_ref();
            writer.write_u16_le(transmit.len() as u16).await?;
            writer.write_all(transmit).await?;
        }
        i += count;
    }
```

This has an iterator that just computes the number of transmits that fit into a packet and the total size, then manually write the header and data.

This does not allocate, but it is probably *slower* because it interacts with the writer more often, whereas the packetizing code above just has 1 big call to write_all per packet.

🤷‍♂️

dignifiedquire: so the fastest should be a bufwrite wrapper with your latest version I suspect

dignifiedquire: but lets leave it as is for now, we can optimize things later, when we see bottlenecks

PR #997, opened 2023-05-11
Save some copies

rklaehn:


src/hp/derp/client.rs
```diff
@@ -224,11 +224,8 @@ impl<R: AsyncRead + Unpin> Client<R> {
                         tracing::warn!("unexpected: dropping short packet from DERP server");
                         continue;
                     }
-                    let (source, data) = parse_recv_frame(&frame_payload)?;
-                    let packet = ReceivedMessage::ReceivedPacket {
-                        source,
-                        data: Bytes::copy_from_slice(data),
```
rklaehn, line 230: This is the saved copy

dignifiedquire (approved):
save them bytes

dignifiedquire (approved):
save them bytes

PR #999, opened 2023-05-11
feat: add configuration for derp regions

dignifiedquire:


src/config.rs
```diff
@@ -0,0 +1,128 @@
...
+fn default_derp_region() -> DerpRegion {
+    DerpRegion {
+        region_id: 1,
+        nodes: vec![DerpNode {
+            name: "default-1".into(),
+            region_id: 1,
+            host_name: "derp.iroh.computer".into(),
+            stun_only: false,
+            stun_port: 3478,
+            ipv4: UseIpv4::Some("35.175.99.113".parse().unwrap()),
```
b5, line 108: move this IP to a constant & add a comment?

dignifiedquire: this is config is the constant

b5: sure, then can you add a comment clarifying these are run by n0 / iroh maintainers? Just something so folks have a sense that this is provisioned-but-default infra

PR #1000, opened 2023-05-12
fix: replacing a connection on the `derp::Server` erroneously removed the new connection as well

ramfox:
When a client connection on the server errors on write or read, we send a message to the `ServerActor` to remove that client. This was causing a vicious cycle:
- The client connection to peer A is not erroring
- However, the client decides to create a new connection & we add that client connection to the server, replacing the original connection to peer A
- the connection to peer A "shutdown". The shutdown process tries to clean up all the peices of the connection (the shutdown can get triggered by an error on the reader & we need to remove the writer, for example). To do this, we send a message to the `ServerActor` to shutdown peer A. But, at this point, the Peer A that we have in the `Server` is actually the new connection! So we shut down the new connection!

We already have pieces in place here to mitigate this mistake: we are already generating unique connection id numbers for each client connection in anticipation of allowing multiple collections to a client anyway, so we use those `conn_num`s to check if we are closing the correct connection down before we remove the client connection.


src/hp/derp/server.rs
```diff
@@ -952,4 +956,109 @@ mod tests {
...
+
+        // create client b and connect it to the server
+        let (new_reader_b, new_writer_b, new_client_b_builder) = make_test_client(key_b);
+        let handler = server.client_conn_handler(Default::default());
+        let handler_task =
+            tokio::spawn(async move { handler.accept(new_reader_b, new_writer_b).await });
+        let new_client_b = new_client_b_builder.build(None).await?;
+        handler_task.await??;
+
+        // assert!(client_b.recv().await.is_err());
```
dignifiedquire, line 1026: what about this assert?

dignifiedquire (approved):
looks good, just some clarification in the test

PR #1001, opened 2023-05-12
ref(Conn/PeerMap): Introduce QuicMappedAddress

flub:
This does a little tidying up of the address we tell the QUIC layer to
use to dial a peer:

- It introduces a newtype for it, making it very clear what you are
  working with and avoiding accidentally using it in the wrong place.

- It adds this to the PeerMap, documenting the peer map a bit better.

- It also change the generation of this address a bit: make it an RFC
  compliant Unique Local Address assigning out own Global Id.  This
  should allow more universally detecting if the address leaks.

### DIFF
--- a/src/hp/magicsock/conn.rs
+++ b/src/hp/magicsock/conn.rs
@@ -6,7 +6,7 @@ use std::{
     ops::Deref,
     pin::Pin,
     sync::{
-        atomic::{AtomicBool, AtomicU16, Ordering},
+        atomic::{AtomicBool, AtomicU16, AtomicU64, Ordering},
         Arc,
     },
     task::{Context, Poll, Waker},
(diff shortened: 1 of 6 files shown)


src/hp/magicsock/conn.rs
```diff
@@ -140,6 +149,7 @@ impl Deref for Conn {
 pub struct Inner {
     actor_sender: flume::Sender<ActorMessage>,
     /// Sends network messages.
+    // TODO: send transmits *together* with their destination
```
dignifiedquire, line 152: the transmit includes the destination, what do you mean with this comment?

flub: I meant that I'd like the lookup of the real destination to be at the time of `poll_send` and include the destination along.  Because that's where the error should be propagated.  To be fair I'm not sure if that makes sense anymore so i'll delete the comment.

PR #1003, opened 2023-05-12
fix hanging tests and allow for single threaded executor

dignifiedquire:


src/provider/mod.rs
```diff
@@ -654,13 +661,13 @@ impl RpcHandler {
     async fn id(self, _: IdRequest) -> IdResponse {
         IdResponse {
             peer_id: Box::new(self.inner.keypair.public().into()),
-            listen_addrs: self.inner.listen_addresses().await.unwrap_or_default(),
+            listen_addrs: self.inner.listen_addresses().unwrap_or_default(),
             version: env!("CARGO_PKG_VERSION").to_string(),
         }
     }
     async fn addrs(self, _: AddrsRequest) -> AddrsResponse {
```
rklaehn, line 668: this does not have to be async anymore, right?

src/provider/mod.rs
```diff
@@ -654,13 +661,13 @@ impl RpcHandler {
     async fn id(self, _: IdRequest) -> IdResponse {
```
rklaehn, line 661: Also not async anymore

rklaehn (approved):
So why was it failing before?

This mix of an actor approach and local caching is a bit weird, even though less async makes it more pleasant to use.

src/provider/mod.rs
dignifiedquire: yes, but the rpc code was upset when I made it not async 🤷

dignifiedquire:
> This mix of an actor approach and local caching is a bit weird,

It is, would love to hear any suggestions on alternatives. This was the least bad one I could come up with.

dignifiedquire:
armv7 should be fixed when we upgrade socket2: https://github.com/rust-lang/socket2/pull/434

flub (approved):

PR #1007, opened 2023-05-12
ref(netcheck): statically declare udp bind addresses

flub:
This is just a PR to keep them smaller.  The main chainge is that the
UDP bind sockets to do STUN on are now directly statically declared.
Functionally this is exactly the same but without the extra state and
function calls involved.

Also adds some doc comments and mostly syntactic refactoring.

### DIFF
--- a/src/hp/netcheck.rs
+++ b/src/hp/netcheck.rs
@@ -4,7 +4,7 @@
 use std::{
     collections::HashMap,
     fmt::Debug,
-    net::{IpAddr, SocketAddr},
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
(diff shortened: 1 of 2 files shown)


src/hp/netcheck.rs
```diff
@@ -162,7 +161,6 @@ impl Client {
                 current_hair_tx: None,
             },
             skip_external_network: false,
-            udp_bind_addr: "0.0.0.0:0".parse().unwrap(),
```
flub, line 165: This is the only place this field was ever initialised, and always to the same.  It was then read back via some functions.  So those functions and this state is removed, which is the main change in this PR.

dignifiedquire (approved):

PR #1008, opened 2023-05-12
Add some metrics counters to the doctor

rklaehn:
Add some metrics to the doctor

Which ones would you find most useful, @ramfox or @dignifiedquire ?

I would make a nice console gui for them then. I guess group by send and recv?

image


dignifiedquire:
something like
```
Derp:
  send: 54 MiB
  recv: 54 MiB
Ipv4:
  send: 54 MiB
  recv: 54 MiB
Ipv6:
  send: 54 MiB
  recv: 54 MiB
```

dignifiedquire:
fyi quinn also has stats, which would be interesting to show

rklaehn:
I will make a nice panel. What you want to see probably depends on the specific scenario, so I doubt that you can come up with something perfect here.

But you should be able to clearly see the derp -> udp switchover etc.

dignifiedquire (approved):

dignifiedquire:
looks like some compile errors

rklaehn:
should be fixed now. Also did clippy.

PR #1016, opened 2023-05-15
ref: Move packet building into conn

rklaehn:
As suggested by @ramfox in https://github.com/n0-computer/iroh/pull/830#discussion_r1191830689

This undoes a lot of the changes to tests that were done in the initial batching PR. The packetize and packet split iter moved to the conn.

### DIFF
--- a/src/hp/derp/client.rs
+++ b/src/hp/derp/client.rs
@@ -72,7 +72,7 @@ impl<R: AsyncRead + Unpin> Client<R> {
     /// Sends a packet to the node identified by `dstkey`
     ///
     /// Errors if the packet is larger than [`super::MAX_PACKET_SIZE`]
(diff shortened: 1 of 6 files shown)


dignifiedquire:
I am not seeing any place where the packets are split on receiving, did you miss that piece?

rklaehn:
> I am not seeing any place where the packets are split on receiving, did you miss that piece?

I did not change anything about this. It is in process_derp_read_result. The only remaining place where PacketSplitIter is still used.

dignifiedquire:
ahh, right :)

dignifiedquire:
did you test performance? I assume no change?

rklaehn:
> did you test performance? I assume no change?

I did not test perf. It does the same thing, just in a different place. Same number of allocations etc.

Although to be fair I think we are not at the point where we have to worry about allocations. Hopefully we will get there eventually...

PR #1020, opened 2023-05-16
fix: process incoming IP packets in a seperate task

dignifiedquire:
This fixes receiving STUN packets during an endpoint update. Before they processing of incoming STUN packets would be interrupted due to the main actor loop waiting on the endpoint update.

### DIFF
--- a/src/hp/magicsock/conn.rs
+++ b/src/hp/magicsock/conn.rs
@@ -20,7 +20,7 @@ use futures::{future::BoxFuture, Stream, StreamExt};
 use quinn::AsyncUdpSocket;
 use rand::{seq::SliceRandom, Rng, SeedableRng};
 use tokio::{
-    sync::{self, Mutex},
+    sync::{self, mpsc, Mutex},
     task::JoinHandle,
     time,
 };
@@ -181,6 +181,8 @@ pub struct Inner {
     closing: AtomicBool,
     /// Close was called.
     closed: AtomicBool,
+    /// Are we currently processing STUN responses (set during a netcheck report).
+    enable_stun_packets: AtomicBool,
 }
 
 impl Inner {
@@ -271,6 +273,7 @@ impl Conn {
             network_send_wakers: std::sync::Mutex::new(None),
             actor_sender: actor_sender.clone(),
             network_sender,
+            enable_stun_packets: AtomicBool::new(false),
         });
 
         let conn = inner.clone();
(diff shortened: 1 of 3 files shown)


src/hp/magicsock/conn.rs
```diff
@@ -181,6 +181,8 @@ pub struct Inner {
     closing: AtomicBool,
     /// Close was called.
     closed: AtomicBool,
+    /// Are we currently processing STUN responses (set during a netcheck report).
+    enable_stun_packets: AtomicBool,
```
flub, line 185: I wonder if we should replace this by a method on the `netcheck::Client` instead of tracking this here.  It seems like a better place to track whether the netcheck client is running or not.

It's a bit difficult though, `netcheck::Client` isn't `Clone` right now, maybe we should make it `Clone`, I think it would simplify things here.

rklaehn:
One change is while before there was a receive_stun_packet fn that would enqueue the msg, now there is a channel that you can grab to send msgs. Was there anything wrong with the approach before, or is this just nicer?

Also, before the netcheck actor had its own message enum with just a single case. You removed that now, but I didn't really mind it.

flub (approved):
I'm not a fan of removing `netcheck::ActorMessage`, I thought it was a very readable struct.  Otherwise this looks fine to me.  I was looking for a reason to make `netcheck::Client` `Clone`, I think now there might be as it would remove some state tracing from where it doesn't need to be.

src/hp/magicsock/conn.rs
rklaehn: Can we just always forward stun packets and let the receiver decide whether they are needed?

dignifiedquire:
> Calling async fns from inside an actor is usually frowned upon.

that seems very unrealistic for most uses I have for this pattern

rklaehn:
> > Calling async fns from inside an actor is usually frowned upon.
> 
> that seems very unrealistic for most uses I have for this pattern

🤷‍♂️ I don't make the rules. https://en.wikipedia.org/wiki/Actor_model

"In response to a message it receives, an actor can: make local decisions, create more actors, send more messages, and determine how to respond to the next message received."

Doing async io inside the message handler can in my opinion be perfectly valid. I am not religious about this. But then it is no longer something that conforms 100% to the actor model.

dignifiedquire:
> But then it is no longer something that conforms 100% to the actor model.

fair, I honestly just call these actors because I have no better name for it 😅

PR #1021, opened 2023-05-16
ref: remove unneeded async

rklaehn:
no real async fns are being called from inside the netcheck actor.


flub (approved):
this is probably going to give me merge conflicts, but if there's no need for them to be async then let them not be async.

dignifiedquire (approved):
less async always better

PR #1022, opened 2023-05-16
ref(netcheck): Log and shut down stun listeners

flub:
This makes sure the listener tasks for the STUN sockets are stopped.
Before they just ran forever.

Also tweaks a lot of logging to give more context.


dignifiedquire (approved):
this is some nice instrumentation :)

PR #1023, opened 2023-05-17
ref(netcheck): Bring back the ActorMessage

flub:


rklaehn (approved):
I like this.

Usually the cheaply cloneable thing that you can use to send msgs is called an actor ref. But msg_sender works as well.

flub:
> Usually the cheaply cloneable thing that you can use to send msgs is called an actor ref. But msg_sender works as well.

Yeah, I want to change that just be the `Client` in a followup PR

PR #1026, opened 2023-05-17
feat(conn): improve shutdown of IO loop

dignifiedquire:


flub (approved):
I'm assuming the big match statement is just copied - didn't review that bit.  cancellation token stuff looks good

src/hp/magicsock/conn.rs
```diff
@@ -2739,6 +2685,94 @@ impl Actor {
...
+    ip_sender: mpsc::Sender<IpPacket>,
+) {
+    loop {
+        tokio::select! {
+            biased;
+            _ = cancel_token.cancelled() => {
+                break;
+            }
+            msg = ip_stream.next() => {
+                match msg {
```
rklaehn, line 2703: You could use a let if to make this not quite as deep...

```rust
let Some(ip_msgs) = msg else { break }
```

rklaehn (approved):

PR #1027, opened 2023-05-17
X hp test fix

rklaehn:
With this tests pass at least on my machine.


dignifiedquire:
still seems to fail on cli, at least cross

rklaehn:
> still seems to fail on cli, at least cross

Yes, I don't think this fixes all things. What I would like to see is at least linux passing. But it is stuck somehow.

image

PR #1028, opened 2023-05-17
ref(netcheck): make netcheck a long-running actor

flub:
This turns the netcheck into a long running actor:

- The `Client` creates the actor which runs in a background tokio task.
- The `Client` can be cloned to get more handles to the netcheck actor.
- Once `Client` is dropped the actor is stopped.
- Now you can always pass stun packets to to netcheck: it manages the state internally.

This last point is the motivation, it simplifies how things are wired up and makes it easier to keep the right state in the netcheck actor where it belongs.  This will also make future changes to netcheck more contained.

It turns the former netcheck actor into more of a full-scale actor.  The `ReportState` is not following this pattern yet.

### DIFF
--- a/src/hp/netcheck.rs
+++ b/src/hp/netcheck.rs
@@ -20,7 +20,7 @@ use rand::seq::IteratorRandom;
 use tokio::{
     net::UdpSocket,
     sync::{self, broadcast, mpsc, oneshot, RwLock},
-    task::JoinSet,
+    task::{AbortHandle, JoinSet},
     time::{self, Duration, Instant},
 };
 use tokio_util::sync::CancellationToken;
@@ -151,13 +151,13 @@ pub struct Client {
 
 #[derive(Debug)]
 struct ClientDropGuard {
-    addr: ActorAddr,
+    task: AbortHandle,
 }
 
 impl Drop for ClientDropGuard {
     fn drop(&mut self) {
-        // If we can't send the actor is already not running so that's fine.
-        self.addr.blocking_send(ActorMessage::Exit).ok();
+        debug!("netcheck actor finished");
+        self.task.abort();
     }
 }
 
@@ -191,10 +191,12 @@ impl Client {
     pub async fn new(port_mapper: Option<portmapper::Client>) -> Result<Self> {
         let mut actor = Actor::new(port_mapper)?;
         let addr = actor.addr();
-        tokio::spawn(async move { actor.main().await });
+        let task = tokio::spawn(async move { actor.main().await });
         Ok(Client {
-            addr: addr.clone(),
-            _drop_guard: ClientDropGuard { addr }.into(),
+            addr: addr,
+            _drop_guard: Arc::new(ClientDropGuard {
+                task: task.abort_handle(),
+            }),
         })
     }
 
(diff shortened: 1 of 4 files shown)


src/hp/netcheck.rs
```diff
@@ -1106,85 +1065,298 @@ impl ProbeReport {
...
+        ActorAddr {
+            sender: self.sender.clone(),
+        }
+    }
+
+    /// Run the actor.
+    ///
+    /// It will now run and handle messages.  Once the connected [`Client`] (including all
+    /// its clones) is dropped this will terminate.
+    async fn main(&mut self) {
```
dignifiedquire, line 1221: can we call this `run` for consistency with the other methods in the codebase?

src/hp/netcheck.rs
```diff
@@ -1106,85 +1065,298 @@ impl ProbeReport {
...
+                ActorMessage::ReportReady { report, derp_map } => {
+                    self.handle_report_ready(*report, derp_map);
+                    self.in_flight_stun_requests.clear();
+                }
+                ActorMessage::ReportAborted => {
+                    self.in_flight_stun_requests.clear();
+                    self.current_check_run.take();
+                }
+                ActorMessage::StunPacket { payload, from_addr } => {
+                    // TODO: drop package if no check is running
```
dignifiedquire, line 1243: still relevant?

src/hp/netcheck.rs
```diff
@@ -1106,85 +1065,298 @@ impl ProbeReport {
...
-                    in_flight.insert(inf.tx, inf);
+        let report_state = self
+            .create_report_state(&derp_map, stun_sock_v4, stun_sock_v6)
+            .await
+            .context("failed to create ReportState")?;
+        let port_mapper = self.port_mapper.clone();
+        let skip_external = self.skip_external_network;
+        let resolver = self.dns_resolver.clone(); // not a cheap clone, maybe Arc it
+        let addr = self.addr();
+        tokio::spawn(async move {
```
dignifiedquire, line 1324: Would be nice to track this task, and make sure it gets aborted on drop as well

src/hp/netcheck.rs
```diff
@@ -1567,6 +1817,7 @@ mod tests {
             .get_report(dm, None, None)
             .await
             .context("failed to get netcheck report")?;
+        dbg!(&r);
```
dignifiedquire, line 1820: stray `dbg`

dignifiedquire (commented):
some small nits, otherwise looks great

src/hp/netcheck.rs
flub: sure, i had both functions for a while so needed a new name for the new func as i was converting... then didn't move it back

src/hp/netcheck.rs
flub: Yeah, I was wondering about that as well.  But kind of wanted to defer that to a followup PR.  Created https://github.com/n0-computer/iroh/issues/1037 for this as there's a few options to consider.

src/hp/netcheck.rs
flub: oh, this was on purpose... do you mind leaving these in or would rather have them removed?  It's much nicer when running the test with --nocapture.

dignifiedquire (approved):

PR #1029, opened 2023-05-17
Drop conn

rklaehn:


src/hp/magicsock/conn.rs
```diff
@@ -478,6 +478,23 @@ impl Conn {
...
+    fn drop(&mut self) {
+        if self.is_closed() {
+            return;
+        }
+        // the actor holds a strong ref to us, so 2 means we're the last conn
+        if Arc::strong_count(&self.inner) > 2 {
+            return;
+        }
+        self.inner.closing.store(true, Ordering::Relaxed);
+        self.inner.closed.store(true, Ordering::SeqCst);
```
dignifiedquire, line 491: This leaves things in a less than ideal place, as the assumption is that things have been cleanly stopped once `closed = true`. So we should improve the internal structure if we are doing shutdown like this.

rklaehn: I just copied the code. But who will ever read these flags after this? The actor task is aborting, and the last Conn has just been dropped.

It would only matter if the actor task is in some sync code (which can not be aborted) and then reads these flags.

rklaehn:
I think this is nicer. It does not matter what you do - once the OwnedJoinHandle gets dropped the task gets aborted. Even if you call close and then drop the future before polling it to completion...

src/tokio_util.rs
```diff
@@ -402,3 +402,27 @@ pub(crate) async fn read_as_bytes(reader: &mut Either<Cursor<Bytes>, File>) -> i
         }
     }
 }
+
+/// A join handle that owns the task it is running, and aborts it when dropped.
+#[derive(Debug)]
+pub(crate) struct OwnedJoinHandle<T>(tokio::task::JoinHandle<T>);
```
flub, line 408: bikeshedding: `OwningJoinHandle`.  The handle owns the task, the handle itself is not owned by something.

Though I'd be tempted to even call this `AbortingJoinHandle`

flub (approved):
I don't care much about the bikeshedding, if you want to merge with current naming things will be fine too.

PR #1031, opened 2023-05-18
Refactor derp handling in conn

dignifiedquire:


src/hp/magicsock/derp_actor.rs
```diff
@@ -0,0 +1,810 @@
...
+use super::conn::{ActorMessage, Inner};
+
+/// How long a non-home DERP connection needs to be idle (last written to) before we close it.
+const DERP_INACTIVE_CLEANUP_TIME: Duration = Duration::from_secs(60);
+
+/// How often `clean_stale_derp` runs when there are potentially-stale DERP connections to close.
+const DERP_CLEAN_STALE_INTERVAL: Duration = Duration::from_secs(15);
+
+pub(super) enum DerpActorMessage {
+    Send {
```
rklaehn, line 33: Can we have small comments about each enum case? I mean in case of `Send` it is self explaning. But for e.g. `NotePreferred` or `MaybeCloseDerpsOnRebind` a bit of text help.

src/hp/magicsock/derp_actor.rs
```diff
@@ -0,0 +1,810 @@
...
+        peer: Option<key::node::PublicKey>,
+    },
+    CloseOrReconnect {
+        region_id: u16,
+        reason: &'static str,
+    },
+    RemovePeerRoutes {
+        peers: Vec<key::node::PublicKey>,
+        region: u16,
+        derp_client: derp::http::Client,
```
rklaehn, line 49: Not quite sure why the client is included here. Comment would be helpful. Is the client passed in every time the actor needs to do something, or does the ownership of this client somehow pass to the actor?

src/hp/magicsock/derp_actor.rs
```diff
@@ -0,0 +1,810 @@
...
+    receiver: mpsc::Receiver<DerpActorMessage>,
+    /// DERP regionID -> connection to a node in that region
+    active_derp: HashMap<u16, ActiveDerp>,
+    /// Contains optional alternate routes to use as an optimization instead of
+    /// contacting a peer via their home DERP connection.  If they sent us a message
+    /// on a different DERP connection (which should really only be on our DERP
+    /// home connection, or what was once our home), then we remember that route here to optimistically
+    /// use instead of creating a new DERP connection back to their home.
+    derp_route: HashMap<key::node::PublicKey, DerpRoute>,
+    msg_sender: flume::Sender<ActorMessage>,
```
rklaehn, line 93: This is to send messages to some actor. But not to self, because that would be `DerpActorMessage`. Maybe it would be helpful to prefix the namespace.

Also, why are some actor connections using flume and some tokio mpsc?

src/hp/magicsock/derp_actor.rs
```diff
@@ -0,0 +1,810 @@
...
+                            self.close_all_derp("conn-close").await;
+                            break;
+                        }
+                    }
+                }
+                _ = cleanup_timer.tick() => {
+                    trace!("tick: cleanup");
+                    self.clean_stale_derp().await;
+                }
+                else => {
```
rklaehn, line 156: Nice. I didn't know that you can pattern match in `select!`. So this is for when self.receiver.recv() returns None?

src/hp/magicsock/endpoint.rs
```diff
@@ -560,15 +559,15 @@ impl Endpoint {
                     let ep = sp.to;
                     let region_id = usize::from(ep.port());
                     // FIXME: this creates a deadlock as it needs to interact with the run loop in the conn::Actor
-                    let region_code = self.get_derp_region(region_id).await.map(|r| r.region_code);
+                    // let region_code = self.get_derp_region(region_id).await.map(|r| r.region_code);
```
rklaehn, line 562: Why is this commented out? We don't want to deal with regions yet?

rklaehn (approved):
Very nice restructuring.

Want me to try it out before we merge to x-hp? I think we want this in any case. We can also merge now and then sort out any regressions on x-hp...

src/hp/magicsock/derp_actor.rs
rklaehn: (going off to change some code in my interprocess stuff now...)

dignifiedquire: > So this is for when self.receiver.recv() returns None?

this will catch anything not matched by previous statements, so in this case `None`

src/hp/magicsock/endpoint.rs
dignifiedquire: because of the `FIXME` above, I haven't figured out a clean way yet, and this is only for CLI pings, which we currently don't thread through yet.

PR #1038, opened 2023-05-19
fix: better handling of ipv4 only setups

dignifiedquire:


dignifiedquire:
The last failure is due to https://github.com/quinn-rs/quinn/issues/1568. It turns out we can't bind to IPv6 currently inside some cross executions because of that issue.

PR #1040, opened 2023-05-19
chore: move cross CI to our runners

Arqu:


Arqu:
FYI tests intentionally left alone in this PR (but verified to work with some of the fix branches)

dignifiedquire (approved):

dignifiedquire:
cli tests on linux were passing consistently, that seems to be a regression.

Arqu:
> cli tests on linux were passing consistently, that seems to be a regression.

There are fixes from @rklaehn's work to make it all pass. It worked because the old runners were slow enough to mostly pass without issues.

PR #1045, opened 2023-05-24
ref(metrics): Document metrics and remove macro_use

flub:
This improves some documentation by explaining the different metric
types and the difference between record and observe.  It's also a bit
opinionated about metric types.

It also restructures the code a bit to make use of the 2018-edition
macro scoping rules, so macros are not magically available and use can
import them from the metrics module, which gives much better context.

### DIFF
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -6,9 +6,12 @@
 //!
 //! To enable metrics collection, call `init_metrics()` before starting the service.
 //!
-//! To record a metric, use the `record!` macro with the metric and the value to record.
-//! To increment a metric by 1, use the `inc!` macro with the metric.
-//! To observe a metric, use the `observe!` macro with the metric and the value to observe.
+//! - To record a **gauge** ( or a **counter**), use the [`record`] macro with a value.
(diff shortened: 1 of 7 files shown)


src/metrics.rs
```diff
@@ -6,9 +6,12 @@
 //!
 //! To enable metrics collection, call `init_metrics()` before starting the service.
 //!
-//! To record a metric, use the `record!` macro with the metric and the value to record.
-//! To increment a metric by 1, use the `inc!` macro with the metric.
-//! To observe a metric, use the `observe!` macro with the metric and the value to observe.
+//! - To record a **gauge** ( or a **counter**), use the [`record`] macro with a value.
+//!   Don't use gauges though.
+//! - To increment a **counter** by 1, use the [`inc`] macro.
+//! - For **histograms** (or **summaries**, but don't use those either) use the [`observe`]
```
Arqu, line 12: Well there are use cases for histograms, and summaries are not properly supported yet.

Arqu (approved):
Lovely, LGTM.

PR #1047, opened 2023-05-24
fix: add entry in peer_map for unknown ping sender

rklaehn:
Note that I really don't know if this is a good idea, I just don't see how ping can work at all otherwise...

### DIFF
--- a/src/hp/magicsock/conn.rs
+++ b/src/hp/magicsock/conn.rs
@@ -1897,8 +1897,8 @@ impl Actor {
         }
 
         let sender = key::node::PublicKey::from(source);
+        let mut unknown_sender = false;
         if self.peer_map.endpoint_for_node_key(&sender).is_none() {
-            let mut abort = true;
             // Disco Ping from seen endpoint without node key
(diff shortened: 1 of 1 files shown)


src/hp/magicsock/conn.rs
```diff
@@ -1966,6 +1962,20 @@ impl Actor {
...
             disco::Message::Ping(ping) => {
                 inc!(MagicsockMetrics::RecvDiscoPing);
+                // if we get here we got a valid ping from an unknown sender
+                // so insert an endpoint for them
+                if unknown_sender {
+                    self.peer_map.insert_endpoint(EndpointOptions {
+                        conn_sender: self.conn.actor_sender.clone(),
+                        conn_public_key: self.conn.public_key.clone(),
+                        public_key: Some(sender.clone()),
+                        derp_addr: if derp_node_src.is_some() {
```
dignifiedquire, line 1972: this check should be `if is_derp` (I think)

rklaehn: isn't that the same? But it looks nicer, so will change.

PR #1049, opened 2023-05-25
fix: handle hairpining timeout properly

dignifiedquire:


flub (approved):

dignifiedquire:
@flub pushed a small additional simplification

src/hp/netcheck.rs
```diff
@@ -793,21 +798,20 @@ impl ReportState {
...
-
-        tokio::select! {
-            biased;
-            _ = self.got_hair_stun.recv() => {
-                Some(true)
-            }
-            _ = hair_timeout => {
-                debug!("hair_check timeout");
-                Some(false)
+        match self.hair_timeout {
```
flub, line 801: doesn't clippy want this as an if-let?  I don't mind though

flub (approved):

src/hp/netcheck.rs
dignifiedquire: it seems happy 😅

Arqu (approved):
Seems to have fixed the issue 👍

PR #1050, opened 2023-05-25
fix(netcheck): hairpin timeout should wait

flub:
notify_one makes sure that future waiters are notified right away.
notify_waiters only wakes currently waiting waiters and is too brittle
for us.  we only have one waiter so let's use notify_one for now until
we refactor this to not use Notify.


flub:
superseded by #1049

PR #1051, opened 2023-05-26
ref(netcheck): Simplify and fix hairpinning

flub:
This removes a lot of the state everywhere to support the hairpinning,
instead using the same state we already have to handle other STUN
responses.  It adds a test to verify hairpinning works.

It also fixes a bug where the IPv6 STUN probe would send packets to an
IPv4-only STUN server.

### DIFF
--- a/src/hp/netcheck.rs
+++ b/src/hp/netcheck.rs
@@ -399,7 +399,7 @@ async fn measure_icmp_latency(
     // STUN and then using that IP.
     let node_addr = get_derp_node_addr(resolver, node, ProbeProto::Ipv4)
         .await
-        .ok_or_else(|| anyhow::anyhow!("no address for node {}", node.name))?;
+        .context("no derp node addr")?;
 
     debug!(
         "ICMP ping start to {} with payload len {} - derp {} {}",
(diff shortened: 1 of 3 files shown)


src/hp/netcheck.rs
```diff
@@ -1457,53 +1461,49 @@ impl Actor {
     /// If there are currently no in-flight stun requests registerd this is dropped,
     /// otherwise forwarded to the probe.
     fn handle_stun_packet(&mut self, pkt: &[u8], src: SocketAddr) {
+        trace!(%src, "received STUN packet");
         if self.in_flight_stun_requests.is_empty() {
             return;
         }
+        dbg!(&self.in_flight_stun_requests);
```
dignifiedquire, line 1468: left over debug?

src/hp/netcheck.rs
```diff
@@ -2096,4 +2097,41 @@ mod tests {
 
         Ok(())
     }
+
+    #[tokio::test]
+    async fn test_hairpin() -> Result<()> {
+        setup_logging();
+
+        // Setup STUN server and create derpmap.
```
dignifiedquire, line 2105: I don't quite get the test setup, could you add a couple of lines of how the flow of packets in here mirrors how hair pinning is supposed to work?

dignifiedquire:
ci tests are sad because of debug statements

PR #1055, opened 2023-05-30
ref: Introduce newtypes for IPv4 and IPv6 bound UdpSocket

flub:
This introduces Udp4Socket and Udp6Socket newtypes for bound IPv4 and
IPv6 sockets.  Since we handle those sockets almost always wrapped in
an Arc so we can clone it to send on it from other places then we recv
from it also includes the Arc wrapping.

This is a trial for the netcheck code, I think it seems nice to have
there.  However I suspect it also makes sense to push this into `Conn`
itself, but would like to hear from someone more familiar with that
code.  In that case the location of those types should probably be
somewhere else as well, please suggest locations.


dignifiedquire:
I am torn on this, if we would want to do this we would likely need to make `RebindingUdpConn` be generic over the socket or sth like that. Which in total would add a decent amount of overhead in code. While I like being able to track the fact that these are v4 vs v6 sockets, I am not sure this is such a big issue that it warrants this. But it is nice to have some additional type safety. So I am torn

flub:
In this PR `bind_local_stun_socket` is also code that's generic over both.  Instead of writing this as truly generic code I chose to instead convert between `UdpSocket` and `Udp[46]Socket` on the boundary.  It already is a type that can handle it generic so no need to reinvent this, and the conversions aren't so bad and read ok.

flub:
PTAL

I've make this a netcheck-internal thing.  Trying to do the generic thing in Conn indeed just leads to a lot of pain, but inside netcheck it's kind of nice.  OTOH if you're still unconvinced I'm also happy to close this PR.

flub:
I still kind of like this, but it's so old that rebasing is probably more work than creating a new one.  Still here for peeking at.

PR #1057, opened 2023-06-01
fix: make sure to use the config by default in iroh doctor report

rklaehn:
- by default, use config
- if explicitily provided, use stun_host and stun_port
- fail if config does not contain a derp map and nothing was explicitly provided

### DIFF
--- a/src/doctor.rs
+++ b/src/doctor.rs
@@ -54,9 +54,12 @@ impl std::str::FromStr for PrivateKey {
 #[derive(Subcommand, Debug, Clone)]
 #[allow(clippy::large_enum_variant)]
 pub enum Commands {
+    /// Report on the current network environment, using either an explicitly provided stun host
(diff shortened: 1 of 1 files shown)


src/doctor.rs
```diff
@@ -54,9 +54,11 @@ impl std::str::FromStr for PrivateKey {
 #[derive(Subcommand, Debug, Clone)]
 #[allow(clippy::large_enum_variant)]
 pub enum Commands {
+    /// Report on the current network environment, using either an explicitly provided stun host
+    /// or the settings from the config file.
     Report {
-        #[clap(long, default_value = "derp.iroh.computer")]
-        host_name: String,
+        #[clap(long)]
+        stun_host: Option<String>,
```
dignifiedquire, line 61: maybe add a note that setting this will disable derp?

Arqu (approved):
LGTM

PR #1058, opened 2023-06-01
Switch spawning subcommands to the duct crate

rklaehn:


flub (approved):
So that macro-removing PR was just to get me mellow enough to introduce another macro?  :stuck_out_tongue:

rklaehn:
> So that macro-removing PR was just to get me mellow enough to introduce another macro? 😛

Well, it is not my macro. But you are right - removed it and replaced it with a fn call.

flub:
oh, didn't even realise they provide a function option.  nice

PR #1060, opened 2023-06-01
tests: Add test utilities to configure logging

flub:
This is useful in many tests, not just netcheck.  Make it more globally available.

It also makes it better to enable by default: it just always logs all modules on the TRACE level.  But it is properly captured by the test runner and only shown if the test fails.

Potentially this is too verbose for some tests.  Variants can be built that allow configuring the logging level or even per-modules logging level as required.  But no need for that just yet.

### DIFF
--- a/src/test_utils.rs
+++ b/src/test_utils.rs
@@ -1,0 +1,40 @@
+//! Internal utilities to support testing.
+
+use tracing::level_filters::LevelFilter;
+use tracing_subscriber::layer::{Layer, SubscriberExt};
+use tracing_subscriber::util::SubscriberInitExt;
+
+/// Configures logging for the current test.
+///
+/// This configures logging that will interact well with tests: logs will be captured by the
+/// test framework and only printed on failure.
+///
+/// The logging is unfiltered, it logs all crates and modules on TRACE level.  If that's too
+/// much consider if your test is too large (or write a version that allows filtering...).
+#[must_use = "The tracing guard must only be dropped at the end of the test"]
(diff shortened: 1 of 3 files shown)


src/test_utils.rs
```diff
@@ -0,0 +1,40 @@
...
+/// This configures logging that will interact well with tests: logs will be captured by the
+/// test framework and only printed on failure.
+///
+/// The logging is unfiltered, it logs all crates and modules on TRACE level.  If that's too
+/// much consider if your test is too large (or write a version that allows filtering...).
+#[must_use = "The tracing guard must only be dropped at the end of the test"]
+pub(crate) fn setup_logging() -> tracing::subscriber::DefaultGuard {
+    let log_layer = tracing_subscriber::fmt::layer()
+        .with_writer(|| TestWriter)
+        .with_filter(LevelFilter::TRACE);
```
rklaehn, line 18: So it is no longer possible to configure log level with RUST_LOG? I find trace to be a bit too much usually...

rklaehn:
I think this will somewhat interfere with my workflow. I like to use println!, but that will now be drowned in all the log msgs. Also, I find it useful to dial the log level using RUST_LOG=info/error/debug to get an overview.

flub:
PTAL: I updated this to use RUST_LOG if it is present but fall back to logging everything on trace level if it is not set.  This allows us full logs on CI but still let's us customise it when manually poking at things.

dignifiedquire (approved):

PR #1065, opened 2023-06-02
fix: online stun test

rklaehn:
maybe it helps against test flakiness of `test_google_stun`

Servers from https://gist.github.com/mondain/b0ec1cf5f60ae726202e

### DIFF
--- a/src/hp/netcheck.rs
+++ b/src/hp/netcheck.rs
@@ -663,10 +663,10 @@ impl ReportState {
                         Some(Ok(probe_report)) => {
                             debug!("finished probe: {:?}", probe_report);
                             match probe_report.probe {
-                                Probe::Https { reg, .. } => {
+                                Probe::Https { region, .. } => {
                                     if let Some(delay) = probe_report.delay {
                                         let mut report = self.report.write().await;
-                                        let l = report.region_latency.entry(reg.region_id).or_insert(delay);
+                                        let l = report.region_latency.entry(region.region_id).or_insert(delay);
                                         if *l >= delay {
                                             *l = delay;
                                         }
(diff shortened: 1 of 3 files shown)


flub (approved):
I'm unsure, i worry the log output will be more messy and as l.google.com is already loadbalanced between datacenters with each datacenter having multiple servers running.  but otoh, who knows?

src/hp/netcheck.rs
```diff
@@ -1822,17 +1822,39 @@ mod tests {
...
-        let derp_port = 0;
-        let mut dm = DerpMap::default_from_node(
-            host_name,
-            stun_port,
-            derp_port,
-            UseIpv4::None,
-            UseIpv6::None,
+
+        let stun_servers = vec![
+            ("stun.l.google.com", 19302, 0),
```
flub, line 1827: can you anchor these at the root?  `stun.l.google.com.`?

rklaehn: ?

rklaehn:
Hm, doesn't seem to help...

```
failures:
    hp::netcheck::tests::test_google_stun
```

src/hp/netcheck.rs
flub: because we do DNS lookups if you specify the stun server by hostname.  not having it anchored at the root means that the local "search" directives (i forget the official name) of your DNS configuration are applied.  e.g. on many home networks that would result in lookups to `stun.l.google.com.lan` or `stun.l.google.com.local`, as well as the intended `stun.l.google.com`.

By using `stun.l.google.com.` you avoid this.

dignifiedquire (approved):

PR #1067, opened 2023-06-02
feat: add api to list collections

dignifiedquire:
This splits out listing content by blobs and collections

Screenshot 2023-06-02 at 12 03 13

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -18,7 +18,7 @@ use iroh::protocol::{GetRequest, RangeSpecSeq};
 use iroh::provider::{Database, Provider, Ticket};
 use iroh::rpc_protocol::*;
 use iroh::rpc_protocol::{
-    ListRequest, ProvideRequest, ProviderRequest, ProviderResponse, ProviderService, VersionRequest,
+    ProvideRequest, ProviderRequest, ProviderResponse, ProviderService, VersionRequest,
 };
 use quic_rpc::transport::quinn::{QuinnConnection, QuinnServerEndpoint};
 use quic_rpc::{RpcClient, ServiceEndpoint};
@@ -123,12 +123,9 @@ enum Commands {
         #[clap(long, default_value_t = ProviderRpcPort::Enabled(DEFAULT_RPC_PORT))]
         rpc_port: ProviderRpcPort,
     },
-    /// List hashes on the running provider.
-    List {
-        /// RPC port of the provider
-        #[clap(long, default_value_t = DEFAULT_RPC_PORT)]
-        rpc_port: u16,
-    },
+    /// List availble content on the provider.
+    #[clap(subcommand)]
+    List(ListCommands),
     /// Validate hashes on the running provider.
     Validate {
         /// RPC port of the provider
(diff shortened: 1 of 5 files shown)


flub (approved):

src/main.rs
```diff
@@ -198,6 +195,22 @@ enum Commands {
...
+        rpc_port: u16,
+    },
+    /// List the available collections on the running provider.
+    Collections {
+        /// RPC port of the provider
+        #[clap(long, default_value_t = DEFAULT_RPC_PORT)]
+        rpc_port: u16,
+    },
+}
+
```
b5, line 213: I'm not a fan of the added depth to the `list` command for two reasons:
1. Don't make the user think about what they want to list. `iroh list` without arguments should list stuff.
2. This breaks the convention established with `iroh get`. Iroh get assumes collections, requires a `--single` flag to fetch a single thing. 

I don't think the `--single` flag is great, but I do think that defaulting to collections is the right choice. a collection with a single item in it is still useful.

**Counter proposal:**
the default for `iroh list` follows the convention of `iroh get`:
* lists collections by default
* a `--blobs` flag lists blobs instead of collections

dignifiedquire: thinking about this some more, I would rather get to a more explicit namespacing, potentially like this

```
iroh collection
  list
  get 
iroh blob
  list 
  get
```

The current `--single` approach is quite odd & confusing imo, and I would really like to avoid extending this further into the api.

rklaehn:
Not sure if the notion of listing collections even makes sense once we have custom collections. If somebody uploads a blob that deserializes as a valid collection, does that count as a collection?

What we really want is to list the GC roots, so the roots that people have added. But I am fine with having this as a temp solution.

b5:
> thinking about this some more, I would rather get to a more explicit namespacing, potentially like this
```
iroh collection
  list
  get 
iroh blob
  list 
  get
```
> The current --single approach is quite odd & confusing imo, and I would really like to avoid extending this further into the api.

This is pushing too much complexity into the user's face. The way to fix the `--single` flag is to remove the need for it entirely: Iroh should just figure out the right thing to do based on the hash I asked for. 

But I see your point @dignifiedquire on not extending the `--single` pattern, but I shouldn't have made such a close association between the two. I totally agree `--single` should be removed, but I think `--blobs` should stick around forever. It's really nice to be able to list only the scalar atoms in a universe, ignoring compound types.

> What we really want is to list the GC roots, **so the roots that people have added**. But I am fine with having this as a temp solution.

Yeah, this crosses into the territory of "my stuff". `iroh list` should list the stuff explicitly added to the node. By the transitive property of fetching a collection I will get some/all of the collection's contents, but I refer to the collection, and want the collection spit out when I list stuff. I can increase the _verbosity_ to say "show me all my stuff in great detail"

As @rklaehn mentions, this gets really funky with custom collections. A problem for another day.

Getting back to the PR, this is a short-term solution, so I'm fine with either taking it as-is and doing a follow-up, or shifting to the `--blobs` flag with no subcommands.

rklaehn:
> As @rklaehn mentions, this gets really funky with custom collections. A problem for another day.

It's not just about custom collections.

E.g. if I create a collection and then add it *as a blob*, is it a collection or a blob?

What if I have a thing that has been added both as a collection and as a blob. The current db will just break on that. Last writer wins, and something breaks...

dignifiedquire:
Going to merge this as is, for the moment it matches what the database does under the hood. When those things get upgraded/change we can update these commands.

PR #1069, opened 2023-06-03
ci: try latest cross

dignifiedquire:


Arqu (approved):
LGTM. I would like to at some point revert to a fixed release not to depend always on main as that could be flakey as stuff moves along.

Think this is the fix that actually brought the improvement https://github.com/cross-rs/cross/pull/1254

PR #1070, opened 2023-06-05
Add back the get --single functionality that has been lost during a refactoring

rklaehn:
Not sure if this should even be the same command. Maybe get blob or get-blob? But we have a bit of a feature explosion.

get (hash / ticket) x (collection / blob)

In any case, the single code path is now separate, which makes things simpler. One question I have is how much, if any, progress we should have when getting single or collection to stdout.

If you just get to stdout without redirecting, the stderr progress and stdout content get mixed and will look very confusing...

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -11,7 +11,7 @@ use indicatif::{
     HumanBytes, HumanDuration, MultiProgress, ProgressBar, ProgressDrawTarget, ProgressState,
     ProgressStyle,
 };
-use iroh::blobs::{Blob, Collection};
+use iroh::blobs::Collection;
 use iroh::config::{Config, CONFIG_FILE_NAME, ENV_PREFIX};
 use iroh::get::get_response_machine::{ConnectedNext, EndBlobNext};
 use iroh::get::{get_data_path, get_missing_range, get_missing_ranges};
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -875,18 +857,129 @@ async fn get_to_dir(get: GetInteractive, out_dir: PathBuf) -> Result<()> {
...
+            .context("unable to create directory {temp_dir}")?;
+        tokio::fs::create_dir_all(&out_dir)
+            .await
+            .context("Unable to create directory {out_dir}")?;
+    }
+    let query = RangeSpecSeq::new([range]);
+    let pb = make_download_pb();
+
+    // collection info, in case we won't get a callback with is_root
+    let collection_info = Some((1, 0));
```
flub, line 900: this is rather cryptic, the comment doesn't really help and the type is so generic I'm not sure what it describes.

src/main.rs
```diff
@@ -875,18 +857,129 @@ async fn get_to_dir(get: GetInteractive, out_dir: PathBuf) -> Result<()> {
...
+        GetInteractive::Ticket {
+            ticket,
+            keylog,
+            derp_map,
+        } => get::run_ticket(&ticket, request, keylog, derp_map).await?,
+        GetInteractive::Hash { opts, .. } => get::run(request, opts).await?,
+    };
+    let connected = response.next().await?;
+    progress!("{} Requesting ...", style("[2/3]").bold().dim());
+    if let Some((count, missing_bytes)) = collection_info {
```
flub, line 913: i don't see how `collection_info` would have been modified by now so isn't this essentially hardcoded?

src/main.rs
```diff
@@ -875,18 +857,129 @@ async fn get_to_dir(get: GetInteractive, out_dir: PathBuf) -> Result<()> {
...
+        init_download_progress(&pb, count, missing_bytes);
+    }
+    let ConnectedNext::StartRoot(curr) = connected.next().await? else {
+        anyhow::bail!("Unexpected StartChild or Closing");
+    };
```
flub, line 918: a let-else in the wild!  nice (/me still learning when they're useful)

src/main.rs
```diff
@@ -1027,41 +1124,31 @@ async fn get_to_dir(get: GetInteractive, out_dir: PathBuf) -> Result<()> {
...
-            .with_key(
-                "eta",
-                |state: &ProgressState, w: &mut dyn std::fmt::Write| {
-                    write!(w, "{:.1}s", state.eta().as_secs_f64()).unwrap()
-                },
-            )
-            .progress_chars("#>-"),
-    );
+/// Get into a file or directory
+async fn get_to_dir(get: GetInteractive, out_dir: PathBuf) -> Result<()> {
```
flub, line 1128: same consistency of doc comment and param name

flub (approved):

flub:
wow, everything is green?!?

src/main.rs
rklaehn: yeah, this code has been split into the single/multi part. I guess there is a bunch of stuff that can go. WIll fix.

src/main.rs
rklaehn: I use them as replacement for the ugly

```rust
let x = match foo {
  Case1(x) => x,
  _ => return
}
```

stuff

PR #1071, opened 2023-06-05
ref(tests): Make all test use test_utils::setup_logging

flub:
A bunch of different logging things were still used and fighting over
the global logging registry.  This migrates them all to
`test_utils::setup_logging` which neatly ties logs to tests by using the
test harness' output capturing.


dignifiedquire (approved):

dignifiedquire:
@flub I believe the failures are due to the way the debugging spans are setup in the test interacting with your log setup. I never got those spans in the quinn_magicsock test to properly work, so likely an easy fix for you.

flub:
Huh, this is much worse than I thought.  I think this is https://github.com/tokio-rs/tracing/issues/1449 which basically says we can't install the tracing subscriber in an async test using set_default (though I got inspired by quinn who does this basically the same way).

It seems you can use the `tracing-futures` crate provides a way of attaching a subscriber that will be installed correctly on polls, but you need to attach it to a runtime or future or stream.  Doing this in tests would essentially mean we have to wrap the test body in an `async {}` block and attach the subscriber to this since I have not found a way to get to the tokio runtime itself.

There are some crates that claim to do this, but they all do this somewhat badly and none really attaches the subscriber correctly but rather do the same global setup hack.

So I guess I could write a proc macro which does the setup correctly and you'd use like this:

```
#[tokio::test]
#[with_subscriber]
async fn test_something() {
     info!("hello");
}
```

which would expand to
```
#[tokio::test]
async fn test_something() {
    use tracing_futures::WithSubscriber;

    let subscriber = setup_subscriber();
    async move {
        info!("hello")
    }.with_subscriber(subscriber).await
}
```

As that would need to be a proc macro it would need to be a separate crate.

flub:
OTOH tracing `set_default` it sets the default subscriber *for the current thread* **using a thread-local**.  Since each test is run in its own thread this should just work.  So maybe all of this is over-engineering and something else is upsetting this.

src/hp/magicsock/conn.rs
```diff
@@ -2850,46 +2850,37 @@ mod tests {
...
-    }
-
     #[tokio::test(flavor = "multi_thread")]
     async fn test_two_devices_roundtrip_quinn_magic() -> Result<()> {
-        setup_logging();
-
-        let devices = Devices {
-            stun_ip: "127.0.0.1".parse()?,
-        };
+        with_logging(async move {
```
flub, line 2855: the entire enclosed code block is just indented, but otherwise unchanged

dignifiedquire:
CI got canceled?

flub:
> CI got canceled?

yeah, stuff hangs again forever in this test.  no idea why.  it worked with the manual async move block, then didn't when wrapping it in a function and now just doesn't want to work anymore.  I'm feeling stupid because I have no clue.

flub:
never mind

Issue #842, opened 2023-03-14
Go back to hex or base32 peerids


rklaehn:
I think the current way of using base64 for peerids is very unergonomic when using the command line. 

1. You can usually not just select the strings with a click, since due to the special chars they are not a single word
2. If you are extra unlucky you will get a peer id that starts with a `-`.

E.g. I got a peer id `-C7yIPRL4Eoy51QS6p6uMyxUsGMkgSWJUmh7gHbqALU` and found it really hard to pass this as an argument to e.g. get.

```
❯ ~/projects_git/iroh/target/release/iroh get --peer -C7yIPRL4Eoy51QS6p6uMyxUsGMkgSWJUmh7gHbqALU --auth-token stigZsDPpuDU-yBld8fAy9EwE0RjfKBy7Z4T3LNxxbM bafkr4iafyu5wgg4em5tz5bxyyxfiqypktgdd4t4ukzwztpbf2c4i5pzove 
error: unexpected argument '-C' found

  note: to pass '-C' as a value, use '-- -C'

Usage: iroh get [OPTIONS] --peer <PEER> --auth-token <AUTH_TOKEN> <HASH>

For more information, try '--help'.
```

🙄 

The concern of these strings being too big for the QR code should be addressed in a different way.

flub:
Yes, this is a real problem and I couldn't figure out how to pass such a PeerId when I had this.  @Arqu also complained about this in the past.  Base32 is a bit smaller than hex, so maybe prefer that?

Arqu:
how about those good ol Qm... or bafy... multihashes :P
Either way, while I've worked around it, it is very annoying to me personally. Just feels *not right* when you encounter those. Don't have a strong preference on what exact encoding to use but anything without special chars would be great.

rklaehn:
I don't see what is wrong with hex. Everybody understands it, and it's not like 5 bits per char vs 4 bits per char is going to make a difference. I am not going to type those strings anyway...

dignifiedquire:
We now use base32 in all CLI output

PR #1072, opened 2023-06-05
Use base32 encoding for both tickets and peer ids

rklaehn:
Not multiformats, just base32 all the time. Use the lower case alphabet and no padding.

Implements https://github.com/n0-computer/iroh/issues/842

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -68,7 +68,10 @@ impl Ord for Hash {
 
 impl Display for Hash {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", hex::encode(self.0.as_bytes()))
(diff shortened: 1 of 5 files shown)


src/util.rs
```diff
@@ -79,7 +68,7 @@ impl Ord for Hash {
 
 impl Display for Hash {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", encode(self.0.as_bytes()))
+        write!(f, "{}", hex::encode(self.0.as_bytes()))
```
dignifiedquire, line 71: why is this hex?

rklaehn: I can make it base32 as well, but since the hashes are never publicly visible I made it hex. Can make it base32lowernopad as well.

dignifiedquire (approved):

PR #1074, opened 2023-06-06
Remove DataSource::File

rklaehn:
Just have 1 case, and fill in the name immediately on construction, just to make things simpler.

Also move `foo/mod.rs` to `foo.rs` to be consistent.


flub (approved):

src/provider.rs
```diff
@@ -1160,49 +1160,37 @@ pub(crate) struct BlobData {
...
+    name: String,
+    /// Path to the file
+    path: PathBuf,
 }
 
 impl DataSource {
     /// Creates a new [`DataSource`] from a [`PathBuf`].
     pub fn new(path: PathBuf) -> Self {
-        DataSource::File(path)
+        let name = path
```
dignifiedquire, line 1173: we should really rename this to `from_path`

PR #1079, opened 2023-06-06
ref(ci): Do not run cargo check separately and test all features

flub:
The cargo check runs do not help with any speedup, they cost >20s, do
not test anything that's not already covered by other tests.  So
remove them.

Updates the cargo test invocations to:

- Use --workspace instead of deprecated --all.  We don't have a
  workspace crate yet though so could just remove it for now, but this
  is a bit more future-proof I guess.

- Use --all-features since we have no features that are incompatible
  with each other.  Let's keep it that way!

- Select the --lib --bins and --tests targets, --all-targets would
  include the benches which we probably don't want to run as part of the
  main test run, but we don't have any yet so that is also just
  future-proofing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,11 +52,8 @@ jobs:
       with:
         toolchain: ${{ matrix.rust }}
       
-    - name: check
-      run: cargo check --all --bins --tests
-      
     - name: tests
-      run: cargo test --all
+      run: cargo test --workspace --all-features --lib --bins --tests
       
   build_and_test_windows:
     name: Build and test (Windows)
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -87,15 +84,8 @@ jobs:
     
     - uses: msys2/setup-msys2@v2
 
-    - name: check
-      run: cargo check --all --bins --tests --target ${{ matrix.target }} 
-      
-    - name: check bench
-      if: matrix.rust == 'nightly'
-      run: cargo check --target ${{ matrix.target }} --benches
```
flub, line 95: We don't have any benches, so i removed this.

flub:
@Arqu any objections?

Arqu (approved):
👍 agree

.github/workflows/ci.yml
```diff
@@ -52,11 +52,8 @@ jobs:
       with:
         toolchain: ${{ matrix.rust }}
       
-    - name: check
-      run: cargo check --all --bins --tests
-      
     - name: tests
-      run: cargo test --all
```
divagant-martian, line 59: I think this is removing the docs tests. There is only one though..

.github/workflows/ci.yml
```diff
@@ -52,11 +52,8 @@ jobs:
...
+      run: cargo test --workspace --all-features --lib --bins --tests
```
divagant-martian, line 56: I think this is removing the docs tests. There is only one though..

PR #1086, opened 2023-06-07
Flub/hp netcheck stun wtf

flub:
what is happening?


.github/workflows/ci.yml
```diff
@@ -18,29 +18,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        name: [ubuntu-latest, macOS-arm-latest]
-        rust: [nightly, stable]
+        name: [ubuntu-latest]
+        rust: [nightly]
```
ramfox, line 22: fwiw, `rust: [stable]` + `name: [ubuntu-latest]` seems to be the matrix that is failing on other branches.

ramfox: ugh, never mind. Ran this with stable & it's the same: no error....

PR #1088, opened 2023-06-08
Feat: Add iroh runtime wrapper and actually use thread per core

rklaehn:
The exact place where we switch on a single threaded io runtime is TBD, but this at least seems to work...

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -80,12 +80,18 @@ mod tests {
     #[tokio::test]
     async fn basics() -> Result<()> {
         setup_logging();
-        transfer_data(vec![("hello_world", "hello world!".as_bytes().to_vec())]).await
+        let rt = test_runtime();
+        transfer_data(
+            vec![("hello_world", "hello world!".as_bytes().to_vec())],
+            rt.handle(),
+        )
+        .await
     }
 
     #[tokio::test]
     async fn multi_file() -> Result<()> {
         setup_logging();
+        let rt = test_runtime();
 
         let file_opts = vec![
             ("1", 10),
@@ -94,12 +100,13 @@ mod tests {
             // overkill, but it works! Just annoying to wait for
             // ("4", 1024 * 1024 * 90),
         ];
-        transfer_random_data(file_opts).await
+        transfer_random_data(file_opts, rt.handle()).await
     }
 
     #[tokio::test]
     async fn many_files() -> Result<()> {
         setup_logging();
+        let rt = test_runtime();
         let num_files = [10, 100, 1000, 10000];
         for num in num_files {
             println!("NUM_FILES: {num}");
@@ -110,7 +117,7 @@ mod tests {
                     (name, 10)
                 })
                 .collect();
-            transfer_random_data(file_opts).await?;
+            transfer_random_data(file_opts, rt.handle()).await?;
         }
         Ok(())
     }
@@ -118,6 +125,7 @@ mod tests {
     #[tokio::test]
     async fn sizes() -> Result<()> {
         setup_logging();
+        let rt = test_runtime();
 
         let sizes = [
             0,
@@ -133,7 +141,7 @@ mod tests {
 
         for size in sizes {
             let now = Instant::now();
-            transfer_random_data(vec![("hello_world", size)]).await?;
+            transfer_random_data(vec![("hello_world", size)], rt.handle()).await?;
             println!("  took {}ms", now.elapsed().as_millis());
         }
 
@@ -142,6 +150,7 @@ mod tests {
 
     #[tokio::test]
     async fn empty_files() -> Result<()> {
+        let rt = test_runtime();
         // try to transfer as many files as possible without hitting a limit
         // booo 400 is too small :(
         let num_files = 400;
@@ -149,7 +158,7 @@ mod tests {
         for i in 0..num_files {
             file_opts.push((i.to_string(), 0));
         }
-        transfer_random_data(file_opts).await
+        transfer_random_data(file_opts, rt.handle()).await
     }
 
     #[tokio::test(flavor = "multi_thread")]
@@ -170,7 +179,7 @@ mod tests {
 
         let rt = test_runtime();
         let provider = provider::Provider::builder(db)
-            .runtime(rt.handle().clone())
+            .runtime(rt.handle())
             .bind_addr(addr)
             .spawn()
             .await?;
@@ -218,7 +227,10 @@ mod tests {
 
     // Run the test creating random data for each blob, using the size specified by the file
     // options
-    async fn transfer_random_data<S>(file_opts: Vec<(S, usize)>) -> Result<()>
+    async fn transfer_random_data<S>(
+        file_opts: Vec<(S, usize)>,
+        rt: &crate::runtime::Handle,
+    ) -> Result<()>
     where
         S: Into<String> + std::fmt::Debug + std::cmp::PartialEq,
     {
@@ -230,11 +242,14 @@ mod tests {
                 (name, content)
             })
             .collect();
-        transfer_data(file_opts).await
+        transfer_data(file_opts, rt).await
     }
 
     // Run the test for a vec of filenames and blob data
-    async fn transfer_data<S>(file_opts: Vec<(S, Vec<u8>)>) -> Result<()>
+    async fn transfer_data<S>(
+        file_opts: Vec<(S, Vec<u8>)>,
+        rt: &crate::runtime::Handle,
+    ) -> Result<()>
     where
         S: Into<String> + std::fmt::Debug + std::cmp::PartialEq,
     {
@@ -267,9 +282,8 @@ mod tests {
         let (db, collection_hash) = provider::create_collection(files).await?;
 
         let addr = "127.0.0.1:0".parse().unwrap();
-        let rt = test_runtime();
         let provider = provider::Provider::builder(db)
-            .runtime(rt.handle().clone())
+            .runtime(rt)
             .bind_addr(addr)
             .spawn()
             .await?;
@@ -377,7 +391,7 @@ mod tests {
         let (db, hash) = create_collection(vec![src.into()]).await.unwrap();
         let mut provider = Provider::builder(db)
             .bind_addr("127.0.0.1:0".parse().unwrap())
-            .runtime(rt.handle().clone())
+            .runtime(rt.handle())
             .spawn()
             .await
             .unwrap();
@@ -449,7 +463,7 @@ mod tests {
         let (db, hash) = create_collection(vec![src0.into(), src1.into()]).await?;
         let provider = Provider::builder(db)
             .bind_addr("127.0.0.1:0".parse().unwrap())
-            .runtime(rt.handle().clone())
+            .runtime(rt.handle())
             .spawn()
             .await?;
         let provider_addr = provider.local_endpoint_addresses().await?;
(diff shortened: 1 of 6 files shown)


src/provider.rs
```diff
@@ -859,14 +862,16 @@ async fn handle_connection<C: CustomGetHandler>(
             events.send(Event::ClientConnected { connection_id }).ok();
             let db = db.clone();
             let custom_get_handler = custom_get_handler.clone();
-            rt.spawn(
+            rt.spawn_tpc(|| {
```
rklaehn, line 865: this is the switch to thread per core for the provider

Arqu:
Tbh this is about 2-3x slower than it used to be.
Also I don't think this works as intended. I might be reading things wrong so I'll dump some thoughts.
- This seems all to be run from a single runtime with 2 worker threads. AFAIK this will limit your whole runtime to two threads and every spawn will just take a share of those two. 
- Spawning per connection might or might not be good overall. I think it's a good strategy to have N threads and then spawn tasks within those but will probably need more finesse than just doing a spawn.
- I don't see that we do any SO_REUSEPORT. Without it (and this goes also for all platforms which don't support load balancing across the reused port) what happens is that the last bind to be executed takes all the traffic (or the first depending on the platform restrictions) but either way, you basically end up with a single threaded runtime, rest is just idling. (this only works on Linux kernels 3.something+ afaik, and otherwise you need to do you own multiplexer on top of the port and load balance internally - not too bad but is ugly code)
- Here https://github.com/n0-computer/beetle/pull/312/files is a very dirty example of some of the tests I've ran where each of these is actually a fully separate run time (which worked better for the tests I've run than just threads in a single runtime)

Arqu:
Another note, once we get to ~same level of perf we had, I think 1 to 1 might still cap out at some point but 1 to N to start improving.

rklaehn:
> Tbh this is about 2-3x slower than it used to be. Also I don't think this works as intended. I might be reading things wrong so I'll dump some thoughts.
> 
> * This seems all to be run from a single runtime with 2 worker threads. AFAIK this will limit your whole runtime to two threads and every spawn will just take a share of those two.
> * Spawning per connection might or might not be good overall. I think it's a good strategy to have N threads and then spawn tasks within those but will probably need more finesse than just doing a spawn.
> * I don't see that we do any `SO_REUSEPORT`. Without it (and this goes also for all platforms which don't support load balancing across the reused port) what happens is that the last bind to be executed takes all the traffic (or the first depending on the platform restrictions) but either way, you basically end up with a single threaded runtime, rest is just idling. (this only works on Linux kernels 3.something+ afaik, and otherwise you need to do you own multiplexer on top of the port and load balance internally - not too bad but is ugly code)
> * Here https://github.com/n0-computer/beetle/pull/312/files is a very dirty example of some of the tests I've ran where each of these is actually a fully separate run time (which worked better for the tests I've run than just threads in a single runtime)

Are you sure? I did some manual checks by running the new `cli_provide_one_file_large` test and looking at the time, and it was identical to x-hp. Or did you mean main?

So, the way this works is that there is a main runtime, which runs accept as well as all the various tasks and timers of the hole punching. Then, once you have a stream pair you move to one of the thread per core runtimes to do the actual IO (reading from the file system and writing to the socket in this case).

Here is where this happens:

```rust
            rt.spawn_tpc(|| {
                async move {
                    let _x = NonSend::new(); // this is just to make the future non send, it does nothing
                    if let Err(err) = handle_stream(db, reader, writer, custom_get_handler).await {
                        warn!("error: {err:#?}",);
                    }
                }
                .instrument(span)
            })
```

Yes, I am aware that this is not fully thread per core since the accept part is still done on a multi threaded tokio runtime. To do real thread per core you would have to basically have an entire quinn endpoint per core. But while that might have worked before we did the hp stuff, now it would be really a can of worms because all the complexity of the magicsocket would have to be duplicated per core.

Arqu:
Right, I was comparing to main!
Yeah tpc is messy if you want to convert and dont start out with that architecture.
Thx for the overview, this might actually be a good approach to get accept/load balancing from a single port easily. 
Think there still is benefit to having dedicated threads as eventually we might want to do 2 threads for hairpinning, one for disk io 5 for network etc. This is mostly dumbspeak without backup at this point, but still think it's a valid observation long term.
I'll try to do some more testing on this over the next few days just to get a better feel on the behavior & perf.

rklaehn:
> Right, I was comparing to main! Yeah tpc is messy if you want to convert and dont start out with that architecture. Thx for the overview, this might actually be a good approach to get accept/load balancing from a single port easily. Think there still is benefit to having dedicated threads as eventually we might want to do 2 threads for hairpinning, one for disk io 5 for network etc. This is mostly dumbspeak without backup at this point, but still think it's a valid observation long term. I'll try to do some more testing on this over the next few days just to get a better feel on the behavior & perf.

Thanks. I would really like to know if this is at least roughly on par with what we had before in `x-hp`. If yes I would like to merge this since it opens up the API design space *a lot* because now we have a way to spawn non send futures.

Whatever happens with the exact architecture later, this is essential since a lot of high perf storage impls just don't do `Send`...

src/lib.rs
```diff
@@ -67,15 +68,30 @@ mod tests {
 
     use super::*;
 
+    /// Pick up the tokio runtime from the thread local and add a
+    /// thread per core runtime.
+    fn test_runtime() -> crate::runtime::Runtime {
+        crate::runtime::Runtime::new(
+            tokio::runtime::Handle::current(),
+            crate::runtime::tpc::Runtime::new("test", 1),
+        )
```
dignifiedquire, line 77: lets make this a little nicer and add a function like this

```rust
impl Runtime {
  pub fn from_currrent(name: &str, size: usize) {
    Self::new(
            tokio::runtime::Handle::current(),
            crate::runtime::tpc::Runtime::new("test", 1),
        )
  }
}
```

src/provider.rs
```diff
@@ -280,7 +280,7 @@ where
     /// get information about it.
     pub async fn spawn(self) -> Result<Provider> {
         trace!("spawning provider");
-        let rt = self.rt.unwrap_or_else(tokio::runtime::Handle::current);
+        let rt = self.rt.unwrap();
```
dignifiedquire, line 283: `expect` with some reason why this shouldn't happen?

dignifiedquire: or ideally actually return with an error?

src/runtime.rs
```diff
@@ -0,0 +1,371 @@
...
+                                .max_blocking_threads(1)
+                                .thread_name(blocking_name)
+                                .build()
+                                .expect("failed to build tokio runtime");
+                            rt.block_on(local.run_until(async move {
+                                while let Ok(Task(f)) = task_receiver.recv_async().await {
+                                    tokio::task::spawn_local(f());
+                                }
+                            }));
+                            tracing::trace!("runtime {} dropped", main_name);
```
dignifiedquire, line 79: this should probably do some `span` tracing magic to make debugging easier, cc @flub

src/runtime.rs
```diff
@@ -0,0 +1,371 @@
...
+            let sender = inner.as_ref().unwrap().clone();
+            sender
+        }
+
+        pub async fn spawn<F, Fut>(&self, f: F)
+        where
+            F: FnOnce() -> Fut + Send + 'static,
+            Fut: Future + 'static,
+        {
+            let f = || f().map(|_| ()).boxed_local();
```
dignifiedquire, line 145: ufff

src/runtime.rs
```diff
@@ -0,0 +1,371 @@
...
+            self.sender()
+                .into_send_async(Task(Box::new(f)))
+                .await
+                .unwrap();
```
dignifiedquire, line 149: please use at least `expect` not just `unwrap` for all these

dignifiedquire: also, realistically all these `send` and `recv` calls can fail, so probably these errors should be propagated to the caller, not just paniced

src/runtime.rs
```diff
@@ -0,0 +1,371 @@
...
+            self.sender().send(Task(Box::new(f))).unwrap();
+        }
+
+        pub async fn run<F, Fut>(&self, f: F) -> Fut::Output
+        where
+            F: FnOnce() -> Fut + Send + 'static,
+            Fut: Future + 'static,
+            Fut::Output: Send + 'static,
+        {
+            let (tx, rx) = futures::channel::oneshot::channel();
```
dignifiedquire, line 168: we already have `flume` as a dependency, so maybe use that for all these, and not mix in channels from futures and std?

src/runtime.rs
```diff
@@ -0,0 +1,371 @@
...
+            tracing_subscriber::registry()
+                .with(tracing_subscriber::fmt::layer().with_writer(std::io::stderr))
+                .with(EnvFilter::from_default_env())
+                .try_init()
+                .ok();
+            let rt = Runtime::new("test", num_cpus::get());
+            let handle = rt.handle().clone();
+            let values = Arc::new(Mutex::new(BTreeSet::new()));
+            let values2 = values.clone();
+            rt.handle().run_sync(|| async move {
```
dignifiedquire, line 227: I believe this is called `_blocking` in other runtimes, so maybe lets use that language instead of `_sync`?

src/runtime.rs
```diff
@@ -0,0 +1,371 @@
...
+            tracing::trace!("shutting down runtime with timeout");
+            // send the shutdown signal by dropping the sender
+            self.handle.sender.lock().unwrap().take();
+            // wait for all threads to finish
+            let start = std::time::Instant::now();
+            while start.elapsed() < duration {
+                if self.handles.iter().all(|h| h.is_finished()) {
+                    break;
+                }
+                std::thread::sleep(std::time::Duration::from_millis(10));
```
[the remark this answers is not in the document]
rklaehn: You could use some more explicit signaling mechanism, but not sure if it is worth the effort. What's the problem with spinning during shutdown?

src/runtime.rs
rklaehn: You don't like the boxing? Or mapping away the result? Not much you can do about the boxing, and mapping away the result should be a noop for things that already return unit and very cheap otherwise.

src/runtime.rs
rklaehn: flume is not an oneshot channel. I assume that a flume channel is more expensive. I can use a tokio::sync::oneshot::channel if you want.

src/runtime.rs
rklaehn: Not sure you want to create a span for the entire duration where the runtime lives. It wouldn't add much info.

I can remove this trace, since I just used it in debugging of tests.

src/runtime.rs
dignifiedquire: it’s just not very pretty, I like simple code..

rklaehn: You can write this in a different way, but I am not sure if it will be simpler. E.g.

```rust
let f = || async { f(); }.boxed_local();
```

src/runtime.rs
dignifiedquire: Sorry, I meant around the stuff in the `spawn_local`, so it is clear which runtime is executing sth

src/runtime.rs
dignifiedquire: I think I like that last option more.

src/runtime.rs
rklaehn: Ah, I see. I can do this. But is it worth it? I would have to clone the name for each task. And what does it gain you? If you want to know on which single threaded runtime you are running you can always output the thread name.

dignifiedquire: true, lets leave it as is for now

PR #1091, opened 2023-06-08
derper: stun improvements

dignifiedquire:
- bind to IPv6 by default, so that dual stack allows receiving requests on both IPv4 and IPv6
- use spawn and spawn_blocking in the stun handler

### DIFF
--- a/src/bin/derper.rs
+++ b/src/bin/derper.rs
@@ -28,7 +28,7 @@ use tokio::{
     task::JoinSet,
 };
 use tokio_rustls_acme::{caches::DirCache, AcmeAcceptor, AcmeConfig};
-use tracing::{debug, debug_span, error, info, warn, Instrument};
+use tracing::{debug, debug_span, error, info, trace, warn, Instrument};
 use tracing_subscriber::{prelude::*, EnvFilter};
 
 type HyperError = Box<dyn std::error::Error + Send + Sync>;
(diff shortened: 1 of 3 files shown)


src/bin/derper.rs
```diff
@@ -692,32 +692,45 @@ async fn serve_stun(host: IpAddr, port: u16) {
...
+                    {
+                        Ok(txid) => {
+                            debug!(%src_addr, %txid, "STUN: received binding request");
+                            let res =
+                                tokio::task::spawn_blocking(move || stun::response(txid, src_addr))
+                                    .await
+                                    .unwrap();
+                            match sock.send_to(&res, src_addr).await {
+                                Ok(len) => {
+                                    debug!(%src_addr, %txid, "STUN: sent {} bytes ({} expected)", len, res.len());
```
flub, line 719: this should be trace level, normally you don't need it because you know a warn would have been shown unless everything succeeded.

src/hp/netcheck.rs
```diff
@@ -1426,7 +1426,7 @@ impl Actor {
 
         let last = self.reports.last.clone();
         let plan = ProbePlan::new(dm, &if_state, last.as_deref());
-
+        dbg!(&plan);
```
flub, line 1429: this isn't test code afaik

src/hp/netcheck.rs
```diff
@@ -1800,65 +1799,74 @@ mod tests {
         Ok(())
     }
 
-    #[tokio::test]
+    #[tokio::test(flavor = "multi_thread")]
```
flub, line 1802: are we sure we want this?

src/bin/derper.rs
dignifiedquire: fixed, it is now checking

flub (approved):

PR #1103, opened 2023-06-13
ref(netcheck): Turn ReportState into more of an actor

### LINKED ISSUE #1012: netcheck: Implement more, test more, monitor more
Netcheck does not have everything implemented yet and is lacking both tests and metrics.  Add all three.

### LINKED ISSUE #1037: netcheck: abort ReportState task when main actor aborts
The main netcheck actor does start a task for ReportState.  However there is no handle to this, it just runs to completion, sending back some messages to the netcheck actor.  It would be nice to also abort this task when the main actor is aborted.  maybe time to introduce a CancellationToken.

Maybe also a good time to make the ReportCheck follow the same actor pattern more for consistency.

### LINKED ISSUE #1100: netcheck: tidy up reportstate by making it actor(s)
It would be easier to follow with less state to be kept around if the reportstate work was split up more.  It would be easier to follow if it followed the same actor pattern as the netcheck actor.

flub:
This moves ReportState into an "Actor" like netcheck is.  It still runs most of the report tasks as futures (haircheck has become an actor, it receives messages).  The main effect of this is that ReportState (now reportgen) does not have any locks around the report anymore.  The control flow does not change significantly (yet) but the amount of state passed around and the amount of local state needing to be stored has reduced significantly.

Closes #1100
Closes #1037 
Closes #1012


iroh-net/src/hp/derp/map.rs
```diff
@@ -57,6 +57,18 @@ impl DerpMap {
 
         dm
     }
+
+    /// Returns the [`DerpNode`] by name.
+    pub fn named_node(&self, node_name: &str) -> Option<&DerpNode> {
```
dignifiedquire, line 60: maybe `find_by_name`?

iroh-net/src/hp/derp/map.rs
```diff
@@ -57,6 +57,18 @@ impl DerpMap {
...
+        for r in self.regions.values() {
+            for n in &r.nodes {
+                if n.name == node_name {
+                    return Some(n);
+                }
+            }
```
dignifiedquire, line 66: could be written as an iterator 

```rust
self.regions
  .values()
  .flat_map(|r| r.nodes.iter())
  .find(|n| n.name == node_name)
```

no strong preference either direction, just wanted to mention it.

iroh-net/src/hp/netcheck.rs
```diff
@@ -111,6 +79,49 @@ impl fmt::Display for Report {
...
+        }
+    }
+
+    /// Returns the maximum latency for all regions.
+    fn max_latency(&self) -> Duration {
+        self.0
+            .values()
+            .max()
+            .copied()
+            .unwrap_or_else(|| Duration::from_millis(100))
```
dignifiedquire, line 105: named constant?

iroh-net/src/hp/netcheck.rs
```diff
@@ -954,38 +618,34 @@ impl Actor {
...
             self.skip_external_network,
-            last.is_some(), // TODO: doesn't need to be passed
             derp_map,
             stun_sock_v4,
             stun_sock_v6,
         );
 
-        self.current_report_run = Some((actor, cancel_token.drop_guard(), response_tx));
+        self.current_report_run = Some(ReportRun {
+            _reportgen: actor,
```
dignifiedquire, line 588: why `_` prefix?

iroh-net/src/hp/netcheck.rs
```diff
@@ -1084,11 +744,8 @@ impl Actor {
                 to_remove.push(*t);
                 continue;
             }
-            for (region_id, d) in &pr.region_latency {
-                let bd = best_recent.entry(*region_id).or_insert(*d);
-                if d < bd {
-                    *bd = *d;
-                }
+            for (region_id, d) in pr.region_latency.iter() {
+                best_recent.update_region(region_id, d);
```
dignifiedquire, line 704: could even make this a dedicated method, `best_recent.merge(&pr.region_latency)`.

iroh-net/src/hp/netcheck/reportgen.rs
```diff
@@ -68,26 +91,24 @@ impl ReportState {
...
             msg_rx,
             netcheck: netcheck.clone(),
-            plan,
-            last: last_report,
+            last_report,
             port_mapper,
             skip_external_network,
             incremental,
-            derpmap,
+            derp_map: derpmap,
```
dignifiedquire, line 105: lets use consistent spelling please

iroh-net/src/hp/netcheck/reportgen.rs
```diff
@@ -310,16 +318,28 @@ impl Actor {
 
         // A collection of futures running probe sets.
         let mut probes = FuturesUnordered::default();
+        let mut derp_nodes_cache: BTreeMap<String, Box<DerpNode>> = BTreeMap::new();
```
dignifiedquire: why are these nodes boxed?

iroh-net/src/hp/netcheck/reportgen.rs
```diff
@@ -310,16 +318,28 @@ impl Actor {
...
+                let derp_node = match derp_nodes_cache.get(probe.node()) {
+                    Some(node) => node.clone(),
+                    None => {
+                        let name = probe.node().to_string();
+                        let node = self
+                            .derp_map
+                            .named_node(&name)
+                            .with_context(|| format!("missing named derp node {}", probe.node()))?;
+                        let node = Box::new(node.clone());
+                        derp_nodes_cache.insert(name, node.clone());
```
dignifiedquire: I am pretty sure that this will clone the full `DerpNode`, If you want ref counting you need to use `Rc` or `Arc`.

iroh-net/src/hp/netcheck/reportgen.rs
```diff
@@ -0,0 +1,1006 @@
...
+        Self {
+            _drop_guard: Arc::new(DropGuard {
+                handle: task.abort_handle(),
+            }),
+        }
+    }
+}
+
+#[derive(Debug)]
+struct DropGuard {
```
dignifiedquire, line 122: don't we have sth reusable for this somewhere now?

iroh-net/src/hp/netcheck/reportgen.rs
```diff
@@ -0,0 +1,1006 @@
...
+    }
+
+    /// Runs the main reportgen actor logic.
+    ///
+    /// This actor runs by:
+    ///
+    /// - Creates a hairpin actor.
+    /// - Creates a captive portal future.
+    /// - Creates ProbeSet futures in a group of futures.
+    /// TODO: pinger
```
dignifiedquire, line 232: what's the plan here?

iroh-net/src/hp/netcheck/reportgen.rs
```diff
@@ -0,0 +1,1006 @@
...
+            trace!(awaiting = ?self.outstanding_tasks, "tick; awaiting tasks");
+            if self.outstanding_tasks.all_done() {
+                debug!("all tasks done");
+                break;
+            }
+            let remaining_time = OVERALL_PROBE_TIMEOUT.saturating_sub(start_time.elapsed());
+            let remaining_probe_time = STUN_PROBE_TIMEOUT.saturating_sub(start_time.elapsed());
+
+            tokio::select! {
+                _ = tokio::time::sleep(remaining_time) => {
```
dignifiedquire, line 262: This seems inefficient (I might be missing sth), you create a new `Sleep` on every loop, but you could just do this

```rust
let total_timer = tokio::time::sleep(OVERALL_PROBE_TIMEOUT);
tokio::pin!(total_timer);

let stun_timer =  tokio::time::sleep(STUN_PROBE_TIMEOUT);
tokio::pin!(total_timer);

loop {
  tokio::select! {
    _ = &mut total_timer => {
    }
    _ = &mut stun_timer => {
    }
  }
}
```

iroh-net/src/hp/netcheck/reportgen.rs
```diff
@@ -0,0 +1,1006 @@
...
+            }
+            let reportcheck = self.addr();
+            tokio::spawn(async move {
+                time::sleep(timeout).await;
+                reportcheck.send(Message::AbortProbes).await.ok();
+            });
+        }
+
+        if let Some(ipp) = ipp {
+            if ipp.is_ipv6() {
```
dignifiedquire, line 445: would be nicer to use a match statement instead of the `if else`

iroh-net/src/hp/netcheck/reportgen.rs
```diff
@@ -0,0 +1,1006 @@
...
+            }
+            match res.1 {
+                Ok((d, ip)) => {
+                    result.delay = Some(d);
+                    // We set these IPv4 and IPv6 but they're not really used
+                    // and we don't necessarily set them both. If UDP is blocked
+                    // and both IPv4 and IPv6 are available over TCP, it's basically
+                    // random which fields end up getting set here.
+                    // Since they're not needed, that's fine for now.
+                    if ip.is_ipv4() {
```
dignifiedquire, line 836: could be a `match` statement

iroh-net/src/hp/netcheck/reportgen.rs
dignifiedquire: ignore, this was on an old commit

iroh-net/src/hp/netcheck/reportgen.rs
dignifiedquire: ignore, this was on an old commit

iroh-net/src/hp/netcheck.rs
flub: It's described in the doc comment of `ReportRun`.  We're only keeping this as a drop_guard: if dropped the reportgen actor will stop.  So we never access this field.

iroh-net/src/hp/netcheck/reportgen.rs
flub: no, it was me who was missing something :)

iroh-net/src/hp/netcheck/reportgen.rs
flub: I just forgot to remove this...

flub:
PTAL

dignifiedquire:
The netsim failure looks real, or this a known issue @Arqu?

divagant-martian (commented):
nowhere near finished but progressing

iroh-net/src/hp/netcheck/reportgen.rs
```diff
@@ -0,0 +1,993 @@
...
+/// Messages to send to the reportstate [`Actor`].
+#[derive(Debug)]
+enum Message {
+    /// Set the hairpinning availability in the report.
+    HairpinResult(bool),
+    /// Check whether executing a probe would still help.
+    // TODO: Ideally we remove the need for this message and the logic is inverted: once we
+    // get a probe result we cancel all probes that are no longer needed.  But for now it's
+    // this way around to ease conversion.
+    ProbeWouldHelp(Probe, Arc<DerpNode>, oneshot::Sender<bool>),
```
divagant-martian, line 147: started working on that here https://github.com/n0-computer/iroh/pull/1180

iroh-net/src/hp/netcheck.rs
```diff
@@ -116,6 +79,49 @@ impl fmt::Display for Report {
...
+        }
+    }
+
+    /// Returns the maximum latency for all regions.
+    fn max_latency(&self) -> Duration {
+        self.0
+            .values()
+            .max()
+            .copied()
+            .unwrap_or_else(|| Duration::from_millis(100))
```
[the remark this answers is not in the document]
flub: now explained in excruciating detail :)

iroh-net/src/hp/netcheck.rs
```diff
@@ -1,79 +1,51 @@
...
+/// latencies we return this as default.  This is the value of the initial STUN probe
+/// delays.  It is only used as time to wait for further latencies to arrive, which *should*
+/// never happen unless there already is at least one latency.  Yet here we are, defining a
+/// default which will never be used.
+const DEFAULT_MAX_LATENCY: Duration = Duration::from_millis(100);
 
 /// A netcheck report.
 ///
-/// Can be obtained by calling [Client::get_report].
+/// Can be obtained by calling [`Client::get_report`].
```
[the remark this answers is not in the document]
flub: It does, but then it does not treat it as a code block.  I once had an in-depth bikeshedding session with someone about whether you should type these backticks or not.  We eventually settled on "do what Rust std does" as the best tie-breaker, and they put code references in backticks.  So since then it's become a habit to me to include the backticks.

:shrug:

iroh-net/src/hp/netcheck.rs
```diff
@@ -1085,12 +366,12 @@ pub(crate) enum ActorMessage {
...
         response_tx: oneshot::Sender<Result<Arc<Report>>>,
     },
-    /// A report produced by [`ReportState`].
+    /// A report produced by the reportgen actor.
     ReportReady {
         report: Box<Report>,
         derp_map: DerpMap,
     },
-    /// [`ReportState`] failed to produce a report.
+    /// The reportgen actor failed to produce a report.
```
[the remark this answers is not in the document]
flub: Very valid point.  I couldn't decide to link to the module, `Actor` or `Client` and ended up doing none which is clearly rubbish.  Linking to the `Actor` now.

iroh-net/src/hp/netcheck.rs
```diff
@@ -1170,11 +454,11 @@ struct Actor {
...
     /// This is used to complete the STUN probe when receiving STUN packets.
     in_flight_stun_requests: HashMap<stun::TransactionId, Inflight>,
-    /// The response channel if there is a check running.
+    /// The reportgen actor currently generating a report.
     ///
-    /// There can only ever be one check running at a time.  If it is running the response
-    /// channel is stored here.
-    current_check_run: Option<oneshot::Sender<Result<Arc<Report>>>>,
+    /// The [`tokio_util::sync::DropGuard`] is to ensure the local STUN listener is shut
+    /// down if it was started.  The finished report is sent on the channel.
```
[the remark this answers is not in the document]
flub: these docs have moved to the `ReportRun` struct.  fixed

iroh-net/src/hp/netcheck.rs
```diff
@@ -1221,64 +506,41 @@ impl Actor {
...
         stun_sock_v4: Option<Arc<UdpSocket>>,
         stun_sock_v6: Option<Arc<UdpSocket>>,
         response_tx: oneshot::Sender<Result<Arc<Report>>>,
     ) {
-        if self.current_check_run.is_some() {
-            response_tx
-                .send(Err(anyhow!("A check is already running")))
-                .ok();
+        if self.current_report_run.is_some() {
+            warn!("ignoring RunCheck request; reportgen actor already running");
```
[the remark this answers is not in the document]
flub: oh, it escaped me that the sender would notice as the sender would be dropped.  Good catch, especially since it can already return errors.

iroh-net/src/hp/netcheck/reportgen.rs
```diff
@@ -0,0 +1,993 @@
...
+const ENOUGH_REGIONS: usize = 3;
+
+/// Holds the state for a single invocation of [`netcheck::Client::get_report`].
+///
+/// Dropping this will cancel the actor and stop the report generation.
+#[derive(Debug, Clone)]
+pub(super) struct Client {
+    // Addr is currently only used by child actors, so not yet exposed here.
+    _drop_guard: Arc<CancelOnDrop>,
+}
```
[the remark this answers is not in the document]
flub: I'm a bit conflicted, it mirrors the behaviour of `netcheck::Client` and I was going for consistency.  I think I'm still mildly preferring this implementation for that reason.

iroh-net/src/hp/netcheck/reportgen.rs
```diff
@@ -0,0 +1,993 @@
...
+/// reply before switching to HTTP probing, on the assumption that outbound UDP is blocked.
+const STUN_PROBE_TIMEOUT: Duration = Duration::from_secs(3);
+
+/// The maximum amount of time netcheck will spend probing with ICMP packets.
+const ICMP_PROBE_TIMEOUT: Duration = Duration::from_secs(1);
+
+/// How long to await for a captive-portal result, chosen semi-arbitrarily.
+const CAPTIVE_PORTAL_DELAY: Duration = Duration::from_millis(200);
+
+/// Timeout for captive portal checks, must be lower than OVERALL_PROBE_TIMEOUT
```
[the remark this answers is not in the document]
flub: Static assertions would be nice I guess, though `PartialOrd` for `Duration` is not `const`, so I don't think we can do this.

iroh-net/src/hp/netcheck/reportgen.rs
```diff
@@ -0,0 +1,993 @@
...
+pub(super) struct Client {
+    // Addr is currently only used by child actors, so not yet exposed here.
+    _drop_guard: Arc<CancelOnDrop>,
+}
+
+impl Client {
+    /// Creates a new actor generating a single report.
+    ///
+    /// The actor starts running immediately and only generates a single report, after which
+    /// it shuts down.  Dropping this handle will abort the actor.
```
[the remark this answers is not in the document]
flub: Ah yes, the witespace debate :)  My fingers type two spaces, they'll probably never change.  It's fine.

(And if it's not fine I'd argue rustfmt needs to be thought how to fix that up)

divagant-martian (commented):
this looks like an overall improvement :tada:

iroh-net/src/hp/netcheck.rs
```diff
@@ -116,6 +79,49 @@ impl fmt::Display for Report {
...
 }
 
+/// Latencies per DERP Region.
+#[derive(Debug, Default, PartialEq, Eq, Clone)]
+pub struct RegionLatencies(HashMap<u16, Duration>);
+
+impl RegionLatencies {
+    fn new() -> Self {
+        Default::default()
+    }
```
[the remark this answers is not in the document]
flub: I'm never really sure, but I usually prefer having it.  Logically creating a new struct is something different from just wanting a struct's default, at least to me, and it's nice to be able to do both.  The stdlib does this in lots of places as well, e.g. `String`.

Also clippy has https://rust-lang.github.io/rust-clippy/master/index.html#/new_without_default in which they suggest to have both rather than just switching entirely to default.

So my conclusion is generally that a new method without parameters is nice and certainly not un-idiomatic in Rust.

iroh-net/src/hp/netcheck/reportgen.rs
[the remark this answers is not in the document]
dignifiedquire: If it is not used I would like to avoid implementing `Clone` for all more complex types. It makes anyone wanting to clone it, think twice :)

iroh-net/src/hp/netcheck.rs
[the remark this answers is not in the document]
dignifiedquire: having both seems fine to me

dignifiedquire (approved):

iroh-net/src/hp/netcheck/reportgen.rs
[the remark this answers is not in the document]
flub: Oh yes, true.  Good idea.  I now I need to worry about overflow semantics though :)

I'd rather leave this for now.  There's a whole bunch of invariants built into lots of parts of netcheck for now that could be handled better by the typesystem.  Even just these constants there are loads more, they just didn't have the comment (which I was just copying around).  They will improve as we refactor and fix things more.

iroh-net/src/hp/netcheck/reportgen.rs
```diff
@@ -0,0 +1,993 @@
...
+
+    /// Creates the future which will perform the portmapper task.
+    ///
+    /// The returned future will run the portmapper, if enabled, resolving to it's result.
+    fn prepare_portmapper_task(
+        &mut self,
+    ) -> MaybeFuture<Pin<Box<impl Future<Output = Option<portmapper::ProbeOutput>>>>> {
+        let mut port_mapping = MaybeFuture::default();
+        if !self.skip_external_network {
+            if let Some(port_mapper) = self.port_mapper.clone() {
```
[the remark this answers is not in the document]
flub: I once more find myself rather conflicted.  Part of me agrees that it would be nice to avoid the clone.  Part of me also worries this makes the code harder to understand and easier to get wrong with later refactors:

The `Option` part of `Option<PortMapper` right now only says "you have a portmapper".  Taking it out changes this to "this bit is mutable state that changes depending on where you are in the code.  I know, the methods already has `&mut self` so what am I complaining about.  Still it makes it easier to get right.

If anything I'd rather remove the `Option` altogether: We have no code path which does not pass in the PortMapper.  So why make this optional at all?  Cloning the thing is pretty cheap I think anyway.

iroh-net/src/hp/netcheck/probe.rs
```diff
@@ -94,9 +109,39 @@ impl Probe {
 /// A set of probes is done once either one of the probes completes, or
 /// the next probe to run wouldn't yield any new information not
 /// already discovered by any previous probe in any set.
+///
+/// Generated probe plans have the following shape:
+///
+/// - Each probe **set** is single identical probe, but multiple tries with delayed starts.
```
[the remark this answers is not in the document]
ramfox: hi hi! In a few places, this included, the conversation is marked as resolved, but the suggestion wasn't taken and there isn't a response as to why. Were these skipped over?

Confused specifically in this case, because there is a grammar error here that should be fixed regardless of phrasing.

iroh-net/src/hp/netcheck/reportgen.rs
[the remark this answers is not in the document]
flub: So I had a go at removing the option, but I don't want to do this for now.  It's a lot of churn, better to be done another time.

If you really feel strongly I can change it into the take, but I'd rather leave it I think until it gets cleaned up a bit better.

iroh-net/src/hp/netcheck/reportgen/hairpin.rs
```diff
@@ -0,0 +1,318 @@
...
+    sender: mpsc::Sender<Message>,
+}
+
+impl Deref for Addr {
+    type Target = mpsc::Sender<Message>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.sender
+    }
+}
```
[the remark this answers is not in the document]
flub: True, though it only removes some lines of code in return for having to look up a bit more how macros work.  I like derive_more's Debug macro since it solves a problem: new fields on a struct don't get forgotten to be added to the manual Debug impl.  But this is just another way or writing things so I'm rather non-plussed about it (this ties in with my general dislike for macros :smile:, it's another language you have to learn)

flub:
@divagant-martian let me know if there's anything remaining you feel so strongly about that you wouldn't want to see it merged without.  Otherwise I'd like to leave it as it is for now, even though there are still improvements to be made.

divagant-martian:
I understand you've been working on this for a long time and would like to merge asap so let's go ahead. We can improve things later  🎉

iroh-net/src/hp/netcheck/probe.rs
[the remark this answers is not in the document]
flub: oh yeah, that's weird now you point it out.  I guess it should show it as outdated.  I promise I don't (intentionally) dismiss reviews without acting on them!  If I click "resolved" I mean I did do the obvious requested fix or something very close to it that I expect to be unobjectionable (perhaps not 100% literally like in this case my editor re-flowed the paragraph).

PR #1106, opened 2023-06-13
Partial thread per core, again

rklaehn:


rklaehn:
https://github.com/n0-computer/iroh/pull/1088 was not against main, but against x-hp-tpc-2 due to the whole x-hp confusion last week. This is identical.

dignifiedquire (approved):

src/runtime.rs
```diff
@@ -0,0 +1,420 @@
+//! The runtime module provides the iroh runtime, consisting of a general purpose
+//! tokio runtime and a set of single threaded runtimes.
+//!
+//! Functions spawned on single threaded runtimes should not panic,
+//! because panics will shut down the worker thread running the function, and
+//! the runtime will not spawn new threads.
+//!
+//! It is best to run the entire program with panic = 'abort'.
```
flub, line 8: should we put this in Cargo.toml?

Arqu: The release builds already do that

flub: great, thanks!

Issue #987, opened 2023-05-09
Request Level Authentication


dignifiedquire:
We need to allow users to checkout authentication on a per request level. 

- Should be entirely optional
- Ideally this has some very simple default implementation like the old ticketing system 
- Needs a hook to allow users to provide fully custom ops, like UCANs or other systems

PR #1109, opened 2023-06-14
Feat: Request Tokens (formerly Opaque Auth)

b5:
Alright, I’m taking a stab at #987.

This PR plumbs opaque request authorization tokens into the protocol
* ~~Increments the protocol version number 2 -> 3~~, keeping version 2
* ~~Adds a reserved for future use "connectionToken", to distinguish connection authorization from request authorization~~
* sends tokens as part of the `get` request payload
* Establishes the requirement that the CLI accept  tokens as base32-encoded binary data
* defines an `AuthorizationHandler` generic on the provider for actually doing authorization
* plumbs Request tokens up into provider events

### Necessary Follow ups
* [ ] restore "ticket tokens": simple auth tokens that `iroh get-ticket` used to consume
* [x] ~~test for declining authorization tokens using a custom `AuthorizationHandler` on the provider~~ default token handler should decline tokens of any kind

This PR is generally a mess with way too much cloning & copying. Looking for some lessons in writing proper rust.


src/protocol.rs
```diff
@@ -18,16 +20,63 @@ use crate::util::Hash;
...
+/// Connection tokens are 32 byte values used to authorize connections.
+//
+/// These are currently unused, but exist to distinguish connection authorization
+/// from request authorization.
+pub type ConnectionToken = [u8; 32];
+
+#[derive(Deserialize, Serialize, Debug, PartialEq, Eq, MaxSize, Clone)]
 pub(crate) struct Handshake {
     pub version: u64,
+    pub conn_token: Option<ConnectionToken>,
```
dignifiedquire, line 34: Please leave this out for now. This will very likely be stored and validated at a different point.

src/protocol.rs
```diff
@@ -18,16 +20,63 @@ use crate::util::Hash;
...
+        Self {
+            version: VERSION,
+            conn_token: None,
+        }
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, From)]
+/// Authorization tokens are opaque byte sequences used to authorize requests.
+pub struct AuthToken {
```
dignifiedquire, line 48: should probably be called something like `RequestAuthToken`

src/protocol.rs
```diff
@@ -18,16 +20,63 @@ use crate::util::Hash;
...
+    bytes: Vec<u8>,
+}
+
+impl AuthToken {
+    pub fn new(bytes: Vec<u8>) -> Result<Self> {
+        ensure!(bytes.len() < MAX_MESSAGE_SIZE, "auth token is too large");
```
dignifiedquire, line 54: This would definitely be too large, lets start with restricting a much lower number, maybe `256` bytes and see how that goes

src/provider.rs
```diff
@@ -117,6 +130,28 @@ impl CustomGetHandler for () {
...
+
+/// Define AuthorizationHandler for () so we can use it as a no-op default.
+impl AuthorizationHandler for () {
+    fn authorize(
+        &self,
+        _db: Database,
+        _token: Option<AuthToken>,
+        _request: Request,
+    ) -> BoxFuture<'static, anyhow::Result<()>> {
+        async move { Ok(()) }.boxed()
```
dignifiedquire, line 151: I think the default handler should check that the auth token is actually `None`.

b5:
I'm going to wait for #1126, then will rebase & add back tests

<details>
<pre>
#[tokio::test]
    async fn test_token_passthrough() {
        let rt = test_runtime();
        let readme = readme_path();
        let (db, hash) = create_collection(vec![readme.into()]).await.unwrap();
        let provider = match Provider::builder(db)
            .bind_addr("[::1]:0".parse().unwrap())
            .runtime(rt.handle())
            .spawn()
            .await
        {
            Ok(provider) => provider,
            Err(_) => {
                // We assume the problem here is IPv6 on this host.  If the problem is
                // not IPv6 then other tests will also fail.
                return;
            }
        };

        let token = Some(AuthToken::from(vec![1, 2, 3, 4, 5, 6]));
        let mut events = provider.subscribe();
        let (tx, rx) = tokio::sync::oneshot::channel();
        tokio::spawn(async move {
            let mut token = None;
            while let Ok(msg) = events.recv().await {
                if let Event::GetRequestReceived { token, .. } = msg {
                    // println!("token: {:?}", token);
                    token = token;
                    break;
                }
            }
            tx.send(token).unwrap();
        });

        let addrs = provider.local_endpoint_addresses().await.unwrap();
        let peer_id = provider.peer_id();
        tokio::time::timeout(Duration::from_secs(10), async move {
            let connection = dial_peer(get::Options {
                addrs,
                peer_id,
                keylog: true,
                derp_map: None,
            })
            .await?;
            let request = GetRequest::all(hash).with_token(token).into();
            let stream = get::run_connection(connection, request);
            aggregate_get_response(stream).await?;
            anyhow::Ok(())
        })
        .await
        .expect("timeout")
        .expect("get failed");

        match rx.await {
            Ok(token) => {
                assert!(Some(AuthToken::from(vec![1, 2, 3, 4, 5, 6])) == token);
            }
            Err(e) => {
                panic!("error receiving token: {:?}", e);
            }
        }
    }
</pre>
</details>

iroh-bytes/src/protocol.rs
```diff
@@ -31,13 +33,77 @@ impl Handshake {
...
+/// user association, etc.
+pub struct RequestToken {
+    bytes: Vec<u8>,
+}
+
+impl RequestToken {
+    pub fn new(bytes: Vec<u8>) -> Result<Self> {
+        ensure!(
+            bytes.len() < MAX_REQUEST_TOKEN_SIZE,
+            "auth token is too large"
```
dignifiedquire, line 52: should be `request token`

iroh-bytes/src/protocol.rs
```diff
@@ -31,13 +33,77 @@ impl Handshake {
     }
```
b5, line 33: This PR originally bumped the protocol version from 2 -> 3. I've reverted the change, keeping the version at 2, because we haven't cut a release with version 2 as the identifier. Others feel free to disagree

dignifiedquire:
> * restore "ticket tokens": simple auth tokens that `iroh get-ticket` used to consume

@b5 what's the plan here?

b5:
> @b5 what's the plan here?

I'm planning a follow-up that re-adds tokens to tickets, and does better error handling around poorly formatted CLI tokens (the base32 requirement is a potential footgun) will leave #987 open until that PR is merged. Sound good?

.github/workflows/ci.yml
```diff
@@ -228,7 +228,7 @@ jobs:
       run: cross check --all --target ${{ matrix.target }}
 
     - name: test
-      run: cross test --all --target ${{ matrix.target }} -- --test-threads=8
+      run: cross test --all --target ${{ matrix.target }} -- --test-threads=4
```
dignifiedquire, line 231: why do we have to keep this change Oo?

iroh-bytes/src/provider.rs
```diff
@@ -466,6 +524,10 @@ async fn handle_stream<E: EventSender>(
         }
     };
 
+    authorization_handler
+        .authorize(db.clone(), request.token().cloned(), &request)
```
dignifiedquire, line 528: Sorry I missed this until now, given that we have to clone this here, the token should not be `Vec<u8>` but rather `Bytes`, so that clones are cheap reference counts.

iroh/src/commands.rs
```diff
@@ -188,16 +190,19 @@ pub enum Commands {
         #[clap(long, default_value_t = DEFAULT_RPC_PORT)]
         rpc_port: u16,
     },
-    /// Fetch the data identified by HASH from a provider.
+    /// Fetch the data identified by HASH from a provider
```
dignifiedquire, line 193: Why do you delete `.` Oo

iroh/src/commands.rs
```diff
@@ -188,16 +190,19 @@ pub enum Commands {
...
         /// The hash to retrieve, as a Blake3 CID
         hash: Blake3Cid,
         /// PeerId of the provider
         #[clap(long, short)]
         peer: PeerId,
-        /// Addresses of the provider.
+        /// Addresses of the provider
         #[clap(long, short)]
         addrs: Vec<SocketAddr>,
+        /// base32-encoded Request token to use for authentication, if any
```
dignifiedquire, line 203: nit: general note, doc comments should start with a capital letter

iroh/tests/provide.rs
```diff
@@ -716,3 +727,106 @@ async fn test_custom_request_collection() {
...
+    let provider = match Node::builder(db)
+        .bind_addr("[::1]:0".parse().unwrap())
+        .custom_auth_handler(CustomAuthHandler)
+        .runtime(&rt)
+        .spawn()
+        .await
+    {
+        Ok(provider) => provider,
+        Err(_) => {
+            // We assume the problem here is IPv6 on this host.  If the problem is
```
dignifiedquire, line 772: Just bind to ipv4 please, and make this not just silently fail. This should always work, if not, this is a bad sign

iroh/tests/provide.rs
```diff
@@ -716,3 +727,106 @@ async fn test_custom_request_collection() {
...
+    .await
+    .expect("timeout")
+    .expect("get failed");
+
+    match rx.await {
+        Ok(token) => {
+            assert!(Some(RequestToken::from(vec![1, 2, 3, 4, 5, 6])) == token);
+        }
+        Err(e) => {
+            panic!("error receiving token: {:?}", e);
```
dignifiedquire, line 829: instead of this, you would write

```rust
let token = rx.await.expect("error receiving token");
```

iroh/tests/provide.rs
```diff
@@ -716,3 +727,106 @@ async fn test_custom_request_collection() {
...
+                        bp_msg
+                    {
+                        // println!("token: {:?}", token);
+                        token = tok;
+                        break;
+                    }
+                }
+            }
+        }
+        tx.send(token).unwrap();
```
dignifiedquire, line 796: Instead of the oneshot, just return the value at the end, and await the handle returned from `tokio::spawn`.

dignifiedquire (approved):
lgtm

.github/workflows/ci.yml
Arqu: TLDR; cross + android makes memory usage grow like crazy. We're already seeing this on other branches too. We're right at the watermark here. Reducing parallelism on these gives more room. 

This is not the way to go about it as eventually we will hit `test-threads=1` and this grows runtime considerably which further clogs CI. I'll tackle this sooner rather than later.

dignifiedquire: 👍

PR #1110, opened 2023-06-14
fix: don't crash the derper

dignifiedquire:


ramfox:
Ensures flushing when `ClientConn` closes. Also, ensures the `ClientConn` closes when it gets an EOF after reading off the connection.

ramfox (approved):

PR #1111, opened 2023-06-15
Crates refactor

dignifiedquire:
Refactors iroh into three crates

- `iroh`
  - main binary
  - provide loop, now called `Node` to prepare for being more generic of different protocols
  - rpc logic
- `iroh-bytes`
  - The current `iroh` protocol, collections, database, provide & get
- `iroh-net`
  - derper binary
  - Holepunching & connectiong mgmt
  - tls

## Things needed for merge

- [ ] update crate names
- [ ] update READMEs
- [x] update `Cargo.toml`s
- [x] update CI to include all feature & create combinations
- [x] green CI
- [x] make metrics modula

## Misc

- Does not use `workspace.dependencies` on purpose. I am trying to limit shared dependencies, and it is much easier to actually read/consume when looking at individual crates, at the expense of updates. 
- There are some things like the runtime and various utilities that could be their own crates. For now I would like to keep them like this to not create too many crates, and only split them out if there is an actual need.


dignifiedquire:
> If we do this refactoring it would make sense to have iroh-io - that I am currently working on - live in this workspace as well.

sounds good 👍

rklaehn:
Just looked at the docs - highly recommended to get an overview. For the iroh-net crate I think it would be useful to move a few more things to top level. Currently it seems more like "a bag of stuff" without a clear purpose and entry point.

image

Iroh-bytes is a bit too big, it is almost "basically iroh without the cli". Also it contains stuff that I would have expected to be relevant also for other iroh subsystems, such as the `PeerId`.

image

PR #1114, opened 2023-06-20
refactor: strip down the iroh runtime

rklaehn:
and use tokio_util::task::LocalPoolHandle instead

https://docs.rs/tokio-util/latest/tokio_util/task/struct.LocalPoolHandle.html

it is basically the same thing, but we can split up iroh more easily without having to have an iroh-runtime crate.


dignifiedquire (approved):
wow, that makes things simpler 😅

rklaehn:
> wow, that makes things simpler 😅

I am totally not emotionally attached to that code... Less code is more good!

PR #1115, opened 2023-06-20
async local io

rklaehn:
Async local io

this does not really make too much sense right now, but will become part of the workspace once the iroh crate split is done.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -79,15 +79,15 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
-version = "1.0.0"
+version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "41ed9a86bf92ae6580e0a31281f65a1b1d867c0cc68d5346e2ae128dddfa6a7d"
+checksum = "3a30da5c5f2d5e72842e00bcb57657162cdabef0931f40e2deb9b4140440cecd"
 
 [[package]]
 name = "anstyle-parse"
-version = "0.2.0"
+version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e765fd216e48e067936442276d1d57399e37bce53c264d6fefbe298080cb57ee"
+checksum = "938874ff5980b03a87c5524b3ae5b59cf99b1d6bc836848df7bc5ada9643c333"
 dependencies = [
  "utf8parse",
 ]
@@ -128,9 +128,9 @@ checksum = "6b4930d2cb77ce62f89ee5d5289b4ac049559b1c45539271f5ed4fdc7db34545"
 
 [[package]]
 name = "arrayvec"
-version = "0.7.3"
+version = "0.7.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8868f09ff8cea88b079da74ae569d9b8c62a23c68c746240b704ee6f7525c89c"
+checksum = "96d30a06541fbafbc7f82ed10c06164cfbd2c401138f6addd8404629c4b16711"
 
 [[package]]
 name = "asn1-rs"
@@ -326,9 +326,9 @@ dependencies = [
 
 [[package]]
 name = "bounded-integer"
-version = "0.5.6"
+version = "0.5.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fc54ef154ce8654be9ee10ed33106bdfd5bff6ee7735c25a160af2290ea938ab"
+checksum = "78a6932c88f1d2c29533a3b8a5f5a2f84cc19c3339b431677c3160c5c2e6ca85"
 
 [[package]]
 name = "build-data"
@@ -445,9 +445,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.3.3"
+version = "4.3.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ca8f255e4b8027970e78db75e78831229c9815fdbfa67eb1a1b777a62e24b4a0"
+checksum = "2686c4115cb0810d9a984776e197823d08ec94f176549a89a9efded477c456dc"
 dependencies = [
  "clap_builder",
  "clap_derive",
@@ -456,9 +456,9 @@ dependencies = [
 
 [[package]]
 name = "clap_builder"
-version = "4.3.3"
+version = "4.3.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "acd4f3c17c83b0ba34ffbc4f8bbd74f079413f747f84a6f89292f138057e36ab"
+checksum = "2e53afce1efce6ed1f633cf0e57612fe51db54a1ee4fd8f8503d078fe02d69ae"
 dependencies = [
  "anstream",
  "anstyle",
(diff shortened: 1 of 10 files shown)


Cargo.lock
```diff
@@ -1844,6 +1939,24 @@ dependencies = [
...
+[[package]]
+name = "native-tls"
+version = "0.2.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "07226173c32f2926027b63cce4bcd8076c3552846cbe7925f3aaffeac0a3b92e"
+dependencies = [
+ "lazy_static",
+ "libc",
+ "log",
+ "openssl",
```
Arqu, line 1951: This is pulling it in

rklaehn:
This is already published. Here are the docs: https://docs.rs/iroh-io/0.1.0/iroh_io/

iroh-io/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+[features]
+http = ["reqwest"]
+tokio-io = ["tokio"]
+default = ["tokio-io"]
+
+[dev-dependencies]
+proptest = "1"
+tempfile = "3"
+tokio = { version = "1", features = ["rt", "test-util", "macros"] }
+warp = { version = "0.3.5", default-features = false }
```
dignifiedquire, line 28: Would be great to switch this to `axum`, given that is already in our dependency tree for the `derper`. (as a follow up)

iroh-io/README.md
```diff
@@ -0,0 +1,3 @@
+# Local (non-`Send`) async IO for rust
+
+Used by the iroh project. For more info see the docs.
```
dignifiedquire, line 3: Can you add the license thing we have in the other readmes please?

iroh-io/src/lib.rs
```diff
@@ -0,0 +1,603 @@
...
+    /// Write the entire fixed size array at the given position.
+    /// This is useful to write a small number of bytes without allocating a [bytes::Bytes].
+    ///
+    /// Except for taking a fixed size array instead of a [bytes::Bytes], this is equivalent to [AsyncSliceWriter::write_at].
+    #[must_use = "io futures must be polled to completion"]
+    fn write_array_at<const N: usize>(
+        &mut self,
+        offset: u64,
+        data: [u8; N],
+    ) -> Self::WriteArrayAtFuture<'_>;
```
dignifiedquire, line 107: Could be nice to provide a default impl that just fallsback to use `write_at`

iroh-io/Cargo.toml
rklaehn: yes, on my todo list, especially given that warp does http range requests wrong.

iroh-io/src/lib.rs
```diff
@@ -0,0 +1,603 @@
...
+    // #[tokio::test]
+    // async fn test_http_range() -> io::Result<()> {
+    //     let url = reqwest::Url::parse("https://ipfs.io/ipfs/bafybeiaj2dgwpi6bsisyf4wq7yvj4lqvpbmlmztm35hqyqqjihybnden24/image").unwrap();
+    //     let mut resource = HttpAdapter::new(url).await?;
+    //     let buf = resource.read_at(0, 100).await?;
+    //     println!("buf: {:?}", buf);
+    //     let buf = resource.read_at(1000000, 100).await?;
+    //     println!("buf: {:?}", buf);
+    //     Ok(())
+    // }
```
dignifiedquire, line 520: what's up with this test?

iroh-io/src/mem.rs
```diff
@@ -0,0 +1,102 @@
+use super::{AsyncSliceReader, AsyncSliceWriter};
+use bytes::{Bytes, BytesMut};
+use futures::future;
+use std::io;
+
+impl AsyncSliceReader for bytes::Bytes {
```
dignifiedquire, line 6: Do you plan on adding impls for `Vec<u8>` and `std::io::Cursor` potentially? might be quite nice

iroh-io/src/tokio_io.rs
```diff
@@ -0,0 +1,289 @@
...
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        match self.project() {
+            AsyncifyProj::Ok(f, h) => f.poll(cx).map(|x| {
+                match x {
+                    Ok((state, r)) => {
+                        // we got a result, so we can store the state
+                        **h = Some(state);
+                        r
+                    }
+                    Err(_) => Err(io::Error::new(
```
dignifiedquire, line 189: would probably be helpful to at least include the error in the message?

dignifiedquire (approved):
Looks good overall. I haven't reviewed the details of the impl, just general structure.

iroh-io/src/mem.rs
rklaehn: Yeah, for Vec it would basically be the same. Cursor I am not sure about since you don't have the notion of a "current position".

iroh-io/src/lib.rs
rklaehn: 😢 default impls would require defaults for associated types, which are unstable

dignifiedquire: 😭

rklaehn:
> Looks good overall. I haven't reviewed the details of the impl, just general structure.

The impls are as simple as possible. There is an entire big can of worms called buffering that I will need to add. But not in this PR

iroh-io/src/mem.rs
dignifiedquire: it would also be good to test that this interface works with https://docs.rs/mmap-rs (and potentially adding a cfged impl to this code) as this would likely be a nice speedup in a bunch of cases.

dignifiedquire: just a thought though

iroh-io/src/lib.rs
rklaehn: It's just a thing to explore the behaviour of one particular http server that supports range requests. Since it is on IPFS I don't want to keep it enabled. I guess I could just disable it instead of commenting it out.

iroh-io/src/mem.rs
rklaehn: Oh boy, that is a can of worms.

So - it is definitely possible to do an impl of this that uses memory mapping. There is a big *but* though: since there is no custom constructor for `Bytes` you would have to copy from the memory mapped region into the Bytes, even if you could ensure that the Bytes will keep the memory mapped buffer alive.

I have opened an issue on bytes::Bytes for this, and so have other people. But so far no success. For other projects I have therefore decided to write my own Bytes equivalent type. But for this crate I decided to go with Bytes since it is ubiquitous and the Bytes/BytesMut combo is quite helpful when doing zero alloc IO (splitting off Bytes from a BytesMut etc) and I did not feel like reimplementing all of this.

This was my issue:
https://github.com/tokio-rs/bytes/issues/571

The other issues:
https://github.com/tokio-rs/bytes/issues/437
https://github.com/tokio-rs/bytes/issues/526

My version of Bytes:
https://github.com/cloudpeers/radixdb/blob/master/src/store/blob_store.rs#L37

rklaehn: By the way: what is the current crate for memory mapped IO?

There is memmap, memmap2, and now this mmap-rs

I thought memmap2 was the current one... 🤷‍♂️

rklaehn: This is memmap2 download stats. It certainly seems to be popular...

image

PR #1117, opened 2023-06-20
feat(iroh-net): Upnp port mapping

divagant-martian:
This is the skateboard version of upnp mappings, which I think should cover a wide range of use cases to begin with.

Random notes on this (that depending on process will end up in one/several issues)

### Tailscale differences (behaviour)
- for probing, they first send a `ssdp:all` request (simple service discovery) this seems unnecessary to me since the type of device we want is an `igd` one. An `ssdp:all` request will be answered from many different devices, let's say for example speakers connected to the local network. I don't see this as useful since afterwards, the responses are filtered to gather the `igd` devices anyway. We don't do that, and instead send the `igd` request only.
- before sending the `igd` request, which is technically sent to a multicast address, ts sends also a unicast request to the likely router address. This is to deal with firewalls that reject the response to the multicast address since it's coming to the unicast address. Including this sounds something that we should do, but would mean to contribute or fork `igd` or move the code to our codebase. The `igd` code is fairly simple so I don't think any of these options is out of scope.
- the local address is for now set to ipv4 unspecified address. ts uses the interfaces address.
- renewing and creating a mapping is done as soon as it's identified (`await`ing) instead of "spawning a task" (goroutine) 
- no `on_change` when the mapping changes.
- "closing" the client in tailscale mean invalidating the mappings and marking the client as closed. However the closed value is only ever used to .. prevent double closing. I think we can simply invalidate the mappings, there is no real use for the closing.

### Other notes
This so far follows the og template but I'd like us to consider a couple points:
- No need to register a `on_change` callback. We probably can instead add a [`tokio::sync::watch`](https://docs.rs/tokio/latest/tokio/sync/watch/index.html) to keep track of the last changed value.
- The client is clone, which is surprising to me. Do we really want multiple port mapping clients to exist and operate concurrently? Depending on this moving this to be more of a service/actor would make sense.
- Updating the mapping is done in a very synchronous, reactive fashion. If we need a mapping and don't have one, we start creating it there. If the mapping is due for renewal, the renewal process starts in that instant (when we realize it's _late_ ). We could instead add the necessary futures to renew the mapping. For creation, we would create the mapping as soon as the local port is known and not when the mapping is requested.
These are all changes that make sense to me, but it would be good to have the opinion of someone with better understanding of how port mapping fits within the code as a whole, in concept. 

### Further notes
- I noticed that the netcheck code that calls the upnp mapping does not wait for it to complete. While this was simple to solve, I'd like to first check out the git history of the code to understand what the current code is supposed to do and adjust accordingly.
- What's the standard for logging?
- What kind of metrics do we want to check? How do we do metrics in general
- Tests will be interesting. I've tested the relevant parts of the code in a toy project and get a mapping, but in practice this probably depends on different kinds of routers and lan setups in general


dignifiedquire:
> * What's the standard for logging?

We use `tracing` crate for logging, some nice examples can be found in the `netcheck.rs` code.

dignifiedquire:
> * What kind of metrics do we want to check? How do we do metrics in general

This is how we define metrics: 
- https://github.com/n0-computer/iroh/blob/main/src/metrics/netcheck.rs
- you can look at `netcheck.rs` to see how we use it.

for a detailed walkthrough, ping @Arqu our metrics guy :D

dignifiedquire:
> * What kind of metrics do we want to check?

I would start with the metrics that tailscale collects, other than that my general approach is to collect 
- success & failure rates
- unexpected events

dignifiedquire:
> * Tests will be interesting.

Yes, maybe we could add some functionality to the `doctor` command that tries to do a port mapping and reports how well it worked, so it is easy for folks to run manual tests.

Arqu:
Dig mostly covered the bare basics of metrics. To have metrics collected run with `iroh --metrics-addr 127.0.0.1:9090 provide ...` then you can access the current snapshot of prometheus metrics on `localhost:9090`. There's a separate exporter you can run to push things up to our infra.

As for what metrics, there is no hard rule, metrics are there to serve us, the important ones will be surfaced more in grafana but most of them are useful. I usually go by add more as you need. Typically you want to cover major events, and output counters along with instrumenting main loops as interesting branches/error cases.

iroh-net/src/hp/portmapper.rs
```diff
@@ -14,41 +21,151 @@ pub struct ProbeResult {
...
+            }
+        }
     }
 
     /// Updates the local port number to which we want to port map UDP traffic.
-    pub async fn set_local_port(&self, _local_port: u16) {
-        // TODO:
+    /// Invalidates the current mapping if any.
+    pub async fn set_local_port(&mut self, local_port: u16) {
+        // TODO(@divagant-martian): this should never be 0
```
dignifiedquire, line 98: maybe use `NonZeroU16`?

iroh-net/src/hp/portmapper.rs
```diff
@@ -14,41 +21,151 @@ pub struct ProbeResult {
...
-    pub fn close(&self) {
-        // TODO:
+    pub async fn close(&mut self) {
+        // TODO(@divagant-martian):
+        // tailscale has a `closed` bool, which is not very pretty.
+        // Easiest way to emulate this is consume self and have the owner have an Option. This does
+        // bring some complexity, so maybe we can just not close (permanently) but simply clean
+        // (invalidate mappings)
+        //
+        // The only place this is ever used is to prevent closing twice...
```
dignifiedquire, line 168: Yeah, closing is not fully figured out, other parts also have the ability to close, but it seems kinda pointless as things can't be restarted. Ideally all the structures just consume `self`  or are able to be restarted.

iroh-net/src/hp/portmapper/upnp.rs
```diff
@@ -0,0 +1,119 @@
...
+            .await?;
+        let external_ip = self.gateway.get_external_ip().await?;
+        self.external_addr.set_ip(external_ip);
+
+        Ok(())
+    }
+
+    /// good_until will return the lease time that the mapping is valid for.
+    pub fn good_until(&self) -> Instant {
+        // assume an hour
```
dignifiedquire, line 86: is this a documented value in the spec, or is this just what tailscale "figured out"

iroh-net/src/hp/portmapper.rs
```diff
@@ -14,41 +21,151 @@ pub struct ProbeResult {
...
-    pub async fn get_cached_mapping_or_start_creating_one(&self) -> Option<SocketAddr> {
-        // TODO:
-        None
+    // TODO(@divagant-martian): fix docs. Also, this will probably only ever return ipv4 addresses.
+    pub async fn get_cached_mapping_or_start_creating_one(&mut self) -> Option<SocketAddr> {
+        // TODO(@divagant-martian): this has a lock ensure just one mapping attempt is underway
+        if let Some(mapping) = &mut self.current_mapping {
+            let now = Instant::now();
+            if now <= mapping.good_until() {
+                // TODO(@divagant-martian): this would go in a goroutine
```
dignifiedquire, line 115: Given this structure, it would probably be nicest to avoid locks etc, to just have a small actor like setup, for the mapping. Even though it is a bit more boilerplate, being able to avoid locks usually pays of in this kind of code in my experience.

dignifiedquire:
> * No need to register a `on_change` callback. We probably can instead add a [`tokio::sync::watch`](https://docs.rs/tokio/latest/tokio/sync/watch/index.html) to keep track of the last changed value.

Sounds good, though should probably check for what it is actually used  where and if we need it.

dignifiedquire:
> * The client is clone, which is surprising to me.

This is purely because of the structure of `netcheck.rs` at the moment. If we can make it work without, cool.

dignifiedquire:
> * Updating the mapping is done in a very synchronous, reactive fashion. If we need a mapping and don't have one, we start creating it there. If the mapping is due for renewal, the renewal process starts in that instant (when we realize it's _late_ ). We could instead add the necessary futures to renew the mapping. For creation, we would create the mapping as soon as the local port is known and not when the mapping is requested.
>   These are all changes that make sense to me, but it would be good to have the opinion of someone with better understanding of how port mapping fits within the code as a whole, in concept.

This sounds good to me in general, but I think it would be good to check with @flub how to best integrate this with his refactor plans for netcheck.

iroh-net/src/hp/portmapper.rs
```diff
@@ -14,41 +21,151 @@ pub struct ProbeResult {
...
-        bail!("not implemented yet")
+    /// Releases the current mapping and clears it from the cache.
+    async fn invalidate_mapping(&mut self) -> Result<(), Error> {
+        if let Some(old_mapping) = self.current_mapping.take() {
+            old_mapping.release().await?;
+        }
+        Ok(())
+    }
+
+    /// UPnP: searchs for an upnp internet gateway device (a router).
```
ramfox, line 44: nit: 'searchs' -> 'searches'

iroh-net/src/hp/portmapper/upnp.rs
```diff
@@ -0,0 +1,119 @@
...
+    /// The external address obtained by this mapping.
+    external_addr: SocketAddrV4,
+    /// Local address used to create this mapping.
+    local_addr: SocketAddrV4,
+    /// Instant in which the mapping was first created or last renewed.
+    created_at: Instant,
+}
+
+impl Mapping {
+    pub(crate) async fn new(local_addr: SocketAddrV4) -> Result<Self, Error> {
```
ramfox, line 31: We usually import `anyhow::Result` directly and use `Result<Self>` (for example).

iroh-net/src/hp/portmapper.rs
```diff
@@ -14,41 +21,151 @@ pub struct ProbeResult {
...
+            Ok(gateway_addr) => {
+                let old_gateway = self
+                    .last_upnp_gateway_addr
+                    .replace((gateway_addr, Instant::now()));
+                if let Some((old_gateway_addr, _last_seen)) = old_gateway {
+                    if old_gateway_addr != gateway_addr {
+                        tracing::debug!(
+                            "upnp gateway changed from {old_gateway_addr} to {gateway_addr}"
+                        );
+                        // TODO(@divagant-martian): tailscale does not invalidate the mappings here. Why?
```
ramfox, line 81: If we've searched for the router and the router has changed, maybe this implies that we don't have access to the old one anymore, so `release`-ing the mapping is maybe not possible, or just not our responsibility?

If our local address has changed, but our router is the same, I understand we want to release the mapping because we want them to not direct the traffic to the old local address. 

Anyways, I think it's fine to not invalidate the mappings in this case.

iroh-net/src/hp/portmapper.rs
```diff
@@ -14,41 +21,151 @@ pub struct ProbeResult {
...
+        } else {
+            match upnp::Mapping::new(SocketAddrV4::new(
+                std::net::Ipv4Addr::UNSPECIFIED,
+                self.local_port,
+            ))
+            .await
+            {
+                Ok(mapping) => {
+                    // TODO(@divagant-martian): too high? seems relevant to the user
+                    info!("upnp port mapping created {mapping}");
```
ramfox, line 136: yea maybe move to debug? Possibly only relevant if you are looking for it.

iroh-net/src/hp/portmapper.rs
ramfox: I don't have anything useful to say about this except to just complain about all the `closed` bools in the go code. I hate them everytime.

ramfox (changes requested):

iroh-net/src/hp/portmapper/upnp.rs
divagant-martian: their value is actually much higher, 2 hours, and supposedly comes from a recommendation from an RFC for another port mapping protocol. an hour seemed reasonable to me, but I can change it to the same 2 hours they use if that's the preference

iroh-net/src/hp/portmapper.rs
divagant-martian: Agree. In general the way portmapping is wired can get more love but it's understandable this was not clear until just now

iroh-net/src/hp/portmapper.rs
divagant-martian: done in https://github.com/n0-computer/iroh/pull/1117/commits/4d2b14bdebd6390ca0974f01c0772d6eb26b2200

iroh-net/src/hp/portmapper/upnp.rs
divagant-martian: done in [32d5a0c](https://github.com/n0-computer/iroh/pull/1117/commits/32d5a0c5022e76bd764d0c8d7431c71fb284e6ad)

iroh-net/src/hp/portmapper.rs
divagant-martian: you are right about releasing the mappings, sorry for the lack of precision: I mean they don't clear the mappings. So we still have `self.current_mapping` as `Some` even if, as you pointed, the router is not there anymore

iroh-net/src/util.rs
```diff
@@ -29,3 +29,43 @@ impl<T> Drop for AbortingJoinHandle<T> {
...
+///
+/// The task can be manually aborted, or aborted once self is dropped.
+/// See [`tokio::task::AbortHandle`].
+pub struct CancelOnDrop {
+    task_name: &'static str,
+    handle: tokio::task::AbortHandle,
+}
+
+impl CancelOnDrop {
+    pub fn new(task_name: &'static str, handle: tokio::task::AbortHandle) -> Self {
```
flub, line 51: ugh, TIL that `tokio::task::Builder` is unstable.  I swear I had used task names in the past...   Which is sad since tokio already has the concept of a task name, if only it was stable.  A `From` impl would have been nice, but I guess we can't without task names.

iroh-net/src/util.rs
```diff
@@ -29,3 +29,43 @@ impl<T> Drop for AbortingJoinHandle<T> {
...
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+/// Holds a handle to abort a task.
+///
+/// The task can be manually aborted, or aborted once self is dropped.
+/// See [`tokio::task::AbortHandle`].
+pub struct CancelOnDrop {
```
flub, line 45: good call to start creating a util for this, we have this in several places by now.

iroh-net/src/util.rs
```diff
@@ -29,3 +29,43 @@ impl<T> Drop for AbortingJoinHandle<T> {
...
+}
+
+impl Drop for CancelOnDrop {
+    fn drop(&mut self) {
+        self.handle.abort();
+        tracing::debug!("{} completed", self.task_name);
+    }
+}
+
+impl std::fmt::Debug for CancelOnDrop {
```
flub, line 65: i think we've been trying to use `derive_more::Debug` instead of manual debug impls where possible.

iroh-net/src/hp/portmapper.rs
```diff
@@ -12,43 +30,435 @@ pub struct ProbeResult {
...
+
+        let service = Service {
+            local_port,
+            last_upnp_gateway_addr: None,
+            rx: service_rx,
+            current_external_address: current_mapping_tx,
+        };
+
+        let handle =
+            util::CancelOnDrop::new("portmap_service", tokio::spawn(async {}).abort_handle());
```
dignifiedquire, line 73: Is this like for later, or what exactly is the point of this empty `async {}`block?

iroh-net/src/hp/portmapper.rs
```diff
@@ -12,43 +30,435 @@ pub struct ProbeResult {
...
+            service_handle: std::sync::Arc::new(handle),
+        };
+        client
+    }
+
+    /// Request a probe to the port mapping protocols.
+    pub fn probe(&self) -> oneshot::Receiver<Result<ProbeResult>> {
+        let (result_tx, result_rx) = oneshot::channel();
+
+        use mpsc::error::TrySendError::*;
```
dignifiedquire, line 87: nit: move into the `if let Err` scope

iroh-net/src/hp/portmapper.rs
```diff
@@ -12,43 +30,435 @@ pub struct ProbeResult {
...
+                }
+                Closed(Message::Probe { result_tx }) => {
+                    (result_tx, anyhow!("Port mapping channel closed"))
+                }
+                Full(_) | Closed(_) => unreachable!("Sent value is a probe."),
+            };
+
+            result_tx
+                .send(Err(e))
+                .expect("receiver counterpart has not been dropped or closed.");
```
dignifiedquire, line 102: I would probably log this, instead of panicing, often times on shutdown these things end up getting dropped. (unless of course you are sure that doesn't happen)

iroh-net/src/hp/portmapper/upnp.rs
dignifiedquire: Happy to go with one hour for now, just add a comment where the number comes from please

divagant-martian:
As I'm now moving this to be and actor-service pair I don't think you guys need to review this for now

iroh-net/src/hp/netcheck.rs
```diff
@@ -120,6 +122,22 @@ impl Report {
     pub fn any_port_mapping_checked(&self) -> bool {
         self.upnp.is_some() || self.pmp.is_some() || self.pcp.is_some()
     }
+
+    /// Updates the portmap related values based on a [`portmapper::ProbeResult`].
+    pub fn update_portmap_probe(&mut self, probe_result: Option<portmapper::ProbeOutput>) {
+        match probe_result {
+            Some(portmapper::ProbeOutput { upnp, pmp, pcp }) => {
```
dignifiedquire, line 129: Suggestion, lest just just store the `ProbeOutput` instead of these three individual fields.

iroh/src/commands/doctor.rs
```diff
@@ -97,6 +98,14 @@ pub enum Commands {
         #[clap(long)]
         local_derper: bool,
     },
+    /// Attempt to get a port mapping to the given local port.
+    PortMap {
+        /// Local port to get a mapping.
+        local_port: NonZeroU16,
+        /// How long to wait for an external port to be ready.
```
dignifiedquire, line 105: nit: `in seconds` or sth like that is missing.

iroh-net/src/hp/portmapper/upnp.rs
```diff
@@ -0,0 +1,140 @@
...
+            .map_err(|_| anyhow::anyhow!("upnp mapping got zero external port"))?;
+
+        Ok(Mapping {
+            gateway,
+            external_ip,
+            external_port,
+        })
+    }
+
+    pub fn half_lifetime(&self) -> Duration {
```
dignifiedquire, line 101: could be `const`

iroh-net/src/hp/portmapper/mapping.rs
```diff
@@ -0,0 +1,233 @@
...
+            };
+            update
+        });
+        old_mapping
+    }
+
+    fn poll(&mut self, cx: &mut std::task::Context<'_>) -> Poll<Event> {
+        // grab the waker if needed
+        if let Some(waker) = &self.waker {
+            if waker.will_wake(cx.waker()) {
```
dignifiedquire, line 113: what exactly does this check do?

dignifiedquire:
When I try and run the doctor command on a network where upnp is disabled it just times out, but looking at the log it should  actually error. Any chance that error could be properly reported?

Screenshot 2023-07-03 at 10 41 11

iroh-net/src/hp/portmapper.rs
```diff
@@ -1,54 +1,522 @@
...
+                    probe.last_upnp_gateway_addr = last_upnp_gateway_addr;
+                    upnp_done = true;
+                },
+                last_pmp = &mut pmp_probing_task, if !pmp_done => {
+                    trace!("tick: pmp probe ready");
+                    probe.last_pmp = last_pmp;
+                    pmp_done = true;
+                },
+                last_pcp = &mut pcp_probing_task, if !pcp_done => {
+                    trace!("tick: pcp probe ready");
```
dignifiedquire, line 223: tick tock, tick tock 🙃

iroh-net/src/hp/portmapper.rs
```diff
@@ -1,54 +1,522 @@
...
+            self.last_pcp = last_pcp;
+        }
+        if last_pmp.is_some() {
+            self.last_pmp = last_pmp;
+        }
+    }
+}
+
+// mainly to make clippy happy
+type ProbeResult = Result<ProbeOutput, String>;
```
dignifiedquire, line 293: Why `String`?

dignifiedquire:
Okay, tried to run this locally and got into the following issue. The `local_addr` ends up being set to `127.0.0.1` when requesting a mapping, resulting in an error, as can be seen here (using the rust-igd examples)

Screenshot 2023-07-03 at 11 08 10

When I use my local lan IP `192.168.x.x` things work as expected in all the examples for `rust-igd`.

iroh-net/src/hp/portmapper.rs
divagant-martian: Anyhow errors aren't clone

divagant-martian:
> When I try and run the doctor command on a network where upnp is disabled it just times out, but looking at the log it should  actually error. Any chance that error could be properly reported?

Not sure what you mean, in your screenshot there is both the log and the error.

Can you elaborate?

dignifiedquire:
> Can you elaborate?

The error is only reported in the logs. If I run without logs only this shows up

Screenshot 2023-07-03 at 12 50 09

iroh-net/src/hp/portmapper/upnp.rs
divagant-martian: pmp mappings seem to not have predefined lifetimes, so I left it as a function to integrate better with that later on. But at least for upnp yeah it could be used as a constant directly or make the fn const. Which way would you suggest with all that in mind?

dignifiedquire: I guess lets leave it as it is then

divagant-martian:
@dignifiedquire ok I see. This is somewhat intentional and I see both approaches making sense. Right now I report whether there is an external address and if it changed on the watcher. In the doctor command, a failure would get a `None` report to a `None` report (so intentionally not really reported". This is the approach tailscale has. 

If you think it makes sense, I could change this option to a result and propagate it up. The only place where this would make a difference tho is in the doctor command

iroh/src/commands/doctor.rs
divagant-martian: done in 552e1c7

iroh-net/src/hp/portmapper/mapping.rs
divagant-martian: an optimization to prevent switching wakers when the task has not moved threads

dignifiedquire:
I see, that makes sense. Given that we can get the errors via logs, I think we can leave it.

iroh-net/src/hp/netcheck.rs
divagant-martian: done in 0846437

divagant-martian:
This should be ready in any case, since I moved to using the same dep as #1157 for the local ip.
Home router is not necessary yet, since the addr in which an upnp gateway answers is a multicast addr following the spec, not it's unicast addr

dignifiedquire (approved):
lets go :)

PR #1119, opened 2023-06-21
use iroh-io everywhere

rklaehn:
this is the first step of pluggable IO

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,9 @@ opt-level = 3
 panic = 'abort'
 incremental = false
 
+[patch.crates-io]
+iroh-io = { path = "./iroh-io" }
+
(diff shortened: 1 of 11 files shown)


Cargo.toml
```diff
@@ -15,6 +15,9 @@ opt-level = 3
 panic = 'abort'
 incremental = false
 
+[patch.crates-io]
+iroh-io = { path = "./iroh-io" }
```
rklaehn, line 19: Without this the dependency iroh-io of bao tree and the local stuff does not get unified. Not sure what to do about that.

rklaehn: actually it is pretty clear - have a dependency on crates.io everywhere.

dignifiedquire (approved):

PR #1121, opened 2023-06-21
refactor: move cli functionality into library format

dignifiedquire:
Depends on #1120

This is a first pass to pull out some of the functionality hidden in `iroh/main` into a more usable format.


iroh-bytes/src/get/io.rs
```diff
@@ -0,0 +1,522 @@
...
+        let response = match self {
+            GetInteractive::Ticket {
+                ticket,
+                keylog,
+                derp_map,
+            } => get::run_ticket(&ticket, request, keylog, derp_map).await?,
+            GetInteractive::Hash { opts, .. } => get::run(request, opts).await?,
+        };
+        let connected = response.next().await?;
+        // progress!("{} Requesting ...", style("[2/3]").bold().dim());
```
ramfox, line 390: Leaving this out makes the CLI output change / not make sense. We have 1/3 & 3/3 but no 2/3 anymore.

dignifiedquire: ahh I missed this 😓 thanks

iroh/src/commands/provide.rs
```diff
@@ -0,0 +1,231 @@
...
+
+async fn provide(
+    db: Database,
+    addr: SocketAddr,
+    key: Option<PathBuf>,
+    keylog: bool,
+    rpc_port: Option<u16>,
+    dm: Option<DerpMap>,
+    rt: &runtime::Handle,
+) -> Result<Node> {
```
b5, line 124: possible to move this up into a new `iroh/src/provide.rs` file?

dignifiedquire: This is very CLI specific, with connecting to RPC etc, so in this form that seems the wrong place. What functionality are you looking for? the core functionality is this

```
    let keypair = get_keypair(key).await?;

    let mut builder = Node::builder(db).keylog(keylog);
    if let Some(dm) = dm {
        builder = builder.derp_map(dm);
    }
    let builder = builder.bind_addr(addr).runtime(rt);
    builder.keypair(keypair).spawn().await?
```
which is already the api that was designed for external consumption.

iroh/src/commands.rs
```diff
@@ -0,0 +1,263 @@
...
+        /// Path to initial file or directory to provide
+        path: Option<PathBuf>,
+        #[clap(long, short)]
+        /// Listening address to bind to
+        #[clap(long, short, default_value_t = SocketAddr::from(crate::node::DEFAULT_BIND_ADDR))]
+        addr: SocketAddr,
+        /// RPC port, set to "disabled" to disable RPC
+        #[clap(long, default_value_t = ProviderRpcPort::Enabled(DEFAULT_RPC_PORT))]
+        rpc_port: ProviderRpcPort,
+    },
```
b5, line 155: I'd vote to move this into `commands/provide.rs`, the current contents of `commands/provide.rs` into `src/provide.rs`, and call a public function from `src/provide.rs` within `commands::provide::run`

dignifiedquire: See comment above, the public api for providing is currently the `Node` builder.

iroh/src/commands/provide.rs
b5: ah ok, I'll experiment with that.

b5 (approved):

PR #1122, opened 2023-06-21
refactor(netcheck): minor code improvements

divagant-martian:
- removes an unnecessary clone
- makes sure the port mapping probe result is used entirely by destructuring it
- changes a log to be emitted only if it's relevant

### DIFF
--- a/iroh-net/src/hp/netcheck.rs
+++ b/iroh-net/src/hp/netcheck.rs
@@ -519,7 +519,7 @@ impl ReportState {
             if let Some(port_mapper) = port_mapper {
                 port_mapping.inner = Some(Box::pin(async move {
                     match port_mapper.probe().await {
-                        Ok(res) => Some((res.upnp, res.pmp, res.pcp)),
+                        Ok(res) => Some(res),
                         Err(err) => {
                             warn!("skipping port mapping: {:?}", err);
                             None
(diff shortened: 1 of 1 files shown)


dignifiedquire (approved):

iroh-net/src/hp/netcheck.rs
```diff
@@ -693,14 +693,16 @@ impl ReportState {
...
+            debug!("aborting {} probes, already done", probes.len());
+        }
         drop(probes);
 
         if let Some(hair_pin) = self.wait_hair_check().await {
             self.report.write().await.hair_pinning = Some(hair_pin);
         }
 
-        if !skip_external_network && port_mapper.is_some() {
+        if do_port_map {
```
flub, line 705: so we start portmapping anyway as long as we're allows to use an external network, but don't always bother waiting for it?  what's the logic for that?

flub (approved):

iroh-net/src/hp/netcheck.rs
divagant-martian: I'm not sure what you are referring to. The logic is akin to what was in here already. In reality we could just do `port_mapping.inner.is_some()`

flub: yeah, i know the logic didn't change.  i was only trying to understand myself why it was this way and given you modified it maybe you knew...

divagant-martian: ah I see. My understanding is that we always finish it actually. At least that's my understanding from tailscale's code. Code as is pretends to do it with the line just below this one without actually polling the future. But in theory, yes, we should finish the probe

dignifiedquire:
ci seems sad

divagant-martian:
> ci seems sad

dump report failed + cargo deny, neither seems related to the changes. Made the sims one run again to be sure

dignifiedquire:
> cargo deny,

trying to fix this here: https://github.com/n0-computer/iroh/pull/1141

flub (approved):

PR #1123, opened 2023-06-22
feat(doctor): add simple port mapping subcommand

divagant-martian:


rklaehn:
New features for the doctor - cool.

What is the expected behaviour? Tried this out, and it just says `Error: No port mapping found`.

divagant-martian:
So far nothing until port mapping is done. But I figure we can start moving this ahead in the meantime

flub (approved):

PR #1124, opened 2023-06-22
refactor: replace warp with axum for the test server

rklaehn:
this also makes the tests not create a file.

With very friendly and polite support from GPT-4

I had to exclude windows from the range request tests because axum has some issues there. But I think this is OK. The purpose of the test is to make sure we properly deal with range requests etc.


b5 (approved):
LGTM, agreed not having windows isn't really a deal-breaker here.

I'd defer to someone from engineering for a proper 👍 before merging

rklaehn:
> LGTM, agreed not having windows isn't really a deal-breaker here.
> 
> I'd defer to someone from engineering for a proper 👍 before merging

Don't be so humble. I think this 👍 is good enough for me. We already have axum as a test dependency, so it's not like it is a controversial choice...

PR #1126, opened 2023-06-22
test: bring back tests

### LINKED ISSUE #1125: Restore tests ignored by Crates Refactor
#1111 commented out a bunch of tests. Let's get them back.

dignifiedquire:
Closes #1125


rklaehn (approved):
maybe we can make the tests prettier later, but having them is better than not having them

rklaehn (approved):

Issue #1084, opened 2023-06-06
netsim integration tests with a NAT on the provide side fail


Arqu:
Sample https://github.com/n0-computer/iroh/actions/runs/5189008447/jobs/9353406369

Issue #1098, opened 2023-06-12
Make connection instantaneous when ticket contains contactable address


flub:
When a connection is created form a ticket which contains a valid and directly contactable address the connection establishment should be instantaneous.  It seems currently this is still blocked on performing the netcheck or something similar, but these operations should be racing each other and the direct connection should win immediately.

flub:
I think at least since #1984 we haven't had complaints about this.  Let's consider it closed.

PR #1128, opened 2023-06-22
fix: improve connectivity 

dignifiedquire:
- make sure to send a full ping (and call me maybe) when doing the initial ping
- handle disco messages when they are coming over derp. before they failed to be extracted due to the packet combinations & the prefixing.

Should improve https://github.com/n0-computer/iroh/issues/1098
Fixes https://github.com/n0-computer/iroh/issues/1084


iroh-net/src/hp/magicsock/endpoint.rs
```diff
@@ -704,8 +704,11 @@ impl Endpoint {
         // Send heartbeat ping to keep the current addr going as long as we need it.
         if let Some(udp_addr) = udp_addr {
             let elapsed = self.last_ping(&udp_addr).map(|l| now - l);
-            // Send a ping if the last ping is either older than 2 seconds or we don't have one.
-            let needs_ping = elapsed.map(|e| e >= Duration::from_secs(2)).unwrap_or(true);
+            // Send a ping if the last ping is either than 2 seconds.
```
ramfox, line 707: "either" => "older"

iroh-net/src/hp/magicsock/udp_actor.rs
```diff
@@ -125,26 +127,26 @@ impl UdpActor {
                                             warn!("ip_sender gone");
                                             break;
                                         };
-                                        continue;
-                                    }
+                                    } else {
+                                        // Foward
```
ramfox, line 131: "Foward" => "Forward"

ramfox (dismissed):
small doc nits otherwise 🔥

ramfox:
question tho, what are these ubuntu cli failures?

Arqu:
FWIW this still fails in netsim for the NAT tests.

Arqu:
You can get tests on your PR running if you modify `.github/workflows/ci.yml` `L334` `sudo python3 main.py --integration --skip intg_derper__1_to_1_NAT_provide,intg_derper__1_to_1_NAT_both sims/integration` into `sudo python3 main.py --integration sims/integration`

It should error out on those tests until they pass.

Arqu:
@dignifiedquire you did the wrong one, `netsim.yml` only runs on main for final numbers, `ci.yml` runs the PR tests.

Arqu:
It's fine to keep it though if we're going to fix nat stuff in this PR

dignifiedquire:
> It's fine to keep it though if we're going to fix nat stuff in this PR

That's the plan, it is fixed in my home test setup, when the both parties are behind different nats.

Arqu:
When running manually on my end, seems to do more now, but still fails connecting to the provider.

iroh-net/src/hp/magicsock/endpoint.rs
```diff
@@ -125,39 +130,113 @@ impl Endpoint {
...
+        }
+
+        // If we found a candidate, set to best addr
+        if let Some(addr) = udp_addr {
+            self.best_addr = Some(AddrLatency {
+                addr,
+                latency: Some(lowest_latency),
+            });
+            self.trust_best_addr_until
+                .replace(*now + Duration::from_secs(60 * 60));
```
ramfox, line 185: maybe trust from `last_ping` time, rather than now? & if that time exceeds an hour it is no longer a `best_addr` candidate?

dignifiedquire: fixed

Arqu:
Seems like the failures are real connectivity issues. (again with a NAT on the provide side)

iroh-net/src/hp/magicsock/endpoint.rs
```diff
@@ -112,52 +116,122 @@ impl Endpoint {
...
+            .keys()
+            .choose_stable(&mut rand::thread_rng())
+            .copied();
+        if let Some(addr) = udp_addr {
+            self.best_addr = Some(AddrLatency {
+                addr,
+                latency: None,
+            });
+
+            self.trust_best_addr_until = Some(*now + Duration::from_secs(15));
```
ramfox, line 201: Okay, so from looking at logs in the failing cli test, there are two main issues i'm seeing.

This issue is apparent on the get side, since as soon as we connect we try to send something. 

Issue One, that needs some more thinking:
The first time we attempt to do a full ping with a `callmemaybe` over derp is the first time we _ever_ connect to the derp server. This requires a full TLS handshake and takes some time to negotiate. But `callmemaybe` is sensitive to time. This is highly likely to never hole punch. 

Issue Two:
This is related to the actual code I'm commenting on 🤣 . Once we set the `best_addr` and `trust_best_addr`, we are potentially condemned to using a bad address for 15 seconds. Plus, since it's considered valid, we won't back it up with a derp address even though we have had no proof (yet) that it is a valid address.

One possible solution would be to keep `self.trust_best_addr_until` `None`, until a `full_ping` comes back successfully. But, this means that if another pong has come back on a different address during this time, we won't ever attempt to use it, because we still have a bad address as the `best_addr`.

I'm proposing, instead, that we never set an address as a `best_addr` unless it has a latency. 

So that would mean just removing these lines. (this fixes the bugs, locally for me, at least).

if we are worried about candidate address flip/flopping, we can keep track of a `candidate_addr`, that is the address we have randomly chosen as the default to use until we get any `pong`s back. We can loop through the addresses, and if none have any latency, we default to using the previous `candiate_addr`.

.github/workflows/ci.yml
```diff
@@ -331,7 +331,7 @@ jobs:
         sudo kill -9 $(pgrep ovs)
         sudo mn --clean
         sudo python3 main.py --integration sims/iroh/iroh.json
-        sudo python3 main.py --integration --skip intg_derper__1_to_1_NAT_provide,intg_derper__1_to_1_NAT_both sims/integration
+        sudo python3 main.py --integration --debug sims/integration
```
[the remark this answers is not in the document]
dignifiedquire: done

iroh-net/src/hp/magicsock/endpoint.rs
dignifiedquire: > are potentially condemned to using a bad address for 15 seconds.

Well only for `5s`, as the pings get sent out and if we time out, it gets removed.

> Plus, since it's considered valid, 

this was the big problem, we drop derp because we think it is valid, forgot my own logic last night..

Arqu (approved):
⭕ 👊

PR #1131, opened 2023-06-23
feat(provider): add 'CollectionAdded' Provider event

b5:
Need this for iroh.network, which listens to an in-process node for provider collection construction.

I'm piggybacking this to also add a test that does minimal node setup, which should form the basis of writing examples later. Learned many things in the process.

### DIFF
--- a/iroh-bytes/src/provider.rs
+++ b/iroh-bytes/src/provider.rs
@@ -34,7 +34,7 @@ pub use ticket::Ticket;
 /// Events emitted by the provider informing about the current status.
 #[derive(Debug, Clone)]
 pub enum Event {
-    /// A new collection has been added via an RPC call
+    /// A new collection has been added
     CollectionAdded {
         /// The hash of the added collection
         hash: Hash,
--- a/iroh-bytes/src/provider.rs
+++ b/iroh-bytes/src/provider.rs
@@ -34,6 +34,11 @@ pub use ticket::Ticket;
 /// Events emitted by the provider informing about the current status.
 #[derive(Debug, Clone)]
 pub enum Event {
+    /// A new collection has been added via an RPC call
+    CollectionAdded {
+        /// The hash of the added collection
+        hash: Hash,
+    },
     /// A new client connected to the node.
     ClientConnected {
         /// An unique connection id.
--- a/iroh/src/commands/provide.rs
+++ b/iroh/src/commands/provide.rs
@@ -175,6 +175,7 @@ async fn get_keypair(key: Option<PathBuf>) -> Result<Keypair> {
     }
 }
 
+/// Makes a an RPC endpoint that uses a QUIC transport
 fn make_rpc_endpoint(
     keypair: &Keypair,
     rpc_port: u16,
--- a/iroh/src/node.rs
+++ b/iroh/src/node.rs
@@ -35,7 +35,7 @@ use quic_rpc::{RpcClient, RpcServer, ServiceConnection, ServiceEndpoint};
 use tokio::sync::{broadcast, mpsc};
 use tokio::task::JoinError;
 use tokio_util::sync::CancellationToken;
-use tracing::{debug, trace};
+use tracing::{debug, trace, warn};
 
 use crate::rpc_protocol::{
     AddrsRequest, AddrsResponse, IdRequest, IdResponse, ListBlobsRequest, ListBlobsResponse,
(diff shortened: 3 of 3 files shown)


iroh/src/commands/provide.rs
```diff
@@ -175,6 +175,7 @@ async fn get_keypair(key: Option<PathBuf>) -> Result<Keypair> {
     }
 }
 
+/// Makes a an RPC endpoint that uses a QUIC transport
 fn make_rpc_endpoint(
```
b5, line 179: went on a wild goose chase looking for a way to construct an RPC endpoint to give to the builder, and at one point wanted to move this, only to discover `node.controller()`. This comment is a momento of hour(s) lost.

iroh/src/node.rs
```diff
@@ -585,13 +585,19 @@ impl RpcHandler {
...
             data_sources,
             Progress::new(progress),
         )
         .await?;
         self.inner.db.union_with(db);
 
+        if let Err(e) = self.inner.events.send(Event::ByteProvide(
+            iroh_bytes::provider::Event::CollectionAdded { hash },
+        )) {
+            warn!("failed to send CollectionAdded event: {:?}", e);
```
b5, line 598: this might be overkill, but I figured I'd err on the side of being pedantic

iroh/src/node.rs
```diff
@@ -728,4 +736,72 @@ mod tests {
...
+            while let Ok(msg) = events.recv().await {
+                match msg {
+                    Event::ByteProvide(e) => {
+                        if let iroh_bytes::provider::Event::CollectionAdded { hash } = e {
+                            got_hash = Some(hash);
+                            break;
+                        }
+                    }
+                }
+            }
```
b5, line 765: This test could be expanded in the future for more extensive provider event testing

iroh/src/node.rs
```diff
@@ -728,4 +736,72 @@ mod tests {
...
+    async fn test_node_add_collection_event() {
+        let db = Database::from(HashMap::new());
+        let node = Builder::with_db(db)
+            .bind_addr((Ipv4Addr::UNSPECIFIED, 0).into())
+            .runtime(&test_runtime())
+            .spawn()
+            .await
+            .unwrap();
+
+        let _drop_guard = node.cancel_token().drop_guard();
```
b5, line 750: do I need to do this?

iroh/src/commands/provide.rs
rklaehn: Sorry about this. So the idea is that you always want a cheap in memory way to control your node via RPC like calls.

Basically like an in memory message channel between you and the node, but nicer than the usual pattern of sending messages that contain a oneshot sender for the reply. This is what the controller is for.

rklaehn (commented):
This seems quite reasonable.

However, please be aware that there is a problem with using the events for any bookkeeping. I frequently have the problem that the many_files test fails because as you optimize things it can produce events very quickly, and the broadcast channel has only a fixed buffer size before it starts dropping messages.

If you want to use these events for anything important, we would have to have a way to prevent the dropping. Unbounded channels, or blocking when you are unable to send.

dignifiedquire:
> I frequently have the problem that the many_files test fails because as you optimize things it can produce events very quickly, and the broadcast channel has only a fixed buffer size before it starts dropping messages.

See failing test in ci, because of exactly that reason

iroh-bytes/src/provider.rs
```diff
@@ -34,6 +34,11 @@ pub use ticket::Ticket;
 /// Events emitted by the provider informing about the current status.
 #[derive(Debug, Clone)]
 pub enum Event {
+    /// A new collection has been added via an RPC call
```
dignifiedquire, line 37: Can we make this more generic, so it always happens, not just on RPC call, like the other events?

Winterhuman:
(Feel free to mark this as offtopic, I couldn't find a repo for iroh.network)

@b5 https://iroh.network shows `<meta name="frontend-version" content="%REACT_APP_GIT_SHA%">` in it's `<head>` element btw, I think that means something's broken.

Arqu:
Dunno what this did, but seems tests run forever. Canceling the current batch, but needs looking into it.

iroh/src/node.rs
dignifiedquire: I am not sure, @flub ?

dignifiedquire (approved):
assuming CI passes, lgtm

PR #1132, opened 2023-06-23
Make the database from which the provider serves generic

rklaehn:
So far this is only the *read* part, the write part will need some more work because it needs the concept of writing files, and making that generic requires a minimal VFS abstraction.

But I would argue that there are several cases where you just want to serve a bunch of existing data via the iroh protocol, and this is very easy to do now. See the InMemDatabase that is used for testing.

The thing that you need to implement is the trait currently named `BaoCollection`. It has just a get function that can do a lookup by hash. What you get is a `BaoCollectionEntry` which can open both an outboard and a data reader.

The BaoCollectionEntry is Send, Sync, 'static and is supposed to be cheaply cloneable. So you can send entries around threads and tasks. It provides two async fns, for opening the outboard or data reader. These futures are *also* Send.

What is **not** Send are the actual readers that you get on success. So once you got the two readers (like open files) you need to stay on a thread. The easiest way to do this is to use https://docs.rs/tokio-util/latest/tokio_util/task/struct.LocalPoolHandle.html#method.spawn_pinned .

The iroh runtime provides a local pool for you that you can use.

This depends on rc versions of bao-tree and iroh-io. I guess that is OK for a PR, but we would have to release those before we do a release.

Edit: to serve from http, all you would have to do would be to do something similar to InMemDatabase, but just containing a BTreeMap<Hash, (Url, Url)>. Then Entry would be just (Hash, Url, Url) and the reader would be an iroh_io::HttpAdapter.


rklaehn:
We have too many handlers and stuff, and it is only going to get more. So we need to reorganize this and maybe even use Box<dyn ...> somewhere. But I would prefer to to this in a separate PR.

iroh-bytes/Cargo.toml
```diff
@@ -13,15 +13,15 @@ rust-version = "1.65"
...
+bao-tree = { version = "0.5.0-rc.1", features = ["tokio_fsm"], default-features = false }
 blake3 = "1.3.3"
 bytes = { version = "1.4", features = ["serde"] }
 data-encoding = "2.3.3"
 derive_more = { version = "0.99.17", git = "https://github.com/JelteF/derive_more", features = ["debug", "display", "from", "try_into"] }
 flume = "0.10.14"
 futures = "0.3.25"
 hex = "0.4.3"
-iroh-io = { path = "../iroh-io" }
+iroh-io = { version = "0.2.0-rc.1" }
```
dignifiedquire, line 24: this should be both, so in development mode it always refers to the local version, otherwise things get quite problematic, especially with releases

iroh-bytes/src/blobs.rs
```diff
@@ -26,6 +26,11 @@ impl Collection {
         })
     }
 
+    /// Serialize this collection to a std Vec<u8>
+    pub fn to_bytes(&self) -> Result<Vec<u8>> {
```
dignifiedquire, line 30: `as_bytes`? `to_` usually takes `self`

iroh-bytes/src/provider.rs
```diff
@@ -672,7 +672,7 @@ pub async fn send_blob<W: AsyncWrite + Unpin + Send + 'static>(
             debug!("done sending blob {} {:?}", name, res);
             res?;
 
-            Ok((SentStatus::Sent, size))
+            Ok((SentStatus::Sent, 0))
```
dignifiedquire, line 675: why is this `0`?

iroh-bytes/src/provider/database.rs
```diff
@@ -31,6 +36,156 @@ pub const FNAME_PATHS: &str = "paths.bin";
 #[derive(Debug, Clone, Default)]
 pub struct Database(Arc<RwLock<HashMap<Hash, DbEntry>>>);
 
+#[derive(Debug, Clone, Default)]
+pub struct InMemDatabase(Arc<HashMap<Hash, (PreOrderMemOutboard, Bytes)>>);
```
dignifiedquire, line 40: maybe make a `database` mod, these files are getting pretty big

iroh-bytes/src/provider/database.rs
```diff
@@ -31,6 +36,156 @@ pub const FNAME_PATHS: &str = "paths.bin";
...
+    }
+}
+
+impl BaoCollection for InMemDatabase {
+    type Outboard = PreOrderMemOutboard<Bytes>;
+    type DataReader = Bytes;
+    type Entry = InMemDatabaseEntry;
+
+    fn get(&self, hash: &Hash) -> Option<Self::Entry> {
+        println!("getting entry for {:?}", hash);
```
dignifiedquire, line 111: these should be probably `debug`s or removed

iroh-bytes/src/provider/database.rs
```diff
@@ -31,6 +36,156 @@ pub const FNAME_PATHS: &str = "paths.bin";
...
+            data: d.clone(),
+        })
+    }
+}
+
+/// An entry for one hash in a bao collection
+///
+/// The entry has the ability to provide you with an (outboard, data)
+/// reader pair. Creating the reader is async and may fail. The futures that
+/// create the readers must be `Send`, but the readers themselves
```
dignifiedquire, line 125: missing `not`?

iroh-bytes/Cargo.toml
rklaehn: yeah, the problem is that bao-tree, which is not in this repo, depends on iroh-io, which is in this repo. This is a bit of a mess. Either pull bao-tree in, make iroh-io a separate repo, or remove the stuff that uses iroh-io from bao-tree and put it somewhere here.

Not quite sure what to do about it.

iroh-bytes/src/blobs.rs
rklaehn: I think into_... takes self. It seems OK according to the conventions. Although I also never really know in some cases...

image

This is an expensive conversion, borrowed -> owned, so to_seems appropriate.

iroh-bytes/src/provider.rs
rklaehn: I didn't have the size at some point. I have it again now, so no I will put it in.

iroh-bytes/src/provider/database.rs
rklaehn: definitely removed.

iroh-bytes/src/provider/database.rs
rklaehn: If I had done this I would have been in even more of a merge hell than I was already... So maybe leave it for the next PR?

iroh-bytes/src/blobs.rs
dignifiedquire: 👍

iroh-bytes/src/provider/database.rs
dignifiedquire: sure

iroh-bytes/src/provider.rs
```diff
@@ -359,23 +364,22 @@ pub async fn read_request(mut reader: quinn::RecvStream, buffer: &mut BytesMut)
 /// close the writer, and return with `Ok(SentStatus::NotFound)`.
 ///
 /// If the transfer does _not_ end in error, the buffer will be empty and the writer is gracefully closed.
-pub async fn transfer_collection<E: EventSender>(
+pub async fn transfer_collection<D: BaoCollection, E: EventSender>(
```
dignifiedquire, line 367: Hmm, I wonder if `BaoCollection` is a good name for this trait. Kinda hard to understand that that is the database.

iroh/src/node.rs
```diff
@@ -540,24 +547,31 @@ impl Future for Node {
...
+    inner: Arc<NodeInner<D>>,
 }
 
-impl RpcHandler {
+impl<D: BaoCollection> RpcHandler<D> {
     fn rt(&self) -> runtime::Handle {
         self.inner.rt.clone()
     }
 
+    fn concrete_db(&self) -> Option<Database> {
```
dignifiedquire, line 559: Can you add a doc comment to this?

iroh/src/node.rs
```diff
@@ -540,24 +547,31 @@ impl Future for Node {
...
+        let db: Box<dyn Any> = Box::new(self.inner.db.clone());
+        db.downcast_ref::<Database>().cloned()
```
dignifiedquire, line 561: when would this return `None`?

iroh-bytes/src/provider.rs
rklaehn: It is "A generic collection of blobs with precomputed outboards" - only the read part. Also not super happy with the name, especially because it collides with the other `Collection`. But Database is also not a good name since that encompasses both the read and write part and will contain more stuff.

BlobMap? BaoMap?

iroh/src/node.rs
rklaehn: My recollection slowly comes back to me. This is a bit of a hack. Some RPC calls only really make sense if you have not a generic BaoCollection but the concrete `Database`.

So some RPC calls have dummy impls unless you have a `Database`. Yeah, I know, not great. But the alternative would have been coming up with traits for all the things, even things that don't really make sense anymore.

iroh/src/node.rs
rklaehn: See above.

iroh-bytes/src/provider.rs
dignifiedquire: I like `BaoMap` it differentiates from `Collection` and makes it clear anything "bao" like is stored it.

iroh/src/node.rs
```diff
@@ -567,15 +581,21 @@ impl RpcHandler {
...
         // collections are always stored internally, so we take everything that is stored internally
         // and try to parse it as a collection
-        let items = self.inner.db.internal().filter_map(|(hash, collection)| {
-            Collection::from_bytes(&collection)
-                .ok()
-                .map(|collection| ListCollectionsResponse {
-                    hash,
-                    total_blobs_count: collection.blobs().len(),
-                    total_blobs_size: collection.total_blobs_size(),
+        let items = if let Some(db) = self.concrete_db() {
```
dignifiedquire, line 584: Why can't I list collections with a generic `BaoCollection`?

iroh/src/node.rs
```diff
@@ -586,11 +606,13 @@ impl RpcHandler {
...
         let (tx, rx) = mpsc::channel(1);
         let tx2 = tx.clone();
-        self.rt().main().spawn(async move {
-            if let Err(e) = self.inner.db.validate(tx).await {
-                tx2.send(ValidateProgress::Abort(e.into())).await.unwrap();
-            }
-        });
+        if let Some(db) = self.concrete_db() {
+            self.rt().main().spawn(async move {
+                if let Err(e) = db.validate(tx).await {
```
dignifiedquire, line 611: `validate` seems potentially very useful for different types of backends, not just the `Database`?

iroh/src/node.rs
rklaehn: You can list blobs, but what even is a collection in a blob store? Everything that has been explicitly added as a collection, or things that were added as blobs and just happen to deserialize as a collection? And what if you happen to change what a collection is using the custom collection stuff?

Once the traits are a bit more fleshed out, you would have to separately track the collections as the "gc roots". But this really doesn't have much to do with what BaoCollection (whatever we end up naming it) is supposed to be - just a thing that you can use to serve requests.

I guess I could have an additional trait containing stuff we also want to have generic but aren't quite as sure about yet. Maybe I will do that.

iroh/src/node.rs
rklaehn: Yeah, I think this will have to go into another trait or be implemented in terms of stuff that the database trait provides. I guess you would only need a way to list all blobs. But it would be a major refactor because the current validate uses sync IO and the new generic IO is by default async. I guess this is why I did this hack...

iroh/src/node.rs
dignifiedquire: I think the confusing piece here is that `Collections` will not be listed under `list_blobs` for `Database` but they would be for `BaoCollection`. So I would really like to find a more consistent approach if we add this abstraction level.

rklaehn: I changed it now so that you have an additional trait that adds blobs and roots fns. blobs lists all blobs including collections.

iroh/src/node.rs
rklaehn: Added another trait that has validate

rklaehn:
@dignifiedquire with the latest additional trait, the state is as such:

- validate and list_blobs are fully generic
- list_collections is generic regarding storage but still assumes that collections are `Collection`. Which I guess is fine since the "custom collections" PR is not yet merged
- only provide is still only implemented for `Database`

I don't see any way to solve the latter without introducing traits for generic write, but I would really like to do that in a different PR to not get into merge hell again.

dignifiedquire (approved):

PR #1133, opened 2023-06-26
feat: add MagicEndpoint to iroh-net

Frando:
This is a first pass at adding a `MagicEndpoint` that combines a magicsock and a quinn endpoint. The API is minimalistic so far and allows to both accept and create connections. I also added two utility functions to extract the ALPN and PeerId on incoming connections, and a maximally-simple example to test things.

An endpoint-like struct that can create and accept connections will be needed for peer-to-peer sockets as e.g. in the upcoming gossipswarm impl. (It could also work with functions only but would involve more code copying I think).

I mostly copied code around from the existing listen and dial functions. No tests or other code is yet converted over to actually use this. Sharing now to get feedback on the API.


iroh-net/src/endpoint.rs
```diff
@@ -0,0 +1,245 @@
...
+            let tls_client_config =
+                tls::make_client_config(&self.keypair, Some(peer_id), alpn_protocols, self.keylog)?;
+            let mut client_config = quinn::ClientConfig::new(Arc::new(tls_client_config));
+            let mut transport_config = quinn::TransportConfig::default();
+            transport_config.keep_alive_interval(Some(Duration::from_secs(1)));
+            client_config.transport_config(Arc::new(transport_config));
+            client_config
+        };
+
+        debug!("connecting to {}: (via {} - {:?})", peer_id, addr, addrs);
```
flub, line 152: i wonder if the `addrs` is really correct here?  it could have even more addresses than we have here in the `addrs` variable i think?

iroh-net/src/endpoint.rs
```diff
@@ -0,0 +1,245 @@
...
+        let client_config = {
+            let alpn_protocols = vec![alpn.to_vec()];
+            let tls_client_config =
+                tls::make_client_config(&self.keypair, Some(peer_id), alpn_protocols, self.keylog)?;
+            let mut client_config = quinn::ClientConfig::new(Arc::new(tls_client_config));
+            let mut transport_config = quinn::TransportConfig::default();
+            transport_config.keep_alive_interval(Some(Duration::from_secs(1)));
+            client_config.transport_config(Arc::new(transport_config));
+            client_config
+        };
```
flub, line 150: This block could also be done at `quinn::Endpoint` creation time with `Endpoint::set_default_client_config`.  In this code-layout this doesn't matter and may even be harder to follow, but I'm not sure if wrapping the endpoint is the right approach.  See my top-level comment.

flub (commented):
I've also been wondering about simplifying the use of the magicsocket.  But I was initially thinking of adding a bit of functionality to `Conn` (or maybe a new `MagicEndpoint, this is a pretty good name) that would have a constructor method to give you a `quinn::Endpoint` which is configured to use this magicsocket.

That way you'd have to call `.add_know_addrs()` and `.get_mapping_addr()` before you can call `.connect()` on a `quinn::Endpoint` making the relation between both sockets a bit more explicit.  But it would not put us in the game of having to wrap all `quinn::Endpoint` APIs.

iroh-net/src/endpoint.rs
Frando: Very much possible, I copied this from `iroh-net/src/client.rs:132` and did not check it at all

iroh-net/src/endpoint.rs
Frando: Can it really? The crypto config contains the `peer_id` that you're connecting to.

iroh-net/src/defaults.rs
```diff
@@ -0,0 +1,30 @@
...
+
+pub fn default_derp_region() -> DerpRegion {
+    // The default derper run by number0.
+    let default_n0_derp = DerpNode {
+        name: "default-1".into(),
+        region_id: 1,
+        host_name: "derp.iroh.network".into(),
+        stun_only: false,
+        stun_port: 3478,
+        ipv4: UseIpv4::Some("35.175.99.113".parse().unwrap()),
```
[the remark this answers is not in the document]
Frando: Ack to the first comment. Wrt to the second, I don't know ;-) Copied it from `iroh/config.rs`.

iroh-net/src/endpoint.rs
flub: oh, i missed that :(

Frando:
> I've also been wondering about simplifying the use of the magicsocket. But I was initially thinking of adding a bit of functionality to `Conn` (or maybe a new `MagicEndpoint`, this is a pretty good name) that would have a method to give you a new `quinn::Endpoint` which is configured to use this magicsocket.

Aha, yeah that would work as well. Somehow, the notion of `connect(peer_id)` is quite appealing to me though.

> That way you'd have to call `.add_know_addrs()` and `.get_mapping_addr()` before you can call `.connect()` on a `quinn::Endpoint` making the relation between both ~sockets~ endpoints a bit more explicit. But it would not put us in the game of having to wrap all `quinn::Endpoint` APIs.

That is true. However, the API surface of `quinn::Endpoint` is really small when omitting the constructors which are taken care of in either case. Method-wise, it's just:
*  `accept`, `connect`: already implemented here
* `local_addr` is mostly useless unless coupled with `get_mapping_addr` I think? So the wrapping would make sense here IMO
* * `set_server_config` and `rebind`: not sure if exposing those in the magicsock setup is needed / makes sense
* `close` and `wait_idle`: those should be exposed

iroh-net/src/endpoint.rs
```diff
@@ -0,0 +1,439 @@
...
+        let connect = self
+            .endpoint
+            .connect_with(client_config, addr, "localhost")?;
+
+        let connection = connect.await.context("failed connecting to provider")?;
+
+        Ok(connection)
+    }
+
+    pub async fn add_known_addrs(
```
flub, line 181: doc comment? (isn't that a lint?)

flub (dismissed):

iroh-net/examples/magic.rs
```diff
@@ -0,0 +1,134 @@
...
+use clap::Parser;
+use ed25519_dalek::SigningKey as SecretKey;
+use iroh_net::{
+    defaults::default_derp_map,
+    endpoint::{accept_conn, MagicEndpoint},
+    hp::derp::{DerpMap, UseIpv4, UseIpv6},
+    tls::{Keypair, PeerId},
+};
+
+const EXAMPLE_ALPN: &[u8] = b"n0/iroh/examples/magic/0";
```
dignifiedquire, line 12: could be cool to make the alpn be also a cli argument

iroh-net/examples/magic.rs
```diff
@@ -0,0 +1,134 @@
...
+            };
+            DerpMap::default_from_node(url, 3478, addr.port(), ip4, ip6)
+        }
+    };
+
+    let endpoint = MagicEndpoint::bind(
+        keypair,
+        args.bind_address,
+        vec![EXAMPLE_ALPN.to_vec()],
+        None,
```
dignifiedquire, line 65: these args get quite unwieldy, probably should be a builder struct

```rust
let endpoint = MagicEndpoint::builder()
  .keypair(keypair)
  .alpns(vec![EXAMPLE_ALPN.to_vec()])
  .derp_map(derp_map)
  .bind(args.bind_address).await?;
```

iroh-net/src/endpoint.rs
dignifiedquire: the lint is currently disabled, there is a lot undocumented atm :(

flub: ah, let's just keep adding issues to the milestone then :) https://github.com/n0-computer/iroh/issues/1139

Frando:
Thanks for the reviews! I added a `MagicEndpointBuilder` and also more docs.

iroh-net/src/defaults.rs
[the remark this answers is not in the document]
dignifiedquire: > More interestingly though: are we fine with hardcoding an IP address here?

It is not great, but okay for the moment I would say

dignifiedquire:
@Frando would be great if you could do the integration/usage of it in `iroh` itself, so we can see how and if it works there as needed for the current use case.

iroh-net/src/endpoint.rs
```diff
@@ -22,6 +22,78 @@ use crate::{
...
+impl MagicEndpointBuilder {
+    /// Set a keypair to authenticate with other peers.
+    ///
+    /// This keypair's public key will be the [PeerId] of this endpoint.
+    ///
+    /// If not set, a new keypair will be generated.
+    pub fn keypair(self, keypair: Keypair) -> Self {
+        Self {
+            keypair: Some(keypair),
+            ..self
```
dignifiedquire, line 44: interesting way to write this, any reason to use this over?

```rust
self.keypair.replace(keypair);
self
```

iroh-net/src/endpoint.rs
```diff
@@ -22,6 +22,78 @@ use crate::{
...
+            transport_config: Some(transport_config),
+            ..self
+        }
+    }
+
+    /// Bind the magic endpoint on the specified socket address.
+    pub async fn bind(self, bind_addr: SocketAddr) -> anyhow::Result<MagicEndpoint> {
+        let keypair = self.keypair.unwrap_or_else(|| Keypair::generate());
+        let tls_server_config =
+            tls::make_server_config(&keypair, self.alpn_protocols, self.keylog)?;
```
dignifiedquire, line 83: does this error if `alpn_protocols` is empty?

Frando:
> @Frando would be great if you could do the integration/usage of it in `iroh` itself, so we can see how and if it works there as needed for the current use case.

Allright, I took a stab at this, the latest commit uses `MagicEndpoint` throughout `iroh`.

iroh-net/src/endpoint.rs
Frando: no reason, changed it

dignifiedquire:
> Allright, I took a stab at this, the latest commit uses `MagicEndpoint` throughout `iroh`.

Very nice and clean :) really like having this abstraction.

dignifiedquire:
@Frando the last integration I would love to see is replacing the structure in `hp/magicsock/conn.rs` tests

Frando:
> @Frando the last integration I would love to see is replacing the structure in `hp/magicsock/conn.rs` tests

I just had a look at this. The tests access more internals of the magic socket than the current API exposes. Most seems to me for validation of those internals. 
* they access the `tracked_endpoints()` of a magic socket to check if some peer id is in there.
* they also update the network map completely whereas the currently-exposed API on the `MagicEndpoint` only allows to `add_known_addr(peer_id, socket_addrs)`
I haven't worked with the internals of the magic socket much yet so not sure if those are needed outside of testing. So, should I ..
* expose `tracked_endpoints` and `set_network_map` on the `MagicEndpoint` or
* make the magic socket `Conn` on the MagicEndpint `pub(crate)` and keep tests mostly as they are
* restructure some of the tests to use less internals and work more with `PeerId`s only (likely possible, but not sure if the current setup is intended)

?

Frando:
And CI passed on the last commit, including CLI tests! Only netsim failures, but seems unrelated (github bucket something something)

dignifiedquire:
I think a first pass just exposing things that are needed using `cfg(test)` is fine. I can take a look in a second round if it makes to restructure the tests.

iroh-net/src/client.rs
```diff
@@ -5,129 +5,37 @@ use std::{
...
-    endpoint.set_default_client_config(client_config);
-    Ok((endpoint, conn))
-}
-
-/// Establishes a QUIC connection to the provided peer.
+/// The PeerId and the ALPN protocol are required. If you happen to know dialable addresses of
+/// the remote endpoint, they can be specified and will be added to the endpoint's peer map.
+/// If no addresses are specified, the endpoint will try to dial the peer through the
+/// configured DERP servers.
 pub async fn dial_peer(
```
dignifiedquire, line 21: I think it would be nice to just make this a helper function on `MagicEndpoint`, eg `MagicEndpoint::dial_peer(..) -> Self`. That way we can reduce the API surface a little bit more, and all interactions just go through `MagicEndpoint`

iroh-net/src/endpoint.rs
```diff
@@ -0,0 +1,542 @@
+use std::{
```
dignifiedquire, line 1: Can we call this `magic_endpoint.rs` please? I am always confused if the structs don't match the filename 😅

iroh-net/src/endpoint.rs
```diff
@@ -0,0 +1,542 @@
...
+    }
+
+    /// Create a quinn endpoint backed by a magicsock.
+    ///
+    /// The *bind_addr* is the address that should be bound locally.  Even though this is an
+    /// outgoing connection a socket must be bound and this is explicit.  The main choice to
+    /// make here is the address family: IPv4 or IPv6.  Otherwise you normally bind to the
+    /// `UNSPECIFIED` address on port `0` thus allowing the kernel to do the right thing.
+    ///
+    /// If *peer_id* is present it will verify during the TLS connection setup that the remote
```
dignifiedquire, line 159: not seeing `peer_id` as an argument

iroh-net/src/endpoint.rs
```diff
@@ -0,0 +1,542 @@
...
+            client_config
+        };
+
+        debug!(
+            "connecting to {}: (via {} - {:?})",
+            peer_id, addr, known_addrs
+        );
+        let connect = self
+            .endpoint
+            .connect_with(client_config, addr, "localhost")?;
```
dignifiedquire, line 273: Can you add a todo, to replace `localhost` with sth that makes more sense? (this is old, and okay for now, but should be checked/fixed eventually)

dignifiedquire (dismissed):
some smaller nits, other than that this is looking good :)

iroh-net/src/client.rs
Frando: Yup, then we can remove the client module alltogether. The other function is only used in a single place (RPC client) and has not much to do with iroh_net as its just creating a quinn endpoint, so I'd just move that to where the RPC client is created for now.

iroh-net/src/endpoint.rs
Frando: Ah right, those were copied over from old docs. As the function is pub(crate) I'd move the docs to the builder.

Frando:
I addressed the last comments and rebased onto main. Let's see what CI says.

One thing I'm not super sure about is the netmap managment. But I think this can be refined later.

iroh-net/examples/magic.rs
```diff
@@ -0,0 +1,137 @@
...
+        Some(url) => {
+            // TODO: This should be done by the DERP client.
+            let derp_port = match url.port() {
+                Some(port) => port,
+                None => match url.scheme() {
+                    "http" => 80,
+                    "https" => 443,
+                    _ => anyhow::bail!(
+                        "Invalid scheme in DERP URL, only http: and https: schemes are supported."
+                    ),
```
dignifiedquire, line 64: I have seen this logic in your other PR, maybe we can add a helper somewhere to keep this consistent?

iroh-net/src/hp/magicsock/conn.rs
```diff
@@ -2649,7 +2658,7 @@ mod tests {
                     nodes: vec![DerpNode {
                         name: "t1".into(),
                         region_id: 1,
-                        host_name: "http://test-node.invalid".parse().unwrap(),
+                        host_name: "https://test-node.invalid".parse().unwrap(),
```
dignifiedquire, line 2661: does this need to be `https`?

iroh-net/src/lib.rs
```diff
@@ -1,7 +1,10 @@
 #![recursion_limit = "256"]
 
-pub mod client;
+pub mod defaults;
 pub mod hp;
+pub mod magic_endpoint;
```
dignifiedquire, line 5: can we make this private, if we reexport the `MagicEndpoint`?

iroh-net/src/magic_endpoint.rs
```diff
@@ -0,0 +1,562 @@
...
+    /// [Self::builder]. See the methods on the builder for documentation of the parameters.
+    async fn bind(
+        keypair: Keypair,
+        bind_addr: SocketAddr,
+        server_config: Option<quinn::ServerConfig>,
+        derp_map: Option<DerpMap>,
+        callbacks: Option<Callbacks>,
+        keylog: bool,
+    ) -> anyhow::Result<Self> {
+        // let (endpoints_update_s, endpoints_update_r) = flume::bounded(1);
```
dignifiedquire, line 195: still needed?

iroh-net/src/magic_endpoint.rs
```diff
@@ -0,0 +1,562 @@
...
+        Ok(data) => match data.protocol {
+            Some(protocol) => std::string::String::from_utf8(protocol).map_err(Into::into),
+            None => anyhow::bail!("no ALPN protocol available"),
+        },
+        Err(_) => anyhow::bail!("unknown handshake type"),
+    }
+}
+
+/// Extract the [`PeerId`] from the peer's TLS certificate.
+pub async fn get_peer_id(connection: &quinn::Connection) -> anyhow::Result<PeerId> {
```
dignifiedquire, line 391: does this need to be public?

iroh-net/src/magic_endpoint.rs
```diff
@@ -0,0 +1,562 @@
...
+    mut conn: quinn::Connecting,
+) -> anyhow::Result<(PeerId, String, quinn::Connection)> {
+    let alpn = get_alpn(&mut conn).await?;
+    let conn = conn.await?;
+    let peer_id = get_peer_id(&conn).await?;
+    Ok((peer_id, alpn, conn))
+}
+
+/// Extract the ALPN protocol from the peer's TLS certificate.
+pub async fn get_alpn(connecting: &mut quinn::Connecting) -> anyhow::Result<String> {
```
dignifiedquire, line 379: does this need to be public?

dignifiedquire (dismissed):
some nits, other than that, lgtm

iroh-net/examples/magic.rs
Frando: This would be obsoleted with https://github.com/n0-computer/iroh/pull/1143 I think because then we don't have to the port parsing here and just pass the URL through

iroh-net/src/lib.rs
Frando: There's also the `MagicEndpointBuilder` struct and `accept_conn` functions in there (and possible `get_alpn` and `get_peer_id`, see below). Not sure if they should all be at the top level?
I think I'd defer this to later, currently pretty much everything in `iroh-net` is public anyway - and once things stabilize a bit we might want to change this overall and actually define the public API surface for `iroh-net`?

iroh-net/src/magic_endpoint.rs
dignifiedquire: lets leave it as is for now then

iroh-net/src/magic_endpoint.rs
```diff
@@ -0,0 +1,561 @@
...
+        Ok(Self {
+            keypair: Arc::new(keypair),
+            conn,
+            endpoint,
+            netmap: Arc::new(Mutex::new(NetworkMap { peers: vec![] })),
+            keylog,
+        })
+    }
+
+    /// Accept a connection on the socket.
```
flub, line 225:
```suggestion
    /// Accept an incoming connection on the socket.
```

iroh-net/src/magic_endpoint.rs
```diff
@@ -0,0 +1,561 @@
...
+        &self.conn
+    }
+    #[cfg(test)]
+    pub(crate) fn endpoint(&self) -> &quinn::Endpoint {
+        &self.endpoint
+    }
+}
+
+/// Accept an incoming connection and extract the client-provided [`PeerId`] and ALPN protocol.
+pub async fn accept_conn(
```
flub, line 368: I'd be tempted to experiment with an extension trait or a wrapper for `Connecting` (and I guess `Accept`).  But we can do that later.

iroh-net/src/hp/magicsock/conn.rs
```diff
@@ -2826,7 +2820,7 @@ mod tests {
         })
     }
 
-    fn setup_logging() {
+    pub fn setup_logging() {
```
flub, line 2823: I kind of wish you copied in `iroh-bytes/src/test_utils.rs` to `iroh-net/src/test_utils.rs` and used it's `setup_logging` call for your functions instead of making this pub...  But I tried to clean this up earlier and utterly failed failed.  For new `#[tokio::test]` tests it should just work though, since they're single-threaded runtimes.  Just don't try change existing tests.

But if you don't want to do that now that's fine, it's my fault for not managing to clean this up before.

flub (approved):
Nice!  Most of my comments are about the public docs from a user point of view.  But nothing is blocking.

iroh-net/src/magic_endpoint.rs
```diff
@@ -0,0 +1,561 @@
...
+        self.callbacks.on_net_info = Some(on_net_info);
+        self
+    }
+
+    /// Bind the magic endpoint on the specified socket address.
+    ///
+    /// The *bind_addr* is the address that should be bound locally.  Even though this is an
+    /// outgoing connection a socket must be bound and this is explicit.  The main choice to
+    /// make here is the address family: IPv4 or IPv6.  Otherwise you normally bind to the
+    /// `UNSPECIFIED` address on port `0` thus allowing the kernel to do the right thing.
```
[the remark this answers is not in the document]
Frando: Good question, this is taken over from the old `dial_peer` functions and the code in `iroh/src/node.rs`
Now that I re-read the code, the `bind_addr` IP address & family  is not even used, only its port is passed to the `magicsock::Conn`. So - it would actually be correct to remove the `bind_addr: SocketAddr` argument and instead add `bind_port: u16`?
@dignifiedquire can you confirm this? Then I'd go ahead and replace `bind_addr` with `bind_port`.

iroh-net/src/magic_endpoint.rs
```diff
@@ -0,0 +1,561 @@
...
+    ///
+    /// Returns a tuple of the IPv4 and the optional IPv6 address.
+    pub fn local_addr(&self) -> anyhow::Result<(SocketAddr, Option<SocketAddr>)> {
+        self.conn.local_addr()
+    }
+
+    /// Get the local endpoints on which the underlying magic socket is listening.
+    ///
+    /// This returns information on the network state of the endpoints in addition to
+    /// their addresses.
```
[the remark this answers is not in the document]
Frando: Good question! I was thinking the `local_endpoints` would also contain addresses obtained via STUN. Is this correct?

If this is the case, then maybe something like this?

```
Get the local and discovered endpoints on which the underlying magic socket is reachable.

This list contains both the locally-bound addresses and the endpoint's publicly-reachable addresses, if they could be discovered through STUN or port mapping.
```

iroh-net/src/magic_endpoint.rs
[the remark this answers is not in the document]
dignifiedquire: My answer is two fold, the current state and the ideal state.

- Current State: The address is effectively ignored and only the port is used, as we bind to IPv4 and IPv6 seperately.
- Should State: Allow to provide sth like this 
```rust
enum BindAddress {
  /// Bind on this port, both ipv4 and ipv6 addr, using `0.0.0.0` and `[::]`
  Port(u16),
  Ips(Ips),
}
enum Ips {
  // Only Ipv4
  Ipv4(SocketV4Addr),
  // Only Ipv6
  Ipv6(SocketV6Addr),
  Both {
    ipv4: SocketV4Addr,
    ipv6: SocketV4Addr,
}

// The `BindAddress::Port` options is equivalent to 
BindAddress::Ips(Ips::Both {
  ipv4: SocketV4Addr::new(Ipv4Addr::UNSPECIFIED, port),
  ipv6: SocketV6Addr::new(Ipv6Addr::UNSPECIFIED, port),
});

// so could probably change this to just have `Ips` and this
impl Ips {
  pub fn from_port(port: u16) -> Self {
    Ips::Both {
      ipv4: SocketV4Addr::new(Ipv4Addr::UNSPECIFIED, port),
      ipv6: SocketV6Addr::new(Ipv6Addr::UNSPECIFIED, port),
    }
  }
}
```

iroh-net/src/magic_endpoint.rs
Frando: Yup I thought about that too. However we'd either have to use `async_trait` or write the futures by hand. I'd defer this to post-merge I think.

iroh-net/src/hp/magicsock/conn.rs
Frando: Yes this sounds good to me too. I also want to expose `run_derp_and_stun` in a test_util module also to other crates (i.e. iroh-gossip) that use iroh-net and need to test things with a local derper. 
However I'd prefer to do this after the merge in a new PR.

iroh-net/src/magic_endpoint.rs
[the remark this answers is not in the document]
dignifiedquire: Yes, the endpoints are all that we have discovered about ourselves

iroh-net/src/magic_endpoint.rs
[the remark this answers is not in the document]
flub: Yes, that makes more sense.  Maybe `Ips` should even be a vec, I think there's nothing wrong with binding to a bunch of specific addresses, e.g. you're on say 20 subnets but only want to bind to the private ones.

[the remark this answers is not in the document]
flub: So I'd be mildly in favour of for now changing this to `u16` and describe that this will change in the future to allow binding to specific socket addresses.

iroh-net/src/magic_endpoint.rs
```diff
@@ -0,0 +1,561 @@
...
+impl MagicEndpoint {
+    /// Build a MagicEndpoint
+    pub fn builder() -> MagicEndpointBuilder {
+        MagicEndpointBuilder::default()
+    }
+
+    /// Connect to a remote endpoint, creating an endpoint on the fly.
+    ///
+    /// The PeerId and the ALPN protocol are required. If you happen to know dialable addresses of
+    /// the remote endpoint, they can be specified and will be added to the endpoint's peer map.
```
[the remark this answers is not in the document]
Frando: I removed the term and took over the description you proposed in `connect`

iroh-net/src/magic_endpoint.rs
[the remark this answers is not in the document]
dignifiedquire: > e.g. you're on say 20 subnets but only want to bind to the private ones

That would be a lot of internal changes, as we only do at most two sockets internally. I would argue, this should be managed by some network configuration/proxy infront if you have multiple private binds.

PR #1137, opened 2023-06-27
fix(netcheck): Make ICMP ping optional

flub:
On Linux you don't have permission to send ICMP pings by default, so
creating the pinger will fail.  This should not make all of netcheck
fail.

### DIFF
--- a/iroh-net/src/hp/netcheck.rs
+++ b/iroh-net/src/hp/netcheck.rs
@@ -566,7 +566,13 @@ impl ReportState {
         };
 
         let pinger = if self.plan.has_https_probes() {
-            Some(Pinger::new().await.context("failed to create pinger")?)
+            match Pinger::new().await {
+                Ok(pinger) => Some(pinger),
+                Err(err) => {
+                    debug!("failed to create pinger: {err:#}");
+                    None
+                }
+            }
         } else {
             None
         };
--- a/iroh-net/src/hp/netcheck/probe.rs
+++ b/iroh-net/src/hp/netcheck/probe.rs
@@ -101,6 +101,7 @@ impl ProbePlan {
     pub fn has_https_probes(&self) -> bool {
         self.keys().any(|k| k.ends_with("https"))
     }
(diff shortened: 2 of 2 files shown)


iroh-net/src/hp/netcheck.rs
```diff
@@ -566,7 +566,13 @@ impl ReportState {
         };
 
         let pinger = if self.plan.has_https_probes() {
-            Some(Pinger::new().await.context("failed to create pinger")?)
+            match Pinger::new().await {
+                Ok(pinger) => Some(pinger),
+                Err(err) => {
+                    debug!("failed to create pinger: {err:#}");
```
dignifiedquire, line 572: Maybe remove the https probes from the plan in this case?

flub: hum, no i think they should stay in since they will keep working.  I'm not really sure why these pingers are triggered by the https probes in the plan, the tailscale code only uses the https and pinger probes if the udp stun fails.  but i decided that changing (or even investigating this) should not be mixed into this PR.

but good call, should create an issue not to forget: https://github.com/n0-computer/iroh/issues/1138

dignifiedquire: Maybe the pinger should be created lazily? The reason is that I went and finished a todo that they had, where the https probes are actually integrated into the plans, instead of being added manually after the fact.

flub:
Can we ignore this failure?  it's a flaky one that I'd like to address separately: https://github.com/n0-computer/iroh/issues/1138

iroh-net/src/hp/netcheck.rs
flub: Yes, can we do this as part of #1138? In either case the fix of making failure to create the pinger non-fatal is needed.

dignifiedquire: sure

dignifiedquire (approved):

PR #1143, opened 2023-06-28
change: use DERP port from host_name URL

Frando:
Now that `DerpNode::host_name` is an URL, the DERP port can be part of that URL, so we don't need a separate `derp_port` field anymore. 

This makes setting custom DERP config or accepting a DERP server from the command line much simpler, because you need only a single URL string and not multiple options.

It looks a bit weird in the test code because oftenly those use `http://derp.invalid:{port}` now. I think we could just put the full IP in the address also and not set `UseIpv4::Some()`.

### DIFF
--- a/iroh-net/examples/magic.rs
+++ b/iroh-net/examples/magic.rs
@@ -52,20 +52,7 @@ async fn main() -> anyhow::Result<()> {
 
     let derp_map = match args.derp_url {
         None => default_derp_map(),
-        Some(url) => {
-            // TODO: This should be done by the DERP client.
-            let derp_port = match url.port() {
-                Some(port) => port,
-                None => match url.scheme() {
-                    "http" => 80,
-                    "https" => 443,
-                    _ => anyhow::bail!(
-                        "Invalid scheme in DERP URL, only http: and https: schemes are supported."
-                    ),
-                },
-            };
-            DerpMap::default_from_node(url, 3478, derp_port, UseIpv4::None, UseIpv6::None)
-        }
+        Some(url) => DerpMap::default_from_node(url, 3478, UseIpv4::None, UseIpv6::None),
     };
 
     let endpoint = MagicEndpoint::builder()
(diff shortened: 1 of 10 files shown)


dignifiedquire (approved):
yesss, much better

dignifiedquire:
looks like some tests are failing though

Frando:
The remaining fail is in ǹetsim` - never worked with those, how do I debug this?

Arqu:
So looking at both, `netsim` has a [derpmap provided](https://github.com/n0-computer/chuck/blob/main/fixtures/derp.config.toml). Unsure where the issue comes from, but in the logs, seems like both the server and client fail handshakes (presumably with the derper). My assumption here is that now that the `derp_port` is obsolete it tries to do just `http` but some port got misplaced and now it fails to establish a connection. 

Note that the test that is failing is where the provide side is behind a NAT, so some confusion might be happening due to the setup. @dignifiedquire already dealt with a similar bug before.

TLDR; either it's a config issue and you might be able to help me understand where and I can update netsim or we re-introduced a bug we closed in https://github.com/n0-computer/iroh/pull/1128

Frando:
Ah, so the netsim config would need to change to this with/after this PR:
https://github.com/n0-computer/chuck/compare/main...Frando:chuck:patch-1

Frando:
Something else: Now that `host_name` is a URL with optional port, `host_name` is not really the fitting term anymore, is it? Maybe we should change it to `derp_url` or `derp_server_url` or something similar right away?

dignifiedquire:
> `host_name` is not really the fitting term anymore

Maybe just `host` or `url`?

dignifiedquire:
merge conflict

Frando:
Rebased and changed the field name to `url`.

I was undecided about naming the field `host` vs `url` because neither is fully correct, but URL seemed "more correct" as we support 3 parts of a typical URL (scheme, host, port) and just not any path parts, but more than just a hostname. But can ofc change again if something else is deemed better.

Arqu:
I'm happy with where it's at. Once approved, we just need to bump the netsim derp map config too.

dignifiedquire (approved):

PR #1145, opened 2023-06-28
ref(tests): Clean up various bits in tests

flub:
- Bump testdir to 0.8, we use multiple testdir!() calls inside a
  single test while expecting to get the same directory.  This was not
  the case but now in 0.8 it is.

- When transferring folders do not use testdir!() directly but a
  subdirectory.  Using testdir!() directly ends up including e.g. the
  iroh_data_dir which should not really be included.

- Simplify make_provider to not require home, all callers used
  testdir!() for that anyway.

- Simplify match_provide_output since it did not use Input anyway.

- Fix test_provide_get_loop to not go put the path inside
  $CARGO_MANIFEST_DIR/tests/fixtures since we do not store any test
  fixtures.  We create all of them on the fly and put them inside
  testdir!().  I think we were always testing this with an empty
  directory before?

- Improve assert_matches_line to not hide the errors from the reader.
  This is the error that is propagated from duct in case the provider
  fails, hiding this is rather confusing.  Improves the error messages
  emitted there too by e.g. showing the failing regex instead of just
  say "EOF, *shrug*".

### DIFF
--- a/iroh/tests/cli.rs
+++ b/iroh/tests/cli.rs
@@ -96,7 +96,8 @@ fn cli_provide_one_file_single() -> Result<()> {
 
 #[test]
 fn cli_provide_folder() -> Result<()> {
-    let dir = testdir!();
+    let dir = testdir!().join("src");
+    std::fs::create_dir(&dir)?;
     let foo_path = dir.join("foo");
     let bar_path = dir.join("bar");
     make_rand_file(1000, &foo_path)?;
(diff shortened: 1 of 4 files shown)


iroh/tests/cli.rs
```diff
@@ -612,22 +601,28 @@ where
...
+                        matches += 1;
+                    } else {
+                        break;
+                    }
+                }
+                Some(Err(err)) => {
+                    panic!("Error from reader: {err:#}");
+                }
+                None => {
+                    panic!("All lines read but no match found for /{rx}/");
```
dignifiedquire, line 624: Any chance we can print all output here, for easier debugging?

flub: the output read so far is already printed above, and getting here means there is no more output afaik so nothing more to print.

dignifiedquire: ah okay, great

dignifiedquire (approved):

PR #1146, opened 2023-06-28
fix(iroh): error when path does not exist

### LINKED ISSUE #1068: iroh provide gives no feedback if path provided is invalid
So this one bit me a few times. If you `iroh provide some/nonexisting/path.x`, iroh will happily start and do nothing. Given that a path was provided it should at least log an error.

Example:
~~Running iroh provide something with a bad derpmap results in the process continuously trying to reach the derper and never finishes creating the collection & ticket.~~

~~Edit: seems to hang in general...~~

dignifiedquire:
Fixes #1068

### DIFF
--- a/iroh/src/commands/provide.rs
+++ b/iroh/src/commands/provide.rs
@@ -5,7 +5,7 @@ use std::{
     str::FromStr,
 };
 
-use anyhow::{Context, Result};
+use anyhow::{ensure, Context, Result};
 use iroh_bytes::{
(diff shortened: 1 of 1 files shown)


flub:
my guess is your test failures are affected by point 5 of https://github.com/n0-computer/iroh/pull/1145

flub:
or not?  some appear to be passing?

flub (approved):

PR #1148, opened 2023-06-28
refactor(iroh-net): refactor derp region handling

dignifiedquire:
See commits for details

- refactor(iroh-net): make region ids always be u16
- refactor: use newtype to avoid derp magic ips
- delete unused file


iroh-net/src/hp/magicsock/conn.rs
```diff
@@ -2293,6 +2299,61 @@ fn log_endpoint_change(endpoints: &[cfg::Endpoint]) {
     });
 }
 
+/// Addresses to which to which we can send. This is either a UDP or a derp address.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum SendAddr {
+    /// UDP, the ip addr.
+    Udp(SocketAddr),
+    /// Derp, region id.
+    Derp(u16),
```
flub, line 2308: I'm pretty tempted to also newtype the `u16` into a `DerpRegion`, not really important here but for the functions that take it as a parameter or structs that store the region id that would probably also make things nicer.

flub (approved):

iroh-net/src/hp/magicsock/conn.rs
dignifiedquire: I thought about it, but wanted to avoid too many types for the moment 😅

iroh-net/src/hp/magicsock/conn.rs
```diff
@@ -2293,6 +2299,61 @@ fn log_endpoint_change(endpoints: &[cfg::Endpoint]) {
...
+}
+
+impl SendAddr {
+    pub fn is_derp(&self) -> bool {
+        matches!(self, Self::Derp(_))
+    }
+
+    pub fn as_udp(&self) -> Option<&SocketAddr> {
```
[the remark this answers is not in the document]
dignifiedquire: yeah, I tend to return references even for `Copy` times 🤷 not sure if that is good or bad

PR #1149, opened 2023-06-29
feat: iroh-gossip

Frando:
## Description

This PR adds  `iroh-gossip`. See https://github.com/n0-computer/iroh-sync/pull/1 for the prehistory.

iroh-gossip exposes a pubsub-like API to broadcast messages on topics. Peers join a topic by connecting to at least one other peer. They can then broadcast messages to all other peers who joined a topic, and subscribe to messages. The protocol is based on [HyParView](https://asc.di.fct.unl.pt/~jleitao/pdf/dsn07-leitao.pdf), which gives us an overlay network of all nodes that joined a topic; and [PlumTree](https://asc.di.fct.unl.pt/~jleitao/pdf/srds07-leitao.pdf) to create an *epidemic broadcast tree* for reliably broadcasting messages to all online peers. 

The new crate is structured as follows:

* The protocol implementation (mod `proto`) as an IO-less state machine. It exposes a single method to handle `InEvent`s and returns an iterator of `OutEvent`s which must be handled by the implementor (register timers, send messages, emit events). The implementation is generic over a `PeerAddress` trait, which should be an identifier for a peer. In our case, it will be a `PeerId`.
  * All protocol messages are namespaced by a `TopicId`, a 32 byte identifier. Topics are seperate swarms and broadcast scopes. The HyParView and PlumTree algorithms both work in the scope of a single topic. Thus, joining multiple topics increases the number of open connections to peers and the size of the local routing table.
  * The high level protocol lives in `state::State`. It mostly dispatches messages based on the topic ID into the matching `topic::State`. This is the actual protocol state for a topic. It, again, mostly dispatches messages onto the actual protocols: `hyparview::State` and `plumtree::State`.
  * The implementations of these two modules quite strictly follows their papers [1](https://asc.di.fct.unl.pt/~jleitao/pdf/dsn07-leitao.pdf) and [2](https://asc.di.fct.unl.pt/~jleitao/pdf/srds07-leitao.pdf). If you find any deviations that are not documented, please report, this is likely a bug.

* The `net` module provides networking for the IO-less protocol impl, through the `MagicEndpoint` from iroh-net, with a (I think) quite straightforward actor implementation.

The `net` module has a smoke test, and the PR also includes a `chat` example that adds signed messages on the trustless base layer.

What this mostly needs is, I think, more testing and validation of the protocol implementation. For this to be practical, I think we should formulate some base usecase scenarios.

The `proto` module has a very simple simulator and some tests that can take environment variables to increase the cluster size. They passed with 10.000 nodes. They also report some stats. The numbers look OK-ish, but I'd love to compare them properly with either other implementations or estimations from the original paper or other resoures. However, those are hard to find.

## Notes & open questions

* Right now the `plumtree` module builds an unbounded cache of all broadcast messages. This will have to change. The simplest might be to move to an LRU cache for now. In the future, we might want to expose a method for the module to request expired messages from a backing storage or so. (The module needs to have a cache of messages to reply for requests by other peers. However these will only come within a reasonable amount of time after message publication / reception.) *this is now implemented with an expiring cache*
* Same for the list of received message IDs - they're also unbounded atm *this is now implemented with an expiring cache*
* We should validate the numbers - either by more simulation tests or by actually comparing them to the paper. However I couldn't find any simulation code from the papers.

## Change checklist

- [x] Self-review.
- [ ]Tests if relevant. *some tests are there, some more might be good*


iroh-gossip/examples/chat.rs
```diff
@@ -0,0 +1,229 @@
...
+        let message =
+            SignedMessage::sign_and_encode(endpoint.keypair(), &Message::AboutMe { name })?;
+        gossip.broadcast(topic, message).await?;
+    }
+
+    // subscribe and print loop
+    tokio::spawn(subscribe_loop(gossip.clone(), topic, names));
+
+    // spawn an input thread that reads stdin
+    // not using tokio here because they recommend this for "technical reasons"
```
dignifiedquire, line 115: 🤣

iroh-gossip/README.md
```diff
@@ -0,0 +1,19 @@
+# iroh-gossip
+
```
dignifiedquire, line 2: please add a short description here

iroh-gossip/src/proto/mod.rs
```diff
@@ -0,0 +1,29 @@
+use std::{fmt, hash::Hash};
```
dignifiedquire, line 1: please rename to `src/proto.rs`

iroh-gossip/src/net/mod.rs
```diff
@@ -0,0 +1,787 @@
+use std::{collections::HashMap, fmt, sync::Arc, time::Instant};
```
dignifiedquire, line 1: please rename to `src/net.rs`

iroh-gossip/src/proto/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+pub mod hyparview;
+pub mod plumtree;
+pub mod state;
+pub mod topic;
+pub mod util;
+
+pub use state::{InEvent, Message, OutEvent, State, Timer, TopicId};
+pub use topic::{Command, Config, Event, IO};
+
+pub trait PeerAddress: Hash + Eq + Copy + fmt::Debug + Serialize + DeserializeOwned {}
```
dignifiedquire, line 14: docs

iroh-gossip/src/proto/state.rs
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    pub fn handle(
+        &mut self,
+        event: InEvent<PA>,
+        now: Instant,
+    ) -> impl Iterator<Item = OutEvent<PA>> + '_ {
+        let event: InEventMapped<PA> = event.into();
+
+        // todo: add command to leave a topic
```
dignifiedquire, line 172: we should probably do this?

iroh-gossip/src/proto/state.rs
```diff
@@ -0,0 +1,253 @@
...
+                        self.config.clone(),
+                        self.rng.clone(),
+                    ));
+                }
+
+                // pass the event to the state handler
+                if let Some(state) = self.states.get_mut(&topic) {
+                    let out = state.handle(event, now);
+                    for event in out {
+                        handle_out_event(topic, event, &mut self.conns, &mut self.outbox);
```
dignifiedquire, line 205: Can any of these block/take longer? if so should this be parallelized potentially?

iroh-gossip/src/proto/state.rs
```diff
@@ -0,0 +1,253 @@
...
+                for (topic, state) in self.states.iter_mut() {
+                    let out = state.handle(event.clone(), now);
+                    for event in out {
+                        handle_out_event(*topic, event, &mut self.conns, &mut self.outbox);
+                    }
+                }
+            }
+        }
+
+        self.outbox.drain(..)
```
dignifiedquire, line 223: what exactly is the point of this?

iroh-gossip/src/proto/util.rs
```diff
@@ -0,0 +1,182 @@
+use rand::{
```
dignifiedquire, line 1: missing unit tests, ideally some proptests for the pieces in this file

iroh-gossip/src/proto/topic.rs
```diff
@@ -0,0 +1,858 @@
...
+                }
+                OutEvent::EmitEvent(Event::NeighborDown(peer)) => {
+                    self.gossip.handle(GossipIn::NeighborDown(*peer), &mut io)
+                }
+                _ => {}
+            }
+        }
+        // Note that this is a no-op because plumtree::handle(NeighborUp | NeighborDown)
+        // above does not emit any OutEvents.
+        self.outbox.extend(io.drain(..));
```
dignifiedquire, line 238: why not make `Outbox` be a `VecDeque` and pass it directly through?

iroh-gossip/src/proto/plumtree.rs
```diff
@@ -0,0 +1,462 @@
...
+        Gossip {
+            id: self.id,
+            content: self.content,
+            round: self.round.next(),
+        }
+    }
+}
+impl fmt::Debug for Gossip {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Gossip")
```
dignifiedquire, line 129: could use `derive_more`

iroh-gossip/src/proto/plumtree.rs
```diff
@@ -0,0 +1,462 @@
...
+    pub dispatch_timeout: Duration,
+    pub optimization_threshold: Round,
+}
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            graft_timeout_1: Duration::from_millis(80),
+            graft_timeout_2: Duration::from_secs(40),
+            dispatch_timeout: Duration::from_millis(50),
+            optimization_threshold: Round(7),
```
dignifiedquire, line 162: would be good to document/describe these defaults

iroh-gossip/src/proto/plumtree.rs
```diff
@@ -0,0 +1,462 @@
...
+            missing_messages: Default::default(),
+            received_messages: Default::default(),
+            graft_timer_scheduled: Default::default(),
+            cache: Default::default(),
+            dispatch_timer_scheduled: false,
+            stats: Default::default(),
+        }
+    }
+
+    // TODO: optimization step from the paper
```
dignifiedquire, line 211: how important is this?

iroh-gossip/src/proto/plumtree.rs
```diff
@@ -0,0 +1,462 @@
...
+        }
+    }
+
+    fn optimize_tree(
+        &mut self,
+        sender: &PA,
+        message: &Gossip,
+        previous_ihaves: VecDeque<(PA, Round)>,
+        io: &mut impl IO<PA>,
+    ) {
```
dignifiedquire, line 317: might be good to have a unit test for this specifically

iroh-gossip/src/proto/mod.rs
```diff
@@ -0,0 +1,29 @@
+use std::{fmt, hash::Hash};
+
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
+
+pub mod hyparview;
+pub mod plumtree;
```
dignifiedquire, line 6: it would be great to have a highlevel description somewhere how plumtree and hyparview are setup to work together/integrate

iroh-gossip/src/proto/plumtree.rs
Frando: It's actually already implemented, forgot to remove the TODO.

iroh-gossip/src/proto/state.rs
Frando: So the `State` assembles a list of `OutEvent`s and then returns an `Iterator` of those `OutEvent`s to the caller. 
The though was: a) keep it an implementation detail how the outevents are stored in the state inbetween, and b) possibly reduce allocations, as the `Vec` that stores the outevents in the state is reused this way, and the caller usually can handle the events right while iterating.

Does that answer your question? Or, what would you suggest instead?

iroh-gossip/src/proto/topic.rs
Frando: `self.outbox` is a `Vec` but could be a `VecDeque` just the same - the issue is that in the lines above, I'm iterating over `self.outbox` and then padding references of some events into the `self.gossip.handle`, therefore I cannot pass `&mut self.outbox` into there as well (because that would be a mutable ref plus an immutable ref, which of course cannot work).

iroh-gossip/src/proto/state.rs
Frando: Yes, good point. I added a `Quit` command.

iroh-gossip/src/proto/state.rs
Frando: Currently I don't think that's easily possible, because they need the mutable references. Also, with current settings the numbers will be quite low usually.
There might be places where we could indeed optimize by parallelizing things though. I'll think a bit more if/where this is feasable. Also, maybe we should run some benchmarks first if its actually worth it?

iroh-gossip/src/proto/plumtree.rs
Frando: Yup, good point. I pushed two commits that document the default values for plumtree and hyparview. Note: They will have to be validated in experiments or practice; their source is either the defaults from the paper; or wild guesses.
Also: They will depend on which network size and sender/receiver distrubtion we are optimizing for. Some implementations derive them from an `expected_network_size` parameter.

iroh-gossip/src/proto/state.rs
dignifiedquire: totally, was more thinking out loud/wondering

iroh-gossip/src/proto/mod.rs
Frando: I added a high level description to the `proto` module docs. Can be extended further for sure, but now there's something there at least.

dignifiedquire:
From my side the last things that are left are the unit tests I mentioned above & resolving merge conflicts, then we can merge this.

iroh-gossip/examples/chat.rs
```diff
@@ -0,0 +1,345 @@
...
+    #[clap(short, long, default_value = "0")]
+    bind_port: u16,
+    #[clap(subcommand)]
+    command: Command,
+}
+
+#[derive(Parser, Debug)]
+enum Command {
+    OpenTopic { topic: String },
+    JoinTicket { ticket: String },
```
divagant-martian, line 62: Could you add some docs to these, since that turns into the descriptions/help message of the commands?

iroh-gossip/examples/chat.rs
```diff
@@ -0,0 +1,345 @@
...
+    MagicEndpoint,
+};
+use once_cell::sync::OnceCell;
+use serde::{Deserialize, Serialize};
+use tokio::sync::mpsc;
+use url::Url;
+
+#[derive(Parser, Debug)]
+struct Args {
+    /// Private key to derive our peer id from
```
divagant-martian, line 40:
```suggestion
    /// Private key to derive our peer id from.
```

iroh-gossip/examples/chat.rs
```diff
@@ -0,0 +1,345 @@
...
+    #[clap(long)]
+    private_key: Option<String>,
+    /// Set a custom DERP server. By default, the DERP server hosted by n0 will be used.
+    #[clap(short, long)]
+    derp: Option<Url>,
+    /// Disable DERP completeley
```
divagant-martian, line 46:
```suggestion
    /// Disable DERP completely.
```

iroh-gossip/examples/chat.rs
```diff
@@ -0,0 +1,345 @@
...
+    #[clap(long)]
+    no_derp: bool,
+    /// Set your nickname
```
divagant-martian, line 49:
```suggestion
    /// Set your nickname.
```

iroh-gossip/examples/chat.rs
```diff
@@ -0,0 +1,345 @@
...
+
+#[derive(Parser, Debug)]
+enum Command {
+    OpenTopic { topic: String },
+    JoinTicket { ticket: String },
+}
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    tracing_subscriber::fmt::init();
```
divagant-martian, line 67: would be good to use the env filter

iroh-gossip/examples/chat.rs
```diff
@@ -0,0 +1,345 @@
...
+        (false, Some(url)) => Some(derp_map_from_url(url)?),
+        (true, None) => None,
+        (true, Some(_)) => bail!("You cannot set --no-derp and --derp at the same time"),
+    };
+    println!("> using DERP servers: {:?}", fmt_derp_map(&derp_map));
+
+    // init a cell that will hold our gossip handle to be used in endpoint callbacks
+    let gossip_cell: OnceCell<GossipHandle> = OnceCell::new();
+    // init a channel that will emit once the initial endpoints of our local node are discovered
+    let (initial_endpoints_tx, mut initial_endpoints_rx) = mpsc::channel(1);
```
divagant-martian, line 89: it looks like this should be a oneshot? (`tokio::sync::oneshot`)

iroh-gossip/examples/chat.rs
```diff
@@ -0,0 +1,345 @@
...
+
+    // pass our initial peer info to the gossip protocol
+    gossip.update_endpoints(&initial_endpoints)?;
+
+    // spawn our endpoint loop that forwards incoming connections to the gossiper
+    tokio::spawn(endpoint_loop(endpoint.clone(), gossip.clone()));
+
+    let mut our_ticket = match &args.command {
+        Command::OpenTopic { topic } => {
+            println!("> opening topic {topic} and waiting for peers to join us...");
```
divagant-martian, line 126: this within the example is confusing. The topic passed to the example is a `utf-8` string, but the real topic is a `hash`. It would be far clearer to print the topic hash along the topic name, since that's what's actually printed on the join side.

Right now I get for example:
```bash
$ <command> <args> open-topic testing
# > trying opening topic testing and waiting for peers to join us...
# > ticket to join us: mhgjrzbn5wlia6agx4lcbyj4jzvbxbigrswzgobkfyyqpqtjv37aa
```

and then on the joining side:
```bash
$ <command> <args> join-ticket mhgjrzbn5wlia6agx4lcbyj4jzvbxbigrswzgobkfyyqpqtjv37aa
# > joining topic mhgjrzbn5wlia6agx4lcbyj4jzvbxbigrswzgobkfyyqpqtjv37a and connecting to []
```

someone trying this example would in principle expect `testing` on the joining side. If this is (likely) not possible to get back, giving the hash on the `topic` command makes this match

iroh-gossip/examples/chat.rs
```diff
@@ -0,0 +1,345 @@
...
+            let topic = blake3::hash(topic.as_bytes()).into();
+            gossip.join(topic, vec![]).await?;
```
divagant-martian, line 128: the example doesn't work for me as is right now.

`gossip.join` hangs in there indefinitely if there are no peers. But I need the ticket to join to begin with, which is printed 24 lines below this. 

To get this to work I had to move adding the `initial_endpoints` addresses and printing the ticket to before joining, to actually get something to use in the join side

iroh-gossip/examples/chat.rs
```diff
@@ -0,0 +1,345 @@
...
+    });
+    println!("> ticket to join us: {our_ticket}");
+
+    let topic = our_ticket.topic;
+
+    // broadcast our name, if set
+    if let Some(name) = args.name {
+        let message = Message::AboutMe { name };
+        let encoded_message = SignedMessage::sign_and_encode(endpoint.keypair(), &message)?;
+        gossip.broadcast(topic, encoded_message).await?;
```
divagant-martian, line 160: it would be good to add a print here to know there is a broadcast going on

divagant-martian (changes requested):
have only reviewed the example since it was an easy entry point. It didn't work for me without a bit of changes to the example itself

iroh-gossip/examples/chat.rs
Frando: The env filter is enabled by default through the `env-filter` feature, which is enabled in iroh-gossip's dev dependencies

iroh-gossip/examples/chat.rs
Frando: Conceptually it should be, yes! However to use tokio's oneshot channel we'd have to wrap it in a `Arc<Mutex<Option>>>` unfortunately, which I thought to be a bit much for this example. Added this comment to explain in the code:
```
    // (not using a oneshot channel here because tokio::oneshot::Sender::send takes self, which
    // does not work in the on_endpoints Fn closure)
```

iroh-gossip/examples/chat.rs
Frando: Oh yes, this is a good point. Thanks for raising, not too much thought has gone into the semantics of this example.
With the most recent comment, I removed the topic-is-a-hashed-string "feature". Instead, you now either pass in a topic id directly as 32 bytes in base32 encoding, or a random topic id will be created.
I think it is much clearer this way for the example.

iroh-gossip/examples/chat.rs
Frando: Oh, sorry for that. Haven't kept up with the example after changing the semantics of `join`. Now fixed in the recent commit.

iroh-gossip/src/proto/plumtree.rs
Frando: I added a small unit test that verifies that the optimization step kicks in on the right messages. A full test with a simulated network would be good to have, too, but quite a bit more of work, so would defer that for later.

Frando:
I successfully rebased upon main!

Also added a couple of cleanups and improved the code structure in the chat example,

Because iroh-gossip conceptually depends on #1248, I added a [commit with an intermediate fix](https://github.com/n0-computer/iroh/pull/1149/commits/7fa47db5b739a4d596c548c553a6199b92663542): If you connect to a peer by just its peer id, it would always fail on current main. This commit here changes that to a warning, which means it will succeed if you previously added addrs or derp_region for that peer, but will fail with timeout and not immediately in case a peer is undialable. This will need a better structure in iroh-net to be fixed properly (see #1248).

iroh-gossip/src/proto/plumtree.rs
```diff
@@ -0,0 +1,500 @@
...
+
+use super::{PeerAddress, IO};
+
+pub enum InEvent<PA> {
+    RecvMessage(PA, Message),
+    Broadcast(Bytes),
+    TimerExpired(Timer),
+    NeighborUp(PA),
+    NeighborDown(PA),
+    PeerDisconnected(PA),
```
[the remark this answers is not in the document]
Frando: Makes sense!

iroh-gossip/src/proto/plumtree.rs
```diff
@@ -0,0 +1,603 @@
...
+        let entry = self
+            .missing_messages
+            .get_mut(&id)
+            .and_then(|entries| entries.pop_front());
+        if let Some((peer, round)) = entry {
+            self.add_eager(peer);
+            let message = Message::Graft(Graft {
+                id: Some(id),
+                round,
+            });
```
divagant-martian, line 425: it would be good to verify that the peer is not already eager to prevent sending extra grafts (in principle we could have received or performed a graft for X/Y reason between the creation of the timer and it's deadline)

iroh-gossip/src/proto/plumtree.rs
```diff
@@ -0,0 +1,603 @@
...
+            !ihaves.is_empty()
+        });
+        self.eager_push_peers.remove(&peer);
+        self.lazy_push_peers.remove(&peer);
+    }
+
+    /// Moves peer into eager set.
+    fn add_eager(&mut self, peer: PA) {
+        self.eager_push_peers.insert(peer);
+        self.lazy_push_peers.remove(&peer);
```
divagant-martian, line 469: I assume this is doing an implicit copy that be prevented by swapping these two lines

iroh-gossip/src/proto/plumtree.rs
```diff
@@ -0,0 +1,603 @@
...
+        }
+    }
+
+    fn optimize_tree(
+        &mut self,
+        sender: &PA,
+        message: &Gossip,
+        previous_ihaves: VecDeque<(PA, Round)>,
+        io: &mut impl IO<PA>,
+    ) {
```
divagant-martian, line 364: my understanding is this should be done with lazy peers, so it would be good to verify the peer is not already eager

divagant-martian (commented):
did a pass over `plumtree.rs`. I think the implementation is legit wrt the paper and my questions/worries are a matter of the algorithm and not your implementation.

As companion to my review I added #1303 

Also, the example now works without further intervention, thanks

iroh-gossip/src/proto/plumtree.rs
```diff
@@ -0,0 +1,603 @@
...
+    /// included message ID.
+    Graft(Graft),
+    /// When receiving IHave, do nothing initially, and request the messages for the included
+    /// message IDs after some time if they aren't pushed eagerly to us.
+    IHave(Vec<IHave>),
+}
+
+#[derive(Serialize, Deserialize, Clone, derive_more::Debug, PartialEq, Eq)]
+pub struct Gossip {
+    id: MessageId,
```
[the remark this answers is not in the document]
Frando: Oh, this is a good point. We should just always verify the message id when received a `Gossip`.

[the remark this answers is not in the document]
Frando: I just pushed a commit that validates messages (plus test): https://github.com/n0-computer/iroh/pull/1149/commits/93a2250d28354364b2e5475cb7bb9cfc72e54baf

iroh-gossip/src/proto/plumtree.rs
```diff
@@ -0,0 +1,603 @@
...
+    hash::Hash,
+    time::Duration,
+};
+
+use bytes::Bytes;
+use derive_more::{Add, From, Sub};
+use indexmap::IndexSet;
+use serde::{Deserialize, Serialize};
+
+use super::{PeerAddress, IO};
```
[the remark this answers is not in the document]
Frando: Yeah, I though the same, there's this comment already on the `PeerAddress` type:
```
/// TODO: Rename to `PeerId`? It does not necessarily refer to a peer's address, as long as the
/// networking layer can translate the value of its concrete type into an address.
```
I was worried a bit about just renaming to `PeerId` because having the same token for a trait (here) and struct (iroh-net and iroh) might be a bit weird. If you or someone else has suggestions, please share :-)

We could do `PeerIdT`, I've seen it at other places, but also was never too fond of that nomenclature. But it's fine, I guess. Or maybe `PeerIdentity` or something like this.

iroh-gossip/src/proto/plumtree.rs
```diff
@@ -0,0 +1,603 @@
...
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
+pub struct IHave {
+    id: MessageId,
+    round: Round,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
+pub struct Graft {
+    id: Option<MessageId>,
+    round: Round,
```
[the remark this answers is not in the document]
Frando: Good point. I rechecked the paper, and there the algorithm when receiving a Graft message is spec'd like this:
```
upon event Receive(GRAFT, mID, round, sender) do
 eagerPushPeers ←− eagerPushPeers ∪ {sender}
 lazyPushPeers ←− lazyPushPeers \ {sender}
 if mID ∈ receivedMsgs do
trigger Send(GOSSIP, sender, m, mID, round, myself)
```
so it uses the `round` of the received graft message as the `round` for the `Gossip` message sent in reply. In my code, the `round` of the `Gossip` message sent in reply to a `Graft` message is as it were when the peer receiving the graft sent that `Gossip` message to its eager peers originally.

I'm trying to make sense of what is better or how it would affect further proceedings, but did not come to a conclusion yet.

iroh-gossip/src/proto/plumtree.rs
Frando: That's a good question, I'm actually not sure: Let's say we receive IHaves A and B both from peer X, and IHave for B also from peer Y (later). The timer for A is triggered, and we send a Graft to peer X. The timer for B is triggered. We want the message content for B. Should we request (graft) it from peer X or not? In your interpretation, we should not, because by now peer X is already eager. However, if peer X is the better candidate to be grafted (because its IHave was received earlier than that of peer Y, wouldn't it be better to just also request message B from peer X (by sending a Graft) instead of also grafting Y (which was slower to deliver to us than X)?

iroh-gossip/src/proto/plumtree.rs
```diff
@@ -0,0 +1,603 @@
...
+    me: PA,
+    config: Config,
+
+    pub(crate) eager_push_peers: HashSet<PA>,
+    pub(crate) lazy_push_peers: HashSet<PA>,
+
+    lazy_push_queue: HashMap<PA, Vec<IHave>>,
+
+    missing_messages: HashMap<MessageId, VecDeque<(PA, Round)>>,
+    received_messages: IndexSet<MessageId>,
```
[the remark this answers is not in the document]
Frando: Good point about not needing `IndexSet`. I'd keep it separate from the cache for now, we might want to keep the list of received messages around for longer than the actual message content. Both will need to be cleared out eventually though (noted in the open tasks in the PR description, will address that next).

iroh-gossip/src/proto/plumtree.rs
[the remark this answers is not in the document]
divagant-martian: this depends on what others would like to do tbh. If we will be computing the hash, we might as well not send the message id to begin with and simply produce it on receive. I'm not arguing in favor or against any option, so happy to hear other opinions

iroh-gossip/src/proto/plumtree.rs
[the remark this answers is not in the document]
divagant-martian: reading it again I have no idea what's the point of sending this to begin with since the information is redundant and redundancy is not verified. Just for the sake of simplicity I would remove it.

For completeness though, I'm going to take a wild guess. I think maybe the authors wanted to have some kind of "conditional" graft. Something along the lines of: graft only if you have this with less rounds than `r` and then decided to move the logic to the sender side with the `IHave`s to avoid simultaneous grafts. This is based on absolutely nothing, like I said .. _wild_ guess

iroh-gossip/src/proto/plumtree.rs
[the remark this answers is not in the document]
divagant-martian: yeah fair :+1:

iroh-gossip/src/proto/plumtree.rs
divagant-martian: I think I'm forgetting that grafting means not only "upgrade our link" but also "gimme the message". In that sense leaving it as it seems fine to me.

However the discussion gets interesting because if we overwhelm peer X with "gimme" requests, it will also make sense to load balance with Y to get the rest. To be clear I think this can be a further optimization we likely don't want to _eargerly_ get in :drum:

iroh-gossip/src/proto/hyparview.rs
```diff
@@ -0,0 +1,707 @@
...
+pub enum Message<PA> {
+    /// Sent to a peer if you want to join the swarm
+    Join(PeerData),
+    /// When receiving Join, ForwardJoin is forwarded to the peer's ActiveView to introduce the
+    /// new member.
+    ForwardJoin(ForwardJoin<PA>),
+    /// A shuffle request is sent occasionally to re-shuffle the PassiveView with contacts from
+    /// other peers.
+    Shuffle(Shuffle<PA>),
+    /// Peers reply to Shuffle requests with a random subset of their PassiveView.
```
divagant-martian, line 81: should mention both passive and active views
```suggestion
    /// Peers reply to [`Message::Shuffle`] requests with random peers from their passive and
    /// active views.
```

iroh-gossip/src/proto/hyparview.rs
```diff
@@ -0,0 +1,707 @@
...
+    ShuffleReply(ShuffleReply<PA>),
+    /// Request to add sender to an active view of recipient. If `highPriority` is set, it cannot
+    /// be denied.
```
divagant-martian, line 84: I assume you had a `highPriority: bool` before. More accurate to current code would be
```suggestion
    /// Request to add sender to the active view of the recipient. If [`Neighbor::priority`] is
    /// high, it cannot be denied.
```

iroh-gossip/src/proto/hyparview.rs
```diff
@@ -0,0 +1,707 @@
...
+    Neighbor(Neighbor),
+    /// Request to disconnect from a peer.
+    /// If `alive` is true, the other peer is not shutting down, so it should be added to the
+    /// passive set.
+    /// If `respond` is true, the peer should answer the disconnect request before shutting down
+    /// the connection.
```
divagant-martian, line 90: Personal preference when mentioning fields I can't directly see
```suggestion
    /// If [`Disconnect::alive`] is true, the other peer is not shutting down, so it should be
    /// added to the passive set.
    /// If [`Disconnect::respond`] is true, the peer should answer the disconnect request before
    /// shutting down the connection.
```

iroh-gossip/src/proto/hyparview.rs
```diff
@@ -0,0 +1,707 @@
...
+/// Will be forwarded in a random walk until `ttl` reaches 0.
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
+pub struct ForwardJoin<PA> {
+    /// The peer that newly joined the swarm
+    peer: PeerInfo<PA>,
+    /// The time-to-live for this message
+    ttl: Ttl,
+}
+
+/// Shuffle messages are send occasionally to shuffle our passive view with peers from other peer's
```
divagant-martian, line 120:
```suggestion
/// Shuffle messages are sent occasionally to shuffle our passive view with peers from other peer's
```

iroh-gossip/src/proto/hyparview.rs
```diff
@@ -0,0 +1,707 @@
...
+    /// Number of peers to which active connections are maintained
+    pub active_view_capacity: usize,
+    /// Number of peers for which contact information is remembered,
+    /// but to which we are not actively connected to.
+    pub passive_view_capacity: usize,
+    /// Number of hops a `ForwardJoin` message is propagated until the new peer's info
+    /// is added to a peer's active view.
+    pub active_random_walk_length: Ttl,
+    /// Number of hops a `ForwardJoin` message is propagated until the new peer's info
+    /// is added to a peer'passive active view.
```
divagant-martian, line 187:
```suggestion
    /// Number of hops a `ForwardJoin` message is propagated until the new peer's info
    /// is added to a peer's passive view.
```

iroh-gossip/src/proto/hyparview.rs
```diff
@@ -0,0 +1,707 @@
...
+        &mut self,
+        from: PA,
+        message: Message<PA>,
+        now: Instant,
+        io: &mut impl IO<PA>,
+    ) {
+        let is_disconnect = matches!(message, Message::Disconnect(Disconnect { .. }));
+        if !is_disconnect && !self.active_view.contains(&from) {
+            self.stats.total_connections += 1;
+        }
```
divagant-martian, line 314: not sure I understand what this is counting :thinking:

iroh-gossip/src/proto/hyparview.rs
```diff
@@ -0,0 +1,707 @@
...
+                alive: true,
+                respond: false,
+            },
+            io,
+        );
+    }
+
+    fn handle_quit(&mut self, io: &mut impl IO<PA>) {
+        let peers = self.active_view.iter().cloned().collect::<Vec<_>>();
+        for peer in peers {
```
divagant-martian, line 350:
```suggestion
        for peer in self.active_view.clone().into_iter() {
```
I added the `IntoIter` implementation in another suggestion. Saves you the collect

iroh-gossip/src/proto/hyparview.rs
```diff
@@ -0,0 +1,707 @@
...
+                Disconnect {
+                    alive: false,
+                    respond: true,
+                },
+                io,
+            );
+        }
+    }
+
+    fn on_join(&mut self, peer: PA, data: PeerData, now: Instant, io: &mut impl IO<PA>) {
```
divagant-martian, line 362: this one is interesting. I think we should first check if the peer is (at least) not in out active view already. Otherwise I can just use this node to spam `ForwardJoin` to a bunch of random nodes simply by repeatedly sending `Join`

iroh-gossip/src/proto/hyparview.rs
```diff
@@ -0,0 +1,707 @@
...
+        // (different from the one from which the request was received)."
+        if !self.active_view.contains(&message.peer.id) {
+            match self
+                .active_view
+                .pick_random_without(&[&sender], &mut self.rng)
+            {
+                None => {
+                    // TODO: I think this is unreachable!() but will have to check, maybe decrease
+                    // to warn
+                    unreachable!("no peers in active view but also did not add this node on forward join, this should not happen");
```
divagant-martian, line 418: this looks unreachable to me too, but I would suggest instead:
```suggestion
                    unreachable!("if the peer was not added, there are at least two peers in our active view.");
```

iroh-gossip/src/proto/hyparview.rs
```diff
@@ -0,0 +1,707 @@
...
+    fn insert_peer_info(&mut self, peer_info: PeerInfo<PA>, io: &mut impl IO<PA>) {
+        let old = self.peer_data.remove(&peer_info.id);
+        let same = matches!(old, Some(old) if old == peer_info.data);
+        if !same {
+            io.push(OutEvent::PeerData(peer_info.id, peer_info.data.clone()));
+        }
+        self.peer_data.insert(peer_info.id, peer_info.data);
+    }
+
+    fn on_shuffle(&mut self, from: PA, shuffle: Shuffle<PA>, io: &mut impl IO<PA>) {
```
divagant-martian, line 464: my interpretation is that shuffle should be responded when it cannot be forwarded as well. From the paper:
>A node q that receives a Shuffle request will first decrease its time to live. If the time to live of the message is greater than zero **and the number of nodes in q’s active view is greater than 1**, the node will select a random node from its active view, different from the one he received this shuffle message from, and simply forwards the Shuffle request. **Otherwise, node q accepts the Shuffle** request and send back

Right now is only forwarded when expired

iroh-gossip/src/proto/hyparview.rs
```diff
@@ -0,0 +1,707 @@
...
+        if shuffle.ttl.expired() {
+            let len = shuffle.nodes.len();
+            for node in shuffle.nodes {
+                self.add_passive(node.id, node.data, io);
+            }
+            let nodes = self
+                .passive_view
+                .shuffled_max(len, &mut self.rng)
+                .into_iter()
+                .map(|id| self.peer_info(&id).unwrap());
```
divagant-martian, line 474: `peer_info` should be bounded in some way, it would be good to handle the `None` case in a more gracious way just for future/bug proofing

iroh-gossip/src/proto/hyparview.rs
```diff
@@ -0,0 +1,707 @@
...
+            );
+            let passive = self.passive_view.shuffled_without_max(
+                &[node],
+                self.config.shuffle_passive_view_count,
+                &mut self.rng,
+            );
+            let nodes = active
+                .iter()
+                .chain(passive.iter())
+                .map(|id| self.peer_info(id).unwrap());
```
divagant-martian, line 525: same here

iroh-gossip/src/proto/hyparview.rs
```diff
@@ -0,0 +1,707 @@
...
+            if respond {
+                let message = Message::Disconnect(Disconnect {
+                    alive: true,
+                    respond: false,
+                });
+                io.push(OutEvent::SendMessage(peer, message));
+            }
+            io.push(OutEvent::DisconnectPeer(peer));
+            io.push(OutEvent::EmitEvent(Event::NeighborDown(peer)));
+            let data = self.peer_data.remove(&peer).unwrap();
```
divagant-martian, line 640: I agree we want to ensure active peer data is present which is likely true right now, but an expect would be appreciated if this explodes
```suggestion
            let data = self.peer_data.remove(&peer).expect("data for peer in active view is known");
```

iroh-gossip/src/proto/util.rs
```diff
@@ -0,0 +1,185 @@
...
+        self.inner.iter()
+    }
+
+    /// Create an iterator over the set in the order of insertion, while skipping the element in
+    /// `without`.
+    pub fn iter_without<'a>(&'a self, value: &'a T) -> impl Iterator<Item = &'a T> {
+        self.iter().filter(move |x| *x != value)
+    }
+}
+
```
divagant-martian, line 103:
```suggestion

impl<T> IntoIterator for IndexSet<T> {
    type Item = T;

    type IntoIter = <indexmap::IndexSet<T> as IntoIterator>::IntoIter;

    fn into_iter(self) -> Self::IntoIter {
        self.inner.into_iter()
    }
}

```

iroh-gossip/src/proto/hyparview.rs
```diff
@@ -0,0 +1,707 @@
...
+                self.add_passive(peer, data, io);
+            }
+        } else {
+            self.passive_view.remove(&peer);
+        }
+    }
+
+    fn handle_shuffle_timer(&mut self, io: &mut impl IO<PA>) {
+        if let Some(node) = self.active_view.pick_random(&mut self.rng) {
+            let active = self.active_view.shuffled_without_max(
```
divagant-martian, line 512: when I see `shuffled_without_max` I read _shuffle and remove the maximum_.  Maybe `shuffled_without_and_capped` or `shuffled_without_capped` would work? Open to other options?

divagant-martian:
pheww finished reviewing hyparview.. moving into the state next

divagant-martian (commented):
only missing `net` for review

iroh-gossip/src/net.rs
```diff
@@ -0,0 +1,819 @@
...
+            rx
+        }
+    }
+}
+
+async fn wait_for_neighbor_up(mut sub: broadcast::Receiver<Event>) -> anyhow::Result<()> {
+    loop {
+        match sub.recv().await {
+            Ok(Event::NeighborUp(_neighbor)) => break Ok(()),
+            Ok(_) | Err(broadcast::error::RecvError::Lagged(_)) => {}
```
Frando, line 513: self-review: if the channel lags, this could lead to an `Event::NeighborUp` to be missed, causing `GossipHandle::join` to never complete. Don't think this can be a problem in practice, because the loop does nothing for messages, so it should just never occur that there's a `RecvError::Lagged` (events would have to be produced faster than a loop that does nothing can consume).

iroh-gossip/src/proto/plumtree.rs
Frando: Fixed in [5a6f2fdd](https://github.com/n0-computer/iroh/pull/1149/commits/5a6f2fddcde3985bc478b64708d39c073a2e5315)

iroh-gossip/src/proto/hyparview.rs
Frando: I think you are correct. Changed accordingly in https://github.com/n0-computer/iroh/pull/1149/commits/f86cde0afca7cefec33c2f7f3cf1886b3d96d8ba

iroh-gossip/src/proto/state.rs
```diff
@@ -0,0 +1,413 @@
...
+        write!(f, "{}", text)
+    }
+}
+impl fmt::Debug for TopicId {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let mut text = data_encoding::BASE32_NOPAD.encode(&self.0);
+        text.make_ascii_lowercase();
+        write!(f, "{}…{}", &text[..5], &text[(text.len() - 2)..])
+    }
+}
```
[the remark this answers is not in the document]
Frando: My reasoning was that `Display` would display the actual content, e.g. if you wanted to share in ticket style, and `Debug` would display something suitable for debug / logging output, where the full 32 bytes usually (at least to me) is only annoying.
We should do the same for all 32 byte tokens / keys in Iroh (e..g `PeerId`). @dignifiedquire opinions?
My low-key preference would be Display is always full 32 bytes, Debug is shortened to be less annoying in logs.  But open to anything as long as we make it consistent and make it easy to do shortened keys in logs.

iroh-gossip/src/proto/state.rs
```diff
@@ -0,0 +1,413 @@
...
+                let quit = matches!(event, topic::InEvent::Command(topic::Command::Quit));
+
+                if let hash_map::Entry::Vacant(e) = self.states.entry(topic) {
+                    e.insert(topic::State::with_rng(
+                        self.me,
+                        self.me_data.clone(),
+                        self.config.clone(),
+                        self.rng.clone(),
+                    ));
+                }
```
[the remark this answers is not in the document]
Frando: Oh yes, good point. I don't think I added this on purpose, it is likely a leftover from before I added the part above that checks if its a join command.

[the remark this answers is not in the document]
Frando: Not only a leftover, tests even relied on this behavior. 
I fixed this now, and also made the `Join` command  be `Join(Vec<PeerId>>)`. When joining with no peers, the state is created and others can join you. When joining with peers, these peers are dialed.

iroh-gossip/src/proto/util.rs
```diff
@@ -0,0 +1,185 @@
...
+#[derive(Debug, Clone)]
+pub(crate) struct IndexSet<T> {
+    inner: indexmap::IndexSet<T>,
+}
+
+impl<T: Hash + Eq + PartialEq> Default for IndexSet<T> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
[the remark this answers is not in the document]
Frando: I had trouble getting `Default` to work without restricting to `T: Default`, which we don't need or want. Will have to retry to see what exactly the error was.

iroh-gossip/src/proto/util.rs
```diff
@@ -0,0 +1,185 @@
...
+    time::{Duration, Instant},
+};
+
+/// A hash set where the iteration order of the values is independent of their
+/// hash values.
+///
+/// This is wrapper around [indexmap::IndexSet] that limits the removal API to
+/// always do shift_remove (preserving the order of other elements) and adds a
+/// couple of utility methods to randomly select elements from the set.
+#[derive(Debug, Clone)]
```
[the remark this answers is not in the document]
Frando: I thought about this too and wasn't sure because a general tendency seems to be to not use `Deref` too much but actually it would make sense here and be fine I think now.

iroh-gossip/src/proto/util.rs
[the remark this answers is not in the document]
Frando: Actually, I cannot reproduce :sweat_smile:

iroh-gossip/src/proto/util.rs
```diff
@@ -0,0 +1,185 @@
...
+    pub fn first(&self) -> Option<(&Instant, &Vec<T>)> {
+        self.0.iter().next()
+    }
+}
+
+impl<T> Default for TimerMap<T> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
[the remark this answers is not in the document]
Frando: Aha, here it was that I didn't get the `derive` for `Default` to work. Here's the error. Cannot restrict T to Default here. Somehow the manual impl just works.
```
   --> iroh-gossip/src/net/util.rs:163:18
    |
163 |             map: TimerMap::default(),
    |                  ^^^^^^^^^^^^^^^^^ the trait `Default` is not implemented for `T`
    |
note: required for `TimerMap<T>` to implement `Default`
   --> iroh-gossip/src/proto/util.rs:133:17
    |
133 | #[derive(Debug, Default)]
    |                 ^^^^^^^ unsatisfied trait bound introduced in this `derive` macro
    = note: this error originates in the derive macro `Default` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider restricting type parameter `T`
    |
159 | impl<T: std::default::Default> Default for Timers<T> {
```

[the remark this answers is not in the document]
Frando: Here's a playground for that: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021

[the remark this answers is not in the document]
divagant-martian: on interesting! gotcha

iroh-gossip/src/proto/hyparview.rs
Frando: Good point. I made the `PeerData` optional and now also handle the missing case properly.

iroh-gossip/src/proto/state.rs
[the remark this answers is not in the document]
Frando: I added a `base32` module and util macro to unify this. Should move to a `iroh-common` crate IMO but that can happen better post-merge.

iroh-gossip/src/proto/util.rs
```diff
@@ -0,0 +1,185 @@
...
+};
+
+/// A hash set where the iteration order of the values is independent of their
+/// hash values.
+///
+/// This is wrapper around [indexmap::IndexSet] that limits the removal API to
+/// always do shift_remove (preserving the order of other elements) and adds a
+/// couple of utility methods to randomly select elements from the set.
+#[derive(Debug, Clone)]
+pub(crate) struct IndexSet<T> {
```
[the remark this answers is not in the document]
Frando: I changed this in the latest commit.

iroh-gossip/src/proto/util.rs
Frando: I added basic unit tests for `IndexSet`, `TimerMap` and the new `base32` utils.
What parts would you want to proptest? Can do, just not super sure if/where it would make sense here.

Frando:
I *think* I addressed all review comments! Thanks again @divagant-martian for the thorough review, this helped a lot in getting here :)

In [this commit](https://github.com/n0-computer/iroh/pull/1149/commits/4ce76a0febde354c7fae27c6b4e7eac6f6c12ac9) I also added a time bound cache to gossip messages and message ids. I think this covers the common case for now. If we also want to limit the cache by size I can add that too, with FIFO semantics? I think the time based cache should work well enough though and ensure that you can reply to grafts from other peers well.

So - from my side, if I'm not forgetting something right now, this should be ready for a final review :-)

iroh-gossip/src/proto/tests.rs
```diff
@@ -0,0 +1,484 @@
...
+//             Some(state) => {
+//                 let peers = state.swarm.active_view.iter().collect::<Vec<_>>();
+//                 eprintln!("{me:?}: {:?}", peers);
+//             }
+//             None => {
+//                 eprintln!("{me:?}: QUIT");
+//             }
+//         }
+//     }
+// }
```
divagant-martian, line 484: is this supposed to remain here?

iroh-gossip/src/proto/tests.rs
```diff
@@ -0,0 +1,484 @@
...
+                .map(|(peer, _topic, _msg)| peer)
+                .collect();
+            for peer in received.iter() {
+                expected.remove(peer);
+            }
+            // eprintln!(
+            //     "tick {tick:3} received {:5} remaining {:5} ",
+            //     received.len(),
+            //     expected.len(),
+            // );
```
divagant-martian, line 345: remove?

divagant-martian (approved):
it's not very reasonable for me to attempt to re-review this to the same depth as before, but on an overview of raised issues, it LGTM!

PR #1151, opened 2023-06-29
fix(iroh-net): allow derp only connections to upgrade

dignifiedquire:
If we have no endpoints, we use the CallMeMaybe to trigger an exchange of potential UDP addresses.

In my tests this allows localhost and transfers between nats, when only connecting via peer id (and derp region), with no default addresses.


ramfox (approved):
lgtm!

btw, I looked into it & we DO add the peerid to the endpoints list when examining a message over derp if we have never seen the peerid before, so that is not the reason we are hanging when trying to send over derp only.

PR #1157, opened 2023-07-03
feat(iroh-net): implement `HomeRouter` detection

dignifiedquire:
- [x] bsd/macos using manual impl because of  https://github.com/shellrow/default-net/issues/34
- [x] linux (using `default_net`)
- [x] windows (using `default_net`)

### DIFF
--- a/iroh-net/src/net/interfaces.rs
+++ b/iroh-net/src/net/interfaces.rs
@@ -369,7 +369,7 @@ impl HomeRouter {
         bsd::likely_home_router()
     }
 
-    #[cfg(any(target_os = "linux", target_os = "android", target_os = "window"))]
+    #[cfg(any(target_os = "linux", target_os = "android", target_os = "windows"))]
     fn get_default_gateway() -> Option<IpAddr> {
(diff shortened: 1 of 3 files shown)


dignifiedquire:
ci failure is an unrelated flake unfortunately

divagant-martian (commented):
compared to TS and looks legit to me. 

Comment - potentially for the future - :
I checked what `default_net` does for the self ip and what `if-addrs` does (`if-addrs` is the dep I was using and removed in favor of `default-net` just aiming for compatibility) and I'm not convinced on which approach is better:

- [`if-addrs`](https://github.com/messense/if-addrs/blob/master/src/posix.rs#L11) uses [`getifaddrs`](https://man7.org/linux/man-pages/man3/getifaddrs.3.html) in posix for example
- [`default_net`](https://github.com/shellrow/default-net/blob/main/src/interface/shared.rs#L4) binds to the unspecified ipv4 address

dignifiedquire:
> and I'm not convinced on which approach is better:

honestly me neither, but this seems to work good enough for now

PR #1162, opened 2023-07-04
[WIP] fix magic endpoint flaky tests

dignifiedquire:
No fixes yet, just more reliably reproducing the issue.

### DIFF
--- a/iroh-net/src/magic_endpoint.rs
+++ b/iroh-net/src/magic_endpoint.rs
@@ -400,6 +400,7 @@ pub async fn get_peer_id(connection: &quinn::Connection) -> anyhow::Result<PeerI
 
 #[cfg(test)]
 mod test {
+    use anyhow::Result;
     use futures::future::BoxFuture;
 
     use super::{accept_conn, MagicEndpoint};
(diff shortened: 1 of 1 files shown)


dignifiedquire:
Some potential issues I am seeing in the logs

```
2023-07-04T18:32:28.367022Z  WARN serve:handle_connection:derp_connection_handler: iroh_net::hp::derp::http::server: server "iroh derp http" io error:
 unable to receive client information

Caused by:
    0: read frame
    1: frame header
    2: unexpected end of file
```

Frando:
I still cannot get it to reproduce for me locally (linux) unfortunately, even with more iterations.

I had another look into the code though and my suspicion is:
* the `magicsocket_endpoint_connect_close ` test calls `MagicEndpoint::close(code, reason)` . this calls `quinn::Endpoint::close` which calls `quinn::Connection::close` for all open connections. the latter makes quinn create a close frame and write that as a last packet into its abstract udp socket which is magicsock AsyncUdpSocket, which pushes the Transmit into a channel
* the close packet now is queued in a channel, either `Actor.network_receiver` or `Actor.derp_actor_sender`
* close packets are not acked (I think?), so for quinn everything is done with the packet successfully written to its abstract udp socket so `quinn::Endpoint::wait_idle` completes (all close frames "sent" = alls conns cleanly shutdown)
* `MagicEndpoint::close` waits for the upper to be done (`wait_idle.await`) and then closes the magicsocket `Conn`
* magisocket `Conn::close` sends shutdown messages to all actors. they are either biased on the channel through which the shtudown arrives or not biased at all, so packets still lying in `network_receiver` or `derp_actor_` channels are potentially not transferred when the derp client shuts down
* this makes the test fail, because then the error code doesn't arrive at the other end

Still unsure, also skimming/reading most of this code now for the first time. If this is correct, the flakyness might point at an actual unclear shutdown scenario with quinn and magicsocket? So if that is the case maybe we'd have to add a more gracefull shutdown message flow to the magicsock, that first tries to actually send all pending outgoing packets to either the peer or the derper, and only then completes shutdown?

iroh-net/src/magic_endpoint.rs
```diff
@@ -455,122 +456,141 @@ mod test {
...
+                let m = recv.read_to_end(100).await?;
+                assert_eq!(m, b"ola");
+
+                send.write_all(b"mundo").await?;
+                send.finish().await?;
+                anyhow::Ok(())
+            });
+
+            p1_accept.await??;
+            p2_connect.await??;
```
flub, line 586: The order between these tasks is not guaranteed.  Therefore it is possible to call `.connect()` before the other has called `.accept()`.  It is not clear to me if this is guaranteed to work if you call connect before accept.

flub:
> I still cannot get it to reproduce for me locally (linux) unfortunately, even with more iterations.
> 
> I had another look into the code though and my suspicion is:
> 
>     * the `magic_endpoint_connect_close ` test calls `MagicEndpoint::close(code, reason)` . this calls `quinn::Endpoint::close` which calls `quinn::Connection::close` for all open connections. the latter makes quinn create a close frame and write that as a last packet into its `AsyncUdpSocket`, which is the `magicsock::Conn`, which pushes the Transmit into its `network_sender` channel sender (goes to `Actor.network_receiver`)
> 
>     * the close packet now is queued in a channel, either `Actor.network_receiver` (or if passed through the actor loop, in `derp_actor_receiver`)
> 
>     * QUIC close frames are not acked (I think?), so for quinn everything is done with the packet successfully written to its abstract udp socket so `quinn::Endpoint::wait_idle` completes (all close frames "sent" = alls conns cleanly shutdown)

A QUIC `CONNECTION_CLOSE` frame MAY elicit **one single** `CONNECTION_CLOSE` frame in response, which effectively serves as an ack.  Either way **both** sides wait in their closing or draining state (draining if they **received** a `CONNECTION_CLOSE`) wait for 3 times PTO (whatever PTO is, too complicated, doesn't matter, some kind of duration).  They do this because otherwise delayed and reordered packets can still cause havoc.

I *think* that Quinn's `Endpoint::wait_idle` waits for this whole waiting to be done, so really it should only return once both sides have fully dropped the connection and are no longer sending *any* sort of packet on them.

>     * `MagicEndpoint::close` waits for the upper to be done (`wait_idle.await`) and then closes the magicsocket `Conn`

So I'm not sure about `MagicEndpoint::close`'s chosen semantics.  At the very least it should have been called `close_and_wait_idle` I guess?  Though maybe providing them separately would have been better.  But that's not really the topic here, more an aside.

>     * magisocket `Conn::close` sends shutdown messages to all actors. they are either biased on the channel through which the shtudown arrives or not biased at all, so packets still lying in `network` or `derp_actor` channels are potentially not transferred when the derp client shuts down
> 
>     * this makes the test fail, because then the error code doesn't arrive at the other end
> 
> 
> if this is correct, the flakyness might point at an actual unclear shutdown scenario with quinn and magicsocket? if that is the case maybe we'd have to add a more gracefull shutdown message flow to the magicsock, that first tries to actually send all pending outgoing packets to either the peer or the derper, and only then completes shutdown?

Because of how `MagicEndpoint::close` is implemented with using `wait_idle` I'm not convinced that this analysis is the right path to explore.  I think it is fine to call `Conn::close` once `quinn::Endpoint::wait_idle` has returned.

iroh-net/src/magic_endpoint.rs
flub: I've done some testing and it seems this isn't true.  I can connect without having to call accept first.  (This does make me worry `Conn` is susceptible to some kind of DOS attack)

PR #1166, opened 2023-07-05
debugging stuff

divagant-martian:


divagant-martian:
closing since we already got debug logs of a failure

PR #1169, opened 2023-07-05
Get rid of MagicEndpoint::dial_peer and refactor dialing in get

rklaehn:
there is now only one dial fn, that basically just passes through the options to the MagicEndpoint. This is used no matter if you use tickets or explicitly given peer and addrs.

### DIFF
--- a/iroh-bytes/src/get.rs
+++ b/iroh-bytes/src/get.rs
@@ -14,6 +14,7 @@ use bao_tree::io::DecodeResponseItem;
 use bao_tree::outboard::PreOrderMemOutboard;
 use bao_tree::{ByteNum, ChunkNum};
 use bytes::BytesMut;
+use iroh_net::tls::Keypair;
 use iroh_net::{hp::derp::DerpMap, tls::PeerId};
 use postcard::experimental::max_size::MaxSize;
 use quinn::RecvStream;
@@ -70,6 +71,7 @@ pub async fn run_ticket(
     derp_map: Option<DerpMap>,
 ) -> Result<get_response_machine::AtInitial> {
     let connection = iroh_net::MagicEndpoint::dial_peer(
+        Keypair::generate(),
         ticket.peer(),
         &crate::P2P_ALPN,
         ticket.addrs(),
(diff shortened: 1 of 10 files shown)


iroh-net/src/magic_endpoint.rs
```diff
@@ -182,15 +182,20 @@ impl MagicEndpoint {
     /// If *keylog* is `true` and the KEYLOGFILE environment variable is present it will be
     /// considered a filename to which the TLS pre-master keys are logged.  This can be useful
     /// to be able to decrypt captured traffic for debugging purposes.
+    ///
+    /// Note that this function is a convenience wrapper around [MagicEndpointBuilder] and [Self::connect].
+    /// It should be used in short lived processes that only need to connect to a single peer.
+    /// Otherwise, it is more efficient to create a [MagicEndpoint] and then
+    /// use [Self::connect] to connect to multiple peers.
     pub async fn dial_peer(
```
rklaehn, line 190: Not sure if there should instead be DialOptions. Getting a bit much...

dignifiedquire (approved):

iroh-bytes/src/provider/ticket.rs
```diff
@@ -92,6 +92,16 @@ impl Ticket {
...
     }
+
+    pub fn get_options(&self, keypair: Keypair) -> get::Options {
+        get::Options {
+            peer_id: self.peer,
+            addrs: self.addrs.clone(),
+            keypair,
+            keylog: true,
+            derp_map: None,
+        }
```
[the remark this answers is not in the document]
rklaehn: it seems like the token is not part of these options but part of the request. So it does not belong to get::Options as far as I can see...

iroh-bytes/src/get.rs
```diff
@@ -62,40 +68,6 @@ impl Stats {
...
-        ticket.peer(),
-        &crate::P2P_ALPN,
-        ticket.addrs(),
-        derp_map,
-        keylog,
-    )
-    .await?;
-
-    let request = if ticket.token().is_some() && request.token().is_none() {
-        // we have a ticket, but no token, so we need to add the token to the request
```
dignifiedquire, line 82: Does this functionality still exist? maybe I am just missing it

dignifiedquire (approved):

PR #1171, opened 2023-07-05
docs: fill out derp documentation

ramfox:


iroh-net/src/hp/derp.rs
```diff
@@ -9,9 +9,12 @@
 //! between very aggressive NATs, firewalls, no IPv6, etc? Well, DERP.
 //! Based on tailscale/derp/derp.go
 
+#![deny(missing_docs, rustdoc::broken_intra_doc_links)]
 pub(crate) mod client;
 pub(crate) mod client_conn;
 pub(crate) mod clients;
+/// An http specific Derp Client and Derp Server. Allows for using tls or non tls connection
+/// upgrades.
```
dignifiedquire, line 17: I think it is nicer to put the module documentation at the beginning of the file using `//!`

iroh-net/src/hp/derp/http.rs
```diff
@@ -1,5 +1,5 @@
 mod client;
-mod mesh_clients;
+pub(crate) mod mesh_clients;
```
dignifiedquire, line 2: why this change?

iroh-net/src/hp/derp/http/client.rs
```diff
@@ -37,46 +37,67 @@ const CONNECT_TIMEOUT: Duration = Duration::from_secs(10);
 const MESH_CLIENT_REDIAL_DELAY: Duration = Duration::from_secs(5);
 
 #[derive(Debug, thiserror::Error)]
+/// Possible connection errors on the [Client]
 pub enum ClientError {
```
dignifiedquire, line 41: I am sorry, but could we unify these to be in the order 1. docs, 2. derive/macros, 3. definition?

```rust
/// Possible connection errors on the [Client]
#[derive(Debug, thiserror::Error)]
pub enum ClientError {
```

iroh-net/src/hp/derp/types.rs
```diff
@@ -101,7 +101,13 @@ impl ServerInfo {
     }
 }
 
+/// A PacketForwarder is can forward a packet to the `dstkey` from the `srckey`.
```
dignifiedquire, line 104: "is can"?

iroh-net/src/hp/derp/http.rs
ramfox: ahh, that was a mistake, changed it back, ty

dignifiedquire (approved):

Issue #964, opened 2023-05-04
Get rid of size prefixes in get request/response protocol


rklaehn:
There are two places where we use `write_lp`: 

- [x] the handshake and 
- [ ] the request.

The handshake is fixed size, so does not need a size prefix. The request is variable size, but it is also the last thing being sent, so the receiver could just read to the end.

dignifiedquire:
> so the receiver could just read to the end.

I would like to avoid this, given buffer allocations and potential memory attacks, it also doesn't cost us much in this case.

rklaehn:
How does a size prefix prevent you from memory attacks?

If you just allocate whatever is the size prefix, somebody can still send you u64::MAX as the size. And if you put a limit on it, you can also put a limit on the buffer that you read into from the stream.

rklaehn:
With https://github.com/n0-computer/iroh/pull/1408 this is implemented.

We still have a size prefix in the response to a CustomGet, but I think that one makes sense to keep. Only question is whether it should be varint encoded.

ramfox:
@rklaehn w/ #1408, are we okay to close?

rklaehn:
CustomGets are gone, so this is done....

PR #1172, opened 2023-07-06
refactor(iroh-bytes): remove handshake and rely on ALPN only for protocol negotiation

dignifiedquire:
Closes the simple part of #964.

### DIFF
--- a/iroh-bytes/src/get.rs
+++ b/iroh-bytes/src/get.rs
@@ -74,7 +74,7 @@ pub mod get_response_machine {
     use std::result;
 
     use crate::{
-        protocol::{read_lp, GetRequest, NonEmptyRequestRangeSpecIter, HANDSHAKE_SIZE},
+        protocol::{read_lp, GetRequest, NonEmptyRequestRangeSpecIter},
         tokio_util::ConcatenateSliceWriter,
     };
 
@@ -194,20 +194,13 @@ pub mod get_response_machine {
                 mut writer,
                 request,
             } = self;
-            let mut out_buffer = BytesMut::zeroed(1024);
-
             // 1. Send Handshake
             {
                 debug!("sending handshake");
                 let handshake = Handshake::new();
-                let used = postcard::to_slice(&handshake, &mut out_buffer)?;
-                debug_assert_eq!(
-                    used.len(),
-                    HANDSHAKE_SIZE,
-                    "serialized handshake had unexpected size"
-                );
+                let used = handshake.to_bytes();
                 writer
-                    .write_all(used)
+                    .write_all(&used)
                     .await
                     .map_err(|e| anyhow!("failed to write handshake {:?}", e))?;
             }
--- a/iroh-bytes/src/get.rs
+++ b/iroh-bytes/src/get.rs
@@ -28,7 +28,7 @@ use tracing::{debug, error};
 pub use crate::util::Hash;
 
 use crate::blobs::Collection;
-use crate::protocol::{write_lp, AnyGetRequest, Handshake, RangeSpecSeq};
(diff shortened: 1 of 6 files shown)


iroh-bytes/src/protocol.rs
```diff
@@ -22,11 +21,14 @@ pub(crate) const MAX_MESSAGE_SIZE: usize = 1024 * 1024 * 100;
...
 pub const VERSION: u64 = 2;
 
-#[derive(Deserialize, Serialize, Debug, PartialEq, Eq, Clone, MaxSize)]
+#[derive(Deserialize, Serialize, Debug, PartialEq, Eq, Clone)]
 pub(crate) struct Handshake {
     pub version: u64,
 }
 
+/// Serialized length of [Handshake].
+pub(crate) const HANDSHAKE_SIZE: usize = 1;
```
rklaehn, line 30: That's a nice handshake size. But what will we do after version 255? Maybe we should do a multihandshake crate 🤣 ...

iroh-bytes/src/protocol.rs
```diff
@@ -284,3 +298,17 @@ impl TryFrom<VarInt> for Closed {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_handshake_ser() {
+        let handshake = Handshake::new();
+        let ser = postcard::to_stdvec(&handshake).unwrap();
+        assert_eq!(ser.len(), HANDSHAKE_SIZE);
```
rklaehn, line 310: The postcard format is stable, so we could actually assert the byte(s) here as well, right?

rklaehn (dismissed):
Save all the bytes!

iroh-bytes/src/protocol.rs
dignifiedquire: Sorry, I actually ended up increasing the size, this comment made me realize that postcard under the hood still uses varints.. Now we have a fully hard coded handshake, with a lot of protocol versions left 😅

iroh-bytes/src/protocol.rs
```diff
@@ -19,19 +18,8 @@ use crate::util::Hash;
...
-    pub version: u64,
-}
-
-impl Handshake {
-    pub fn new() -> Self {
-        Self { version: VERSION }
-    }
-}
+/// The ALPN used with quic for the iroh bytes protocol.
+pub const ALPN: [u8; 13] = *b"/iroh-bytes/2";
```
flub, line 22: you want to lose the `n0/` prefix?

flub (approved):

iroh-bytes/src/get.rs
```diff
@@ -193,25 +192,15 @@ pub mod get_response_machine {
...
-                write_lp(&mut writer, used).await?;
-            }
-
-            // 2. Send Request
+            // 1. Send Request
             {
                 debug!("sending request");
                 // wrap the get request in a request so we can serialize it
                 let request_bytes = postcard::to_stdvec(&request)?;
                 write_lp(&mut writer, &request_bytes).await?;
```
rklaehn, line 200: at some point we can get rid of this lp as well... But not yet.

rklaehn (approved):

iroh-bytes/src/protocol.rs
dignifiedquire: yeah, seems unnecessary to me,

flub: but why leave the leading `/`?

dignifiedquire: personal preference, not a strong one though

PR #1173, opened 2023-07-06
refactor: pluggable metrics

Arqu:
Still need to refactor some of the constants & complete the move of the netcheck, magicsock and portmap metrics.

Would like a first 👀 on the 🍝  I've made here.


iroh-metrics/src/core.rs
```diff
@@ -119,17 +141,29 @@ pub trait MObserver {
...
     if CORE.is_enabled() {
-        match c {
-            Collector::Iroh => CORE.iroh_metrics().record(m, v),
-            Collector::Magicsock => CORE.magicsock_metrics().record(m, v),
-            Collector::Netcheck => CORE.netcheck_metrics().record(m, v),
-            Collector::Portmap => CORE.portmap_metrics().record(m, v),
-        };
+        let cc = c.to_string();
+        let mn = m.name().to_string();
+        tokio::task::spawn(async move {
```
dignifiedquire, line 151: why spawn?

iroh-metrics/src/core.rs
```diff
@@ -138,30 +172,28 @@ where
...
     if CORE.is_enabled() {
-        match c {
-            Collector::Iroh => CORE.iroh_metrics().observe(m, v),
-            Collector::Magicsock => CORE.magicsock_metrics().observe(m, v),
-            Collector::Netcheck => CORE.netcheck_metrics().observe(m, v),
-            Collector::Portmap => CORE.portmap_metrics().observe(m, v),
-        };
+        let cc = c.to_string();
+        let mn = m.name().to_string();
+        tokio::task::spawn(async move {
```
dignifiedquire, line 182: why spawn if you are in an async fn?

iroh-metrics/src/macros.rs
```diff
@@ -115,18 +116,79 @@ macro_rules! make_metric_recorders {
                         )+
                     }
                 }
+
+                pub fn run(self, rx: std::sync::mpsc::Receiver<$crate::core::MMsg>) {
+                    tokio::task::spawn_blocking(move || {
```
dignifiedquire, line 121: `spawn_blocking` is for inside of async fns, should probably use a tokio channel if you need to wait on them inside of async fns

dignifiedquire:
The sync/async mix is a bit confusing and potentially problematic, let me know if you have a plan on how to resolve or if you have specific questions.

iroh-metrics/src/core.rs
Arqu: Not an async func, but need to await stuff. Making the above async too makes things a lot worse/incompatible with `once_cell::Lazy`

iroh-metrics/src/core.rs
Arqu: Yeah this passes only because it's not really used. Have to remove the async.

iroh-metrics/src/macros.rs
Arqu: Ok, so this is the part that's messed up and some next lever rust-fu might be able to help. 

Since everything in metrics is basically a singleton, `Lazy` forces me to have everything be `Sync`. Thus I have to rely on `SyncSender` from the `std`.

Alongside that If I run multiple of these collectors, they seem to starve the scheduler completely, because they probably just alternate awaiting for a message that won't come just yet. Moving it to a blocking spawn ensures it's separate from the rest and can run on it's own accord.

iroh-metrics/src/core.rs
dignifiedquire: but then you should use https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.block_on

iroh-metrics/Cargo.toml
```diff
@@ -20,6 +20,7 @@ hyper = { version = "0.14.25", features = ["server", "client", "http1", "tcp"] }
 
 [dev-dependencies]
 tokio = { version = "1", features = ["io-util", "sync", "rt", "net", "fs", "macros", "time", "test-util"] }
+tokio-test = "0.4.2"
```
dignifiedquire, line 23: is this actually used?

Arqu: Think so, the docs require a tokio runtime and this is one of the lighter/better options for docs from my google search when I initially did it.

dignifiedquire: the docs don’t requrie tokio anymore no async in metrics

iroh/src/commands/doctor.rs
```diff
@@ -275,16 +275,16 @@ impl Gui {
...
-        let metrics = &CORE;
-        if metrics.is_enabled() {
-            let mm = metrics.magicsock_metrics();
-            let send_ipv4 = HumanBytes(mm.send_ipv4.get());
-            let send_ipv6 = HumanBytes(mm.send_ipv6.get());
-            let send_derp = HumanBytes(mm.send_derp.get());
-            let recv_data_derp = HumanBytes(mm.recv_data_derp.get());
-            let recv_data_ipv4 = HumanBytes(mm.recv_data_ipv4.get());
-            let recv_data_ipv6 = HumanBytes(mm.recv_data_ipv6.get());
+    async fn update_counters(target: &ProgressBar) {
```
dignifiedquire, line 278: doesn't need to be `async`

dignifiedquire (approved):

PR #1174, opened 2023-07-06
fix: cross builds

Arqu:
Every now and then a cross docker container hangs around (failed job or something gets borked) and it keeps eating up RAM. Worst offender at 75GB for a single lingering container.

This ensures a clean slate and bumps the runners up to a reasonable number, so things should be faster and more stable.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -232,7 +232,7 @@ jobs:
       run: cross check --all --target ${{ matrix.target }}
 
     - name: test
-      run: cross test --all --target ${{ matrix.target }} -- --test-threads=16
+      run: cross test --all --target ${{ matrix.target }} -- --test-threads=12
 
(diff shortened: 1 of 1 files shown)


dignifiedquire (approved):

.github/workflows/ci.yml
```diff
@@ -220,6 +220,11 @@ jobs:
     - name: Install rust stable
       uses: dtolnay/rust-toolchain@stable
 
+    - name: Cleanup Docker
+      continue-on-error: true
+      run: |
+        docker kill $(docker ps -q)
```
dignifiedquire, line 226: isn't there a chance this kills containers that are still in use?

Arqu: No, there's a mutex lock on every run for every box, so no chance of overlap.

PR #1175, opened 2023-07-06
docs: iroh-net toplevel

flub:
Add some top-level docs for iroh-net

### DIFF
--- a/iroh-net/src/lib.rs
+++ b/iroh-net/src/lib.rs
@@ -1,16 +1,16 @@
+//! iroh-net provides connectivity for iroh.
+//!
+//! This crate is a collection of tools to establish direct connectivity between peers.  At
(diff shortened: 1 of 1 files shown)


iroh-net/src/lib.rs
```diff
@@ -1,16 +1,16 @@
+//! iroh-net provides connectivity for iroh.
+//!
+//! This crate is a collection of tools to establish direct connectivity between peers.  At
```
divagant-martian, line 3: isn't the `direct` technically false? as a lot of traffic is relayed?

dignifiedquire (approved):

iroh-net/src/lib.rs
flub: good call

PR #1176, opened 2023-07-06
ref: Change module structure and visibility and require docs

flub:
This modifies the modules inside the crate structure and visibility.
Specifically it moves the `config` and `command` modules from the iroh
*crate* into the iroh *binary*.  This could be a little confusing.
This means they are no longer part of the public API of the iroh
crate (and do not need to be documented).

It also fills in a few odd bits of missing doc comments and makes doc
comments required for public APIs on the crate.

When reviewing: the crux are the changes to `main.rs` and `lib.rs`.

### DIFF
--- a/iroh/src/lib.rs
+++ b/iroh/src/lib.rs
@@ -1,12 +1,8 @@
 //! Send data over the internet.
+#![deny(missing_docs, rustdoc::broken_intra_doc_links)]
 
 pub use iroh_bytes as bytes;
 pub use iroh_net as net;
 
 pub mod node;
 pub mod rpc_protocol;
-
-#[cfg(feature = "cli")]
-pub mod commands;
-#[cfg(feature = "cli")]
-pub mod config;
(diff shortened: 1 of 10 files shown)


iroh/src/lib.rs
```diff
@@ -1,12 +1,8 @@
...
 pub use iroh_bytes as bytes;
 pub use iroh_net as net;
 
 pub mod node;
 pub mod rpc_protocol;
-
-#[cfg(feature = "cli")]
-pub mod commands;
-#[cfg(feature = "cli")]
-pub mod config;
```
divagant-martian, line 12: looks cleaner to me

divagant-martian (commented):
Don't think I'm in position to approve, but it makes sense to me to not expose cli stuff inside the library under a feature. If it's required by the lib and main ,it can be moved to the shared code (lib) later

dignifiedquire (approved):

PR #1127, opened 2023-06-22
[WIP] iroh sync

dignifiedquire:
Opening this for tracking purposes

PR #1177, opened 2023-07-06
[WIP] feat: Add iroh-sync and integrate with gossip

Frando:
This PR starts with a rebase of #1127 on top of #1149

The main addition so far is `iroh/src/sync/live.rs` which integrates iroh-sync and iroh-gossip to provide live synchronization of a iroh-sync document among peers.
There's also an example in `iroh/examples/sync.rs` that showcases live editing a document with peers.

There is no integration yet with `iroh-bytes` which also means that the content of entries inserted by peers is missing (only the hash is there). Integration with iroh-bytes will be next.

Also there's lots of TODOs in the code. 

~~And something seems to be off for the initial sync (that actually uses iroh-sync). It does not work reliably right now, looking into debugging next.~~ a commit was missing, now fixed

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -620,6 +620,64 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6548a0ad5d2549e111e1f6a11a6c2e2d00ce6a3dafe22948d67c2b443f775e52"
 
 [[package]]
+name = "crossbeam"
+version = "0.8.2"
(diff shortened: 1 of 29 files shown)


Arqu:
Canceled the CI run, windows tests were running for 1.5 hrs.

Frando:
Closing in favor of #1216 which includes all commits from this PR.

PR #1178, opened 2023-07-06
fix(iroh): pass derp-map on get-options

dignifiedquire:
The derp map was not properly set when using `get --ticket`, which resulted in issues in the netsim tests, as now derp node was known on the `get` side.


Arqu (approved):
LGTM

PR #1180, opened 2023-07-07
wip remove the would help 

divagant-martian:


flub (commented):
So my high level comment is that this unwraps the probe-would-help question into the main loop.  making the main loop have more state to track and more logic to understand.  it's not really the way i was hoping to see this evolve.  the less the main loop has to track the better to me.

When I wrote this comment I was thinking of inverting the probe-would-help logic instead: on every probe result that arrives we could check which probes are still outstanding and cancel them if they are no longer going to help.

Doing it that way would allow the probes logic to be moved out to their own sub-module entirely - i'm very happy how this worked out with the hearpin check, all logic is contained there now.

Maybe it involves making the probe runner an "actor" in the same style as netcheck/reportgen/hairpin so that reportgen can send messages like "abort all IPv6", "abort all for region 1" or combinations of such.  In that case probably the new probes actor can itself already track what has been reported and what needs to be cancelled, moving probe-would-help entirely out of reportgen which would be great.

Maybe there is a pure-futures solution to this where all this logic is running in a probes future but it doesn't need to become an actor.  if it doesn't need to receive or send messages that's possible.

So in summary I would like to see reportgen know less about the individual tasks, not more.  It should only drive them and do any coordination between them that can't be done otherwise.  And I think probes can move to their own module and have their own bit of state including the probe-would-help logic.


But these are just my thoughts, feel free to have other opinions.

divagant-martian:
you explanation is great and I actually think this aligns with that thought, in the sense that splitting the probes is smaller, easier to track logical steps will allow to identify what part of the state is needed to move the whole probe logic to its own struct. 

Will probably continue this later when there is more time

PR #1186, opened 2023-07-07
feat(ci): allow running netsim from another branch

Arqu:


.github/workflows/netsim.yml
```diff
@@ -51,11 +51,24 @@ jobs:
       run: |
         cargo build --profile optimized-release --workspace --all-features
 
+    - name: Detect chuck branch
+      if: github.event_name == 'issue_comment'
+      id: detect_chuck_branch
+      run: |
+        a='${{ github.event.comment.body }}'
+        branch="main"
+        if [[ $a == '/netsim branch'* ]];
```
flub, line 60: so the entire comment body must be only `/netsim branch <branch_name>` and nothing more?

flub (approved):

.github/workflows/netsim.yml
Arqu: In theory you can add random stuff beyond but it only matches the 1st 2 parts for the propper branching and the 3rd for the value, rest is ignored.

PR #1188, opened 2023-07-07
refactor(metrics): remove macros for metric definitions

dignifiedquire:
- Only a single `Metric` type per metric group
- Only a single `Metric` trait needed
- No more string matching internally -> more typesafety at compile time
- No more macro needed to generate metric groups
- `record` is now `inc_by` to match the language from `prometheus_client`
- `inc!` and `inc_by!` shorthand macros are back


iroh-metrics/src/core.rs
```diff
@@ -10,30 +12,129 @@ static CORE: OnceCell<Core> = OnceCell::new();
...
     registry: Registry,
-    metrics_map: HashMap<&'static str, Box<dyn Metric>>,
+    metrics_map: ErasedSyncSet,
+}
+/// Open Metrics [`Counter`] to measure discrete events.
+///
+/// Single monotonically increasing value metric.
+#[derive(Debug, Clone)]
+pub struct Counter {
+    /// The actual prometheus counter.
```
Arqu, line 24: Not the biggest fan of re-wrapping every metric type, and some of the others can get quite ugly, but not end of the world to do.

Arqu (approved):
Nice! This has some next level rust-foo I'm soaking up here.
IMHO ended up fairly good on the ergonomics side, very close to what we initially had.

iroh-metrics/src/core.rs
dignifiedquire: will figure it out with the next one we actually add 😅

PR #1199, opened 2023-07-10
feat: add DerpMap::default_from_url

Frando:
## Description

Now that #1143 is merged, the default Derp map can be configured from a single URL for the common case of using DNS and the default STUN port. This PR adds a `DerpMap::default_from_url` method that just takes a Derp server URL as argument. When writing examples that accept a Derp server from e.g. a CLI argument this makes the Derp map construction simpler.

## Change checklist

- [x] Self-review.
- [x] Documentation updates


iroh-net/examples/magic.rs
```diff
@@ -52,9 +52,7 @@ async fn main() -> anyhow::Result<()> {
 
     let derp_map = match args.derp_url {
         None => default_derp_map(),
-        Some(url) => {
-            DerpMap::default_from_node(url, DEFAULT_DERP_STUN_PORT, UseIpv4::None, UseIpv6::None)
-        }
+        Some(url) => DerpMap::default_from_url(url),
```
dignifiedquire, line 55: I wonder if we should use `From<Url>` instead

rklaehn (approved):
Looks reasonable.

One unrelated thing: I tried out the example and had the server stop once because of a timeout. I think accept_conn producing an error should not terminate the accept loop.

```rust
let (peer_id, alpn, conn) = accept_conn(conn).await?;
```

should be
```rust
let Ok((peer_id, alpn, conn)) = accept_conn(conn).await else {
  error!("accept_conn errored");
  continue;
};

iroh-net/examples/magic.rs
rklaehn: I find `From` instances sometimes a bit hard to discover for such things.

flub: I agree with `From<Url>` though, it's very much the interface you expect from rust.  You can help discoverability, but also once you know about `From` you know to go look for it and really do expect it for any conversion that does not require extra input.

Frando: Maybe let's just do both? `From<Url>` but also keep `DerpMap::default_from_url` as public method?

dignifiedquire: Why keep both? The `From` impl is public as well and shows up nicely in the docs

Frando: I was wondering about docs to note that a DNS lookup will be performed and default STUN port will be assumed. There's no way to add docs to trait impls, right?
Maybe also here it doesn't matter much. Your call :)

PR #1202, opened 2023-07-10
Simplify iroh-bytes get

rklaehn:
## Description

Reduce the API surface of the iroh-bytes get module and simplify it a bit. No actual changes, just moving stuff that is specific to iroh to the iroh library, and adding a helper method for creating the start state of the get finite state machine.

Also moving around the various unfortunate util parts of iroh-bytes a bit. They are now in an `util` submodule. We might be able to fully remove them later once the db write abstraction part is completely done.

## Notes & open questions

~Note: there is going to be more of these, but I want to do it step by step.~
Note: some of the stuff in iroh-bytes/util/io might make it into bao-tree and be removed here.

## Change checklist

- [x] Self-review.


dignifiedquire:
I would expect API users to likely want to use those helper functions, hiding them completely in the CLI seems unfortunate

rklaehn:
> I would expect API users to likely want to use those helper functions, hiding them completely in the CLI seems unfortunate

But they are very specific to one particular use case, and probably also subject to change. So maybe put them in iroh, but not just in the CLI but also in library mode?

dignifiedquire:
> But they are very specific to one particular use case, and probably also subject to change. So maybe put them in iroh, but not just in the CLI but also in library mode?

Sounds like a good first step 👍

rklaehn:
Just moved stuff around in iroh-bytes some more. This is how the top level of iroh-bytes looks now:
image

rklaehn:
This is the latest top level docs page.

image

dignifiedquire (approved):

Issue #1097, opened 2023-06-12
Make connection instantanious when derper is available


flub:
When a derper server is available the connection establishment should be instant: initial traffic should flow via the derper connection and once holepunching established a direct connection traffic should transparently switch over.  It seems currently there are some delays.

Frando:
I was trying out the iroh_net magic.rs example with the default derper, listen and connect running on the same computer. it works, but connection establishment took about 12 seconds for me (compiled in debug mode). The network I'm in is a bit slowish, but not a lot. So, I wondered, and took logs and browsed a bit. The magicsock logs are a bit hard to read for me still, but below are some initial questions/potential findings.

here's a gist with logs for listen and connect (removed ansi codes, my IP and the date up to seconds for better readability): https://gist.github.com/Frando/d624c7e12f59f0c9b233fdbb15918be4

Why do we only establish the Derp connection after the netstate report finished? in my example run the report generation took about 1 second, and only afterwards the connection to the Derper is being established.  Edit: It seems MagicEndpoint::add_known_addr -> Conn::set_netmap waits for a netcheck to be complete and only afterwards we can call Conn::get_mapping_addr?
It seems the single connect establishes multiple connections to the derper with multiple handshakes etc. Why is that? (grep for e.g. connect_0 in the log_connect above)
It seems that sending quinn quic packets only starts after the disco phase is done - isn't the magicsock conn supposed to start sending immediately over the http/tls conn to the derper?

Arqu:
FWIW according to https://perf.iroh.computer these have been back to working as before since https://github.com/n0-computer/iroh/commit/a0ae22851453ea9e277adba8d52fe55f90edcef3
Haven't tested manually though.

flub:
> FWIW according to https://perf.iroh.computer these have been back to working as before since [a0ae228](https://github.com/n0-computer/iroh/commit/a0ae22851453ea9e277adba8d52fe55f90edcef3) Haven't tested manually though.

huh, that's really surprising.  I'll still investigate this, seems to odd.

dignifiedquire:
My latest tests show that while there might still be delays, there are no clear logical issues anymore, so I would suggest we close this for now.

dignifiedquire:
closing for now, please open a new issues about specifc perf issues

PR #1203, opened 2023-07-10
fix(iroh-net): better logic for initial derp connection

dignifiedquire:
## Description

- avoid double derp connection
- do not block on establishing the derp connection

Should help with some issues seen in  #1097

## Change checklist

- [x] Self-review.


Frando (approved):
LGTM and testing manually I didn't get any double connects anymore. 

However I am still not too familiar with the inner workings of MagicEndpoint, so sanity check from @flub would be good to have.

PR #1213, opened 2023-07-11
[WIP] feat: implement firewall setup for windows

### LINKED ISSUE #911: Implement Windows firewall rules
The `windows` crate has the required APIs likely. 

The actual rules we can orient ourselves on https://github.com/tailscale/tailscale/blob/main/wf/firewall.go

Ref
- https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/NetworkManagement/WindowsFirewall/index.html
- https://learn.microsoft.com/en-us/windows/win32/api/netfw/nn-netfw-inetfwopenports
- https://stackoverflow.com/questions/8488227/create-firewall-rule-to-open-port-per-application-programmatically-in-c-sharp
- https://github.com/microsoft/windows-rs/blob/a672f5f450b16c947840ab89fdc5ddd16d1fd0d0/crates/libs/windows/src/Windows/Win32/NetworkManagement/WindowsFirewall/impl.rs#L1471

dignifiedquire:
Still a work in progress. But the general shape of the code is there.

- Implements a generic wrapper around windows firewall management in `iroh_net::firewall::fwpm`. Based on https://github.com/inetaf/wf
- Implements a basic set of the firewall rules in `iroh_net::firewall::windows` based on https://github.com/tailscale/tailscale/blob/main/wf/firewall.go

Closes #911


dignifiedquire:
We actually don't need this, and can't quite use it either:

- Running these requires admin privileges, so folks would have to run iroh as administrator, which we do not want
- Running iroh from main on windows works as expected, there is a popup asking the user if internet access is allowed, and if it is, everything works as expected.

flub:
Previously someone did not manage to get this popup.  They were speculating this was because of UDP traffic (this was before holepunching).  Are you sure you switch to the non-relayed-via-derper transport?

dignifiedquire:
> Are you sure you switch to the non-relayed-via-derper transport?

Yes, I also confirmed that the firewall rules allowed UDP traffic.

dignifiedquire:
> Previously someone did not manage to get this popup. They were speculating this was because of UDP traffic (this was before holepunching)

I don't exactly know how that happened, but this might be more related to the way the wrapping application was managing this. In any case, this code I wrote would not solve that problem, as it would be even more complex to execute (eg force the running application to be execute with full administrative privileges)

PR #1214, opened 2023-07-11
fix: make sure to clean up any lingering processes

Arqu:
## Description

Occasionally there are lingering iroh/derper nodes running from tests (probably netsim failed runs) which might not be cleaned up properly (and might impact future runs on those boxes). This makes sure it kills anything remaining.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.


flub (approved):
this looks like a great reason to use cgroups in some capacity.  So that an entire CI run is contained in some resource groups and you can make sure everything spawned in that groups is killed at the end.

systemd probably has facilities for this, though I'd also have to spend a while figuring those out.  Anyway, only an idea to maybe use in the future.  I'm kind of surprised this isn't part of the github runner already and that you have to deal with this here.

PR #1215, opened 2023-07-12
Use boxed dyn callbacks to get rid of some type parameters

rklaehn:
## Description

Get rid of some type parameters where they do not provide much benefit.

We can not get rid of the database parameter, since that has type members for all the futures, and getting rid of this would mean boxing all futures. But for the request authentication handler and the custom get handler we can just box them.

## Notes & open questions

In the builder, should we leave it up to the caller to create an Arc<dyn ...> or do this ourselves? I prefer to let the caller do it
because this is just during setup, and they might have already arced it for some other reason.

## Change checklist

- [x] Self-review.


flub (approved):
I think it's fine to live the Arc dyn creation to the caller.

PR #1216, opened 2023-07-12
[WIP] feat: iroh-sync

Frando:
## Description

This PR adds `iroh-sync`, a document synchronization protocol, to iroh, and integrates with `iroh-net`, `iroh-gossip` and `iroh-bytes`.

At the core is the `iroh-sync` crate, with a set reconciliation algorithm implemented by @dignifiedquire. See https://github.com/n0-computer/iroh-sync/ for its prehistory. *TODO: More details*

The PR is structured as follows:

* `iroh-sync` crate contains the basic data structures. It has a dependency on `iroh_bytes` at the moment, but *only* for its `Hash` type. It does not perform any IO, the implementation is in-memory only. Its basic data structures is a `Replica`, which is the local instance of a multi-writer key value store.
 * in `iroh::sync`, there are `connect_and_sync` and `handle_connection` functions that integrate with `iroh_net` to perform the set reconciliation operation between two peers
  * `iroh::sync::live::LiveSync` is the handle to an actor that integrates sync with [gossip](#1149 ): Peers join an iroh-gossip topic for each open document, and the `LiveSync` then watches for changes in these documents and broadcasts the new entry to other peer over `iroh-gossip`. The entries are signed, so incoming entries are validated and then inserted into the local replica.
  * `iroh::sync::content` module integrates the replicas with `iroh-bytes` by providing a `Doc` abstraction that combines a replica with a way to fetch and insert blobs into a local `iroh-bytes` database. 

Finally, `iroh/examples/sync.rs` has a REPL to interact with documents. You can open documents, invite peers to join the doc, and insert / list / get entries. The example also provides basic persistence for documents by serializing them into a file on disc.

Here's a rough list of changes from the original PR in #1177:

* removes content support from iroh-sync
* adds a quick-and-dirty writable database to iroh-bytes (will be replaced with a better generic writable database soon)
* adds a `Downloader` to queue get requests for individual hashes from individual peers
* adds a `BlobStore` that combines the writable db with the downloader
* adds a `Doc` abstraction that combines an iroh-sync `Replica` with a `BlobStore` to download content from peers on-demand
* updates the sync repl example to plug it all together
* adds very basic persistence to `Replica` (encode to bytes) and uses this in the repl example
* made the REPL a proper REPL
* rebased on current main and #1149 

## Notes & open questions

* I did not yet update the tests in iroh-sync, so `cargo test` doesn't compile atm. Only the `sync` example works.
* right now, it only attempts to download from the author of a key. This will fail if the peer is offline or not reachable (through the derper to which you are connected or through addresses obtained from the gossip layer). This is of course not a final solution. We will need some heuristics about which peers to try to download from.
* I added a `Doc` abstraction that combines a replica with a content blob store. We will have to think if we want this and what API it should expose. It's quite nice from a user perspective I think.
* right now, there is a `DownloadMode` that is either `Always` or `Manual`. If `Manual`, then no content downloads are triggered. If `Auto`, it will always try to immediately download the content for newly added keys. We might also want to add a `Filtered` mode or so that executes a callback to allow app level heuristics on what to download
* The `WritableFileDatabase` is a quick-and-dirty intermediate solution until the [`generic-db-rewrite`](https://github.com/n0-computer/iroh/compare/main...generic-db-write) arrives
* The `Downloader` is not as quick and dirty, but should be replaced with functionality from iroh-bytes or moved to there and extended to really work well with fetching stuff from multiple peers, possibly in parallel in the future

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


iroh-sync/src/ranger.rs
```diff
@@ -401,11 +401,12 @@ where
 
     /// Processes an incoming message and produces a response.
     /// If terminated, returns `None`
-    pub fn process_message(&mut self, message: Message<K, V>) -> Option<Message<K, V>> {
+    pub fn process_message(&mut self, message: Message<K, V>) -> (Vec<K>, Option<Message<K, V>>) {
         let mut out = Vec::new();
 
         // TODO: can these allocs be avoided?
         let mut items = Vec::new();
+        let mut inserted = Vec::new();
```
dignifiedquire, line 409: this seems potentially expensive, should probably be a callback instead

iroh/src/sync.rs
```diff
@@ -0,0 +1,218 @@
...
+                        debug!("starting sync for {}", namespace);
+                        if let Some(msg) = r.sync_process_message(message) {
+                            send_sync_message(writer, msg).await?;
+                        } else {
+                            break;
+                        }
+                        replica = Some(r);
+                    }
+                    None => {
+                        // TODO: this should be possible.
```
dignifiedquire, line 141: No it shouldn't, the receiver needs to setup the namespace/doc first, even if empty, to sync it.

Frando:
This PR is now outdated because it is not rebased. Closing in favor of #1333 .

PR #1217, opened 2023-07-12
Move database to iroh

rklaehn:
## Description

This is something I wanted to do for a long time. This moves the concrete database and everything related to it into iroh. This includes the Collection type. So iroh-bytes does not know anything about what a collection is. This is merely defined in the iroh crate itself.

## Notes & open questions

~Should the stuff be in iroh the library or just iroh the binary. I think iroh the library is fine for a first step. The main idea is to get iroh-bytes nice and lean so we can document it well and write a spec.~

=> it makes sense to have this in iroh the library behind a feature flag

~Possible next step: moving InMemDatabase into iroh as well, and hiding both InMemDatabase and the persistent database behind feature flags. But this is big enough already.~

=> have done this as well

- what should be the iroh default collection? I tend to think that the most basic thing you might want to do is to just group some blobs without caring at all about metadata. For that a good default would be collection = `Vec<blake3::Hash>`, period.  I think this would make a good default.

Currently I have a default that just disables collections entirely, which also works.

## Change checklist

- [x] Self-review.

To review, forget about looking at the code. It's just lots of stuff being moved around. The way to do it is to run `cargo doc` and `cargo doc --no-default-features` and look at the crate structure.


iroh-bytes/src/provider.rs
```diff
@@ -127,7 +115,7 @@ pub enum ValidateProgress {
         /// the hash of the entry
         hash: Hash,
         /// the path of the entry on the local file system
-        path: Option<PathBuf>,
+        path: Option<String>,
```
dignifiedquire, line 118: Why is this a `String` if it is a file location?

dignifiedquire: but then the doc comment is wrong

rklaehn:
@dignifiedquire I thought about this a bit.

Really not fond of the idea of having the iroh production persistent db as a part of iroh-bytes behind a feature flag.

It is possible. But now that it is all neatly separated by traits I would prefer a separate crate in the iroh workspace, or have it in `iroh` the library crate.

It is going to change a lot, and it is very likely also going to do things unrelated to iroh-bytes stuff, such as storing documents and auth info. Keeping it separate would allow iroh-bytes to be just about the bao data exchange protocol.

dignifiedquire:
I guess this is fine for now. We can always move it somewhere else later if we don't like it.

rklaehn:
Now it is basically like this. There are 3 feature flags: `flat-db`, `mem-db` and `iroh-collection`. They are all on by default.

mem-db is the in memory blob store. Not sure if it is worth hiding it behind a feature flag since it is basically just a hashmap, but 🤷 

flat-db is the current persistent blob store. this one comes with lots of special logic about paths that is subject to change and so definitely worth the feature flag

iroh-collection is the definition of the iroh collection. This is now no longer baked into the node.

**node** is not behind a feature flag (can change that), so you can create e.g. a node with just an in memory db and no notion of collections at all, or with your own notion of collections.

So e.g. `mem-db` & `iroh-collections` means you are compatible with iroh nodes in terms of collections but don't have persistence.

iroh/src/node.rs
```diff
@@ -42,6 +29,13 @@ use quic_rpc::server::RpcChannel;
 use quic_rpc::transport::flume::FlumeConnection;
 use quic_rpc::transport::misc::DummyServerEndpoint;
 use quic_rpc::{RpcClient, RpcServer, ServiceConnection, ServiceEndpoint};
+use std::fmt::Debug;
+use std::future::Future;
+use std::net::{Ipv4Addr, SocketAddr};
+use std::pin::Pin;
+use std::sync::Arc;
+use std::task::Poll;
+use std::time::Duration;
```
dignifiedquire, line 38: general nit, please keep imports sorted as 

- std
- newline
- external crates
- newline
- internal (`self`, `crate`)

dignifiedquire:
Haven't reviewed the code in detail, but assuming this is pretty much only moving the existing code into the new structure.

iroh/src/node.rs
```diff
@@ -453,6 +450,7 @@ impl Callbacks {
         self.0.write().await.push(cb);
     }
 
+    #[allow(dead_code)]
```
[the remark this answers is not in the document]
rklaehn: if you build it with no features, this is never used, and you get a clippy warning.

iroh/src/node.rs
rklaehn: will do, just hard to keep this clean with all the moving stuff around...

rklaehn:
> Haven't reviewed the code in detail, but assuming this is pretty much only moving the existing code into the new structure.

Yes, that is correct. Spent the last days not writing much code, just moving stuff around and staring at CI...

iroh/src/node.rs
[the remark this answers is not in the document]
rklaehn: The only place where send gets called is here, in provide0

```
        self.inner
            .callbacks
            .send(Event::ByteProvide(
                iroh_bytes::provider::Event::CollectionAdded { hash },
            ))
            .await;
```

but this part is disabled when the flat-db feature is off, because we don't have generic write yet. So it is unused when compiling wiht --no-default-features

PR #1219, opened 2023-07-12
feat(iroh): pass a callback to subscribe

### LINKED ISSUE #1139: lossy node events results in flaky tests
The `iroh::node::Node` struct has a `.subscribe()` method which returns a broadcast receiver with 256 slots.  Since events are emitted for every blob it is hard to ensure you receive all events.  Yet for some applications, i.e. our tests, it is required to not miss any events.

We could enable this system to provide backpressure and slow down the provider.  One option is to use an callback to receive events, then it is up to the user to decide how to handle the events and whether they should provide backpressure or be lossy.

dignifiedquire:
## Description

This allows buffer management and channel construction to be more flexible on the callers side.

Closes #1139

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-bytes/src/provider.rs
+++ b/iroh-bytes/src/provider.rs
@@ -498,12 +498,15 @@ pub async fn transfer_collection<D: BaoMap, E: EventSender, C: CollectionParser>
     let mut c = if !just_root {
         // use the collection parser to parse the collection
         let (c, stats) = collection_parser.parse(0, &mut data).await?;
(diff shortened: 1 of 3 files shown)


flub (approved):

b5 (approved):
I made an attempt to update iroh.network & confirm the callback subscription approach still works, but I have at least an hour or two of catchup before I can sync up to the latest changes. Instead of blocking this PR, I'll file issues later once I get a chance to catch things up

PR #1222, opened 2023-07-13
first pass: `my_derp` and `add_known_addrs`

### LINKED ISSUE #1221: feat: integrate the notion of multiple derp regions into the code
* [x] add EU region as a potential region in the config
* [x] adjust code to expect a derp region when we add a peer to our address book
* [x] ensure we won't break anything by expecting to be able to dial by region 
* [x] add `region` as a flag in the `iroh get` cli
* [x] add `region` as part of the get ticket
* [x] adjust expectations for `iroh doctor` calls to include derp region
* [x] add `iroh doctor derp-regions` command that will return the information you need to adjust your derp configuration, including latencies for nodes
* [x] add `docs/regions.md` file that shows the default region configuration & explains the expectations around them

ramfox:
ensure we can display our own home derp, and that we add associate a derp region with a peer id

## Description

closes #1221 

#### what happens if the peer you want to talk to is in a region you do not understand?
Hi, I've been going back and forth a few times with this. Can I get some feedback.

Now that we have multiple derp regions, it's possible that the regions you have configured are NOT the region that the node you are trying to communicate is in.

So, how should we handle the case where we cannot connect to that specific derp region?

a) When we add a peer to our address book,  the derp region is required (along with the PublicKey & ALPNs), and we check our DERP configuration. If we don't have that derp region in our configuration, we fail when we attempt to add that peer to our address book. In this case, we won't ever attempt to dial the peer if no DERP region is provided or the DERP region provided does not exist in our configuration, even if we have a valid UDP connection
b) We do not require a derp region when you add a peer to the address book and we don't check if any derp region provided exists in our configuration. If no UDP address nor DERP region exists when we go to connect, we log a warning, but fail silently. Whatever timeouts exist on QUIC will timeout when no connection can be made. If you don't have logging, you will not understand why this failed.
c) You are not required to provide a DERP region, but if you DO, we will fail if that derp region does not exist in our configuration. The implication here is that if you have provided a DERP region, you expect hole punching to occur, and it is an error if you cannot attempt to hole punch in the current conditions. If you do NOT provide a derp region, you are accepting that hole punching is not possible or you have some outside information that the UDP addresses are accessible.

This last one seems like the best option to me & it's how I am planning to continue with the PR, but would love some feedback.

#### region 0
Currently, `region_id` 1 is expected to be NA, and `region_id` 2 is expected to be EU. This is currently only documented here: https://www.notion.so/number-zero/Derp-Regions-a63c309839e64d9ab19c23778a7b9c37

We also have a bit of a testing problem. Since the region_ids should be standardized, it seems we should leave the id 0 as a testing id.

#### `impl From<Url> for DerpMap`
Now that it is we have the possibility of multiple regions, we can no longer assume the default region id for any given `Url`, so we can no longer implement `From` on `DerpMap` for `Url`

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


flub:
I think I could choose b.

If we don't know the DERP region in our config we might still be able to connect directly to it, it may not even need holepunching at all.

If we do not have any direct address and can not reach the peer's home derp server [0] then we'd have to return either a `quinn::ConnectionError::Timeout` or `quinn::ConnectionError::TransportError` (ConnectionError if possible, `NO_VIABLE_PATH` might be suitable but currently a bit of a stretch since this isn't a real quic path (yet)) at connection time.

[0] Also, I think not knowing the derp server in our own config might be equivalent to the lightly broader "can not reach that derp server" which we also have to handle anyway.

dignifiedquire:
I think we should require at least a derp region or a UDP address, if neither is provided we don't have any chance of connecting and should fail fast.

ramfox:
> I think we should require at least a derp region or a UDP address, if neither is provided we don't have any chance of connecting and should fail fast.

> If we do not have any direct address and can not reach the peer's home derp server [0] then we'd have to return either a quinn::ConnectionError::Timeout or quinn::ConnectionError::TransportError (ConnectionError if possible, NO_VIABLE_PATH might be suitable but currently a bit of a stretch since this isn't a real quic path (yet)) at connection time.

Okay, so if we have no UDP addresses, and no derp server OR if that derp server is not one we have in configuration we error. (no UDP && ( no DERP || unknown DERP )) -> error

However, once we attempt to dial in the `Conn`, if we discover that our connection to the derp server fails us, we don't have a good way of reporting those errors, because all attempts to send packets happen through `poll_send`, which immediately sends the transmits off to the `Actor`, which more or less eats the errors. There is likely a good strategy to report these errors back up to `poll_send` (maybe some ugh mutex locked `error` field that we check and `take` each time we `poll_send`?), but I'd like a longer discussion about the implications of this. My main concern when reporting these connection errors, is that I'm not totally sure how it's supposed to work when we use one `magicsock::Conn` for multiple connections.

I'm going to move forward on this PR with the following scenarios in mind:
- If you do not have UDP addresses and no DERP region, we error
- If you do not have UDP addresses and the DERP region is unknown to us, we error
- If you have any UDP address and no DERP region or the DERP region is unknown to us, we log a warning, but we keep the current behavior that exists on main.

ramfox:
I can't believe I did this, but I added a lock 😢 😢 😢 

Now that `0` is a valid derp region (we use it in testing), we need a new way to represent that we are not connected to any derp region. So now our derp region is represented by `Option<u16>` (and not `Atomicu16`).  And now that we have the possibility of multiple derp regions, or no derp region at all, it seemed like this was a more logical way of describing what is going on.

If there is a strong feeling about this, PLEASE FEEL FREE TO CHANGE IT BACK.

iroh-net/src/hp/magicsock/conn.rs
```diff
@@ -202,16 +202,35 @@ pub struct Inner {
     /// None (or zero regions/nodes) means DERP is disabled.
     pub(super) derp_map: tokio::sync::RwLock<Option<DerpMap>>,
     /// Nearest DERP region ID; 0 means none/unknown.
-    my_derp: AtomicU16,
+    /// None means we are not connected or do not yet know the connection status.
+    // I'm sorry.
+    my_derp: tokio::sync::RwLock<Option<u16>>,
```
ramfox, line 207: the dreaded new lock.

Please feel free to reverse this decision.

ramfox: ... I've already reversed the decision.

ramfox:
Okay, I need to put this down for the night, but the `cli` tests are WAY more flakey now. Maybe it has something to do with how we now output the derp region to the console?

iroh-net/src/hp/magicsock/conn.rs
dignifiedquire: what an emotional journey

iroh-net/docs/derp_regions.md
```diff
@@ -0,0 +1,53 @@
...
+						    region_id: 1,
+						    url: format!("https://derp.iroh.network")
+						        .parse()
+						        .unwrap(),
+						    stun_only: false,
+						    stun_port: 3478,
+						    ipv4: UseIpv4::Some(std::net::Ipv4Addr::new(35, 175, 99, 113)),
+						    ipv6: UseIpv6::None,
+						    stun_test_ip: None,
+						}
```
dignifiedquire, line 23: sth is off in this formatting 😅

iroh-net/src/bin/derper.rs
```diff
@@ -167,7 +167,7 @@ struct Config {
     /// The UDP port on which to serve STUN. The listener is bound to the same IP (if any) as
     /// specified in the `addr` field. Defaults to [`DEFAULT_DERP_STUN_PORT`].
     stun_port: u16,
-    /// Certificate hostname. Defaults to [`DEFAULT_DERP_HOSTNAME`].
+    /// Certificate hostname. Defaults to [`DEFAULT_NA_DERP_HOSTNAME`].
```
dignifiedquire, line 170: we should probably just drop the default

iroh-net/src/defaults.rs
```diff
@@ -3,30 +3,38 @@ use std::collections::HashMap;
 
 use crate::hp::derp::{DerpMap, DerpNode, DerpRegion, UseIpv4, UseIpv6};
 
-/// Hostname of the default Derp.
-pub const DEFAULT_DERP_HOSTNAME: &str = "derp.iroh.network";
+/// Hostname of the default NA Derp.
+pub const DEFAULT_NA_DERP_HOSTNAME: &str = "derp.iroh.network";
```
dignifiedquire, line 7: @Arqu can we get this to be sth like `na1.` or similar?

iroh-net/src/defaults.rs
```diff
@@ -37,3 +45,26 @@ pub fn default_derp_region() -> DerpRegion {
...
+            .parse()
+            .unwrap(),
+        stun_only: false,
+        stun_port: DEFAULT_DERP_STUN_PORT,
+        ipv4: UseIpv4::Some(DEFAULT_EU_DERP_IPV4),
+        ipv6: UseIpv6::None,
+        stun_test_ip: None,
+    };
+    DerpRegion {
+        region_id: 2,
```
dignifiedquire, line 65: should probably be constants for the region ids

iroh-net/src/hp/derp/map.rs
```diff
@@ -43,10 +44,10 @@ impl DerpMap {
         dm.regions.insert(
             1,
```
dignifiedquire, line 45: this is wrong, this must be the right region as well

iroh-net/src/magic_endpoint.rs
```diff
@@ -292,12 +305,32 @@ impl MagicEndpoint {
     ///
     /// This updates the magic socket's *netmap* with these addresses, which are used as candidates
     /// when connecting to this peer (in addition to addresses obtained from a derp server).
+    ///
+    /// If no UDP addresses are added, and `derp_region` is `None`, it will errpr.
```
dignifiedquire, line 309: typo

iroh/src/commands/doctor.rs
```diff
@@ -600,6 +619,124 @@ async fn port_map(local_port: NonZeroU16, timeout: Duration) -> anyhow::Result<(
...
+    let mut fail = Vec::new();
+    while let Some(region_details) = set.join_next().await {
+        let region_details = region_details?;
+        if region_details.latency.is_some() {
+            success.push(region_details);
+        } else {
+            fail.push(region_details);
+        }
+    }
+    success.sort_by(|a, b| {
```
dignifiedquire, line 675: I think you can use https://doc.rust-lang.org/std/primitive.slice.html#method.sort_by_key here

b5:
At @ramfox's request I've run the test suite on windows, and run provide from a windows machine to a mac. While running the test suite I received two "allow this to happen" requests from the windows firewall, which I said yes to. Test suite passed, transfer worked.

iroh-net/src/magic_endpoint.rs
```diff
@@ -292,12 +305,32 @@ impl MagicEndpoint {
...
     pub async fn add_known_addrs(
         &self,
         peer_id: PeerId,
+        derp_region: Option<u16>,
         endpoints: &[SocketAddr],
     ) -> anyhow::Result<()> {
-        const DEFAULT_DERP_REGION: u16 = 1;
+        match (endpoints.is_empty(), derp_region) {
+            (true, None) => {
+                anyhow::bail!(
```
Frando, line 319: This errors if `derp_region` is `None` and `endpoints` is empty. `add_known_addrs` is currently always called by `connect`. Now if I call `connect` once with a derp region and/or a list of UDP addresses, those regions and addresses are added to the addressbook (netmap). If I then call `connect` a  second time with only a peer id, it will error even though the magicsock actually does know how to connect to the peer in question, because the addresses/regions from the first call are stored and remembered in the netmap.

This behavior is being used in #1149. When new peer info is received via a gossip `Join` or `ForwardJoin` message, this info (addresses and after this PR regions) is added to the endpoint with a direct call to `add_known_addrs`. If the gossip layer later on wants to connect to the peer, `connect` is called only with the peer id, without addresses or regions. This works, because the addresses/regions are stored in the netmap.

The nice thing about this is that we do not have to keep a separate registry of peer info in the gossip layer but instead can just use the addressbook of the magicsock.

These lines would change this behavior. Can we instead first check if the current netmap of the magicsock actually has any regions/addresses for the peerid and only error if this is empty as well?

Frando: We could also move all these checks into the conn actor, and pass only the new info via a `Conn::add_peer_info` and `ActorMessage::AddPeerInfo` or so that would replace `ActorMessage::SetNetworkMap`. And then we could add `ActorMessage::CanDialPeer(oneshot::Sender<anyhow::Result<()>>` or similar to check if a peer is dialable and return those warnings/errors. Maybe not for this PR. But it would be nice to centralize the netmap management and not keep a copy of the netmap in the MagicEndpoint. I think the current setup might be a little brittle.

iroh-net/src/hp/derp/http/mesh_clients.rs
```diff
@@ -112,10 +112,11 @@ mod tests {
             .try_init()
             .ok();
 
-        for i in 0..10 {
-            println!("TEST_MESH_NETWORK: round {i}");
-            test_mesh_network_once().await?;
-        }
+        // TODO(ramfox): figure out why this fails on later rounds
```
dignifiedquire, line 115: Can you file an issue and link it here, so we don't loose track of it

iroh-net/src/magic_endpoint.rs
dignifiedquire: > Can we instead first check if the current netmap of the magicsock actually has any regions/addresses for the peerid and only error if this is empty as well?

yeah, we probably should do that

dignifiedquire: Honestly we probably should not error out here ever, but actually just try to dial with whatever we have. But this will need more refinement and testing over time

dignifiedquire (approved):
Generally looks good to me, except the dialing restrictions, we should figure those out to make sure things still work.

iroh-net/src/bin/derper.rs
```diff
@@ -236,7 +236,7 @@ impl Default for Config {
             private_key: key::node::SecretKey::generate(),
             addr: "[::]:443".parse().unwrap(),
             stun_port: DEFAULT_DERP_STUN_PORT,
-            hostname: DEFAULT_DERP_HOSTNAME.into(),
+            hostname: NA_DERP_HOSTNAME.into(),
```
flub, line 239: lol, i read this as "Not Available derp hostname" initially...

iroh-net/src/hp/magicsock/conn.rs
```diff
@@ -206,14 +206,25 @@ pub struct Inner {
 }
 
 impl Inner {
+    /// Returns the derp region we are connected to, that has the best latency.
+    ///
+    /// If `0`, then we are not connected to any derp region.
```
flub, line 211: I would rather prefer an `Option` instead of using `0` as special value.

iroh-net/src/hp/magicsock/conn.rs
```diff
@@ -506,6 +517,18 @@ impl Conn {
         Ok(())
     }
 
+    /// Returns the DERP region with the best latency.
+    ///
+    /// If `None`, then we currently have no verified connection to a DERP node in any region.
+    pub async fn my_derp(&self) -> Option<u16> {
+        let my_derp = self.inner.my_derp();
+        if my_derp == 0 {
```
flub, line 525: oh, here you do this... why not in the inner?

iroh-net/src/defaults.rs
```diff
@@ -3,37 +3,73 @@ use std::collections::HashMap;
...
 use crate::hp::derp::{DerpMap, DerpNode, DerpRegion, UseIpv4, UseIpv6};
 
-/// Hostname of the default Derp.
-pub const DEFAULT_DERP_HOSTNAME: &str = "derp.iroh.network";
+/// Hostname of the default NA Derp.
+pub const NA_DERP_HOSTNAME: &str = "derp.iroh.network";
+/// IPv4 of the default NA Derp.
+pub const NA_DERP_IPV4: std::net::Ipv4Addr = std::net::Ipv4Addr::new(35, 175, 99, 113);
+/// NA region id
+pub const NA_REGION_ID: u16 = 1;
```
flub, line 11: We probably should put this in a `#[repr(u16)]` enum instead and use `as u16`

iroh-net/src/hp/netcheck.rs
```diff
@@ -63,7 +63,7 @@ pub struct Report {
     pub hair_pinning: Option<bool>,
     /// Probe indicating the presence of port mapping protocols on the LAN.
     pub portmap_probe: Option<portmapper::ProbeOutput>,
-    /// or 0 for unknown
+    /// `0` for unknown
```
flub, line 66: also would love to turn this into an option

iroh-net/src/hp/netcheck/reportgen/probes.rs
```diff
@@ -467,15 +471,21 @@ fn sort_regions<'a>(dm: &'a DerpMap, last: &Report) -> Vec<&'a DerpRegion> {
...
-        if da.is_none() {
-            // Zero can't sort before anything else.
-            return std::cmp::Ordering::Less;
+        match (da, db) {
+            (Some(_), None) => {
+                // Non-zero sorts before zero.
+                std::cmp::Ordering::Greater
+            }
+            (None, Some(_)) => {
+                // Zero can't sort before anything else.
```
flub, line 480:
```suggestion
                // Zero can't sort before anything else, but keep sort stable by region id.
```

iroh-net/src/hp/netcheck/reportgen/probes.rs
```diff
@@ -467,15 +471,21 @@ fn sort_regions<'a>(dm: &'a DerpMap, last: &Report) -> Vec<&'a DerpRegion> {
...
+                a.region_id.cmp(&b.region_id)
+            }
+            (None, None) => std::cmp::Ordering::Equal,
```
flub, line 483: should this also not use the region_id to guarantee an order?

iroh-net/src/hp/netcheck/reportgen/probes.rs
```diff
@@ -724,4 +674,213 @@ mod tests {
...
+            println!("{} round", i);
+            println!("expected:");
+            println!("{expected_plan}");
+            println!("actual:");
+            println!("{plan}");
+            // Just in case there's a bug in the Display impl:
+            assert_eq!(plan, expected_plan, "{}", i);
+
+            // The readable error:
+            assert_eq!(plan.to_string(), expected_plan.to_string(), "{}", i);
```
flub, line 883: these two asserts seem to have flipped around?  also not sure if the diff is now just entirely hidden by the round number or not.

iroh-net/src/magic_endpoint.rs
```diff
@@ -245,19 +245,32 @@ impl MagicEndpoint {
...
     /// Connect to a remote endpoint.
     ///
     /// The PeerId and the ALPN protocol are required. If you happen to know dialable addresses of
     /// the remote endpoint, they can be specified and will be used to try and establish a direct
     /// connection without involving a DERP server. If no addresses are specified, the endpoint
     /// will try to dial the peer through the configured DERP servers.
+    ///
+    /// If the `derp_region` is not `None` and the configured DERP servers do not include a DERP node from the given `derp_region`, it will error.
+    ///
+    /// If no UDP addresses and no DERP region is provided, it will error.
```
flub, line 264: time to make more type-safe signatures maybe.  but thanks for calling this out!

flub (approved):

iroh-net/docs/derp_regions.md
```diff
@@ -0,0 +1,61 @@
...
+
+A DERP region is a collection of DERP nodes in a relatively similar geographical area. All the DERP nodes in a single DERP region are connected to every other node in that region through a mesh network. They can all act as packet forwarders for each other. When an Iroh peer starts up, it does a latency test to see which region it is “closest to”. That region is considered the Iroh peer’s home region.
+
+A peer may be connected to multiple regions, but it will advertise its home region as the one best used to hole-punch or relay packets through.
+
+When attempting to connect to another peer, it is not required to pass in that peer’s DERP region, but if the other peer is behind a NAT, the only way to attempt to hole punch directly through to it is by passing down a DERP region. It is also possible that the DERP region that the remote Iroh peer is connected to is one that you have no dialing information about. However, that is unlikely to happen if you use the default regions and nodes we have specified here. These also exist as defaults in the code.
+
+We currently have 2 derp regions, with a single node in each.
+
+Region ID `65535` is reserved for testing and experiments.
```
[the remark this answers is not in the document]
flub: https://github.com/n0-computer/iroh/issues/1246

iroh-net/src/hp/derp/http/mesh_clients.rs
flub: https://github.com/n0-computer/iroh/issues/1247

iroh-net/src/magic_endpoint.rs
flub: https://github.com/n0-computer/iroh/issues/1248

iroh-net/src/defaults.rs
flub: https://github.com/n0-computer/iroh/issues/1249

iroh-net/src/hp/netcheck/reportgen/probes.rs
flub: https://github.com/n0-computer/iroh/pull/1250

iroh-net/src/hp/netcheck/reportgen/probes.rs
flub: no, that's bogus

PR #1223, opened 2023-07-13
Feature flagged db

rklaehn:
## Description

This is an alternative to https://github.com/n0-computer/iroh/pull/1217 that moves the 2 concrete database implementations into feature flagged submodules of iroh-bytes.

## Notes & open questions

I fear that this will end up a maze of optional dependencies as the database gets evolved. Also, this will make it hard to have the database do multiple things including things that are unrelated to iroh-bytes. E.g. you might want to store documents and authentication info in the database as well, but that does not really work well in iroh-bytes.

## Change checklist

- [x] Self-review.


rklaehn:
Closing this and moving the db to iroh as originally planned.

https://github.com/n0-computer/iroh/pull/1217#issuecomment-1633864155

PR #1209, opened 2023-07-10
[WIP] feat(iroh-net): PCP/NAT-PMP

divagant-martian:
## Description

Implement probing and mapping for PCP/NAT-PMP

## Notes & open questions

TBD

## Change checklist

- [ ] Self-review.
- [ ] Documentation updates if relevant.
- [ ] Tests if relevant.

PR #1225, opened 2023-07-13
chore(iroh-net): fix warnings and errors due to incompatible deps after underlying dep update

divagant-martian:
## Description

- Working on #1209, adding new deps resulted in a `Cargo.lock` update, which brings some incompatible versions between `default-net` and `rtnetlink` as well a a deprecation warning for `rustls`
- New feature flags are needed in `tokio`
- From Ci, also `ucd-parse` needs rust `1.70` or higher

## Notes & open questions

it would be great if `rtnetlink` re-exported the libs that are exposed in its public api

## Change checklist

- [x] Self-review.


flub (approved):
ugh, the msrv is unfortunate

Issue #1224, opened 2023-07-13
Remove iroh-net dependency from iroh-bytes


rklaehn:
Currently there is a dependency to iroh-net in iroh-bytes.

If you take a closer look you will find that iroh-bytes does not use anythin from iroh-net. There is just a helper function dial that has to know about things like DerpMap and PeerId.

The only iroh-bytes specific thing this function does is to set the ALPN. This can also be done in iroh, I think.

It is possible to use iroh-bytes with a normal, non derped/magic quinn connection. People that want to do that might not want to pull in iroh-net.

PR #1226, opened 2023-07-13
refactor: move dialing utils into iroh from iroh-bytes

rklaehn:
this allows iroh-bytes to not have a dependency on iroh-net

rationale: see https://github.com/n0-computer/iroh/issues/1224 fixes https://github.com/n0-computer/iroh/issues/1224

## Description

it seems that the tiny dialing util in iroh-bytes is not worth having a dep on iroh-net. So move it into iroh so we don't have to have this dependency.

## Notes & open questions

Rationale for this change, see #1224

## Change checklist

- [x] Self-review.


dignifiedquire (approved):
Cool, I am surprised this was the only dep we had on `iroh-net`, but very glad

dignifiedquire:
ci is sad, because of dependencies

PR #1228, opened 2023-07-13
ref(logging): Use nested tracing spans

flub:
## Description

This does a fair amount of logging-only changes.  Mostly it hooks up
the tracing spans to nest correctly.  This results in seeing all the
outer contexts when reading messages (at the cost of shifting plain
output to the right).

As a side effect logging the Conn name is not needed on many methods
since it is already present in the parent span.  So this removes a lot
of duplicate logging.

## Change checklist

- [x] Self-review.
- [ ] ~~Tests if relevant.~~


dignifiedquire (approved):
I accept..even though I really hate the `.instrument(..)` syntax..

flub:
> I accept..even though I really hate the `.instrument(..)` syntax..

That syntax does make the nesting work though, that's why I switched to it in places in favour of the proc-macro.

Maybe we need to introduce our own `spawn()` function that hides this.  It could give us some more niceties.

PR #1231, opened 2023-07-13
ref(derp): Default DNS name rooted at the DNS root

flub:
## Description

This makes sure that the default DNS name starts at the DNS root so
that DNS resolvers do not start searching their search domains.

## Change checklist

- [x] Self-review.
- [ ] ~~Tests if relevant.~~


flub:
Yes, that works.  HTTP just does DNS under the hood like anyone else.  Also, tests pass :)

dignifiedquire (approved):

divagant-martian:
this means the constant can be used in at least one more place I think

dignifiedquire:
@flub merge conflicts

PR #1232, opened 2023-07-13
tests: change some tests to no longer access the disk

rklaehn:
## Description

Now that it is reasonably easy to work with in memory databases, adapt tests to run in memory where possible.

We must still have tests that access disk, but tests that are entirely unrelated to this should not touch the file system.

Note that the tests still use the iroh (name, hash) collections, just not write to disk.

## Change checklist

- [x] Self-review.


flub (approved):
nice!

PR #1233, opened 2023-07-14
ci: cancel workflows when a new commit is pushed

rklaehn:
## Description

This should help with load on our test runners. Copied from https://stackoverflow.com/questions/66335225/how-to-cancel-previous-runs-in-the-pr-when-you-push-new-commitsupdate-the-curre

## Notes & open questions

How fast will this cancel the jobs on our own boxes? No idea how that works really...

## Change checklist

- [x] Self-review.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,10 @@ on:
     branches:
       - main
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 env:
   RUST_BACKTRACE: 1
   RUSTFLAGS: -Dwarnings


rklaehn:
Did a quick test, it seems to work with a small delay. Also, it triggers re-test if you remove some commits.

flub (approved):

Arqu:
Nice, the delay is expected, even manually canceling jobs takes a bit of time due to the architecture of GH runners. 
For those interested, there is no "push" mechanism to send updates to the runners, rather there's an endpoint that gets polled from the runners periodically. This means the poll loop if working correctly will take X amount of time to hit it again. Also under certain situations those runners get slower to poll (ie some heavy load/continuous tests) and then it takes a little bit longer.

Arqu:
Also let's just keep an eye out how this interacts with main and manually (comment based) runs.
ie multiple merged PRs on main or running CI for a PR + invoking `/netsim`

PR #1234, opened 2023-07-14
docs: Add docs to RPC protocol

rklaehn:
## Description

Just add some basic docs to the RPC protocol so it does not look so naked in docs.rs. Let's not bikeshed this too much - it is going to change a lot.

### DIFF
--- a/iroh/src/rpc_protocol.rs
+++ b/iroh/src/rpc_protocol.rs
@@ -28,8 +28,9 @@ pub use iroh_bytes::provider::{ProvideProgress, ValidateProgress};
 pub struct ProvideRequest {
     /// The path to the data to provide.
     ///
-    /// This should be an absolute path.
-    /// The node and the cli should be on the same filesystem.
+    /// This should be an absolute path valid for the file system on which
+    /// the node runs. Usually the cli will run on the same machine as the
+    /// node, so this should be an absolute path on the cli machine.
     pub path: PathBuf,
 }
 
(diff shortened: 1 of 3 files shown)


iroh/src/rpc_protocol.rs
```diff
@@ -13,8 +21,15 @@ use serde::{Deserialize, Serialize};
...
 
+/// A request to the node to provide the data at the given path
+///
+/// Will produce a stream of [`ProvideProgress`] messages.
 #[derive(Debug, Serialize, Deserialize)]
 pub struct ProvideRequest {
+    /// The path to the data to provide.
+    ///
+    /// This should be an absolute path.
+    /// The node and the cli should be on the same filesystem.
```
flub, line 32: nitpicking, but does the cli access this path?  is it not sufficient that this path exists on the node?

iroh/src/rpc_protocol.rs
```diff
@@ -55,14 +76,25 @@ impl ServerStreamingMsg<ProviderService> for ListBlobsRequest {
     type Response = ListBlobsResponse;
 }
 
+/// List all collections
+///
+///
```
flub, line 81: ...?  so much suspense!

iroh/src/rpc_protocol.rs
```diff
@@ -73,53 +105,67 @@ impl ServerStreamingMsg<ProviderService> for ListCollectionsRequest {
...
 pub struct ShutdownRequest {
+    /// Force shutdown
     pub force: bool,
 }
 
 impl RpcMsg<ProviderService> for ShutdownRequest {
     type Response = ();
 }
 
+/// A request to get the id of the node
```
flub, line 131: is this the PeerId?

flub (approved):

iroh/src/rpc_protocol.rs
rklaehn: I hate this call, that is why. Not sure what to write here that makes sense since the whole call does not make that much sense.

Issue #1236, opened 2023-07-14
Allow having tickets for single blobs


rklaehn:
Currenty tickets imply downloading a collection and all its children.

We should extend tickets to allow downloading single blobs.

rklaehn:
this is done, but I am still not entirely happy with the cil commands. To be continued in https://github.com/n0-computer/iroh/issues/1239

PR #1237, opened 2023-07-14
docs: Iroh examples

rklaehn:
## Description

Add a minimal example to spin up an iroh node and serve a single blob.
Also an example to share a collection from memory.

Do examples count as docs?

## Change checklist

- [x] Self-review.

Implements https://github.com/n0-computer/iroh/issues/1236


iroh/examples/collection/src/main.rs
```diff
@@ -0,0 +1,63 @@
+//! An example that serves an iroh collection from memory.
```
b5, line 1: I've added suggested changes to this file in a branch: https://github.com/n0-computer/iroh/tree/b5/iroh-examples

PTAL, and feel free to cherry-pick 331ea3a991c6eb35a25d60498424e1e717ab4e88

b5 (approved):
After some small patches, LGTM!

Issue #1240, opened 2023-07-14
Write get examples


rklaehn:
We now https://github.com/n0-computer/iroh/pull/1237 have examples for the provider. What is missing are examples how to get data.

When writing these examples, it might also make sense to add a few more utility methods in iroh_bytes::get. There are already some utility methods in the tests and on the fsm states, but it would probably be a good idea to have something more discoverable.

Examples should include:

- stream blob from a source
- stream blob range from source
- blob to file?
- collection to files

rklaehn:
To be followed up with https://github.com/n0-computer/iroh/issues/1265

PR #1242, opened 2023-07-17
More get examples

rklaehn:
## Description

This provides more examples for how to get data.

We now have:
- get blob by manually driving the FSM
- get blob using a helper method that converts the FSM into a `impl Stream<io::Result<Bytes>>`
- get iroh collection children using a helper method that converts the FSM into a `impl Stream<io::Result<Bytes>>`

## Notes & open questions

Which one of the helper methods is worth going into some porcelain feature of iroh-bytes?

I think you can see why providing porcellain is not that easy. There are so many different ways to wrap the FSM, and even
if you provide a dozen different (e.g. AsyncWrite, Stream<Bytes>, AsyncRead, with different callbacks) anybody with a less than trivial use case will probably have to use the fsm directly.

What might work is to provide porcelain just for the "download a blob" use case, and good examples for using the FSM for the "download a collection" use case.

## Change checklist

- [x] Self-review.


dignifiedquire (approved):

rklaehn:
Implements https://github.com/n0-computer/iroh/issues/1240

PR #1244, opened 2023-07-17
ref(reportgen): Improve logging to be less confusing

flub:
## Description

Logging a generic error in the .send() method is pretty useless for
tracing the error.  So stop doing that and propagate the error up,
handle it at all the callers.

Generally also improve log messages a bit and fix up some forgotten
renaming.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh-net/src/hp/netcheck/reportgen.rs
```diff
@@ -721,9 +729,16 @@ async fn run_probe(
             would_help_tx,
         ))
         .await
+        .map_err(|err| {
+            error!("Failed to check of probe would help: {err:#}");
```
divagant-martian, line 733:
```suggestion
            error!("Failed to check if probe would help: {err:#}");
```

dignifiedquire (approved):

PR #1245, opened 2023-07-17
Flat examples

rklaehn:
## Description

This removes the clippy test for examples when running with --no-default-features. Examples usually use some features, and without this the examples have to be their own crates, which I think is not so nice.

## Notes & open questions

Note: there might still be some examples that will have to have their own crate, but I would like to have small examples that are just one file.

## Change checklist

- [x] Self-review.


dignifiedquire:
The examples should be listed in `iroh/Cargo.toml` with `required-features` set to the features they need

dignifiedquire (approved):

PR #1250, opened 2023-07-17
fix(netcheck): Stable derp-region sorting

flub:
## Description

Make sure derp-regions are always sorted in the same way by using the
region id if we don't have any relevant latencies.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


dignifiedquire:
can you please add a testcase to make sure this now covers all cases?

flub:
lol, **of course** we still got it wrong.  now with tests.  and maybe correct?

dignifiedquire (approved):

PR #1251, opened 2023-07-17
feat: disable bailing out when temp dir is missing

rklaehn:
## Description

Note that this will lead to the download being repeated from scratch even if the data is already there. We can change that
later. Basically once you get the collection you would have to repeat the step of trying to figure out what is missing, and
then send another request.

## Notes & open questions

- Is this behaviour OK?
- we should probably rethink single blob downloads. They were added as an afterthought, but now it seems they are being used more often.

## Change checklist

- [x] Self-review.


dignifiedquire (approved):
I like it

PR #1260, opened 2023-07-17
feat: add metrics to the derp server

### LINKED ISSUE #926: implement metrics for derp

### LINKED ISSUE #1135: documented example for configuring iroh to use a local derper
We need a documented example setup for running a local derper with either self-signed certs or without https. This is a nontrivial thing to do at the moment, and requires careful `derper` flag composition to avoid conflicting configuration.

The example should show the reader how to have all parts of iroh running locally, including the derp server. ~~bonus points if it can be expressed as a rust example~~ nm, can't (reasonably) be done as a rust example

ramfox:
## Description

closes #926 
closes #1135 

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh-metrics/src/derpserver.rs
+++ b/iroh-metrics/src/derpserver.rs
@@ -14,18 +14,18 @@ pub struct Metrics {
     /// Bytes received from a `FrameType::SendPacket`
     pub bytes_recv: Counter,
 
-    /// `FrameType::SendPacket` sent
+    /// `FrameType::SendPacket` sent, that are not disco messages
     pub send_packets_sent: Counter,
-    /// `FrameType::SendPacket` received
+    /// `FrameType::SendPacket` received, that are not disco messages
     pub send_packets_recv: Counter,
-    /// `FrameType::SendPacket` dropped
+    /// `FrameType::SendPacket` dropped, that are not disco messages
     pub send_packets_dropped: Counter,
 
     /// `FrameType::SendPacket` sent that are disco messages
     pub disco_packets_sent: Counter,
     /// `FrameType::SendPacket` received that are disco messages
     pub disco_packets_recv: Counter,
-    /// `FrameType::SendPacket` dropped that are also disco messages
+    /// `FrameType::SendPacket` dropped that are disco messages
     pub disco_packets_dropped: Counter,
 
     /// Packets of other `FrameType`s sent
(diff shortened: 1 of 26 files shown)


iroh-metrics/src/derpserver.rs
```diff
@@ -0,0 +1,130 @@
+use struct_iterable::Iterable;
+
+use crate::core::{Counter, Metric};
+
+/// Metrics tracked for the DERP server
+#[allow(missing_docs)]
+#[derive(Debug, Clone, Iterable)]
+pub struct Metrics {
```
dignifiedquire, line 8: This can now just live in the `iroh-net` crate, since the metrics refactor.

iroh-net/src/hp/derp/clients.rs
```diff
@@ -63,19 +65,45 @@ impl Client {
...
+        }
+        res
     }
 
     pub fn send_disco_packet(&self, packet: Packet) -> Result<(), SendError> {
-        try_send(&self.conn.client_channels.disco_send_queue, packet)
+        let res = try_send(&self.conn.client_channels.disco_send_queue, packet);
+        if res.is_ok() {
+            inc!(Metrics, disco_packets_sent);
+        }
```
dignifiedquire, line 79: why not mark as `dropped`

iroh-net/src/hp/derp/clients.rs
```diff
@@ -63,19 +65,45 @@ impl Client {
     }
 
     pub fn send_packet(&self, packet: Packet) -> Result<(), SendError> {
-        try_send(&self.conn.client_channels.send_queue, packet)
+        let res = try_send(&self.conn.client_channels.send_queue, packet);
+        if res.is_ok() {
+            inc!(Metrics, send_packets_sent);
+        }
```
dignifiedquire, line 71: why not mark as `dropped`

ramfox: I'll leave a comment. If we cannot send the packet here, it's still possible that we are able to forward it.

iroh-net/src/hp/derp/clients.rs
ramfox: left a comment explaining that this is not the last time we attempt to send the packet, so it isn't considered "dropped" yet

Issue #1264, opened 2023-07-18
Setup CI for ffi bindings


dignifiedquire:
In order to test and make sure the bindings are actually working CI needs to test them. 

First round is to test 

- [x] `iroh-ffi` 
- [ ] swift bindings

Ref https://github.com/n0-computer/iroh/pull/948

dignifiedquire:
this is moved to https://github.com/n0-computer/iroh-ffi/

PR #1270, opened 2023-07-18
chore(ci): add ffi bindings to CI

Arqu:
## Description

Getting started on adding ffi bindings to CI.
Ref https://github.com/n0-computer/iroh/issues/1264

## Notes & open questions

What platforms should we cover, do we just target macOS for now?
Is this even correct?

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.


.github/workflows/bindings.yml
```diff
@@ -0,0 +1,40 @@
...
+  pull_request:
+  push:
+    branches:
+      - main
+
+env:
+  RUST_BACKTRACE: 1
+  RUSTFLAGS: -Dwarnings
+  RUSTDOCFLAGS: -Dwarnings
+  MSRV: "1.70"
```
dignifiedquire, line 13: current MSRV is `1.66`

Arqu: Ahh, nice, missed that PR.

.github/workflows/bindings.yml
```diff
@@ -0,0 +1,71 @@
...
+  MSRV: "1.66"
+
+jobs:
+  build_and_test_ffi_bindings:
+    name: Build and test ffi bindings
+    runs-on: ${{ matrix.name }}
+    strategy:
+      fail-fast: false
+      matrix:
+        name: [macOS-latest]
```
Arqu, line 22: should we cover more platforms or is macOS just fine for now?

.github/workflows/bindings.yml
```diff
@@ -0,0 +1,71 @@
...
+
+      - name: Install ${{ matrix.rust }}
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ matrix.rust }}
+
+      - name: Create a copy of iroh.h
+        run: |
+          cd iroh-ffi
+          cp iroh.h iroh.compare.h
```
dignifiedquire, line 38: alternatively you can use git diff

Arqu: Eh, since `iroh.h` is in `.gitignore` it doesn't seem to diff it. Also think cmp is slightly faster.

PR #1271, opened 2023-07-18
test(iroh): only use flat db when enabled

dignifiedquire:


rklaehn (approved):
Yeah, the places where the flat db is used in tests are getting less and less

PR #1272, opened 2023-07-18
fix(iroh): atomically write keyfile

dignifiedquire:
@rklaehn saw some data corruption when reading the key from disk, this should avoid that.


iroh/src/commands/provide.rs
```diff
@@ -164,15 +165,30 @@ async fn get_keypair(key: Option<PathBuf>) -> Result<Keypair> {
...
                 Ok(keypair)
             } else {
                 let keypair = Keypair::generate();
                 let ser_key = keypair.to_openssh()?;
                 if let Some(parent) = key_path.parent() {
                     tokio::fs::create_dir_all(parent).await?;
                 }
-                tokio::fs::write(key_path, ser_key).await?;
+                // write to tempfile
+                let (file, temp_file_path) = tempfile::NamedTempFile::new()
```
flub, line 177: I think you need `NamedTempFile::new_in()` to put it in the same directory as the destination.  Otherwise the tempfile could be on a different filesystem and copying is still not atomic.

flub (changes requested):

iroh/src/commands/provide.rs
dignifiedquire: fixed

iroh/src/commands/provide.rs
```diff
@@ -164,15 +165,35 @@ async fn get_keypair(key: Option<PathBuf>) -> Result<Keypair> {
...
             } else {
                 let keypair = Keypair::generate();
                 let ser_key = keypair.to_openssh()?;
-                if let Some(parent) = key_path.parent() {
-                    tokio::fs::create_dir_all(parent).await?;
-                }
-                tokio::fs::write(key_path, ser_key).await?;
+
+                // Try to canoncialize if possible
+                let key_path = key_path.canonicalize().unwrap_or(key_path);
```
flub, line 175: I don't think this is really needed.  The directory could be a symlink or something but in either case `.parent().join()` should give you something in the same directory.

Probably doesn't do any harm either?  So I don't mind if you'd like to keep it.

dignifiedquire: I am not sure, wanted to be on the safe side with things like `"."`

dignifiedquire:
MSRV failure is unrelated

rklaehn (approved):

dignifiedquire:
> How big is the keyfile? I think tiny writes (below 512 bytes) are atomic on most file systems.

387 bytes on my system

PR #1280, opened 2023-07-19
feat(tests): Improve test_utils to warn about mutli-runtime tests

flub:
## Description

This detects if setup_logging() is going to blow up in your face and warns
against it.

Also adds closed spans to the log output.

## Notes & open questions

~~It works this time?  I'm as surprised as anything.~~

It did not work, reverted this to be just the other improvements of the logging stuff under test.

## Change checklist

- [x] Self-review.
- [ ] ~~Tests if relevant.~~


iroh-net/src/magicsock.rs
```diff
@@ -3085,48 +3085,48 @@ pub(crate) mod tests {
...
-        };
-
-        for _ in 0..10 {
-            let (derp_map, _, cleanup) = run_derp_and_stun(devices.stun_ip).await?;
-            println!("setting up magic stack");
-            let m1 = MagicStack::new(derp_map.clone()).await?;
-            let m2 = MagicStack::new(derp_map.clone()).await?;
-
-            let cleanup_mesh = mesh_stacks(vec![m1.clone(), m2.clone()]).await?;
+        crate::test_utils::with_logging(async move {
```
flub, line 3088: All the test code here is untouched, only wrapped in this function call and async block.

iroh-net/src/test_utils.rs
```diff
@@ -1,38 +1,98 @@
...
+///    }.with_subscriber(testing_subscriber()).await
+/// }
+/// ```
+pub(crate) fn testing_subscriber() -> impl tracing::Subscriber {
     let var = std::env::var_os("RUST_LOG");
     let trace_log_layer = match var {
         Some(_) => None,
         None => Some(
             tracing_subscriber::fmt::layer()
+                .with_span_events(FmtSpan::CLOSE)
```
dignifiedquire, line 82: what does this do?

flub: It also shows a log line when a span is closed.  I find this really nice as you very clearly see things finish.

rklaehn: Every thing that makes spans more readable is welcome. By the way - is there a way to use RUST_LOG to get rid of the span info in each log line? I find it a bit overwhelming sometimes...

```
2023-07-19T12:47:00.734742Z DEBUG conn{name=magic-78c000c451139ad8}:derp.actor:recv_detail:client-connect{key=PublicKey(78c000c451139ad8a8a77e9b5fe752b18320a9c551c0959ee71a85f24c82de15)}: rustls::client::hs: Using ciphersuite TLS13_AES_256_GCM_SHA384    
```

flub: though often you only have to go scan all the way to the right for the few things you care about.  that's a little mitigating i guess.

flub:
@dignifiedquire @rklaehn could I get either an LGTM or more changes requested?

iroh-net/src/test_utils.rs
```diff
@@ -1,38 +1,98 @@
...
 #[must_use = "The tracing guard must only be dropped at the end of the test"]
 #[allow(dead_code)]
 pub(crate) fn setup_logging() -> tracing::subscriber::DefaultGuard {
+    if let Ok(handle) = tokio::runtime::Handle::try_current() {
+        match handle.runtime_flavor() {
+            RuntimeFlavor::CurrentThread => (),
+            RuntimeFlavor::MultiThread => {
+                panic!(
+                    "setup_logging() does not work in a multi-threaded tokio runtime, \
+                        use with_logging() instead"
```
dignifiedquire, line 38: `with_logging` is commented out?

flub: yes, it doesn't want to work :(  needs more debugging

dignifiedquire: then maybe change the panic message, it's a bit confusing atm 😅

flub: done

dignifiedquire (approved):

Issue #1183, opened 2023-07-07
reenable flaky tests in magicendpoint


dignifiedquire:

matheus23:
As far as I can tell, these tests are now enabled again:
https://github.com/n0-computer/iroh/blob/1d3f3fa8125761df509b94621eb173b91ed42904/iroh-net/src/endpoint.rs#L1161-L1164

PR #1282, opened 2023-07-19
fix(tests): bring back MagicEndpoint connect-close test

flub:
## Description

This brings back the MagicEndpoint connect-close test and attempts to
make it more reliable.

- Most importantly the client sends one message to the server before
  the server closes the connection.  This is a crucial bit of
  synchronisation for the interaction to be deterministic.

- Both client and server run in a task again.  Keeps things a bit more
  local.

- The endpoints are now created in the spawned tasks, this ensures we
  get nice span prefixes on all the logging for each thus
  distinguishing the log output.

- The run_derp_and_stun function is moved into test_utils.  Maybe it
  belongs better there?

- The run_derp_and_stun function now returns a drop guard for cleanup.
  This ensures that whatever happens to the test, regardless of
  assertions, the cleanup will run.  It does mean for a successful
  test the test will no longer delay exiting before the cleanup is
  finished (was already not the case for the STUN server).  This is
  fine though, the servers both bind to 127.0.0.1:0 so do not affect
  anything else while they are still shutting down.

- The run_derp_and_stun function now disables IPv6 in the returned
  DerpMap.  Previously it would trigger a DNS lookup.  But we only
  bind on IPv4 so skip this noise.

- A couple of minor logging output adjustments.

Part of #1183.

If you're reading this commit message because you're looking the
test added here having just failed.  Let me know!  I'd love to
debug it further.

## Notes & open questions

This may not fix it, we can't know until we try.  But if it doesn't it will have good logging to help me further.  Which is why I'm cautiously just submitting this one test and would like to see it on CI for a while.  

## Change checklist

- [x] Self-review.
- [ ] ~~Tests if relevant.~~


iroh-net/src/magic_endpoint.rs
```diff
@@ -432,153 +432,172 @@ pub async fn get_peer_id(connection: &quinn::Connection) -> anyhow::Result<PeerI
...
-
-//         p2_accept.await.unwrap();
-//         p1_connect.await.unwrap();
-
-//         cleanup().await;
-//     }
-// }
+#[cfg(test)]
+mod tests {
+    // use futures::future::BoxFuture;
```
dignifiedquire, line 437: remove?

iroh-net/src/test_utils.rs
```diff
@@ -54,3 +60,68 @@ impl std::io::Write for TestWriter {
...
+/// The returned `oneshot::Sender<()>` is a drop guard: the server will be shut down when
+/// dropped.  You could also explicitly send a value to achieve the same if you so desire.
+///
+/// The returned `u16` is the region ID of the DERP server in the returned [`DerpMap`], it
+/// is always `Some` as that is how the [`MagicEndpoint::connect`] API expects it.
+///
+/// [`MagicEndpoint::connect`]: crate::magic_endpoint::MagicEndpoint
+pub async fn run_derp_and_stun(
+    stun_ip: IpAddr,
+) -> Result<(DerpMap, Option<u16>, oneshot::Sender<()>)> {
```
dignifiedquire, line 75: maybe make the oneshot an actual dropguard?

iroh-net/src/magic_endpoint.rs
flub: it's part of the still-commented-out other test.  but sure I could remove it.

iroh-net/src/magic_endpoint.rs
```diff
@@ -432,153 +432,172 @@ pub async fn get_peer_id(connection: &quinn::Connection) -> anyhow::Result<PeerI
 
 // TODO: Reenable when not flaky
```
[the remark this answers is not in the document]
flub: well, we don't know if the test isn't flaky anymore :)  plus this still applies to the other test...

iroh-net/src/test_utils.rs
flub: it crossed my mind.  though i don't know what it would really gain other than a newtype.

maybe a nice version might be a type that's a dropguard as well as has a `.cleanup().await` that waits for tidy completion if you really want.  though i'm again wondering what the benefit is, each test gets a separate runtime anyway so as long as this cleanup is executed it's fine, even if it runs after the test.

so i really think it would be just a newtype, and i didn't really consider that worth it for a test.  but i guess we can do this.

flub:
ptal

PR #1292, opened 2023-07-24
fix(iroh-net): Stop quinn from erroring when closing a connection

flub:
## Description

Quinn spawns a driver task for each endpoint.  This task only terminates when the endpoint is completely dropped.  Crucially however at this point it will still poll the underlying socket once more and if the socket errors at this time it will report the error.  Only when the driver task terminates does it drop the socket, which must free it's resources on drop.

This implementation does not do anything to make the MagicSocket shut down cleanly on drop: dropping it drops the AbortingTaskHandles, in turn aborting the associated tasks of the MagicSocket.

## Notes & open questions

There are some options to shut down the `MagicSock` gracefully:

- The simplest is to store the quinn endpoint in an Option in the `MagicEndpoint`.  It allows you to drop it before closing the `MagicSock`.  However you're still left with not knowing when the quinn driver really has finished so are still guessing at when to close it.
- We could implement `Drop` on the `MagicSock`.  However drop is synchronous so the current `close` implementation would only do the sync part and signal the `MagicSock` actor to shutdown without waiting, which is totally fine for a `Drop`.  The main problem here is that the `MagicSock` actor inbox is limited so when sending a message sync it is not guaranteed to succeed.  If that's the case then the `MagicSock` tasks would just never terminate.

If we'd just take the current implementation you still don't know when the `MagicSock` is really done.  We could add a method which would return a future which would complete when the `MagicSock`, and thus `MagicEndpoint` is really done.

I guess one question for which I don't know things well enough for is: do we gain anything by doing a graceful shutdown of `MagicSock`?

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


dignifiedquire:
> I guess one question for which I don't know things well enough for is: do we gain anything by doing a graceful shutdown of `MagicSock`?

Given that `iroh` will be used as a library, it should be possible to shut it down cleanly, so when starting a new one, nothing lingers. It is very unfortunate that `quinn` does not expose a clean way to do so, maybe we can ask them if they would consider adding an API for this? 

 I do not think we need it to be restartable. I think API wise, doing it on `async close` is the best solution, to avoid having to push potentially locking/blocking work into a `drop` call.

dignifiedquire:
@flub should we close this for now?

flub:
No, I would like to keep it open.  I do think it matters to get expected behaviour, we did have flaky tests that were impossible to resolve without this so I think other real uses might suffer this too.

I'll get back to this once I'm back from iroh.network land.

PR #1295, opened 2023-07-24
deps: switch back to derive_more og

dignifiedquire:
A `1.0.0-beta.1` was just released, so we don't need to use the forked version anymore

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -173,13 +173,13 @@ dependencies = [
 
 [[package]]
 name = "async-trait"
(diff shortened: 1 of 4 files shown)


flub (approved):
Sneaking in a full upgrade of all dependencies I see... :)
Maybe mention that in the commit message as well.

dignifiedquire:
> Sneaking in a full upgrade of all dependencies I see... :)

It's just the lockfile, so it doesn't actually upgrade anything for library users :P

PR #1299, opened 2023-07-26
refactor(iroh-net): Always attach tracing spans to spawned tasks

flub:
## Description

This is another commit that keeps fine-tuning our logging.

- Most importantly this consistently attaches spans when tasks are
  spawned.  This threads togehther a lot more context.

- As a general rule spawned tasks get info_span!(), otther spans get
  debug_span!()

- protmapper report is logged on info in reportgen now, just as the
  other sub-reports.

- MagicSock::poll_recv logs number of datagrams on trace now instead
  of info.  This is expected to be a lot.  A new metric is added to
  keep track of this as well.

## Change checklist

- [x] Self-review.
- [ ] ~~Tests if relevant.~~

### DIFF
--- a/iroh-net/src/bin/derper.rs
+++ b/iroh-net/src/bin/derper.rs
@@ -31,7 +31,7 @@ use reqwest::Url;
 use serde::{Deserialize, Serialize};
 use tokio::net::{TcpListener, UdpSocket};
 use tokio_rustls_acme::{caches::DirCache, AcmeConfig};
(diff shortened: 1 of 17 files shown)


dignifiedquire (approved):
very nice

Arqu (approved):
Since we're starting to commit to it, I like it. I'll take a look on how to pick up these logs and digest on the infra side.

PR #1302, opened 2023-07-26
example(sync): add watch command

dignifiedquire:
Adds a simple watch command to show notifications when changes come in 
Screenshot 2023-07-26 at 20 59 05


Frando:
Very nice, thank you

PR #1305, opened 2023-07-27
fix tests in iroh sync and some cleanups

dignifiedquire:


iroh-sync/src/sync.rs
```diff
@@ -456,11 +456,20 @@ impl Replica {
...
         inner.peer.put(id, signed_entry.clone());
-        for cb in &inner.on_insert {
+        drop(inner);
+        let on_insert = self.on_insert.read();
+        for cb in &*on_insert {
             cb(InsertOrigin::Local, signed_entry.clone());
         }
     }
 
+    /// Hashes the given data and inserts it.
```
Frando, line 466: Maybe make clear in the comment that the data is not stored/inserted anyhwere, only its hash. Also, maybe return the hash.

Frando (approved):

Frando:
If you could split the commits between iroh-gossip and the rest that would ease splitting this between #1149  and #1127 later on. But also not critically, would also make as-is and rewrite later.

iroh-sync/src/sync.rs
dignifiedquire: done

PR #1306, opened 2023-07-27
feat: metrics for iroh-gossip and iroh-sync

Frando:
## Description

This PR adds metrics for iroh-gossip and iroh-sync. The commit on iroh-gossip should move to #1149 but this will happen not today to keep things easy for the demo.

Also adds a `stats` command to the repl to display current metrics.

foo

## Change checklist

- [x] Self-review.
- [ ] ~~Tests if relevant.~~


dignifiedquire (approved):
nice

PR #1307, opened 2023-07-27
fix: remove usage of unbounded channels

dignifiedquire:
uses flume channels to allow for combined sync and async usage


Frando:
Thanks!

One question though: `to_actor_tx` now is a bounded channel. `Downloader::start_download` is a sync fn that calls `to_actor_tx.send()`, a sync method that will potentially block synchronously if the channel is full. `Downloader::start_download` is called from `Replica::on_insert`, also all sync. Doesn't this mean that, if calling `Replica::insert` from an async context (which is the case currently usually), this will block the async executor if the channel is full?

PR #1309, opened 2023-07-27
sync: implement more extensive fetch methods

dignifiedquire:


Frando:
This is merged now, right?

Frando:
yup, it is! weird why github doesn't pick it up

PR #1312, opened 2023-07-27
feat: hammer sync example

Arqu:
## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/examples/sync.rs
+++ b/iroh/examples/sync.rs
@@ -413,7 +413,7 @@ async fn handle_fs_command(cmd: FsCmd, doc: &Doc) -> anyhow::Result<()> {
             dir_path,
             mut key_prefix,
         } => {
-            if key_prefix.ends_with("/") {
+            if key_prefix.ends_with('/') {
                 key_prefix.pop();
             }
             let root = canonicalize_path(&dir_path)?.canonicalize()?;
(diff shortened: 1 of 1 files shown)


Arqu:
Dont merge yet, working on some extensions

dignifiedquire:
clippy sad

iroh/examples/sync.rs
```diff
@@ -502,6 +546,17 @@ pub enum Cmd {
     WatchCancel,
     /// Show stats about the current session
     Stats,
+    /// Stress test with the hammer
+    Hammer {
```
Frando, line 550: Now that we're at four arguments, I'd prefer to turn them into options. Then you can also add default values if needed.
`#[clap(short, long)]` etc

iroh/examples/sync.rs
```diff
@@ -349,6 +349,50 @@ async fn handle_command(
...
+        } => {
+            println!(
+                "> Hammering with prefix {prefix} for {threads} x {count} messages of size {size} bytes",
+                prefix = prefix,
+                threads = threads,
+                count = count,
+                size = size,
+            );
+            let mut bytes = vec![0; size];
+            bytes.fill(97);
```
Frando, line 366: If we also want to test content downloading with this (do we?) maybe let's set their content to an increasing number or add an option for that behavior. Can also come later though.

Issue #1285, opened 2023-07-20
Panic somewhere deep in iroh_net


rklaehn:
While playing around with iroh I encountered this panic (by the way: we should abort the process on panic instead of getting in some weird state):

```
thread 'tokio-runtime-worker' panicked at 'error -1', iroh-net/src/net/interfaces/bsd.rs:501:13
```

Not sure if this is fatal or not, but it comes from here:

```rust
        if err != 0 {
            // If the sysctl failed because the data got larger
            // between the two sysctl calls, try a few times
            // before failing. (golang.org/issue/45736).
            const MAX_TRIES: usize = 3;
            if err == libc::ENOMEM && round < MAX_TRIES {
                continue;
            }
            panic!("error {err}");
            // return nil, os.NewSyscallError("sysctl", err);
        }
```

The fn returns a RouteError, so I guess you could just return a special RouteError instead of panicking.

PR #1313, opened 2023-07-27
fix(iroh-net): do not panic on RIB issues

dignifiedquire:
## Description

Fixes the panic we have seen in `bsd.rs` and should allow at least see the issue in logging, if there are more problems.

## References
- https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/sysctl.3.html
- https://doc.rust-lang.org/std/io/struct.Error.html#method.last_os_error

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh-net/src/net/interfaces/bsd.rs
+++ b/iroh-net/src/net/interfaces/bsd.rs
@@ -8,6 +8,7 @@ use std::{
 };
 
 use once_cell::sync::Lazy;
+use tracing::warn;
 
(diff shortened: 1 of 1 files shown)


flub (approved):

rklaehn (approved):

rklaehn:
Fixes https://github.com/n0-computer/iroh/issues/1285

PR #1320, opened 2023-08-01
Persistence rewrite

rklaehn:
## Description

This completely rewrites the persistence and enables mutable databases, including persistence of partially downloaded items.

Basically where before there was a trait `BaoMap` with the corresponding entry type `BaoMapEntry`, we now also have `BaoPartialMap` and `BaoPartialMapEntry`. These deal with pairs(*) of partially downloaded data and outboards.

In addition to the flat database, there is also a fully mutable memory database. The existing memory database has been retained but renamed to test database.

The lifecycle of data is roughly this: When downloading a blob, a temp entry is created or looked up. Then as we download we write to this temp entry (data and optional outboard) until it is complete. Then it is moved to a complete entry as atomically as possible. In the mem db, this can of course be done really atomically. In the flat db this is not possible because you can't atomically move/rename two files.

## Notes & open questions

- @b5 do we need conversion tools from the existing flat file format?

## Change checklist

- [x] Rip out old fs resume code and replace it with a flat db in a temp dir
- [x] Self-review.
- [x] Tests if relevant.


Frando (commented):
I did a first read-over, and the design looks very solid to me! I mostly left a couple of naming comments for now.

iroh-bytes/src/util/progress.rs
```diff
@@ -0,0 +1,421 @@
...
+    }
+}
+
+/// A conveneince type for sending progress messages.
+pub struct TokioProgressSender<T> {
+    sender: tokio::sync::mpsc::Sender<T>,
+    id: std::sync::Arc<std::sync::atomic::AtomicU64>,
+}
+
+impl<T> std::fmt::Debug for TokioProgressSender<T> {
```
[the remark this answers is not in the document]
rklaehn: No, this would require T to be Debug, but we don't really care about T in the Debug for this thing.

iroh/examples/hello-world.rs
```diff
@@ -22,7 +21,7 @@ pub fn setup_logging() {
 async fn main() -> anyhow::Result<()> {
     setup_logging();
     // create a new, empty in memory database
-    let mut db = mem::Database::default();
+    let mut db = iroh::database::test::Database::default();
```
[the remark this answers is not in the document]
rklaehn: Yeah, not sure about the name.

So basically this is the most simple database you can imagine that can be plugged into a node to serve stuff. The downside of this thing compared to the mem db is that it is immutable once shared, so you can only serve static data. But I still like it and want to keep it around.

It is currently mostly used in testing, hence the name. If you have a better name I am all ears...

iroh-bytes/src/provider.rs
```diff
@@ -52,9 +66,44 @@ pub trait BaoMap: Clone + Send + Sync + 'static {
...
     /// is an entry. Creating an entry should be cheap, any expensive ops should
     /// be deferred to the creation of the actual readers.
+    ///
+    /// It is not guaranteed that the entry is complete. A BaoMapMut would return
+    /// here both complete and partial entries, so that you can share partial entries.
     fn get(&self, hash: &Hash) -> Option<Self::Entry>;
 }
 
+/// A partial entry
+pub trait BaoPartialMapEntry<D: BaoPartialMap>: BaoMapEntry<D> {
```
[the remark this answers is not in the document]
rklaehn: I might try that and see how I like it.

iroh-bytes/src/provider.rs
```diff
@@ -52,9 +66,44 @@ pub trait BaoMap: Clone + Send + Sync + 'static {
...
+    /// A future that resolves to an writeable outboard
+    fn outboard_mut(&self) -> BoxFuture<'_, io::Result<D::OutboardMut>>;
+    /// A future that resolves to a writer that can be used to write the data
+    fn data_writer(&self) -> BoxFuture<'_, io::Result<D::DataWriter>>;
+}
+
+/// A mutable bao map
+pub trait BaoPartialMap: BaoMap {
```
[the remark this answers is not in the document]
rklaehn: Thanks. Not much left.

One thing I am not quite sure about is whether the get and get_partial fns should be async. Currently I think they can be sync, which means that at least the knowledge whether we have a hash or not must be quickly available in memory.

I can see scenarios where that would not be the case, e.g. you want to just rely on the file system for a membership test. But I think it is nice to be able to assume that this will be fast.

iroh-bytes/src/provider.rs
```diff
@@ -616,8 +713,107 @@ pub async fn send_blob<D: BaoMap, W: AsyncWrite + Unpin + Send + 'static>(
...
+/// The mutable part of a BaoDb
+pub trait BaoDb: BaoReadonlyDb + BaoPartialMap {
+    /// list partial blobs in the database
+    fn partial_blobs(&self) -> Box<dyn Iterator<Item = Hash> + Send + Sync + 'static>;
+
+    /// extract a file to a local path
+    ///
+    /// `hash` is the hash of the file
+    /// `target` is the path to the target file
+    /// `stable` is true if the file can be assumed to be retained unchanged in the file system
```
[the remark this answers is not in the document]
rklaehn: It gets less important if the reflink thing works. Then we can always reflink no matter if the target is stable or not.

But I guess we want to keep it for the case where the file system does not support CoW, and also so the DB itself remains small.

iroh-bytes/src/util.rs
```diff
@@ -96,16 +97,22 @@ impl Ord for Hash {
...
-        let mut res = [b'b'; 59];
-        // write the encoded bytes
-        data_encoding::BASE32_NOPAD.encode_mut(&self.as_cid_bytes(), &mut res[1..]);
-        // convert to string, this is guaranteed to succeed
-        let t = std::str::from_utf8_mut(res.as_mut()).unwrap();
-        // hack since data_encoding doesn't have BASE32LOWER_NOPAD as a const
-        t.make_ascii_lowercase();
-        // write the str, no allocations
-        f.write_str(t)
+        if f.alternate() {
```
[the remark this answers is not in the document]
rklaehn: I could throw this out again. Was just annoyed during debugging that I could not correlate hashes with b3sum output. See https://github.com/n0-computer/iroh/issues/1330

iroh-bytes/src/provider.rs
```diff
@@ -616,8 +713,107 @@ pub async fn send_blob<D: BaoMap, W: AsyncWrite + Unpin + Send + 'static>(
...
+            debug!("blob not found {}", hex::encode(name));
             Ok((SentStatus::NotFound, 0))
         }
     }
 }
+
+/// The mutable part of a BaoDb
+pub trait BaoDb: BaoReadonlyDb + BaoPartialMap {
+    /// list partial blobs in the database
+    fn partial_blobs(&self) -> Box<dyn Iterator<Item = Hash> + Send + Sync + 'static>;
```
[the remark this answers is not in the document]
rklaehn: This part of the API is the least settled yet. Sometimes people just want the number of hashes. Not really sure.

iroh-bytes/src/provider.rs
```diff
@@ -52,9 +66,44 @@ pub trait BaoMap: Clone + Send + Sync + 'static {
...
+    ///
+    /// We need to know the size of the partial entry. This might produce an
+    /// error e.g. if there is not enough space on disk.
+    fn get_or_create_partial(&self, hash: Hash, size: u64) -> io::Result<Self::PartialEntry>;
+
+    /// Get an existing partial entry
+    fn get_partial(&self, hash: &Hash) -> Option<Self::PartialEntry>;
+
+    /// Upgrade a partial entry to a complete entry
+    fn insert_complete_entry(&self, entry: Self::PartialEntry) -> BoxFuture<'_, io::Result<()>>;
```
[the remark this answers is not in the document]
rklaehn: Did this. We could also call it upgrade or something?

b5:
> @b5 do we need conversion tools from the existing flat file format?

I don't think we need to do a conversion tool for this go-round, _but_ all subsequent changes we should aim to provide migrations for

iroh-bytes/src/provider.rs
```diff
@@ -220,6 +270,74 @@ pub enum ProvideProgress {
...
+        /// The number of bytes written
+        bytes_written: u64,
+        /// The number of bytes read
+        bytes_read: u64,
+        /// The time it took to transfer the data
+        elapsed: Duration,
+    },
+    /// We are done with the whole operation
+    AllDone,
+    ///
```
dignifiedquire, line 319: missing comments

iroh-bytes/src/provider.rs
```diff
@@ -616,8 +734,107 @@ pub async fn send_blob<D: BaoMap, W: AsyncWrite + Unpin + Send + 'static>(
...
+        target: PathBuf,
+        stable: bool,
+        progress: impl Fn(u64) -> io::Result<()> + Send + Sync + 'static,
+    ) -> BoxFuture<'_, io::Result<()>>;
+
+    /// import a file from a local path
+    ///
+    /// `data` is the path to the file
+    /// `stable` is true if the file can be assumed to be retained unchanged in the file system. If
+    /// `stable` is false, the file will be copied.
```
dignifiedquire, line 766: I wonder if `should_copy` would be a better name

iroh/Cargo.toml
```diff
@@ -49,14 +49,14 @@ tempfile = { version = "3.4", optional = true }
 tracing-subscriber = { version = "0.3", features = ["env-filter"], optional = true }
 data-encoding = "2.4.0"
 url = { version = "2.4", features = ["serde"] }
-
+rand = "0.8"
```
dignifiedquire, line 52: please put it into the right section/order of the file

iroh/Cargo.toml
```diff
@@ -70,12 +70,14 @@ regex = { version = "1.7.1", features = ["std"] }
 testdir = "0.8"
 tokio = { version = "1", features = ["macros", "io-util", "rt"] }
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
-iroh-io = { version = "0.2.1" }
 postcard = "1"
 proptest = "1.2.0"
 tempfile = "3.4"
 genawaiter = { version = "0.99", features = ["futures03"] }
 
+[profile.release]
```
dignifiedquire, line 78: I don't think this needs to be here, I think only the ones in the top level are picked up

iroh/src/commands.rs
```diff
@@ -60,6 +61,65 @@ pub struct Cli {
...
+                    .server_streaming(ShareRequest {
+                        hash,
+                        recursive,
+                        peer,
+                        addrs: addr,
+                        derp_region,
+                        token: token.cloned(),
+                        out: out.map(|x| x.display().to_string()),
+                        in_place,
+                        force: true, // todo: use force from Commands::Share
```
dignifiedquire, line 114: what about this todo?

iroh/src/commands.rs
```diff
@@ -257,6 +338,56 @@ pub enum Commands {
...
         single: bool,
     },
+    /// Download data to the running provider's database and provide it.
+    ///
+    /// In addition to downloading the data, you can also specify an optional output directory
+    /// where the data will be exported to after it has been downloaded.
+    Share {
+        /// Hash to get, required unless ticket is specified
+        #[clap(long, conflicts_with = "ticket", required_unless_present = "ticket")]
+        #[clap(long)]
```
dignifiedquire, line 348: why clap(long) twice?

iroh/src/commands/list.rs
```diff
@@ -14,6 +14,12 @@ pub enum Commands {
         #[clap(long, default_value_t = DEFAULT_RPC_PORT)]
         rpc_port: u16,
     },
+    /// List the available blobs on the running provider.
+    IncompleteBlobs {
```
dignifiedquire, line 18: will this list incomplete collections?

iroh/src/database/mem.rs
```diff
@@ -1,121 +1,495 @@
-//! An in memory implementation of [BaoMap] and [BaoReadonlyDb], useful for
-//! testing and short lived nodes.
+//! A full in memory database for iroh-bytes
+//!
+//! Main entry point is [Database].
```
dignifiedquire, line 3: why do we need the test db if we have this?

iroh/src/rpc_protocol.rs
```diff
@@ -42,9 +45,43 @@ impl ServerStreamingMsg<ProviderService> for ProvideRequest {
     type Response = ProvideProgress;
 }
 
+///
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
dignifiedquire, line 49: missing comments

iroh/src/util/fs.rs
```diff
@@ -0,0 +1,121 @@
...
+    } else {
+        // A single file, use the file name as the name of the blob.
+        vec![DataSource {
+            name: canonicalize_path(root.file_name().context("path must be a file")?)?,
+            path: root,
+        }]
+    })
+}
+
+/// converts a canonicalized relative path to a string, returning an error if
```
dignifiedquire, line 88: doc comments should be sentences, so starting with capitalized and ending with a `.` (not just here, in a bunch of places)

iroh/src/commands.rs
rklaehn: merge oversight probably...

iroh/src/commands/list.rs
rklaehn: Collections are blobs, so yes.

The concept of collections does not really exist anymore except in a few places.

iroh/src/util/fs.rs
rklaehn: Not sure I like this rule. It takes more than a capitalized first word to make something a complete sentence. E.g. for a comment on a struct or function, you would have to start with "This struct" or "This function". Together with mandatory docs it feels very redundant and verbose. But I will give it a try.

iroh/tests/provide.rs
```diff
@@ -417,56 +418,13 @@ async fn test_server_close() {
         .expect("supervisor failed");
 }
 
-#[cfg(feature = "flat-db")]
-#[tokio::test]
-async fn test_blob_reader_partial() -> Result<()> {
```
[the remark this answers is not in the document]
rklaehn: I have no idea anymore what the test was supposed to test. I think it tested a behaviour of a complete download function - basically that it would behave correctly when abandoned in the middle.

But now that we have the finite state machine with explicit state transitions having such a test does not really make sense anymore.

iroh/src/node.rs
```diff
@@ -7,18 +7,35 @@
 //! To shut down the node, call [`Node::shutdown`].
 use std::fmt::Debug;
 use std::future::Future;
+use std::io;
```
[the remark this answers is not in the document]
rklaehn: Yes, I was going to do this. Already discussed with @Frando how to make a rpc request protocol more modular and split up the handlers.

The question is if this should happen in this PR or a subsequent one. I would prefer doing it later in a separate PR, otherwise _somebody_ will be in merge hell.

iroh/src/database/mem.rs
rklaehn: Not sure about the name, see the discussion with @Frando. But I like this one because it is lock free and very simple. It is the smallest possible impl that can be used to share stuff via iroh-bytes.

iroh/src/commands.rs
rklaehn: This was quite useful during development. Otherwise it would just do nothing all the time. I enabled this now, but the question is if we should have it at all. Maybe just remove the force mechanism entirely?

iroh-bytes/src/provider.rs
rklaehn: I think I will just make an enum...

rklaehn:
I removed the force flag entirely. I don't know when the user would ever want it, and for our own needs we can just delete the data. Well, we can't because we don't have delete, but we should fix that asap...

iroh-bytes/src/provider.rs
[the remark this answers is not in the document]
rklaehn: I will leave them sync for now. We can always make them async later without much overhead, but not much point doing it before we have a use case.

iroh-bytes/src/baomap.rs
```diff
@@ -0,0 +1,288 @@
...
+/// A partial entry
+pub trait PartialMapEntry<D: PartialMap>: MapEntry<D> {
+    /// A future that resolves to an writeable outboard
+    fn outboard_mut(&self) -> BoxFuture<'_, io::Result<D::OutboardMut>>;
+    /// A future that resolves to a writer that can be used to write the data
+    fn data_writer(&self) -> BoxFuture<'_, io::Result<D::DataWriter>>;
+}
+
+/// A mutable bao map
+pub trait PartialMap: Map {
```
dignifiedquire, line 74: why is the mutable version called `Partial`?

iroh-bytes/src/baomap.rs
```diff
@@ -0,0 +1,288 @@
...
+    /// The outboard type to write data to the partial entry.
+    type OutboardMut: bao_tree::io::fsm::OutboardMut;
+    /// The writer type to write data to the partial entry.
+    type DataWriter: iroh_io::AsyncSliceWriter;
+    /// A partial entry. This is an entry that is writeable and possibly incomplete.
+    ///
+    /// It must also be readable.
+    type PartialEntry: PartialMapEntry<Self>;
```
dignifiedquire, line 82: if `Partial<..>` is the mutable version, wouldn't it be clearer to use `Readable<..>` and `Writable<..>` (or `Mut<..>`) or sth like that for naming? the `Partial` naming indicates to me nothing about the mutability, just about the completion state.

iroh-bytes/src/baomap.rs
rklaehn: It is not just mutable as in you can add and remove entries, but it allows you to create partial (data, outboard) pairs that start empty and then slowly get filled up.

Now the normal Map is allowed to hand out partial entries since last week, since we want to be able to share stuff we don't yet have downloaded completely. Nevertheless it seemed like partial would be a better name than MapMut.

iroh-bytes/src/baomap.rs
```diff
@@ -0,0 +1,288 @@
...
+
+    /// Get an existing partial entry.
+    fn get_partial(&self, hash: &Hash) -> Option<Self::PartialEntry>;
+
+    /// Upgrade a partial entry to a complete entry.
+    fn insert_complete(&self, entry: Self::PartialEntry) -> BoxFuture<'_, io::Result<()>>;
+}
+
+/// Extension of BaoMap to add misc methods used by the rpc calls.
+pub trait ReadonlyStore: Map {
```
dignifiedquire, line 98: nit: maybe `ReadableStore`? 

Also what's the difference in your naming between `Map`s and `Store`s?

iroh-bytes/src/baomap.rs
```diff
@@ -0,0 +1,288 @@
...
+///
+/// An import operation involves computing the outboard of a file, and then
+/// either copying or moving the file into the database.
+#[allow(missing_docs)]
+#[derive(Debug)]
+pub enum ImportProgress {
+    /// Found a path
+    ///
+    /// This will be the first message for an id
+    Found { id: u64, path: PathBuf },
```
dignifiedquire, line 161: what does an `id` reference? a blob? or is this an event id?

iroh-bytes/src/baomap.rs
```diff
@@ -0,0 +1,288 @@
...
+}
+
+/// The import mode describes how files will be imported.
+///
+/// This is a hint to the import trait method. For some implementations, this
+/// does not make any sense. E.g. an in memory implementation will always have
+/// to copy the file into memory. Also, a disk based implementation might choose
+/// to copy small files even if the mode is `Reference`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum ImportMode {
```
dignifiedquire, line 191: I guess the plan is to add `Reflink` as an option here when we have support for it?

iroh-bytes/src/baomap.rs
```diff
@@ -0,0 +1,288 @@
...
+    Progress { id: u64, offset: u64 },
+    /// Done exporting
+    Done { id: u64 },
+}
+
+/// Progress updates for the provide operation
+#[derive(Debug, Serialize, Deserialize)]
+pub enum ValidateProgress {
+    /// started validating
+    Starting {
```
dignifiedquire, line 252: maybe `Start` as well to keep consistent with `ExportProgress`?

iroh-bytes/src/baomap.rs
rklaehn: A map is just a map, a store adds all the stuff you need for a full blown store, such as validation and export.

iroh-bytes/src/baomap.rs
```diff
@@ -0,0 +1,288 @@
...
+        /// location of the entry.
+        ///
+        /// In case of a file, this is the path to the file.
+        /// Otherwise it might be an url or something else to uniquely identify the entry.
+        path: Option<String>,
+        /// the size of the entry
+        size: u64,
+    },
+    /// We got progress ingesting item `id`.
+    Progress {
```
dignifiedquire, line 271: so this happens many times for each `Entry`? maybe then the names would be a little better as 

- `EntryStart`
- `EntryProgress`
- `EntryDone`

iroh-bytes/src/baomap.rs
```diff
@@ -0,0 +1,288 @@
...
+    Done {
+        /// The unique id of the entry.
+        id: u64,
+        /// An error if we failed to validate the entry.
+        error: Option<String>,
+    },
+    /// We are done with the whole operation.
+    AllDone,
+    /// We got an error and need to abort.
+    Abort(RpcError),
```
dignifiedquire, line 287: is there a case where an error happens and we do not need to abort?

iroh-bytes/src/baomap.rs
rklaehn: This is just an arbitrary id that allows you to save space in subsequent and very frequent progress calls like the update... ones. You wouldn't want to serialize the entire path in every update call, otherwise sending the update events over the wire would take more time than the actual process you are reporting progress on.

iroh-bytes/src/get.rs
```diff
@@ -445,24 +458,28 @@ pub mod fsm {
...
         }
 
-        /// Write the entire blob to a slice writer
-        pub async fn write_all<D: AsyncSliceWriter>(
-            self,
-            data: D,
-        ) -> result::Result<AtEndBlob, DecodeError> {
-            self.write_all_with_outboard::<D, D>(None, data).await
+        /// The geometry of the tree we are currently reading.
+        pub fn tree(&self) -> &bao_tree::BaoTree {
```
dignifiedquire, line 462: There are a couple of external types that are used in the public api here (BaoTree, blake3::Hash, RangeSet2). Please make sure they are reexported, so it is easy to pull them in in.

iroh-bytes/src/baomap.rs
rklaehn: Yes. But also if you have 2 options an enum is more clear than a flag. For the flag you never know if true means on or off.

Not sure how to deal with reflink: have a separate option or just treat it as a fast way to copy?

iroh/examples/rpc.rs
```diff
@@ -0,0 +1,86 @@
...
+}
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    setup_logging();
+    // let iroh_data_dir = std::env::current_dir()?.join(".iroh");
+    // tokio::fs::create_dir_all(&iroh_data_dir).await?;
+    // // create a new persistent database
+    // let db = iroh::database::flat::Store::load(iroh_data_dir.clone(), iroh_data_dir).await?;
+    // run(db).await
```
dignifiedquire, line 82: is this an example of how it would look like with the fs based backend?

iroh/src/commands.rs
```diff
@@ -257,6 +336,52 @@ pub enum Commands {
         #[clap(long, default_value_t = false)]
         single: bool,
     },
+    /// Download data to the running provider's database and provide it.
+    ///
+    /// In addition to downloading the data, you can also specify an optional output directory
+    /// where the data will be exported to after it has been downloaded.
+    Share {
```
dignifiedquire, line 343: Is there a way to share sth that I have already downloaded (into the iroh db/store)?

iroh-bytes/src/baomap.rs
rklaehn: Yeah, this one is much older than the others.

You can spend a lot of time bikeshedding whether this should be `Start` (imperative) or `Starting` (which for me seems like a state name and not a good name for a state transition event). But I guess `Start` it is because it is short...

iroh-bytes/src/baomap.rs
rklaehn: In some cases there can be many different progress types. E.g. if you import you might want to have progress for the outboard computation / hashing as well as for the copying into the store.

But `EntryCopyStart`, `EntryCopyProgress`, `EntryCopyDone` would sound a bit long. I thought that the fact that the msg has an id is sufficient to make it clear that it refers to an entry.

But I am not wed to the names.

iroh-bytes/src/util/progress.rs
```diff
@@ -0,0 +1,422 @@
...
+}
+
+impl IdGenerator for IgnoreProgressSender<()> {
+    fn new_id(&self) -> u64 {
+        0
+    }
+}
+
+///
+pub struct With<
```
dignifiedquire, line 178: missing docs, would be useful to have an explanation what exactly this is for

iroh-bytes/src/util/progress.rs
```diff
@@ -0,0 +1,422 @@
...
+    }
+
+    fn blocking_send(&self, msg: U) -> std::result::Result<(), ProgressSendError> {
+        let msg = (self.1)(msg);
+        self.0.blocking_send(msg)
+    }
+}
+
+///
+pub struct FilterWith<I, U, F>(I, F, PhantomData<U>);
```
dignifiedquire, line 233: missing docs

iroh-bytes/src/util/progress.rs
```diff
@@ -0,0 +1,422 @@
...
+    fn try_send(&self, msg: T) -> std::result::Result<(), ProgressSendError> {
+        TokioProgressSender::try_send(self, msg)
+    }
+
+    fn blocking_send(&self, msg: T) -> std::result::Result<(), ProgressSendError> {
+        TokioProgressSender::blocking_send(self, msg)
+    }
+}
+
+/// A conveneince type for sending progress messages.
```
dignifiedquire, line 327: typo

iroh-bytes/src/util/progress.rs
```diff
@@ -0,0 +1,422 @@
...
+
+    /// try to send a message and drop it if the receiver is full.
+    /// this will only fail if the receiver is dropped.
+    /// It can be used to send progress messages where delivery is not important, e.g.
+    /// a self contained progress message.
+    pub(crate) fn try_send(&self, msg: impl Into<T>) -> std::result::Result<(), ProgressSendError> {
+        let msg = msg.into();
+        match self.sender.try_send(msg) {
+            Ok(_) => Ok(()),
+            Err(tokio::sync::mpsc::error::TrySendError::Full(_)) => Ok(()),
```
dignifiedquire, line 404: shouldn't this report the fact that it was unable to deliver the message?

iroh/examples/rpc.rs
rklaehn: Yes, I was not quite sure whether to do this as one example with cli args or as two examples. Also, the commented out code is out of date.

So one or two examples?

iroh/src/node.rs
```diff
@@ -653,6 +665,32 @@ impl<D: BaoMap + BaoReadonlyDb, C: CollectionParser> RpcHandler<D, C> {
...
+                    return Err(io::Error::new(
+                        io::ErrorKind::NotFound,
+                        "no partial entry found",
+                    ));
+                };
+                io::Result::Ok(ListIncompleteBlobsResponse {
+                    hash,
+                    size: 0,
+                    expected_size: entry.size(),
+                    path: "".to_owned(),
```
dignifiedquire, line 687: shouldn't `path` be optional if this is used like this?

iroh/src/commands.rs
rklaehn: If it is already in the store it is already shared. You mean you want to export something you already have without providing a peer to download it from?

iroh/src/node.rs
```diff
@@ -653,6 +665,32 @@ impl<D: BaoMap + BaoReadonlyDb, C: CollectionParser> RpcHandler<D, C> {
         })
     }
 
+    fn list_incomplete_blobs(
+        self,
+        _msg: ListIncompleteBlobsRequest,
+    ) -> impl Stream<Item = ListIncompleteBlobsResponse> + Send + 'static {
+        let db = self.inner.db.clone();
+        let local = self.inner.rt.local_pool().clone();
+        futures::stream::iter(db.partial_blobs()).filter_map(move |hash| {
```
dignifiedquire, line 674: can `partial_blobs()` block? (or rather it's iterator impl? if so wouldn't it be better to run the whole iterator on a `spawn_blocking` and send the results via a channel?

iroh/src/node.rs
```diff
@@ -710,33 +749,571 @@ impl<D: BaoMap + BaoReadonlyDb, C: CollectionParser> RpcHandler<D, C> {
...
+                let collection = Collection::from_bytes(&bytes).context("invalid collection")?;
+                for Blob { hash, name } in collection.blobs() {
+                    let path = path.join(pathbuf_from_name(name));
+                    if let Some(parent) = path.parent() {
+                        tokio::fs::create_dir_all(parent).await?;
+                    }
+                    tracing::trace!("exporting blob {} to {}", hash, path.display());
+                    let id = progress.new_id();
+                    let progress1 = progress.clone();
+                    db.export(*hash, path, mode, move |offset| {
```
dignifiedquire, line 805: does this block?

iroh/src/node.rs
```diff
@@ -710,33 +749,571 @@ impl<D: BaoMap + BaoReadonlyDb, C: CollectionParser> RpcHandler<D, C> {
...
+        let valid_from_outboard = bao_tree::io::fsm::valid_ranges(&mut outboard).await?;
+        let valid: RangeSet2<ChunkNum> = valid_from_data.intersection(&valid_from_outboard);
+        let total_valid: u64 = valid
+            .iter()
+            .map(|x| match x {
+                RangeSetRange::Range(x) => x.end.to_bytes().0 - x.start.to_bytes().0,
+                RangeSetRange::RangeFrom(_) => 0,
+            })
+            .sum();
+        log!("valid_from_data: {:?}", valid_from_data);
```
dignifiedquire, line 966: what does `log!` do?

iroh/src/node.rs
rklaehn: The current idea is that fns like get, get_partial, and these iterators don't block.

So while the data can be on disk, the information about what you have must be in memory. If you want to have the ability to perform IO you would probably turn partial_blobs into a stream or a future that returns a stream.

iroh/src/node.rs
```diff
@@ -710,33 +749,571 @@ impl<D: BaoMap + BaoReadonlyDb, C: CollectionParser> RpcHandler<D, C> {
...
+                    })
+                    .await?;
+                }
+            }
+            #[cfg(not(feature = "iroh-collection"))]
+            anyhow::bail!("recursive export not supported without iroh-collection feature");
+        } else if let Some(parent) = path.parent() {
+            tokio::fs::create_dir_all(parent).await?;
+            let id = progress.new_id();
+            let entry = db.get(&hash).context("entry not there")?;
```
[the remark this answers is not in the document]
rklaehn: nope, see above. I had thought about making this async, but I think in most sane impls you would want get to be able to execute without io. I can make it a future to be... future proof. But it's a bit of a can of worms. Either you do the horribly verbose type member thing and cross your fingers, or you accept that every single get call will produce a BoxFuture that usually just will be futures::future::Ready(x).boxed().

iroh/src/node.rs
rklaehn: No, in this case export is a possibly long running async fn that returns a BoxFuture. If there was blocking io it would have to be dealt with inside the store.

iroh/src/node.rs
rklaehn: It's just so I can adjust the log level for the entire fn in a single line. `use tracing::trace as log;`

iroh/src/baomap/flat.rs
```diff
@@ -0,0 +1,1210 @@
...
+                data: Either::Right((data_path, size)),
+                outboard: Either::Right(outboard_path),
+            },
+        })
+    }
+
+    fn insert_complete(&self, entry: Entry) -> BoxFuture<'_, io::Result<()>> {
+        let hash = entry.hash.into();
+        let Either::Right((temp_data_path, size)) = entry.entry.data else {
+            todo!()
```
dignifiedquire, line 213: what about these `todo`s?

iroh/src/baomap/flat.rs
```diff
@@ -0,0 +1,1210 @@
...
+        let items = inner.complete.keys().copied().collect::<Vec<_>>();
+        Box::new(items.into_iter())
+    }
+
+    fn roots(&self) -> Box<dyn Iterator<Item = Hash> + Send + Sync + 'static> {
+        unimplemented!()
+    }
+
+    fn validate(&self, _tx: mpsc::Sender<ValidateProgress>) -> BoxFuture<'_, anyhow::Result<()>> {
+        unimplemented!()
```
dignifiedquire, line 481: what about this?

iroh/src/baomap/flat.rs
```diff
@@ -0,0 +1,1210 @@
...
+
+    fn export(
+        &self,
+        hash: Hash,
+        target: PathBuf,
+        mode: ExportMode,
+        progress: impl Fn(u64) -> io::Result<()> + Send + Sync + 'static,
+    ) -> BoxFuture<'_, io::Result<()>> {
+        let this = self.clone();
+        tokio::task::spawn_blocking(move || this.export_sync(hash, target, mode, progress))
```
dignifiedquire, line 498: shouldn't these `spawn_blocking` calls use the passed through `runtime::Handle`?

iroh/src/baomap/flat.rs
rklaehn: this is basically `unreachable!`. I just forgot to change it. I am using the same Entry type for partial and non-partial, but a partial entry can never have the left side. Two ways to solve this - either change it so there is a separate PartialEntry type, or return an io error "impossible" here.

iroh/src/baomap/flat.rs
rklaehn: I don't have a runtime handle in State currently. I guess I would have to add it and explicitly pass in a runtime in the store constructor.

iroh/src/baomap/flat.rs
rklaehn: Had punted on this, see https://github.com/n0-computer/iroh/issues/1338

I guess an anyhow::bail might be nicer.

iroh/src/baomap/flat.rs
```diff
@@ -0,0 +1,1210 @@
...
+                let temp_data_path = self
+                    .0
+                    .options
+                    .partial_path
+                    .join(format!("{}.temp", hex::encode(uuid)));
+                // copy the data, since it is not stable
+                progress.try_send(ImportProgress::CopyProgress { id, offset: 0 })?;
+                let size = std::fs::copy(&path, &temp_data_path)?;
+                // report the size only after the copy is done
+                progress.blocking_send(ImportProgress::Size { id, size })?;
```
dignifiedquire, line 590: this will panic, if the sync method is not called inside a tokio runtime, when using the tokio progress sender. you might want to use `flume` instead of `tokio::mpsc` for the progress sender, to avoid this

iroh/src/baomap/flat.rs
```diff
@@ -0,0 +1,1210 @@
...
+
+impl ReadonlyStore for Store {
+    fn blobs(&self) -> Box<dyn Iterator<Item = Hash> + Send + Sync + 'static> {
+        let inner = self.0.state.read().unwrap();
+        let items = inner.complete.keys().copied().collect::<Vec<_>>();
+        Box::new(items.into_iter())
+    }
+
+    fn roots(&self) -> Box<dyn Iterator<Item = Hash> + Send + Sync + 'static> {
+        unimplemented!()
```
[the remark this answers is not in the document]
rklaehn: This is related to gc/deletion. Made an issue but have not yet implemented it. See https://github.com/n0-computer/iroh/issues/1336

iroh/src/baomap/flat.rs
```diff
@@ -0,0 +1,1210 @@
...
+            Self::Outboard(hash) => write!(f, "{}.{}", hex::encode(hash), OUTBOARD_EXT),
+            Self::Meta(name) => write!(f, "{}.meta", hex::encode(name)),
+        }
+    }
+}
+
+impl FromStr for FileName {
+    type Err = ();
+
+    fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
```
dignifiedquire, line 1117: would be nice to have a test for this

iroh/src/baomap/flat.rs
```diff
@@ -0,0 +1,1210 @@
...
+use iroh_bytes::util::progress::{IdGenerator, ProgressSender};
+use iroh_bytes::{Hash, IROH_BLOCK_SIZE};
+use iroh_io::{AsyncSliceReader, AsyncSliceWriter, File};
+use rand::Rng;
+use range_collections::RangeSet2;
+use tokio::sync::mpsc;
+use tracing::trace_span;
+
+use super::flatten_to_io;
+
```
dignifiedquire, line 33: It would be really helpful to have some documentation describing the on disk data structure, and what it relies on

iroh/src/node.rs
rklaehn: In general, things like spawn_blocking where they are needed are left to the store impl. We don't want to call spawn_blocking just in case and then slow down an in mem impl.

iroh/src/baomap/mem.rs
```diff
@@ -0,0 +1,497 @@
...
+
+impl MutableMemFile {
+    /// Create a new empty file
+    pub fn with_capacity(capacity: usize) -> Self {
+        Self(Arc::new(RwLock::new(BytesMut::with_capacity(capacity))))
+    }
+
+    /// Freeze the data, returning the content
+    ///
+    /// Note that this will clear other references to the data.
```
dignifiedquire, line 54: why clear those, instead of freezing them?

iroh/src/baomap/mem.rs
```diff
@@ -0,0 +1,497 @@
...
+
+    type SyncFuture<'a> = futures::future::Ready<io::Result<()>>;
+
+    fn sync(&mut self) -> Self::SyncFuture<'_> {
+        futures::future::ok(())
+    }
+}
+
+#[derive(Debug, Clone, Default)]
+///
```
dignifiedquire, line 182: docs

iroh/src/baomap/mem.rs
```diff
@@ -0,0 +1,497 @@
...
+                .into_iter(),
+        )
+    }
+
+    fn roots(&self) -> Box<dyn Iterator<Item = Hash> + Send + Sync + 'static> {
+        Box::new(std::iter::empty())
+    }
+
+    fn validate(&self, _tx: mpsc::Sender<ValidateProgress>) -> BoxFuture<'_, anyhow::Result<()>> {
+        todo!()
```
dignifiedquire, line 315: todo?

iroh/src/baomap/mem.rs
```diff
@@ -0,0 +1,497 @@
...
+        }
+        .boxed()
+    }
+}
+
+impl baomap::Store for Store {
+    fn import(
+        &self,
+        data: std::path::PathBuf,
+        _mode: ImportMode,
```
dignifiedquire, line 408: why does this ignore the `ImportMode`?

iroh/src/baomap/mem.rs
```diff
@@ -0,0 +1,497 @@
...
+        _progress: impl ProgressSender<Msg = ImportProgress> + IdGenerator,
```
dignifiedquire, line 409: why no progress?, not even start and finished?

iroh/src/baomap/mem.rs
```diff
@@ -0,0 +1,497 @@
...
+            .complete
+            .insert(hash.into(), (bytes, outboard));
+        hash.into()
+    }
+
+    fn export_sync(
+        &self,
+        hash: Hash,
+        target: PathBuf,
+        _mode: ExportMode,
```
dignifiedquire, line 455: why ignored?

iroh/src/baomap/mem.rs
```diff
@@ -0,0 +1,497 @@
...
+        _progress: impl Fn(u64) -> io::Result<()> + Send + Sync + 'static,
```
dignifiedquire, line 456: why ignored?

iroh/src/baomap/readonly_mem.rs
```diff
@@ -0,0 +1,293 @@
...
+    pub fn get(&self, hash: &Hash) -> Option<Bytes> {
+        let entry = self.0.get(hash)?;
+        Some(entry.1.clone())
+    }
+
+    fn export_sync(
+        &self,
+        hash: Hash,
+        target: PathBuf,
+        _mode: ExportMode,
```
dignifiedquire, line 109: no mode and no progress check?

iroh/src/baomap/mem.rs
rklaehn: You are taking the BytesMut out of the Arc<Mutex<...>> using std::mem::swap. Freeze takes a BytesMut, not a &mut BytesMut, so this is the only way I could find to make this zero copy.

iroh/src/baomap/mem.rs
rklaehn: ImportMode does not really make sense for a mem store. It can not support ImportMode::Reference

iroh/src/baomap/mem.rs
rklaehn: ExportMode does not really make sense for a mem store. It can not support ExportMode::Reference

iroh/src/baomap/readonly_mem.rs
rklaehn: mode: see above. progress: to support progress you would no longer be able to use std::fs::write.

iroh/src/baomap/readonly_mem.rs
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    fn partial_blobs(&self) -> Box<dyn Iterator<Item = Hash> + Send + Sync + 'static> {
+        Box::new(std::iter::empty())
+    }
+}
+
+impl MapEntry<Store> for PartialEntry {
+    fn hash(&self) -> blake3::Hash {
+        unreachable!()
```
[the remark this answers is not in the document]
rklaehn: PartialEntry are not supported for readonly_mem. They are an uninhabited type, can not exist.

```rust
pub enum PartialEntry {}
```

iroh/src/baomap/mem.rs
dignifiedquire: right, but I think it should error on that, rather than silently ignore it

iroh/src/baomap/readonly_mem.rs
dignifiedquire: Doesn't progress at least give start and end? which we could easily emit at least. Also pretty sure `fs::write` is just `File::open` + `write_all`, which should support progress

iroh/src/baomap/flat.rs
dignifiedquire: I mean `unreachable!` with a comment why it won't happen works well enough I would say

iroh/src/commands.rs
dignifiedquire: I am thinking of this flow 

- `iroh get foobar`
- some time passes and someone asks me to share `foobar`
- `iroh ??? foobar` (what do I do here? is this just provide, but with what if I only stored it in irohs db? or is that not an option?)

iroh/examples/rpc.rs
dignifiedquire: I think one example with a flag is good

iroh-bytes/src/baomap.rs
dignifiedquire: I don't mind either, as long as it is consistent

iroh-bytes/src/baomap.rs
dignifiedquire: Yeah, they do get long. Maybe just add sth to the comment in that regard, that it gets emitted once or multiple times per X

iroh/src/node.rs
dignifiedquire: Okay, so the requirement here is on the store implementation to make sure this doesn't have any bad blocking effects. Maybe add this to the docs of the trait, so it is not lost?

iroh/src/commands.rs
rklaehn: Ah, ok. iroh get is either directly downloading or creating an ephemeral node in process. It does not even talk via rpc. So it is completely separate, and there is nothing you can do except call `iroh add` on the downloaded folder.

We should rethink the ergonomics of the whole get vs. share thing at some point, but for now that is how it works.

iroh/src/baomap/mem.rs
dignifiedquire: Maybe `TryReference`?

iroh/src/commands.rs
dignifiedquire: okay, thanks

iroh/src/baomap/flat.rs
rklaehn: OK, will do `unreachable!` now and then maybe have a separate entry type later.

iroh/src/baomap/readonly_mem.rs
rklaehn: Yeah, it was just me being lazy. Initially I did not support export at all for this store type, but then I added it.

iroh/src/baomap/mem.rs
rklaehn: So make it a hint, and change the name so that it is clear that is a hint. I can live with that.

iroh/src/node.rs
rklaehn: For Map::get I would argue that this is already the case. But I will make it clear also for the iterators.

iroh/src/baomap/flat.rs
rklaehn: I think this is not the end of the world, given that we are always in tokio land. But I will see if writing a FlumeProgressSender is straightforward.

iroh/src/node.rs
rklaehn: removed it entirely. incomplete blobs always live in the same place.

dignifiedquire (approved):
Looks good to me overall. At least I caught everything that I will catch given the size of the diff 😅

rklaehn:
I am sure there will be more things. But it is definitely better than what it replaces...

PR #1322, opened 2023-08-02
Fast bao tree

rklaehn:
## Description

Use the newly published iroh-blake3 and bao-tree to get much faster hashing of 16kb subtrees. Factor of 4 improvement due to SIMD use in blake3.

Ideally we would move back to upstream blake3 eventually, once https://github.com/BLAKE3-team/BLAKE3/pull/329 is merged.

To make the switch and the subsequent switch back easier I removed the dependency on blake3 and instead reexported our forked blake3 as blake3 in bao-tree.

## Notes & open questions

Is there a nicer way to do all this reexport business?

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-bytes/Cargo.toml
+++ b/iroh-bytes/Cargo.toml
@@ -13,8 +13,7 @@ rust-version = "1.66"
 
 [dependencies]
 anyhow = { version = "1", features = ["backtrace"] }
-bao-tree = { version = "0.5.0", features = ["tokio_fsm"], default-features = false }
-blake3 = "1.3.3"
+bao-tree = { version = "0.6.0", features = ["tokio_fsm"], default-features = false }
 bytes = { version = "1.4", features = ["serde"] }
 data-encoding = "2.3.3"
 derive_more = { version = "1.0.0-beta.1", features = ["debug", "display", "from", "try_into"] }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -243,11 +243,13 @@ dependencies = [
 
 [[package]]
 name = "bao-tree"
-version = "0.5.2"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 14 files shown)


rklaehn:
image

iroh-bytes/Cargo.toml
```diff
@@ -13,8 +13,7 @@ rust-version = "1.66"
 
 [dependencies]
 anyhow = { version = "1", features = ["backtrace"] }
-bao-tree = { version = "0.5.0", features = ["tokio_fsm"], default-features = false }
-blake3 = "1.3.3"
```
dignifiedquire, line 17: I believe there are a couple of other users (`iroh` crate at least) that use `blake3` directly, should be unified, to avoid duplicate dependencies

rklaehn: Not sure what to do about this. The way it is now,

- bao-tree uses iroh-blake3 instead of blake3
- bao-tree reexports this as blake3
- both iroh-bytes and iroh use bao-tree and the reexport of blake3

If some other crate uses both iroh and upstream blake3, it will work but you will have a duplicate dependency. But what can we do about it? Encourage people to use our fork or reexport blake3 in iroh somewhere?

dignifiedquire: It is fine if folks depending on iroh endup with duplicate deps, but at least in our tree it should be clean

dignifiedquire (approved):

iroh-bytes/Cargo.toml
rklaehn: There is the case where people might want to convert from a blake3::Hash to an iroh_bytes::Hash. That won't work unless they are using iroh-blake3. But this should be rare, so maybe fix it only if there is no chance of upstreaming the blake3 changes.

The blake3 changes are *really* minimal, so I hope we can get them in.

PR #1329, opened 2023-08-03
refactor(iroh-net): keep `DerpMap` fixed

### LINKED ISSUE #1182: Change derpmap to have Arc<DerpNode> as values
The derpmap is an expensive struct yet it gets cloned around all the time.  Furthermore especially in netcheck we jump from `DerpNode` to it's name as a string to pass around so it is cheaper.  To only later look up this name and find the `DerpNode` again.  This all would be a lot nicer if we globally made the `DerpMap` contain `Arc<DerpNode>` entries.

dignifiedquire:
## Description

The `DerpMap` wasn't being changed anyway at runtime, only for initial config.

Closes #1182 

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh-net/src/derp/map.rs
+++ b/iroh-net/src/derp/map.rs
@@ -27,7 +27,7 @@ impl DerpMap {
         ids
     }
 
-    /// Returns an `Iteratore` over all known regions.
+    /// Returns an `Iterator` over all known regions.
     pub fn regions(&self) -> impl Iterator<Item = &DerpRegion> {
         self.regions.values()
     }
--- a/iroh-net/src/derp/map.rs
+++ b/iroh-net/src/derp/map.rs
@@ -4,6 +4,7 @@ use std::{
     collections::HashMap,
     fmt,
     net::{IpAddr, Ipv4Addr, Ipv6Addr},
+    sync::Arc,
 };
 
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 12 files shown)


iroh-net/src/magicsock.rs
```diff
@@ -208,9 +212,8 @@ pub(self) struct Inner {
     /// If the last netcheck report, reports IPv6 to be available.
     pub(self) ipv6_reported: Arc<AtomicBool>,
 
-    // Sigh, a lock it is..
     /// None (or zero regions/nodes) means DERP is disabled.
-    pub(self) derp_map: tokio::sync::RwLock<Option<DerpMap>>,
```
rklaehn, line 213: 🎆

rklaehn (approved):
Looks good. Less locks is more better.

But why exactly don't we need set_derp_map anymore?

dignifiedquire:
> But why exactly don't we need set_derp_map anymore?

it came from the original tailscale design, but we don't actually use it, other than on startup

dignifiedquire:
Looks like some initial setup pieces are not quite working yet with this change..

iroh-net/src/derp/map.rs
```diff
@@ -26,6 +27,37 @@ impl DerpMap {
         ids
     }
 
+    /// Returns an `Iteratore` over all known regions.
```
divagant-martian, line 30:
```suggestion
    /// Returns an `Iterator` over all known regions.
```
it felt very Italian :ok_hand:

dignifiedquire: 🤣

ramfox (approved):
I do think that in the (distant) future we will likely need something that allows us to add new derp regions to very long running nodes without restarting them, this is much simpler and all we need for now!

PR #1333, opened 2023-08-04
feat: add iroh-sync and integrate into iroh node

Frando:
## Description

This PR adds `iroh-sync`, a document synchronization protocol, to iroh, and integrates with `iroh-net`, `iroh-gossip` and `iroh-bytes`. 

* At the core is the `iroh-sync` crate, with a set reconciliation algorithm implemented by @dignifiedquire. See [the old iroh-sync repo](https://github.com/n0-computer/iroh-sync/) for the prehistory and #1216 for the initial PR (fully included in this PR, and by now outdated)
* Iroh sync is integrated in the iroh node, with iroh-gossip, in the RPC interface, and the CLI. 
  * `LiveSync` is the handle to an actor that integrates sync with [gossip](#1149 ) to broadcast and receive document updates from peers. For each open document a gossip swarm is joined with a `TopicId` derived from the doc namespace.
  * mod `download` contains the new downloader. It will be improved in #1344 .
  *  mod `client` is the new high-level RPC client. It currently only has methods for dealing with docs and sync, other things should be added once we merged this. CLI commands for sync are in `commands/sync.rs`. Will be much better with #1356 .
  *  `examples/sync.rs`  has a REPL to modify and sync docs. It does a full setup without using the iroh console. Also includes code to sync directories, and a hammer command for load testing.
* The PR also introduces `iroh::client::Iroh`, a wrapper around the RPC client, and   `iroh::client::Doc`, a wrapper around RPC client for a single document

#### Should likely happen before merge:

* [x] Make `iroh_sync::Store:::list_authors` and `list_replicas` return iterators `iroh-sync` *fixed in #1366 *
* [ ] Add `iroh_sync::Store::close_replica`
* [x] `ContentStatus` in `on_insert` callback is reported as `Ready` if the content is still `baomap::PartialEntry` (in-process download) *fixed in a8e809367245504787168d354f1714b11305920d*

#### Can happen after merge, but  before `0.6` release

* [ ] Implement `AuthorImport` and `AuthorShare` RPC & CLI commands
* [ ] sync store `list_namespaces` and `list_authors` internally collect, return iterator instead
* [ ] Fix cross-compiles to arm/android. See https://github.com/cross-rs/cross/issues/1311
* [ ] Ensure that fingerpring calculation is efficient and/or cached for large documents. Currently calculating the initial fingerprint iterates once over all entries in a document.
* [ ] Make content downloads be more reliable
   * [ ] Add some way to download content from peers independent of the first insertion event for a remote entry. The downloader with retries is tracked in #1334 and 1344, but independent of that, we still would currently only ever try to queue a download when the `on_insert` callback triggers, which is only once. There should be a way, even if manual for now, to try to download missing content in a replica from peers.
   * [ ] during `iroh-sync` sync include info if content is available for each entry
* [ ] Add basic peer management and persistence. Currently live sync will stop to do anything after a node restart.
    * [ ] Persist the addressbook of peers for a document, to reconnect when restarting the node
    * [ ] Implement `PeerAdd` and `PeerList` RPC & CLI commands. The latter needs changes in `iroh-net` to expose information of currently-connected peers and their peer info.
* [ ] Make read-only replicas possible
* [ ] Improve reliablity of replica sync. 
     * sync is triggered on each `NeighborUp` event from gossip. check that we don't sync too much.
     * maybe include peer info in gossip messages, to queue syncs with those (but not all at once)
     * track and exchange the timestamp of last full sync for peers, to know if you missed gossiped message and react accordingly
     * add more tests with peers coming and leaving

#### Open questions

* [ ] `iroh_sync::EntrySignature` should the signatures include a namespace prefix?
* [ ] do we want the 1:1 mapping of `NamespaceId`and gossip `TopicId`, or would the topic id as a hash be better?

#### Other TODOs collected from the code

* [ ] Port `hammer` and `fs` commands from REPL example to iroh cli
* [ ] docs/naming: settle terminology about keypairs, private/secret/signing keys, public keys/identifiers and make docs and symbols consistent
* [ ] Make `bytes_get` streaming in the RPC interface
* [ ] Allow to unset the subscription on a replica
* [ ] `iroh-sync` shouldn't depend on `iroh-bytes` only for `Hash` type -> #1354
* [ ] * [ ] Move `sync::live::PeerSource` to iroh-net or even better -> #1354
* [ ] `StoreInstance::put` propagate error and verify timestamp is reasonable.
* [ ] `StoreInstance::get_range` implement inverted range
* [ ] `iroh_sync`: Remove some items only used in tests (marked with #[cfg(test)])
* [ ] `iroh_sync` fs store: verify get method fetches all keys with this namespace
* [ ] `ranger::SimpleStore::get_range`: optimize
* [ ] `ranger::Peer` avoid allocs?
* [ ] `fs::StoreInstance::get_fingerprint` optimize
* [ ] `SyncEngine::doc_subscribe` remove unwrap, handle error

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.


iroh/src/database/flat/writable.rs
Arqu, line 1: IIUC this is no longer used and should be dropped.

iroh/examples/sync.rs
```diff
@@ -0,0 +1,1057 @@
...
+        (false, Some(url)) => Some(DerpMap::from_url(url, 0)),
+        (true, None) => None,
+        (true, Some(_)) => bail!("You cannot set --no-derp and --derp at the same time"),
+    };
+    println!("> using DERP servers: {}", fmt_derp_map(&derp_map));
+
+    // build our magic endpoint and the gossip protocol
+    let (endpoint, gossip, initial_endpoints) = {
+        // init a cell that will hold our gossip handle to be used in endpoint callbacks
+        let gossip_cell: OnceCell<Gossip> = OnceCell::new();
```
flub, line 138: not that this needs to change, but i believe the "async way" is to use a oneshot channel for this.

iroh/src/metrics.rs
```diff
@@ -18,6 +29,13 @@ impl Default for Metrics {
             requests_total: Counter::new("Total number of requests received"),
             bytes_sent: Counter::new("Number of bytes streamed"),
             bytes_received: Counter::new("Number of bytes received"),
+            download_bytes_total: Counter::new("Total number of content bytes downloaded"),
+            download_time_total: Counter::new("Total time in ms spent downloading content bytes"),
```
flub, line 33: doesn't seem that useful to me.  a histogram of download speeds or something might be more useful.  but just removing this would also be fine i think

iroh-gossip/src/net.rs
```diff
@@ -469,13 +479,21 @@ impl Actor {
 
     async fn handle_in_event(&mut self, event: InEvent, now: Instant) -> anyhow::Result<()> {
         let me = *self.state.me();
-        debug!(me = ?me, "handle in_event  {event:?}");
+        if matches!(event, InEvent::TimerExpired(_)) {
+            trace!(me = ?me, "handle in_event  {event:?}");
```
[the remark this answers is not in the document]
Frando: Those were not added in this PR but in #1149. `me` is a field because it is a simple type (32 byte peer id) whereas `event` is a complex type (enum with tuple variantes and potentially a lot of data in them) so I thought debug-printing makes more sense. But also I'm not really sure when to use fields vs debug logging in tracing. I created a separate PR for this as its unrelated to this PR: https://github.com/n0-computer/iroh/pull/1390

iroh-sync/src/lib.rs
```diff
@@ -0,0 +1,40 @@
+//! Multi-dimensional key-value documents with an efficient synchronization protocol
+//!
+//! The crate operates on [Replicas](Replica). A replica contains an unlimited number of
+//! [Entrys][Entry]. Each entry is identified by a key, its author, and the replica's
```
flub, line 4:
```suggestion
//! [Entrys](Entry). Each entry is identified by a key, its author, and the replica's
```
?

But if cargo-doc is happy then it's fine i guess...

iroh-sync/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+
+[dependencies]
+anyhow = "1.0.71"
+blake3 = { package = "iroh-blake3", version = "1.4.3"}
+crossbeam = "0.8.2"
+derive_more = { version = "1.0.0-beta.1", features = ["debug", "display", "from", "try_into"] }
+ed25519-dalek = { version = "2.0.0", features = ["serde", "rand_core"] }
+flume = "0.10"
+iroh-bytes = { version = "0.5.0", path = "../iroh-bytes" }
+iroh-metrics = { version = "0.5.0", path = "../iroh-metrics", optional = true }
```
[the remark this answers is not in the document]
Frando: I don't know, but this is the same as on `main`: https://github.com/n0-computer/iroh/blob/main/iroh/Cargo.toml#L25

iroh/Cargo.toml
```diff
@@ -50,24 +56,30 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"], optional = tr
...
+shell-words = { version = "1.1.0", optional = true }
+shellexpand = { version = "3.1.0", optional = true }
+rustyline = { version = "12.0.0", optional = true }
+itertools = "0.11.0"
+
 [features]
 default = ["cli", "metrics"]
 cli = ["clap", "config", "console", "dirs-next", "indicatif", "multibase", "quic-rpc/quinn-transport", "tempfile", "tokio/rt-multi-thread", "tracing-subscriber", "flat-db", "mem-db", "iroh-collection"]
-metrics = ["iroh-metrics"]
+metrics = ["iroh-metrics", "flat-db", "mem-db", "iroh-collection"]
```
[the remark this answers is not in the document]
Frando: Good point. I think this is a leftover from the initial hacky integration of iroh-sync. Removed, let's see if CI says everything fine.

iroh-sync/src/keys.rs
```diff
@@ -0,0 +1,276 @@
...
+
+    /// Returns the [`Author`] byte representation.
+    pub fn to_bytes(&self) -> [u8; 32] {
+        self.signing_key.to_bytes()
+    }
+
+    /// Returns the [`AuthorId`] byte representation.
+    pub fn id_bytes(&self) -> [u8; 32] {
+        self.signing_key.verifying_key().to_bytes()
+    }
```
flub, line 36: Seems redundant since `AuthorId::as_bytes` already exists

iroh/examples/hello-world.rs
```diff
@@ -22,12 +22,17 @@ async fn main() -> anyhow::Result<()> {
     setup_logging();
     // create a new, empty in memory database
     let mut db = iroh::baomap::readonly_mem::Store::default();
+    // create an in-memory doc store (not used in the example)
```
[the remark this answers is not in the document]
Frando: Because the `iroh::node::Node` is now generic over the doc store. What we could do is to add a `NoopDocStore` or `DisabledDocStore` that would return errors for all method calls, and make that the default for the generic. WDYT? Or other ideas on how to deal with "we need a generic for this but want a disabled default"?

iroh-sync/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+version = "0.5.1"
+edition = "2021"
+readme = "README.md"
+description = "Iroh sync"
+license = "MIT/Apache-2.0"
+authors = ["n0 team"]
+repository = "https://github.com/n0-computer/iroh"
+
+[dependencies]
+anyhow = "1.0.71"
```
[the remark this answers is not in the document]
Frando: fixe

iroh-sync/Cargo.toml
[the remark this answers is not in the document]
flub: fair enough :shrug:

iroh/Cargo.toml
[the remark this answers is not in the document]
Frando: Removed in the latest commit and CI is as green as it gets :+1:

iroh-sync/src/ranger.rs
```diff
@@ -0,0 +1,1322 @@
...
+
+pub trait RangeKey: Sized + Ord + Debug {
+    /// Is this key inside the range?
+    fn contains(&self, range: &Range<Self>) -> bool {
+        contains(self, range)
+    }
+}
+
+/// Default implementation of `contains` for `Ord` types.
+pub fn contains<T: Ord>(t: &T, range: &Range<T>) -> bool {
```
flub, line 58: why is this available both directly and via the trait?

iroh-sync/src/ranger.rs
```diff
@@ -0,0 +1,1322 @@
...
+        Fingerprint::new(&[][..])
+    }
+
+    pub fn new<T: AsFingerprint>(val: T) -> Self {
+        val.as_fingerprint()
+    }
+}
+
+pub trait AsFingerprint {
+    fn as_fingerprint(&self) -> Fingerprint;
```
flub, line 90: isn't this `to_fingerprint` since it returns something owned?

iroh-gossip/src/proto/util.rs
```diff
@@ -56,7 +56,7 @@ macro_rules! idbytes_impls {
             }
         }
 
-        impl<T: Into<[u8; 32]>> From<T> for $ty {
+        impl<T: Into<[u8; 32]>> std::convert::From<T> for $ty {
```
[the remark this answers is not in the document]
Frando: Good catch. Fixed in the latest commit.

iroh-sync/src/ranger.rs
```diff
@@ -0,0 +1,1322 @@
...
+                    let range = Range::new(x.clone(), y.clone());
+                    ranges.push(range);
+                    start_index += 1;
+                }
+
+                for range in ranges.into_iter() {
+                    let chunk: Vec<_> = self
+                        .store
+                        .get_range(range.clone(), self.limit.clone())?
+                        .collect();
```
flub, line 575: do you need to allocate this?  Seems like you only get the `len` from it?

iroh/src/download.rs
```diff
@@ -0,0 +1,386 @@
...
+
+/// A download queue for iroh-bytes
+///
+/// Spawns a background task that handles connecting to peers and performing get requests.
+///
+/// TODO: Support retries and backoff - become a proper queue...
+/// TODO: Support collections, likely become generic over C: CollectionParser
+#[derive(Debug, Clone)]
+pub struct Downloader {
+    pending_downloads: Arc<Mutex<HashMap<Hash, DownloadFuture>>>,
```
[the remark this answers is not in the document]
Frando: I think I agree. The downloader will be overhauled in #1344 so not sure if it's worth changing it here.

iroh/src/metrics.rs
Frando: This was added because @Arqu said "only do counters in metrics if you can" and @b5 wanted to have download speed stats for the demo. So I'd defer to them on what to do here :)

iroh-sync/src/keys.rs
Frando: Yup I agree, @dignifiedquire added it, are you fine with removing it again?

Frando: Removed it

iroh-sync/src/ranger.rs
dignifiedquire: Unfortunately you can't `count` an iterator without consuming it, so I haven't found a way around it.

iroh-sync/src/ranger.rs
dignifiedquire: Because they take different arguments, and the standalone version is useful to implement the trait on other types.

iroh-net/src/key.rs
```diff
@@ -50,6 +50,18 @@ impl PublicKey {
...
     }
 
+    /// Construct a `PublicKey` from a slice of bytes.
+    ///
+    /// # Warning
+    ///
+    /// This will return a [`SignatureError`] if the bytes passed into this method do not represent
+    /// a valid `ed25519_dalek` curve point. Will never fail for bytes return from [`Self::as_bytes`].
+    /// See [`VerifyingKey::from_bytes`] for details.
+    pub fn from_bytes(bytes: &[u8; 32]) -> Result<Self, SignatureError> {
```
dignifiedquire, line 60: Is this really needed when we have `TryFrom<&[u8>`?

iroh-sync/src/keys.rs
```diff
@@ -0,0 +1,283 @@
...
+    ///
+    /// Will return an error if the input bytes do not represent a valid [`ed25519_dalek`]
+    /// curve point. Will never fail for a byte array returned from [`Self::as_bytes`].
+    /// See [`VerifyingKey::from_bytes`] for details.
+    pub fn from_bytes(bytes: &[u8; 32]) -> anyhow::Result<Self> {
+        Ok(NamespaceId(VerifyingKey::from_bytes(bytes)?))
+    }
+}
+
+impl fmt::Display for Author {
```
dignifiedquire, line 150: you can avoid these manual impls by using `derive_more::Display` much easier to do

iroh/src/commands/sync.rs
```diff
@@ -0,0 +1,263 @@
...
+        }
+        Ok(())
+    }
+}
+
+#[derive(Debug, Clone, Parser)]
+pub enum Docs {
+    List,
+    Create,
+    // Import {
```
dignifiedquire, line 75: why is this commented out?

iroh/src/commands/sync.rs
```diff
@@ -0,0 +1,263 @@
...
+    StartSync {
+        peers: Vec<PeerSource>,
+    },
+    Share {
+        mode: ShareMode,
+    },
+    /// Set an entry
+    Set {
+        /// Author of this entry.
+        author: AuthorId,
```
dignifiedquire, line 124: should be the same as with get `#[clap(short, long)]` otherwise this is a positional argument

iroh/src/commands/sync.rs
```diff
@@ -0,0 +1,263 @@
...
+        /// If true, old entries will be included. By default only the latest value for each key is
+        /// shown.
+        #[clap(short, long)]
+        old: bool,
+
+        /// Also print the content for each entry (but only if smaller than 1MB and valid UTf-8)
+        #[clap(short, long)]
+        content: bool,
+    },
+    List {
```
dignifiedquire, line 151: this should probably have the option to filter by author?

iroh/src/commands/sync.rs
```diff
@@ -0,0 +1,263 @@
...
+    pub async fn run(self, iroh: Iroh, doc_id: NamespaceId) -> anyhow::Result<()> {
+        let doc = iroh.get_doc(doc_id)?;
+        match self {
+            Doc::StartSync { peers } => {
+                doc.start_sync(peers).await?;
+                println!("ok");
+            }
+            Doc::Share { mode } => {
+                let ticket = doc.share(mode).await?;
+                // println!("key:    {}", hex::encode(ticket.key));
```
dignifiedquire, line 171: delete?

iroh/src/rpc_protocol.rs
```diff
@@ -256,6 +269,333 @@ pub struct VersionResponse {
     pub version: String,
 }
 
+// peer
+
+/// todo
```
dignifiedquire, line 274: lets either fill in the `todo`s or remove them

iroh/src/node.rs
```diff
@@ -257,16 +295,24 @@ where
     /// This will create the underlying network server and spawn a tokio task accepting
     /// connections.  The returned [`Node`] can be used to control the task as well as
     /// get information about it.
-    pub async fn spawn(self) -> Result<Node<D>> {
+    pub async fn spawn(self) -> Result<Node<D, S>> {
         trace!("spawning node");
         let rt = self.rt.context("runtime not set")?;
 
+        // TODO: this should actually run globally only once.
```
dignifiedquire, line 302: we should do sth about this before merging

iroh/src/node.rs
```diff
@@ -257,16 +295,24 @@ where
...
+        init_metrics_collection();
+
         let (endpoints_update_s, endpoints_update_r) = flume::bounded(1);
         let mut transport_config = quinn::TransportConfig::default();
         transport_config
             .max_concurrent_bidi_streams(MAX_STREAMS.try_into()?)
             .max_concurrent_uni_streams(0u32.into());
 
+        // init a cell that will hold our gossip handle to be used in endpoint callbacks
+        let gossip_cell: OnceCell<Gossip> = OnceCell::new();
```
dignifiedquire, line 313: you can use a oneshot here

iroh/src/node.rs
```diff
@@ -392,27 +475,15 @@ where
...
-                        let collection_parser = collection_parser.clone();
-                        let rt2 = rt.clone();
-                        let callbacks = callbacks.clone();
-                        rt.main().spawn(iroh_bytes::provider::handle_connection(connecting, db, callbacks, collection_parser, custom_get_handler, auth_handler, rt2));
-                    } else {
-                        tracing::error!("unknown protocol: {}", alpn);
-                        continue;
-                    }
-                }
+                    rt.main().spawn(handle_connection(connecting, alpn, handler.inner.clone(), gossip.clone(), collection_parser.clone(), custom_get_handler.clone(), auth_handler.clone()));
```
dignifiedquire, line 485: the error from the spawn is never handled, should at least print it

iroh-net/src/key.rs
Frando: For me having a from_bytes method seems convenient. Eg all the `ed22591_dalek` types have it as well. I can also remove it though..

iroh/src/client.rs
```diff
@@ -0,0 +1,203 @@
+//! Client to an iroh node. Is generic over the connection (in-memory or RPC).
+//!
+//! TODO: Contains only iroh sync related methods. Add other methods.
+
+// TODO: fill out docs
+#![allow(missing_docs)]
```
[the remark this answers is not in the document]
Frando: Done

iroh/src/client.rs
```diff
@@ -0,0 +1,203 @@
...
+            .rpc
+            .rpc(DocStartSyncRequest {
+                doc_id: self.id,
+                peers,
+            })
+            .await??;
+        Ok(())
+    }
+
+    // TODO: add stop_sync
```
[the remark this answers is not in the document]
Frando: I added the full impl.

iroh/src/commands/sync.rs
Frando: leftover, removed

iroh/src/commands/sync.rs
Frando: Changed. However also this is all overhauled in #1356

iroh/src/commands/sync.rs
Frando: removed

iroh/src/rpc_protocol.rs
Frando: removed them for now

iroh/src/node.rs
Frando: Actually I think we can just remove the TODO:
`init_metrics_collection` calls `iroh_metrics::core::Core::try_init(|reg, metrics| { ... }).ok()`, whereas `Core::try_init` tries to set a `OnceCell`, which will fail for subsequent invocations, which is silenced by the `ok`.
So - not sure how/if we want to do this differently?

iroh/src/node.rs
Frando: logging the error now.

iroh/src/commands/sync.rs
Frando: Added

iroh-sync/src/ranger.rs
dignifiedquire: maybe

iroh-sync/src/keys.rs
dignifiedquire: https://docs.rs/derive_more/1.0.0-beta.3/derive_more/derive.Display.html

Frando: OK, can look into this. However I'm not sure if I do like it better to put actual code into the derive attributes, for me that doesn't really look simpler than the impl here.

iroh/src/node.rs
dignifiedquire: The only worry I have with this setup, is that we now can not add any metrics in `main.rs` eg, because the first try_init will succeed and follow ups will not, so only the metrics in here will be added. It is probably be fine but want to call it out.

cc @Arqu

iroh/src/client.rs
```diff
@@ -0,0 +1,203 @@
...
+        Ok(doc)
+    }
+
+    pub async fn list_docs(&self) -> Result<impl Stream<Item = Result<NamespaceId>>> {
+        let stream = self.rpc.server_streaming(DocsListRequest {}).await?;
+        Ok(flatten(stream).map_ok(|res| res.id))
+    }
+
+    pub fn get_doc(&self, id: NamespaceId) -> Result<Doc<C>> {
+        // TODO: Check if doc exists?
```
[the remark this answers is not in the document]
Frando: I added this. Needed a new RPC method - I chose `DocInfoRequest`, where the `DocInfoResponse` is currently empty, but that would be where we'd add some info on if the document is sync'ed currently etc.
I kept the signature as currently, and added `try_get_doc() -> Result<Option<Doc>>`. Usually you will want to error if the doc doesn't exist, so I think this is easier to use this way.

iroh/src/node.rs
Frando: Well, you would have to add the metrics in `main.rs` before spawning the node. This should be easily doable.
I added a commit that clarifies the current setup with better method names and comments: https://github.com/n0-computer/iroh/pull/1333/commits/503ff23b6439a4787bd4d02e7d605c65cbfd8022
Or do you have another idea on how to proceed with this?

iroh-sync/src/keys.rs
dignifiedquire: not super important

iroh/src/node.rs
dignifiedquire: I think comments about this are good enough for now 👍

Arqu: Yeah I think having comments is the best option currently. Due to the metrics setup we can only define them once depending where your entrypoint is. The benefit is that you can randomly drop metrics wherever you like without having to pass weirdo references everywhere.

dignifiedquire (approved):
a long time coming, and finally a big PR that is not from me <3

PR #1337, opened 2023-08-08
Db write mod rename

rklaehn:
As discussed yesterday and requested by @Frando , make an additional module baomap instead of having long trait names.

Also rename Database to Store.

Note that this is only moving stuff around, not a single LOC is actually changed.


rklaehn:
This was requested in a PR review, and I also like it. So I am going to just merge it.

PR #1345, opened 2023-08-11
refactor: improve path handling in iroh dir

Frando:
## Description

Currently the handling of paths and files within the IROH_DATA_DIR was a bit ad-hoc. This adds an enum in `config.rs` to only use well-defined paths.

## Change checklist

- [x] Self-review.
- [x] ~~Tests if relevant~~.

### DIFF
--- a/iroh/src/commands/provide.rs
+++ b/iroh/src/commands/provide.rs
@@ -49,12 +49,14 @@ pub async fn run(
         );
     }
 
-    let blob_dir = IrohPaths::BaoFlatStore.env_path()?;
+    let blob_dir = IrohPaths::BaoFlatStoreComplete.with_env()?;
+    let partial_blob_dir = IrohPaths::BaoFlatStorePartial.with_env()?;
     tokio::fs::create_dir_all(&blob_dir).await?;
-    let db = flat::Store::load(&blob_dir, &blob_dir, rt)
+    tokio::fs::create_dir_all(&partial_blob_dir).await?;
+    let db = flat::Store::load(&blob_dir, &partial_blob_dir, rt)
         .await
         .with_context(|| format!("Failed to load iroh database from {}", blob_dir.display()))?;
-    let key = Some(IrohPaths::Keypair.env_path()?);
+    let key = Some(IrohPaths::Keypair.with_env()?);
     let token = opts.request_token.clone();
     let provider = provide(db.clone(), rt, key, opts).await?;
     let controller = provider.controller();
--- a/iroh/src/commands/provide.rs
+++ b/iroh/src/commands/provide.rs
@@ -19,7 +19,7 @@ use quic_rpc::{transport::quinn::QuinnServerEndpoint, ServiceEndpoint};
 use tokio::io::AsyncWriteExt;
 use tracing::{info_span, Instrument};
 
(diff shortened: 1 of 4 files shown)


iroh/src/commands/provide.rs
```diff
@@ -49,20 +49,12 @@ pub async fn run(
...
 
-    let mut iroh_data_root = iroh_data_root()?;
-    if !iroh_data_root.is_absolute() {
-        iroh_data_root = std::env::current_dir()?.join(iroh_data_root);
-    }
-    tokio::fs::create_dir_all(&iroh_data_root).await?;
-    let db = flat::Store::load(&iroh_data_root, &iroh_data_root, rt)
+    let blob_dir = IrohPaths::BaoFlatStore.env_path()?;
+    tokio::fs::create_dir_all(&blob_dir).await?;
+    let db = flat::Store::load(&blob_dir, &blob_dir, rt)
```
dignifiedquire, line 54: maybe also make a separate dir for the partial blobs?

iroh/src/config.rs
```diff
@@ -22,6 +23,55 @@ pub const CONFIG_FILE_NAME: &str = "iroh.config.toml";
...
+        let s: &str = self.into();
+        Path::new(s)
+    }
+}
+impl IrohPaths {
+    /// Get the path for this [`IrohPath`] by joining the name to `IROH_DATA_DIR` environment variable.
+    pub fn env_path(self) -> Result<PathBuf> {
+        let mut root = iroh_data_root()?;
+        if !root.is_absolute() {
+            root = std::env::current_dir()?.join(root);
```
dignifiedquire, line 68: shouldn't this use `https://doc.rust-lang.org/std/path/struct.Path.html#method.canonicalize` instead?

iroh/src/config.rs
```diff
@@ -22,6 +23,55 @@ pub const CONFIG_FILE_NAME: &str = "iroh.config.toml";
...
+        }
+    }
+}
+impl FromStr for IrohPaths {
+    type Err = anyhow::Error;
+    fn from_str(s: &str) -> Result<Self> {
+        Ok(match s {
+            "keypair" => Self::Keypair,
+            "blobs.v0" => Self::BaoFlatStore,
+            _ => bail!("unknown file or directory"),
```
dignifiedquire, line 47: would be good to have a basic test for this

Frando: added a encode-parse roundtrip test

iroh/src/config.rs
Frando: This comes from current main, I only copied it from `src/commands/provide.rs`: https://github.com/n0-computer/iroh/pull/1345/files#diff-7de268e3454ececdfb93202f8f3bb012d38673ddcdca12fd5f7c5e2a6778dc5bL53 
so dunno for sure. Does `canonicalize` join to `env::current_dir()` for relative paths? And - is that actually what we want?

Frando:
failure is unrelated (windows box fails to setup)

iroh/src/config.rs
dignifiedquire: hmm, okay lets leave it as is for now

dignifiedquire (approved):

PR #1346, opened 2023-08-11
feat: implement minimal subscription for sync

dignifiedquire:


Frando:
LGTM. I'd merge this now so that the todo app stuff and FFI can be directly on the sync-integration PR. I can then also look into adding removal of the on_insert callbacks, I think that'd be the most important of the TODOs in the PR.

PR #1353, opened 2023-08-14
feat(sync): emit event when entry content is downloaded from peers

Frando:
## Description

Adds `LiveEvent`s when the content of an entry is downloaded. Also adds availability status of content to the live events.
Now the end-to-end sync test passes without any `sleep`!

## Notes & open questions

* I wasn't sure if the entry `RecordIdentifier` should be included in the event, or only the content `Hash`. Right now the PR does the latter, but could easily switch to the former if it makes it more usable.
* The `SyncEngine` now doesn't do anything over the `LiveSync`. We might want to unify the two structs.  But would do in a seperate PR.
* We might want to add more events later

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/sync/live.rs
+++ b/iroh/src/sync/live.rs
@@ -601,7 +601,7 @@ impl<S: store::Store, B: baomap::Store> Actor<S, B> {
 
                 // A new entry was inserted from initial sync or gossip. Queue downloading the
                 // content.
-                let content_status = if let Some(_bao_entry) = self.bao_store.get(&hash) {
+                let content_status = if self.bao_store.get(&hash).is_some() {
                     ContentStatus::Ready
                 } else {
                     if let Some(from) = from {
--- a/iroh/src/sync/live.rs
+++ b/iroh/src/sync/live.rs
@@ -142,6 +142,7 @@ pub type OnLiveEventCallback =
 
 /// Events informing about actions of the live sync progres.
 #[derive(Serialize, Deserialize, Debug, Clone)]
+#[allow(clippy::large_enum_variant)]
 pub enum LiveEvent {
     /// A local insertion.
     InsertLocal {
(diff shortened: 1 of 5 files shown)


iroh/src/sync/live.rs
```diff
@@ -544,29 +586,52 @@ impl<S: store::Store> Actor<S> {
...
                 }
             }
             InsertOrigin::Sync(peer_id) => {
+                let from = peer_id.and_then(|id| PeerId::from_bytes(&id).ok());
+                let entry = signed_entry.entry();
+                let hash = *entry.record().content_hash();
+
+                // A new entry was inserted from initial sync or gossip. Queue downloading the
+                // content.
+                let content_status = if let Some(_bao_entry) = self.bao_store.get(&hash) {
```
dignifiedquire, line 604: `.is_some()` is a bit cleaner

dignifiedquire: would also be nice, if there was a method for this on the store, eg `.contains`, that could look this up without having to move any memory

iroh/src/sync/live.rs
```diff
@@ -544,29 +586,52 @@ impl<S: store::Store> Actor<S> {
...
+                    ContentStatus::Ready
+                } else {
+                    if let Some(from) = from {
```
dignifiedquire, line 607: when do we have no `from`? this would mean we never attempt to download those

dignifiedquire (approved):
some nits, otherwise this is a great addition

iroh/src/sync/live.rs
Frando: Would leave that for a separate PR. Also I just noticed that partial entries are not properly covered.

iroh/src/sync/live.rs
Frando: Actually we always have `from`. Will remove the Option in a separate commit.

PR #1356, opened 2023-08-15
feat: Iroh console (REPL) and restructured CLI

Frando:
## Description

This adds a REPL to the main `iroh` binary. It builds upon #1216.

* The REPL embeds all commands that operate via the RPC client - which, currently, is everything but `provide` (which should be called `start`), `get` and `doctor`.

* The REPL has two new, REPL-only commands: `set-doc` and `set-author`. `set-doc` changes the state of the REPL: author and document will be displayed above the input line, and the document commands (`set`, `get`, `list` , `share` etc) will be available top-level.

* Most of the changes  in `src/commands.rs` only move code and the `rpc_port` option around, the individual commands are not changed in this PR.

## Notes & open questions

* The document and author IDs  in the `list` commands are currently printed in hex. Should change to base32.

* I'm not yet super sure about the `set-doc` and `set-author` commands. Another path might be to dig more into a pwd-like structure and have a `cd` command or so. This could then also move further into documents. I'm not sure how the author fits in here.

* When in the level of a document, there's a conflict between the `doc list` command (which is just `list` then) and the top-level `list` command (to list blobs and collections). Not sure yet what the solution is. For now I embedded only the `sync` commands on the doc level, but I think I'd prefer to have the global set of commands not change between levels. Maybe we just rename the top-level `list` command to `blobs` and group the other blob-related commands (`add', a tbd `get` via RPC, possibly `export`) there.

* The REPL embeds all the existing RPC CLI commands. For this I changed the structure of the `src/cli/commands.rs` to a) split between `RpcCommands` and `FullCommands` , the latter are the ones that start an actual iroh node (plus doctor, wasn't sure about that for now). The former all work atop the RPC client. For them to not create a new RPC client for each REPL command, I moved the `--rpc-port` option to the top level. This is not super correct, because it does not apply to `provide`, `get`, `doctor`. Clap does not allow to scope an argument to a set of subcommands by default, see [this discussion](https://github.com/clap-rs/clap/discussions/5070#discussioncomment-6721310). Still thinking about what the cleanest solution is.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


b5:
Great stuff overall @Frando! I've been writing a Quickstart guide for working with the REPL (preview [here](https://iroh-computer-git-nextjs-number-0.vercel.app/docs/quickstart), and have a few notes aimed at cutting down the number of steps a first time user needs to take to "get to the good stuff"

1. The REPL should automatically create and "set" an author on startup when one isn't detected. This matches the keypair creation that happens when an iroh node starts. Setting a key allows us to defer the entire conversation about authors when first introducing users to documents.
2. I think we can dramatically simplify the command list & structure:

Current PR:
```
Commands:
  start-sync  Start to synchronize a document with peers
  share       Share a document and print a ticket to share with peers
  set         Set an entry
  get         Get entries by key
  list        List all entries in the document
  set-doc     Open a document
  set-author  Set the active author for doc insertion
  close       Close the open document
  exit        Quit the Iroh console
  author      Manage document authors
    list        List authors
    create      Create a new author
    help        Print this message or the help of the given subcommand(s)
  docs        Manage documents
    list        List documents
    create      Create a new document
    import      Import a document from peers
    help        Print this message or the help of the given subcommand(s)
  doc         Manage a single document
    start-sync  Start to synchronize a document with peers
    share       Share a document and print a ticket to share with peers
    set         Set an entry
    get         Get entries by key
    list        List all entries in the document
    help        Print this message or the help of the given subcommand(s)
  help        Print this message or the help of the given subcommand(s)
```

After:
```
Commands:
  sync        Start to synchronize a document with peers
  share       Share a document and print a ticket to share with peers
  set         Set an entry
  get         Get entries by key
  keys        List all entries in the document
  open        Open a document
  close       Close the active document, remove update subscriptions
  exit        Quit the Iroh console
  init        Create a new document
  import      Import a document from peers
  docs        List documents
  author      Manage document authors
    switch      Set the active author for doc insertion
    list        List authors
    create      Create a new author
    help        Print this message or the help of the given subcommand(s)
  help        Print this message or the help of the given subcommand(s)
```

Notable changes:
* I like your suggestion of `cd` instead of "doc-set". being able to `cd` _into_ documents would require we settle on a separator character for keys, which we should think carefully about. I think the command is still useful if it's just `change document`, and can't seek into doc prefixes.
* remove `doc` command entirely. the console itself is for working with documents
* switch `docs` command to be about listing, move all subcommands of `docs` up

Frando:
Thanks!

> The REPL should automatically create and "set" an author on startup when one isn't detected. This matches the keypair creation that happens when an iroh node starts. Setting a key allows us to defer the entire conversation about authors when first introducing users to documents.

Currently we do not have a built-in way to store that notion of a default author per node. I can think of 3 ways to do this:
* do a `default_author.pubkey` file in the `IROH_DATA_DIR`. Works well and is easy, but doesn't scale to other local state which we will get soon (e.g. a list of peers per document to join on node startup)
*  store it in a document. that would reduce the external state to a single pubkey with the namespace of this "node config doc". however it would expose a risk to share this document. which might also be a feature, but we should not store any secrets there.
* we could expose  `set_local_key/get_local_key` methods on the `iroh_sync::store::Store` trait and use those in `iroh` to set/retrieve the default author

> I think we can dramatically simplify the command list & structure:

that looks simpler indeed!

Frando:
We lean towards using the first option for now: store the default author's pubkey in a file in the iroh data dir. However, I have a new question: Should the "default author" for the repl be part of the node's state (so accessible over RPC) or a "console-only state" (read from the filesystem when starting the repl)? Leaning to the latter for now.

Frando:
With the recent changes, the PR now exposes the following strucure:

* `doc` is nested in both repl and cli
* `blob` and `node` are now nested too
* document author and document id are read from env vars `IROH_DOC` and `IROH_AUTHOR`  or from `-d`and `-a` opts on doc commands. if neither is set, an error is thrown with useful help text. in the repl, both can also be set via `set-author` and `set-doc` commands. `set-author` is persisted between sessions, `set-doc` is not.


Here's the help pages for both the CLI and the Console:



## CLI

```

$ iroh
Iroh is a tool for syncing bytes.
https://iroh.computer/docs

Usage: iroh [OPTIONS] <COMMAND>

Commands:
  console  Start the Iroh console
  start    Start a Iroh node
  get      Fetch data from a provider
  doctor   Diagnostic commands for the derp relay protocol
  doc      Manage documents
  author   Manage document authors
  blob     Manage blobs
  node     Manage a running Iroh node
  help     Print this message or the help of the given subcommand(s)

Options:
  -h, --help     Print help
  -V, --version  Print version

Options for console, doc, author, blob, node:
      --rpc-port <RPC_PORT>  RPC port of the Iroh node [default: 4919]

Options for start, get, doctor:
      --keylog                       Log SSL pre-master key to file in SSLKEYLOGFILE environment variable
      --metrics-addr <METRICS_ADDR>  Bind address on which to serve Prometheus metrics
      --cfg <CFG>

$ iroh help doc
Manage documents

Usage: iroh doc <COMMAND>

Commands:
  init    Create a new document
  import  Import a document from peers
  list    List documents
  sync    Start to synchronize a document with peers
  share   Share a document and print a ticket to share with peers
  set     Set an entry
  get     Get entries by key
  keys    List all entries in the document
  help    Print this message or the help of the given subcommand(s)

Options:
  -h, --help  Print help

```

## Console

```
$ iroh console
Welcome to the Iroh console!
Type `help` for a list of commands.

author:g633vys7
> help
Usage: <COMMAND>

Commands:
  set-doc     Set the active document
  set-author  Set the active author
  doc         Manage documents
  author      Manage document authors
  blob        Manage blobs
  node        Manage a running Iroh node
  exit        Quit the Iroh console
  help        Print this message or the help of the given subcommand(s)

author:g633vys7
> doc list
jpkdphd3gtoumwr5kjat547wacx6s7fgqtxs57jabpysqbl7n45a
j5tfpwkm4vb2o3iz2y3nsvf7teq4a7f4uufh3freqgdtdfmoyetq
pg4mlvf6efyxyepbmbi7wxi2sbxn522y7mbugfsne7oou6qtcxuq
ucipyhquqbmpsoqycoflaqf6nz5ahek4kepwaalfjv2cjkgh5ssa
366a42zagf6lbhpgn3w5qrqnwe7eu4gtoy2ecmxpes5xclqmvjqa
4cimkj5wn5y2bkfxxbylkxzobgpi5eaq45qdwhyq2ofgwusqoimq

author:g633vys7
> set-doc jpkdphd3gtoumwr5kjat547wacx6s7fgqtxs57jabpysqbl7n45a

author:g633vys7 doc:jpkdphd3
> doc set foo bazoo
@g633vys7…: foo = 3nfsmswq… (5 B)

author:g633vys7 doc:jpkdphd3
>
```

Frando:
I added `doc watch` and `node stats` commands in the recent commits.

Frando:
I pushed a few more commits.
Now `doc init --switch` and `author create --switch` work in the Iroh console to create & switch to the newly created doc/author! Should be great for the tutorials.

The `--switch` flag and the `doc switch` / `author switch` commands are only supported in the repl and will produce an error if used in the CLI.

for the cli: the init / create commands now only print the doc id/author id, so this should work:
```
export IROH_DOC=`iroh doc init`
export IROH_AUTHOR=`iroh author create`
iroh doc set foo bar
```
would be the same as this in the console:
```
doc init --switch
author create --switch
doc set foo bar
```
and there's useful errors now if author or doc is missing

iroh/src/commands.rs
```diff
@@ -55,66 +68,161 @@ pub struct Cli {
...
+    Rpc(#[clap(subcommands)] RpcCommands),
+}
+
+#[allow(clippy::large_enum_variant)]
+#[derive(Subcommand, Debug, Clone)]
+pub enum FullCommands {
+    /// Start a Iroh node
+    ///
+    /// If PATH is a folder all files in that folder will be served.  If no PATH is
+    /// specified reads from STDIN.
```
dignifiedquire, line 124: what about not serving anything? which will kinda be the default for using `sync`

iroh/src/commands.rs
```diff
@@ -159,132 +267,85 @@ impl Cli {
...
+        Ok(())
+    }
+}
+
+impl RpcCommands {
+    pub async fn run(self, client: RpcClient, env: ConsoleEnv) -> Result<()> {
+        match self {
+            Self::Node { command } => command.run(client).await,
+            Self::Blob { command } => command.run(client).await,
+            Self::Sync(command) => command.run(client, env).await,
```
dignifiedquire, line 341: would be nice to unify this to be also `Sync:: { command }`

iroh/src/commands/repl.rs
```diff
@@ -0,0 +1,168 @@
...
+#[derive(Debug, Parser)]
+pub enum ReplCmd {
+    #[clap(flatten)]
+    Rpc(#[clap(subcommand)] RpcCommands),
+    /// Quit the Iroh console
+    #[clap(alias = "quit")]
+    Exit,
+}
+
+fn try_parse_cmd<C: Subcommand>(s: &str) -> anyhow::Result<C> {
```
dignifiedquire, line 112: Can this handle sth like `doc set my_key hello world`? or at least `doc set my_key "hello world"`?

Frando: the latter should work, the former not. feel invited to try out some combinations. it uses the [shell_words](https://docs.rs/shell-words/latest/shell_words/) crate to split / escape the command line.

iroh/src/commands/sync.rs
```diff
@@ -1,124 +1,103 @@
...
 
 #[derive(Debug, Clone, Parser)]
-pub enum Doc {
-    StartSync {
-        peers: Vec<PeerSource>,
+pub enum DocCommands {
+    /// Set the active document (only works within the Iroh console).
+    Switch { id: NamespaceId },
+    /// Create a new document.
+    Init {
```
dignifiedquire, line 48: why "init" and not "new"?

Frando: @b5 proposed `init`. I had `create` before. Personally not sure what the best is between `new`, `create`, `init`

iroh/src/commands/sync.rs
```diff
@@ -147,33 +134,80 @@ pub enum Doc {
...
         prefix: Option<String>,
     },
+    /// Watch for changes and events on a document
+    Watch {
+        /// Document to operate on.
+        ///
+        /// Required unless the document is set through the IROH_DOC environment variable.
+        /// Within the Iroh console, the active document can also set with `doc switch`.
+        #[clap(short, long)]
+        doc_id: Option<NamespaceId>,
```
dignifiedquire, line 144: nit: I wonder if just `doc` would be a bit cleaner as argument name

iroh/src/commands.rs
Frando: Yes thought about that - I didn't change anything about the `start`  command apart from renaming from `provide` to `start` and adjusting docs a little. Maybe let's leave that for a followup.

iroh/src/commands/sync.rs
```diff
@@ -226,25 +262,96 @@ impl Doc {
...
+    /// List authors.
+    #[clap(alias = "ls")]
+    List,
+}
+
+impl AuthorCommands {
+    pub async fn run(self, iroh: &Iroh, env: ConsoleEnv) -> Result<()> {
+        match self {
+            Self::Switch { id } => {
+                env.set_author(id)?;
```
dignifiedquire, line 319: maybe this should print sth?

iroh/src/commands/sync.rs
```diff
@@ -226,25 +262,96 @@ impl Doc {
...
+            }
+            Self::List => {
+                let mut stream = iroh.list_authors().await?;
+                while let Some(author_id) = stream.try_next().await? {
+                    println!("{}", author_id);
```
dignifiedquire, line 324: could we print out an indication if we have a secret key for the author? maybe sth like

```
@<author_id>   write ✔️ 
@<author_id>   write ✖️ 
```

iroh/src/commands/sync.rs
```diff
@@ -226,25 +262,96 @@ impl Doc {
...
+    let author = fmt_short(id.author().as_bytes());
+    let hash = entry.record().content_hash();
+    let hash = fmt_short(hash.as_bytes());
+    let len = HumanBytes(entry.record().content_len());
     format!("@{author}: {key} = {hash} ({len})",)
 }
 
-fn fmt_hash(hash: impl AsRef<[u8]>) -> String {
+fn fmt_short(hash: impl AsRef<[u8]>) -> String {
     let mut text = data_encoding::BASE32_NOPAD.encode(&hash.as_ref()[..5]);
```
dignifiedquire, line 355: isn't this already implemented somewhere else?

iroh/src/config.rs
```diff
@@ -163,6 +262,144 @@ impl Config {
     }
 }
 
+/// Environment for CLI and REPL
+///
+/// This is cheaply cloneable and has interior mutability. If not running in the console
+/// environment, [Self::set_doc] and [Self::set_author] will lead to an error, as changing the
+/// environment is only supported within the console.
+#[derive(Clone, Debug)]
+pub struct ConsoleEnv(Arc<Mutex<ConsoleEnvInner>>);
```
dignifiedquire, line 271: it seems that most commands will be reading from this, so maybe worth using an `RwLock`?

iroh/src/commands.rs
dignifiedquire: okay, lets have an issue about this though, as I think this is important to do sooner rather than later

iroh/src/commands/sync.rs
Frando: It only lists authors for which we have a secret key. `iroh_sync::Store::create_author/list_authors` only ever cares about full author keypairs. do we exepct this to change?
to list non-write authors, we'd have to iterate over all entries of all documents, wouldn't we?

dignifiedquire: you are write, lets just document that a little clearer I guess 😅

iroh/src/commands/sync.rs
dignifiedquire: @b5 thoughts?

iroh/src/commands.rs
Frando: See https://github.com/n0-computer/iroh/issues/1403

iroh/src/config.rs
Frando: Fixed

iroh/src/commands/sync.rs
Frando: Fixed

iroh/src/commands/sync.rs
Frando: Fixed

iroh/src/commands.rs
Frando: Done :+1:

dignifiedquire (approved):

PR #1359, opened 2023-08-15
feat: methods to check if a hash is complete or partial

Frando:
## Description

This adds the methods `contains_complete` and `contains_partial` to `baomap::Map`, and `is_complete` to `baomap::Map::Entry`.

## Change checklist

- [x] Self-review.
- ~~Tests if relevant.~~

### DIFF
--- a/iroh/src/baomap/flat.rs
+++ b/iroh/src/baomap/flat.rs
@@ -266,6 +266,10 @@ impl MapEntry<Store> for PartialEntry {
         }
         .boxed()
     }
+
+    fn is_complete(&self) -> bool {
+        false
+    }
 }
 
 impl PartialMapEntry<Store> for PartialEntry {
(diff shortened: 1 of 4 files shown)


iroh/src/baomap/flat.rs
```diff
@@ -590,6 +600,7 @@ impl Map for Store {
                 hex::encode(entry.uuid)
             );
             Some(Entry {
+                is_complete: false,
```
rklaehn, line 603: hash first reads nicer I think

iroh/src/baomap/flat.rs
```diff
@@ -601,6 +612,16 @@ impl Map for Store {
             None
         }
     }
+
+    fn contains_complete(&self, hash: &Hash) -> bool {
```
rklaehn, line 616: You think it is worth it to have these two? I was thinking you would just call get for a membership test. Or maybe have a fn that takes a hash and returns an enum Complete/Partial/NotFound?

rklaehn (approved):

iroh/src/baomap/flat.rs
Frando: I changed it to an enum. Thought about this too and agree that it makes more sense than to have to methods.

PR #1365, opened 2023-08-16
test: Introduce iroh-test with common logging infrastructure

flub:
## Description

This code was copy-pasted around too much.  This takes a simple and
clean approach to this by creating a new crate which is used as a
dev-dependency of all the other crates.

The upside is there is no messing about with cargo features and
delicate matching versions.

The main downside is that it can not provide testing utilities which
build on the code in the other crates.  Most notable this is currently
the testing derp server.

## Notes & open questions

This skirts around the tricky bits of adding such a common module: The
testing derp server is still copy-pasted around.  I think this is
fine, because solving that is ugly.  Let's see if any other code ends
up in this crate or if the majority needs a more complicated solution
anyway.

This also doesn't introduce a general *published* iroh-common crate
(or whatever it would be named).  This might also be useful but
affects the public API and introduces other questions.  There's no
harm in considering this to be an entirely independent problem and to
later add such a module as well as this one.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


ramfox (approved):
F-yes, this was so annoying, thank you for taking the initiative to fix it.

iroh-test/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "iroh-test"
+version = "0.5.1"
+edition = "2021"
+readme = "README.md"
+license = "MIT OR Apache-2.0"
+authors = ["n0 team"]
+repository = "https://github.com/n0-computer/iroh"
+publish = false
```
dignifiedquire, line 9: I believe we have to publish it, otherwise we can't publish the other crates, dev dependencies need to be available

iroh-gossip/Cargo.toml
```diff
@@ -34,10 +34,11 @@ once_cell = "1.18.0"
 genawaiter = { version = "0.99.1", default-features = false, features = ["futures03"] }
 
 [dev-dependencies]
-tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 clap = { version = "4", features = ["derive"] }
-url = "2.4.0"
+iroh-test = { path = "../iroh-test" }
 rand_chacha = "0.3.1"
+tracing-subscriber = { version = "0.3", features = ["env-filter"] }
```
dignifiedquire, line 40: is this direct dependency still needed?

flub: Yes, because of examples.  But good catch, some others are no longer needed and I cleaned them up.

iroh-test/Cargo.toml
flub: oh, that's sad

flub: fixed

iroh-gossip/Cargo.toml
```diff
@@ -34,10 +34,11 @@ once_cell = "1.18.0"
 genawaiter = { version = "0.99.1", default-features = false, features = ["futures03"] }
 
 [dev-dependencies]
-tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 clap = { version = "4", features = ["derive"] }
-url = "2.4.0"
+iroh-test = { path = "../iroh-test" }
```
dignifiedquire, line 38: these should have the current version as well, makes it easier when I publish things

iroh-net/Cargo.toml
```diff
@@ -90,6 +90,7 @@ pretty_assertions = "1.4"
 rand_chacha = "0.3.1"
 tokio = { version = "1", features = ["io-util", "sync", "rt", "net", "fs", "macros", "time", "test-util"] }
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
+iroh-test = { path = "../iroh-test" }
```
dignifiedquire, line 93: version

iroh-test/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "iroh-test"
+version = "0.5.1"
+edition = "2021"
+readme = "README.md"
+license = "MIT OR Apache-2.0"
+authors = ["n0 team"]
+repository = "https://github.com/n0-computer/iroh"
+publish = true
+
```
dignifiedquire, line 10: missing rust version

iroh-test/README.md
```diff
@@ -0,0 +1,4 @@
+# iroh-test
+
+This crate contains testing support code for iroh.  It is only used
+for running the iroh test suite.
```
dignifiedquire, line 4: should have the license footer as in the other readmes

dignifiedquire: also needs license files

iroh/Cargo.toml
```diff
@@ -64,6 +64,7 @@ anyhow = { version = "1", features = ["backtrace"] }
 bytes = "1"
 duct = "0.13.6"
 genawaiter = { version = "0.99", features = ["futures03"] }
+iroh-test = { path = "../iroh-test" }
```
dignifiedquire, line 67: version

iroh-test/Cargo.toml
dignifiedquire, line 10: missing description

iroh-test/README.md
flub: > also needs license files

We only have license files in the workspace though?

dignifiedquire: hmm, you are right, then lets leave it as is for now

iroh-test/Cargo.toml
```diff
@@ -3,11 +3,15 @@ name = "iroh-test"
...
 edition = "2021"
 readme = "README.md"
+description = "Internal utilities to support testing of iroh."
 license = "MIT OR Apache-2.0"
 authors = ["n0 team"]
 repository = "https://github.com/n0-computer/iroh"
 publish = true
 
+# Sadly this also needs to be updated in .github/workflows/ci.yml
+rust-version = "1.66"
```
dignifiedquire, line 13: https://github.com/n0-computer/iroh/blob/main/iroh/Cargo.toml#L13 this is 1.67

dignifiedquire (approved):

PR #1372, opened 2023-08-17
docs(iroh-bytes): Improve range-spec docs

flub:
## Description

This attempts to improve the range-spec docs to be a bit more verbose.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-bytes/src/protocol/range_spec.rs
+++ b/iroh-bytes/src/protocol/range_spec.rs
@@ -83,6 +83,8 @@ impl RangeSpec {
     ///
     /// The [`RangeSet2`] is the same as a [`RangeSet`] but is used because it can store up
     /// to two 2 span boundaries without allocating.
+    ///
+    /// [`RangeSet`]: range_collections::RangeSet
     pub fn to_chunk_ranges(&self) -> RangeSet2<ChunkNum> {
         // this is zero allocation for single ranges
         // todo: optimize this in range collections


flub:
Now ready for full review.  Please check correctness (some examples have changed!) and how easy it is to understand.

flub (commented):
LGTM I think, but let's wait for @rklaehn

divagant-martian:
for the record I tried to make the examples doc test ones, but gave up trying to compare the objects. Something like 
```rust
        let spec = RangeSpec::from([2, 5, 3, 1]);
        let ranges = spec.to_chunk_ranges();
        let first_range: RangeSet2<_> = (ChunkNum(2)..ChunkNum(7)).into();
        let second_range: RangeSet2<_> = (ChunkNum(10)..ChunkNum(11)).into();
        let mut iter = ranges.iter();
        assert_eq!(iter.next(), Some(first_range));
        assert_eq!(iter.next(), Some(second_range));
        assert_eq!(iter.next(), None);
```

I think the current state is a really good improvement tbh, a doc test would make it clearer but it's not necessary

iroh-bytes/src/protocol/range_spec.rs
```diff
@@ -97,27 +112,45 @@ impl fmt::Debug for RangeSpec {
...
+///
+/// To select chunks in a collection this is encoded as a sequence of `(blob_offset, range_spec)`
+/// tuples. Offsets are interpreted in an accumulating fashion.
+///
+/// ## Example:
+///
+/// Supose two [`RangeSpec`]s `range_a` and `range_b`.
+///
+/// - `[(0, range_a), (2, empty), (3, range_b), (1, empty)]` encodes:
+///   - Select `range_a` for children in the range `[0, 2)`
```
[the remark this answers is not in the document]
divagant-martian: whatever you guys decide

iroh-bytes/src/protocol/range_spec.rs
```diff
@@ -1,32 +1,47 @@
-//! Specifications for individual ranges within a blob, and for ranges for a
-//! collection and its children.
+//! Specifications for ranges selection in blobs and collections.
+//!
+//! The [`RangeSpec`] allows specifying which BAO chunks inside a single blob should be
+//! selected.
+//!
+//! The [`RangeSpecSeq`] builds on top of this to select blob chunks in an entire
+//! collection.
```
rklaehn, line 7: "...and its childen" ?

iroh-bytes/src/protocol/range_spec.rs
```diff
@@ -1,32 +1,47 @@
...
 ///
-/// All values except for the first one must be non-zero. The first value may be zero.
-/// Values are bao chunk numbers, not byte offsets.
+/// - `[2, 5, 3, 1]` encodes five spans, of which two are selected:
+///   - `[0, 0+2) = [0, 2)` is deselected.
+///   - `[2, 2+5) = [2, 7)` is selected.
+///   - `[7, 7+3) = [7, 10)` is deselected.
+///   - `[10, 10+1) = [10, 11)` is selected.
+///   - `[11, inf)` is deselected.
+///   Iterating such a [`RangeSpec`] would then produce the [`RangeSet2`] `RangeSet{2..7, 10..11}`
```
rklaehn, line 27: Converting from a RangeSpec to a RangeSet is not really iterating. Maybe "This RangeSpec can be converted to a RangeSet containing the selected ranges" ?

iroh-bytes/src/protocol/range_spec.rs
```diff
@@ -1,32 +1,47 @@
...
-/// The range 10..33 would be encoded as `[10, 23]`
-/// The empty range would be encoded as the empty array `[]`
-/// A full interval .. would be encoded as `[0]`
-/// A half open interval 15.. would be encoded as `[15]`
+/// ## Examples:
 ///
-/// All values except for the first one must be non-zero. The first value may be zero.
-/// Values are bao chunk numbers, not byte offsets.
+/// - `[2, 5, 3, 1]` encodes five spans, of which two are selected:
+///   - `[0, 0+2) = [0, 2)` is deselected.
```
rklaehn, line 22: "not selected" instead of "deselected"? You either select something or you don't. Using "selected" and "deselected" for me sounds like there is two different things you can do, when there is really just one thing you can either do or not do.

iroh-bytes/src/protocol/range_spec.rs
```diff
@@ -41,25 +56,25 @@ impl RangeSpec {
         Self(res)
     }
 
-    /// An empty range spec
+    /// A [`RangeSpec`] deselecting the entire blob.
```
rklaehn, line 59: selecting nothing from the blob?

It is called EMPTY because RangeSpec's are really just compact sets of integers, so this is the empty set.

iroh-bytes/src/protocol/range_spec.rs
```diff
@@ -97,27 +112,45 @@ impl fmt::Debug for RangeSpec {
     }
 }
 
-/// A compressed sequence of range specs
+/// A chunk range specification for a collection of blobs.
```
rklaehn, line 115: "sequence of blobs"? You could use this for specifying ranges for any sequence of blobs, and calling it a sequence makes it clear why the thing is called RangeSpec *Seq* ?

iroh-bytes/src/protocol/range_spec.rs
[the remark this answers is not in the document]
rklaehn: I think blobs is fine. They are not all children. It is really just a way to efficiently specify ranges for a sequence. It just so happens that the first element is the collection and all subsequent elements are children, but you might use this also for situations where you have a list of hashes from somewhere else.

iroh-bytes/src/protocol/range_spec.rs
```diff
@@ -167,7 +200,11 @@ impl RangeSpecSeq {
         Self(res)
     }
 
-    /// An infinite iterator of range specs
+    /// An infinite iterator of range specs for blobs in the collection.
+    ///
+    /// Each item yielded by the iterator is the [`RangeSpec`] for a blob in the collection.
```
rklaehn, line 205: I would leave out the notion of collections entirely. It is really simplest to think of this as just a fancy/efficient way to write down a (possibly infinite) sequence of RangeSets.

The way this is mapped to the notion of collections is that the first element is the collection itself, and all subsequent elements are children of the collection. But e.g. if we ever have collections in collections you might have a different scheme.

iroh-bytes/src/protocol/range_spec.rs
```diff
@@ -177,19 +214,23 @@ impl RangeSpecSeq {
...
     }
 }
 
 static EMPTY_RANGE_SPEC: RangeSpec = RangeSpec::EMPTY;
 
-/// An infinite iterator of range specs
+/// An infinite iterator yielding [`RangeSpec`]s for each blob in a collection.
 ///
-/// default is what to use if the children of this RequestRangeSpec are empty.
+/// The first item yielded is the [`RangeSpec`] for the first blob in the collection, the
```
rklaehn, line 232: This is not correct. The first item yielded is the RangeSpec for the collection itself.

This is needed e.g. when you already have the collection and don't want to download it again.

But as I mentioned above, I think it is best to leave out the notion of collections and just describe this as a sequence of `RangeSet`s.

rklaehn:
I think we should probably make a note somewhere why the thing is called `RangeSet2`. So basically in a few places we accept generic `RangeSet`s. RangeSet2 is just a RangeSet that can contain up to 2 boundaries without allocating.

iroh-bytes/src/protocol/range_spec.rs
rklaehn: thinking about this some more I think we should really leave the notion of a collection out of this, and explain how this is used to encode collections and its children in a different place.

iroh-bytes/src/protocol/range_spec.rs
flub: +1

iroh-bytes/src/protocol/range_spec.rs
flub: so i agree that "sequence of blobs" is probably a better description since when using "collection" you already are assuming special treatment of the first item/blob in the collection, which is not what I intended.  I fear "sequence of RangeSets" is a bit too abstract.

flub:
> I think we should probably make a note somewhere why the thing is called `RangeSet2`. So basically in a few places we accept generic `RangeSet`s. RangeSet2 is just a RangeSet that can contain up to 2 boundaries without allocating.

this would be great to add!

iroh-bytes/src/protocol/range_spec.rs
rklaehn: So, how about this: we remove collection / children from both the docs and the API, and then add another section where we describe where a RangeSpecSeq is used in the case of collections and children (in a subsequent PR)?

I also would like to make the canonical repr mandatory. Without that you can't have structural equality. But better in the next PR.

flub: sounds good

rklaehn (approved):

iroh-bytes/src/protocol/range_spec.rs
flub: fixed

iroh-bytes/src/protocol/range_spec.rs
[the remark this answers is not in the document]
flub: changed to "sequence of blobs"

iroh-bytes/src/protocol/range_spec.rs
flub: fixed

iroh-bytes/src/protocol/range_spec.rs
flub: fixed and added

iroh-bytes/src/protocol/range_spec.rs
flub: no more collections mentioned

flub:
@rklaehn one question not yet answered is the `RangeSpecSeq::single` API (this was only mentioned in chat).  Why does it return an offset if it checks that this offset is 0 anyway?

rklaehn:
> @rklaehn one question not yet answered is the `RangeSpecSeq::single` API (this was only mentioned in chat). Why does it return an offset if it checks that this offset is 0 anyway?

I think it is just a bug. It should return the offset of the single non-zero element, instead it only returns Some only if the offset of the single non-zero element is 0, which makes it redundant.

But maybe merge this and we can do a second pass?

PR #1373, opened 2023-08-17
[WIP] refactor(iroh-net): unify key handling

dignifiedquire:
## Description

Moves all keys being used to `ed25519` keys, and only converts them to `crypto_box` keys only when needed. 
This is a breaking change to the derp protocol, as now only `ed25519` keys are sent over the wire. 

## Notes & open questions

This is more expensive as the conversions are not cached at the moment, we should do so either in this PR or in a follow up.

Depends on https://github.com/RustCrypto/nacl-compat/pull/137 as otherwise the upgrade of `ed25519` breaks the code.


iroh-net/src/derp/client.rs
```diff
@@ -61,7 +61,7 @@ impl Client {
     ///
     /// Errors if the packet is larger than [`super::MAX_PACKET_SIZE`]
     pub async fn send(&self, dstkey: PublicKey, packet: Bytes) -> Result<()> {
-        debug!(%dstkey, len=packet.len(), "[DERP] send");
+        debug!(?dstkey, len = packet.len(), "[DERP] send");
```
flub, line 64: Could we impl Display again on the keys?  I think before we had these as a short printable version, ~8 hex chars, so that it would be suitable for logging purposes.  It makes logging much easier to read.

iroh-net/src/disco.rs
```diff
@@ -25,9 +25,9 @@ use std::{
 
 use anyhow::{anyhow, ensure, Result};
 
-use crate::net::ip::to_canonical;
+use crate::{key, net::ip::to_canonical};
 
-use super::{key, stun};
+use super::{key::PublicKey, stun};
```
flub, line 30: this `super` is a bit weird, importing from both super::key and crate::key which is the same module.  I'd say move everything to `crate::`.

iroh-net/src/key/encryption.rs
```diff
@@ -0,0 +1,153 @@
...
+        let plaintext = self
+            .0
+            .decrypt(&nonce.into(), ciphertext)
+            .map_err(|e| anyhow!("decryption failed: {:?}", e))?;
+
+        Ok(plaintext)
+    }
+}
+
+impl crate::key::Keypair {
```
flub, line 62: maybe `super::Keypair`?

iroh-net/src/key.rs
```diff
@@ -1,2 +1,183 @@
...
+    Signature, SigningKey as SecretKey, VerifyingKey as PublicKey, PUBLIC_KEY_LENGTH,
+};
+use serde::{Deserialize, Serialize};
+use ssh_key::LineEnding;
+
+pub use self::encryption::SharedSecret;
+
+/// A keypair.
+#[derive(Clone, Debug)]
+pub struct Keypair {
```
flub, line 20: nit, should this be `KeyPair` to be consistent rust naming?

iroh-net/src/key/encryption.rs
```diff
@@ -0,0 +1,153 @@
...
+
+pub(crate) const NONCE_LEN: usize = 24;
+
+fn public_ed_box(key: &ed25519_dalek::VerifyingKey) -> crypto_box::PublicKey {
+    crypto_box::PublicKey::from(key.to_montgomery())
+}
+
+fn secret_ed_box(key: &ed25519_dalek::SigningKey) -> crypto_box::SecretKey {
+    crypto_box::SecretKey::from(key.to_scalar())
+}
```
flub, line 15: how expensive are these two functions?  maybe leave a note about that in (doc) comments so people understand the tradeoffs made here and what to keep in mind when refactoring in the future

flub (commented):
nice simplification overall!

maybe we can use alternate formatting (`:#`) for short-loggable versions of Display for everything?  Though I'm not sure that's easy to use with tracing fields so maybe that doesn't help much and a `.short_fmt()` or so method is better.

iroh-net/src/derp/client.rs
dignifiedquire: only keypair is a newtype atm, ao only for that

iroh-net/src/key/encryption.rs
dignifiedquire: super doesn‘t work, that is the top level already

iroh-net/src/key/encryption.rs
dignifiedquire: they are not super cheap unfortunately..though I need to measure them

iroh-net/src/key.rs
dignifiedquire: maybe, i also don’t like having Keypair, SecretKey and PublicKey, so I might combine this with SecretKey

dignifiedquire: called `SecretKey` now

iroh-net/src/key/encryption.rs
dignifiedquire: caching added

iroh-net/src/derp/client.rs
dignifiedquire: for now both debug and display print the long version

iroh-net/src/key.rs
```diff
@@ -1,2 +1,333 @@
...
+impl Display for PublicKey {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let mut text = data_encoding::BASE32_NOPAD.encode(self.as_bytes());
+        text.make_ascii_lowercase();
+        write!(f, "{text}")
+    }
+}
+
+/// A secret key.
+// TODO: rename to `SecretKey`.
```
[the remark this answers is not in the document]
dignifiedquire: fixed

iroh-net/src/key.rs
```diff
@@ -1,2 +1,333 @@
...
+
+pub use self::encryption::SharedSecret;
+use self::encryption::{public_ed_box, secret_ed_box};
+
+/// A public key.
+#[derive(Clone, Copy, Hash, PartialEq, Eq)]
+pub struct PublicKey {
+    public: VerifyingKey,
+    /// Cached version of `crypto_box::PublicKey` matching `public`.
+    public_crypto_box: [u8; 32],
```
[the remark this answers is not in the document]
dignifiedquire: fixed

iroh-gossip/examples/chat.rs
```diff
@@ -87,15 +87,12 @@ async fn main() -> anyhow::Result<()> {
...
+    // parse or generate our secret key
+    let secret_key = match args.secret_key {
+        None => SecretKey::generate(),
+        Some(key) => parse_secret_key(&key)?,
     };
-    println!(
-        "> our private key: {}",
-        base32::fmt(keypair.secret().to_bytes())
-    );
+    println!("> our private key: {}", base32::fmt(secret_key.to_bytes()));
```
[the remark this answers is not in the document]
dignifiedquire: fixed

flub (approved):
lgtm I think, very thorough job on the renaming.  I almost feel bad suggesting another name change... but feel free to disagree with that.

iroh-gossip/src/net.rs
```diff
@@ -41,21 +41,21 @@ const TO_ACTOR_CAP: usize = 64;
 const IN_EVENT_CAP: usize = 1024;
 
 /// Events emitted from the gossip protocol
-pub type Event = proto::Event<PeerId>;
+pub type Event = proto::Event<PublicKey>;
```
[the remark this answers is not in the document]
dignifiedquire: I am not saying no, but I would like to discuss & change in a follow up, can you make a PR for the bikeshedding to start?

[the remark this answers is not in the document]
flub: https://github.com/n0-computer/iroh/issues/1388

PR #1375, opened 2023-08-18
feat: get `NodeInfo`s from the `MagicEndpoint`

ramfox:
## Description

Call `MagicEndpoint::node_infos()` to get a `Vec<NodeInfo>`, or list connection information on all nodes we know about.

`NodeInfo` is aliased from `super::magicsock::EndpointInfo`.

```rust
/// The type of connection we have to the node.
#[derive(Debug, Clone, Eq, PartialEq)]
pub enum ConnectionType {
    /// Direct UDP connection
    Direct(SocketAddr),
    /// Relay connection over DERP
    Relay(u16),
    /// We have no verified connection to this peer
    None,
}

/// Details about an Endpoint
#[derive(Debug, Clone, Eq, PartialEq)]
pub struct EndpointInfo {
    /// The id in the peer_map
    pub id: usize,
    /// The public key of the endpoint.
    pub public_key: PublicKey,
    /// Derp region, if available.
    pub derp_region: Option<u16>,
    /// List of addresses at which this node might be reachable, plus any latency information we
    /// have about that address.
    pub addrs: Vec<(SocketAddr, Option<Duration>)>,
    /// The type of connection we have to the peer, either direct or over relay.
    ///
    /// Depending on the expiry time, we may still have a direct connection to the peer
    /// but our best address for the peer may have timed out.
    /// Until we verify that connection, we assume that we may only have a relay
    /// connection to that peer.
    pub conn_type: ConnectionType,
    /// The latency of the `conn_type`.
}

```

## Notes & open questions

Please let me know if the above `EndpointInfo` + `ConnectionType` is enough for our UI and sync purposes!

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/src/magic_endpoint.rs
+++ b/iroh-net/src/magic_endpoint.rs
@@ -20,7 +20,7 @@ use crate::{
     tls,
 };
 
-type PeerInfo = super::magicsock::EndpointInfo;
+type NodeInfo = super::magicsock::EndpointInfo;
 
 /// Builder for [MagicEndpoint]
 #[derive(Debug)]
@@ -291,10 +291,10 @@ impl MagicEndpoint {
 
     /// Get information on all the peers we have connection information about.
     ///
-    /// Includes the peer's [`PublicKey`], potential DERP region, it's addresses with any known
+    /// Includes the node's [`PublicKey`], potential DERP region, it's addresses with any known
     /// latency, and its [`ConnectionType`], which let's us know if we are currently communicating
-    /// with that peer over a `Direct` (UDP) or `Relay` (DERP) connection.
-    pub async fn peer_infos(&self) -> anyhow::Result<Vec<PeerInfo>> {
+    /// with that node over a `Direct` (UDP) or `Relay` (DERP) connection.
+    pub async fn node_infos(&self) -> anyhow::Result<Vec<NodeInfo>> {
         self.msock.tracked_endpoints().await
     }
 
--- a/iroh-net/src/magic_endpoint.rs
+++ b/iroh-net/src/magic_endpoint.rs
@@ -292,7 +292,7 @@ impl MagicEndpoint {
     /// Get information on all the peers we have connection information about.
     ///
     /// Includes the node's [`PublicKey`], potential DERP region, it's addresses with any known
-    /// latency, and its [`ConnectionType`], which let's us know if we are currently communicating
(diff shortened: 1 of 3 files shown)


b5:
We've been shifting away from `Peer` as a term since https://github.com/n0-computer/iroh/pull/137  , what do you think about `NodeInfo`?

iroh-net/src/magic_endpoint.rs
```diff
@@ -287,6 +289,15 @@ impl MagicEndpoint {
         self.msock.my_derp().await
     }
 
+    /// Get information on all the peers we have connection information about.
+    ///
+    /// Includes the node's [`PublicKey`], potential DERP region, it's addresses with any known
+    /// latency, and its [`crate::magicsock::ConnectionType`], which let's us know if we are currently communicating
+    /// with that node over a `Direct` (UDP) or `Relay` (DERP) connection.
+    pub async fn node_infos(&self) -> anyhow::Result<Vec<NodeInfo>> {
```
Frando, line 297: nit: maybe `active_connections` or so? `node_infos` to me sounds a bit more static than what this method does. but also not sure myself, it's fine to leave it as is.

dignifiedquire:
`NodeInfo` is very confusing to me, I think it should be at least sth like `RemoteNodeInfo`, otherwise I would assume it is the local node

iroh-net/src/magic_endpoint.rs
ramfox: I did some reading. We *do* delete endpoints, but it is currently only triggered by user action. Any time we `set_network_map`, we remove any connections that were made to us "in the wild". 

But the only method that calls `set_network_map` is `add_known_addrs`. Which... is an interesting side effect that we haven't really surfaced. I'm not sure that we intended to have every time we add a new address to the netmap, that it should clear all connections made by other nodes to us.

However, since this happens only on user action, we are likely carrying around a bunch of "inactive" connections, so until / unless we change this, I don't think `active_connections` is the best name.

@b5 I think `ConnectionInfo` might be the right name for the `EndpointInfo` outside of the `MagicSock`.

which would make this `connection_infos` & I will add the above context into the function docs for both `connection_infos` and `add_known_addrs`.

ramfox:
Out of band discussion, we settled on `ConnectionInfo`

ramfox:
closing this PR in favor of #1435

Issue #1358, opened 2023-08-15
Magicsock: optimize reception of UDP packets from derper on linux


rklaehn:
When udp packets are received from the derper, on linux there is the possibility of 

a) using multiple slots in poll_recv
b) using multiple udp packets per slot

While we use a), we don't use b). We just put 1 packet into each slot and set the segment_size to the size of the entire packet. I have some doubts that this works at all, but even if it works it is less than optimal.

Each slot should be filled as much as possible before moving on to the next slot - that way the total number of udp packets you can handle in a single poll_recv call is maximized.

rklaehn:
Related to https://github.com/n0-computer/iroh/issues/1357

flub:
I think in general we might be losing the GSO optimsation on the send side.  Which may be the same as you're describing here.  I wanted to create an issue but need to check if this is really true first.

rklaehn:
I doubt this is very critical perf wise.

But in principle on the linux side you could suck tons of UDP packets from the TCP stream and then fill a **lot** of slots in a single poll_recv call. There are 32 slots, and each slot can hold up to ~80 UDP packets.

flub:
> I doubt this is very critical perf wise.

Maybe because you're on mac and there you never get segmentation offloading... But on linux this is know to have a huge perf impact and segmentation offloading is required to get speeds that are in the ballpark of TCP.

rklaehn:
> > I doubt this is very critical perf wise.
> 
> Maybe because you're on mac and there you never get segmentation offloading... But on linux this is know to have a huge perf impact and segmentation offloading is required to get speeds that are in the ballpark of TCP.

Yeah, I mean it is not perf critical because once you use the derper perf is not going to be that great anyway... :-)

rklaehn:
There is something that causes the derper to constantly disconnect. We need to fix this before even thinking about doing this issue... https://github.com/n0-computer/iroh/issues/1381

dignifiedquire:
what is left here to do?

PR #1376, opened 2023-08-18
fix(iroh-net): Split GSO packts received from derper

### LINKED ISSUE #1357: Check that magicsock via derp works from linux to mac - handling of segment_size
When working on some quinn testing infrastructure in quic-rpc, I noticed the following:

- on linux, you will often get transmits with the `segment_size` field set and `contents` larger than a single UDP packet.
- when in derper mode, as far as I can see the contents will just be sent to the derper as single blobs. See https://github.com/n0-computer/iroh/blob/1c3a3f131d2f394f18118a6a153ff8f0f29e9a83/iroh-net/src/magicsock.rs#L1220 .

There are two issues with this:
1. on linux, the receive buffers that you get from quinn in poll_recv are really large, almost 100000 bytes. But on mac they are quite small, something like 1500 bytes. So why can't the place where the receive buffer is filled from whatever we get from derp here https://github.com/n0-computer/iroh/blob/1c3a3f131d2f394f18118a6a153ff8f0f29e9a83/iroh-net/src/magicsock.rs#L749 panic?
2. in my experiments in quic-rpc I found that quinn gets confused if the packet boundaries get mixed up, and you will get errors such as "packet auth failed". E.g. you send 2400 bytes with a segment size of 1200 bytes, that's two packets. But if you now put this into the buffer on the recv side and set the stride to the entire part len, as we do here https://github.com/n0-computer/iroh/blob/1c3a3f131d2f394f18118a6a153ff8f0f29e9a83/iroh-net/src/magicsock.rs#L1159 , the receiver sees some weird single udp packet.

flub:
## Description

The datagrams received over the derper channel could be multiple datagrams combined by Generic Segmentation Offloading (GSO).  Normally these would go over the network as individual datagrams and the receiving machine would use GRO if it had it and all would look normal for the application code reading from the sockets.

However when sending datagrams over the derper channel they arrive at the receiver in whichever way the GSO of the sender assembled them. We could stop that from happening, but we might as well use GSO/GRO over the derper channel so we benefit from it there too.

So instead this manually implements partial GRO on the receiver side. It splits up the received datagrams into small enough chunks so they can be received by the host's IoSlices (quinn configures these smaller depending on the host's limit so save wasting memory).  This can still read multiple datagrams in as one packet, but the RecvMeta's stride is set correctly and Quinn handles this fine.  So it delays splitting into individual datagrams as long as possible.

Note that this is kind of a hot-fix.  We might want to optimise GSO/GRO paths through magicsock a bit more and this could end up being refactored in different ways.  See also #1358.

Fixes #1357

## Notes & open questions

This is my first draft of the logic needed here, not yet tested.  But it is good enough to give it a spin in the manual testing and see if it works.

## Todo

- [ ] Unit tests to verify.
- [ ] Try and remove the mutex, if possible :sob: .
- [ ] See if code de-duplication is possible.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/src/magicsock.rs
+++ b/iroh-net/src/magicsock.rs
@@ -23,6 +23,7 @@ use bytes::Bytes;
 use futures::future::BoxFuture;
 use iroh_metrics::{inc, inc_by};
 use quinn::AsyncUdpSocket;
+use quinn_udp::RecvMeta;
 use rand::{seq::SliceRandom, Rng, SeedableRng};
 use tokio::{
     sync::{self, mpsc, Mutex},
@@ -216,6 +217,19 @@ pub(self) struct Inner {
     pub(self) derp_map: Option<DerpMap>,
     /// Nearest DERP region ID; 0 means none/unknown.
     my_derp: AtomicU16,
+
+    /// Raw datagrams which need to be returned on the next .poll_recv().
+    ///
+    /// Since we use GSO (Generic Segementation Offloading) on the send path we end up
+    /// sending concatenated datagrams over the DERP connection as well.  Since we have no
+    /// GRO (Generic Read Offloading) stack on the receiving side when using the DERP
+    /// connection we will receive the concatenated datagrams in the magicsock even if the
+    /// local machine does not support this, e.g. MacOS.  In this case Quinn will not give
+    /// us large enough buffers in .poll_recv() to put the entire received payload and we
+    /// need to split up the payload.  If we had to split the payload we store the
+    /// not-yet-received datagrams here and the next .poll_recv() will return these before
+    /// reading more from the [`Inner::network_recv_ch`].
+    cached_recv_datagrams: std::sync::Mutex<Option<(RecvMeta, Bytes)>>,
 }
 
 impl Inner {
@@ -351,6 +365,7 @@ impl MagicSock {
             ipv6_reported: Arc::new(AtomicBool::new(false)),
             derp_map,
             my_derp: AtomicU16::new(0),
+            cached_recv_datagrams: std::sync::Mutex::new(None),
         });
 
         let udp_state = quinn_udp::UdpState::default();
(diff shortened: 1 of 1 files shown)


iroh-net/src/magicsock.rs
```diff
@@ -742,13 +782,30 @@ impl AsyncUdpSocket for MagicSock {
...
+                            mut bytes,
+                            mut meta,
                             source,
                         } => {
-                            buf_out[..bytes.len()].copy_from_slice(&bytes);
+                            let copy_len = if bytes.len() > buf_out.len() {
+                                // split bytes on segment boundary, leaving the largest
+                                // possible number of segments for buf.  Store the remaining
+                                // segements for next round.
+                                let segment_count = buf_out.len() / meta.stride;
```
dignifiedquire, line 793: what if `meta.stride` is larger than `buf_out`?

iroh-net/src/magicsock.rs
```diff
@@ -742,13 +782,30 @@ impl AsyncUdpSocket for MagicSock {
...
                         NetworkReadResult::Ok {
-                            bytes,
-                            meta,
+                            mut bytes,
+                            mut meta,
                             source,
                         } => {
-                            buf_out[..bytes.len()].copy_from_slice(&bytes);
+                            let copy_len = if bytes.len() > buf_out.len() {
+                                // split bytes on segment boundary, leaving the largest
```
dignifiedquire, line 790: the splitting is already done on the IP recv path, so I would only run this path if the source is `derp`

iroh-net/src/magicsock.rs
```diff
@@ -717,6 +732,31 @@ impl AsyncUdpSocket for MagicSock {
...
         for (buf_out, meta_out) in bufs.iter_mut().zip(metas.iter_mut()) {
+            // first read from self.inner.cached_recv_datagrams
+            {
+                if self.inner.is_closed() {
+                    break;
+                }
+                let mut cached = self.inner.cached_recv_datagrams.lock().unwrap();
+                if let Some((mut meta, mut bytes)) = cached.take() {
+                    let copy_len = if bytes.len() > buf_out.len() {
+                        let segment_count = buf_out.len() / meta.stride;
```
rklaehn, line 743: Unfortunately the docs about RecvMeta are rather silent about the meaning of stride.

https://docs.rs/quinn-udp/latest/quinn_udp/struct.RecvMeta.html

But is this guaranteed to be a whole number, or could you have a situation where e.g. buf is 1500 bytes and stride is 1200 bytes, so you get 1 1200 byte packet and 1 300 byte packet?

At least on the send side I saw something like this - segment_size being 1200 and the total size being something that is not a multiple of the segment_size.

See the log in https://github.com/n0-computer/iroh/issues/1357#issuecomment-1679110798

rklaehn: I implemented the "split on send" thing here: https://github.com/n0-computer/iroh/pull/1380

iroh-net/src/magicsock.rs
flub: that would be a mis-configuration of quinn.  The QUIC MTU is 1200 bytes and quinn should make sure that it calls recv with enough space.

Though you're right that is is a corner case that should be handled.  Because what we should do in that case is return a partial datagram and return the rest of the datagram on the next poll.

iroh-net/src/magicsock.rs
flub: So RecvMeta docs are updated in the quinn main branch.

Your understanding is correct: the last datagram could be smaller than the stride/segment-size (i'm not sure why they call this something different on the send and receive paths).

flub:
So this approach has the problem that the send path of the derper does not tell us what the right segment/stide size should be.  The derper send path would have to be updated too and we'd change the derper protocol for that.

We may as well go with #1380 for the hotfix approach, even though I was hoping we would be able to at least not use the existing GSO when fixing this.

PR #1378, opened 2023-08-18
feat(iroh-net): add `DEV_DERP_ONLY` env variable for testing the derp relay

ramfox:
## Description
This forces the `MagicSock` to only ever *send* messages through a derp node.

However, this does not prevent the node from receiving messages over the udp sockets.

Hole-punching attempts and responses are prevented. 

------

Adjusts the `addr_for_send` method, ensuring that we only ever return the `DerpRegion` (if we know it) as a possible option for how to communicate with the given peer.

It also prevents `send_pings` and `start_pings`.

Both the provider and getter must use the `DEV_DERP_ONLY` flag for this to function. There is some bad interaction that causes blocks / stalling when we attempt to hole punch and use the relay.

-----

Fixes a bug around discovering derp regions. Now, when we receive a message from a `PublicKey` from a particular `DerpRegion`, and we previously had no derp information about this `PublicKey` before, we add that region ID to the `Endpoint` associated with that `PublicKey`.

Also fixes some tests that previously ignored adding derp information.

Now, if you run `DEV_DERP_ONLY=true cargo test`, all tests will pass with any connections strictly passing through the derp relay.

## Change checklist

- [x] Self-review.


iroh-net/src/magicsock.rs
```diff
@@ -287,6 +287,9 @@ impl MagicSock {
     ///
     /// [`Callbacks::on_endpoint`]: crate::magicsock::conn::Callbacks::on_endpoints
     pub async fn new(opts: Options) -> Result<Self> {
+        #[cfg(features = "derp-only")]
+        debug!("creating MagicSock that will only send packets over a derp relay connection.");
```
flub, line 291: why not make this `warn!`?  i mean, debug is fine too, but more prominent wouldn't hurt

flub (approved):
looks fine, but tests appear unhappy

iroh/src/main.rs
```diff
@@ -52,6 +52,7 @@ async fn main_impl() -> Result<()> {
     #[cfg(feature = "metrics")]
     let metrics_fut = init_metrics_collection(cli.metrics_addr, &rt);
 
+    tracing::debug!(target: "derponly", "{config:?}");
```
dignifiedquire, line 55: why print out the full config in this case?

ramfox: 🤦‍♀️ I could have sworn i removed that, sigh

iroh-net/src/magicsock.rs
ramfox: great idea

rklaehn:
How hard would it be to make this runtime configurable? It seems you would have to just have a flag check in `addr_for_send`. 

Something like 

```rust
if self.config.derp_only {
    return (None, self.derp_addr, false);
}
// the other stuff
```

I always find feature flags a bit of a maze, and also I can see scenarios where you would want to set this for an existing binary like `iroh` cli to troubleshoot something.

rklaehn (approved):

rklaehn:
> How hard would it be to make this runtime configurable? It seems you would have to just have a flag check in `addr_for_send`.
> 
> Something like
> 
> ```rust
> if self.config.derp_only {
>     return (None, self.derp_addr, false);
> }
> // the other stuff
> ```
> 
> I always find feature flags a bit of a maze, and also I can see scenarios where you would want to set this for an existing binary like `iroh` cli to troubleshoot something.

Never mind, I think a feature flag is fine. Making this a cli feature would require more thought about what it means to be "derp-only"

iroh-net/src/magicsock/endpoint.rs
```diff
@@ -151,6 +152,14 @@ impl Endpoint {
 
     /// Returns the address(es) that should be used for sending the next packet.
     /// Zero, one, or both of UDP address and DERP addr may be non-zero.
+    #[cfg(feature = "derp-only")]
+    fn addr_for_send(&mut self, _now: &Instant) -> (Option<SocketAddr>, Option<u16>, bool) {
```
rklaehn, line 156: I got a slight preference for doing the feature flag distinction only inside the fn. Just need to do something like `let _ = now` to get rid of the unused warning.

rklaehn:
Two things:

1. maybe it is better to translate the feature flag into a `const bool` and then from then on only use the const bool. It seems like you get fewer unused code warnings that way. See this: https://github.com/n0-computer/iroh/commit/fc41dfd78f5194898f8f03de5838ea64aa75aacc for inspiration.

2. I am still not entirely happy with using a feature flag for this. At least not the way it is done here. Feature flags are supposed to be additive, and this is a subtractive feature flag. So --all-features actually does less than the default feature set.

Not sure what to do about it. Use something else after all (env var?), or make it an additive feature flag that is enabled by default, like `use-direct-connections = []` that you would have to disable using --no-default-features?

flub (changes requested):
I don't think making `--all-features` a footgun like this is a good idea.

dignifiedquire:
From a feature flag perspective, in an ideal world we would have

- `derp-connections`: enables the usage of derp 
- `udp-connections`: enables the usage of udp based direct connections

They would be additive and could be used entirely independently, buuut this would require a lot of refactoring. So maybe we make this a runtime/config flag after all..

flub:
While having these two modes is great, I'm not sure they need to be feature flags.  I think feature flags only start making sense if the resulting code ends up being significantly smaller.  But we're just returning `None` somewhere, so feature flags are only additional maintenance and use burden.

dignifiedquire:
I meant that they would actually disable the whole machinery if implemented properly, which is why it would be a large refactor.

rklaehn:
> I meant that they would actually disable the whole machinery if implemented properly, which is why it would be a large refactor.

Ah, you mean with them both off magicsock would just be a quinn socket? Not sure how you would even do that, and if you do it it would be a large refactor.

I don't think it is a good idea. Better to have crates that just use quinn, like iroh-bytes, and then give people the option to plug in a real quinn endpoint or a magic endpoint.

ramfox:
Agree with all these criticisms. It’s been annoying working through the feature implications, including messing with `—all-features`. No bueno.

The purpose of this PR was to create a relatively simple way to test that the DERP protocol works as a relay *within* the `MagicEndpoint` flow (in the case where we have a insurmountable NAT problem). 

So part of the disconnect is I'm trying to solve for a *dev* issue when any adjustment we make to the code can be interpreted / used as a feature.

Eg, adding a field to the config or adding a feature flag implies some other use beyond testing. 

So, I’m thinking, env variable as per dig’s mention. `DEV_DERP_ONLY`. Name can be bike shedded.

This will prevent the features snafus and the unused code warnings. The code change would be a one-liner, it’ll be easy test manually, and we can add a few `DEV_DERP_ONLY` modes in critical tests, that way we can catch issues in the derp protocol that our *superior* 🤣 hole punching has been masking.

Thoughts?

ramfox:
Latest issues: https://github.com/n0-computer/iroh/issues/1404

Basically, I have proof that if we do "derp only" mode on both sides with no attempts at hole punching, I was able to send ~300mbs of data no problem (took at least a minute, but no issues).

When I tried the scenario where the Getter attempted to hole punch, we started getting errors. I outlined my theory in the above issue, but haven't investigated yet. Since this is the exact scenario we want to ensure works I want to take another day to try to solve the issue.

The current `derp_only` branch includes a temp commit that prevents hole punching attempts when using `DEV_DERP_ONLY` mode, if others want to try it out.

That should work just fine. But if you check out the previous commit [0aefc81](https://github.com/n0-computer/iroh/pull/1378/commits/0aefc814e67188d8f2dd873891c7a7ec541e461f), you will get errors when using `DEV_DERP_ONLY`.

iroh-net/src/magicsock.rs
```diff
@@ -1,5 +1,18 @@
 //! Implements a socket that can change its communication path while in use, actively searching for the best way to communicate.
 //!
+//! ### `DEV_DERP_ONLY` env var:
```
rklaehn, line 3: This is a compile time switch via an env var. I think this should note that the DEV_DERP_ONLY env var needs to be preset at compile time.

https://doc.rust-lang.org/std/macro.option_env.html

iroh-net/src/magicsock/endpoint.rs
```diff
@@ -149,9 +149,23 @@ impl Endpoint {
...
+
+    /// Adds a derp addr for this endpoint
+    pub fn add_derp_addr(&mut self, region: u16) {
+        self.derp_addr = Some(region);
+    }
+
     /// Returns the address(es) that should be used for sending the next packet.
     /// Zero, one, or both of UDP address and DERP addr may be non-zero.
     fn addr_for_send(&mut self, now: &Instant) -> (Option<SocketAddr>, Option<u16>, bool) {
+        if std::option_env!("DEV_DERP_ONLY").is_some() {
```
rklaehn, line 165: can we move the env check into a fn?

rklaehn (approved):
Looks fine so far, but I would like the check to be in a fn and to have a comment about this being a compile time switch.

flub (approved):
There's already a comment about it, but I assume that this env-var being compile-time is intentional.

iroh-net/src/magicsock/endpoint.rs
```diff
@@ -149,9 +149,23 @@ impl Endpoint {
...
+            trace!("in `DEV_DERP_ONLY` mode, giving the DERP address as the only viable address for this endpoint");
```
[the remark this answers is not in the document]
ramfox: I'm moving this particular one to `debug` because it happens so often that if we printed it as `info` it would flood the user.

The others are moved to `warn`. How does that sound?

PR #1380, opened 2023-08-21
fix(iroh-net): split packets on send

rklaehn:
this seems to be the safest way to fix possible GSO/GRO issues on all platforms (Linux and ???) that support them.

## Description

This is a fix for the possible issues described in https://github.com/n0-computer/iroh/issues/1357 that does the simplest possible thing - just split up the packets before sending them to the derper.

Note that we still get GRO/GSO when we have a direct connection. And since the stream to the derper is a byte stream, unless we flush after each send this should still be rather efficient. The only downside is that the derper has to deal with more individual packets.

## Notes & open questions

Does this fix the issue?

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/src/magicsock.rs
+++ b/iroh-net/src/magicsock.rs
@@ -1176,6 +1176,27 @@ impl Actor {
         out
     }
 
+    /// Split a number of transmits into individual packets.
+    ///
+    /// For each transmit, if it has a segment size, it will be split into
+    /// multiple packets according to that segment size. If it does not have a
+    /// segment size, the contents will be sent as a single packet.
+    fn split_packets(transmits: Vec<quinn_udp::Transmit>) -> Vec<Bytes> {
+        let mut res = Vec::with_capacity(transmits.len());
+        for transmit in transmits {
+            if let Some(segment_size) = transmit.segment_size {
+                let len = transmit.contents.len();
+                for min in (0..len).step_by(segment_size) {
+                    let max = (min + segment_size).min(len);
+                    res.push(transmit.contents.slice(min..max));
+                }
+            } else {
+                res.push(transmit.contents);
+            }
+        }
+        res
+    }
+
     async fn send_network(&mut self, transmits: Vec<quinn_udp::Transmit>) {
         trace!(
             "sending:\n{}",
@@ -1217,11 +1238,7 @@ impl Actor {
                 match ep.get_send_addrs().await {
                     Ok((Some(udp_addr), Some(derp_addr))) => {
                         let res = self.send_raw(udp_addr, transmits.clone()).await;
-                        self.send_derp(
-                            derp_addr,
-                            public_key,
-                            transmits.into_iter().map(|t| t.contents).collect(),
-                        );
+                        self.send_derp(derp_addr, public_key, Self::split_packets(transmits));
 
                         if let Err(err) = res {
                             warn!("failed to send UDP: {:?}", err);
@@ -1231,7 +1248,7 @@ impl Actor {
                         self.send_derp(
                             derp_addr,
                             public_key.clone(),
-                            transmits.into_iter().map(|t| t.contents).collect(),
+                            Self::split_packets(transmits),
                         );
                     }
                     Ok((Some(udp_addr), None)) => {
--- a/iroh-net/src/magicsock.rs
+++ b/iroh-net/src/magicsock.rs
@@ -3358,4 +3358,55 @@ pub(crate) mod tests {
 
         Ok(())
     }
+
+    #[test]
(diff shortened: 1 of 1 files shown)


rklaehn:
@Arqu once you have a way to test things, can you check if this solves the issue?

iroh-net/src/magicsock.rs
```diff
@@ -1176,6 +1176,27 @@ impl Actor {
...
+    /// multiple packets according to that segment size. If it does not have a
+    /// segment size, the contents will be sent as a single packet.
+    fn split_packets(transmits: Vec<quinn_udp::Transmit>) -> Vec<Bytes> {
+        let mut res = Vec::with_capacity(transmits.len());
+        for transmit in transmits {
+            if let Some(segment_size) = transmit.segment_size {
+                let len = transmit.contents.len();
+                for min in (0..len).step_by(segment_size) {
+                    let max = (min + segment_size).min(len);
+                    res.push(transmit.contents.slice(min..max));
```
flub, line 1191: isn't this a lot easier using `transmit.contents.chunks(segment_size)`?

iroh-net/src/magicsock.rs
```diff
@@ -1176,6 +1176,27 @@ impl Actor {
...
+                }
+            } else {
+                res.push(transmit.contents);
+            }
+        }
+        res
+    }
```
flub, line 1198: this whole method is perfect for a few unittests...

flub (commented):
I think generally this approach is probably the right call for fixing this.  we need to do some GSO-like thing properly on the derper channel anyway later.

iroh-net/src/magicsock.rs
rklaehn: It might be that I just reimplemented that one...

rklaehn: it's not quite as straightforward if you want to use the Bytes in a 0 copy way, but I pushed an update.

Arqu:
> @Arqu once you have a way to test things, can you check if this solves the issue?

iroh-net/src/magicsock.rs
```diff
@@ -1176,6 +1176,26 @@ impl Actor {
...
+    /// For each transmit, if it has a segment size, it will be split into
+    /// multiple packets according to that segment size. If it does not have a
+    /// segment size, the contents will be sent as a single packet.
+    fn split_packets(transmits: Vec<quinn_udp::Transmit>) -> Vec<Bytes> {
+        let mut res = Vec::with_capacity(transmits.len());
+        for transmit in transmits {
+            let contents = transmit.contents;
+            if let Some(segment_size) = transmit.segment_size {
+                for chunk in contents.chunks(segment_size) {
+                    res.push(contents.slice_ref(chunk));
```
flub, line 1190: this `slice_ref` is probably a bit too lose for what we want.

But maybe rather than figure out what we're doing wrong we should start with copying Quinn's own implementation of this: https://github.com/quinn-rs/quinn/blob/f07a40d7f1da99253408fb1ab3db91eef3fe07e6/quinn-proto/src/tests/util.rs#L516-L537

rklaehn: slice_ref just converts the `&[u8]` back to a `Bytes`. The question is if chunks does the right thing.

rklaehn: This seems to be the exact same thing, except done in a loop. It is basically taking from the start, at most segment_size, until there is nothing left.

let end = segment_size.min(transmit.contents.len());

flub: oh, `slice_ref` uses pointer offsets.  that's probably fine then

rklaehn: slice_ref is just some weird API to get around the fact that the APIs for slices is much richer than for Bytes, so you often have a &[u8] that you *know* is part of the original Bytes, and you don't want to use slice.to_vec().into() but do actual 0 copy.

rklaehn:
I did the following to validate this:

- log in to a linux box
- enforce derper use
image
- configure a local derper
- run a local derper
- try with split_packets being a noop - not splitting => transfer does not work
- try with split_packets as in this PR - splitting on segment_size => transfer does work

flub (approved):

iroh-net/src/magicsock.rs
```diff
@@ -1176,6 +1176,26 @@ impl Actor {
...
+                }
+            } else {
```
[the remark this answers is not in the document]
dignifiedquire: right, I forgot about that sadness

PR #1382, opened 2023-08-21
fix(iroh-net): fix bug when reading the frame header

rklaehn:
the leftover bytes that might be already in the buffer were not properly considered

## Description

Minimal fix for the frame reading bug I found

## Notes & open questions

Does it actually work?

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


rklaehn:
implemented in the tokio-codec refactoring

Issue #1381, opened 2023-08-21
Intermittent derper disconnects when transferring large amounts of data


rklaehn:
While investigating https://github.com/n0-computer/iroh/issues/1357 I did some manual derper only tests, both mac -> mac and linux -> linux.

When transferring a large amount of data, the derper drops connections to clients in regular intervals, leading to very slow transfers. Below is an example of a derper log.

It seems that the derper gets errors in the frame parsing code, causing it to drop the connection to the client. The connection will then be reestablished from the client side a few seconds later, and the process begins anew.

This could either be the client sending malformed packets and the derper correctly disconnecting them, or the derper being unable to parse some valid packets. In any case it slows down the transfer to a crawl.

```
2023-08-21T12:09:47.602154Z  WARN iroh_net::derp::client_conn: connection manager for PublicKey(874d9a50e9b3449bf06af6aceae81b558f8e6cd023bb419d743713e6d84b1814) 34: writer closed in error frame header size 3417188410 exceeds reader limit of 1048576
2023-08-21T12:09:47.602173Z  WARN iroh_net::derp::clients: Could not find client for PublicKey(e0cbae283a789a8b6f9bbe115954807d370ecde72eb8f0b47e3e879b0099df6d), dropping packet
2023-08-21T12:09:47.602178Z  WARN iroh_net::derp::clients: pruning connection PublicKey(874d9a50e9b3449bf06af6aceae81b558f8e6cd023bb419d743713e6d84b1814)
2023-08-21T12:09:47.633353Z  WARN iroh_net::derp::server: send packet: no way to reach client PublicKey(874d9a50e9b3449bf06af6aceae81b558f8e6cd023bb419d743713e6d84b1814), dropped packet
2023-08-21T12:09:47.633368Z  WARN iroh_net::derp::server: send packet: no way to reach client PublicKey(874d9a50e9b3449bf06af6aceae81b558f8e6cd023bb419d743713e6d84b1814), dropped packet
2023-08-21T12:09:47.633373Z  WARN iroh_net::derp::server: send packet: no way to reach client PublicKey(874d9a50e9b3449bf06af6aceae81b558f8e6cd023bb419d743713e6d84b1814), dropped packet
2023-08-21T12:09:47.633377Z  WARN iroh_net::derp::server: send packet: no way to reach client PublicKey(874d9a50e9b3449bf06af6aceae81b558f8e6cd023bb419d743713e6d84b1814), dropped packet
[...]

Issue #1247, opened 2023-07-17
Fix test_mesh_network


flub:
This is possibly masking a bug

https://github.com/n0-computer/iroh/blob/1dadcf5465fc4314edd860b026df9cc1fe9d5f67/iroh-net/src/hp/derp/http/mesh_clients.rs#L115

ramfox:
After finally forcing through a run that fails w/ logs. Here is what I've found.

The issue occurs sometimes when forwarding a packet from one node to another.

Sending a packet from Bob to Alice (Bob -> derp server -> derp server -> Alice)
1) `SendPacket` is sent from Bob to the derp server - 49 bytes
2) Bob's derp server receives the `SendPacket` with the correct number of bytes
3) Bob's derp server sees it has a packet forwarder for Alice, so it sends a `ForwardPacket` message to Alice's derp server
81 bytes
4) Alice's derp server receives an `Unknown` packet (read off the wire as the first u8 of the packet) and interprets the number of bytes at 16 (read as the next u32), reads only 16 bytes off the wire & it drops the message (since it is unknown)
5) Since there are more than 16 bytes coming in, Alice's derp server continues to read off the wire
6) this time it receives an `Unknown` packet with what it believes are 823853558, which is over the number of bytes allowed in a packet. It drops the connection to the other derp server.
7) We timeout

We are also getting this panic, that I don't fully understand:
```
thread 'derp::http::mesh_clients::tests::test_mesh_network' panicked at 'tried to drop a ref to Id(322007373356990473), but no such span exists!', /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tracing-subscriber-0.3.17/src/registry/sharded.rs:350:21
```

I'm unclear about exactly which write or read is causing the issue! Looking into it further.

dignifiedquire:
> ```
> thread 'derp::http::mesh_clients::tests::test_mesh_network' panicked at 'tried to drop a ref to Id(322007373356990473), but no such span exists!', /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tracing-subscriber-0.3.17/src/registry/sharded.rs:350:21
> ```

this is related to debug spans, cc @flub

PR #1386, opened 2023-08-21
refactor(iroh-net): use tokio-util::codec for derp protocol implementation

### LINKED ISSUE #1384: Less error prone way to read frames
Currently you got a reader and a buffer that **needs** to stay together wherever you read stuff. If you read only from the reader, it works most of the time in small tests, but in the real world you will get a bug whenever there are some bytes in the buffer.

This style is very error prone. Here is a proposed solution:

we have a DerpFrameReader which takes a reader and a buffer and implements Stream<io::Result<(FrameType, Bytes)>> or something. Internally it does all the usual buffer stuff, but then it splits off the bytes.

If the returned Bytes go out of scope before the next call to `.next()`, this will be just as zero copy as the (&mut reader, &mut buffer) version, but much less error prone, because you get an immutable Bytes where you don't have to remember to drag it along, clear it etc.

dignifiedquire:
We have been having issues multiple times with the buffer management, this cleans up things and should make them more reliable. The core change is to use `tokio_util::codec` to implement the derp protocol. Additionally there are a couple smaller fixes in the derper code and tests, for issues that where discovered while doing this.

Fixes #1384, #1381, #1247


iroh-net/src/derp/codec.rs
```diff
@@ -0,0 +1,276 @@
...
+pub(crate) enum WriteFrame<'a> {
+    ServerKey {
+        key: PublicKey,
+    },
+    ClientInfo {
+        client_public_key: PublicKey,
+        encrypted_message: Vec<u8>,
+    },
+    ServerInfo {
+        encrypted_message: Vec<u8>,
```
rklaehn, line 29: &'a [u8]

rklaehn (commented):
*very* nice improvement

iroh-net/src/derp/codec.rs
```diff
@@ -0,0 +1,276 @@
...
+pub(crate) struct DerpCodec;
+
+#[derive(Debug)]
+pub(crate) struct Frame {
+    pub(crate) typ: FrameType,
+    pub(crate) content: Bytes,
+}
+
+#[derive(Debug)]
+pub(crate) enum WriteFrame<'a> {
```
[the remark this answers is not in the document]
dignifiedquire: yeah, planning to do sth like that, and to normalize the slice types as well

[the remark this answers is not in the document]
rklaehn: I tried it out in https://github.com/n0-computer/iroh/pull/1387 . Seems to work just fine. You just have to change a lot of code if you go this way because currently the parsing is all over the place.

iroh-net/src/derp/codec.rs
```diff
@@ -0,0 +1,276 @@
...
+    ServerInfo {
+        encrypted_message: Vec<u8>,
+    },
+    SendPacket {
+        dst_key: PublicKey,
+        packet: &'a [u8],
+    },
+    RecvPacket {
+        src_key: PublicKey,
+        content: Bytes,
```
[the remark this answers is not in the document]
dignifiedquire: unified to `Bytes`

iroh-net/src/derp/codec.rs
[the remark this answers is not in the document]
dignifiedquire: > You just have to change a lot of code if you go this way because currently the parsing is all over the place.

That's fine, will make it much less error prone and easier to test (and change if needed in the future)

iroh-net/src/derp.rs
```diff
@@ -184,180 +186,116 @@ impl From<FrameType> for u8 {
...
-    let mut buf = BytesMut::new();
     // the client is untrusted at this point, limit the input size even smaller than our usual
     // maximum frame size, and give a timeout
-    let (frame_type, _) = read_frame_timeout(
-        &mut reader,
-        256 * 1024,
-        &mut buf,
-        Some(Duration::from_secs(10)),
+
+    // TODO: variable recv size: 256 * 1024
```
rklaehn, line 238: I don't think there is a way to do this with the tokio_codec framework...

dignifiedquire: there is, we need to set this configurable on the derpcodec struct, and then set it and set it back

rklaehn:
This is not just a refactor but also fixes a bug with frame reading. Without this the derper is basically unusable as a relay.

This PR is just the fix without the refactor. It also fails the tests. Maybe that is a hint regarding the cause: https://github.com/n0-computer/iroh/actions/runs/5935069054?pr=1382

iroh-net/src/key.rs
```diff
@@ -17,14 +18,20 @@ pub use self::encryption::SharedSecret;
...
-#[derive(Clone, Copy, Hash, PartialEq, Eq)]
+#[derive(Clone, Copy, PartialEq, Eq)]
 pub struct PublicKey {
     public: VerifyingKey,
     /// Cached version of `crypto_box::PublicKey` matching `public`.
     /// Stored as raw array, as `crypto_box::PublicKey` is not `Copy`.
     public_crypto_box: [u8; 32],
 }
 
+impl Hash for PublicKey {
```
ramfox, line 29: now that we determined this wasn't the issue, should we go back to using the derive `Hash`?

dignifiedquire: no, this impl is still better, as it improves perf, and hashes the important piece

ramfox (approved):
Looks good! Huge improvement to the DERP code!

PR #1387, opened 2023-08-22
experiment: maybe we can use the big frame enum for both reading and writing...

rklaehn:


rklaehn:
We are doing this on the main [fix-derp-tokio-codec](https://github.com/n0-computer/iroh/tree/fix-derp-tokio-codec) branch

PR #1391, opened 2023-08-22
WIP experiment to track down the bug in fix-derp-tokio-codec

rklaehn:
Copy all the enum stuff, but don't do the codec stuff yet.


rklaehn:
The experiment was successful. The bug was bc the trick to push some extra bytes to the encoder by adding them to read_buffer_mut does not work.

PR #1397, opened 2023-08-24
test(iroh-net) add simple roundtrip test for frame encoding

rklaehn:
## Description

Add a simple roundtrip test for the frame encoding

## Notes & open questions

This adds just a happy case test. I was going to also add failure tests, but I noticed that I don't really understand the protocol well enough. E.g. which variable size content must be below MAX_PACKET_SIZE.

E.g. for the message types that contain an encrypted message: is there some size limit on the encrypted_message, and what exactly is it?

For RecvPacket, what is the size limit on the content? Is it MAX_PACKET_SIZE or MAX_PACKET_SIZE - PUBLIC_KEY_LENGTH?

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh-net/src/derp/codec.rs
```diff
@@ -676,3 +618,142 @@ pub(super) async fn recv_frame<S: Stream<Item = anyhow::Result<Frame>> + Unpin>(
...
+mod tests {
+    use proptest::prelude::*;
+    use tokio_util::codec::{FramedRead, FramedWrite};
+
+    use crate::derp::codec::DerpCodec;
+
+    use super::*;
+
+    #[tokio::test]
+    async fn test_basic_read_write() -> anyhow::Result<()> {
```
rklaehn, line 632: these are just moved around

rklaehn:
It's green now, but I had multiple failures with the arm cross compile. Maybe running the proptests in the arm crosscompile is just too slow?

dignifiedquire:
> MAX_PACKET_SIZE

This is the maximum of a single packet being fowarded or sent

> MAX_FRAME_SIZE

This should be the maximum size of a full frame

rklaehn:
> > MAX_PACKET_SIZE
> 
> This is the maximum of a single packet being fowarded or sent
> 
> > MAX_FRAME_SIZE
> 
> This should be the maximum size of a full frame

So there should be a test in the top of `from_bytes` that makes sure that content <= MAX_FRAME_SIZE - 5, and in all cases where a packet is involved ( `SendPacket`, `RecvPacket` and `ForwardPacket` ) there should be a test that packet.len() < MAX_PACKET_SIZE?

dignifiedquire:
yes

rklaehn:
Damn, even when running only in release mode this test fails...

https://github.com/n0-computer/iroh/actions/runs/5962328310/job/16173318109?pr=1397

rklaehn:
I guess this is ready for review, given that we might change the entire protocol with https://github.com/n0-computer/iroh/pull/1400 , and if we do we still want such tests...

flub (approved):

iroh-net/src/derp/codec.rs
```diff
@@ -677,6 +682,7 @@ mod tests {
     }
 }
 
+/// these test are slow in debug mode, so only run them in release mode
 #[cfg(test)]
 #[cfg(not(debug_assertions))]
```
dignifiedquire, line 687: don't we compile with debug in release mode currently?

rklaehn: I think we compile with debug symbols, but it seems enabling debug symbols does not set debug_assertions to true...

In any case we should do the ignore thing instead in a subsequent PR.

rklaehn: ❯ cargo test --release test_basic_read_write

running 1 test
test derp::codec::tests::test_basic_read_write ... ok

Just to confirm: the test is actually being run when run with --release...

PR #1400, opened 2023-08-24
feat(iroh-net): use postcard encoding for the derp protocol

dignifiedquire:
We actually entirely avoid length prefixing of individual frames, we only need length prefixes on the individual frame pieces, which makes this quite elegant.

This is a breaking change to the DERP protocol.

### DIFF
--- a/iroh-net/Cargo.toml
+++ b/iroh-net/Cargo.toml
@@ -15,7 +15,7 @@ rust-version = "1.67"
 anyhow = { version = "1", features = ["backtrace"] }
 aead = { version = "0.5.2", features = ["bytes"] }
 backoff = "0.4.0"
-bytes = "1"
+bytes = { version = "1", features = ["serde"] }
 crypto_box = { version = "0.9.1", features = ["serde", "chacha20"] }
 curve25519-dalek = "4.0.0"
 default-net = "0.16.2"
--- a/iroh-net/src/derp/client.rs
+++ b/iroh-net/src/derp/client.rs
@@ -554,7 +554,7 @@ pub enum ReceivedMessage {
     },
 }
 
-pub(crate) async fn send_packet<S: Sink<Frame, Error = std::io::Error> + Unpin>(
+pub(crate) async fn send_packet<S: Sink<Frame, Error = anyhow::Error> + Unpin>(
     mut writer: S,
     rate_limiter: &Option<RateLimiter>,
     dst_key: PublicKey,
@@ -579,7 +579,7 @@ pub(crate) async fn send_packet<S: Sink<Frame, Error = std::io::Error> + Unpin>(
     Ok(())
 }
 
-pub(crate) async fn forward_packet<S: Sink<Frame, Error = std::io::Error> + Unpin>(
+pub(crate) async fn forward_packet<S: Sink<Frame, Error = anyhow::Error> + Unpin>(
     mut writer: S,
     src_key: PublicKey,
     dst_key: PublicKey,
--- a/iroh-net/src/derp/codec.rs
+++ b/iroh-net/src/derp/codec.rs
@@ -1,9 +1,9 @@
 use std::time::Duration;
 
 use anyhow::{bail, ensure, Context};
-use bytes::{Buf, BufMut, Bytes, BytesMut};
-use ed25519_dalek::PUBLIC_KEY_LENGTH;
+use bytes::{Buf, Bytes, BytesMut};
 use futures::{Sink, SinkExt, Stream, StreamExt};
+use serde::{Deserialize, Serialize};
 use tokio_util::codec::{Decoder, Encoder};
 
 use super::types::ClientInfo;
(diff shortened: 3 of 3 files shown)


rklaehn:
OK, it makes the code simpler. But it seems that the checks for `MAX_PACKET_SIZE` are gone. Were they not needed, or is there a way to get them back?

Also, verbose as the code was before, it was immediately obvious what the wire format was. Now that it not the case anymore. E.g. I don't know what the value of the enum type discriminator is and where and how it writes length prefixes.

dignifiedquire:
> I think to do this we would need a test for each packet type that ensures that the wire format remains stable in the presence of small changes. E.g. just changing the order of the enum would probably change the type discriminator byte on the wire, but obviously mere roundtrip tests would not catch this...

I think we should have these tests for all our protocols anyway, because even without this it is easy to accidentally change the wire format.

dignifiedquire:
> it was immediately obvious what the wire format was. Now that it not the case anymore. E.g. I don't know what the value of the enum type discriminator is and where and how it writes length prefixes.

True, though I think this is something that some documentation can solve, and combined with the knowledge how `postcard`s data model works, it should become quite clear.

dignifiedquire:
> But it seems that the checks for `MAX_PACKET_SIZE` are gone. Were they not needed, or is there a way to get them back?

Hmm, they are gone inside this code path, they really should be part of constructing the `enum` variant, but unfortunately there is no easy way to do that.

rklaehn:
> > But it seems that the checks for `MAX_PACKET_SIZE` are gone. Were they not needed, or is there a way to get them back?
> 
> Hmm, they are gone inside this code path, they really should be part of constructing the `enum` variant, but unfortunately there is no easy way to do that.

You can make a newtype for Bytes that only allows anything up to a certain size, and does all the usual conversions so that it behaves mostly like a `Bytes`. But - there goes your simplification.

Maybe you can also use something like serde::serialize_with and serde::deserialize_with, but that won't prevent you from creating instances in memory that are too big - you will only know once you attempt to deserialize...

rklaehn:
So where are the length prefixes now?

Also, I noticed that the `MAGIC` is gone. Wasn't there some weird reason that the MAGIC is actually needed on the wire to distinguish such packets from other things?

dignifiedquire:
> So where are the length prefixes now?

inside the variable length sized items (`Bytes` and `String`)

rklaehn:
> > So where are the length prefixes now?
> 
> inside the variable length sized items (`Bytes` and `String`)

That means that you have to know the details of each packet type to be able to skip through the stream, because you need to know where the varint is based on the packet type.

In the protocol before you only needed to know the 5 byte header to skip over the stream, and were able to just skip/ignore message types you don't know.

I think if you want to e.g. do things like write a wireshark filter the old protocol style is more easy to deal with...

@Arqu WDYT?

ramfox:
There are 2 different `MAGIC` consts, `derp::MAGIC` & `disco::MAGIC`. The disco one we need helps us distinguish between disco messages and normal messages.

The derp one is only used in `ServerKey`, no where else.

rklaehn:
> There are 2 different `MAGIC` consts, `derp::MAGIC` & `disco::MAGIC`. The disco one we need helps us distinguish between disco messages and normal messages.
> 
> The derp one is only used in `ServerKey`, nowhere else.

I guess the one that I found weird that is needed after all is the derp::MAGIC one then.

Frando:
> > > So where are the length prefixes now?
> > 
> > 
> > inside the variable length sized items (`Bytes` and `String`)
> 
> That means that you have to know the details of each packet type to be able to skip through the stream, because you need to know where the varint is based on the packet type.
> 
> In the protocol before you only needed to know the 5 byte header to skip over the stream, and were able to just skip/ignore message types you don't know.
> 
> I think if you want to e.g. do things like write a wireshark filter the old protocol style is more easy to deal with...
> 
> @Arqu WDYT?

I agree to this. A 4-byte `u32` length prefix really doesn't hurt I'd say, and it makes the protocol much more approachable when working with other tools (e.g. wireshark). So I'd vote to put back the length prefix.
Could also improve performance, because with a length prefix you'd only try decoding with `postcard` once you know that you have a full message ready. With the current PR, you'd always try and fail if not enough data is there, which potentially means more unneeded parsing attempts.

Apart from that, I think this PR is very nice. The good thing about `postcard` is that it's such a straightforward and simple format, that with the right docs (potentially autogenerated even?) it is really as nice and simple as a manual protocol codec would be. E.g. if the message itself is a big enum, then the first 4 bytes of the frame are the enum discriminant as `u32be` - which is exactly what you'd do in manual impl as frame type.

ramfox:
> I guess the one that I found weird that is needed after all is the derp::MAGIC one then.
Agreed, it doesn't seem necessary.

rklaehn:
So IMHO the current protocol is just fine, and we should not change it just to save a few LOC. I like postcard, but I don't see the big benefit in this case.

Add some byte level tests for the current protocol, and we're done.

dignifiedquire:
closing for now

PR #1401, opened 2023-08-24
refactor(iroh): simplify and cleanup the sync algorithm

dignifiedquire:
switches to also use tokio::codec under the hood now.

Also introduces a max message length of 1GiB for now.


Frando:
LGTM! If we decide for having  a length prefix (see discussion in #1400 ) then let's add this, and then merge into the main sync pr from my side.

Issue #1316, opened 2023-07-28
Use copy on write file systems when importing and exporting files into the iroh repo


rklaehn:
This might be the best of both worlds, as cheap as a move, but safer because it is copy on write.

Here is a crate that claims to support this: https://docs.rs/reflink/0.1.3/reflink/fn.reflink_or_copy.html

rklaehn:
Implemented in https://github.com/n0-computer/iroh/pull/1581

PR #1405, opened 2023-08-25
feat(iroh): use reflink to avoid copying files into/out of the flat store

rklaehn:
## Description

By using reflink, we can avoid having to copy files into/out of the flat store. For file systems that support COW, this should be basically instantaneous.

The reflink crate provides a function that automatically falls back to copying.

## Notes & open questions

Does this cause trouble on any of the more obscure platforms, or is the fallback safe?

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh/src/baomap/flat.rs
+++ b/iroh/src/baomap/flat.rs
@@ -749,7 +749,17 @@ impl Store {
                     .join(format!("{}.temp", hex::encode(uuid)));
                 // copy the data, since it is not stable
                 progress.try_send(ImportProgress::CopyProgress { id, offset: 0 })?;
-                let size = std::fs::copy(&path, &temp_data_path)?;
+                let size = if let Some(size) = reflink::reflink_or_copy(&path, &temp_data_path)? {
+                    tracing::trace!("copied {} to {}", path.display(), temp_data_path.display());
+                    size
+                } else {
+                    tracing::trace!(
+                        "reflinked {} to {}",
+                        path.display(),
+                        temp_data_path.display()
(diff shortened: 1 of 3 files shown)


iroh/src/baomap/flat.rs
```diff
@@ -865,7 +875,13 @@ impl Store {
             tracing::info!("copying {} to {}", source.display(), target.display());
             progress(0)?;
             // todo: progress
-            std::fs::copy(&source, &target)?;
+            let size = if let Some(size) = reflink::reflink_or_copy(&source, &target)? {
+                tracing::trace!("copied {} to {}", source.display(), target.display());
+                size
+            } else {
+                tracing::trace!("reflinked {} to {}", source.display(), target.display());
+                std::fs::metadata(&target)?.len()
```
rklaehn, line 883: we do this just to get the size to report progress. Maybe not do it?

rklaehn:
Implements https://github.com/n0-computer/iroh/issues/1316

rklaehn:
sigh... seems like none of the crates work well for some targets...

iroh/src/baomap/flat.rs
dignifiedquire: It's not particularly expensive, and progress bars are a great thing. So anything that is really gained from not doing it?

rklaehn:
Asked the crate author of the forked crate in https://github.com/cargo-bins/reflink-copy/issues/16

dignifiedquire:
@rklaehn can we close this for now?

PR #1407, opened 2023-08-25
Wire test util

rklaehn:
## Description

Add test utils that make it more pleasant to write wire format tests, and write a few wire format tests.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh-test/Cargo.toml
```diff
@@ -13,6 +13,7 @@ publish = true
 rust-version = "1.67"
 
 [dependencies]
-tokio = { version = "1", features = ["rt"] }
+anyhow = "1.0.75"
```
flub: Can we leave this "1" like we do in most cargo.toml files?

PR #1408, opened 2023-08-25
feat(iroh-bytes): remove unneeded u64 length prefix

rklaehn:
## Description

we used to write the length prefix here. But this is not needed since we immediately close the stream after writing the request. And since this is a request/response protocol this is unlikely to change.

quinn::ReadStream::read_to_end offers a size limit where we can just use MAX_MESSAGE_SIZE to be on the safe side.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-bytes/src/get.rs
+++ b/iroh-bytes/src/get.rs
@@ -255,12 +255,6 @@ pub mod fsm {
                     return Err(ConnectedNextError::RequestTooBig);
                 }
 
-                // write u64 length prefix
-                writer
-                    .write_u64_le(request_bytes.len() as u64)
-                    .await
-                    .map_err(ConnectedNextError::from_io)?;
-
                 // write the request itself
                 writer
                     .write_all(&request_bytes)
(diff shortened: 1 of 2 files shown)


rklaehn:
Relevant discussion in discord: https://discord.com/channels/949724860232392765/950683937661935667/1144586154440663130

TLDR:

- if we ever wanted to extend this protocol to be not a `request -> response stream protocol` but a `(request, request update stream) -> response stream` protocol, the length prefix makes sense.
- due to some fundamental properties of the bao streaming format, I think it is very unlikely that we would want to extend this protocol to be a bidi protocol for some request variants

```
there are some cases where bidi streaming is inevitable. E.g. some update where progress flows back. But in this particular case I would agree with you. Just close the stream without an error and do a new request as soon as there has been a large change to data availability.
the bao protocol has the nice property that it is completely deterministic. for one request, there is only one valid sequence of bytes as a response. And if you don't have the data the only allowed thing is to cut short the response.
so no matter what your node state is, all valid responses are prefixes of the complete response.
once you got bidi, with msgs arriving after some delay, all that determinism goes out of the window.
```

flub (approved):

PR #1409, opened 2023-08-25
Iroh bytes wire tests

rklaehn:
## Description

A few more wire format tests for the various kinds of Requests of iroh-bytes.

## Notes & open questions

Do we want the length prefix for the hash?

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-bytes/src/protocol.rs
+++ b/iroh-bytes/src/protocol.rs
@@ -468,14 +468,14 @@ pub struct CustomGetRequest {
 /// A request
 #[derive(Deserialize, Serialize, Debug, PartialEq, Eq, Clone)]
 pub struct GetRequest {
+    /// Optional Request token
+    token: Option<RequestToken>,
     /// blake3 hash
     pub hash: Hash,
     /// The range of data to request
     ///
     /// The first element is the parent, all subsequent elements are children.
     pub ranges: RangeSpecSeq,
-    /// Optional Request token
-    token: Option<RequestToken>,
 }
 
 impl GetRequest {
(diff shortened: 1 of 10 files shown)


iroh-bytes/src/util.rs
```diff
@@ -252,4 +254,15 @@ mod tests {
         let encoded = hash.to_string();
         assert_eq!(encoded.parse::<Hash>().unwrap(), hash);
     }
+
+    #[test]
+    fn hash_wire_format() {
+        let hash = Hash::from([0xab; 32]);
+        let serialized = postcard::to_stdvec(&hash).unwrap();
+        let expected = parse_hexdump(r"
+            20 # length prefix
```
rklaehn, line 263: This was unexpected. I thought hashes would be serialized as 32 bytes without length prefix. Should we get rid of the length prefix?

iroh-bytes/src/protocol.rs
```diff
@@ -636,3 +636,82 @@ impl TryFrom<VarInt> for Closed {
...
+            ",
+            ),
+            (
+                Request::from(CustomGetRequest {
+                    token: Some(token),
+                    data: Bytes::from(&b"hello"[..]),
+                }),
+                r"
+                    01 # enum variant for CustomGetRequest
+                    01 # a token
```
rklaehn, line 703: It would be nice to put the token in front for all cases or at least be somewhat consistent.

also, we really need these tests. If you reorder the fields you obviously change the format, and we don't want to do that accidentally.

dignifiedquire (approved):

PR #1410, opened 2023-08-25
fix(iroh-bytes): Hash should be serialized as array not bytes

dignifiedquire:


iroh-bytes/Cargo.toml
```diff
@@ -43,6 +43,7 @@ tracing-futures = "0.2.5"
 http-body = "0.4.5"
 iroh-test = { path = "../iroh-test" }
 proptest = "1.0.0"
+serde_test = "1.0.176"
```
rklaehn, line 46: is it enough to use just "1"?

iroh-bytes/src/util.rs
```diff
@@ -156,7 +160,13 @@ impl Serialize for Hash {
     where
         S: Serializer,
     {
-        serializer.serialize_bytes(self.0.as_bytes())
+        // Fixed-length structures, including arrays, are supported in Serde as tuples
+        // See: https://serde.rs/impl-serialize.html#serializing-a-tuple
+        let mut s = serializer.serialize_tuple(32)?;
+        for item in self.0.as_bytes() {
```
rklaehn, line 166: this seems horribly inefficient. But I guess it is fine since in our bulk data path we don't use this.

rklaehn (approved):

iroh-bytes/Cargo.toml
dignifiedquire: this is how `cargo add` does it, which I don't think changing has a lot of benefit

iroh-bytes/src/util.rs
dignifiedquire: There is no other way, at least not in serde: 

https://docs.rs/serde/latest/src/serde/ser/impls.rs.html#163-168

PR #1413, opened 2023-08-26
feat(iroh-bytes): add initial query request

rklaehn:
## Description

This adds a new request type that allows the requester to get availability information about a blob or collection without having to download the data.

## Notes & open questions

Merge this before implementing so we keep the protocol stable?

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-bytes/src/protocol.rs
+++ b/iroh-bytes/src/protocol.rs
@@ -433,6 +433,8 @@ pub enum Request {
     Get(GetRequest),
     /// A get request that allows the receiver to create a collection
     CustomGet(CustomGetRequest),
+    /// A query request for a blob or collection
+    Query(QueryRequest),
 }
 
 impl Request {
@@ -441,6 +443,7 @@ impl Request {
         match self {
             Request::Get(get) => get.token(),
             Request::CustomGet(get) => get.token.as_ref(),
+            Request::Query(query) => query.token(),
         }
     }
 
@@ -449,6 +452,7 @@ impl Request {
         match &mut self {
             Request::Get(get) => get.token = value,
             Request::CustomGet(get) => get.token = value,
+            Request::Query(query) => query.token = value,
         }
         self
     }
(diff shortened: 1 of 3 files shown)


rklaehn:
First part of https://github.com/n0-computer/iroh/issues/1414

Frando:
Question: Do we want to be able to batch these? I can imagine that oftenly you'd want to query for a lot of hashes. With this architecture, you'd send a query for each hash on a separate stream, so likely sequentially. When sending a batch request, the provider can potentially batch resolve the availability in parallel? Not sure, but thought it's worth to bring up the question.

iroh-bytes/src/protocol.rs
```diff
@@ -517,6 +521,81 @@ impl GetRequest {
...
+///
+/// The answer to a query request is a set of chunks that are available, encoded
+/// as a [`RangeSpecSeq`].
+///
+/// The provider has to compute the availability information only for the requested
+/// ranges, not for the entire blob or collection. The response must contain
+/// availability information for the requested ranges, but may contain availability
+/// information for additional ranges if expedient. E.g. if the provider has
+/// a blob completely available, it may send `RangeSet2::all()` for that blob
+/// instead of limiting the response to the requested ranges.
```
flub, line 537: Do we think the size of a response would get too big if we always respond with all the availability for the entire blob/collection?  It would make the protocol a lot simpler.

iroh-bytes/src/protocol.rs
```diff
@@ -517,6 +521,81 @@ impl GetRequest {
...
+///
+/// E.g. availability might increase when the provider is itself downloading
+/// data from other providers, or decrease when the provider is deleting data.
+#[derive(Deserialize, Serialize, Debug, PartialEq, Eq, Clone)]
+pub struct QueryRequest {
+    /// blake3 hash
+    pub hash: Hash,
+    /// The range of data we request availability information for
+    ///
+    /// The first element is the parent, all subsequent elements are children.
```
flub, line 554: What if it's a single blob being requested? This description is a bit misleading?

iroh-bytes/src/protocol.rs
```diff
@@ -517,6 +521,81 @@ impl GetRequest {
...
+    /// Request a blob or collection with specified ranges
+    pub fn new(hash: Hash, ranges: RangeSpecSeq) -> Self {
+        Self {
+            hash,
+            ranges,
+            token: None,
+        }
+    }
+
+    /// Request a collection and all its children
```
flub, line 570: This can also be used on a single blob?

It's a bit confusing together with `.single()` below

iroh-bytes/src/protocol.rs
```diff
@@ -517,6 +521,81 @@ impl GetRequest {
...
+    pub fn all(hash: Hash) -> Self {
+        Self {
+            hash,
+            token: None,
+            ranges: RangeSpecSeq::all(),
+        }
+    }
+
+    /// Request just a single blob
```
flub, line 579: this means "the first blob"?

flub:
> Question: Do we want to be able to batch these? I can imagine that oftenly you'd want to query for a lot of hashes. With this architecture, you'd send a query for each hash on a separate stream, so likely sequentially. When sending a batch request, the provider can potentially batch resolve the availability in parallel? Not sure, but thought it's worth to bring up the question.

Batching on one stream vs creating many streams is something I often wonder and still haven't benchmarked.  In theory it's nicer to use many streams, the sender can create an entire bunch of requests in a tight loop and quinn should start packing these streams together into packets so you don't really suffer from the many streams.  That's the theory at least.  As I said I don't know how well it works in practice.

Also the quinn folks once mentioned they were wondering about implementing something like tcp_cork on a stream and/or connection to even further improve this: the ability to say to quinn "only start sending things once you have full packets or i uncork".  That would improve that kind of behaviour even more.

flub:
> Merge this before implementing so we keep the protocol stable?

Not sure I follow?  You want to get this into today's release?

Afaik this is a new request type so should be ok to add later is what we were saying last week i think?

iroh-bytes/src/protocol.rs
rklaehn: The issue is that depending on the impl, computing availability can be quite expensive. Either you go straight to the bytes, then computing availability is just hashing the file again. Or you cache it, then it might be incorrect.

That is the rationale for specifying a range of interest, in addition to the response size.

Assuming you got the data cached, reducing the response to the range that the requester has asked for is a single op, so I don't think it is overly complex. We just need to specify more precisely what the response should be - it should be the smallest RangeSpec that describes the availability in the intervals that were asked for.

iroh-bytes/src/protocol.rs
rklaehn: It's the same as with requests - if you only ask for the first element, it is a blob request.

rklaehn:
> > Merge this before implementing so we keep the protocol stable?
> 
> Not sure I follow? You want to get this into today's release?
> 
> Afaik this is a new request type so should be ok to add later is what we were saying last week i think?

I had just briefly considered having this request in there. But it's fine to do it later.

rklaehn:
> Question: Do we want to be able to batch these? I can imagine that oftenly you'd want to query for a lot of hashes. With this architecture, you'd send a query for each hash on a separate stream, so likely sequentially. When sending a batch request, the provider can potentially batch resolve the availability in parallel? Not sure, but thought it's worth to bring up the question.

For the entire protocol, I have assumed that batching is done by either creating a collection containing related items, or opening multiple quinn streams. The latter should be cheap for an existing connection.

Batching would be a straightforward extension of the protocol, reusing things like RangeSpecSeq etc., but it definitely requires more thought. So I would prefer if we try how far we can get without batching with cheap quinn substreams - just like in quic-rpc a RPC call creates it's own substream.

rklaehn:
> Batching on one stream vs creating many streams is something I often wonder and still haven't benchmarked. In theory it's nicer to use many streams, the sender can create an entire bunch of requests in a tight loop and quinn should start packing these streams together into packets so you don't really suffer from the many streams. That's the theory at least. As I said I don't know how well it works in practice.

Anecdotal evidence, but quic-rpc is using a stream per RPC call, and nevertheless you can get several 100000 RPC calls per second giving a fast connection.

If you look at RPC (1 stream per request) vs bidi (1 stream for all requests), RPC par isn't that much worse than bidi seq. This is what I base my gut feeling on that streams are cheap.

```
> cargo test --release quinn_channel_bench -- --nocapture
running 1 test
RPC seq 22_567 rps                                                              
RPC par 193_088 rps                                                             
bidi seq 252_977 rps                                                            
test quinn_channel_bench ... ok
```

iroh-bytes/src/protocol.rs
flub: sounds good

dignifiedquire:
what's the state of this?

PR #1418, opened 2023-08-28
refactor: Reduce amount of info logging

flub:
## Description

There is a bunch of netcheck stuff that happens periodically and does
not need info logging for this.

Similarly this reduces the log level of some very regular logging in
the derp client, magicsock and iroh.

Lastly log_concise_report is updated to not do so many allocations
while building the log message.

## Notes & open questions

This does not yet (knowlying) address the issue of the ios debug
console becoming unusable due to logging.

## Change checklist

- [x] Self-review.
- [x] ~~Tests if relevant.~~

### DIFF
--- a/iroh-sync/src/store/fs.rs
+++ b/iroh-sync/src/store/fs.rs
@@ -516,13 +516,14 @@ impl crate::ranger::Store<RecordIdentifier, SignedEntry> for StoreInstance {
             {
                 let mut record_table = write_tx.open_multimap_table(RECORDS_TABLE)?;
                 let key = (k.namespace_bytes(), k.author_bytes(), k.key());
-                let record = v.entry().record();
+                let record = v.entry();
+                let hash = record.content_hash();
                 let value = (
                     timestamp,
                     &v.signature().namespace_signature().to_bytes(),
                     &v.signature().author_signature().to_bytes(),
                     record.content_len(),
-                    record.content_hash().as_bytes(),
+                    hash.as_bytes(),
                 );
                 record_table.insert(key, value)?;
             }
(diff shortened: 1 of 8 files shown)


iroh-net/src/netcheck/reportgen.rs
```diff
@@ -336,7 +334,7 @@ impl Actor {
     }
 
     fn handle_probe_report(&mut self, probe_report: ProbeReport) {
-        info!("finished probe: {:?}", probe_report);
+        debug!("finished probe: {:?}", probe_report);
```
b5, line 337: I see this one a lot, agree it should be debug level

iroh-net/src/magicsock.rs
```diff
@@ -2202,13 +2202,6 @@ impl Actor {
         // remove moribund nodes in the next step below.
         for n in &self.net_map.as_ref().unwrap().peers {
             if self.peer_map.endpoint_for_node_key(&n.key).is_none() {
-                // info!(
-                //     "inserting peer's endpoint {:?} - {:?} {:#?} {:#?}",
-                //     self.inner.public_key, // our own node's public key
-                //     n.key.clone(),         // public key of node endpoint being created
-                //     self.peer_map,         // map of all endpoints we know about
-                //     n,                     // node being added
-                // );
```
b5, line 2211: why is this commented out? maybe just delete outright? We have git for remembering formatting

b5 (approved):
Agree I don't think this is going to fix the issue I was seeing, but it might, and it's a welcome set of changes.

I have one nit about commented out code that I think should be deleted, beyond that, LGTM

iroh-net/src/magicsock.rs
flub: yeah, i wondered the same hence deleting it here.  i think it was left there by accident before (by me)

PR #1419, opened 2023-08-28
fix(iroh): Try to fix flaky test_token_passthrough test

flub:
## Description

This seems to fail often on CI on the emulated platforms when the
machines are busy.  And the failures we see are timeouts.  So very
likely increasing this timeout should help.

Does a few smaller fixes:

- Only bind on localhost instead of exposing us to the world.  Tests
  should not expose themselves to the world.

- Make sure that the receiver dropping is caught in a panic.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/tests/provide.rs
+++ b/iroh/tests/provide.rs
@@ -877,7 +877,7 @@ async fn test_token_passthrough() -> Result<()> {
     let rt = test_runtime();
     let expected = b"hello".to_vec();
     let (db, hash) = create_test_db([("test", expected.clone())]);
-    let addr = "0.0.0.0:0".parse().unwrap();
+    let addr = SocketAddr::from((Ipv4Addr::LOCALHOST, 0));
     let node = test_node(db, addr)
         .custom_auth_handler(Arc::new(CustomAuthHandler))
         .runtime(&rt)
@@ -894,7 +894,7 @@ async fn test_token_passthrough() -> Result<()> {
                     if let iroh_bytes::provider::Event::GetRequestReceived { token: tok, .. } =
                         bp_msg
                     {
-                        events_sender.send(tok).ok();
+                        events_sender.send(tok).expect("receiver dropped");
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


iroh/tests/provide.rs
```diff
@@ -877,7 +877,7 @@ async fn test_token_passthrough() -> Result<()> {
     let rt = test_runtime();
     let expected = b"hello".to_vec();
     let (db, hash) = create_test_db([("test", expected.clone())]);
-    let addr = "0.0.0.0:0".parse().unwrap();
+    let addr = SocketAddr::from((Ipv4Addr::LOCALHOST, 0));
```
rklaehn, line 880: this is just drive by boyscouting, right?

iroh/tests/provide.rs
```diff
@@ -894,7 +894,7 @@ async fn test_token_passthrough() -> Result<()> {
                     if let iroh_bytes::provider::Event::GetRequestReceived { token: tok, .. } =
                         bp_msg
                     {
-                        events_sender.send(tok).ok();
+                        events_sender.send(tok).expect("receiver dropped");
```
rklaehn, line 897: Why does this fix the test?

rklaehn (approved):

iroh/tests/provide.rs
flub: not sure what you mean, but this is not crucial to the fix as the description says.  it is an orthogonal fix because I think no test should ever open public ports

iroh/tests/provide.rs
flub: it doesn't.  as per description it ensures we get a panic if this ever is an issue in one of the flakes.  this is more belt-and-braces as i expect the timeout to work, but with flaky tests you can't be explicit enough.

iroh/tests/provide.rs
rklaehn: But this is no change, no? It's just a nicer way to write the same thing without parse and unwrap.

Issue #1334, opened 2023-08-07
Downloader for iroh-bytes


Frando:
The [iroh-sync PR](https://github.com/n0-computer/iroh/pull/1216) adds a [downloader](https://github.com/n0-computer/iroh/blob/sync-gossip-bytes/iroh/src/download.rs). In its current state it is unfinished, and cannot do retries and doesn't have a proper queue.

Features we need from the downloader from a high level iroh sync point of view:

* Enqueue downloads as `Hash`, `PeerId` pairs
* Get notified when a download is finished (through a future)

This high level API exists already in the PR, with public methods `push(hash, peer_id)` and `async finished(hash: Hash)`. The actual actor below misses many features though. We want:

* Retries with incremental backoff
* Limits on parallel downloads and connection attempts (e.g. max 5 concurrent dials)
* Being able to queue a hash with a list of peers in an order with different timeouts (e.g. try from peer 1 immediately, from peer 2 in 100ms, and from peer 3 in 1 second) - aborting/skipping of course if a download succeeded

--

copying some notes from the [now-merged downloader improvements PR](https://github.com/n0-computer/iroh/pull/1216): 

The `Downloader` currently acts immediately on new requests and doesn't do any retries. It also does not have a proper queue internally.

We want to be able to queue downloads with retries and different timeouts. My idea was to add a time-based queue, maybe using the `TimerMap` from `iroh_gossip::util`. Then I thought to change the API to expose `download(hashes: Vec<Hash>, peer: Peer, delay: Option<Instant>)`. The actor would maintain a timer queue of pending requests and their delay, and poll that for the next operation, check if the operation is still needed, and if so enqueue the future to actually do the download. The actor should also keep track of running downloads, and have a limit of how many peers are being dialed or connected to in parallel (configurable). Code-wise, I thought to add at least a `DownloadState` for each pending download or so, and then similar to the current code have indexes into those for `PeerId` and `Hash` (and `Instant` if using something like the `TimerMap`).

With this, we can then:
* On new entry message via gossip, enqueue download from 
  * peer we recv'd from with a very small delay
  * from our other current active gossip neighbors with a larger delay
  * from the author with a larger delay plus random offset (*note: for this we need to add the peer info (address&region), with a signature, to the gossip message in addition to the `SignedEntry`)
* Queue more retries if all fail up to some limit

[...]

PR #1420, opened 2023-08-28
feat(iroh): downloader

divagant-martian:
## Description

Adds the `Downloader` as specified in #1334 plus some backchannel convos
Features include:
- Support collections
- Add delays to downloads
- Add download retries with an incremental backoff 
- Keeping peers for a bit longer than necessary in hopes they will be useful again
- Having the concept of intents and deduplicating downloads efforts
- Cancelling download intents
- Limiting how many concurrent requests are done in total
- Limiting how many concurrent requests are done per peer
- Limiting the number of open connections in total
- Basic error management in the form of deciding whether a peer should be dropped, the request should be dropped, or if the request should be retried

### TODOs
- A remaining TODO in the code is whether something special should be done when dropping quic connections
- Should downloads have a timeout?
- ~<sup>I know I've said this a hundred times with a hundred different things but would love to test this as well under stress scenarios and a large number of peers. don't hate me</sup>~
In reality after abstracting away all the IO most scenarios can be simulated easily. What would remain for a _much_ later time when the  need and opportunity for real case testing scenario arises is to tune the concurrency parameters

#### Downloading Ranges
There was the requirement of downloading a Hash, a range of a Hash, a collection and (not mentioned but potentially implied) ranges of collections. There is no support for ranges right now because of the great duplication of the `get` code in order to take advantage of proper errors added in #1362. In principle, adding ranges should be really easy. This is because it's an extension of the `DownloadKind` and would simply need calculating the missing ranges not based on the difference between what we have and the whole range, but the given range. I would prefer to find a way to deduplicate the get code before doing this extension.
Also, as far as I can tell, there is no need for this yet.
#### Prioritizing candidates per role: `Provider` and `Candidate`
A nice extension, as discussed at some point, is to differentiate candidates we know have the data, from those that _might_ have the data. This has added benefit that when a peer is available to perform another download under the concurrency limits, a hash we know they have could be downloaded right away instead of waiting for the delay. At this point making this doesn't make sense because we will likely attempt a download before the peer has retrieved the data themselves. To implement this, we would need to add the notifications of fully downloaded hashes as available into gossip first.
#### Leveraging the info from gossip
When declaring that a hash `X` should be downloaded, it's also an option to query gossip for peers that are subscribed to the topic to which `X` belongs to use them as candidates. This could be done connecting the `ProviderMap` to `gossip`. For now I don't see the need to do this.

### Open questions about Future work
- In line with the described work from above, the registry only allows querying for peer candidates to a hash since that's as good as it gets in terms of what we know from a remote right now. It's not clear to me if we would want to change this to have better availability information with #1413 in progress.
- More future work: downloading a large data set/blob from multiple peers would most likely require us to do a three step process:
  1. understanding the data layout/size.
  2. splitting the download.
  3. actually performing the separate downloads.
  Generally curious how this will end. My question here is whether we should do this for every download, or just on data that we expect to be big. Is there any way to obtain such hint without relying on a query every single time?

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh/src/downloader.rs
```diff
@@ -0,0 +1,979 @@
...
+mod test;
+
+/// Delay added to a request when it's first received.
+const INITIAL_REQUEST_DELAY: std::time::Duration = std::time::Duration::from_millis(500);
+/// Number of retries initially assigned to a request.
+const INITIAL_RETRY_COUNT: u8 = 4;
+/// Duration for which we keep peers connected after they were last useful to us.
+const IDLE_PEER_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
+/// Capacity of the channel used to comunicate between the [`Downloader`] and the [`Service`].
+const SERVICE_CHANNEL_CAPACITY: usize = 120;
```
dignifiedquire, line 58: why not a power of two?

iroh/src/downloader.rs
```diff
@@ -0,0 +1,979 @@
...
+    pub max_open_connections: usize,
+}
+
+impl Default for ConcurrencyLimits {
+    fn default() -> Self {
+        // these numbers should be checked against a running node and might depend on platform
+        ConcurrencyLimits {
+            max_concurrent_requests: 50,
+            max_concurrent_requests_per_peer: 4,
+            max_open_connections: 100,
```
dignifiedquire, line 117: this seems quite high, we might want to default to sth lower here, especially on mobile

iroh/src/downloader.rs
```diff
@@ -0,0 +1,979 @@
...
+        }
+    }
+
+    /// Get the ranges this download is requesting.
+    // NOTE: necessary to extend downloads to support ranges of blobs ranges of collections.
+    #[allow(dead_code)]
+    fn ranges(&self) -> RangeSpecSeq {
+        match self {
+            DownloadKind::Blob { .. } => RangeSpecSeq::from_ranges([RangeSet2::all()]),
+            DownloadKind::Collection { .. } => RangeSpecSeq::all(),
```
dignifiedquire, line 168: what's the difference between these two `RangeSpecSeq`, the read is if they are the same for the uninitiated 😅

iroh/src/downloader.rs
```diff
@@ -0,0 +1,979 @@
...
+}
+
+// TODO(@divma): do we care about failure reason? do we care about success data reporting?
+type DownloadResult = Result<(), ()>;
+
+/// Handle to interact with a download request.
+#[derive(Debug)]
+pub struct DownloadHandle {
+    /// Id used to identify the request in the [`Downloader`].
+    id: u64,
```
dignifiedquire, line 180: should this use the `Id` type alias?

iroh/src/downloader.rs
divagant-martian: the first one means the range `all` just once (so a blob), whereas the second one means the `all` range in the whole span

iroh/src/downloader.rs
divagant-martian: what do you suggest? I have no idea what the overhead of new connections is in general to pick a reasonable number, maybe we can start with something like 30?

iroh/src/downloader.rs
divagant-martian: I think because my brain saw a 128 in the code and then moved it to a constant and 8 and look similar? jk aside I'll change it to 128 again :)

iroh/src/downloader.rs
divagant-martian: yep. Done in 12c63dd5

iroh/src/downloader.rs
divagant-martian: fixed in 12c63dd5

iroh/src/downloader.rs
```diff
@@ -0,0 +1,979 @@
...
+use futures::{future::LocalBoxFuture, stream::FuturesUnordered, FutureExt, StreamExt};
+use iroh_bytes::{
+    baomap::{range_collections::RangeSet2, Store},
+    collection::CollectionParser,
+    protocol::RangeSpecSeq,
+    Hash,
+};
+use iroh_net::{key::PublicKey, MagicEndpoint};
+use tokio::sync::{mpsc, oneshot};
+use tokio_util::{sync::CancellationToken, time::delay_queue};
```
Frando, line 45: cool, TIL about `tokio_util::time::delay_queue` :thumbsup:

Frando (commented):
This is a very nice PR! The docs make it quite straightforward to read and understand. Thanks for that!

I did not get through to the end, will continue tomorrow or later tonight. I did not see anything that would block a merge, apart from minor nits. Will try to read through the tests still.

I think we will want to move away from the default 500ms delay ASAP, this just doesn't feel good - but that can come after merge with #1470 I'd say.

iroh/src/downloader.rs
```diff
@@ -0,0 +1,979 @@
...
+                    info.next_peer = next_peer
+                }
+                // increasing the retries by one accounts for multiple intents for the same request in
+                // a conservative way
+                info.remaining_retries += 1;
+                trace!(?kind, ?info, "intent registered with scheduled request");
+            }
+            None => {
+                let intents = HashMap::from([(id, sender)]);
+                self.schedule_request(kind, INITIAL_RETRY_COUNT, next_peer, intents)
```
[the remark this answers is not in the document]
divagant-martian: yeah basically here we would check, is the best peer a connected provider? if so, start right away (provided there is capacity to do so), also, we would start s download right away when a download finishes and the peer is a provider for a required hash

iroh/src/downloader.rs
dignifiedquire: I see, thanks

iroh/src/downloader.rs
dignifiedquire: > I think because my brain saw a 128 in the code and then moved it to a constant and 8 and 0 look similar? jk aside I'll change it to 128 again :)

🤣 



> fixed in [12c63dd](https://github.com/n0-computer/iroh/commit/12c63dd5a6950a339e23c6a8eaabbc8c666953ed)

thanks, makes my brain happy

iroh/src/downloader.rs
```diff
@@ -0,0 +1,979 @@
...
+            remaining_retries,
+            cancellation,
+            peer,
+        };
+        let cancellation = info.cancellation.clone();
+        self.current_requests.insert(kind.clone(), info);
+
+        let get = self.getter.get(kind.clone(), conn);
+        let fut = async move {
+            // TODO(@divma): timeout?
```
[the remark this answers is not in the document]
divagant-martian: that would probably go into the huge error mapping in the `get.rs` file, along with (if necessary) some new type of download failure. In reality what we need to figure out is what kind of errors we want to identify and surface and which ones we group together (in the fashion of, for example, `FailureAction::RetryLater` where we mean .. don't suface this, just go do the thing again later)

TLDR, I would agree with it being follow-up

iroh/src/downloader.rs
```diff
@@ -0,0 +1,979 @@
...
+    fn schedule_request(
+        &mut self,
+        kind: DownloadKind,
+        remaining_retries: u8,
+        next_peer: Option<PublicKey>,
+        intents: HashMap<Id, oneshot::Sender<DownloadResult>>,
+    ) {
+        // this is simply INITIAL_REQUEST_DELAY * attempt_num where attempt_num (as an ordinal
+        // number) is maxed at INITIAL_RETRY_COUNT
+        let delay = INITIAL_REQUEST_DELAY
```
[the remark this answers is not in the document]
divagant-martian: also not sure, current maximum accumulated delay amounts to 10seconds if my math is right. This is accumulated delay, which does not take into account the time it takes to do a download. max total delay of 10seconds sounds ok to me in general

iroh/src/downloader.rs
```diff
@@ -0,0 +1,979 @@
...
+    fn ranges(&self) -> RangeSpecSeq {
+        match self {
+            DownloadKind::Blob { .. } => RangeSpecSeq::from_ranges([RangeSet2::all()]),
+            DownloadKind::Collection { .. } => RangeSpecSeq::all(),
+        }
+    }
+}
+
+// TODO(@divma): do we care about failure reason? do we care about success data reporting?
+type DownloadResult = Result<(), ()>;
```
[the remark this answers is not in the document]
divagant-martian: I would prefer us to have an explicit error type to model what the caller needs to be surfaced to decide what to do next. If the answer is "always return on error" (`?`ing on it), then I would go with the `anyhow` error instead

iroh/src/downloader.rs
```diff
@@ -0,0 +1,979 @@
...
+        msg_rx: mpsc::Receiver<Message>,
+    ) -> Self {
+        Service {
+            getter,
+            providers: ProviderMap::default(),
+            dialer,
+            concurrency_limits,
+            msg_rx,
+            peers: HashMap::default(),
+            goodbye_peer_queue: delay_queue::DelayQueue::default(),
```
dignifiedquire, line 436: love the name :)

iroh/src/downloader.rs
```diff
@@ -0,0 +1,979 @@
...
+    /// Main loop for the service.
+    async fn run(mut self) {
+        loop {
+            // check if we have capacity to dequeue another scheduled request
+            let at_capacity = self
+                .concurrency_limits
+                .at_requests_capacity(self.in_progress_downloads.len());
+
+            tokio::select! {
+                Some((peer, conn_result)) = self.dialer.next() => {
```
dignifiedquire, line 453: shouldn't this have a fallback for the `select!` statement, given that you use `Some()` matching on the branches?

iroh/src/downloader.rs
```diff
@@ -0,0 +1,979 @@
...
+                    let request_info = self.scheduled_requests.remove(&kind).expect("is registered");
+                    self.on_scheduled_request_ready(kind, request_info);
+                }
+                Some(expired) = self.goodbye_peer_queue.next() => {
+                    let peer = expired.into_inner();
+                    self.peers.remove(&peer);
+                    trace!(%peer, "tick: goodbye peer");
+                }
+            }
+            #[cfg(test)]
```
dignifiedquire, line 480: maybe `cfg(debug_assertions)`?

iroh/src/downloader.rs
```diff
@@ -0,0 +1,979 @@
...
+        let dialer = iroh_gossip::net::util::Dialer::new(endpoint);
+
+        let create_future = move || {
+            let concurrency_limits = ConcurrencyLimits::default();
+            let getter = get::IoGetter {
+                store,
+                collection_parser,
+            };
+
+            let service = Service::new(getter, dialer, concurrency_limits, msg_rx);
```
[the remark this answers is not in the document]
divagant-martian: I prefer `service` because it does not marry with the Actor model and it's clear enough on intent, but I'm fine with adjusting if the team agrees

iroh/src/downloader.rs
divagant-martian: sounds good, tho that would require me to mos the function somewhere else since it's currently under test

But yeah agree

iroh/src/downloader.rs
```diff
@@ -0,0 +1,979 @@
...
+
+        // update the active requests for this peer
+        let ActiveRequestInfo {
+            intents,
+            peer,
+            mut remaining_retries,
+            ..
+        } = info;
+
+        let peer_info = self.peers.get_mut(&peer).expect("peer is connected");
```
dignifiedquire, line 711: Is this guranteed to be true? couldn't we have disconnected in the meantime?

iroh/src/downloader.rs
```diff
@@ -0,0 +1,979 @@
...
+            Err(FailureAction::AbortRequest(reason)) => {
+                debug!(%peer, ?kind, %reason, "aborting request");
+                for sender in intents.into_values() {
+                    let _ = sender.send(Err(()));
+                }
+            }
+            Err(FailureAction::DropPeer(reason)) => {
+                debug!(%peer, ?kind, %reason, "peer will be dropped");
+                if let Some(_connection) = peer_info.conn.take() {
+                    // TODO(@divma): this will fail open streams, do we want this?
```
dignifiedquire, line 744: lets at least log this in a way that we can see if this is an issue

iroh/src/downloader.rs
```diff
@@ -0,0 +1,979 @@
...
+
+        if let ConnState::NotConnected = state {
+            if !self.at_connections_capacity() {
+                // peer is not connected, not dialing and concurrency limits allow another connection
+                debug!(%peer, "dialing peer");
+                self.dialer.queue_dial(*peer);
+                Some(*peer)
+            } else {
+                trace!(%peer, "required peer not dialed to maintain concurrency limits");
+                None
```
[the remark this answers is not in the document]
divagant-martian: if we land here we already did just that
If we are here it means this is the best peer, so any other peer is either not connected or does not have the capacity

iroh/src/downloader.rs
```diff
@@ -0,0 +1,979 @@
...
+                    Some(connection.clone())
+                } else {
+                    None
+                }
+            }
+            PeerState::Idle { drop_key } => {
+                // peer is no longer idle
+                self.goodbye_peer_queue.remove(drop_key);
+                info.state = PeerState::Busy {
+                    active_requests: NonZeroUsize::new(1).expect("clearly non zero"),
```
dignifiedquire, line 898: clearly :)

iroh/src/downloader.rs
divagant-martian: maybe the connection was lost, but the peer should only be removed from the mapping after being idle, so having active quests = 0. If we get here the peer has active requests > 0

I will check in any case

iroh/src/downloader.rs
divagant-martian: we log it in the previous line. Something else you think we should log or that should be added?

iroh/src/downloader.rs
dignifiedquire: Sorry, I meant in general for all the `Some(..)` matches in the `select!`, I usually have fallback clause to break out of the loop and log an error, if any of them actually return `None`, to make sure it shuts down if the streams are shut down

iroh/src/downloader.rs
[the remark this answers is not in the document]
dignifiedquire: ah I see, I missed that, 👍

iroh/src/downloader.rs
divagant-martian: (also to note, maybe useful for @b5) that we can set different defaults per platform

b5: let's start with 25 (the split of diva's 30 & dig's 20 😄, but also 5 docs worth of gossip connections)

iroh/src/downloader.rs
divagant-martian: ah I see, yes I'll add that, thanks

iroh/src/downloader/get.rs
```diff
@@ -0,0 +1,561 @@
...
+                FailureAction::DropPeer(e.into())
+            }
+            e @ quinn::ConnectionError::ApplicationClosed(_) => {
+                // > The peer closed the connection
+                // peer might be disconnecting or otherwise unavailable, drop it
+                FailureAction::DropPeer(e.into())
+            }
+            e @ quinn::ConnectionError::Reset => {
+                // > The peer is unable to continue processing this connection, usually due to having restarted
+                // TODO(@divma): peer is unavailable but might be available later, maybe retry?
```
dignifiedquire, line 110: I think retry makes sense here

iroh/src/downloader/get.rs
```diff
@@ -0,0 +1,561 @@
...
+                FailureAction::DropPeer(e.into())
+            }
+            e @ quinn::ConnectionError::TimedOut => {
+                // > Communication with the peer has lapsed for longer than the negotiated idle timeout
+                // TODO(@divma): my understanding is that quinn should be configured to ping often
+                // enough to prevent this. If the peer's connfiguration and our configuration allow
+                // this to happen maybe the peer is not really usable
+                FailureAction::DropPeer(e.into())
```
dignifiedquire, line 118: I think we should attempt to retry here, if we we are on a slow connection eg derping our way through, so I would want to retry at least once to see if this can't be made to work

iroh/src/downloader/get.rs
```diff
@@ -0,0 +1,561 @@
...
+            }
+        }
+    }
+}
+
+impl From<quinn::ReadError> for FailureAction {
+    fn from(value: quinn::ReadError) -> Self {
+        match value {
+            quinn::ReadError::Reset(_)
```
dignifiedquire, line 132: can't we retry here, for the same reasons as above?

iroh/src/downloader/get.rs
```diff
@@ -0,0 +1,561 @@
...
+            | quinn::ReadError::ConnectionLost(_)
```
dignifiedquire, line 133: couldn't we retry here? seems like this could be a temporary issue

iroh/src/downloader/get.rs
```diff
@@ -0,0 +1,561 @@
...
+            }
+        }
+    }
+}
+
+impl From<quinn::WriteError> for FailureAction {
+    fn from(value: quinn::WriteError) -> Self {
+        match value {
+            quinn::WriteError::Stopped(_)
+            | quinn::WriteError::ConnectionLost(_)
```
dignifiedquire, line 148: same as for read

iroh/src/downloader/get.rs
```diff
@@ -0,0 +1,561 @@
...
+impl From<std::io::Error> for FailureAction {
+    fn from(value: std::io::Error) -> Self {
+        // generally consider io errors recoverable
+        // we might want to revisit this at some point
+        FailureAction::RetryLater(value.into())
+    }
+}
+
+/// Get a blob or collection
+pub async fn get<D: Store, C: CollectionParser>(
```
dignifiedquire, line 248: @rklaehn would be good if you could review the below `get_*` functions, given you are the most familiar with fetching blob

iroh/src/downloader/test.rs
```diff
@@ -0,0 +1,200 @@
...
+            .into_iter()
+            .all(|r| r.is_ok()),
+        "all downloads should succeed"
+    );
+
+    // verify that the request was sent just once
+    getter.assert_history(&expected_history);
+}
+
+/// Test that when the downloader receives a flood of requests, with only one peer to handle them,
```
dignifiedquire, line 169: love adding doc comments to tests, never seen this, but great idea

iroh/src/downloader/test/invariants.rs
```diff
@@ -0,0 +1,83 @@
...
+    #[track_caller]
+    fn chech_concurrency_limits(&self) {
+        let ConcurrencyLimits {
+            max_concurrent_requests,
+            max_concurrent_requests_per_peer,
+            max_open_connections,
+        } = &self.concurrency_limits;
+
+        // check the total number of active requests to ensure it stays within the limit
+        assert!(self.in_progress_downloads.len() <= *max_concurrent_requests);
```
dignifiedquire, line 28: maybe add some string comments on the `assert!`s, otherwise sometimes pretty hard to parse the failures

dignifiedquire (approved):
Very nice work. Left all the useful comments I can give for this round 😅

iroh/src/downloader.rs
dignifiedquire: Maybe the connection id?

iroh/src/downloader.rs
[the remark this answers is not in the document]
divagant-martian: I gave it a bit more thought. A download can be retried many times, so it's hard to point to a single error. In that sense, the anyhow error works fine imo

iroh/src/downloader.rs
divagant-martian: I'm now more or less certain that this not necessary. 

Reading the docs of `tokio::select` when an enabled branch is ready, and the pattern doesn't match it's simply considered disabled and another branch is polled. Basically the only case in which we would land in a default branch is if all patters fail (which means, if all branches are disabled). Since there is a branch that does not do pattern matching (the one that polls the message channel), we should never reach that

lmk if this sounds convincing enough

Here are the docs I'm basing this, for completeness

> The select! macro accepts one or more branches with the following pattern:
> 
> `<pattern> = <async expression> (, if <precondition>)? => <handler>,`
>
> (...more docs here..)
> 
> The complete lifecycle of a select! expression is as follows:
>
>  1. Evaluate all provded `<precondition>` expressions. If the precondition returns false, disable the branch for the remainder of the current call to select!. Re-entering `select!` due to a loop clears the "disabled" state.
    2.  Aggregate the `<async expression>`s from each branch, including the disabled ones. If the branch is disabled, `<async expression>` is still evaluated, but the resulting future is not polled.
    3.  Concurrently await on the results for all remaining `<async expression>`s.
    4.  Once an `<async expression>` returns a value, attempt to apply the value to the provided `<pattern>`, if the pattern matches, evaluate `<handler>` and return. If the pattern does not match, disable the current branch and for the remainder of the current call to select!. Continue from step 3.
    5. If all branches are disabled, evaluate the else expression. If none is provided, panic.

iroh/src/downloader.rs
divagant-martian: done in 7e704d73

iroh/src/downloader.rs
divagant-martian: done in aba53f2e

iroh/src/downloader.rs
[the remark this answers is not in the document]
divagant-martian: done in aba53f2e

iroh/src/downloader.rs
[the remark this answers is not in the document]
divagant-martian: I think your comment is of high enough value to be included in the code, so I did just that in aba53f2e, removing the `TODO`

iroh/src/downloader.rs
```diff
@@ -0,0 +1,979 @@
...
+            intents,
+            remaining_retries,
+            delay_key,
+            next_peer,
+        };
+        debug!(?kind, ?info, "request scheduled");
+        self.scheduled_requests.insert(kind, info);
+    }
+
+    /// Gets the [`quinn::Connection`] for a peer if it's connected an has capacity for another
```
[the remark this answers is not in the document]
divagant-martian: done in aba53f2e

iroh/src/downloader.rs
divagant-martian: I changed the expect message to better reflect this logic in [aba53f2e](https://github.com/n0-computer/iroh/pull/1420/commits/aba53f2e60dd5d79f7588a9ea152237b847e0548#diff-d4ea9132b751f09c6923e77e3e0fa5a8dfaa3971da42cb12d99837f06f900505L712)

iroh/src/downloader/get.rs
divagant-martian: I just realized the inner error is already mapped so we should use that instead

iroh/src/downloader/test/invariants.rs
divagant-martian: done in https://github.com/n0-computer/iroh/commit/aba53f2e60dd5d79f7588a9ea152237b847e0548

iroh/src/downloader/get.rs
divagant-martian: done in 4a7e4b2e

iroh/src/downloader/get.rs
divagant-martian: done in 4a7e4b2e

iroh/src/downloader/get.rs
divagant-martian: done in 4a7e4b2e

iroh/src/downloader/get.rs
divagant-martian: done in 4a7e4b2e

iroh/src/downloader/get.rs
divagant-martian: done in 4a7e4b2e

Frando (approved):
Looks great, and I did not find anything that @dignifiedquire did not yet mention. 

I really like the test setup with `check_invariants` and the mocked impls.

Let's get this in :-)

PR #1430, opened 2023-08-30
feat(iroh-sync) include timestamps in the sync and remove history

dignifiedquire:
(PR text updated by @Frando)

## Description

This is needed, to ensure both sides will end up with the actual latest version. Also stops storing old versions, as they are not used anymore, and things get very complicated otherwise.

This leads to a couple of changes (all in this PR):

* The store always overwrites if an entry with identical (`namespace`, `author`, `key`) and later timestamp than current is inserted
* The `GetFilter` doesn't have `latest` option anymore and is now a enum
* RPC gets `get_one` to select an entry by author and key

## Notes & open questions

* This change should inform our docs and discussion about author sharing: Using the same author privatekey on different nodes now would mean that concurrent writes while offline would result in the older of the entries (by wall clock time) being deleted and lost without recovery after sync. This means, I think, that author keys should not be shared between devices, and profile identity has to be established across author keys on an upper layer or within the app.

* We don't support deletes in iroh bytes yet, but with this change we definitely have to. After an update (overwrite) the old blob currently is still persisted but lost in space without any reference to its hash.

TODOs:
* Review method naming on `store` trait. Maybe  ~~`get`~~ `get_range` and ~~`get_by_key_and_author`~~`get`?
* More tests

### DIFF
--- a/iroh-sync/src/store/fs.rs
+++ b/iroh-sync/src/store/fs.rs
@@ -5,7 +5,6 @@ use std::{cmp::Ordering, collections::HashMap, path::Path, sync::Arc};
 use anyhow::{bail, Result};
 use ouroboros::self_referencing;
 use parking_lot::RwLock;
-use rand_core::CryptoRngCore;
 use redb::{
     AccessGuard, Database, MultimapRange, MultimapTableDefinition, MultimapValue,
     ReadOnlyMultimapTable, ReadTransaction, ReadableMultimapTable, ReadableTable, StorageError,
@@ -180,10 +179,9 @@ impl super::Store for Store {
         Ok(Some(author))
     }
 
-    fn new_author<R: CryptoRngCore + ?Sized>(&self, rng: &mut R) -> Result<Author> {
-        let author = Author::new(rng);
-        self.insert_author(author.clone())?;
-        Ok(author)
+    fn import_author(&self, author: Author) -> Result<()> {
+        self.insert_author(author)?;
+        Ok(())
     }
 
     fn list_authors(&self) -> Result<Self::AuthorsIter<'_>> {
--- a/iroh-sync/src/store/fs.rs
+++ b/iroh-sync/src/store/fs.rs
@@ -2,13 +2,12 @@
 
 use std::{cmp::Ordering, collections::HashMap, path::Path, sync::Arc};
 
-use anyhow::{bail, Result};
+use anyhow::Result;
 use ouroboros::self_referencing;
 use parking_lot::RwLock;
 use redb::{
-    AccessGuard, Database, MultimapRange, MultimapTableDefinition, MultimapValue,
-    ReadOnlyMultimapTable, ReadTransaction, ReadableMultimapTable, ReadableTable, StorageError,
-    TableDefinition,
+    AccessGuard, Database, Range as TableRange, ReadOnlyTable, ReadTransaction, ReadableTable,
+    StorageError, TableDefinition,
 };
 
 use crate::{
@@ -20,7 +19,7 @@ use crate::{
     },
 };
 
-use self::ouroboros_impl_range_all_iterator::BorrowedMutFields;
+use self::ouroboros_impl_range_iterator::BorrowedMutFields;
 
 /// Manages the replicas and authors for an instance.
 #[derive(Debug, Clone)]
@@ -45,20 +44,19 @@ const NAMESPACES_TABLE: TableDefinition<&[u8; 32], &[u8; 32]> =
     TableDefinition::new("namespaces-1");
 
 // Records
-// Multimap
+// Table
 // Key: ([u8; 32], [u8; 32], Vec<u8>) # (NamespaceId, AuthorId, Key)
-// Values:
-//    (u64,  [u8; 32], [u8; 32], u64, [u8; 32])
+// Value:
+//    (u64, [u8; 32], [u8; 32], u64, [u8; 32])
 //  # (timestamp, signature_namespace, signature_author, len, hash)
 
 type RecordsId<'a> = (&'a [u8; 32], &'a [u8; 32], &'a [u8]);
 type RecordsValue<'a> = (u64, &'a [u8; 64], &'a [u8; 64], u64, &'a [u8; 32]);
-type RecordsRange<'a> = MultimapRange<'a, RecordsId<'static>, RecordsValue<'static>>;
-type RecordsTable<'a> = ReadOnlyMultimapTable<'a, RecordsId<'static>, RecordsValue<'static>>;
+type RecordsRange<'a> = TableRange<'a, RecordsId<'static>, RecordsValue<'static>>;
+type RecordsTable<'a> = ReadOnlyTable<'a, RecordsId<'static>, RecordsValue<'static>>;
 type DbResult<T> = Result<T, StorageError>;
 
-const RECORDS_TABLE: MultimapTableDefinition<RecordsId, RecordsValue> =
-    MultimapTableDefinition::new("records-1");
+const RECORDS_TABLE: TableDefinition<RecordsId, RecordsValue> = TableDefinition::new("records-1");
 
 impl Store {
     /// Create or open a store from a `path` to a database file.
@@ -70,7 +68,7 @@ impl Store {
         // Setup all tables
         let write_tx = db.begin_write()?;
         {
-            let _table = write_tx.open_multimap_table(RECORDS_TABLE)?;
+            let _table = write_tx.open_table(RECORDS_TABLE)?;
             let _table = write_tx.open_table(NAMESPACES_TABLE)?;
             let _table = write_tx.open_table(AUTHORS_TABLE)?;
         }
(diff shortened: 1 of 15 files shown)


iroh/src/commands/sync.rs
```diff
@@ -195,62 +186,44 @@ impl DocCommands {
...
-                };
-                if let Some(author) = author {
-                    filter = filter.with_author(author);
+                let filter = match (author, prefix) {
+                    (None, None) => GetFilter::All,
+                    (None, Some(prefix)) => GetFilter::Prefix(prefix.as_bytes().to_vec()),
+                    (Some(author), None) => GetFilter::Author(author),
+                    (Some(author), Some(prefix)) => {
+                        GetFilter::AuthorAndPrefix(author, prefix.as_bytes().to_vec())
+                    }
```
dignifiedquire, line 224: maybe this could be a method on `GetFilter` sth like `From<(Option<AuthorId>, &[u8])> for GetFilter`

iroh-sync/src/sync.rs
```diff
@@ -503,6 +534,7 @@ impl RecordIdentifier {
         out.extend_from_slice(self.namespace.as_bytes());
         out.extend_from_slice(self.author.as_bytes());
         out.extend_from_slice(&self.key);
+        out.extend(&self.timestamp.to_be_bytes())
```
dignifiedquire, line 537: should be extend_from_slice

iroh-sync/src/store/memory.rs
```diff
@@ -127,137 +106,98 @@ impl super::Store for Store {
...
+    type Item = Result<SignedEntry>;
     fn next(&mut self) -> Option<Self::Item> {
-        match &mut self.inner {
-            GetIterInner::All(iter) => iter.next().map(|x| x.map(|(_id, entry)| entry)),
-            GetIterInner::Latest(iter) => iter.next().map(|x| x.map(|(_id, entry)| entry)),
-            GetIterInner::Single(iter) => iter.next(),
-        }
+        self.0.next().map(|res| {
+            res.map(|(_id, entry)| entry)
+                .map_err(|_| unreachable!("never errors"))
```
dignifiedquire, line 123: you can just use `expect`

iroh-sync/src/store/memory.rs
```diff
@@ -509,43 +454,51 @@ impl crate::ranger::Store<RecordIdentifier, SignedEntry> for ReplicaStoreInstanc
     }
 
     fn put(&mut self, k: RecordIdentifier, v: SignedEntry) -> Result<(), Self::Error> {
-        // TODO: propagate error/not insertion?
+        assert_eq!(k.timestamp(), v.timestamp(), "inconsistent timestamp"); // TODO: error
+                                                                            // TODO: propagate error/not insertion?
         if v.verify().is_ok() {
-            let timestamp = v.entry().record().timestamp();
             // TODO: verify timestamp is "reasonable"
```
dignifiedquire, line 460: given the amount of reliance we have now on this, we should really do some sanity checks here

iroh-sync/src/store/fs.rs
```diff
@@ -273,75 +230,68 @@ impl super::Store for Store {
...
-        )
+        author: AuthorId,
+        prefix: impl AsRef<[u8]>,
+    ) -> Result<RangeIterator<'_>> {
+        let author = author.as_bytes();
+        let start = (namespace.as_bytes(), author, prefix.as_ref());
+        let mut end_prefix = prefix.as_ref().to_vec();
+        // move to next possible key after the prefix, which marks the end of the range
+        // TODO: is there a better way to define the range end?
+        for char in end_prefix.iter_mut().rev() {
```
dignifiedquire, line 271: this won't work for a key that is all 255

iroh-sync/src/store/fs.rs
```diff
@@ -600,18 +561,14 @@ impl Iterator for RangeLatestIterator<'_> {
...
                 };
                 if fields.filter.matches(&id) {
-                    let last = next.1.last();
-                    let value = match last? {
-                        Ok(value) => value,
-                        Err(err) => return Some(Err(err.into())),
-                    };
-                    let (timestamp, namespace_sig, author_sig, len, hash) = value.value();
-                    let record = Record::new(timestamp, len, hash.into());
+                    // TODO: remove timestamp
```
dignifiedquire, line 570: ?

iroh-sync/src/sync.rs
```diff
@@ -601,7 +630,7 @@ impl Record {
 
     /// Serialize this record into a mutable byte array.
     pub(crate) fn as_bytes(&self, out: &mut Vec<u8>) {
-        out.extend_from_slice(&self.timestamp.to_be_bytes());
+        // Note: the timestamp is not serialized as it is transmitted through the key
```
dignifiedquire, line 633: delete

iroh-sync/src/store/fs.rs
```diff
@@ -424,12 +375,12 @@ impl crate::ranger::Store<RecordIdentifier, SignedEntry> for StoreInstance {
     fn put(&mut self, k: RecordIdentifier, v: SignedEntry) -> Result<()> {
         // TODO: propagate error/not insertion?
         if v.verify().is_ok() {
-            let timestamp = v.entry().record().timestamp();
+            let timestamp = k.timestamp();
```
dignifiedquire, line 378: this is missing the timestamp check on insert, need to makes sure it is newer than the existing one

iroh-sync/src/store/fs.rs
Frando: Yup. To solve this, we need to add a seperator byte between author and key. Maybe also between namespace and author (or not if we're positive that `[255; 32]` never occurs in practice as èd25591` pubkey.

iroh-sync/src/store/fs.rs
Frando: Leftover comment

iroh-sync/src/store/fs.rs
Frando: Right. For an all-255 prefix, we should increment the last byte of the author then instead.

iroh/src/commands/sync.rs
Frando: Made a method instead, not sure if the `From` for a tuple is discoverable enough.

iroh-sync/src/store/fs.rs
Frando: Fixed and added a unit test

iroh-sync/src/store/memory.rs
Frando: Hm, what kinds of sanity checks are you thinking about? Timestamp not being too far in the future from the device's wall clock?

Frando:
Thanks for the review!

2 open questions still:

* What sanity checks do we want to do on the timestamp?
* Do you know if `[255u8; 32]` is a possible `ed25519` pubkey? If so, we'd have to change the end range calculations. If we know that this pubkey will never occur, we're fine with the current code I think.

Frando:
> Do you know if [255u8; 32] is a possible ed25519 pubkey? If so, we'd have to change the end range calculations. If we know that this pubkey will never occur, we're fine with the current code I think.

`[255u8; 32]` actually *is* a valid `ed25519_dalek::VerifyingKey`.

I changed the algorithm to use open-ended ranges if we cannot increment the author or namespace id bytes. Should be fine this way, I think.

iroh-sync/src/store/memory.rs
dignifiedquire: > Timestamp not being too far in the future from the device's wall clock?

yes: for reference: https://willowprotocol.org/more/timestamps-really

the suggestion is 10min into the future

rklaehn:
> > Do you know if [255u8; 32] is a possible ed25519 pubkey? If so, we'd have to change the end range calculations. If we know that this pubkey will never occur, we're fine with the current code I think.
> 
> `[255u8; 32]` actually _is_ a valid `ed25519_dalek::VerifyingKey`.
> 
> I changed the algorithm to use open-ended ranges if we cannot increment the author or namespace id bytes. Should be fine this way, I think.

If the storage engine supports open-ended ranges, why not always use them instead of trying to increment the bytes?

Frando:
> > > Do you know if [255u8; 32] is a possible ed25519 pubkey? If so, we'd have to change the end range calculations. If we know that this pubkey will never occur, we're fine with the current code I think.
> > 
> > 
> > `[255u8; 32]` actually _is_ a valid `ed25519_dalek::VerifyingKey`.
> > I changed the algorithm to use open-ended ranges if we cannot increment the author or namespace id bytes. Should be fine this way, I think.
> 
> If the storage engine supports open-ended ranges, why not always use them instead of trying to increment the bytes?

Our keys in redb are `namespace` (32bytes), `author` (32bytes), `key` (any number of bytes), concatenated together.
If we want to select by author and key prefix, the range goes from `namespace:author:prefix` to the same byte string incremented by one. Alternative would be to not do this but stop the iterator when leaving the prefix. 
Or am I missing something?

b5 (approved):
Proxy approval for dig

PR #1435, opened 2023-08-31
feat: get list of `ConnectionInfo`s or an individual node's `ConnectionInfo`

ramfox:
## Description

Display connection information for all nodes we have connected with or have supplied our node with addresses for.

Display connection information on a particular node, using its node id.

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh/src/commands.rs
+++ b/iroh/src/commands.rs
@@ -9,12 +9,11 @@ use iroh::dial::Ticket;
 use iroh::rpc_protocol::*;
 use iroh_bytes::{protocol::RequestToken, util::runtime, Hash};
 use iroh_net::key::{PublicKey, SecretKey};
-use iroh_net::magicsock::ConnectionType;
 
 use crate::config::{ConsoleEnv, NodeConfig};
 
 use self::provide::{ProvideOptions, ProviderRpcPort};
-use self::sync::{AuthorCommands, DocCommands};
+use self::sync::{fmt_short, AuthorCommands, DocCommands};
 
 const DEFAULT_RPC_PORT: u16 = 0x1337;
 const MAX_RPC_CONNECTIONS: u32 = 16;
@@ -323,11 +322,10 @@ impl NodeCommands {
         match self {
             Self::Connections => {
                 let mut connections = client.server_streaming(ConnectionsRequest).await?;
-                println!("Connections:");
+                println!("node id\t\tderp region\tconn type\tlatency");
                 while let Some(Ok(Ok(ConnectionsResponse { node_info }))) = connections.next().await
                 {
                     fmt_connection(node_info);
-                    println!();
                 }
             }
             Self::Shutdown { force } => {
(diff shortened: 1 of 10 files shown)


b5:
ZOMG I WANT THIS.

Howeve, given that you had to censor the screenshot, I'd say right off the bat we should go with listing by `NodeID`, and omitting IP addresses. wdyt?

ramfox:
> ZOMG I WANT THIS.
> 
> Howeve, given that you had to censor the screenshot, I'd say right off the bat we should go with listing by `NodeID`, and omitting IP addresses. wdyt?

Screen Shot 2023-08-30 at 22 04 06

What do you think is the "right" language when we don't have any latency info?

b5:
Can we use short NodeIDs, like we do for doc & author ids in the console? We could also do this in a follow-up.

> What do you think is the "right" language when we don't have any latency info?

`with a latency of Unknown`, using "unknown" is the right thing to do here imho

iroh/src/commands.rs
```diff
@@ -318,6 +321,15 @@ pub enum NodeCommands {
 impl NodeCommands {
     pub async fn run(self, client: RpcClient) -> Result<()> {
         match self {
+            Self::Connections => {
+                let mut connections = client.server_streaming(ConnectionsRequest).await?;
+                println!("Connections:");
+                while let Some(Ok(Ok(ConnectionsResponse { node_info }))) = connections.next().await
```
ramfox, line 327: @rklaehn can I get some RPC help?

I don't understand how I ended up with double `Result`s. Did I do something incorrectly, or is expected?

ramfox:
@b5 
Screen Shot 2023-08-30 at 22 44 39

```
Connections:
	hpf6qa4h…
	derp region: 1
	direct connection
	5175250ns latency

	nzx6mxi6…
	derp region: 1
	relay connection
	unknown latency

	xce2h4zq…
	derp region: 1
	relay connection
	unknown latency

	xuko33wo…
	derp region: 1
	direct connection
	3489375ns latency
```

If you write out one example of how you think it should look, I'll just do it, because it doesn't look right yet but i don't know how to adjust it.

ramfox:
```
Connections:
	xuko33wo…
	derp region: 1
	connection type: direct
	latency: 3489375ns 
```
?

b5:
zomg you fast. the problem is you're not doing one entry per line. aim for a table layout:

```
nodeId    derp region    conn type    latency
xuko33wo…  1             direct       56.5678ms
```

Huge win here is each row will be a constant-ish length except for the last value (latency), which makes the formatting much easier, just use `\t` characters between values.

latency should be displayed in `ms` not `ns`.

ramfox:
Screen Shot 2023-08-30 at 23 37 51

will look at other ways to display latency tomorrow!

iroh/src/commands.rs
Frando: This is expected behavior. The first `Result` is around a connection error while reading the response, and the second `Result` is the `RpcResult` that you are sending as `Item` in the stream.

Frando (commented):
Looks good to me!

Maybe let's add a `node connection NODE_ID` command as well that shows all details for a single node id, including IP addresses. This info is available at the node, so I think it makes sense to expose it in some way. Useful for debugging etc.

dignifiedquire:
https://github.com/manlge/human-time makes nicer timestamps

dignifiedquire:
https://github.com/Nukesor/comfy-table can render easy very nice tables :)

dignifiedquire:
> Can extend this functionality to the console.

lets add this to the repl for sure

dignifiedquire:
We should add this to the `client` api as well, this will be very useful from the API

ramfox:
### `iroh node connections`
CLI
Screen Shot 2023-08-31 at 23 15 21

Console
Screen Shot 2023-08-31 at 23 15 59

Adjusted the `node id` field to show the full ID, otherwise you wouldn't be able to fetch any individual node connection information with `iroh node connection NODE_ID`

### `iroh node connection NODE_ID`
CLI
Screen Shot 2023-08-31 at 23 15 31

Console
Screen Shot 2023-08-31 at 23 16 10

dignifiedquire (approved):

PR #1438, opened 2023-08-31
fix(iroh-net): dns fallback to default config

### LINKED ISSUE #1436: iroh-net: dns resolver issue: "No such file or directory" error on iOS, when running on real hardware
iroh works fine in the simulator with the latest stuff, but crashes on an actual iPhone for any `iroh-ffi` build that references a commit after 93147ac49a960ef0f4646c41a5c951faf5df2278 on `main`

The  error I see: 
```
thread 'main-runtime' panicked at 'unable to create DNS resolver: ResolveError { kind: Io(Os { code: 2, kind: NotFound, message: "No such file or directory" }) }', /Users/b5/.cargo/git/checkouts/iroh-6b78e5b19f41ff57/33cf40b/iroh-net/src/dns.rs:5:50
```

I've stepped through the commits, and have confirmed the issue is introduced in 93147ac49a960ef0f4646c41a5c951faf5df2278, the prior commit is fine.

dignifiedquire:
On at least iOS the resolve.conf is not available, so need to fallback to a default configuration in that case.

Based on the tested approach in DeltaChat

Fixes #1436

### DIFF
--- a/iroh-net/src/dns.rs
+++ b/iroh-net/src/dns.rs
@@ -1,9 +1,25 @@
+use anyhow::Result;
 use once_cell::sync::Lazy;
-use trust_dns_resolver::TokioAsyncResolver;
+use trust_dns_resolver::{config, AsyncResolver, TokioAsyncResolver};
 
-pub static DNS_RESOLVER: Lazy<TokioAsyncResolver> = Lazy::new(|| {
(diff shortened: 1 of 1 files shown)


flub (approved):

b5:
Have tested this on real hardware, can confirm this fixes!

PR #1439, opened 2023-08-31
feat(iroh-sync): validate entries in upper sync layer

Frando:
## Description

* Moves validation of entries from the store to `sync.rs`
* Skip validation for locally-inserted entries (not needed, signature was created by us)
* Reject entries that are more than 10 minutes in the future of our local device's wall clock time

## Notes & open questions

* The validation uses `SystemTime::now()`. Maybe we could make `Replica` generic over a `TimeSource`, with a default impl `SystemTimeSource` that uses `SystemTime`. This would allow tests to plug in a mock time source. But if its only for tests, it might not be worth it. I don't think there's many  realistic scenarios where you'd want to swap `SystemTime` for something else - apart from platform support, where `SystemTime` has a few gaps (e.g. wasm). So maybe let's keep it simple for now and add the generic once a need outside of tests arise.

* Do we want to make the "max 10 minutes in the future" const configurable?

## Change checklist

- [x ] Self-review.
- [x] Tests if relevant.


iroh-sync/src/ranger.rs
```diff
@@ -400,13 +400,17 @@ where
...
+    /// (which means the entry will be dropped and not stored).
     pub fn process_message<F>(
         &mut self,
         message: Message<K, V>,
-        cb: F,
+        validate_cb: F,
     ) -> Result<Option<Message<K, V>>, S::Error>
     where
-        F: Fn(K, V),
+        F: Fn(&S, K, V) -> bool,
```
dignifiedquire, line 413: Could we pass references for all here? That way we don't always have to clone them

iroh-sync/src/ranger.rs
```diff
@@ -979,9 +983,9 @@ mod tests {
             rounds += 1;
             alice_to_bob.push(msg.clone());
 
-            if let Some(msg) = bob.process_message(msg, |_, _| {}).unwrap() {
```
dignifiedquire, line 982: would be great to have a test in here, that checks that the validation function works as expected

iroh-sync/src/store/memory.rs
```diff
@@ -426,10 +426,7 @@ impl crate::ranger::Store<RecordIdentifier, SignedEntry> for ReplicaStoreInstanc
         Ok(self.with_records(|records| {
             records.and_then(|r| {
                 let v = r.get(&(key.author(), key.key().to_vec()))?;
-                if v.timestamp() == key.timestamp() {
-                    return Some(v.clone());
-                }
-                None
+                Some(v.clone())
```
dignifiedquire, line 429: I really dislike this current interface but haven't found a good alternative. I think it this be a source of errors though, if we return values without checking that the timestamp matches

dignifiedquire (approved):
looks good other than my comments

iroh-sync/src/store/memory.rs
Frando: You mean this flow:
* Insert entry `e1`. Clone the resulting `RecordIdentifier`, `r1`
* Insert entry `e2` with identical author and key.
* Get entry `r1`. You get `e2` while you expected to get `e1`?

I'm not sure. Now that there's now history ever, I would say this is expected behavior. Or did you mean something else?

iroh-sync/src/ranger.rs
Frando: see latest commit

iroh-sync/src/store/memory.rs
Frando: Actually, should we move the timestamp back into the `Record`?
It was moved from `Record` to `RecordIdentifier` in #1430 - but maybe it makes more sense semantically to have them in the `Record`. We have both record and id available at all places that needs them anyway.

dignifiedquire: conceptually the timestamp really should be in the `Record`, so yes if we can do it such that we move it back that would be the best case imo

PR #1444, opened 2023-09-01
refactor(iroh-sync): process items one by one without preprocessing

rklaehn:
Discussed this with AljoscheMeyer on discord. I think in the long term I want to rely on these things being sorted, but for now this just processes the ranges individually.

"That depends on how strict do you define your protocol. One approach is to require no such things and simply process arbitrary ranges individually as they come in. The other extreme is to consider overlapping or unordered ranges as invalid. I don't believe that's a good idea, as it adds conceptual complexity.

https://discord.com/channels/949724860232392765/1117853022207352952/1147046914597060638

Also adds a few utility fns on Range

### DIFF
--- a/iroh-sync/src/ranger.rs
+++ b/iroh-sync/src/ranger.rs
@@ -5,6 +5,7 @@
 use std::collections::BTreeMap;
 use std::fmt::Debug;
 use std::marker::PhantomData;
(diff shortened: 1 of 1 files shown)


rklaehn:
Not sure how relevant this is anymore, so gonna close it.

PR #1445, opened 2023-09-01
refactor(iroh-sync): `RangeEntry` trait, byte newtypes for author/namestamp, timestamp into record

Frando:
## Description

Several refactors in `iroh-sync` that emerged out of recent discussion and clarifies the architecture.

* make ranger generic over `E: RangeEntry` in place of `K, V`
* moves the timestamp out of `RecordIdentifier` and into `Record`
* includes the timestamp and the content hash in the fingerprint
* Rename the public key types to `AuthorPublicKey` and `NamespacePublicKey`
* Introduce newtypes `AuthorId` and `NamespaceId` over `[u8; 32]` and use them in most places where we used the public key so far
* in `SignedEntry` keys are now stored as these newtypes and a trait implemented on the store instance is used to resolve the expanded form for crypto operations (works on the fly but uses the cached key if available)
* Remove `ranger::store::AllIterator` because it was always identical to the `RangeIterator`
* #1446 
* #1448 

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-sync/src/ranger.rs
+++ b/iroh-sync/src/ranger.rs
@@ -2,12 +2,20 @@
 //! "Range-Based Set Reconciliation" by Aljoscha Meyer.
 //!
 
-use std::collections::BTreeMap;
+use std::convert::Infallible;
 use std::fmt::Debug;
 use std::marker::PhantomData;
-use std::{cmp::Ordering, convert::Infallible};
+use std::{cmp::Ordering, collections::BTreeMap};
 
 use serde::{Deserialize, Serialize};
+
+pub trait RangeEntry: Debug + Clone + PartialOrd {
+    type Key: RangeKey + PartialEq + Clone + Default + Debug;
+    type Value: Debug + Clone;
+    fn key(&self) -> &Self::Key;
+    fn value(&self) -> &Self::Value;
+    fn as_fingerprint(&self) -> Fingerprint;
+}
 
 /// Stores a range.
 ///
@@ -80,21 +88,11 @@ impl Debug for Fingerprint {
 impl Fingerprint {
     /// The fingerprint of the empty set
     pub fn empty() -> Self {
-        Fingerprint::new(&[][..])
+        Fingerprint(*blake3::hash(&[]).as_bytes())
     }
 
-    pub fn new<T: AsFingerprint>(val: T) -> Self {
+    pub fn new<T: RangeEntry>(val: T) -> Self {
         val.as_fingerprint()
-    }
-}
-
-pub trait AsFingerprint {
-    fn as_fingerprint(&self) -> Fingerprint;
-}
-
-impl<T: AsRef<[u8]>> AsFingerprint for T {
-    fn as_fingerprint(&self) -> Fingerprint {
-        Fingerprint(blake3::hash(self.as_ref()).into())
     }
 }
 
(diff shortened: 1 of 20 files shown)


iroh-sync/src/ranger.rs
```diff
@@ -168,205 +164,73 @@ impl<K, V> MessagePart<K, V> {
...
-{
+pub struct Peer<E: RangeEntry, S: Store<E>> {
     store: S,
     /// Up to how many values to send immediately, before sending only a fingerprint.
     max_set_size: usize,
     /// `k` in the protocol, how many splits to generate. at least 2
     split_factor: usize,
 
-    _phantom: PhantomData<(K, V)>, // why???
+    _phantom: PhantomData<E>, // why???
```
rklaehn, line 227: why why? E is not used directly, so you have to fake use using PhantomData.

iroh-sync/src/sync.rs
```diff
@@ -532,14 +566,12 @@ impl Entry {
 /// The indentifier of a record.
 #[derive(Default, Clone, Serialize, Deserialize)]
 pub struct RecordIdentifier {
```
rklaehn, line 568: Any Vec<u8> with a length >= 64 should be castable to a RecordIdentifier, right? We are not doing any checks other than the length. So I think RecordIdentifier can be done differently. Basically an unsized type.

But... next PR.

rklaehn (approved):
That is quite big. But I really like the change.

iroh-sync/src/ranger.rs
dignifiedquire: well, `Store` uses it, so I think it is dumb that this is needed..

iroh-sync/src/ranger.rs
```diff
@@ -629,11 +491,143 @@ where
...
 mod tests {
     use proptest::prelude::*;
-    use std::{cell::RefCell, fmt::Debug, rc::Rc};
+    use std::{cell::RefCell, collections::BTreeMap, convert::Infallible, fmt::Debug, rc::Rc};
     use test_strategy::proptest;
 
     use super::*;
 
+    #[derive(Debug)]
+    pub struct SimpleStore<K, V> {
```
dignifiedquire, line 500: maybe not needed to be `pub` if used in tests here only?

iroh-sync/src/ranger.rs
```diff
@@ -629,11 +491,143 @@ where
...
+    {
+        type Key = K;
+
+        fn key(&self) -> &Self::Key {
+            &self.0
+        }
+
+        fn as_fingerprint(&self) -> Fingerprint {
+            let mut hasher = blake3::Hasher::new();
+            hasher.update(format!("{:?}", self.0).as_bytes());
```
dignifiedquire, line 525: this should include the value as well probably, to be more consistent in testing this

iroh-sync/src/store/fs.rs
```diff
@@ -390,23 +395,29 @@ impl crate::ranger::Store<RecordIdentifier, SignedEntry> for StoreInstance {
...
+            fp ^= el.as_fingerprint();
         }
 
         Ok(fp)
     }
 
-    fn put(&mut self, k: RecordIdentifier, v: SignedEntry) -> Result<()> {
+    fn put(&mut self, e: SignedEntry) -> Result<()> {
+        // let k = e.id();
+        // let v = e;
```
dignifiedquire, line 406: delete?

iroh-sync/src/store/pubkeys.rs
```diff
@@ -0,0 +1,55 @@
...
+}
+
+impl PublicKeyStore for () {
+    fn public_key(&self, id: &[u8; 32]) -> Result<VerifyingKey, SignatureError> {
+        VerifyingKey::from_bytes(id)
+    }
+}
+
+/// In-memory key storage
+#[derive(Debug, Clone, Default)]
```
dignifiedquire, line 41: not necessary for this PR, but would be good to have this be configurable in max number of keys stored

iroh-sync/src/sync.rs
```diff
@@ -242,6 +241,11 @@ impl<S: ranger::Store<RecordIdentifier, SignedEntry> + 'static> Replica<S> {
         self.inner.read().namespace.id()
     }
 
+    // /// Get the namespace identifier for this [`Replica`] in byte represenation.
+    // pub fn namespace(&self) -> NamespaceId {
+    //     self.inner.read().namespace.id().into()
+    // }
```
dignifiedquire, line 247: delete?

dignifiedquire (approved):
really like the overall changes

iroh-sync/src/ranger.rs
Frando: Added a comment

Issue #1440, opened 2023-08-31
possible bug? adding a new address to the magic sock will remove connections made to us


ramfox:
We only prune endpoints after a call to `set_network_map`. In that method, we overwrite the internal `MagicSock::netmap` & remove any `MagicSock::peer_map` `Endpoints` that are not present in the latest `netmap`.

This removes any connection information about nodes that have connected to us in the wild.

We may very well want to prune these connections in this way.

However, we use `set_network_map` to add new addresses explicitly to the `MagicSock` / `MagicEndpoint` (using `MagicEndpoint::add_known_addr`). This means anytime we add a new address, we are pruning any "natural" connections made. This seems like an unintended side effect, I don't think a user would expect that any time they add an explicit connection, they are potentially closing connections to other (potentially active) nodes.

Frando:
We should change the data flow so that `MagicEndpoint::add_known_addr` sends a message with the peer info to the actor, and does not fully replace the netmap.

flub:
@Frando so can this be closed now or is #1447 not the whole story?

Frando:
I would guess but am not sure, maybe @ramfox can confirm that the situation that triggered this issue does not occur anymore.

ramfox:
closed by #1447 !

PR #1447, opened 2023-09-02
refactor(iroh-net): remove `NetworkMap`

Frando:
## Description

While having another look at the code when thinking about #1440 I realized that the `NetworkMap`is never used apart from serving as a source of entries for the `PeerMap`.

So I removed it. Looks much simpler now, and tests passed for me locally. Not sure if I haven't missed something.

## Notes & open questions

* We never remove entries from the `peer_map`. This is already the case before this PR if I'm not missing something, because we never removed entries from the `NetworkMap`, and only the latter would ever have triggered a removal of entries from the `peer_map`.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.


dignifiedquire (approved):
we need to figure out how to handle deletion & cleanup over time, but that wasn't handled properly before any way

PR #1448, opened 2023-09-02
refactor(iroh-sync): store `RecordIdentifier` as bytes

Frando:
## Description

Builds on #1445. Had an idea - we can make the `RecordIdentifier` be a newtype wrapper around `Bytes`. With this we can expose the `RecordIdentifier` directly as a simple continous byte slice.

I'm not too sure if we gain much over the previous form – Maybe some performance when cloning oftenly. And @rklaehn wanted to have a simple byte representation for store improvements. 

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-sync/src/sync.rs
+++ b/iroh-sync/src/sync.rs
@@ -14,6 +14,7 @@ use std::{
 
 #[cfg(feature = "metrics")]
 use crate::metrics::Metrics;
+use bytes::{Bytes, BytesMut};
 use derive_more::Deref;
 #[cfg(feature = "metrics")]
 use iroh_metrics::{inc, inc_by};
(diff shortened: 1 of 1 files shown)


dignifiedquire:
I like it, should make the key cloning we have in the stores a lot cheaper

rklaehn:
I think this is great. What you might gain is to simplify the memory store and bring it more in line with how the redb file store works. E.g. instead of having a multi level map, you can just do the mem record store as a `BTreeMap<Bytes, RecordValue>` and then do the exact same range tricks as you have to do in the file store anyway.

rklaehn (approved):

PR #1468, opened 2023-09-08
feat(iroh-net): Implement the https probe.

flub:
## Description

This is needed to figure out the closest derp server when UDP is
blocked.  Now that we have multiple derp servers we'll start to rely
on this working.

## Notes & open questions

The Initial commit introduces the feature and can be manually tested against a derp server.  However it comes with a failing test.

The 2nd commit breaks things but is aiming at being able to run a full derp server, including the http(s) services in tests:

It moves out the server part of `src/bin/derper.rs` into `src/derp/derper.rs`.  This latter does not try to do anything related to UX or touch the filesystem.  It's configuration is rather different but works from just a struct.  `src/bin/derper.rs` will have to do all the file reading and convert things into the right shape.  This way this PR will also not affect any configuration file formats.

This part is incomplete so far.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/src/bin/derper.rs
+++ b/iroh-net/src/bin/derper.rs
@@ -231,8 +231,9 @@ struct TlsConfig {
     cert_dir: Option<PathBuf>,
     /// The port on which to serve a response for the captive portal probe over HTTP.
     ///
-    /// The listener is bound to the same IP as specified in the `addr` field. Defaults to 80.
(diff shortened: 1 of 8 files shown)


dignifiedquire:
Closing for now, please reopen when you restart this work

PR #1469, opened 2023-09-08
Use latest version of bao-tree

rklaehn:
## Description

The new version of bao-tree has chunk level granularity of requests, and a slightly different way to handle ranges. This tries it out.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh/tests/cli.rs
```diff
@@ -293,7 +293,7 @@ fn cli_provide_tree_resume() -> Result<()> {
         let get_output = get.unchecked().run()?;
         assert!(get_output.status.success());
         let matches = explicit_matches(match_get_stderr(get_output.stderr)?);
-        assert_eq!(matches, vec!["65.98 KiB"]);
+        assert_eq!(matches, vec!["69.79 KiB"]);
```
rklaehn, line 296: no idea why this got more. need to investigate.

rklaehn:
What is missing is a range canonicalization function in bao-tree in support of the "range outside the valid range" functionality to ensure we return the minimum range. See https://github.com/n0-computer/bao-tree/issues/36

Other than that this is ready to review.

iroh/tests/cli.rs
```diff
@@ -470,9 +474,9 @@ fn make_provider_in(
             rpc_port.unwrap_or("disabled"),
         ],
     )
-    .stderr_null()
+    // .stderr_null()
     // .stderr_file(std::io::stderr().as_raw_fd()) // for debug output
-    .env("RUST_LOG", "debug")
+    // .env("RUST_LOG", "iroh_bytes=debug,iroh_net=warn,iroh=debug,warn")
```
dignifiedquire, line 479: should this be comitted?

dignifiedquire (approved):
one nit, other than that lgtm

Issue #1464, opened 2023-09-05
Switching from Wifi to Cellular with a running iroh node breaks connections


b5:
this issue tracks network interface switching detection. When switching is working properly the magicSocket should transparently handle changes in the underlying connection topology & things should "just work"

dignifiedquire:
I wrote up the handling in tailscale, which we should pretty much just mirror: https://number-zero.notion.site/Network-Change-Monitoring-7ab0ac445c414f158b66df38611a533c?pvs=4

b5:
closed by #1472

PR #1472, opened 2023-09-11
feat(iroh-net): implement network monitoring

dignifiedquire:
## Description

Implements https://github.com/n0-computer/iroh/issues/1464

## Outstanding work

- [x] Integration of monitoring
- [x] OS Specific monitoring
  - [x] macos
  - [x] windows
  - [x] linux
  - [x] android (unusable dummy)
- [x] sleep detection (wall time jumps)
- [x] testing: only manual for now, and a smoke test, hard to test otherwise
- [x] ~~metrics~~: I don't think we need those for now
- [x] cleanup & shutdown

## Notes

Android sucks, seriously google what are you doing. It is not possible to use rtnetlink anymore on android, and the `ConnectivityManager` can only be used from Java/Kotlin. So for now we don't do anything on android. 
The solution is to inject events from the `ConnectivityManager` through a public API, like tailscale does.. 😭 😭 😠 

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh-net/src/net/netmon/actor.rs
```diff
@@ -0,0 +1,257 @@
...
+        if !time_jumped && old_state == &new_state {
+            debug!("no changes detected");
+            return Ok(());
+        }
+
+        let mut is_major = is_major_change(old_state, &new_state);
+        // Check for time jumps
+        if !is_major && time_jumped {
+            is_major = true;
+        }
```
divagant-martian, line 171: is this not the same as 
```suggestion
        let is_major = is_major_change(old_state, &new_state) || time_jumped;
```

iroh-net/src/net/netmon/actor.rs
```diff
@@ -0,0 +1,257 @@
...
+        || s1.pac != s2.pac
+    {
+        return true;
+    }
+
+    for (iname, i) in &s1.interface {
+        let Some(ips) = s1.interface_ips.get(iname) else {
+            // inconsistent dataset, ignore
+            continue;
+        };
```
divagant-martian, line 218: it's out of context for the current pr but it looks like we could store the ips along with the interfaces, saving us storing the name twice every time?

iroh-net/src/net/netmon/actor.rs
```diff
@@ -0,0 +1,257 @@
...
+                cb(is_major).await;
+            });
+        }
+
+        Ok(())
+    }
+
+    /// Reports whether wall time jumped more than 150%
+    /// of `POLL_WALL_TIME_INTERVAL`, indicating we probably just came out of sleep.
+    fn check_wall_time_advance(&mut self) -> bool {
```
divagant-martian, line 190: tailscale mentions this being expensive on mobile and not really useful since phones don't "sleep" as laptops. Do you think we should do it anyway?

dignifiedquire: you are right, I forgot to disable this for mobile

iroh-net/src/net/netmon/actor.rs
dignifiedquire: removed, much cleaner now

iroh-net/src/net/netmon/actor.rs
```diff
@@ -0,0 +1,257 @@
...
+            return true;
+        }
+    }
+
+    false
+}
+
+/// Checks wheter `a` and `b` are equal after ignoring uninteresting
+/// things, like link-local, loopback and multicast addresses.
+fn prefixes_major_equal(a: &[IpNet], b: &[IpNet]) -> bool {
```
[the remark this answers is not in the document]
dignifiedquire: removed

iroh-net/src/net/netmon/actor.rs
```diff
@@ -0,0 +1,257 @@
...
+        } else {
+            false
+        };
+
+        self.wall_time = now;
+        jumped
+    }
+}
+
+fn is_major_change(s1: &State, s2: &State) -> bool {
```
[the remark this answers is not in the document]
dignifiedquire: removed the `equal_filtered`

divagant-martian (approved):
lgtm within my limited knowledge

dignifiedquire:
there is now: https://github.com/n0-computer/iroh/pull/1845

PR #1477, opened 2023-09-13
feat: streaming blob reads over RPC

Frando:
## Description

This makes getting blobs over RPC streaming. Before we'd alloc the full blob in-memory, which of course is a no-go.

## Notes & open questions

Client API methods will be renamed in the upcoming API refactor.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/node.rs
+++ b/iroh/src/node.rs
@@ -27,7 +27,7 @@ use iroh_bytes::get::Stats;
 use iroh_bytes::protocol::GetRequest;
 use iroh_bytes::provider::ShareProgress;
 use iroh_bytes::util::progress::{FlumeProgressSender, IdGenerator, ProgressSender};
-use iroh_bytes::util::{RpcError, RpcResult};
+use iroh_bytes::util::RpcResult;
 use iroh_bytes::{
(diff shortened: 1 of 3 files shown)


iroh/src/node.rs
```diff
@@ -1141,29 +1146,48 @@ impl<D: BaoStore, S: DocStore, C: CollectionParser> RpcHandler<D, S, C> {
...
-                .await
-                .map_err(anyhow::Error::from)?
-                .read_to_end()
-                .await
-                .map_err(anyhow::Error::from)?;
-            Result::<_, anyhow::Error>::Ok(data)
+    fn bytes_get(
+        self,
+        req: BytesGetRequest,
+    ) -> impl Stream<Item = RpcResult<BytesGetResponse>> + Send + 'static {
```
rklaehn, line 1152: This works fine and should not be an issue in this case.

But if quic-rpc would optionally give you a sender instead of requiring a signature that returns a stream the additional channel could be avoided.

rklaehn (approved):
Looks quite reasonable...

PR #1478, opened 2023-09-13
refactor: extend Iroh client, improve API consistency

Frando:
## Description

* Extend Iroh client surface
* Split Iroh client struct into node, blobs, docs
* Use Iroh client in CLI (no more raw RPC calls)
* Improve naming consistency throughout the RPC interface
* Use `iroh_net::NodeAddr` in ticket and get commands
* Improve arguments of `BlobGetRequest`

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/src/magic_endpoint.rs
+++ b/iroh-net/src/magic_endpoint.rs
@@ -31,7 +31,11 @@ pub struct NodeAddr {
 
 impl NodeAddr {
     /// Create a new [`NodeAddr`] from its parts.
-    pub fn from_parts(node_id: PublicKey, derp_region: Option<u16>, endpoints: Vec<SocketAddr>) -> Self {
+    pub fn from_parts(
+        node_id: PublicKey,
+        derp_region: Option<u16>,
+        endpoints: Vec<SocketAddr>,
+    ) -> Self {
         Self {
             node_id,
             derp_region,
--- a/iroh-net/src/magic_endpoint.rs
+++ b/iroh-net/src/magic_endpoint.rs
@@ -31,7 +31,7 @@ pub struct NodeAddr {
 
 impl NodeAddr {
     /// Create a new [`NodeAddr`] from its parts.
-    pub fn new(node_id: PublicKey, derp_region: Option<u16>, endpoints: Vec<SocketAddr>) -> Self {
+    pub fn from_parts(node_id: PublicKey, derp_region: Option<u16>, endpoints: Vec<SocketAddr>) -> Self {
         Self {
             node_id,
             derp_region,
@@ -311,7 +311,7 @@ impl MagicEndpoint {
         let addrs = self.local_endpoints().await?;
         let derp = self.my_derp().await;
         let addrs = addrs.into_iter().map(|x| x.addr).collect();
-        Ok(NodeAddr::new(self.peer_id(), derp, addrs))
+        Ok(NodeAddr::from_parts(self.peer_id(), derp, addrs))
     }
 
     /// Get information on all the nodes we have connection information about.
--- a/iroh-net/src/magic_endpoint.rs
+++ b/iroh-net/src/magic_endpoint.rs
@@ -4,6 +4,7 @@ use std::{net::SocketAddr, sync::Arc, time::Duration};
 
 use anyhow::{anyhow, ensure, Context, Result};
 use quinn_proto::VarInt;
+use serde::{Deserialize, Serialize};
 use tracing::{debug, trace};
 
 use crate::{
@@ -18,7 +19,7 @@ use crate::{
 pub use super::magicsock::EndpointInfo as ConnectionInfo;
 
 /// Address information for a node.
-#[derive(Debug)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 pub struct NodeAddr {
     /// The node's public key.
     pub node_id: PublicKey,
@@ -26,6 +27,17 @@ pub struct NodeAddr {
     pub derp_region: Option<u16>,
     /// Socket addresses where this node might be reached directly.
     pub endpoints: Vec<SocketAddr>,
+}
+
+impl NodeAddr {
+    /// Create a new [`NodeAddr`] from its parts.
+    pub fn new(node_id: PublicKey, derp_region: Option<u16>, endpoints: Vec<SocketAddr>) -> Self {
+        Self {
+            node_id,
+            derp_region,
+            endpoints,
+        }
+    }
 }
 
 /// Builder for [MagicEndpoint]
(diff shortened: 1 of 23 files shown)


iroh-net/src/magic_endpoint.rs
```diff
@@ -28,6 +29,17 @@ pub struct NodeAddr {
     pub endpoints: Vec<SocketAddr>,
 }
 
+impl NodeAddr {
+    /// Create a new [`NodeAddr`] from its parts.
+    pub fn new(node_id: PublicKey, derp_region: Option<u16>, endpoints: Vec<SocketAddr>) -> Self {
```
dignifiedquire, line 34: maybe call this `from_parts`? given the description 😅

iroh/src/client.rs
```diff
@@ -44,9 +56,71 @@ where
...
+pub struct DocsClient<C> {
+    rpc: RpcClient<ProviderService, C>,
+}
+
+impl<C> DocsClient<C>
+where
+    C: ServiceConnection<ProviderService>,
+{
     /// Create a new document author.
     pub async fn create_author(&self) -> Result<AuthorId> {
```
dignifiedquire, line 125: I wonder if we want to add another layer (or potentially to the top layer)

- `iroh.docs.authors.new()` or 
- `iroh.authors.new()`

iroh/src/client.rs
```diff
@@ -60,7 +134,7 @@ where
     }
 
     /// Create a new document.
-    pub async fn create_doc(&self) -> Result<Doc<C>> {
+    pub async fn create(&self) -> Result<Doc<C>> {
```
dignifiedquire, line 137: didn't we settle on `new` for creation at some point? cc @b5

iroh/src/client.rs
```diff
@@ -80,21 +154,21 @@ where
...
-    pub async fn list_docs(&self) -> Result<impl Stream<Item = Result<NamespaceId>>> {
+    pub async fn list(&self) -> Result<impl Stream<Item = Result<NamespaceId>>> {
         let stream = self.rpc.server_streaming(DocListRequest {}).await?;
         Ok(flatten(stream).map_ok(|res| res.id))
     }
 
     /// Get a [`Doc`] client for a single document. Return an error if the document cannot be found.
-    pub async fn get_doc(&self, id: NamespaceId) -> Result<Doc<C>> {
-        match self.try_get_doc(id).await? {
+    pub async fn get(&self, id: NamespaceId) -> Result<Doc<C>> {
```
dignifiedquire, line 163: This is a very odd method. The rust convention if there is a `try_foo` and `foo` is to panic if the thing is missing, so I think we should only have `get` and it should return `Result<Option>`

iroh/src/client.rs
```diff
@@ -104,37 +178,99 @@ where
...
+#[derive(Debug, Clone)]
+pub struct BlobsClient<C> {
+    rpc: RpcClient<ProviderService, C>,
+}
+
+impl<C> BlobsClient<C>
+where
+    C: ServiceConnection<ProviderService>,
+{
+    /// Read a single blob as a streaming [`BlobReader`].
```
dignifiedquire, line 193: I would describe this just as as `Stream the contents of a blob` or sth like that. The `BlobReader` is not really important, as it is just the type for the stream.

iroh/src/client.rs
```diff
@@ -104,37 +178,99 @@ where
...
+    /// Read all bytes of single blob.
     ///
-    /// Note: This reads the full blob into memory.
-    pub async fn get_bytes(&self, hash: Hash) -> Result<Bytes> {
-        let mut stream = self.get_bytes_stream(hash).await?;
-        stream.read_to_end().await
+    /// This allocates a buffer for the full blob. Use only if you know that the blob you're
+    /// reading is small. If not sure, use [`Self::read`] and check the size with
+    /// [`BlobReader::size`] before calling [`BlobReader::read_to_end`].
+    pub async fn read_to_end(&self, hash: Hash) -> Result<Bytes> {
```
dignifiedquire, line 203: Maybe `read_to_bytes` matching the `read_to_vec` convention I have seen used before?

iroh/src/client.rs
```diff
@@ -104,37 +178,99 @@ where
...
     }
 
-    /// Get statistics of the running node.
-    pub async fn stats(&self) -> Result<HashMap<String, CounterStats>> {
-        let res = self.rpc.rpc(StatsGetRequest {}).await??;
-        Ok(res.stats)
+    /// Validate hashes on the running node.
+    ///
+    /// If `repair` is true, repair the store by removing invalid data.
+    pub async fn validate(
```
dignifiedquire, line 231: shouldn't this be on the `node` part of things?

iroh/src/client.rs
```diff
@@ -104,37 +178,99 @@ where
...
 
-    /// Get information about the different connections we have made
-    pub async fn connections(&self) -> Result<impl Stream<Item = Result<ConnectionInfo>>> {
-        let stream = self.rpc.server_streaming(ConnectionsRequest {}).await?;
-        Ok(flatten(stream).map_ok(|res| res.conn_info))
+    /// Download a blob from another node and add it to the local database.
+    pub async fn get(
+        &self,
+        req: BlobGetRequest,
+    ) -> Result<impl Stream<Item = Result<GetProgress>>> {
```
dignifiedquire, line 246: If I was using this api, I would expect the result here to be the same as `.read()` tbh

iroh/src/commands.rs
```diff
@@ -332,53 +327,49 @@ pub enum NodeCommands {
...
                     );
                 }
             }
             Self::Status => {
-                let response = client.rpc(StatusRequest).await?;
-
+                let response = iroh.node.status().await?;
                 println!("Listening addresses: {:#?}", response.listen_addrs);
-                println!("PeerID: {}", response.peer_id);
+                println!("Node public key: {}", response.peer_id);
```
dignifiedquire, line 358: `peer_id` -> `node_id` I think

iroh/src/dial.rs
```diff
@@ -116,9 +106,9 @@ impl Ticket {
         self.hash
     }
 
-    /// The [`PublicKey`] of the provider for this ticket.
-    pub fn peer(&self) -> PublicKey {
-        self.peer
+    /// The [`NodeAddr`] of the provider for this ticket.
+    pub fn peer(&self) -> &NodeAddr {
```
dignifiedquire, line 110: should be called `node_addr()`

iroh/src/client.rs
Frando: We did for the CLI, yes. I'm happy to change here too, but wondering: Doesn't `new` here violate Rust naming conventions, as it's not a constructor but a method taking `&self`?

iroh/src/client.rs
Frando: What about `download`?

iroh/src/client.rs
Frando: My reasoning for a method that returns just `Result<Doc>` is that I would think that in the vast majority of cases you have a document ID and not found equals error, so skipping the `ok_or_else` or `context` call makes application code nicer to read.

No nice naming scheme comes to mind, so I changed it for now to always return `Result<Option<Doc>>`.

iroh/src/client.rs
Frando: Not sure - it deals with blobs only. No strong opinion.

iroh/src/client.rs
Frando: Changed it to `iroh.authors.create` for now.

iroh/src/client.rs
Frando: Changed it to download in a recent commit. Reads nice IMO.

iroh/src/client.rs
dignifiedquire: I‘d rather be consistent, and use new tbh, what do others think?

dignifiedquire (approved):
this is overall an improvement in either case, so lets merge this and settle any other changes to the API in a new issue/PR

Issue #1481, opened 2023-09-15
iroh-net: Persist peer information


divagant-martian:
To ease start-up operations it would be useful to have a set of known peers on start up.
There are a couple of questions surrounding this issue:
1. Where to store this information
2. What information to store
3. When to persist this information

divagant-martian:
Some thoughts about the relevant questions
## 1. Where to store the information
It was suggested to use the current db. Noting this here to confirm. I have not yet checked where do we start the db and how we interact with it
## 2. What information to store:
I was pointed to the `PeerMap`: https://github.com/n0-computer/iroh/blob/c6935bddbef652982d58145c7936cb9efb1023a3/iroh-net/src/magicsock/endpoint.rs#L1029-L1035

This is my understanding so far, let me know if anything is not correct:
- The "id" (an `usize`) is a runtime-only id, as far as I can tell it's only ever used to manage pending pings. Since we have the following mappings:
  - `by_node_key: {PublicKey -> Id}`
  - `by_id: {Id -> Endpoint}` 
  
  from intuition and from this I would assume the real identifier is the `PublicKey`, I would thus not persist the "id".
- These are the contents of an `Endpoint`: lots of interesting things here
  https://github.com/n0-computer/iroh/blob/c6935bddbef652982d58145c7936cb9efb1023a3/iroh-net/src/magicsock/endpoint.rs#L51-L84
  - `id: usize` as mentioned above I would not persist this
  - `conn_sender: mpsc::Sender<ActorMessage>:` This seems to be a back reference to the actor holding the `PeerMap`. This can be filled by the actor on start up 
  - `quic_mapped_addr: QuicMappedAddr`: this seems to be relevant only on runtime as it can be `generate`d so I wouldn't consider this to be necessary to be persisted.
  - `conn_public_key: PublicKey` trying to get rid of this in #1482 
  - `public_key: PublicKey` should be persisted
  - `last_full_ping: Option<Instant>` set to `None`
  - `derp_region: Option<u16>` persisted
  - `best_addr: Option<AddrLatency>` set to `None`?
  - `best_addr_at: Option<Instant>` set to `None`? 
  - `trust_best_addr_until: Option<Instant>` set to `None`?
  - `endpoint_state: HashMap<SendAddr, EndpointState>` set to `default()`
  - `is_call_me_maybe_ep: HashMap<SocketAddr, bool>` set to `default()`
  - `pending_cli_pings: Vec<PendingCliPing>` set to `default()`
[...]

PR #1482, opened 2023-09-15
(iroh-net): remove our public key from the endpoint

divagant-martian:
## Description

While trying to understand the information we store for peers for #1481 I noticed this can be removed. It's more efficient in space in a couple of ways:
- It's not duplicated for every known peer
- Does not need to be cloned for every sent ping
- The `ActorMessage::SendDiscoMessage` that was removed had a size of 512bytes, making it the largest message of the enum. By replacing it with the two new ones the size of the `ActorMessage` size is reduced from 512 to 248. The whole change reduces the size of `Endpoint` from 784 to 560.

## Notes & open questions

One place where this might be potentially less efficient is in calling `public_key()` on the `secret_key` for every sent ping. If this computation is expensive we can just cache the public key and use it there. Maybe @dignifiedquire knows if this should be done or if it's fine like this

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/src/magicsock.rs
+++ b/iroh-net/src/magicsock.rs
@@ -852,10 +852,15 @@ enum ActorMessage {
         derp_region: u16,
         endpoint_id: usize,
     },
-    SendDiscoMessage {
+    SendPing {
         dst: SendAddr,
         dst_key: PublicKey,
-        msg: disco::Message,
+        tx_id: stun::TransactionId,
+    },
+    SendCallMeMaybe {
+        dst: SendAddr,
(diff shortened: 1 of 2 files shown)


dignifiedquire (approved):
this is good, the public_key is already cached as it just calls to https://github.com/dalek-cryptography/curve25519-dalek/blob/c157a1ed6defda81c8113531c9db6d7a85ff7f5a/ed25519-dalek/src/signing.rs#L165 under the hood

divagant-martian:
ohh ofc, thanks! merging

PR #1486, opened 2023-09-17
refactor(iroh-net): remove `iroh-net::config::Node` since limited to its used fields it's redundant

divagant-martian:
## Description
the `Node` type contains:
https://github.com/n0-computer/iroh/blob/f2f3ead655dee94bd01c9f4fdcc8730457f51ffd/iroh-net/src/config.rs#L157-L174
from these fields:
- `name` is always `None`, so unused.
- `addresses` is always calculated from the the `endpoints` so it can be removed.
After this, the only remaining fields are the key, endpoints and derp region. This is equivalent to the `NodeAddr` type, making `Node` redundant
https://github.com/n0-computer/iroh/blob/f2f3ead655dee94bd01c9f4fdcc8730457f51ffd/iroh-net/src/magic_endpoint.rs#L20-L29

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/src/magicsock.rs
+++ b/iroh-net/src/magicsock.rs
@@ -2235,12 +2235,15 @@ impl Actor {
 
     #[instrument(skip_all)]
     fn add_known_addr(&mut self, addr: NodeAddr) {
+        let NodeAddr {
+            node_id: key,
+            derp_region,
+            endpoints,
+        } = addr;
         let n = config::Node {
-            name: None,
-            addresses: addr.endpoints.iter().map(|e| e.ip()).collect(),
-            endpoints: addr.endpoints.clone(),
-            key: addr.node_id,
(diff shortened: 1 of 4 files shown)


iroh-net/src/magicsock.rs
```diff
@@ -2235,12 +2235,15 @@ impl Actor {
...
         let n = config::Node {
-            name: None,
-            addresses: addr.endpoints.iter().map(|e| e.ip()).collect(),
-            endpoints: addr.endpoints.clone(),
-            key: addr.node_id,
-            derp: addr.derp_region,
+            endpoints,
+            key,
+            derp_region,
         };
```
divagant-martian, line 2247: from this I realized this can be removed completely

Frando (approved):
LGTM!

Issue #1487, opened 2023-09-17
decide what fields of type `PublicKey` are called


divagant-martian:
Right now we have a variety of names for fields of type `PublicKey`. IMO, some of those are consistent, some others not so much.
We have:
- `public_key`
- `peer`
- `peer_id`
- `node_key`
- `node_id`
- `*key` (`srckey`, `dstkey`, `client_key` etc)
The ones I think do not really fit are the `node_key`, and `node_id` since it's not clear if a node and a peer are the same thing or different concepts.

**why this matters**: code is easier to understand when concepts have a consistent naming

flub:
I totally agree with "why this matters"!  Let's improve this.

Surprisingly I think `node` has grown on me, to me it means "a node in an iroh network".  The "peer" term is maybe more associated with iroh nodes with whom you're directly communicating, i.e. have an open connection.  The other benefit of the "peer" term is that it emphasises the equality of the participants.  But perhaps that difference it too subtle and we should only pick one.

The first time I struggled with the `node` terminology was in the derper where you also have "derp node" which is not an iroh node.  I've tried to consistently call those `derp_node` in code I've been touching but that's not complete yet.

flub:
Also, as we have a `NodeAddr` type prominently now maybe `node_id` for `PublicKey` is not that crazy?

divagant-martian:
`NodeAddr` was in fact renamed to `PeerAddr` :sweat_smile:, your comment about derp nodes both adds to the confusion and helps clarify it, ty. For Derp "servers" that don't actually participate actively in the network I think `node` works fine (`derp_node` even better). For `iroh` ones I'm leaning towards `peers` basically because it's been used in the whole gossip and sync code. As long as it's clear what is what I'm totally in favor of any of both options

flub:
So if I understand you right:

- `peer` is preferred.
- `peer_id` can be acceptable?
- `node` & `node_id` can be fine if the context is clear.
- keep derp mesh nodes `derp_node`.
- avoid the other names

Is that a reasonable conclusion?  If so, when do we close this issue?

divagant-martian:
I think the issue is mainly about consensus. So I would consider this closed once the team agrees on something and then whatever remnants we have can be slowly addressed. The starting point is to at least not introduce further inconsistencies

divagant-martian:
there seems to be little to no interest in general in keeping this kind of consistency so closing, thanks for the discussion :pray:

PR #1488, opened 2023-09-18
feat(iroh-net): persist known peer info

divagant-martian:
## Description

Rough list of changes
- add an option to the `MagicEndpointBuilder` to set the path in which peers are saved. This is used to read peers on start up and store them on an interval and on shutdown.
- add `KnownPeers` to ease serializing and deserializing peer info
- move the logic of `add_known_addr` to the `PeerMap` to facilitate instantiating from a known peer list
- Add an `IrohPath` for the peers.
- Add an option to the `Node`s builder to set the path, passed down to the endpoint.
- For serialization it makes sense to have an object modelling the peer addressing info. This is the `AddrInfo`, which now replaces the custom `IrohInfo`

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh-gossip/src/net.rs
```diff
@@ -1,18 +1,14 @@
...
-use std::{
-    collections::HashMap, fmt, future::Future, net::SocketAddr, sync::Arc, task::Poll,
-    time::Instant,
-};
-
 use anyhow::{anyhow, Context};
 use bytes::{Bytes, BytesMut};
 use futures::{stream::Stream, FutureExt};
 use genawaiter::sync::{Co, Gen};
+use iroh_net::magic_endpoint::AddrInfo as IrohInfo;
```
Frando, line 7: I'd say remove the `as IrohInfo` and just use `AddrInfo`

iroh-net/src/magic_endpoint.rs
```diff
@@ -422,6 +438,21 @@ impl MagicEndpoint {
         Ok(())
     }
 
+    /// Get addressing information of this [`MagicEndpoint`].
+    pub async fn my_addr_info(&self) -> Result<AddrInfo> {
```
Frando, line 442: I see that `magicsock/endpoint.rs` now has `addr_info` which returns exactly what we return here. Maybe route to that to have 1, not 2 async calls to the actor here. But can also be a followup.

Frando (commented):
Looks good!

iroh-net/src/magic_endpoint.rs
```diff
@@ -391,16 +408,15 @@ impl MagicEndpoint {
     /// If no UDP addresses are added, and the given `derp_region` cannot be dialed, it will error.
     pub async fn add_known_addrs(
         &self,
-        node_id: PublicKey,
+        peer: PublicKey,
```
dignifiedquire, line 411: weren't we calling these `node_id`s? why the change back?

iroh-net/src/magicsock.rs
```diff
@@ -400,6 +409,24 @@ impl MagicSock {
             .instrument(info_span!("derp.actor")),
         );
 
+        // load the peer data
+        let peer_map = match peers_path.as_ref() {
+            Some(path) if path.exists() => {
+                match PeerMap::load_from_file(path, actor_sender.clone()) {
```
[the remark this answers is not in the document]
dignifiedquire: maybe just add a bulk add addr method?

iroh-net/src/magic_endpoint.rs
```diff
@@ -17,18 +18,32 @@ use crate::{
...
-/// Builder for [MagicEndpoint]
+impl PartialEq for AddrInfo {
+    fn eq(&self, other: &Self) -> bool {
+        if self.derp_region != other.derp_region {
+            return false;
+        }
+        if self.endpoints.len() != other.endpoints.len() {
+            return false;
+        }
+        let mut self_eps = self.endpoints.clone();
```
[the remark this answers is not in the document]
divagant-martian: I did that but changes felt intrusive to everything that calls connect. But agree it's a better option in general

iroh-net/src/magicsock/endpoint.rs
```diff
@@ -1112,6 +1193,15 @@ impl PeerMap {
         self.by_ip_port.insert(*ipp, id);
     }
 
+    /// Saves the known peer info to the given path, returning the number of peers persisted.
+    pub(super) fn save_to_file(&self, path: &Path) -> anyhow::Result<usize> {
+        let known_peers = self.known_peers();
+        let count = known_peers.len();
+        let serialized = postcard::to_stdvec(&known_peers)?;
+        std::fs::write(path, serialized)?;
```
dignifiedquire, line 1201: should this be async?

iroh-net/src/magic_endpoint.rs
divagant-martian: I asked about this in #1487 
We are using peer_id through the gossip and sync code so at this point it feels more consistent than node to me. As long as we pick one I'm happy to oblige

iroh-net/src/magicsock/endpoint.rs
```diff
@@ -1112,6 +1193,15 @@ impl PeerMap {
         self.by_ip_port.insert(*ipp, id);
     }
 
+    /// Saves the known peer info to the given path, returning the number of peers persisted.
+    pub(super) fn save_to_file(&self, path: &Path) -> anyhow::Result<usize> {
+        let known_peers = self.known_peers();
```
dignifiedquire, line 1198: This allocates the whole thing in memory, effectively duplicating memory usage for this moment. I wonder if we could serialize it in an iterator/streaming fashion. eg sth like this

```rust
let mut file = std::fs::File::open(path)?;
file.write_all(HEADER)?;
let mut buffer = Vec::new();
for peer in &self.known_peers() {
  let ser = postcard::to_slice(peer, &mut buffer)?;
  file.write_all(ser)?;
}
file.flush()?;
```

dignifiedquire: (and make `self.known_peers()` return an iterator instead of a vec)

iroh/src/config.rs
```diff
@@ -58,6 +60,7 @@ impl From<&IrohPaths> for &'static str {
             IrohPaths::BaoFlatStorePartial => "blobs-partial.v0",
             IrohPaths::DocsDatabase => "docs.redb",
             IrohPaths::Console => "console",
+            IrohPaths::PeerData => "peers.postcard",
```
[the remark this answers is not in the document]
dignifiedquire: I'd rather do sth like `peers.bin`, but more a personal preference. Maybe we should include a version number?

iroh-net/src/magic_endpoint.rs
dignifiedquire: I see, I don't mind either way, as long as we are somewhat consistent about it 😅

iroh-net/src/magicsock/endpoint.rs
dignifiedquire: Also, we shouldn't directly write to the path, but rather to a `tmp` file (fine in the same location) and rename when we have successfully written, to avoid having errors during writing leaving corrupted files behind

iroh-net/src/magicsock.rs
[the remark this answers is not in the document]
divagant-martian: the endpoint is passed because the value is stored for each endpoint, but it's not used and peers are not sent via this (or any channel) so I don't think there is possibility of deadlock. 

This seems dependent on trusting that the inner code does not use the channel, but I think the sync/non async coloring saves us here:  `load_from_file` is non-async, so any async call would need to be spawned outside this(effectively doing what you suggest). Any other non-async call of the `try_send` fashion would fail in that scenario. So I think this is safe

iroh/src/config.rs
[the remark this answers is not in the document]
divagant-martian: since we have already introduced strum, I'll move that test to iterate the values so that future paths are automatically added.

About the version number, what would be the purpose of this? is it for us to understand what info is in there? otherwise we don't store or read a peer data version that would help in an upgrade for example. Maybe we can tackle upgrading a peer data format version once it changes

the `.postcard` was more of a hunch I guess, I'm fine with any

iroh-gossip/src/net.rs
divagant-martian: done in [afa69d0](https://github.com/n0-computer/iroh/pull/1488/commits/afa69d0c4e051eac5a6f9ff20a8fc8df8ce7f4d7)

iroh-net/src/magic_endpoint.rs
```diff
@@ -154,6 +172,14 @@ impl MagicEndpointBuilder {
         self
     }
 
+    /// Optionally set the path where peer info should be stored.
```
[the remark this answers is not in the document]
divagant-martian: done in d036c3c

iroh-net/src/magicsock/endpoint.rs
divagant-martian: didn't get to do this but will look into it

iroh-net/src/magicsock/endpoint.rs
divagant-martian: didn't get to do this but will look into it

iroh/src/config.rs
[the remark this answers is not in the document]
divagant-martian: this is done in https://github.com/n0-computer/iroh/pull/1507

iroh-net/src/magicsock/endpoint.rs
```diff
@@ -1018,6 +1043,53 @@ pub(super) struct PeerMap {
...
+        })
+    }
+
+    /// Create a new [`PeerMap`] from data stored in `path`.
+    pub fn load_from_file(
+        path: impl AsRef<Path>,
+        msock_sender: mpsc::Sender<ActorMessage>,
+    ) -> anyhow::Result<Self> {
+        let mut me = PeerMap::default();
+        let contents = std::fs::read(path)?;
```
Frando, line 1061: We could read in chunks here. However the file would need to get *really* big for this to be an issue. A `PeerAddr` will take less than 100 bytes even with a few socket addresses. So if you store 10.000 peers that'd do a ~1MB allocation here.
I think it's fine to keep as-is now, and we could move to chunked reading easily in a follow-up.

Frando (approved):
LGTM! Let's get this in :+1: 

We might want to add a full test with a "recovering" endpoint (shutdown, restart, reconnect) to test the save-on-shutdown and read-on-bin, but that can also be a followup I'd say.

PR #1491, opened 2023-09-18
feat(sync): track incoming sync requests, allow subscriptions without sync, close inactive replicas

Frando:
## Description

This PR adds a few missing parts for sync in iroh:

* Allow to subscribe to replicas if they're not in the set of syncing replicas (this was not possible before).
* Properly track open replicas in the `LiveSync` and close them once they're not needed anymore. Therefore adds `iroh_sync::store::close_replica` to remove a replica instance from the store-internal cache of open replicas. This is needed because otherwise it grows unbounded (for anchor nodes especially).
* Pass incoming connections to the `LiveSync` actor and handle them there. Before, incoming connections for sync were *always* processed if a replica with the requested `NamespaceId`  is found in the local replica store. This PR changes this mechanism so that incoming sync connections are passed into the sync engine first. This leads to the changes that motivate this PR:
    * Incoming sync requests are only answered if the document (replica) in question was opened for sync before (via a call to `start_sync`). Note that this is *not* persisted at the moment, but should be (separate change though).
    * We can track the sync status for outgoing *and incoming* sync flows. This will reduce the number of sync connections, because currently we will oftenly double-sync in both directions (unneeded, because the sync is always bidirectional). Now, we abort if a sync is already in progress.
* Expose events about successfull and failed sync runs  to the upper layer via the document subscriptions with a new `LiveEvent:.SyncFinished` that reports all available information

## Notes & open questions

Do we want to track history of syncs internally or leave it up to the app via the events? Not sure myself.

The diff of `sync_engine/live.rs` is a little hard to read, because there's a change from `TopicId` to `NamespaceId` in a few places that is mostly unrelated. Sorry for the mixup, but it makes sense to track everything by `NamespaceId` and it ended up in a commit here.

Logging is much improved in this PR. 

Maybe we can improve the code flow a little to make things more obvious.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-sync/src/store.rs
+++ b/iroh-sync/src/store.rs
@@ -1,6 +1,7 @@
 //! Storage trait and implementation for iroh-sync documents
 
 use anyhow::Result;
+use iroh_bytes::Hash;
 use rand_core::CryptoRngCore;
 use serde::{Deserialize, Serialize};
 
@@ -23,6 +24,11 @@ pub trait Store: std::fmt::Debug + Clone + Send + Sync + 'static {
 
     /// Iterator over entries in the store, returned from [`Self::get_many`]
     type GetIter<'a>: Iterator<Item = Result<SignedEntry>>
+    where
+        Self: 'a;
+
+    /// Iterator over all content hashes in the store, returned from [`Self::content_hashes`]
+    type ContentHashesIter<'a>: Iterator<Item = Result<Hash>>
     where
         Self: 'a;
 
@@ -84,6 +90,9 @@ pub trait Store: std::fmt::Debug + Clone + Send + Sync + 'static {
         author: AuthorId,
         key: impl AsRef<[u8]>,
     ) -> Result<Option<SignedEntry>>;
+
+    /// Get all content hashes of all replicas in the store.
+    fn content_hashes(&self) -> Result<Self::ContentHashesIter<'_>>;
 }
 
 /// Filter a get query onto a namespace
--- a/iroh-sync/src/store/fs.rs
+++ b/iroh-sync/src/store/fs.rs
@@ -5,6 +5,7 @@ use std::{cmp::Ordering, collections::HashMap, path::Path, sync::Arc};
 use anyhow::Result;
 use derive_more::From;
 use ed25519_dalek::{SignatureError, VerifyingKey};
+use iroh_bytes::Hash;
 use ouroboros::self_referencing;
 use parking_lot::RwLock;
 use redb::{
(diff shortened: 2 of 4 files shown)


iroh-sync/src/net/codec.rs
```diff
@@ -260,10 +293,16 @@ mod tests {
...
         let bob_task = tokio::task::spawn(async move {
-            run_bob::<store::memory::Store, _, _>(
+            run_bob::<store::memory::Store, _, _, _, _>(
                 &mut bob_writer,
                 &mut bob_reader,
-                bob_replica_store_task,
+                |namespace, _| {
+                    futures::future::ready(
+                        bob_replica_store_task
+                            .open_replica(&namespace)
```
dignifiedquire, line 302: this is a blocking call, we should probably mark it as such

iroh-sync/src/net/codec.rs
```diff
@@ -459,10 +498,12 @@ mod tests {
         let (mut bob_reader, mut bob_writer) = tokio::io::split(bob);
         let bob_store = bob_store.clone();
         let bob_task = tokio::task::spawn(async move {
-            run_bob::<S, _, _>(
+            run_bob::<S, _, _, _, _>(
                 &mut bob_writer,
                 &mut bob_reader,
-                bob_store,
+                |namespace, _| {
+                    futures::future::ready(bob_store.open_replica(&namespace).map(Into::into))
```
dignifiedquire, line 505: same here

iroh/src/commands/sync.rs
```diff
@@ -247,6 +247,9 @@ impl DocCommands {
                         LiveEvent::ContentReady { hash } => {
                             println!("content ready: {}", fmt_short(hash.as_bytes()))
                         }
+                        LiveEvent::SyncFinished(details) => {
+                            println!("sync finished: {details:?}")
```
dignifiedquire, line 251: maybe `sync finished for {fmt_short(..)}`?

dignifiedquire (approved):

iroh-sync/src/net/codec.rs
Frando: deferred for later: https://github.com/n0-computer/iroh/issues/1502

iroh-sync/src/net/codec.rs
Frando: deferred to https://github.com/n0-computer/iroh/issues/1502

PR #1493, opened 2023-09-19
refactor: Use `iroh_net::PeerAddr` more

Frando:
## Description

Replaces custom types with `iroh_net::PeerAddr`

## Notes & open questions

na

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -564,8 +564,8 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9ab77dbd8adecaf3f0db40581631b995f312a8a5ae3aa9993188bb8f23d83a5b"
 dependencies = [
  "crossterm",
- "strum",
- "strum_macros",
+ "strum 0.24.1",
+ "strum_macros 0.24.3",
  "unicode-width",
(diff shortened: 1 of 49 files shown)


divagant-martian (commented):
I actually this more or less just this in #1481 based on #1486 with a different take, since I need to serialize the `(derp_region, endpoints)` by itself (just like gossip does), thus using it there as well. 

Can you take a look? I think in principle our object is the same and I would then add the convenience methods you created like `with_derp_region` to extend the api in the same fashion

Frando:
Alright, yes this is pretty similar. We can bikeshed if we want to keep the `NodeAddr` type, which would then be `NodeAddr { node_id: PublicKey, info: AddrInfo }` and use that for `connect` as in this PR, or change `connect` to take `connect(node_id: PublicKey, addr_info: AddrInfo, alpn: &str)`.

But let's get the (much more important) #1488 in first and then I can update this PR afterwards.

dignifiedquire (approved):

divagant-martian (approved):

PR #1495, opened 2023-09-19
feat(sync): Add `close_replica` and enable subscriptions without active sync

Frando:
## Description

Based on #1491 

This adds two related features that were missing so far:

* Add `iroh_sync::store::close_replica` to remove a replica instance from the store-internal cache of open replicas. This is needed because otherwise it grows unbounded (for anchor nodes especially).
* Properly handle open replicas in the `LiveSync`, and through this allow to subscribe to replica events for replicas that are not in the set of actively syncing replicas. This was not possible before (you could only subscribe to replicas that are active for sync)
* Add test for subscribing without sync

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/sync_engine/live.rs
+++ b/iroh/src/sync_engine/live.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::HashMap,
+    collections::{HashMap, HashSet},
     fmt,
     net::SocketAddr,
     str::FromStr,
@@ -9,6 +9,7 @@ use std::{
 
 use crate::downloader::{DownloadKind, Downloader};
 use anyhow::{anyhow, bail, Result};
+use flume::r#async::RecvStream;
 use futures::{
     future::{BoxFuture, Shared},
     stream::{BoxStream, FuturesUnordered, StreamExt},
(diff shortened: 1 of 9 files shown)


iroh/src/sync_engine/live.rs
```diff
@@ -506,8 +528,16 @@ impl<S: store::Store, B: baomap::Store> Actor<S, B> {
         Ok(())
     }
 
+    fn get_replica_if_syncing(&self, namespace: &NamespaceId) -> Option<Replica<S::Instance>> {
+        if !self.syncing_replicas.contains(namespace) {
+            None
+        } else {
+            self.replica_store.open_replica(namespace).ok().flatten()
```
dignifiedquire, line 535: should probably return the error or at least log it

dignifiedquire (approved):
lgtm, other than my small notes

Frando:
Thanks! Addressed the comments. As this is based on #1491 - Should I merge into that one now or wait for that to be reviewed, to keep reviewing easier?

b5:
I just ran this up in iOS land, and it's fixed a bug I was experiencing with the node crashing on an attempt to load keys!

dignifiedquire:
feel free to pick your merge order, I don’t mind for review

PR #1496, opened 2023-09-19
GC

rklaehn:
## Description

Adds the concept of GC.

There is an additional persisted table roots, which is name -> (hash, format) where `name` is any blob. Currently `format` is either blob or collection, but in the future this might become a multiformat or something.

The GC computes the set of all live hashes from these roots and user-provided extra roots. This is the mark phase, which mostly involves IO for opening collections and the like.

Then it does a second phase which is the sweep phase. In this phase it just looks at all blobs and deletes the ones which are not in the life set. This involves the actual deletion, which may or may not be expensive depending on the file system.

Anything hash that is added between the mark phase and the sweep phase is protected against being deleted, and will only be considered during the next mark phase.

To protect stuff against deletion there are two mechanisms:

- named roots, see above, which are persisted to disk
- temp pins, which are also (hash, format) pairs but are purely in memory

While adding files or during sync, temp pins can be used to protect temporary data from gc.

Todo:

- [x] actually persist the roots table
- [x] change RPC api to list named roots and to manipulate named roots
- [ ] think more about all the possible race conditions
- [x] format blob/collection vs multiformats format
- [x] document integration
- [x] add GC to the mem version

## Notes & open questions

All that I need from iroh-docs is an iterator of all hashes as (hash, Format::Blob) pairs. ~The exact timing needs some thought for this to be safe. I think it needs to happen after the life set is cleared, which makes the current impl buggy.~

I added the ability to delete blobs even if they are pinned. Not sure if everybody agrees, but I think you should be able to do this. Imagine you got something that you really don't want to have due to legal reasons.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-bytes/src/baomap.rs
+++ b/iroh-bytes/src/baomap.rs
@@ -138,10 +138,10 @@ pub trait ReadableStore: Map {
     /// This function should not block to perform io. The knowledge about
     /// existing blobs must be present in memory.
     fn blobs(&self) -> Box<dyn Iterator<Item = Hash> + Send + Sync + 'static>;
-    /// list all roots (collections or other explicitly added things) in the database
+    /// list all tags (collections or other explicitly added things) in the database
     ///
     /// This function should not block to perform io. The knowledge about
-    /// existing roots must be present in memory.
+    /// existing tags must be present in memory.
     fn tags(&self) -> Box<dyn Iterator<Item = (Bytes, Cid)> + Send + Sync + 'static>;
 
     /// Temp pins
--- a/iroh-bytes/src/baomap.rs
+++ b/iroh-bytes/src/baomap.rs
@@ -11,7 +11,7 @@ use crate::{
 };
 use bao_tree::{blake3, ChunkNum};
 use bytes::Bytes;
-use futures::{future::BoxFuture, stream::LocalBoxStream, FutureExt, StreamExt};
+use futures::{future::BoxFuture, stream::LocalBoxStream, StreamExt};
 use genawaiter::rc::{Co, Gen};
 use iroh_io::AsyncSliceReader;
 use range_collections::RangeSet2;
@@ -195,19 +195,10 @@ pub trait Store: ReadableStore + PartialMap {
     fn import_bytes(&self, bytes: Bytes, format: BlobFormat) -> BoxFuture<'_, io::Result<TempTag>>;
 
     /// Set a named pin
-    fn set_tag(&self, name: Bytes, hash: Option<Cid>) -> BoxFuture<'_, io::Result<()>> {
-        let _ = name;
-        let _ = hash;
-        async move { Ok(()) }.boxed()
-    }
+    fn set_tag(&self, name: Bytes, hash: Option<Cid>) -> BoxFuture<'_, io::Result<()>>;
 
     /// Create a temporary pin for this store
-    fn temp_tag(&self, cid: Cid) -> TempTag {
-        TempTag {
-            cid,
-            liveness: None,
-        }
-    }
+    fn temp_tag(&self, cid: Cid) -> TempTag;
 
     /// Traverse all roots recursively and mark them as live.
     ///
@@ -258,28 +249,18 @@ pub trait Store: ReadableStore + PartialMap {
     }
 
     /// Clear the live set.
-    fn clear_live(&self) {
-        let _ = self;
-    }
+    fn clear_live(&self);
 
     /// Add the given hashes to the live set.
     ///
     /// This is used by the gc mark phase to mark roots as live.
-    fn add_live(&self, live: impl IntoIterator<Item = Hash>) {
-        let _ = live;
-    }
+    fn add_live(&self, live: impl IntoIterator<Item = Hash>);
 
     /// True if the given hash is live.
-    fn is_live(&self, hash: &Hash) -> bool {
-        let _ = hash;
-        false
-    }
+    fn is_live(&self, hash: &Hash) -> bool;
 
     /// physically delete the given hash from the store.
-    fn delete(&self, hash: &Hash) -> BoxFuture<'_, io::Result<()>> {
-        let _ = hash;
-        async move { Ok(()) }.boxed()
-    }
+    fn delete(&self, hash: &Hash) -> BoxFuture<'_, io::Result<()>>;
 }
 
 /// A trait for things that can track liveness of cids.
@@ -334,6 +315,14 @@ impl TempTag {
     /// The format of the pinned item
     pub fn format(&self) -> BlobFormat {
         self.cid.1
+    }
+
+    /// Keep the item alive until the end of the process
+    pub fn leak(mut self) {
+        // set the liveness tracker to None, so that the refcount is not decreased
+        // during drop. This means that the refcount will never reach 0 and the
+        // item will not be gced until the end of the process.
+        self.liveness = None;
     }
 }
 
--- a/iroh-bytes/src/baomap.rs
+++ b/iroh-bytes/src/baomap.rs
@@ -1,22 +1,17 @@
 //! Traits for in-memory or persistent maps of blob with bao encoded outboards.
-use std::{collections::BTreeSet, io, path::PathBuf};
+use std::{collections::BTreeSet, io, path::PathBuf, sync::Arc};
 
 use crate::{
     collection::CollectionParser,
     util::{
         progress::{IdGenerator, ProgressSender},
-        RpcError,
+        Cid, Format, RpcError,
     },
     Hash,
 };
-use anyhow::Context;
 use bao_tree::{blake3, ChunkNum};
 use bytes::Bytes;
-use futures::{
-    future::{self, BoxFuture, LocalBoxFuture},
-    stream::{self, LocalBoxStream},
-    Future, FutureExt, StreamExt,
-};
+use futures::{future::BoxFuture, stream::LocalBoxStream, FutureExt, StreamExt};
 use genawaiter::rc::{Co, Gen};
 use iroh_io::AsyncSliceReader;
 use range_collections::RangeSet2;
@@ -25,20 +20,6 @@ use tokio::sync::mpsc;
 
 pub use bao_tree;
 pub use range_collections;
-
-/// A format identifier
-///
-/// Should we make this an u64 and use https://github.com/multiformats/multicodec/blob/master/table.csv?
-///
-/// That table is so weird. There is so much unrelated stuff in there, so the smallest value we would be
-/// able to use for iroh collections would be 2 bytes varint encoded or something...
-#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
-pub enum Format {
-    /// Hash refers to a blob
-    Blob,
-    /// Hash refers to an iroh collection (format here?)
-    Collection,
-}
 
 /// The availability status of an entry in a store.
 #[derive(Debug, Clone, Eq, PartialEq)]
@@ -161,7 +142,11 @@ pub trait ReadableStore: Map {
     ///
     /// This function should not block to perform io. The knowledge about
     /// existing roots must be present in memory.
-    fn roots(&self) -> Box<dyn Iterator<Item = (Hash, Format)> + Send + Sync + 'static>;
+    fn roots(&self) -> Box<dyn Iterator<Item = (Bytes, Cid)> + Send + Sync + 'static>;
+
+    /// Temp pins
+    fn temp_pins(&self) -> Box<dyn Iterator<Item = Cid> + Send + Sync + 'static>;
+
     /// Validate the database
     fn validate(&self, tx: mpsc::Sender<ValidateProgress>) -> BoxFuture<'_, anyhow::Result<()>>;
 
@@ -200,13 +185,14 @@ pub trait Store: ReadableStore + PartialMap {
         &self,
         data: PathBuf,
         mode: ImportMode,
+        format: Format,
         progress: impl ProgressSender<Msg = ImportProgress> + IdGenerator,
-    ) -> BoxFuture<'_, io::Result<(Hash, u64)>>;
+    ) -> BoxFuture<'_, io::Result<(PinnedCid, u64)>>;
 
     /// This trait method imports data from memory.
     ///
     /// It is a special case of `import` that does not use the file system.
-    fn import_bytes(&self, bytes: Bytes) -> BoxFuture<'_, io::Result<Hash>>;
+    fn import_bytes(&self, bytes: Bytes, format: Format) -> BoxFuture<'_, io::Result<PinnedCid>>;
 
     /// Traverse all roots recursively and mark them as live.
     ///
@@ -214,17 +200,17 @@ pub trait Store: ReadableStore + PartialMap {
     ///
     /// Not polling this stream to completion is dangerous, since it might lead
     /// to some live data being missed.
-    /// 
+    ///
     /// The implementation of this method should do the minimum amount of work
     /// to determine the live set. Actual deletion of garbage should be done
     /// in the gc_sweep phase.
     fn gc_mark<'a>(
         &'a self,
         cp: impl CollectionParser + 'a,
-        extra_roots: impl Iterator<Item = io::Result<(Hash, Format)>> + 'a,
+        extra_roots: impl IntoIterator<Item = io::Result<Cid>> + 'a,
     ) -> LocalBoxStream<'a, GcMarkEvent> {
         Gen::new(|co| async move {
-            if let Err(e) = gc_task(self, cp, extra_roots, &co).await {
+            if let Err(e) = gc_mark_task(self, cp, extra_roots, &co).await {
                 co.yield_(GcMarkEvent::Error(e)).await;
             }
         })
@@ -245,8 +231,9 @@ pub trait Store: ReadableStore + PartialMap {
                 if !self.is_live(&hash) {
                     if let Err(e) = self.delete(&hash).await {
                         co.yield_(GcSweepEvent::Error(e.into())).await;
+                    } else {
+                        count += 1;
                     }
-                    count += 1;
                 }
             }
             co.yield_(GcSweepEvent::CustomInfo(format!("deleted {} blobs", count)))
@@ -258,14 +245,21 @@ pub trait Store: ReadableStore + PartialMap {
     /// Add the given hashes to the live set.
     ///
     /// This is used by the gc mark phase to mark roots as live.
-    fn add_live<'a>(&'a self, live: impl IntoIterator<Item = Hash> + 'a) {
+    fn add_live(&self, live: impl IntoIterator<Item = Hash>) {
         let _ = live;
+    }
+
+    /// Set a named pin
+    fn set_root(&self, name: &[u8], hash: Option<Cid>) -> BoxFuture<io::Result<()>> {
+        let _ = name;
+        let _ = hash;
+        async move { Ok(()) }.boxed()
     }
 
     /// True if the given hash is live.
     fn is_live(&self, hash: &Hash) -> bool {
         let _ = hash;
-        true
+        false
     }
 
     /// physically delete the given hash from the store.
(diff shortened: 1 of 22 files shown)


iroh/src/commands/provide.rs
```diff
@@ -108,6 +113,45 @@ pub async fn run(
...
+                match item {
+                    GcMarkEvent::CustomInfo(text) => {
+                        tracing::info!("{}", text);
+                    }
+                    GcMarkEvent::CustomWarning(text, _) => {
+                        tracing::warn!("{}", text);
+                    }
+                    GcMarkEvent::Error(err) => {
+                        tracing::error!("Fatal error during GC mark {}", err);
+                        continue 'outer;
```
ramfox, line 131: this immediately starts the GC over again, just making sure that's what we want to do on error & not sleep for 10 seconds before we try again.

iroh/src/commands/provide.rs
```diff
@@ -108,6 +113,45 @@ pub async fn run(
...
+                match item {
+                    GcSweepEvent::CustomInfo(text) => {
+                        tracing::info!("{}", text);
+                    }
+                    GcSweepEvent::CustomWarning(text, _) => {
+                        tracing::warn!("{}", text);
+                    }
+                    GcSweepEvent::Error(err) => {
+                        tracing::error!("Fatal error during GC mark {}", err);
+                        continue 'outer;
```
ramfox, line 147: same as above question

iroh/src/rpc_protocol.rs
```diff
@@ -162,29 +167,40 @@ impl ServerStreamingMsg<ProviderService> for ListIncompleteBlobsRequest {
...
 
-impl ServerStreamingMsg<ProviderService> for ListCollectionsRequest {
-    type Response = ListCollectionsResponse;
+impl ServerStreamingMsg<ProviderService> for ListRootsRequest {
+    type Response = ListRootsResponse;
+}
+
+/// List all collections
+///
+/// Lists all collections that have been explicitly added to the database.
```
ramfox, line 193: i know this is a WIP, but figured i'd just mention this comment hasn't been updated

iroh/src/commands/delete.rs
```diff
@@ -0,0 +1,29 @@
...
+    pub async fn run(self, client: RpcClient) -> Result<()> {
+        match self {
+            Commands::Roots { name } => {
+                let name = if name.starts_with("\"") && name.ends_with("\"") && name.len() > 1 {
+                    name[1..name.len() - 1].as_bytes().to_vec()
+                } else {
+                    hex::decode(name)?
+                }
+                .into();
+                let response = client.rpc(SetRootRequest { name, value: None }).await?;
```
ramfox, line 21: I'm assuming that when you set the value to `None`, that is how you mark it for deletion?

It's a bit confusing because I'm now associating the word "root" w/ what we keep, but it looks like it can also be associated with what we delete.

I think a better way of describing my confusion is, it's hard to tell what you are actually doing by sending this `SetRootRequest` unless you already have an understanding of the whole system. Might be solved with a comment, but might be work a re-name

iroh/src/commands/list.rs
```diff
@@ -13,8 +15,8 @@ pub enum Commands {
     Blobs,
     /// List the available blobs on the running provider.
     IncompleteBlobs,
-    /// List the available collections on the running provider.
-    Collections,
+    /// List the available roots on the running provider.
+    Roots,
```
ramfox, line 19: can roots be both `blobs` & `collections`?

ramfox: ahhh i see from other parts of the code that yes they can be. 

I understand why you mentioned that naming has been challenging for this PR. I'll have to mull this over.

iroh/src/rpc_protocol.rs
```diff
@@ -676,7 +692,8 @@ pub enum ProviderRequest {
     Version(VersionRequest),
     ListBlobs(ListBlobsRequest),
     ListIncompleteBlobs(ListIncompleteBlobsRequest),
-    ListCollections(ListCollectionsRequest),
+    ListCollections(ListRootsRequest),
```
ramfox, line 695: for consistency's sake, this should read `ListRoots(ListRootsRequest)`

iroh/src/rpc_protocol.rs
```diff
@@ -717,7 +734,7 @@ pub enum ProviderResponse {
     Version(VersionResponse),
     ListBlobs(ListBlobsResponse),
     ListIncompleteBlobs(ListIncompleteBlobsResponse),
-    ListCollections(ListCollectionsResponse),
+    ListCollections(ListRootsResponse),
```
ramfox, line 737: same as above `ListRoots(ListRootsResponse)`

iroh/src/commands/provide.rs
rklaehn: well, I was not quite sure what to do here. Exiting the loop would also be an option, but then you might lose this task and not notice. It should probably still do the delay.

iroh/src/commands/delete.rs
rklaehn: Yes, I did not want to have a full set of requests for a CRUD API, so I did this, a request that can both set and delete.

iroh/src/commands/list.rs
rklaehn: Yes, if we end up doing the ipfs format flag you could even have any multiformat here.

iroh/src/commands/delete.rs
rklaehn: We are going to rename roots to tags.

iroh/src/commands/provide.rs
```diff
@@ -108,6 +123,46 @@ pub async fn run(
         )
     };
 
+    let db2 = db.clone();
```
Frando, line 126: Why is the GC task running in the binary directly, and not within the node? I'd say let's move it to there, otherwise it will not run when starting a node embedded via Rust or FFI.
We can add configuration to the builder to enable GC, with something like this maybe?
```
NodeBuilder::with_gc(self, gc: GcConfig)
enum GcConfig {
    Disabled,
    Interval(Duration)
}
```

rklaehn: moved it

iroh-bytes/src/baomap.rs
```diff
@@ -140,7 +142,11 @@ pub trait ReadableStore: Map {
     ///
     /// This function should not block to perform io. The knowledge about
     /// existing roots must be present in memory.
```
dignifiedquire, line 144: still `roots`?

iroh-bytes/src/baomap.rs
```diff
@@ -140,7 +142,11 @@ pub trait ReadableStore: Map {
...
-    fn roots(&self) -> Box<dyn Iterator<Item = Hash> + Send + Sync + 'static>;
+    fn tags(&self) -> Box<dyn Iterator<Item = (Bytes, Cid)> + Send + Sync + 'static>;
+
+    /// Temp pins
```
dignifiedquire, line 147: can we call them temporary tags? I really dislike the pinning name, it is very overloaded, especially in rust

iroh-bytes/src/baomap.rs
```diff
@@ -179,13 +185,278 @@ pub trait Store: ReadableStore + PartialMap {
...
+    fn import_bytes(
+        &self,
+        bytes: Bytes,
+        format: BlobFormat,
+    ) -> BoxFuture<'_, io::Result<PinnedCid>>;
+
+    /// Set a named pin
+    fn set_tag(&self, name: Bytes, hash: Option<Cid>) -> BoxFuture<'_, io::Result<()>> {
+        let _ = name;
+        let _ = hash;
```
dignifiedquire, line 204: is this a place holder method, or why this odd assignment?

iroh-bytes/src/util.rs
```diff
@@ -13,10 +13,41 @@ pub mod io;
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+pub enum BlobFormat {
+    /// Hash refers to a blob
+    Raw,
+    /// Hash refers to an iroh collection (format here?)
+    Collection,
+}
+
+/// A hash and format pair
+pub type Cid = (Hash, BlobFormat);
```
dignifiedquire, line 31: can we not call these `cid`s? given that this is not the same as multiformats cids

iroh-bytes/src/util.rs
```diff
@@ -13,10 +13,41 @@ pub mod io;
...
+#[derive(PartialEq, Eq, Copy, Clone, Hash)]
 pub struct Hash(blake3::Hash);
 
+impl fmt::Debug for Hash {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("Hash").field(&DD(self.to_hex())).finish()
+    }
+}
+
+struct DD<T: fmt::Display>(T);
```
dignifiedquire, line 43: DD?

iroh-net/src/util.rs
```diff
@@ -10,7 +10,7 @@ use futures::FutureExt;
 
 /// A join handle that owns the task it is running, and aborts it when dropped.
 #[derive(Debug, derive_more::Deref)]
-pub(crate) struct AbortingJoinHandle<T>(tokio::task::JoinHandle<T>);
+pub struct AbortingJoinHandle<T>(pub tokio::task::JoinHandle<T>);
```
dignifiedquire, line 13: why make the field pub?

iroh/src/baomap/flat.rs
```diff
@@ -799,7 +938,94 @@ impl Store {
...
+        }
+        state.outboard.remove(&hash);
+        state.data.remove(&hash);
+        drop(state);
+        if let Some(data) = data {
+            if let Err(cause) = std::fs::remove_file(data) {
+                tracing::warn!("failed to delete data file: {}", cause);
+            }
+        }
+        if let Some(external) = external {
```
dignifiedquire, line 978: does this delete external files?  or is this just the outboard?

iroh/src/baomap/mem.rs
```diff
@@ -335,7 +337,11 @@ impl ReadableStore for Store {
         )
     }
 
-    fn roots(&self) -> Box<dyn Iterator<Item = Hash> + Send + Sync + 'static> {
+    fn tags(&self) -> Box<dyn Iterator<Item = (Bytes, Cid)> + Send + Sync + 'static> {
+        Box::new(std::iter::empty())
```
dignifiedquire, line 341: wouldn't it be useful to have this for the in memory version as well?

iroh/src/commands/delete.rs
```diff
@@ -0,0 +1,53 @@
+use anyhow::Result;
+use clap::Subcommand;
+use iroh::client::quic::Iroh;
+use iroh_bytes::Hash;
+
+#[derive(Subcommand, Debug, Clone)]
+pub enum Commands {
```
dignifiedquire, line 7: shouldn't the delete commands be under their respective existing commands? (or for tag a new one)

iroh/src/node.rs
```diff
@@ -389,6 +415,17 @@ where
             downloader,
         );
 
+        let gc_task = if let GcPolicy::Interval(gc_period) = self.gc_policy {
+            tracing::info!("Starting GC task with interval {}s", gc_period.as_secs());
+            let db = self.db.clone();
+            let cp = self.collection_parser.clone();
+            let task = rt
+                .local_pool()
```
dignifiedquire, line 423: is it a good idea for this to be on the local executor?

iroh-bytes/src/baomap.rs
rklaehn: Agree that temp pins does not make sense if the other things are not called pins. But temp tags also is not great, since these are unnamed.

iroh-bytes/src/baomap.rs
rklaehn: This is a method that does nothing. So there are two ways to make clippy happy. 1. have _name, _hash for the fn args, or do this. If you do the former then if you implement the fn it will use these names. I prefer to have the names in the signature not starting with _

iroh-bytes/src/baomap.rs
```diff
@@ -179,13 +185,278 @@ pub trait Store: ReadableStore + PartialMap {
...
+            }
+            co.yield_(GcSweepEvent::CustomInfo(format!("deleted {} blobs", count)))
+                .await;
+        })
+        .boxed_local()
+    }
+
+    /// Clear the live set.
+    fn clear_live(&self) {
+        let _ = self;
```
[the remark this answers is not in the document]
rklaehn: see above

iroh-bytes/src/util.rs
rklaehn: well, this is the question.

These are basically a blake3 cid (hash type and size is fixed, but format is variable). Depending on how we decide regarding the format, these might actually serialize as multiformats cids.

iroh-net/src/util.rs
rklaehn: Because I now use this from iroh to make sure the optional gc task is aborted on drop. This would be another one for the util crate.

iroh/src/baomap/flat.rs
rklaehn: There is a file that contains the paths for the external files. It deletes that file, not the external file itself. Deleting external files would be quite evil.

iroh/src/baomap/mem.rs
rklaehn: yeah, I did not touch the mem version at all so far.

iroh/src/commands/delete.rs
rklaehn: ?

We got the following hierarchy:

blob add
blob download
blob list <what to list>

this adds

blob delete <what to delete>

you want to add a new top level thing for tags, so

tag delete

?

iroh/src/node.rs
rklaehn: not sure. My rationale was that the only thing in any way expensive it does is io. For the mark phase the expensive thing is reading the collections into mem and parsing them, where I think the reading into mem is the expensive part.

For the sweep phase the expensive thing is to delete stuff, which depending on the file system might be very fast (modern file systems) or not (windows, sometimes).

I don't think any of this would benefit from being multithreaded in terms of performance. And as far as I understand it this does not mean that one thread of the local pool is permanently blocked, just that this task will be pinned to one thread of the local pool. But the thread will still be able to do other things, e.g. while this thing is waiting for the next gc.

iroh-bytes/src/util.rs
rklaehn: Something like this:

```
enum BlobFormat {
  Raw,
  Collection,
  Multiformat(u64),
}

impl BlobFormat {
  fn to_multiformat(&self) -> u64 {
    ...
  }
  fn from_multiformat(value: u64) -> Self {
    ...
  }
}

struct Blake3Cid(Hash, BlobFormat);

impl Blake3Cid {
  pub fn as_cid_bytes(&self) -> Vec<u8> {
    ...
  }
  pub fn from_cid_bytes(bytes: &[u8]) -> Option<Self> {
    ...
  }
}
```

iroh-bytes/src/baomap.rs
rklaehn: renamed it to temp tags for now, but I am not 100% happy with the name

rklaehn: I guess the exact name for the thing depends on the outcome of the naming discussion below.

What you get is basically a (hash, format) pair that keeps the corresponding data safe from gc until it gets dropped. It is not just a tag, but also contains the hash and format, so it can be used interchangeably with an actual (hash, format) pair.

iroh/src/node.rs
rklaehn: I think there are some extreme cases where io can actually benefit from being multithreaded. But typically if you perform multithreaded io you are just making things worse. So unless having a long running task in a pinned thread pool is in some way evil (I would not know why), I think this is the way to go.

iroh-bytes/src/util.rs
rklaehn: I was planning to replace this tuple with a proper struct called `Blake3Cid` or something that has the various ser/de methods.

iroh/src/commands/delete.rs
dignifiedquire: yeah, that was my thinking

iroh-bytes/src/util.rs
dignifiedquire: happy to keep it, but a comment what the point of it is would be nice

iroh-bytes/src/util.rs
rklaehn: made a tuple struct now. I think we can confine all cid craziness to a helper struct. The only nod to multiformats is the values we use in format.

iroh/src/commands/delete.rs
b5: this one is tough. IMHO the question of weather tags should be their own thing or not boils down to **how tags & documents relate.**. Do documents have a tag? It could be either implicit (the docs system won't let you delete stuff docs need, aka "implict tag") or explicit (every document gets a tag when you join). If all documents have a tag, then I think tags should be a top-level concept.

But if that's true,  I think we should take a step back, and consider renaming the concept of `tags` to `names` and just turning this into a petname system. We need a way to name stuff, and stuff we name expresses user intent about cache retention. So let's lean into that and have one concept that says "stuff you name is stuff you keep". If I have two chickens, one named Wilber, and the other one doesn't have a name, you can guess which chicken I would "evict from cache" first.

rklaehn: so, I am pretty sure you don't want to pin every single document you have ever interacted with in any way. why should documents get an "everything I contain is important" excemption compared to collections?

So yes, I would think that docs should also be taggable, with the default behaviour for an untagged doc being just like an untagged collection.

> I have two chickens, one named Wilber, and the other one doesn't have a name, you can guess which chicken I would "evict from cache" first.

🐔 🍗 🤣 

I am OK with calling tags names, provided that names are still just bytes. We can make it a convention that you usually use utf-8 strings, but I really don't want to enforce that.

Another problem with names is that people will immediately think about something like IPNS...

rklaehn:
Added auto tag creation and made a newtype wrapper for tags.

Unfortunately clippy thinks that bytes::Bytes are mutable...

iroh/src/commands/delete.rs
b5: > I am OK with calling tags names, provided that names are still just bytes.

Yep. Super into that convention, which I think we should use everywhere (collection keys if we do them, document keys, and now tag/name names)

> Another problem with names is that people will immediately think about something like IPNS...

I wouldn't say "everyone", but I think you're right that the word `name` generally evokes the notion of a universal name. Humans generally have 1 "name". **I change my vote back to "tags"**

rklaehn: So should we move tags to the top level of the cmd hierarchy in anticipation of them being relevant for docs in the future?

iroh-bytes/src/util.rs
```diff
@@ -1,22 +1,158 @@
...
 pub mod runtime;
 
+/// A format identifier
+///
+/// Should we make this an u64 and use ?
+///
+/// That table is so weird. There is so much unrelated stuff in there, so the smallest value we would be
+/// able to use for iroh collections would be 2 bytes varint encoded or something...
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct BlobFormat(u64);
```
dignifiedquire, line 25: I would strongly prefer to ignore multiformats entirely and use a `u8`, with `u8::MAX` being reserved for future extensions.

rklaehn: Usually I agree. But in this case I am not so sure. This costs us almost nothing. We use 0x55 instead of 0 for "raw blob" and whatever other value we choose for collections instead of 1. Other than that this costs us nothing.

We are using postcard, so it's not like we are totally allergic to varints. And for cases where speed of serialization matters, BlobFormat serializes as just an u64.

In the doc db, we just got a 40 byte (hash, format) pair instead of a 33 byte pair, but then rounding up to word boundaries is probably a good idea anyway.

We could do some kind of permutation table thing where we replace 0x55 <-> 0 and the multiformats collection format id with 1, but what for?

And if we do an u8 that means we won't be able to ever represent all possible multiformats values, but have to have an explicit list of supported formats and mappings to our enum.

dignifiedquire (approved):
Haven't reviewed for structural correctness of the GC, but general integration, naming and setup look good to me now. 

More tests are certainly needed in a followup 😅

PR #1503, opened 2023-09-20
fix: print enum variant name for RPC debug logs

Frando:
## Description

Right now in the debug logs we print the following for each incoming RPC request:
```
handling rpc request: quic_rpc::transport::flume::FlumeServerEndpoint<iroh::rpc_protocol::ProviderRequest, iroh::rpc_protocol::ProviderResponse>
```
which is very long and not useful because it contains no info at all about incoming request.

Before that, we printed the full Debug impl, which contained too much (especially bytes).

This PR changes it to just print the enum variant name, so e.g.

```
handling rpc request: BlobRead
```

Also removes a debug leftover `println`.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-sync/src/sync.rs
+++ b/iroh-sync/src/sync.rs
@@ -98,7 +98,6 @@ impl<S: ranger::Store<SignedEntry> + PublicKeyStore + 'static> Replica<S> {
     /// the receiver to be received from.
     // TODO: Allow to clear a previous subscription?
     pub fn subscribe(&self) -> Option<flume::Receiver<(InsertOrigin, SignedEntry)>> {
-        println!("subscribing..");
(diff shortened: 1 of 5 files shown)


iroh-sync/src/sync.rs
```diff
@@ -98,7 +98,6 @@ impl<S: ranger::Store<SignedEntry> + PublicKeyStore + 'static> Replica<S> {
     /// the receiver to be received from.
     // TODO: Allow to clear a previous subscription?
     pub fn subscribe(&self) -> Option<flume::Receiver<(InsertOrigin, SignedEntry)>> {
-        println!("subscribing..");
```
b5, line 101: thank you!

dignifiedquire (approved):

PR #1504, opened 2023-09-20
feat: doc import & export commands

b5:
## Description

* Rough outline porting over the "fs" commands from the repl example into the main CLI & console.
* This ends up requiring a set_hash command on the core side, which I think is great.
* I'm also trying to keep all the import / export logic in the iroh crate. don't think any of this belongs in core

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-bytes/src/util.rs
+++ b/iroh-bytes/src/util.rs
@@ -304,6 +304,17 @@ impl<'de> Deserialize<'de> for Hash {
     }
 }
 
+/// A convenience struct that contains metadata about a blob
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct BlobInfo {
+    /// name of the blob
+    pub name: String,
+    /// size of the blob
+    pub size: u64,
+    /// hash of the blob
+    pub hash: Hash,
+}
+
 struct HashVisitor;
 
 impl<'de> de::Visitor<'de> for HashVisitor {
--- a/iroh/src/commands/sync.rs
+++ b/iroh/src/commands/sync.rs
@@ -285,37 +285,33 @@ impl DocCommands {
                 }
                 let root = canonicalize_path(&path)?.canonicalize()?;
                 let tag = tag_from_file_name(&root)?;
-                let files = walkdir::WalkDir::new(&root).into_iter();
-                // TODO: parallelize
                 let mut counts = (0, 0);
-                for file in files {
-                    let file = file?;
-                    if file.file_type().is_file() {
-                        let relative = file.path().strip_prefix(&root)?.to_string_lossy();
-                        if relative.is_empty() {
-                            print!("invalid file path: {:?}", file.path());
-                            continue;
-                        }
-                        let key = if prefix.is_empty() {
-                            relative.to_string()
-                        } else {
-                            format!("{prefix}/{relative}")
-                        };
-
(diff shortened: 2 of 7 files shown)


iroh-bytes/src/util.rs
```diff
@@ -304,6 +304,17 @@ impl<'de> Deserialize<'de> for Hash {
     }
 }
 
+/// A convenience struct that contains metadata about a blob
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct BlobInfo {
```
ramfox, line 309: possibly unnecessary

iroh/src/commands/sync.rs
```diff
@@ -223,6 +269,74 @@ impl DocCommands {
...
+            } => {
+                let doc = get_doc(iroh, env, doc).await?;
+                let author = env.author(author)?;
+                let mut prefix = prefix.unwrap_or_else(|| String::from(""));
+
+                if prefix.ends_with('/') {
+                    prefix.pop();
+                }
+                let root = canonicalize_path(&path)?.canonicalize()?;
+                let tag = tag_from_file_name(&root)?;
```
ramfox, line 287: We may want this to be `SetTagOption::Auto`, but otherwise the dir name (or file name if the path is to a file) is the tag

iroh/src/commands/sync.rs
```diff
@@ -223,6 +269,74 @@ impl DocCommands {
...
+                    counts.1 += entry.size;
+                    // adjust the key so that it does not leak the entire directory structure of
+                    // the importer's machine
+                    let key: Vec<u8> = PathBuf::from(entry.name.clone())
+                        .strip_prefix(root_prefix.clone())?
+                        .to_str()
+                        .map(|p| p.as_bytes())
+                        .ok_or(anyhow!("could not convert path to bytes"))?
+                        .into();
+                    doc.set_hash(author, key, entry).await?;
```
ramfox, line 314: sending the `entry` here, which contains the `Hash` and the size of the blob. This means we do not have to load the entry from the `bao_store` to discover the size. However, this also mean that you can "spoof" the document if you wanted to, and add a size that does not actually relate to the content from the hash.

iroh/src/commands/sync.rs
```diff
@@ -223,6 +269,74 @@ impl DocCommands {
...
+                }
+                println!(
+                    "Imported {} entries totaling {}",
+                    counts.0,
+                    HumanBytes(counts.1)
+                );
+            }
+            Self::Export { doc, key, path } => {
```
ramfox, line 322: Okay, seems like we can have more flexible functionality with export, if we wanted:

- add `prefix` flag, this would allow us to "round trip" import:
```
$ iroh docs import --path ./DIR_NAME
$ iroh docs export --prefix DIR_NAME --path ./NEW_DIR
```

- add `author` flag / read author from `IROH_AUTHOR`, allows you to get entries from one specific author

Do we want these?

b5: meh, I find the `--prefix` on export confusing, because it overlaps with the provided path. Instead I'd like the export command to automatically create directories to satisfy the provided `--path` argument.

Let's skip the `author` flag until someone asks for it

ramfox:
closing in favor of https://github.com/n0-computer/iroh/pull/1563

PR #1505, opened 2023-09-20
fix: avoid double conns, better state tracking

Frando:
## Description

On `main` the test `sync_full_basic` is flakey. This PR (hopefully) fixes it.

The reason was: We have the situation that two peers initiate connections to each other at (roughly) the same time. In #1491 this was sometimes prevented, but not reliably.

This PR fixes it by:
* When connecting, we set a `SyncState::Dialing` for the `(namespace, peer)`
* When accepting, if our own state is `Dialing` for the incoming request for `(namespace, peer)` we compare our peer id with that of the incoming request, and abort if ours is higher (doesn't matter which way, we care about a predictable outcome only
* Through this, only one of the two simoultanoues connections will survive
* Also added a `Abort` frame to the wire protocol to transfer to inform the dialer about the reason of the declined request, which is either "we do double sync, and will take the other conn" (`AlreadySyncing`) or "this replica is not here" (`NotAvailable`)

This PR also:
* Further improves logging
* Improves errors

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-sync/src/net.rs
+++ b/iroh-sync/src/net.rs
@@ -1,9 +1,9 @@
 //! Network implementation of the iroh-sync protocol
 
-use std::{future::Future, net::SocketAddr};
+use std::{future::Future, net::SocketAddr, result::Result};
 
-use anyhow::{Context, Result};
 use iroh_net::{key::PublicKey, magic_endpoint::get_peer_id, MagicEndpoint};
+use serde::{Deserialize, Serialize};
 use tracing::debug;
 
 use crate::{
@@ -30,18 +30,23 @@ pub async fn connect_and_sync<S: store::Store>(
     peer: PublicKey,
     derp_region: Option<u16>,
     addrs: &[SocketAddr],
-) -> Result<()> {
-    debug!(peer = ?peer, "sync (via connect): start");
+) -> Result<(), ConnectError> {
+    debug!(?peer, "sync[dial]: connect");
     let namespace = doc.namespace();
     let connection = endpoint
         .connect(peer, SYNC_ALPN, derp_region, addrs)
         .await
-        .context("failed to establish connection")?;
-    debug!(?peer, ?namespace, "sync (via connect): connected");
-    let (mut send_stream, mut recv_stream) = connection.open_bi().await?;
+        .map_err(ConnectError::connect)?;
+    debug!(?peer, ?namespace, "sync[dial]: connected");
+    let (mut send_stream, mut recv_stream) =
+        connection.open_bi().await.map_err(ConnectError::connect)?;
     let res = run_alice::<S, _, _>(&mut send_stream, &mut recv_stream, doc, peer).await;
-    send_stream.finish().await?;
-    recv_stream.read_to_end(0).await?;
+
+    send_stream.finish().await.map_err(ConnectError::close)?;
+    recv_stream
+        .read_to_end(0)
+        .await
+        .map_err(ConnectError::close)?;
 
     #[cfg(feature = "metrics")]
     if res.is_ok() {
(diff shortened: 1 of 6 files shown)


iroh-sync/src/net/codec.rs
```diff
@@ -1,6 +1,6 @@
-use std::future::Future;
+use std::{future::Future, result::Result};
```
divagant-martian, line 1: same here

iroh/src/sync_engine/live.rs
```diff
@@ -389,11 +396,20 @@ struct Actor<S: store::Store, B: baomap::Store> {
...
-    running_sync_connect:
-        FuturesUnordered<BoxFuture<'static, (NamespaceId, PublicKey, SyncReason, Result<()>)>>,
+    running_sync_connect: FuturesUnordered<
+        BoxFuture<
+            'static,
+            (
+                NamespaceId,
+                PublicKey,
+                SyncReason,
+                std::result::Result<(), ConnectError>,
```
divagant-martian, line 406: also here is not necessary

iroh/src/sync_engine/live.rs
```diff
@@ -389,11 +396,20 @@ struct Actor<S: store::Store, B: baomap::Store> {
...
+            ),
+        >,
+    >,
     /// Running sync futures (from accept).
     running_sync_accept: FuturesUnordered<
-        BoxFuture<'static, std::result::Result<(NamespaceId, PublicKey), SyncError>>,
+        BoxFuture<'static, std::result::Result<(NamespaceId, PublicKey), AcceptError>>,
```
divagant-martian, line 412: same here

iroh/src/sync_engine/live.rs
```diff
@@ -508,9 +524,9 @@ impl<S: store::Store, B: baomap::Store> Actor<S, B> {
                 }
                 Some((namespace, res)) = self.pending_joins.next() => {
                     if let Err(err) = res {
-                        error!("failed to join gossip for {namespace:?}: {err:?}");
+                        error!(?namespace, ?err, "failed to join gossip");
```
divagant-martian, line 527: maybe we want `%err` ?

iroh/src/sync_engine/live.rs
```diff
@@ -720,21 +755,54 @@ impl<S: store::Store, B: baomap::Store> Actor<S, B> {
         namespace: NamespaceId,
         peer: PublicKey,
         reason: SyncReason,
-        result: Result<()>,
+        result: std::result::Result<(), ConnectError>,
```
divagant-martian, line 758: also here

iroh/src/sync_engine/live.rs
```diff
@@ -720,21 +755,54 @@ impl<S: store::Store, B: baomap::Store> Actor<S, B> {
...
+                self.on_sync_finished(namespace, peer, Origin::Connect(reason), Ok(()))
+                    .await;
+            }
+        }
     }
 
     async fn on_sync_via_accept_finished(
         &mut self,
-        res: std::result::Result<(NamespaceId, PublicKey), SyncError>,
+        res: std::result::Result<(NamespaceId, PublicKey), AcceptError>,
```
divagant-martian, line 791: here as well

divagant-martian (approved):
the only things that jump to me is doing explicit imports of `std::result::Result` if you are using `anyhow::Result`, setting the error will overwrite the generic so that's not necessary (if that's the reason)

iroh-sync/src/net.rs
```diff
@@ -1,9 +1,9 @@
 //! Network implementation of the iroh-sync protocol
 
-use std::{future::Future, net::SocketAddr};
+use std::{future::Future, net::SocketAddr, result::Result};
```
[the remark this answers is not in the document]
Frando: no reason, editor did it, will remove

dignifiedquire (approved):

iroh/src/sync_engine/live.rs
divagant-martian: removed in 884fe52

iroh/src/sync_engine/live.rs
divagant-martian: removed in 884fe52

iroh/src/sync_engine/live.rs
divagant-martian: removed in 884fe52

iroh-sync/src/net/codec.rs
divagant-martian: removed in 884fe52

iroh-sync/src/net.rs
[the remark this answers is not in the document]
divagant-martian: removed in 884fe52

PR #1506, opened 2023-09-21
refactor(*):  rework `NodeAddr`

divagant-martian:
## Description

- Adds a `AddrInfo` that contains the addressing information of peers. This is necessary as it's own type since it's serialized both in gossip and in #1488 
- Renames `NodeAddr` to `PeerData` (better name suggestions are well received) this helps consolidate the "peer" terminology already present throughout gossip, sync, downloader, and many other places
- Use this type in different parts of the code. Some other types that can be replaced by this are left to #1493 

## Notes & open questions

This applies the suggestion of doing `connect` using the full type. It's debatable if this is better than `connect(PublicKey, AddrInfo)` but both are imo an improvement over the current way since the idea if "what does iroh need to connect to a peer" is now fully described in a type

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/src/magic_endpoint.rs
+++ b/iroh-net/src/magic_endpoint.rs
@@ -20,17 +20,17 @@ pub use super::magicsock::EndpointInfo as ConnectionInfo;
 
 /// A peer and it's addressing information.
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
-pub struct PeerData {
+pub struct PeerAddr {
     /// The node's public key.
     pub peer_id: PublicKey,
     /// Addressing information to connect to [`Self::peer_id`].
     pub addr_info: AddrInfo,
 }
 
-impl PeerData {
-    /// Create a new [`PeerData`] with empty [`AddrInfo`].
+impl PeerAddr {
+    /// Create a new [`PeerAddr`] with empty [`AddrInfo`].
     pub fn new(peer_id: PublicKey) -> Self {
-        PeerData {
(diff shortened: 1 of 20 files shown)


iroh-net/src/magic_endpoint.rs
```diff
@@ -18,28 +18,92 @@ use crate::{
 
 pub use super::magicsock::EndpointInfo as ConnectionInfo;
 
-/// Address information for a node.
+/// A peer and it's addressing information.
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
-pub struct NodeAddr {
+pub struct PeerData {
```
Frando, line 23: Why not `PeerAddr`? To me having `addr`  in there suggests "this is what you need to connect", which is what it is.
So if we decided for `Peer`, not `Node`, then I'd say let's do `PeerAddr`.
Returning `PeerAddr` from `endpoint.my_addr()` sounds good too.

Frando:
LGTM! Would change the naming of the combined type to include `Addr`.
Otherwise this does everything that #1493 did, right? Only difference I've seen is that #1493 also replaced the `PeerSource` type in `iroh/sync_engine/live.rs`. If you'd add this over here, we can just close #1493, wdyt?

Frando (approved):
Approving, with the suggestion to change the naming to `PeerAddr` or similar.
Finer details or further can be done in followups but let's get the actual change in to have a combined type we use in the interface :+1:

PR #1514, opened 2023-09-22
tests: big sync test (wip)

Frando:
## Description

This adds a sync test with more nodes. It passes with 3 nodes (the default). Set `NODES=4` or higher as env variable and it will fail.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


Frando:
Closing, see #1530

PR #1516, opened 2023-09-23
Collection = Vec<Hash>

rklaehn:
## Description

As discussed on Friday, this turns collections into just sequences of hashes

## Notes & open questions

~Just wanted to check how much effort this is. Not ready to merge. But the effort seems very manageable...~

How should the two concepts be named.

1. a collection as in a sequence of links
2. a thing that contains names and can therefore be used to map a directory tree

Todo:

- [x] add total size to metadata
- [x] remove IrohCollectionParser from iroh, add LinkSeqCollectionParser to iroh-bytes, and make it the default
- [x] add more convenient APIs to work with 2 initial blobs
- [x] skip size prefix for links

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/collection.rs
+++ b/iroh/src/collection.rs
@@ -21,6 +21,19 @@ pub struct Collection {
 }
 
 impl Collection {
+    ///
+    pub fn from_parts(links: &[Hash], names: &[String]) -> anyhow::Result<Self> {
+        let blobs = links
+            .iter()
+            .zip(names.iter())
+            .map(|(hash, name)| Blob {
+                name: name.clone(),
+                hash: *hash,
+            })
+            .collect();
+        Self::new(blobs, 0)
+    }
+
     /// Create a new collection from a list of blobs and total size of the raw data
     pub fn new(blobs: Vec<Blob>, total_blobs_size: u64) -> anyhow::Result<Self> {
         let mut blobs = blobs;
@@ -37,6 +50,16 @@ impl Collection {
     /// Serialize this collection to a std `Vec<u8>`
     pub fn to_bytes(&self) -> Result<Vec<u8>> {
         Ok(postcard::to_stdvec(self)?)
+    }
+
+    /// Get the links to the blobs in this collection
+    pub fn links(&self) -> Vec<Hash> {
+        self.blobs.iter().map(|x| x.hash).collect()
+    }
+
+    /// Get the names of the blobs in this collection
+    pub fn names(&self) -> Vec<String> {
+        self.blobs.iter().map(|x| x.name.clone()).collect()
     }
 
     /// Deserialize a collection from a byte slice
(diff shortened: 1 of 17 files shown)


iroh/src/collection.rs
```diff
@@ -159,19 +182,9 @@ impl CollectionParser for IrohCollectionParser {
...
-            let hashes = collection
-                .into_inner()
-                .into_iter()
-                .map(|x| x.hash)
-                .collect::<Vec<_>>()
-                .into_boxed_slice();
+            let hashes = postcard::from_bytes::<Box<[Hash]>>(&data)?;
             let res: Box<dyn LinkStream> = Box::new(ArrayLinkStream { hashes, offset: 0 });
-            Ok((res, stats))
+            Ok((res, CollectionStats::default()))
```
rklaehn, line 187: todo: the total size needs to be part of the metadata

rklaehn:
This was easier than expected. The only big downside so far is that the total_blobs_size is in the metadata and therefore no longer available when parsing just the collection (=seq of links) blob.

However, I think using this for bookkeeping is probably unwise anyway.

rklaehn:
There is a naming issue here. A collection as far as iroh-bytes is concerned is just a sequence of links. A collection in iroh is a thing that you can map to a fs (with names and stuff). Before these were kinda the same, but now they aren't.

So I guess one of them need a new name, and that would be the "sequence of links". This will fade into the background or might be hardcoded in the future.

rklaehn:
@Arqu can you comment on this: due to the metadata now being in the first leaf, the TransferCollectionStarted event now no longer contains the total_blob_size. I think this is no big deal, since for bookkeeping and billing you probably want to use something else anyway.

But am I right, or does this has a big impact on iroh-net?

Arqu:
> @Arqu can you comment on this: due to the metadata now being in the first leaf, the TransferCollectionStarted event now no longer contains the total_blob_size. I think this is no big deal, since for bookkeeping and billing you probably want to use something else anyway.
> 
> But am I right, or does this has a big impact on iroh.network?

Good eye for tagging me on this. You're very lucky actually :). The work on iroh.network is currently exactly around this spot. Yes it's going to be moved around and potentially have a completely different implementation so don't worry about it in this PR.
@flub Just tagging you here for reference, we can touch on this in our meeting.

iroh/tests/cli.rs
```diff
@@ -244,7 +244,7 @@ fn cli_provide_tree_resume() -> Result<()> {
         let get_output = get.unchecked().run()?;
         assert!(get_output.status.success());
         let matches = explicit_matches(match_get_stderr(get_output.stderr)?);
-        assert_eq!(matches, vec!["112.84 KiB"]);
+        assert_eq!(matches, vec!["112.88 KiB"]);
```
dignifiedquire, line 247: how come this is larger in total now?

dignifiedquire (approved):

iroh/tests/cli.rs
rklaehn: You got an additional indirection. Instead of having n links, you have n+1 links. That gives you an extra 32 bytes, or 0.032 kilobytes. I guess it gets rounded up...

Issue #1498, opened 2023-09-20
Prettify blob and collection related iroh console commands


rklaehn:
For some commands you get nice progress bars like in the console:

Here is blob add:
![image](https://github.com/n0-computer/iroh/assets/248257/eea5725d-0bad-471a-bdb4-5e872f4eb811)

For other commands you get the raw output of the progress stream:

Here is blob share:
![image](https://github.com/n0-computer/iroh/assets/248257/5f2b0ddf-8526-43aa-90ac-55163fa8d5b3)

We should be able to reuse the nice progress stuff from the cli commands...

rklaehn:
I think this is done with the stats pr merge. Reopen if I am wrong.

PR #1526, opened 2023-09-26
Console download progress

rklaehn:
## Description

This brings the console download GUI in line with iroh get. You get a nice progress bar instead of just an event dump.

In addition I fixed(?) a few other things. The different download stages 1/3 connecting, 2/3 requesting 3/3 downloading are now only shown transiently as the progress bar message. The only thing that is permanent is "requesting <hash>" and the (successful or unsuccessful) outcome.

To make this possible I had to change the test regexes to only look for the permanent output, which I think is not bad. We really don't want to test the transient outputs in all the tests. Maybe there should be one test that does that that is specifically made to test the "console UI".

This is one thing I found that is related to https://github.com/n0-computer/iroh/issues/1498

## Notes & open questions

For debugging the event dump is not that bad. Maybe we need a mode where you always get the event dump?

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/commands.rs
+++ b/iroh/src/commands.rs
@@ -685,6 +685,7 @@ const PROGRESS_STYLE: &str =
 
 fn make_download_pb() -> ProgressBar {
     let pb = ProgressBar::hidden();
+    pb.set_draw_target(ProgressDrawTarget::stderr());
     pb.enable_steady_tick(std::time::Duration::from_millis(50));
     pb.set_style(
         ProgressStyle::with_template(PROGRESS_STYLE)
@@ -709,7 +710,6 @@ fn init_download_progress(pb: &ProgressBar, count: u64, missing_bytes: u64) -> R
     ));
     pb.set_length(missing_bytes);
     pb.reset();
-    pb.set_draw_target(ProgressDrawTarget::stderr());
 
     Ok(())
 }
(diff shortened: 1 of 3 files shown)


rklaehn:
@Arqu I changed the command output. Before it was like this:

image

now it is like this:

image

the difference is that the intermediate states [1/3] Connecting and [2/3] Requesting are no longer permanent, but get erased by the progress bar gui lib.

I think this is better, but it interferes with the screen scraping done by netsim.

Arqu:
> @Arqu I changed the command output. Before it was like this:
> 
> image
> 
> now it is like this:
> 
> image
> 
> the difference is that the intermediate states [1/3] Connecting and [2/3] Requesting are no longer permanent, but get erased by the progress bar gui lib.
> 
> I think this is better, but it interferes with the screen scraping done by netsim.

hmmm that part should not be relevant. I'll take a look right away. I'll set up netsim on a branch and we can test against it.

rklaehn:
Maybe it is a real error, but the netsim error is this:

```
File "/root/actions-runner/_work/iroh/chuck/netsim/reports_csv.py", line [18](https://github.com/n0-computer/iroh/actions/runs/6310635094/job/17133000070?pr=1526#step:12:19)3, in <module>
    reported_throughput = json_d['sum']['reported_mbits']
```

which is kinda what you would expect if it was trying to parse the output but fails.

Arqu:
It's fixed.

rklaehn:
> It's fixed. The console output is not detected in logs, guess I need to redirect something else to the logs too eventually. But generally just updated it to be less brittle.

thank you!

flub (approved):

Issue #1528, opened 2023-09-26
 Audit usage of unwraps


Frando:
`thread 'sync_big' panicked at 'called Result::unwrap() on an Err value: ()', iroh/iroh-net/src/magicsock.rs:1175:28`. This happened at the end of the test, after calling `node.shutdown()`

divagant-martian:
I'm splitting this as they are unrelated and closed by different solutions https://github.com/n0-computer/iroh/issues/1531

flub:
We should still audit other places for this.  But it's painfully manual and error-prone.

PR #1529, opened 2023-09-26
fix(iroh-net): Do not unwrap sending on response channel

flub:
## Description

This response is not critical.  If someone dropped the receiver it's
just fine, it meant they didn't care about the response anymore.

This has occurred during shutdown.

Part of #1528

## Change checklist

- [x] Self-review.
- [x] ~~Tests if relevant.~~

### DIFF
--- a/iroh-net/src/magicsock.rs
+++ b/iroh-net/src/magicsock.rs
@@ -1172,7 +1172,7 @@ impl Actor {
             }
             ActorMessage::AddKnownAddr(addr, s) => {
                 self.add_known_addr(addr);
-                s.send(()).unwrap();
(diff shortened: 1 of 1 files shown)


divagant-martian:
just to note this change should be done everywhere a channel sends a response. Is this the only instance?

flub:
> just to note this change should be done everywhere a channel sends a response. Is this the only instance?

Agreed, and it probably isn't the only place.  The referenced issue does consider the wider picture, this PR fixes only a known failing part.

dignifiedquire (approved):
lets merge this, and add an issue to review channel sends

PR #1535, opened 2023-09-27
feat(sync): Implement prefix deletion

Frando:
## Description

This PR implements prefix deletion as specified in [Willow](https://willowprotocol.org/specs/data-model)

## Notes & open questions

There's an important implication of prefix deletion in our willow data model that we'll have to be very explicit about in docs and tutorials:
```rust
doc.set_bytes(author, b"foobar", b"important data").await?;
doc.set_bytes(author, b"foo", b"crap").await?;
let x = doc.get(author, b"foobar").await?;
// x == None
```
In explanation: The willow data model defines that adding an entry removes all *older entries that have the new entry's key as their prefix*.

What this means is that apps need to model their keyspace carefully. For example, the current file system import/export demo (#1504) will not work as intended, because of the above. 

The recommended solution is that for usecases that need nested keys without them being affected by prefix deletion is to append a nul byte to those keys.

Quoting @AljoschaMeyer from Discord:
> In defense of prefix-based deletion and its useability: the approach was inspired by regular file-system semantics. If I create `/food/pizza.jpg` and `/food/croissant.jpg`, but later created an empty text file /food, it would delete the previously existing /food directory, and hence everything in it.
> Prefix-based deletion was the simplemost solution we could find to replicate this behavior. In particular, we do not need to introduce the concepts of nested directories and leaf files (or the formal counterpart to file system paths: strings of strings).
> The idea was that anyone who'd create an application layer where users deal with actual filesystems would add a small conversion layer: the user-facing api has (nested) directories, (optionally) containing leaf files. These are converted to willow paths in a way that adds a null-byte at the end of every leaf file (and prefixes every directory with a path separator such as /). Willow then replicates exactly the deletion behavior that the users expect from a file system. 

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


dignifiedquire:
One thing we should make sure to either unify or clarify is the difference (or not) between `delete_prefix("foo")` and `set("foo", "")`

Frando:
> One thing we should make sure to either unify or clarify is the difference (or not) between `delete_prefix("foo")` and `set("foo", "")`
Yup. They are exactly the same with the current implementation. So should we just note this in the docs?

dignifiedquire:
I think having the same behaviour is the right call, just need to makes sure it is tested & explained in the docs.

Frando:
I pushed a commit that keeps the behavior, but makes the store return `None` for entries with the empty hash (and skip them in the iterators).

iroh-sync/src/store/fs.rs
```diff
@@ -513,6 +518,68 @@ impl crate::ranger::Store<SignedEntry> for StoreInstance {
         let iter2 = RangeIterator::empty(&self.store.db)?;
         Ok(iter.chain(iter2))
     }
+
+    // TODO: Not horrible
```
dignifiedquire, line 522: this should be fixed before merging

iroh-sync/src/sync.rs
```diff
@@ -733,16 +798,54 @@ pub struct Record {
...
+    }
+
+    /// Create a tombstone record with the timestamp set to now.
+    pub fn empty_current() -> Self {
+        Self::new_current(Hash::EMPTY, 0)
+    }
+
+    /// Return `true` if the entry is empty.
+    pub fn is_empty(&self) -> bool {
+        self.len == 0 && self.hash == Hash::EMPTY
```
dignifiedquire, line 846: we use || in other places, maybe here too?

iroh-sync/src/sync.rs
```diff
@@ -1297,6 +1395,64 @@ mod tests {
...
+
+        // delete
+        let deleted = replica.wipe_at_prefix(b"foo", &alice)?;
+        assert_eq!(deleted, 2);
+        assert_eq!(store.get_one(myspace.id(), alice.id(), b"foobar")?, None);
+        assert_eq!(store.get_one(myspace.id(), alice.id(), b"fooboo")?, None);
+        assert_eq!(store.get_one(myspace.id(), alice.id(), b"foo")?, None);
+
+        Ok(())
+    }
```
dignifiedquire, line 1454: we will need some more extenstive tests here, especially for the db part, making sure we don’t delete the wrong things

dignifiedquire (commented):
looks good overall, but we also need a test syncing nodes checking deletions match up

iroh-sync/src/store/fs.rs
Frando: Do you have an idea how to solve this efficiently without moving the complete database structure to a tree based one?

dignifiedquire: I meant avoiding the allocations for now, not the algorithm itself, that will need more structural changes from the work @rklaehn will do

iroh-sync/src/sync.rs
dignifiedquire: we should probably do a consistency check at some point, that if the hash is `EMPTY` the length is set to `0`

Frando: This is mostly covered already:
* `replica.insert()` with `hash::EMPTY` or `len: 0` returns an error so that you're forced to use `wipe_at_prefix` That is debatable - should we allow it instead?
* `wipe_at_prefix` sets `len: 0` and `hash::EMPTY`
* leaves `insert_remote_entry` for gossiped entries - I'll add a check there and return an error for entries with hash::Empty and not len 0.

Frando: Added the validation on `insert_remote_entry`

iroh-sync/src/sync.rs
Frando: Do you have some specific test cases in mind here?

iroh-sync/src/store/fs.rs
Frando: I removed the allocations.

Frando:
Changes since @dignifiedquire's review:
* Added a test that syncs a prefix wipe
* Added validation so that "invalid empty" entries are rejected (empty hash but len not 0, or len 0 but hash not empty hash)
* Removed allocations from the ParentIterator in `fs` impl

iroh-sync/src/sync.rs
dignifiedquire: mostly around the prefix logic, edge cases etc adding and removing things a lot

dignifiedquire (approved):

Frando:
So I wrote a test, found a bug in the range calculation, and fixed it.
This should still get more tests. But now we at least have one test that caught edge case.
Should we go ahead and merge, and then likely look into some proptests or similar? Or does someone have ideas what we should test manually?

Frando:
Last commits have some renames - the bikeshed is over, for now: we settled for `del`.

iroh/src/commands/sync.rs
```diff
@@ -197,6 +215,21 @@ impl DocCommands {
...
             }
+            Self::Del {
+                doc,
+                author,
+                prefix,
+            } => {
+                let doc = get_doc(iroh, env, doc).await?;
+                let author = env.author(author)?;
+                let key = prefix.as_bytes().to_vec();
+                let removed = doc.del(author, key).await?;
```
dignifiedquire, line 226: what about confirmation?

iroh/Cargo.toml
```diff
@@ -68,6 +68,7 @@ colored = { version = "2.0.4", optional = true }
 # Examples
 ed25519-dalek = { version = "2.0.0", features = ["serde", "rand_core"], optional = true }
 reflink-copy = { version = "0.1.8", optional = true }
+dialoguer = { version = "0.11.0", default-features = false }
```
dignifiedquire, line 71: should be optional and enabled with the `cli` feature

iroh/src/commands/sync.rs
```diff
@@ -197,6 +216,31 @@ impl DocCommands {
...
+                {
+                    let key = prefix.as_bytes().to_vec();
+                    let removed = doc.del(author, key).await?;
+                    println!("Deleted {removed} entries.");
+                    println!(
+                        "Inserted an empty entry for author {} with key {prefix}.",
+                        fmt_short(author)
+                    );
+                } else {
+                    println!("Aborted.")
```
dignifiedquire, line 241: maybe `Canceled`, Aborted sounds like the whole program got killed 🤷

dignifiedquire (approved):

PR #1539, opened 2023-09-27
fix: mark initially created endpoints inactive

### LINKED ISSUE #1538: iroh-net: Add way to add non-active peers to the peer map
Currently, when a peer map is being loaded from disk all the peers become active and are pinged.

We just want to add the info from the peer map without it becoming active. But it seems that currently there is no way to do this. There is only `PeerMap::add_peer_addr`, which adds the peer as an active peer.

We need a way to add these peers as passive peers.

dignifiedquire:
## Description

By default endpoints are now inactive, and will not be pinged. 

Closes #1538

### DIFF
--- a/iroh-net/src/magicsock.rs
+++ b/iroh-net/src/magicsock.rs
@@ -1272,6 +1272,7 @@ impl Actor {
                     msock_sender: self.inner.actor_sender.clone(),
                     public_key: dm.src,
                     derp_region: Some(region_id),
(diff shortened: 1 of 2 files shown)


divagant-martian (approved):

divagant-martian:
code looks good plus can confirm this addresses the behaviour we were seeing about loaded from disk peers :rocket:

Issue #1531, opened 2023-09-27
panic on max connections exceeded


divagant-martian:
`thread '<unnamed>' panicked at 'max_open_connections exceeded',  iroh/iroh/src/downloader.rs:528:18`

PR #1544, opened 2023-09-28
feat(iroh, iroh-gossip): introduce `log-self` feature

divagant-martian:
## Description

Aff a feature flag to turn on logging our own `peer_id` in short form and plug this in in sync, gossip and the downloader. This is enabled by default (not sure if we want this for releases thought. Having it as a feature flag at least gives us the option to chose.

This is what I did to debug #1531 and it was really useful

## Notes & open questions

Log sample:
#### Sync
```log
2023-09-28T20:51:18.908126Z DEBUG sync{me=oywqb57ovmdry243}: iroh::sync_engine::live: sync[dial]: start namespace=NamespaceId(q47zmyim5r6isz22…) peer=PublicKey(2jnygwapdm26wwa2) reason=DirectJoin last_state=None
```
#### Gossip
```logs
2023-09-28T20:58:42.854730Z DEBUG gossip{me=zooj7iazcl7rsoal}: iroh_gossip::net: handle out_event EmitEvent(TopicId(wno5nwxtkhhtnqsm…), Received(GossipEvent { content: <33b>, delivered_from: PublicKey(7a7kkzndvbt6eulu), scope: Neighbors }))
```

#### Downloader
```
2023-09-28T21:00:11.107411Z DEBUG downloader{me=4vabufwku3wselbl}: iroh::downloader: download completed peer=2jnygwapdm26wwa26tvcprm3m5vqajmhwz7lx5xizwx637ad5sea kind=Blob { hash: Hash(224746ea89d286220e0770f89cda2ab138143b00e384dd795d5a13b77b094822) }
```

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-gossip/src/net.rs
+++ b/iroh-gossip/src/net.rs
@@ -13,6 +13,8 @@ use tokio::{
     task::JoinHandle,
 };
 use tracing::{debug, trace, warn};
+#[cfg(feature = "log-self")]
+use tracing::{debug_span, Instrument};
 
 use self::util::{read_message, write_message, Dialer, Timers};
 use crate::proto::{self, PeerData, Scope, TopicId};
@@ -85,6 +87,9 @@ impl Gossip {
         let (to_actor_tx, to_actor_rx) = mpsc::channel(TO_ACTOR_CAP);
         let (in_event_tx, in_event_rx) = mpsc::channel(IN_EVENT_CAP);
         let (on_endpoints_tx, on_endpoints_rx) = watch::channel(Default::default());
+
+        #[cfg(feature = "log-self")]
+        let me = endpoint.peer_id().fmt_short();
         let actor = Actor {
             endpoint,
             state,
@@ -101,7 +106,10 @@ impl Gossip {
             subscribers_topic: Default::default(),
         };
         let actor_handle = tokio::spawn(async move {
-            if let Err(err) = actor.run().await {
+            let fut = actor.run();
+            #[cfg(feature = "log-self")]
+            let fut = fut.instrument(debug_span!("gossip", %me));
+            if let Err(err) = fut.await {
                 warn!("gossip actor closed with error: {err:?}");
                 Err(err)
             } else {
@@ -333,7 +341,6 @@ struct Actor {
 
 impl Actor {
     pub async fn run(mut self) -> anyhow::Result<()> {
-        let me = *self.state.me();
         loop {
             tokio::select! {
                 biased;
@@ -341,7 +348,7 @@ impl Actor {
                     match msg {
                         Some(msg) => self.handle_to_actor_msg(msg, Instant::now()).await?,
                         None => {
-                            debug!(?me, "all gossip handles dropped, stop gossip actor");
+                            debug!("all gossip handles dropped, stop gossip actor");
                             break;
                         }
                     }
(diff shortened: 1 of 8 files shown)


iroh-gossip/src/net.rs
```diff
@@ -101,7 +106,10 @@ impl Gossip {
             subscribers_topic: Default::default(),
         };
         let actor_handle = tokio::spawn(async move {
-            if let Err(err) = actor.run().await {
+            let fut = actor.run();
+            #[cfg(feature = "log-self")]
+            let fut = fut.instrument(trace_span!("gossip", %me));
```
flub, line 111: since this is a spawned future I'd strongly recommend to use an `info_span`.  And also to not make it depend on the feature.  Every task should have it's own `info_span` attached as a general rule.

The span should also cover the bit of error handing in the lines below that are still inside this spawned task.

flub (commented):
Why not always do this instead of the extra feature and conditional compilation?  If it's useful, which it clearly is, then let's just always do it.

iroh/src/downloader.rs
```diff
@@ -237,6 +241,12 @@ impl Downloader {
 
             let service = Service::new(getter, dialer, concurrency_limits, msg_rx);
 
+            #[cfg(feature = "log-self")]
+            {
+                service.run().instrument(trace_span!("downloader", %me))
```
flub, line 246: same here, it should always attach the span and be `info_span` as it's run in a task.  Again try to do it around the entire spawned future and not just the `service.run()` method.

iroh/src/sync_engine/live.rs
```diff
@@ -193,9 +196,12 @@ impl<S: store::Store> LiveSync<S> {
             to_actor_rx,
             to_actor_tx.clone(),
         );
-        let span = debug_span!("sync", %me);
+
         let task = rt.main().spawn(async move {
-            if let Err(err) = actor.run().instrument(span).await {
+            let fut = actor.run();
+            #[cfg(feature = "log-self")]
+            let fut = fut.instrument(trace_span!("sync", %me));
```
flub, line 203: same note, `info_span` and cover entire future that's spawned into the task.

Arqu:
This might / will be PII eventually, so having the option to scrub it is nice.

dignifiedquire:
> This might / will be PII eventually

unlikely, peerids are sent over the network and shared as identifiers. 
if we really need to scrub PII we should do so on themlog storage side, as there is already much more in the logs, like ip addresses

Frando:
Yeah, let's add this. Test logs with multiple nodes are basically unusable without. I did parts of this in #1530 but let's get this in!

@flub has more experience with `tracing` then I do so don't have much to add on the implementation front.

divagant-martian:
can people comment on deciding to make this a compile time flag? @Frando @dignifiedquire @flub (heck we are four here let's just bring @ramfox  and @b5 ) my thinking is that logs are already very hard to read.  Your own peer id is useful only when you are dealing with multiple nodes in the single run. 

This is not the case of users in general, and I would not include this in release builds by default for example. Basically I think it's very nice to be able to have cleaner logs by just removing a feature because this adds "a lot" to already noisy logs

divagant-martian:
agreed feature is not necessary

iroh-gossip/src/net.rs
divagant-martian: ok just to be clear I don't understand what this changes, not sure what the effect of this is

flub:
> agreed feature is not necessary

for completeness: everyone agreed the logging added here is super useful and we should totally have it.  but we prefer it always-on, so without the crate-level feature to control it.

i don't mind if you want to re-open this and remove the crate-feature or re-create the PR.  though I think it's a reasonable refactor to do to this PR so re-opening might be easier.

iroh-gossip/src/net.rs
flub: See also https://discord.com/channels/949724860232392765/950683937661935667/1156506832567808060 where @Frando is also asked me this question and I (hopefully!) gave a similar answer.

Issue #1541, opened 2023-09-28
revitalize metrics


Arqu:
We collect some stats but don't really utilize them. Need to collect more and bring them back into our workflow.

Arqu:
We got metrics and logs for anchors and derpers up on the new grafana. This will be extended in the future, but for now things are back in action.

PR #1546, opened 2023-09-29
Stats

rklaehn:
## Description

Add some very detailed stats about where exactly the transfer_collection fn spends it's time.
DRY the progress reporting in get, to make sure the same code is used when piping to console and when writing to a file

## Notes & open questions

We got lots of info now, but where exactly should it go. Events?

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -246,9 +246,9 @@ dependencies = [
 
 [[package]]
 name = "bao-tree"
-version = "0.8.0"
+version = "0.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9282899b0068a806644365e8416beb676203c471ade24b711c825817a2968e9f"
+checksum = "477137d698cec66b733dc0d9e6a6422b351bf9b549d9150a5fbee2870538d49f"
 dependencies = [
  "bytes",
  "futures",
@@ -2067,9 +2067,9 @@ dependencies = [
 
 [[package]]
 name = "iroh-io"
-version = "0.2.2"
+version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e9c47e66a8ee652d554f00e085a0a6559b6c029036aab918a412aad1541b3fea"
+checksum = "19ace4f69567bfeb726672bab901d3e81be0c01119d860b2a10b7efb1553b880"
 dependencies = [
  "bytes",
  "futures",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -246,9 +246,9 @@ dependencies = [
 
 [[package]]
 name = "bao-tree"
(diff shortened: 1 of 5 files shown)


rklaehn:
@Arqu the question is what should be done with the gathered stats. Do you really need them all, and if yes, where should they go?

There are events TransferCollectionCompleted and TransferAborted. I guess we could add them there. As discussed, you really also want to keep track of data for aborted collections, otherwise people can cheat.

Arqu:
> @Arqu the question is what should be done with the gathered stats. Do you really need them all, and if yes, where should they go?
> 
> There are events TransferCollectionCompleted and TransferAborted. I guess we could add them there. As discussed, you really also want to keep track of data for aborted collections, otherwise people can cheat.

`TransferCollection[Started, Completed, Aborted]` are solid ones to push some of the data as that is the most useful for `iroh.network`. I also agree we don't really need all of it there, unless I'm missing something we only need the sizes there.

We will probably end up with *two* pipes for metrics data. One is the one we already have and will serve dev metrics and end up in grafana (which all of your new metrics should pipe into & I can take care as part of https://github.com/n0-computer/iroh/issues/1541). The second one will be for the actual byte & resource accounting which will probably be a separate bus and I wouldn't worry about that here, getting the sizes propagated to the `TransferCollection` events is plenty.

Arqu:
Also another mention here in regards to the PR, we do have the possibility to bring back tracing spans as part of the metrics pipe which would practically also give us these timings without having to manually sprinkle stuff in code. But they are resource expensive (both perf and bw wise) and will need some time to be brought back. But we could make it a feature toggle.

rklaehn:
> Also another mention here in regards to the PR, we do have the possibility to bring back tracing spans as part of the metrics pipe which would practically also give us these timings without having to manually sprinkle stuff in code. But they are resource expensive (both perf and bw wise) and will need some time to be brought back. But we could make it a feature toggle.

I have been burned before regarding tracing spans.

https://github.com/tokio-rs/tracing/issues/515
https://github.com/tokio-rs/tracing/issues/458

This has all been fixed, so I am fine with using tracing to get the total amount of time spent for a transfer. But wrapping a tracing span around every single write or read op would end up just measuring the speed of the tracing lib.

I think just measuring the time manually for each IO op and then putting the result somewhere seems safer.

Arqu:
Yep, that's exactly what I meant, and yes, we've been burned by this exact thing ~1y ago.

rklaehn:
I mean, it is not surprising. Using the tracing lib, a counter increase means taking a lock, doing a dictionary lookup for the dynamic counter name, and then increasing the actual counter.

Increasing a local counter is basically free in comparison. No lock, do dict lookup.

dignifiedquire (approved):

PR #1550, opened 2023-09-29
feat: add blobs from byte streams

Frando:
## Description

This adds methods in the client, RPC and node to add blobs from byte streams.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-bytes/src/baomap.rs
+++ b/iroh-bytes/src/baomap.rs
@@ -11,12 +11,12 @@ use crate::{
 };
 use bao_tree::{blake3, ChunkNum};
 use bytes::Bytes;
-use futures::{future::BoxFuture, stream::LocalBoxStream, StreamExt};
+use futures::{future::BoxFuture, stream::LocalBoxStream, Stream, StreamExt};
 use genawaiter::rc::{Co, Gen};
 use iroh_io::AsyncSliceReader;
 use range_collections::RangeSet2;
 use serde::{Deserialize, Serialize};
-use tokio::sync::mpsc;
+use tokio::{io::AsyncRead, sync::mpsc};
 
 pub use bao_tree;
 pub use range_collections;
@@ -181,7 +181,7 @@ pub trait Store: ReadableStore + PartialMap {
     ///
     /// Returns the hash of the imported file. The reason to have this method is that some database
     /// implementations might be able to import a file without copying it.
-    fn import(
+    fn import_file(
         &self,
         data: PathBuf,
         mode: ImportMode,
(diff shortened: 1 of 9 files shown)


iroh-bytes/src/baomap.rs
```diff
@@ -181,19 +181,38 @@ pub trait Store: ReadableStore + PartialMap {
...
     ) -> BoxFuture<'_, io::Result<(TempTag, u64)>>;
 
-    /// This trait method imports data from memory.
+    /// Import data from memory.
     ///
     /// It is a special case of `import` that does not use the file system.
     fn import_bytes(&self, bytes: Bytes, format: BlobFormat) -> BoxFuture<'_, io::Result<TempTag>>;
 
+    /// Import data from a stream of bytes.
+    fn import_stream(
```
dignifiedquire, line 198: would it make sense for the non file based imports to allow adding a name manually?

iroh/src/client.rs
```diff
@@ -323,6 +366,75 @@ where
...
+        });
+        Self {
+            stream: Box::pin(stream),
+        }
+    }
+    /// Finish writing the stream, ignoring all intermediate progress events.
+    ///
+    /// Returns a hash and a size. When importing a single blob, this is the hash and size of that
+    /// blob. When importing a collection, this is the hash of the collection and the total size of
+    /// all imported blobs.
```
dignifiedquire, line 405: is the size including or excluding the size of the collection blob?

Frando: It is excluding the size of the collection blob. Added this to the docs. Or should we change the behavior? Would need a little more work to report the collection blob size as well, but can do.

iroh-bytes/src/baomap.rs
Frando: This always adds raw blobs which don't have names. Adding tags and wrapping collections happens at the `node.rs` layer.

What we could do is add a`WrapOption` to the RPC message and client interface which would be 
```rust
enum WrapOption {
  NoWrap,
  Wrap { name: String }
```
which would allow to wrap the single blob in a collection with a single entry by passing the name here. Do we want/need this?

We might want to add a client & rpc API to create collections from a list of hashes instead maybe? Also not sure.

dignifiedquire: I meant to be able to tag the raw blobs, is that not sth we want to do?

Frando: That is possible already: The [client API](https://github.com/n0-computer/iroh/pull/1550/files#diff-481403521f6c5ced02ca5150086fc28be9ad2f8e0e518c3742222f3c8b504b84R276) has a `tag` option, and it is handled in the [rpc handler in `node.rs`](https://github.com/n0-computer/iroh/pull/1550/files#diff-4bd93a0bc8c0813c38e9d897deb531fef9ca4517415c3f3f873a9cb3dbefa595R1412)

Frando:
I added a test. Think this is ready to go from my side.

dignifiedquire (approved):

PR #1560, opened 2023-10-02
fix: actually allow to disable DERP

### LINKED ISSUE #1558: Bug: `node::Builder::disable_derp` is broken
.. as in, has  no effect. We moved the default in `MagicEndpoint` to be the default derp map (and not no derp). However, we never call `MagicEndpointBuilder::disable_derp()`, even if you call `node::Builder::disable_derp`. This needs to be fixed.

Frando:
## Description

`node::Builder::disable_derp` is broken. We do not call `MagicEndpointBuilder::disable_derp` for that, and at some point we changed the magic endpoint to default to the default, n0-supplied derp servers.

This fixes this through a `DerpMode` enum that is much more explicit I think.

Fixes #1558 

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/src/magic_endpoint.rs
+++ b/iroh-net/src/magic_endpoint.rs
@@ -173,7 +173,7 @@ impl MagicEndpointBuilder {
     /// will result in an error.
     ///
     /// [`bind`]: MagicEndpointBuilder::bind
-    pub fn set_derp(mut self, derp_mode: DerpMode) -> Self {
+    pub fn derp_mode(mut self, derp_mode: DerpMode) -> Self {
         self.derp_mode = derp_mode;
         self
     }
@@ -600,7 +600,7 @@ mod tests {
             async move {
                 let ep = MagicEndpoint::builder()
                     .alpns(vec![TEST_ALPN.to_vec()])
-                    .enable_derp(derp_map)
+                    .derp_mode(DerpMode::Custom(derp_map))
                     .bind(0)
                     .await
                     .unwrap();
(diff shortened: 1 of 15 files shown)


iroh-net/src/magic_endpoint.rs
```diff
@@ -253,13 +240,14 @@ impl MagicEndpointBuilder {
...
     pub async fn bind(self, bind_port: u16) -> Result<MagicEndpoint> {
-        ensure!(
-            self.derp_map
-                .as_ref()
-                .map(|m| !m.is_empty())
-                .unwrap_or(true),
-            "Derp server enabled but DerpMap is empty",
-        );
+        let derp_map = match self.derp_mode {
+            DerpMode::Disabled => DerpMap::default(),
```
dignifiedquire, line 244: can we make sth like `DerpMap::empty()` or sth like that for this? `default` reads very confusing here

iroh/examples/sync.rs
```diff
@@ -130,13 +132,14 @@ async fn run(args: Args) -> anyhow::Result<()> {
     println!("> our secret key: {}", secret_key);
 
     // configure our derp map
-    let derp_map = match (args.no_derp, args.derp) {
-        (false, None) => Some(default_derp_map()),
-        (false, Some(url)) => Some(DerpMap::from_url(url, 0)),
-        (true, None) => None,
+    // configure our derp map
```
flub, line 135: dup comment

flub (approved):
agreed on `DerpMap::emtpy()`, that's been confusing for a while.  but overall this i much better than what i did before!

iroh-net/src/magic_endpoint.rs
Frando: fixed

iroh/examples/sync.rs
Frando: fixed

PR #1562, opened 2023-10-02
chore(baomap): drop all baomap logging to debug or lower

b5:
## Description

We've decided that `INFO` logging level & above should be reserved for node operators that are at least familiar with iroh's different subsystems, but _not_ engineering internals.

This is less than ideal on node startup: 
Screenshot 2023-10-02 at 3 53 19 PM

So, drop to debug level.

## Change checklist

- [ ] Self-review.
- [ ] ~~Tests if relevant.~~


flub (approved):
lgtm, we can always bring some back if this proved to be too broad.

dignifiedquire (approved):

PR #1563, opened 2023-10-02
feat: `doc import` & `doc export` commands

ramfox:
## Description

- adds doc import / doc export commands on the cli
- `doc import` can import a file or directory
- `doc export` exports a single entry as a file
- adds `DocClient::doc_set_hash` that sets an entry in a document by using the hash, key, and size of the entry.

 
## Notes & open questions

Re-implementation of https://github.com/n0-computer/iroh/pull/1504, but it has drifted so far that it is now easier to start a new PR.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh/src/commands/sync.rs
```diff
@@ -354,3 +480,252 @@ async fn print_entry(doc: &Doc, entry: &Entry, content: bool) -> anyhow::Result<
...
+        Self {
+            task: Arc::new(Mutex::new(Some(task))),
+            sender: sender.clone(),
+        }
+    }
+
+    fn found(&mut self, name: String, id: u64, size: u64) {
+        if !self.sender.is_disconnected() {
+            self.sender
+                .send(ImportProgressEvent::Found { name, id, size })
```
dignifiedquire, line 672: all these `.send`s should probably be `.send_async`

iroh/src/commands/sync.rs
```diff
@@ -354,3 +480,252 @@ async fn print_entry(doc: &Doc, entry: &Entry, content: bool) -> anyhow::Result<
...
+    mut doc_set_progress: impl Stream<Item = Result<DocSetProgress>> + Unpin,
+) -> Result<(u64, u64)> {
+    let mut mp = ImportProgressState::new(doc_id);
+    let mut task_mp = mp.clone();
+    // new task to iterate through the stream of added files
+    // and send them through the `doc_set_updates` `UpdateSink`
+    // to get added to the doc
+    //
+    // TODO: catch errors and signal stop if error occurs in task
+    let add_progress_task = tokio::spawn(async move {
```
ramfox, line 519: should i be spawning from something other than the ambient tokio runtime?

iroh/src/commands/sync.rs
```diff
@@ -223,6 +285,70 @@ impl DocCommands {
...
+                        WrapOption::NoWrap,
+                    )
+                    .await?;
+                let (updates, progress) = doc.set_hash_streaming(author).await?;
+                let root_prefix = match root.parent() {
+                    Some(p) => p.to_path_buf(),
+                    None => PathBuf::new(),
+                };
+                let (entries, size) =
+                    import_coordinator(doc.id(), root_prefix, prefix, stream, updates, progress)
```
ramfox, line 319: probably needs a better name. engine? It takes the entries off of the stream from `iroh.blobs.add_from_path` and sends them down to the doc using the `doc.set_hash_streaming` sink, and then aggregates the results & powers the progress bars.

iroh/src/commands/sync.rs
```diff
@@ -354,3 +480,273 @@ async fn print_entry(doc: &Doc, entry: &Entry, content: bool) -> anyhow::Result<
...
+            sender: sender.clone(),
+        }
+    }
+
+    async fn found(&mut self, name: String, id: u64, size: u64) {
+        if !self.sender.is_disconnected() {
+            self.sender
+                .send_async(ImportProgressEvent::Found { name, id, size })
+                .await
+                .expect("checked");
```
rklaehn, line 689: Why the check for is_disconnected?

Isn't this the same as `send_async(...).await.ok()`?

iroh-sync/src/sync.rs
```diff
@@ -786,6 +786,26 @@ impl Record {
...
 
+/// Progress updates for the streaming set entry by hash operation.
+#[derive(Debug, Serialize, Deserialize)]
+pub enum DocSetProgress {
+    /// Finished adding entry of the given `key` to the document
+    Done {
+        /// Id for this entry, unique for this import
+        id: u64,
+        /// The key of this entry
+        key: Vec<u8>,
```
rklaehn, line 797: Maybe not for this PR, but we might want to do a newtype for this. This is the key path into the document, correct?

iroh/src/commands/sync.rs
```diff
@@ -98,6 +117,49 @@ pub enum DocCommands {
...
+    Export {
+        /// Document to operate on.
+        ///
+        /// Required unless the document is set through the IROH_DOC environment variable.
+        /// Within the Iroh console, the active document can also be set with `doc switch`.
+        #[clap(short, long)]
+        doc: Option<NamespaceId>,
+        /// Key to the entry (parsed as UTF-8 string)
+        #[clap(short, long)]
+        key: String,
```
rklaehn, line 158: I don't think this needs to happen in this PR. But we can import entire directories, but export only individual blobs, correct? It would be nicer if we could also export entire directories. We could maybe make this an option, and if it is None export the entire doc.

iroh/src/commands/sync.rs
```diff
@@ -354,3 +480,273 @@ async fn print_entry(doc: &Doc, entry: &Entry, content: bool) -> anyhow::Result<
...
+        Some(name) => name
+            .to_os_string()
+            .into_string()
+            .map(|t| t.into())
+            .map_err(|e| anyhow!("{e:?} contains invalid Unicode")),
+        None => bail!("the given `path` does not have a proper directory or file name"),
+    }
+}
+
+pub async fn import_coordinator(
```
rklaehn, line 500: It would be nice to have either some docs or a struct. Otherwise it is a bit hard to know what the two integers are that are returned.

I think we have (total_number, total_size) in a number of places, maybe we can make a struct in iroh-bytes...

iroh/src/commands/sync.rs
```diff
@@ -354,3 +480,273 @@ async fn print_entry(doc: &Doc, entry: &Entry, content: bool) -> anyhow::Result<
...
+        match res? {
+            DocSetProgress::Done { id, size, .. } => {
+                mp.done(id).await;
+                entries += 1;
+                total_size += size;
+            }
+            DocSetProgress::AllDone => {
+                mp.all_done().await;
+                add_progress_task.abort();
+                return Ok((entries, total_size));
```
rklaehn, line 594: I would prefer if we would break here instead of return. So

```rust
break Ok((entries, total_size));
```

and 

```rust
break Err(anyhow!("..."))
```

below.

Then aborting the task can happen afterwards, and at the end just return res.

Alternatively you could wrap the add_progress_task in an AbortingJoinHandle, which we got in some utils somewhere.

iroh/src/commands/sync.rs
```diff
@@ -354,3 +480,273 @@ async fn print_entry(doc: &Doc, entry: &Entry, content: bool) -> anyhow::Result<
...
+        .ok_or(anyhow!("could not convert path to bytes"))?
+        .to_vec();
+    let mut key = prefix.into_bytes().to_vec();
+    key.extend(suffix);
+    Ok(key)
+}
+
+#[derive(Debug, Clone)]
+struct ImportProgressState {
+    task: Arc<Mutex<Option<JoinHandle<()>>>>,
```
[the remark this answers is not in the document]
rklaehn: Can you try taking out the join handle here, so that ImportProgressState is basically just a newtype for the sender?

new would then return Self and an AbortingJoinHandle<()>, and you just have to keep the latter around until you're done. That would allow getting rid of both the Arc and the Mutex.

iroh/src/commands/sync.rs
rklaehn: Coordinator seems fine. An engine I would associate with a stateful thing that stays around for some time.

iroh/src/commands/sync.rs
```diff
@@ -98,6 +117,49 @@ pub enum DocCommands {
...
+        #[clap(short, long)]
+        author: Option<AuthorId>,
+        /// Prefix to add to imported entries (parsed as UTF-8 string). Defaults to no prefix
+        #[clap(long)]
+        prefix: Option<String>,
+        /// Path to a local file or directory to import
+        ///
+        /// Pathnames will be used as the document key
+        #[clap(long)]
+        path: String,
```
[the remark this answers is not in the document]
b5: I agree, should be a positional argument.

iroh/src/commands/sync.rs
ramfox: ohh thank you yes that's better

iroh/src/commands/sync.rs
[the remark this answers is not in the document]
ramfox: 🔥 I have also been frustrated by this

iroh/src/commands/sync.rs
ramfox: yea i had the same thought: https://github.com/n0-computer/iroh/pull/1504#discussion_r1340695276

The conclusion was, let's wait for someone to ask for it.

iroh/src/commands/sync.rs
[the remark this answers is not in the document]
ramfox: that's so much better, thank you for the suggestion!!!

ramfox:
Okay, updated w/ PR suggestions, but when testing directories with a ton of files and lots of nesting (a git repo for example), it started hanging. When adding that same directory using `blobs add`, it imported without any problems. Checking into it now.

rklaehn:
Code looks fine for me now. Much better without the Mutex<Option<Task>>...

ramfox:
Stats on three different refactors trying to reduce the time to importing the `iroh.computer` git repo to a doc

**32376** entries totaling **730.56 MiB**

### "Linear"
#### [e86639a](https://github.com/n0-computer/iroh/pull/1563/commits/e86639adaeaedc8c449ba858d05ecbd557b5d8a7) after each blob has been added, send the entry to the doc & wait for success before moving on 
WITH PROGRESS BARS 

`IROH_DATA_DIR=~/Work/iroh/data_dir_0 cargo run --bin iroh doc import  -a  -d   61.90s user 10.91s system 7% cpu 15:31.21 total`

### "Streaming"
#### [9fdc158](https://github.com/n0-computer/iroh/pull/1563/commits/9fdc15862a61a53ec4ab7366f2bad4819bb721d4) stream from the `blobs_add_progress` into the `import_progress` sink
NO PROGRESS BARS
`IROH_DATA_DIR=~/Work/iroh/data_dir_0 cargo run --bin iroh doc import  -a  -d   41.88s user 7.45s system 6% cpu 13:40.82 total`

### "Parallelized"
#### [840b908](https://github.com/n0-computer/iroh/pull/1563/commits/840b908666b896f857e72d1e3c1abd76c5ed4bdb) Pull chunks off of the `blob_add_progress` stream & `doc_set_hash` the entries in separate tasks in parallel
NO PROGRESS BARS 

- chunks of 10 in parallel
`IROH_DATA_DIR=data_dir_0 cargo run --bin iroh doc import ~/Work/iroh.computer  34.02s user 6.04s system 4% cpu 13:58.68 total`

- chunks of 100 in parallel
`IROH_DATA_DIR=data_dir_1 cargo run --bin iroh doc import ~/Work/iroh.computer  14.20s user 2.85s system 2% cpu 13:52.61 total`

## halp
So, I'm obviously doing something wrong while attempting to add the entries in parallel, because the chunk size is having no effect on the overall time.

Overall, the "best" UX experience right now is [e86639a](https://github.com/n0-computer/iroh/pull/1563/commits/e86639adaeaedc8c449ba858d05ecbd557b5d8a7), since the progress bars show neatly. This takes actual minutes longer when importing the git repo, but is almost negligible when importing more "reasonable" use cases, eg sharing a directory of photos.

To illustrate: for a directory of 423 photos totaling ~350MB, it took 11.6 seconds to import & with the progress bars you can see all the entries zooming by. (for reference, adding the files to just the blob store using `iroh blob add` takes 6.5 seconds)

**I would like to revert to [e86639a](https://github.com/n0-computer/iroh/pull/1563/commits/e86639adaeaedc8c449ba858d05ecbd557b5d8a7) for the release on monday, so we can have the import feature on time. Then I'd love to pair or get some guidance on what I am doing wrong when trying to parallelize.**

dignifiedquire:
the above error was likely due to my iroh node being a missmatch

iroh/src/util/fs.rs
```diff
@@ -171,11 +172,70 @@ pub async fn load_secret_key(key_path: PathBuf) -> anyhow::Result<SecretKey> {
...
+    path_content_info0(path)
+}
+
+fn path_content_info0(path: impl AsRef<Path>) -> anyhow::Result<PathContent> {
+    let mut files = 0;
+    let mut size = 0;
+    let path = path.as_ref();
+
+    if path.is_dir() {
+        for entry in read_dir(&path).context("read_dir err")? {
```
dignifiedquire, line 195: https://crates.io/crates/walkdir is probably a good deal faster

iroh/src/commands/sync.rs
```diff
@@ -337,6 +338,21 @@ impl DocCommands {
                 }
                 let root = canonicalize_path(&path)?.canonicalize()?;
                 let tag = tag_from_file_name(&root)?;
+
+                println!("Preparing import...");
+                // get information about the directory or file we are trying to import
+                // and confirm with the user that they still want to import the file
+                let PathContent { size, files } = path_content_info(root.clone())?;
```
dignifiedquire, line 345: this needs to be spawn_blocking call

iroh/src/util/fs.rs
ramfox: switched to `walkdir` & got a stack overflow when adding a dir w/ 30k entries, so switched it back. lmk if you have an insight.

dignifiedquire: Oo, I would definitely file an issue, that is surprising

b5 (approved):
smoke tested on a 127 file import, single file export, looks great!

PR #1564, opened 2023-10-02
feat(iroh-sync): store peers per doc

divagant-martian:
## Description

Store peers that have been useful for sync

## Notes & open questions

A live test consisted on:
- `node_a` creates and shares a ticket (with only `node_a` in it)
- `node_b` joins the ticket
- `node_c` joins the ticket. At this point all three peers are aware of the other two
- shutdown all nodes
- bring `node_c` back again
- join `node_c` using the ticket from `node_a` (which is gone)
- `node_c` has cached `node_b` as useful and dials it

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-sync/src/store/fs.rs
+++ b/iroh-sync/src/store/fs.rs
@@ -329,7 +329,7 @@ impl super::Store for Store {
         let read_tx = self.db.begin_read()?;
         let peers_table = read_tx.open_multimap_table(NAMESPACE_PEERS_TABLE)?;
         let mut peers = Vec::with_capacity(super::PEERS_PER_DOC_CACHE_SIZE.get());
-        for result in peers_table.get(namespace.as_bytes())? {
+        for result in peers_table.get(namespace.as_bytes())?.rev() {
             let (_nanos, &peer) = result?.value();
             peers.push(peer);
         }
--- a/iroh-sync/src/store/fs.rs
+++ b/iroh-sync/src/store/fs.rs
@@ -19,7 +19,7 @@ use crate::{
     sync::{
         Author, Entry, EntrySignature, Namespace, Record, RecordIdentifier, Replica, SignedEntry,
     },
-    AuthorId, NamespaceId,
+    AuthorId, NamespaceId, PeerIdBytes,
 };
 
 use super::{pubkeys::MemPublicKeyStore, PublicKeyStore};
@@ -115,6 +115,7 @@ impl super::Store for Store {
     type ContentHashesIter<'a> = ContentHashesIterator<'a>;
     type AuthorsIter<'a> = std::vec::IntoIter<Result<Author>>;
     type NamespaceIter<'a> = std::vec::IntoIter<Result<NamespaceId>>;
+    type PeersIter<'a> = std::vec::IntoIter<Result<PeerIdBytes>>;
 
     fn open_replica(&self, namespace_id: &NamespaceId) -> Result<Option<Replica<Self::Instance>>> {
         if let Some(replica) = self.replicas.read().get(namespace_id) {
@@ -236,6 +237,12 @@ impl super::Store for Store {
 
     fn content_hashes(&self) -> Result<Self::ContentHashesIter<'_>> {
         ContentHashesIterator::create(&self.db)
+    }
+
+    fn register_useful_peer(&self, namespace: &NamespaceId, peer: crate::PeerIdBytes) {}
+
+    fn get_sync_peers(&self, namespace: &NamespaceId) -> Result<Self::PeersIter<'_>> {
+        Ok(vec![].into_iter())
     }
 }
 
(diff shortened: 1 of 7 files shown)


divagant-martian:
test is meh right now but would like an opinion on the core @Frando thanks for the patience, sync and redb are both foreign to me

iroh/src/sync_engine/live.rs
```diff
@@ -773,7 +797,16 @@ impl<S: store::Store, B: baomap::Store> Actor<S, B> {
             }
         }
         let state = match result {
-            Ok(_) => SyncState::Finished,
+            Ok(_) => {
+                // register the peer as useful for the document
+                if let Err(e) = self
+                    .replica_store
+                    .register_useful_peer(namespace, *peer.as_bytes())
```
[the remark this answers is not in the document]
divagant-martian: Not sure. Peers are pruned to keep the size as better ones come in and the peer is good for sync regardless of downloads. I would think if sync decided to remove peers for failed downloads these would be pruned as the replacements are added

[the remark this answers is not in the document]
divagant-martian: It probably makes sense to add a removal functionality if sync itself starts to actively dislike peers in a way

iroh-sync/src/store/fs.rs
```diff
@@ -237,6 +250,72 @@ impl super::Store for Store {
...
+            let mut peers_table = write_tx.open_multimap_table(NAMESPACE_PEERS_TABLE)?;
+            let namespace_peers = peers_table.get(namespace.as_bytes())?;
+
+            // find any previous entry for the same peer to remove it
+            let mut prev_peer_nanos = None;
+            // calculate the len in the same loop since calling `len` is another fallible operation
+            let mut len = 0;
+            // the oldest entry in the table, candidate for removal depending on len
+            let mut oldest_entry: Option<(Nanos, PeerIdBytes)> = None;
+            for result in namespace_peers {
```
dignifiedquire, line 270: I don't think you need to iterate over all values, I think you can just take the first one. Per the docs of https://docs.rs/redb/latest/redb/trait.ReadableMultimapTable.html#tymethod.get the iterator is sorted in ascending order, and the order for tuples is that they are element wise. So if the timestamp is the first element in the value, the oldest value should be the first one of the iterator.

divagant-martian: Oh nice I mistook this for descending. I still need to iterate the values to find a previous entry for the same peer though. But candidate for removal can make use of this fact, ty!

dignifiedquire: Also MultimapValue implments https://doc.rust-lang.org/nightly/core/iter/trait.DoubleEndedIterator.html so it is cheap to get the last element as well

divagant-martian: done in [d84c374](https://github.com/n0-computer/iroh/pull/1564/commits/d84c374814628da767f458d1b84062e9a059800a)

iroh-sync/src/store/fs.rs
```diff
@@ -237,6 +250,95 @@ impl super::Store for Store {
...
+
+                    if oldest_peer == peer {
+                        // oldest peer is the current one, so replacing the entry for the peer will
+                        // maintain the size
+                        drop(namespace_peers);
+                        peers_table.remove(namespace, (oldest_nanos, oldest_peer))?;
+                        peers_table.insert(namespace, (nanos, peer))?;
+                    } else {
+                        // calculate the len in the same loop since calling `len` is another fallible operation
+                        let mut len = 1;
```
dignifiedquire, line 289: why does this star at 1?

dignifiedquire (changes requested):
one question, which might just be me not understanding otherwise lgtm

iroh-sync/src/store/fs.rs
```diff
@@ -237,6 +250,95 @@ impl super::Store for Store {
...
+        }
+        write_tx.commit()?;
+
+        Ok(())
+    }
+
+    fn get_sync_peers(&self, namespace: &NamespaceId) -> Result<Option<Self::PeersIter<'_>>> {
+        let read_tx = self.db.begin_read()?;
+        let peers_table = read_tx.open_multimap_table(NAMESPACE_PEERS_TABLE)?;
+        let mut peers = Vec::with_capacity(super::PEERS_PER_DOC_CACHE_SIZE.get());
```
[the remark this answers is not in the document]
divagant-martian: I tried to check this based on the other todos but I'm worried this will keep some underlying lock around for as long as the response is kept. Not sure, what do you think?

iroh-sync/src/store/fs.rs
divagant-martian: Because sigue it's some the iterator already advanced by one

dignifiedquire: ah yes 😅

iroh-sync/src/store/fs.rs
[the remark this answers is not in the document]
dignifiedquire: well it would just keep the read tx around, so that should be fine, or at least the caller can decide if immediately to colllect or not

dignifiedquire (approved):

PR #1565, opened 2023-10-03
test(iroh-sync): fix `test_content_hashes_iterator_memory`

divagant-martian:
## Description

the content of this test is the same as `test_basics_memory` so it;s doing a duplicated test scenario. Name suggests this is what should be here instead

## Notes & open questions

na

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.


Frando (approved):
oh, thanks for catching this

PR #1567, opened 2023-10-03
fix(net): stop deleting endpoints we want to keep

dignifiedquire:
also simplifies the general logic around endpoint management

### DIFF
--- a/iroh-net/src/magicsock/endpoint.rs
+++ b/iroh-net/src/magicsock/endpoint.rs
@@ -459,30 +459,7 @@ impl Endpoint {
             return msgs;
         }
         self.last_full_ping.replace(now);
-
-        // first cleanout out all old endpoints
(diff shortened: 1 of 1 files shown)


iroh-net/src/magicsock/endpoint.rs
```diff
@@ -1312,13 +1244,6 @@ pub struct EndpointInfo {
     pub latency: Option<Duration>,
 }
 
-#[derive(Default, Debug, PartialEq, Eq, Clone, Copy, Hash)]
-enum Index {
```
Frando, line 1316: Thanks for removing this! I was reading in the code and wondering that it was never used apart from a deletion flag.

Frando (approved):
LGTM! Apart from fixing the bug also a nice code simplification. We'll want more of those in `iroh_net` :)

PR #1569, opened 2023-10-03
fix: avoid blockage in endpoint handling

### LINKED ISSUE #1568: test: sync_subscribe times out
Running this test on current `main` fails on my machine (macos) reliably

```
     Running tests/sync.rs (target/debug/deps/sync-6ea38533d3f95595)

running 2 tests
test sync_subscribe ... FAILED
test sync_subscribe_stop ... ok

failures:

---- sync_subscribe stdout ----
Error: deadline has elapsed


failures:
    sync_subscribe

test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 2 filtered out; finished in 6.20
```

dignifiedquire:
- avoid using `tokio::sync::watch` has it has a history of causing random blockages
- do not recursively call into the magicsocket to retrieve endpoints

Closes #1568

### DIFF
--- a/iroh-gossip/src/net.rs
+++ b/iroh-gossip/src/net.rs
@@ -9,7 +9,7 @@ use rand::rngs::StdRng;
 use rand_core::SeedableRng;
 use std::{collections::HashMap, future::Future, sync::Arc, task::Poll, time::Instant};
 use tokio::{
-    sync::{broadcast, mpsc, oneshot, watch},
+    sync::{broadcast, mpsc, oneshot},
     task::JoinHandle,
 };
 use tracing::{debug, error_span, trace, warn, Instrument};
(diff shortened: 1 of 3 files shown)


iroh-gossip/src/net.rs
```diff
@@ -229,10 +231,14 @@ impl Gossip {
     ///
     /// This will be sent to peers on Neighbor and Join requests so that they can connect directly
     /// to us.
+    ///
+    /// This is only best effort, and will drop new events if backed up.
```
Frando, line 235: That's why I opted for `Watch` in the first place: We only ever care for the latest value that was sent. Isn't there another alternative to `Watch` that keeps that behavior? Could just be something like a ringbuf with capacity 1.
But also fine to keep as is for now, there won't be more than 64 endpoint updates before the actor pulls them out.

Frando (approved):

iroh-gossip/src/net.rs
dignifiedquire: I don't trust watch.. I also looked at the tokio issue tracker and there are way too many cases where subtle misuse of it results in deadlocks

PR #1579, opened 2023-10-05
perf: no more channels for UDP send/recv

Frando:
## Description

Do not use channels for the hot path of sending and receiving QUIC UDP packets.

* Put `PeerMap` into a `Mutex` so that we can access it for the address remapping directly in the `AsyncUdpSocket` `poll_send/poll_recv` methods
* `poll_recv` and `poll_send` the underlying UDP socket directly in `MagicSock as AsyncUdpSocket` and only forward Derp, Disco and Stun packages over channels

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


iroh-net/src/magicsock.rs
```diff
@@ -285,6 +287,242 @@ impl Inner {
...
+                            io::ErrorKind::Other,
+                            "no IPv6 connection",
+                        )));
+                    }
+                    let conn = if addr.is_ipv6() {
+                        self.pconn6.as_ref().unwrap()
+                    } else {
+                        &self.pconn4
+                    };
+                    let n = ready!(conn.poll_send(&self.udp_state, cx, &transmits))?;
```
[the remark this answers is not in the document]
Frando: good catch, might be. made it `warn` log instead in the latest commit and not return.

it is a good question when we should return an error to quinn and when to fail silently..?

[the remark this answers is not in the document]
dignifiedquire: ideally we should only error to quinn when all three options fail

Frando:
So I looked into how quinn parses packets, and found a reliable way to make it ignore the non-quic packets (stun and disco) that we pass right through to quinn with the approach in this PR. By setting `grease_quic_bit` to false in the endpoint config, quinn ignores all packets that do not have the quic `FIXED_BIT` set to 1. So if we set the first byte of non-quic packets to 0, they are ignored directly (and without much perf overhead). I'd guess that's a fine approach to keep then (because it lets us keep the buffers without restructuring them if there's a non-quic packet inside).

We still have failures though, which I'll investigate next.

iroh-net/src/magicsock.rs
```diff
@@ -278,6 +286,664 @@ impl Inner {
...
+        let dest = transmits[0].destination;
+        for transmit in transmits.iter() {
+            if transmit.destination != dest {
+                break;
+            }
+            n += 1;
+        }
+        // TODO: This is the remaining alloc on the hot path for send.
+        // Unfortunately I don't see a way around this because we have do modify the transmits.
+        let mut transmits = transmits[..n].to_vec();
```
dignifiedquire, line 350: we could use a fixed buffer, that might help

iroh-net/src/magicsock.rs
```diff
@@ -278,6 +286,664 @@ impl Inner {
...
+        let dest = QuicMappedAddr(dest);
+
+        match self.peer_map.get_send_addrs_for_quic_mapped_addr(&dest) {
+            // TODO: Handle ping actions
```
dignifiedquire, line 354: this is done, isn't it?

iroh-net/src/magicsock.rs
```diff
@@ -278,6 +286,664 @@ impl Inner {
...
+                if let Some(addr) = udp_addr {
+                    // rewrite target addresses.
+                    for t in transmits.iter_mut() {
+                        t.destination = addr;
+                    }
+                    match ready!(self.poll_send_udp(addr, &transmits, cx)) {
+                        Ok(n) => {
+                            debug!(peer = %public_key.fmt_short(), ?addr, ?n, "sent udp");
+                            // truncate the transmits to be sent over derp, otherwise we'd
+                            // potentially send the same transmits multiple times over derp.
```
dignifiedquire, line 378: the system relies on double sending over both udp and derp, if there are both addresses available, so this needs to be removed

iroh-net/src/magicsock.rs
```diff
@@ -278,6 +286,664 @@ impl Inner {
...
+                        )
+                    });
+                    Poll::Ready(Err(err))
+                } else {
+                    Poll::Ready(Ok(n))
+                }
+            }
+            None => {
+                error!(addr=%dest, "no endpoint for mapped address");
+                Poll::Ready(Ok(n))
```
dignifiedquire, line 420: this should error, we can't send anything as we have no connection option

iroh-net/src/magicsock.rs
```diff
@@ -278,6 +286,664 @@ impl Inner {
...
+                match self
+                    .peer_map
+                    .get_quic_mapped_addr_for_ip_port(&SendAddr::Udp(meta.addr))
+                {
+                    None => {
+                        warn!(peer=?meta.addr, len = meta.len, ?count, "no peer state found, skipping");
+                    }
+                    Some(quic_mapped_addr) => {
+                        trace!(peer = ?meta.addr, len = meta.len, ?count, "recv ok, peer state found");
+                        num_quic_msgs += meta.len / meta.stride;
```
dignifiedquire, line 528: I think this number will be wrong, this will round down, if the `meta.stride` doesn't evenly divide `meta.len`, so should round up if you do the calculation like this

dignifiedquire: probably best, to just manually count in the above loop

iroh-net/src/magicsock.rs
```diff
@@ -278,6 +286,664 @@ impl Inner {
...
+        if num_msgs > 0 {
+            inc_by!(MagicsockMetrics, recv_datagrams, num_msgs as _);
+            trace!("received {} datagrams", num_msgs);
+            return Poll::Ready(Ok(num_msgs));
+        }
+
+        Poll::Pending
+    }
+
+    /// Handles a discovery message and reports whether `msg`f was a Tailscale inter-node discovery message.
```
dignifiedquire, line 607: typo "`msg`f"

iroh-net/src/magicsock.rs
Frando: To clarify my thought: The double sending should still be working fine with this change. If `poll_send_udp`returns `Poll::Ready(n)` we'd only send those `n` transmits to Derp, but we return this `n` from `poll_send` so on the next call (which happens immediately) we'll get passed the next transmits after `n`. Procedure repeats. So if my understanding is correct without the truncate we'd double-send to Derp the transmits following `n`(because they'll be sent on this and the next call to `poll_send` from quinn). Which might be just fine though.

dignifiedquire: I see, okay, then we should do this, makes sense 👍

Frando: Will restore!

iroh-net/src/magicsock.rs
Frando: Added a fixed buffer, so one alloc less. Instead there's a `parking_lot::Mutex::lock` but it should have no contention ever so fast.

We will still allocate when sending over Derp, because the derp actor expectes a `Vec<Bytes>`.  Not sure if there's a way to reduce allocs here as well, likely not - would defer to followup.

iroh-net/src/magicsock.rs
Frando: Changed to that.

iroh-net/src/magicsock.rs
```diff
@@ -345,9 +348,18 @@ impl Inner {
...
-        // TODO: This is the remaining alloc on the hot path for send.
-        // Unfortunately I don't see a way around this because we have do modify the transmits.
-        let mut transmits = transmits[..n].to_vec();
+
+        // Copy the transmits into an owned buffer, because we will have to modify the send
+        // addresses to translate from the quic mapped address to the actual UDP address.
+        // To avoid allocating on each call to `poll_send`, we use a fixed buffer.
+        let mut transmits = {
+            let mut buf = self.send_buffer.lock();
+            buf.clear();
```
dignifiedquire, line 357: no need to clear, you can just overwrite

Frando: how do I overwrite when the new length is greater than the old length?

iroh-net/src/magicsock/endpoint.rs
```diff
@@ -1608,9 +1621,12 @@ impl EndpointState {
...
                 let elapsed = now.duration_since(last_ping);
                 let needs_ping = elapsed > DISCO_PING_INTERVAL;
-                if !needs_ping {
-                    debug!("ping is too new: {}ms", elapsed.as_millis());
-                }
+                // TODO: remove!
+                // This logs "ping is too new" for each send whenever the endpoint does *not* need
+                // a ping. Pretty sure this is not a useful log, but maybe there was a reason?
+                // if !needs_ping {
+                //     debug!("ping is too new: {}ms", elapsed.as_millis());
```
dignifiedquire, line 1628: it was useful to understand why pings would not be sent, maybe set it to trace?

iroh-net/bench/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "bench"
```
dignifiedquire, line 2: maybe we should call this `iroh-net-bench` or sth like that?

iroh-net/src/magicsock.rs
```diff
@@ -241,42 +238,777 @@ struct Inner {
...
+    /// UDP IPv6 socket
+    pconn6: Option<RebindingUdpConn>,
+    /// Netcheck client
+    net_checker: netcheck::Client,
+    /// The state for an active DiscoKey.
+    disco_info: parking_lot::Mutex<HashMap<PublicKey, DiscoInfo>>,
+    udp_state: quinn_udp::UdpState,
+
+    // Send buffer used in `poll_send_udp`
+    send_buffer: parking_lot::Mutex<Vec<quinn_udp::Transmit>>,
```
dignifiedquire, line 254: nit: I think the length of this doesn't actually change over time, so we could potentially use `Box<[Transmit]>` instead

dignifiedquire: @Frando I filed https://github.com/quinn-rs/quinn/pull/1692 so hopefully we can get rid of this buffer entirely

iroh-net/src/magicsock.rs
```diff
@@ -241,42 +238,777 @@ struct Inner {
...
+            return Poll::Ready(Ok(num_msgs));
+        }
+
+        Poll::Pending
+    }
+
+    /// Handles a discovery message and reports whether `msg` was a Tailscale inter-node discovery message.
+    ///
+    /// For messages received over DERP, the src.ip() will be DERP_MAGIC_IP (with src.port() being the region ID) and the
+    /// derp_node_src will be the node key it was received from at the DERP layer. derp_node_src is None when received over UDP.
```
dignifiedquire, line 649: this comment is outdated

iroh-net/src/magicsock.rs
```diff
@@ -241,42 +238,777 @@ struct Inner {
...
+                            ep.handle_call_me_maybe(cm);
+                        }
+                    });
+                true
+            }
+        }
+    }
+
+    /// di is the DiscoInfo of the source of the ping.
+    /// derp_node_src is non-zero if the ping arrived via DERP.
```
dignifiedquire, line 749: outdated

iroh-net/src/magicsock.rs
```diff
@@ -241,42 +238,777 @@ struct Inner {
...
+                    .last_ping_time
+                    .map(|s| s.elapsed() < Duration::from_secs(5))
+                    .unwrap_or_default();
+            di.last_ping_from.replace(src_addr);
+            di.last_ping_time.replace(Instant::now());
+            (di.node_key, likely_heart_beat)
+        };
+        debug_assert_eq!(sender, &node_key);
+
+        // If we got a ping over DERP, then derp_node_src is non-zero and we reply
```
dignifiedquire, line 766: outdated

iroh-net/src/magicsock.rs
```diff
@@ -677,6 +1387,15 @@ struct DiscoInfo {
     last_ping_time: Option<Instant>,
 }
 
+#[derive(Debug)]
+enum DerpRecvResult {
+    Error(io::Error),
+    Ok {
+        meta: quinn_udp::RecvMeta,
+        bytes: Bytes,
+    },
```
dignifiedquire, line 1396: would be nicer to make this an actual `Result`

```rust
type DerpRecvResult = io::Result<(quinn_udp::RecvMeta, Bytes)>;
```

iroh-net/src/magicsock/endpoint.rs
```diff
@@ -260,12 +259,21 @@ impl Endpoint {
                 }
             }
 
+            let addr = pong.from.as_socket_addr();
+            let trust_best_addr_until = pong.pong_at + Duration::from_secs(60 * 60);
```
dignifiedquire, line 263: we should really change that number (but lets do that in a follow up)

iroh-net/src/magicsock/endpoint.rs
```diff
@@ -1143,6 +1166,152 @@ pub(super) struct PeerMap {
...
+        self.inner.lock().node_count()
+    }
+
+    pub(super) fn write<T>(&self, f: impl FnOnce(&mut PeerMapInner) -> T) -> T {
+        let mut inner = self.inner.lock();
+        f(&mut inner)
+    }
+
+    pub(super) fn read<T>(&self, f: impl FnOnce(&PeerMapInner) -> T) -> T {
+        let inner = self.inner.lock();
```
dignifiedquire, line 1203: would it be worth it to use a `RwLock` instead of a `Mutex`?

dignifiedquire (approved):
some small comments left, but other than, looks good

iroh-net/src/magicsock.rs
rklaehn: If this is always or frequently a single Bytes we could reach for mallvec for such issues.

iroh-net/src/magicsock.rs
Frando: Sounds good. Let's defer to that then for now.

iroh-net/src/magicsock.rs
Frando: did that

PR #1580, opened 2023-10-05
perf(iroh-net): simplify stun::is check

dignifiedquire:
This is run on every packet, and showed up in some profiling as not being as cheap as it should be

### DIFF
--- a/iroh-net/src/stun.rs
+++ b/iroh-net/src/stun.rs
@@ -65,13 +65,15 @@ pub fn response(tx: TransactionId, addr: SocketAddr) -> Vec<u8> {
     buffer
 }
 
+// Copied from stun_rs
+// const MAGIC_COOKIE: Cookie = Cookie(0x2112_A442);
+const COOKIE: [u8; 4] = 0x2112_A442u32.to_be_bytes();
+
 /// Reports whether b is a STUN message.
 pub fn is(b: &[u8]) -> bool {
-    let cookie: [u8; 4] = b[4..8].try_into().unwrap();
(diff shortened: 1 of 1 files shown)


iroh-net/src/stun.rs
```diff
@@ -65,13 +65,15 @@ pub fn response(tx: TransactionId, addr: SocketAddr) -> Vec<u8> {
     buffer
 }
 
+// Copied from stun_rs
+// const MAGIC_COOKIE: Cookie = Cookie(0x2112_A442);
+const COOKIE: [u8; 4] = 0x2112_A442u32.to_be_bytes();
```
divagant-martian, line 70: why the need to copy the constant?

iroh-net/src/stun.rs
```diff
@@ -508,6 +510,11 @@ mod tests {
         assert_eq!(got_tx, tx);
     }
 
+    #[test]
+    fn test_stun_cookie() {
+        assert_eq!(stun_rs::MAGIC_COOKIE, COOKIE);
+    }
```
divagant-martian, line 516: we can do this in compile time

iroh-net/src/stun.rs
dignifiedquire: because I can't get it out of the library as constant time

iroh-net/src/stun.rs
dignifiedquire: the equality check/conversion is not constant time, so not sure how to do that

PR #1581, opened 2023-10-05
feat(iroh): use reflink if possible

rklaehn:
reflink-copy is a clone of the abandoned reflink crate. Last time there were some failures on android, so let's see if it works now...

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1910,6 +1910,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "ioctl-sys"
+version = "0.8.0"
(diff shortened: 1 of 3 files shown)


rklaehn:
If the tests pass on all platforms, should we enable this by default?

dignifiedquire:
Seems very reasonable to enable by default. Did you try this out manually if this works as expected?

rklaehn:
Tried last time. Worked well. Didn't try this time yet.

dignifiedquire (approved):

PR #1585, opened 2023-10-06
Gc tests

rklaehn:
## Description

Add a number of basic tests as well as a stress test to gc

## Questions

I had some consistent windows failures on Friday. But now it seems to work on windows. Not sure what is going on - was windows CI just broken on Friday?

## Missing (this or next PR):

- [ ] test that docs are protected
- [ ] actually change gc default to enabled

## Change checklist

- [x] Self-review.
- [x] ~Documentation updates if relevant.~
- [x] Tests if relevant.


iroh-bytes/src/util.rs
```diff
@@ -124,6 +130,18 @@ pub enum SetTagOption {
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
 pub struct HashAndFormat(pub Hash, pub BlobFormat);
 
+impl HashAndFormat {
+    /// Create a new hash and format pair, using the default (raw) format.
+    pub fn raw(hash: Hash) -> Self {
```
rklaehn, line 135: just for convenience

iroh/src/baomap/flat.rs
```diff
@@ -895,42 +895,65 @@ impl Store {
         Ok((tag, size))
     }
 
+    fn import_bytes_sync(&self, data: Bytes, format: BlobFormat) -> io::Result<TempTag> {
```
rklaehn, line 898: Use finalize_import_sync from import_bytes_sync for DRY

iroh/src/baomap/flat.rs
```diff
@@ -895,42 +895,65 @@ impl Store {
...
+            std::fs::write(&temp_outboard_path, outboard)?;
+            Some(temp_outboard_path)
+        } else {
+            None
+        };
+        // before here we did not touch the complete files at all.
+        // all writes here are protected by the temp tag
+        let complete_io_guard = self.0.complete_io_mutex.lock().unwrap();
+        // move the data file into place, or create a reference to it
+        let new = match file {
```
rklaehn, line 945: This was a bug before. You *must* pass through the format.

if somebody adds a hashseq using import_stream or import_reader, the resulting temp tag *must still* be with format HashSeq.

iroh/tests/gc.rs
```diff
@@ -0,0 +1,440 @@
+#![cfg(all(feature = "mem-db", feature = "iroh-collection"))]
```
dignifiedquire, line 1: What about also running the tests with the `fs` db? Would probably be good to make sure deletions there work similarly good

iroh/tests/gc.rs
```diff
@@ -0,0 +1,440 @@
...
+    baomap::mem::Store,
+) {
+    let rt = test_runtime();
+    let bao_store = baomap::mem::Store::new(rt.clone());
+    let node = wrap_in_node(bao_store.clone(), rt).await;
+    (node, bao_store)
+}
+
+async fn step() {
+    tokio::time::sleep(Duration::from_millis(100)).await;
```
dignifiedquire, line 71: why not use tokios timer https://docs.rs/tokio/latest/tokio/time/fn.pause.html#auto-advance testing functionality, this actually worked quite well for me when testing other timer based behaviour?

iroh/tests/gc.rs
rklaehn: The "gc_flat_stress" and "gc_flat_partial" tests are using the flat file db.

The idea was to use a "perfect" db in the smoke tests to see if the gc_loop / driver is implemented correctly, and then use db specific tests like gc_flat_xxx to make sure that the db actually works.

I guess I can make the tests generic.

iroh/tests/gc.rs
rklaehn: I hate it.

It depends on the notion that every single place in your code base and your dependencies that deals with time is using the tokio stuff, otherwise it will break in weird ways. Have a single place in your own code or your library code where you compute differences between `SystemTime`, and it is going to fail. But not outright fail. Fail in a subtle way.

It's another of these attempts of the tokio people to make tokio the default stdlib of rust. Is Lennart Poettering of systemd a secret tokio contributor?

I had thought about having an event stream from the db that you could use to detect when a GC has been performed instead of a fixed delay. I did not do it bc I thought that would lock in the fact that we are using a certain GC method (mark and sweep). But I would prefer to do this over using that tokio stuff.

dignifiedquire: > Is Lennart Poettering of systemd a secret tokio contributor?

🤣

PR #1590, opened 2023-10-06
Allow using b3sum output to identify hashes

rklaehn:
part 1 of my secret masterplan to use hex encoding everywhere...

## Description

Allows you to use hex blake3 hashes without any `bafkr4i`

## Notes & open questions

Can we do this for the output as well?

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-bytes/src/util.rs
+++ b/iroh-bytes/src/util.rs
@@ -124,7 +124,7 @@ pub enum SetTagOption {
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
 pub struct HashAndFormat(pub Hash, pub BlobFormat);
 
-/// Hash type used throught.
+/// Hash type used throughout.
 #[derive(PartialEq, Eq, Copy, Clone, Hash)]
 pub struct Hash(blake3::Hash);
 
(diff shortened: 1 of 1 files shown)


iroh-bytes/src/util.rs
```diff
@@ -124,7 +124,7 @@ pub enum SetTagOption {
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
 pub struct HashAndFormat(pub Hash, pub BlobFormat);
 
-/// Hash type used throught.
+/// Hash type used througoht.
```
dignifiedquire, line 127: still wrong

iroh-bytes/src/util.rs
```diff
@@ -256,6 +256,14 @@ impl FromStr for Hash {
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         let sb = s.as_bytes();
+        if sb.len() == 64 {
```
dignifiedquire, line 259: does b3sum pad?

rklaehn: what would there be to pad? a 32 byte hash is always 64 hex chars, or not?

```
❯ head -c 10 /dev/urandom | b3sum
e93842b50651bacd7b51beed6f4549ae6cdba26e657c11c1874e145f4af41055  -
```

rklaehn: Since we allow all multiformats, I guess there could be some multiformat that is 64 byte long. But I think it is pretty unlikely (impossible?) to have a multiformat that is 64 byte long and all hex chars.

dignifiedquire: > 64 hex chars, or not 

ah yes I think so, it was base64 with its many variants of padding

rklaehn: that's exactly why hex is so nice...

b5 (approved):

PR #1593, opened 2023-10-06
fix(iroh-net): improve peermap storage

dignifiedquire:
- cleanup the peermap before storing
- add more cases when endpoint addrs are removed


dignifiedquire:
this needs more thought, this actually deletes too many endpoints

PR #1595, opened 2023-10-07
Add ability to optionally connect just by node id

rklaehn:
## Description

This adds a fn to the magicsock to connect just by node id. Since a node id is insufficient to connect, a configurable discovery mechanism is added that allows finding direct addresses or at least the region id for a node id.

An example exercising the discovery mechanism has been added in https://github.com/n0-computer/iroh-examples/pull/2

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.


Nuhvi:
I will do my best on Mainline, but ... please take your time :)

Nuhvi:
Worth also researching how private trackers in BitTorrent do rank peers or give them credit or whatever ... I have very little knowledge ther, but while Bao-tree gives you novel stuff, maybe a lot of the value comes from the convention measures Trackers have been using for years.

rklaehn:
> Worth also researching how private trackers in BitTorrent do rank peers or give them credit or whatever ... I have very little knowledge ther, but while Bao-tree gives you novel stuff, maybe a lot of the value comes from the convention measures Trackers have been using for years.

I think ranking is pretty straightforward. You do a probe and measure if there is an answer at all, and if there is you measure how fast. This will rank the peers as seen from the tracker. What more than "this peer probably has the data" can you do?

Of course this really does not prevent a malicious peer from answering quickly to the tracker and not answering at all to other peers...

rklaehn:
> I will do my best on Mainline, but ... please take your time :)

Yeah, this is just a POC. It would be good if we had a full story for content routing. Not everything has to be super polished, but there should be no showstoppers...

Nuhvi:
> What more than "this peer probably has the data" can you do?

I have no idea, but ... will read around and see if there was something that private trackers managed to do.

iroh-net/src/key.rs
```diff
@@ -116,8 +116,8 @@ impl<'de> Deserialize<'de> for PublicKey {
         D: serde::Deserializer<'de>,
     {
         if deserializer.is_human_readable() {
-            let s: &str = serde::Deserialize::deserialize(deserializer)?;
-            Self::from_str(s).map_err(serde::de::Error::custom)
+            let s = String::deserialize(deserializer)?;
```
dignifiedquire, line 119: why was this needed?

iroh-net/src/magic_endpoint.rs
```diff
@@ -412,6 +427,39 @@ impl MagicEndpoint {
...
+                    info,
+                };
+                self.add_node_addr(peer_addr)?;
+                self.msock.get_mapping_addr(node_id).await.ok_or_else(|| {
+                    anyhow!("Failed to retrieve the mapped address from the magic socket. Unable to dial node {node_id:?}")
+                })?
+            }
+        };
+
+        debug!("connecting to {}: (via {})", node_id, addr);
```
dignifiedquire, line 459: would be good to log how the addr resolution was done (for later debugging sessions..)

iroh-net/src/key.rs
rklaehn: This is unrelated to the discovery. Without this I had issues deserializing from toml using our toml lib. Deserializing as a `&str` is more restrictive than deserializing as a String since it is zero copy, so you can not do it from an owned String, but only from a &str.

iroh-net/src/magicsock/derp_actor.rs
```diff
@@ -502,6 +502,12 @@ impl DerpActor {
             // TODO: spawn
             f();
         }
+        // notify the actor that it should republish its addr info
```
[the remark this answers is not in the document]
rklaehn: I was not quite sure when this is triggered.

If this is when the own derp region is changed, this is relevant for the addr info and needs to trigger a publish. If it is just when you connect to another node, it is not relevant.

I need to hook up the publish every time anything about the own addr info changes.

iroh-net/src/magic_endpoint.rs
rklaehn: You mean just "was already in peer map" / "was using the discovery mechanism" or more fine grained? The former can be done as it is, for the latter you would have to change the discovery trait.

But - proper attribution of where you got discovery info would be a larger topic. E.g. what if you get the same info from multiple sources, like mdns and pkarr? Just pick one at random? Do you keep attribution in the peer map as well?

rklaehn: I added the simple option - every time discovery resolve it is used it will be logged at debug level - the discovery trait requires Debug.

iroh-net/src/magicsock/derp_actor.rs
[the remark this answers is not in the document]
dignifiedquire: yeah I don't think you need this here, publishing on endpoint changes should be enough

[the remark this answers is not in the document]
rklaehn: Removed it.

dignifiedquire (approved):

PR #1596, opened 2023-10-09
feat(iroh): show content as hex when utf8 fails

divagant-martian:
## Description

use hex encoding in the repl when decoding as utf8 fails

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/commands/sync.rs
+++ b/iroh/src/commands/sync.rs
@@ -354,14 +354,14 @@ impl AuthorCommands {
 /// Format the content. If an error occurs it's returned in a formatted, friendly way.
 async fn fmt_content(doc: &Doc, entry: &Entry, mode: DisplayContentMode) -> Result<String, String> {
     let read_failed = |err: anyhow::Error| format!("<failed to get content: {err}>");
-    let decode_failed = |_err: std::string::FromUtf8Error| "<invalid UTF-8>".to_string();
+    let encode_hex = |err: std::string::FromUtf8Error| format!("0x{}", hex::encode(err.as_bytes()));
 
     match mode {
         DisplayContentMode::Auto => {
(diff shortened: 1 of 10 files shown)


divagant-martian:
@dignifiedquire can you give this a go to see if this works for your needs?

dignifiedquire (approved):
yes, this helps a lot

iroh/src/commands/sync.rs
```diff
@@ -354,14 +354,14 @@ impl AuthorCommands {
 /// Format the content. If an error occurs it's returned in a formatted, friendly way.
 async fn fmt_content(doc: &Doc, entry: &Entry, mode: DisplayContentMode) -> Result<String, String> {
     let read_failed = |err: anyhow::Error| format!("<failed to get content: {err}>");
-    let decode_failed = |_err: std::string::FromUtf8Error| "<invalid UTF-8>".to_string();
+    let encode_hex = |err: std::string::FromUtf8Error| format!("0x{}", hex::encode(err.as_bytes()));
```
rklaehn, line 357: Not really a showstopper, but how do you distinguish between the actual text `0x1234` and the values 0x1234 formatted like this?

To make the representation unique so that you can do a roundtrip, you would have to add quotes to the text format.

divagant-martian: this is the reason why I asked if you should be able to explicitly say give me hex, but ultimately if your content is hex encoded and added as a value, it's valid utf8 and I doubt the distinction matters

divagant-martian: Another alternative would be to `utf8: texthere` and `hex: encodedhere`might be clearer

dignifiedquire (approved):

PR #1597, opened 2023-10-09
[wip] peermap cleanup take two

divagant-martian:
## Description

TBD

## Notes & open questions

I'm not convinced this is exactly what we want. If we think about the peer map as a connectionless connection manager then it makes sense to rapidly prune peers we have not heard from recently. If we think about this as an address book there is interest in keeping peers for a longer time, even if they are not used. 

For example, in the case of sync, peers are stored if they were recently used, but just their peer id. If the user shutdowns, does some work without joining the network and then comes back online, Iroh will have been running that time and will have pruned the contact info of the peers that where stored precisely for this case. 

Need to think more about this

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


dignifiedquire:
can we close this in favor of #1653 ?

dignifiedquire:
@divagant-martian please reopen if there is sth more you wanted to do here

PR #1599, opened 2023-10-09
fix(derper): update config to auto generate keys

Arqu:
## Description

Deploying a derper node is a bit cumbersome right now if you don't run with default config. You have to provide a `secret_key` but have no easy path to generate one so either you dance to create a fresh config and then modify it or need to use some external thing to generate keys for you. Either way it's cumbersome.

This just adds an extra handler to inject a secret key if not present already in the supplied config.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/src/bin/derper.rs
+++ b/iroh-net/src/bin/derper.rs
@@ -284,10 +284,24 @@ impl Config {
         if !path.as_ref().is_file() {
             bail!("config-path must be a valid toml file");
         }
-        let config_ser = tokio::fs::read_to_string(path)
+        let mut config_ser = tokio::fs::read_to_string(&path)
             .await
             .context("unable to read config")?;
-        let config = toml::from_str(&config_ser).context("unable to decode config")?;
+        let mut secret_key_injected = false;
+        if !config_ser.contains("secret_key") {
+            config_ser = format!(
+                r#"
+                secret_key = "{}"
+                {}"#,
+                SecretKey::generate(),
+                config_ser
+            );
+            secret_key_injected = true;
+        }
+        let config: Self = toml::from_str(&config_ser).context("unable to decode config")?;
+        if secret_key_injected {
+            config.write_to_file(path).await?;
+        }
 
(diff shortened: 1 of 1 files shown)


iroh-net/src/bin/derper.rs
```diff
@@ -284,10 +284,24 @@ impl Config {
         if !path.as_ref().is_file() {
             bail!("config-path must be a valid toml file");
         }
-        let config_ser = tokio::fs::read_to_string(path)
+        let mut config_ser = tokio::fs::read_to_string(&path)
             .await
             .context("unable to read config")?;
-        let config = toml::from_str(&config_ser).context("unable to decode config")?;
+        let mut secret_key_injected = false;
+        if !config_ser.contains("secret_key") {
```
dignifiedquire, line 291: I would strongly prefer to inject this after toml parsing, there is actually a nice way to do this. 

```rust
#[serde(default = "SecretKey::generate")]
secret_key: SecretKey,
```

will do exactly that.

Ref https://serde.rs/field-attrs.html#default

Arqu: I swear serde is more powerful than any AI we have so far :D

iroh-net/src/bin/derper.rs
```diff
@@ -284,10 +285,13 @@ impl Config {
...
             bail!("config-path must be a valid toml file");
         }
-        let config_ser = tokio::fs::read_to_string(path)
+        let config_ser = tokio::fs::read_to_string(&path)
             .await
             .context("unable to read config")?;
-        let config = toml::from_str(&config_ser).context("unable to decode config")?;
+        let config: Self = toml::from_str(&config_ser).context("unable to decode config")?;
+        if !config_ser.contains("secret_key") {
+            config.write_to_file(path).await?;
```
dignifiedquire, line 293: one last complain, lets add a log line for this event

dignifiedquire (approved):

PR #1601, opened 2023-10-09
refactor(iroh-net): store a single pong

divagant-martian:
## Description

There seems to be no reason why we should store more than one `PongReply` since only the last is ever used

## Notes & open questions

If at some point we do need to check back on pong history this can be easily done with a `VecDeque`

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.


dignifiedquire (approved):
lgtm for now, if we ever need more we can add them back

PR #1603, opened 2023-10-09
perf: bring default scale of tokio runtime back

dignifiedquire:
I believe this should be removed and the default of num_cpus from tokio should be here

### DIFF
--- a/iroh/src/main.rs
+++ b/iroh/src/main.rs
@@ -12,7 +12,6 @@ use crate::commands::Cli;
 fn main() -> Result<()> {
     let rt = tokio::runtime::Builder::new_multi_thread()
         .thread_name("main-runtime")
-        .worker_threads(2)
         .enable_all()
         .build()?;
     rt.block_on(main_impl())?;


Arqu:
Yeah so this is purely theory time on my end from some past tpc experience.

The runtime spawned here doesn't really do all that *much* and giving it more threads just increases overhead while not really helping. The way things are set up is that a lot of the things now run in a local pool (see `L#28/29`) which are then used to spawn tasks on those. Worker threads being `2` here means we actually give it 2 threads to have a real multi threaded runtime but don't really need more than that, just avoiding pinning it all to a single worker thread.

What we should look into is the general usage of spawn around the codebase. Given the setup we should rarely do `rt.main().spawn` unless we really want it at the top level. Same goes for `tokio::spawn || tokio::task::spawn`. 
Doing a quick search we actually have quite a bit of those around the codebase. Having stuff block there might negatively impact the rest. I see a lot of `tokio::task::spawn` in `magicsock.rs`, see a `rt.main().spawn_blocking` for the metrics service (though spawn blocking should not create issues with tokio) etc.

dignifiedquire:
this needs more thought

PR #1606, opened 2023-10-10
fix(iroh-net): reverse ip-port mapping stores only direct addresses in the peermap

divagant-martian:
## Description

We were storing reverse mappings for `SendAddr::Derp` in the `PeerMap`. This means basically treating the DERP as a peer itself. We concluded this is something we shouldn't do. To fix the issue, the better solution is to enforce it via the type system. As such, we now only do reverse mapping from socket addresses.

As a note, I added a new type `IpPort`. Using `SocketAddr`s as key to hashmaps often goes wrong with ipv6 because of the flow control & class fields it contains. These fields can't be assumed stable even within the same connection: sending a the socket address made from an (ipv6, port) and receiving a response can make it look like the socket address is different (same ip, same port, different flow labels) Ignoring these fields has some limitations in extremely exceptional cases (link local addresses with a single interface) As a note, ignoring these fields is what tailscale does (all they store is an ip and a port). This makes communications over ipv6 work better in the common case so I think it's worth doing

## Notes & open questions

Some references
- flow label [RFC 2460 section 6](https://datatracker.ietf.org/doc/html/rfc2460#section-6)
- traffic class [RFC 2460 section 7](https://datatracker.ietf.org/doc/html/rfc2460#section-7)
- scope id [RFC 4007](https://www.rfc-editor.org/rfc/rfc4007)

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/src/magicsock.rs
+++ b/iroh-net/src/magicsock.rs
@@ -1295,7 +1295,7 @@ impl Actor {
                     derp_region: Some(region_id),
                     active: true,
                 });
-                self.peer_map.set_endpoint_for_ip_port(&ipp, id);
+                // TODO(@divma): removed here
                 let ep = self.peer_map.by_id_mut(&id).expect("inserted");
                 ep.quic_mapped_addr
             }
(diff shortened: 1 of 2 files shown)


iroh-net/src/magicsock.rs
```diff
@@ -2268,48 +2268,31 @@ impl Actor {
...
-                if let Some(ep) = self.peer_map.endpoint_for_node_key_mut(&di.node_key) {
-                    if ep.add_candidate_endpoint(src, dm.tx_id) {
-                        debug!("disco: ping got duplicate endpoint {} - {}", src, dm.tx_id);
-                        return;
-                    }
-                    (di.node_key, true)
-                } else {
-                    (di.node_key, false)
-                }
+                assert!(!src.is_derp());
```
dignifiedquire, line 2283: Making this an assert was probably a bad idea, maybe we can make it a `debug_assert` and print an error if this happens?

dignifiedquire (approved):

iroh-net/src/magicsock.rs
divagant-martian: done in df9e1a4 !

Issue #1576, opened 2023-10-05
clean up  output for `node connections`


ramfox:
1) because we keep track of peer addresses between sessions, we now show information on peers that we aren't actually connected to
2) relay connections never seem to show any latencies

Also, sometimes one side of the connection will show that there is a direct conn and the other side will believe that there is only a relay connection, need to make sure that if this a bug it's fixed.

dignifiedquire:
> This means we never actually send pings to the derp regions, so we never get a latency for the region.

netcheck does this I believe

PR #1607, opened 2023-10-11
fix(iroh-net): enforce storing a single derp region per peer

divagant-martian:
## Description

The Endpoint contained `SendAddr` instead of `ip`, `port` pairs. This is problematic for a couple of reasons:
- it makes it harder to reason about direct addresses and relay addresses
- makes it harder to properly update the derp-related information
- in principle, we shouldn't be storing multiple derp regions per peer

So this PR enforces storing a single derp region per peer. Some relevant note with these changes:

## Notes & open questions

- this proves we never ping a peer via their relay address. I left some TODOs that I intend to address in a later PR when it's clear how we should be doing this
- I also noticed updating a peer's region is different between "updating from node" to receiving a ping. Maybe this is the right thing to do but it's also peculiar. If a peer pinged us via relay we update their derp region, if we add a ticket or receive info from a gossip peer exchange we only update the region if we don't have any. To me this makes sense, but again, it draws my attention. I left a comment about it in the code
Basically TLDR, next step is adding ping via relays with some appropriate tests. This should at a minimum solve the second part of #1576 

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/src/magicsock.rs
+++ b/iroh-net/src/magicsock.rs
@@ -1295,7 +1295,7 @@ impl Actor {
                     derp_region: Some(region_id),
                     active: true,
                 });
-                self.peer_map.set_endpoint_for_ip_port(&ipp, id);
+                // TODO(@divma): removed here
                 let ep = self.peer_map.by_id_mut(&id).expect("inserted");
                 ep.quic_mapped_addr
             }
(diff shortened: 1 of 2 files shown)


dignifiedquire:
this looks pretty good, what is missing for this to be ready?

divagant-martian:
@dignifiedquire was built on top of #1606 so was waiting for that. Otherwise I think it's ready. I'll update the description with the relevant info

dignifiedquire (approved):

PR #1609, opened 2023-10-11
Fix flakiness of gc_hashseq test

rklaehn:
## Description

The issue was that a temp tag was not inserted into the live set. This is not a problem if it happens outside a gc. But if it happens right between a gc mark phase and a gc sweep phase, the temp pin will not be considered, and the data associated with the temp pin might be deleted.

I knew that this had to be done, and I think I even did it at some point, but it must have been lost in some refactoring.

## Notes & open questions

There are a number of places where temp pins are not used. See https://github.com/n0-computer/iroh/issues/1611 . But I think this should be done in a separate PR since this definitely fixes something.

This increases the gc frequency so much that gc is running continuously with no pause whatsoever for some tests. That was the only way to get the failure mentioned above at least semi-reliably.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/baomap/flat.rs
+++ b/iroh/src/baomap/flat.rs
@@ -1030,7 +1030,7 @@ impl Store {
     fn delete_sync(&self, hash: Hash) -> io::Result<()> {
         let mut data = None;
         let mut outboard = None;
-        let mut external = None;
+        let mut paths = None;
         let mut partial_data = None;
         let mut partial_outboard = None;
         let complete_io_guard = self.0.complete_io_mutex.lock().unwrap();
@@ -1043,7 +1043,7 @@ impl Store {
                 outboard = Some(self.owned_outboard_path(&hash));
             }
             if !entry.external.is_empty() {
-                external = Some(self.0.options.paths_path(hash));
+                paths = Some(self.0.options.paths_path(hash));
             }
         }
         if let Some(partial) = state.partial.remove(&hash) {
@@ -1056,16 +1056,19 @@ impl Store {
         state.data.remove(&hash);
         drop(state);
         if let Some(data) = data {
+            tracing::debug!("deleting data {}", data.display());
             if let Err(cause) = std::fs::remove_file(data) {
                 tracing::warn!("failed to delete data file: {}", cause);
             }
         }
-        if let Some(external) = external {
+        if let Some(external) = paths {
+            tracing::debug!("deleting paths file {}", external.display());
             if let Err(cause) = std::fs::remove_file(external) {
-                tracing::warn!("failed to delete partial outboard file: {}", cause);
+                tracing::warn!("failed to delete paths file: {}", cause);
             }
         }
         if let Some(outboard) = outboard {
+            tracing::debug!("deleting outboard {}", outboard.display());
             if let Err(cause) = std::fs::remove_file(outboard) {
                 tracing::warn!("failed to delete outboard file: {}", cause);
             }
--- a/iroh/src/baomap/flat.rs
+++ b/iroh/src/baomap/flat.rs
@@ -764,14 +764,6 @@ impl baomap::Store for Store {
     }
 
     fn temp_tag(&self, tag: HashAndFormat) -> TempTag {
-        let mut state = self.0.state.write().unwrap();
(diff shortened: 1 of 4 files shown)


rklaehn:
This is far from being done, there is also this: https://github.com/n0-computer/iroh/issues/1611

But it does fix the test I hope.

iroh/src/baomap/mem.rs
```diff
@@ -556,6 +556,14 @@ impl baomap::Store for Store {
...
     fn temp_tag(&self, tag: HashAndFormat) -> TempTag {
+        let mut state = self.0.state.write().unwrap();
+        // add to live set immediately. This prevents this blob from being deleted
+        // if this happens between a mark and a sweep phase.
+        state.live.insert(tag.0);
+        // increase the ref count
+        let entry = state.temp.entry(tag).or_default();
+        // panic if we overflow an u64
+        *entry = entry.checked_add(1).unwrap();
         TempTag::new(tag, Some(self.0.clone()))
```
dignifiedquire, line 567: So the above seems to be the same as the code in `on_clone`. But the `clone` called here is not triggering the callback, as you implemented `LivenessTracker` on `Inner`, not on `Arc<Inner`. Why do this duplication/separation?

iroh/src/baomap/flat.rs
```diff
@@ -763,8 +763,16 @@ impl baomap::Store for Store {
...
+    fn temp_tag(&self, tag: HashAndFormat) -> TempTag {
+        let mut state = self.0.state.write().unwrap();
+        // add to live set immediately. This prevents this blob from being deleted
+        // if this happens between a mark and a sweep phase.
+        state.live.insert(tag.0);
+        // increase the ref count
+        let entry = state.temp.entry(tag).or_default();
+        // panic if we overflow an u64
+        *entry = entry.checked_add(1).unwrap();
+        TempTag::new(tag, Some(self.0.clone()))
```
dignifiedquire, line 775: same `clone` and `on_clone` issue potentially as in the mem version

iroh/src/baomap/mem.rs
rklaehn: Sigh.

Yes, the code in on_clone does the exact same thing, except for adding the hash to the live set. I could just remove this and do the refcount increase in TempTag::new(), but then I would have to grab the mutex twice.

I think this is how I introduced this bug in the first place. I tried to DRY creation and refcount increase and forgot to copy over the live set insertion...

Any idea how to DRY this without needlessly double locking?

rklaehn: The things I tried to DRY these things ended up making things worse. E.g. you could have a fn on the state (where you already have the &mut) that does the refcount increase. But is that really worth having another fn or another struct to wrap the temp table?

dignifiedquire: I am less worried about the DRYness of the code, more about this potentially being done twice due to `Arc::clone` vs `Inner::clone`

rklaehn: If you were to increase the refcount by 2 on construction, that data could never be erased because it would not go back to 0. It would immediately be caught by the tests.

But I got another idea. Maybe not have the temp tags in the live set, but just check both sets in is_live? Will try that.

rklaehn: Nah, that does not work. Temp has a HashAndFormat as key, and is_live requires just a hash, so the test would have to be like this:

```
        let res = state.live.contains(hash)
            || state.temp.contains_key(&HashAndFormat::raw(*hash))
            || state.temp.contains_key(&HashAndFormat::hash_seq(*hash));
```

I mean, it works...

dignifiedquire: wouldn't it make more sense to store the raw hash in the set and the format in the value part? inclusion checks the format shouldn't matter asfaiu

rklaehn: for the temp tags it does matter if it is hashseq or raw. a temp tag with a format of hashseq protects the children too.

What you could do is to make the temp map a map from hash to (raw_count, hash_seq_count). But you need separate counts.

dignifiedquire (approved):

PR #1612, opened 2023-10-11
refactor(iroh-sync): Add actor to iroh-sync, remove deadlocks

Frando:
## Description

* Add an actor to `iroh_sync` that processes async requests for operations on Replicas
* Major refactoring of how we access the iroh sync API from iroh: Before this our `live::Actor` would easily deadlock, because inserting entries from gossip, and processing insert events coming back from the replica, happened in the same actor loop. Also, because iroh-sync is fully synchronous, we have to take care to not lock or block the async executor. So, to solve both of these issues, all operations on replicas and the iroh sync store are moved into a separate actor which runs in a loop on a `std::thread` and just blocks while waiting for actions from a channel. This, in turn, moves some state keeping from iroh to iroh-sync, which is a good thing I think.
* Remove Clone and all locks from `Replica`. As they are now usually owned by the `iroh_sync::actor` actor, no need for locks. We can get unique mut refs instead.

* Change the subscription model on `Replica`s to be a `Vec<flume::Sender<Event>>`. This now allows to directly subscribe to replica events from multiple places. This allows us to have subscription to replica events from the client not go through any actor. The replica events are merged together with the events from the `sync_engine::live` actor, and that's it. Should improve latency and ease work on the actors. Because we merge the channels for replica events and live actor events for client subscriptions, this means, that closing a replica from the client does not end the subscription stream, because the subscription on the live actor will remain active. To end the subscription, the client would need to drop the receiver. I think this is fine for now. What we might want to do instead is to split the subscription from the client to two methods and channels, `subscribe_insert_events` and `subscribe_sync_events` or so, then we could close them separately. However the one for sync events would still stay open indefinitely, because you wouldn't want to drop and recreate, I think, in case you do `join` / `leave` / `join`. Or do we?

* Track the open state of Replicas in the `iroh_sync::actor`. I opted for a simple implementation to start: The actor counts calls to `open` and decrements on calls to `close` and closes the replica once the count reaches zero. This works fine, however for replicas opened from the RPC client, because there's no async drop, I spawn a tokio task in drop to send the close call to the node. This means that if a client is force-killed, the replica would remain open indefinitely. It would be better to solve this cleaner - the only idea I had so far was to give out something like `ReplicaDescriptor`s on `open` and then send require regular keep-alive calls from the RPC client. I think it's fine to defer this change to a followup (which will be straightforward with the architecture in place now) because the impl in this PR is already quite an improvement over the state in `main` or #1612, where we don't ever close replicas.
* event subscriptions for replica insert events are now handled directly in the `iroh_sync::actor`, which is much nicer IMO. For the RPC subscription this event stream is merged with a subscription for sync events from the `iroh::sync_engine::live` actor. 

## Notes & open questions

Replicas may remain open indefinitely if the RPC client dies without calling `Doc::close`. This should be fixed in a followup and will need work in quic-rpc to get a notification once connections close.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh-gossip/src/net.rs
```diff
@@ -345,10 +348,14 @@ struct Actor {
 
 impl Actor {
     pub async fn run(mut self) -> anyhow::Result<()> {
+        let mut i = 0;
```
dignifiedquire, line 351: ❤️  I know the feeling so well when you have to add a counter to your actor loops

iroh-sync/src/net.rs
```diff
@@ -103,11 +152,55 @@ where
...
+        peer = %peer.fmt_short(),
+        namespace = %namespace.fmt_short(),
+        ?t_process,
+        ?t_connect,
+        sent = %progress.num_sent,
+        recv = %progress.num_recv,
+        "sync[accept]: done, success"
+    );
+
+    let timings = Timings {
```
Frando, line 169: We can remove this again also. Added for debugging purposes. But might be useful nevertheless..

iroh-sync/src/ranger.rs
```diff
@@ -338,7 +346,9 @@ where
 
             // Store incoming values
             for (entry, content_status) in values {
+                println!("NOW VALIDATE");
```
dignifiedquire, line 349: I think this needs to get cleaned up ;)

iroh-sync/src/sync.rs
```diff
@@ -1377,6 +1409,8 @@ mod tests {
     ) -> Result<()> {
         let alice_peer_id = [1u8; 32];
         let bob_peer_id = [2u8; 32];
+        let mut alice_state = SyncOutcome::default();
+        let mut bob_state = SyncOutcome::default();
```
dignifiedquire, line 1413: should test that the outcome is the expected

iroh/src/sync_engine/live.rs
```diff
@@ -162,65 +166,104 @@ fn entry_to_content_status(entry: EntryStatus) -> ContentStatus {
     }
 }
 
-/// Handle to a running live sync actor
+/// Handle to a https://pad.riseup.net/p/Z3Y9dpq19FjgwQSFGvbLrunning live sync actor
```
dignifiedquire, line 169: I don't think this belongs here

iroh/src/sync_engine/live.rs
```diff
@@ -162,65 +166,104 @@ fn entry_to_content_status(entry: EntryStatus) -> ContentStatus {
...
-    task: Shared<BoxFuture<'static, Result<(), Arc<JoinError>>>>,
+pub struct LiveSync {
+    sync_actor_tx: mpsc::Sender<ToSyncActor>,
+    task: Shared<BoxFuture<'static, ()>>,
 }
 
-impl<S: store::Store> LiveSync<S> {
-    /// Start the live sync.
+impl LiveSync {
+    /// Start the live sync.https://pad.riseup.net/p/Z3Y9dpq19FjgwQSFGvbL
```
dignifiedquire, line 177: more riseup pads?

iroh/src/sync_engine/rpc.rs
```diff
@@ -180,6 +180,7 @@ impl<S: Store> SyncEngine<S> {
         let tag = bao_store
             .import_bytes(value.into(), BlobFormat::RAW)
             .await?;
+        // TODO: Use sync handle
```
dignifiedquire, line 183: alle the store operations probably should? that way there is no need anymore for the spawn_blocking calls either

iroh-sync/src/net.rs
dignifiedquire: I like it, alternatively we could make these into metrics

iroh/src/sync_engine/live.rs
Frando: fixed

iroh/src/sync_engine/rpc.rs
Frando: Yes, that's the plan I think. Will do.

iroh/src/sync_engine/live.rs
Frando: fixed.

iroh-sync/src/ranger.rs
Frando: fixed

iroh/src/sync_engine/rpc.rs
Frando: Actually, I'm not that sure anymore: Redb is basically a global rwlock, correct?
By passing everything through a single channel onto a single thread, we basically make that a mutex. So no concurrent reads can happen on the database. Whereas before, with transforming the sync interators into streams with `spawn_blocking`, this could at least theoretically happen in parallel if no thread is writing at the time. Depending on how many threads the blocking pool has.

So do we want a threadpool of our own here in the end?

Maybe these thoughts are premature, and it makes sense architecture wise to have all interactions go through the channel interface, to have a clean point between sync and async interactions. And we could then still add a threadpool for non-mutating access to the replica later...

iroh-sync/src/sync.rs
Frando: :+1:  Added to always verify that num_sent matches num_recv from the other peer. Also added an explicit check to one of the tests, `test_replica_sync`. Could do more, but maybe in a followup.

iroh/src/sync_engine/rpc.rs
dignifiedquire: my initial sense would be for now to make everything go to the actor, and potentially parallelise in there if possible later down the road

iroh-sync/src/actor.rs
```diff
@@ -0,0 +1,654 @@
...
+    states: ReplicaStates<S>,
+    event_tx: flume::Sender<Event>,
+    action_rx: flume::Receiver<Action>,
+    content_status_callback: Option<ContentStatusCallback>,
+}
+
+impl<S: store::Store> Actor<S> {
+    fn run(&mut self) -> Result<()> {
+        loop {
+            let Ok(action) = self.action_rx.recv() else {
```
dignifiedquire, line 415: could be written as `while let Ok(action)`

iroh-sync/src/actor.rs
```diff
@@ -0,0 +1,654 @@
...
+            }
+            Action::ImportReplica { namespace, reply } => {
+                let id = namespace.id();
+                send_reply(reply, self.store.new_replica(namespace).map(|_| id))
+            }
+            Action::ListAuthors { reply } => iter_to_channel(
+                reply,
+                self.store
+                    .list_authors()
+                    .map(|a| a.map(|a| a.map(|a| a.id()))),
```
dignifiedquire, line 451: so you like `map`s ? ;)

iroh/src/downloader/get.rs
```diff
@@ -223,9 +223,6 @@ pub async fn get<D: Store>(
         BlobFormat::Raw => get_blob(db, conn, &hash).await,
         BlobFormat::HashSeq => get_hash_seq(db, conn, &hash).await,
     };
-    if let Err(e) = stats.as_ref() {
-        tracing::error!("get failed: {e:?}");
-    }
```
dignifiedquire, line 228: why is this removed?

iroh/src/sync_engine.rs
```diff
@@ -2,106 +2,246 @@
...
-    pub(crate) live: LiveSync<S>,
+    pub(crate) sync: SyncHandle,
+    to_live_actor: mpsc::Sender<ToLiveActor>,
+    tasks_fut: Shared<BoxFuture<'static, ()>>,
+
+    // TODO:
+    // After the latest refactoring we don't need the store here anymore because all interactions
+    // go over the [`SyncHandle`]. Removing the store removes the `S: Store` generic from the
+    // `SyncEngine`, in turn removing the `S: Store` generic from [`iroh::node::Node`]. Yay!
+    // As this changes the code in many lines, I'd defer it to a follwup.
```
dignifiedquire, line 50: 🎉

iroh/src/sync_engine/rpc.rs
```diff
@@ -236,17 +218,19 @@ impl<S: Store> SyncEngine<S> {
...
-        self.rt.main().spawn_blocking(move || {
-            let ite = store.get_many(doc_id, filter);
-            let ite = inline_result(ite).map_ok(|entry| DocGetManyResponse { entry });
-            for entry in ite {
-                if let Err(_err) = tx.send(entry) {
-                    break;
-                }
+        let sync = self.sync.clone();
+        // we need to spawn a task to send our request to the sync handle, because the method
+        // itself must be sync.
```
dignifiedquire, line 223: why must this function be sync?

iroh/src/sync_engine/rpc.rs
```diff
@@ -255,17 +239,9 @@ impl<S: Store> SyncEngine<S> {
             author,
             key,
         } = req;
-        let replica = self.get_replica(&doc_id)?;
-        let entry = self.store.get_one(replica.namespace(), author, key)?;
+        let entry = self.sync.get_one(doc_id, author, key).await?;
+        // let replica = self.get_replica(&doc_id)?;
+        // let entry = self.store.get_one(replica.namespace(), author, key)?;
```
dignifiedquire, line 244: delete?

iroh/tests/sync.rs
```diff
@@ -30,38 +38,55 @@ fn test_runtime() -> runtime::Handle {
...
-    Node::builder(db, store).runtime(&rt).bind_addr(addr)
+    Node::builder(db, store)
+        .secret_key(secret_key)
+        .derp_mode(DerpMode::Disabled)
+        .runtime(&rt)
+        .bind_addr(addr)
 }
 
-async fn spawn_node(
+fn spawn_node(
```
dignifiedquire, line 52: why?

iroh-sync/src/store/fs.rs
```diff
@@ -83,7 +82,7 @@ impl Store {
         // Setup all tables
         let write_tx = db.begin_write()?;
         {
-            let _table = write_tx.open_table(RECORDS_TABLE)?;
+            let _ = write_tx.open_table(RECORDS_TABLE)?;
```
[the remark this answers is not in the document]
Frando: leftover, will remove

iroh/src/downloader/get.rs
Frando: because it would log an error for each failed download attempt, while there might still be other peers queued, and I thought an `error!` log for each failed attempt is too much. if the final attempt fails, a `warn!` is logged already in `downloader.rs`.

iroh-sync/src/sync.rs
```diff
@@ -123,44 +135,40 @@ impl<S: ranger::Store<SignedEntry> + PublicKeyStore + 'static> Replica<S> {
...
-                Some(r)
+                *on_insert_sender = Some(sender);
+                true
             }
         }
     }
 
     /// Remove the subscription.
-    pub fn unsubscribe(&self) -> bool {
+    pub fn unset_event_sender(&self) -> bool {
```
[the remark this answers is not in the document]
Frando: because I thought it more fitting with `set_event_sender`. but this will be changed again in an upcoming commit, so let's see then :)

iroh/src/sync_engine/rpc.rs
Frando: because quic-rpc forces us to return `impl Stream`, not `impl Future`

iroh/tests/sync.rs
Frando: The function is not `async fn` so that we can take a `&mut` borrow on the `rng` without capturing that `&mut` lifetime in the returned future. This allows to call it in a loop while still collecting the futures before awaiting them alltogether (see [`spawn_nodes`])

Will add the explanation as a comment.

dignifiedquire (approved):

PR #1613, opened 2023-10-11
feat(iroh-sync): Sync propagation

Frando:
## Description

This PR implements sync propagation. 

* When a set-reconcilation sync completes, `iroh-sync` now returns a struct `AuthorHeads` which contains a map with an entry for each author for which we *received* entries in the sync operation, and the *latest timestamp* of all entries we received. So it is a kind of "cheapo vector clock", indicating "what's the latest entry we received in this sync op for each author"
* This `AuthorHeads` struct is then sent as a gossip message * to neighbors only*  as a `SyncReport`
* When receiving a sync report, a node checks if the `AuthorHeads` have any news to offer (i.e. if it contains a timestamp for an author newer than the latest existing entry for this author in the local store). If so, the node starts a sync request *to the node we received the report from*

Through this, the sync operations should propagate, and through checking the `AuthorHeads` if there are actually news for us loops should be avoided.

The PR also contains a much-needed refactoring for better state handling in the `live` actor.

## Notes & open questions

My biggest question is if that propagation model of sending the`AuthorHeads` to neighbors will scale in practice.

There is a test included that passes mostly, but sometimes not on CI (timeout of 60s reached before receiving all events). It really should not take 60s to sync 10 nodes together on localhost.

Some notes:
* For big sync operations with many messages we might want to limit the `AuthorHeads` to some number of authors
* We will want to debounce and merge sending sync reports over some interval
* We might want to delay sending sync reports, or not send sync reports for very recent entries,to not "race" with gossip

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-sync/src/heads.rs
+++ b/iroh-sync/src/heads.rs
@@ -28,9 +28,14 @@ impl AuthorHeads {
         self.heads.len()
     }
 
+    /// Whether this [`AuthorHeads`] is emty.
+    pub fn is_empty(&self) -> bool {
+        self.heads.is_empty()
+    }
+
     /// Get the timestamp for an author.
     pub fn get(&self, author: &AuthorId) -> Option<Timestamp> {
-        self.heads.get(author).map(|t| *t)
+        self.heads.get(author).copied()
     }
 
     /// Can this state offer newer stuff to `other`?
@@ -38,7 +43,7 @@ impl AuthorHeads {
         let mut updates = 0;
         for (author, ts_ours) in self.iter() {
             if other
-                .get(&author)
+                .get(author)
                 .map(|ts_theirs| *ts_ours > ts_theirs)
                 .unwrap_or(true)
             {
@@ -127,7 +132,7 @@ mod tests {
         let decoded = AuthorHeads::decode(&encoded)?;
         assert_eq!(decoded.len(), 6);
         let expected: AuthorHeads = (0u64..6)
-            .map(|n| (AuthorId::from(&[9 - n as u8; 32]), start + (9 - n as u64)))
+            .map(|n| (AuthorId::from(&[9 - n as u8; 32]), start + (9 - n)))
             .collect();
         assert_eq!(expected, decoded);
         Ok(())
@@ -135,13 +140,19 @@ mod tests {
 
     #[test]
     fn author_heads_compare() -> Result<()> {
-        let a = [(AuthorId::from(&[0u8; 32]), 5), (AuthorId::from(&[1u8; 32]), 7)];
-        let b = [(AuthorId::from(&[0u8; 32]), 4), (AuthorId::from(&[1u8; 32]), 6), (AuthorId::from(&[2u8; 32]), 7)];
+        let a = [
+            (AuthorId::from(&[0u8; 32]), 5),
+            (AuthorId::from(&[1u8; 32]), 7),
+        ];
+        let b = [
+            (AuthorId::from(&[0u8; 32]), 4),
+            (AuthorId::from(&[1u8; 32]), 6),
+            (AuthorId::from(&[2u8; 32]), 7),
+        ];
         let a: AuthorHeads = a.into_iter().collect();
         let b: AuthorHeads = b.into_iter().collect();
         assert_eq!(a.has_news_for(&b), NonZeroU64::new(2));
         assert_eq!(b.has_news_for(&a), NonZeroU64::new(1));
         Ok(())
-
     }
 }
--- a/iroh-sync/src/heads.rs
+++ b/iroh-sync/src/heads.rs
@@ -1,6 +1,6 @@
 //! Author heads
 
-use std::collections::BTreeMap;
+use std::{collections::BTreeMap, num::NonZeroU64};
 
 use anyhow::Result;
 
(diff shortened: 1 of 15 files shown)


iroh-sync/src/heads.rs
```diff
@@ -0,0 +1,53 @@
+//! Author heads
+
+use std::collections::BTreeMap;
+
+use serde::{Deserialize, Serialize};
+
+use crate::AuthorId;
+
+/// Timestamps of the latest entry for each author.
+// TODO: If we want to encode this in gossip messages we have to enforce a max length.
```
dignifiedquire, line 10: aren't these already sent on gossip messages?

iroh-sync/src/store.rs
```diff
@@ -132,6 +140,24 @@ pub trait Store: std::fmt::Debug + Clone + Send + Sync + 'static {
...
+            let latest = self.get_latest(namespace)?;
+            let mut heads = AuthorHeads::default();
+            for e in latest {
+                let (author, timestamp, _key) = e?;
+                heads.insert(author, timestamp);
+            }
+            heads
+        };
+        let has_news_for_us = heads.has_news_for(&our_heads);
+        Ok(has_news_for_us)
```
dignifiedquire, line 158: it might be interesting (at least for tracking purposes) to not just return `true` or `false`, but rather the number of keys that we have news for

iroh-sync/src/store.rs
```diff
@@ -132,6 +140,24 @@ pub trait Store: std::fmt::Debug + Clone + Send + Sync + 'static {
     /// Get all content hashes of all replicas in the store.
     fn content_hashes(&self) -> Result<Self::ContentHashesIter<'_>>;
 
+    /// Get the latest timestamp and key for each author in a namespace
+    fn get_latest(&self, namespace: NamespaceId) -> Result<Self::LatestIter<'_>>;
```
dignifiedquire, line 144: Isn't this the same as `get_all` without the values? Might be clearer to call this `get_all_keys()` or similar

iroh-sync/src/heads.rs
Frando: WDYM *these* here?

iroh-sync/src/store.rs
Frando: No, this only gets *one entry per author*, not *one entry per key*. Will try to improve the docs.

iroh/src/sync_engine/live.rs
```diff
@@ -546,35 +510,94 @@ impl<B: iroh_bytes::store::Store> LiveActor<B> {
...
+        }
+
+        let msg = match postcard::to_stdvec(op) {
+            Ok(msg) => msg,
+            Err(err) => {
+                error!(?err, ?op, "Failed to serialize message:");
+                return;
+            }
+        };
+        // TODO: We should debounce and merge these neighbor announcements likely.
```
dignifiedquire, line 563: yeah, definitely, if you have a fast producer this will trigger a lot of traffic

iroh/src/sync_engine/state.rs
```diff
@@ -0,0 +1,231 @@
...
+}
+
+/// Contains an entry for each active (syncing) namespace, and in there an entry for each peer we
+/// synced with.
+#[derive(Default)]
+pub struct NamespaceStates(HashMap<NamespaceId, NamespaceState>);
+
+#[derive(Default)]
+struct NamespaceState {
+    peers: HashMap<PublicKey, PeerState>,
```
dignifiedquire, line 54: maybe `BtreeMap` as suggested by @rklaehn for ordering stability?

iroh/src/sync_engine/state.rs
```diff
@@ -0,0 +1,231 @@
...
+impl Default for SyncState {
+    fn default() -> Self {
+        Self::Idle
+    }
+}
+
+/// Contains an entry for each active (syncing) namespace, and in there an entry for each peer we
+/// synced with.
+#[derive(Default)]
+pub struct NamespaceStates(HashMap<NamespaceId, NamespaceState>);
```
dignifiedquire, line 50: maybe `BtreeMap` as suggested by @rklaehn for ordering stability?

dignifiedquire (approved):
some notes, but overall looks good to me

iroh-sync/src/heads.rs
dignifiedquire: The "this" in the "TODO" 😅

iroh-sync/src/store.rs
dignifiedquire: ohh, I see, I thought this was per author & per key

iroh-sync/src/store.rs
Frando: Good idea. Changed it to return `Option<NonZeroU64>`. However it returns the number of *authors*, not keys, that there's updates for.

iroh/src/sync_engine/live.rs
Frando: We should. Let's do it in a followup though?

iroh-sync/src/heads.rs
Frando: Fixed in the latest commits. We now only encode the `n` most recent `(AuthorId, Timestamp)` pairs until we reach the gossip max message size.

iroh/src/sync_engine/live.rs
dignifiedquire: sounds good

dignifiedquire (approved):

Issue #1611, opened 2023-10-11
Integrate tagging with downloader


rklaehn:
I think the downloader is performing naked downloads, where a hash is not protected by a temp tag. This will have to be changed, otherwise a download could be deleted right when it is being downloaded.

The inner API should not allow this in the first place, maybe.

Also, the nomenclature needs to be changed, since there is a lot of mentions of Collections where now we just care about HashSeqs (where iroh collections are a special case of this)

CC @divagant-martian

PR #1615, opened 2023-10-11
Integrate tagging with downloader

rklaehn:
## Description

- Some code sharing between the mem and flat bao stores, and turning BlobFormat into an enum
- Create temp tags in the downloader and keep them alive while the download is ongoing

Fixes https://github.com/n0-computer/iroh/issues/1611

## Notes & open questions

Note: this looks worse than it is. I changed BlobFormat to an enum and changed HashAndFormat from a tuple struct to a normal struct. That is the majority of the changes.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


divagant-martian (approved):
from what I con on good conscience review, this lgtm!

PR #1617, opened 2023-10-11
fix(cli,console): default to 'hash' mode for the keys command

b5:
## Description
The purpose of the `doc keys` command is to list keys. Having a default of `auto` that shows content, which makes the output unreadable from a "fast review of the keys in this doc" use case:
Screenshot 2023-10-11 at 1 20 19 PM

one default change keeps both the option, and makes the output useful again:

Screenshot 2023-10-11 at 1 22 26 PM

This leaves `doc get` untouched, which should absolutely keep the `auto` default.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/commands/sync.rs
+++ b/iroh/src/commands/sync.rs
@@ -130,7 +130,7 @@ pub enum DocCommands {
         /// Optional key prefix (parsed as UTF-8 string)
         prefix: Option<String>,
         /// How to show the contents of the keys.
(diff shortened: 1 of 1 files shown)


divagant-martian (approved):
any makes sense depending on use case so I don't think there is a perfect solution :woman_shrugging: lgtm
in the future there might be interest on doing something similar with watch, which has no selection right now

Issue #1628, opened 2023-10-12
`doc list` should show all author entries for keys


ramfox:
we also need to decide how this affects `doc get KEY`, which should theoretically show multiple entries, one for each author

dignifiedquire:
should be controlled through queries once #1701 lands

ramfox:
closed by #1701

PR #1629, opened 2023-10-12
fix: `doc export` exports the latest entry at a given key

ramfox:
## Description

We currently only export via the latest key & the author id, but in the CLI and console, we don't currently have a way of viewing other authors. 

This needs to be fixed asap for demo purposes, and we should have a longer discussion about how we expose the other peers' author ids to the CLI and console: https://github.com/n0-computer/iroh/issues/1628

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh/src/client.rs
+++ b/iroh/src/client.rs
@@ -621,6 +621,27 @@ where
         Ok(res.entry.map(|entry| entry.into()))
     }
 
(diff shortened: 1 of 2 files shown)


divagant-martian (approved):
tried it out, it works. Checked the code, lgtm!

PR #1630, opened 2023-10-12
fix(iroh-sync): actually transfer newer entries for identical keys

Frando:
## Description

We had a bug that just was never tested for: When both peers have an entry with identical ranger keys (i.e. identical namespace, author and key), when peer A received an entry from peer B, it would never send their matching entry, even if their entry has a higher timestamp.

This PR fixes this! The actual fix is a single line: https://github.com/n0-computer/iroh/pull/1630/files#diff-be755583e5a892a8b9a6329e59cfc17e9633447fb7992db74eef670ba1508ccbR378

In addition, the PR adds both a manual test for this and a proptest strategy, which are both properly failing without the change.

It also renames the prefix methods on the ranger store trait that were introduced in #1535 to have a clearer naming (which helped me while tracking this down), and improves naming and logic around the prefix in other places.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


dignifiedquire (approved):
very nice catch

divagant-martian:
should the title be changed?

PR #1631, opened 2023-10-12
perf: improve derp connection establishment

dignifiedquire:
- make sure that TLS session resumption is possible, by caching the TLS config in the derper
- parallelise dials to derpers in the same region
- expand the `doctor derp-regions` command

### DIFF
--- a/iroh-net/src/derp/http/client.rs
+++ b/iroh-net/src/derp/http/client.rs
@@ -557,7 +557,7 @@ impl Client {
             ));
         }
 
-        warn!("tls handshake done: {:?}", start.elapsed());
+        trace!("tls handshake done: {:?}", start.elapsed());
         let start = Instant::now();
 
         debug!("starting upgrade");
@@ -569,7 +569,7 @@ impl Client {
             }
         };
 
-        warn!("derp upgrade done: {:?}", start.elapsed());
+        trace!("derp upgrade done: {:?}", start.elapsed());
         let start = Instant::now();
 
         debug!("connection upgraded");
@@ -586,7 +586,7 @@ impl Client {
                 .await
                 .map_err(|e| ClientError::Build(e.to_string()))?;
 
-        warn!("derp handshake: {:?}", start.elapsed());
+        trace!("derp handshake: {:?}", start.elapsed());
         let start = Instant::now();
         if *self.inner.is_preferred.lock().await && derp_client.note_preferred(true).await.is_err()
         {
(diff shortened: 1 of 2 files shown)


iroh-net/src/derp/http/client.rs
```diff
@@ -620,28 +645,49 @@ impl Client {
...
-            let conn = self.dial_node(node).await;
-            match conn {
-                Ok(conn) => return Ok((conn, node.clone())),
-                Err(e) => first_err = Some(e),
             }
         }
-        let err = first_err.unwrap();
-        Err(err)
+
+        Err(first_err.unwrap())
```
rklaehn, line 680: seems a bit arbitrary to just return the first error.

rklaehn (approved):

iroh-net/src/derp/http/client.rs
```diff
@@ -624,24 +640,44 @@ impl Client {
...
+                let target = target.clone();
+                async move {
+                    if node.stun_only {
+                        return Err(ClientError::StunOnlyNodesFound(target));
+                    }
+                    let conn = this.dial_node(&node).await;
+                    match conn {
+                        Ok(conn) => Ok((conn, node)),
+                        Err(e) => Err(e),
+                    }
```
flub, line 659: optional, but isn't this `this.dial_node(&node).await.map(|c| (c, node))`?

iroh-net/src/derp/http/client.rs
```diff
@@ -624,24 +640,44 @@ impl Client {
...
+        while let Some(res) = dials.next().await {
+            match res {
+                Ok((conn, node)) => {
+                    // return on the first successfull one
+                    trace!("dialed region");
+                    return Ok((conn, node));
+                }
+                Err(e) => {
+                    if first_err.is_none() {
+                        first_err = Some(e);
```
flub, line 674: related to @rklaehn 's comment below, unless you return a Vec of errors you'll always have an arbitrary choice.  Probably easier is to `tracing::error!()` log all the error cases here?  or at least `warn!`.

flub (approved):

iroh-net/src/derp/http/client.rs
dignifiedquire: fixed

iroh-net/src/derp/http/client.rs
dignifiedquire: fixed

iroh-net/src/derp/http/client.rs
dignifiedquire: logging them at least now

PR #1632, opened 2023-10-12
fix(iroh-net): ping via relay, enable relay ping in derp only mode

divagant-martian:
## Description

- pings the derp region when we ping the udp addresses
- handles pongs via derp
- enables ping and pongs via relay in derp only mode

## Notes & open questions

I left two nodes in my machine, in derp only mode running and they never became direct peers so I consider this a safe change wrt the derp only env flag

This fixes the second part of #1576 
Output example
```bash
> node connections
# node id                                               region  conn type  latency     
# xqcdy7bioujckakup5iicwfgppljpffbmelwssj4hngf5u3satga  1       relay      166ms,859μs 
```

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.


dignifiedquire (approved):
makes a lot of sense

PR #1640, opened 2023-10-16
refactor(iroh-sync): move replica event channels to iroh_sync, track open/close state of replicas

Frando:
## Description

This PR contains changes on top of #1612. Made a separate PR to ease reviewing. I'd merge this into #1612 after a review here, and then hopefully merge #1612 into main right after.

Has the following changes:

* Change the subscription model on `Replica`s to be a `Vec<flume::Sender<Event>>`. This now allows to directly subscribe to replica events from multiple places. This allows us to have subscription to replica events from the client not go through any actor. The replica events are merged together with the events from the `sync_engine::live` actor, and that's it. Should improve latency and ease work on the actors. Because we merge the channels for replica events and live actor events for client subscriptions, this means, that closing a replica from the client does not end the subscription stream, because the subscription on the live actor will remain active. To end the subscription, the client would need to drop the receiver. I think this is fine for now. What we might want to do instead is to split the subscription from the client to two methods and channels, `subscribe_insert_events` and `subscribe_sync_events` or so, then we could close them separately. However the one for sync events would still stay open indefinitely, because you wouldn't want to drop and recreate, I think, in case you do `join` / `leave` / `join`. Or do we?

* Track the open state of Replicas in the `iroh_sync::actor`. I opted for a simple implementation to start: The actor counts calls to `open` and decrements on calls to `close` and closes the replica once the count reaches zero. This works fine, however for replicas opened from the RPC client, because there's no async drop, I spawn a tokio task in drop to send the close call to the node. This means that if a client is force-killed, the replica would remain open indefinitely. It would be better to solve this cleaner - the only idea I had so far was to give out something like `ReplicaDescriptor`s on `open` and then send require regular keep-alive calls from the RPC client. I think it's fine to defer this change to a followup (which will be straightforward with the architecture in place now) because the impl in this PR is already quite an improvement over the state in `main` or #1612, where we don't ever close replicas.
* event subscriptions for replica insert events are now handled directly in the `iroh_sync::actor`, which is much nicer IMO. For the RPC subscription this event stream is merged with a subscription for sync events from the `iroh::sync_engine::live` actor. 

## Notes & open questions

included in description..

One more thought: Now that the `Replica`s are usually only used in the `iroh_sync::actor::Actor`, which is a single loop with `&mut` access, we could think about removing the `Clone` impl and allo locks from the `Replica`. If you wanted to use `Replica`s outside of the actor, you'd have to lock them yourself. Didn't think it through fully, but possibly worthwhile to investigate.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-sync/src/sync.rs
+++ b/iroh-sync/src/sync.rs
@@ -112,6 +112,9 @@ impl Subscribers {
     pub fn send(&mut self, event: Event) {
         self.0.retain(|sender| sender.send(event.clone()).is_ok())
     }
+    pub fn len(&self) -> usize {
+        self.0.len()
+    }
     pub fn send_with(&mut self, f: impl FnOnce() -> Event) {
         if !self.0.is_empty() {
             self.send(f())
@@ -182,7 +185,7 @@ impl<S: ranger::Store<SignedEntry> + PublicKeyStore + 'static> Replica<S> {
 
     /// Get the number of current event subscribers.
     pub fn subscribers_count(&self) -> usize {
-        self.inner.subscribers.lock().0.len()
+        self.inner.subscribers.lock().len()
     }
 
     /// Set the content status callback.
--- a/iroh-sync/src/sync.rs
+++ b/iroh-sync/src/sync.rs
@@ -23,7 +23,7 @@ use derive_more::Deref;
 #[cfg(feature = "metrics")]
 use iroh_metrics::{inc, inc_by};
 
-use parking_lot::RwLock;
+use parking_lot::{Mutex, RwLock};
 
 use ed25519_dalek::{Signature, SignatureError};
 use iroh_bytes::Hash;
@@ -100,6 +100,28 @@ pub struct SyncOutcome {
     pub num_sent: usize,
 }
 
+#[derive(Debug, Default)]
+struct Subscribers(Vec<flume::Sender<Event>>);
+impl Subscribers {
+    pub fn subscribe(&mut self, sender: flume::Sender<Event>) {
+        self.0.push(sender)
+    }
+    pub fn unsubscribe(&mut self, sender: &flume::Sender<Event>) {
+        self.0.retain(|s| !s.same_channel(sender));
+    }
+    pub fn send(&mut self, event: Event) {
+        self.0.retain(|sender| sender.send(event.clone()).is_ok())
+    }
+    pub fn send_with(&mut self, f: impl FnOnce() -> Event) {
+        if !self.0.is_empty() {
+            self.send(f())
+        }
+    }
+    pub fn clear(&mut self) {
+        self.0 = vec![];
+    }
+}
+
 /// Local representation of a mutable, synchronizable key-value store.
 #[derive(derive_more::Debug, Clone)]
 pub struct Replica<S: ranger::Store<SignedEntry> + PublicKeyStore> {
@@ -110,7 +132,7 @@ pub struct Replica<S: ranger::Store<SignedEntry> + PublicKeyStore> {
 struct InnerReplica<S: ranger::Store<SignedEntry> + PublicKeyStore> {
     namespace: Namespace,
     peer: RwLock<Peer<SignedEntry, S>>,
-    on_insert_sender: RwLock<Option<flume::Sender<Event>>>,
+    subscribers: Mutex<Subscribers>,
     #[debug("ContentStatusCallback")]
     content_status_cb: RwLock<Option<ContentStatusCallback>>,
     closed: AtomicBool,
@@ -124,7 +146,8 @@ impl<S: ranger::Store<SignedEntry> + PublicKeyStore + 'static> Replica<S> {
             inner: Arc::new(InnerReplica {
                 namespace,
                 peer: RwLock::new(Peer::from_store(store)),
-                on_insert_sender: RwLock::new(None),
+                subscribers: Default::default(),
+                // on_insert_sender: RwLock::new(None),
                 content_status_cb: RwLock::new(None),
                 closed: AtomicBool::new(false),
             }),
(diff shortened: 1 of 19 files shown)


iroh-sync/src/sync.rs
```diff
@@ -100,6 +100,28 @@ pub struct SyncOutcome {
...
+struct Subscribers(Vec<flume::Sender<Event>>);
+impl Subscribers {
+    pub fn subscribe(&mut self, sender: flume::Sender<Event>) {
+        self.0.push(sender)
+    }
+    pub fn unsubscribe(&mut self, sender: &flume::Sender<Event>) {
+        self.0.retain(|s| !s.same_channel(sender));
+    }
+    pub fn send(&mut self, event: Event) {
+        self.0.retain(|sender| sender.send(event.clone()).is_ok())
```
dignifiedquire, line 113: I don't think this should be `retain` unless I am missing sth

iroh-sync/src/sync.rs
```diff
@@ -100,6 +100,28 @@ pub struct SyncOutcome {
...
+    }
+    pub fn send_with(&mut self, f: impl FnOnce() -> Event) {
+        if !self.0.is_empty() {
+            self.send(f())
+        }
+    }
+    pub fn clear(&mut self) {
+        self.0 = vec![];
```
dignifiedquire, line 121: why not `self.0.clear()`?

iroh-sync/src/sync.rs
```diff
@@ -100,6 +100,28 @@ pub struct SyncOutcome {
...
+#[derive(Debug, Default)]
+struct Subscribers(Vec<flume::Sender<Event>>);
+impl Subscribers {
+    pub fn subscribe(&mut self, sender: flume::Sender<Event>) {
+        self.0.push(sender)
+    }
+    pub fn unsubscribe(&mut self, sender: &flume::Sender<Event>) {
+        self.0.retain(|s| !s.same_channel(sender));
+    }
+    pub fn send(&mut self, event: Event) {
```
dignifiedquire, line 112: shouldn't this only require `&self`?

iroh-sync/src/sync.rs
```diff
@@ -110,7 +132,7 @@ pub struct Replica<S: ranger::Store<SignedEntry> + PublicKeyStore> {
 struct InnerReplica<S: ranger::Store<SignedEntry> + PublicKeyStore> {
     namespace: Namespace,
     peer: RwLock<Peer<SignedEntry, S>>,
-    on_insert_sender: RwLock<Option<flume::Sender<Event>>>,
+    subscribers: Mutex<Subscribers>,
```
dignifiedquire, line 135: given that sending will happen much more than updating subscribers, might want to go for an `RwLock`

iroh-sync/src/sync.rs
Frando: Well, my idea was: `flume::Sender::send` only returns an error if the receiver is dropped. Through `retain`, all senders which are disconnected are removed. Wdyt?

iroh-sync/src/sync.rs
Frando: See comment above, the `retain` removes all disconnected senders. This allows to simply drop a receiver without having to explicitly unsubscribe from the replica, because the disconnected sender will be removed on the next emitted event.

iroh-sync/src/sync.rs
Frando: See above, if we require explicit unsubscribes we can do this, if we want to auto-remove disconnected senders we need `&mut` for each send.

iroh/src/sync_engine.rs
```diff
@@ -245,3 +262,63 @@ pub(crate) fn entry_to_content_status(entry: EntryStatus) -> ContentStatus {
...
+    fn from_replica_event(ev: iroh_sync::Event, content_status_cb: &ContentStatusCallback) -> Self {
+        match ev {
+            iroh_sync::Event::Insert { origin, entry, .. } => match origin {
+                InsertOrigin::Local => Self::InsertLocal {
+                    entry: entry.into(),
+                },
+                InsertOrigin::Sync { from, .. } => Self::InsertRemote {
+                    content_status: content_status_cb(entry.content_hash()),
+                    entry: entry.into(),
+                    // TODO: unwrap
```
dignifiedquire, line 318: sad face

iroh-sync/src/sync.rs
```diff
@@ -135,33 +158,31 @@ impl<S: ranger::Store<SignedEntry> + PublicKeyStore + 'static> Replica<S> {
...
+    pub fn unsubscribe(&self, sender: &flume::Sender<Event>) {
+        self.inner.subscribers.lock().unsubscribe(sender)
     }
 
-    /// Remove the subscription.
-    pub fn unset_event_sender(&self) -> bool {
-        self.inner.on_insert_sender.write().take().is_some()
+    /// Get the number of current event subscribers.
+    pub fn subscribers_count(&self) -> usize {
+        self.inner.subscribers.lock().0.len()
```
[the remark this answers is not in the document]
Frando: changed

iroh/src/sync_engine/live.rs
```diff
@@ -421,31 +364,26 @@ impl<B: iroh_bytes::store::Store> LiveActor<B> {
     }
 
     async fn shutdown(&mut self) -> anyhow::Result<()> {
-        for (namespace, _state) in self.states.drain() {
-            self.subscriptions.remove(&namespace);
-        }
+        // cancel all subscriptions
+        self.subscribers = Default::default();
```
dignifiedquire, line 368: `self.subscribers.clear()`?

iroh-sync/src/sync.rs
dignifiedquire: right, that makes sense, but it requires now always taking write lock..tradeoffs

iroh/src/sync_engine.rs
Frando: fixed

PR #1642, opened 2023-10-16
ref(iroh-net): Improve how STUN probes are run

flub:
## Description

This improves the branches taken when stun probes need to be run:

- If we don't have an appropriate socket, abort the entire probe set.

- If we didn't send the probe fully, error.

- In case of a plain send error: log it.

## Notes & open questions

This is probably more how this code should look like.

@dignifiedquire feel free to take this over and make any tweaks you
feel like as I'm no supposed to distract myself with this... ;)

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/src/netcheck/reportgen.rs
+++ b/iroh-net/src/netcheck/reportgen.rs
@@ -776,37 +776,40 @@ async fn run_probe(
     let mut result = ProbeReport::new(probe.clone());
 
     match probe {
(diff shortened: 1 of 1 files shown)


flub:
Let's get this merged.  Though I can't approve my own PR...

flub:
@dignifiedquire ping

dignifiedquire (approved):

Issue #1643, opened 2023-10-16
Do not eat memory when using documents


dignifiedquire:
The current code seems to load `values` in documents entirely into memory, even when the content is only added and read in a streaming fashion.

Frando:
This is a known issue to me with `client.docs.set_bytes`, but wouldn't know of other places where this happens?

dignifiedquire:
if you look #1645 I made a tests that does not use any of the buffering APIs, and that still blows up to almost 1GiB memory usage (adding a value of 512MiB)

dignifiedquire:
Closing as stale

PR #1645, opened 2023-10-16
test: add test for streaming content in iroh sync

dignifiedquire:
## Description

Test that helps to detect #1643

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ env:
   RUST_BACKTRACE: 1
   RUSTFLAGS: -Dwarnings
   RUSTDOCFLAGS: -Dwarnings
-  MSRV: "1.70"
+  MSRV: "1.71"
 
 jobs:
   build_and_test_nix:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -52,9 +52,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.1"
+version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ea5d730647d4fadd988536d06fecce94b7b4f2a7efdae548f1cf4b63205518ab"
+checksum = "b2969dcb958b36655471fc61f7e416fa76033bdd4bfed0678d8fee1e2d07a1f0"
 dependencies = [
  "memchr",
 ]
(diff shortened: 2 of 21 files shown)


Frando:
Wrote this test here that only uses iroh-bytes over RPC, and it also peaks at >512MB memory usage

```rust
#[tokio::test]
async fn bytes_big_value() -> Result<()> {
    setup_logging();
    let rt = test_runtime();
    let mut rng = test_rng(b"bytes_big_value");
    let node = spawn_node(rt, 0, &mut rng).await?;
    let client = node.client();
    let dir = testdir::testdir!();
    let num_chunks = 1024 * 512;
    let chunk_size = 1024;
    let size = num_chunks * chunk_size;
    let chunk = vec![2u8; chunk_size];

    let file_name = dir.join("data");
    let mut file = tokio::fs::File::create(&file_name).await?;
    for _ in 0..num_chunks {
        file.write_all(&chunk).await?;
    }
    file.flush().await?;
    drop(file);

    println!("adding from path");
    let outcome = client
        .blobs
        .add_from_path(
            file_name.to_path_buf(),
            true,
            iroh::rpc_protocol::SetTagOption::Auto,
            iroh::rpc_protocol::WrapOption::NoWrap,
        )
        .await?.finish().await?;
    let hash = outcome.hash;
    println!("read back");
    let mut file = tokio::fs::File::open(file_name).await?;
    let mut reader = client.blobs.read(hash).await?;

    let mut chunk_a = vec![0u8; 1024];
    let mut chunk_b = vec![0u8; 1024];

    let mut left_to_read = size;
    loop {
        let chunk_size = left_to_read.min(1024);
        reader.read_exact(&mut chunk_a[..chunk_size]).await?;
        file.read_exact(&mut chunk_b[..chunk_size]).await?;
        assert_eq!(&chunk_a[..chunk_size], &chunk_b[..chunk_size]);
        left_to_read -= chunk_size;

        if left_to_read == 0 {
            break;
        }
    }
    println!("done");
    node.shutdown();
    Ok(())
}

```

Frando:
`iroh_io` does allocate for each `read_at` call, because the API does not take a mutable buffer, but returns `Bytes`:
https://github.com/n0-computer/iroh-io/blob/main/src/tokio_io.rs#L236
```rust
    fn read_at(mut self, offset: u64, len: usize) -> JoinHandle<(Self, io::Result<Bytes>)> {
       // [..]
        spawn_blocking(move || {
            // len is just the expected len, so if it is too big, we should not allocate
            // the entire size.
            let mut buf = Vec::with_capacity(len.min(MAX_PREALLOC));
            let res = inner(&mut self.0, offset, len, &mut buf);
            (self, res.map(|_| buf.into()))
        })
    }
```

These `Bytes` are then sent as-is over the RPC in [`iroh::node::RpcHandler::blob_read`](https://github.com/n0-computer/iroh/blob/main/iroh/src/node.rs#L1332). So while this can be a lot of allocs, I don't see where it would ever alloc the full blob size. Where is the leak!?

dignifiedquire:
updated the test to actually store data on disk. now max memory allocations in a single spot are around 3MiB so as expected :)

Issue #1646, opened 2023-10-16
iroh-net: valgrind error: invalid atomic reads


dignifiedquire:
This looks like an issue in https://github.com/rust-lang/futures-rs/blob/master/futures-util/src/stream/futures_unordered/ready_to_run_queue.rs and more specifically in the way `drop` is implemented. 

This might or might not be responsible for the `malloc` crashes we have been seeing in CI

```
==1716606== Thread 25 magic_endpoint::
==1716606== Invalid read of size 8
==1716606==    at 0x38B867: core::sync::atomic::atomic_load (atomic.rs:3185)
==1716606==    by 0x38FBA7: core::sync::atomic::AtomicPtr<T>::load (atomic.rs:1325)
==1716606==    by 0xBAB2CB: futures_util::stream::futures_unordered::ready_to_run_queue::ReadyToRunQueue<Fut>::dequeue (ready_to_run_queue.rs:60)
==1716606==    by 0xBA9FD7: futures_util::stream::futures_unordered::ready_to_run_queue::ReadyToRunQueue<Fut>::clear (ready_to_run_queue.rs:102)
==1716606==    by 0xE9BA39: <futures_util::stream::futures_unordered::ready_to_run_queue::ReadyToRunQueue<Fut> as core::ops::drop::Drop>::drop (ready_to_run_queue.rs:119)
==1716606==    by 0xEFD092: core::ptr::drop_in_place<futures_util::stream::futures_unordered::ready_to_run_queue::ReadyToRunQueue<iroh_net::derp::http::client::Client::dial_region::{{closure}}::{{closure}}::{{closure}}>> (mod.rs:497)
==1716606==    by 0x94989D: alloc::sync::Arc<T,A>::drop_slow (sync.rs:1749)
==1716606==    by 0xF42203: <alloc::sync::Arc<T,A> as core::ops::drop::Drop>::drop (sync.rs:2405)
==1716606==    by 0xF02599: core::ptr::drop_in_place<alloc::sync::Arc<futures_util::stream::futures_unordered::ready_to_run_queue::ReadyToRunQueue<iroh_net::derp::http::client::Client::dial_region::{{closure}}::{{closure}}::{{closure}}>>> (mod.rs:497)
[...]

PR #1647, opened 2023-10-16
fix: avoid FuturesUnordered

dignifiedquire:
## Description

Workaround for #1646  and likely makes more sense in the context of Tokio usage anyway.

In general `JoinSet` is simply used instead.

### DIFF
--- a/iroh-gossip/src/net/util.rs
+++ b/iroh-gossip/src/net/util.rs
@@ -4,10 +4,11 @@ use std::{collections::HashMap, io, pin::Pin, time::Instant};
 
 use anyhow::{anyhow, bail, ensure, Context, Result};
 use bytes::{Bytes, BytesMut};
-use futures::{future::BoxFuture, stream::FuturesUnordered, FutureExt, StreamExt};
+use futures::future::BoxFuture;
 use iroh_net::{key::PublicKey, MagicEndpoint, PeerAddr};
 use tokio::{
     io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt},
+    task::JoinSet,
     time::{sleep_until, Sleep},
 };
 use tokio_util::sync::CancellationToken;
@@ -88,7 +89,7 @@ pub type DialFuture = BoxFuture<'static, (PublicKey, anyhow::Result<quinn::Conne
 #[derive(Debug)]
 pub struct Dialer {
     endpoint: MagicEndpoint,
-    pending: FuturesUnordered<DialFuture>,
+    pending: JoinSet<(PublicKey, anyhow::Result<quinn::Connection>)>,
     pending_peers: HashMap<PublicKey, CancellationToken>,
 }
 
@@ -113,16 +114,14 @@ impl Dialer {
         let cancel = CancellationToken::new();
         self.pending_peers.insert(peer_id, cancel.clone());
         let endpoint = self.endpoint.clone();
-        let fut = async move {
+        self.pending.spawn(async move {
             let res = tokio::select! {
                 biased;
                 _ = cancel.cancelled() => Err(anyhow!("Cancelled")),
                 res = endpoint.connect(PeerAddr::new(peer_id), alpn_protocol) => res
             };
             (peer_id, res)
-        }
-        .boxed();
-        self.pending.push(fut.boxed());
+        });
     }
 
     /// Abort a pending dial
@@ -141,7 +140,12 @@ impl Dialer {
     pub async fn next_conn(&mut self) -> (PublicKey, anyhow::Result<quinn::Connection>) {
         match self.pending_peers.is_empty() {
             false => {
-                let (peer_id, res) = self.pending.next().await.unwrap();
+                let (peer_id, res) = self
+                    .pending
+                    .join_next()
+                    .await
+                    .expect("not canceled")
+                    .unwrap();
                 self.pending_peers.remove(&peer_id);
                 (peer_id, res)
             }
@@ -162,10 +166,15 @@ impl futures::Stream for Dialer {
         mut self: Pin<&mut Self>,
         cx: &mut std::task::Context<'_>,
     ) -> std::task::Poll<Option<Self::Item>> {
-        match self.pending.poll_next_unpin(cx) {
-            std::task::Poll::Ready(Some((peer_id, result))) => {
+        match self.pending.poll_join_next(cx) {
+            std::task::Poll::Ready(Some(Ok((peer_id, result)))) => {
                 self.pending_peers.remove(&peer_id);
                 std::task::Poll::Ready(Some((peer_id, result)))
+            }
+            std::task::Poll::Ready(Some(Err(e))) => {
+                // Should not happen unless the task paniced or got canceled
+                // TODO: is this what we want to do here?
+                panic!("{:?}", e);
             }
             _ => std::task::Poll::Pending,
         }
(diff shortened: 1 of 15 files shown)


dignifiedquire:
CI is sad because MSRV bump to 1.71 is needed due to the new dep.

Arqu:
Ran tests with valgrind last night. Made 52 loops before it died. No errors reported. LGTM

rklaehn:
So what exactly is the difference in terms of how the futures are being polled? Or is this just trying something until it works?

iroh-net/src/netcheck/reportgen.rs
```diff
@@ -253,6 +252,7 @@ impl Actor {
 
                 _ = &mut probe_timer => {
                     warn!("probes timed out");
+                    probes.abort_all();
```
flub, line 255: The probes are aborted on line 303 and they get a log message there.  if you abort here i'm not sure if you get the log message.  (admittedly the log message there is a bit misleading in case of a timeout, maybe it could be tweaked)

Then `self.handle_abort_probes()` makes sure that we break out of the loop.  It's designed this way because it should not matter whether we fall in this condition by the timeout or by the actor message, and this way everything is done by the same code.

dignifiedquire:
> So what exactly is the difference in terms of how the futures are being polled? Or is this just trying something until it works?

this is not about how futures are polled, it is about how the `unsafe` code in `FuturesUnordered` produces invalid memory reads and writes :(

iroh-net/src/netcheck/reportgen.rs
```diff
@@ -564,9 +572,7 @@ impl Actor {
     ///     failure permanent.  Probes in a probe set are essentially retries.
     ///   - Once there are [`ProbeReport`]s from enough regions, all remaining probes are
     ///     aborted.  That is, the main actor loop stops polling them.
-    async fn prepare_probes_task(
-        &mut self,
-    ) -> Result<FuturesUnordered<Pin<Box<impl Future<Output = Result<ProbeReport>>>>>> {
+    async fn prepare_probes_task(&mut self) -> Result<JoinSet<Result<ProbeReport>>> {
```
flub, line 575: This no longer prepares a future but will spawn the futures on tasks and they'll start running right away.  Maybe rename the function to `start_probe_tasks` or `spawn_probe_tasks` and update the first line of the docstring as well?

iroh-net/src/netcheck/reportgen.rs
```diff
@@ -599,42 +605,44 @@ impl Actor {
...
-                        Err(ProbeError::AbortSet(err, probe)) => {
+                        Ok(Err(ProbeError::AbortSet(err, probe))) => {
                             debug!(?probe, "probe set aborted: {:#}", err);
+                            set.abort_all();
                             return Err(err);
                         }
+                        Err(err) => {
+                            warn!("fatal probe set error, aborting: {:#}", err);
+                            return Err(anyhow::anyhow!(err));
+                        }
```
flub, line 640: If a single probe panics (or is cancelled) I think it's probably fine to continue the other probes.  Not really sure if all of them would panic as they run the same code... but really none should panic so I guess I'd still choose to carry on.

Not sure if we can log the panic in a reasonable way though.

iroh-net/src/netcheck/reportgen.rs
```diff
@@ -266,11 +266,19 @@ impl Actor {
                 }
 
                 // Drive the probes.
```
flub, line 268: maybe update this comment to say something like "wait for probe tasks to finish" as this no longer drives the future.

rklaehn:
> > So what exactly is the difference in terms of how the futures are being polled? Or is this just trying something until it works?
> 
> this is not about how futures are polled, it is about how the `unsafe` code in `FuturesUnordered` produces invalid memory reads and writes :(

Oh boy...

dignifiedquire:
> Oh boy... Did you open an issue?

of course : https://github.com/n0-computer/iroh/issues/1646 and https://github.com/rust-lang/futures-rs/issues/2781

iroh-gossip/src/net/util.rs
```diff
@@ -162,11 +166,16 @@ impl futures::Stream for Dialer {
...
-            std::task::Poll::Ready(Some((peer_id, result))) => {
+        match self.pending.poll_join_next(cx) {
+            std::task::Poll::Ready(Some(Ok((peer_id, result)))) => {
                 self.pending_peers.remove(&peer_id);
                 std::task::Poll::Ready(Some((peer_id, result)))
             }
+            std::task::Poll::Ready(Some(Err(e))) => {
+                // Should not happen unless the task paniced or got canceled
+                // TODO: is this what we want to do here?
+                panic!("{:?}", e);
```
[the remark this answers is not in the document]
dignifiedquire: maybe, usually panics in one part of the code result in the other parts not quite working as intended anymore

iroh-net/src/netcheck/reportgen.rs
dignifiedquire: kept the logging, can `continue` now

iroh-net/src/netcheck/reportgen.rs
dignifiedquire: renamed

iroh-net/src/netcheck/reportgen.rs
dignifiedquire: removed

iroh-net/src/netcheck/reportgen.rs
```diff
@@ -266,11 +266,19 @@ impl Actor {
...
-                        None => self.handle_abort_probes(),
+                        Some(Ok(Ok(report))) => self.handle_probe_report(report),
+                        Some(Ok(Err(_))) => (),
+                        Some(Err(e)) => {
+                            warn!("fatal probes error: {:?}", e);
+                            probes.abort_all();
+                            self.handle_abort_probes();
+                        }
+                        None => {
+                            probes.abort_all();
```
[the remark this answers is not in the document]
dignifiedquire: removed

iroh-net/src/netcheck/reportgen.rs
```diff
@@ -266,11 +266,19 @@ impl Actor {
...
+                set_result = probes.join_next(), if self.outstanding_tasks.probes => {
                     match set_result {
-                        Some(Ok(report)) => self.handle_probe_report(report),
-                        Some(Err(_)) => (),
-                        None => self.handle_abort_probes(),
+                        Some(Ok(Ok(report))) => self.handle_probe_report(report),
+                        Some(Ok(Err(_))) => (),
+                        Some(Err(e)) => {
+                            warn!("fatal probes error: {:?}", e);
+                            probes.abort_all();
```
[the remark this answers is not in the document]
dignifiedquire: continues now

iroh-gossip/src/net/util.rs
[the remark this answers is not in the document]
dignifiedquire: now only logs

iroh-gossip/src/net/util.rs
```diff
@@ -141,7 +140,12 @@ impl Dialer {
     pub async fn next_conn(&mut self) -> (PublicKey, anyhow::Result<quinn::Connection>) {
         match self.pending_peers.is_empty() {
             false => {
-                let (peer_id, res) = self.pending.next().await.unwrap();
+                let (peer_id, res) = self
+                    .pending
+                    .join_next()
+                    .await
+                    .expect("not canceled")
```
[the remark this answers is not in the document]
dignifiedquire: fixed, to try to find the first working one

flub (approved):

PR #1653, opened 2023-10-17
fix(iroh-net): direct address management

divagant-martian:
## Description

Prunes peers and addresses with the same strategy: keep at most X inactive ones. We will need to use this for a bit to decide if further efforts are needed or if this pruning is too aggressive but for now I think this is a sweet spot between "connection management" and "address book management"

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


iroh-net/src/magicsock/endpoint.rs
```diff
@@ -44,6 +44,15 @@ const UPGRADE_INTERVAL: Duration = Duration::from_secs(60);
...
 const TRUST_UDP_ADDR_DURATION: Duration = Duration::from_millis(6500);
 
+/// Number of addresses that are not active that we keep around per peer.
+///
+/// See [`Endpoint::prune_direct_addresses`].
+const MAX_INACTIVE_DIRECT_ADDRESSES: usize = 5;
+
+/// Number of peers that are inactive for which we keep info about. This limit is enforced
+/// periodically via [`PeerMap::prune_inactive`].
+const MAX_INACTIVE_PEERS: usize = 30;
```
dignifiedquire, line 54: I wonder if we should make this number configurable, large nodes might want to keep more than this around

dignifiedquire (approved):
I think this approach aligns very well with the current goals & the way connections are used. Nice work

PR #1658, opened 2023-10-18
fix(iroh-net): temp fix for progress bar when downloading a hash seq

rklaehn:
## Description

Fixes the download progress bar when downloading a hash seq where the size is not known.

## Notes & open questions

Is this good enough? The only way to get a **proper** progress for a hash seq where you don't know the overall size is to do a verified size request before and give the sum of the sizes to the progress logic. But that would be another roundtrip. Fine for GUI stuff, but in many cases we don't really care, so I would hate to do this all the time.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -201,6 +201,28 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-stream"
+version = "0.3.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cd56dd203fef61ac097dd65721a419ddccb106b2d2b70ba60a6b529f03961a51"
+dependencies = [
+ "async-stream-impl",
+ "futures-core",
+ "pin-project-lite",
+]
+
+[[package]]
+name = "async-stream-impl"
+version = "0.3.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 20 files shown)


Cargo.lock
```diff
@@ -1971,16 +1971,16 @@ dependencies = [
 
 [[package]]
 name = "iana-time-zone"
-version = "0.1.58"
+version = "0.1.57"
```
dignifiedquire, line 1974: these seem to be downgrades, I think you need to update your lock file

iroh-bytes/src/store/flat.rs
```diff
@@ -1197,7 +1197,7 @@ impl Store {
         meta_path: PathBuf,
         rt: crate::util::runtime::Handle,
     ) -> anyhow::Result<Self> {
-        tracing::debug!(
+        tracing::info!(
```
dignifiedquire, line 1200: should this really be `info`?

rklaehn: Yes. I constantly need to be reminded where the damn store is so I can nuke it. Compared to all the other stuff we log at info level, this is more important.

dignifiedquire (approved):

PR #1660, opened 2023-10-18
iroh-bytes: Get api improvements

rklaehn:
## Description

A number of small improvements of the get side of the iroh-bytes API, needed for the tracker and the progress rework

## Notes & open questions

I want to get those in separately to the above 2 PRs

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-bytes/src/util.rs
+++ b/iroh-bytes/src/util.rs
@@ -129,6 +129,38 @@ impl HashAndFormat {
     }
 }
 
+impl Display for HashAndFormat {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let mut slice = [0u8; 64];
+        hex::encode_to_slice(self.hash.as_bytes(), &mut slice).unwrap();
+        write!(f, "{}", std::str::from_utf8(&slice).unwrap())?;
+        if self.format.is_hash_seq() {
+            write!(f, "s")?;
+        }
+        Ok(())
+    }
+}
+
+impl FromStr for HashAndFormat {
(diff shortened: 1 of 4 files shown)


iroh-bytes/src/util.rs
```diff
@@ -129,6 +129,38 @@ impl HashAndFormat {
     }
 }
 
+impl Display for HashAndFormat {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let mut slice = [0u8; 64];
+        hex::encode_to_slice(self.hash.as_bytes(), &mut slice).unwrap();
+        write!(f, "{}", std::str::from_utf8(&slice).unwrap())?;
+        if self.format.is_hash_seq() {
+            write!(f, "s")?;
```
dignifiedquire, line 138: why use a suffix, aren't we usually using prefixes for things like this?

iroh-bytes/src/util.rs
```diff
@@ -129,6 +129,38 @@ impl HashAndFormat {
...
+        }
+        Ok(())
+    }
+}
+
+impl FromStr for HashAndFormat {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
```
dignifiedquire, line 147: should have a test for roundtripping

iroh-bytes/src/util.rs
rklaehn: prefix works as well. Just needed something.

dignifiedquire (approved):

PR #1663, opened 2023-10-18
iroh-net: doctor improvements

rklaehn:
## Description

- doctor accept can accept multiple connections at the same time (only the first one will get a GUI).
- gui on server side is properly cleared

## Notes & open questions

This is useful when running doctor accept on a server, for async dogfooding.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.


rklaehn:
@b5 this came out of the dogfooding session yesterday. It would be nice if you could just leave the accept running indefinitely on a server.

dignifiedquire (approved):

Issue #1635, opened 2023-10-13
Default to BTreeSet / BTreeMap unless something is performance critical


rklaehn:
A HashSet with randomized element order is a great way to introduce nondeterminism into a code base for no good reason. In some places you want nondeterminism, but on average I think it makes things quite annoying.

E.g. this output. AddrInfo is unchanged, but you don't really see it from the output because the iteration order of direct_addresses is randomized.

```
listening on PeerAddr { peer_id: PublicKey(ks7dxisghqs3mvfe), info: AddrInfo { derp_region: Some(2), direct_addresses: {82.76.245.118:44204, 192.168.1.141:44204} } }
peer addr: ks7dxisghqs3mvfe4qrmn6hcohkgghraby6x6sfsr3zf2sna7bsa
listening on PeerAddr { peer_id: PublicKey(ks7dxisghqs3mvfe), info: AddrInfo { derp_region: Some(2), direct_addresses: {192.168.1.141:44204, 82.76.245.118:44204} } }
peer addr: ks7dxisghqs3mvfe4qrmn6hcohkgghraby6x6sfsr3zf2sna7bsa
listening on PeerAddr { peer_id: PublicKey(ks7dxisghqs3mvfe), info: AddrInfo { derp_region: Some(2), direct_addresses: {82.76.245.118:44204, 192.168.1.141:44204} } }
peer addr: ks7dxisghqs3mvfe4qrmn6hcohkgghraby6x6sfsr3zf2sna7bsa
```

PR #1672, opened 2023-10-19
iroh-net: Use BTreeSet for AddrInfo

rklaehn:
## Description

rationale: this way the serialized representation of AddrInfo and all the things that contain AddrInfo is unique instead of constantly toggling when there are 2 or more addrs.)

## Notes & open questions

Is there any place that will only work if the addrs are randomized? I don't think so, and I also don't think we should have it. But you never know...

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


rklaehn:
See also https://github.com/n0-computer/iroh/issues/1635

dignifiedquire (approved):
this should be fine

PR #1674, opened 2023-10-19
refactor(net): improve derp client handling

dignifiedquire:
## Description

- Currently derp messages can be lost, due to using `tokio::select` and`futures::select_all` over non cancelation safe futures. This refactors the code to fix this.
- Additionally this refactors `derp::http::Client` to an actor like structure and removes the need for locks in it
- Bumps MSRV to `1.72` due to updated dependencies
- Refactor `derp::Client` to not use locks either

With this all of the `derp` code is stripped of locks :)

### DIFF
--- a/iroh-net/src/derp/client.rs
+++ b/iroh-net/src/derp/client.rs
@@ -3,17 +3,16 @@ use std::net::SocketAddr;
 use std::sync::Arc;
 use std::time::Duration;
 
-use anyhow::{bail, ensure, Context, Result};
+use anyhow::{anyhow, bail, ensure, Context, Result};
 use bytes::Bytes;
 use futures::stream::Stream;
 use futures::{Sink, SinkExt, StreamExt};
 use tokio::io::{AsyncRead, AsyncWrite};
 use tokio::sync::mpsc;
-use tokio::sync::Mutex;
-use tokio::task::JoinHandle;
 use tokio_util::codec::{FramedRead, FramedWrite};
 use tracing::{debug, info_span, Instrument};
 
+use super::codec::PER_CLIENT_READ_QUEUE_DEPTH;
 use super::{
     codec::{
         recv_frame, write_frame, DerpCodec, Frame, FrameType, MAX_PACKET_SIZE,
@@ -23,6 +22,7 @@ use super::{
 };
 
 use crate::key::{PublicKey, SecretKey};
+use crate::util::AbortingJoinHandle;
 
 const CLIENT_RECV_TIMEOUT: Duration = Duration::from_secs(120);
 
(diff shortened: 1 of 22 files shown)


iroh-net/src/magicsock/derp_actor.rs
```diff
@@ -373,12 +458,25 @@ impl DerpActor {
...
             if *i == self.conn.my_derp() {
                 continue;
             }
-            if ad.last_write.duration_since(now) > DERP_INACTIVE_CLEANUP_TIME {
-                to_close.push(*i);
+            let (os, or) = oneshot::channel();
+            match s.send(ActiveDerpMessage::GetLastWrite(os)).await {
+                Ok(_) => match or.await {
+                    Ok(last_write) => {
+                        if last_write.duration_since(now) > DERP_INACTIVE_CLEANUP_TIME {
```
ramfox, line 469: Just trying to remember, is the idea with `clean_stale_derp` and other methods that close the derp clients, that if we end up needing to dial that derper later, we can just add it back as an `active_derp` when we need it?

iroh-net/src/magicsock/derp_actor.rs
```diff
@@ -49,33 +48,108 @@ pub(super) enum DerpActorMessage {
...
+    GetPeerRoute(PublicKey, oneshot::Sender<Option<derp::http::Client>>),
+    GetClient(oneshot::Sender<derp::http::Client>),
+    NotePreferred(bool),
+    Shutdown,
+}
+
+impl ActiveDerp {
+    async fn run(mut self) {
+        if let Err(err) = self.c.connect().await {
+            // TODO: what to do?
```
ramfox, line 80: if the error if `ClientError::Timeout`, probably just do a backoff and try to dial again later?

The other option is to just shut down the active derp and remove it from the `DerpActor` & then wait until we need to dial that derper again

iroh-net/src/magicsock/derp_actor.rs
dignifiedquire: yes

iroh-net/src/magicsock/derp_actor.rs
dignifiedquire: fixed

iroh-net/src/derp/client.rs
```diff
@@ -144,7 +144,7 @@ impl Client {
     }
 
     /// The local address that the [`Client`] is listening on.
-    pub async fn local_addr(&self) -> Result<SocketAddr> {
+    pub fn local_addr(&self) -> Result<SocketAddr> {
```
rklaehn, line 147: :+1:

iroh-net/src/derp/client_conn.rs
```diff
@@ -45,7 +45,7 @@ pub(crate) struct ClientConnManager {
     // TODO: replace with rate limiter, also, this should probably be on the ClientSets, not on
     // the client itself
     // replace_limiter: RateLimiter,
-    io_handle: JoinHandle<Result<()>>,
+    io_handle: AbortingJoinHandle<Result<()>>,
```
rklaehn, line 48: :+1: AbortingJoinHandle should really be the default.

dignifiedquire:
Pushed the last round of lock removal

ramfox (approved):
Looks good to me! 🙌

iroh-net/src/magicsock/derp_actor.rs
```diff
@@ -75,74 +82,190 @@ enum ActiveDerpMessage {
...
+    async fn handle_derp_msg(
+        &mut self,
+        msg: Result<(ReceivedMessage, usize), ClientError>,
+    ) -> ReadResult {
+        match msg {
+            Err(err) => {
+                debug!(
+                    "[{:?}] derp.recv(derp-{}): {:?}",
+                    self.derp_client, self.region, err
+                );
```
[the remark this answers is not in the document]
dignifiedquire: you are right, pushing improved logging in a moment

ramfox (approved):
Fab!

PR #1675, opened 2023-10-19
refactor(iroh-net): Proper abstraction around `best_addr`

Frando:
## Description

This refactors how we use and calculate `best_addr`. It should not have any functional changes, only refactoring to make things more robust for the future.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/src/magicsock/endpoint.rs
+++ b/iroh-net/src/magicsock/endpoint.rs
@@ -23,6 +23,9 @@ use crate::{
 
 use super::{metrics::Metrics as MagicsockMetrics, ActorMessage, QuicMappedAddr, SendAddr};
 
+mod best_addr;
+use best_addr::{BestAddr, BestAddrClearReason, BestAddrSource, BestAddrState};
+
 /// How long we wait for a pong reply before assuming it's never coming.
 const PING_TIMEOUT_DURATION: Duration = Duration::from_secs(5);
 
@@ -86,11 +89,7 @@ pub(super) struct Endpoint {
     /// The fallback/bootstrap path, if non-zero (non-zero for well-behaved clients).
     derp_region: Option<(u16, EndpointState)>,
     /// Best non-DERP path.
-    best_addr: Option<AddrLatency>,
-    /// Time best address re-confirmed.
-    best_addr_at: Option<Instant>,
-    /// Time when best_addr expires.
-    trust_best_addr_until: Option<Instant>,
+    best_addr: BestAddr,
     /// [`EndpointState`] for this peer's direct addresses.
     direct_addr_state: HashMap<IpPort, EndpointState>,
     is_call_me_maybe_ep: HashMap<SocketAddr, bool>,
@@ -136,9 +135,7 @@ impl Endpoint {
             derp_region: options
                 .derp_region
                 .map(|region| (region, EndpointState::default())),
-            best_addr: None,
-            best_addr_at: None,
-            trust_best_addr_until: None,
+            best_addr: Default::default(),
             sent_ping: HashMap::new(),
             direct_addr_state: HashMap::new(),
             is_call_me_maybe_ep: HashMap::new(),
@@ -153,8 +150,9 @@ impl Endpoint {
 
     /// Returns info about this endpoint
     pub fn info(&self) -> EndpointInfo {
-        let (conn_type, latency) = if self.is_best_addr_valid(Instant::now()) {
-            let addr_info = self.best_addr.as_ref().expect("checked");
+        // TODO: I think we should return direct here for outdated.
+        let (conn_type, latency) = if self.best_addr.is_valid(Instant::now()) {
+            let addr_info = self.best_addr.addr_info().expect("checked");
             (ConnectionType::Direct(addr_info.addr), addr_info.latency)
         } else if let Some((region_id, relay_state)) = self.derp_region.as_ref() {
             let latency = relay_state.recent_pong().map(|pong| pong.latency);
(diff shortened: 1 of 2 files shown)


iroh-net/src/magicsock/endpoint/best_addr.rs
```diff
@@ -0,0 +1,202 @@
...
+
+#[derive(Debug)]
+pub(super) enum BestAddrState<'a> {
+    Valid(&'a AddrLatency),
+    Outdated(&'a AddrLatency),
+    Empty,
+}
+
+#[derive(Debug)]
+pub enum BestAddrClearReason {
```
dignifiedquire, line 48: just `ClearReason`,  to avoid `best_addr::BestAddrClearReson` double

Frando: fixed!

dignifiedquire (approved):
lgtm but should get review from @divagant-martian before merging

iroh-net/src/magicsock/endpoint/best_addr.rs
```diff
@@ -0,0 +1,202 @@
+//! The [`BestAddr`] is the currently active best address for UDP sends.
```
divagant-martian, line 1: this file is not pub but it would be good to document its contents for us

iroh-net/src/magicsock/endpoint/best_addr.rs
```diff
@@ -0,0 +1,202 @@
...
+
+use std::{
+    net::SocketAddr,
+    time::{Duration, Instant},
+};
+
+use iroh_metrics::inc;
+use tracing::{debug, info, trace};
+
+use super::{AddrLatency, TRUST_UDP_ADDR_DURATION};
```
divagant-martian, line 11: These two only exist for the best address, maybe move them to this file? Having a separate file is a really good call, helping cleanup `endpoint.rs` and my moving we can further tidy it

iroh-net/src/magicsock/endpoint/best_addr.rs
```diff
@@ -0,0 +1,202 @@
...
+        confirmed_at: Instant,
+        trust_until: Instant,
+    ) -> Self {
+        let inner = BestAddrInner {
+            addr: AddrLatency { addr, latency },
+            confirmed_at,
+            trust_until: Some(trust_until),
+        };
+        Self(Some(inner))
+    }
```
divagant-martian, line 69:
```suggestion
    }
    
```

iroh-net/src/magicsock/endpoint/best_addr.rs
```diff
@@ -0,0 +1,202 @@
...
+            inc!(MagicsockMetrics, num_direct_conns_added);
+            if has_derp {
+                // we no longer rely on the relay connection, decrease the relay connection
+                // count
+                inc!(MagicsockMetrics, num_relay_conns_removed);
+            }
+        }
+    }
+
+    pub fn state(&self, now: Instant) -> State {
```
divagant-martian, line 168: since the best address is per peer, these logs are probably more useful in the caller (where you already have similar logs)

iroh-net/src/magicsock/endpoint.rs
```diff
@@ -203,36 +201,40 @@ impl Endpoint {
...
-                    // Address is current and can be used
-                    (Some(best_addr.addr), None, false)
-                }
+        // Update our best addr from candidate addresses (only if it is empty and if we have recent
+        // pongs).
+        self.update_best_addr_from_candidates();
+        match self.best_addr.state(*now) {
+            // we have a valid address: use it!
+            best_addr::State::Valid(best_addr) => {
+                trace!(addr = %best_addr.addr, latency = ?best_addr.latency, "best_addr is set and valid, use best_addr only");
```
divagant-martian, line 210: can you add `self.public_key.fmt_short()` here to know what peer we are working on?

iroh-net/src/magicsock/endpoint.rs
```diff
@@ -203,36 +201,40 @@ impl Endpoint {
...
-                    self.derp_region()
-                } else {
-                    None
-                };
-
-                trace!(udp_addr = ?addr, derp_region = ?derp_region, ?should_ping, "best_addr is unset, use candidate addr and derp");
-                (addr, derp_region, should_ping)
+            // we have an outdated address: use it, but use derp as well.
+            best_addr::State::Outdated(best_addr) => {
+                trace!(addr = %best_addr.addr, latency = ?best_addr.latency, "best_addr is set but outdated, use best_addr and derp");
```
divagant-martian, line 215: same here

iroh-net/src/magicsock/endpoint.rs
```diff
@@ -203,36 +201,40 @@ impl Endpoint {
...
+                (Some(best_addr.addr), self.derp_region(), true)
+            }
+            // we have no best address: use a random canidate if available, and derp as backup.
+            best_addr::State::Empty => {
+                let addr = self
+                    .direct_addr_state
+                    .keys()
+                    .choose_stable(&mut rand::thread_rng())
+                    .map(|ipp| SocketAddr::from(*ipp));
+                trace!(udp_addr = ?addr, "best_addr is unset, use candidate addr and derp");
```
divagant-martian, line 225: here as well

iroh-net/src/magicsock/endpoint.rs
```diff
@@ -203,36 +201,40 @@ impl Endpoint {
...
             }
         }
     }
 
-    /// Determines a potential best addr for this endpoint. And if the endpoint needs a ping.
-    fn get_candidate_udp_addr(&mut self) -> (Option<SocketAddr>, bool) {
+    /// Update our best_addr (if empty) with the candidate udp addr with the lowest latency.
+    ///
+    /// Returns true if the address was updated.
+    fn update_best_addr_from_candidates(&mut self) -> bool {
```
divagant-martian, line 234: this return value is never used in the code, we can remove it

iroh-net/src/magicsock/endpoint.rs
```diff
@@ -153,8 +150,9 @@ impl Endpoint {
 
     /// Returns info about this endpoint
     pub fn info(&self) -> EndpointInfo {
-        let (conn_type, latency) = if self.is_best_addr_valid(Instant::now()) {
-            let addr_info = self.best_addr.as_ref().expect("checked");
+        // TODO: I think we should return direct here for outdated.
```
divagant-martian, line 153: to me it makes sense what you say. When it's outdated we return both this address and the derp and metrics are only updated when the address is cleared. So even there, we are counting an outdated conn as a a direct one

iroh-net/src/magicsock/endpoint.rs
```diff
@@ -712,18 +665,11 @@ impl Endpoint {
...
-                    self.best_addr = None;
-                    // no longer relying on a direct connection, remove conn count
-                    inc!(MagicsockMetrics, num_direct_conns_removed);
-                    if self.derp_region.is_some() {
-                        // we now rely on a relay connection, add a relay count
-                        inc!(MagicsockMetrics, num_relay_conns_added);
-                    }
-                }
-            }
+            self.best_addr.clear_if_equals(
```
divagant-martian, line 668: this is a lot nicer but we lost the information of the peer whose best address is being dropped

divagant-martian (commented):
overall I think it's a great improvement, and moving it to another file seems like the right call to me. I found a bug -not caused by your changes- but that we should address, and otherwise just some comments about info we lost in logging

iroh-net/src/magicsock/endpoint.rs
Frando: These log lines are only called from function that have instrumentation spans over them. So the log lines should look like this:
```
2023-10-23T09:20:32.398115Z TRACE poll_send{me=pp6ebwzioaeldcnl}:get_send_addrs{peer=cx6rezjzqm3ayple}: iroh_net::magicsock::endpoint: best_addr is set and valid, use best_addr only addr=172.20.0.1:60955 latency=Some(17.247288ms)
```

iroh-net/src/magicsock/endpoint.rs
```diff
@@ -249,45 +251,17 @@ impl Endpoint {
...
-            let addr = pong.from.as_socket_addr();
-            let trust_best_addr_until = pong.pong_at + Duration::from_secs(60 * 60);
-
-            info!(
-               %addr,
-               latency = ?lowest_latency,
-               trust_for = ?trust_best_addr_until.duration_since(Instant::now()),
-               "new best_addr (candidate address with most recent pong)"
+            self.best_addr.insert(
+                pong.from.as_socket_addr(),
```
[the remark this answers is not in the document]
Frando: Ah, true that `SendAddr::as_socket_addr` maps Derp addresses to fake addresses. However, in this place I think it is correct (even though not enforced on the type level), because we only use pongs and thus `SendAddr`s out of `Endpoint::direct_addr_state`, which should only ever contain pongs for direct (UDP) addresses. So I do think that this is correct here, unless I'm missing something.

I already thought of abstracing the direct addr state in a similar manner to this PR.

In any case - if the behavior is the same as it is on `main`, let's leave this to a followup and verify there that no fake Derp addresses can ever end up in the `best_addr`.

iroh-net/src/magicsock/endpoint.rs
```diff
@@ -203,36 +201,40 @@ impl Endpoint {
...
 
-    /// Determines a potential best addr for this endpoint. And if the endpoint needs a ping.
-    fn get_candidate_udp_addr(&mut self) -> (Option<SocketAddr>, bool) {
+    /// Update our best_addr (if empty) with the candidate udp addr with the lowest latency.
+    ///
+    /// Returns true if the address was updated.
+    fn update_best_addr_from_candidates(&mut self) -> bool {
+        if !self.best_addr.is_empty() {
+            return false;
+        }
```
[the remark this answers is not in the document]
Frando: Renamed to `assign_best_addr_from_candidates_if_empty`

iroh-net/src/magicsock/endpoint.rs
```diff
@@ -297,25 +271,22 @@ impl Endpoint {
             debug!("full ping: no full ping done");
             return true;
         }
-        if !self.is_best_addr_valid(*now) {
+        if !self.best_addr.is_valid(*now) {
             debug!("full ping: best addr expired");
```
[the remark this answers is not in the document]
Frando: :+1: Changed to match on `best_addr.state()`in the latest commit, reads much clearer also.

iroh-net/src/magicsock/endpoint/best_addr.rs
Frando: yup removed them here

iroh-net/src/magicsock/endpoint.rs
Frando: removed

iroh-net/src/magicsock/endpoint.rs
Frando: I changed it so that we return `Direct` if best_addr is set but expired.

Actually the more proper fix might be to return something like `ConnectionType::Both` because if the `best_addr` is set but expired, we send over both Derp and the Udp address. But not sure how important this is to report to the outside.

iroh-net/src/magicsock/endpoint.rs
Frando: see above

iroh-net/src/magicsock/endpoint.rs
Frando: see above

iroh-net/src/magicsock/endpoint.rs
Frando: In the most recent commit I changed this to return `ConnectionType::Mixed`  in that case. Also put it all into a single `match` statement for clarity.

iroh-net/src/magicsock/endpoint.rs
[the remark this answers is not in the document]
divagant-martian: ah yes, I fixed that in fact. Probably the reason why it's still there is because we store a pong for the derper (in the derp_region) as well

divagant-martian (approved):

Issue #1673, opened 2023-10-19
iroh-net: reduce size of crypto keys


rklaehn:
We have a lot of places where we are dragging iroh_net:: key ::PublicKey (size 224) around just to use it as keys.

The reason these things are so fat is that they contain some information that is expensive to compute. But I think in a lot of places this extra information is not needed and is just taking up space. This is probably costing us something wherever we use message passing because we end up with giant enums, but this is hard to track down in perf traces. Also every single key -> whatever map suffers.

How about this: we use a global map / intern table for deduplication, and e.g. a PublicKey becomes just a newtype for an [u8; 32]?

rklaehn:
Ok, I tried removing Copy from PublicKey so we can do some fancy stuff in the background. After changing every single file of our code, I gave up. Will have to be something else...

rklaehn:
Second attempt is having a cache but keeping the thing `Copy`. Not so nice internally, but you should never really see it.

PR #1677, opened 2023-10-19
feat(iroh-net): cache for crypto keys

rklaehn:
## Description

An attempt to get down the size of a key to something more manageable, see https://github.com/n0-computer/iroh/issues/1673

## Notes & open questions

Size of the cache and TTL. I think 1 minute is fine, if you don't perform any crypto ops on a key for 1 minute it means you are not connected to it. Size is more difficult, but the items are not that big so there is not much harm in having a big size.

Note two: this should help a lot for the derper.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh-net/src/key.rs
```diff
@@ -38,7 +58,7 @@ impl Serialize for PublicKey {
     where
         S: serde::Serializer,
     {
-        self.public.serialize(serializer)
+        self.public().serialize(serializer)
```
rklaehn, line 61: Why can't this be self.0.serialize?

iroh-net/Cargo.toml
```diff
@@ -78,6 +78,8 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"], optional = tr
 # metrics
 iroh-metrics = { version = "0.7.0", path = "../iroh-metrics", default-features = false }
 bounded_join_set = "0.1.0"
+lazy_static = "1.4.0"
```
dignifiedquire, line 81: `once_cell` please, we already use that

iroh-net/src/key.rs
dignifiedquire: you need to use `serialize_bytes`, see https://github.com/dalek-cryptography/ed25519-dalek/blob/main/src/verifying.rs#L630

iroh-net/Cargo.toml
```diff
@@ -79,6 +79,8 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"], optional = tr
 # metrics
 iroh-metrics = { version = "0.7.0", path = "../iroh-metrics", default-features = false }
 bounded_join_set = "0.1.0"
+ttl_cache = "0.5.1"
+serde_bytes = "0.11.12"
```
rklaehn, line 83: we can also write the weird visitors ourselves if you don't want the dependency. But it's small and from @dtolnay, so I'd rather not.

flub (commented):
Have you considered putting the cache into the magicsock struct rather than having it global?  It would make the change more invasive probably, but you have more flexibility.  E.g. you could potentially allow customising it etc.

dignifiedquire (approved):

Issue #1696, opened 2023-10-20
iroh-net: looking for derpers and mapped addresses as direct addresses in the peer map


divagant-martian:
what I see:
```log
2023-10-20T21:26:33.026066Z  WARN poll_recv{me=2afwy33ocwrot26y}: iroh_net::magicsock: recv packets: no peer state found, skipping src=x.x.x.x:55762 count=3363 len=28
2023-10-20T23:43:57.526962Z  WARN iroh_net::magicsock: recv packets: no peer state found, skipping src=34.253.75.5:3478 count=2140 len=44
```
the second one is the EU derper, the first one is my ip with a port that is not one of the ones I'm listening to, so maybe a mapped port.

I get this just by spinning up a node on a fresh setup.

Tracked this via git bisect to #1579 

Maybe everything works fine and these are just annoying warning. It's not clear to me what the effect of these really is

PR #1698, opened 2023-10-21
fix(net): correct packet math for poll_recv

dignifiedquire:
## Description

- `meta.len` is the length of the actually received data (not `buf.len()`)
- `meta.stride` is the length of an individual packet (not `meta.len`)

This was triggering https://github.com/n0-computer/iroh/issues/1696 because for STUN packets the additional (empty) buffer was considered to be QUIC packets, resulting in the warning.

### DIFF
--- a/iroh-net/src/magicsock.rs
+++ b/iroh-net/src/magicsock.rs
@@ -528,26 +528,28 @@ impl Inner {
             // find disco and stun packets and forward them to the actor
             loop {
                 let end = start + meta.stride;
-                if end > buf.len() {
+                if end > meta.len {
                     break;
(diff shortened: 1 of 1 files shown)


Frando (approved):
Great find!

divagant-martian:
I have no idea if it's hard to mock or not, but would it make sense to test that stun packages are correctly received on this end? If we were reporting stun packages as udp ones then stun ones were failing and CI didn't notice

dignifiedquire:
the stun packets were correctly reported, there were just additional packets reported, which is why tests didn’t pick this up

PR #1710, opened 2023-10-23
fix(net): correctly order binding of UDP socket 

dignifiedquire:
this fixes the case where a socket bound to `0` would only be available for usage later on.

This was seen when stun requests would time out, as the socket was not fully ready yet

TLDR: binding to `0` was probably taking longer, and we were setting `non_blocking` too early, so we weren't properly waiting for the binding to be setup (which is why it eventually recovered)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -41,9 +41,9 @@ dependencies = [
 
 [[package]]
 name = "ahash"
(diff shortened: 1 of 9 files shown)


dignifiedquire:
Superseeded by #1755

PR #1711, opened 2023-10-23
feat(iroh): add ticket prefixes and a `doctor ticket-inspect` command

divagant-martian:
## Description

Adds a `doc:` and `blob:` prefix to the doc and blob tickets respectively. This is mandatory and identifies the kind of iroh ticket.

## Notes & open questions

doctor sample run (happy path):
```
doctor ticket-inspect blob:edjoqrkky753mdwphqrxbr2wjzkizv4hkb5s5ovav73c2zfuaja7aaibaqalkpj5jcqzsaqawu6t2sfbvubqbqfiaaf4ivyaycuaadgek4adtjetziwzcpcdqzemuxmwei2fp5wtmfjcy4qysovyzjhahfao3aaa
```
```
Blob ticket:
Ticket {
    node: PeerAddr {
        peer_id: PublicKey(2luekswh7o3a5tz4),
        info: AddrInfo {
            derp_region: Some(
                1,
            ),
            direct_addresses: {
                181.61.61.72:34144,
                181.61.61.72:61486,
                192.168.0.12:11204,
                192.168.0.25:11204,
            },
        },
    },
    format: Raw,
    hash: Hash(
        39a493ca2d913c438648ca5d96223457f6d361522c721893ab8ca4e03940ed80,
    ),
    token: None,
}
```
doctor sample run (unhappy path):
```
 doctor ticket-inspect blob:EDJOQRKKY753MDWPHQRXBR2WJZKIZV4HKB5S5OVAV73C2ZFUAJA7AAIBAQALKPJ5JDQIUAQAWU6T2SFO4ABQBQFIAAGMIVYAYCU'
Error: failed parsing blob ticket

Caused by:
    0: decoding failed: invalid length at 98
    1: invalid length at 98
```

another unhappy path:
```
doctor ticket-inspect asds:EDJOQRKKY753MDWPHQRXBR2WJZKIZV4HKB5S5OVAV73C2ZFUAJA7AAIBAQALKPJ5JDQIUAQAWU6T2SFO4ABQBQFIAAGMIVYAYCUAAGOEK4ADTJETZIWZCPCDQZEMUXMWEI2FP5WTMFJCY4QYSOVYZJHAHFAO3AAA'
```
```
Error: unrecogized ticket prefix

Caused by:
    Matching variant not found
```

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


divagant-martian:
ping with love @b5

iroh/src/ticket.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    /// Deserialize from postcard bytes.
+    fn from_bytes(bytes: &[u8]) -> Result<Self, postcard::Error> {
+        postcard::from_bytes(bytes)
+    }
+
+    /// Serialize to string.
+    fn serialize(&self) -> String {
+        let mut out = Self::KIND.to_string();
+        out.push(':');
```
b5, line 74: ok, I want to propose we use `_` instead of `:` to separate. my reason: I can't double-click the ticket to get the entire string if we use the colon. If we use the underscore it'll work

divagant-martian: curious I thought this would happen but weirdly it works on my terminal

b5 (approved):

iroh/src/ticket/blob.rs
```diff
@@ -0,0 +1,151 @@
...
+        Ok(Self {
+            hash,
+            format,
+            node: peer,
+            token,
+        })
+    }
+
+    /// Deserializes from bytes.
+    pub fn from_bytes(bytes: &[u8]) -> Result<Self> {
```
dignifiedquire, line 61: This shadowing of method names is a bit tricky. Maybe add a `verify_ticket` method to the trait (which is called when doing `from_bytes`, with a default impl of `Ok(())` and implement the verification you are doing here with that?

dignifiedquire:
contains an empty file of `node.rs`  that can be removed

iroh/src/ticket/blob.rs
```diff
@@ -0,0 +1,151 @@
...
+    hash: Hash,
+    /// Optional Request token.
+    token: Option<RequestToken>,
+}
+
+impl IrohTicket for Ticket {
+    const KIND: Kind = Kind::Blob;
+}
+
+impl std::str::FromStr for Ticket {
```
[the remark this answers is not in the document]
divagant-martian: sadly no, due to orphan rules we need to own either the type over which the trait is implemented, or the implemented trait, and here we only own the bound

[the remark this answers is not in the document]
dignifiedquire: :(

dignifiedquire (approved):

divagant-martian:
in light of the changes done in #1738 I'm thinking there is no need to group the tickets together so not to overwrite the authorship of the code. What would you prefer @dignifiedquire to keep them together like this or where they currently are on main? Revert to where they are should be quick

PR #1718, opened 2023-10-24
refactor(iroh-net): split `endpoint` module and reduce `PeerMap` surface

Frando:
## Description

Refactors to prepare for *actual* refactors of `PeerMap`:

* Split the `endpoint` module into `peer_map` and `endpoint` to give them a managaeable size
* Access to `PeerMap` is now all through methods, no more direct access to the `PeerMapInner` and `Endpoint`s in there, `PeerMap::read` and `PeerMap::write` are removed
* Move some pure logic from `MagicSockInner` to `PeerMap`

This should contain *no* functional changes, only pure refactoring. IMO this gives us a better structure to start with actually refactoring `PeerMap` and `Endpoint` to be less bug-prone (through properly abstracted data structures with a limited API surface).

Best way to review is to go commit-by-commit.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


iroh-net/src/magicsock/peer_map.rs
```diff
@@ -0,0 +1,956 @@
...
+            (!peer_addr.info.is_empty()).then_some(peer_addr)
+        })
+    }
+
+    /// Create a new [`PeerMap`] from data stored in `path`.
+    fn load_from_file(path: impl AsRef<Path>) -> anyhow::Result<Self> {
+        let path = path.as_ref();
+        ensure!(path.is_file(), "{} is not a file", path.display());
+        let mut me = PeerMapInner::default();
+        let contents = std::fs::read(path)?;
```
dignifiedquire, line 304: why is this not async?

iroh-net/src/magicsock/peer_map/endpoint.rs
```diff
@@ -108,10 +102,10 @@ pub struct PendingCliPing {
 
 #[derive(Debug)]
 pub(super) struct Options {
-    pub(super) public_key: PublicKey,
-    pub(super) derp_region: Option<u16>,
+    pub public_key: PublicKey,
+    pub derp_region: Option<u16>,
```
dignifiedquire, line 106: why is this `pub` if the struct itself is `pub(super)`?

iroh-net/src/magicsock/peer_map/endpoint.rs
```diff
@@ -878,7 +872,7 @@ impl Endpoint {
         self.pending_cli_pings.clear();
     }
 
-    fn last_ping(&self, addr: &SendAddr) -> Option<Instant> {
+    pub fn last_ping(&self, addr: &SendAddr) -> Option<Instant> {
```
dignifiedquire, line 875: `pub(super)`?

Frando: the whole `Endpoint` struct is only `pub(super)`, so no need to add it to all methods I thought?

iroh-net/src/magicsock/peer_map/endpoint.rs
Frando: changed to `pub(super)`

iroh-net/src/magicsock/peer_map/endpoint.rs
Frando: changed it all to `pub(super)`

dignifiedquire (approved):
haven't reviewed each moved code snippet, assuming this is just copy & paste

PR #1719, opened 2023-10-24
fix unused export found in nightly rust

dignifiedquire:
Actual issue was a double reexport which went unnoticed in older rustc versions. 

Thanks to @bvanjoi in https://github.com/rust-lang/rust/issues/117120#issuecomment-1777113406 for pointing out the problem


rklaehn (approved):
if it works...

PR #1720, opened 2023-10-24
Use an option for last_probe

rklaehn:
## Description

`Instant::now() - AVAILABILITY_TRUST_DURATION` will panic if the uptime of the system is less than AVAILABILITY_TRUST_DURATION.

This will happen frequently with freshly spun up machines for CI.

## Notes & open questions

I refactored last_probe to an option. I was expecting to get an error message somewhere and having to fix something, but I got nothing.

Where is last_probe ever updated?

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


dignifiedquire (commented):
this seems odd, according to this code itwill never be set to `Some`

rklaehn:
> this seems odd, according to this code itwill never be set to `Some`

Yes, see note.

PR #1722, opened 2023-10-24
fix(iroh-net): correctly set the time in which a probe is created

divagant-martian:
## Description

- the time a probe is done should be `now`
- remove a `Default` impl with side effects
- still initialize the first probe as expired when possible

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.


iroh-net/src/portmapper.rs
```diff
@@ -415,12 +413,20 @@ impl Service {
         rx: mpsc::Receiver<Message>,
     ) -> (Self, watch::Receiver<Option<SocketAddrV4>>) {
         let (current_mapping, watcher) = CurrentMapping::new();
+        let mut full_probe = Probe::empty();
+        if let Some(in_the_past) = full_probe
+            .last_probe
+            .checked_sub(AVAILABILITY_TRUST_DURATION)
+        {
+            // we want to do a first full probe, so set is as expired on start-up
+            full_probe.last_probe = in_the_past;
```
rklaehn, line 422: So this means that we won't do a probe on startup for boxes where uptime < AVAILABILITY_TRUST_DURATION.

If that is acceptable, yes this works fine.

rklaehn (approved):

iroh-net/src/portmapper.rs
divagant-martian: yes, that would be the side effect, delaying the first probe. I find it acceptable too

PR #1724, opened 2023-10-24
refactor(iroh-net): improve `PeerMap` and `Endpoint` abstractions

Frando:
## Description

* refactor: Make the abstractions around `PeerMap` and `Endpoint` more solid:
  * Better getter API on `PeerMap`, making sure that no mismatch between the indexes and endpoints can emerge
  * No more public acccess to endpoint fields from the peermap (only methods)
* fix: Improvments to tracing spans, fields, levels
* fix: Actually send out pings after handling call-me-maybe messages (we did not do that)
* tests: add a test that recreates a node in a loop and connects to the same server node over derp

## Notes & open questions

The new test passes for me locally but still is super slow (~40 seconds for a loop of 5 times bind & connect & transfer 20 bytes). Something is still not correct. It should never take 10 seconds to connect and transfer 20 bytes.

I think I need some help here - my guess is now that something in the derper is not correct, but not sure.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/src/disco.rs
+++ b/iroh-net/src/disco.rs
@@ -224,7 +224,6 @@ impl Pong {
 impl CallMeMaybe {
     fn from_bytes(ver: u8, p: &[u8]) -> Result<Self> {
         ensure!(ver == V0, "invalid version");
-        // ensure!(p.len() % EP_LENGTH == 0, "invalid entries");
 
         let num_entries = p.len() / EP_LENGTH;
         let mut m = CallMeMaybe {
@@ -365,7 +364,12 @@ mod tests {
             Test {
                 name: "call_me_maybe",
                 m: Message::CallMeMaybe(CallMeMaybe { my_number: Vec::new(), clear_pongs: false }),
-                want: "03 00",
+                want: "03 00 00",
+            },
+            Test {
+                name: "call_me_maybe_clear_pongs",
+                m: Message::CallMeMaybe(CallMeMaybe { my_number: Vec::new(), clear_pongs: true }),
+                want: "03 00 01",
             },
             Test {
                 name: "call_me_maybe_endpoints",
@@ -376,7 +380,7 @@ mod tests {
                     ],
                     clear_pongs: false
                 }),
-                want: "03 00 00 00 00 00 00 00 00 00 00 00 ff ff 01 02 03 04 37 02 20 01 00 00 00 00 00 00 00 00 00 00 00 00 34 56 15 03",
+                want: "03 00 00 00 00 00 00 00 00 00 00 00 ff ff 01 02 03 04 37 02 20 01 00 00 00 00 00 00 00 00 00 00 00 00 34 56 15 03 00",
             },
         ];
         for test in tests {
--- a/iroh-net/src/disco.rs
+++ b/iroh-net/src/disco.rs
@@ -143,6 +143,8 @@ pub struct Pong {
 pub struct CallMeMaybe {
     /// What the peer believes its endpoints are.
     pub my_number: Vec<SocketAddr>,
+    /// If true invalidate all previous pongs. This is sent after a total state reset.
+    pub clear_pongs: bool,
 }
 
 impl Ping {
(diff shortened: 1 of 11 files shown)


iroh-net/src/magic_endpoint.rs
```diff
@@ -759,6 +759,9 @@ mod tests {
                 let client_secret_key = client_secret_key.clone();
                 let fut = async move {
                     info!("client binding");
+                    info!("!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!1");
```
dignifiedquire, line 762: beautiful

iroh-net/src/disco.rs
```diff
@@ -143,6 +143,8 @@ pub struct Pong {
 pub struct CallMeMaybe {
     /// What the peer believes its endpoints are.
     pub my_number: Vec<SocketAddr>,
+    /// If true invalidate all previous pongs. This is sent after a total state reset.
+    pub clear_pongs: bool,
```
rklaehn, line 147: Can't we infer if the pongs need to be cleared somehow?

If we really have to have this field, I would prefer to have the field name be a description of why this `CallMeMaybe` is different than the normal ones rather than an instruction about what to do...

Frando: > Can't we infer if the pongs need to be cleared somehow?
I spend some time thinking about this and didn't find anything. Only idea I had was lowering the maximum age of the most recent pong of the best_addr for it to be considered valid (currently 6 seconds). Other ideas welcome.

iroh-net/src/magicsock.rs
```diff
@@ -942,6 +986,52 @@ impl Inner {
...
+                endpoints.last_endpoints_time
+            );
+
+            let actor_sender = self.actor_sender.clone();
+
+            tokio::task::spawn(async move {
+                actor_sender
+                    .send(ActorMessage::ReStun("refresh-for-peering"))
+                    .await
+                    .unwrap();
```
rklaehn, line 1005: .ok() if we don't care if it succeeds or not?

Can this also be a try_send without the task?

iroh-net/src/magicsock.rs
```diff
@@ -1121,29 +1204,30 @@ impl MagicSock {
             closing: AtomicBool::new(false),
             closed: AtomicBool::new(false),
             derp_recv_receiver,
-            network_recv_wakers: std::sync::Mutex::new(None),
-            network_send_wakers: std::sync::Mutex::new(None),
+            network_recv_wakers: parking_lot::Mutex::new(None),
```
rklaehn, line 1207: why parking lot? Perf wise the stdlib ones are comparable since some time. Because of the deadlock detection feature?

iroh-net/src/magicsock.rs
Frando: This was like this before.

I thought if we could do try_send, but am not sure - if the actor channel were full, the restun wouldn't happen. But might still happen anyway after a timeout? @dignifiedquire maybe knows more?

iroh-net/src/magicsock.rs
Frando: Actually no real reason, just that it cannot panic? Can change back to `std`. We have a mix of both mutexes in here, maybe we should settle for one.

dignifiedquire: I used parking_lot in a couple of places because it has better support for mapping locks. That's the only reason I found to use them over std

iroh-net/src/magicsock.rs
dignifiedquire: We shouldn't `unwrap`. `try_send` makes more sense to me, given this is non critical to do

iroh-net/src/magicsock.rs
rklaehn: It used to be that parking lot was insanely fast compared to stdlib. And of course that you don't have to do the annoying unwrap everywhere. But the stdlib perf has catched up, so these days most of the time I use std. But the deadlock detector of parking lot is neat.

iroh-net/src/disco.rs
Frando: Moved `clear_pongs` feature to #1742

iroh-net/src/magicsock.rs
Frando: Removed the unwrap but kept the task for now.

iroh-net/src/magicsock.rs
```diff
@@ -1002,7 +1002,7 @@ impl Inner {
                 actor_sender
                     .send(ActorMessage::ReStun("refresh-for-peering"))
                     .await
-                    .unwrap();
+                    .ok()
```
rklaehn, line 1005: Still would like to turn this into a try_send if possible, but that can happen later? Maybe a todo: investigate if try_send is ok here?

rklaehn (approved):

iroh-net/src/magicsock.rs
Frando: Yes I have a followup already that changes this. Will push to a new PR.

PR #1736, opened 2023-10-26
fix(net): do not dial regions in parallel

dignifiedquire:
These are supposed to be dialed in order, not in parallel.

This should also fix the warnings in the logs about aborted dials.


rklaehn:
So what does dialing a region mean anyway? And in what order will they be dialed?

dignifiedquire:
Yes this doesn't really trigger currently, as we do not have multiple nodes in a region setup. But if we do, it is still better to dial one by one, as we only need one. 

The order is based on how the config is done.

rklaehn (approved):

PR #1738, opened 2023-10-26
Made sure the following types also work with a "human readable" ser/de

rklaehn:
## Description

Made sure the following types also work with a "human readable" ser/de

- Hash
- HashAndFormat
- PublicKey
- Ticket

## Notes & open questions

Shall we just switch to hex for hashes so they are identical to b3sum? Sorry, but I had to ask...

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


dignifiedquire (approved):
cool, I didn't know serde allowed configuration like this

PR #1739, opened 2023-10-26
fix(iroh): handle rpc args in any position

### LINKED ISSUE #1639: cli: rpc-port argument is confusing and partially broken
Currently when running `iroh --rpc-port PORT start` it will be ignored, but for all other commands this is the position it has to be. For it to be accepted `iroh start --rpc-port PORT` has to be used.

dignifiedquire:
It is now possible to do any of these 

```
> iroh start --rpc-port 1234
> iroh --rpc-port 1234 start
> iroh console --rpc-port 1234
> iroh --rpc-port 1234 console
```

Closes #1639

### DIFF
--- a/iroh/src/commands/node.rs
+++ b/iroh/src/commands/node.rs
@@ -1,8 +1,6 @@
 use std::{
-    fmt,
     net::{Ipv4Addr, SocketAddr, SocketAddrV4},
     path::PathBuf,
-    str::FromStr,
     sync::Arc,
 };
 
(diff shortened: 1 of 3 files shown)


iroh/src/commands/node.rs
```diff
@@ -28,7 +28,7 @@ use super::{BlobAddOptions, MAX_RPC_CONNECTIONS, MAX_RPC_STREAMS};
 #[derive(Debug)]
 pub struct StartOptions {
     pub addr: SocketAddr,
-    pub rpc_port: RpcPort,
+    pub rpc_port: u16,
```
rklaehn, line 31: So we can't start with rpc port disabled anymore?

dignifiedquire: For now I have removed the ability, yes. Lets work on getting the portless version for rpc supported soon to improve on this.

Arqu:
Something doesn't sit right, netsim works with this though it doesn't pass in `--rpc-port` at start.

Arqu (approved):
Other than it magically working when it maybe shouldn't, LGTM.

PR #1742, opened 2023-10-26
feat: clear pongs on initial connect via Derp

Frando:
## Description

TLDR: If you connect to a node, abort, and then reconnect from a new magic socket over Derp, the other node will likely reply over UDP (because that's what its state indicates to do) but you won't be able to receive those messages. To solve, upon reconnecting send a message to the other node "please reset my address state".

Explanation of the problem:

* Node A binds
* Node B binds, connects to A by nodeid and derp region
* Node A and B communicate, do call-me-maybe, upgrade to UDP
* Node A now has a UDP `best_addr` set for node A. It is renewed on each ping-pong (every 1s) and has a valid time of 6s currently
* Node B shutdown
* Node B respawn, reconnect to Node A by nodeid and derp region
* Node A recvs packets by node B over Derp **but replies over UDP because it still has a best_addr set**
* Node B either doesn't receive the packets (because it bound to a new port) or discards them (because it does not know to which peer the addr belongs)
* Node A still has no reason to switch to Derp until its best_addr times out

To solve this situation, this PR does...

* Add a `clear_pongs` optional field to the `CallMeMaybe` disco message
* When connecting to a so-far unused node over Derp, first send a `CallMeMaybe` message with `clear_pongs: true`
* Even send this message right away, before we determined our own endpoints (currently all `CallMeMaybe` messages are held back and only sent out after we completed STUN, which takes a long time unforunately, but that's yet a different issue)
* When receiving `clear_pongs: true` , the node will reset the pong state of all addrs of the endpoint entry in the peer map for that node, and invalidate the `best_addr`, which means that it will send over Derp until a new ping-pong completes

The change in the wire protocol is backwards comaptible.

I'm very open to ideas how to solve this situation without a new message, but I couldn't come up with any alternative so far.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/src/disco.rs
+++ b/iroh-net/src/disco.rs
@@ -143,6 +143,8 @@ pub struct Pong {
 pub struct CallMeMaybe {
     /// What the peer believes its endpoints are.
     pub my_number: Vec<SocketAddr>,
+    /// If true invalidate all previous pongs. This is sent after a total state reset.
+    pub clear_pongs: bool,
 }
 
 impl Ping {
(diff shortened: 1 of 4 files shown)


rklaehn:
So what is the problem with just resetting the pong state on every CallMeMaybe? Let's say A and B are connected via UDP and everything is great. Under what circumstances would A get a CallMeMaybe from B in that case?

So the reset_pongs is basically true for the first msg B sends to A and then false afterwards, so if we had a `call_me_maybe_count: u64`, it would be `call_me_maybe_count == 0`?

Frando:
> So what is the problem with just resetting the pong state on every CallMeMaybe? Let's say A and B are connected via UDP and everything is great. Under what circumstances would A get a CallMeMaybe from B in that case?

Not sure. We didn't do that previously. I'll have to check first under which conditions call me maybes are sent.

Maybe it could work to invalidate pongs for all endpoints not included in the call-me-maybe.

Frando:
Closing in favor of #1752

PR #1743, opened 2023-10-26
clippy: warn on unsused async fn

dignifiedquire:
Details: https://rust-lang.github.io/rust-clippy/master/index.html#/unused_async

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,2 @@
 [build]
-rustflags = ["-Wmissing_debug_implementations"]
+rustflags = ["-Wmissing_debug_implementations", "-Wclippy::unused-async"]
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,2 @@
 [build]
-rustflags = ["-Wmissing_debug_implementations", "-Wclippy::unused-async"]
+rustflags = []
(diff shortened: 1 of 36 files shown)


Frando:
This shouldn't pass, right? E.g. https://github.com/n0-computer/iroh/blob/main/iroh-net/src/magicsock.rs#L2011

rklaehn:
I can not put in words how much I love this lint.

So many times where you have some code that you wish to be sync but you can't do that because it calls some code that is async.

This alone puts the cost/benefit ratio of clippy solidly in the positive despite all my complaints.

rklaehn (approved):

dignifiedquire:
@rklaehn I added some exceptions in the `iroh-bytes` code, is that the right way, or should I change those functions?

rklaehn:
> @rklaehn I added some exceptions in the `iroh-bytes` code, is that the right way, or should I change those functions?

They are async for 2 reasons:

1. it used to be a trait
2. if we want to lazily load hash seqs, we would need them to be async

So I think they should remain async because of 2. Eventually we will want to lazily load them, but I did not see much urgency since so far we don't have HashSeqs with billions of entries.

dignifiedquire:
will rebase and merge once the current work around magicsock settles down a little bit 😅

PR #1752, opened 2023-10-27
refactor(iroh-net): call-me-maybe improvements (no more tasks for queue, better logic on recv)

Frando:
## Description

Removes the task spawning for queued call-me-maybes. Instead just keep them in a hashmap. This is possible now because sending call-me-maybes is sync anyway. We will lose them if the derp channel is full, which might be a problem, but that is already a problem on main.

Additionally removes `ActorMessage::ReStun` and instead moves the `EndpointUpdateState` onto `Inner`. One less thing to go over the main channel - not needed, because the `EndpointUpdateState` already is a concurrent data structure (with a mutex added for the `next_update`).

Also improves (hopefully) the logic when receiving a call-me-maybe: It now clears the `recent_pong` for all addresses not included in the call-me-maybe, and clears the `best_addr` trust if not included in the call-me-maybe.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/src/magicsock.rs
+++ b/iroh-net/src/magicsock.rs
@@ -266,6 +266,9 @@ struct Inner {
     /// List of CallMeMaybe disco messages that should be sent out after the next endpoint update
     /// completes
     pending_call_me_maybes: parking_lot::Mutex<HashMap<PublicKey, u16>>,
+
+    /// Indicates the update endpoint state.
+    endpoints_update_state: EndpointUpdateState,
 }
 
 impl Inner {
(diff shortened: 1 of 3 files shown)


iroh-net/src/magicsock.rs
```diff
@@ -2090,26 +2046,30 @@ impl Actor {
...
 
-        if self.inner.endpoints.lock().set(&eps) {
+        if self.inner.endpoints.write().set(&eps) {
             log_endpoint_change(&eps);
             if let Some(ref cb) = self.inner.on_endpoints {
                 cb(&eps[..]);
             }
+            for (public_key, region_id) in self.inner.pending_call_me_maybes.lock().drain() {
+                self.inner
+                    .send_or_queue_call_me_maybe(region_id, public_key);
```
dignifiedquire, line 2056: shouldn't this force sending instead of potentially requeuing? not sure in either direction

dignifiedquire (approved):
looks good, one question

iroh-net/src/magicsock.rs
Frando: Yea it won't requeue because endpoints.set was just called which means endpoints.is_fresh will always be true.
We could add a separate send_call_me_maybe method for clarity though.

dignifiedquire: I think that would be good, to make the intent clear (given the complexity of this code.. )

Frando: Changed it to separate functions.

dignifiedquire (approved):

PR #1755, opened 2023-10-27
fix: do not block on dropping UDP sockets

dignifiedquire:
Observed `drop(std::net::UdpSocket)` to block for up to 1s on macOS

> This makes me sad 😭 

- Introduce `net::UdpSocket` wrapper that handles the issue
- Improve binding logic 
- Make sure to use the new `UdpSocket` everywhere in `iroh-net`


rklaehn:
can you avoid the whole wrapper affair by just cloning the socket (into/from raw fd whatever)? It is plausible that the libc::close will only block for 1s when the *last* reference to the fd is removed, so that way you could just isolate this bullshit to the Drop impl of RebindingUdpSocket

dignifiedquire:
> the Drop impl of RebindingUdpSocket

No because we don't use `RebindingUdpSocket` in hairpin.rs

rklaehn:
> > the Drop impl of RebindingUdpSocket
> 
> No because we don't use `RebindingUdpSocket` in hairpin.rs

Ah, so it is not wrapped in an arc at all?

but then you could just directly move it into a spawn_blocking?

And for the RebindingUdpSocket itself, forget about the refcount stuff, there is this nice thing: https://doc.rust-lang.org/std/sync/struct.Arc.html#method.into_inner

dignifiedquire:
sure I could do a manual thing in hairpin.rs, but I would like to avoid having this absurd logic in multiple places, which is why I wrote the wrapper

rklaehn (approved):
I disapprove that you have to do this, but... 🤷‍♂️

rklaehn:
Still not sure about the wrapper. If you have to do this manually in 2 places, and only for some OSes, that seems preferable to the wrapper. But both options really suck...

iroh-net/src/net/network.rs
```diff
@@ -0,0 +1,46 @@
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
+
+/// Network selection between Ipv4 and Ipv6.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum Network {
```
flub, line 5: As this is now fully `pub`, can we **please, pretty please** rename this to `IpFamily`.  This name has been bothering me since the start...

iroh-net/src/net/network.rs
```diff
@@ -0,0 +1,46 @@
...
+        match value {
+            IpAddr::V4(_) => Self::Ipv4,
+            IpAddr::V6(_) => Self::Ipv6,
+        }
+    }
+}
+
+impl Network {
+    /// Returns the matching default address.
+    pub fn default_addr(&self) -> IpAddr {
```
flub, line 23: should this not be better called `unspecified_addr`?

iroh-net/src/net/udp.rs
```diff
@@ -0,0 +1,150 @@
...
+        Self::bind(Network::Ipv4, port).await
+    }
+
+    /// Bind only Ipv6 on any interface.
+    pub async fn bind_v6(port: u16) -> Result<Self> {
+        Self::bind(Network::Ipv6, port).await
+    }
+
+    /// Bind only Ipv4 on localhost.
+    pub async fn bind_local_v4(port: u16) -> Result<Self> {
```
flub, line 28: This is binding to `*` just like the `bind_v4` one does?

iroh-net/src/net/udp.rs
```diff
@@ -0,0 +1,150 @@
...
+
+        if addr.port() != 0 {
+            let local_addr = socket.local_addr()?;
+            ensure!(
+                local_addr.port() == addr.port(),
+                "wrong port bound: {:?}: wanted: {} got {}",
+                network,
+                addr.port(),
+                local_addr.port(),
+            );
```
flub, line 108: In what situations does this assertion ever fail?  I mean, I would never have thought of writing such an assertion and am horrified this might be needed.

iroh-net/src/net/udp.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    fn deref(&self) -> &Self::Target {
+        self.0.as_ref().expect("only removed on drop")
+    }
+}
+
+impl Drop for UdpSocket {
+    fn drop(&mut self) {
+        let std_sock = self.0.take().expect("not yet dropped").into_std();
+        tokio::task::spawn_blocking(move || {
```
flub, line 145: this panics if it's being dropped outside of a tokio runtime.  That's... scary.

Could we explicitly use `tokio::runtime::Handle::try_current` and do a sync drop if not in a runtime?  Or would that defeat the point?

flub (commented):
I thought there was a way to deny use of a particular struct, like `tokio::net::UdpSocket` or `std::net::UdpSocket` but can't seem to find it now.  cargo-deny can only deny entire crates.  Maybe I imagined this.

iroh-net/src/magicsock.rs
```diff
@@ -2399,19 +2360,20 @@ fn new_re_stun_timer(initial_delay: bool) -> time::Interval {
 
 /// Initial connection setup.
 async fn bind(port: u16) -> Result<(RebindingUdpConn, Option<RebindingUdpConn>)> {
-    let ip6_port = if port != 0 { port + 1 } else { 0 };
+    let pconn4 = RebindingUdpConn::bind(port, Network::Ipv4)
+        .await
+        .context("bind IPv4 failed")?;
+    let ip4_port = pconn4.local_addr()?.port();
+    let ip6_port = ip4_port + 1;
```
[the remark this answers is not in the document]
dignifiedquire: That was always the original intention (and we did so before, if we didn't bind to port 0`) The idea is to not choose just random ports, but try and allocate two ports together. The underlying code always falls back to using `0` if it is already used

iroh-net/src/netcheck/reportgen/hairpin.rs
```diff
@@ -100,37 +85,31 @@ enum Message {
...
         match self.run_inner().await {
             Ok(_) => debug!("hairpin actor finished successfully"),
             Err(err) => error!("Hairpin actor failed: {err:#}"),
         }
     }
 
-    async fn run_inner(&mut self) -> Result<()> {
-        let sock = UdpSocket::bind("0.0.0.0:0")
+    async fn run_inner(self) -> Result<()> {
+        let socket = UdpSocket::bind_v4(0)
```
flub, line 102: I did also wonder in the past if this would not be better served by keeping this socket around for a long time.  The netcheck actor stays alive for as long as the magicsock is, so storing this socket there and handing a clone to each hairpin could be an improvement.  Especially now we know this is rather expensive on macos maybe that could be worth the additional complexity.  But if we think so perhaps create an issue and do it in a separate PR.

iroh-net/src/net/udp.rs
dignifiedquire: uups

iroh-net/src/net/udp.rs
dignifiedquire: I had a lot of fun debugging network bindings not doing what I expect, and I agree this should never happen..but my paranoia is strong at this point..

iroh-net/src/net/udp.rs
dignifiedquire: ah yes, I should do that

iroh-net/src/netcheck/reportgen/hairpin.rs
dignifiedquire: > But if we think so perhaps create an issue and do it in a separate PR.

Yes, lets defer this. I agree it would be better, but doing so required more changes to the code than I wanted when I tried to do it.

iroh-net/src/magicsock.rs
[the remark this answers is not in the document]
flub: yeah, i know this was already from before but was curious why.  thanks for the reasoning.

I guess I was surprised it wasn't trying to use the same port as the IPv4 one, if the reason is to keep them close by using the same port on IPv4 and IPv6 would be nicer?

iroh-net/src/net/network.rs
dignifiedquire: done

iroh-net/src/net/udp.rs
dignifiedquire: fixed

iroh-net/src/net/network.rs
dignifiedquire: fixed

iroh-net/src/magicsock.rs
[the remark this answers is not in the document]
dignifiedquire: > it wasn't trying to use the same port as the IPv4 one,

This is based on the tailscale code which did not do this. Last time I did research on reusing ports it was possible, but just not super reliable on OS combinations, this setup is trying to work around all odd and broken things when it comes to ipv4/ipv6 setups..

iroh-net/src/net/udp.rs
dignifiedquire: fixed

flub (approved):

PR #1758, opened 2023-10-30
feat(iroh): pass a runtime to Doc client to spawn close task on drop

dignifiedquire:
We spawn a tokio task when dropping a client `Doc` handle to notify the node that the document was closed. This does not work in the FFI currently because the FFI does not run in a global tokio context. Instead we now always pass a `runtime::Handle` into the client which can be used to spawn tasks independently of running in a tokio context.


dignifiedquire (commented):
lgtm

Frando (approved):

Issue #1388, opened 2023-08-22
PublicKey & SecretKey bikeshedding


flub:
# Background

The cryptographic keys and identities have been reduced to just two types right now:

- `PublicKey`
- `SecretKey`

This `PublickKey` type is used for everything: identifying an iroh node (or peer, formerly done by `PeerId`) as well as encrypting messages to the iroh node (formerly done by the node key).  The `SecretKey` is the obvious secret part for the node itself.  This does not require much explanation (which is the point!).

# Issue

However in some places this naming arguably reduces readability a little:

- types like `Event<PublicKey>`
- parameters like `from: PublicKey`

One could argue that the former `PeerId` was more readable in these places.  However `PeerId` was purposefully removed since this was confusing wrt what is a "peer", what is an "iroh node" etc.

# Proposal

One possible suggestion is to name the new types slightly different:

- `NodeId`
- `NodeSecret`

The hope here is that these names make more sense to someone reading code using the public API without much background.  These terms try to convey a little more meaning than the current more generic names.

----

Other suggestions are of course welcome, this is a bikeshedding issue!

dignifiedquire:
Lets add 

```rust
pub type NodeId = PublicKey;
pub type NodeSecret = SecretKey;
```

to `iroh-net `.

Arguments and functions should use `NodeId` if it is in reference to a a node, and `PublicKey` if it is only in reference to cryptographic references.
eg 
- `encrypt(key: PublicKey)`
- but `send_to(node: NodeId)`

dignifiedquire:
A first pass has been done, if you see peers flying around please make sure to rename them to nodes.

PR #1765, opened 2023-11-01
refactor(iroh-net): rename PeerAddr to NodeAddr, introduce NodeId alias

divagant-martian:
## Description
First step on dealing with the whole peer_id vs node_id. Changes include:
- renames `PeerAddr` to `NodeAddr` and it's inner fields from `peer_id` to `node_id`
- introduces the `NodeId` type alias as stated in #1388 
- moves every field and parameter defined as `peer_id: PublicKey` to `node_id: NodeId`. As far as I can tell everywhere touched is not doing any crypto stuff.

## Notes & open questions

The one place where this is not changed is the tls code as it does deal with crypto ops. Later refactors should update the code to use `key: PublicKey` instead of `peer_id: PublicKey` for example.

In the future it might be worth to actually rename `PublicKey` to `NodeId` or creating a newtype `NodeId` that can only be obtained from a `PublicKey` but has no crypto ops available. Not doing any of that since I know it's controversial.

Also note that this is intentionally kept partial. Reasons for this:
- the state already has a mixed peer_id/node_id state, so we are not passing from a consistent state into an inconsistent one, but from a less desirable and inconsistent one to a more desirable, still inconsistent one
- keeping the PR short to facilitate a meaningful review

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/src/magic_endpoint.rs
+++ b/iroh-net/src/magic_endpoint.rs
@@ -20,18 +20,18 @@ pub use super::magicsock::EndpointInfo as ConnectionInfo;
 
 /// A peer and it's addressing information.
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
-pub struct PeerAddr {
+pub struct NodeAddr {
     /// The node's public key.
-    pub peer_id: PublicKey,
-    /// Addressing information to connect to [`Self::peer_id`].
+    pub node_id: PublicKey,
+    /// Addressing information to connect to [`Self::node_id`].
     pub info: AddrInfo,
 }
 
-impl PeerAddr {
-    /// Create a new [`PeerAddr`] with empty [`AddrInfo`].
-    pub fn new(peer_id: PublicKey) -> Self {
-        PeerAddr {
-            peer_id,
+impl NodeAddr {
+    /// Create a new [`NodeAddr`] with empty [`AddrInfo`].
+    pub fn new(node_id: PublicKey) -> Self {
+        NodeAddr {
+            node_id,
             info: Default::default(),
         }
     }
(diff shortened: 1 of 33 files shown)


iroh-net/src/magic_endpoint.rs
```diff
@@ -422,28 +422,28 @@ impl MagicEndpoint {
     /// If the `derp_region` is not `None` and the configured DERP servers do not include a DERP node from the given `derp_region`, it will error.
     ///
     /// If no UDP addresses and no DERP region is provided, it will error.
-    pub async fn connect(&self, peer_addr: PeerAddr, alpn: &[u8]) -> Result<quinn::Connection> {
-        self.add_peer_addr(peer_addr.clone())?;
+    pub async fn connect(&self, node_addr: NodeAddr, alpn: &[u8]) -> Result<quinn::Connection> {
+        self.add_peer_addr(node_addr.clone())?;
```
dignifiedquire, line 426: should this method be renamed?

iroh-net/src/magicsock/peer_map.rs
```diff
@@ -258,10 +258,10 @@ impl PeerMap {
 impl PeerMapInner {
     /// Get the known peer addresses stored in the map. Peers with empty addressing information are
     /// filtered out.
-    fn known_peer_addresses(&self) -> impl Iterator<Item = PeerAddr> + '_ {
+    fn known_peer_addresses(&self) -> impl Iterator<Item = NodeAddr> + '_ {
```
dignifiedquire, line 261: rename method?

iroh-net/src/magicsock/peer_map.rs
```diff
@@ -258,10 +258,10 @@ impl PeerMap {
 impl PeerMapInner {
```
dignifiedquire, line 258: rename?

dignifiedquire (approved):
changes look good to me in general, what's the plan with things like `PeerMap`?

Frando:
There's also still `MagicEndpoint::peer_id` I think.
What about a `s/peer_id/node_id` across the whole repo to also rename variables etc?

PR #1701, opened 2023-10-23
[WIP] Querys & Views

dignifiedquire:
## Description

Implementation of #1667

Issue #1667, opened 2023-10-18
Proposal: Document Views


b5:
I've had time to play with iroh a bunch now that prefix-based deletion is in, and I must say, it's been very fun learning how iroh works 😜. With that said, I think prefix deletion _and_ author ID's as key prefixes are too much for a user to get their head around on first go. As a user when I hear "key value store", I expect all keys to be unique _based solely on the key I provide_.

We have some new UI that at least illustrates the problem:

<img width="990" alt="Screenshot 2023-10-18 at 12 30 43 PM" src="https://github.com/n0-computer/iroh/assets/1154390/66dc5559-ef0b-4ccb-86e0-59d6de858c7b">

In this example there are two `foo` keys in the same document, because they have different authors. What's more, because I am neither of these authors, I cannot edit these keys, which is super confusing. As someone who works at n0, I fully understand why I can't edit stuff in someone else's subspace, so I'm not advocating we change anything about the way iroh works, instead I want a new "views" feature that gives me a "merged" view of the document, which is a union of all known replica keys, using timestamps to break ties.

Iroh is missing a true "shared" view of a document. I need this ASAP because it's too hard to introduce users to both the author-replica thing _and_ the prefix thing at the same time. I think the right answer is to start with teaching prefix deletion, but have the default display be this merged view that will feel like a true key-value store. That shared view will naturally invite the "but wait, just anyone can edit the db!?" question, which will open us up to talk about how there are ways to change the view to match individual authors, or the full-blast "all" view.

## Views
The term view here means "view" in the classical database sense:

> a [view](https://en.wikipedia.org/wiki/View_(database)) is a virtual [table](https://en.wikipedia.org/wiki/Table_(database)) representing the result of a [database](https://en.wikipedia.org/wiki/Database) [query](https://en.wikipedia.org/wiki/Information_retrieval).

We already have the concept of _filters_ which perform largely the same function, but co-mingle querying with views:

https://github.com/n0-computer/iroh/blob/8169053a0605f848ac3bade7b09833b4636333a6/iroh-sync/src/store.rs#L141-L155

I think the `All` and `Author` are views, with `Key` being an implicit query on the `All` view, `AuthorAndPrefix` is an implict query on the `Author` view.

## Merged View
A merged view takes the following steps:

1. construct the union of all entries across all authors, sorting by `key` field
[...]

PR #1766, opened 2023-11-02
feat(iroh-sync): Queries and "views"

Frando:
## Description

Implementation of #1667, extends #1701

With the following changes after discussions in Discord:

* Remove the notion of views, instead embed the info in the query. If the high-level concept of views makes more sense for people from an API perspective, we can restore it in the client api. However under the hood the query details are different per view, this is why here I expose a single `Query` struct
* Add a query builder with a typestate to only allow possible combinations
* Add an index to the redb (fs) document store to make queries that are sorted by key, or that are filtered by key but not by author, efficient. This also is what allows to do queries for the "latest only" entry per key, without allocating the full result set.

In the process I did a refactoring of the redb store to be safer to use. Especially, I moved the `ouroboros` self-referencing stuff into a `ranges` module, and encapsulated in an inner type to keep the self-referencing compilcations scoped. Also did something I had mind for a while: Add some type-safe abstractions around the range bounds constructions that are used when selecting on redb tables. All this turned out quite nice, IMO.

Also contains the changes from #1772 : 
* Renames `get_one` to `get_exact`
* Add flag to `get_exact` whether to include empty entries or not
* Add `get_one` to `iroh::client::Doc` to get a single entry with the same query mechanisms as `get_many`

Open questions and tasks:
* Naming review of the query builder methods
* Integration of the query parameters in the console

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.


dignifiedquire:
I think you need to rebase 😅 lots of conflicts

iroh-sync/src/store.rs
```diff
@@ -143,233 +142,243 @@ pub trait Store: std::fmt::Debug + Clone + Send + Sync + 'static {
...
+    pub fn key(key: impl AsRef<[u8]>) -> QueryBuilder<FlatQuery> {
+        Self::all().key_exact(key)
     }
-}
 
-impl From<std::ops::Range<u64>> for Range {
-    fn from(value: std::ops::Range<u64>) -> Self {
-        Range::Exclusive(value.start, value.end)
+    /// Create a [Query::all] query filtered by a key prefix.
+    pub fn prefix(prefix: impl AsRef<[u8]>) -> QueryBuilder<FlatQuery> {
```
dignifiedquire, line 305: maybe use `key_exact` and `key_prefix` for these too?

iroh-sync/src/store.rs
```diff
@@ -143,233 +142,243 @@ pub trait Store: std::fmt::Debug + Clone + Send + Sync + 'static {
...
+}
+
+/// Fields by which the query can be sorted
+#[derive(Debug, Default, Serialize, Deserialize)]
+pub enum SortBy {
+    /// Sort by key.
+    Key,
+    /// Sort by author.
+    #[default]
+    Author,
```
dignifiedquire, line 349: this is author and then key, right?

Frando:
> I think you need to rebase 😅 lots of conflicts

yes will do, started hacking away on #1701

iroh-sync/src/store/fs.rs
```diff
@@ -59,10 +64,22 @@ const NAMESPACES_TABLE: TableDefinition<&[u8; 32], &[u8; 32]> =
...
 type RecordsValue<'a> = (u64, &'a [u8; 64], &'a [u8; 64], u64, &'a [u8; 32]);
+
 type RecordsRange<'a> = TableRange<'a, RecordsId<'static>, RecordsValue<'static>>;
 type RecordsTable<'a> = ReadOnlyTable<'a, RecordsId<'static>, RecordsValue<'static>>;
-type DbResult<T> = Result<T, StorageError>;
+type RecordsReader<'a> = TableRangeReader<'a, RecordsId<'static>, RecordsValue<'static>>;
+
+// Records by key
+// Key: (NamespaceId, Key, AuthorId)
+// Value: same as in RECORDS_TABLE
```
dignifiedquire, line 77: should we be duplicating the full value here? seems like we could should store a reference to the records table (not sure how to and if that is actually better..)

iroh-sync/src/store.rs
Frando: Yup, renamed for clarity in the latest commit.

iroh-sync/src/store.rs
Frando: done

iroh-sync/src/store/fs.rs
Frando: Yeah, not sure myself. It's one extra lookup vs. 156 bytes (value size) per entry.

In the most recent commit, I changed it to not duplicate the value but instead do the extra lookup.

iroh-sync/src/store/fs/migrations.rs
```diff
@@ -0,0 +1,58 @@
...
+                    *e = (timestamp, key.to_vec());
+                }
+            })
+            .or_insert_with(|| (timestamp, key.to_vec()));
+    }
+    let len = heads.len();
+    for ((namespace, author), (timestamp, key)) in heads {
+        latest_table.insert((&namespace, &author), (timestamp, key.as_slice()))?;
+    }
+    info!("Migration finished (inserted {len} entries)");
```
dignifiedquire, line 36: maybe print the migration number?

iroh-sync/src/store/fs.rs
```diff
@@ -148,11 +112,15 @@ impl Store {
...
             let mut latest_table = write_tx.open_table(LATEST_TABLE)?;
             let _table = write_tx.open_multimap_table(NAMESPACE_PEERS_TABLE)?;
-            let _table = write_tx.open_table(RECORDS_BY_KEY_TABLE)?;
+            let mut by_key_table = write_tx.open_table(RECORDS_BY_KEY_TABLE)?;
 
             // migration 001: populate latest table if it was empty before
             if latest_table.is_empty()? && !records_table.is_empty()? {
-                migration_001_populate_latest_table(&records_table, &mut latest_table)?;
+                migrations::migration_001_populate_latest_table(&records_table, &mut latest_table)?;
+            }
```
dignifiedquire, line 129: we should be committing the transactions between individual migrations

iroh-sync/src/store/fs/bounds.rs
```diff
@@ -0,0 +1,169 @@
+use std::ops::Bound;
+
+use bytes::Bytes;
+
+use crate::{store::KeyFilter, AuthorId, NamespaceId};
+
+use super::{RecordsByKeyId, RecordsByKeyIdOwned, RecordsId, RecordsIdOwned};
+
+/// Bounds on the recors table.
```
dignifiedquire: typo

iroh-sync/src/store/fs/bounds.rs
```diff
@@ -0,0 +1,169 @@
+use std::ops::Bound;
+
```
dignifiedquire, line 2: all these bounds are quite important, I wonder if we could write some proptests to make sure they behave as expected

iroh-sync/src/store.rs
```diff
@@ -180,9 +180,10 @@ pub trait Store: std::fmt::Debug + Clone + Send + Sync + 'static {
 #[derive(Debug, Default)]
 pub struct QueryBuilder<K> {
     kind: K,
-    filter_author: AuthorMatcher,
-    filter_key: KeyMatcher,
-    limit_offset: LimitOffset,
+    filter_author: AuthorFilter,
```
dignifiedquire, line 183: any specific reasoning for Matcher -> Filter?

iroh-sync/src/sync.rs
```diff
@@ -1943,56 +1943,167 @@ mod tests {
...
+        );
+
+        assert_eq!(
+            query(
+                &store,
+                namespace,
+                Query::all()
+                    .sort_by(SortBy::KeyAuthor, SortDirection::Asc)
+                    .limit(2)
+                    .offset(1)
```
dignifiedquire, line 2081: I think there is no test with key_prefix mixed with offset/limit, that would be good to have

iroh-sync/src/store.rs
```diff
@@ -172,35 +176,250 @@ pub trait Store: std::fmt::Debug + Clone + Send + Sync + 'static {
...
-    Key(Vec<u8>),
-    /// Filter for key prefix
-    Prefix(Vec<u8>),
-    /// Filter by author
-    Author(AuthorId),
-    /// Filter by key prefix and author
-    AuthorAndPrefix(AuthorId, Vec<u8>),
+/// A query builder for document queries.
+#[derive(Debug, Default)]
+pub struct QueryBuilder<K> {
```
dignifiedquire, line 181: we should move the whole query stuff into its own file

iroh-sync/src/sync.rs
```diff
@@ -838,6 +838,11 @@ impl RecordIdentifier {
         &self.0
     }
 
+    /// Get this [`RecordIdentifier`] as [Bytes].
+    pub fn to_bytes(&self) -> Bytes {
```
dignifiedquire, line 842: should be `as_bytes` following the convention I believe

iroh-sync/src/sync.rs
```diff
@@ -1903,14 +1916,214 @@ mod tests {
...
+        replica.hash_and_insert("hi/moon", &a2, "a1")?;
+        replica.hash_and_insert("hi", &a3, "a3")?;
+
+        let expect = |rows: &[(&str, &Author)]| {
+            rows.into_iter()
+                .map(|(key, author)| (key.to_string(), author.id()))
+                .collect::<Vec<_>>()
+        };
+
+        assert_eq!(
```
dignifiedquire, line 1964: could write the below a little more readable like this

```rust

struct Test<'a> {
  query: Query,
  res: Vec<(&'static str, &'a AuthorId)>,
}

let tests = [
  Test {  
     query: Query::single_latest_per_key(), 
     res: vec![("hi", &a3), ("hi/moon", &a2), ("hi/world", &a1)],
  },
  // ... 
];

for test in tests {
  assert_eq!(
    query(&store, namespace, test.query),
    expect(&test.res),
    "query: {:#?}", test.query
  );
}
```

iroh-sync/src/store/fs/migrations.rs
Frando: Fixed in latest commit

iroh-sync/src/store/fs.rs
Frando: Fixed in latest commit

iroh-sync/src/store/fs/bounds.rs
Frando: Fixed

iroh-sync/src/store.rs
Frando: No special reason. Can revert. I think I renamed some variables to `filter` and then thought it to match better with `Filter`. But no preference on my side I think.

iroh-sync/src/store/fs/migrations.rs
```diff
@@ -1,17 +1,49 @@
...
+    Execute(usize),
+}
 
 /// migration 001: populate the latest table (which did not exist before)
-pub fn migration_001_populate_latest_table(
-    records_table: &Table<RecordsId<'static>, RecordsValue<'static>>,
-    latest_table: &mut Table<LatestKey<'static>, LatestValue<'static>>,
-) -> Result<()> {
-    info!("Starting migration: 001_populate_latest_table");
+pub fn migration_001_populate_latest_table(tx: &WriteTransaction) -> Result<MigrateOutcome> {
```
dignifiedquire, line 40: no need for `pub`?

iroh-sync/src/sync.rs
Frando: `as_bytes` already exists and returns `&[u8]`. This `to_bytes` returns [`Bytes`] to skip the extra allocation when recreating a `Bytes`. Open for better ideas on naming, but `as_bytes` is taken..

iroh-sync/src/sync.rs
Frando: Came up with something similar independently when writing more tests a minute ago ;) but I think I like your's even better, will switch

iroh-sync/src/store.rs
dignifiedquire: I think filter is fine, was just curious

iroh-sync/src/store.rs
Frando: Done

iroh-sync/src/sync.rs
dignifiedquire: could do 
- `impl AsRef<[u8]>` for the `&[u8]` and
- `fn as_bytes(&self) -> Bytes`

iroh-sync/src/store/fs/migrations.rs
Frando: Done

iroh-sync/src/sync.rs
dignifiedquire: :)

iroh-sync/src/store.rs
dignifiedquire: I think you only moved the query stuff in the `fs` piece, not here

iroh-sync/src/sync.rs
Frando: did something inbetween ;)

iroh-sync/src/sync.rs
Frando: sounds good, changed to that

iroh-sync/src/sync.rs
Frando: Added a test for that

iroh-sync/src/store/fs/bounds.rs
Frando: One a first thought I can't find a way to write meaningful proptests here. Do you have an idea how they should look or what they should test for?

dignifiedquire: I don't 😓 mostly thinking out loud

Frando: I added some manual tests for the edgecases.

Frando:
I merged main and solved the conflicts from #1770.

Also added `sort` and `desc` options to `doc keys` console command.

This is now ready to go from my side.

dignifiedquire (approved):

PR #1770, opened 2023-11-04
feat(iroh-sync): read only replicas

### LINKED ISSUE #1767: Read-only replica sharing
currently `doc share read` gives an error, because it's not implemented. We should, you know, implement that.

divagant-martian:
## Description

Closes #1767 
Updates sync, the sync db, commands and tickets to enable read only replicas

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-sync/src/actor.rs
+++ b/iroh-sync/src/actor.rs
@@ -15,7 +15,7 @@ use tracing::{debug, error, error_span, trace, warn};
 
 use crate::{
     ranger::Message,
-    store::{self, GetFilter},
+    store::{self, GetFilter, ImportNamespaceOutcome},
     Author, AuthorHeads, AuthorId, Capability, CapabilityKind, ContentStatus,
     ContentStatusCallback, Event, NamespaceId, NamespaceSecret, PeerIdBytes, Replica, SignedEntry,
     SyncOutcome,
@@ -485,13 +485,16 @@ impl<S: store::Store> Actor<S> {
                 let id = author.id();
                 send_reply(reply, self.store.import_author(author).map(|_| id))
             }
-            Action::ImportNamespace {
-                capability: namespace,
-                reply,
-            } => {
-                let id = namespace.id();
-                send_reply(reply, self.store.import_namespace(namespace).map(|_| id))
-            }
+            Action::ImportNamespace { capability, reply } => send_reply_with(reply, self, |this| {
+                let id = capability.id();
+                let outcome = this.store.import_namespace(capability.clone())?;
+                if let ImportNamespaceOutcome::Upgraded = outcome {
+                    if let Ok(replica) = this.states.replica(&id) {
+                        replica.merge_capability(capability)?;
+                    }
+                }
+                Ok(id)
+            }),
             Action::ListAuthors { reply } => iter_to_channel(
                 reply,
                 self.store
--- a/iroh-sync/src/actor.rs
+++ b/iroh-sync/src/actor.rs
@@ -603,7 +603,10 @@ impl<S: store::Store> Actor<S> {
                 this.store.remove_replica(&namespace)
             }),
             ReplicaAction::ExportSecretKey { reply } => {
-                let res = self.states.replica(&namespace).map(|r| r.secret_key());
+                let res = self
+                    .states
+                    .replica(&namespace)
+                    .and_then(|r| Ok(r.secret_key()?.clone()));
                 send_reply(reply, res)
             }
             ReplicaAction::GetState { reply } => send_reply_with(reply, self, move |this| {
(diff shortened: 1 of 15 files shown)


iroh-sync/src/actor.rs
```diff
@@ -481,7 +485,10 @@ impl<S: store::Store> Actor<S> {
                 let id = author.id();
                 send_reply(reply, self.store.import_author(author).map(|_| id))
             }
-            Action::ImportNamespace { namespace, reply } => {
+            Action::ImportNamespace {
+                capability: namespace,
```
Frando, line 489: maybe rename to `capability` for clarity

iroh-sync/src/store/fs.rs
```diff
@@ -48,9 +48,16 @@ const AUTHORS_TABLE: TableDefinition<&[u8; 32], &[u8; 32]> = TableDefinition::ne
...
 // Key: [u8; 32] # NamespaceId
 // Value: #[u8; 32] # Namespace
-const NAMESPACES_TABLE: TableDefinition<&[u8; 32], &[u8; 32]> =
+const NAMESPACES_TABLE_V1: TableDefinition<&[u8; 32], &[u8; 32]> =
     TableDefinition::new("namespaces-1");
 
+// Namespaces
+// Table
+// Key: [u8; 32] # NamespaceId
+// Value: #[u8; 32] # Namespace
```
Frando, line 57:
```suggestion
// Value: (u8, [u8; 32]) # CapabilityKind, NamespaceSecret
```

iroh-sync/src/store/memory.rs
```diff
@@ -108,8 +107,8 @@ impl super::Store for Store {
             .into_iter())
     }
 
-    fn import_namespace(&self, namespace: Namespace) -> Result<()> {
-        self.namespaces.write().insert(namespace.id(), namespace);
+    fn import_namespace(&self, capability: Capability) -> Result<()> {
+        self.namespaces.write().insert(capability.id(), capability);
```
Frando, line 111: I think this would delete the secret key when importing a read capability for an existing namespace, which is  likely not the desired outcome.

iroh-sync/src/store/fs.rs
```diff
@@ -149,11 +186,13 @@ impl Store {
...
     /// Stores a new namespace
-    fn insert_namespace(&self, namespace: Namespace) -> Result<()> {
+    fn insert_namespace(&self, capability: Capability) -> Result<()> {
         let write_tx = self.db.begin_write()?;
         {
             let mut namespace_table = write_tx.open_table(NAMESPACES_TABLE)?;
-            namespace_table.insert(namespace.id().as_bytes(), &namespace.to_bytes())?;
+            let id = capability.id().to_bytes();
+            let (kind, bytes) = capability.raw();
+            namespace_table.insert(&id, (kind, &bytes))?;
```
Frando, line 195: This would delete a namespace secret when importing a write capability for an existing read-only namespace, which is likely not the desired outcome.

Frando (changes requested):
Looks good! 

One question though:
If reading the current code correctly, importing a read capability (e.g. from a ticket) for an existing, writable namespace would delete a namespace's secret key. This is not the desired outcome, I'd say. What should happen IMO:
* Import write capability for existing namespace with read capability: Upgrade to write capability
* Import read capability for existing namespace with write capability: No change (but also no error)

Would be great to change this and add a test.

Frando:
I added code to merge capabilities. When importing a write capability for a namespace for which you have a read-only capability currently, it will now merge the capabilities in the store. however, if an actual Replica instance is opened in the sync actor, it will not yet update the capability to write, which is a problem. How should we solve this? Currently, the store does not keep a copy of the replica (it is non-cloneable, this was changed in the actor pr).

What we could do is that the actor would both do store.import_namespace and replica.update_capability or so. wdyt?

iroh-sync/src/store/fs.rs
Frando: fixed

iroh-sync/src/sync.rs
```diff
@@ -1928,6 +1952,38 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_replica_capability() -> Result<()> {
+        let mut rng = rand_chacha::ChaCha12Rng::seed_from_u64(1);
+        let store = store::memory::Store::default();
```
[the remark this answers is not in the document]
Frando: fixed

Frando:
I pushed another commit that takes care of upgrading capabilities of open replicas.

iroh-sync/src/store/memory.rs
Frando: Fixed

iroh-sync/src/store/fs.rs
Frando: Fixed

iroh-sync/src/actor.rs
Frando: Fixed

dignifiedquire (approved):

Frando (approved):

PR #1773, opened 2023-11-06
fix(iroh-sync): fix panic in send

rklaehn:
## Description

this was referring to entries by index, which will fail if two items are removed in one iteration.

We are now iterating in reverse order, so removal won't shift the remaining elements.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


dignifiedquire (approved):
maybe add a comment why `.rev()`?

dignifiedquire (approved):

PR #1775, opened 2023-11-07
fix(iroh-sync): prevent panic in namespace migration

Frando:
## Description

The migration as commited with #1770 panics when starting iroh with an existing data dir. The reason is that redb panics when deleting table in a transaction  where they were opened as well. See: https://github.com/cberner/redb/issues/715

This fixes this by moving the table deletion to a separate transaction.

The limitation and potential panic in redb was fixed in https://github.com/cberner/redb/pull/716 so once we upgrade to the next (still unreleased) version of redb, this can be removed again.

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh-sync/src/store/fs/migrations.rs
+++ b/iroh-sync/src/store/fs/migrations.rs
@@ -99,6 +99,14 @@ fn migration_002_namespaces_populate_v2(tx: &WriteTransaction) -> Result<Migrate
 }
 
 /// Delete the v1 namespaces table.
+///
+/// This should be part of [`migration_002_namespaces_populate_v2`] but due to a limitation in
+/// [`redb`] up to v1.3.0 a table cannot be deleted in a transaction that also opens this table.
+/// Therefore the table deletion has to be in a separate transaction.
+///
+/// This limitation was removed in https://github.com/cberner/redb/pull/716 so this can be merged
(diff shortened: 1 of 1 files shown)


dignifiedquire (approved):

iroh-sync/src/store/fs/migrations.rs
```diff
@@ -95,12 +95,23 @@ fn migration_002_namespaces_v2(tx: &WriteTransaction) -> Result<MigrateOutcome>
         namespaces_v2.insert(&id, (raw_kind, &raw_bytes))?;
         entries += 1;
     }
-    tx.delete_table(NAMESPACES_TABLE_V1)?;
     Ok(MigrateOutcome::Execute(entries))
 }
 
+/// Delete the v1 namespaces table.
+fn migration_003_namespaces_delete_v1(tx: &WriteTransaction) -> Result<MigrateOutcome> {
```
dignifiedquire, line 102: can you add a comment why this is its own migration? this is quite surprising

dignifiedquire (approved):
lgtm, modulu my one comment

iroh-sync/src/store/fs/migrations.rs
Frando: :+1: added a commit with a comment

PR #1778, opened 2023-11-08
fix(iroh-bytes): ensure to flush file to disk

dignifiedquire:
The content was not always on disk before emitting progress events and returning, resulting in broken progress updates and unfinished writes.


dignifiedquire:
tested using https://github.com/n0-computer/iroh-ffi/pull/63 which was flaky/not passing before

PR #1780, opened 2023-11-08
refactor: common base library

Frando:
## Description

This adds a new crate `iroh_base`. The following types were moved into the new crate:

* `Hash`, `HashAndFormat` and `BlobFormat` from `iroh_bytes::util`
* `base32` encoding/decoding util functions. They lived at various places before.
* `RpcError` and `RpcResult` from `iroh_bytes::util`

## Notes & open questions

* Naming: Shall it be `iroh_common` or `iroh_base`? I do not mind. If anyhow has a preference, please raise. I can change easily.
* Anything else that should be moved over? Anything that I moved that should *not* be moved?

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


dignifiedquire:
please make sure to reexport the types from `iroh/src/lib.rs` so only `iroh` needs to be a dependency

Frando:
> please make sure to reexport the types from `iroh/src/lib.rs` so only `iroh` needs to be a dependency

Added a `pub use iroh_base as base`. Or should we export the individual types directly?

dignifiedquire:
> Added a `pub use iroh_base as base`. Or should we export the individual types directly?

I think individually makes probably more sense, to avoid moving types in there/out there a breaking change down the line

dignifiedquire:
I am generally fine with the name, I don't love it, but I don't have any alternatives 😅

Frando:
> I am generally fine with the name, I don't love it, but I don't have any alternatives 😅

Well, `iroh-common` was the most discussed contender so far. I don't really have a preference between the two I think.

Frando:
> > Added a `pub use iroh_base as base`. Or should we export the individual types directly?
> 
> I think individually makes probably more sense, to avoid moving types in there/out there a breaking change down the line

I realized that the `iroh_base::hash::*` types are already reexported in `iroh_bytes`, which is reexported from `iroh`, so they are available as `iroh::bytes::{Hash, HashAndFormat, BlobFormat}`, which is fine I'd say?

The types from `iroh_bytes::rpc` are re-exported as `iroh::rpc_protocol::{RpcResult, RpcError}`, which is also fine.

Leaves the `base32` which I reexported from `iroh` directly now.

Frando:
`iroh_base` currently has a dependency on `bao_tree`, but only for the `Hash` type, which is a reexport from `iroh_blake3`. Should we add a dep to `iroh_blake3` instead? Downside is that the versions would have to be kept in sync vigilantly.

dignifiedquire:
not sure, @rklaehn thoughts?

Frando:
`iroh_sync` now fails to compile because rustc claims that `bytes::Bytes` doesn't implement `serde::Serialize` and `Deserialize`. I have no idea why that is the case. The `serde` feature is enabled for `bytes` in `iroh-sync/Cargo.toml`.

Anyone has an idea!?

iroh-sync/Cargo.toml
```diff
@@ -31,7 +31,7 @@ rand_core = "0.6.4"
 serde = { version = "1.0.164", features = ["derive"] }
 strum = { version = "0.25", features = ["derive"] }
 url = "2.4"
-bytes = { version = "1.4", feature = ["serde"] }
+bytes = { version = "1.4", features = ["serde"] }
```
Frando, line 34: :laughing:

dignifiedquire (approved):

PR #1783, opened 2023-11-08
feat(iroh): store rpc port in iroh data dir

### LINKED ISSUE #1281: Prevent multiple iroh daemon processes from using the same IROH_DATA_DIR
It just causes a lot of weirdness to have 2 processes with the same private key.

dignifiedquire:
- prevents iroh from being started in the same iroh data dir
- retrieves the rpc port from the rpc.lock file
- removes  the `--rpc-port` argument from all commands other than `start`

This also changes the way in put handling is done on `start`, as I discovered some issues with the current state of things.

The way it works now is
```sh
# no adding of content
> iroh start
# adding from stdin
> iroh start --source STDIN
# adding from file
> iroh start --source ./path/to/file
```

Closes #1281

### DIFF
--- a/iroh/src/util/path.rs
+++ b/iroh/src/util/path.rs
@@ -28,7 +28,7 @@ pub enum IrohPaths {
     /// Path to store known peer data.
     PeerData,
     #[strum(serialize = "rpc.lock")]
-    /// Path to RPC lock file, containint the RPC port if running.
+    /// Path to RPC lock file, containing the RPC port if running.
     RpcLock,
 }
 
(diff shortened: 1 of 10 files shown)


iroh/src/util/path.rs
```diff
@@ -27,6 +27,9 @@ pub enum IrohPaths {
     #[strum(serialize = "peers.postcard")]
     /// Path to store known peer data.
     PeerData,
+    #[strum(serialize = "rpc.lock")]
+    /// Path to RPC lock file, containint the RPC port if running.
```
ramfox, line 31: containint -> containing

ramfox (approved):
dogfooded and it works as intended!

Arqu (approved):

PR #1793, opened 2023-11-10
feat: add `Doc::import_file` and `Doc::export_file`

### LINKED ISSUE #1781: Import & export files and directories directly into docs
## High-level feature

We want to support exporting and importing files and directories into documents. This should be directly from the `Doc` client, e.g. `doc.import_from_path()` and `doc.export_to_path()`.

### Exporting from docs

What we miss (and will want anyway) is `blobs.export_to_path` in client, rpc and node - this should be straightforward to add though. Exporting a "path tree" from a document can then simply be done on the client level in a `export_to_path` function that runs a query and exports each entry's content hash.

### Importing into docs

We already support importing files and directories recursively directly into the `iroh_bytes` blobs store through `client.blobs.add_from_path()`.  

We also already support setting document entries directly to hashes. However this is not enough, because all blobs added through `client.blobs` are tagged (either with a manual tag or an auto tag) and this is not wanted for blobs added to documents, as it will prevent garbage collection to delete blobs once their document entry is deleted or overwritten.

I see two ways to solve this:

**Exposing tempt tags to the client** - `blobs.add_from_path` iterates over a file tree and imports each file and gives them a temporary tag. At the end the collection itself is tagged and the temporary tags are deleted. What we could do is to add an option to omit the final step and instead keep the temp tags and return the tags over the RPC. The client could then return a stream of temp tags to drop once the hashes are added to the document. The node should drop the temp tags by itself maybe after a timeout if the RPC client did not return them (e.g. because it died) so that those blobs would be deleted and not lie around dangling (unreferenced).

The upside is that this mechanism might be useful for other usecases as well - I can imagine people wanting to add blobs incrementally over some time, and then add something that references them once the operation finishes (doc or collection) - not necessarily only blobs in a single directory tree. The downside is that it will likely involve some work to keep track of the temp tags. @rklaehn can likely tell more. 

**Add `DocImportFromPath` to the RPC** - this would be the alternative, basically copy the BlobAddFromPath rpc methods to to `DocAddFromPath` and implement likewise. Then the question of temp tags can be handled in the node, in the same way as we already do for blob trees.

### The magic last byte

Because we employ Willow's model of prefix deletion, path-like usecases of docs need to terminate leaf (non-directory) entries with a character that is otherwise invalid in the path, so that importing `somef` does not delete `somefile`. A simple way might be to just add a nul byte to the path. However, apps might have different needs (e.g. use these suffix for more info - append the mime type? add some id?). So I was thinking that maybe we should support passing a callback that converts from file path to key for import, and from key to file path for export.

### API

... could e.g. look like this

```rust
impl DocClient {
    fn import_from_path(&self, author: AuthorId, path: Path, path_to_key: impl Fn(Path) -> Bytes) -> impl Stream<Item = Progress>{}
    fn export_from_prefix(&self, query: Query, key_to_path:  impl Fn(&[u8)) -> PathBuf)  -> impl Stream<Item = Progress> {}
}
``

### PR DESCRIPTION
**Pull request by ramfox**
## Description

closes #1781 

### canonicalize_path:

There is now a `canonicalized_path_to_string` method that takes a path (that it expects to be canonicalized already) and turns it into a string.

What was previously `canonicalize_path` is now renamed to `relative_canonicalized_path_to_string`, and calls `canonicalized_path_to_string`, with the `must_be_relative` param set to `true`

### DIFF
--- a/iroh/src/client.rs
+++ b/iroh/src/client.rs
@@ -4,19 +4,20 @@
 
 use std::collections::HashMap;
 use std::io::{self, Cursor};
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::pin::Pin;
 use std::result::Result as StdResult;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 use std::task::{Context, Poll};
 
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, Context as AnyhowContext, Result};
 use bytes::Bytes;
 use futures::stream::BoxStream;
 use futures::{SinkExt, Stream, StreamExt, TryStreamExt};
 use iroh_bytes::provider::AddProgress;
 use iroh_bytes::store::ValidateProgress;
+// use iroh_bytes::util::progress::FlumeProgressSender;
 use iroh_bytes::util::runtime;
 use iroh_bytes::Hash;
 use iroh_bytes::{BlobFormat, Tag};
@@ -35,12 +36,13 @@ use crate::rpc_protocol::{
     BlobListCollectionsResponse, BlobListIncompleteRequest, BlobListIncompleteResponse,
     BlobListRequest, BlobListResponse, BlobReadRequest, BlobReadResponse, BlobValidateRequest,
     CounterStats, DeleteTagRequest, DocCloseRequest, DocCreateRequest, DocDelRequest,
-    DocDelResponse, DocDropRequest, DocGetExactRequest, DocGetManyRequest, DocImportRequest,
-    DocLeaveRequest, DocListRequest, DocOpenRequest, DocSetHashRequest, DocSetRequest,
-    DocShareRequest, DocStartSyncRequest, DocStatusRequest, DocSubscribeRequest, DocTicket,
-    DownloadProgress, ListTagsRequest, ListTagsResponse, NodeConnectionInfoRequest,
-    NodeConnectionInfoResponse, NodeConnectionsRequest, NodeShutdownRequest, NodeStatsRequest,
-    NodeStatusRequest, NodeStatusResponse, ProviderService, SetTagOption, ShareMode, WrapOption,
+    DocDelResponse, DocDropRequest, DocGetExactRequest, DocGetManyRequest, DocImportFileRequest,
+    DocImportProgress, DocImportRequest, DocLeaveRequest, DocListRequest, DocOpenRequest,
+    DocSetHashRequest, DocSetRequest, DocShareRequest, DocStartSyncRequest, DocStatusRequest,
+    DocSubscribeRequest, DocTicket, DownloadProgress, ListTagsRequest, ListTagsResponse,
+    NodeConnectionInfoRequest, NodeConnectionInfoResponse, NodeConnectionsRequest,
+    NodeShutdownRequest, NodeStatsRequest, NodeStatusRequest, NodeStatusResponse, ProviderService,
+    SetTagOption, ShareMode, WrapOption,
 };
 use crate::sync_engine::LiveEvent;
 
@@ -624,6 +626,47 @@ where
         Ok(())
     }
 
+    /// Add an entry from a file path
+    pub async fn import_file(
+        &self,
+        author: AuthorId,
+        path: impl AsRef<Path>,
+        path_to_key: impl Fn(PathBuf) -> Bytes + Send + Sync + 'static,
+    ) -> Result<DocImportFileProgress> {
+        self.ensure_open()?;
+        let path = path.as_ref().to_path_buf();
+        let key = path_to_key(path.clone());
+        let stream = self
+            .0
+            .rpc
+            .server_streaming(DocImportFileRequest {
+                doc_id: self.id(),
+                author_id: author,
+                path,
+                key,
+            })
+            .await?;
+        Ok(DocImportFileProgress::new(stream))
+    }
+
+    /// Export an entry as a file to a given path.
+    pub async fn export_file(
+        &self,
+        query: impl Into<Query>,
+        key_to_path: impl Fn(&[u8]) -> PathBuf + Send + Sync + 'static,
+    ) -> Result<()> {
+        self.ensure_open()?;
+        let entry = self.get_one(query).await?.context("Entry not found")?;
+        let path = key_to_path(entry.key());
+        if let Some(dir) = path.parent() {
+            tokio::fs::create_dir_all(dir).await?;
+        }
+        let mut file = tokio::fs::File::create(path).await?;
+        let mut reader = self.read(&entry).await?;
+        tokio::io::copy(&mut reader, &mut file).await?;
+        Ok(())
+    }
+
     /// Read the content of an [`Entry`] as a streaming [`BlobReader`].
     pub async fn read(&self, entry: &Entry) -> Result<BlobReader> {
         self.ensure_open()?;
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: iroh/src/client.rs [hunk 1]
```diff
@@ -624,6 +626,47 @@ where
         Ok(())
     }
 
+    /// Add an entry from a file path
+    pub async fn import_file(
+        &self,
+        author: AuthorId,
+        path: impl AsRef<Path>,
+        path_to_key: impl Fn(PathBuf) -> Bytes + Send + Sync + 'static,
```
dignifiedquire, line 634: - are these bounds really needed, given that we immediately execute it?
- alternatively how about making the signature 

```rust
pub async fn import_file(
        &self,
        author: AuthorId,
        key: Bytes,
        path: impl AsRef<Path>,
)  -> Result<DocImportFileProgress> { }
```

iroh/src/client.rs
```diff
@@ -624,6 +626,47 @@ where
...
+    ) -> Result<()> {
+        self.ensure_open()?;
+        let entry = self.get_one(query).await?.context("Entry not found")?;
+        let path = key_to_path(entry.key());
+        if let Some(dir) = path.parent() {
+            tokio::fs::create_dir_all(dir).await?;
+        }
+        let mut file = tokio::fs::File::create(path).await?;
+        let mut reader = self.read(&entry).await?;
+        tokio::io::copy(&mut reader, &mut file).await?;
```
dignifiedquire, line 666: file needs flushing (to be safe..)

iroh/src/util/fs.rs
```diff
@@ -225,10 +226,48 @@ fn path_content_info0(path: impl AsRef<Path>) -> anyhow::Result<PathContent> {
...
+/// Helper function that translates a key that was derived from a path on a filesystem back to that
+/// path.
+pub fn key_to_path(key: impl AsRef<[u8]>) -> anyhow::Result<PathBuf> {
+    let key = std::str::from_utf8(key.as_ref())?;
+    if key.is_empty() {
+        return Ok(PathBuf::new());
+    }
+    // assumes the last byte is the null byte that was appended using
+    // the [`path_to_key`] helper function
+    Ok(PathBuf::from(&key[..key.len() - 1]))
```
dignifiedquire, line 250: this should verify that it was a nul byte

iroh/src/util/fs.rs
```diff
@@ -225,10 +226,48 @@ fn path_content_info0(path: impl AsRef<Path>) -> anyhow::Result<PathContent> {
...
+        .ok_or(anyhow::anyhow!("could not convert path to bytes"))?;
+    let mut bytes = BytesMut::from(buf);
+    bytes.put(&b"\0"[..]);
+    Ok(bytes.freeze())
+}
+
+/// Helper function that translates a key that was derived from a path on a filesystem back to that
+/// path.
+pub fn key_to_path(key: impl AsRef<[u8]>) -> anyhow::Result<PathBuf> {
+    let key = std::str::from_utf8(key.as_ref())?;
```
dignifiedquire, line 244: we shouldn't assume utf8 here, this should use OsString, and the nul byte should be removed before conversion, to avoid accidental issues

iroh/src/client.rs
```diff
@@ -751,6 +794,142 @@ where
...
 }
 
+/// Progress stream for doc import operations.
+#[derive(derive_more::Debug)]
+pub struct DocImportFileProgress {
+    #[debug(skip)]
+    stream: Pin<Box<dyn Stream<Item = Result<DocImportProgress>> + Send + Unpin + 'static>>,
+}
+
+impl DocImportFileProgress {
```
Frando, line 804: This duplication is what lead me to the thought of exposing temp tags to the client. So that the existing impl for blobs could be reused. But I see there's other not-so-easy issues with exposing temp tags, so well...

iroh/src/client.rs
```diff
@@ -624,6 +626,47 @@ where
...
+                author_id: author,
+                path,
+                key,
+            })
+            .await?;
+        Ok(DocImportFileProgress::new(stream))
+    }
+
+    /// Export an entry as a file to a given path.
+    pub async fn export_file(
```
Frando, line 653: We should think about potential consequences that `import_file` runs in the node and `export_file` in the client. E.g. they could potentially run under different unix system users or permissions, and it might be unexpected for `import_file` to work for a path but `export_file` not (if e.g. the user running the console has less permissions than the node).

Not saying this needs to change, but something to think about. Maybe it does make sense to move `export_file` to the node as well, could also help with perf at least for big files to not have to copy them over the rpc conn.

iroh/src/node.rs
```diff
@@ -954,6 +955,121 @@ impl<D: BaoStore> RpcHandler<D> {
...
+        } = msg;
+        // Check that the path is absolute and exists.
+        anyhow::ensure!(root.is_absolute(), "path must be absolute");
+        anyhow::ensure!(
+            root.exists(),
+            "trying to add missing path: {}",
+            root.display()
+        );
+
+        // todo: do we assume copy?
```
Frando, line 1020: I think we should expose the import mode to the rpc & client. Otherwise docs are second-class citizens with reagard to dealing with "big data" compared to blobs imported via Blobs rpc.

iroh/src/node.rs
```diff
@@ -954,6 +955,121 @@ impl<D: BaoStore> RpcHandler<D> {
...
+        //     SetTagOption::Named(tag) => {
+        //         self.inner
+        //             .db
+        //             .set_tag(tag.clone(), Some(*hash_and_format))
+        //             .await?;
+        //         tag
+        //     }
+        //     SetTagOption::Auto => self.inner.db.create_tag(*hash_and_format).await?,
+        // };
+        let tag = self.inner.db.create_tag(*hash_and_format).await?;
```
Frando, line 1046: Ah, I think this we should *not* do :) This creates a permanent tag, so the blob cannot be garbage collected unless the tag is deleted. Adding the hash to the document already protects the blob from being garbage collected (until it is deleted or overwritten). So this line should be removed and instead the temp tag simply be dropped after `doc_set_hash()` completed.

@rklaehn is dropping temp tags enough or does it has to be deleted somehow to free the blob from tag protection?

iroh/src/util/fs.rs
```diff
@@ -225,10 +226,48 @@ fn path_content_info0(path: impl AsRef<Path>) -> anyhow::Result<PathContent> {
...
+        .map(|p| p.as_bytes())
+        .ok_or(anyhow::anyhow!("could not convert path to bytes"))?;
+    let mut bytes = BytesMut::from(buf);
+    bytes.put(&b"\0"[..]);
+    Ok(bytes.freeze())
+}
+
+/// Helper function that translates a key that was derived from a path on a filesystem back to that
+/// path.
+pub fn key_to_path(key: impl AsRef<[u8]>) -> anyhow::Result<PathBuf> {
```
Frando, line 243: For both these functions users will usually want to strip/apply a path prefix because they likely wouldn't want or should treat doc keys as *absolute* paths. So the utility of these functions in this way is questionable to me now that I see them. Or there should be a way to pass in a path prefix to add/strip I think.

iroh/src/client.rs
ramfox: fully agree.  this iteration was my "get it done and ask for help" version. If we change the params to remove the `query` and `key_to_path` method, in favor of `key` and `path` (as was recommended to me on my call this morning w/ dig and rudi), I can then use the rpc `server_streaming` patterns that we already have set up!

appreciate the thoughtful feedback

iroh/src/client.rs
ramfox: yea there is alot of duplication i don't know enough about all of the consequences of exposing temp tags right now, so I'm not terribly helpful here.

iroh/src/node.rs
ramfox: great, will add this in

iroh/src/node.rs
ramfox: thank you, this is exactly the advice i needed, because i really don't have a grasp on the consequences of tags right now! figured i'd do it wrong and have someone yell at me about it, lol.

iroh/src/util/fs.rs
ramfox: great, thank you. I wrote a version of that (adding and stripping prefixes) in the `iroh import` and `iroh export` code paths already, so we can now just move those to the util crate 🙌

iroh/src/node.rs
```diff
@@ -963,6 +965,147 @@ impl<D: BaoStore> RpcHandler<D> {
...
+        };
+
+        let (temp_tag, size) = self
+            .inner
+            .db
+            .import_file(root, import_mode, BlobFormat::Raw, import_progress)
+            .await?;
+
+        let hash_and_format = temp_tag.inner();
+        let HashAndFormat { hash, .. } = *hash_and_format;
```
ramfox, line 1040: @Frando @rklaehn Is this enough to handle tags correctly? Do I need to do more than just let the `temp_tag` drop?

iroh/src/util/fs.rs
```diff
@@ -225,10 +228,115 @@ fn path_content_info0(path: impl AsRef<Path>) -> anyhow::Result<PathContent> {
...
+            &key[prefix.len()..]
+        } else {
+            anyhow::bail!("key {:?} does not begin with prefix {:?}", key, prefix);
+        }
+    } else {
+        key
+    };
+
+    // convert to a path
+    let path = PathBuf::from(OsStr::from_bytes(key));
```
ramfox, line 263: okay, so this is obviously wrong because it only works for unix

I'm not sure the right thing to do for both the conversion btw path to bytes and bytes to path is, because I'm not sure I understand how to get from a &[u8] to &[u16] that is needed to for the windows version of OsString.

I also don't understand what the expectations are if we, for example, create a key from a path on one platform, but then convert that same key to a path on a different platform.

Can I get some advice? Is it worth this frustration? Is it easier/better to just fail if we don't have utf-8? Need some help!

iroh/src/node.rs
Frando: I think this is the case, yes, see the drop impl for  `TempTag` which informs the "liveness tracker" of the store that the tag is dropped.

rklaehn: Yes, I think this is correct. You are keeping the temp tag around until you store it in the document. As soon as it is in the document, it will be protected by the document.

If you want to make it more explicit, you can explicitly call drop on the temp tag after the add to the doc.

iroh/src/util/fs.rs
dignifiedquire: okay, this is all not very great, but for now we should simplify probably

the import steps should looks sth like this
- `canoncialize` the path (this removes things like `..` and symbolic links)
- iterate over each component with `.components()`
- for each component check that it is valid utf8 (yes this is a limitation for now, but utf16 or other crazy things are just too hard to deal with for now)
- escape any `/` with `\/`
- join all components into a `String` with `/` as the seperator

- on the out step replace `/` with the OS specific seperator
- unescape `\/` to `/`

rklaehn: Something roughly similar exists in `util::fs::canonicalize_path`. It does not escape / though but just assumes that they don't exist and fails otherwise. This is what is used to create the path metadata for collections.

ramfox: okay, adapted from `canonicalize_path`:

There is now a `canonicalized_path_to_string` method that takes a path (that it expects to be canonicalized already) and turns it into a string

`canonicalize_path` is renamed to `relative_canonicalized_path_to_string`, and calls `canonicalized_path_to_string`, with the `must_be_relative` param set to `true`

These are a mouthful, but they are specific. TBH, I didn't even look at `canonicalize_path` closely before because I thought the function was just a util function to "canonicalize a path", not a function that turned a canonicalized path into a string.

I opted for keeping the restriction that these helper functions expect a canonicalized paths already, because I didn't want to complicate them by having them touch the actual file system or have to potentially add `async` functionality to helper util functions.

iroh/src/util/fs.rs
```diff
@@ -338,8 +360,11 @@ mod tests {
     }
 
     #[test]
-    fn test_canonicalize_path() {
-        assert_eq!(super::canonicalize_path("foo/bar").unwrap(), "foo/bar");
+    fn test_relative_canonicalized_path_to_string() {
+        assert_eq!(
+            super::relative_canonicalized_path_to_string("foo/bar").unwrap(),
+            "foo/bar"
+        );
```
dignifiedquire, line 367: can you add some tests for crazy inputs to these methods please?

dignifiedquire, line 367: can you add some tests for crazy inputs to these methods please?

dignifiedquire (approved):

PR #1795, opened 2023-11-10
feat(iroh): allow full and short hash printing in the cli

dignifiedquire:
this allows to view the full hash of the value

### DIFF
--- a/iroh/src/commands/sync.rs
+++ b/iroh/src/commands/sync.rs
@@ -12,7 +12,7 @@ use colored::Colorize;
 use dialoguer::Confirm;
 use futures::{Stream, StreamExt, TryStreamExt};
 use indicatif::{HumanBytes, HumanDuration, MultiProgress, ProgressBar, ProgressStyle};
-use iroh_base::base32::fmt_short;
+use iroh_base::base32::{fmt, fmt_short};
 use tokio::io::AsyncReadExt;
 
 use iroh::{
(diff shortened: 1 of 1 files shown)


iroh/src/commands/sync.rs
```diff
@@ -40,6 +40,8 @@ pub enum DisplayContentMode {
     Content,
     /// Display the hash of the content.
     Hash,
+    /// Displays the full hash of the content.
+    HashFull,
```
b5, line 44: I think this should just change the behaviour of `Hash` instead of introducing a new mode. We could then adjust the description of `Auto` to "diplays a shortened content hash"

```rs
pub enum DisplayContentMode {
    /// Displays the content if small enough, otherwise it displays a shortened content hash.
    Auto,
    /// Display the content unconditionally.
    Content,
    /// Display the hash of the content.
    Hash,
}

dignifiedquire: `Hash` is currently used in listings, so that won't work unless we want to show the content or the full hash there

b5: ah I see. then I think we should invert the meaning of "hash", and introduce `ShortHash` to keep the current listing behaviour. I think my biggest sticking point is "hash should be hash"

dignifiedquire: done

b5 (approved):

PR #1799, opened 2023-11-10
refactor(iroh): restructure cli modules to match command structure

dignifiedquire:
Only moves code around


Arqu (approved):

ramfox (approved):
lol this is so much more logical 🙌

PR #1800, opened 2023-11-10
feat(cli): get node with rpc and stable node stats sort

Arqu:
## Description

Only remaining thing is how to catch an early abort to properly do `clear_rpc(...).await?;`

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


Arqu:
Refactor of cli get underway.

PR #1802, opened 2023-11-10
refactor(iroh): add `--start` option to CLI commands

dignifiedquire:
## Description

Changes to CLI commands:

* All RPC commands now have a `--start` option that starts a node, runs the command, and then shuts down the node
* `iroh get` is now `iroh blob get`, and needs `--start` flag to run standalone
* `iroh start SOURCE` is now `iroh start --add SOURCE`
* Improvements to the node startup code and error handling

## Notes and open questions

We could remove the `--add SOURCE` option from `iroh start` if there was a way to to keep running the node after starting it with a command. Could also be something like an `--exec` option. But let's leave that to a followup.


Arqu:
If this passes (which is expected) https://github.com/n0-computer/iroh/actions/runs/6889635559/job/18741024267
Just merge the netsim PR and then your PR.

dignifiedquire:
> If this passes (which is expected)

looks like it is not passing

Frando (approved):

PR #1810, opened 2023-11-15
fix: drop temp tag after doc insert

Frando:
## Description

When importing files into docs, the temp tag should only be dropped *after* the hash was inserted into the doc. Otherwise gc could potentially delete the blob if hitting in between the tag drop and the completion of the doc insert.


ramfox (approved):
great catch! thank you!!!!

PR #1811, opened 2023-11-15
fix: use `path_to_key` helper function in `iroh doc import`

ramfox:
The `std::util::fs::path_to_key` method will properly format paths and turn them into keys that we can use when working between the filesystem and a doc.

It will specifically do these three things:
- remove any `root` dir that you want it to remove
- add any `prefix` you provide
- append the null byte `\0` to the end of the key, to prevent any unexpected overwrites that may occur when you set entries on a document.


Frando (commented):
LGTM! Maybe also use `key_to_path` in export, and add a test?

Frando (approved):

ramfox:
> LGTM! Maybe also use key_to_path in export, and add a test?

I should definitely add a test.

As for using `key_to_path` in `iroh export`: right now the `--out` flag takes a path & we use that path to save, we don't use the given key at all. We don't currently do export for multiple files, but we should definitely use `key_to_path` when we end up implementing that.

PR #1812, opened 2023-11-16
refactor: improve code and args for CLI start

Frando:
## Description

Changes for #1802 

CLI command changes:

* `iroh start` takes `--add` not `--source`
* `iroh blob add` takes required  `SOURCE` arg and not optional `--source` opt

Internal changes:
* reduce code duplication around starting a node
* always start the metrics server (for both `iroh start` and `iroh CMD --start`)
* make sure to remove the rpc lock file if aborting with error
* cleanups

## Change checklist

- [x] Self-review.
- ~~[ ] Tests if relevant.~~

### DIFF
--- a/iroh/src/commands.rs
+++ b/iroh/src/commands.rs
@@ -1,18 +1,15 @@
+use std::path::PathBuf;
 use std::str::FromStr;
-use std::{net::SocketAddr, path::PathBuf};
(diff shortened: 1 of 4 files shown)


dignifiedquire (approved):
very nice

PR #1821, opened 2023-11-19
chore(deps): Update dependencies

flub:
## Description

This updates dependencies of iroh 0.4.1 release, most notably quinn fixes platform support for android without patches.

## Notes & open questions

- Not all dependencies are updated.  Only a few select ones which make things better
  for deltachat.

- Note that the base of this PR is `maint-0.4` which is a branch currently pointing
  at the 0.4.1 commit.

- This PR also bumps the version number already to 0.4.2, we'd have to publish a new
  release with this.

- This ignores a rust advisory, the way iroh uses the code was determined not to be
  problematic and adopting the new API is rather invasive and not suitable for a
  maintenance branch.  The main branch has updated this code.

- Netsim CI is disabled, provides not much value for a maintenance branch and it
  has been updated for newer versions of iroh.  Keeping it compatible with this
  old version is not worth it.

- Only stable rust is tested in CI now.  The proc-macro2 crate version used in
  Cargo.lock does not work with current nightly, which is fine.

- This disables the CI check on the Windows/GNU build.  There is a linker problem
  picking up an MSVC library.  Rather than debug this I think we care much more
  about MSVC.

## Change checklist

- [x] Self-review.
- [x] ~~Tests if relevant.~~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ on:
 env:
   RUST_BACKTRACE: 1
   RUSTFLAGS: -Dwarnings
-  MSRV: "1.63"
+  MSRV: "1.67"
 
 
 jobs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest]
-        rust: [nightly, stable]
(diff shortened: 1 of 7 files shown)


flub:
@dignifiedquire now CI is happy again.  Admittedly I disabled some of it...  Do you think it would be reasonable to publish 0.4.2?

dignifiedquire (approved):

PR #1829, opened 2023-11-23
refactor(iroh-bytes): remove explicitly passing the runtime to the flat store

rklaehn:
## Description

the flat store will now assume that there is an ambient tokio runtime and pick that up to do spawn_blocking in all async ops. I hate this, but this is how tokio wants to be used, so what can you do.

It makes creating a store much simpler.

## Notes & open questions

-

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


flub:
If you want to pass in the runtime in so that the caller can control this you could still make this optional by having a `with_rt` kind if constructor which allows one to pass in the runtime if they really want to have control over this.  Whether that's worth it really depends on what your goal is here I guess.

rklaehn:
> If you want to pass in the runtime in so that the caller can control this you could still make this optional by having a `with_rt` kind if constructor which allows one to pass in the runtime if they really want to have control over this. Whether that's worth it really depends on what your goal is here I guess.

The issue is that as soon as you use any tokio code, it is using just the static `tokio::task::spawn_blocking` anyway. So passing the runtime explicitly just provides you a false sense of security.

E.g. every single AsyncXXX trait impl for tokio::fs::File just uses tokio::task::spawn_blocking, so will pick up the ambient runtime. https://docs.rs/tokio/latest/src/tokio/fs/file.rs.html#606

So given the amount of mental overhead this causes it just does not seem worth it at this time.

flub:
> The issue is that as soon as you use any tokio code, it is using just the static `tokio::task::spawn_blocking` anyway. So passing the runtime explicitly just provides you a false sense of security.

ah, yes.  good point

> E.g. every single AsyncXXX trait impl for tokio::fs::File just uses tokio::task::spawn_blocking, so will pick up the ambient runtime. https://docs.rs/tokio/latest/src/tokio/fs/file.rs.html#606
> 
> So given the amount of mental overhead this causes it just does not seem worth it at this time.

I guess you could enter a different runtime every time you call in any other tokio function, but that's super painful and error prone so entirely unusually in practice i guess.

dignifiedquire (approved):

PR #1833, opened 2023-11-23
[WIP] feat: protocol handlers

Frando:
## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-gossip/examples/chat.rs
+++ b/iroh-gossip/examples/chat.rs
@@ -1,6 +1,6 @@
 use std::{collections::HashMap, fmt, str::FromStr, sync::Arc};
 
-use anyhow::{bail, Context};
+use anyhow::bail;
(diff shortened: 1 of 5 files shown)


Frando:
This does not work out well this way. Instead our current idea is to expose the endpoint events not as a single callback, but as an (overflowing, non-lagging) broadcast stream. This would very much simplify the composition of protocols, as users won't have to merge their event listeners into a single callback, which needs to be passed to `MagicEndpointBuilder` (thus before `MagicEndpoint` exists, leading to chicken-and-egg problems)

dignifiedquire:
https://github.com/n0-computer/iroh/pull/1847 should help to improve the startup story

Frando:
Closing this. Needs more thought and is not on the immediate agenda.

PR #1841, opened 2023-11-24
Move collection to iroh-bytes and add metadata header

rklaehn:
## Description

This changes the collection type a bit to make it more idiomatic rust. E.g. it now implements traits to use it like a... collection, like FromIterator. It also moves collection (back) to iroh-bytes so it can be used from tools like sendme without having to depend on all of iroh.

This collection format is intended to be a minimum viable metadata format that we can support for a long time - just a sequence of strings with a string for each blob in the HashSeq. It is useful for tools like sendme and for gateways, so I think this is valuable enough to have as a separate format to e.g. document snapshots.

## Notes & open questions

Should this be sequences of blobs? Weirdly, in this case I am OK with a sequence of strings. If somebody wants names to be arbitrary blobs they can always define another metadata format.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5928,18 +5928,18 @@ dependencies = [
 
 [[package]]
 name = "zerocopy"
-version = "0.7.25"
+version = "0.7.31"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


dignifiedquire (approved):

dignifiedquire:
please rebase before merge

PR #1845, opened 2023-11-26
feat(net): add `Magicsock::network_change`

dignifiedquire:
This allows notifying `iroh-net` when the network has likely changed.

Usually this is only needed on android, as on all other OSes this is automatically detected.

### DIFF
--- a/iroh-net/src/magic_endpoint.rs
+++ b/iroh-net/src/magic_endpoint.rs
@@ -558,6 +558,15 @@ impl MagicEndpoint {
     }
 
     /// Call to notify the system of potential network changes.
+    ///
+    /// On many systems iroh is able to detect network changes by itself, however
+    /// some systems like android do not expose this functionality to native code. 
+    /// Android does however provide this functionality to Java code.  This
+    /// function allows for notifying iroh of any potential network changes like
+    /// this.
+    ///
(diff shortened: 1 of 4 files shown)


dignifiedquire:
> Is it fine to call this on non-Android platforms as well because it is debounced anyway? If both the iroh-using app and iroh detect network change, this would not result in double STUN request, right?

It should be fine now, yes. The original version I had would not have been, but now that it is fed properly through the internal detector, feel free to call it unconditionally

iroh-net/src/magic_endpoint.rs
```diff
@@ -557,6 +557,11 @@ impl MagicEndpoint {
         Ok(())
     }
 
+    /// Call to notify the system of potential network changes.
```
flub, line 560:
```suggestion
    /// Call to notify the system of potential network changes.
    ///
    /// On many systems iroh is able to detect network changes by itself, however
    /// some systems like android do not expose this functionality to native code. 
    /// Android does however provide this functionality to Java code.  This
    /// function allows for notifying iroh of any potential network changes like
    /// this.
    ///
    /// Even when the network did not change, or iroh was already able to detect
    /// the network change itself, there is no harm in calling this function.
```

flub (approved):

PR #1847, opened 2023-11-27
refactor: streamline local endpoint discovery

dignifiedquire:
To avoid the issues seen when setting up an `iroh-net` node, this allows to dynamically subscribe to endpoint changes. It also removes the other callbacks on the magicsock, which where unused.

### DIFF
--- a/iroh-gossip/examples/chat.rs
+++ b/iroh-gossip/examples/chat.rs
@@ -110,9 +110,6 @@ async fn main() -> anyhow::Result<()> {
         .await?;
     println!("> our node id: {}", endpoint.node_id());
 
-    // wait for a first endpoint update so that we know about our endpoint addresses
-    let _endpoints = endpoint.local_endpoints().await?;
-
     let my_addr = endpoint.my_addr().await?;
     // create the gossip protocol
     let gossip = Gossip::from_endpoint(endpoint.clone(), Default::default(), &my_addr.info);
--- a/iroh-gossip/examples/chat.rs
+++ b/iroh-gossip/examples/chat.rs
@@ -1,4 +1,4 @@
-use std::{collections::HashMap, fmt, str::FromStr, sync::Arc};
+use std::{collections::HashMap, fmt, str::FromStr};
 
 use anyhow::{bail, Context};
 use bytes::Bytes;
@@ -17,7 +17,6 @@ use iroh_net::{
 };
 use once_cell::sync::OnceCell;
 use serde::{Deserialize, Serialize};
-use tokio::sync::Notify;
 use url::Url;
 
 /// Chat over iroh-gossip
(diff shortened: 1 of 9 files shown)


iroh-net/src/magic_endpoint.rs
```diff
@@ -380,6 +352,12 @@ impl MagicEndpoint {
         self.msock.local_endpoints().await
     }
 
+    /// The same as `local_endpoints`, except it waits for the first endpoint update.
```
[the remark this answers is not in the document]
dignifiedquire: the goal is that you can register on this again and again, but I guess it doesn't quite work the way I had hoped..

iroh-net/src/magic_endpoint.rs
```diff
@@ -380,6 +352,11 @@ impl MagicEndpoint {
         self.msock.local_endpoints().await
     }
 
+    /// Returns the next non empty version of `local_endpoints`.
+    pub async fn ensure_local_endpoints(&self) -> Result<Vec<config::Endpoint>> {
```
flub, line 356: why not build this into the `local_endpoints()` call instead of adding a new API?  The first time it is called it could ensure that it waits - it already is async.

The address-related API of MagicEndpoint is already large, and this addition does not make it clearer.  This is supposed to be the high-level-easy-to-use API...

Septias:
How is progress here? This is kind of a blocker for https://github.com/deltachat/deltachat-core-rust/pull/5041

iroh-gossip/examples/chat.rs
```diff
@@ -106,35 +105,17 @@ async fn main() -> anyhow::Result<()> {
     // init a cell that will hold our gossip handle to be used in endpoint callbacks
     let gossip_cell: OnceCell<Gossip> = OnceCell::new();
```
Septias, line 106: unused variable I guess

iroh-net/src/magic_endpoint.rs
dignifiedquire: yes, agreed, fixed

iroh-gossip/examples/chat.rs
dignifiedquire: fixed

iroh-net/src/magicsock.rs
```diff
@@ -2032,11 +1990,15 @@ impl Actor {
...
-            if let Some(ref cb) = self.inner.on_endpoints {
-                cb(&eps[..]);
-            }
+        let updated = self
+            .inner
+            .endpoints
+            .update(DiscoveredEndpoints::new(eps))
+            .is_ok();
+        if updated {
+            let eps = self.inner.endpoints.read();
```
flub, line 1999: I think this code is fine.  But seeing `.read()` here did make me go round and check all the docs on what exactly we're reading, who we're notifying and why we must have a lock.  Instinctively I would have cloned the endpoints before storing and then worked of the clone.  It's much of a muchness though.

iroh-net/src/magicsock.rs
```diff
@@ -1316,14 +1272,19 @@ impl MagicSock {
...
-        self.inner
-            .actor_sender
-            .send(ActorMessage::LocalEndpoints(s))
-            .await?;
-        let res = r.await?;
-        Ok(res)
+        {
+            // check if we have some value already
+            let current_value = self.inner.endpoints.read();
+            if !current_value.is_empty() {
```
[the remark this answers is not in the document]
dignifiedquire: it is expected on startup for this to be empty

iroh-net/src/magicsock.rs
dignifiedquire: I did it this way, to avoid having to clone in all cases, and only if we actually need to do so

iroh-net/src/magicsock.rs
```diff
@@ -1282,6 +1282,12 @@ impl MagicSock {
                 return Ok(current_value.clone().into_iter().collect());
             }
         }
+
+        self.local_endpoints_change().await
+    }
+
+    /// Waits for local endpoints to change and returns the new ones.
+    pub async fn local_endpoints_change(&self) -> Result<Vec<config::Endpoint>> {
```
flub, line 1290: I can't help but bikeshed, something about `_change` doesn't feel right.  It's almost like you want to return a `Watchable<Vec<config::Endpoint>>`, but that would expose a foreign crate on the API which is no good.

Maybe just `local_endpoints_updated()` is better already?  I think the tense of `change` is also bothering me.

dignifiedquire: both `changed` and `updated` sound to me like it should return a boolean to me

dignifiedquire: maybe `next_local_endpoints`, matching the stream api, as it behaves similarly like that

flub: `next_local_endpoints` gives me the expectation of an `Option`... naming is hard.  maybe putting the verb at the front is better? `updated_local_endpoints`

let's not make this a blocker, it's not worth it.  and whatever we'll come up with now probably won't be great anyway.  maybe someone will realise a great name while using it and then we can change the name as long as we're not 1.0.

PR #1854, opened 2023-11-29
refactor: client `Entry` with methods to read content

Frando:
## Description

This adds a `iroh::client::Entry` struct which wraps the `iroh_sync::Entry` and adds methods to read the content of the entry. Reading the entry needs a reference to either a `Doc<C>` or the `Iroh<C>` client (both work equally).

@flub mentioned that this is missing from the expected API. I started this at the hack session but didn't get very far but now it's in a good shape IMO.

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh/examples/client.rs
+++ b/iroh/examples/client.rs
@@ -6,9 +6,9 @@
 //! run this example from the project root:
 //!     $ cargo run --example client
 use indicatif::HumanBytes;
-use iroh::node::Node;
+use iroh::{client::Entry, node::Node};
 use iroh_base::base32;
-use iroh_sync::{store::Query, Entry};
+use iroh_sync::store::Query;
 use tokio_stream::StreamExt;
 
 #[tokio::main]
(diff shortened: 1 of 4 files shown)


dignifiedquire:
why not store the client on the `Entry`?

Frando:
> why not store the client on the `Entry`?

That's what I did first. It makes the `Entry` generic over `C: ServiceConnection<ProviderService>`. And because it holds an `RpcClient<C>` it also cannot implement `Deserialize`. It would also make the `LiveEvent` generic over `C` and also not deserializable. So I thought to maybe stop making everything generic at that level, because `LiveEvent<C>` felt wrong. We could do two `Entry` structs, e.g. `EntryClient` and `Entry` or so, but to me the version with passing a client into the methods seems simpler. Open to other thoughts so!

dignifiedquire:
yeah that makes sense, this api is nice enough :)

dignifiedquire (approved):

Issue #1662, opened 2023-10-18
iroh-bytes: consolidate storage paths for flat store


rklaehn:
So we could have a dir blobs.v1 that would then contain the various subdirs there or just smash it all together.
Having partial and non partial in different paths is a bit of a risk anyway, since then they might be on different drives, and we rely on atomic rename.

https://discord.com/channels/949724860232392765/950683937661935667/1164162436048375818

PR #1855, opened 2023-11-30
feat(bytes): switch to a single directory for the flat store

rklaehn:
## Description

Have the constructor for the bao db take a single directory. Internally it will still have 3 directories, although this is not strictly speaking necessary because the files can be distinguished by their extension.

Implements https://github.com/n0-computer/iroh/issues/1662

## Notes & open questions

~This does not yet include a migration. Do we need one, or do we declare that the number of changes is just too big for this release? It's not super hard to write, but we would then have to drag it around for a while.~

Added a migration

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-bytes/src/store/flat.rs
+++ b/iroh-bytes/src/store/flat.rs
@@ -1173,17 +1173,27 @@ impl Store {
         Ok(())
     }
 
+    /// Path to the directory where complete files and outboard files are stored.
+    pub(crate) fn complete_path(root: &Path) -> PathBuf {
+        root.join("complete")
+    }
+
+    /// Path to the directory where partial files and outboard are stored.
+    pub(crate) fn partial_path(root: &Path) -> PathBuf {
+        root.join("partial")
+    }
+
+    /// Path to the directory where metadata is stored.
+    pub(crate) fn meta_path(root: &Path) -> PathBuf {
+        root.join("meta")
+    }
+
     /// scan a directory for data
-    pub(crate) fn load_sync(
-        complete_path: PathBuf,
-        partial_path: PathBuf,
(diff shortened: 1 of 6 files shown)


iroh-bytes/src/store/flat.rs
```diff
@@ -1173,17 +1173,27 @@ impl Store {
         Ok(())
     }
 
+    /// Path to the directory where complete files and outboard files are stored.
+    pub(crate) fn complete_path(root: &Path) -> PathBuf {
```
rklaehn, line 1177: These are pub(crate) intentionally since this is an internal detail. This means that some tests have to manually duplicate the code, but I think that is fine.

iroh/src/commands/start.rs
```diff
@@ -124,6 +124,49 @@ where
     Ok(())
 }
 
+/// Migrate the flat store from v0 to v1. This can not be done in the store itself, since the
+/// constructor of the store now only takes a single directory.
+fn migrate_flat_store_v0_v1() -> anyhow::Result<()> {
```
dignifiedquire, line 129: can you add a test for this, I know it is annoying but we had issues with migrations a couple of times in the wild now
the main cases to cover are (1) doesn't fail/works on a fresh new setup (2) works with an old version (3) doesn't do anything on a new version

rklaehn: I added 2. I think every other cli test test 1. Not sure what to do about 3. I guess I could look at the log and make sure the lines indicating that a migration is happening are not appearing. But that would be super fragile. So short of adding some kind of event for this I would not know how to test this.

dignifiedquire (approved):

PR #1858, opened 2023-12-01
refactor: Upgrade to hyper 1.0

flub:
## Description

This upgrades our direct use of hyper to version 1.0 of the crate.

## Notes & open questions

This is adds a little duplication to our deps for now since reqwest is still
using hyper 1.0.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -229,7 +229,7 @@ version = "0.16.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fdb8867f378f33f78a811a8eb9bf108ad99430d7aad43315dd9319c827ef6247"
 dependencies = [
- "http",
+ "http 0.2.11",
  "log",
  "url",
  "wildmatch",
@@ -252,9 +252,9 @@ dependencies = [
  "bitflags 1.3.2",
  "bytes",
  "futures-util",
- "http",
- "http-body",
- "hyper",
+ "http 0.2.11",
+ "http-body 0.4.5",
+ "hyper 0.14.27",
  "itoa",
  "matchit",
  "memchr",
@@ -278,8 +278,8 @@ dependencies = [
  "async-trait",
  "bytes",
  "futures-util",
- "http",
- "http-body",
+ "http 0.2.11",
+ "http-body 0.4.5",
  "mime",
  "rustversion",
  "tower-layer",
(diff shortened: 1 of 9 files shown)


iroh-metrics/Cargo.toml
```diff
@@ -15,12 +15,16 @@ rust-version = "1.72"
...
-hyper = { version = "0.14.25", features = ["server", "client", "http1", "tcp"] }
+anyhow = "1.0.75"
 erased_set = "0.7"
+http-body-util = "0.1.0"
+hyper = { version = "1", features = ["server", "http1"] }
+hyper-util = { version = "0.1.1", features = ["tokio"] }
+once_cell = "1.17.0"
+prometheus-client = { version = "0.22.0", optional = true }
 struct_iterable = "0.1"
+tokio = "1"
```
dignifiedquire, line 26: this pull in all features, please minimize those as needed

iroh-metrics/src/service.rs
```diff
@@ -1,43 +1,45 @@
...
-        .serve(make_service_fn(move |_conn| async move {
-            let handler = make_handler();
-            Ok::<_, io::Error>(service_fn(handler))
-        }))
-        .await
+    let listener = TcpListener::bind(metrics_addr).await?;
+    loop {
+        let (stream, _addr) = listener.accept().await?;
+        let io = hyper_util::rt::TokioIo::new(stream);
+        tokio::spawn(async move {
```
dignifiedquire, line 20: why do we need to spawn now?

Cargo.lock
```diff
@@ -252,9 +252,9 @@ dependencies = [
  "bitflags 1.3.2",
  "bytes",
  "futures-util",
- "http",
- "http-body",
- "hyper",
+ "http 0.2.11",
+ "http-body 0.4.5",
+ "hyper 0.14.27",
```
dignifiedquire, line 257: this means we have a variety of `hyper` versions now, increasing our dependency graph :/ have you checked how many crates we depend on/that need upgrading until we are back to one version?

flub: reqwest and igd for released bits.

console-subscriber -> tonic -> axum for dev dependencies.  axum already exists.

iroh-metrics/src/service.rs
flub: Because we need to write the accepting loop by hand, previously `Server.serve()` did the spawning but they took away the whole `Server` thing and are only given lower-level bits now.

iroh-metrics/Cargo.toml
flub: ccb5c50f

Cargo.lock
dignifiedquire: filed https://github.com/sbstp/rust-igd/issues/58

dignifiedquire: reqwest: https://github.com/seanmonstar/reqwest/issues/2039

dignifiedquire: tonic: https://github.com/hyperium/tonic/issues/1579

dignifiedquire: looks like it will be a while until those libs upgrade :(

iroh-metrics/src/service.rs
dignifiedquire: I see

dignifiedquire (approved):

PR #1862, opened 2023-12-05
feat: usage metrics reporting

Arqu:
## Description

Simple extension to the metrics collector to have a dedicated usage reporting flow.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/src/derp/server.rs
+++ b/iroh-net/src/derp/server.rs
@@ -9,7 +9,8 @@ use std::time::Duration;
 use anyhow::{Context as _, Result};
 use futures::SinkExt;
 use hyper::HeaderMap;
-use iroh_metrics::inc;
+use iroh_metrics::core::UsageStatsReport;
+use iroh_metrics::{inc, report_usage_stats};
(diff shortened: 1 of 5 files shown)


iroh-net/src/derp/server.rs
```diff
@@ -500,8 +501,18 @@ where
...
                            tracing::trace!("create client: {:?}", client_builder.key);
                            let key = client_builder.key;
+
+                           report_usage_stats(&UsageStatsReport::new(
+                                "derp_accepts".to_string(),
+                                self.key.to_string(),
+                                1,
+                                None, // TODO(arqu): attribute to user id; possibly with the re-introduction of request tokens or other auth
+                                Some(key.to_string()),
+                            )).await.unwrap_or_else(|e| tracing::warn!("error reporting usage stats: {:?}", e));
```
flub, line 514: For metrics-like APIs it might be better to make the functions infallible and handle the error reporting in the function itself.  I assume this will end up being strewn around the code in many places?

flub (approved):

PR #1865, opened 2023-12-06
chore(ci): Configure sccache to use local caches

flub:
## Description

Because we use self-hosted runners we do not want the sccache action
to query the github action cache service for each compilation request.
Instead it should be using the local cache.

This means the first time something is compiled on a runner it will
not yet be in the cache, even if another runner already did compile
this.  However because the cache size is set reasonably large over
time all the common dependencies will end up cached on each
self-hosted runner.

Maybe it will be possible to run a self-hosted caching service
sometime, but for now relying on a large local cache is expected to be
reasonable.

## Notes & open questions

- This will not start showing results right away, as caches on the
  self-hosted runners need to be filled.  They are full by now though
  and show improvements across the board.
  
- Please check that all local runners have 100G space for a cache in
  `~/.cache/sccache` or `%LOCALAPPDATA%\Mozilla\sccache`.

## Change checklist

- [x] Self-review.
- ~~Tests if relevant.~~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -142,8 +142,8 @@ jobs:
 
     # TODO: disabled while we do not have Powershell 7+ on the runner.
     # See https://github.com/actions/toolkit/issues/1179
-    # - name: Install sccache
-    #   uses: mozilla-actions/sccache-action@v0.0.3
+    - name: Install sccache
+      uses: mozilla-actions/sccache-action@v0.0.3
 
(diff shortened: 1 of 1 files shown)


flub:
we could probably easily half the size of the cache based on current numbers

Arqu:
Can I take you up on the offer to drop the cache to something like 60G, the arm runner has a 100G disk in total. All the others are fine

Arqu (approved):

PR #1867, opened 2023-12-06
Use cargo-nextest to run the tests

### LINKED ISSUE #1864: Proposal: Improve CI runtime
The build-and-test jobs are taking more than 10 minutes.  Here some proposals to speed this up:

- Configure sccache on the self-hosted runners:
  
  We run on self-hosted runners but use `SCCACHE_GHA_ENABLED`.  We should instead set a `SCCACHE_CACHE_SIZE=80G` and use the (default) local cache mechanism.  This should mean that soon all runners will have a local cache with all dependencies and we'll spend no more time on compiling dependencies.

- Reduce the number of steps:
  
  We do not need a separate `check` step, it achieves nothing.  Instead run `test` straight away.

- Split default-features and no-features into separate jobs
  
  Because these jobs are compiling iroh with  different feature flags we can parallelise these more by moving the tests to two jobs: `test-default-features`, `test-no-features`.  Only worth it if we have enough runners.  Maybe we should get more runners if that's needed.

- Switch to cargo-nextest

flub:
## Description

This balances the tests accross multiple cores better and speeds up our test run.

Closes #1864 

## Notes & open questions

This completes everything suggested in #1864.  Runtimes can now be
further improved by getting more runners.  Probably 2 more of each
for linux, mac and windows is a good start.

## Change checklist

- [x] Self-review.
- ~~Tests if relevant.~~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -134,9 +134,10 @@ jobs:
         rustup target add ${{ matrix.target }}
         rustup set default-host ${{ matrix.target }}
 
-    # Seems like taiki-e/install-action does not want to work on
-    # self-hosted windows runners for now.  So install manually.
     - name: Install cargo-nextest
+      # uses: n0-computer/install-action@win
+      # with:
+      #   tool: nextest
       shell: powershell
       run: |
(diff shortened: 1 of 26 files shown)


flub:
Total run time reduced from 1h50 to 1h10 :tada:

dignifiedquire (approved):
very cool

PR #1870, opened 2023-12-06
feat(sync) store and get local download policies per document

divagant-martian:
## Description

Implements **local** download policies. 

Download policies are have a general starting point of either everything or nothing, to which an arbitrary number of exceptions can be added. This allows to solve the original use case of "do not download node_modules". 
Exceptions are done in the form of exact matches or prefixes. 

I looked into perhaps having a `.gitignore` format but every single gitignore impl I found requires a working git setup, which in turns needs at least a file system setup. Of course, we have neither. I also looked into general globing, but again all implementations require a file system and return a visitor over files, again not suitable for our use case. As a middleground, exact matches and prefixes are consistent with current primitives and allow a (for now) general enough approximation to download policies. _If_ the need for something more flexible arises, this is an easy change to make.

## Notes & open questions

This is connected to the console as our only current user interface. The interface for download policies is ugly, but I see no better alternative. According to recent discussions, this is a developer (API) interface, much more than an console user api. It would be good anyway, to start working on the more general ux (tauri app) where complex settings are this can be presented in a much better way.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh-sync/src/store.rs
```diff
@@ -175,6 +181,23 @@ pub trait Store: std::fmt::Debug + Clone + Send + Sync + 'static {
 
     /// Get peers to use for syncing a document.
     fn get_sync_peers(&self, namespace: &NamespaceId) -> Result<Option<Self::PeersIter<'_>>>;
+
+    /// Set the download policy for a document.
+    fn set_download_policy(&self, namespace: &NamespaceId, policy: DownloadPolicy) -> Result<()>;
+    /// Get the download policy for a document.
+    fn get_download_policy(&self, namespace: &NamespaceId) -> Result<DownloadPolicy>;
```
dignifiedquire, line 188: why is this duplicated from the `DownloadPolicyStore` trait?

iroh-sync/src/store.rs
```diff
@@ -175,6 +181,23 @@ pub trait Store: std::fmt::Debug + Clone + Send + Sync + 'static {
 
     /// Get peers to use for syncing a document.
     fn get_sync_peers(&self, namespace: &NamespaceId) -> Result<Option<Self::PeersIter<'_>>>;
+
+    /// Set the download policy for a document.
+    fn set_download_policy(&self, namespace: &NamespaceId, policy: DownloadPolicy) -> Result<()>;
```
dignifiedquire, line 186: why is this not on `DownloadPolicyStore`?

divagant-martian: I thought about it as a read only approach. It's intentional but yeah could be ready and write

iroh-sync/src/store.rs
divagant-martian: That's a good point. I guess both can be moved completely to the alt trait

iroh/src/commands/doc.rs
```diff
@@ -570,6 +612,41 @@ impl DocCommands {
                     println!("Aborted.")
                 }
             }
+            Self::DlPolicy(DlPolicyCmd::Set { doc, kind, except }) => {
+                let doc = get_doc(iroh, env, doc).await?;
+                let download_policy = match kind {
+                    FetchKind::Everything => DownloadPolicy::EverythingExcept(except),
+                    FetchKind::Nothing => DownloadPolicy::NothingExcept(except),
+                };
+                if doc.set_download_policy(download_policy).await.is_err() {
```
dignifiedquire, line 621: maybe log the error?

iroh/src/commands/doc.rs
```diff
@@ -44,6 +44,45 @@ pub enum DisplayContentMode {
...
+        doc: Option<NamespaceId>,
+        /// Set the general download policy for this document.
+        kind: FetchKind,
+        /// Add an exception to the download policy.
+        /// An exception must be formated as <matching_kind>:<encoding>:<pattern>.
+        ///
+        /// - <matching_kind> can be either `prefix` or `exact`.
+        ///
+        /// - <encoding> can be either `utf8` or `hex`.
+        #[clap(short, long, value_name = "matching_kind>:<encoding>:<pattern")]
```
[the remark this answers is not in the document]
divagant-martian: Clap adds them. So of I add them you get double on the edges

dignifiedquire (approved):

dignifiedquire (approved):

PR #1872, opened 2023-12-07
chore(ci): Minimise stomping over build artefacts in target

flub:
## Description

This changes the sequence in the test step to minimise re-compiling crates because of different feature sets.

## Notes & open questions

An alternative to https://github.com/n0-computer/iroh/pull/1871

The downside of this approach is that adding more runners will not help
further speedup.

The upside is that as we currently don't have more runners we don't lose
in wall-clock time.  Plus the yaml is less complex due to less duplication.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -142,8 +142,8 @@ jobs:
 
     # TODO: disabled while we do not have Powershell 7+ on the runner.
     # See https://github.com/actions/toolkit/issues/1179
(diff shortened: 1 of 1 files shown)


flub:
This seems to save about 1m of compilation time compared with the baseline (ubuntu, nightly)

flub:
we're going the other route.  more runners!

PR #1884, opened 2023-12-13
chore(ci): test derper deploy flow

Arqu:
## Description

Deploys the test derper on a daily basis at 4am. Can also be manually triggered to bump.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/.github/ansible/redeploy-derper.yml
+++ b/.github/ansible/redeploy-derper.yml
@@ -1,0 +1,25 @@
+---
+- name: Update derper node
+  hosts: derper
+  become: yes
+
+  tasks:
+  - name: Make sure derper is stopped
+    ansible.builtin.systemd:
+      state: stopped
+      enabled: yes
+      name: derper
+  - name: Fetch derper binary
+    get_url:
+      url: https://vorc.s3.us-east-2.amazonaws.com/derper-linux-amd64-latest
+      mode: '0755'
+      force: yes
+      dest: /usr/local/bin/derper
+  - name: Allow ports
+    shell:
+      cmd: sudo setcap CAP_NET_BIND_SERVICE=+eip /usr/local/bin/derper
+  - name: Make sure derper is started
+    ansible.builtin.systemd:
+      state: restarted
+      enabled: yes
+      name: derper
--- a/.github/ansible/redeploy-derper.yml
+++ b/.github/ansible/redeploy-derper.yml
@@ -11,7 +11,7 @@
       name: derper
   - name: Fetch derper binary
     get_url:
-      url: https://vorc.s3.us-east-2.amazonaws.com/derper-linux-amd64-latest
+      url: https://vorc.s3.us-east-2.amazonaws.com/derper-linux-amd64-{{ derper_version }}
       mode: '0755'
       force: yes
       dest: /usr/local/bin/derper
--- a/.github/ansible/redeploy-derper.yml
+++ b/.github/ansible/redeploy-derper.yml
@@ -4,11 +4,6 @@
   become: yes
 
   tasks:
-  - name: Make sure derper is stopped
-    ansible.builtin.systemd:
-      state: stopped
-      enabled: yes
-      name: derper
   - name: Fetch derper binary
     get_url:
       url: https://vorc.s3.us-east-2.amazonaws.com/derper-linux-amd64-{{ derper_version }}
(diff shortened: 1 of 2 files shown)


.github/ansible/redeploy-derper.yml
```diff
@@ -0,0 +1,25 @@
...
+  become: yes
+
+  tasks:
+  - name: Make sure derper is stopped
+    ansible.builtin.systemd:
+      state: stopped
+      enabled: yes
+      name: derper
+  - name: Fetch derper binary
+    get_url:
```
flub, line 13: they recommend
```suggestion
    ansible.builtin.get_url:
```
apparently

flub: So I can't really find how atomically this writes to the destination.  Which seems such a basic question which is not answered.  But what happens when the download fails halfway?  Is the original file left untouched?

Also, how does that work for the entire flow here?  Is the restart not happening in that case and this service is left stopped?

.github/ansible/redeploy-derper.yml
```diff
@@ -0,0 +1,25 @@
...
+      url: https://vorc.s3.us-east-2.amazonaws.com/derper-linux-amd64-{{ derper_version }}
+      mode: '0755'
+      force: yes
+      dest: /usr/local/bin/derper
+  - name: Allow ports
+    shell:
+      cmd: sudo setcap CAP_NET_BIND_SERVICE=+eip /usr/local/bin/derper
+  - name: Make sure derper is started
+    ansible.builtin.systemd:
+      state: restarted
```
flub, line 23: it seems to me that this is the only systemd interaction needed.  The stopping earlier can only result in a stopped service if the download fails I'd expect (I'm guessing at how ansible works here).  But there should be nothing wrong with downloading the binary while the service is still running.

.github/workflows/test_derper.yml
```diff
@@ -0,0 +1,70 @@
...
+        env:
+            # Using self-hosted runners so use local cache for sccache and
+            # not SCCACHE_GHA_ENABLED.
+            RUSTC_WRAPPER: "sccache"
+            RUST_BACKTRACE: full
+            RUSTV: stable
+        steps:
+        - uses: actions/checkout@v2
+        - name: Install rust
+          run: rustup toolchain install stable
```
flub, line 32: why choose this over the `dtolnay/rust-toolchain@stable` action we use everywhere else?

.github/workflows/test_derper.yml
```diff
@@ -0,0 +1,70 @@
...
+        
+
+    deploy_derper:
+        runs-on: ubuntu-latest
+        if: github.ref_name=='main'
+        needs: build_derper
+        steps:
+        - uses: actions/checkout@v2
+        - name: Run Staging Deploy Playbook
+          uses: arqu/action-ansible-playbook@master
```
flub, line 63: why forked?

flub (approved):
oops, didn't mean to block this.

.github/workflows/test_derper.yml
Arqu: I don't trust random actions on other peoples repos and tied to main/master.

.github/workflows/test_derper.yml
Arqu: Cause I copy and pasted it from the iroh ci.yml file 😢 
I'll update what I find.

.github/ansible/redeploy-derper.yml
Arqu: Let's try 🤷

.github/ansible/redeploy-derper.yml
Arqu: Yes a failed step would leave the derper stopped in this case.

Also from the docs
`unsafe_writes`
```
By default this module uses atomic operations to prevent data corruption or inconsistent reads from the target files, but sometimes systems are configured or just broken in ways that prevent this. One example is docker mounted files, which cannot be updated atomically from inside the container and can only be written in an unsafe manner.
```

.github/workflows/test_derper.yml
flub: :+1: though, should probably clone them to n0-computer org

.github/ansible/redeploy-derper.yml
flub: yeah, i also found `unsafe_writes` which i interpreted as a very weak hint that by default they try and do the write atomically.  but this isn't described at all :(

Indeed, let's try it and assume it only replaces the file atomically if the download succeeds.  It'd just be nice to document these kind of things...

PR #1885, opened 2023-12-13
fix: do not block on network change

dignifiedquire:
There was a deadlock due to the usage of the inner actor message loop in handling network changes, which would trigger a deadlock. 

Additionally the following two fixes are included
- clearing endpoint state when connectivity is changed
- drop incoming messages when the magicsock is overloaded

### DIFF
--- a/iroh-net/src/magicsock.rs
+++ b/iroh-net/src/magicsock.rs
@@ -1664,7 +1664,7 @@ impl Actor {
     }
 
     async fn handle_network_change(&mut self, is_major: bool) {
-        warn!("link change detected: major? {}", is_major);
+        debug!("link change detected: major? {}", is_major);
 
         if is_major {
             // Clear DNS cache
@@ -3117,7 +3117,7 @@ pub(crate) mod tests {
             let delay = rand::thread_rng().gen_range(10..=500);
             Duration::from_millis(delay)
         };
-        let rounds = 20;
+        let rounds = 10;
 
         let m1_t = m1.clone();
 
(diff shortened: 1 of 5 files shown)


iroh-net/src/magicsock.rs
```diff
@@ -2934,7 +2930,7 @@ pub(crate) mod tests {
             };
         }
 
-        for i in 0..10 {
+        for i in 0..100 {
```
Arqu, line 2933: can we bring this back down to 10 when we want to merge?

Arqu (approved):
LGTM but pls drop 100 > 10, saves about a minute of CI time.

iroh-net/src/magicsock.rs
dignifiedquire: yeah, I just wanted to make sure this is stable first

iroh-net/src/magicsock/derp_actor.rs
```diff
@@ -233,10 +235,10 @@ impl ActiveDerp {
                             src: source,
                             buf: data,
                         };
-                        self.msg_sender
-                            .send(ActorMessage::ReceiveDerp(res))
-                            .await
-                            .ok();
+                        if let Err(err) = self.msg_sender.try_send(ActorMessage::ReceiveDerp(res)) {
```
dignifiedquire, line 238: I have seen this happen a lot in shutdown, so I think this seems like a good idea, but also means we are not slowing down here, when the magicsock is overloaded.

PR #1893, opened 2023-12-15
Disable gc_partial test and increase timeouts

rklaehn:
let's disable it for now.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


flub:
I've seen these fail on macos plenty of times today as i was working on mac today as well.  let's disable it for all platforms?

flub (approved):

PR #1896, opened 2023-12-16
chore(ci): Try out hosted runners

flub:
We don't have enough runners.  Let's see how much the slowdown is when using hosted runners.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


flub:
without larger machines this won't work

PR #1897, opened 2023-12-18
Move tickets around

rklaehn:
## Description

This moves the ticket code around.

The ticket trait goes into iroh-base and becomes public
Blob and node ticket go to iroh-net
Doc tickets remain in iroh itself

This also makes the trait completely open, kind is not an enum but a `&'static str`. Postcard encoding is just encouraged via docs, not enforced.

The purpose of this is to be able to use tickets such as the node ticket from [dumbpipe](https://github.com/n0-computer/dumbpipe) and the blob ticket from [sendme](https://github.com/n0-computer/sendme).

## Notes & open questions

~Versioning?~ => Added some very basic versioning/variant scheme - just a single variant enum so the postcard bytes have a variant identifier.

I am currently reexporting the tickets from iroh-net in iroh, just so you have all 3 ticket kinds in one place. Not sure if that is necessary. Also, should the trait be called ticket and the individual impl also called ticket?

This is valid in itself, but the way PublicKey seriaizes is broken. We should merge https://github.com/n0-computer/iroh/pull/1900 first.

Note: yes, the additional version adds a byte. But the bugfix in #1900 removes one, so we're net zero bytes... :-)

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


dignifiedquire:
this needs some rebase/merge, a bit many changes currently 😅

rklaehn:
Yeah, I kinda dropped this and went back to main to fix the PublicKey serde thing. Should be better now.

iroh-base/src/ticket.rs
```diff
@@ -0,0 +1,63 @@
...
+///
+/// The serialization format for converting the ticket from and to bytes is left
+/// to the implementer. We recommend using [postcard] for serialization.
+///
+/// [postcard]: https://docs.rs/postcard/latest/postcard/
+pub trait Ticket: Sized {
+    /// String prefix describing the kind of iroh ticket.
+    ///
+    /// This should be lower case ascii characters.
+    fn kind() -> &'static str;
```
dignifiedquire, line 21: why not an associated constant?

iroh-net/Cargo.toml
```diff
@@ -86,6 +86,7 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"], optional = tr
 
 # metrics
 iroh-metrics = { version = "0.11.0", path = "../iroh-metrics", default-features = false }
+strum = { version = "0.25.0", features = ["derive"] }
```
flub, line 89: I'm not a big fan of making sections in this list... because as shown here items end up in the wrong place.

Can you either remove the sections and sort alphabetically or move this to the right secion (still sorted alphabetically i guess).

iroh-net/src/ticket/blob.rs
```diff
@@ -0,0 +1,189 @@
...
+        .unwrap();
+
+        let ticket = Ticket {
+            node: NodeAddr::from_parts(node_id, None, vec![]),
+            format: BlobFormat::Raw,
+            hash,
+        };
+        let base32 = base32::parse_vec(ticket.to_string().strip_prefix("blob").unwrap()).unwrap();
+        let expected = parse_hexdump("
+            00 # discriminator for variant 0
```
flub, line 179: nice to include a test for this!

flub (approved):
Personally I'd have gone with `BlobTicket`, `NodeTicket` etc as naming instead of having everything called `Ticket`.  Compiler errors tend to be easier to understand that way.  But :shrug:.  Naming is hard.

rklaehn:
> Personally I'd have gone with `BlobTicket`, `NodeTicket` etc as naming instead of having everything called `Ticket`. Compiler errors tend to be easier to understand that way. But 🤷. Naming is hard.

Not sure either. We were using the packet as discriminator before, so I stuck with it. But I don't like it much either. I will try your way.

iroh-net/src/ticket/blob.rs
rklaehn: Without this you could easily break things by moving fields around in the ticket.

PR #1899, opened 2023-12-18
fix(iroh): do not exit after commands

dignifiedquire:
Otherwise the console exits after each command


divagant-martian (approved):
looks just like before the changes so should be good

rklaehn (approved):
sorry about that 😊

PR #1911, opened 2023-12-20
Add an additional event FoundLocal.

rklaehn:
## Description

This event needs to be considered when computing how much data needs to be downloaded in case of resume.

Also change hashseq download to assume that at least the hashseq itself is complete. We currently do not support resume if the download of the hashseq itself was interrupted.

## Notes & open questions

1. I think this provides the required info to get good progress bars even in case of resume. But I don't know for sure until I have hacked it into sendme.
2. Sendme does 2 roundtrips and has accurate progress bars. Iroh does 1 request and has a 2 level progress bar because the total size is not known in advance. Should we change iroh to do what sendme does?

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-bytes/src/provider.rs
+++ b/iroh-bytes/src/provider.rs
@@ -155,6 +155,17 @@ pub enum AddProgress {
 /// Progress updates for the get operation.
 #[derive(Debug, Serialize, Deserialize)]
 pub enum DownloadProgress {
+    /// Data was found locally.
+    FoundLocal {
+        /// child offset
+        child: u64,
+        /// The hash of the entry.
+        hash: Hash,
+        /// The size of the entry in bytes.
+        size: u64,
+        /// The ranges that are available locally.
+        valid_ranges: RangeSpec,
+    },
     /// A new connection was established.
     Connected,
     /// An item was found with hash `hash`, from now on referred to via `id`.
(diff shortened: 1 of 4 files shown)


rklaehn:
Note for reviewers: there are not as many changes as it looks like from the PR.

Basically I changed
```rust
let finishing = if let Some(entry) = db.get(root_hash) {
```

to
```rust
    let finishing =
        if let PossiblyPartialEntry::Complete(entry) = db.get_possibly_partial(root_hash) {
```

and that changed the indentation level for all the code below.

iroh-bytes/src/provider.rs
```diff
@@ -155,6 +155,17 @@ pub enum AddProgress {
...
 #[derive(Debug, Serialize, Deserialize)]
 pub enum DownloadProgress {
+    /// Data was found locally.
+    FoundLocal {
+        /// child offset
+        child: u64,
+        /// The hash of the entry.
+        hash: Hash,
+        /// The size of the entry in bytes.
+        size: u64,
```
dignifiedquire, line 165: is this the full size, or the size that resides locally?

rklaehn: No, it is the full size. That is one thing we know as soon as we get any data. To compute the size that is local you need to use this and valid_ranges.

dignifiedquire: I can see how the ranges could be useful, if you want to show a visualization which pieces of a file are already available on disk. But the more common use case seems to be just showing `x/y` has already been downloaded.

dignifiedquire: so maybe at least have a simple helper method on getting the local size?

PR #1914, opened 2023-12-20
feat: basic MDNS discovery

dignifiedquire:
## Description

This uses a simple scheme to discover dialing information of local nodes via MDNS. 
The service is using a hashed version of the node id, to avoid broadcasting the node id directly.

## Notes & open questions

This is not a compile time feature, but just a runtime feature, seems okay to me, but others might have different opinions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -637,6 +637,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "concurrent-queue"
+version = "2.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d16048cd947b08fa32c24458a22f5dc5e835264f689f4f5653210c69fd107363"
+dependencies = [
+ "crossbeam-utils",
+]
+
+[[package]]
 name = "config"
 version = "0.13.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2045,6 +2054,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "if-addrs"
+version = "0.10.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cabb0019d51a643781ff15c9c8a3e5dedc365c47211270f4e8f82812fedd8f0a"
+dependencies = [
+ "libc",
+ "windows-sys 0.48.0",
+]
+
+[[package]]
 name = "igd"
 version = "0.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2378,6 +2397,7 @@ dependencies = [
  "iroh-metrics",
  "iroh-test",
  "libc",
+ "mdns-sd",
  "netlink-packet-core",
  "netlink-packet-route",
  "netlink-sys",
@@ -2659,6 +2679,19 @@ name = "md5"
 version = "0.7.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "490cc448043f947bae3cbee9c203358d62dbee0db12107a74be5c30ccfd09771"
+
+[[package]]
+name = "mdns-sd"
+version = "0.10.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c3a2299267f664ed19a64614562def624c91dfed4d682c1512645db2b050f12d"
+dependencies = [
+ "flume 0.11.0",
+ "if-addrs",
+ "log",
+ "polling",
+ "socket2 0.4.10",
+]
 
 [[package]]
 name = "memalloc"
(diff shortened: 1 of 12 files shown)


iroh-net/src/magic_endpoint.rs
```diff
@@ -135,6 +144,7 @@ impl Default for MagicEndpointBuilder {
             keylog: Default::default(),
             discovery: Default::default(),
             peers_path: None,
+            use_mdns: true,
```
rklaehn, line 147: Keep it off by default for 1 release until we have some experience with it? It might cause trouble.

iroh-net/src/magic_endpoint.rs
```diff
@@ -300,6 +321,53 @@ impl MagicEndpoint {
...
+            let mdns = ServiceDaemon::new().context("mdns")?;
+            let node_id = secret_key.public();
+            // Hash the node id, to avoid publicly sharing it.
+            let service_id = hash::Hash::new(node_id);
+            let service_hostname = "iroh-discovery.local.";
+            let instance_name = format!("{}-iroh", service_id);
+            let (local_ipv4_addr, local_ipv6_addr) = msock.local_addr()?;
+            let port = local_ipv4_addr.port();
+
+            let my_addrs: Vec<_> = msock
```
rklaehn, line 334: What if this info changes? Maybe I missed it, but where is the part that changes this on endpoint change?

rklaehn:
Ah, if we do that tickets that contain nothing but a node id become useful. So we should change NodeTicket::validate and BlobTicket::validate to not consider such tickets as invalid.

iroh-net/src/magic_endpoint.rs
dignifiedquire: there is no updating happening right now, I should add at least a todo for it, yes

iroh-net/src/magic_endpoint.rs
dignifiedquire: yeah, also need to add that option on `iroh::Node`

dignifiedquire: fixed

rklaehn:
Still missing 2 things:

- issue or todo about dealing with endpoint change
- allowing tickets with neither ip addrs nor derp url

dignifiedquire:
> * issue or todo about dealing with endpoint change

this is already being done, the impl itself now monitors ip addr changes

rklaehn:
> > * issue or todo about dealing with endpoint change
> 
> this is already being done, the impl itself now monitors ip addr changes

So this is the .enable_addr_auto(); feature, and we don't hook it up with our own IP addr detection? So we got 2 things running to detect IP changes, this built in thing and ours which also tries to figure out public IPs.

I have not really an idea how this works in detail, but could there be a problem that these two things interfere with each other?

dignifiedquire:
> I have not really an idea how this works in detail, but could there be a problem that these two things interfere with each other?

I don't think there is any risk there, as they just like at the network interfaces of the machine. The only thing that could happen is that they are inconsistent, which is not ideal, but also not a big issue. 

Now it would be nicer to be able to hook up our existing machinery into the library, but that is currently not exposed, and seems like sth to worry about for much later, if ever.

iroh-net/src/magic_endpoint.rs
```diff
@@ -300,6 +321,54 @@ impl MagicEndpoint {
         // the packet if grease_quic_bit is set to false.
         endpoint_config.grease_quic_bit(false);
 
+        let mdns_service = if use_mdns {
+            let mdns = ServiceDaemon::new().context("mdns")?;
+            let node_id = secret_key.public();
+            // Hash the node id, to avoid publicly sharing it.
+            let service_id = hash::Hash::new(node_id);
```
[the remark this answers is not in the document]
dignifiedquire: This should be a cryptographic hash (it is blake3). This is needed to ensure that the node ids are not broadcasted locally.

iroh-net/src/magic_endpoint.rs
```diff
@@ -135,6 +144,7 @@ impl Default for MagicEndpointBuilder {
             keylog: Default::default(),
             discovery: Default::default(),
             peers_path: None,
+            use_mdns: false,
```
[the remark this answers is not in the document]
dignifiedquire: likely

Cargo.lock
```diff
@@ -2044,6 +2053,16 @@ dependencies = [
  "unicode-normalization",
 ]
 
+[[package]]
+name = "if-addrs"
```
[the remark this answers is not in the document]
dignifiedquire: it is not ideal, I agree. probably should change the library to allow interface details to simply be injected

rklaehn:
It is a bit unfortunate that we can't have this be just a discovery mechanism that you can plug into the existing one. I can imagine many scenarios where you want to combine multiple discovery mechanisms. Hardcoding them all is clearly not great.

But I noticed that the discovery trait might need an update to allow this. E.g. discovered nodes should be a stream, and each discovered node should come with some kind of provenance so you know where you know it from.

dignifiedquire:
while fun, this needs more thought and work before we can merge this

dignifiedquire:
> But I noticed that the discovery trait might need an update to allow this. E.g. discovered nodes should be a stream, and each discovered node should come with some kind of provenance so you know where you know it from.

yes, for properly pluggable discovery mechanisms, we would need to extend the existing hooks quite a bit

Issue #1901, opened 2023-12-18
`netcheck::tests::test_udp_blocked` failure


divagant-martian:
seen here https://github.com/n0-computer/iroh/actions/runs/7253050090/job/19758824459

divagant-martian:
and again here https://github.com/n0-computer/iroh/actions/runs/7253555147/job/19760392300

flub:
Sadly these logs are useless: let's try and improve them again: https://github.com/n0-computer/iroh/pull/1902

flub:
Logs with more logging:

https://github.com/n0-computer/iroh/actions/runs/7273863276/job/19818670553
https://github.com/n0-computer/iroh/actions/runs/7273863276/job/19818672125

divagant-martian:
fixed on f2e1c44

PR #1915, opened 2023-12-20
fix(iroh-net): Try fix flaky udp_blocked test

flub:
## Description

This test is flaky because somehow the executor is entirely blocked
for a solid 5 seconds after which the entire netcheck report is timing
out.  Thus the part of the netcheck report that is being tested never
executes.

What blocks the executor however is a mystery to me.  This fix assumes
that it is generating the backtrace in the log, because there simply
isn't anything else that it can be.  Thus the relevant change is no
longer producing the backtrace.  The error message is already pretty
obvious.

To  go further, this disables the anyhow backtrace feature entirely.
This will probably make our lifes a bit harder in some situations but
there are many places where they end up in the log otherwise and each
time this happens it blocks an executor thread for a long time.

This also tweaks the error reporting to not generate so much logging
noise when the captive portal fails in an expected way.

## Notes & open questions

An attempt at fixing #1901 

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/src/netcheck.rs
+++ b/iroh-net/src/netcheck.rs
@@ -949,6 +949,11 @@ mod tests {
         // create raw ICMP pings and we'll have to silenty accept this test is useless (if
         // we could, this would be a skip instead).
         let have_pinger = Pinger::new().is_ok();
+        if !have_pinger {
+            error!("pinger disabled, test effectively skipped");
+        } else {
+            info!("pinger enabled");
+        }
 
         // This is the test: we will fall back to sending ICMP pings.  These should
         // succeed when we have a working pinger.
--- a/iroh-net/src/netcheck/reportgen.rs
+++ b/iroh-net/src/netcheck/reportgen.rs
@@ -249,6 +249,7 @@ impl Actor {
                 break;
             }
             tokio::select! {
+                biased;
                 _ = &mut total_timer => {
                     trace!("tick: total_timer expired");
                     bail!("report timed out");
@@ -533,6 +534,7 @@ impl Actor {
             MaybeFuture {
                 inner: Some(Box::pin(async move {
                     tokio::time::sleep(CAPTIVE_PORTAL_DELAY).await;
+                    debug!("Captive portal check started after {CAPTIVE_PORTAL_DELAY:?}");
                     let captive_portal_check = tokio::time::timeout(
                         CAPTIVE_PORTAL_TIMEOUT,
                         check_captive_portal(&dm, preferred_derp)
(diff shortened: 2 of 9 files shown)


iroh-net/src/netcheck.rs
```diff
@@ -949,6 +949,11 @@ mod tests {
         // create raw ICMP pings and we'll have to silenty accept this test is useless (if
         // we could, this would be a skip instead).
         let have_pinger = Pinger::new().is_ok();
+        if !have_pinger {
+            error!("pinger disabled, test effectively skipped");
+        } else {
+            info!("pinger enabled");
```
divagant-martian, line 955: while useful this will pop up to the user, maybe both should be debug

iroh-net/src/netcheck/reportgen.rs
```diff
@@ -249,6 +249,7 @@ impl Actor {
                 break;
             }
             tokio::select! {
+                biased;
```
divagant-martian, line 252: what's your intuition here?

iroh-net/src/netcheck/reportgen.rs
```diff
@@ -541,7 +543,14 @@ impl Actor {
...
                         Ok(Err(err)) => {
-                            warn!("check_captive_portal error: {:?}", err);
+                            let err: Result<reqwest::Error, _> = err.downcast();
+                            match err {
+                                Ok(req_err) if req_err.is_connect() => {
+                                    debug!("check_captive_portal failed: {req_err:#}");
+                                }
+                                Ok(req_err) => warn!("check_captive_portal error: {req_err:#}"),
+                                Err(any_err) => warn!("check_captive_portal error: {any_err:#}"),
+                            }
```
divagant-martian, line 553: same with these, while useful yo use I don't think they mean much for end users

iroh-net/src/netcheck.rs
flub: This is in a test.  I very much want to see this.

divagant-martian: yeah in a test it makes sense

iroh-net/src/netcheck/reportgen.rs
flub: Generally every select should be biased, unless there is a compelling reason not too.  Not using biased will result in a random number being generated every time which is a bit of a waste usually.

It would probably have helped here as well, since the failure mode from the timers expiring would be a lot more obvious: what it did now was randomly poll some of the other futures first before finally coming round to "oh, guess we should have stopped".  The log output would have been easier to understand.

dignifiedquire (approved):

iroh-net/src/netcheck/reportgen.rs
flub: Here I jumped through some hoops to reduce the expected errors to debug but leave unexpected errors as errors.  It's hard to capture everything and maybe there's still a bit too much on WARN, but I'd like real failures to still show up.

flub:
going to merge this as it keeps kicking PR's out of the merge queue.  happy to follow up to more comments in a followup PR @divagant-martian

PR #1916, opened 2023-12-20
fix(iroh-net): do not prune addrs that are just added

dignifiedquire:
This pruning would delete all direct addrs that were just added, as they had no pings or so attached


divagant-martian (approved):
makes sense!

Issue #1919, opened 2023-12-21
Proposal: handling flaky tests


flub:
# Objective

Ensure flaky tests do not slow down PRs and take up too many CI resources while not neglecting them entirely.

# Background

Despite our best efforts we regularly have flaky tests.  This is problematic because we use a merge queue normally, which means after a PR has all tests passing the tests are run once again when in the merge queue.  If a test then suddenly fails in the merge queue it is removed from the merge queue and needs to be manually put in the merge queue to try again.  Given enough flaky tests PRs can go through several rounds of this before they manage to be merged, clogging up CI resources.

# Proposal

## Ruthlessly removing flaky tests

To keep the merge queue flowing we **must** remove every single instance of a flaky tests.  Each failure of tests in the merge queue should result in either:

* Fixing the test if it was caused by the PR itself.
* Marking the test as `#[ignore = "flaky"]`
  * Creating an issue to resolve this flaky test, labelled `flaky-test`

Simply re-scheduling the PR in the merge queue without doing either of these must never occur.

## Fixing flaky tests

The flaky test was also written for a reason, we can not simply forget about it.  We also have to accept that it needs running in CI to show up.  We should create a new github actions workflow for flaky tests:

* This "flaky workflow" must run the entire test suite, including normal and flaky tests to create the most realistic environment for it to remain flaky.
  * The `--ignored` argument to the test runner allows doing this.
* This "flaky workflow" should run on a cron job, failures should nag us somewhere.
  * Maybe nag us in a discord channel if possible?
  * The cronjob must not take up too many CI resources but still be regular.  Daily is probably a good start.
* It should be possible to manually trigger this flaky workflow.
  * Some issues can only be debugged on CI.  We must allow doing so.
* PRs should be able to target the flaky workfow.
  * When working on flaky tests a PR would waste CI resources if it needs to both run the normal CI workflow and the flaky workflow.
  * Maybe a label on the PR could be used to disable the normal CI workflow and enable the flaky workflow.

Arqu:
👏 like the proposal very much.

ramfox:
Sounds good to me!

flub:
The only thing missing now is some notification for the flaky workflow failures.

dignifiedquire:
@flub where are we with the notifications/reruns?

flub:
reruns are happening: https://github.com/n0-computer/iroh/actions/workflows/flaky.yaml (some are even passing!)

[...]

PR #1922, opened 2023-12-22
chore(ci): Set up a separate workflow for flaky tests

flub:
## Description

This sets up a separate workflow which runs all tests, including flaky ones.  It can be triggered by setting the "flaky-test" label on a PR, which will disable most of the normal CI workflow and instead run just the tests including flaky ones.

The flaky workflow can also be manually triggered on a specific branch.

## Notes & open questions

- The flaky workflow only runs the tests, none of the other jobs.
- Cross-compiled jobs aren't even included.  Maybe they should be but right now they don't run tests at all so is pointless.
- On any labelling change for the PR the jobs in the workflows will be cancelled if they're still running.  I don't think we generally label PRs so that's probably fine.
- Flaky tests are going to be carnage for a while!
- We have no notification for the cron job results yet.  Not sure how to approach this.  I'd be fine doing this later.
- The `workflow_dispatch` version (aka manual trigger) is untested, I think we have to wait until it is merged to main to be able to test that?
- The UI will now always show all jobs: the skipped flaky jobs on normal PRs and the missing normal jobs on flaky PRs.  That's probably fine?  A bit noisier though.
- The set of "required jobs" has changed again and will need patching up in branch protection.

This implements the bulk of https://github.com/n0-computer/iroh/issues/1919

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


flub:
@Arqu could you have an initial look at this and point out all silly things I did the hard way?  I think this is remarkably close already.

If you have any ideas on notifications for the cron job I'd be happy as well.

flub:
Since the tests run in the flaky workflow are so similar to the tests in normal CI, maybe we should check if this is doable with a reusable-workflow instead?  It could take inputs for the few differences (if conditions and test args).

flub:
I rather like this `workflow_call` version.

But now it opens the door to only having a single `ci` workflow instead of a separate `flaky` workflow.

Upsides:
- There would only be one entry in the "Actions" list for workflows for a PR.  Might be simpler?

Downsides:
- The ci action would have a bunch of inputs with defaults.  manually triggering could be a bit weird.

Arqu:
Will take a look tonight.

.github/workflows/flaky.yaml
```diff
@@ -0,0 +1,46 @@
...
+        default: 'main'
+        type: string
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  tests:
+    if: "contains(github.event.pull_request.labels.*.name, 'flaky-test') || contains(github.event, 'workflow_dispatch')"
```
Arqu, line 42: This can be done `github.event_name == 'workflow_dispatch'` for a bit more precision.
Also doesn't cover the `schedule` event and will basically skip it altogether on every trigger.

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,182 @@
+# Run all tests, with or without flaky tests.
+
+name: Tests
```
Arqu, line 3: This is nice.

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,182 @@
...
+                ;;
+            default)
+                echo "FEATURES=" >> "$GITHUB_ENV"
+                ;;
+            *)
+                exit 1
+        esac
+
+    - name: check features
+      if: ${{ ! inputs.flaky }}
```
Arqu, line 90: is `inputs.xyz` the correct syntax here. Afaik these should be subclassed to `github.event.inputs` or `github.inputs`

.github/workflows/ci.yml
```diff
@@ -20,163 +21,12 @@ env:
...
-
-    - uses: msys2/setup-msys2@v2
-
-    - name: tests
-      run: |
-        cargo nextest run --workspace ${{ env.FEATURES }} --lib --bins --tests --target ${{ matrix.target }}
-      env:
-        RUST_LOG: "TRACE"
+  tests:
+    if: "! contains(github.event.pull_request.labels.*.name, 'flaky-test')"
```
Arqu, line 25: I think this will fail outside of a PR context as the `github.event.pull_request` object will be `null`.

.github/workflows/ci.yml
```diff
@@ -2,6 +2,7 @@ name: CI
 
 on:
   pull_request:
+    types: [ 'labeled', 'opened', 'synchronize', 'reopened' ]
```
Arqu, line 5: Nice idea to use labels to control flow. We could do a `no-netsim` flag to skip netsim for in progress PRs later.

.github/workflows/tests.yaml
flub: i believe `inputs` is correct: https://docs.github.com/en/actions/learn-github-actions/contexts#inputs-context

.github/workflows/ci.yml
flub: made it so they always run outside of PR context

flub:
@Arqu PTAL, thanks for the suggestions

.github/workflows/tests.yaml
Arqu: Nice, I used to think it was always scoped to the `github` context.

Arqu:
LGTM now. I guess we need to rebase this and then align branch protections.

flub:
@Arqu merged master (and fixed another bug...)

should be good to merge and adjust the branch protections

flub:
I've also made https://github.com/n0-computer/iroh/wiki/Flaky-Tests as a reference for how this is to be used.

Arqu (approved):

PR #1937, opened 2024-01-08
fix(ci): stop reusing concurrency labels across jobs

Arqu:
## Description

Jobs are getting canceled because multiple separate flows have the same concurrency group.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -16,7 +16,7 @@ on:
 
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yaml
```diff
@@ -14,11 +14,6 @@ on:
         type: string
         default: 'main'
 
-
```
Arqu, line 17: Seems like these start to interfere because they are nesting from the general CI flow. 
Should be fine to keep them running to completion since either the parent group will cancel them on newer runs or you manually triggered it anyways.

flub (approved):
good catch. i was just cargo-culting that around without knowing how it works...

PR #1938, opened 2024-01-08
chore(ci): add name to tests step

Arqu:
## Description

Seems like the lack of job name makes it invisible on the branch protections, hence can't make the step required for the nested runs. Raw `Tests` job is available.  This is just theory for now.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,7 @@ env:
 
 jobs:
   tests:
+    name: CI Test Suite
     if: "github.event_name != 'pull_request' || ! contains(github.event.pull_request.labels.*.name, 'flaky-test')"
     uses: './.github/workflows/tests.yaml'
 


dignifiedquire (approved):

.github/workflows/ci.yml
```diff
@@ -22,6 +22,7 @@ env:
 
 jobs:
   tests:
+    name: CI Test Suite
```
flub, line 25: fwiw since this is used directly in the UI the horizontal space matters.  I'd have made it `tests` to keep the same.

Arqu: > fwiw since this is used directly in the UI the horizontal space matters. I'd have made it `tests` to keep the same.

Agree, I just took something else as it overlaps with `test / Tests` and GH UI is confusing around these. I'll make an issue and revise on another pass.

flub: ah yeah.  that branch protection UI is pretty rubbish too :(

PR #1949, opened 2024-01-12
fix(iroh): add timestamp method to Entry RPC struct

b5:
## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/client.rs
+++ b/iroh/src/client.rs
@@ -871,6 +871,11 @@ impl Entry {
         self.0.key()
     }
 
+    /// Get the timestamp of this entry.
+    pub fn timestamp(&self) -> u64 {
+        self.0.timestamp()
+    }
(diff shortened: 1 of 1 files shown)


iroh/src/client.rs
```diff
@@ -871,6 +871,11 @@ impl Entry {
         self.0.key()
     }
 
+    /// Get the timestamp of this entry.
+    pub fn timestamp(&self) -> u64 {
```
flub, line 875: I'm rather sad this comes as a u64.  Perhaps that's already a mistake in iroh-sync?  Or just implementation details leaking out?  But for a user-level Rust API I'd much rather see something like `time::OffsetDateTime` or so.

dignifiedquire (approved):

PR #1954, opened 2024-01-15
Store metadata in redb

### LINKED ISSUE #1942: Slow boot time with a large repo size
Iroh has a very slow boot time, like more than 1 hour, when the node has a very large repo size.

This is likely due to disk access.  We should evaluate what happens at startup and see if we can make this faster.

dignifiedquire:
This stores metadata in redb instead of in the file system. A part of the data can be inferred from the file system (e.g. partial and complete hashes), another part is exclusively in the redb database (tags).

The upside of this is that it reduces startup delay due to scanning the file system. Also, manipulation of tags should be much faster now since it does not involve IO anymore.

The downside is that now the file system can become inconsistent with the metadata, since there is redundant information.

Closes #1942


iroh-bytes/src/store/flat.rs
```diff
@@ -591,39 +653,31 @@ impl Map for Store {
...
     }
 }
 
 impl ReadableStore for Store {
-    fn blobs(&self) -> Box<dyn Iterator<Item = Hash> + Send + Sync + 'static> {
-        let inner = self.0.state.read().unwrap();
-        let items = inner.complete.keys().copied().collect::<Vec<_>>();
-        Box::new(items.into_iter())
+    fn blobs(&self) -> io::Result<DbIter<Hash>> {
+        // TODO: should return Result
```
dignifiedquire, line 664: this is fixed, so no need to return empty iters on error anymore, should return the errors

iroh-bytes/src/store/flat.rs
```diff
@@ -591,39 +653,31 @@ impl Map for Store {
...
+            return Ok(Box::new(std::iter::empty()));
+        };
+
+        // TODO: avoid allocation
+        let items: Vec<_> = {
+            let Ok(full_table) = read_tx.open_table(COMPLETE_TABLE) else {
+                return Ok(Box::new(std::iter::empty()));
+            };
+            let Ok(iter) = full_table.iter() else {
+                return Ok(Box::new(std::iter::empty()));
```
dignifiedquire, line 675: same as above

iroh-bytes/src/store/flat.rs
```diff
@@ -591,39 +653,31 @@ impl Map for Store {
...
+    fn blobs(&self) -> io::Result<DbIter<Hash>> {
+        // TODO: should return Result
+        let Ok(read_tx) = self.0.db.begin_read() else {
+            return Ok(Box::new(std::iter::empty()));
+        };
+
+        // TODO: avoid allocation
+        let items: Vec<_> = {
+            let Ok(full_table) = read_tx.open_table(COMPLETE_TABLE) else {
+                return Ok(Box::new(std::iter::empty()));
```
dignifiedquire, line 672: same as above

iroh-bytes/src/store/flat.rs
```diff
@@ -1201,21 +1251,82 @@ impl Store {
...
+
+        if needs_migration {
+            res.add_from_disk()?;
+        }
+
+        Ok(res)
+    }
+
+    fn add_from_disk(&self) -> anyhow::Result<()> {
+        // TODO: write results to DB
```
dignifiedquire, line 1308: this is fixed I believe?

iroh-bytes/src/store/flat.rs
```diff
@@ -1201,21 +1251,82 @@ impl Store {
...
+
     /// scan a directory for data
     pub(crate) fn load_sync(path: &Path) -> anyhow::Result<Self> {
         tracing::info!("loading database from {}", path.display(),);
         let complete_path = Self::complete_path(path);
         let partial_path = Self::partial_path(path);
         let meta_path = Self::meta_path(path);
+        let db_path = Self::db_path(path);
+
+        let needs_migration = !db_path.exists();
```
dignifiedquire, line 1267: If the migration is canceled or errors, the db file will exists, but the migration won't be done. Maybe check for all tables to exist to determine if the migration is needed, and only create the tables in the migration inside the transaction there.

iroh-bytes/src/store/flat.rs
```diff
@@ -1430,38 +1544,30 @@ impl Store {
...
+            let mut complete_table = txn.open_table(COMPLETE_TABLE)?;
+            let mut partial_table = txn.open_table(PARTIAL_TABLE)?;
+            let mut tags_table = txn.open_table(TAGS_TABLE)?;
+            for (hash, entry) in complete {
+                complete_table.insert(hash, entry)?;
+            }
+            for (hash, entry) in partial {
+                partial_table.insert(hash, entry)?;
+            }
+            for (tag, target) in tags {
```
dignifiedquire, line 1564: shouldn't this delete the tags and external path files on successs?

iroh-bytes/src/store/flat.rs
rklaehn: yeah...

iroh-bytes/src/store/flat.rs
rklaehn: yeah.

iroh-bytes/src/store/flat.rs
rklaehn: I guess I could also use the tags.meta as an indicator. That is gone afterwards.

rklaehn: ended up just using an entirely new directory so the distinction is clear.

iroh/src/commands/start.rs
```diff
@@ -137,10 +137,18 @@ fn migrate_flat_store_v0_v1() -> anyhow::Result<()> {
...
-        tracing::info!("migrating from v0 to v1");
-        std::fs::create_dir_all(blobs_v1)?;
+    let v0_exists = complete_v0.exists() || partial_v0.exists() || meta_v0.exists();
+    let v1_exists = blobs_v1.exists();
+    if !v0_exists {
+        tracing::debug!("skipping migration from v0 to v1, nothing to migrate");
+        return Ok(());
     }
+    if v1_exists {
+        tracing::debug!("skipping migration from v0 to v1, already migrated");
```
dignifiedquire, line 147: this should be v1 to v2

iroh-bytes/src/store/flat.rs
```diff
@@ -591,39 +653,26 @@ impl Map for Store {
...
 }
 
 impl ReadableStore for Store {
-    fn blobs(&self) -> Box<dyn Iterator<Item = Hash> + Send + Sync + 'static> {
-        let inner = self.0.state.read().unwrap();
-        let items = inner.complete.keys().copied().collect::<Vec<_>>();
-        Box::new(items.into_iter())
+    fn blobs(&self) -> io::Result<DbIter<Hash>> {
+        let Ok(read_tx) = self.0.db.begin_read() else {
+            return Ok(Box::new(std::iter::empty()));
```
dignifiedquire, line 665: return error

iroh/src/commands/start.rs
```diff
@@ -169,6 +176,27 @@ fn migrate_flat_store_v0_v1() -> anyhow::Result<()> {
     Ok(())
 }
 
+fn migrate_flat_store_v1_v2() -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
rklaehn: Not quite sure where to put it. Can you describe what you would want them for?

I don't want to expose details about previous versions of our database too much. That is also why in IrohPaths there is no BlobStoreV0 or BlobStoreV1, but these paths are hardcoded as strings.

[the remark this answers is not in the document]
dignifiedquire: this can be a follow up PR (but ideally part of this release)

I am not sure where exactly to put this, but `blobstorev1` paths have been used for example in the ffi, so that code needs the migration as well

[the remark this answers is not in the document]
dignifiedquire: ref: https://github.com/n0-computer/iroh-ffi/blob/main/src/node.rs#L216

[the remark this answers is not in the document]
rklaehn: So the ffi has it's own iroh node? Yeah, in this case I can see that it needs to be exposed. I am sure there is a good reason, but why does iroh-ffi not include the iroh node?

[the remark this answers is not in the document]
dignifiedquire: > iroh-ffi not include the iroh node?

how would it do that? the `Node::builder` is the expected way to run an iroh node from code

rklaehn:
So now we have a version field in the DB, that starts at 2.
The blobs.v1 is renamed to just blobs.
None of this is exposed, the idea is that the db will take care of versioning within it's directory from now on.

We just need to do the rename in iroh-ffi as well if we want to support v1 compat, but I would prefer to do that manually instead of exposing something.

rklaehn:
It seems that GC is quite a bit slower on this branch compared to main:

gc_flat_stress on this branch:
```
running 1 test
GC mark phase took 131.125µs
GC sweep phase took 659.351166ms
GC mark phase took 180.334µs
GC sweep phase took 314.981834ms
GC mark phase took 138.542µs
GC sweep phase took 41.084µs
test flat::gc_flat_stress ... ok
```

gc_flat_stress on main:
```
running 1 test
GC mark phase took 160.333µs
GC sweep phase took 55.864125ms
GC mark phase took 68.292µs
GC sweep phase took 16.541µs
GC mark phase took 25.083µs
GC sweep phase took 5.833µs
test flat::gc_flat_stress ... ok
```

Why? No idea, but it seems that while this is still fine on a fast dev mac, it puts it over the threshold to failure on the ci mac boxes.

rklaehn:
So it seems that creating a write txn for each delete makes deleting hash by hash much slower.

In general all ops that before were a map lookup (basically free) are now a read txn and all ops that before were a map write and a fs op for the actual content are now a write txn including (presumably) a full disk sync plus the io op itself.

Maybe batching deletes works, but we can't make the batch too large since during the delete the complete_io_guard has to be held otherwise there is a race between files being deleted and being newly added.

iroh-bytes/src/store/flat.rs
```diff
@@ -825,11 +825,10 @@ impl super::Store for Store {
         state.live.contains(hash) || state.temp.contains(hash)
     }
 
-    fn delete(&self, hash: &Hash) -> BoxFuture<'_, io::Result<()>> {
-        tracing::debug!("delete: {:?}", hash);
+    fn delete(&self, hashes: Vec<Hash>) -> BoxFuture<'_, io::Result<()>> {
```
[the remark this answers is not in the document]
rklaehn: We are creating boxed futures left and right, also thousands whenever we use tokio io. So I am not sure worrying about allocations makes any sense anymore once you use tokio async io.

Remember that every single tokio file io call is at least one spawn_blocking.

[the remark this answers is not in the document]
dignifiedquire: 😭

dignifiedquire:
I guess the bug was in the test helper 😅

rklaehn:
OK, I have second thoughts about merging this as is. It completely removes the outboard cache, hereby _completely_ changing the performance characteristics of the thing.

The former version of the flat store had a consistent concept. It might not have been perfect, but it worked for a large set of use cases. This is some weird hybrid between the former concept and something else.

image

dignifiedquire:
this outboard cache was a pretty big footgun given it wasn't limited in size asfaict, so if we bring it back it should be at  minimum  size limited

rklaehn:
> this outboard cache was a pretty big footgun given it wasn't limited in size asfaict, so if we bring it back it should be at minimum size limited

I don't think so. Outboards 1/256 of data size, so for a 1 TiB disk even if you had all outboards in memory it would be just 4 GiB. Ok, could become a problem if you have a RpI 3 with a giant external hard drive...

ppodolsky:
> I don't think so. Outboards 1/256 of data size, so for a 1 TiB disk even if you had all outboards in memory it would be just 4 GiB. Ok, could become a problem if you have a Raspberry Pi 3 with a giant external hard drive...

Ha-ha, exactly me. 8GB RAM Pi with attached 200TB HDD :D Seems I'm quite unlucky customer.

rklaehn:
> > I don't think so. Outboards 1/256 of data size, so for a 1 TiB disk even if you had all outboards in memory it would be just 4 GiB. Ok, could become a problem if you have a Raspberry Pi 3 with a giant external hard drive...
> 
> Ha-ha, exactly me. 8GB RAM Pi with attached 200TB HDD :D Seems I'm quite unlucky customer. May it worth to use some sort of LRU for dat cache?

OK, seriously, thank you for being such a demanding customer.

I am working on a larger refactoring. Basically I want to store small files inline in the redb always, and also store small outboards in the redb once they are complete. This should reduce both the number of files and the number of file operations for many use cases.

Could you do me a favour and compute some sort of histogram of your file sizes, or just do something like

```bash
stat -f "%N %z" * # (Mac)
stat -c "%n %s" * # (Linux)
```

and send me the result?

That would help a lot.

ppodolsky:
I have done this: `find . -type f -print0 | xargs -0 ls -l | awk '{size[int(log($5)/log(2))]++}END{for (i in size) printf("%10d %3d\n", 2^i, size[i])}' | sort -n`

``
         0  11
        32   2
       256   2
       512   1
      2048  78
      4096 983
      8192 2370
     16384 24614
     32768 31862
     65536 61111
    131072 111704
    262144 148639
    524288 160883
   1048576 132321
   2097152 87118
   4194304 51952
   8388608 24728
  16777216 9846
  33554432 4255
  67108864 1883
 134217728 561
 268435456 106
 536870912  45
1073741824  12
```

dignifiedquire:
closing in favor of ongoing PRs refactoring the store in a more step-by-step fashion

PR #1957, opened 2024-01-16
Add protocol level authentication

dignifiedquire:
Brings back the general concept of authenticating requests. 

This is a protocol level breaking change, which is why all ALPN version numbers have been increased.

Currently there are two authenticators implemented 

- `NoAuthenticator`: does nothing
- `IdAuthenticator`: sends a single UUID for all requests, which allows attribution of incoming traffic to this id. By default this authenticator is use in the iroh cli, and the uuid can be set in the config or on the console using `IROH_UUID="my-uuid"`

## Open Questions

- Should gossip authenticate each message, or can we reduce this?

### DIFF
--- a/iroh-base/src/auth.rs
+++ b/iroh-base/src/auth.rs
@@ -1,0 +1,85 @@
+//! Authentication related types and tooling.
+
+use std::ops::Deref;
+use std::sync::Arc;
+
+use anyhow::Result;
+use serde::{Deserialize, Serialize};
+
+use crate::hash::Hash;
+
+/// The error code sent using quinn when aborting due to authentication errors.
+pub const REJECTED_CODE: u32 = 10;
+
+#[derive(Debug, Clone)]
+pub struct Authenticator(Arc<dyn DynAuthenticator>);
+
+impl Deref for Authenticator {
+    type Target = dyn DynAuthenticator;
+
+    fn deref(&self) -> &Self::Target {
+        self.0.deref()
+    }
+}
+
+impl<A: DynAuthenticator> From<A> for Authenticator {
+    fn from(a: A) -> Self {
+        Authenticator(Arc::new(a))
+    }
+}
+
+pub trait DynAuthenticator: Sync + Send + std::fmt::Debug + 'static {
+    fn request(&self, request: Request) -> Result<Option<Token>>;
+    fn respond(&self, request: Request, token: &Option<Token>) -> Result<AcceptOutcome>;
+}
+
+/// A minimal authenticator that does nothing.
+#[derive(Debug, Clone)]
+pub struct NoAuthenticator;
+
+impl DynAuthenticator for NoAuthenticator {
+    fn request(&self, _request: Request) -> Result<Option<Token>> {
+        Ok(None)
+    }
+
+    fn respond(&self, _request: Request, _token: &Option<Token>) -> Result<AcceptOutcome> {
+        Ok(AcceptOutcome::Accept)
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct Request {
+    pub id: u64,
+    pub data: RequestData,
+}
+
+#[derive(Debug, Clone)]
+pub enum RequestData {
+    Gossip {
+        /// Topic ID (raw because of dependencies)
+        topic: [u8; 32],
+    },
+    Bytes(BytesRequestData),
+    Sync {
+        /// Namespace ID (raw, because of dependencies)
+        namespace: [u8; 32],
+    },
+}
+
+#[derive(Debug, Clone)]
+pub enum BytesRequestData {
+    Get { hash: Hash },
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum AcceptOutcome {
+    Accept,
+    Reject,
+}
+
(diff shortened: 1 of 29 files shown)


iroh-bytes/src/get.rs
```diff
@@ -83,8 +84,12 @@ pub mod fsm {
     }
 
     /// The entry point of the get response machine
-    pub fn start(connection: quinn::Connection, request: GetRequest) -> AtInitial {
-        AtInitial::new(connection, request)
+    pub fn start(
+        connection: quinn::Connection,
+        request: GetRequest,
+        auth: Authenticator,
```
flub, line 90: If you want to be fancy you could request `impl DynAuthenticator` and then implement it for `Option<Authenticator>` or maybe for `Option<dyn DynAuthenticator>` (not entirely sure what works and what doesn't without trying it out).  It would give a more natural API here as a user would probably guess to pass `None` instead of having to find `NoAuthenticator` and importing it.

iroh-base/src/auth.rs
```diff
@@ -0,0 +1,93 @@
...
+use anyhow::Result;
+use serde::{Deserialize, Serialize};
+
+use crate::hash::Hash;
+
+/// The error code sent using quinn when aborting due to authentication errors.
+pub const REJECTED_CODE: u32 = 10;
+
+#[derive(Debug, Clone)]
+pub struct Authenticator(Arc<dyn DynAuthenticator>);
```
rklaehn, line 17: Good to have this dyn by default. This does not warrant a type parameter.

iroh-base/src/auth.rs
```diff
@@ -0,0 +1,93 @@
...
+
+impl<A: DynAuthenticator> From<A> for Authenticator {
+    fn from(a: A) -> Self {
+        Authenticator(Arc::new(a))
+    }
+}
+
+type BoxFuture<T> = Pin<Box<dyn Future<Output = T> + Send + 'static>>;
+
+pub trait DynAuthenticator: Sync + Send + std::fmt::Debug + 'static {
```
rklaehn, line 35: It would be good to have doc comments here to describe the flow. So for each request the request fn is called, which may or may not produce a token. And then response is called with the request and the token from above, and tells us whether to proceed. What is the purpose of the split compared to just having request -> ok/not ok?

iroh-base/src/auth.rs
```diff
@@ -0,0 +1,93 @@
...
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum AcceptOutcome {
+    Accept,
+    Reject,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct Token {
```
Frando, line 89: If we want to be able to evolve this format without breaking the wire format, maybe make it an enum with a single variant for now?

iroh-base/src/auth.rs
```diff
@@ -0,0 +1,93 @@
...
+    fn from(a: A) -> Self {
+        Authenticator(Arc::new(a))
+    }
+}
+
+type BoxFuture<T> = Pin<Box<dyn Future<Output = T> + Send + 'static>>;
+
+pub trait DynAuthenticator: Sync + Send + std::fmt::Debug + 'static {
+    fn request(&self, request: Request) -> BoxFuture<Result<Option<Token>>>;
+    fn respond(&self, request: Request, token: &Option<Token>) -> BoxFuture<Result<AcceptOutcome>>;
```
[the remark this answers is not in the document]
dignifiedquire: yes, exactly, thank you for writing the docs for me :)

iroh-base/src/auth.rs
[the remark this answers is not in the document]
dignifiedquire: I would say lets not add it for now, we can easily add it later down the line if we see the need for it, as it doesn't change the wire format.

iroh-base/src/auth.rs
dignifiedquire: my assumption was that if we need to change the token we need to bump the protocol version anyway

iroh-bytes/src/get.rs
dignifiedquire: this doesn't quite work unfortunately, tried it

iroh-base/src/auth.rs
dignifiedquire: added some docs

iroh-base/src/auth.rs
```diff
@@ -0,0 +1,93 @@
...
+        _request: Request,
+        _token: &Option<Token>,
+    ) -> BoxFuture<Result<AcceptOutcome>> {
+        Box::pin(future::ready(Ok(AcceptOutcome::Accept)))
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct Request {
+    pub id: u64,
```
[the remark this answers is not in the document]
dignifiedquire: added a comment

iroh-base/src/auth.rs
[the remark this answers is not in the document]
dignifiedquire: renamed, let me know what you think of the new names

dignifiedquire:
Needs more design

PR #1959, opened 2024-01-17
feat(iroh): add blobs.create_collection api endpoint

### LINKED ISSUE #1950: Feature Request: New RPC Method: `iroh.blobs.add_collection(hashes: BTreeMap<String, Hash>)`
Working over RPC, there's no way to replicate https://github.com/n0-computer/iroh/blob/eab55bf97acff7df3163240c7dc438520e4874f0/iroh/src/node.rs#L1248-L1279 , because we don't have access to `self.inner.db`

I think it would be nice to have a new method on the `blobs` RPC command:

```rust
iroh.blobs.add_collection(meta: BTreeMap<String,Hash>, tag: ) -> Result<Hash>
```

This would error if any of the hashes in the collection isn't already present in the store (except fro the meta & collection root blob), and return the root hash of the collection

dignifiedquire:
Closes #1950

### DIFF
--- a/iroh/src/rpc_protocol.rs
+++ b/iroh/src/rpc_protocol.rs
@@ -11,7 +11,7 @@ use std::{collections::BTreeMap, net::SocketAddr, path::PathBuf};
 
 use bytes::Bytes;
 use derive_more::{From, TryInto};
-use iroh_bytes::util::Tag;
+use iroh_bytes::{format::collection::Collection, util::Tag};
 pub use iroh_bytes::{get::db::DownloadProgress, BlobFormat, Hash};
 use iroh_net::{
     key::PublicKey,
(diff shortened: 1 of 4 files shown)


iroh/src/rpc_protocol.rs
```diff
@@ -274,6 +274,28 @@ impl RpcMsg<ProviderService> for DeleteTagRequest {
...
+    pub tag: SetTagOption,
+}
+
+/// A response to a create collection request
+#[derive(Debug, Serialize, Deserialize)]
+pub struct CreateCollectionResponse {
+    /// The resulting hash.
+    pub hash: Hash,
+    /// The resulting tag.
+    pub tag: Tag,
```
rklaehn, line 292: 👍

rklaehn (approved):
This works as intended, but I am not sure it is a great API.

The only way to create a collection is to make multiple individual blobs that all have a tag so they don't get immediately deleted, then create a collection out of them. But then you got all these tags that you probably don't want anymore.

Ideally we would have a way to extend the concept of temp tags over the iroh/iroh-client boundary. But maybe that can be done in a subsequent PR.

dignifiedquire:
> Ideally we would have a way to extend the concept of temp tags over the iroh/iroh-client boundary. But maybe that can be done in a subsequent PR.

Yes I agree, but I wanted to have this functionality sooner rather than later, as designing a solution for `TempTag`s over RPC is non trivial.

rklaehn:
> > Ideally we would have a way to extend the concept of temp tags over the iroh/iroh-client boundary. But maybe that can be done in a subsequent PR.
> 
> Yes I agree, but I wanted to have this functionality sooner rather than later, as designing a solution for `TempTag`s over RPC is non trivial.

Understood. People want to use this. But let's revisit it at some point.

Frando:
> Ideally we would have a way to extend the concept of temp tags over the iroh/iroh-client boundary. But maybe that can be done in a subsequent PR.

Yes, this would be great to have I think. Currently the docs RPC API copies things from the blobs api only because otherwise we could not add entries to docs without each having a tag. We could remove this potentially if temp tags could cross the RPC boundary.

PR #1962, opened 2024-01-19
feat(iroh): add more rpc methods

dignifiedquire:
- add `blobs.read_at`: Allows to efficiently read parts of a blob via RPC
- add `blobs.get_collection`: read a collection

### DIFF
--- a/iroh/src/node.rs
+++ b/iroh/src/node.rs
@@ -1503,7 +1503,7 @@ impl<D: BaoStore> RpcHandler<D> {
 
         async fn read_loop<M: Map>(
             offset: u64,
-            len: usize,
+            len: Option<usize>,
             entry: Option<impl MapEntry<M>>,
             tx: flume::Sender<RpcResult<BlobReadAtResponse>>,
             max_chunk_size: usize,
@@ -1516,6 +1516,8 @@ impl<D: BaoStore> RpcHandler<D> {
             }))
             .await?;
             let mut reader = entry.data_reader().await?;
+
+            let len = len.unwrap_or_else(|| (size - offset) as usize);
 
             let (num_chunks, chunk_size) = if len <= max_chunk_size {
                 (1, len)
--- a/iroh/src/node.rs
+++ b/iroh/src/node.rs
@@ -56,14 +56,15 @@ use crate::rpc_protocol::{
     BlobAddPathRequest, BlobAddPathResponse, BlobAddStreamRequest, BlobAddStreamResponse,
     BlobAddStreamUpdate, BlobDeleteBlobRequest, BlobDownloadRequest, BlobListCollectionsRequest,
     BlobListCollectionsResponse, BlobListIncompleteRequest, BlobListIncompleteResponse,
-    BlobListRequest, BlobListResponse, BlobReadRequest, BlobReadResponse, BlobValidateRequest,
-    CreateCollectionRequest, CreateCollectionResponse, DeleteTagRequest, DocExportFileRequest,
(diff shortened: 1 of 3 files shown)


Frando:
Cool I thought to add this quite a while ago already. Is there a reason to keep both `BlobRead` and `BlobReadAt` rpcs? Might be simpler to just have a single `BlobRead` RPC with `offset: Option<usize>, limit: Option<usize>`?

iroh/src/node.rs
```diff
@@ -1480,6 +1481,77 @@ impl<D: BaoStore> RpcHandler<D> {
...
+                is_complete: entry.is_complete(),
+            }))
+            .await?;
+            let mut reader = entry.data_reader().await?;
+
+            let (num_chunks, chunk_size) = if len <= max_chunk_size {
+                (1, len)
+            } else {
+                (
+                    (len as f64 / max_chunk_size as f64).ceil() as usize,
```
rklaehn, line 1526: weird to use f64 here... It is shorter than the integer variant, but will fail if len or max_chunk_size is > 2^53.

iroh/src/node.rs
```diff
@@ -1480,6 +1481,79 @@ impl<D: BaoStore> RpcHandler<D> {
...
+                tx.send_async(RpcResult::Err(err.into())).await.ok();
+            }
+        });
+
+        async fn read_loop<M: Map>(
+            offset: u64,
+            len: Option<usize>,
+            entry: Option<impl MapEntry<M>>,
+            tx: flume::Sender<RpcResult<BlobReadAtResponse>>,
+            max_chunk_size: usize,
```
rklaehn, line 1509: Do we need the max_chunk_size? Makes the code more complex.

At least in the AsyncSliceReader the idea is that you would only ask for sizes that you are fine with handling as a single Bytes. That is why the len parameter is `usize`, not `u64`.

iroh/src/node.rs
dignifiedquire: how would I round up otherwise?

iroh/src/node.rs
dignifiedquire: it's needed to limit the chunk size sent over the wire , like in the regular read

iroh/src/node.rs
rklaehn:
```rust
len / max_chunk_size + if len % max_chunk_size == 0 { 0 } else { 1 }
```

or

```rust
len / max_chunk_size + (len % max_chunk_size !=0) as usize
```

or something. With a bit of luck the optimizer will do only 1 integer division for both division and rest.

iroh/src/node.rs
rklaehn: I don't quite get it. The read_at API returns a `Bytes`, so if you call it with some large size, you will at one point get a Bytes containing all the data in memory. So what's the benefit of this chunking for the purpose of implementing AsyncReadAt?

dignifiedquire: the rpc `read_at` api does not return the full `Bytes`, but rather streams it, so no we will not get the full data in memory with the current strategy

iroh/src/node.rs
rklaehn: Note that I don't worry about speed here. It's just weird to use fp for integer ops, and it will fail in weird ways as soon as you exceed the range of integers that can be losslessly represented in a fp num.

dignifiedquire: ah yes, that is of course nice

dignifiedquire:
> Might be simpler to just have a single `BlobRead` RPC wit

done

rklaehn (approved):
avoid the floats!

PR #1967, opened 2024-01-22
example(iroh-net): minimal use of unreliable datagram

dignifiedquire:
Show a minimal test case of using the unreliable datagram extension of quic with iroh-net


flub (approved):
I think @ramfox and me were thinking of having fewer cli options and instead having examples with as little customisation as possible.  In that case this would be split off into a separate example and repeated code just be repeated.  But also, we can do this splitting off separately, this example has not yet been looked at.  So we can merge this for now.

Issue #1929, opened 2024-01-05
audit our examples


b5:
a user on [our discord](https://discord.com/channels/1161119546170687619/1161119546644627528/1192883199555149895) has pointed out that our examples need a cleanup to match recent changes. We should do an audit of examples & make sure they're up to date by confirming each example works as described.

* [x] `/iroh/examples/client.rs`
* [x] `/iroh/examples/collection.rs`
* [ ] `/iroh/examples/dump-blob-fsm.rs`
* [x] `/iroh/examples/hello-world.rs`
* [ ] `/iroh/examples/stream.rs`
* [ ] `/iroh-net/examples/magic.rs`
* [ ] `/iroh-gossip/examples/chat.rs`

s2dd:
/iroh/iroh/examples/collections has hardcoded 
`
// print the ticket, containing all the above information
    println!("in another terminal, run:");
    println!("\t$ cargo run -- get --ticket {}", ticket);
    // wait for the node to finish, this will block indefinitely
    // stop with SIGINT (ctrl+c)
    node.await?;
`

where i today need use `console` and there issue  `blob get <ticket>` produced from example.

s2dd:
Stated in the README.md
"At the top of each example file is a comment describing how to run the example."
more then often outdated ...

ramfox:
for `/iroh-net/examples/magic.rs` specifically:
I believe when a user looks at an example called "magic" they believe that it's meant to magically connect via just `node_id`, but I'm pretty sure the intention here is to just show how to use `MagicEndpoint`

we should:
- [ ] write docs that talk about the intention of the example
- [ ] mention that if you supply a direct addr only on `connect`, you may not be able to connect if you require any hole-punching

for the listen command:
- [ ] print out all of our known addrs
- [ ] print out all of our derp urls
- [ ] print out what command should be run in another terminal to `connect` correctly
- [ ] allow overrides for `bind-port` and `derp-url` are "global", but should really be moved to `listen` (since the `connect` command has a `derp-url` parameter that isn't global as well)

for the connect command:
- [ ] don't allow the command to run if you don't have one of `addr` or `derp_url`

However, if we *are* supposed to show that we can magically connect via just node id, we need to add default `derp_urls` to the `connect` command.

s2dd:
if no `derp_url` use `default_derp_url`

besides node-id , addr , maybe a ticket with `derp_url`embeded = calling-card

ramfox:
[...]

PR #1968, opened 2024-01-23
fix(examples): Adjust and add examples

ramfox:
## Description

Fixes incorrect information in our `iroh/examples`

Also adds an example that shows the most basic way to use iroh to fetch content.

related to #1929 

## Notes & open questions

Unfortunately, we don't seem to have very simple APIs for fetching (that isn't just exporting the content to an external path). Working through this.

Also adds:
- `BlobDownloadProgress` - this takes a stream of `DownloadProgress` and puts it into a form that will poll the stream by calling `finish`
- adjusts the `blob_download0` function to not eat certain errors

## Change checklist

- [x] Self-review.


iroh/examples/collection.rs
```diff
@@ -58,7 +58,7 @@ async fn main() -> anyhow::Result<()> {
     }
     // print the ticket, containing all the above information
     println!("in another terminal, run:");
-    println!("\t$ cargo run -- get --ticket {}", ticket);
+    println!("\t$ cargo run -- blob get {} --start", ticket);
```
dignifiedquire, line 61: this is missing the `--example collection`

iroh/examples/fetch.rs
```diff
@@ -0,0 +1,105 @@
...
+        .with(tracing_subscriber::fmt::layer().with_writer(std::io::stderr))
+        .with(EnvFilter::from_default_env())
+        .try_init()
+        .ok();
+}
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    setup_logging();
+    println!("ran example");
```
dignifiedquire, line 31: stray println?

iroh/examples/fetch.rs
```diff
@@ -0,0 +1,105 @@
...
+    if args.len() != 2 {
+        bail!("expected one argument [BLOB_TICKET], get a ticket by running `cargo run --example hello_world` in a separate terminal.");
+    }
+
+    // deserialize ticket string into a ticket
+    let ticket =
+        iroh::ticket::BlobTicket::from_str(&args[1]).context("failed parsing blob ticket")?;
+
+    // create a new, empty in memory database
+    let db = iroh_bytes::store::readonly_mem::Store::default();
```
dignifiedquire, line 44: shouldn't be readonly probably

iroh/examples/hello-world.rs
```diff
@@ -45,7 +45,7 @@ async fn main() -> anyhow::Result<()> {
     }
     // print the ticket, containing all the above information
     println!("in another terminal, run:");
-    println!("\t$ cargo run -- get --ticket {}", ticket);
+    println!("\t$ cargo run -- blob get {} --start", ticket);
```
dignifiedquire, line 48: missing `--example`

iroh/src/node.rs
```diff
@@ -1135,7 +1135,15 @@ impl<D: BaoStore> RpcHandler<D> {
...
                 progress.clone(),
             )
-            .await?;
+            .await
+            {
+                Ok(stats) => stats,
+                Err(e) => {
+                    let str_err = e.to_string();
+                    progress.send(DownloadProgress::Abort(e.into())).await?;
+                    bail!("{str_err}");
```
dignifiedquire, line 1144: same as below

iroh/src/node.rs
```diff
@@ -1165,10 +1173,14 @@ impl<D: BaoStore> RpcHandler<D> {
...
-                    db.set_tag(tag, Some(haf)).await?;
+                    if let Err(err) = db.set_tag(tag, Some(haf)).await {
+                        progress.send(DownloadProgress::Abort(err.into())).await?;
+                    }
                 }
                 SetTagOption::Auto => {
-                    db.create_tag(haf).await?;
+                    if let Err(err) = db.create_tag(haf).await {
+                        progress.send(DownloadProgress::Abort(err.into())).await?;
+                    }
```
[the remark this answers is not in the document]
ramfox: fixed this and adjusted the error handling in this section in general.

We never await the handle that this task is running on, so there is nothing that catches those errors. So there is no reason to return them, we now just send the error on the progress channel, and log any errors from the `send`.

iroh/src/node.rs
```diff
@@ -1165,15 +1182,24 @@ impl<D: BaoStore> RpcHandler<D> {
...
+                        }
+                        return;
+                    }
                 }
             }
             drop(temp_pin);
-            progress.send(DownloadProgress::AllDone).await?;
-            anyhow::Ok(())
+            let _ = progress.send(DownloadProgress::AllDone).await;
         });
```
dignifiedquire, line 1203: we should also return the errors from the closure and check for errors here, currently this function will exit with `Ok` even though it failed

ramfox: I figured that we weren't `await`ing the closure because the intended use case was to return once the download process was set up and expect users to listen for errors on the progress stream.

I refactored so that this was closer to the other methods that rely on progress streams for communciation.

dignifiedquire (approved):
lgtm

PR #1969, opened 2024-01-23
fix(iroh): do not establish connection if content already exists locally

### LINKED ISSUE #1947: `blob_download` unnecessarily tries to connect to a peer even if the requested blob is found locally.
I ran into this issue when testing the RPC interface, by taking a `BlobTicket`, deconstructing it, constructing a `BlobDownloadRequest` and calling  `BlobsClient::download`. It's trivial to work around, simply call `BlobsClient::read` or `BlobsClient::read_to_bytes` first and use that result if you don't get an error. Perhaps this behaviour is intended, but it's a little unintuitive.

dignifiedquire:
Avoid establishing a connection in `blobs.download` if content is already locally available.

Closes #1947

### DIFF
--- a/iroh-bytes/src/get/db.rs
+++ b/iroh-bytes/src/get/db.rs
@@ -3,6 +3,7 @@ use std::path::PathBuf;
 use std::time::Duration;
 
 use bytes::Bytes;
+use futures::Future;
 use futures::{future::LocalBoxFuture, FutureExt, StreamExt};
 use iroh_base::{hash::Hash, rpc::RpcError};
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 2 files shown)


flub (approved):

dignifiedquire:
@Arqu @flub any ideas why this breaks netsim?

Arqu:
Netsim output says it timed out & the logs don't really yell about anything. I guess it hangs for some reason?

TLDR; just tested this out manually. Transfer hangs. 
```
tty1
export IROH_DATA_DIR=/foo/blah-1
./target/release/iroh start --add ./target/release/iroh
# copy ticket
tty2
export IROH_DATA_DIR=/foo/blah-2
./target/release/iroh blob get --start blobticket --out STDOUT > /dev/null
```
This just hangs infinitely 
Screenshot 2024-01-23 at 12 43 25

rklaehn (approved):

iroh-bytes/src/get/db.rs
```diff
@@ -33,26 +34,34 @@ use tracing::trace;
 ///
 /// This considers data that is already in the store, and will only request
 /// the remaining data.
-pub async fn get_to_db<D: BaoStore>(
+pub async fn get_to_db<
+    D: BaoStore,
+    C: FnOnce() -> F,
```
rklaehn, line 39: :+1: for FnOnce here - you wouldn't want to spam tons of connections...

Issue #1966, opened 2024-01-22
Flaky test: test_two_devices_roundtrip_quinn_magic


flub:
These tests are sometimes very slow on windows, to the point they get cancelled:

- `iroh_net::magicsock::tests::test_two_devices_roundtrip_quinn_magic`
- `iroh_net::magicsock::tests::test_two_devices_roundtrip_network_change`
- `iroh_net::magicsock::tests::test_two_devices_setup_teardown`

`iroh_net::magic_endpoint::tests::magic_endpoint_derp_connect_loop`
is also slow and sometimes fails on windows.  Possibly related as well.

flub:
Split off #2049 and #2050.  The rest here was fixed in #2034

PR #1971, opened 2024-01-23
chore(ci): Run tests from the PR branch, not main

flub:
## Description

This fixes the tests to run from the PR branch and not from main.

Additionally it splits the tests running into three steps:

- Building the tests.
- Listing ignored tests.
- Running tests.

This makes the output easier to consume and does not add any slowdown.
Listing ignored tests is useful since nextest only says how many tests
are ignored rather than listing them all.

Finally it fixes the features testing by moving the CRATES_LIST environment
variable to the right workflow.

## Notes & open questions

Flaky tests tracked in #1966 

## Change checklist

- [x] Self-review.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -16,7 +16,7 @@ on:
       git-ref:
         description: 'Which git ref to checkout'
         type: string
-        default: 'main'
+        default: ${{ github.ref }}
 
 env:
   RUST_BACKTRACE: 1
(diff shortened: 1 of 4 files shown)


Arqu (approved):
I mean LGTM, also cool moving the crates list into tests. I guess they don't propagate eiter?

flub:
> I mean LGTM, also cool moving the crates list into tests. I guess they don't propagate eiter?

yeah, they don't propagate.  i considered seeing if it could be generated from a cargo command but decided to leave it as it already was.

.github/workflows/tests.yaml
```diff
@@ -16,13 +16,14 @@ on:
       git-ref:
         description: 'Which git ref to checkout'
         type: string
-        default: 'main'
+        default: ${{ github.ref }}
```
szepeviktor, line 19: `github.ref` is "refs/heads/main"
`github.ref_name` is "main"

[_source_](https://github.com/szepeviktor/byte-level-care/actions/runs/7576269316/job/20634685938)

flub: yes, but is there a reason to prefer `main` over `refs/heads/main`?  the latter seems more specific to me

PR #1972, opened 2024-01-23
Add better debug instance for AddrInfo

rklaehn:
## Description

Adds a better debug output for AddrInfo.

Before:
AddrInfo { derp_url: Some(Url { scheme: \"https\", cannot_be_a_base: false, username: \"\", password: None, host: Some(Domain(\"derp.example.com\")), port: None, path: \"/\", query: None, fragment: None }), direct_addresses: {1.2.3.4:1234} }

After:

AddrInfo { derp_url: Some(https://derp.example.com/), direct_addresses: {1.2.3.4:1234} }

## Notes & open questions

- Is there a way to make derive_more do this?
- should we make a newtype for DerpUrl that does this and also arc?

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/src/magic_endpoint.rs
+++ b/iroh-net/src/magic_endpoint.rs
@@ -103,7 +103,6 @@ impl<T: std::fmt::Display> std::fmt::Debug for DD<T> {
     }
 }
 
-
 impl AddrInfo {
     /// Return whether this addressing information is empty.
     pub fn is_empty(&self) -> bool {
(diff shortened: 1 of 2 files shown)


iroh-net/src/magic_endpoint.rs
```diff
@@ -78,14 +78,31 @@ impl From<(PublicKey, Option<Url>, &[SocketAddr])> for NodeAddr {
...
+impl Debug for AddrInfo {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("AddrInfo")
+            .field("derp_url", &self.derp_url.as_ref().map(DD))
+            .field("direct_addresses", &self.direct_addresses)
+            .finish()
+    }
+}
+
+struct DD<T: std::fmt::Display>(T);
```
flub, line 98: not really a need for a new type here, you can directly format the field instead (we already do this somewhere else as well, maybe somewhere in netcheck, because i recall doing exactly the same to a URL before)

PR #1979, opened 2024-01-25
fix(iroh): improve and test blob share

### LINKED ISSUE #1978: `iroh blob share` is broken
on main as of 2024-01-25 `iroh blob share HASH` returns 
```
Error: failed to retrieve blob info

Caused by:
    Expected header frame
```

dignifiedquire:
- adds `iroh.blobs.share`
- adds `iroh.blobs.status`
 
Closes #1978

### DIFF
--- a/iroh/src/client.rs
+++ b/iroh/src/client.rs
@@ -1706,7 +1706,7 @@ mod tests {
         assert_eq!(ticket.hash(), import_outcome.hash);
 
         let status = client.blobs.status(import_outcome.hash).await?;
-        assert_eq!(status.is_complete, true);
+        assert!(status.is_complete);
         assert_eq!(status.size, size);
 
         Ok(())
--- a/iroh/src/client.rs
+++ b/iroh/src/client.rs
@@ -21,6 +21,7 @@ use iroh_bytes::store::ValidateProgress;
 // use iroh_bytes::util::progress::FlumeProgressSender;
 use iroh_bytes::Hash;
 use iroh_bytes::{BlobFormat, Tag};
+use iroh_net::ticket::BlobTicket;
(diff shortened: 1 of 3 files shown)


iroh/src/client.rs
```diff
@@ -433,6 +434,51 @@ where
...
+        Ok(BlobStatus {
+            size: reader.size,
+            is_complete: reader.is_complete,
+        })
+    }
+}
+
+/// Status information about a blob.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct BlobStatus {
```
rklaehn, line 477: Would also prefer an enum here. E.g. "partial" is a pretty wide range from 1/1TiB to 999GiB/1TiB. We might want to add some more info here.

dignifiedquire: but that information is not known, the store doesn't expose this easily

rklaehn: Yes, I know. And it is not free to compute. What I'm saying is that with an enum it is easier to evolve it as more information becomes available.

dignifiedquire: done

iroh/src/client.rs
```diff
@@ -433,6 +434,51 @@ where
         self.rpc.rpc(BlobDeleteBlobRequest { hash }).await??;
         Ok(())
     }
+
+    /// Share a blob.
+    pub async fn share(
+        &self,
+        hash: Hash,
+        blob_format: BlobFormat,
+        no_derp: bool,
```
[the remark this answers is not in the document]
dignifiedquire: done

rklaehn (approved):

PR #1981, opened 2024-01-27
move some examples down to `iroh-bytes`

### LINKED ISSUE #1929: audit our examples

ramfox:
## Description

closes #1929 

### `provide-bytes`
This example adds either a single blob or collection to the database and provides it. It gives instructions on how to run the fetch side of the example for `fetch-fsm` and `fetch-stream`.

We could make this example more "advanced" by having the user provide a path & serve the blob or collection of that content, but I wanted to keep its simple at first.

### `fetch-fsm`
This example fetches a blob or collection using the `get` finite state machine directly, printing the contents to the terminal.

### `fetch-stream`
This example fetches a blob or collection using a helper function to turn the `get` finite state machine into a stream. It prints the contents to the terminal.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.


iroh-base/src/hash.rs
```diff
@@ -360,6 +360,21 @@ mod tests {
...
+        let hash1 = Hash::new("hello");
+        let hash2 = Hash::new("world");
+        let mut ser = postcard::to_stdvec(&hash1).unwrap();
+        let mut ser2 = postcard::to_stdvec(&hash2).unwrap();
+        ser.append(&mut ser2);
+        let hashes: Box<[Hash]> = postcard::from_bytes(&ser).unwrap();
+        assert_eq!(hash1, hashes[1]);
+        assert_eq!(hash2, hashes[2]);
+    }
+
```
ramfox, line 377: `let hashes: Box<[Hash]> = postcard::from_bytes(&ser).unwrap();` was taken from the previous `dump-blob-stream` example code, but it always failed when I wrote the `fetch-fsm` and `fetch-stream` examples. 

Should we be expecting this to work?

iroh-bytes/examples/provide-bytes.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+    let format = {
+        if args[1] != "blob" && args[1] != "collection" {
+            anyhow::bail!(
+                "expected either 'blob' or 'collection' for FORMAT argument, got {}",
+                args[1]
+            );
+        }
+        args[1].clone()
+    };
```
ramfox, line 46: I split "blob" and "collection" up to make the instructions easier for the user when running the example, but I'm not sure this is the best way to do it.

Could possibly split this into `provide-blob` and `provide-collection`.

iroh-base/src/hash.rs
dignifiedquire: cc @rklaehn

rklaehn: How can this ever have worked? A Box<[Hash]> would be a varint encoded "2" and then 2 hashes, not just 2 concatenated hashes. I have no idea what this was supposed to test, so I would just drop it.

rklaehn (approved):
Looks reasonable.

This stuff is pretty low level so it does not really belong in iroh.

iroh-bytes/examples/provide-bytes.rs
rklaehn: I think it is fine.

PR #1982, opened 2024-01-29
refactor(iroh-net): A bunch of logging improvements

flub:
- Make sure the DerpMap has a nicer Debug impl that shows the map with URLs instead of structs.

- Do not log individual packets on debug level, downgrade those to trace.

- Log derp URLs and URLs rather than structs in various places.

- Log sending of DISCO messages on debug, these are generally interesting and do not transfer bulk data.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


dignifiedquire (approved):

flub:
ok, fine.  let's just merge this if it's already approved.  i was thinking of collecting more things but whatever

PR #1983, opened 2024-01-29
fix(iroh-net): improve hole punching when relying on call-me-maybe

dignifiedquire:
- only consider active endpoints for call-me-maybe checks
- insert call me maybe addrs into the peer map


iroh-net/src/magic_endpoint.rs
```diff
@@ -673,7 +672,8 @@ mod tests {
                     let mut buf = [0u8, 5];
                     stream.read_exact(&mut buf).await.unwrap();
                     info!("Accepted 1 stream, received {buf:?}.  Closing now.");
-                    ep.close(7u8.into(), b"bye").await.unwrap();
+                    // close the stream
+                    conn.close(7u8.into(), b"bye");
```
flub, line 676: by not awaiting this you're never closing this?

dignifiedquire: there is no way to actually wait for a connection close asfaik

iroh-net/src/magicsock/peer_map/endpoint.rs
```diff
@@ -438,7 +438,7 @@ impl Endpoint {
             })
             .collect();
         let ping_needed = !pings.is_empty();
-        let have_endpoints = !self.direct_addr_state.is_empty();
+        let have_alive_endpoints = self.direct_addr_state.values().any(|e| e.is_alive());
```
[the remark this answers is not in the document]
dignifiedquire: why?

flub (approved):

Issue #1935, opened 2024-01-08
cli and provide tests are largely flaky


flub:
There are a whole bunch of flaky integration tests in the following files:

- `iroh/tests/cli.rs`
- `iroh/tests/provide.rs`

I suspect all of these are related to filesystem latencies.  We probably need a reliable way to run `sync(1)`/`fsync(2)` on the filesystem to fix all of these.

flub:
https://github.com/n0-computer/iroh/actions/runs/7829002247/job/21360042881?pr=2009

flub:
Maybe we should just sprinkle shelling out to the `sync(1)` command around to see if this helps quickly.  As ugly as it may be, we'd get these tests back.  If there's an equivalent for this on Windows.

matheus23:
> https://github.com/n0-computer/iroh/actions/runs/7829002247/job/21360042881?pr=2009

The next time we should download the logs & upload them here.
They've expired since, and they'd be really valuable actually.

divagant-martian:
most of these tests are no longer flaky, and grouping them all together gives us a poor view of our understanding of what's wrong with each. That being said, I'll close this and open new issue for the two remaining cli flaky tests

PR #1984, opened 2024-01-31
fix(iroh-net): Improve direct connectivity establishment speed and reliablity

### LINKED ISSUE #1955: direct connection not chosen for dumbpipe
I’m trying out iroh-net by means of dumbpipe (use case: `git fast-export | dumbpipe connect ...` plus `dumbpipe listen | git fast-import`). While data are flowing nicely over the derper, I can’t seem to get a direct connection going — and I also see no need to use your precious bandwidth ;-)

full log (of the startup) at https://gist.github.com/rkuhn/1170c5e27ace2aac71168d9d1ad47c52

I have fixed the magic port to 5858 (IPv4), which implies 5859 (IPv6). Both ports are open in the firewall where `dumbpipe listen` is running, and that host has a public IP (Hetzner). The log is from the other end — unfortunately it isn’t possible to log on the other end without trashing stdout (at least `git fast-import` crashes then).

It seems that something is amiss with liveness checking:

```
2024-01-16T12:45:45.035836Z DEBUG magicsock{me=fewl2qhaymncv6ht}:actor:disco{node=ebosiergt6xnkiyc}: iroh_net::magicsock::peer_map::endpoint: send stayin alive ping dst=[2a01:4f9:4a:5150::2]:5859 since_last_ping=Some(5.0007875s)
2024-01-16T12:45:45.035910Z DEBUG magicsock{me=fewl2qhaymncv6ht}:actor:disco{node=ebosiergt6xnkiyc}: iroh_net::magicsock::peer_map::endpoint: start ping tx=563c0a30c68e1d8a8a5057e0 dst=UDP([2a01:4f9:4a:5150::2]:5859) purpose=StayinAlive
2024-01-16T12:45:45.036199Z DEBUG magicsock{me=fewl2qhaymncv6ht}:actor:handle_ping_action: iroh_net::magicsock: ping sent (polled) dst=Udp([2a01:4f9:4a:5150::2]:5859) tx=563c0a30c68e1d8a8a5057e0 purpose=StayinAlive
2024-01-16T12:45:45.036963Z DEBUG magicsock{me=fewl2qhaymncv6ht}:actor:disco{node=ebosiergt6xnkiyc}: iroh_net::magicsock::peer_map::endpoint: pong not received in timeout tx=8d7389efcf4ff350d75d4368 addr=UDP(172.17.0.1:5858)
```

flub:
## Description

This includes various fixes and improvements in holepunching behaviour,
improving both how fast a direct connection is established and how
reliably it is established.

- Improve sending of call-me-maybe: this sometimes wasn't sent if
  there were no known addresses for an endpoint because no pings
  were sent and there was an active connection on via the derp
  path.  Instead we now send it when needed, regardless of pings
  or active connections.  To avoid sending it all the time when
  no direct connection can be established we only send it once
  every heartbeat cycle, i.e. every 5s.  In the worst case (like
  when UDP pings get lost) this means a call-me-maybe will be sent
  after just under 2 heartbeat intervals, and should be regular
  every heartbeat interval afterwards.  So in the face of packet
  loss this will still establish a connection eventually.

- Not every scenario will result in a direct connection in both
  directions after each endpoint has sent one call-me-maybe
  request, because direct connections are asymmetric and each
  side has to receive their own disco pong before it will use
  the direct connection.  To alleviate this and speed up the
  direct connection on the return path in cases were only one
  side manages to establish the direct path, this now also sends
  a ping message upon receipt of a ping.  This opens the direct
  path much quicker in both directions.

- When queuing a call-me-maybe and the last netcheck report is
  deemed too old, the call-me-maybe is not immediately sent.
  Instead a new netcheck is kicked off and the queued
  call-me-maybes should be sent once the report is in, so that
  they are sent with accurate local addresses.  However they
  were never sent after the report finished.  This is now fixed
  and they are sent after the report is ready.

- When a ping timeout occurs, we were clearing the best_addr, i.e.
  the selected direct path, if the ping was for the same path as
  currently used.  However that could have been for a timeout of
  an earlier ping that was still stopped by a firewall or NAT.  It
  happens that a later ping is sent and receives a pong, before the
  timeout of that earlier ping.  In that case we should not clear
  the direct path in use.

- We used to prune addresses even if they were only just received from a
  call-me-maybe.  This did not allow us to ping all the addresses from a
  call-me-maybe breaking hole punching.

  We take two measures:

  - Do not prune messages that were recently "alive", which means they
    received a ping, pong or call-me-maybe recently.

  - We increase the number of allowed inactive addresses.  They used to
    be 100.  If someone switches between several networks regularly, we
    would like to still know those previous networks as they might start
    working again.

- When looking for a destination address for an endpoint, the
  addr_for_send() function uses un-confirmed UDP address if available.
  However if we were given an IPv6 address (probably via a ticket,
  effectively via add_node_addr()) we should only select the IPv6
  address if netcheck repored we have working IPv6 ourself.

- The netcheck report was wrongly overwriting whether an IP
  family was usable on dual stack hosts.  Resulting in the magic
  socket being confused about which connectivity it had.

This also begins to clean up some terminology: "endpoint" is used for
many different things.  Here the term "path" is consistently used
when referring to a specific destination address of a node, whenter
over the derper or a direct UDP socket.  This naming cleanup however
is not complete as that would obscure the fixes too much.  More to
be done later.

Closes #1955 

## Notes & open questions

This also changes the first address of our QuicMappedAddresses to
start at 1.  Not that 0 is reserved, but it is rather odd for people
who are used to IPv4 to see 0 being a real node.  And it's not like
you're short of addresses in IPv6.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh-net/src/magicsock/peer_map/endpoint.rs
```diff
@@ -437,22 +453,35 @@ impl Endpoint {
...
-        if send_call_me_maybe && (ping_needed || !have_endpoints) {
+        // If we were asked for a call-me-maybe we should send it if we don't have a best
+        // addr.  We may not be sending any pings, because we just received a call-me-maybe
+        // and already send pings just now, which remains valid for 5 seconds.
+        //
+        // TODO: double check if there really is any point in still checking beyond
+        // send_call_me_maybe, it posssibly was already checked at the call site that we
+        // don't have a best addr.
+        //
+        // TODO: this could be problematic if we can not punch: we keep sending out
```
dignifiedquire, line 480: we should check when we sent the last call-me-maybe probably, and use that as a hint if we need to send one

iroh-net/src/magicsock/peer_map/endpoint.rs
```diff
@@ -199,68 +199,86 @@ impl Endpoint {
...
             }
         }
     }
 
     /// Update our best_addr (if empty) with the candidate udp addr with the lowest latency.
     fn assign_best_addr_from_candidates_if_empty(&mut self) {
+        // TODO(flub): I'm unsure how this function can be responsible to ever fill in the
+        // best_path.  We'd have to have a path that has a known latency, yet no pong should
+        // have been received on that path as otherwise the pong handler would have already
+        // selected the path.
```
dignifiedquire, line 248: the point of this function is to select the best guess in the face of missing (potentially no latency) information

iroh-net/src/magicsock/peer_map/endpoint.rs
```diff
@@ -424,38 +456,47 @@ impl Endpoint {
...
+        //
+        // TODO: this could be problematic if we can not punch: we keep sending out
+        // call-me-maybe messages.
+
+        // The if condition here is really trying to be: send call me maybe if we have just
+        // sent some pings, or did send some pings recently.  However maybe we also need to
+        // send it if we don't have a BestADdr yet.
+
+        // if send_call_me_maybe && (have_ping_msgs || !have_alive_endpoints) {
+        if send_call_me_maybe {
```
dignifiedquire, line 499: so this only relies on the caller to determine this now?

iroh-net/src/netcheck.rs
```diff
@@ -686,64 +682,6 @@ impl Actor {
 
         r
     }
-
-    fn log_concise_report(&self, r: &Report, dm: &DerpMap) {
```
dignifiedquire, line 690: why no nice report log anymore?

flub:
Currently there are two main things left:

- Call-me-maybe messages are sent too often.  Resulting in too many pings.  The issue is that they need to be sent whenever we have no info about the endpoint, because otherwise we can never learn about the other endpoint.  I'm thinking of adding an exponential backoff for them: initially send call-me-maybe every 0.5-1 second, then as connections do not get established slow it down up to every 30ish seconds.

- Old ping timeouts reset the best-addr state.  This is kind of papered over with the `assign_best_addr_from_candidates_if_empty` function, but I think it should not be cleared in the first place.

iroh-net/src/netcheck.rs
flub: I think this was mostly a go thing, this was essentially a manual implementation of Rust's `Debug`.  So I switched it to `Debug`, we get the same information equally clearly.

dignifiedquire: ok

dignifiedquire:
> * I'm thinking of adding an exponential backoff for them

sounds good

iroh-net/src/magicsock/peer_map/endpoint.rs
flub: Yes, the logic of whether pings are being sent is wrong.  We need to send a call-me-maybe even if we don't know any UDP addresses to ping.  The alive endpoints (paths really) don't help much either, a path is considered alive too quickly: if it received a ping it is alive.  But an endpoint can only ever be used to send if it received a pong (at which point it would be in best_addr, so that's already checked by the callers as they wouldn't be asking for this if they had one)

iroh-net/src/magicsock/peer_map/endpoint.rs
flub: Yep, I think that's the approach i want to take.  see the top-level comment on the pr

iroh-net/src/magicsock/peer_map/endpoint.rs
flub: yes, but this should not happen.  When we receive a pong it should be promoted to best_addr.

However I have seen this happening: when pasts pings timeout they clear the best_addr state.  Even though there was a more recent successful ping.  That bug needs to be fixed, but this papered over it.

I don't think I'll remove this though, that's a bit risky probably as this is a good fallback.

flub:
> > * I'm thinking of adding an exponential backoff for them
> 
> sounds good

ok, glad you seem to agree because i think this is the main thing to make call-me-maybe interactions work well and the biggest change.

flub:
> > * Old ping timeouts reset the best-addr state.
> 
> they should only clear if no newer information has been attained since they were sent

yep, and that's a bug that needs to be fixed.

iroh-net/src/netcheck/reportgen.rs
```diff
@@ -366,9 +366,9 @@ impl Actor {
                 Probe::Https { .. } | Probe::Icmp { .. } => (),
             }
         }
-        self.report.ipv4_can_send = probe_report.ipv4_can_send;
-        self.report.ipv6_can_send = probe_report.ipv6_can_send;
-        self.report.icmpv4 = probe_report.icmpv4;
+        self.report.ipv4_can_send |= probe_report.ipv4_can_send;
```
dignifiedquire, line 369: I think this is wrong, if our report says we cant send anymore we shouldn’t keep this true

flub: Here the `probe_report` is of a single probe: the `Ipv6Stun` probe will always report `ipv4_can_send` to be false, just like the `Ipv4Stun` probe will always report `ipv6_can_send` to be false.  As we assemble the entire netcheck report we need to combine these results from individual probes together, without trampling over each other's results.  So I believe this is correct.

dignifiedquire: ahh okay, you are right

iroh-net/src/magicsock/peer_map/endpoint.rs
```diff
@@ -199,68 +223,93 @@ impl Endpoint {
...
             }
         }
     }
 
-    /// Update our best_addr (if empty) with the candidate udp addr with the lowest latency.
+    /// Fixup best_addr from candidates.
+    ///
+    /// If somehow we end up in a state where we failed to set a best_addr, while we do have
+    /// valid candidates, this will chose a candidate and set best_addr again.  Most likely
+    /// this is a bug elsewhere though.
```
flub, line 279: all changes in this function are a re-write for readability, should not contain functional changes

iroh-net/src/magicsock.rs
```diff
@@ -2484,7 +2511,7 @@ impl Iterator for PacketSplitIter {
 pub(crate) struct QuicMappedAddr(SocketAddr);
 
 /// Counter to always generate unique addresses for [`QuicMappedAddr`].
-static ADDR_COUNTER: AtomicU64 = AtomicU64::new(0);
+static ADDR_COUNTER: AtomicU64 = AtomicU64::new(1);
```
dignifiedquire, line 2514: why?

iroh-net/src/magicsock/peer_map/endpoint.rs
```diff
@@ -199,68 +226,92 @@ impl Endpoint {
...
-        for (ipp, state) in self.direct_addr_state.iter() {
-            if let Some(pong) = state.recent_pong() {
-                // Lower latency, or when equal, prefer IPv6.
-                if pong.latency < lowest_latency
-                    || (pong.latency == lowest_latency && ipp.ip().is_ipv6())
-                {
-                    lowest_latency = pong.latency;
-                    last_pong.replace(pong);
+
+        // The lowest acceptable latency for an endpoint path.  If the latency is higher
```
dignifiedquire, line 287: the hightest?

iroh/tests/cli.rs
```diff
@@ -478,6 +478,7 @@ fn run_cli(
 
 #[cfg(feature = "cli")]
 #[test]
+#[ignore = "flaky"]
```
dignifiedquire, line 481: :( do we have an issue for this?

dignifiedquire (approved):

iroh-net/src/magicsock.rs
dignifiedquire: nvm, found your comment, fine by me

iroh/tests/cli.rs
flub: #1935 

All these tests are fundamentally not reliable sadly.

PR #1985, opened 2024-01-31
redb store - inline blobs and outboards

rklaehn:
## Description

Inline small blobs and small outboards in the redb.

## Notes & open questions

- What's the best limit for when to inline outboards?
- Separate table for the blobs, or put them into the partial and complete table
  - currently my gut feeling is that mixing large items like inline data and small items like entries will lead to bad performance for entries queries. 

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


rklaehn:
This is superseded by bao-file

PR #1988, opened 2024-01-31
fix: ci release builds

Arqu:
## Description

After the initial merge of https://github.com/n0-computer/iroh/pull/1987 made some fixes.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -83,7 +83,7 @@ jobs:
         - name: attach artifacts
           uses: actions/upload-artifact@v4
           with:
-            name: iroh-${RELEASE_OS}-${RELEASE_ARCH}-${GITHUB_SHA::7}
+            name: iroh-${{env.RELEASE_OS }}--${{env.RELEASE_ARCH}}-${{GITHUB_SHA::7}}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -58,7 +58,7 @@ jobs:
         env:
           # Using self-hosted runners so use local cache for sccache and
           # not SCCACHE_GHA_ENABLED.
-          RUSTC_WRAPPER: "sccache"
+          # RUSTC_WRAPPER: "sccache"
```
dignifiedquire, line 61: why remove?

Arqu: Cause the net benefit in this case is 0 and it would take me more time to make it work for windows.

Windows is not set up the same way so it starts complaining. Instead of doing a custom unset for windows or separate job, simply disabling it works as well. 
These are jobs that are only rarely run manually and on tagged releases. They run with a different build profile so those caches we have from PRs are worth nothing in this case, might even be net negative where it pushes out some of those deps. By the time we do a follow up release, the deps are incredibly stale and most likely pushed out due to storage limits.

PR #1993, opened 2024-02-05
refactor(iroh-net): Introduce a minimal DerpUrl

flub:
## Description

This addresses two issues for now: some very minimal normalisation and
nicer Debug printing.

The type is purely a wrapper and does extremely minimal extra things
to the underlying Url.  There is no validation happening and no added
contraints to the URLs.  It is still possible to construct a DerpUrl
that will fail, e.g. by using the mailto: scheme.  The users of the
DerpUrl are expected to fail in such cases.

## Notes & open questions

I was initially trying to keep this crate-local to iroh-net.  However
iroh-net has a surprisingly large public API and that would be painful
and largely defeat the point to leave all those public APIs as Url.
Thus it is a public API.

As a consequence there are a few uses of this outside of the iroh-net
crate as well.  I think they are reasonable.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/src/derp.rs
+++ b/iroh-net/src/derp.rs
@@ -14,6 +14,7 @@ use std::fmt;
 use std::ops::Deref;
 use std::str::FromStr;
 
+use anyhow::Context;
 use serde::{Deserialize, Serialize};
 use url::Url;
 
@@ -70,12 +71,15 @@ impl From<Url> for DerpUrl {
     }
 }
 
+/// This is a convenience only to directly parse strings.
+///
+/// If you need more control over the error first create a [`Url`] and use [`DerpUrl::from`]
+/// instead.
 impl FromStr for DerpUrl {
-    // Be aware, we are re-exporting another crate's public type in our API.
-    type Err = url::ParseError;
+    type Err = anyhow::Error;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let inner = Url::from_str(s)?;
+        let inner = Url::from_str(s).context("invalid URL")?;
         Ok(DerpUrl::from(inner))
     }
 }
@@ -120,10 +124,12 @@ mod tests {
     use super::*;
 
     #[test]
-    fn test_derp_url_debug() {
+    fn test_derp_url_debug_display() {
         let url = DerpUrl::from(Url::parse("https://example.com").unwrap());
 
         assert_eq!(format!("{url:?}"), r#"DerpUrl("https://example.com./")"#);
+
+        assert_eq!(format!("{url}"), "https://example.com./");
     }
 
     #[test]
--- a/iroh-net/src/derp.rs
+++ b/iroh-net/src/derp.rs
@@ -11,7 +11,7 @@
 #![deny(missing_docs, rustdoc::broken_intra_doc_links)]
 
 use std::fmt;
-use std::ops::{Deref, DerefMut};
+use std::ops::Deref;
 use std::str::FromStr;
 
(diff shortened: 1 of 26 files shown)


flub:
See https://github.com/n0-computer/iroh/pull/1994 for the cargo-deny failure

rklaehn (approved):
I think there is some custom debug instance for AddrInfo that is no longer needed, assuming that AddrInfo also uses a DerpUrl.

```rust

impl Debug for AddrInfo {
        f.debug_struct("AddrInfo")
            .field("derp_url", &self.derp_url.as_ref().map(DD))
            .field("direct_addresses", &self.direct_addresses)
            .finish()
    }
}
```

iroh-net/src/derp.rs
```diff
@@ -28,3 +36,109 @@ pub use self::server::{
...
+        Self(url)
+    }
+}
+
+impl FromStr for DerpUrl {
+    // Be aware, we are re-exporting another crate's public type in our API.
+    type Err = url::ParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let inner = Url::from_str(s)?;
```
dignifiedquire, line 78: this doesn't do the domain ending check above, why?

iroh-net/src/derp.rs
```diff
@@ -28,3 +36,109 @@ pub use self::server::{
...
+/// [`DerefMut`]: std::ops::DerefMut
+impl Deref for DerpUrl {
+    type Target = Url;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl fmt::Debug for DerpUrl {
```
dignifiedquire, line 97: missing `Display` impl for `DerpUrl`

iroh-net/src/derp.rs
```diff
@@ -28,3 +36,109 @@ pub use self::server::{
     ClientConnHandler, MaybeTlsStream as MaybeTlsStreamServer, PacketForwarderHandler, Server,
 };
 pub use self::types::{MeshKey, PacketForwarder};
+
+/// A URL identifying a DERP server.
+///
+/// This is but a wrapper around [`Url`], with a few custom tweaks:
+///
+/// - A DERP URL is never a relative URL, so an implicit `.` is added at the end of the
```
dignifiedquire, line 44: this is never checked in the constructors

iroh-net/src/derp.rs
flub: Oops, this was meant to use `DerpUrl::from` to make sure this check happened.

iroh-net/src/derp.rs
flub: There is a derive `derive_more::Display`.  I've added a test for that functionality now as well.

iroh-net/src/derp.rs
```diff
@@ -28,3 +36,109 @@ pub use self::server::{
...
+                url.set_host(Some(&domain)).ok();
+            }
+        }
+        Self(url)
+    }
+}
+
+impl FromStr for DerpUrl {
+    // Be aware, we are re-exporting another crate's public type in our API.
+    type Err = url::ParseError;
```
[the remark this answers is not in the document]
flub: We'd actively take away a lot of functionality though.  Admittedly you can get that back by constructing the URL manually first, so maybe that's fine.

Changed this.

iroh-net/src/derp.rs
dignifiedquire: lol, I was just blind I guess

PR #1995, opened 2024-02-05
feat(iroh-net): add DerpMode::StunOnly

dignifiedquire:
This allows to use the default derp map, but only using them for stun, not relay.


dignifiedquire:
not needed...

PR #2000, opened 2024-02-06
refactor: cleanup ProgressSliceWriter

Frando:
## Description

The `ProgressSliceWriter` is currently duplicated code in `iroh_bytes` and `iroh`. Also there's two versions of it, one that takes a fallible and one that takes an infallible one.

This PR removes the code duplication and makes the naming understandable.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-bytes/src/util/progress.rs
+++ b/iroh-bytes/src/util/progress.rs
@@ -442,11 +442,11 @@ impl<W: AsyncSliceWriter + 'static, F: FnMut(u64, usize) + 'static> AsyncSliceWr
     }
 }
 
-/// A slice writer that adds an asynchronous progress callback.
+/// A slice writer that adds a fallible progress callback.
 ///
 /// This wraps any `AsyncSliceWriter`, passes through all operations to the inner writer, and
-/// calls the passed `on_write` callback whenever data is written. `on_write` must return a future
-/// which is awaited before continuing with the write.
+/// calls the passed `on_write` callback whenever data is written. `on_write` must return an
+/// `io::Result`, and can abort the download by returning an error.
 #[derive(Debug)]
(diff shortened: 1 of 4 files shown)


iroh-bytes/src/util/progress.rs
```diff
@@ -391,3 +393,112 @@ impl From<ProgressSendError> for std::io::Error {
...
+    }
+}
+
+/// A slice writer that adds an asynchronous progress callback.
+///
+/// This wraps any `AsyncSliceWriter`, passes through all operations to the inner writer, and
+/// calls the passed `on_write` callback whenever data is written. `on_write` must return a future
+/// which is awaited before continuing with the write.
+#[derive(Debug)]
+pub struct FallibleProgressSliceWriter<W, F>(W, F);
```
dignifiedquire, line 451: the naming is confusing, why is this not `AsyncProgressSliceWriter`?

Frando: The comment is wrong. I first misread the code, then wrote the comment, then understood the code :D will fix

iroh-bytes/src/util/progress.rs
```diff
@@ -391,3 +393,112 @@ impl From<ProgressSendError> for std::io::Error {
...
+
+impl<W: AsyncSliceWriter, F: Fn(u64, usize) -> io::Result<()> + 'static>
```
rklaehn, line 453: The description above says that F is an async fn. But it is just a sync, but fallible fn. So I think either the description or the code is wrong.

Frando: Very correct ;) pushed a fix. wrote docs before rereading the code, and then forgot to update.

rklaehn (approved):

PR #2003, opened 2024-02-06
refactor: download and export structure and progress events

Frando:
## Description

This PR has a number of cleanups regarding download and export functionality:
* We currently have an enum `DownloadProgress` which contains events for both downloading a blob and for exporting a blob from the node's blob store to files on the local filesystem. This PR moves the export progress events into an `ExportProgress` enum.
* The code to export blobs to the local file system is moved from `iroh` to `iroh-bytes`
* Code duplication around export progress is reduced and a couple of RPC and client structs are cleaned up accordingly

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-bytes/src/export.rs
+++ b/iroh-bytes/src/export.rs
@@ -26,61 +26,66 @@ use crate::{
 pub async fn export<D: BaoStore>(
     db: &D,
     hash: Hash,
-    out: PathBuf,
+    outpath: PathBuf,
     recursive: bool,
     mode: ExportMode,
     progress: impl ProgressSender<Msg = ExportProgress> + IdGenerator,
 ) -> anyhow::Result<()> {
-    let path = PathBuf::from(&out);
     if recursive {
-        tokio::fs::create_dir_all(&path).await?;
-        let collection = Collection::load(db, &hash).await?;
-        for (name, hash) in collection.into_iter() {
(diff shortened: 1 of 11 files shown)


iroh-bytes/src/export.rs
```diff
@@ -0,0 +1,129 @@
...
+pub async fn export<D: BaoStore>(
+    db: &D,
+    hash: Hash,
+    out: PathBuf,
+    recursive: bool,
+    mode: ExportMode,
+    progress: impl ProgressSender<Msg = ExportProgress> + IdGenerator,
+) -> anyhow::Result<()> {
+    let path = PathBuf::from(&out);
+    if recursive {
```
dignifiedquire, line 35: should we error if neither `recursive` nor a parent path is found?

Frando: The non-recursive case always exports only to a single file, and the parent path is only used to create the parent as a directory if it doesn't exist. So I think we can simply only do the mkdir if parent is set, but always export? Exporting a blob to `/myfile` is a valid usecase I think, no reason to disallow it?

Frando: I pushed a commit that both reduces code duplication and implements the above (allow exports to a file in root).

dignifiedquire (approved):

PR #2005, opened 2024-02-07
refactor: Fallible store traits

rklaehn:
## Description

This makes the store traits fallible, which is a precondition for wiring up a store that uses IO for the blob index, such as the redb store.

There will be more changes to the traits, but this is necessary in any case and also a bit painful, so let's get it out of the way.

This also needs to extend the list-x responses with an error case, unfortunately.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh/src/node.rs
```diff
@@ -782,21 +783,85 @@ impl<D: BaoStore> RpcHandler<D> {
...
-        let db = self.inner.db.clone();
-        futures::stream::iter(db.blobs()).filter_map(move |hash| {
-            let db = db.clone();
-            async move {
-                let entry = db.get(&hash)?;
-                let hash = entry.hash().into();
-                let size = entry.outboard().await.ok()?.tree().size().0;
-                let path = "".to_owned();
-                Some(BlobListResponse { hash, size, path })
+        Gen::new(|co| async move {
```
rklaehn, line 862: I guess this pattern is so common that we might want to do a helper for it. But maybe not just yet. It isn't that long.

iroh/src/sync_engine.rs
```diff
@@ -243,11 +243,12 @@ impl SyncEngine {
...
-pub(crate) fn entry_to_content_status(entry: EntryStatus) -> ContentStatus {
+pub(crate) fn entry_to_content_status(entry: io::Result<EntryStatus>) -> ContentStatus {
     match entry {
-        EntryStatus::Complete => ContentStatus::Complete,
-        EntryStatus::Partial => ContentStatus::Incomplete,
-        EntryStatus::NotFound => ContentStatus::Missing,
+        Ok(EntryStatus::Complete) => ContentStatus::Complete,
+        Ok(EntryStatus::Partial) => ContentStatus::Incomplete,
+        Ok(EntryStatus::NotFound) => ContentStatus::Missing,
+        Err(_) => ContentStatus::Missing,
```
rklaehn, line 251: Not sure if it is OK to just return missing on error, but anything else would require major changes in the sync engine.

iroh/src/node.rs
```diff
@@ -805,67 +870,67 @@ impl<D: BaoStore> RpcHandler<D> {
...
-                })
+        Gen::new(move |co| async move {
+            if let Err(e) = self.blob_list_collections_impl(&co).await {
+                co.yield_(BlobListCollectionsResponse::IoError(e.into()))
+                    .await;
             }
         })
     }
 
+    // fn blob_list_collections(
```
dignifiedquire, line 893: keep or delete?

rklaehn: Oops. Delete.

iroh/src/sync_engine.rs
dignifiedquire: maybe add a debug log at least?

iroh/src/rpc_protocol.rs
```diff
@@ -155,13 +155,18 @@ pub struct BlobListRequest;
...
+    Item {
+        /// Location of the blob
+        path: String,
+        /// The hash of the blob
+        hash: Hash,
+        /// The size of the blob
+        size: u64,
+    },
+    /// An error occurred while listing blobs
+    IoError(RpcError),
```
dignifiedquire, line 169: why is this not `Result<BlobListResponse, RpcError>` this new enum makes the (public facing) api quite confusing

iroh/src/sync_engine.rs
```diff
@@ -243,11 +243,15 @@ impl SyncEngine {
...
     match entry {
-        EntryStatus::Complete => ContentStatus::Complete,
-        EntryStatus::Partial => ContentStatus::Incomplete,
-        EntryStatus::NotFound => ContentStatus::Missing,
+        Ok(EntryStatus::Complete) => ContentStatus::Complete,
+        Ok(EntryStatus::Partial) => ContentStatus::Incomplete,
+        Ok(EntryStatus::NotFound) => ContentStatus::Missing,
+        Err(cause) => {
+            tracing::warn!("Error while checking entry status: {cause:?}");
+            ContentStatus::Missing
```
dignifiedquire, line 253: one other thought, maybe it might make sense to introduce `ContentStatus::Unknown` for cases like this at some point

dignifiedquire: right, but if the report is `missing` you won't know to do that, right?

rklaehn: Yeah, the proper way to deal with this is to make this a Result<ContentStatus> or have some other way to signal to the sync engine that "something's wrong, please stop what you're doing".

But I would prefer to do this in a subsequent PR.

iroh/src/client.rs
```diff
@@ -420,7 +421,7 @@ where
     /// List all collections.
     pub async fn list_collections(
         &self,
-    ) -> Result<impl Stream<Item = Result<BlobListCollectionsResponse>>> {
+    ) -> Result<impl Stream<Item = Result<RpcResult<BlobListCollectionsResponse>>>> {
```
dignifiedquire, line 424: shouldn't we flatten these results? or is there real value nesting these?

rklaehn: I am not a big fan of flattening these.

One is a network error, the other is an internal error on the remote side. Sure, you can flatten them and make it an anyhow, but I think they are very different things that you also might want to report in different ways.

dignifiedquire: internally I would agree, but from a publics api perspective it is very confusing, especially given that the outer one is already a `anyhow::Result`

dignifiedquire: if you care, you can always downcast back as a consumer

dignifiedquire (approved):

PR #2007, opened 2024-02-07
refactor: Batch write trait

rklaehn:
## Description

To take into account the fact that the size we get initially in the iroh-bytes protocol is not verified, I want to lazily move from mem storage to file storage only once some limit is *actually* exceeded.

This requires combining outboard and data file and treating them a bit more as one file to which you write batches.

For now, just add the trait and keep the impl the same.

## Notes & open questions

Should the BaoBatchWriter move to bao-tree io? I think in the long term yes, but for now let's keep it here to make things simple?

Note: this kind of code *really* wants 1.75 impl return types in traits.... I asked on the public discord if somebody has some major objections, otherwise I would just go with it.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-bytes/src/get.rs
+++ b/iroh-bytes/src/get.rs
@@ -57,7 +57,10 @@ impl Stats {
 pub mod fsm {
     use std::{io, result};
 
-    use crate::protocol::{GetRequest, NonEmptyRequestRangeSpecIter, Request, MAX_MESSAGE_SIZE};
+    use crate::{
+        protocol::{GetRequest, NonEmptyRequestRangeSpecIter, Request, MAX_MESSAGE_SIZE},
+        store::BatchWriter,
+    };
 
     use super::*;
 
@@ -490,6 +493,19 @@ pub mod fsm {
             Ok(res)
         }
 
+        /// Write the entire blob to a slice writer and to an optional outboard.
+        ///
+        /// The outboard is only written to if the blob is larger than a single
+        /// chunk group.
+        pub async fn write_all_batch<B>(self, batch: B) -> result::Result<AtEndBlob, DecodeError>
+        where
+            B: BatchWriter,
+        {
+            let (content, _size) = self.next().await?;
+            let res = content.write_all_batch(batch).await?;
+            Ok(res)
+        }
+
         /// The hash of the blob we are reading.
         pub fn hash(&self) -> Hash {
             (*self.stream.hash()).into()
@@ -684,6 +700,39 @@ pub mod fsm {
                 }
             };
             Ok((done, res))
+        }
+
+        /// Write the entire blob to a batch writer.
+        pub async fn write_all_batch<B>(self, writer: B) -> result::Result<AtEndBlob, DecodeError>
+        where
+            B: BatchWriter,
+        {
+            let mut writer = writer;
+            let mut buf = Vec::new();
+            let mut content = self;
+            let size = content.tree().size().0;
+            loop {
+                match content.next().await {
+                    BlobContentNext::More((next, item)) => {
+                        let item = item?;
+                        match &item {
+                            BaoContentItem::Parent(_) => {
+                                buf.push(item);
+                            }
+                            BaoContentItem::Leaf(_) => {
+                                buf.push(item);
+                                let batch = std::mem::take(&mut buf);
+                                writer.write_batch(size, batch).await?;
+                            }
+                        }
+                        content = next;
+                    }
+                    BlobContentNext::Done(end) => {
+                        assert!(buf.is_empty());
+                        return Ok(end);
+                    }
+                }
+            }
         }
 
         /// Write the entire blob to a slice writer and to an optional outboard.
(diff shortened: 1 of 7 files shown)


iroh-bytes/src/get.rs
```diff
@@ -491,6 +494,16 @@ pub mod fsm {
             Ok(res)
         }
 
+        /// Write the entire stream for this blob to a batch writer.
+        pub async fn write_all_batch<B>(self, batch: B) -> result::Result<AtEndBlob, DecodeError>
+        where
+            B: BaoBatchWriter,
+        {
+            let (content, _size) = self.next().await?;
```
dignifiedquire, line 502: would it make sense to report the total written size in return value?

iroh-bytes/src/store/traits.rs
```diff
@@ -103,23 +110,178 @@ pub trait Map: Clone + Send + Sync + 'static {
...
+/// to the implementation.
+pub trait BaoBatchWriter {
+    /// Write a batch of bao content items to the underlying storage.
+    ///
+    /// The batch is guaranteed to be sorted as data is received from the network.
+    /// So leafs will be sorted by offset, and parents will be sorted by pre order
+    /// traversal offset. There is no guarantee that they will be consecutive
+    /// though.
+    ///
+    /// The size is the size the remote side told us. It is not guaranteed
```
dignifiedquire, line 130: this is the total size of all batch items?

iroh-bytes/src/store/traits.rs
```diff
@@ -103,23 +110,178 @@ pub trait Map: Clone + Send + Sync + 'static {
...
+    /// necessary in typical io code.
+    ///
+    /// TODO: return impl Future<...> once we got 1.75
+    fn write_batch(
+        &mut self,
+        size: u64,
+        batch: Vec<BaoContentItem>,
+    ) -> LocalBoxFuture<'_, io::Result<()>>;
+
+    /// Sync the writer
```
dignifiedquire, line 147: maybe specify to what this syncs, like that it persists or so, and not synced to another node ;)

iroh-bytes/src/store/traits.rs
```diff
@@ -103,23 +110,178 @@ pub trait Map: Clone + Send + Sync + 'static {
...
+            }
+            Ok(())
+        }
+        .boxed_local()
+    }
+
+    fn sync(&mut self) -> LocalBoxFuture<'_, io::Result<()>> {
+        async move {
+            self.data.sync().await?;
+            self.outboard.sync().await?;
```
dignifiedquire, line 268: could these be polled using `join`?

iroh-bytes/src/store/traits.rs
```diff
@@ -103,23 +110,178 @@ pub trait Map: Clone + Send + Sync + 'static {
...
+}
+
+impl<D, O> BaoBatchWriter for CombinedBatchWriter<D, O>
+where
+    D: AsyncSliceWriter,
+    O: OutboardMut,
+{
+    fn write_batch(
+        &mut self,
+        _size: u64,
```
dignifiedquire, line 246: why is the size not needed in this impl?

iroh-bytes/src/store/traits.rs
rklaehn: No, this is the size in bytes of the entire blob. We just can't fully trust it yet. But in the previous impl we did trust it.

This determines what the stuff on the wire means, so this is an *upper bound* on the size, so are guaranteed to only get stuff within 0..size.

rklaehn: Trying to clarify this more in the comments.

dignifiedquire: maybe call it blob_size or sth like that

iroh-bytes/src/get.rs
rklaehn: This returns a state of the iroh-bytes state machine. If you follow the state machine to the end without an error, you get a Stats object which contains this information.

But it is hard to come up with stats that are not too expensive to compute and that are useful for everybody. So to get very precise info you would have to write a wrapper for the batch writer, just like the FallibleProgressBatchWriter, that computes some detailed stats.

```
pub struct Stats {
    /// The number of bytes written
    pub bytes_written: u64,
    /// The number of bytes read
    pub bytes_read: u64,
    /// The time it took to transfer the data
    pub elapsed: Duration,
}
```

iroh-bytes/src/store/traits.rs
rklaehn: yeah, but this thing is temporary anyway. The idea of the batch writer in the long term is to be able to batch things including sync, and to leave it up to the impl what needs to be synced.

So the long term impl would be spawn_blocking(|| { sync all the files }) or something.

iroh-bytes/src/store/traits.rs
rklaehn: We are not quite there yet in terms of tolerating wrong sizes.

The OutboardMut has a Tree, which has a size. So the OutboardMut needs to know in advance what the size is. It gets this info from the PartialEntry, which in this PR still has a size.

The fun part will be one of the next PRs, where we start using this size field here.

dignifiedquire (approved):

PR #2011, opened 2024-02-08
feat(iroh): Add --log-fd flag on unix

flub:
## Description

This allows you to write log output to a specific filedescriptor in
the shell: iroh --log-fd 3 ... 3>some/file.log

Thus not interfering with normal stdout and stderr output and getting
the full functionality of this.

## Notes & open questions

This is splitting off some peripheral stuff from
https://github.com/n0-computer/iroh/pull/1984

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh/src/main.rs
+++ b/iroh/src/main.rs
@@ -29,11 +29,17 @@ async fn main_impl() -> Result<()> {
 
     #[cfg(unix)]
     if let Some(log_fd) = cli.log_fd {
+        use std::fs::File;
+        use std::mem::ManuallyDrop;
         use std::os::unix::io::FromRawFd;
 
-        let f = unsafe { std::fs::File::from_raw_fd(log_fd) };
+        // SAFETY: We take ownership but ensure it is never dropped, thus we never close the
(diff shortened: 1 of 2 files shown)


iroh/src/main.rs
```diff
@@ -25,11 +25,23 @@ fn main() -> Result<()> {
 
 async fn main_impl() -> Result<()> {
     let lp = tokio_util::task::LocalPoolHandle::new(num_cpus::get());
+    let cli = Cli::parse();
+
+    #[cfg(unix)]
+    if let Some(log_fd) = cli.log_fd {
+        use std::os::unix::io::FromRawFd;
+
+        let f = unsafe { std::fs::File::from_raw_fd(log_fd) };
```
dignifiedquire, line 34: should this be `manuallydrop`? from `from_raw_fd` takes ownership and closes the descriptor when the `File` is dropped

dignifiedquire, line 34: please add a comment describing why this is `safe` to do

flub: I believe the ownership is passed into the `.with_witer()` and tracing keeps it alive?

dignifiedquire (approved):

PR #2013, opened 2024-02-08
fix(iroh): do not remove the rpc lockfile if an iroh node is already running

ramfox:
## Description

Previously, we were running `RpcStatus::clear`, every time we ran the `iroh start` command (`start::run_with_command`).

This meant that even when we correctly exited the `iroh start` command in with the error "iroh is already running", we would then remove the rpc lock file, even though iroh was still running in a different process.

Any subsequent calls to `iroh console` or any other iroh commands that expected an iroh node to be running, would assume that a node was *no longer running* because the rpc lock file was gone.

This PR adds a new concrete `AlreadyRunningError` that we return from `start_node`. Before we run `RpcStatus::clear`, we check for the `AlreadyRunningError`, and do not clear the lockfile if that error is present.

## Notes & open questions

Open to ways to make this a bit more elegant, or opinions on a better way to set up the error.

## Change checklist

- [ ] Self-review.
- [x] Tests if relevant.


dignifiedquire (approved):

flub (approved):

flub:
> I have a commit that adjusts the code to:
> 
>     * run `iroh_data_dir()` once in `main.rs`
> 
>     * cascades the `iroh_data_dir` down through the code from there
> 
>     * remove any other references to calling `iroh_data_dir` anywhere else
> 
> 
> Would love some more eyes and opinions tho.

excellent choice

dignifiedquire (approved):
much much better than what we currently have!

PR #2018, opened 2024-02-12
refactor(iroh-bytes): Take advantage of impl T in trait, update bao-tree and iroh-io

rklaehn:
## Description

bao-tree and iroh-io have been republished with new versions that require 1.75 and use impl trait in trait. This updates these two dependencies and also refactors the store traits to take advantage of 1.75 features, removing a ton of boxes along the way.

Instead of `BoxFuture<'a, T>` we use `impl Future<Output = T> + Send`
Instead of `LocalBoxFuture<'a, T>` we use `impl Future<Output = T>`

Explicit lifetimes are in some cases no longer necessary because the future returned by impl Future<...> will implicitly capture the lifetimes of all parameters.

See https://rust-lang.github.io/rfcs/3498-lifetime-capture-rules-2024.html

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -311,9 +311,9 @@ dependencies = [
 
 [[package]]
 name = "bao-tree"
-version = "0.9.1"
+version = "0.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "155e7e0c896695a9049badd7bf2b915d29230e24dc82a7c7ef065eded072404f"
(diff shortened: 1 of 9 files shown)


dignifiedquire (approved):
deleting code is always nice

rklaehn:
> deleting code is always nice

I've been waiting for this for so long. Without this, the whole async syntax feels incomplete...

PR #2019, opened 2024-02-12
feat(iroh-net): DNS queries: lookup ipv6 & ipv4 in parallel

### LINKED ISSUE #2006: Netcheck does not query AAAA records for the derp server
When resolving the derp URL netcheck seems to query only A records, even on a machine which has a full public IPv6 address.

dignifiedquire:
Closes #2006

### DIFF
--- a/iroh-net/src/dns.rs
+++ b/iroh-net/src/dns.rs
@@ -1,6 +1,6 @@
 use anyhow::Result;
 use once_cell::sync::Lazy;
-use trust_dns_resolver::{config, AsyncResolver, TokioAsyncResolver};
+use trust_dns_resolver::{AsyncResolver, TokioAsyncResolver};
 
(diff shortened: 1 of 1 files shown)


flub (approved):
Nice, tested it and it now gets the AAAA records and my dual-stack host even connects to the derper over ipv6 as it has a 2ms faster ping.

dignifiedquire:
> rather than one single query for both A and AAAA records.

Is that even possible? I didn't see anything in `trust-dns`s options to allow for such a thing.

flub:
> > rather than one single query for both A and AAAA records.
> 
> Is that even possible? I didn't see anything in `trust-dns`s options to allow for such a thing.

I assumed yes.  But a quick look suggests that's completely wrong and this isn't possible.  So yes, we're doing the best thing we can already.  TIL

PR #2023, opened 2024-02-13
refactor(iroh-bytes): Simplify store traits

rklaehn:
## Description

Simplify and rename some store traits, taking advantage of return position impl trait in trait.

The main change is that the map entries are no longer for one specific map impl, but just generic map entries, so the type parameter goes away. Other than that, the PartialXXX traits are renamed to XXXMut, which is more accurate. An immutable entry can be partial. Since MapEntry does not guarantee that it is complete (anymore), the main difference between MapEntry and MapEntryMut is not that the latter allows writing.

Edit: one additional change is to introduce BaoBlobSize, which is a size together with an information about whether the size is validated or not. Why not just (u64, bool)? Because later we might add more info to the unverified enum case.

## Notes & open questions

Note: this is not the end of it. Just trying to do this incrementally.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-bytes/src/store/mem.rs
+++ b/iroh-bytes/src/store/mem.rs
@@ -21,8 +21,8 @@ use super::PossiblyPartialEntry;
 use super::TempCounterMap;
 use crate::{
     store::{
-        EntryStatus, ExportMode, ImportMode, ImportProgress, Map, MapEntry, PartialMap,
-        PartialMapEntry, ReadableStore, ValidateProgress,
+        EntryStatus, ExportMode, ImportMode, ImportProgress, Map, MapEntry, MapEntryMut,
+        MapMut, ReadableStore, ValidateProgress,
     },
     util::{
         progress::{IdGenerator, IgnoreProgressSender, ProgressSender},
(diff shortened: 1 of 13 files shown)


iroh-bytes/src/store/mem.rs
```diff
@@ -215,11 +216,11 @@ impl MapEntry<Store> for Entry {
         self.outboard.tree().size().0
     }
 
-    async fn outboard(&self) -> io::Result<PreOrderOutboard<MemFile>> {
+    async fn outboard(&self) -> io::Result<impl Outboard> {
         Ok(self.outboard.clone())
     }
 
-    async fn data_reader(&self) -> io::Result<MemFile> {
+    async fn data_reader(&self) -> io::Result<impl AsyncSliceReader> {
```
dignifiedquire, line 223: this makes it impossible to store this in a struct without boxing it are we okay with that?

rklaehn: > this makes it impossible to store this in a struct without boxing it are we okay with that?

I think it is fine.

Once you have impl traits you can't really do manual state machines in a nice way anymore, so the way to "store this in a struct" is to have an async block and let rustc make a struct for you. But that is the style you end up doing anyway if you don't want to go crazy.

You can still have functions, including async fns, that take an unnamed type. And you can put your logic into an async block and spawn it, which basically boxes it anyway.

I have tried the explicit associated type thing now for a while, and is just so painful to compose them. E.g. just a thin wrapper to add logging or any kind of logic requires either pages of boilerplate or... boxing. So 🤷...

dignifiedquire (approved):

Frando (approved):
:+1: Nothing stood out to me, apart from the already resolved question of whether we want to use RPITIT, but I think with I agree with rklaehn's reasoning above that you usually will need lots of boilerplate or boxing anyways. And the code reads much nicer with the impl Trait than with the associated traits. So: LGT;M!

PR #2025, opened 2024-02-14
fix(iroh-net): correctly report sent transmits in poll_send

dignifiedquire:
If udp and derp errored (or are not available) but we have sent pings, this would wrongly report that all transmits were sent.

### DIFF
--- a/iroh-net/src/magicsock.rs
+++ b/iroh-net/src/magicsock.rs
@@ -303,7 +303,6 @@ impl Inner {
         if transmits.is_empty() {
             return Poll::Ready(Ok(n));
         }
-
         trace!(
             "sending:\n{}",
             transmits.iter().fold(
@@ -344,6 +343,7 @@ impl Inner {
 
         let dest = QuicMappedAddr(dest);
 
+        let mut transmits_sent = 0;
         match self
             .node_map
             .get_send_addrs_for_quic_mapped_addr(&dest, self.ipv6_reported.load(Ordering::Relaxed))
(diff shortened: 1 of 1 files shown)


ramfox (approved):

iroh-net/src/magicsock.rs
```diff
@@ -388,11 +390,10 @@ impl Inner {
                     }
                 }
 
-                let n = transmits.len();
-
                 // send derp
                 if let Some(ref derp_url) = derp_url {
                     self.try_send_derp(derp_url, public_key, split_packets(&transmits));
+                    transmits_sent = transmits.len();
```
flub, line 396: this seems to be mostly a wish to be fair, given how `try_send_derp` can drop things on the floor if it feels like it.  but for the purpose of this fix it's probably fine to leave as it.

iroh-net/src/magicsock.rs
```diff
@@ -408,14 +409,18 @@ impl Inner {
...
                         send_derp = ?derp_url,
                         "sent transmits"
                     );
-                    Poll::Ready(Ok(n))
+                    assert!(
+                        transmits_sent <= transmits_len,
+                        "too many sent {} > {}",
+                        transmits_sent,
+                        transmits_len
+                    );
```
flub, line 422: curious why you feel the need for this assert?  It's also the only reason you keep `transmits_len` around.

flub (approved):

iroh-net/src/magicsock.rs
dignifiedquire: this was from debugging this, probably can remove it

iroh-net/src/magicsock.rs
flub: i guess that's a fair way to look at it

Frando (approved):

PR #2026, opened 2024-02-15
fix(iroh-net): race ipv4 and ipv6 dns resolution

dignifiedquire:
This now races the resolution, with a fixed timeout on each, returning as many results as found during that time.

Ref https://github.com/n0-computer/iroh/issues/2006

### DIFF
--- a/iroh-net/src/dns.rs
+++ b/iroh-net/src/dns.rs
@@ -23,7 +23,11 @@ fn get_resolver() -> Result<TokioAsyncResolver> {
     Ok(resolver)
 }
 
-/// Resolve IPv4 and IPv6 in parallel
+/// Resolve IPv4 and IPv6 in parallel.
+///
+/// `LookupIpStrategy::Ipv4AndIpv6` will wait for ipv6 resolution timeout, even if it is
+/// not usable on the stack, so we manually query both lookups concurrently and time them out
+/// individually.
 pub(crate) async fn lookup_ipv4_ipv6<N: IntoName + TryParseIp + Clone>(
     host: N,
     timeout: Duration,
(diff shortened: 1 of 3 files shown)


iroh-net/src/dns.rs
```diff
@@ -14,23 +17,83 @@ fn get_resolver() -> Result<TokioAsyncResolver> {
...
         trust_dns_resolver::system_conf::read_system_conf().unwrap_or_default();
     // lookup IPv4 and IPv6 in parallel
-    options.ip_strategy = trust_dns_resolver::config::LookupIpStrategy::Ipv4AndIpv6;
+    options.ip_strategy = trust_dns_resolver::config::LookupIpStrategy::Ipv4thenIpv6;
 
     let resolver = AsyncResolver::tokio(config, options);
     Ok(resolver)
 }
 
+/// Resolve IPv4 and IPv6 in parallel
```
flub, line 26: This should probably elaborate a little why we don't use the `Ipv4AndIpv6` strategy.

iroh-net/src/dns.rs
```diff
@@ -14,23 +17,83 @@ fn get_resolver() -> Result<TokioAsyncResolver> {
...
+    match res {
+        (Ok(Ok(ipv4)), Ok(Ok(ipv6))) => {
+            let res = ipv4
+                .into_iter()
+                .map(|ip| IpAddr::V4(ip.0))
+                .chain(ipv6.into_iter().map(|ip| IpAddr::V6(ip.0)))
+                .collect();
+            Ok(res)
+        }
+        (Ok(Ok(ipv4)), Err(_err)) => {
```
flub, line 46:
```suggestion
        (Ok(Ok(ipv4)), Err(_timeout)) => {
```
is perhaps a little bit more readable?
(in all places were a timeout err is matched)

flub (approved):
I think this is a reasonable tradeoff and a quick improvement.

I guess the right thing is to rework reportgen so that it can start probe sets as soon as a result is returned and return a stream of results.  But let's do the simple thing first.

iroh-net/src/dns.rs
dignifiedquire: added some details

iroh-net/src/dns.rs
dignifiedquire: done

PR #2027, opened 2024-02-15
fix(iroh-net): Ensure netcheck finishes once it has results

### LINKED ISSUE #2021: netcheck report does not finish early enough
I have a netcheck report where we have established connectivity to the derper and figured out our IPv4 STUN addresses very quickly, yet keep waiting for all probes much, much to long.  In fact it logs `Have enough probe reports, aborting further probes soon` and yet it waits forever for to actually abort.  This wreaks havoc with good connection establishment.

Splitting this off to keep #1984 somewhat limited in scope.

Relevant log: https://gist.github.com/flub/1d503f757bc1e275e417d0d918b66604

flub:
## Description

Once netcheck has managed to establish an external IP address there is
no need to further wait on the portmapper.  The results are already
useful.  If later the portmapper manages to get something done the
magicsock will react to that and make use of it.

## Notes & open questions

Fixes #2021

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/src/netcheck/reportgen.rs
+++ b/iroh-net/src/netcheck/reportgen.rs
@@ -170,6 +170,12 @@ struct Actor {
     last_report: Option<Arc<Report>>,
     /// The portmapper client, if there is one.
     port_mapper: Option<portmapper::Client>,
+    /// Whether the actor should try to reach things other than localhost.
+    ///
+    /// This is set to true in tests to avoid probing the local LAN's router etc.
+    ///
+    /// TODO: currently this only skips the portmapper, but still does STUN, HTTP and ICMP
+    /// probes.  Furthermore none of our tests actually use this feature.
     skip_external_network: bool,
     /// The DERP configuration.
     derp_map: DerpMap,
@@ -266,7 +272,6 @@ impl Actor {
                     self.report.portmap_probe = pm;
                     port_mapping.inner = None;
                     self.outstanding_tasks.port_mapper = false;
-                    trace!("portmapper future done");
                 }
 
                 // Check for probes finishing.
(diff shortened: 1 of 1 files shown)


dignifiedquire (approved):
code looks good, what impact have you seen in testing?

iroh-net/src/netcheck/reportgen.rs
```diff
@@ -170,6 +170,12 @@ struct Actor {
     last_report: Option<Arc<Report>>,
     /// The portmapper client, if there is one.
     port_mapper: Option<portmapper::Client>,
+    /// Whether the actor should try to reach things other than localhost.
+    ///
+    /// This is set to true in tests to avoid probing the local LAN's router etc.
+    ///
+    /// TODO: currently this only skips the portmapper, but still does STUN, HTTP and ICMP
+    /// probes.  Furthermore none of our tests actually use this feature.
```
dignifiedquire, line 178: honestly, just delete it, we can bring it back if we ever need to

flub:
Pinging @divagant-martian to check my understanding of the portmapper is correct.

I think what happens is that on the very first netcheck the portmapper has not yet run to completion and that slows down the first netcheck, often only to not get a portmapper result anyway.  Later on there are cached reports that are returned.  As all this is reporting is a bunch of bools in an option I think it's fine to not wait for this the first time.

iroh-net/src/netcheck/reportgen.rs
flub: yeah, fair enough.  i'll do that in a separate PR though.

flub:
@divagant-martian I'm going to interpret the thumbs-up as approval.  Can always revert if that was wrong.

PR #2032, opened 2024-02-17
fix(console): prevent deadlock in `author new --switch`

### LINKED ISSUE #2024: bug: in the console `author new --switch` hangs
To repro:

`iroh console --start`

then in the console:

`author new --switch`

The new author id is printed to the console, but the console never returns to a state to receive more input. You can't even `Ctrl-C` to interrupt the command nor can you `Ctrl-D` to close the console.

divagant-martian:
## Description

Fixes #2024

This was caused by a read attempt over a lock nested inside a write lock. Thw write lock being `let mut inner = self.0.write()` and the read lock in `self.iroh_data_dir()`

## Notes & open questions

I checked the rest of the code in #2013 and I think `author new --switch` is the only command affected

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/config.rs
+++ b/iroh/src/config.rs
@@ -268,15 +268,13 @@ impl ConsoleEnv {
     /// Will error if not running in the Iroh console.
     /// Will persist to a file in the Iroh data dir otherwise.
     pub fn set_author(&self, author: AuthorId) -> anyhow::Result<()> {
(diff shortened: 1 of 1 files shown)


dignifiedquire (approved):
the beauty of locks..

PR #2034, opened 2024-02-20
fix(iroh-net): Fix some flaky magicsock tests

flub:
## Description

Fix flaky tests in iroh-net (hopefully).  The main issue was that the
testing code meshing together the sockets had a race-condition and
would sometimes fail.

Other improvements:

- Logging improvements of various parts.  These tests should now
  provide better info on failure.

- Add alternative formatting of NodeId to use the short format.

- Use NodeId in the public API, rather than PublicKey.

- Move multi-threaded logging setup to the testing utils crate now
  that it can work properly.  More code will have to switch.

## Notes & open questions

The local_endpoints_change API is really difficult to use correctly.
Even the way this testing code uses it now is wrong.  Instead of the
two current APIs there should be a single API which returns a stream.
I'll do this as a separate followup PR though.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/src/key.rs
+++ b/iroh-net/src/key.rs
@@ -227,7 +227,11 @@ impl Debug for PublicKey {
 
 impl Display for PublicKey {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", base32::fmt(self.as_bytes()))
+        if f.alternate() {
+            write!(f, "{}", base32::fmt_short(self.as_bytes()))
+        } else {
+            write!(f, "{}", base32::fmt(self.as_bytes()))
+        }
     }
 }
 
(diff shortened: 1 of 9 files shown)


iroh-net/src/key.rs
```diff
@@ -227,7 +227,11 @@ impl Debug for PublicKey {
 
 impl Display for PublicKey {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", base32::fmt(self.as_bytes()))
+        if f.alternate() {
```
dignifiedquire, line 230: when/how is this triggered?

dignifiedquire: shouldn’t it be the other way around then,? # prints longer more detailed messages

flub: I could be rather tempted to switch this around and do the short format by default and the long format for alternate.  That would reduce rather a lot of verbosity in our code.

dignifiedquire: the issue is that, even externally, we rely on `to_string()` roundtriping

flub: yeah, it'd probably break things if this can't be round-tripped.  so maybe we'll have to settle for this.

dignifiedquire (approved):

PR #2041, opened 2024-02-22
refactor(iroh-net): Improve API to retrieve local endpoints

flub:
## Description

The API to retrieve local endpoints was split in two between the
current endpoints and updates to it.  This was very hard to use right
without any race conditions.

This replaces it with a single API which returns a stream.  This is a
little bit wordier to use if you just need the current local item, but
much harder to misuse.

## Notes & open questions

It would be possible to add a convenient `.current_local_endpoints()`
API if we wanted.  Not sure this is really worth it.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/src/magicsock.rs
+++ b/iroh-net/src/magicsock.rs
@@ -24,6 +24,7 @@ use std::{
     io,
     net::{IpAddr, Ipv6Addr, SocketAddr},
     path::PathBuf,
+    pin::Pin,
     sync::{
         atomic::{AtomicBool, AtomicU16, AtomicU64, Ordering},
         Arc,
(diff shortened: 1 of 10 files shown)


iroh-net/src/magicsock.rs
```diff
@@ -1419,6 +1431,46 @@ impl MagicSock {
...
+                Poll::Ready(Some(discovered)) => {
+                    if discovered.is_empty() {
+                        // When we start up we might initially have empty local endpoints as
+                        // the magic socket has not yet figured this out.  Later on this set
+                        // should never be emtpy.  However even if it was the magicsock
+                        // would be in a state not very useable so skipping those events is
+                        // probably fine.
+                        // To make sure we install the right waker we loop rather than
+                        // returning Poll::Pending immediately here.
+                        continue;
```
dignifiedquire, line 1463: why not return `Poll::Pending` here?

dignifiedquire (approved):
thank you for improving this horrible api!

PR #2045, opened 2024-02-27
feat: node discovery via DNS

### LINKED ISSUE #1248: magic_endpoint: allow dialing with just PeerId
The magic endpoint stores known addresses of peers in the PeerMap.  This should allows it to keep state across dials and allows successful dials with just a PeerId.  We should allow this.

See https://github.com/n0-computer/iroh/pull/1222#discussion_r1264970532

Frando:
## Description

This enables global node discovery over DNS, i.e. dialing nodes by just their node id.

Current setup is as follows:

* When dialing a node only by its NodeId, the new `DnsDiscovery` service is invoked. It will lookup a TXT record at (by default) `_iroh_node.b32encodednodeid.testdns.iroh.link` over regular DNS or DNS-over-http. Right now the Cloudflare DNS servers are configured. At `testdns.iroh.link` we run a custom [DNS server](https://github.com/n0-computer/iroh-dns-server/tree/main)
* Nodes publish their Derp address to this DNS server through Pkarr signed packets. This is an intermediate step, we decided that the publishing by default should not happen by the nodes directly but mediated through the Derp servers. Work for the latter happens in #2052 

This PR thus allows for the following:
```sh
# terminal/computer 1
$ iroh console --start
Iroh is running
Node ID: qp2znfedwdij4llc5noizwfemfgba7bzxozvr4bp7hfsdmwqbpua
$ blob add ./myfile
...
Blob: o5uanh5s2zwn2sucy47puqidsfx2advxos7kajq3ajwitcwobhba
...

# terminal/computer 2
iroh console --start
blob get o5uanh5s2zwn2sucy47puqidsfx2advxos7kajq3ajwitcwobhba --node qp2znfedwdij4llc5noizwfemfgba7bzxozvr4bp7hfsdmwqbpua
```

## Notes & open questions

* Misses node configuration in the CLI for the node origin domain (right now hardcoded to `testdns.iroh.link`). How do we want to expose this - CLI flag? Or in the config file? I'd say the latter.

* Offload publishing to the Derpers - see #2052 

* Right now the records published via pkarr have a TTL of 30s - the iroh-dns-server will use that TTL as-is when serving the records over DNS. both can/should change?

* We can also *very* easily allow to lookup nodes not only by NodeId, but by any domain name. In the `iroh-dns` crate I included an example `resolve` that does just that. By setting a `CNAME` record you can even use any domain and simply point to the record hosted at the `testdns.iroh.link` server.
So if, on your custom domain, you added a record like this
```
_iroh_node.frando.n0.computer CNAME _iroh_node.qp2znfedwdij4llc5noizwfemfgba7bzxozvr4bp7hfsdmwqbpua.iroh.link.
```
You can use this with the example to resolve to the node id and derp addresses:
```
cargo run --example resolve -- domain frando.n0.computer
```

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

Closes #1248


iroh-net/src/magic_endpoint.rs
```diff
@@ -468,21 +453,42 @@ impl MagicEndpoint {
     /// connection without involving a DERP server. If no addresses are specified, the endpoint
     /// will try to dial the peer through the configured DERP servers.
     ///
-    /// If the `derp_url` is not `None` and the configured DERP servers do not include a DERP node from the given `derp_url`, it will error.
+    /// If the `derp_url` is not `None` and the configured DERP servers do not include a DERP node from the given `derp_url`,
+    /// it will use the configures [`Discovery`] service to retrieve DERP or direct addresses. If
```
flub, line 457:
```suggestion
    /// it will use the configured [`Discovery`] service to retrieve DERP or direct addresses. If
```

b5:
Does this close #1248?

iroh-net/src/magic_endpoint.rs
Frando: Moved to #2056

Frando:
> Does this close #1248?

Yes

iroh-net/src/discovery/dns.rs
```diff
@@ -0,0 +1,63 @@
...
+pub const N0_TESTDNS_NODE_ORIGIN: &str = "testdns.iroh.link";
+
+/// DNS node discovery.
+///
+/// The DNS discovery looks up node addressing information over the Domain Name System.
+/// Node information is resolved via an _iroh_node.z32encodednodeid TXT record.
+///
+/// The content of this record is expected to be a DNS attribute string, with a required
+/// `node=` attribute containing the base32 encoded node id and a derp_url attribute containing the
+/// node's home Derp server.
```
ramfox, line 22:
```suggestion
/// `node=` attribute containing the base32 encoded node id and a relay_url attribute containing the
/// node's home iroh-relay server.
```

iroh-cli/src/commands/blob.rs
```diff
@@ -236,11 +236,6 @@ impl BlobCommands {
                     return Err(anyhow::anyhow!("The input arguments refer to a collection of blobs and output is set to STDOUT. Only single blobs may be passed in this case."));
                 }
 
-                if node_addr.info.is_empty() {
-                    return Err(anyhow::anyhow!(
-                        "no relay url provided and no direct addresses provided"
-                    ));
-                }
```
rklaehn, line 243: :+1:

iroh-net/src/discovery/pkarr_publish.rs
```diff
@@ -0,0 +1,116 @@
+//! A discovery service which publishes node information to a [Pkarr] relay.
+//!
+//! This service only implements the [`Discovery::publish`] method and does not provide discovery.
+//! It encodes the node information into a DNS packet in the format resolvable by the
+//! [`super::dns::DnsDiscovery`], which means a single _iroh_node TXT record, under the z32 encoded
+//! node id as origin domain.
+//!
+//! [pkarr]: https://pkarr.org
+
+// TODO: Decide what to do with this module once publishing over relays land. Either remove, or
```
rklaehn, line 10: I would say keep it and have it disabled by default.

iroh-net/src/discovery/pkarr_publish.rs
```diff
@@ -0,0 +1,116 @@
...
+
+    /// Set the TTL for pkarr packets, in seconds.
+    ///
+    /// Default value is 30 seconds.
+    pub fn set_ttl(&mut self, ttl: u32) {
+        self.ttl = ttl;
+    }
+
+    /// Publish [`AddrInfo`] about this node to a pkarr relay.
+    pub async fn publish_addr_info(&self, info: &AddrInfo) -> Result<()> {
```
rklaehn, line 66: This does not do republishing. If I remember correctly, the pkarr crate also does not do it. So this will not republish. So the info in the DHT will vanish after a few days.

In iroh-pkarr-node-discovery I spawn a task that republishes. See https://github.com/n0-computer/iroh-examples/blob/main/iroh-pkarr-node-discovery/src/lib.rs#L109

Or is the publishing triggered in regular intervals by the magicsock?

iroh-net/src/discovery/pkarr_publish.rs
```diff
@@ -0,0 +1,116 @@
...
+        let info = NodeInfo::new(
+            self.secret_key.public(),
+            info.relay_url.clone().map(Url::from),
+        );
+        if self.last_published.read().as_ref() == Some(&info) {
```
rklaehn, line 71: This is to only publish if the info is new, right?

iroh-net/src/discovery/pkarr_publish.rs
rklaehn: By the way: in the iroh-pkarr-node-discovery crate I just have a small initial delay before publishing in the task, that way if the addr changes frequently before settling down the transient addrs are not published.

Frando: This only publishes to a pkarr relay over http, not to mainline, so I don't think it needs republishing. If we were to change that to publish to mainline directly we would need republishing, true.

rklaehn: So the idea is that the pkarr relay will continue publishing this information even if it is very old? Or will it somehow detect that the node has not been seen for a while and discard the info?

E.g. you start a test that creates a large number of random node ids. We might do this, but even if we don't, _somebody_ will. So then you are spamming the DHT with thousands of publish requests that are no longer relevant, and risk being rate limited by the DHT nodes.

I guess the derper has all the info it needs to check if info is still relevant.

But still I somehow like the idea of very low frequency republishing to make things self-healing.

E.g. you publish, the publish gets lost because the derper crashed. Now you are in a state where you think your info is publicly known, but it won't be until your network changes (which might be never for a static node).

In general, whenever there is some system that requires reliable delivery of every single message I think it becomes much more fragile compared to if you have a way to avoid this. E.g. a progress update stream that requires the remote side to sum the deltas compared to a progress stream that occasionally sends the full current value to reset the state.

I would also like it if we would retain the ability to publish directly to the DHT via pkarr, at least as an option. So this is not a regression compared to the stuff I hacked with @arnazeh.

iroh-net/src/dns/node_info.rs
```diff
@@ -0,0 +1,267 @@
...
+    } else {
+        Name::parse(IROH_NODE_TXT_LABEL, Some(&name))
+    }
+}
+
+/// Encode a [`NodeId`] in [`z-base-32`] encoding.
+///
+/// [z-base-32]: https://philzimmermann.com/docs/human-oriented-base-32-encoding.txt
+pub fn to_z32(node_id: &NodeId) -> String {
+    z32::encode(node_id.as_bytes())
```
rklaehn, line 60: At the risk of opening a giant rabbit hole or the mother of all bike shed discussions: Can we just make the string representation of our node ids be zbase32 as well?

I think that this is going to become a source of confusion especially when people start manually debugging the DNS records using nslookup/dig. And there is really nothing wrong with zbase32. It's not like our encoding (lower case base32) is completely standard either.

Also there are some valid reasons why the z encoding is better in the admittedly rare cases where people will type those by hand. https://philzimmermann.com/docs/human-oriented-base-32-encoding.txt

I am not going to argue about using hex again, since the 63 char subdomain fragment limit is a very convincing argument...

Frando: I agree that it would be much nicer to have the same encoding everywhere, and that a different encoding for the DNS names than how we print NodeIds in general is quite unfortunate.

* The cleanest would be switch all our string encodings from lowerbase32 to zbase32. This would mean, though, that with the next release all existing tickets, hashes, nodeids would break. How much pain would that cause? I'm not sure. Maybe yes? Depends on how and where and in which ways Iroh is already in production use I guess?
* If we switched only the `to_string()/from_string()` of `NodeId`, we wouldn't have much breakage I think, however then we would move the inconsistency from "two ways to encode a NodeId" towards "NodeIds and hashes are encoded with different formats", which is also not great.

iroh-net/src/discovery/pkarr_publish.rs
Frando: The derper would only once publish the info it receives onto a pkarr relay. Our pkarr relay/DNS server stores and serves everything, forever (or until replaced by update). So also for publish-via-relay: no republishing yet because no mainline DHT integration, thus not needed.

I do agree though that there's no downside and some upsides to eventual republishing. It will increase load a bit on our DNS server/pkarr relay, but well. Maybe once every 5 minutes or so? Not sure.

> I would also like it if we would retain the ability to publish directly to the DHT via pkarr,
I am supportive of this! But would like to move to followups and not do in this PR.

iroh-net/src/dns/node_info.rs
dignifiedquire: lets not do any of this as part of this PR, please make a new issue if you want to bikeshed encodings, and we can consider this independently

iroh-net/src/discovery/dns.rs
```diff
@@ -0,0 +1,63 @@
...
+    discovery::{Discovery, DiscoveryItem},
+    MagicEndpoint, NodeId,
+};
+use anyhow::Result;
+use futures::{future::FutureExt, stream::BoxStream, StreamExt};
+
+use crate::dns;
+
+/// The n0 testing DNS node origin
+pub const N0_TESTDNS_NODE_ORIGIN: &str = "testdns.iroh.link";
```
dignifiedquire, line 13: needs updating, @Arqu what's the state of using the new url?

iroh-net/src/dns/node_info.rs
```diff
@@ -0,0 +1,267 @@
...
+    /// Create a new [`NodeInfo`] from its parts.
+    pub fn new(node_id: NodeId, relay_url: Option<Url>) -> Self {
+        Self { node_id, relay_url }
+    }
+
+    /// Convert this node info into a DNS attribute string.
+    ///
+    /// It will look like this:
+    /// `node=b32encodednodeid relay=https://myrelay.example`
+    pub fn to_attribute_string(&self) -> String {
```
dignifiedquire, line 111: should be `as_attribute_string` according to the convention I believe

dignifiedquire:
@Frando I think this should also have an integration test, running the actual iroh-dns server

iroh-net/src/discovery/pkarr_publish.rs
Frando: See #2151 for a change to do republishing. I am quite unsure though what the republish interval should be. WDYT? Let's take the discussion to 2151 though.

iroh-net/src/dns/node_info.rs
Frando: Rereading the docs at https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv - I think `to_` is correct because this create a new owned `String`?

iroh-net/src/discovery/dns.rs
Frando: Changed to `dns.iroh.link`. Our server is running under both `testdns.iroh.link` and `dns.iroh.link` for now.

Frando:
> @Frando I think this should also have an integration test, running the actual iroh-dns server

Added an online test that uses `dns.iroh.link` for now. It also has an integration test with a very simple test server, see `iroh-net/src/discovery.rs`.

An offline test with spawning an instance of the actual iroh-dns server code I will add once we have a PR to move the server from https://github.com/n0-computer/iroh-dns-server/ into the iroh repo.

iroh-net/src/dns/node_info.rs
dignifiedquire: you are right

iroh-dns-server/Cargo.toml
```diff
@@ -0,0 +1,54 @@
+[package]
+name = "iroh-dns-server"
+version = "0.1.0"
+edition = "2021"
+description = "A pkarr relay and DNS server"
+license = "MIT OR Apache-2.0"
+authors = ["Frando <franz@n0.computer>", "n0 team"]
+repository = "https://github.com/n0-computer/iroh-dns-server"
```
dignifiedquire, line 8: needs updating

iroh-dns-server/Cargo.toml
```diff
@@ -0,0 +1,54 @@
+[package]
+name = "iroh-dns-server"
+version = "0.1.0"
```
dignifiedquire, line 3: maybe we align the version with `iroh`?

Frando: fixed

dignifiedquire (approved):

PR #2046, opened 2024-02-28
refactor(iroh-net): Avoid using .unwrap() calls

### LINKED ISSUE #1528: Audit usage of unwraps

flub:
## Description

This is an attempt at an audit of all the .unwrap() calls in iroh-net.
A few are replaced by .expect() with clear reasons why they're
reasonable.  Some are replaces by non-panicking versions which will
never take the failure branch because of the invariants that are
supposed to be upheld.

Closes #1528

## Notes & open questions

This was surprisingly horrible and long to do.  Oh well.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/src/magicsock.rs
+++ b/iroh-net/src/magicsock.rs
@@ -251,14 +251,14 @@ impl Inner {
     ///
     /// If `None`, then we are not connected to any derp region.
     fn my_derp(&self) -> Option<DerpUrl> {
-        self.my_derp.read().unwrap().clone()
+        self.my_derp.read().expect("not poisoned").clone()
     }
 
     /// Sets the derp node with the best latency.
     ///
     /// If we are not connected to any derp nodes, set this to `None`.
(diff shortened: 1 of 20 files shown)


iroh-net/src/magicsock.rs
```diff
@@ -1403,8 +1403,8 @@ impl MagicSock {
             .actor_sender
             .send(ActorMessage::RebindAll(s))
             .await
-            .unwrap();
-        r.await.unwrap();
+            .ok();
+        r.await.ok();
```
dignifiedquire, line 1407: maybe at leas log these errors?

dignifiedquire (approved):

iroh-net/src/magicsock.rs
flub: done

PR #2051, opened 2024-02-29
refactor(iroh-bytes): Rewrite the blob store to use redb

rklaehn:
## Description

Rewrite the blob store to use redb

This does a number of things:
1. rewrite the blob store to use redb.
2. inline small data blobs as well as small outboards in the redb db
3. properly handle remotes lying about sizes
4. make sure all redb interactions, which can be IO, are done in an actor on a thread

It also adds a consistency check of the store as a feature.

Since this is a totally new new store format, it provides functions to import from a flat store, `import_flat_store` rather than having a migration. This just takes the paths of some flat store directories and adsorbs all the data in that location.

## Notes & open questions

- Open question: should the inline data and inline outboard store be combined into one?
- Todo: audit the places where the actor terminates. If it does, the store is dead. We don't want to do this unless there is something really bad going on, on the other hand it does not make much sense to continue with a broken store.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


rklaehn:
@ppodolsky, since it seems that you already looked at this a bit, you are welcome to review it...

iroh-bytes/Cargo.toml
```diff
@@ -58,10 +59,11 @@ tokio = { version = "1", features = ["macros", "test-util"] }
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 rcgen = "0.12.0"
 rustls = { version = "0.21.0", default-features = false, features = ["quic"] }
+tempfile = "3.10.0"
 
 [features]
-default = ["flat-db"]
-flat-db = ["reflink-copy"]
+default = ["redb-db"]
+redb-db = ["reflink-copy", "redb"]
```
dignifiedquire, line 66: maybe use a more generic term, like `disk-db`? not sure

rklaehn: yeah, I was also thinking that the thing should a proper name. Like some character from a certain well known anime show or something...

rklaehn: Maybe just filestore? The notable thing compared to the mem store is that it is using individual files. That is probably more permanent than using redb.

dignifiedquire: yes, filestore makes sense to me

dignifiedquire: want to do that rename before merging?

iroh-bytes/src/store/redb/util.rs
```diff
@@ -0,0 +1,123 @@
...
+    // todo: should we fail here or just log a warning?
+    // remove could fail e.g. on windows if the file is still open
+    std::fs::remove_file(path)?;
+    Ok(data)
+}
+
+/// A wrapper for a flume receiver that allows peeking at the next message.
+#[derive(Debug)]
+#[allow(dead_code)]
+pub(super) struct PeekableFlumeReceiver<T> {
```
dignifiedquire, line 69: would be nice to have same basic unit tests for helpers like this

iroh-bytes/src/store/redb/validate.rs
```diff
@@ -0,0 +1,362 @@
+//! Validation of the store's contents.
+//!
+//! This performs a full consistency check. Eventually it will also validate
```
dignifiedquire, line 3: would be useful to have a short list of things that it does check

iroh-bytes/src/store/redb/validate.rs
```diff
@@ -0,0 +1,362 @@
...
+        ($($arg:tt)*) => {
+            send!(ValidateLevel::Error, None, $($arg)*)
+        };
+    }
+        macro_rules! entry_error {
+        ($hash:expr, $($arg:tt)*) => {
+            send!(ValidateLevel::Error, Some($hash), $($arg)*)
+        };
+    }
+        // first, dump the entire data content at trace level
```
dignifiedquire, line 68: shouldn't this be gated on the log level, currently this is executed no matter the log level

iroh-bytes/src/store/redb/validate.rs
```diff
@@ -0,0 +1,362 @@
...
+            }
+            Err(cause) => {
+                error!("failed to iterate tags: {}", cause);
+            }
+        }
+
+        // perform consistency check for each entry
+        info!("validating blobs");
+        // set of a all hashes that are referenced by the blobs table
+        let mut entries = BTreeSet::new();
```
dignifiedquire, line 153: this could become quite large, so it would be good to at least warn about this in the docs, that this can eat a lot of memory depending on the size of your store

iroh-bytes/src/util/mem_or_file.rs
```diff
@@ -0,0 +1,104 @@
...
+    /// Get the size of the MemOrFile
+    pub fn size(&self) -> u64 {
+        match self {
+            MemOrFile::Mem(mem) => mem.as_ref().len() as u64,
+            MemOrFile::File((_, size)) => *size,
+        }
+    }
+}
+
+impl ReadAt for MemOrFile<Bytes, File> {
```
dignifiedquire, line 32: it's a bit confusing that `size` is not implemented for this, but only for the `(_, u64)` version

iroh-bytes/src/util/mem_or_file.rs
```diff
@@ -0,0 +1,104 @@
...
+    pub fn as_ref(&self) -> MemOrFile<&M, &F> {
+        match self {
+            MemOrFile::Mem(mem) => MemOrFile::Mem(mem),
+            MemOrFile::File(file) => MemOrFile::File(file),
+        }
+    }
+
+    /// True if this is a Mem
+    pub fn is_mem(&self) -> bool {
+        self.mem().is_some()
```
dignifiedquire, line 67: `matches!(self, ..)` seems more intuitive

iroh-bytes/src/util/sparse_mem_file.rs
```diff
@@ -0,0 +1,122 @@
...
+/// A file that is sparse in memory
+///
+/// It is not actually using sparse storage to make reading faster, so it will
+/// not conserve memory. It is just a way to remember the gaps so we can
+/// write it to a file in a sparse way later.
+#[derive(derive_more::Debug)]
+pub struct SparseMemFile {
+    /// The data, with gaps filled with zeros
+    #[debug("{} bytes", data.len())]
+    data: Vec<u8>,
```
dignifiedquire, line 16: maybe use `Box<[u8]>`, or can you resize this?

iroh/src/commands/start.rs
```diff
@@ -200,14 +159,33 @@ pub(crate) async fn start_node(
...
+            partial: iroh_data_root.join("blobs.v1").join("partial"),
+            meta: iroh_data_root.join("blobs.v1").join("meta"),
+        })
+        .await?;
+    if v0 || v1 {
+        tracing::info!("flat data was imported - reapply inline options");
+        bao_store
+            .update_inline_options(InlineOptions::default(), true)
+            .await?;
+    }
```
dignifiedquire, line 187: is the above the way I would implement migration in other places, like the node.rs in iroh-ffi?

iroh/src/node.rs
```diff
@@ -470,15 +471,20 @@ where
     }
 
     async fn gc_loop(db: D, ds: S, gc_period: Duration, callbacks: Callbacks) {
+        let mut live = BTreeSet::new();
         tracing::debug!("GC loop starting {:?}", gc_period);
         'outer: loop {
+            if db.gc_start().await.is_err() {
+                tracing::error!("Error starting GC, skipping GC to be safe");
```
dignifiedquire, line 478: shouldn't the error be logged?

iroh/tests/cli.rs
```diff
@@ -150,7 +150,8 @@ fn copy_dir_all(src: impl AsRef<Path>, dst: impl AsRef<Path>) -> anyhow::Result<
     Ok(len)
 }
 
-#[cfg(feature = "flat-db")]
+#[cfg(feature = "redb-db")]
+#[allow(dead_code)]
```
dignifiedquire, line 154: why dead code?

iroh/tests/cli.rs
```diff
@@ -297,51 +228,51 @@ fn cli_provide_tree_resume() -> Result<()> {
...
-        })?;
-        let get = make_get_cmd(&get_iroh_data_dir, &ticket, Some(tgt.clone()));
-        let get_output = get.unchecked().run()?;
-        assert!(get_output.status.success());
-        let matches = explicit_matches(match_get_stderr(get_output.stderr)?);
-        assert_eq!(matches, vec!["65.98 KiB"]);
-        compare_files(&src, &tgt)?;
-        std::fs::remove_dir_all(&tgt)?;
-    }
+    // {
```
dignifiedquire, line 231: what's the plan here?

iroh/tests/gc.rs
```diff
@@ -205,74 +217,84 @@ mod flat {
...
+        let client = node.client();
+        let doc = client.docs.create().await?;
+        let author = client.authors.create().await?;
+        let temp_path = dir.join("temp");
+        std::fs::create_dir_all(&temp_path)?;
+        let mut to_import = Vec::new();
+        for i in 0..10000 {
+            let data = create_test_data(16 * 1024 * 3 + 1);
+            let path = temp_path.join(format!("file{}", i));
+            std::fs::write(&path, &data)?;
```
dignifiedquire, line 258: please use `async` methods if you are inside an async function :)

iroh-bytes/src/store/redb/util.rs
rklaehn: Yeah, will add as soon as I use this one. It is currently unused, will be used in the batch branch.

iroh-bytes/src/store/redb/validate.rs
rklaehn: The log level is not part of the rpc msg. Filtering happens on the client side. I think for something like fsck this is fine. Compared to the other stuff this does it won't matter in terms of perf. And validate is special anyway since it locks the db for an extended period.

iroh-bytes/src/util/sparse_mem_file.rs
rklaehn: No, it can grow. The place where it grows is a bit hidden. In `write_at` I use the WriteAt instance of Vec<u8> that will extend the vec if you write behind the end.

iroh-bytes/src/util/mem_or_file.rs
rklaehn: Will do, unless clippy forces me to revert it...

iroh/src/commands/start.rs
rklaehn: Yes, it's not really migration but import. You just tell it the location of a flat store and it will consume it and integrate it.

Once we are in redb land we would have internal migrations, but for the flat -> redb step it does not make sense because so many things change.

iroh/tests/cli.rs
rklaehn: See above.

iroh/src/commands/start.rs
dignifiedquire: makes sense, would be good to add that maybe to the PR description, so we have it in the changelog/a place to look up what the recommended structure is

iroh-bytes/Cargo.toml
rklaehn: Yeah.

iroh/tests/gc.rs
rklaehn: I hate them but OK.

iroh-bytes/src/store/redb/util.rs
rklaehn: Removed the whole thing.

iroh-bytes/src/store/bao_file.rs
```diff
@@ -0,0 +1,1245 @@
...
+#[derive(Default)]
+pub(crate) struct CompleteMemOrFileStorage {
+    pub data: MemOrFile<Bytes, (File, u64)>,
+    pub outboard: MemOrFile<Bytes, (File, u64)>,
+}
+
+impl Debug for CompleteMemOrFileStorage {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("CompleteMemOrFileStorage")
+            .field("data", &self.data.as_ref().map_mem(|x| x.len()))
```
dignifiedquire, line 93: maybe instead of this manual impl, add a nicer debug impl for `MemOrFile`?

iroh-bytes/src/store/bao_file.rs
```diff
@@ -0,0 +1,1245 @@
...
+    fn default() -> Self {
+        BaoFileStorage::Complete(Default::default())
+    }
+}
+
+impl BaoFileStorage {
+    /// Take the storage out, leaving an empty storage in its place.
+    ///
+    /// Be careful to put somethign back in its place, or you will lose data.
+    #[allow(dead_code)]
```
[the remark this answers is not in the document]
dignifiedquire: thanks :)

iroh-bytes/src/store/file/util.rs
```diff
@@ -0,0 +1,64 @@
+use std::{
+    fs::OpenOptions,
+    io::{self, Write},
+    path::Path,
+};
+
+pub(crate) struct ProgressReader<R, F: Fn(u64) -> io::Result<()>> {
```
dignifiedquire, line 7: small comment what the point of this struct is would be nice

iroh/tests/cli.rs
```diff
@@ -150,7 +150,8 @@ fn copy_dir_all(src: impl AsRef<Path>, dst: impl AsRef<Path>) -> anyhow::Result<
     Ok(len)
 }
 
-#[cfg(feature = "flat-db")]
+#[cfg(feature = "file-db")]
+#[allow(dead_code)]
```
dignifiedquire, line 154: you can drop the allow dead code now?

iroh/tests/gc.rs
```diff
@@ -205,74 +217,84 @@ mod flat {
...
+    async fn redb_doc_import_stress() -> Result<()> {
+        let _ = tracing_subscriber::fmt::try_init();
+        let dir = testdir!();
+        let bao_store = iroh_bytes::store::file::Store::load(dir.join("store")).await?;
+        let node = wrap_in_node(bao_store.clone(), Duration::from_secs(10)).await;
+        let client = node.client();
+        let doc = client.docs.create().await?;
+        let author = client.authors.create().await?;
+        let temp_path = dir.join("temp");
+        std::fs::create_dir_all(&temp_path)?;
```
dignifiedquire, line 253: async

iroh-bytes/src/store/file/util.rs
```diff
@@ -0,0 +1,64 @@
...
+        "overwriting file {} with {} bytes",
+        path.display(),
+        data.len()
+    );
+    // std::fs::create_dir_all(path.parent().unwrap()).unwrap();
+    // tracing::error!("{}", path.parent().unwrap().display());
+    // tracing::error!("{}", path.parent().unwrap().metadata().unwrap().is_dir());
+    let mut file = OpenOptions::new().write(true).create(true).open(path)?;
+    file.write_all(data)?;
+    // todo: figure out the consequences of not syncing here
```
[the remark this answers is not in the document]
rklaehn: Yeah, currently I am syncing. But I would like to avoid it. So I want to figure out if it would be safe not to sync. Will reword.

iroh-bytes/src/store/mem.rs
```diff
@@ -512,155 +208,253 @@ impl super::Store for Store {
...
-        let state = self.0.state.read().unwrap();
-        // a blob is live if it is either in the live set, or it is temp tagged
-        state.live.contains(hash) || state.temp.contains(hash)
+    async fn gc_start(&self) -> io::Result<()> {
+        Ok(())
     }
 
     async fn delete(&self, hashes: Vec<Hash>) -> io::Result<()> {
-        let mut state = self.0.state.write().unwrap();
+        let mut state = self.write_lock();
```
dignifiedquire, line 226: did you check that it is safe to use a sync mutex inside these async calls?

iroh-bytes/src/store/file/validate.rs
```diff
@@ -0,0 +1,469 @@
...
+                        );
+                    }
+                },
+                _ => {
+                    warn!("unexpected file in data directory: {}", path.display());
+                }
+            }
+        }
+        if repair {
+            info!("repairing - removing orphaned files and inline data");
```
dignifiedquire, line 438: would be nice to print some stats at the end, like "removed 18 orphaned files"

iroh-bytes/src/store/bao_file.rs
```diff
@@ -0,0 +1,1238 @@
...
+
+    /// Persist the batch to disk, creating a FileBatch.
+    fn persist(&self, paths: DataPaths) -> io::Result<FileStorage> {
+        let mut data = create_read_write(&paths.data)?;
+        let mut outboard = create_read_write(&paths.outboard)?;
+        let mut sizes = create_read_write(&paths.sizes)?;
+        self.data.persist(&mut data)?;
+        self.outboard.persist(&mut outboard)?;
+        self.sizes.persist(&mut sizes)?;
+        data.sync_all()?;
```
dignifiedquire, line 229: nit: wouldn't it make more sense to call these pair wise `.persist()` followed by the matching `sync_all` call?

iroh-bytes/src/store/bao_file.rs
```diff
@@ -0,0 +1,1238 @@
...
+        batch: &[BaoContentItem],
+    ) -> std::io::Result<()> {
+        let tree = BaoTree::new(ByteNum(size), IROH_BLOCK_SIZE);
+        for item in batch {
+            match item {
+                BaoContentItem::Parent(parent) => {
+                    if let Some(offset) = tree.pre_order_offset(parent.node) {
+                        let o0 = offset
+                            .checked_mul(64)
+                            .expect("u64 overflow multiplying to hash pair offset");
```
dignifiedquire, line 271: why panic, instead of returning an error?

iroh-bytes/src/store/bao_file.rs
```diff
@@ -0,0 +1,1238 @@
...
+                            .write_all_at(o0, parent.pair.0.as_bytes().as_slice())?;
+                        self.outboard
+                            .write_all_at(o0 + 32, parent.pair.1.as_bytes().as_slice())?;
+                    }
+                }
+                BaoContentItem::Leaf(leaf) => {
+                    let o0 = leaf.offset.0;
+                    // divide by chunk size, multiply by 8
+                    let index = (leaf.offset.0 >> (tree.block_size().0 + 10)) << 3;
+                    tracing::debug!(
```
dignifiedquire, line 370: maybe `trace`?

iroh-bytes/src/store/bao_file.rs
```diff
@@ -0,0 +1,1238 @@
...
+
+    /// Create a new mutable mem storage.
+    pub fn incomplete_mem() -> Self {
+        Self::IncompleteMem(Default::default())
+    }
+
+    /// Call sync_all on all the files.
+    fn sync_all(&self) -> io::Result<()> {
+        match self {
+            Self::Complete(_) => Ok(()),
```
dignifiedquire, line 439: why is there no need to propagate the `sync_all` to the potential files in this?

iroh-bytes/src/store/bao_file.rs
```diff
@@ -0,0 +1,1238 @@
...
+}
+
+/// A cheaply cloneable handle to a bao file, including the hash and the configuration.
+#[derive(Debug, Clone)]
+pub struct BaoFileHandle {
+    pub(crate) storage: Arc<RwLock<BaoFileStorage>>,
+    config: Arc<BaoFileConfig>,
+    hash: Hash,
+    #[cfg(test)]
+    pub(crate) id: u64,
```
dignifiedquire, line 472: maybe add a comment for what this id is needed in tests

iroh-bytes/src/store/bao_file.rs
```diff
@@ -0,0 +1,1238 @@
...
+                    // create the paths. This allocates 3 pathbufs, so we do it
+                    // only when we need to.
+                    let paths = self.config.paths(&self.hash);
+                    // *first* switch to file mode, *then* write the batch.
+                    //
+                    // otherwise we might allocate a lot of memory if we get
+                    // a write at the end of a very large file.
+                    let mut file_batch = mem.persist(paths)?;
+                    // if let Some(cb) = self.config.on_file_create.as_ref() {
+                    //     cb(&self.hash)?;
```
dignifiedquire, line 773: what about this?

iroh-bytes/src/store/bao_file.rs
```diff
@@ -0,0 +1,1238 @@
...
+                }
+            }
+            BaoFileStorage::IncompleteFile(file) => {
+                // already in file mode, just write the batch
+                file.write_batch(size, batch)?;
+                Ok(HandleChange::None)
+            }
+            BaoFileStorage::Complete(_) => {
+                // we are complete, so just ignore the write
+                // unless there is a bug, this would just write the exact same data
```
dignifiedquire, line 787: shouldn't this at least print a warning? or is this expected to happen regularly?

iroh-bytes/src/store/bao_file.rs
```diff
@@ -0,0 +1,1238 @@
...
+    }
+
+    async fn load(&mut self, node: TreeNode) -> io::Result<Option<(blake3::Hash, blake3::Hash)>> {
+        let Some(offset) = self.tree.pre_order_offset(node) else {
+            return Ok(None);
+        };
+        let offset = offset * 64;
+        let content = self.data.read_at(offset, 64).await?;
+        Ok(Some(if content.len() != 64 {
+            (blake3::Hash::from([0; 32]), blake3::Hash::from([0; 32]))
```
dignifiedquire, line 866: why is this returning an "empty" hash when the content length is not a multiple of 64?

iroh-bytes/src/store/file/import_flat_store.rs
```diff
@@ -0,0 +1,387 @@
...
+    /// File is storing data for the hash
+    Data(Hash),
+    /// File is storing a partial outboard
+    PartialOutboard(Hash, [u8; 16]),
+    /// File is storing an outboard
+    ///
+    /// We can have multiple files with the same outboard, in case the outboard
+    /// does not contain hashes. But we don't store those outboards.
+    Outboard(Hash),
+    #[allow(dead_code)]
```
dignifiedquire, line 39: why dead code?

iroh/tests/gc.rs
```diff
@@ -450,13 +474,11 @@ mod flat {
 
     ///
     #[tokio::test]
-    async fn gc_flat_stress() -> Result<()> {
+    async fn gc_redb_stress() -> Result<()> {
```
dignifiedquire, line 477: maybe call these `gc_file_` to match the rename?

iroh-bytes/src/store/file/tests.rs
```diff
@@ -0,0 +1,892 @@
+use std::io::Cursor;
+// use std::os::unix::fs::PermissionsExt;
```
dignifiedquire, line 2: delete?

iroh-bytes/src/store/mem.rs
rklaehn: This is the mem version. All the things it does are manipulating in memory set and maps protected by a rwlock. There isn't a single await anywhere. Async mutexes are only needed if the mutex spans an await point, right?

iroh-bytes/src/store/file/tests.rs
```diff
@@ -0,0 +1,892 @@
...
+        };
+        let actual = entry
+            .data_reader()
+            .await
+            .unwrap()
+            .read_to_end()
+            .await
+            .unwrap();
+        assert_eq!(actual, small);
+        drop(tt);
```
dignifiedquire, line 93: why the explicit drop?

iroh-bytes/src/store/file/validate.rs
rklaehn: Yeah, added that in the batch branch.

iroh-bytes/src/store/file/tests.rs
```diff
@@ -0,0 +1,892 @@
...
+        let (_outboard, hash) = raw_outboard(&small);
+        let res = db.get(&hash).await.unwrap();
+        assert_matches!(res, None);
+        let tt = db
+            .import_bytes(small.clone(), BlobFormat::Raw)
+            .await
+            .unwrap();
+        let res = db.get(&hash).await.unwrap();
+        let Some(entry) = res else {
+            panic!("entry not found");
```
dignifiedquire, line 83:
```rust
let entry = res.expect("entry not found");
```

iroh-bytes/src/store/file/tests.rs
```diff
@@ -0,0 +1,892 @@
...
+                    data_location: DataLocation::Owned(SIZE),
+                    ..
+                })
+            );
+        }
+        drop(entry);
+        // // sync so we know the msg sent on drop is processed
+        // db.sync().await.unwrap();
+        // let state = db.entry_state(hash).await.unwrap();
+        // assert_matches!(state.mem, None);
```
dignifiedquire, line 241: what about these?

iroh-bytes/src/store/bao_file.rs
rklaehn: You mean because it will be faster? Yes, might be. You could also do it in parallel, but really I was going to leave optimizations like that for later and just go for hopefully correct code.

iroh-bytes/src/store/file/tests.rs
```diff
@@ -0,0 +1,892 @@
...
+        let expected = EntryState::<Bytes>::Complete {
+            data_location: DataLocation::Owned(SIZE),
+            outboard_location: OutboardLocation::Owned,
+        };
+        assert_eq!(tt.hash(), &hash);
+        assert_eq!(actual.db, Some(expected));
+        assert_eq!(large, std::fs::read(db.owned_data_path(&hash)).unwrap());
+        assert_eq!(
+            outboard,
+            std::fs::read(db.owned_outboard_path(&hash)).unwrap()
```
dignifiedquire, line 300: use to `async` force ;)

iroh-bytes/src/store/bao_file.rs
rklaehn: this error would never happen. Even for a tree with data size u64::max, the offset will be something like data_size / chunk_group_size, so way below u64::max.

Maybe just using checked_mul here is overly paranoid and it should just be * 64.

iroh-bytes/src/store/file/tests.rs
```diff
@@ -0,0 +1,892 @@
...
+    // {
+    //     let path = tempdir.path().join("forbidden.data");
+    //     let data = random_test_data(1024);
+    //     std::fs::write(&path, &data).unwrap();
+    //     std::fs::set_permissions(&path, PermissionsExt::from_mode(0o0)).unwrap();
+    //     let cause = db
+    //         .import_file(path, ImportMode::Copy, BlobFormat::Raw, np())
+    //         .await
+    //         .unwrap_err();
+    //     assert_eq!(cause.kind(), io::ErrorKind::PermissionDenied);
```
dignifiedquire, line 552: why is this commented out?

iroh-bytes/src/store/file/tests.rs
```diff
@@ -0,0 +1,892 @@
...
+    // }
+    drop(tempdir);
+}
+
+// #[cfg(unix)]
```
dignifiedquire, line 557: same with these

iroh-bytes/src/store/bao_file.rs
rklaehn: A complete file is immutable. Both the data and the outboard. So any attempts to write to a complete file are noops that are just there to make things simpler.

iroh-bytes/src/store/bao_file.rs
rklaehn: The place where the cb happens has moved further outside. This can be removed.

iroh-bytes/src/store/file.rs
```diff
@@ -0,0 +1,2483 @@
...
+    ReadOnlyTables, ReadableTables, Tables, BLOBS_TABLE, INLINE_DATA_TABLE, INLINE_OUTBOARD_TABLE,
+};
+
+use super::{
+    bao_file::{raw_outboard_size, BaoFileConfig, BaoFileHandle, CreateCb, DropCb},
+    temp_name, BaoBatchWriter, EntryStatus, ExportMode, ExportProgressCb, ImportMode,
+    ImportProgress, MapEntry, ReadableStore, TempCounterMap, ValidateProgress,
+};
+
+use super::BaoBlobSize;
```
dignifiedquire, line 119: nit: group with other `super` imports

iroh-bytes/src/store/bao_file.rs
rklaehn: If the file is partial, and you read beyond the end of the file, you basically pretend the file is padded with 0 until the final size.

This is harmless since it will just make a part of the tree invalid, which is perfectly fine and expected for a partial outboard.

iroh-bytes/src/store/file.rs
```diff
@@ -0,0 +1,2483 @@
...
+    fn union(self, that: DataLocation<X, u64>) -> io::Result<Self> {
+        Ok(match (self, that) {
+            (DataLocation::External(mut paths, size), DataLocation::External(b_paths, ..)) => {
+                paths.extend(b_paths);
+                paths.sort();
+                paths.dedup();
+                DataLocation::External(paths, size)
+            }
+            (DataLocation::External(_, _), b @ DataLocation::Owned(_)) => b,
+            (a, _b) => a,
```
dignifiedquire, line 147: this seems inconsistent, if I union `Internal` with `External` I get `Internal` but if I call it the otherway around I get `External`

iroh-bytes/src/store/file/import_flat_store.rs
rklaehn: This is copied from the old flat store. Not sure. Probably it can be removed.

iroh-bytes/src/store/file.rs
```diff
@@ -0,0 +1,2483 @@
...
+        match self {
+            DataLocation::Inline(_) => None,
+            DataLocation::Owned(size) => Some(*size),
+            DataLocation::External(_, size) => Some(*size),
+        }
+    }
+}
+
+impl<I, E> DataLocation<I, E> {
+    #[allow(dead_code)]
```
dignifiedquire, line 162: why dead?

iroh-bytes/src/store/file.rs
```diff
@@ -0,0 +1,2483 @@
...
+                paths.sort();
+                paths.dedup();
+                DataLocation::External(paths, size)
+            }
+            (DataLocation::External(_, _), b @ DataLocation::Owned(_)) => b,
+            (a, _b) => a,
+        })
+    }
+
+    #[allow(dead_code)]
```
dignifiedquire, line 151: why dead?

iroh-bytes/src/store/file.rs
```diff
@@ -0,0 +1,2483 @@
...
+            DataLocation::Inline(_) => None,
+            DataLocation::Owned(size) => Some(*size),
+            DataLocation::External(_, size) => Some(*size),
+        }
+    }
+}
+
+impl<I, E> DataLocation<I, E> {
+    #[allow(dead_code)]
+    fn discard_extra_data(&self) -> DataLocation<(), ()> {
```
dignifiedquire, line 163: maybe make this take `self` as well avoiding the `clone`?

iroh-bytes/src/store/file.rs
```diff
@@ -0,0 +1,2483 @@
...
+/// Outboards are implementation specific to the store and as such are always owned.
+///
+/// Only complete outboards can be inlined.
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub(crate) enum OutboardLocation<I = ()> {
+    /// Outboard is in the inline_outboard table.
+    Inline(I),
+    /// Outboard is in the canonical location in the data directory.
+    Owned,
+    /// Outboard is not needed,
```
dignifiedquire, line 191: typo: `,`

iroh-bytes/src/store/file.rs
```diff
@@ -0,0 +1,2483 @@
...
+            {
+                Ok(a)
+            }
+            (Self::Partial { .. }, b @ Self::Complete { .. }) =>
+            // complete wins over partial
+            {
+                Ok(b)
+            }
+            (Self::Partial { size: a_size }, Self::Partial { size: b_size }) =>
+            // keep known size from either entry
```
dignifiedquire, line 267: why not prefer the larger one?

iroh-bytes/src/store/file.rs
```diff
@@ -0,0 +1,2483 @@
...
+                })
+            }
+        }
+    }
+}
+
+impl redb::RedbValue for EntryState {
+    type SelfType<'a> = EntryState;
+
+    type AsBytes<'a> = SmallVec<[u8; 128]>;
```
dignifiedquire, line 280: what's the reason for `128` capacity here?

iroh-bytes/src/store/file.rs
```diff
@@ -0,0 +1,2483 @@
...
+
+impl InlineOptions {
+    /// Do not inline anything, ever.
+    pub const NO_INLINE: Self = Self {
+        max_data_inlined: 0,
+        max_outboard_inlined: 0,
+    };
+    /// Always inline everything
+    pub const ALWAYS_INLINE: Self = Self {
+        max_data_inlined: u64::MAX,
```
dignifiedquire, line 323: I heard someone didn't like sentinel values, so why not make this `Option<u64>` and in this case set it to `None`?

iroh-bytes/src/store/file/tests.rs
```diff
@@ -0,0 +1,892 @@
...
+        assert_eq!(large, std::fs::read(db.owned_data_path(&hash)).unwrap());
+        assert_eq!(
+            outboard,
+            std::fs::read(db.owned_outboard_path(&hash)).unwrap()
+        );
+    }
+    drop(tempdir);
+}
+
+/// Import mem cases, small (data inline, outboard none), mid (data file, outboard inline), large (data file, outboard file)
```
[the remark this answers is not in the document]
rklaehn: They should be completely independent of each other. So I guess I didn't want to have so many different tests.

iroh-bytes/src/store/file.rs
```diff
@@ -0,0 +1,2483 @@
...
+pub struct Entry(BaoFileHandle);
+
+impl super::MapEntry for Entry {
+    fn hash(&self) -> Hash {
+        self.0.hash()
+    }
+
+    fn size(&self) -> BaoBlobSize {
+        let size = self.0.current_size().unwrap();
+        tracing::info!("redb::Entry::size() = {}", size);
```
dignifiedquire, line 415: maybe debug or trace?

iroh-bytes/src/store/file.rs
```diff
@@ -0,0 +1,2483 @@
...
+        BaoBlobSize::new(size, self.is_complete())
+    }
+
+    fn is_complete(&self) -> bool {
+        self.0.is_complete()
+    }
+
+    async fn available_ranges(&self) -> io::Result<bao_tree::ChunkRanges> {
+        todo!()
```
dignifiedquire, line 424: is there an issue/tracking place for this?

iroh-bytes/src/store/file/tests.rs
rklaehn: This was my attempt to cover some of the not so happy cases, but it turned out that using PermissionExt to set some thing readonly did not have the desired effect on windows.

Maybe I can reenable them only on unix or come up with another way to trigger an error.

iroh-bytes/src/store/file.rs
```diff
@@ -0,0 +1,2483 @@
...
+    },
+    /// Internal method: shutdown the actor.
+    Shutdown,
+}
+
+/// Predicate for filtering entries in a redb table.
+pub(crate) type FilterPredicate<K, V> =
+    Box<dyn Fn(u64, AccessGuard<K>, AccessGuard<V>) -> Option<(K, V)> + Send + Sync>;
+
+/// Paramerers for importing from a flat store
```
dignifiedquire, line 598: typo

iroh-bytes/src/store/file.rs
```diff
@@ -0,0 +1,2483 @@
...
+}
+
+#[cfg(test)]
+#[derive(Debug)]
+pub(crate) struct EntryStateResponse {
+    mem: Option<BaoFileHandle>,
+    db: Option<EntryState<Bytes>>,
+}
+
+///
```
dignifiedquire, line 616: sad face

iroh-bytes/src/store/file.rs
```diff
@@ -0,0 +1,2483 @@
...
+    async fn new(path: PathBuf, options: Options) -> io::Result<Self> {
+        // spawn_blocking because StoreInner::new creates directories
+        let rt = tokio::runtime::Handle::try_current()
+            .map_err(|_| io::Error::new(io::ErrorKind::Other, "no tokio runtime"))?;
+        let inner =
+            tokio::task::spawn_blocking(move || StoreInner::new_sync(path, options, rt)).await??;
+        Ok(Self(Arc::new(inner)))
+    }
+
+    ///
```
dignifiedquire, line 641: sad face (some more below)

iroh-bytes/src/store/file.rs
rklaehn: At some point it was useful, now nobody uses it anymore. Poor thing. Will remove.

iroh-bytes/src/store/file.rs
rklaehn: Same as previous.

iroh-bytes/src/store/file.rs
```diff
@@ -0,0 +1,2483 @@
...
+
+    pub async fn complete(&self, hash: Hash) -> OuterResult<()> {
+        self.tx
+            .send_async(ActorMessage::OnComplete { hash })
+            .await?;
+        Ok(())
+    }
+
+    ///
+    #[allow(dead_code)]
```
dignifiedquire, line 900: why dead?

iroh-bytes/src/store/file.rs
rklaehn: It's dead anyway.

iroh-bytes/src/store/file.rs
```diff
@@ -0,0 +1,2483 @@
...
+        self.write().unwrap().inc(content);
+    }
+
+    fn on_drop(&self, content: &HashAndFormat) {
+        self.write().unwrap().dec(content);
+    }
+}
+
+impl StoreInner {
+    pub fn new_sync(
```
dignifiedquire, line 704: why are all the `StoreInner` methods public?

iroh-bytes/src/store/file.rs
rklaehn: Then you would be in so much trouble.

This is the validated size, which you will at some point set once you get the last chunk. This is currently not done at all, but I want to keep it in there so I don't have to perform a migration once I do it.

iroh-bytes/src/store/file.rs
```diff
@@ -0,0 +1,2483 @@
...
+    async fn import_file(
+        &self,
+        path: PathBuf,
+        mode: ImportMode,
+        format: BlobFormat,
+        progress: impl ProgressSender<Msg = ImportProgress> + IdGenerator,
+    ) -> io::Result<(crate::TempTag, u64)> {
+        let this = self.0.clone();
+        Ok(
+            tokio::task::spawn_blocking(move || {
```
dignifiedquire, line 1309: why assume the runtime here, and in earlier code check for a current handle?

iroh-bytes/src/store/file.rs
rklaehn: Just a guess. The idea is that an EntryState that has a DataLocation::External with a typical path should fit in without allocation. If you don't care about external paths you can make this much smaller, since a typical EntryState that does not refer to external paths will be smaller.

iroh-bytes/src/store/file.rs
```diff
@@ -0,0 +1,2483 @@
...
+        let on_file_create: CreateCb = Arc::new(move |hash| {
+            // todo: make the callback allow async
+            tx2.send(ActorMessage::OnMemSizeExceeded { hash: *hash })
+                .ok();
+            Ok(())
+        });
+        let on_drop: DropCb = Arc::new(move |hash, refcount| {
+            // the count includes the one that is being dropped, so we check for 2.
+            // it will be 1 when the last one is being dropped.
+            if refcount == 2 {
```
dignifiedquire, line 1404: wouldn't this still be `1`, and the count is `0` after the drop?

iroh-bytes/src/store/file.rs
```diff
@@ -0,0 +1,2483 @@
...
+                tx3.send(ActorMessage::HandleDropped { hash: *hash }).ok();
```
dignifiedquire, line 1405: should these errors be logged?

iroh-bytes/src/store/file.rs
rklaehn: Hm, this is inclusive, so setting this to u64::MAX actually means inlining everything including the unlikely u64::max sized blob. The downside is that you can't disable inlining the single possible blob that has a size of 0.

iroh-bytes/src/store/file.rs
```diff
@@ -0,0 +1,2483 @@
...
+            ActorMessage::OnComplete { hash } => {
+                let txn = db.begin_write()?;
+                self.on_complete(&mut Tables::new(&txn)?, hash)?;
+                txn.commit()?;
+            }
+            ActorMessage::HandleDropped { hash } => {
+                self.handle_dropped(hash);
+            }
+            ActorMessage::Sync { tx } => {
+                tx.send(()).ok();
```
dignifiedquire, line 2200: nothing to be done at all here?

iroh-bytes/src/store/file.rs
```diff
@@ -0,0 +1,2483 @@
...
+                let path = path_options.owned_data_path(hash);
+                let file = overwrite_and_sync(&path, &data)?;
+                MemOrFile::File((file, data_size))
+            }
+            MemOrFile::File(data) => MemOrFile::File((data, data_size)),
+        }
+    };
+    // inline outboard if needed, or write to file if needed
+    let outboard = if outboard_size == 0 {
+        Default::default()
```
dignifiedquire, line 2455: what's the default?

rklaehn: An empty outboard. There is a default instance for MemOrFile when the Mem part has a default.

iroh-bytes/src/store/file.rs
dignifiedquire: both are validated? eg `a_size: Some(1024)` and `b_size: Some(2048)`, currently you choose the left one, so it depends on the calling order, is that the correct logic?

iroh-bytes/src/store/mem.rs
dignifiedquire: and if the locks are only held for a few ms, as you still end up locking up a whole thread

iroh-bytes/src/store/bao_file.rs
dignifiedquire: I was more thinking about error cases than performance, but I am undecided which ordering in an error case is the better one

iroh-bytes/src/store/bao_file.rs
dignifiedquire: maybe change the doc comment then to sth like `propagate to all mutable files`

iroh-bytes/src/store/bao_file.rs
```diff
@@ -0,0 +1,1238 @@
...
+    pub(crate) storage: Arc<RwLock<BaoFileStorage>>,
+    config: Arc<BaoFileConfig>,
+    hash: Hash,
+    #[cfg(test)]
+    pub(crate) id: u64,
+}
+
+impl Drop for BaoFileHandle {
+    fn drop(&mut self) {
+        if let Some(cb) = self.config.on_drop.as_ref() {
```
Frando, line 477: Maybe let's change this so that the `OnDrop` callback is only called for the last drop? Don't think anyone using this would be interested in the drops before the last drop. Then the check in `file.rs:1458` can be removed, because it will only be called for the last_drop. Then you could use `Arc::get_mut` here to check if the last handle is being dropped, and only then call the `cb`.

iroh-bytes/src/store/file.rs
```diff
@@ -0,0 +1,2483 @@
...
+/// The actor for the redb store.
+///
+/// It is split into the database and the rest of the state to allow for split
+/// borrows in the message handlers.
+#[derive(derive_more::Deref, derive_more::DerefMut)]
+struct Actor {
+    db: redb::Database,
+    #[deref]
+    #[deref_mut]
+    state: ActorState,
```
Frando, line 1122: Not too sure if I like the `deref` here - it caused a little irritation when seeing `self.msgs` or similar on the `impl Actor` - but maybe it's fine also, a comment wouldn't hurt though.

iroh-bytes/src/store/file.rs
rklaehn: No, but it is so far unused, so just gonna remove it for now.

iroh-bytes/src/store/file.rs
rklaehn: Really the code should be

match (a_size, b_size) {
  Some(a), Some(b) if a != b { panic!("bao is broken"); }
}

Will add something along these lines.

iroh-bytes/src/store/file.rs
rklaehn: They used to be directly on Store.

iroh-bytes/src/store/file.rs
rklaehn: We don't want the last drop but the second to last drop. We know that 1 entry is in the entry map, so we want to do the cb if all *other* clones are done.

iroh-bytes/src/store/file.rs
rklaehn: I don't think so. If this fails it means that the actor is dead, and that means that cleaning up the entries is no longer necessary anyway.

You would get that if you manually call shutdown on the store despite it being shared somewhere...

iroh-bytes/src/store/file.rs
rklaehn: Yeah, the main purpose of sync is to just allow code to ensure that all messages before sync have been processed, and that we are not in a transaction.

Since this is not yet the batch branch, we are doing one transaction at a time. So all that remains is to just notify when the sync has been processed.

iroh-bytes/src/store/file.rs
```diff
@@ -0,0 +1,2483 @@
...
+}
+
+impl super::MapEntryMut for Entry {
+    async fn batch_writer(&self) -> io::Result<impl BaoBatchWriter> {
+        Ok(self.0.writer())
+    }
+}
+
+#[derive(derive_more::Debug)]
+pub(crate) struct Import {
```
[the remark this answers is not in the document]
rklaehn: I have a lot of places where I use content for a HashAndFormat because hash_and_format is too verbose and hash is weird because then you need to do `hash.hash()`.

I guess source is a reasonable name for ImportSource.

At some point I called the HashAndFormat thing `cid`, but then people would hate me because that is overloaded in ipfs land...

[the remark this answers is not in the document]
rklaehn: Ok, how about `content_id`?

iroh-bytes/src/store/bao_file.rs
rklaehn: The issue is that we are interested in the second to last drop, because one entry is in the entry map in the actor. And just firing the cb on the second to last drop seems arbitrary.

iroh-bytes/src/store/file.rs
rklaehn: why?

external/external -> 
external/owned -> owned (b)
owned/owned -> owned (a)
owned/external -> owned (a)

anyway, spelled it out a bit clearer now.

dignifiedquire:
```
(external, internal) => external
(internal, external) => internal
```

iroh-bytes/src/store/file.rs
```diff
@@ -1602,10 +1602,16 @@ impl ActorState {
                             })?
                             .to_owned();
                         // we can not reference external files, so we just copy them. But this does not have to happen in the actor.
-                        self.rt.spawn_blocking(move || {
-                            tx.send(export_file_copy(temp_tag, path, size, target, progress))
-                                .ok();
-                        });
+                        if path == target {
```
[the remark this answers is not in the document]
rklaehn: Then you would break it as well. But that's on you then. When using external files the user promises to keep the files unchanged.

If you do
`export <hash1> foo.bin`
`export <hash2> foo.bin`

you are basically violating that promise.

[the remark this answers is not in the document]
ppodolsky: I know it is a very poor man solution, but may it worth to add nested size comparision here to detect 99.99% of such cases and add warning?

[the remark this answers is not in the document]
rklaehn: no, there are a million ways how you can modify files and void the contract. This would just avoid one, and only in some cases.

what you could do is to check if the target is in the set of all external files and in that case refuse. It would be costly though, and it would still not help with symlinks and hardliks.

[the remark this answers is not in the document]
pasha-pplx: I meant cases when you export file to existing file path, not all possible cases. I can easily imagine cases when people do something like `iroh export-file <doc>/<very-complicated-key-name>.zip ~/archive.zip` with different files and warning would cover most of such cases.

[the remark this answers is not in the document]
ppodolsky: I meant cases when you export file to existing file path, not all possible cases. I can easily imagine cases when people do something like iroh export-file <doc>/<very-complicated-key-name>.zip ~/archive.zip with different files and warning would cover most of such cases.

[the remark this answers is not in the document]
rklaehn: How about just a warning if you are going to override something? Would make sense in any case, and when you use it from code you would not pay for it.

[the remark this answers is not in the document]
ppodolsky: Yep, warning would be nice

iroh-bytes/src/store/file/test_support.rs
```diff
@@ -0,0 +1,399 @@
...
+        outboard: Vec<u8>,
+        /// Sizes
+        #[debug("sizes")]
+        sizes: Vec<u8>,
+    },
+}
+
+impl Store {
+    /// Get the complete state of an entry, both in memory and in redb.
+    #[cfg(test)]
```
dignifiedquire, line 54: why not make the whole module `cfg[test]`?

rklaehn: It is used from the cli tests, which are not in the same crate.

dignifiedquire (approved):

PR #2052, opened 2024-03-01
(WIP) feat: Publish self-announces as pkarr signed packets through the derper

Frando:
## Description

**Work in progress**

build upon #2045 

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


ramfox:
Did some dogfooding!

used a local iroh-relay to test w/ the following config:
```
# iroh-relay-config.toml
pkarr_relay = "https://testdns.iroh.link/pkarr"
addr = "[::]:3340"
enable_relay = true
enable_stun = true
stun_port = 3478
hostname = "my.relay.network"
```

Works with two local nodes dialing by `node_id`!

Can also confirm that if the relay server is not "pkarr_relay" enabled, the normal relay interactions still work, there is no problem when sending a `PkarrPacket` from the node to the relay server.

Frando:
I'm closing this for now. We can revive if we ever decide to do this, but for now this is not planned.

PR #2053, opened 2024-03-01
refactor: move `iroh_net` base types to `iroh_base`

Frando:
## Description

This moves the basic types from iroh_net to iroh_base, so that crates do not have to depend on iroh_net if they just need to interface with its basic types.

It moves the following:
* `iroh_net::key` module with the `NodeId`/`PublicKey` and `SecretKey` types, plus the crypto dependencies for those types. Because this adds weight to iroh_base they are behind a non-default `key` feature there.
* `iroh_net::ticket::{BlobTicket, NodeTicket}` - they were only in iroh_net because we couldn't move them because of their dependendy on `NodeId`
* `iroh_net::AddrInfo` and `NodeAddr` - are used by the tickets, and also very simple types

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


dignifiedquire (approved):

iroh-base/Cargo.toml
```diff
@@ -26,6 +26,19 @@ serde = { version = "1", features = ["derive"] }
...
+derive_more = { version = "1.0.0-beta.1", features = ["debug", "display"], optional = true }
+ed25519-dalek = { version = "2.0.0", features = ["serde", "rand_core"], optional = true }
+once_cell = { version = "1.18.0", optional = true }
+rand = { version = "0.8", optional = true }
+rand_core = { version = "0.6.4", optional = true }
+ssh-key = { version = "0.6.0", features = ["ed25519", "std", "rand_core"], optional = true }
+ttl_cache = { version = "0.5.1", optional = true }
+crypto_box = { version = "0.9.1", features = ["serde", "chacha20"], optional = true }
+zeroize = { version = "1.5", optional = true }
+url = { version = "2.5.0", features = ["serde"], optional = true }
```
flub, line 40: pretty minor, but i'm not a fan of keeping things in blocks like this.  they start to overlap, get outdated etc.  in the long run a single alphabetically-sorted list of dependencies is almost always nicer (and `cargo add` will keep it sorted, which it can't like this).

Frando: let's do a followup then that unifies this in the iroh codebase, we have those blocks in a couple of Cargo.tomls in the repo I think.

PR #2054, opened 2024-03-02
feat(iroh): expose GetSyncPeers

dignifiedquire:


rklaehn (approved):
I don't see anything wrong with this, other than the fact that it adds yet another case to the giant list of RPC methods...

PR #2055, opened 2024-03-04
feat(sync): allow asynchronously retrieving the content status of entries

### LINKED ISSUE #2047: sync actor: make `ContentStatusCallback` async
This is a prerequisite to moving the store to its own actor
NOTE: I might update the issue as my understanding of the problem progresses

divagant-martian:
## Description

Fixes #2047 and removes `entry_status_sync` from the blob store, which was the remaining ToDo after #2043 

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.


iroh-sync/src/actor.rs
```diff
@@ -219,13 +220,28 @@ impl SyncHandle {
             let span = error_span!("sync", %me);
             let _enter = span.enter();
 
-            if let Err(err) = actor.run() {
-                error!("Sync actor closed with error: {err:?}");
-            }
+            rt.block_on(async move {
+                if let Err(err) = actor.run().await {
```
dignifiedquire, line 224: This puts the whole actor loop (which uses a bunch of blocking things internally asfaik) inside the tokio runtime. Why not pass the runtime to the point where we actually need to wait for the future, and just block at that point?

divagant-martian: gave this more thought and I think you are right, since the actor run in it's own thread, this should wither be a  spawn blocking, which puts the blocking code in the executor anyway, or blocking only on the calls to the replica that require it.

Frando: With #2051 the iroh-bytes store will also run with a actor in a std::sync thread.
so maybe using a flume channel for the entry status callback actually isn't a hack but the proper solution?
Replicas already perform sync IO, so a sync send+recv on a flume channel isn't any different I'd say.
So the runtime wouldn't be needed at all, the replica methods could stay sync and the `EntryContentStatus`  callback would call a new method on the baostore like `entry_status_channel(hash: &Hash) -> oneshot::Receiver<()>` and calling `blocking_recv` on the receiver.

cc @rklaehn

PR #2056, opened 2024-03-04
feat(iroh-net): combine discovery services and add heuristics when to start discovery

Frando:
## Description

* Changes the `resolve` method of the discovery trait to return a `Stream` of `DiscoveryItems`
* Adds a `CombinedDiscovery` to combine multiple discovery services
* Always uses the discovery service when connecting to a node to which we don't have any address yet
* Start the quinn connect as soon as a first discovery result came in, but continue the discovery until we have a successful connection
* Move the discovery trait from module `magicsock` to new module `discovery`

## Notes & open questions

* Should we also attempt a discovery if we have *some* address already?
* Add tests

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh-net/src/discovery.rs
```diff
@@ -0,0 +1,93 @@
...
+pub trait Discovery: std::fmt::Debug + Send + Sync {
+    /// Publish the given [`AddrInfo`] to the discovery mechanisms.
+    ///
+    /// This is fire and forget, since the magicsock can not wait for successful
+    /// publishing. If publishing is async, the implementation should start it's
+    /// own task.
+    ///
+    /// This will be called from a tokio task, so it is safe to spawn new tasks.
+    /// These tasks will be run on the runtime of the [`super::MagicEndpoint`].
+    fn publish(&self, _info: &AddrInfo) {}
```
dignifiedquire, line 26: why are these not `async`?

iroh-net/src/discovery.rs
```diff
@@ -0,0 +1,93 @@
...
+    ///
+    /// Must be microseconds since the unix epoch.
+    pub last_updated: Option<u64>,
+    /// The adress info for the node being resolved.
+    pub addr_info: AddrInfo,
+}
+
+/// A discovery service that combines multiple discovery sources.
+#[derive(Debug, Default)]
+pub struct CombinedDiscovery {
```
dignifiedquire, line 58: would be nice to have an `From<Iterator>` impl, so I can write

```rust
let discovery = CombinedDiscovery::from([Box::new(Service1), Box::new(Service2)]);
```

iroh-net/src/discovery.rs
rklaehn: I can answer this at least for the original discovery trait.

They are intended to be "fire and forget". At the site where you call these you do not want to wait for them to complete. E.g. if we publish to a HTTP PUT endpoint we don't want to wait for successful publish before continuing, since that would probably create havoc inside the magicsock and in any case not help anybody.

E.g. in case of the pkarr discovery, this just puts the record somewhere and starts a loop to publish and republish to the DHT. Publishing to the DHT can take seconds, you don't want to wait for that. And what are you going to do if it fails? E.g. you are offline.

dignifiedquire: well from a task management perspective I would have expected a handling more like this

```rust
// in the magicsock
let publish_task = tokio::task::spawn(discovery.publish());
// track task or drop depending on the management
```

because this way you can eg cancel things on shutdown

rklaehn: The task is owned by the discovery service, so it's the job of the discovery service to cancel the task on drop. I think there is not necessarily is a 1:1 relation between tasks that need to be spawned and publish calls.

E.g. in the pkarr discovery there is just a single long lived republish task owned by the discovery, and calling publish just replaces the value to be published with the new one.

dignifiedquire: how would I communicate that I want the publish to end then? there is no shutdown method or anything like that if this is service is long running

Frando: Should we maybe pass `CancellationToken`s into both `resolve` and `publish`? This would give full flexibility to both. Downside is that implementors would have to use `tokio::select!` likely. Still, even for the resolve case it could make implementations simpler. Depending on how a resolver works, detecting that the returned stream was dropped may not be straightforward or instant. E.g. with channels, you might only realize once you try to sent, and so you might have been doing uneeded work already up to that point.

Frando:
I pushed tests.

Frando:
I added more docs and tests, and implemented a `DiscoveryTask` with the heuristics as discussed in Discord:

* connect called with NodeAddr
* If NodeAddr contains direct_addrs or derp_url: Add to magicsock
* If no direct_addrs or derp_url provided, and no info in magicsock, start discovery and wait for first result
* If we haven't received recently on any path from the node, start task **in background**:
  * Check: Did we receive recently? If so, done.
  * If user provided new addresses, wait for a delay of 500ms
  * Check again: Did we receive recently? If so - done. If not - start discovery.
* start connect_quinn. This starts *in parallel* to the discovery (maybe) still running.
* Quinn will retry automatically for up to 10s in our config.
* Abort discovery task once a connection is made

This is ready to review now. I will rebase the DNS discovery on top of this also I think.

iroh-net/src/discovery.rs
Frando: Or we could add an optional `fn shutdown()` to the discovery trait which will be called in `MagicEndpoint::close` and can be used by the discovery service to abort all tasks it started.

dignifiedquire: I like the cancellation token approach

iroh-net/src/discovery.rs
```diff
@@ -0,0 +1,536 @@
...
+pub(super) struct DiscoveryTask {
+    on_first_rx: oneshot::Receiver<Result<()>>,
+    task: JoinHandle<()>,
+}
+
+impl DiscoveryTask {
+    /// Start a discovery task.
+    pub fn start(ep: MagicEndpoint, node_id: NodeId) -> Result<Self> {
+        if ep.discovery().is_none() {
+            bail!("No discovery services configured");
```
dignifiedquire, line 124: could use ensure

dignifiedquire (approved):
some nits but overall nice improvements

iroh-net/src/discovery.rs
flub: When I saw the doc comment here I was wondering why the signature does not give you a handle to the runtime to spawn instead of having it just in the documentation that "tokio spawn is fine".

Given the cancellation discussion, maybe it could even be a `JoinSet` that you're politely asked to spawn into, and then you get aborting for free when the magicsock drops the joinset.

iroh-net/src/discovery.rs
```diff
@@ -0,0 +1,536 @@
...
+    pub provenance: &'static str,
+    /// Optional timestamp when this node address info was last updated.
+    ///
+    /// Must be microseconds since the unix epoch.
+    pub last_updated: Option<u64>,
+    /// The adress info for the node being resolved.
+    pub addr_info: AddrInfo,
+}
+
+/// A discovery service that combines multiple discovery sources.
```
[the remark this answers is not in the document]
flub: maybe even call it `ConcurrentDiscovery`?

iroh-net/src/discovery.rs
Frando: Leaving this for a followup and created an issue here: https://github.com/n0-computer/iroh/issues/2066

PR #2057, opened 2024-03-06
feat(iroh-net): ICMPv6 probe support in netcheck

flub:
## Description

Now that we support IPv6 it becomes worth it to add an ICMPv6 probe.
This adds the probe, the two main features are:
- latency from the probe is used for the derper selection.
- the information is used in network change detection.

Note that this also tweaks the definition of this for network change
detection a bit.  If one of the netcheck reports did not include any
information about ICMP availability then it is ignored in the
comparison.  This makes sense because the report stops before ICMP
probes are run if things are decent, so just because in one report
some STUN UDP packets got lost and the ICMP probe did get to run
does not mean anything changed.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


iroh-net/src/config.rs
```diff
@@ -63,8 +63,11 @@ pub struct NetInfo {
     /// Whether the host has UDP internet connectivity.
     pub working_udp: Option<bool>,
 
-    /// Whether ICMPv4 works. Empty means not checked.
-    pub working_icm_pv4: Option<bool>,
+    /// Whether ICMPv4 works, empty means not checked.
+    pub working_icmp_v4: Option<bool>,
+
+    /// Whether ICMPv6 works, empty means not checked.
```
dignifiedquire, line 69: maybe `None means not checked`

iroh-net/src/netcheck/reportgen.rs
```diff
@@ -893,22 +780,109 @@ async fn run_probe(
         }
     }
 
-    trace!(probe = ?probe, "probe successful");
+    trace!("probe successful");
     Ok(result)
 }
 
+/// Run a STUN IPv4 or IPv6 probe.
+async fn run_stun_probe(
```
dignifiedquire, line 788: maybe add a debug_assert or such to make sure that `derp_addr` and `Probe` match in respect to ipv4/ipv6

iroh-net/src/netcheck/reportgen.rs
```diff
@@ -1053,3 +1042,283 @@ async fn measure_https_latency(_node: &DerpNode) -> Result<(Duration, IpAddr)> {
...
+                        report.mapping_varies_by_dest_ip = Some(false);
+                    }
+                }
+                Some(ipp @ SocketAddr::V6(_)) => {
+                    report.ipv6 = true;
+                    report
+                        .derp_v6_latency
+                        .update_derp(derp_node.url.clone(), latency);
+                    report.global_v6 = Some(ipp);
+                    // TODO: Should we track mapping_varies_by_dest_ip for IPv6 too?  Would
```
dignifiedquire, line 1080: would be interesting to track, and print a big log message if it happens, I am really curious if this happens in practice

iroh-net/src/netcheck/reportgen.rs
flub: 55499cbd

iroh-net/src/ping.rs
```diff
@@ -17,32 +33,59 @@ impl Debug for Inner {
...
+    }
+
+    /// Lazily create the ping client.
+    ///
+    /// We do this because it means we do not bind a socket until we really try to send a
+    /// ping.  It makes it more transparent to use the pinger.
+    fn get_client(&self, kind: ICMP) -> Result<Client> {
+        let client = match kind {
+            ICMP::V4 => {
+                let mut opt_client = self.0.client_v4.lock().unwrap();
```
dignifiedquire, line 57: is it a good idea to use a sync mutex inside the async call stack from netcheck?

flub: This is without an await while holding the lock so the way tokio recommends it AFAIK.

The only other places this lock is held is in the other probes running at the same time.  In practice this will be the same probe for a different derper.  I think they're fine with blocking on each other (they run as a tokio TaskSet currently, though used to be FuturesUnordered before we figured out that panicked - in either case I still think that's fine).

iroh-net/src/netcheck/reportgen.rs
flub: 14ada614

Yeah, why not.  This was an old todo, possibly even copied from tailscale.  I assumed it meant a lot of changes, but it really isn't so added it.

iroh-net/src/ping.rs
dignifiedquire: Well the taskset doesn't stop the cpu core to be blocked, when a mutex is held, so I don't think this will work out (FuturesUnordered would have the same issues)

flub: I'm not sure I follow.  Two cpu cores get to this mutex, one acquires it and one blocks for a short time after which it acquires it as well.  I don't think this does any harm for the tiny amount of non-blocking code bing run while holding the lock.

flub:
This is finally ready for review.

iroh-net/src/ping.rs
flub: If both tasks are running on the same core nothing will ever block I think.  Both tasks would get here, find the lock unlocked, lock, create or get the client, and unlock and carry on.  It would only be an issue if there's a lock held across an await point.

Is there another easy and better way to lazily create these clients that I am missing?

dignifiedquire: you are right, sorry 😅 this stuff is hard

dignifiedquire (approved):

PR #2058, opened 2024-03-06
refactor(iroh-net): remove unneeded async interactions with the magicsock actor

Frando:
## Description

While working on #2056 I spotted that we use the actor inbox with return channels for information that is readily available already on the shared inner magicsock. This removes the unneeded complexity and thus simplifies `get_mapping_addr`, `endpoint_info` and `endpoint_infos` to return the information non-async and infallible. Yay!

## Change checklist

- [x] Self-review.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2429,6 +2429,7 @@ dependencies = [
  "bytes",
  "clap",
  "criterion",
+ "crypto_box",
  "curve25519-dalek",
  "data-encoding",
  "default-net",
--- a/iroh-net/src/magicsock.rs
+++ b/iroh-net/src/magicsock.rs
@@ -1251,25 +1251,13 @@ impl MagicSock {
     }
 
     /// Retrieve connection information about nodes in the network.
-    pub async fn tracked_endpoints(&self) -> Result<Vec<EndpointInfo>> {
-        let (s, r) = sync::oneshot::channel();
-        self.inner
-            .actor_sender
-            .send(ActorMessage::TrackedEndpoints(s))
-            .await?;
-        let res = r.await?;
-        Ok(res)
+    pub fn tracked_endpoints(&self) -> Vec<EndpointInfo> {
+        self.inner.node_map.endpoint_infos(Instant::now())
     }
 
     /// Retrieve connection information about a node in the network.
-    pub async fn tracked_endpoint(&self, node_key: PublicKey) -> Result<Option<EndpointInfo>> {
-        let (s, r) = sync::oneshot::channel();
-        self.inner
-            .actor_sender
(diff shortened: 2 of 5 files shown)


iroh-net/src/magicsock.rs
```diff
@@ -1319,18 +1307,11 @@ impl MagicSock {
...
-            .inner
-            .actor_sender
-            .send(ActorMessage::GetMappingAddr(*node_key, s))
-            .await
-            .is_ok()
-        {
-            return r.await.ok().flatten().map(|m| m.0);
-        }
-        None
+    pub fn get_mapping_addr(&self, node_key: &PublicKey) -> Option<SocketAddr> {
```
rklaehn, line 1310: The new code gets the mapping addr immediately, the old code waits until all msgs currently in the queue are processed.

I guess this does not matter?

iroh/src/node.rs
```diff
@@ -1441,31 +1441,25 @@ impl<D: BaoStore> RpcHandler<D> {
...
         });
         rx.into_stream()
     }
 
-    async fn node_connection_info(
+    fn node_connection_info(
         self,
         req: NodeConnectionInfoRequest,
-    ) -> RpcResult<NodeConnectionInfoResponse> {
+    ) -> impl Future<Output = RpcResult<NodeConnectionInfoResponse>> {
```
dignifiedquire, line 1459: why the change from `async` to `impl Future`?

iroh-net/src/magicsock.rs
Frando: Ah, that is a good question. Let's see what tests say. The relevant interaction though is `add_node_addr`, which also is sync. So I think it should be fine.

iroh/src/node.rs
Frando: Because otherwise clippy says `Warning: unused `async` for function with no await statements`. Also maybe it makes it obvious that there is no await. Can also add a clippy allow, if preferred? Don't mind.

Frando: Changed it to return `std::future::Ready` directly.

Cargo.lock
```diff
@@ -2429,6 +2429,7 @@ dependencies = [
  "bytes",
  "clap",
  "criterion",
+ "crypto_box",
```
flub, line 2432: why is this changing?

flub (approved):

Cargo.lock
Frando: I don't know?

Frando: Maybe current main has this wrong after #2053

rklaehn (approved):

PR #2060, opened 2024-03-06
refactor(iroh-net): Clean up peer_map, node_map and endpoint names

flub:
## Description

This cleans up a bunch of naming:

- It documents how we decided to use NodeId in #1388.

- It fixes up a few more of those PublicKey uses to use NodeId.

- It renames the peer_map module to  node_map, the primary data
  structure in there is the NodeMap as we migrated away from naming
  things peer a while ago as well.

- It tries to carve out the "Endpoint" naming for a synonym of a
  "Node".

- As part of that it uses "Path" as meaning "one way of connecting to
  a node".  Either via a direct IP address or via a derper.

## Change checklist

- [ ] Self-review.


iroh-net/src/lib.rs
```diff
@@ -31,6 +31,17 @@ pub mod util;
...
+///
+/// Each [`MagicEndpoint`] is uniquely identified by this.
+///
+/// This is equivalent to [`iroh_base::key::PublicKey`].  By convention we will (or should)
+/// use `PublicKey` as type name when performing cryptographic operations, but use `NodeId`
+/// when referencing a node.  E.g.:
+///
+/// - `encrypt(key: PublicKey)`
+/// - `send_to(node: NodeId)`
 pub use iroh_base::key::PublicKey as NodeId;
```
Frando, line 45: Can you move the type alias, including the docs, to `iroh_base` and pub use it here?

It was an oversight on my side in # to not move the type alias as well.

divagant-martian (approved):

iroh-net/src/lib.rs
flub: done

PR #2064, opened 2024-03-08
tests(iroh-net): Make test more reliable

flub:
## Description

This refactors this test which has had reports of being flaky a bit.
It tiedies up some logging but most significantly:

- The server task is now aborted properly regardless of how the test
  execution works out.

- The clients are no longer moved to a task.  The server is
  handling them sequentially anyway so there's not much to gain from
  having one of the clients sit around waiting to be accepted.

- Most importantly, and possibly the relevant bugfix, both clients
  were using the same NodeId.  This *may* just about work out if both
  are not connecting at the same time as it would maybe look like a
  migrating client to the server.  But combined with the clients
  before being spawned in tasks this would have confused the server's
  PathState quite a lot.  I don't believe these clients should have
  shared their NodeId.

- A (generous) timeout is added to the test.  The flake issue mentions
  that this is sometimes just very slow.  There really is no reason
  for this to be slow so consider these failures so that we get
  debugging info from them in the form of logs.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,0 @@
-[build]
-rustflags = []
--- a/iroh-net/src/magic_endpoint.rs
+++ b/iroh-net/src/magic_endpoint.rs
@@ -535,6 +535,7 @@ mod tests {
 
     use std::time::Instant;
 
+    use iroh_test::CallOnDrop;
     use rand_core::SeedableRng;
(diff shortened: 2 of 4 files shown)


dignifiedquire (approved):
code looks good, lets hope it helps

PR #2067, opened 2024-03-11
fix: properly shutdown sync actor

Frando:
## Description

Make sure that we join the sync actor thread once the last handle is dropped.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-sync/src/actor.rs
+++ b/iroh-sync/src/actor.rs
@@ -4,6 +4,7 @@ use std::{
     collections::{hash_map, HashMap},
     num::NonZeroU64,
     sync::Arc,
+    thread::JoinHandle,
 };
 
 use anyhow::{anyhow, Context, Result};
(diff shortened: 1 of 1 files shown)


rklaehn:
Not for this PR, but we should document the tradeoff of calling shutdown explicitly vs. relying on the last drop, for both stores.

iroh-sync/src/actor.rs
```diff
@@ -485,6 +491,19 @@ impl SyncHandle {
     }
 }
 
+impl Drop for SyncHandle {
+    fn drop(&mut self) {
+        // this means we're dropping the last reference
+        if let Some(handle) = Arc::get_mut(&mut self.join_handle) {
```
rklaehn, line 497: Ah, this is nicer than relying on the refcount...

rklaehn (approved):
Can you name the thread while you're at it?

dignifiedquire (approved):

PR #2075, opened 2024-03-13
fix(iroh-net): Work around broken windows DNS configuration

### LINKED ISSUE #2069: test_icmp_probe_eu_derper flaky on windows
It seems that on windows sometimes we can't do this probe. The derper looks healthy, and if it was down all the other tests would fail as well. We have seen this not just in tests but also in real life.

Possibly related: https://github.com/n0-computer/dumbpipe/issues/17

image

flub:
## Description

This actively refuses to use the `fec0:0:0:ffff::1`, `fec0:0:0:ffff::2` and `fec0:0:0:ffff::3` DNS servers if the system has them configured.

Windows by default adds 3 IPv6 site-local anycast addresses to the DNS servers: `fec0:0:0:ffff::1`, `fec0:0:0:ffff::2` and `fec0:0:0:ffff::3`.  Supposedly Microsoft DNS servers by default listen on those.  These are present as soon as there's an IPv6 interface configured it seems, even for a loopback interface which is extremely common if not the default.

Our hickory-resolver loads the system configuration, which includes these 3 IPv6 DNS servers.  When it needs to make a DNS query it selects a random nameserver and tries this.  If that fails it will try another one.  For the next query there is bias, it will remember which servers to avoid or use.  So if you get lucky and your first query falls on an actual DNS server then you are good.  If you get unlucky recovering is a bit of a tussle because:

Inside netcheck we do DNS queries with a 1s timeout, this because all the probes have a 3s timeout.  However hickory-resolver has a 5s timeout configured, so it's queries stay alive longer than ours.  This means almost all subsequent DNS queries will end up reusing an existing connection to one of those bad servers if you are unlucky to land on one.  The interplay of these timeouts and the connection reuse make recovering to a good DNS server a rather tough prospect for netcheck.  It probably would eventually, given enough netcheck runs (which run at intervals of ~30s).

The odds of these nameservers being the sole way of having working DNS is basically zero.  The odds of these nameservers breaking the resolver are about 50%.  So remove these deprecated things.

## Notes & open questions

Unfortunately the resolver returned by `get_resolver()` does not have an API that allows to test it.  But the test would basically be the inverse logic of the logic that removes the bad servers so perhaps not that useful anyway.

Closes #2069 
Closes https://github.com/n0-computer/dumbpipe/issues/17

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/src/dns.rs
+++ b/iroh-net/src/dns.rs
@@ -41,7 +41,7 @@ fn get_resolver() -> Result<TokioAsyncResolver> {
         config.add_search(name.clone());
     }
     for nameserver_cfg in system_config.name_servers() {
-        if WINDOWS_BAD_SITE_LOCAL_DNS_SERVERS.contains(&nameserver_cfg.socket_addr.ip()) {
+        if !WINDOWS_BAD_SITE_LOCAL_DNS_SERVERS.contains(&nameserver_cfg.socket_addr.ip()) {
             config.add_name_server(nameserver_cfg.clone());
         }
     }
--- a/iroh-net/src/dns.rs
+++ b/iroh-net/src/dns.rs
@@ -103,7 +103,7 @@ mod tests {
 
     #[test]
     fn test_get_resolver() {
-        let resolver = get_resolver();
+        let _resolver = get_resolver();
         panic!("boom");
     }
 
--- a/iroh-net/src/dns.rs
+++ b/iroh-net/src/dns.rs
@@ -102,7 +102,7 @@ mod tests {
     use super::*;
 
     #[test]
-    async fn test_get_resolver() {
+    fn test_get_resolver() {
         let resolver = get_resolver();
         panic!("boom");
     }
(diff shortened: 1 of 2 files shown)


iroh-net/src/dns.rs
```diff
@@ -41,7 +41,7 @@ fn get_resolver() -> Result<TokioAsyncResolver> {
         config.add_search(name.clone());
     }
     for nameserver_cfg in system_config.name_servers() {
-        if WINDOWS_BAD_SITE_LOCAL_DNS_SERVERS.contains(&nameserver_cfg.socket_addr.ip()) {
+        if !WINDOWS_BAD_SITE_LOCAL_DNS_SERVERS.contains(&nameserver_cfg.socket_addr.ip()) {
```
dignifiedquire, line 44: 💙

flub: :sweat: :sob: :sunglasses:

dignifiedquire (approved):
https://www.destroyallsoftware.com/talks/wat

PR #2076, opened 2024-03-13
feat: split CLI implementation into a new iroh-cli crate

dignifiedquire:


iroh-cli/src/commands/rpc.rs
```diff
@@ -36,7 +36,7 @@ pub enum RpcCommands {
     /// Manage blobs
     ///
     /// Blobs are immutable, opaque chunks of arbitrary-sized data.
-    /// For more on blobs see https://iroh.computer/docs/layers/blobs
+    /// For more on blobs see <https://iroh.computer/docs/layers/blobs>
```
dignifiedquire, line 39: @flub no please revert, this is the help that gets printed on the CLI, we don't want angle brackets there

flub: oops, I misunderstood.  reverted the commit now

dignifiedquire: thanks

PR #2078, opened 2024-03-13
refactor(iroh): move rpc handling into its own module

dignifiedquire:
And some more minor node cleanups.


iroh/examples/rpc.rs
```diff
@@ -35,10 +35,16 @@ fn make_rpc_endpoint(
     secret_key: &SecretKey,
 ) -> anyhow::Result<impl ServiceEndpoint<ProviderService>> {
     let rpc_addr = SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::LOCALHOST, DEFAULT_RPC_PORT));
-    let rpc_quinn_endpoint = quinn::Endpoint::server(
-        iroh::node::make_server_config(secret_key, 8, 1024, vec![RPC_ALPN.to_vec()])?,
-        rpc_addr,
+    let mut transport_config = quinn::TransportConfig::default();
+    transport_config.max_concurrent_bidi_streams(8u64.try_into().unwrap());
```
flub, line 39: no unwraps please.  You can write this as `8u32.into()` I think?  If you really, really must use `.expect()`.

iroh/src/commands/start.rs
```diff
@@ -256,12 +256,17 @@ async fn make_rpc_endpoint(
     iroh_data_root: &Path,
 ) -> Result<impl ServiceEndpoint<ProviderService>> {
     let rpc_addr = SocketAddrV4::new(Ipv4Addr::LOCALHOST, rpc_port);
-    let server_config = iroh::node::make_server_config(
+    let mut transport_config = quinn::TransportConfig::default();
+    transport_config
+        .max_concurrent_bidi_streams(MAX_RPC_STREAMS.try_into().unwrap())
```
flub, line 261: Again, quinn `VarInt` has `From<u8>`, `From<u16>`, and `From<u32>`.  Let's change the type of the constant to avoid the unwrap.

dignifiedquire: side note, horrible api on quinns side..

iroh/examples/rpc.rs
dignifiedquire: fixed

b5 (approved):

PR #2079, opened 2024-03-14
refactor: remove derp meshing

### LINKED ISSUE #2074: Remove unused relay meshing code
We are removing parts of the DERP protocol we do not use.

If we need to reimplement load-balancing to our relay servers, we will likely have that be part of the relay server implementation, rather than part of the DERP protocol itself.

ramfox:
## Description

Since the advent of the `DerpUrl`, we no longer use any meshing. We are currently on a mission to remove unused and unnecessary code, so the derper meshing has to go!

closes #2074

## Notes

We have removed three message types from the protocol:
`PeerPresent`, `ClosePeer`, and `WatchConns`

These messages were only used to communicate between meshed derpers. Now, if a server recieves on of those messages from a client, the server will error on that connection.

This also removes the `PacketForwarder` trait, since derp nodes will no longer act as packet forwarders for each other, they will only ever relay packets between clients.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


rklaehn:
Red is a beautiful colour...

dignifiedquire (approved):
nice red code

PR #2082, opened 2024-03-14
feat(iroh-net): Use the local endpoints info when closing derps

flub:
## Description

When the network has changed we used to close all the derp connections
unconditionally.  This passes in the local endpoints information so
that the derp http actor will only close those derp connections for
which we no longer have the local endpoint.  That's a lot less
disruptive than always closing them.

## Notes & open questions

The relevant rebinding test is still marked flaky.  I've only seen it
pass but will monitor in the flaky tests before considering enabling.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


dignifiedquire (approved):

flub:
#2086 for flaky tests

PR #2083, opened 2024-03-14
refactor(iroh-net): remove rebinding

flub:
We still handle major network changes, but we do not pretend to handle rebinding anymore.

There are a few reasons and implications that I'll write up properly tomorrow.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/src/magicsock.rs
+++ b/iroh-net/src/magicsock.rs
@@ -94,12 +94,6 @@ const SAVE_NODES_INTERVAL: Duration = Duration::from_secs(30);
 /// Maximum duration to wait for a netcheck report.
 const NETCHECK_REPORT_TIMEOUT: Duration = Duration::from_secs(10);
 
-#[derive(Debug, Copy, Clone, PartialEq, Eq)]
-enum CurrentPortFate {
-    Keep,
-    Drop,
-}
-
 /// Contains options for `MagicSock::listen`.
 #[derive(derive_more::Debug)]
 pub struct Options {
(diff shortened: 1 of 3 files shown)


iroh-net/src/magicsock.rs
```diff
@@ -1869,15 +1826,6 @@ impl Actor {
         inc!(MagicsockMetrics, update_endpoints);
 
         debug!("starting endpoint update ({})", why);
-        if self.no_v4_send && !self.inner.is_closed() {
```
dignifiedquire, line 1872: should we error out here? seems in this case we are quite dead in the water

iroh-net/src/magicsock.rs
```diff
@@ -1699,10 +1657,17 @@ impl Actor {
         debug!("link change detected: major? {}", is_major);
 
         if is_major {
-            // Clear DNS cache
             DNS_RESOLVER.clear_cache();
             self.inner.re_stun("link-change-major");
-            self.rebind_all().await;
```
dignifiedquire, line 1705: I think I prefer having this in a method

dignifiedquire: you are not calling `update_net_info` on this path anymore, is that an issue?

flub: Moved it to a method.

And good spot!  I did indeed remove that, because this is already done by the `re_stun()` call just above.  In fact that method should not have been allowed to be called, I've documented that in the docstring for now.

iroh-net/src/magicsock.rs
flub: I'm not sure there's anything for us to do really.  Trying to send and receive data will result in an error at which point the application code has to give up on the magicsock like it would have with a normal socket.

The other option is to short-cut this into a fatal error that's propagated in `.poll_send()` or `.poll_recv()`, but it achieves just the same.

dignifiedquire (approved):

PR #2085, opened 2024-03-14
feat(iroh-bytes)!: refactor downloader queue and add progress reporting

Frando:
## Description

This PR contains changes to the downloader:

* Remove the `Role::Provider` vs `Role::Candidate` distinction. We added this back when we did not have content propagation in docs, and it now does not make much sense for our architecture. Either we know/assume a node has something, or not. The "inbetween" state did not make sense anymore IMO.
* Rework the queuing logic to be based on a simple queue of pending downloads. Before, if a download could not be started because the concurrenty limits were reached, the download was considered failed and inserted, with a delay, into the retry queue. Now, if a download cannot be started, we just wait until a slot frees up, and then start it. Note that the queue is, for now, quite simple - if the next download in the queue cannot be started (e.g. because all provider nodes are currently busy with other downloads), we do not try to start the second-top download in the queue, but instead wait until a slot is freed up. We could certainly optimize this by "jumping the queue" in certain cases, this would however also need more logic to make sure that downloads cannot be "forgotten". Therefore, for now the queue is handled strictly in order.
* The retry behavior is refactored: We now retry nodes (not downloads as before) up to a retry limit, with an increasing timeout. If a download can only proceed with a retrying node, it is parked and the next item in the queue is processed. The download is unparked if the retrying node successfully connects.
* Add progress reporting to downloads managed through the downloader. For this I wrote a `SharedProgress` handler that allows to subscribe to already running downloads: If an intent is registered for hash A, and this download is started, and while it is running, another intent is registered for the same hash, it will now receive an `DownloadProgress::InitialState` which contains a `TransferProgress` which functions as a reducer for the progress events This can be used from the client even, and further events can be reduced/merged into that struct. The PR contains a test for this concurrent progress reporting.
* Expose the downloader in the iroh node. Download requests via the RPC API can now set a `DownloadMode` enum either to `Direct` or to `Queued`: the former will behave as currently (issue an iroh-bytes request directly, without a queue or concurrency limits) and the latter will add the download to the downloader queue.

## Breaking changes

Changes in `iroh`:
* The  `BlobDownloadRequest`  has a new field `mode` to select between direct and queued downloads, and now contains a list of `nodes` in place of a single `node` before

Changes in `iroh_bytes`: 
* `Role` enum is removed
* `Downloader::queue` now takes a `DownloadRequest` with more options than before
* `DownloadProgress` has a new variant `InitialState` which is emitted when attaching to an already-running download
* `ConcurrencyLimits` gained a new field

Other changes:
* `SetTagOption` was moved from `iroh`  to `iroh-bytes`

## Notes & open questions

* Another followup improves content downloading in docs: #2127 .

* A few more tests around the queuing behavior would be great

* I have a partially done followup which adds a hook for content discovery to the downloader

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh-bytes/src/get/progress.rs
```diff
@@ -0,0 +1,206 @@
...
+            _ => {}
+        }
+    }
+}
+
+/// The id of a blob in a transfer
+#[derive(
+    Debug, Copy, Clone, Ord, PartialOrd, Eq, PartialEq, std::hash::Hash, Serialize, Deserialize,
+)]
+pub enum BlobId {
```
rklaehn, line 183: Nice, might be worth moving this further into the get part of iroh-bytes for clarity.

iroh-bytes/src/downloader/invariants.rs
```diff
@@ -73,15 +74,15 @@ impl<G: Getter<Connection = D::Connection>, D: Dialer> Service<G, D> {
...
-    /// Checks that the scheduled requests match the queue that handles their delays.
-    #[track_caller]
-    fn check_scheduled_requests_consistency(&self) {
-        assert_eq!(
-            self.scheduled_requests.len(),
-            self.scheduled_request_queue.len(),
-            "scheduled_request_queue and scheduled_requests are out of sync"
-        );
-    }
+    // /// Checks that the scheduled requests match the queue that handles their delays.
```
rklaehn, line 77: What's with this? No longer needed?

iroh-bytes/src/downloader/progress.rs
```diff
@@ -0,0 +1,262 @@
...
+use crate::{
+    export::ExportProgress,
+    get::{db::DownloadProgress, progress::TransferState},
+    util::progress::{FlumeProgressSender, IdGenerator, ProgressSendError, ProgressSender},
+};
+
+use super::DownloadKind;
+
+/// The channel that can be used to subscribe to progress updates.
+pub type ProgressSubscriber = FlumeProgressSender<DownloadProgress>;
```
rklaehn, line 21: Once it is merged we can make this a BoxedProgressSender so we don't have flume in here...

iroh/src/node/rpc.rs
```diff
@@ -990,50 +970,94 @@ impl<D: BaoStore> Handler<D> {
...
-    hash_and_format: HashAndFormat,
-    out: DownloadLocation,
-    tag: SetTagOption,
-    progress: impl ProgressSender<Msg = DownloadProgress> + IdGenerator,
+async fn download<D>(
+    db: &D,
+    endpoint: MagicEndpoint,
+    downloader: &Downloader,
+    req: BlobDownloadRequest,
+    progress: FlumeProgressSender<DownloadProgress>,
```
rklaehn, line 978: To avoid this whole id mapping thing, couldn't you have a global id generator? Then the ids for different downloads would never overlap, and you could join them without having to do the id mapping...

Frando: This is what I had until a couple of commits ago. However! After the download finishes, the export runs, outside of the scope of the downloader. See the changes in `iroh/src/node/rpc.rs`.
So the flow roughly like this:
```rust
let progress =  FlumeProgressSender::new();
let handle = downloader.queue(progress.clone());
handle.await?;
progress.send(NetworkDone).await;
export_download(progress.clone()).await?;
progress.send(AllDone).await;
```
the `export_download` again creates ids with `progress.new_id()`. Without the id mapping, these progress ids could be the same ones that were emitted during the download.

I couldn't come up with a nice way to solve this, that's why I resorted to the id map. I agree that it's not too nice, but it is the most flexible if we want to keep the current premise of "just pass in any progress sender".

I'm open to other ideas...

iroh-bytes/src/downloader/invariants.rs
Frando: Ah right, the delay queue doesn't exist anymore. As noted in the description above, no retries atm because we first need to decide how they should work.

Left this in earlier to see if there's a similar assert to be made now. Will think through and either remove or replace

iroh/src/node/rpc.rs
Frando: With #2113 a download now only gets progress IDs from within the downloader, so we can remove the whole id mapping foo!

iroh-bytes/src/get/progress.rs
Frando: Moved it up!

iroh-bytes/src/downloader.rs
```diff
@@ -248,20 +333,26 @@ impl Downloader {
...
+        let intent_id = IntentId(self.next_id.fetch_add(1, Ordering::SeqCst));
         let (sender, receiver) = oneshot::channel();
         let handle = DownloadHandle {
-            id,
-            kind: kind.clone(),
+            id: intent_id,
+            kind,
             receiver,
         };
+        // let callbacks = IntentCallbacks {
```
Frando, line 346: remove

iroh-bytes/src/downloader.rs
```diff
@@ -248,20 +333,26 @@ impl Downloader {
...
+        // };
         let msg = Message::Queue {
-            kind,
-            id,
-            sender,
-            nodes,
+            on_finish: sender,
+            request,
+            intent_id, // kind,
+                       // nodes: req.nodes,
```
Frando, line 354: remove

iroh-bytes/src/downloader/invariants.rs
Frando: Replaced with an actual check.

iroh-bytes/src/downloader/test/dialer.rs
```diff
@@ -87,6 +88,7 @@ impl TestingDialer {
         assert_eq!(self.0.read().dial_history, history)
     }
 
+    #[allow(unused)]
```
dignifiedquire, line 91: why keep around?

iroh/src/node/rpc.rs
```diff
@@ -1037,31 +1020,138 @@ impl<D: BaoStore> Handler<D> {
...
 where
     D: BaoStore,
-    C: FnOnce() -> F,
-    F: Future<Output = Result<quinn::Connection>>,
 {
-    let stats =
-        iroh_bytes::get::db::get_to_db(&db, get_conn, &hash_and_format, progress.clone()).await?;
+    ensure!(!nodes.is_empty(), "No nodes to download from provided.");
+    let mut last_err = None;
+    for node in nodes {
```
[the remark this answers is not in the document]
Frando: So this is new functionality introduced in this PR (in #2128 which is now merged into this PR, to be exact). We could parallize this in theory - I mostly added it because it is a common usecase for the downloader, and to keep the RPC APIs for queued vs direct downloads aligned, added this.
* If we keep it serial, the list of nodes is considered ordered, in the order to try the nodes
* We could parallize it, but would then need a good "cutoff point" on when to chose the "winning" request and abort the others (if all succeed). It would also slightly increase the connection overhead, especially if you include many node ids.

So I think this being serial is the simplest for now. I will check that the behavior is documented, and add that "if you want to try connecting to multiple nodes in parallel, use `DownloadMode::Queued`?

Or would you suggest to always parallelize here as well?

iroh-bytes/src/downloader/invariants.rs
```diff
@@ -21,8 +21,9 @@ impl<G: Getter<Connection = D::Connection>, D: Dialer> Service<G, D> {
     fn check_concurrency_limits(&self) {
         let ConcurrencyLimits {
             max_concurrent_requests,
-            max_concurrent_requests_per_node: max_concurrent_requests_per_peer,
+            max_concurrent_requests_per_node,
             max_open_connections,
+            ..
```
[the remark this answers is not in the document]
Frando: good catch, added a check for `max_concurrent_dials_per_hash`

iroh-bytes/src/downloader/progress.rs
```diff
@@ -0,0 +1,186 @@
...
+impl IdGenerator for BroadcastProgressSender {
+    fn new_id(&self) -> u64 {
+        self.id_gen.fetch_add(1, Ordering::SeqCst)
+    }
+}
+
+impl ProgressSender for BroadcastProgressSender {
+    type Msg = DownloadProgress;
+
+    async fn send(&self, msg: Self::Msg) -> std::result::Result<(), ProgressSendError> {
```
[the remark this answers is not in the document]
Frando: no reason, autocomplete or copypaste foo

iroh-bytes/src/util/progress.rs
```diff
@@ -94,6 +94,21 @@ pub trait ProgressSender: std::fmt::Debug + Clone + Send + Sync + 'static {
...
+    /// without a reference on `self`.
+    ///
+    /// The default implementation boxes the future. Some channels may optimize this
+    /// without needing to box, e.g [`FlumeProgressSender::into_send`] does not box.
+    #[must_use]
+    fn into_send(
+        self,
+        msg: Self::Msg,
+    ) -> impl Future<Output = ProgressSendResult<()>> + Send + 'static {
+        async move { self.send(msg).await }.boxed()
```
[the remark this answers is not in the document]
Frando: actually this is not needed at all anymore, will remove
(likely the box was needed before this moved to RPITIT, but not sure)

iroh-bytes/src/downloader/test/dialer.rs
Frando: removed

iroh-bytes/src/downloader.rs
```diff
@@ -454,540 +503,495 @@ struct Service<G: Getter, D: Dialer> {
...
+    /// Information about pending and active requests
+    requests: HashMap<DownloadKind, RequestInfo>,
+    /// State of running downloads
+    active_requests: HashMap<DownloadKind, ActiveRequestInfo>,
+    /// Tasks for currently running transfers.
+    in_progress_downloads: JoinSet<(DownloadKind, InternalDownloadResult)>,
+    /// Progress tracker
+    progress_tracker: ProgressTracker,
+    /// Registered intents with progress senders and result callbacks.
+    intents: HashMap<IntentId, IntentCallbacks>,
```
divagant-martian, line 521: Why not store this in `RequestInfo`? if the intents are already being hashed there

iroh-bytes/src/downloader.rs
```diff
@@ -454,540 +503,495 @@ struct Service<G: Getter, D: Dialer> {
...
                         Ok((kind, result)) => {
-                            trace!("tick: download completed");
-                            self.on_download_completed(kind, result);
+                            trace!(kind=%kind.fmt_short(), "tick: transfer completed");
+                            self.on_download_completed(kind, result).await;
                         }
-                        Err(e) => {
-                            warn!("download issue: {:?}", e);
+                        Err(err) => {
+                            warn!(?err, "transfer task paniced");
```
divagant-martian, line 573:
```suggestion
                            warn!(?err, "transfer task panicked");
```

However, it seems to be a `JoinError` can be both a cancel and panic

iroh-bytes/src/downloader.rs
```diff
@@ -454,540 +503,495 @@ struct Service<G: Getter, D: Dialer> {
...
-            } else {
-                trace!(node = %node.node_id, "required node not dialed to maintain concurrency limits");
-                None
+            .or_insert_with(|| RequestInfo::new(intent_id, tag));
+
+        // queue the transfer (if not running) or attach to transfer progress (if already running)
+        if self.active_requests.contains_key(&kind) {
+            // the transfer is already running, so attach the progress sender
+            if let Some(on_progress) = &intent_callbacks.on_progress {
+                // this is async because it send the current state over the progress channel.
```
divagant-martian, line 663:
```suggestion
                // this is async because it sends the current state over the progress channel
```

iroh-bytes/src/downloader.rs
```diff
@@ -454,540 +503,495 @@ struct Service<G: Getter, D: Dialer> {
...
-            let intents = &mut occupied_entry.get_mut().intents;
-            intents.remove(&id);
-            // if this was the last intent associated with the request cancel it
-            if intents.is_empty() {
-                download_removed = true;
-                occupied_entry.remove().cancellation.cancel();
-            }
-        } else if let Entry::Occupied(mut occupied_entry) = self.scheduled_requests.entry(kind) {
-            // remove the intent from the associated request
+    /// remove it from the scheduled requests, or cancel the future.o send abort message on progress sender
```
divagant-martian, line 682: funky docs here

iroh-bytes/src/downloader.rs
```diff
@@ -454,540 +503,495 @@ struct Service<G: Getter, D: Dialer> {
...
-            Some((kind, info))
-        } else {
-            None
+        if no_more_intents {
+            if let Entry::Occupied(occupied_entry) = self.active_requests.entry(kind) {
+                occupied_entry.remove().cancellation.cancel();
+            } else {
+                self.queue.remove(&kind);
+            }
+            self.providers.remove_hash(&kind.hash());
```
divagant-martian, line 715: However unlikely, won't this remove the providers for the other kind if present?

iroh-bytes/src/downloader.rs
```diff
@@ -1084,88 +1053,142 @@ impl<G: Getter<Connection = D::Connection>, D: Dialer> Service<G, D> {
...
-    provider_hashes_by_node: HashMap<NodeId, VecDeque<Hash>>,
-}
-
-struct ProviderIter<'a> {
-    inner: Option<std::collections::hash_map::Iter<'a, NodeId, Role>>,
+#[derive(Debug)]
+enum NextStep {
+    StartTransfer(NodeId),
+    Dial(NodeId),
+    DialAfterIdleDisconnect(NodeId, delay_queue::Key),
```
divagant-martian, line 1060: `DialQueuedDisconnect`?
open to other names. "After idle disconnect" sounds to me like we need to wait for the idle timeout, disconnect, and then dial. Which ofc is not what we want

iroh-bytes/src/downloader.rs
```diff
@@ -1084,88 +1053,142 @@ impl<G: Getter<Connection = D::Connection>, D: Dialer> Service<G, D> {
...
+            if hashes.is_empty() {
+                self.remove_node(node);
+            }
+        }
+    }
+}
+
+/// Helper struct to merge multiple tags for a blob.
+#[derive(Debug, Default)]
+struct TagSet {
```
divagant-martian, line 1148: Not a suggestion, but a question. Would it make sense to have this around the tag api/db code? seems useful enough and perhaps more pertinent to have there

iroh-bytes/src/downloader.rs
Frando: :+1: We don't cancel the tasks in here though, we only use a cancellation token so I think panic is the only reason this can get triggered.

iroh-bytes/src/downloader.rs
Frando: fixed

iroh-bytes/src/get/progress.rs
```diff
@@ -0,0 +1,177 @@
...
+                    self.root.child_count = Some(children);
+                } else {
+                    // I think it is an invariant of the protocol that `FoundHashSeq` is only
+                    // triggered for the root hash.
+                }
+            }
+            DownloadProgress::Progress { id, offset } => {
+                if let Some(blob) = self.get_by_progress_id(id) {
+                    blob.progress = BlobProgress::Progressing(offset);
+                }
```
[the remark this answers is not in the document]
Frando: Yes, your understanding is correct. I added `warn!` logs for the else cases so we can notice them. As this is only progress reporting I think I wouldn't assert them, but open to change it to debug_assert.

iroh-bytes/src/downloader.rs
```diff
@@ -454,540 +503,495 @@ struct Service<G: Getter, D: Dialer> {
...
+                    debug!(kind=%kind.fmt_short(), "abort download: out of providers");
+                    let _ = self.queue.pop_front();
+                    let info = self.requests.remove(&kind).expect("queued downloads exist");
+                    self.finalize_download(kind, info.intents, Err(DownloadError::NoProviders));
                 }
             }
+        }
+    }
+
+    fn next_step(&self, kind: &DownloadKind) -> NextStep {
```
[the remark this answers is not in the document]
Frando: :+1: added more docs

iroh-bytes/src/downloader.rs
```diff
@@ -454,540 +503,495 @@ struct Service<G: Getter, D: Dialer> {
...
+        let mut next_to_dial = None;
+        for node in candidates {
+            match self.node_state(node) {
+                NodeState::Connected(info) => {
+                    let active_requests = info.active_requests();
+                    if self
+                        .concurrency_limits
+                        .node_at_request_capacity(active_requests)
+                    {
+                        has_exhausted = true;
```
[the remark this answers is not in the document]
Frando: Added this comment to the code:
```       
         // Track if we have at least one provider node which is currently at its request capacity.
        // If this is the case, we will never return [`NextStep::OutOfProviders`] but [`NextStep::Wait`]
        // instead, because we can still try that node once it has finished its work.
```
does this make it clear? if not please post back and I'll refine the docs

iroh-bytes/src/downloader.rs
```diff
@@ -454,540 +503,495 @@ struct Service<G: Getter, D: Dialer> {
...
+
+        if !available.is_empty() {
+            available.sort_unstable_by_key(|(_node, req_count)| *req_count);
+            let (node, _) = available.last().expect("just checked");
+            NextStep::StartTransfer(**node)
+        } else if let Some(node) = next_to_dial {
+            let at_dial_capacity = has_dialing
+                && self
+                    .concurrency_limits
+                    .at_dials_per_hash_capacity(currently_dialing);
```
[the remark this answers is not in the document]
Frando: Not sure I understand your comment correctly: My understanding is:
* We iterate over the providers for this download, and count the number of nodes from the provider list which are currently being dialed
* If this number is higher than `at_dials_per_hash_capacity`, we don't queue another dial
This seems correct to me?

iroh-bytes/src/downloader.rs
Frando: Good catch, added a check.

iroh-bytes/src/downloader.rs
```diff
@@ -454,540 +503,495 @@ struct Service<G: Getter, D: Dialer> {
...
+    /// This triggers the intent return channels, and removes the download from the progress tracker
+    /// and provider map.
+    fn finalize_download(
+        &mut self,
+        kind: DownloadKind,
+        intents: HashSet<IntentId>,
+        result: ExternalDownloadResult,
+    ) {
+        self.progress_tracker.remove(&kind);
+        self.providers.remove_hash(&kind.hash());
```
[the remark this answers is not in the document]
Frando: added a check!

iroh-bytes/src/downloader.rs
Frando: SGTM, moved to `iroh-bytes/src/util.rs`

iroh-bytes/src/downloader.rs
Frando: Agreed and changed!
This was a bit different in an earlier design, now it makes sense to inline this in the request info. Also simplifies the code and removes some lines.

iroh-bytes/src/downloader.rs
```diff
@@ -274,159 +359,119 @@ impl Downloader {
...
     /// cancelled.
-    Cancel { id: Id, kind: DownloadKind },
-    /// Declare that nodes have certains hash and can be used for downloading. This feeds the [`ProviderMap`].
-    PeersHave { hash: Hash, nodes: Vec<NodeInfo> },
+    CancelIntent { id: IntentId, kind: DownloadKind },
 }
 
-/// Information about a request being processed.
+#[derive(derive_more::Debug)]
+struct IntentCallbacks {
```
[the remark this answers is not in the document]
Frando: Renamed to `IntentHandlers`, wdyt?

iroh-bytes/src/get/progress.rs
```diff
@@ -0,0 +1,177 @@
...
+                size,
+            } => {
+                let blob_id = child;
+                let blob = self.get_or_insert_blob(blob_id, hash);
+                if blob.size.is_none() {
+                    blob.size = Some(BaoBlobSize::Verified(size));
+                }
+                blob.progress = BlobProgress::Progressing(0);
+                self.progress_id_to_blob.insert(progress_id, blob_id);
+                self.current = Some(blob_id);
```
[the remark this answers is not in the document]
Frando: Currently iroh-bytes only ever downloads one child blob after the other, linearily. The `self.current` is only used to get the current child offset, and is only used by the CLI to say `downloading blob 17 of 53` or so.
So I'd think this is fine as-is, if we start to download multiple blobs in parallel, the progress reporting will need some refactor anyway.

iroh-bytes/src/downloader/progress.rs
```diff
@@ -0,0 +1,186 @@
...
+        &mut self,
+        kind: DownloadKind,
+        subscribers: impl IntoIterator<Item = ProgressSubscriber>,
+    ) -> BroadcastProgressSender {
+        let inner = Inner {
+            subscribers: subscribers.into_iter().collect(),
+            state: TransferState::new(kind.hash()),
+        };
+        let shared = Arc::new(Mutex::new(inner));
+        self.running.insert(kind, Arc::clone(&shared));
```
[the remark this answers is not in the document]
divagant-martian: I think that's good enough long term

iroh-bytes/src/downloader.rs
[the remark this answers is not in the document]
divagant-martian: this makes it clear

iroh-bytes/src/downloader.rs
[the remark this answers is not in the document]
divagant-martian: The confusing part to me is dialing _for this hash_, which we don't track (because I don't think we care in general). However this is tricky to verify and maybe we should try it as a best effort thing, more than a hard limit treated as an invariant., and document it as such.

Consider the following scenario:

- `at_dials_per_hash_capacity = 2`
- `ProviderMap := {HashX <-> {PeerA, PeerB}}`
- `PeerA` is being dialed (for `HashX`)
- `PeerB` is being dialed (for `HashX`)
- `HashX` is at capacity
- queue now `HashY`, with `PeerC`, which is useful for `HashX` as well
- `ProviderMap := {HashX <-> {PeerA, PeerB, PeerC}, HashY <-> {PeerC}}`
- dial `PeerC` for `HashY`, well within capacity
- `HashX` is now above capacity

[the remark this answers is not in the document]
Frando: Right, I understand the problem now, thank you!

I am thinking about two potentially viable options:

* What would happen if we remove the `at_dials_per_hash_capacity` limit? My thought for adding it was to not use all available connections capacity for dialing a single hash. However, in the current code flow, we only ever dial providers for the current head of the queue. When one of the dials succeeds, we start a transfer with that node, and move to the next item in the queue. The remaining running dials will not be aborted, but once they succeed, their state will be idle, so if the next item needs to start dials, it will return `NextStep::DialQueuedDisconnect`, which will terminate idle nodes as needed. The remaining issue *can* be that if hash A starts many dials, they won't be aborted until each connection is established, so the next hash in the queue might have to wait until those  move to connected-and-idle state to abort them.
So what we could do is mark dials as `required`, and abort non-required dials before terminating idle connections on `NextStep::DialQueuedDisconnect`. This is fine, but we still might not want to start 50 parallel dials at all if we happen to have many providers for a hash. So, option B:

* We leave things exactly as they are! `at_dials_per_hash_capacity` is not checked as an invariant, but only at the codepoint you linked, so only to determine whether a new dial should be started or not. Which is fine - the dial will either succeed at the connection be "up for picks" for the next element in the queue. Or it will fail, which will free a slot in `at_dials_per_hash_capacity` for all hashes that care about the node.

WDYT?

divagant-martian:
@Frando what do think about moving the progress reporting out to another PR. That part is mostly self-contained and as far as I can tell has been reviewed by everyone and it's good to go. The queuing is a different issue that I think needs more work still, since in terms of reliability of downloads the current state in the PR is much worse than before without retries

Frando:
> @Frando what do think about moving the progress reporting out to another PR. That part is mostly self-contained and as far as I can tell has been reviewed by everyone and it's good to go. The queuing is a different issue that I think needs more work still, since in terms of reliability of downloads the current state in the PR is much worse than before without retries

I will have a look tomorrow, not sure OTOH how much work the splitting is but should be doable.

I'm not too sure though if we are worse in practice than before. We always retry for 10 seconds in our configuration (default quinn timeout). We also continue using a node if it returned not found. We do not retry a node that "failed" for more than 10 seconds, or had an unrecoverable failure (bad protocol, or bad quic). I do think we should add retries here,, but I'm not sure if it is essential.
On `main`, we handle some more retries, but we have another problem I think which is be fixed by this PR: If a lot of downloads are queued, and the resource limits are reached, newly queued downloads end up in the retry queue with increasing timeouts, so will even still wait if the slots are freed up, and might be retried while the limits are still reached, making them exceed their retry budget.

Frando:
Here is a PR on top of this one which adds retries: #2165

iroh-bytes/src/get/progress.rs
```diff
@@ -0,0 +1,177 @@
...
+            DownloadProgress::Found {
+                id: progress_id,
+                child,
+                hash,
+                size,
+            } => {
+                let blob_id = child;
+                let blob = self.get_or_insert_blob(blob_id, hash);
+                if blob.size.is_none() {
+                    blob.size = Some(BaoBlobSize::Verified(size));
```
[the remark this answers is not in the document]
divagant-martian: @rklaehn could you please check this?

iroh-bytes/src/downloader.rs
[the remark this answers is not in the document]
divagant-martian: yes second option aligns with my suggestion: treat it as a best effort thing and document it

iroh-bytes/src/get/progress.rs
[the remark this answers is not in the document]
Frando: Sorry, skipped this comment somehow. When rereading things, I think `Unverified` would be correct. It is the size the remote told us.

(`Found` should be renamed to `RemoteFound` or so to make clear that this convey information from the remote peer, and not from our local store, where it already is well-named `FoundLocal`)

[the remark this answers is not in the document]
Frando: I pushed a commit which changes this to `Unverified`. I think this is correct, but I also am not fully sure what the implications are, cc @rklaehn

iroh-bytes/src/downloader.rs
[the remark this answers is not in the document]
Frando: I pushed a commit that adds more docs to `at_dials_per_hash_capacity` and the logic in `next_step`.

iroh-bytes/src/downloader.rs
```diff
@@ -454,540 +506,535 @@ struct Service<G: Getter, D: Dialer> {
...
+        // get general request info
+        let request_info = self.requests.remove(&kind).expect("request was active");
 
+        let ActiveRequestInfo { node, temp_tag, .. } = active_request_info;
+
+        // get node info
         let node_info = self
             .nodes
             .get_mut(&node)
             .expect("node exists in the mapping");
```
Frando, line 749: @ppodolsky triggered this `expect` when using this branch. I have a suspicion and am writing a test to confirm.

Frando: This is now fixed.

divagant-martian (approved):
Minor comments. TBH this was a less thorough review just due to reviewer exhaustion, which I assume are PR owner are suffering as well. Taking into account we just did a release, I think we have time to deal with bugs if any in other PRs

iroh-bytes/src/downloader/invariants.rs
```diff
@@ -113,13 +115,27 @@ impl<G: Getter<Connection = D::Connection>, D: Dialer, S: Store> Service<G, D, S
                 "all queued requests have request info"
             );
         }
+
+        // check that all parked hashes should be parked
+        for entry in self.queue.iter_parked() {
+            assert!(
+                matches!(self.next_step(entry), NextStep::Park),
+                "next step for parked node ist WaitForNodeRetry"
```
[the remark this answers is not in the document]
Frando: The comment is wrong, the enum variant was renamed earlier but not in the comment. Also slopyy language. Rephrased to "all parked downloads evaluate to the correct next step".

iroh-bytes/src/downloader/test.rs
```diff
@@ -345,3 +361,134 @@ async fn long_queue() {
         res.expect("all downloads to succeed");
     }
 }
+
+#[tokio::test]
+async fn fail_while_running() {
```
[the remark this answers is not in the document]
Frando: Your intuition was correct. The branch previously had a bug (spotted by @ppodolsky) where `DropPeer` would not first check if there are other downloads running but drop the peer unconditionally. The test recreates that scenario, and passes since the bug was fixed.
I added this comment to the test:
```rust
/// If a download errors with [`FailureAction::DropPeer`], make sure that the peer is not dropped
/// while other transfers are still running.
```

>  Does it matter whether the peer was or not removed (DropPeer) after both are done?

Currently, a peer will only be dropped after a `FailureAction::DropPeer` if no other transfer is running. My intuition for that choice: 
* If the other transfers after the failed one succeed without error, the condition that caused DropPeer was likely transient, so we just keep the peer and ignore the previous DropPeer
* If the failure condition remains, the other transfers will fail with DropPeer as well, and after the last transfer failed, the peer will be dropped.

As I'm typing this, the following scenario would be possible in the current implementation, but likely should be avoided:
* Queue hashA, hashB with nodeA. hashA, hashB are started
* hashA fails with DropPeer -> peer is not dropped because hashB is still running
* Queue hashC with nodeA. hashC is started
* hashB fails with DropPeer -> peer is not dropped  because hashC is still running
* Queue hashD with nodeA. hashD is started
* ... repeat as above

In plain words: If I keep queuing new dowloads to a bad node in some frequency, it might happen that the peer is never dropped even though it is clearly faulty. Downloads will still succeed if other providers are available, but cycles will be waisted and resource capacity needlessly consumed.

I think I will add a test for this case. A fix could be to mark a node as `failed` (or count the failures) and not queue new downloads for suchly marked nodes (but do not drop them while things are running). After a successful transfer, the `failed` marker is cleared.

iroh-bytes/src/downloader.rs
```diff
@@ -501,14 +542,20 @@ struct Service<G: Getter, D: Dialer, DB: Store> {
...
+    /// Configuration for retry behavior.
+    retry_config: RetryConfig,
     /// Channel to receive messages from the service's handle.
     msg_rx: mpsc::Receiver<Message>,
-    /// Active connections
-    nodes: HashMap<NodeId, ConnectionInfo<D::Connection>>,
-    /// Queue to manage dropping nodes.
+    /// Nodes to which we have an active or idle connection
+    connected_nodes: HashMap<NodeId, ConnectionInfo<D::Connection>>,
+    /// Nodes that failed and for which we track retries
```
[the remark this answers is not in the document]
Frando: It's both. Adjusted the comment.

PR #2088, opened 2024-03-18
[wip] feat(iroh-net): shutdown magicsock on quinn endpoint shutdown

divagant-martian:
## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -51,9 +51,9 @@ dependencies = [
 
 [[package]]
 name = "allocator-api2"
-version = "0.2.18"
+version = "0.2.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5c6cb57a04249c6480766f7f7cef5467412af1490f8d1e243141daddada3264f"
+checksum = "0942ffc6dcaadf03badf6e6a2d0228460359d5e34b57ccdc720b7382dfbd5ec5"
 
 [[package]]
 name = "android-tzdata"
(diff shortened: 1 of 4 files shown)


iroh-net/src/magicsock/relay_actor.rs
```diff
@@ -276,7 +276,7 @@ impl ActiveRelay {
 }
 
 pub(super) struct RelayActor {
-    conn: Arc<Inner>,
+    msock: Arc<MagicSockInner>,
```
flub, line 279: +100 for all this renaming :)

Cargo.lock
flub, line 1: These changes all seem to be downgrading, they are probably a mistake?

divagant-martian:
Sadly, while this allows quinn to make use of the socket until it's actually done with it, we do close operations that are now not actually guaranteed to finish. Although the only one we currently test against it saving peers, there are other shutdown operations going on as well that matter and that we will want to make sure are correctly done in the future.

My conclusion is that this is not really worth doing: Yes, it adjusts to quinn's quirkiness but in turns makes it worse for the MagicEndpoint.

Thus I'm closing this.

Hopefully in the future we can revisit, and change quinn to do a better job in the `close` story, so that it's not left in an intermediate unusable state, or at least make the endpoint driver shutdown on close instead of on drop. These are what I see as entry points to the work to be made, but the needed effort is probably larger than just that

flub:
Would a conceptual way to get this working be to add some sort of API that allows you to know when the magicsocket is finally done?  E.g. `MagicEndpoint::notify_closed() -> oneshot::Receiver<()>` or something similar?  That way you could drop the thing and still be told when everything is finally finished.

Not saying you need to pick this up again, it's been a lot of work already.  I'm more interested in understanding the problem space better.

divagant-martian:
I've gone this path already. Doing so would mean potentially storing the `quinn::Endpoint` in an option do be able to drop it while allowing the `MagicEndpoint` to still exist. But then we need to deal everywhere in the code with ensuring this is `Some`. Many methods do not return an error, so then we would need to add a panic warning to every method in that list. Same with similar options as and `Arc` or `Weak`, etc. 

This is not really the same case as `UdpSocket` which contains an option, since we know for sure the option is taken only on `drop`, while here it would be in `close`.

To me this feels like leaking a somewhat poor design up to our code

PR #2094, opened 2024-03-19
feat(iroh): improve various aspects of the api

dignifiedquire:
- easier access to the client via deref on the node
- simply await add and download progress results
- general simplifications in the examples
- split up client implementation across individual mods.


iroh/src/client.rs
```diff
@@ -578,6 +575,37 @@ impl Stream for BlobAddProgress {
...
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        loop {
+            match self.stream.poll_next_unpin(cx) {
+                Poll::Pending => return Poll::Pending,
+                Poll::Ready(None) => {
+                    return Poll::Ready(Err(anyhow!("Response stream ended prematurely")))
+                }
+                Poll::Ready(Some(Err(err))) => return Poll::Ready(Err(err)),
+                Poll::Ready(Some(Ok(msg))) => match msg {
```
Frando, line 589: Misses the `AddProgress::Found` case to add to `self.current_total_size` (see the removed code from `async fn finish`)

dignifiedquire: that's always done now using the atomic u64, see the stream mapping in the constructor

Frando (approved):
Nice cleanup, we need those. Did not check the copy-pastes into the client modules, assuming they are unchanged apart from the blob progress handlers, which I reviewed.

PR #2100, opened 2024-03-20
fix(iroh): Properly shut down the store on control-c

rklaehn:
## Description

fix(iroh): Properly shut down the store on control-c
    
without this, the (flaky) cli_provide_persistence test fails, because
control-c interrupts the store before the write transaction is committed.

## Notes & open questions

Should we change shutdown in node to be async, and remove the impl Future for Node?

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-bytes/src/store/file.rs
+++ b/iroh-bytes/src/store/file.rs
@@ -650,7 +650,9 @@ pub(crate) enum ActorMessage {
     /// This will be called periodically and can be used to do misc cleanups.
     GcStart { tx: oneshot::Sender<()> },
     /// Internal method: shutdown the actor.
-    Shutdown,
+    ///
+    /// Can have an optional oneshot sender to signal when the actor has shut down.
+    Shutdown { tx: Option<oneshot::Sender<()>> },
 }
 
 impl ActorMessage {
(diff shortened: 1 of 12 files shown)


iroh-bytes/src/store/file.rs
```diff
@@ -1153,12 +1155,21 @@ impl StoreInner {
     fn temp_file_name(&self) -> PathBuf {
         self.path_options.temp_file_name()
     }
+
+    async fn shutdown(&self) {
```
rklaehn, line 1159: God how I hate all this boilerplate. But no macros please. At some point I will write an actor framework to solve this...

iroh-bytes/src/store/file.rs
```diff
@@ -1469,16 +1484,19 @@ impl Actor {
...
+                }
                 break;
             }
             match msg.category() {
                 MessageCategory::TopLevel => {
-                    let msg = msgs.recv().expect("just peeked");
                     self.state.handle_toplevel(&self.db, msg)?;
                 }
                 MessageCategory::ReadOnly => {
+                    msgs.push_back(msg).expect("just recv'd");
```
dignifiedquire, line 1499: how does this `push_back` construction work?

rklaehn: It's just a single option in front of the flume channel to make it peekable. You can push back if you just pulled. Makes structuring some code nicer. If the option is already full this will fail, but we just pulled from it so we know it is not full...

dignifiedquire (approved):

PR #2103, opened 2024-03-20
refactor(iroh-bytes): Add the ability to box progress senders so we have a concrete type.

rklaehn:
## Description

refactor(iroh-bytes): Add the ability to box progress senders so we have a concrete type.

This is necessary to send progress senders via actor messages, but we also probably don't want to use a concrete progress sender type in the actor messges.

Quite a bit of plumbing, but the only public API changes are a new fn boxed() on ProgressSender that returns a concrete type `BoxedProgressSender<M>`.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-bytes/src/util/progress.rs
+++ b/iroh-bytes/src/util/progress.rs
@@ -2,9 +2,10 @@
 //!
 //! The main entry point is the [ProgressSender] trait.
 use bytes::Bytes;
-use futures::Future;
+use derive_more::Deref;
+use futures::{future::BoxFuture, Future, FutureExt};
 use iroh_io::AsyncSliceWriter;
-use std::{io, marker::PhantomData};
+use std::{io, marker::PhantomData, sync::Arc};
 
 /// A general purpose progress sender. This should be usable for reporting progress
 /// from both blocking and non-blocking contexts.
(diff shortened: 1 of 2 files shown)


iroh-bytes/src/util/progress.rs
```diff
@@ -121,6 +122,107 @@ pub trait ProgressSender: std::fmt::Debug + Clone + Send + Sync + 'static {
...
+}
+
+impl<T> std::fmt::Debug for BoxedProgressSender<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_tuple("BoxedProgressSender").field(&self.0).finish()
+    }
+}
+
+/// Boxable progress sender
+trait BoxableProgressSender<T>: IdGenerator + std::fmt::Debug + Send + Sync + 'static {
```
Frando, line 151: Why do we need this new trait? Can't we just implement `ProgressSender` on `BoxedProgressSender`?

rklaehn: I need something dynable inside the BoxedProgressSender. And ProgressSender is not dynable, "can not be made into an object" bla bla bla.

I think the reasons are:

- It has msg as an associated type, not as a type parameter
- send returns an impl Future
- the with_filter and with_filter_map also are not good

I guess you could solve this by always returning a BoxedProgressSender from with_filter and with_filter_map, always returning a boxed future from send, and making the associated type a type parameter. But that would suck - way too much boxing in cases you don't need it...

Issue #2104, opened 2024-03-20
Too much `send derp: message dropped, channel to actor is full` happens during intense import on low-profile servers.


ppodolsky:
At a certain point, Iroh, built from the master branch, starts to spam with 'send derp: message dropped, channel to actor is full' 2000-3000 times per second during import. This excessive output and corresponding attemts to fit into full channel reduces the overall performance of the import process, also making Iroh almost unresponsive.

Related discussion: https://discord.com/channels/949724860232392765/950683937661935667/1220100350607101997

flub:
This has a sibling issue already: https://github.com/n0-computer/iroh/issues/1977 which covers the flow in the other direction IIUC.  DISCO messages received from the relay and going to the magicsock.

flub:
There was a conversation on discord about this, these are my personal conclusions from it:

- We need to add backpressure to the channel so that `.poll_send()` refuses to send the packet instead of accepting it and then dropping it right away.  This will slow down the flow from quinn directly.  Of course when doing that we need to make sure the socket becomes writable again when the channel has space again.
- This also affects hole-punching: disco messages need to be relayed to establish holepunching.  If they get dropped because of an overflowing channel then we will not holepunch which in turn means we never take the pressure of this channel and it never recovers.
- This means we may need to consider adding two channels to the derp client: one for quic payloads and one for disco messages.  Disco messages are rather important, if the queue is filled up with quic messages then we risk losing disco messages.  However on the flipside someone could try and attack you by making you flood disco messages, taking up valuable relay capacity.

PR #2105, opened 2024-03-21
fix(iroh-net): improve backpressure handling

dignifiedquire:
Feed backpressure in the form of `Poll::Pending` and waking up wakers back through `poll_send` when sending to derp. 

Should help with #2104, but that might have other problems as well.


iroh-net/src/magicsock.rs
```diff
@@ -348,18 +350,42 @@ impl Inner {
...
-                    self.try_send_derp(derp_url, public_key, split_packets(&transmits));
-                    transmits_sent = transmits.len();
-                    derp_sent = true;
+                    match self.try_send_derp(derp_url, public_key, split_packets(&transmits)) {
+                        Poll::Ready(sent) => {
+                            derp_sent = sent;
+                            transmits_sent = transmits.len();
+                        }
+                        Poll::Pending => {
+                            self.network_send_wakers.lock().replace(cx.waker().clone());
```
flub, line 371: huh, the derp actor already wakes this.  I guess this was always intended to be like this but was somehow overlooked?

iroh-net/src/magicsock.rs
```diff
@@ -348,18 +350,42 @@ impl Inner {
...
+                    // Handle no addresses being available
+                    warn!(node = %public_key.fmt_short(), "failed to send: no UDP or DERP addr");
+                    return Poll::Ready(Err(io::Error::new(
+                        io::ErrorKind::NotConnected,
+                        "no UDP or Derp address available for node",
+                    )));
+                }
+
+                if (udp_addr.is_none() || udp_pending) && (derp_url.is_none() || derp_pending) {
+                    // Handle backpressure
```
flub, line 387: Could we elaborate here on the tradeoff we're making?  IIUC we only return pending if we didn't manage to send to either.  That means we're accepting that some packets will only flow via one path and be dropped on the other.

However IIRC when we decide to send to both paths we do this because we're not certain that the UDP path is working.  So should we not always return pending if derp is pending in that case?

flub: Does this condition not also result in always returning pending when we have neither a derp nor a udp path?  However just below we're supposed to return a not-connected error in that case.

flub (commented):
I'm not sure if I fully understand what happens when both udp and derp return pending.  The waker will be stored in both the udp and derp code paths and the calling task will be awoken when either becomes writable again.  At that point the sender is going to send again and in the worse case only one of the paths will be able to send at that time, the other still returning pending.

iroh-net/src/magicsock.rs
dignifiedquire: it got lost in a refactoring I think

iroh-net/src/magicsock.rs
dignifiedquire: > Does this condition not also result in always returning pending when we have neither a derp nor a udp path?

this is errored out in the above condition

dignifiedquire:
> At that point the sender is going to send again and in the worse case only one of the paths will be able to send at that time, the other still returning pending.

This is how I think it will work best, we try again, and update the waker, which might have changed, if we still can't send on one path.

iroh-net/src/magicsock.rs
dignifiedquire: >  That means we're accepting that some packets will only flow via one path and be dropped on the other 

Yes, that's because we treat both connections as lossy, and sending is only best effort any way

flub: > > Does this condition not also result in always returning pending when we have neither a derp nor a udp path?
> 
> this is errored out in the above condition

oh yes, sorry i missed that.

flub: > > That means we're accepting that some packets will only flow via one path and be dropped on the other
> 
> Yes, that's because we treat both connections as lossy, and sending is only best effort any way

So your take is that we should not make sending via derp required when both paths are chosen?  I'm not 100% convinced that's the right choice.  In this situation either the udp path still works and recovers very quickly and will be marked the only path again.  Or it no longer works and will take longer and need holepunching etc again.  This could take up to 5s in the best case.  So I think I'm still in favour of requiring a derp send at this point, but if you really do not want to do this then I'll approve anyway.

But please update the comment making clear which way is chosen and why.  I think it would be good to have it described right there rather than having to find this conversation again.

iroh-net/src/magicsock.rs
```diff
@@ -900,7 +927,7 @@ impl Inner {
         Poll::Ready(Ok(()))
     }
 
-    fn try_send_derp(&self, url: &DerpUrl, node: PublicKey, contents: DerpContents) -> bool {
+    fn try_send_derp(&self, url: &DerpUrl, node: PublicKey, contents: DerpContents) -> Poll<bool> {
```
Frando, line 930: Should rename this to `poll_send_derp` now that it returns `Poll`? And maybe even make it take the `Context` as argument, register the waker in there and return `Poll<Result<()>>` instead of `Poll<bool>`, would align it more to other `poll_send` methods.

dignifiedquire: we call it from places where there is no context available, which is why I am not passing it as argument

iroh-net/src/magicsock.rs
dignifiedquire: added a comment

iroh-net/src/magicsock.rs
dignifiedquire: renamed

flub (approved):

PR #2107, opened 2024-03-21
feat(iroh-bytes): bring back validation

rklaehn:
## Description

feat(iroh-bytes): bring back validation

This brings back validation, and also separates validation and consistency check and validation.

Rationale:
- consistency check is highly dependent on the store impl. E.g. for an in memory impl it might not even make sense.
- validate is much more expensive, but can be mostly implemented in a store impl independent way. It just needs access to the data.

consistency check output at very high verbosity:
(basically dumps the db and then validates each entry)
image

validate output:
(validates all complete files, lists ranges for all partial files)
image

## Notes & open questions

- Note: repairing by downgrading complete entries to partial entries is not yet implemented. It is a bit tricky.
- Note: both are implemented as a RPC command for a running store, and in iroh doctor for a separate file.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


iroh-bytes/src/store/traits.rs
```diff
@@ -428,6 +433,141 @@ pub trait Store: ReadableStore + MapMut {
 
     /// Shutdown the store.
     fn shutdown(&self) -> impl Future<Output = ()> + Send;
+
+    /// Validate the database
```
dignifiedquire, line 437: maybe add some notes how this differs from the consistentency check?

iroh-bytes/src/store/traits.rs
```diff
@@ -428,6 +433,141 @@ pub trait Store: ReadableStore + MapMut {
...
+                        expected_chunk_range, actual_chunk_ranges
+                    ))
+                };
+                tx.send(ValidateProgress::EntryDone { id, error }).await?;
+                drop(ranges);
+                drop(entry);
+                io::Result::Ok((hash, incomplete))
+            })
+        })
+        .buffer_unordered(validate_parallelism)
```
dignifiedquire, line 509: sorry, but no, this will segfault iroh again..

iroh-bytes/src/store/traits.rs
```diff
@@ -428,6 +433,141 @@ pub trait Store: ReadableStore + MapMut {
...
+                    id,
+                    ranges: RangeSpec::new(&actual_chunk_ranges),
+                })
+                .await?;
+                drop(ranges);
+                drop(entry);
+                io::Result::Ok(())
+            })
+        })
+        .buffer_unordered(validate_parallelism)
```
dignifiedquire, line 551: same

iroh-cli/src/commands/blob.rs
```diff
@@ -435,48 +451,128 @@ impl DeleteCommands {
...
+            format!("{}: {} ({})", level_text, message, hash.to_hex())
+        } else {
+            format!("{}: {}", level_text, message)
+        };
+        let styled = match level {
+            ReportLevel::Trace => style(text).dim(),
+            ReportLevel::Info => style(text),
+            ReportLevel::Warn => style(text).yellow(),
+            ReportLevel::Error => style(text).red(),
+        };
```
dignifiedquire, line 479: maybe reuse the helper below?

iroh-bytes/src/store/traits.rs
rklaehn: Do you have an issue for that?

dignifiedquire: https://github.com/n0-computer/iroh/issues/1646 all things are linked there

dignifiedquire (approved):

PR #2112, opened 2024-03-22
refactor(iroh-sync): rip out the mem implementation of the doc store

rklaehn:
## Description

refactor(iroh-sync): rip out the mem implementation of the doc store

This would be a prerequisite for doing major changes on how docs work. Having 2 implementations that need to be kept in sync would slow that down way too much.

## Notes & open questions

How costly is creating a mem redb instance? Is this a big deal? Are we creating a mem instance and then throwing it away in the builder, or not?

Todo: remove the tests associated with the mem store.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


iroh-sync/src/store/fs.rs
```diff
@@ -115,13 +115,28 @@ pub struct Store {
     pubkeys: MemPublicKeyStore,
 }
 
+impl Default for Store {
+    fn default() -> Self {
```
dignifiedquire, line 119: this is pretty confusing, I think  I would prefer

- `Store::memory()`
- `Store::persistent(path)`

similar to how I did it on the node builder.

iroh-sync/src/store.rs
```diff
@@ -16,9 +16,7 @@ use crate::{
     AuthorId, Capability, CapabilityKind, Entry, NamespaceId, PeerIdBytes,
 };
 
-#[cfg(feature = "fs-store")]
 pub mod fs;
```
dignifiedquire, line 19: Should we rename this, `fs` seems like the wrong name? If we drop the trait we could export it just as `Store`

rklaehn: Not sure about dropping the trait, I think in the longer term we do want a trait here. Still unclear whether dropping the trait is a temp thing to refactor faster or a permanent thing...

dignifiedquire (approved):
I like it

PR #2113, opened 2024-03-22
refactor: make export a seperate operation from download

Frando:
## Description

This changes two things about how we expose Iroh in the RPC API:

* Add `iroh.blobs.export()` / `BlobExportRequest` to export a blob from the internal blob store to the local filesystem. We prominently missed this in our API.
* Add `blob export` to the CLI. The CLI also keeps the "export after download" functionality, it just issues a second RPC request for the export now.
* Remove the `out: DownloadLocation` field from `BlobDownloadRequest` and thus the functionality to automatically export a blob after the download finishes. Instead, users may call `iroh.blobs.export` themselves now after the download finishes. This does not make a difference performance-wise (apart from a single RPC roundtrip), because the export operation was already running sequentially after the download before this PR.

The untangling of export and download will simplify #2085 quite a bit, because at the moment the fact that the export progress events were emitted from outside the downloader introduced a complication in how we can emit progress events from the downloader (we had to remap the progress IDs to avoid conflicts). With this change, a download will only get progress event from the download itself, so this is solved much more simple.

## Notes & open questions

There's no progress reporting for the CLI yet. But we also didn't have that before, so I think it's fine to do as a followup.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-cli/src/commands/blob.rs
+++ b/iroh-cli/src/commands/blob.rs
@@ -16,7 +16,7 @@ use indicatif::{
 use iroh::bytes::{
     get::{db::DownloadProgress, Stats},
     provider::AddProgress,
-    store::{ExportMode, ValidateLevel, ValidateProgress},
+    store::{ExportFormat, ExportMode, ValidateLevel, ValidateProgress},
     BlobFormat, Hash, HashAndFormat, Tag,
 };
 use iroh::net::{key::PublicKey, relay::RelayUrl, NodeAddr};
@@ -263,8 +263,12 @@ impl BlobCommands {
                             true => ExportMode::TryReference,
                             false => ExportMode::Copy,
                         };
+                        let format = match recursive {
+                            true => ExportFormat::Collection,
+                            false => ExportFormat::Blob,
+                        };
                         tracing::info!("exporting to {} -> {}", path.display(), absolute.display());
-                        let stream = iroh.blobs.export(hash, absolute, recursive, mode).await?;
+                        let stream = iroh.blobs.export(hash, absolute, format, mode).await?;
                         // TODO: report export progress
                         stream.await?;
                     }
@@ -296,12 +300,16 @@ impl BlobCommands {
                             true => ExportMode::TryReference,
                             false => ExportMode::Copy,
                         };
+                        let format = match recursive {
+                            true => ExportFormat::Collection,
+                            false => ExportFormat::Blob,
+                        };
                         tracing::info!(
                             "exporting {hash} to {} -> {}",
                             path.display(),
                             absolute.display()
                         );
-                        let stream = iroh.blobs.export(hash, absolute, recursive, mode).await?;
+                        let stream = iroh.blobs.export(hash, absolute, format, mode).await?;
                         // TODO: report export progress
                         stream.await?;
                     }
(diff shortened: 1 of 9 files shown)


iroh/src/rpc_protocol.rs
```diff
@@ -139,6 +118,38 @@ impl ServerStreamingMsg<ProviderService> for BlobDownloadRequest {
 #[derive(Debug, Clone, Serialize, Deserialize, derive_more::From, derive_more::Into)]
 pub struct BlobDownloadResponse(pub DownloadProgress);
 
+/// A request to the node to download and share the data specified by the hash.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct BlobExportRequest {
+    /// The hash of the blob to export.
+    pub hash: Hash,
+    /// Set to true if the `hash` referes to a collection and all children of the collection should
```
rklaehn, line 126: Not a fan of "recursive" here. hash seqs are just 1 level deep, so if somebody is going to put HashSeqs in HashSeqs this won't do what they expect. Not sure what the best name is though.

rklaehn: Make it another enum, with values Blob and Collection? (this won't work for arbitrary HashSeqs)

iroh/src/client/blobs.rs
```diff
@@ -520,6 +554,92 @@ impl Future for BlobDownloadProgress {
...
+        self.stream.poll_next_unpin(cx)
+    }
+}
+
+impl Future for BlobExportProgress {
+    type Output = Result<BlobExportOutcome>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        loop {
+            match self.stream.poll_next_unpin(cx) {
```
rklaehn, line 620: I like the style of this API, but I don't like the manual poll impl. Any way to get the same thing without having to manually write poll, or at least reuse this somehow? I don't want our entire codebase to be full of this stuff. Writing new RPC methods is boilerplatey enough as it is.

rklaehn: We can also postpone this and make an issue.

iroh-cli/src/commands/blob.rs
```diff
@@ -216,53 +233,81 @@ impl BlobCommands {
...
+                        if matches!(format, BlobFormat::HashSeq) {
+                            ensure!(!absolute.is_dir(), "output must not be a directory");
+                        }
+                        let recursive = format == BlobFormat::HashSeq;
+                        let mode = match stable {
+                            true => ExportMode::TryReference,
+                            false => ExportMode::Copy,
+                        };
+                        tracing::info!("exporting to {} -> {}", path.display(), absolute.display());
+                        let stream = iroh.blobs.export(hash, absolute, recursive, mode).await?;
```
rklaehn, line 267: Not for this time, but this is not a generic HashSeq export but an export that will only work for a specific type of HashSeq, a collection.

rklaehn (approved):

iroh/src/client/blobs.rs
Frando: Yes let's move this to a followup. This applies to all these progress streams. I'm quite confident we can find an abstraction to reduce the boilerplate here.

Created an issue: https://github.com/n0-computer/iroh/issues/2114

iroh/src/rpc_protocol.rs
Frando: Sounds good. Added an enum.

PR #2117, opened 2024-03-22
refactor!: avoid using futures crate directly

dignifiedquire:
Due to the low maintenance and absurd high amount of `unsafe` code in parts of the `futures` crate, we are trying to avoid usage of it.

Usages are replaced with

- `futures-lite` : general `Future` and `Stream` tooling
- `futures-sink`: `Sink` trait
- `futures-buffered`: faster and safer version of `Futures{Un}Ordered`
- If must be `futures-util` for sink specific things that are missing

## Breaking Changes

- `iroh::node::Node` does not implement `Future` anymore
- `iroh::node::Node::shutdown()` is now `async` and can be awaited upon to wait for the node to exit
- `iroh_net::util::AbortingJoinHandle`s inner field is not public anymore, use the `From<JoinHandle>` implementation to contruct it

## Followups

- [x] Apply this to `bao-tree`: https://github.com/n0-computer/bao-tree/pull/49
- [x] Apply this to `iroh-io`: https://github.com/n0-computer/iroh-io/pull/6
- [x] Apply this to `quic-rpc`: https://github.com/n0-computer/quic-rpc/pull/73


Frando:
What about we do an `iroh_base::futures` module with reexports of the types from these crates? It is rather annoying to work with that many crates for basic concurrency utils. And we might change crates again, and then have to go through all iroh crates again.

iroh-cli/src/commands/#start.rs#
flub, line 1: hi emacs.  not the file we wanted in our repo...

flub (commented):
I think I'd be -1 on iroh_base::futures.

Maybe it's worth seeing if we can get futures denied by cargo-deny?

iroh-cli/src/commands/#start.rs#
dignifiedquire: 😭

flub:
is this still alive?  should we get it done?

dignifiedquire:
> is this still alive? should we get it done?

this is blocked on https://github.com/n0-computer/quic-rpc/pull/71

rklaehn:
Unfortunately quic-rpc is using quite a few things from futures

- sinks
- streams

Changing the futures dependency in quic-rpc to futures-lite would mean that it does not play well with crates using the full futures crate anymore, which would be not very nice. Also, if we just use futures-sink instead of futures directly aren't we still depending on code from the futures project? So what do we gain?

I can switch quic-rpc to a combination of futures-lite and importing futures-sink. But I seriously don't see the benefit for all the work. I started, but it's a lot. There is no Either in futures_lite, no select, no SinkExt in futures_sink etc.

rklaehn:
@dignifiedquire I pushed a fix to your compile errors

iroh/src/rpc_protocol.rs
```diff
@@ -572,7 +572,7 @@ impl Msg<ProviderService> for DocSubscribeRequest {
 impl TryServerStreamingMsg<ProviderService> for DocSubscribeRequest {
     type Item = DocSubscribeResponse;
     type ItemError = RpcError;
-    type CreateError = anyhow::Error;
+    type CreateError = RpcError;
```
rklaehn, line 575: You can't have anyhow::Error here because it needs to go over the wire.

iroh/src/rpc_protocol.rs
```diff
@@ -1224,6 +1224,7 @@ pub enum ProviderResponse {
     DocGetDownloadPolicy(RpcResult<DocGetDownloadPolicyResponse>),
     DocSetDownloadPolicy(RpcResult<DocSetDownloadPolicyResponse>),
     DocGetSyncPeers(RpcResult<DocGetSyncPeersResponse>),
+    StreamCreated(RpcResult<StreamCreated>),
```
rklaehn, line 1227: You need to have a way to send the message when the stream is created, otherwise there is no way to distinguish between "error on stream creation" and "error producing first item".

dignifiedquire: got it, thanks

iroh/src/node.rs
```diff
@@ -250,15 +253,6 @@ impl<D: BaoStore> Node<D> {
     }
 }
 
-/// The future completes when the spawned tokio task finishes.
-impl<D> Future for Node<D> {
```
rklaehn, line 254: I don't mind losing this at all.

rklaehn (approved):
It would be nice if you could constrain futures_lite to the imports. But if you do use futures_lite::future that collides with the import of std::future that you sometimes have to do to use the anemic standardized futures api. So 🤷. If futures_lite would reexport the std futures stuff this would not be an issue, but I guess they did not want to do that.

Also I really don't get what was wrong with the BoxFuture and BoxStream type aliases in futures.

PR #2120, opened 2024-03-23
feat: update redb to v2

Frando:
## Description

`redb` 2.0.0 was released: https://github.com/cberner/redb/releases/tag/v2.0.0

most relevant for us: 
> 2.0.0 uses a new file format that optimizes len() to be constant time. This means that it is not backwards compatible with 1.x. To upgrade, consider using a pattern like that shown in the [upgrade_v1_to_v2](https://github.com/cberner/redb/blob/222a37f4600588261b0983eebcd074bb69d6e5a0/tests/backward_compatibility.rs#L282-L299) test.

and some nice API changes that will simplify the iroh-sync codebase (no more ouroboros for self-refential structs needed!)

> * Remove lifetimes from read-only tables
> * Remove lifetime from WriteTransaction and ReadTransaction

and some further API changes, see the notes.

---

* [x] Update iroh-bytes to redb v2
  * Did some crude search&replaces to remove now unneeded lifetime bounds. Compiles now, let's see what test says.
* [x] Migration for iroh-bytes db
   * Code adapted from here https://github.com/n0-computer/migrate-bao-store-redb/blob/main/src/main.rs
* [x] Update  iroh-sync to redb v2
   * This will be a bit more involved as iroh-bytes because we hit more API changes, but also great because we can remove `ouroborous` because iterators can now own a range
* [x] Migration for iroh-sync db

The migrations likely need a dependency to redb v1 to read the old database, so at least for one release cycle we'll have to depend on both redb v2 and v1.

## Notes & open questions

This currently includes an extra-safeguard by preserving the old databases for sync and blobs as `docs.redb.backup-redb-v1` and `blobs.db-backup-redb-v1` - do we want this or delete the backups after successfull migration?

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.


iroh-bytes/src/store/fs/migrate_redb_v1_v2.rs
```diff
@@ -0,0 +1,321 @@
+use std::path::{Path, PathBuf};
+
+use anyhow::Result;
+use redb_v1::ReadableTable;
+use tempfile::NamedTempFile;
+use tracing::info;
+
+pub fn run(source: impl AsRef<Path>) -> Result<redb::Database> {
+    let source = source.as_ref();
+    // create the database to a tempfile
```
dignifiedquire, line 10: I think we should use a tempfile next to the source directory, thinking about multiple disk setups, and the db being large, you want to make sure you stay on the same disk

iroh-bytes/src/store/fs/migrate_redb_v1_v2.rs
```diff
@@ -0,0 +1,321 @@
...
+        let old_tags = rtx.open_table(old::TAGS_TABLE)?;
+        let mut new_tags = wtx.open_table(new::TAGS_TABLE)?;
+        let len = old_tags.len()?;
+        info!("migrate tags table ({len} rows)");
+        for (i, entry) in old_tags.iter()?.enumerate() {
+            let (key, value) = entry?;
+            let key = key.value();
+            let value: crate::HashAndFormat = value.value().into();
+            if i > 0 && i % 100 == 0 {
+                info!("    row {i:>6} of {len}");
```
dignifiedquire, line 44: every 100 seems a lot, maybe every 1000?

iroh-sync/src/store/fs/migrate_v1_v2.rs
```diff
@@ -0,0 +1,139 @@
...
+    let target = NamedTempFile::new()?;
+    let target = target.into_temp_path();
+    info!("migrate {} to {}", source.display(), target.display());
+    let old_db = redb_v1::Database::open(source)?;
+    let new_db = redb::Database::create(&target)?;
+
+    let rtx = old_db.begin_read()?;
+    let wtx = new_db.begin_write()?;
+
+    migrate_table!(rtx, wtx, old::AUTHORS_TABLE, new::AUTHORS_TABLE);
```
dignifiedquire, line 65: why not a function?

iroh-sync/src/store/fs/migrate_v1_v2.rs
```diff
@@ -0,0 +1,139 @@
...
+            }
+        }
+        info!("migrate {name} done");
+    }};
+}
+
+pub fn run(source: impl AsRef<Path>) -> Result<redb::Database> {
+    let source = source.as_ref();
+    // create the database to a tempfile
+    let target = NamedTempFile::new()?;
```
dignifiedquire, line 56: same as above

iroh-sync/src/store/fs/migrate_v1_v2.rs
Frando: It would need a lot of nested generics and I was a bit lazy. But can change I think.

rklaehn:
FYI I tried this out locally, works fine except that now my store is stuck on 2.0 :-)

rklaehn (approved):

rklaehn:
We should probably wait a bit more with merging this in case there are some issues with redb v2. One week or so?

dignifiedquire (approved):

dignifiedquire:
> I'm wondering - would this also be a good time to remove old db migrations in iroh-sync, and panic with a message like "You need to update to iroh v0.12 before updating to v0.13"?

seems reasonable

iroh-sync/src/store/fs/migrate_v1_v2.rs
```diff
@@ -0,0 +1,140 @@
...
+        let mut new_table = $wtx.open_multimap_table($new)?;
+        let name = $new.name();
+        let len = old_table.len()?;
+        info!("migrate {name} ({len} rows)");
+        let ind = (len as usize / 1000) + 1;
+        for (i, entry) in old_table.iter()?.enumerate() {
+            let (key, values) = entry?;
+            let key = key.value();
+            if i > 0 && i % 1000 == 0 {
+                info!("    {name} {i:>ind$}/{len}");
```
ppodolsky, line 42: Screenshot 2024-04-04 at 17 05 04

Dunno how it is supposed to work, but it looks like this

rklaehn:
I looked at redb issues. Nothing was opened in the last week, and also no closed issues related to the 1.0 -> 2.0 transition. So maybe now is time to merge it?

PR #2121, opened 2024-03-25
fix: Remove dead code detected by the rust beta compiler

flub:
## Description

The beta compiler identifies these as dead code.

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh-bytes/src/store/traits.rs
+++ b/iroh-bytes/src/store/traits.rs
@@ -2,14 +2,14 @@
 use std::{collections::BTreeSet, io, path::PathBuf};
 
 use bao_tree::{
-    io::fsm::{BaoContentItem, Outboard, OutboardMut},
+    io::fsm::{BaoContentItem, Outboard},
     BaoTree, ByteNum, ChunkRanges,
 };
 use bytes::Bytes;
-use futures::{future, Future, Stream, StreamExt};
+use futures::{Future, Stream, StreamExt};
 use genawaiter::rc::{Co, Gen};
 use iroh_base::rpc::RpcError;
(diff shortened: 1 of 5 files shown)


iroh-sync/src/ranger.rs
```diff
@@ -40,13 +40,13 @@ pub trait RangeEntry: Debug + Clone {
...
+    // fn is_prefix_of(&self, other: &Self) -> bool;
 
-    /// Returns true if `other` is a prefix of `self`.
-    fn is_prefixed_by(&self, other: &Self) -> bool {
-        other.is_prefix_of(self)
-    }
+    // /// Returns true if `other` is a prefix of `self`.
+    // fn is_prefixed_by(&self, other: &Self) -> bool {
+    //     other.is_prefix_of(self)
+    // }
```
flub, line 49: @Frando What is the deal with this trait?

flub: oh, it's used in tests

flub: @Frando I think I've made this happy for the compiler.  But could you make sure it does the right thing?

flub:
@Frando  could you, err,  approve in that case?

iroh-bytes/src/store/traits.rs
```diff
@@ -216,43 +216,6 @@ impl<W: BaoBatchWriter, F: Fn(u64, usize) -> io::Result<()> + 'static> BaoBatchW
     }
 }
 
-/// A combined batch writer
```
rklaehn, line 219: Yeah, this is dead since we have a proper batch writer. In some branch I removed this already.

rklaehn (approved):

Frando (approved):

PR #2122, opened 2024-03-25
fix(ci): Also test iroh-cli for cargo features

flub:
## Description

We really ought to use something like cargo-metadata to pull this
information from the cargo.toml file.  Maybe one day one of us will be
motivated enough to do this.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


dignifiedquire (approved):
thanks

Arqu (approved):
👍

PR #2123, opened 2024-03-25
fix: do not allow connecting to ourself

Frando:
## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/src/magic_endpoint.rs
+++ b/iroh-net/src/magic_endpoint.rs
@@ -411,6 +411,14 @@ impl MagicEndpoint {
 
         let NodeAddr { node_id, info } = node_addr;
 
+        // Connecting to ourselves is not supported.
+        if node_id == self.node_id() {
+            bail!(
+                "Connecting to ourself is not supported ({} is the node id of this node)",
+                node_id.fmt_short()
+            );
+        }
+
         // Get the mapped IPv6 address from the magic socket. Quinn will connect to this address.
         let (addr, discovery) = match self.msock.get_mapping_addr(&node_id) {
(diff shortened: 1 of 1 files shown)


iroh-net/src/magic_endpoint.rs
```diff
@@ -411,6 +411,14 @@ impl MagicEndpoint {
 
         let NodeAddr { node_id, info } = node_addr;
 
+        // Connecting to ourselves is not supported.
+        if node_id == self.node_id() {
```
flub, line 415: Should you not do this before calling `.add_node_addr()`?

flub (commented):
I wonder if it's possible to deny this on the `MagicSock` layer.  But that's possibly rather more difficult to do.

iroh-net/src/magic_endpoint.rs
Frando: Should we fail for `add_node_addr` with node_id == self.node_id too?

flub: yes, i think that would also be good.  in which case you could just rely on that to fail here.

flub (approved):
In lieu of being able to do it entirely in magicsock this is already good enough.

iroh-net/src/magic_endpoint.rs
Frando: Added this. We can't only check there, because add_node_addr is only called if the NodeAddr has non-empty AddrInfo. But we also don't want to start discovery for ourself is AddrInfo is empty, so two checks for now.

PR #2131, opened 2024-03-30
fix(iroh): shutdown sync engine on iroh node shutdown

dignifiedquire:
- make sure the sync engine is shutdown in time
- enable `fs-store` by default


Frando (approved):
LGTM

PR #2132, opened 2024-03-31
feat(iroh): implement basic author api

dignifiedquire:
- `author.export`
- `author.import`
- `author.delete`

### DIFF
--- a/iroh/src/client/authors.rs
+++ b/iroh/src/client/authors.rs
@@ -1,9 +1,12 @@
 use anyhow::Result;
 use futures::{Stream, TryStreamExt};
-use iroh_sync::AuthorId;
+use iroh_sync::{Author, AuthorId};
 use quic_rpc::{RpcClient, ServiceConnection};
 
-use crate::rpc_protocol::{AuthorCreateRequest, AuthorListRequest, ProviderService};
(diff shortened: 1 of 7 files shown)


iroh/src/client/authors.rs
```diff
@@ -28,4 +31,79 @@ where
...
+    ///
+    /// Warning: This contains sensitive data.
+    pub async fn export(&self, author: AuthorId) -> Result<Option<Author>> {
+        let res = self.rpc.rpc(AuthorExportRequest { author }).await??;
+        Ok(res.author)
+    }
+
+    /// Import the given author.
+    ///
+    /// Warning: This contains sensitive data.
```
Frando, line 45: remove

Frando (approved):

iroh/src/client/authors.rs
dignifiedquire: why? importing contains sensitive data as well

Frando: Ah, I thought it was a copy-paste error, because I related the `contain` to a return value.
So fine to leave as-is. Or we rephrase the comments to state what and where sensitive data is involved.

PR #2135, opened 2024-04-02
chore: move metrics init into CLI

Arqu:
## Description

Moved the metrics init into CLI so the lib doesn't init the full metrics collection on its own and block other lib users to do their custom setup on top.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


Arqu:
Closing in favor of https://github.com/n0-computer/iroh/pull/2136

PR #2140, opened 2024-04-03
docs(gossip): add comment about message uniqueness

Septias:
This would have spared me some debugging if I knew beforehand.


dignifiedquire (approved):
thanks

Issue #2149, opened 2024-04-04
RFC: `_iroh` DNS records and pkarr signed packets


Frando:
This is a proposal for spec on how Iroh uses the [Domain Name System (DNS)](https://datatracker.ietf.org/doc/html/rfc1035) and [Pkarr](https://github.com/Nuhvi/pkarr/) for node discovery.

Implementation in iroh is at #2045, and [`iroh-dns-server`](https://github.com/n0-computer/iroh-dns-server/), and a first discussion on the design happened in n0-computer/iroh-dns-server#5.

This is a first draft without much focus on edge cases, but wanted to condense the discussion into something we can finalize.

## Node discovery via DNS

When connecting to an unknown `NodeId`, the DNS discovery mechanism in Iroh will perform a DNS query to discover relays or addresses for the node. The DNS discovery is configured with a *origin domain* (which defaults to *dns.iroh.link*, a server run by n0). Iroh will then perform a DNS query through the DNS resolver (which defaults to using the host system's nameservers): 

`_iroh.<z32-node-id>.<origin-domain> TXT`

* `_iroh` is the record name defined in this spec
* `<z32-node-id>` is the [z32](https://crates.io/crates/z32) encoding of the 32-byte long `NodeId` (which is a string of 52 characters)
* `<origin-domain>` is the configured origin domain
* `TXT`  is the queried record type

The returned TXT records must contain a string value of the form `key=value`, as defined in [RFC1464](https://www.rfc-editor.org/rfc/rfc1464). If a record contains multiple character strings, they will be concatenated first. All characters after the first `=` character form the value string, and those before the first `=` are the key. The attributes from all valid TXT records will be merged, and each key can therefore have either one or multiple values.

This spec defines the following attributes:

* `relay=<url>`: The home relay for this node, e.g. `https://euw1-1.derp.iroh.network`.
* `addr=<addr> <addr> ..` A space-seperated list of socket addresses for this Iroh node. Each address is an IPv4 or IPv6 address with a port (e.g. `1.2.3.4:7367` or `[::1]:3521`

## Node announces via `pkarr`

Nodes announce their address information in [Pkarr](https://github.com/Nuhvi/pkarr/) signed packets. The TXT records, as described above, are added to the `answers` section of a DNS section as specified by pkarr. Their name must be `_iroh.<z32-node-id>.` The z32 encoded node ID must be the root name, no other origin but `.` (the single dot) is permitted.

[...]

PR #2150, opened 2024-04-04
refactor(dns): use one DNS record per attribute

Frando:
Adapts #2045 to match #2149 

* change the DNS and pkarr encoding for the NodeInfo to use one DNS record for each attribue
* remove the node= attribute as the node id is contained in the DNS name

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/src/dns/node_info.rs
+++ b/iroh-net/src/dns/node_info.rs
@@ -161,7 +161,7 @@ impl NodeInfo {
         &self,
         origin: &str,
         ttl: u32,
-    ) -> Result<Vec<hickory_proto::rr::Record>> {
+    ) -> Result<impl Iterator<Item = hickory_proto::rr::Record>> {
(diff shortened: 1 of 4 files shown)


iroh-net/src/dns/node_info.rs
```diff
@@ -251,17 +180,137 @@ pub(crate) fn parse_hickory_node_info_name(name: &hickory_proto::rr::Name) -> Op
     Some(node_id)
 }
 
-fn fmt_attr(label: &str, value: impl fmt::Display) -> String {
-    format!("{label}={value}")
+struct NodeAttrs {
+    node_id: NodeId,
+    attrs: BTreeMap<String, Vec<String>>,
```
dignifiedquire, line 185: (can be a follow up) can we make the keys be an enum, for better type saftey in our code?

dignifiedquire (approved):
looks reasonable

Issue #1437, opened 2023-08-31
Optimize document store and sync algorithm


rklaehn:
The sync algorithm currently reads large parts of the store in to memory. E.g.

```rust
// TODO: This is hugely inefficient and needs to be optimized
// For an identity range that includes everything we allocate a vec with all entries of
// the replica here.
let local_values: Vec<_> = self
    .store
    .get_range(range.clone())?
    .collect::<Result<_, _>>()?;
```

this will be fine for a few 1000 entries, but it won't scale for really large documents. We should make sure that iroh provides few but load-bearing primitives, and currently iroh docs (unlike e.g. blobs) are not load-bearing. As soon as you build something big on top, they will crumble.

I think we can come up with a trait for the store that allows implementing an efficient mem store and disk store, and refactor the sync algo (`process_message`) to use this trait.

rklaehn:
I think the first thing we must do is to make sure that keys are basically newtypes for `&[u8]`. That way we ensure that we can use a lex sorted kv store or a u8 radix tree and get the correct order.

PR #2152, opened 2024-04-04
perf(iroh-sync): avoid allocating a full range of values during sync

dignifiedquire:
Ref #1437

This does not optimize for speed yet, but should drastically lower the memory requirements for large documents.


rklaehn (approved):
Yeah, that should help a bit...

PR #2155, opened 2024-04-05
(wip) feat(iroh-sync): Inline small content

Frando:
## Description

WIP

* Refactor interface between iroh-sync and a content store, use a trait instead of arc'ed callbacks
* Inline small content (< 1024 bytes) during set reconciliation and in gossip messages
* Store incoming inlined content in the content store

## Notes & open questions

* I updated the `sync_full_basic` test as a smoke test.  Most other integration tests are still broken because we don't emit  `ContentReady` events if the content is already available (marked as `Complete` ) in the `RemoteInsert` event. `ContentReady`  is only emitted if the content is not yet available at the time when the RemoteInsert is emitted.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-bytes/src/store/fs.rs
+++ b/iroh-bytes/src/store/fs.rs
@@ -1399,6 +1399,15 @@ impl super::Store for Store {
         Ok(tokio::task::spawn_blocking(move || this.import_bytes_sync(data, format)).await??)
     }
 
+    fn import_bytes_sync(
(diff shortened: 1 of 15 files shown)


dignifiedquire:
Closing, as this will be done as part of the reworking to the updated willow protocol in documents

PR #2156, opened 2024-04-05
[WIP] feat: use zerocopy types for defining underlying docs fs types

dignifiedquire:
First round of using zerocopy in the sync db. 

## Open tasks

- this is missing a migration currently
- in a next round, after redb@v2 is merged, it should be possible to reduce some of the current allocations further (hopefully)


dignifiedquire:
Will be included in future work

Issue #2143, opened 2024-04-04
iroh-bytes: cleaner shutdown


rklaehn:
During shutdown of the bao file store, occasionally you get errors like this:

2024-04-04T08:11:08.488738Z ERROR iroh::node::builder: Error send error: sending on a closed channel starting GC, skipping GC to be safe

This can be triggered by running GC very often, as is done in some tests.

This is due to an unclean shutdown. The store actor thread is already stopped, but the gc loop is still running. Until it also has stopped it will produce such garbage warnings.

dignifiedquire:
seeing alot of these when running regular tests

```
rror: sending on a closed channel starting GC, skipping GC to be safe
2024-04-04T15:54:19.582569Z ERROR iroh::node::builder: Error send error: sending on a closed channel starting GC, skipping GC to be safe
2024-04-04T15:54:19.582573Z ERROR iroh::node::builder: Error send error: sending on a closed channel starting GC, skipping GC to be safe
2024-04-04T15:54:19.582577Z ERROR iroh::node::builder: Error send error: sending on a closed channel starting GC, skipping GC to be safe
2024-04-04T15:54:19.582581Z ERROR iroh::node::builder: Error send error: sending on a closed channel starting GC, skipping GC to be safe
2024-04-04T15:54:19.582585Z ERROR iroh::node::builder: Error send error: sending on a closed channel starting GC, skipping GC to be safe
2024-04-04T15:54:19.582588Z ERROR iroh::node::builder: Error send error: sending on a closed channel starting GC, skipping GC to be safe
2024-04-04T15:54:19.582647Z ERROR iroh::node::builder: Error send error: sending on a closed channel starting GC, skipping GC to be safe
2024-04-04T15:54:19.582652Z ERROR iroh::node::builder: Error send error: sending on a closed channel starting GC, skipping GC to be safe
2024-04-04T15:54:19.582656Z ERROR iroh::node::builder: Error send error: sending on a closed channel starting GC, skipping GC to be safe
2024-04-04T15:54:19.582659Z ERROR iroh::node::builder: Error send error: sending on a closed channel starting GC, skipping GC to be safe
2024-04-04T15:54:19.582663Z ERROR iroh::node::builder: Error send error: sending on a closed channel starting GC, skipping GC to be safe
2024-04-04T15:54:19.582667Z ERROR iroh::node::builder: Error send error: sending on a closed channel starting GC, skipping GC to be safe
[...]

PR #2158, opened 2024-04-08
fix(iroh): do not shut down node on internal rpc error

### LINKED ISSUE #2157: iroh shuts down early due to internal rpc error
We are getting an error on the internal RPC, possibly due to the "remote" future being dropped. And that is causing the entire node to shut down.

Options to fix this are

1. try to make sure we don't get that error.
2. have more explicit shutdown and don't shut down on internal rpc error at all.

```
ERROR node: last controller dropped due to error EarlyClose, shutting down me=...
```

rklaehn:
## Description

fix(iroh): do not shut down node on internal rpc error

Internal rpc errors can happen, e.g. when the client side gets dropped. No reason to shut down the entire node. We already have a cancellation token, so it's not like we rely on this for shutdown.

Fixes #2157, #2143

## Notes & open questions

Note: I also sneaked in a fix for #2143. Basically when we are unable to notify the db that a new gc epoch has started, we just shot down the gc loop, since clearly the store is not well (probably shut down).

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/node/builder.rs
+++ b/iroh/src/node/builder.rs
@@ -458,8 +458,8 @@ where
                         Ok((msg, chan)) => {
                             handler.handle_rpc_request(msg, chan);
                         }
-                        Err(_) => {
-                            info!("last controller dropped, shutting down");
+                        Err(e) => {
+                            info!("internal rpc request error: {:?}", e);
                             break;
                         }
                     }


iroh/src/node/builder.rs
```diff
@@ -458,8 +458,8 @@ where
                         Ok((msg, chan)) => {
                             handler.handle_rpc_request(msg, chan);
                         }
-                        Err(_) => {
-                            info!("last controller dropped, shutting down");
+                        Err(e) => {
+                            info!("internal rpc request error: {:?}", e);
                             break;
```
dignifiedquire, line 463: you are still breaking, how does this fix things?

rklaehn: oops, i wasn't intending to. got this back somehow while switching branches.

dignifiedquire (approved):

PR #2160, opened 2024-04-08
feat: mark iroh-cli's iroh binary as workspace default

flub:
# Description

This makes sure that when you call `cargo run` in the workspace root
the `iroh` binary from the iroh-cli crate will be chosen if no
additional --bin parameter is passed.

## Notes & open questions

If we are de-prioritising the iroh cli maybe this is the wrong default
to set?

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


dignifiedquire (approved):
thanks, I forgot about this

dignifiedquire:
> If we are de-prioritising the iroh cli maybe this is the wrong default
> to set?

no, this is good, as long as it still lives here

rklaehn:
Thank you. Wanted to figure out how to do this today!

PR #2161, opened 2024-04-09
feat(iroh-net): add `MagicEndpoint::conn_type_stream` returns a stream that reports connection type changes for a `node_id`

ramfox:
## Description
`MagicEndpoint::conn_type_stream` returns a `Stream` that reports changes for a `magicsock::Endpoint` with a given `node_id` in a `magicsock::NodeMap`.

It will error if no address information for that `node_id` exists in the `NodeMap`.

This PR also adjusts the `Endpoint::info()` method to use the same `ConnectionType` that gets reported to the stream.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/src/magic_endpoint.rs
+++ b/iroh-net/src/magic_endpoint.rs
@@ -402,6 +402,19 @@ impl MagicEndpoint {
         self.connect(addr, alpn).await
     }
 
+    /// Returns once we have a direct connection to the peer at `node_id` or timed out.
+    ///
+    /// Can be called before or after `MagicEndpoint::connect`
+    pub async fn wait_for_direct_connection(
+        &self,
+        node_id: &PublicKey,
+        timeout: Duration,
+    ) -> Result<()> {
+        let r = self.msock.notify_direct_conn(node_id);
+        tokio::time::timeout(timeout, r).await??;
+        Ok(())
+    }
+
     /// Connect to a remote endpoint.
     ///
     /// A [`NodeAddr`] is required. It must contain the [`NodeId`] to dial and may also contain a
@@ -630,7 +643,7 @@ mod tests {
     use rand_core::SeedableRng;
     use tracing::{error_span, info, info_span, Instrument};
 
-    use crate::test_utils::run_relay_server;
+    use crate::{magicsock::ConnectionType, test_utils::run_relay_server};
 
     use super::*;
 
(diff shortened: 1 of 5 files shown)


iroh-net/src/magicsock/node_map/endpoint.rs
```diff
@@ -286,6 +294,28 @@ impl Endpoint {
...
 
+    /// Returns a [`oneshot::Receiver`] that will be alerted once we have a valid `best_addr`
+    ///
+    /// If we already have a valid `best_addr` for this endpoint, the [`oneshot::Receiver`]
+    /// will recv immediately.
+    ///
+    /// It is possible we will never have a viable `best_addr`, in which case
+    /// no alert will ever be issued.
+    pub fn notify_has_best_addr(&mut self) -> oneshot::Receiver<()> {
+        self.assign_best_addr_from_candidates_if_empty();
```
flub, line 305: Do you really want to call this from here?  I'm kind of -1 because it is logic that influences how the magicsock behaves but is called from a place that's not in the normal flow.  The magicsock will already call this from somewhere else when it thinks this is needed.

iroh-net/src/magic_endpoint.rs
```diff
@@ -971,4 +990,70 @@ mod tests {
         p1_connect.await.unwrap();
         p2_connect.await.unwrap();
     }
+
+    #[tokio::test]
+    async fn magic_endpoint_wait_for_direct_conn() {
+        let _logging_guard = iroh_test::logging::setup();
+        let (relay_map, relay_url, _relay_guard) = run_relay_server().await.unwrap();
+        let mut rng = rand_chacha::ChaCha8Rng::seed_from_u64(42);
```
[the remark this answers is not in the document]
ramfox: 🤣 TBH, this was taken from another test & I didn't change it. I'm assuming it was originally 42 because 42 is the answer to life the universe and everything

iroh-net/src/magicsock/node_map/endpoint.rs
ramfox: good point. if we go back to that flow, I'll remove it.

iroh-net/src/magicsock/node_map/endpoint.rs
```diff
@@ -171,6 +174,7 @@ impl Endpoint {
             direct_addr_state: BTreeMap::new(),
             last_used: options.active.then(Instant::now),
             last_call_me_maybe: None,
+            conn_type: Watchable::new(ConnectionType::None),
```
ramfox, line 177: my biggest concern with this approach is how "heavy" it might be to have a `Watchable` for each `node_map::Endpoint`

iroh-net/src/magic_endpoint.rs
```diff
@@ -971,4 +981,96 @@ mod tests {
...
+        assert_eq!(ep2_nodeid, got_id);
+
+        res_ep1.await.unwrap().unwrap();
+        res_ep2.await.unwrap().unwrap();
+
+        let conn_info = ep1.connection_info(ep2.node_id()).unwrap();
+        match conn_info.conn_type {
+            ConnectionType::Direct(_) => {}
+            _ => panic!("expected direct connection, got {:?}", conn_info.conn_type),
+        }
```
flub, line 1074: I'm mildly worried that due to #2169 this could make this test flaky.  The `res_ep1` resulting in an `Ok` is already sufficient as assert maybe?

iroh-net/src/magicsock/node_map.rs
```diff
@@ -389,6 +405,25 @@ impl NodeMapInner {
             .map(|ep| ep.info(Instant::now()))
     }
 
+    /// Returns a stream of [`ConnectionType`].
+    ///
+    /// Sends the current [`ConnectionType`] whenever any changes to the
+    /// connection type for `public_key` has occured.
+    ///
+    /// Errors:
```
flub, line 413: FWIW rust doc conventions/rules/style-guides/whatever use `# Errors` for this.

iroh-net/src/magicsock/node_map.rs
```diff
@@ -536,6 +571,29 @@ impl NodeMapInner {
...
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let this = &mut *self;
+        if let Some(initial_conn_type) = this.initial.take() {
+            return Poll::Ready(Some(initial_conn_type));
+        }
+        match Pin::new(&mut this.inner).poll_next(cx) {
+            Poll::Pending => Poll::Pending,
+            Poll::Ready(Some(conn_type)) => Poll::Ready(Some(conn_type)),
+            Poll::Ready(None) => Poll::Ready(None),
+        }
```
flub, line 593: Isn't this the same as directly `Pin::new(&mut this.inner).poll_next(cx)`?

iroh-net/src/magicsock/node_map/endpoint.rs
flub: It does acquire an additional `RwLock` write handle for each time you set it.  Same would be for a tokio `Watch`.  Though I think there would be very little contention for this, so it's probably fine?

flub (approved):

iroh-net/src/magic_endpoint.rs
ramfox: oooo good call. I don't think we need this assertion anymore, I was using it when I was relying on the oneshot senders and wanted extra assurance that everything was sent at the right time.

iroh-net/src/magicsock/node_map.rs
ramfox: oops thank you, that was what i was going for

iroh-net/src/magicsock/node_map.rs
ramfox: lmao yep thank you.

PR #2163, opened 2024-04-09
fix(iroh-cli):  doctor relay-urls, correct connection logic

dignifiedquire:


rklaehn (approved):
So the issue was that after round 1 this assertion was failing?

```rust
assert!(!client.is_connected().await?);
```

because the clients were already connected?

dignifiedquire:
yes, because of recent changes the connection is established more eager

PR #2164, opened 2024-04-09
perf(iroh-net): simplify relay handshake

dignifiedquire:
The relay handshake is quite expensive currently

- TLS 1.3 + HTTP 1 UPGRADE
- Server sends FrameType::ServerKey
- Client sends FrameType::ClientInfo
- Server sends FrameType::ServerInfo

This simplifies the protocol to 

- TLS 1.3 + HTTP 1 UPGRADE
- Client sends FrameType::ClientInfo

information changes
- using a signature, instead of a shared secret to ensure the client identity. 
- server key is not sent (use certificate pinning if fixed relays are important)
- remove unused configuration in the info frames
- change magic from `derp` to `relay`
- increase protocol version to `3`
- drop `serverinfo` in favor of fixed config on the client
- enable tcp nodelay
- switch to AES for encryption in favor over chacha
- the server simply aborts when the client version doesn't match (matching what we do in other protocols)

Benchmarks:
https://gist.github.com/dignifiedquire/30131dbe8b87fb799a971068899656ef

BREAKING: 
This breaks the relay handshake, and so requires updated relay nodes to work with.

### DIFF
--- a/iroh-net/src/relay/codec.rs
+++ b/iroh-net/src/relay/codec.rs
@@ -175,9 +175,10 @@ pub(super) async fn recv_client_key<S: Stream<Item = anyhow::Result<Frame>> + Un
         signature,
     } = buf
     {
-        client_public_key.verify(&message, &signature).context("invalid signature")?;
-        let info: ClientInfo =
-            postcard::from_bytes(&message).context("deserialization")?;
+        client_public_key
+            .verify(&message, &signature)
+            .context("invalid signature")?;
+        let info: ClientInfo = postcard::from_bytes(&message).context("deserialization")?;
         Ok((client_public_key, info))
     } else {
         anyhow::bail!("expected FrameType::ClientInfo");
@@ -251,9 +252,7 @@ impl Frame {
                 client_public_key: _,
                 message,
                 signature: _,
-            } => {
-                MAGIC.as_bytes().len() +  PUBLIC_KEY_LENGTH + message.len() + Signature::BYTE_SIZE
-            },
+            } => MAGIC.as_bytes().len() + PUBLIC_KEY_LENGTH + message.len() + Signature::BYTE_SIZE,
             Frame::ServerInfo { message } => message.len(),
             Frame::SendPacket { dst_key: _, packet } => PUBLIC_KEY_LENGTH + packet.len(),
             Frame::RecvPacket {
@@ -276,7 +275,7 @@ impl Frame {
             Frame::ClientInfo {
                 client_public_key,
                 message,
-                signature
+                signature,
             } => {
                 dst.put(MAGIC.as_bytes());
                 dst.put(client_public_key.as_ref());
@@ -328,7 +327,8 @@ impl Frame {
         let res = match frame_type {
             FrameType::ClientInfo => {
                 ensure!(
-                    content.len() >= PUBLIC_KEY_LENGTH + Signature::BYTE_SIZE + MAGIC.as_bytes().len(),
+                    content.len()
+                        >= PUBLIC_KEY_LENGTH + Signature::BYTE_SIZE + MAGIC.as_bytes().len(),
                     "invalid client info frame length: {}",
                     content.len()
                 );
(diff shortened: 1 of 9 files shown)


iroh-net/src/relay/codec.rs
```diff
@@ -63,12 +58,8 @@ const NOT_PREFERRED: u8 = 0u8;
 #[derive(Debug, PartialEq, Eq, num_enum::IntoPrimitive, num_enum::FromPrimitive, Clone, Copy)]
 #[repr(u8)]
 pub(crate) enum FrameType {
-    /// 8B magic + 32B public key + (0+ bytes future use)
-    ServerKey = 1,
-    /// 32b pub key + 24B nonce + chachabox(bytes)
+    /// magci + 32b pub key + 24B nonce + bytes
```
flub, line 61:
```suggestion
    /// magic + 32b pub key + 24B nonce + bytes
```

flub (commented):
IIUC the idea is to deploy this under a new URL?

I think this looks fine, there's still a TODO though

ramfox:
Haven't done a read-through yet, but just looking at the protocol changes: we haven't been using rate limiters for the client yet, but we should talk about what we might do to replace it in the future if we no longer have a `ServerInfo`. Or is this something we can get away with never doing?

iroh-net/src/relay/codec.rs
```diff
@@ -16,11 +16,8 @@ pub const MAX_PACKET_SIZE: usize = 64 * 1024;
 
 const MAX_FRAME_SIZE: usize = 1024 * 1024;
 
-/// The DERP magic number, sent in the FrameType::ServerKey frame
-/// upon initial connection
-///
-/// 8 bytes: 0x44 45 52 50 f0 9f 94 91
-const MAGIC: &str = "DERP🔑";
+/// The Relay magic number, sent in the FrameType::ClientInfo frame upon initial connection.
+const MAGIC: &str = "RELAY🔑";
```
ramfox, line 20: oooo spicy

iroh-net/src/relay/codec.rs
```diff
@@ -158,9 +150,8 @@ pub(crate) async fn send_client_key<S: Sink<Frame, Error = std::io::Error> + Unp
 /// Reads the `FrameType::ClientInfo` frame from the client (its proof of identity)
 /// upon it's initial connection.
 pub(super) async fn recv_client_key<S: Stream<Item = anyhow::Result<Frame>> + Unpin>(
-    secret_key: SecretKey,
     stream: S,
-) -> anyhow::Result<(PublicKey, ClientInfo, SharedSecret)> {
+) -> anyhow::Result<(PublicKey, ClientInfo)> {
     // the client is untrusted at this point, limit the input size even smaller than our usual
     // maximum frame size, and give a timeout
```
ramfox, line 157: These and the TODO below are no longer relevant after the `codec` refactor

iroh-net/src/relay/http/client.rs
```diff
@@ -812,6 +805,8 @@ impl Actor {
             .map_err(|_| ClientError::ConnectTimeout)?
             .map_err(ClientError::DialIO)?;
 
+        tcp_stream.set_nodelay(true)?;
```
ramfox, line 808: 🔥

iroh-net/src/relay/codec.rs
```diff
@@ -35,16 +32,14 @@ pub(super) const PER_CLIENT_READ_QUEUE_DEPTH: usize = 512;
 /// NOTE: we are techincally running a modified version of the protocol.
 /// `FrameType::PeerPresent`, `FrameType::WatchConn`, `FrameType::ClosePeer`, have been removed.
 /// The server will error on that connection if a client sends one of these frames.
-pub(super) const PROTOCOL_VERSION: usize = 2;
+pub(super) const PROTOCOL_VERSION: usize = 3;
```
ramfox, line 35:
```suggestion
/// We have split with the DERP protocol significantly starting with our relay protocol 3
/// `FrameType::PeerPresent`, `FrameType::WatchConn`, `FrameType::ClosePeer`, `FrameType::ServerKey`, and `FrameType::ServerInfo` have been removed.
/// The server will error on that connection if a client sends one of these frames.
/// This materially affects the handshake protocol, and so relay nodes on version 3 will be unable to communicate
/// with nodes running earlier protocol versions.
pub(super) const PROTOCOL_VERSION: usize = 3;
```

ramfox (approved):
looks good! I just have some doc change suggestions. 

very exciting

rklaehn:
I just looked at the benchmarks and did not see any breathtaking improvements. Things seemed pretty much unchanged or even a bit worse. Is that expected?

Anyway, eliminating an entire roundtrip seems like a big win in any case, even if it does not show up in the benches.

dignifiedquire:
in US case we drop around 200ms by the end for connection, do not sure what you mean

rklaehn:
> in US case we drop around 200ms by the end for connection, do not sure what you mean

Probably should not comment without first ☕ ... I looked at the latency numbers.

iroh-net/src/relay/codec.rs
dignifiedquire: are you sure?

ramfox: Oh, nope. I'm not. My bad. I misunderstood what was happening in the `Decoder` impl.

dignifiedquire:
> BREAKING:
This breaks the relay handshake, and so requires updated relay nodes to work with.


from the description above 😅

PR #2167, opened 2024-04-10
feat: pkarr relay with DNS server

Frando:
## Description

Imports https://github.com/n0-computer/iroh-dns-server into this repo.

See n0-computer/iroh-dns-server#5 for previous review/discussion.

Now includes an integration smoke test in `iroh-dns-server/src/lib.rs`.

## Notes & open questions

I *think* I addressed most review points that came up in the initial review. Prominently still open is:

* The `redb` store is used from async context but only exposes a sync interface (redb default). I think this is fine for medium load. However for better performance we should reuse transactions, which likely means we need an actor on a separate thread, as we do in iroh-bytes and iroh-sync.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


rklaehn:
> The redb store is used from async context but only exposes a sync interface (redb default). I think this is fine for medium load. However for better performance we should reuse transactions, which likely means we need an actor on a separate thread, as we do in iroh-bytes and iroh-sync.

Let's add the batching asap, since I am currently finding out that adding it to a big code base like iroh-sync is no fun at all.

But it *does not* have to happen in this PR.

rklaehn (approved):
This just adds the iroh-dns-server to the iroh repo.

It has been relatively thoroughly reviewed in https://github.com/n0-computer/iroh-dns-server/pull/5

So I think we should just merge this and do small adjustments in subsequent PRs.

PR #2170, opened 2024-04-10
deps: remove unused dependencies

Frando:
I just stumbled upon the very nice tool [`cargo machete`](https://blog.benj.me/2022/04/27/cargo-machete/) which finds unused dependencies. So let's remove them!

Ran only `cargo check` locally - let's see if CI also says that all is fine.


rklaehn (approved):
if it works, it works!

PR #2172, opened 2024-04-11
refactor(iroh-sync): Doc store batching

rklaehn:
## Description

refactor(iroh-sync): Doc store batching

try to reuse a read and especially a write transaction for some time.

Todo:

- [x] fix doc gc. currently there is no connection between the doc store and the gc loop.
- [x] trigger commit of open txn after a delay even if there is no other activity

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh-sync/src/sync.rs
```diff
@@ -242,40 +239,28 @@ pub enum CapabilityError {
...
             // on_insert_sender: RwLock::new(None),
             content_status_cb: None,
             closed: false,
         }
     }
 
-    /// Mark the replica as closed, prohibiting any further operations.
-    ///
-    /// This method is not public. Use [store::Store::close_replica] instead.
-    pub(crate) fn close(&mut self) {
```
dignifiedquire, line 274: why is this rmeoved?

rklaehn: This was called by close_replica, but is not needed anymore.

It is setting the closed flag to true and clears the subscribers, but it was called on an instance that was immediately dropped afterward, so it really did nothing as far as I can see. Maybe it did something at some point, but not anymore after some refactoring.

iroh-sync/src/actor.rs
```diff
@@ -775,24 +823,33 @@ impl Actor {
...
-        self.get_mut(namespace).map(|state| &mut state.replica)
-    }
-
-    fn get_mut(&mut self, namespace: &NamespaceId) -> Result<&mut OpenReplica> {
-        self.0.get_mut(namespace).context("replica not open")
+    fn replica<'a, 'b>(
+        &'a mut self,
+        namespace: NamespaceId,
+        store: &'b mut Store,
+    ) -> Result<Replica<StoreInstance<'b>, &'a mut ReplicaInfo>> {
```
Frando, line 830: Maybe add a type alias for `Replica<StoreInstance<'b>, &'a mut ReplicaInfo>`, this is used a lot

iroh-sync/src/ranger.rs
```diff
@@ -238,30 +237,30 @@ pub trait Store<E: RangeEntry>: Sized {
...
-    fn is_empty(&self) -> Result<bool, Self::Error>;
+    fn is_empty(&mut self) -> Result<bool, Self::Error>;
 
     /// Calculate the fingerprint of the given range.
-    fn get_fingerprint(&self, range: &Range<E::Key>) -> Result<Fingerprint, Self::Error>;
+    fn get_fingerprint(&mut self, range: &Range<E::Key>) -> Result<Fingerprint, Self::Error>;
 
     /// Insert the given key value pair.
-    fn put(&mut self, entry: E) -> Result<(), Self::Error>;
+    fn kv_put(&mut self, entry: E) -> Result<(), Self::Error>;
```
Frando, line 255: I'd prefer `entry_put` here instead of introducing a new `kv` term.

iroh-sync/src/ranger.rs
```diff
@@ -271,16 +270,16 @@ pub trait Store<E: RangeEntry>: Sized {
...
-    fn prefixes_of(&self, key: &E::Key) -> Result<Self::ParentIterator<'_>, Self::Error>;
+    fn prefixes_of(&mut self, key: &E::Key) -> Result<Self::ParentIterator<'_>, Self::Error>;
 
     /// Get all entries in the store
-    fn all(&self) -> Result<Self::RangeIterator<'_>, Self::Error>;
+    fn all(&mut self) -> Result<Self::RangeIterator<'_>, Self::Error>;
 
     /// Remove an entry from the store.
-    fn remove(&mut self, key: &E::Key) -> Result<Option<E>, Self::Error>;
+    fn kv_remove(&mut self, key: &E::Key) -> Result<Option<E>, Self::Error>;
```
Frando, line 282: See above: I'd prefer `entry_remove`

iroh-sync/src/ranger.rs
```diff
@@ -293,53 +292,10 @@ pub trait Store<E: RangeEntry>: Sized {
         prefix: &E::Key,
         predicate: impl Fn(&E::Value) -> bool,
     ) -> Result<usize, Self::Error>;
-}
-
-#[derive(Debug)]
-pub struct Peer<E: RangeEntry, S: Store<E>> {
```
Frando, line 299: Cool that is neat to make the methods just default impls on the `Store`, I think I like this.

iroh-sync/src/ranger.rs
```diff
@@ -412,23 +369,25 @@ where
                             }
                             Err(err) => Some(Err(err)),
                         })
+                        .collect::<Result<Vec<_>, _>>()?;
+                    // add the content status in a second pass
```
Frando, line 373: This adds a new allocation over all items. I guess the borrow checker complains otherwise?
Fine to leave as-is, this code paths will have to optimized further anyway. And #2141 will also touch it.

iroh-sync/src/ranger.rs
```diff
@@ -611,25 +570,100 @@ where
...
         Ok(InsertOutcome::Inserted { removed })
     }
+}
 
-    /// List all existing key value pairs.
-    // currently unused outside of tests
-    #[cfg(test)]
-    pub fn all(&self) -> Result<impl Iterator<Item = Result<E, S::Error>> + '_, S::Error> {
-        self.store.all()
+impl<E: RangeEntry, S: Store<E>> Store<E> for &mut S {
```
Frando, line 581: It is a bit unforunate that rust does not provide a way to auto-impl this, but oh well, so be it...

Frando (commented):
Didn't get to the redb store impl yet, but so far the changes in actor and ranger looks good to me!

iroh-sync/src/actor.rs
```diff
@@ -58,12 +62,17 @@ enum Action {
         #[debug("reply")]
         reply: flume::Sender<Result<(NamespaceId, CapabilityKind)>>,
     },
+    #[display("ContentHashes")]
+    ContentHashes {
+        #[debug("reply")]
+        reply: oneshot::Sender<Result<Vec<Hash>>>,
```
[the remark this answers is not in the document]
rklaehn: The only way to make this streaming is to create a real snapshot. But yeah, given that this is only called once per gc it could be done.

iroh-sync/src/ranger.rs
rklaehn: yeah, this is to fight the borrow checker. But allocations are not that expensive compared to db ops.

iroh-sync/src/ranger.rs
rklaehn: the intention was to distinguish between the "willow put" and the (currently underlying) kv store style put.

iroh-sync/src/actor.rs
```diff
@@ -608,6 +640,9 @@ impl Actor {
                     .map(|a| a.map(|a| a.map(|a| a.id()))),
             ),
             Action::ListReplicas { reply } => iter_to_channel(reply, self.store.list_namespaces()),
+            Action::ContentHashes { reply } => {
+                send_reply_with(reply, self, |this| this.store.content_hashes()?.collect())
```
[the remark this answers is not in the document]
rklaehn: we can make an owned iterator that contains a snapshot and can be sent to the receiver.

iroh-sync/src/sync.rs
Frando: Yup, this was important before we had the actor as replicas were cloned freely then. Now the actor owns the only instance so close equals drop.

iroh-sync/src/actor.rs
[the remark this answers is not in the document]
rklaehn: made it take a snapshot. This requires closing the current write transaction, so I don't want to do this all the time. But in this case I think it is OK.

iroh-sync/src/actor.rs
```diff
@@ -522,6 +533,12 @@ impl SyncHandle {
         rx.await?
     }
 
+    pub async fn content_hashes(&self) -> Result<ContentHashesIterator> {
```
Frando, line 536: Maybe add a comment that this is an iterator onto the actual redb table, which means writes are blocked until the iterator is dropped, right?

iroh-sync/src/store/fs.rs
```diff
@@ -341,73 +390,70 @@ impl Store {
...
+        )
     }
 
     /// Get all content hashes of all replicas in the store.
-    pub fn content_hashes(&self) -> Result<ContentHashesIterator> {
-        let read_tx = self.db.begin_read()?;
-        ContentHashesIterator::new(&read_tx)
+    pub fn content_hashes(&mut self) -> Result<ContentHashesIterator> {
+        // make sure the current transaction is committed
+        self.flush()?;
```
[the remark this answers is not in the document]
rklaehn: QueryIterator has a lifetime and only needs a reference to a Table<a'>. ContentHashesIterator needs a *copy* of a table.

iroh-sync/src/store/fs.rs
```diff
@@ -341,73 +390,70 @@ impl Store {
...
+        let records = tables.records;
+        ContentHashesIterator::all(records)
     }
 
     /// Get the latest entry for each author in a namespace.
-    pub fn get_latest_for_each_author(&self, namespace: NamespaceId) -> Result<LatestIterator> {
-        let tx = self.db.begin_read()?;
-        LatestIterator::new(&tx, namespace)
+    pub fn get_latest_for_each_author(&mut self, namespace: NamespaceId) -> Result<LatestIterator> {
+        // TODO: find out why this requires snapshot to work!
```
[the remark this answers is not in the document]
rklaehn: yeah, outdated!

iroh-sync/src/actor.rs
rklaehn: I don't think this is true. According to the redb docs, "Read-only transactions may exist concurrently with writes".

It's a snapshot though, so you won't see writes of course.

image

iroh-sync/src/store/fs.rs
```diff
@@ -208,45 +266,43 @@ impl Store {
...
-        namespace_id: &NamespaceId,
-    ) -> Result<Replica<StoreInstance>, OpenError> {
-        if self.open_replicas.read().contains(namespace_id) {
-            return Err(OpenError::AlreadyOpen);
-        }
+    /// This just calls load_replica_info and then creates a new replica with the info.
+    pub fn open_replica(&mut self, namespace_id: &NamespaceId) -> Result<Replica, OpenError> {
+        let info = self.load_replica_info(namespace_id)?;
+        let instance = StoreInstance::new(*namespace_id, self);
+        Ok(Replica::new(instance, Box::new(info)))
```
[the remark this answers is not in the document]
Frando: I think you can impl Deref yourself:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=34ee29e60cf4799afbd14afd5d3a27b3

iroh-sync/src/store/fs.rs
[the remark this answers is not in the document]
Frando: Ah, this is because we are sending the `ContentHashesIterator` over the channel boundary directly, whereas the QueryIterator is iterated within the actor, and each item pushed into a flume channel. Correct?
Maybe you can add a comment somewhere that mentions these two different approaches.
Did you think about what the perfomance implications are for each approach?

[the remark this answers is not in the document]
rklaehn: > Ah, this is because we are sending the ContentHashesIterator over the channel boundary directly, whereas the QueryIterator is iterated within the actor, and each item pushed into a flume channel. Correct?

Yes.

If the QueryIterator was an owned iterator it would have to be created on a read transaction, since only ReadOnlyTables are without lifetimes. So the iterator would be the same speed, but you might have to constantly switch between read and write transactions.

For GC I think making a snapshot is OK, but if you would make one for each query you would defeat the purpose of the batching.

Frando (approved):
The approach and implementation look good to me, and all my review comments have been addressed. I can't see anything missing from the PR. So I'm approving, and we have still some time until next release to discover potential issues or performance pitfalls during dogfooding.

PR #2180, opened 2024-04-12
refactor(iroh-net): Remove incremental state from reportgen actor

flub:
## Description

The reportgen actor was carrying this "incremental" state around in
two ways: as an Option in last_report and as a boolean in incremental.
It's simpler to have less state and only carry it as an Option.

## Change checklist

- [x] Self-review.
- ~~[ ] Documentation updates if relevant.~~
- ~~[ ] Tests if relevant.~~


rklaehn (approved):
👍 this avoids having to deal with the state where incremental is false but last_report is Some. Make invalid states unrepresentable!

Issue #2142, opened 2024-04-03
Repeated warning: WARN iroh_bytes::store::fs::tables: failed to delete Sizes data-


dignifiedquire:
Seeing this come up quite often:

```
2024-04-03T18:43:54.531868Z  INFO iroh_bytes::store::fs::tables: deleting Sizes
2024-04-03T18:43:54.531927Z  WARN iroh_bytes::store::fs::tables: failed to delete Sizes data-3/blobs/data/10ffa91002b0bbb3fa0e596ee5808f6e09f8fa1e57e47291f865fac52127d5ba.sizes4: No such file or directory (os error 2)
```

PR #2199, opened 2024-04-16
fix(iroh-bytes): do not log redundant file delete error

rklaehn:
## Description

fix(iroh-bytes): do not log when a file can not be deleted...

because it is not there. this makes the delete op idempotent, and also helps in case the file was not there in the first place.

Fixes https://github.com/n0-computer/iroh/issues/2142

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-bytes/src/store/fs/tables.rs
+++ b/iroh-bytes/src/store/fs/tables.rs
@@ -149,12 +149,15 @@ impl DeleteSet {
                 BaoFilePart::Sizes => options.owned_sizes_path(hash),
             };
             if let Err(cause) = std::fs::remove_file(&path) {
-                tracing::warn!(
-                    "failed to delete {:?} {}: {}",
-                    to_delete,
-                    path.display(),
-                    cause
-                );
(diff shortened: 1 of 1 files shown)


iroh-bytes/src/store/fs/tables.rs
```diff
@@ -149,12 +149,15 @@ impl DeleteSet {
...
             };
             if let Err(cause) = std::fs::remove_file(&path) {
-                tracing::warn!(
-                    "failed to delete {:?} {}: {}",
-                    to_delete,
-                    path.display(),
-                    cause
-                );
+                // Ignore NotFound errors, if the file is already gone that's fine.
+                if cause.kind() == std::io::ErrorKind::NotFound {
```
dignifiedquire, line 153: shouldn't this be `!=`?

rklaehn: who says reviews don't catch bugs...

dignifiedquire (approved):

PR #2201, opened 2024-04-17
feat(iroh-net): extend discovery NodeInfo to allow direct addrs

rklaehn:
## Description

feat(iroh-net): extend discovery NodeInfo to allow direct addrs

This should be compatible with what is sketched in https://github.com/n0-computer/iroh/issues/2149

For now we leave the direct addresses empty, but having this would allow tools like https://github.com/n0-computer/iroh-examples/tree/main/iroh-pkarr-node-discovery to publish everything (optionally) without having to reimplement the conversion from/to signed packets.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


Frando (approved):
Looks good to me

PR #2209, opened 2024-04-18
fix(iroh-net): do not persist invalid node addresses

Frando:
## Description

Changes the logic of which addresses are persisted in the magic endpoint:
* For nodes that were used in the current session, only the paths that were successfully used will be saved. This means that invalid paths will be omitted. If all paths are invalid, and no relay URL is known, the node will not be persisted at all.
* For nodes that were not used at all, all known information will be persisted. This is in place to not forget all info if a node starts, loads the addressbook, and then shuts down again without doing anything.

## Notes & open questions

Still needs a test for the new behavior, will do tomorrow

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/src/magicsock/node_map.rs
+++ b/iroh-net/src/magicsock/node_map.rs
@@ -92,11 +92,9 @@ impl NodeMap {
         }
     }
 
-    /// Get the known node addresses stored in the map. Nodes with empty addressing information are
-    /// filtered out.
-    #[cfg(test)]
-    pub fn known_node_addresses(&self) -> Vec<NodeAddr> {
-        self.inner.lock().known_node_addresses().collect()
+    /// Get the known node addresses which should be persisted.
+    pub fn node_addresses_for_storage(&self) -> Vec<NodeAddr> {
+        self.inner.lock().node_addresses_for_storage().collect()
     }
 
     /// Add the contact information for a node.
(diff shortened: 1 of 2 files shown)


dignifiedquire:
seems sensible, cc @divagant-martian and @flub who have both stared at this code before 😅

flub:
I have largely mixed feelings about this, because it is extremely difficult to come up with something that always works reasonably well.  While it is true that storing all the junk paths is pointless, it is far from clear which paths are junk and which are not generally.  You could say to only store publicly reachable addresses, but that would not work for a local cluster etc etc.

One immediate concern I have with this implementation is that it mixes two very different philosophies.  It tries to both store just what worked, but then also stores everything if it didn't work.  Generally the derp approach is "flood things till it works" with little care for the harm that could do.  But now we're starting to react to that harm (rightfully so!).

Maybe taking a step back is better: we are storing these so that we can more reliably re-connect to a node we know existed.  Now we have iroh-dns however!  do we really still need to store this?

dignifiedquire:
> do we really still need to store this?

yes, I am wondering if we should only store addresses we have confirmed good, and drop everything else

divagant-martian:
> Now we have iroh-dns
do we somehow register nodes in a vpn-fashion dns (like tailscale's MagicDNS)? Otherwise how do you contact a node without a server name

I think the approach makes sense, since we ping all the known paths for a node it is a safe assumptions that the unused ones are (at least provisionally) invalid, while for uncontacted nodes they are all unverified.

Maybe worth noting too that this idea mostly comes from tailscale where a single node entity can have multiple addresses that are simultaneously valid. I'm not sure this applies to our case in general

flub:
in chat we discussed maybe just going with this, and consider stop storing altogether once we also have mdns discovery on top of iorh-dns.

flub (approved):

PR #2219, opened 2024-04-22
fix(ci): release builds

Arqu:
## Description

Runs the usual release flow, but now also creates a draft release and attaches all binaries to it.

Also closes: https://github.com/n0-computer/iroh/issues/2232

Sample: https://github.com/n0-computer/iroh/releases/tag/untagged-b36c22bc1e19cbd75396

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


.github/workflows/release.yml
```diff
@@ -30,6 +30,7 @@ env:
   MSRV: "1.75"
   SCCACHE_CACHE_SIZE: "50G"
   BIN_NAMES: "iroh,iroh-relay,iroh-dns-server"
+  RELEASE_VERSION: ${{ github.event.inputs.release_version }}
```
flub, line 33: Given you have a whole script a little lower to set this in several ways I assume this is a leftover of an earlier version and should be removed?

.github/workflows/release.yml
```diff
@@ -138,13 +145,18 @@ jobs:
...
           run: |
             if [ "${{ matrix.name }}" = "ubuntu-arm-latest" ]; then
               export CARGO_TARGET_AARCH64_UNKNOWN_LINUX_MUSL_LINKER=aarch64-linux-gnu-gcc
               export CC=aarch64-linux-gnu-gcc
             fi
             cargo build --profile optimized-release --all-features --target ${{ matrix.cargo_targets }}
+        
+        - name: build release
+          if: matrix.os == 'windows-latest'
+          run: cargo build --profile optimized-release --all-features --target ${{ matrix.cargo_targets }}
```
flub, line 159: I don't get why windows gets its own section here, the cargo command is exactly the same as the unix version above?

.github/workflows/release.yml
```diff
@@ -204,36 +216,40 @@ jobs:
...
+              Copy-Item -Path "target/${{ matrix.cargo_targets }}/optimized-release/$BIN_NAME.exe" -Destination "$staging/"
+              Set-Location -Path "$staging"
+              Compress-Archive -Path * -DestinationPath "../$staging.zip"
+              $ASSETS += "$staging.zip"
+              Set-Location -Path ..
+            }
+            $ASSETS = $ASSETS -join ','
+            Add-Content -Path $env:GITHUB_ENV -Value "ASSET=$ASSETS"
+
+        - uses: n0-computer/actions-upload-release-asset@main
```
flub, line 251: I'm in general a bit careful about using actions from a branch rather than a tag, which can still be moving like .e.g. `v1`.  But the repo doesn't seem to have any tags.

flub (approved):

PR #2227, opened 2024-04-23
fix(iroh-net)!: improve magicsock's shutdown story

divagant-martian:
## Description

- Waits for connections to actually be closed by calling `quinn::Endpoint::wait_idle` after close.
- Consumes the `MagicEndpoint` in an effort to express the fact that closing an endpoint is a terminal operation.
- Adds logging to closing the connections in case it's deemed necessary
- Changes `MagicSock::poll_recv` to return `Poll::Pending` on close instead of an error. More info on the logic behind this change below.
- Changes `MagicSock::poll_send` to not error if nothing is being sent.

## Breaking Changes

`MagicEndpoint::close` now consumes the endpoint.

## Notes & open questions

Shutdown `MagicSock` when quinn is done with it has proven something we can't do reliably.
- Drop based solutions (close when the `MagicSock` is dropped) are unreliable because the socket is held by multiple components on `quinn`, including spawned detached tasks. There is no really any way to trigger the dropping of these tasks/components from outside `quinn`, and would require quite en effort to change it to happen inside `quinn`.
- Close based solutions were rejected. The objective here was to stop polling the socket when the endpoint was both closed (close called) and all connections gracefully finalized. The reasoning here is that `quinn` both receives and sends frames after close to read new connection attempts and gracefully reject them. This is a fair argument on their side, despite it being clearly a limitation for a reliable freeing of resources from `quinn`.
- Taking into account the fact that the socket _will_ be polled after closed, both to send and receive, returning an error from these will always produce the logging error `quinn::endpoint: I/O error: connection closed` we have been chasing, _even_ after the `quinn::Endpoint` has been dropped. Therefore changing `poll_recv` to return `Poll::Pending` addresses this annoyance.
- Note that the part about _gracefully_ shutting down is actually done by calling `quinn::Endpoing::wait_idle` and that the (now averted) log error in `quinn` doesn't really tells us whether shutdown was or not gracefull.
- Note that the above point creates an API disparity between `poll_send` and `poll_recv`: one will error on close, the other will return `Pending`. I find this OK if `MagicSock` is simply a part of our implementation, but right now it's also part of our public API. I wonder if the `MagicSock` makes sense to users on its own or if we should remove it from the public API.
- NOTE that conversation with `quinn`'s maintainers to get a better api that allows to understand when all resources as freed has started. But this is playing the long game and we need to solve this on our end for now.

## Change checklist

- [x] Self-review.
- [ ] ~Tests if relevant.~

### DIFF
--- a/iroh-net/src/magic_endpoint.rs
+++ b/iroh-net/src/magic_endpoint.rs
@@ -4,7 +4,7 @@ use std::{net::SocketAddr, path::PathBuf, sync::Arc, time::Duration};
 
 use anyhow::{anyhow, bail, ensure, Context, Result};
 use derive_more::Debug;
-use futures::StreamExt;
+use futures::{future, StreamExt};
 use quinn_proto::VarInt;
 use tokio_util::sync::{CancellationToken, WaitForCancellationFuture};
 use tracing::{debug, trace};
@@ -560,10 +560,40 @@ impl MagicEndpoint {
     ///
     /// Returns an error if closing the magic socket failed.
     /// TODO: Document error cases.
-    pub async fn close(&self, error_code: VarInt, reason: &[u8]) -> Result<()> {
-        self.cancel_token.cancel();
-        self.endpoint.close(error_code, reason);
-        self.msock.close().await?;
+    pub async fn close(self, error_code: VarInt, reason: &[u8]) -> Result<()> {
+        let MagicEndpoint {
+            msock,
+            endpoint,
+            cancel_token,
+            ..
+        } = self;
+        cancel_token.cancel();
+        endpoint.close(error_code, reason);
+
+        {
+            // waiting for connections to close can be slow: this timeout decide whethers we should
+            // inform the user of a slow operation
+            let logging_timeout = tokio::time::sleep(Duration::from_millis(500));
+            futures::pin_mut!(logging_timeout);
+
+            let wait_idle = endpoint.wait_idle();
+            futures::pin_mut!(wait_idle);
+
+            // race the timeout and closing. Nothing else needs to be done if closing finishes
+            // first. If the timeouts is done first, give the user some logs
+            if let future::Either::Right(((), wait_idle_fut)) =
+                future::select(wait_idle, logging_timeout).await
+            {
+                tracing::info!("Closing connections");
+                wait_idle_fut.await;
+                tracing::info!("Connections closed");
+            }
+        }
+        // In case this is the last clone of `MagicEndpoint`, dropping the `quinn::Endpoint` will
+        // make it more likely that the underlying socket is not polled by quinn anymore after this
+        drop(endpoint);
+
+        msock.close().await?;
         Ok(())
     }
 
@@ -582,8 +612,8 @@ impl MagicEndpoint {
     }
 
     #[cfg(test)]
-    pub(crate) fn magic_sock(&self) -> &MagicSock {
-        &self.msock
+    pub(crate) fn magic_sock(&self) -> MagicSock {
+        self.msock.clone()
     }
     #[cfg(test)]
     pub(crate) fn endpoint(&self) -> &quinn::Endpoint {
--- a/iroh-net/src/magic_endpoint.rs
+++ b/iroh-net/src/magic_endpoint.rs
@@ -4,7 +4,7 @@ use std::{net::SocketAddr, path::PathBuf, sync::Arc, time::Duration};
 
 use anyhow::{anyhow, bail, ensure, Context, Result};
 use derive_more::Debug;
-use futures::{future, StreamExt};
+use futures::StreamExt;
 use quinn_proto::VarInt;
 use tokio_util::sync::{CancellationToken, WaitForCancellationFuture};
 use tracing::{debug, trace};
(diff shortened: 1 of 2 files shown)


iroh-net/src/magic_endpoint.rs
```diff
@@ -560,10 +560,40 @@ impl MagicEndpoint {
...
+            let wait_idle = endpoint.wait_idle();
+            futures::pin_mut!(wait_idle);
+
+            // race the timeout and closing. Nothing else needs to be done if closing finishes
+            // first. If the timeouts is done first, give the user some logs
+            if let future::Either::Right(((), wait_idle_fut)) =
+                future::select(wait_idle, logging_timeout).await
+            {
+                tracing::info!("Closing connections");
+                wait_idle_fut.await;
```
dignifiedquire, line 588: when testing, did you observe this often? or is this more a rare thing that it takes longer than 500ms?

flub:
>     * Drop based solutions (close when the `MagicSock` is dropped) are unreliable because the socket is held by multiple components on `quinn`, including spawned detached tasks. There is no really any way to trigger the dropping of these tasks/components from outside `quinn`, and would require quite en effort to change it to happen inside `quinn`.

I haven't gone through all the pain of trying to implement this, so might have this wrong.  Could you explain why the dropping of `MagicSock` by Quinn is not a suitable signal that it is done with it and can really be closed?  I still don't fully grasp this, apologies.  Is this mostly around the difficulty of achieving that drop inside of our `MagicEndpoint` while the `MagicEndpoint` itself is not yet dropped?

The most recent comment at https://github.com/quinn-rs/quinn/pull/1828#issuecomment-2073594181 also seems to think that dropping of the `AsyncUdpSocket` is a signal that all resources are released.

Also I think once Quinn drops the `AsyncUdpSocket` you can also be sure there's no more clone of the socket lingering around in some cloned connection, endpoint or whatever internal tasks.

iroh-net/src/magic_endpoint.rs
```diff
@@ -560,10 +560,40 @@ impl MagicEndpoint {
     ///
     /// Returns an error if closing the magic socket failed.
     /// TODO: Document error cases.
-    pub async fn close(&self, error_code: VarInt, reason: &[u8]) -> Result<()> {
-        self.cancel_token.cancel();
-        self.endpoint.close(error_code, reason);
-        self.msock.close().await?;
+    pub async fn close(self, error_code: VarInt, reason: &[u8]) -> Result<()> {
```
flub, line 563: +1 to consuming `self`

iroh-net/src/magic_endpoint.rs
```diff
@@ -560,10 +560,40 @@ impl MagicEndpoint {
...
+            let logging_timeout = tokio::time::sleep(Duration::from_millis(500));
+            futures::pin_mut!(logging_timeout);
+
+            let wait_idle = endpoint.wait_idle();
+            futures::pin_mut!(wait_idle);
+
+            // race the timeout and closing. Nothing else needs to be done if closing finishes
+            // first. If the timeouts is done first, give the user some logs
+            if let future::Either::Right(((), wait_idle_fut)) =
+                future::select(wait_idle, logging_timeout).await
```
flub, line 585: Given we want to get rid of future sometime maybe using `tokio::select!` is better?  It's also more often used in our code.  Feel free to disagree however.

iroh-net/src/magic_endpoint.rs
```diff
@@ -560,10 +560,40 @@ impl MagicEndpoint {
...
+            {
+                tracing::info!("Closing connections");
```
flub, line 587: why not always log these messages?  are they that invasive?

iroh-net/src/magicsock.rs
```diff
@@ -287,6 +287,11 @@ impl Inner {
         let bytes_total: usize = transmits.iter().map(|t| t.contents.len()).sum();
         inc_by!(MagicsockMetrics, send_data, bytes_total as _);
 
+        let mut n = 0;
+        if transmits.is_empty() {
```
flub, line 291: part of me is curious how often this happens and would be tempted to log this on debug or trace level.  maybe also together with the `self.is_closed()` result as a field.

iroh-net/src/magicsock.rs
```diff
@@ -484,6 +485,7 @@ impl Inner {
         Ok(sock)
     }
 
+    // NOTE: Receiving on a [`Self::closed`] socket will return [`Poll::Pending`] undefinitely.
```
flub, line 488: This should be mentioned in the doc comment of the `MagicSock` itself, or at least of it's `AsyncUdpSocket` impl.

divagant-martian:
Signal on drop has two problems mainly:
1. `MagicEndpoint: Clone` which makes for this horrible footgun:
```rust
let ep1 = MagicEndpoint::new();
let ep2 = ep1.clone();
ep2.close().await // this is a deadlock
```
Keeping track of clones is not easy so even in the presence of a big warning to users, I find such an API honestly unacceptable. It's not like a function `xyz must be called before` or `should run inside a tokio runtime` which are much more easily verifiable. It would be a `you can only call this after dropping all clones of `MagicEndpoint`
2. Clearly when the `AsyncUdpSocket` is dropped.. well it's dropped. The thing is that we can't reliably ensure this happened even if we drop the `quinn::Endpoint`, because it's the last thing to go. In fact, take a code such as this (where the `MagicSock` had no changes to `poll_*`
```rust
let msock = MagicSock::new();
let quinn_ep = quinn::Endpoint::new_with_abstract_socket(msock.clone());
drop(quinn_ep)
msock.close().await; // here the endpoint has been dropped but more likely than not, the `AsyncUdpSocket` **hasn't** been dropped yet. So more often than not you will _still_ get that error bubbled up as the quinn error log.

The main problem is for sure the first one and it makes it IMO simply not worth it going such a convoluted path when a far simpler solutions suffices

iroh-net/src/magicsock.rs
divagant-martian: just thought it made sense in preventing undesirable errors, but it's not like I've seen it before

iroh-net/src/magic_endpoint.rs
divagant-martian: I had always used `tokio::select!` before, until Rudiger for some things in another repo asked me to use the `futures` function. For such small cases I know prefer it for a couple of reasons:
1. no macro, so the code inside is correctly formatted by `cargo fmt`. Of course this is a lesser reason
2. Saves me dealing with cancellation safety. `future::select` returns the future that didnt' complete, which is exactly what I'm taking advantage of here. `tokio::select!` would simply drop the future, we would likely need to have this in a loop and carefully think how to wrap the future so that it's not just recreated on the iteration - because I doubt `wait_idle` is cancel-safe -

Of course we don't need to deal with any of this if we just log it all the time

iroh-net/src/magic_endpoint.rs
dignifiedquire: we could probably just drop the log level to debug as a compromise

iroh-net/src/magic_endpoint.rs
divagant-martian: removed in c1238e3, going with `debug` as a compromise

dignifiedquire (approved):

iroh-net/src/magicsock.rs
divagant-martian: done in c8fa0de

iroh-net/src/magicsock.rs
divagant-martian: logs added in c8fa0de

flub:
@divagant-martian thanks for the clear summary on the problems of drop!

PR #2231, opened 2024-04-25
[wip] feat: willow implementation

Frando:
## Description

Work in progress! Implementation of https://willowprotocol.org/ for iroh

*Protocol*

The `protocol` module contains the data structures and basic interactions as defined by the Willow specs. It is IO-less and not generic, all types are concrete.

* [x] [Willow data model](https://willowprotocol.org/specs/data-model/index.html#data_model) `proto::willow`
   * [x] Use willow-rs (see #2616 )
*  [Meadowcap](https://willowprotocol.org/specs/meadowcap/index.html#meadowcap) `proto::meadowcap`
     * [x] Use willow-rs (see #2616 )
* [x] [Groupings ](https://willowprotocol.org/specs/grouping-entries/index.html#grouping_entries) `proto::groupings`
   * [x] Use willow-rs (see #2616 )
* Willow Geneal Sync Protocol (WGSP) `proto::sync` 
  * [x]  Most message type structs written down
  * [x]  Private Area Intersection
  * [ ] Cryptographic fingerprint (uses only XOR atm)
  * [ ] Protocol-compliant encoding

*Session*

The `session` module contains the implementation of the WGPS protocol session.

* [x] Session setup
* [x] Private area intersection
* [x] Set reconciliation
* [ ] Resource control
  * [x] Binding resources
  * [ ] Freeing resources
  * [x] Issue initial guarantees
  * [x] Respect issued guarantees
  * [ ] Issue new guarantees
* [x] Live data mode between two peers
  * [x] Subscribe to new inserts and send to active sessions
  * [x] Only send entries for matcing shared areas of interest
* [ ] Verified streaming for payloads with bao outboards embedded
* [ ] Payload requests
* [ ] ~~Live data mode in a swarm setting~~

*Net*

The `net` module opens sessions over iroh-net QUIC connections.

* [x] Commitment exchange
* [x] Separate channels for each logical channel
* [x] Basic session lifecycle (allow to close sessions)
* [x] Properly terminate connections gracefully (fixed by #2599)
* [x] Properly deal with simultaenous connections by only using one of them deterministicaly (fixed by #2599)
* [x] Do not drop intents if the session is about to be closed when submitting a new intent  (fixed by #2599)

*Store*

The `store` module contains backing stores for entries and keys

* Store traits and generic functionality
  * [x] Entry store trait
  * [x] Key store trait
  * [x] Use iroh-blobs for payload storage
  * [ ] Store capabilities by hash and do reference counting
  * [ ] Expose hook to prevent GC of payloads and/or do reference counting for payloads
* [x] PoC memory store
* [ ] move store to own thread & sessions to local pools
* [ ] redb store
  * [ ] port @rklaehn's redb PoC
  * [ ] figure out if the traits need to change for our delayed commit strategy
  * [ ] add `commit` to the store trait
* [ ] subscriptions
  * [ ] reliable/persistent subscriptions

*Engine*

- [x] PeerManager (1 session peer peer only)
- [x] Intents
  - [ ] changing between `ReconcileOnce` and `Live` modes

*Other*

- [ ] How can builder APIs be done nicely over FFI

## Breaking Changes

Once we integrate, this will be a very much breaking changes for everything around docs: This will break not only APIs, but also the protocol and the storage. We cannot offer automatic migration even, because signatures change. How exactly we deal with this situation is tbd. Likely we will offer an out-of-tree tool to migrate data from the old iroh-docs to the new willow iroh-docs.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


iroh-willow/src/net/codec.rs
```diff
@@ -0,0 +1,55 @@
...
+
+impl Decoder for WillowCodec {
+    type Item = Message;
+    type Error = anyhow::Error;
+    fn decode(&mut self, src: &mut BytesMut) -> Result<Option<Self::Item>, Self::Error> {
+        if src.len() < 4 {
+            return Ok(None);
+        }
+        let bytes: [u8; 4] = src[..4].try_into().unwrap();
+        let frame_len = u32::from_be_bytes(bytes) as usize;
```
dignifiedquire, line 20: https://docs.rs/bytes/latest/bytes/buf/trait.Buf.html#method.get_u32 is easier

Frando: :+1: copied this from `iroh-sync` :smile:

dignifiedquire: oh no, that means I wrote this probably 🤣

Frando: this is gone after refactors (not using tokio codec anymore)

Cargo.toml
```diff
@@ -26,6 +30,14 @@ opt-level = 3
...
 incremental = false
 
+[profile.dev.package.ed25519-dalek]
+opt-level = 3
+
+[profile.dev.package.curve25519-dalek]
+opt-level = 3
+
+[profile.dev.package.iroh-blake3]
+opt-level = 3
```
dignifiedquire, line 40: you can do

```toml

[profile.dev.package]
ed25519-dalek = { opt-level = 3 }
curve25519-dalek = { opt-level = 3 }
iroh-blake3 = { opt-level = 3 }


```

iroh-willow/src/form.rs
```diff
@@ -0,0 +1,155 @@
+//! Types for forms for entries
```
dignifiedquire, line 1: I like the idea of these, but `form` seems a little confusing as a name, I don't have a better one (yet) though

iroh-willow/src/net.rs
```diff
@@ -0,0 +1,671 @@
...
+        // channel id.
+        Role::Alfie => {
+            channels
+                .map(|ch| {
+                    let conn = conn.clone();
+                    async move {
+                        let (mut send, recv) = conn.open_bi().await?;
+                        send.write_u8(ch.id()).await?;
+                        trace!(?ch, "opened bi stream");
+                        Result::<_, anyhow::Error>::Ok((ch, Some((send, recv))))
```
dignifiedquire, line 137: you can write `anyhow::Ok` as a shorthand

iroh-willow/src/net.rs
```diff
@@ -0,0 +1,671 @@
...
+        LogicalChannelSenders {
+            intersection: pai.0,
+            reconciliation: rec.0,
+            static_tokens: stt.0,
+            aoi: aoi.0,
+            capability: cap.0,
+            data: dat.0,
+        },
+        LogicalChannelReceivers {
+            intersection_recv: pai.1.into(),
```
dignifiedquire, line 197: why have `_recv` on these, but not `_send` on the `LogicalChannelSenders`?

iroh-willow/src/net.rs
Frando: Worked in most places but not all, not sure why.

iroh-willow/src/net.rs
```diff
@@ -0,0 +1,671 @@
...
+    send_stream.finish().await?;
+    trace!("close");
+    Ok(())
+}
+
+async fn exchange_commitments(
+    send_stream: &mut SendStream,
+    recv_stream: &mut RecvStream,
+) -> anyhow::Result<InitialTransmission> {
+    let our_nonce: AccessChallenge = rand::random();
```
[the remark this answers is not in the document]
Frando: Changed it to pass the nonce into the `net` functions.
In the upper layer (`PeerManager`) it's not yet pluggable.

iroh-willow/src/proto/pai.rs
```diff
@@ -0,0 +1,173 @@
...
+            FragmentKit::Complete(namespace_id, path) => {
+                let mut pairs = vec![];
+                for prefix in path.all_prefixes() {
+                    pairs.push((namespace_id, prefix));
+                }
+                FragmentSet::Complete(pairs)
+            }
+            FragmentKit::Selective(namespace_id, subspace_id, path) => {
+                let mut primary = vec![];
+                let mut secondary = vec![];
```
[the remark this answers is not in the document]
Frando: fixed

iroh-willow/tests/basic.rs
```diff
@@ -0,0 +1,341 @@
...
+#[tokio::test(flavor = "multi_thread")]
+async fn peer_manager_update_intent() -> Result<()> {
+    iroh_test::logging::setup_multithreaded();
+    let mut rng = create_rng("peer_manager_update_intent");
+
+    let [alfie, betty] = spawn_two(&mut rng).await?;
+    let (namespace, _alfie_user, betty_user) = setup_and_delegate(&alfie, &betty).await?;
+    let betty_node_id = betty.node_id();
+
+    insert(&betty, namespace, betty_user, &[b"foo"], "foo 1").await?;
```
Voronar, line 127: In that test case, if I change the path (both for inserted entry and interest area) to any path, longer than one component (e.g. `&[b"to", b"bar"]`), all session intents do work as expected, but actual entries are not synchronized for Alfie (her store is empty at the end of the test).

Frando: Thanks for playing with this branch and spotting this!

It is very likely exactly as you make it out, a bug in the area-to-range conversion.

I will have a look in the next days. We're also planning to move towards using willow-rs for the data model types, so latest then this will be fixed for sure.

Frando:
I merged #2616 into this branch! `iroh-willow` is now based on `willow-rs`.

Will copy-paste the PR description of #2616 because it contains context and notes/TODOs that shouldn't get lost:

## Description

Adopts [`willow-rs`](https://github.com/earthstar-project/willow-rs) in `iroh-willow`:

* Use the types from [`willow-data-model`](https://docs.rs/willow-data-model/latest/willow_data_model/). Our implementations are replaced by type aliases onto the willow-data-model types with the generics filled in.
* Use [`meadowcap`](https://docs.rs/willow-data-model/latest/willow_data_model/). Our implementations are replaced by type aliases onto the meadowcap types with the generics filled in.

The PR is large and not well readable because I adapted the module structure to be closer to willow-rs as well in the same go (a bit non-ideal, but I was already having to go over most imports once, so did it in one go). I will merge this as soon as all tests are green, and then refine in the main willow PR (#2231).

## Breaking Changes

Many public types from iroh-willow are now type aliases instead of structs.

## Notes & open questions


Based on #2231  and currently uses willow-rs git dependency with two PRs merged in: https://github.com/earthstar-project/willow-rs/pull/45 and https://github.com/earthstar-project/willow-rs/pull/46

### Generics

Type aliases work well in many regards, but they don't really hide all the generics well, they appear in compiler error messages and rust-analyzer doc hovers etc and make all that quite unreadable. We talked about this a bit, and will have to improve the story at least for the public client API, and potentially also for ourselves working in the internals.

* We can newtype wrap everything, then the generics are gone. But the cost is that we'd have to reimplement and forward all methods.
* Maybe there's ways to reduce the generics in iroh-willow itself, by using a zerosized struct that implements a trait with associated types instead. 

### Encodings

This PR does **not yet** include spec-compliant encodings for WGPS messages, it still uses postcard with serde as an interim solution. To make this work with the willow-rs types which don't implement serde, I added wrapper types that takes the willow-encoding and maps them into serde. The proper encodings will still have to be written, same for the integration of ufotofu encoding with our in-memory pipe.

### Key types

We currently have `UserId` and `UserPublicKey` types (and same for namespace): `Id` is a wrapper around `[u8; 32]`, wheres `PublicKey` is the `ed25519_dalek::VerifyingKey`. Converting from Id to PublicKey is fallible (because not all 32 bytes are valid ed pubkeys) and non-free (it involves numeric operations). Before this PR, we used `PublicKey` in meadowcap and `Id` in data-model. This is no longer possible because the generics have to map on the same time. We now use `Id` everywhere and convert to `PublicKey` on demand.

This is non-ideal because the conversion will happen far more often than needed. However willow currently assumes a non-fallible `successor()` function when a `UserId` (`UserPublicKey`) is used as a subspace id.

What we likely want is an in-memory LRU cache for the converted versions. Or a type where the numeric pubkey is optional, but that would either need mutable access to the key for operations, or interior mutability.

iroh-willow/tests/basic.rs
```diff
@@ -0,0 +1,435 @@
...
+#[tokio::test(flavor = "multi_thread")]
+async fn peer_manager_two_intents() -> Result<()> {
+    iroh_test::logging::setup_multithreaded();
+    let mut rng = create_rng("peer_manager_two_intents");
+
+    let [alfie, betty] = spawn_two(&mut rng).await?;
+    let (namespace, _alfie_user, betty_user) = setup_and_delegate(&alfie, &betty).await?;
+    let betty_node_id = betty.node_id();
+
+    insert(&betty, namespace, betty_user, &[b"foo", b"1"], "foo 1").await?;
```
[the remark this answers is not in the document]
Frando: Thanks! It was a bug, and is now fixed. Didn't handle the case well where no `ReconciliationSendPayload` messages were sent after a `ReconciliationSendEntry`. Also added a test so that it doesn't break again.

iroh-willow/tests/basic.rs
Voronar, line 1:
```rust
async fn delegate_path(
    path: Path,
    alfie: Peer,
    betty: Peer,
    user_alfie: UserId,
    user_betty: UserId,
    namespace_id: NamespaceId,
    session_mode: SessionMode,
) -> anyhow::Result<()> {
    let restricted_area = Area::new_path(path.clone());

    let cap_for_betty = alfie
        .delegate_caps(
            CapSelector::any(namespace_id),
            AccessMode::Read,
            DelegateTo::new(user_betty, RestrictArea::Restrict(restricted_area.clone())),
        )
        .await?;

    betty.import_caps(cap_for_betty).await?;

    let entry = EntryForm::new_bytes(namespace_id, path.clone(), "foo");

    alfie.insert_entry(entry, user_alfie).await?;

    let cap_selector = CapSelector::new(
        namespace_id,
        UserSelector::Any,
        AreaSelector::ContainsArea(restricted_area.clone()),
    );

    let interest = Interests::builder()
        .add_area(cap_selector.clone(), vec![restricted_area])
        .build();

    let init = SessionInit::new(interest, session_mode);

    let mut intent = betty.sync_with_peer(alfie.node_id(), init).await.unwrap();

    intent.complete().await.unwrap();

    intent.close().await;

    Ok(())
}

#[tokio::test(flavor = "multi_thread")]
async fn subsequent_cap_delegations() -> Result<()> {
    iroh_test::logging::setup_multithreaded();
    let mut rng = create_rng("subsequent_cap_delegations");

    let [alfie, betty] = spawn_two(&mut rng).await?;

    let user_alfie = alfie.create_user().await?;
    let user_betty = betty.create_user().await?;

    let namespace_id = alfie
        .create_namespace(NamespaceKind::Owned, user_alfie)
        .await?;

    let shared_path1 = Path::from_bytes(&[b"1"])?;
    let shared_path2 = Path::from_bytes(&[b"2"])?;

    tokio::spawn({
        let alfie = alfie.clone();
        let betty = betty.clone();

        async move {
            delegate_path(
                shared_path1,
                alfie,
                betty,
                user_alfie,
                user_betty,
                namespace_id,
                SessionMode::Continuous,
            )
            .await
            .unwrap();
        }
    });

    loop {
        let entries: Vec<_> = betty
            .get_entries(namespace_id, Range3d::new_full())
            .await?
            .try_collect()
            .await?;

        tracing::info!("len1: {:#?}", entries);

        if entries.len() == 1 {
            break;
        }

        sleep(Duration::from_secs(1)).await;
    }

    tokio::spawn({
        let alfie = alfie.clone();
        let betty = betty.clone();

        async move {
            delegate_path(
                shared_path2,
                alfie,
                betty,
                user_alfie,
                user_betty,
                namespace_id,
                SessionMode::Continuous,
            )
            .await
            .unwrap();
        }
    });

    loop {
        let entries: Vec<_> = betty
            .get_entries(namespace_id, Range3d::new_full())
            .await?
            .try_collect()
            .await?;

        tracing::info!("len2: {:#?}", entries);

        if entries.len() == 2 {
            break;
        }

        sleep(Duration::from_secs(1)).await;
    }

    Ok(())
}

#[tokio::test(flavor = "multi_thread")]
async fn mutual_cap_delegations() -> Result<()> {
    iroh_test::logging::setup_multithreaded();
    let mut rng = create_rng("mutual_cap_delegations");

    let [alfie, betty] = spawn_two(&mut rng).await?;

    let user_alfie = alfie.create_user().await?;
    let user_betty = betty.create_user().await?;

    let namespace_id = alfie
        .create_namespace(NamespaceKind::Owned, user_alfie)
        .await?;

    let alfie_path = Path::from_bytes(&[b"a"])?;
    let betty_path = Path::from_bytes(&[b"b"])?;

    tokio::spawn({
        let alfie = alfie.clone();
        let betty = betty.clone();
        let alfie_path = alfie_path.clone();

        async move {
            delegate_path(
                alfie_path,
                alfie,
                betty,
                user_alfie,
                user_betty,
                namespace_id,
                SessionMode::ReconcileOnce,
            )
            .await
            .unwrap();
        }
    });

    loop {
        let entries: Vec<_> = betty
            .get_entries(namespace_id, Range3d::new_full())
            .await?
            .try_collect()
            .await?;

        tracing::info!("alfie path delegated to betty: {:?}", entries);

        if entries.iter().any(|e| e.entry().path() == &alfie_path) {
            break;
        }

        sleep(Duration::from_secs(1)).await;
    }

    tokio::spawn({
        let alfie = alfie.clone();
        let betty = betty.clone();
        let betty_path = betty_path.clone();

        async move {
            delegate_path(
                betty_path,
                betty,
                alfie,
                user_betty,
                user_alfie,
                namespace_id,
                SessionMode::ReconcileOnce,
            )
            .await
            .unwrap();
        }
    });

    loop {
        let entries: Vec<_> = alfie
            .get_entries(namespace_id, Range3d::new_full())
            .await?
            .try_collect()
            .await?;

        tracing::info!("Betty path delegated to Alfie: {:?}", entries);

        if entries.iter().any(|e| e.entry().path() == &betty_path) {
            break;
        }

        sleep(Duration::from_secs(1)).await;
    }

    Ok(())
}
```

I've added two test cases that don't work for some reason. Maybe it's my fault (willow API incorrect usage) or some bug in the willow implementation.

The first test case is subsequent capability delegation.
And the second one is mutual capability delegation.
In both cases, entry synchronization doesn't work as I expect from it (in case of successful scenario, tests should be terminated).

iroh-willow/src/store.rs
```diff
@@ -0,0 +1,144 @@
...
+        // so significant.
+        let token = capability.authorisation_token(&entry, secret_key)?;
+        let authorised_entry = AuthorisedEntry::new_unchecked(entry, token);
+        let inserted = self
+            .entries()
+            .ingest_entry(&authorised_entry, EntryOrigin::Local)?;
+        Ok((authorised_entry, inserted))
+    }
+
+    pub fn create_namespace(
```
Voronar, line 103: Working with namespaces having only their IDs is not very convenient. It would be nice to have some kind of association of namespace ID with some developer provided (in form of opaque data container) readable name like enum variant or constant string. Otherwise, one have to create its own association map with some persistence logic, partially duplicating namespace store as result of that.

iroh-willow/src/store/auth.rs
```diff
@@ -0,0 +1,295 @@
...
+    }
+    pub fn get_write_cap(
+        &self,
+        selector: &CapSelector,
+    ) -> Result<Option<WriteCapability>, AuthError> {
+        let cap = self.caps.get_write_cap(selector)?;
+        Ok(cap)
+    }
+
+    pub fn get_read_cap(
```
Voronar, line 47: It would be nice to have a unique ID for the capability to query it exactly by the ID, because a capability selector could return more than one capability. Also, when we want to list capabilities (e.g. in some admin UI) it would be pleasant to have some readable names associated with the capabilities IDs (as I mentioned in a comment about namespace store ergonomics here https://github.com/n0-computer/iroh/pull/2231#pullrequestreview-2366756832).

matheus23:
what the fuck github

matheus23:
LOL except I can't ?????

matheus23:
Anyhow. I didn't intend it to close *right now*, but intended on closing today.

The implementation is getting extracted in the same vein as `iroh-docs`, `iroh-blobs` and `iroh-gossip`.
Work is going to continue at https://github.com/n0-computer/iroh-willow

PR #2233, opened 2024-04-25
feat(iroh-bytes): add copy fallback for Export::TryReference

### LINKED ISSUE #2229: Fallback to std::fs::copy in `export` to make cross-mountpoint export possible
Now export tries to use std::fs::rename for exporting blobs. It works only if both src and dst are on the same mountpoint. It would be nice if in the case of std::fs::rename's error, Iroh falls back to std::fs::copy.

dignifiedquire:
When encountering a failure in renaming, due to cross mount point issues, fallback to copy and delete.

Closes #2229

### DIFF
--- a/iroh-bytes/src/store/fs.rs
+++ b/iroh-bytes/src/store/fs.rs
@@ -86,7 +86,7 @@ use redb::{AccessGuard, DatabaseError, ReadableTable, StorageError};
 use serde::{Deserialize, Serialize};
 use smallvec::SmallVec;
 use tokio::io::AsyncWriteExt;
-use tracing::trace_span;
+use tracing::{trace_span, warn};
 
 mod import_flat_store;
 mod migrate_redb_v1_v2;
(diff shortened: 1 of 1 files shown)


iroh-bytes/src/store/fs.rs
```diff
@@ -1643,10 +1646,40 @@ impl ActorState {
...
+                                            Ok(_) => {
+                                                let entry = EntryState::Complete {
+                                                    data_location: DataLocation::External(
+                                                        vec![target],
+                                                        size,
+                                                    ),
+                                                    outboard_location,
+                                                };
+                                                // Delete old file
+                                                if let Err(err) = std::fs::remove_file(&path) {
```
rklaehn, line 1662: You are not supposed to actually delete stuff here. If the transaction gets rejected the store would be in an inconsistent state. The entry would still point to internal data, but that internal data would no longer be there.

The proper way to delete things is to just _mark_ them for deletion by adding this file to the delete_after_commit set.

Unfortunately this is a bit convoluted, but keeping a file system and a transactional DB in sync is a PITA.

                                                 ```
-                                                // Delete old file
-                                                if let Err(err) = std::fs::remove_file(&path) {
-                                                    warn!("failed to delete {} after exporting it: {:?}", path.display(), err);
-                                                }
-
                                                 drop(guard);
                                                 tables.blobs.insert(temp_tag.hash(), entry)?;
+                                                tables
+                                                    .delete_after_commit
+                                                    .insert(*temp_tag.hash(), [BaoFilePart::Data]);
                                                 drop(temp_tag);
                                                 ```

dignifiedquire: thanks, fixed

iroh-bytes/src/store/fs.rs
```diff
@@ -1611,65 +1611,93 @@ impl ActorState {
...
-                                        data_location: DataLocation::External(vec![target], size),
-                                        outboard_location,
-                                    };
-                                    drop(guard);
-                                    tables.blobs.insert(temp_tag.hash(), entry)?;
-                                    drop(temp_tag);
-                                    tx.send(Ok(())).ok();
-                                }
-                                Err(e) => {
+                        }
```
rklaehn, line 1633: Not sure, maybe this whole procedure is long enough to warrant a fn?

rklaehn (approved):

iroh-bytes/src/store/fs.rs
dignifiedquire: I tried, and ran into borrowchk issues and then aborted 😓

PR #2234, opened 2024-04-25
feat(iroh-net)!: Improve initial connection latency

flub:
## Description

When the magic socket changes the underlying path the congestion controller has all the wrong ideas about the latency.  This results in widely varying latencies and only slowly settling of them.  This change resets the values of the congestion controller whenever the magic socket path changes.  The result is a much faster stable connection.

Renames the Pong::src field to ping_observed_addr to try and make this field a bit more intuitive to read.  This helps especially around logging pongs and the UDP address of the pong sender next to each other.

The mesh_stacks function in the tests has the relay server removed.  The point of the function is to hook up the magic endpoints using direct addresses, so the relay is redundant.  Furthermore the way the server was set up in the tests made it not functional, so it never did anything.  So things are simplified without it.

## Breaking Changes

- `MagicEndpoint::accept` now returns `magic_endpoint::Accept` rather than Quinn's `Accept` type.
- `magic_endpoint::Connecting` replaces `quinn::Connecting`.  This is the type returned by `.await`ing the `Accept` future.
- `magic_endpoint::accept_conn` and `magic_endpoint::get_alpn` have been removed.  You now accept the connection by directly awaiting the futures returned.  To retrieve the ALPN use the new `Connecting::alpn` method.

## Notes & open questions

To review, start at the new `rtt_actor.rs`, that's the important bit.  Then look at `MagicEndpoint` how it adds it to `.connect()` (simple) and `.accept()` (complex).  The latter requires copying a bunch of Quinn structs into magic_endpoint.  This also simplifies how the interact with the alpn a little.  Everything else is how making the rest of iroh adapt to those changes.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

## TODO

- ~~[ ] Use stream-util instead of futures-concurrency~~


iroh-net/examples/listen.rs
```diff
@@ -62,9 +62,10 @@ async fn main() -> anyhow::Result<()> {
         "\tcargo run --example connect -- --node-id {me} --addrs \"{local_addrs}\" --relay-url {relay_url}\n"
     );
     // accept incoming connections, returns a normal QUIC connection
-    while let Some(conn) = endpoint.accept().await {
-        // accept the connection and extract the `node_id` and ALPN
-        let (node_id, alpn, conn) = iroh_net::magic_endpoint::accept_conn(conn).await?;
+    while let Some(mut conn) = endpoint.accept().await {
+        let alpn = conn.alpn().await?;
+        let conn = conn.await?;
+        let node_id = iroh_net::magic_endpoint::get_remote_node_id(&conn)?;
```
dignifiedquire, line 68: why did this code get worse, can't we keep the helper?

iroh-net/src/magic_endpoint.rs
```diff
@@ -649,25 +678,130 @@ impl MagicEndpoint {
...
+    inner: quinn::Connecting,
+    magic_ep: MagicEndpoint,
+}
+
+impl Connecting {
+    /// Convert into a 0-RTT or 0.5-RTT connection at the cost of weakened security.
+    pub fn into_0rtt(self) -> Result<(quinn::Connection, quinn::ZeroRttAccepted), Self> {
+        match self.inner.into_0rtt() {
+            Ok((conn, zrtt_accepted)) => {
+                // If we can't notify the rtt-actor that's not great but not critical.
```
dignifiedquire, line 720: wrong comment?

iroh-net/src/magic_endpoint.rs
```diff
@@ -649,25 +678,130 @@ impl MagicEndpoint {
...
+impl Future for Connecting {
+    type Output = Result<quinn::Connection, quinn::ConnectionError>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut std::task::Context<'_>) -> Poll<Self::Output> {
+        let this = self.project();
+        match this.inner.poll(cx) {
+            Poll::Pending => Poll::Pending,
+            Poll::Ready(Err(err)) => Poll::Ready(Err(err)),
+            Poll::Ready(Ok(conn)) => {
+                // If we can't notify the rtt-actor that's not great but not critical.
```
dignifiedquire, line 785: wrong comment?

iroh-net/src/magic_endpoint.rs
```diff
@@ -1118,9 +1255,10 @@ mod tests {
         let _ep2_guard = CallOnDrop::new(move || {
             ep2_abort_handle.abort();
         });
-        async fn accept(ep: MagicEndpoint) -> (PublicKey, String, quinn::Connection) {
+        async fn accept(ep: MagicEndpoint) -> NodeId {
```
dignifiedquire, line 1258: that's surprising, the `Connection` gets dropped immediately here?

iroh-net/src/magic_endpoint/rtt_actor.rs
```diff
@@ -0,0 +1,153 @@
...
+    ) {
+        let key = self.connection_events.insert(conn_type_changes);
+        self.connections.insert(key, (connection, node_id));
+        self.tick.notify_one();
+    }
+
+    /// Performs the congestion controller reset for a magic socket path change.
+    fn do_reset_rtt(&mut self, item: Option<(stream_group::Key, ConnectionType)>) {
+        match item {
+            Some((key, new_conn_type)) => match self.connections.get(&key) {
```
dignifiedquire, line 120: shouldn't this be checking the `new_conn_type`?

flub: I considered that, though I don't think this matters.  Every time the path changes we should reset this.  If we do direct -> mixed -> direct then we should fix the reason for this so it stay in direct (which we already did).  So I think keeping the logic here simple is preferred.

dignifiedquire: maybe add a comment in this regard?

iroh-net/src/magic_endpoint/rtt_actor.rs
```diff
@@ -0,0 +1,153 @@
...
+use tokio::sync::{mpsc, Notify};
+use tokio::task::JoinHandle;
+use tokio::time::Duration;
+use tracing::{debug, error, info_span, trace, warn, Instrument};
+
+use crate::magicsock::{ConnectionType, ConnectionTypeStream};
+
+#[derive(Debug)]
+pub(super) struct RttHandle {
+    pub(super) _handle: JoinHandle<()>,
```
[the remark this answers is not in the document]
flub: The actor is terminated by dropping the mpsc sender.  I added the JoinHandle here with the intention of allowing panic propagation.  But then didn't use it because that's very painful to do in something like the `MagicEndpoint`, i.e. an object that is supported by long-running tasks but itself does not have a task. 

Maybe there's a world where it checks on each method called whether any of it's actors have errored or paniced and propagates that error/panic.  And that's why I added a Joinhandle here.  But that change shouldn't be included in this PR.

iroh-net/src/magic_endpoint.rs
flub: Yeah, the original test did not rely on this connection at all, it just wants to assert on the NodeId.  So I simplified this to make that clear.  :shrug:

iroh-net/src/magic_endpoint.rs
flub: This comment applies to all the `let Ok(...) = ...` statements that follow.  But clearly it's confusing.  Any suggestion on how to explain this better?

iroh-net/src/magic_endpoint.rs
flub: similar, applies to the following let-ok statements

iroh-net/src/magic_endpoint.rs
dignifiedquire: maybe a helper function that these calls? 

```rust
fn try_send_rtt_msg(conn, magic_ep) {
  let Ok(peer_id) = get_remote_node_id(&conn) else {
    warn!(?conn, "failed to get remote node id");
    return;
  };
  let Ok(conn_type_changes) = this.magic_ep.conn_type_stream(&peer_id) else {
    warn!(?conn, "failed to create conn_type_stream");
    return;
  };
  let rtt_msg = RttMessage::NewConnection {
    connection: conn.weak_handle(),
   conn_type_changes,
   node_id: peer_id,
  };
  if let Err(err) = this.magic_ep.rtt_actor.msg_tx.try_send(rtt_msg) {
    warn!(?conn, "rtt-actor not reachable: {err:#}");
  }
  Ok(())
}
```

iroh-net/examples/listen.rs
flub: heh, you think this is worse?  To me the helper function was a long-standing eyesore.  This way we provide the same API as upstream Quinn does, with the `.alpn()` method extra.

If we did `.alpn()` as a helper function, you could write:
```rust
let conn = ep.accept().await?;
let node_id = get_remote_nodeid(&conn)?;
let alpn = get_alpn(&conn)?;
```

We could keep the helper function.  But then there's multiple ways to do something, which I'm not super keen on.  Though I guess if you know you need all three it's reasonable, but I've also seen code that did only need the connection and used the helper function because that's what they thought it should do.

dignifiedquire: hmm I guess the nicest version would be to 

```rust
let alpn = conn.alpn().await?;
let conn = conn.await?;
let node_id = conn.remote_node_id()?;
```

dignifiedquire: but overall, I agree, now that I look at it, this is actually better than before

dignifiedquire:
looks like `iroh-sync` still pulls in `quinn`

iroh-net/examples/listen.rs
flub: Once and if we wrap `Connection` we can even do:
```rust
let conn = ep.accept().await?;
// optionally, if you need either of these
let alpn = conn.alpn().await?;
let node_id = conn.remote_node_id()?;
```
which I think would be the nicest.  And you could still get the alpn from the intermediate `Connecting` future as well if you need it earlier.

dignifiedquire: yes, agreed

flub: ok, re-exported `Connection` now.  Are we leave the rest of this API as now in the PR?

flub:
PTAL, I believe this is ready now.

dignifiedquire (approved):

dignifiedquire:
ci sad

PR #2235, opened 2024-04-25
fix(iroh-gossip): do not enable "metrics" feature for iroh-net by default

link2xt:
Otherwise it is impossible to disable metrics support in iroh-net
if iroh-gossip is used.


Frando (approved):
LGTM!

Issue #2169, opened 2024-04-10
connections keep switching away from direct connection and switching back


flub:
On slower connections we noticed we keep switching connections all the time.  Even though we would expect them to stay direct.  E.g. transcript from a modified `iroh doctor`:

```
Accepted connection from fltgqbqd37gta4bvy22zknq62m5xp4rucuijng7ds7up7ixl4daq
connection type changed to: relay
connection type changed to: direct
connection type changed to: mixed
connection type changed to: direct
17.978148733s: rtt: 332.081712ms
connection type changed to: mixed
connection type changed to: direct
connection type changed to: mixed
29.551192275s: rtt: 337.195327ms
connection type changed to: direct
connection type changed to: relay
connection type changed to: direct
connection type changed to: relay
Test finished after 40.893338786s: connection lost
```

flub:
I suspect this is mostly because we have lossy ping/pongs:
```
... INFO ... iroh_net::magicsock::node_map::best_addr: 87: clearing best_addr reason=PongTimeout has_relay=true old_addr=213.143.125.104:57096
```

flub:
this is a bit silly in a way, because we know there's a lot of traffic flowing around at this point.  traffic which is being acked by quic in both directions.  We only downgrade the connection because this pong is treated as more important somehow, and missing it forces downgrading our connection even if data is flowing over it.

We might be able to improve this by not considering the loss of a pong so important if other data is flowing as well.

ramfox:
Agreed. It seems sort of silly to decide that a connection is dead if we are literally getting UDP packets from that addr at the same time as we timeout on a pong.

PR #2238, opened 2024-04-25
fix(iroh_net): less agressive best_addr clearing on pong timeout

divagant-martian:
## Description

Clears the best address only if within the time the pong should have been received, nothing else was seen so that the address is considered not alive in that period. This should result in less aggressive best_addr clearing.

Specific changes include:
- remove `BestAddr::clear_if_addr_older` since it is no longer needed.
- change `BestAddr::clear*` return since it was never read anyway.
- check for alive withing the `PING_TIMEOUT_DURATION` before clearing the best address on timeout
- misc docs fixes in code related to being alive/active

## Breaking Changes

n/a

## Notes & open questions

n/a

## Change checklist

- [x] Self-review.
- [ ] ~Tests if relevant.~

### DIFF
--- a/iroh-net/src/magicsock/node_map/best_addr.rs
+++ b/iroh-net/src/magicsock/node_map/best_addr.rs
@@ -29,6 +29,10 @@ impl BestAddrInner {
             .map(|trust_until| trust_until >= now)
             .unwrap_or(false)
     }
+
+    fn addr(&self) -> SocketAddr {
+        self.addr.addr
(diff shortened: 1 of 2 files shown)


dignifiedquire (approved):
lgtm

dignifiedquire:
does this fix https://github.com/n0-computer/iroh/issues/2169?

flub:
awesome, thanks!

PR #2247, opened 2024-04-26
refactor(iroh-net)!: remove the magicsock module from the public api

divagant-martian:
## Description

`magicsock` main's component, the `MagicSock` is not useful on it's own. Having it on our public API forces us to treat as breaking changes modifications that for a `MagicEndpoint` user are implementation details. To avoid this, this PR makes the following changes:
- expose all types that are part of the public functions of the `MagicSock` as `pub use`
- explicitly re-export every publicly exposed by `magicsock` from `magic_endpoint` (so that the types involved in the prev change can be used from `magic_endpoint` directly)
- changes the description of the `magic_endpoint` module to not include the `MagicSock` and make it more expressive at the same time.
- `MagicSock` had a bunch of methods that were simply methods of `Inner` re-exposed. Now that `MagicSock` is private, we can get rid of those all and simply use `Deref` to `Inner` without exposing `Inner` via trait associated types.
- That being said, renames `MagicSock` to `magicsock::Handle` which is what it really is, leaving only the new and close methods, which are the ones that handle creation and destruction of the associated tasks.
- Misc rename of fields of type `MagicSock` to `msock`, which I think aligns better to new naming as is more readable/clear.
- Simplifies the `Timer` code. The removed parts were considered used because they were publicly accessible. Becoming private showed parts that were dead code.
- `local_addr` no longer returns an error (it seems to be it never did), but it became more clear in the process of accessing the `MagicSock`(`Inner`) methods via `Deref`.
- `MagicEndpoint::local_addr` no longer returns an error

## Breaking Changes

- module `magicsock` is no longer public
- `iroh_net::magicsock::ConnectionType` -> `iroh_net::magic_endpoint::ConnectionType`
- `iroh_net::magicsock::ControlMsg` -> `iroh_net::magic_endpoint::ControlMsg`
- `iroh_net::magicsock::ConnectionInfo` -> `iroh_net::magic_endpoint::ConnectionInfo`
- `iroh_net::magicsock::ConnectionTypeStream` -> `iroh_net::magic_endpoint::ConnectionTypeStream`
- `iroh_net::magicsock::DirectAddrInfo` -> `iroh_net::magic_endpoint::DirectAddrInfo`
- `iroh_net::magicsock::LocalEndpointsStream` -> `iroh_net::magic_endpoint::LocalEndpointsStream`
- `iroh_net::magicsock::MagicSock` no longer accessible. Use `MagicEndpoint` instead.
- `iroh_net::magicsock::Options` no longer public. Used internally only.
- `iroh_net::magicsock::PacketSplitIter` no longer public. This is an implementation detail of internal use only.
- `iroh_net::magicsock::Timer` no longer public. This is an implementation detail of internal use only.
- `iroh_net::magicsock::UdpSocket` no longer public. This is an implementation detail of internal use only.

## Notes & open questions

ended up using `cargo public-api -p iroh-net diff` to get what was _added_ and make some of those not-for-public enums and structs private again. While really useful, checking this is very annoying. Hopefully we don't have to do it often.

## Change checklist

- [x] Self-review.
- [ ] ~Tests if relevant.~

### DIFF
--- a/iroh-net/src/magic_endpoint.rs
+++ b/iroh-net/src/magic_endpoint.rs
@@ -15,7 +15,7 @@ use crate::{
     discovery::{Discovery, DiscoveryTask},
     dns::{default_resolver, DnsResolver},
     key::{PublicKey, SecretKey},
-    magicsock::{self, ConnectionTypeStream, MagicSockHandle},
+    magicsock::{self, ConnectionTypeStream, Handle},
     relay::{RelayMap, RelayMode, RelayUrl},
     tls, NodeId,
 };
@@ -230,7 +230,7 @@ pub fn make_server_config(
 #[derive(Clone, Debug)]
 pub struct MagicEndpoint {
     secret_key: Arc<SecretKey>,
-    msock: MagicSockHandle,
+    msock: Handle,
     endpoint: quinn::Endpoint,
     keylog: bool,
     cancel_token: CancellationToken,
@@ -252,7 +252,7 @@ impl MagicEndpoint {
         keylog: bool,
     ) -> Result<Self> {
         let secret_key = msock_opts.secret_key.clone();
-        let msock = magicsock::MagicSockHandle::new(msock_opts).await?;
+        let msock = magicsock::MagicSock::new(msock_opts).await?;
         trace!("created magicsock");
 
         let mut endpoint_config = quinn::EndpointConfig::default();
@@ -595,7 +595,7 @@ impl MagicEndpoint {
     }
 
     #[cfg(test)]
-    pub(crate) fn magic_sock(&self) -> MagicSockHandle {
+    pub(crate) fn magic_sock(&self) -> Handle {
         self.msock.clone()
     }
     #[cfg(test)]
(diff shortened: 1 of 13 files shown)


rklaehn:
I don't have time for a thorough review, but I very much agree with this, if only to reduce the confusion when talking about MagicEndpoint and MagicSocket.

> ended up using cargo public-api -p iroh-net diff to get what was added and make some of those not-for-public enums and structs private again. While really useful, checking this is very annoying. Hopefully we don't have to do it often.

Didn't know this. Nice!

iroh-net/src/magic_endpoint.rs
```diff
@@ -15,12 +17,12 @@ use crate::{
...
     dns::{default_resolver, DnsResolver},
     key::{PublicKey, SecretKey},
-    magicsock::{self, ConnectionTypeStream, MagicSock},
+    magicsock::{self, Handle},
     relay::{RelayMap, RelayMode, RelayUrl},
     tls, NodeId,
 };
 
-pub use super::magicsock::{ConnectionInfo, LocalEndpointsStream};
+pub use super::magicsock::*;
```
dignifiedquire, line 25: I know it is a pain, but please avoid the glob reexport, so it is easy to see what is being reexported

divagant-martian: removed in cb151e2 and explicitly added in af56639

PR #2250, opened 2024-04-29
test: add test for iroh-dns-server with mainline

Frando:
## Description

This adds a test for iroh-dns-server with mainline fallback resolution.
Needed to rework the config structure a little to not duplicate code too much.

## Breaking Changes

are documented in #2180 

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-dns-server/src/config.rs
+++ b/iroh-dns-server/src/config.rs
@@ -71,8 +71,20 @@ pub struct MainlineConfig {
     pub enabled: bool,
     /// Set custom bootstrap nodes.
     ///
+    /// Addresses can either be `domain:port` or `ipv4:port`.
+    ///
     /// If empty this will use the default bittorrent mainline bootstrap nodes as defined by pkarr.
-    pub bootstrap: Vec<SocketAddr>,
+    pub bootstrap: Option<Vec<String>>,
+}
+
+/// Configure the bootstrap servers for mainline DHT resolution.
+#[derive(Debug, Serialize, Deserialize, Default)]
+pub enum BootstrapOption {
+    /// Use the default bootstrap servers.
+    #[default]
+    Default,
+    /// Use custom bootstrap servers.
+    Custom(Vec<String>),
 }
 
 #[allow(clippy::derivable_impls)]
@@ -80,7 +92,7 @@ impl Default for MainlineConfig {
     fn default() -> Self {
         Self {
             enabled: false,
-            bootstrap: vec![],
+            bootstrap: None,
         }
     }
 }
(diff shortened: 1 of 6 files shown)


iroh-dns-server/src/config.rs
```diff
@@ -69,12 +69,19 @@ impl MetricsConfig {
 pub struct MainlineConfig {
     /// Set to true to enable the mainline lookup.
     pub enabled: bool,
+    /// Set custom bootstrap nodes.
+    ///
+    /// If empty this will use the default bittorrent mainline bootstrap nodes as defined by pkarr.
+    pub bootstrap: Vec<SocketAddr>,
```
rklaehn, line 75: OK, glad we got something else to put into the mainline config section...

iroh-dns-server/src/config.rs
```diff
@@ -69,12 +69,19 @@ impl MetricsConfig {
...
 }
 
 #[allow(clippy::derivable_impls)]
 impl Default for MainlineConfig {
     fn default() -> Self {
-        Self { enabled: false }
+        Self {
+            enabled: false,
+            bootstrap: vec![],
```
rklaehn, line 83: Empty means default? How do you get the default mainline bootstrap nodes?

iroh-dns-server/src/store.rs
```diff
@@ -55,12 +58,26 @@ impl ZoneStore {
...
+    ///
+    /// Optionally set custom bootstrap nodes. If `bootstrap` is empty it will use the default
+    /// mainline bootstrap nodes.
+    pub fn with_mainline_fallback(self, bootstrap: &[SocketAddr]) -> Self {
+        let pkarr_client = if bootstrap.is_empty() {
+            PkarrClient::default()
+        } else {
+            let bootstrap = bootstrap
+                .iter()
+                .map(|addr| addr.to_string())
```
rklaehn, line 73: Can these be URLs, or why does pkarr require strings here? Hopefully that can be changed in pkarr 2.0

iroh-dns-server/src/config.rs
rklaehn: I know it is annoying, but can we make this an enum Default/Custom instead of using empty as "use the defaults"?

rklaehn (approved):

iroh-dns-server/src/store.rs
Frando: Good point. I misread the pkarr source - these can be either "ip:port" or "domain:port" strings. changed to use strings on our end as well.

PR #2254, opened 2024-04-30
feat(iroh-base): allow the addr info of tickets to be empty

rklaehn:
## Description

feat(iroh-base): allow the addr info of tickets to be empty

Without this, a ticket that does not contain any addresses or relay URL can not be parsed. But we need to allow this if we want to use short tickets.

## Notes & open questions

Note: I did not remove the error type Validation. That way we don't have breaking changes, and we might need it again in the future.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-base/src/ticket/blob.rs
+++ b/iroh-base/src/ticket/blob.rs
@@ -45,9 +45,6 @@ impl Ticket for BlobTicket {
     fn from_bytes(bytes: &[u8]) -> std::result::Result<Self, ticket::Error> {
         let res: TicketWireFormat = postcard::from_bytes(bytes).map_err(ticket::Error::Postcard)?;
         let TicketWireFormat::Variant0(res) = res;
-        if res.node.info.is_empty() {
-            return Err(ticket::Error::Verify("addressing info cannot be empty"));
-        }
         Ok(res)
     }
 }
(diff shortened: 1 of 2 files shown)


iroh/src/ticket/doc.rs
```diff
@@ -36,9 +36,6 @@ impl ticket::Ticket for DocTicket {
     fn from_bytes(bytes: &[u8]) -> Result<Self, ticket::Error> {
         let res: TicketWireFormat = postcard::from_bytes(bytes).map_err(ticket::Error::Postcard)?;
         let TicketWireFormat::Variant0(res) = res;
-        if res.nodes.is_empty() {
```
dignifiedquire, line 39: this should be kept, we do want a node to be still there

dignifiedquire:
Side note: I was thinking that we should allow parsing a node id as a `NodeTicket` if it is exactly 32 bytes long, this would allow to use Node Tickets and Node IDs interchangeably in more places

rklaehn:
> Side note: I was thinking that we should allow parsing a node id as a `NodeTicket` if it is exactly 32 bytes long, this would allow to use Node Tickets and Node IDs interchangeably in more places

Not sure about that. We do guarantee ticket stability by having the variant byte. We do not guarantee node id stability. We might change encoding to zbase32 in the future or use a different hash.

For a tool I would just write an `NodeIdOrTicket enum`. Maybe we should have such a helper.

PR #2255, opened 2024-04-30
fix(iroh-net): reconfirm best addr when receiving data on it

### LINKED ISSUE #2169: connections keep switching away from direct connection and switching back

dignifiedquire:
When receiving data via UDP, we were not extending the validity of the current `best_addr`. This would trigger expiry of the `best_addr`, even though it was actively being used. 

This fix, extends the validity, everytime we successfully receive a UDP packet on the current `best_addr`.

Closes #2169

### DIFF
--- a/iroh-net/src/magicsock/node_map/best_addr.rs
+++ b/iroh-net/src/magicsock/node_map/best_addr.rs
@@ -39,6 +39,7 @@ impl BestAddrInner {
 pub(super) enum Source {
     ReceivedPong,
     BestCandidate,
+    Udp,
 }
 
 impl Source {
@@ -47,6 +48,7 @@ impl Source {
             Source::ReceivedPong => from + TRUST_UDP_ADDR_DURATION,
             // TODO: Fix time
             Source::BestCandidate => from + Duration::from_secs(60 * 60),
+            Source::Udp => from + TRUST_UDP_ADDR_DURATION,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


iroh-net/src/magicsock/node_map/best_addr.rs
```diff
@@ -141,6 +143,22 @@ impl BestAddr {
         }
     }
 
+    /// Reset the expiry, if the passed in addr matches the currently used one.
+    pub fn reconfirm_if_used(&mut self, addr: SocketAddr, source: Source, confirmed_at: Instant) {
+        match self.0.as_mut() {
+            None => {
+                // Nothing to do
+                // TODO: do we want to store this regardless?
```
flub, line 151: no, something else is supposed to set best-addr in that case.  at least, i think so

flub (approved):

flub:
nit, the description should probably describe what this fixes and why it fixes it instead of just relying on the issue link

dignifiedquire:
> nit, the description should probably describe what this fixes and why it fixes it instead of just relying on the issue link

added more details to the description

iroh-net/src/magicsock/node_map/best_addr.rs
dignifiedquire: dropped

PR #2262, opened 2024-05-02
feat(iroh-net): Expose DNS resolver

Frando:
## Description

* allow to set a custom DNS resolver in the node builder
* slight refactor of pkarr/dns test utils
* add test using test servers for node discovery to `iroh`

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/src/discovery.rs
+++ b/iroh-net/src/discovery.rs
@@ -558,15 +558,13 @@ mod tests {
 /// publish to. The DNS and pkarr servers share their state.
 #[cfg(test)]
 mod test_dns_pkarr {
-    use std::net::SocketAddr;
     use std::time::Duration;
 
     use anyhow::Result;
     use iroh_base::key::SecretKey;
-    use url::Url;
 
     use crate::{
-        discovery::{dns::DnsDiscovery, pkarr_publish::PkarrPublisher, ConcurrentDiscovery},
+        discovery::pkarr_publish::PkarrPublisher,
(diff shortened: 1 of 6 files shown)


iroh-net/src/discovery.rs
```diff
@@ -692,24 +684,17 @@ mod test_dns_pkarr {
     }
 
     async fn ep_with_discovery(
-        relay_map: RelayMap,
-        nameserver: SocketAddr,
-        node_origin: &str,
-        pkarr_relay: &Url,
+        relay_map: &RelayMap,
```
flub, line 687: why change this to a reference if it then immediately clones it?  by keeping it as owned it's more obvious to the caller what happens to it and they get to do the clone, if needed, explicitly.

flub: oh, i guess this is just a test util.  but still, i find that an odd pattern even if it is for tests.  if you really like the convenience of calling it with borrowed data maybe using a generic with `Clone` or `ToOwned` is better.

flub (approved):

rklaehn (approved):
very nice!

iroh-net/src/discovery.rs
Frando: I agree in general but thought for a test only util it's ok and makes writing tests a bit simpler IMO. But no strong opinion, can also change back to take by value.

flub: yeah, i approved anyway as i noticed it's a test.  up to you

PR #2263, opened 2024-05-02
refactor(iroh)!: cleanup public API

dignifiedquire:
## Description

Cleanup and improve public api, see individual commits for details.

## Breaking Changes

Many, but hopefully less in the future.

- `iroh`:
  - renamed:
    - `ProviderService` -> `RpcService`
    - `iroh::client`
      - `mem::Iroh` -> `MemIroh`
      - `mem::Doc` -> `MemDoc`
      - `quic::Iroh` -> `QuicIroh`
      - `quic::Doc` -> `QuicDoc`
      - `blobs::BlobReader` -> `blobs::Reader`
      - `blobs::BlobAddProgress` -> `blobs::AddProgress`
      - `blobs::BlobAddOutcome` -> `blobs::AddOutcome`
      - `blobs::BlobDownloadProgress` -> `blobs::DownloadProgress`
      - `blobs::BlobDownloadOutcome` -> `blobs::DownloadOutcome`
      - `blobs::BlobExportProgress` -> `blobs::ExportProgress`
      - `docs::DocImportFileProgress` -> `docs::ImportFileProgress`
      - `docs::DocExportFileProgress` -> `docs::ExportFileProgress`
      - `docs::DocImportFileOutcome` -> `docs::ImportFileOutcome`
      - `docs::DocExportFileOutcome` -> `docs::ExportFileOutcome`
    - `rpc_protocol::NodeStatusResponse` -> `client::node::NodeStatus`
    - `rpc_protocol::ListTagsResponse` -> `client::tags::TagInfo`
    - `rpc_protocol::BlobListResponse` -> `client::blobs::BlobInfo`
    - `rpc_protocol::BlobListIncompleteResponse` -> `client::blobs::IncompleteBlobInfo`
    - `rpc_protocol::BlobListCollectionResponse` -> `client::blobs::CollectionInfo`
    - `rpc_protocol::DownloadMode` -> `client::blobs::DownloadMode`
  - moved:
    - `DocTicket` into `iroh-sync`
  - removed:
    - `ticket` module
    - `dial` module
  - made private:
    - `sync_engine`
    - `client::rpc_protocol`
    - `client::quic::RPC_ALPN`
    - `client::quic::connect_raw`
  - added
    - `client::node::Client::id`
    - `client::blobs::Client::download_with_opts`
    - `client::blobs::Client::download_hash_seq`


divagant-martian:
Really like the `iroh::clients::doc`, `iroh::client::blobs`, `iroh::client::node` approach. The only one change I don't love is the now `iroh::client::{MemDoc, QuicDoc, QuicIroh, etc}`. IMO `iroh::client::{mem, quic}` was pretty clean. But it's not a hill I'll die on.

Also, to confirm, removing access to the rpc is based on the assumption that only we use it, right? If that's the case (and even if not) here is a quick search of non fork repos using `rpc_protocol`. It looks to me like everything people are using is either an import with a new (more reasonable path) or achievable in a less convoluted way, so it LGTM

dignifiedquire:
> IMO `iroh::client::{mem, quic}` was pretty clean. But it's not a hill I'll die on.

I agree, the reason I changed it, was that I found it confusing to have eg `client::mem` and `client::docs`, as this looke like `mem` was just another part of the API

dignifiedquire:
> Maybe also rename `ProviderRequest` and `ProviderResponse`?

keeping with the rust convention of avoiding double names, I changed them to `rpc_protocol::{Request, Response}`

.github/workflows/ci.yml
```diff
@@ -123,8 +123,25 @@ jobs:
...
+  check_docs:
+    timeout-minutes: 30
+    name: Checking docs
+    runs-on: ubuntu-latest
+    env:
+      RUSTC_WRAPPER: "sccache"
+      SCCACHE_GHA_ENABLED: "on"
+    steps:
+    - uses: actions/checkout@v4
+    - uses: dtolnay/rust-toolchain@stable
```
flub, line 135: If you use `with: toolchain` then this should be `@master`

dignifiedquire: but then this is not against a fixed tag and would change everytime there is a change on the action, no?

flub: true, it's the way the action is designed though.  i think it makes sense to use as designed, until we find out it hurts us.  @stable is also a moving tag which moves every 6 weeks.

divagant-martian (approved):
lgtm

PR #2264, opened 2024-05-04
feat: Update from `default-net` to rebranded `netdev`

matheus23:
## Description

- Upgrades from `default-net` v0.20 to `netdev` v0.25, which is simply a rebrand of the original default-net
- Allows depending on `system-configuration` version 0.6 instead of 0.5.1 downstream. This fixes iOS compliation for us.

## Breaking Changes

~~Would need to check if/how much the `default-net` types were exposed in the API.~~
@divagant-martian says this is what changes in the API:
> ```diff
> Changed items in the public API
> ===============================
> -pub iroh_net::net::interfaces::IpNet::V4(netdev::ip::Ipv4Net)
> +pub iroh_net::net::interfaces::IpNet::V4(default_net::ip::Ipv4Net)
> -pub iroh_net::net::interfaces::IpNet::V6(netdev::ip::Ipv6Net)
> +pub iroh_net::net::interfaces::IpNet::V6(default_net::ip::Ipv6Net)
> ```

## Notes & open questions

~~Not sure yet if it fixes our iOS problems.~~ It's required to fix our iOS build.
This should also make this crates.io patch obsolete: https://github.com/n0-computer/iroh-ffi/blob/524e5c0bab219291ddca6fa03f4038302eb4f27f/Cargo.toml#L62

## Change checklist

- [x] Self-review.
- [x] Documentation updates if relevant. (no mention of `default_net` in docs)
- [ ] Tests if relevant.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1042,23 +1042,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7e962a19be5cfc3f3bf6dd8f61eb50107f356ad6270fbb3ed41476571db78be5"
 
 [[package]]
-name = "default-net"
-version = "0.20.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7ba429d84a27fa854c66fd2e29eb1cdf6d38bbfd4495bd9f522f12a7f21e05bf"
-dependencies = [
- "dlopen2",
- "libc",
- "memalloc",
- "netlink-packet-core",
- "netlink-packet-route",
- "netlink-sys",
- "once_cell",
- "system-configuration",
- "windows 0.48.0",
-]
-
-[[package]]
 name = "der"
 version = "0.7.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


divagant-martian:
```diff
Changed items in the public API
===============================
-pub iroh_net::net::interfaces::IpNet::V4(netdev::ip::Ipv4Net)
+pub iroh_net::net::interfaces::IpNet::V4(default_net::ip::Ipv4Net)
-pub iroh_net::net::interfaces::IpNet::V6(netdev::ip::Ipv6Net)
+pub iroh_net::net::interfaces::IpNet::V6(default_net::ip::Ipv6Net)
```
interfaces should probably not even be exposed as part of the public api, but in any case, here is what changed

matheus23:
This fixes our iOS build together with the instructions from here: https://iroh.computer/docs/examples/ios-starter#add-the-system-configuration-framework

dignifiedquire:
> It seems like it's been resolved and was released in version 0.16.
> `iroh-net` already depends on version 0.20 at the moment, so should the macOS logic be adjusted, or is the original issue not _actually_ fixed?

yeah, I think this just never got rechecked after that fix was done

matheus23:
I'd be happy to contribute a change that removes the logic that's possibly subsumed by the upstream fix, but don't have a macOS device to test on :confused:

PR #2272, opened 2024-05-07
[WIP]: experiment with collections

dignifiedquire:


iroh/src/util/collection.rs
```diff
@@ -202,6 +232,35 @@ impl Collection {
...
+
+        let links_bytes = std::iter::once(res_meta.hash)
+            .chain(links)
+            .collect::<HashSeq>();
+
+        // delete meta_tag
+        let tags = crate::client::tags::Client {
+            rpc: blobs.rpc.clone(),
+        };
+        tags.delete(res_meta.tag).await?;
```
rklaehn, line 257: Shouldn't this happen after the hashseq is added? If not, what's protecting the meta_bytes if a GC would happen between here and the add_stream_hash_seq call?

rklaehn (changes requested):
I like the idea of removing the concept of collections from ~~iroh-bytes~~ iroh-blobs.

Other than moving things around this does not seem to change anything fundamentally.

iroh/src/util/collection.rs
dignifiedquire: fixed

rklaehn:
I like the idea to move collection out of iroh-bytes. Also the idea of moving the export logic to the client side of the rpc boundary.

But I don't like moving it into iroh.

You should be able to work with collections without having to import all of iroh, including gossip, docs, ... Collections are fundamentally an iroh-blobs concept.

- So I would say: keep the simplification of the export API
- Keep collections in iroh-blobs under a feature flag, or move them into a separate crate, but not in iroh

WDYT?

dignifiedquire:
how would you implement them then over RPC? if in `iroh-blobs` implemented, the only trait it can know about is the `Store`  parts, but that is not enough to move them to client only, which the follow on step for me would be

PR #2275, opened 2024-05-08
feat(iroh)!: unify node api

dignifiedquire:
Unifies the way to interact with the node to the rpc layer

## Breaking Changes

- moved
  - `iroh::client::Node::stats`  -> `iroh::client::Client::stats`
  - `iroh::client::Node::connections`  -> `iroh::client::Client::connections`
  - `iroh::client::Node::connection_info`  -> `iroh::client::Client::connection_info`
  - `iroh::client::Node::status`  -> `iroh::client::Client::status`
  - `iroh::client::Node::id`  -> `iroh::client::Client::node_id`
  - `iroh::client::Node::shutdown`  -> `iroh::client::Client::shutdown`
- added 
  - `iroh::client::Client::my_relay`
  - `iroh::client::Client::my_addr`
- removed
  - `iroh::node::Node::ticket`, use `iroh::client::blobs::Client::share` instead

### DIFF
--- a/iroh/src/client/node.rs
+++ b/iroh/src/client/node.rs
@@ -5,24 +5,19 @@ use std::{collections::BTreeMap, net::SocketAddr};
 use anyhow::Result;
 use futures_lite::{Stream, StreamExt};
 use iroh_base::key::PublicKey;
-use iroh_net::{magic_endpoint::ConnectionInfo, NodeAddr, NodeId};
-use quic_rpc::{RpcClient, ServiceConnection};
+use iroh_net::{magic_endpoint::ConnectionInfo, relay::RelayUrl, NodeAddr, NodeId};
+use quic_rpc::ServiceConnection;
 use serde::{Deserialize, Serialize};
 
 use crate::rpc_protocol::{
-    CounterStats, NodeConnectionInfoRequest, NodeConnectionInfoResponse, NodeConnectionsRequest,
-    NodeIdRequest, NodeShutdownRequest, NodeStatsRequest, NodeStatusRequest, RpcService,
+    CounterStats, NodeAddrRequest, NodeConnectionInfoRequest, NodeConnectionInfoResponse,
+    NodeConnectionsRequest, NodeIdRequest, NodeRelayRequest, NodeShutdownRequest, NodeStatsRequest,
+    NodeStatusRequest, RpcService,
 };
 
-use super::flatten;
+use super::{flatten, Iroh};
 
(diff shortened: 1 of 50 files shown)


iroh/src/client/node.rs
```diff
@@ -54,11 +49,23 @@ where
     }
 
     /// Get the id of this node.
-    pub async fn id(&self) -> Result<NodeId> {
+    pub async fn node_id(&self) -> Result<NodeId> {
         let id = self.rpc.rpc(NodeIdRequest).await??;
```
divagant-martian, line 53: why go over rpc for this? was the previous version (`self.inner.secret_key.public()`) somehow wrong?

divagant-martian (commented):
After a lot of confusion about why the new `node_id` and `my_relay` calls were async and returning an error, I think these can be kept for the `Node` since it relies on a `MemClient` so it should be fine to query the shared state in the underlying `MemClient`. The way I see it, in a generic `Iroh<C>` is fine (and necessary) to deal with these calls, and it's also fine to accept some benefits from using a `Node` directly like access to this info without going over rpc/incurring in another await point + potential error

iroh/src/node.rs
```diff
@@ -211,25 +198,6 @@ impl<D: BaoStore> Node<D> {
...
-
-    /// Return the [`NodeAddr`] for this node.
-    pub async fn my_addr(&self) -> Result<NodeAddr> {
-        self.inner.endpoint.my_addr().await
-    }
-
-    /// Get the relay server we are connected to.
-    pub fn my_relay(&self) -> Option<RelayUrl> {
-        self.inner.endpoint.my_relay()
-    }
```
divagant-martian, line 231: I think it would be better to keep this

dignifiedquire:
@divagant-martian I agree with you, but with the `deref` I am not sure how to make this work, as I will have conflicts. And giving the functions a different signature for the memory case would only work with something like specialization.

divagant-martian:
so my understanding is that the overloaded method takes precedence and method resolution should still use be able to identify which is which since the name is the same but the signature is not (Result + async) I gave it a go here https://github.com/n0-computer/iroh/pull/2280. tbh I still find it confusing to deal with so many layers of method delegation with the same name, so I _hope_ I updated the right one

dignifiedquire:
thanks @divagant-martian looks like your changes worked :)

divagant-martian (approved):
Lgtm

PR #2279, opened 2024-05-10
feat(iroh-net): Own the public QUIC API

flub:
## Description

This export a lot of quinn types directly from
iroh-net::magic_endpoint.  These are all the types we need to interact
with iroh-net in our own code.  The goal is that users should not need
to figure out how to add their own (iroh-)quinn dependency to use
iroh-net, instead all types should be provided by iroh-net.

Not all APIs are re-exported however, to avoid exporting too much.
Hopefully what iroh itself needs is a reasonable indication of what is
needed, we can always add more.

## Notes & open questions

- This ducks re-writing the iroh-blobs examples.
- This leaves quinn as a dependency of iroh itself, this is still needed because quic-rpc likewise does not provide all the needed types in it's public API.  We should probably do a similar effort there.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh-cli/Cargo.toml
```diff
@@ -40,12 +40,12 @@ human-time = "0.1.6"
 indicatif = { version = "0.17", features = ["tokio"] }
 iroh = { version = "0.15.0", path = "../iroh", features = ["metrics"] }
 iroh-metrics = { version = "0.15.0", path = "../iroh-metrics" }
+iroh-net = { version = "0.15.0", path = "../iroh-net" }
```
dignifiedquire, line 43: no need for this, you can just do `iroh::net`

iroh-gossip/src/net.rs
```diff
@@ -229,10 +229,15 @@ impl Gossip {
...
-    /// Handle an incoming [`quinn::Connection`].
+    /// Handle an incoming [`Connection`].
     ///
     /// Make sure to check the ALPN protocol yourself before passing the connection.
-    pub async fn handle_connection(&self, conn: quinn::Connection) -> anyhow::Result<()> {
+    ///
+    /// [`Connection`]: iroh_net::magic_endpoint::Connection
+    pub async fn handle_connection(
+        &self,
+        conn: iroh_net::magic_endpoint::Connection,
```
divagant-martian, line 239: we already do a bunch of `iroh_net` imports in these files. Why not do the same and keep the docs and params simply referring the import? The new state is a tad too verbose for my liking. What do you think?

iroh-cli/Cargo.toml
flub: good point, done in 0c303f1

divagant-martian (commented):
really important change to get in soon. Any reason to handle the typing and docs in the qualified way vs doing imports?

iroh-gossip/src/net.rs
flub: I went with a direct replacement as I was not editing the local code: if `quinn::SomeThing` was used I replaced it without imports as well.  It is true the new full name is a lot longer.  It will probably also change to `iroh_net::endpoint` sometime soon as well which is at least a bit shorter.

I will attempt some cleanup.

flub: 290e447 is more aggressive with imports, hope this is better

divagant-martian (approved):
LGTM.

Related, do we not export these types in the public api's of the individual crates? should we? not sure on either

dignifiedquire (approved):

PR #2284, opened 2024-05-13
feat(iroh): allow custom protocols with optional manual accept

Frando:
## Description

allow to manually accept connections and use custom protocols together with iroh core

see the included example which uses both a custom protocol and core iroh protocols.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/node/builder.rs
+++ b/iroh/src/node/builder.rs
@@ -6,7 +6,7 @@ use std::{
     time::Duration,
 };
 
-use anyhow::{bail, Context, Result};
+use anyhow::{Context, Result};
 use futures_lite::StreamExt;
 use iroh_base::key::SecretKey;
 use iroh_blobs::{
@@ -29,7 +29,7 @@ use quic_rpc::{
 use serde::{Deserialize, Serialize};
 use tokio::sync::mpsc;
 use tokio_util::{sync::CancellationToken, task::LocalPoolHandle};
-use tracing::{debug, error, error_span, info, trace, warn, Instrument};
+use tracing::{debug, error_span, info, trace, warn, Instrument};
 
 use crate::{
     client::RPC_ALPN,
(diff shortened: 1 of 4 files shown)


iroh/src/node/builder.rs
```diff
@@ -328,6 +350,19 @@ where
...
 
+    /// Set the mode in which the node accepts incoming connections.
+    ///
+    /// If set to [`AcceptMode::Manual`], incoming connections have to be accepted by the
+    /// application code in a loop. Connections that should be handled by iroh can be passed to
+    /// [`Node::handle_connection`]. This allows to handle custom protocols.
+    ///
+    /// The default is [`AcceptMode::Auto`], which accepts all iroh protocol automatically and does
+    /// not allow to use custom protocols.
+    pub fn accept_mode(mut self, accept_mode: AcceptMode) -> Self {
```
rklaehn, line 361: I like passing the mode instead of having accept_manual().

rklaehn (commented):
I like this approach better than some elaborate scheme with ALPN handlers. I think the complexity is acceptable for advanced use cases, and having some handler trait scheme will break down in many cases.

One thing I was wondering: you can already get the endpoint from the node. So what would happen if somebody would just grab an endpoint and then call accept on it, in the current code? Who wins? Is it round robin?

https://docs.rs/iroh/latest/iroh/node/struct.Node.html#method.magic_endpoint

Frando:
> One thing I was wondering: you can already get the endpoint from the node. So what would happen if somebody would just grab an endpoint and then call accept on it, in the current code? Who wins? Is it round robin?

Yes, it would be round-robin I think, same as `quinn::Endpoint::accept`. 

You could already do this now by the way - the endpoint is already exposed (as you linked). You really shouldn't though because behavior would be unpredictable.

We could remove Node::accept in the PR, it would be identical to calling node.magic_endpoint().accept.

Frando:
Closing in favor of #2357

PR #2285, opened 2024-05-13
switch directions

dignifiedquire:


iroh/examples/custom-protocol.rs
```diff
@@ -26,7 +26,19 @@ async fn main() -> Result<()> {
     // create a new node
     let alpns = vec![EXAMPLE_ALPN.to_vec()];
     let node = iroh::node::Node::memory()
-        .accept_mode(AcceptMode::Manual { alpns })
+        .accept_mode(AcceptMode::Manual {
+            alpns,
+            cb: Box::new(|node, mut connecting| {
```
rklaehn, line 31: What if your protocol handlers need an endpoint to do something time based outside this callback?

When manually driving the loop this is easy, just get the endpoint from the node. With this approach, I guess you would have to grab the endpoint from the node on the first call of this fn, and then put it into some Arc<Mutex<...>>. But then, is it always the same node and always the same endpoint? Is it safe to store the node somewhere? The function signature does not say.

Imagine you wanted to use gossip as a separate custom protocol. Using it using this callback would be a major pain, since the constructor needs an endpoint.

https://docs.rs/iroh-gossip/latest/iroh_gossip/net/struct.Gossip.html#method.from_endpoint

PR #2289, opened 2024-05-14
docs(iroh-net): Minor tweaks in the public iroh_net::dns module

flub:
## Description

This is a bunch of entirely too much nitpcking on the dns module's
documentation, as this is all public:

- Describe clearly what this is for and how it works.

- Summary line is a single full sentence ending in a full stop.

- Summary line is in 3rd person.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/src/dns/node_info.rs
+++ b/iroh-net/src/dns/node_info.rs
@@ -1,5 +1,33 @@
-//! This module contains functions and structs to lookup node information from DNS
-//! and to encode node information in Pkarr signed packets.
+//! Support for handling DNS resource records for dialing by [`NodeId`].
+//!
+//! Dialing by [`NodeId`] is supported by iroh nodes publishing [Pkarr] records to DNS
+//! servers or the Mainline DHT.  This module supports creating and parsing these records.
+//!
+//! DNS records are published under the following names:
+//!
+//! `_iroh.<z32-node-id>.<origin-domain> TXT`
+//!
+//! - `_iroh` is the record name as defined by [`IROH_TXT_NAME`].
+//!
+//! - `<z32-node-id>` is the [z-base-32] encoding of the [`NodeId`].
+//!
+//! - `<origin-domain>` is the domain name of the publishing DNS server, `dns.iroh.link` is
+//!   operated by number0.
+//!
+//! - `TXT` is the DNS record type.
+//!
+//! The returned TXT records must contain a string value of the form `key=value` as defined
+//! in [RFC1464].  The following attributes are defined:
+//!
+//! - `relay=<url>`: The home [`RelayUrl`] of this node.
+//!
+//! - `addr=<addr> <addr>`: A space-separated list of sockets addresses for this iroh node.
+//!   Each address is an IPv4 or IPv6 address with a port.
+//!
+//! [Pkarr]: https://app.pkarr.org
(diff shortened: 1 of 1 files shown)


iroh-net/src/dns/node_info.rs
```diff
@@ -1,5 +1,33 @@
...
+//!
+//! DNS records are published under the following names:
+//!
+//! `_iroh.<z32-node-id>.<origin-domain> TXT`
+//!
+//! - `_iroh` is the record name as defined by [`IROH_TXT_NAME`].
+//!
+//! - `<z32-node-id>` is the [z-base-32] encoding of the [`NodeId`].
+//!
+//! - `<origin-domain>` is the domain name of the publishing DNS server, `dns.iroh.link` is
```
divagant-martian, line 14: any objection to calling N0_DNS_NODE_ORIGIN directly here?

iroh-net/src/dns/node_info.rs
```diff
@@ -16,32 +44,34 @@ use url::Url;
...
 
-/// Lookup node info by domain name
+/// Looks up node info by DNS name.
 ///
-/// The domain name must either contain an _iroh TXT record or be a CNAME record that leads to
-/// an _iroh TXT record.
-pub async fn lookup_by_domain(resolver: &TokioAsyncResolver, domain: &str) -> Result<NodeAddr> {
-    let attrs = TxtAttrs::<IrohAttr>::lookup_by_domain(resolver, domain).await?;
+/// The resource records returned for `name` must either contain an `_iroh` TXT record or be
+/// a CNAME record that leads to an `_iroh` TXT record.
```
divagant-martian, line 67: Even if unlikely to change here I think it's clearer to reference `IROH_TXT_NAME` here

iroh-net/src/dns/node_info.rs
```diff
@@ -69,15 +99,15 @@ pub fn from_z32(s: &str) -> Result<NodeId> {
     Ok(node_id)
 }
 
-/// Node info contained in a DNS _iroh TXT record.
+/// Information about and iroh node which is contained in an `_iroh` TXT resource record.
```
divagant-martian, line 102:
```suggestion
/// Information about the iroh node which is contained in an `_iroh` TXT resource record.
```

or maybe you meant `an iroh node`?

Also, same comment about `IROH_TXT_NAME`

divagant-martian (commented):
I like it. Only nitpick to your nitpicks is that using `_iroh` if at any point requires us to change it is doing test search for a potentially common thing. Referencing the constants is clearer and easier to maintain imo

iroh-net/src/dns/node_info.rs
flub: yeah, i means `an iroh node`.  but `the iroh node` also works

iroh-net/src/dns/node_info.rs
flub: much better indeed, thanks for the pointer

iroh-net/src/dns/node_info.rs
flub: good call, i've replaced these with the constant.

divagant-martian (approved):
love the clarity and the standardization of style

Frando:
Thanks so much for this @flub , this is a great improvement to these docs.

Didn't spot anything on a quick readover but had to leave it for now, if I spot anything later I'll do a followup PR.

Issue #2086, opened 2024-03-15
DNS still flaky on windows


flub:
It seems the windows CI is still giving us timeouts on DNS resolution.  This is weird.

flub:
Let's try to close this again.  The staggered DNS approach seems successful.

PR #2290, opened 2024-05-14
feat(iroh-net): small improvements to dns behaviour

### LINKED ISSUE #2277: Improve DNS lookup behaviour
IIUC the hickory DNS configuragion we use ends up with doing DNS requests using UDP with a 5 seconds timeout with two attempts, i.e. one retry.  These retries are sequential, so the full timeout is 10 seconds.

If the first DNS request is lost, this is essentially fatal for us, netcheck as a whole has a timeout of 5s so will not be able to resolve a relay URL.  The relay client has no specific timeout I think, so might be more successful in connecting (e.g. when connecting to a relay extracted from a NodeAddr).

Given the unreliability of UDP we should probably adopt a similar strategy as how netcheck does it's probes for DNS: perform multiple requests in parallel, but stagger their start time.  The aim is that most requests never send the second query, but should things be slow or a request is lost there are backup requests happening faster than after 5 seconds.

Some rather ad-hoc testing gave lookup times in the range of 50ms - 200ms on various public DNS servers.  Which leads me to suggest the following strategy:

- 1st request at T+0ms
- 2nd request at T+200ms
- 3rd request at T+300ms

And each request with a 3s timeout.

This may be a solution to #2086 as well.

divagant-martian:
## Description

Closes #2277

- According to the issue, we do sequencial queries with a relatively high
  timeout. However, from a quick check of hickory's docs, the default is a
  number of concurrent requests of 2. In summary, this was never actually an
  issue. To ensure we keep this behaviour even if the default changes, set the
  number of concurrent requests to 3.. just to gives us more room. 
- It also reduces the timeout for lookups to make it smaller than netcheck's.
- Change confusing comment that stated we configure Ipv4 AND Ipv6 when we
  actually avoid it.

## Breaking Changes

n/a

## Notes & open questions

From reading hickory's code and adding there some dirty log statements, this is
(and was) working as intended.

## Change checklist

- [x] Self-review.
- [ ] ~Tests if relevant.~

### DIFF
--- a/iroh-net/src/dns.rs
+++ b/iroh-net/src/dns.rs
@@ -69,7 +69,7 @@ fn create_default_resolver() -> Result<TokioAsyncResolver> {
 
     // ensure multiple name servers are queried concurrently
     options.num_concurrent_reqs = 3;
-    options.timeout = std::time::Duration::from_millis(450);
+    options.timeout = std::time::Duration::from_millis(4500);
 
     // see [`lookup_ipv4_ipv6`] for info on why we avoid LookupIpStrategy::Ipv4AndIpv6
     options.ip_strategy = hickory_resolver::config::LookupIpStrategy::Ipv4thenIpv6;
--- a/iroh-net/src/dns.rs
+++ b/iroh-net/src/dns.rs
@@ -68,7 +68,7 @@ fn create_default_resolver() -> Result<TokioAsyncResolver> {
     }
 
     // lookup IPv4 and IPv6 in parallel
-    options.ip_strategy = hickory_resolver::config::LookupIpStrategy::Ipv4thenIpv6;
+    options.ip_strategy = hickory_resolver::config::LookupIpStrategy::Ipv4AndIpv6;
 
     let resolver = AsyncResolver::tokio(config, options);
     Ok(resolver)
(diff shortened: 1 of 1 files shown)


flub:
How are the concurrent requests done by default?  When i looked at logs I never saw concurrent requests happening.  How long before it starts the next request?

The aim is to try and get especially the windows DNS resolution to be more robust.  I think the a schema like described in #2277 would have benefits because we race several requests against each other.  How does hickory's behaviour compare to that?

I also worry about setting the timeout as low as 450ms, from my limited experimentation it is rather likely we would time out valid responses.  I'm much more comfortable with a timeout of >2s I think.

iroh-net/src/dns.rs
```diff
@@ -67,7 +67,11 @@ fn create_default_resolver() -> Result<TokioAsyncResolver> {
         }
     }
 
-    // lookup IPv4 and IPv6 in parallel
+    // ensure multiple name servers are queried concurrently
+    options.num_concurrent_reqs = 3;
```
flub, line 71: Sorry, only just realised this comment.  It queries multiple DNS servers concurrently, but we typically only have a single DNS server.  We want (I think) to be robust in case the UDP packet to this server is lost and thus do multiple concurrent requests to the same server, but with staggered starts so that you don't always pay the price of the multiple requests if your DNS server is fast enough.

flub:
Maybe this PR could also attempt removing the flaky label of the windows DNS tests?  See #2086

divagant-martian:
Closing this to tackle it in a fresh one

PR #2298, opened 2024-05-15
feat(iroh-net)!: Implement http proxy support

### LINKED ISSUE #2295: Iroh fails when running through an outbound HTTPS proxy
* this user is running a proxy that forces outbound HTTP traffic onto a port that isn't 443
* router blocks outbound traffic on 443, it should follow redirects, apparently in the way recent versions of `reqwest` do
* log: https://gist.github.com/zicklag/d6e2fb694650cabcb556258f30dd7405
* from the user: "it looks like it fails with a couple timeouts, which is usually what happens when something tries to go outbound on 443 and the router blocks it."

dignifiedquire:
Headers are based on how `curl` and env variables on`reqwest`.

Setting any of 

- `HTTP_PROXY`
- `http_proxy`
- `HTTPS_PROXY`
- `https_proxy`

will make all relay code use these to proxy outgoing connections.

Closes #2295 

## Breaking Changes

- Added `iroh_net::endpoint::Builder::proxy_url`
- Added `iroh_net::endpoint::Builder::proxy_from_env`
- Added `iroh_net::relay::http::ClientError::Proxy` enum variant

## TODOs

- [x] config & parsing env variables
- [x] the todos in the code
- [x] https proxy 
- [x] testing: tested manually on two machines using `squid`

### DIFF
--- a/iroh-net/Cargo.toml
+++ b/iroh-net/Cargo.toml
@@ -54,7 +54,7 @@ quinn-udp = { package = "iroh-quinn-udp", version = "0.4" }
 rand = "0.8"
 rand_core = "0.6.4"
 rcgen = "0.11"
-reqwest = { version = "0.11.19", default-features = false, features = ["rustls-tls"] }
+reqwest = { version = "0.12.4", default-features = false, features = ["rustls-tls"] }
 ring = "0.17"
 rustls = { version = "0.21.11", default-features = false, features = ["dangerous_configuration"] }
 serde = { version = "1", features = ["derive", "rc"] }
--- a/iroh-net/Cargo.toml
+++ b/iroh-net/Cargo.toml
@@ -19,6 +19,7 @@ workspace = true
 axum = { version = "0.7.4", optional = true }
 aead = { version = "0.5.2", features = ["bytes"] }
 anyhow = { version = "1" }
+base64 = "0.22.1"
 backoff = "0.4.0"
 bytes = "1"
 netdev = "0.25"
@@ -87,8 +88,6 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"], optional = tr
 # metrics
 iroh-metrics = { version = "0.16.0", path = "../iroh-metrics", default-features = false }
 strum = { version = "0.26.2", features = ["derive"] }
-pin-project-lite = "0.2.14"
-base64 = "0.22.1"
 
 [target.'cfg(any(target_os = "linux", target_os = "android"))'.dependencies]
 netlink-packet-core = "0.7.0"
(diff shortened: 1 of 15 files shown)


dignifiedquire:
> Is it a good idea for a library to read environment variables?

`reqwest` is already used, and does exactly this. They do cover a wider variety of supported variables, like `NO_PROXY` and system specific configurations which we can add over time, but for now this allows existing users who need this to work with iroh, where they can't currently

zicklag:
Environment variables for proxy configuration are _extremely_ common. Almost every tool that supports proxies supports it through environment variables, especially if it has a CLI.

Fundamentally I think it makes sense because it is an _environment_ configuration, that is the same for every tool on the machine, in almost every case, and it doesn't seem valuable to make the user comprehend separate configurations for every app they use that needs to access the network if it isn't necessary.

zicklag:
I see your point.

Maybe we should have a `Proxy` config struct that could be passed into the `ClientBuilder` or something like that. We could have a `Proxy::from_env()` function, and if it helps, you could have proxies disabled by default in the library.

Then the Iroh CLI will configure the client with the `Proxy::from_env()`.

It would actually be handy to have the `Proxy` struct, and the environment variable parsing logic in a separate crate so that it could be re-used.

dignifiedquire:
the environment now needs to be read explicitly in `iroh-net`, but `iroh` will by default read it

iroh-net/src/relay/http/client.rs
```diff
@@ -275,6 +291,18 @@ impl ClientBuilder {
...
+    pub fn proxy_url(mut self, url: Url) -> Self {
+        self.proxy_url.replace(url);
+        self
+    }
+
+    /// Disable http proxy entirely, no matter the environment variables.
+    pub fn no_proxy(mut self) -> Self {
+        self.no_proxy = true;
+        self
+    }
```
flub, line 304: This (and the field) are unused?  Left over from a previous version?

iroh-net/Cargo.toml
```diff
@@ -87,6 +87,8 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"], optional = tr
 # metrics
 iroh-metrics = { version = "0.16.0", path = "../iroh-metrics", default-features = false }
 strum = { version = "0.26.2", features = ["derive"] }
+pin-project-lite = "0.2.14"
+base64 = "0.22.1"
```
flub, line 91: These are not for the metrics section.  maybe we should just remove the "sections" and simply keep one sorted list.

We also already have `pin_project` in the deps.  We should probably not have both, at least not directly, indirectly we get both anyway I think.  Personally I much prefer the attribute-style over this function-style as it integrates much better with tooling generally, no structs inside macros.  And the compilation cost isn't that much.

flub (approved):
I like the approach of the builder in `Endpoint`.  Only minor note I have is that we now have two pin-project crates...

iroh-net/src/relay/http/client.rs
dignifiedquire: removed

PR #2299, opened 2024-05-15
feat: node-wide default author for documents

Frando:
## Description

This adds the notion of a node-wide default author for documents.

The default author can be retrieved via `iroh.authors.default()`. On persistent nodes, on first start a new author is created and its public key is saved to `IROH_DATA_DIR/default-author` (as a base32 string). For in-memory nodes, a new author is created.

When trying to delete the default author via `iroh.authors.delete`, an error is returned, stating that the default author may not be deleted.

The default author can not be changed. 

If the `default-author` file is deleted manually, a new default author will be created on the next start.

## Breaking Changes

* Added `AuthorClient::default()`, which returns author id of a single default author per node 

## Notes & open questions

There are certainly other ways to go around the interactions between the iroh-local notion of a default author and the author database in the iroh-docs store. Quoting from discord (this PR implements option B.1):

> if you delete the default author with the author api, what should happen?
>* A: default author remains in place, further operations with it fail
>  * A.1: on next start new default author is created automatically
>  * A.2: next start fails, says "delete default_author file by hand"
>* B: deletion of default author fails. 
>  * B1: that's the way it is, only way to delete it is to delete `default-author` file by hand and restart the node
>  * B2: add `set_default_author` rpc call
>* C: a new default author is created and set always & automatically

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/client/authors.rs
+++ b/iroh/src/client/authors.rs
@@ -6,8 +6,8 @@ use iroh_docs::{Author, AuthorId};
 use quic_rpc::{RpcClient, ServiceConnection};
 
 use crate::rpc_protocol::{
-    AuthorCreateRequest, AuthorDeleteRequest, AuthorExportRequest, AuthorImportRequest,
-    AuthorListRequest, RpcService,
+    AuthorCreateRequest, AuthorDefaultRequest, AuthorDeleteRequest, AuthorExportRequest,
+    AuthorImportRequest, AuthorListRequest, RpcService,
 };
 
 use super::flatten;
@@ -25,6 +25,12 @@ where
     /// Create a new document author.
     pub async fn create(&self) -> Result<AuthorId> {
         let res = self.rpc.rpc(AuthorCreateRequest).await??;
+        Ok(res.author_id)
+    }
+
+    /// Get the default document author of this node.
+    pub async fn default(&self) -> Result<AuthorId> {
+        let res = self.rpc.rpc(AuthorDefaultRequest).await?;
         Ok(res.author_id)
     }
 
@@ -53,6 +59,8 @@ where
     /// Deletes the given author by id.
     ///
     /// Warning: This permanently removes this author.
+    ///
+    /// Deleting the default author is not supported.
     pub async fn delete(&self, author: AuthorId) -> Result<()> {
         self.rpc.rpc(AuthorDeleteRequest { author }).await??;
         Ok(())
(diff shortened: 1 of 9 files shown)


dignifiedquire:
> * add `set_default_author` rpc call

I think this would be good to have, but we can add it later when needed

Frando:
> I think this would be good to have, but we can add it later when needed
added this in the most recent commit!

iroh/src/docs_engine.rs
```diff
@@ -46,20 +48,77 @@ pub struct Engine {
...
+                }
+            }
+        }
+    }
+    pub async fn save(&self, docs_store: &SyncHandle, author_id: AuthorId) -> anyhow::Result<()> {
+        if docs_store.export_author(author_id).await?.is_none() {
+            bail!("The author does not exist");
+        }
+        match self {
+            Self::Mem => {}
```
dignifiedquire, line 100: this is surprising, why not store it in memory?

Frando: It is stored on the `DocsEngine` struct already, so `save` is a noop for mem storage. Will add a comment.

Frando: OK pushed a new commit that makes this nicer (and actually work).

iroh/src/client/authors.rs
```diff
@@ -23,11 +23,40 @@ where
...
+    pub async fn default(&self) -> Result<AuthorId> {
+        let res = self.rpc.rpc(AuthorGetDefaultRequest).await?;
+        Ok(res.author_id)
+    }
+
+    /// Set the node-wide default author.
+    ///
+    /// If the author does not exist, an error is returned.
+    ///
+    /// This is a noop on memory nodes. On peristent node, the author id will be saved to a file in
```
dignifiedquire, line 51: why is this a noop, shouldn't this change the default on a memory node as well?

iroh/src/docs_engine.rs
```diff
@@ -95,13 +103,24 @@ impl Engine {
             .instrument(error_span!("sync", %me)),
         );
 
-        Self {
+        let default_author = match DefaultAuthor::load(default_author_storage, &sync).await {
+            Ok(author) => author,
+            Err(err) => {
```
dignifiedquire, line 108: why exit in this case, instead of generating a new one?

iroh/src/client/authors.rs
Frando: Fixed the comment.

iroh/src/docs_engine.rs
Frando: This only returns an error if parsing the default author file failed. If the default author file is non-existent, a default author is created already without returning an error, see line 337ff.

dignifiedquire (approved):

PR #2302, opened 2024-05-16
feat: emit PendingContentReady event

Frando:
## Description

This adds a new `LiveEvent` variant: `PendingContentReady`.

It is emitted once, after sync was started, the list of pending content blobs becomes empty. This is the case if either all blobs from this sesion have been downloaded, or failed.

I think this could help with simple examples like https://gist.github.com/dignifiedquire/efbd1a7a1da729494adb088d72f1ceaa#file-main-sync-rs-L84

It should not be used as a guarantee that all content is now available locally, and it should be used with a timeout usually, because it may very legally never be emitted (e.g. if the remote does not have all blobs themselves, or closes the connections, etc).

## Breaking Changes

* `LiveEvent` now has a new variant `PendingContentReady`. This event is emitted  after a sync completed and `Self::SyncFinished` was emitted. It signals that all currently pending downloads have been completed. Receiving this event does not guarantee that all content in the document is available. If blobs failed to download, this event will still be emitted after all operations completed.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-cli/src/commands/doc.rs
+++ b/iroh-cli/src/commands/doc.rs
@@ -609,6 +609,9 @@ impl DocCommands {
                         LiveEvent::NeighborDown(peer) => {
                             println!("neighbor peer down: {peer:?}");
                         }
+                        LiveEvent::PendingContentReady => {
+                            println!("all pending content is now ready")
+                        }
                     }
                 }
             }
--- a/iroh/src/client/docs.rs
+++ b/iroh/src/client/docs.rs
@@ -547,6 +547,14 @@ pub enum LiveEvent {
     NeighborDown(PublicKey),
     /// A set-reconciliation sync finished.
     SyncFinished(SyncEvent),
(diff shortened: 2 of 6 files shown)


dignifiedquire (approved):

dignifiedquire:
```
2024-05-17T10:38:10.257646Z DEBUG gossip{me=icdmm64o2lmqfyxw}: iroh_gossip::net: handle out_event SendMessage(PublicKey(4dxirwpnfj5gb3av), Message { topic: TopicId(lofyckr47fnbhjaq), message: Gossip(Gossip(Gossip { id: MessageId(nazywnpxm4nuce77), content: <239b>, scope: Swarm(Round(0)) })) })
thread 'sync_restart_node' panicked at iroh\tests\sync.rs:1316:9:
Failed to receive or match all events: Event didn't match any matcher: PendingContentReady
```

this looks like a real failure?

PR #2304, opened 2024-05-16
examples: minimize use of raw base32 in examples

marceline-cramer:
## Description

I noticed that in the example code (particularly `iroh-net`'s code), there are a few explicit uses of `base32` encoding and decoding that are unnecessary. In particular, `SecretKey` is manually `base32`-encoded in the networking examples, although the `Display` implementation uses that same encoding. I figure that the examples were just written before the utility encoding/decoding code in the main libraries were and weren't updated when that utility code was added. I hope that catching this is helpful!

## Breaking Changes

None.

## Notes & open questions

None.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


dignifiedquire (approved):
thanks a lot, those examples were probably written before we had all these niceties :)

PR #2308, opened 2024-05-19
fix(iroh-gossip): do not drop existing peer connection when we get incoming one

### LINKED ISSUE #2307: iroh-gossip fails to establish connection if it is established simultaneously from both sides
Here if we get incoming connection and already have a connection with the same peer, old connection is dropped:
https://github.com/n0-computer/iroh/blob/370075c6d5689ad4349664eb9b51ec0a5e7e4e81/iroh-gossip/src/net.rs#L425
This is fine because connection is just a reference and it got cloned into `connection_loop` already.

But here we create new `send_tx` and evict existing one from the map:
https://github.com/n0-computer/iroh/blob/370075c6d5689ad4349664eb9b51ec0a5e7e4e81/iroh-gossip/src/net.rs#L428

Previously started connection loop detects that its `send_rx` side got closed and breaks:
https://github.com/n0-computer/iroh/blob/370075c6d5689ad4349664eb9b51ec0a5e7e4e81/iroh-gossip/src/net.rs#L620

Then existing connection gets dropped and only incoming connection remains.

If two peers try to establish connection to each other simultaneously, this is what happens. First, node establishes outgoing connection. Then it gets incoming connection and closes its outgoing connection because `send_rx` got closed. At this point "connection closed without error" is printed.

Because this happens on the other side, the other side also closes its outgoing connection. For the first peer it is incoming connection. When it detects that incoming connection got closed, it logs 
```
iroh_gossip::net: connection closed with error connection lost

Caused by:
    closed by peer: 0
```

Then both sides have no connection anymore.

Not accepting incoming connection does not work either, then we have two outgoing connection and don't listen on the incoming ones. I think we should just keep incoming connection even if it is replaced (`send_tx` gets dropped) until it is closed by the other side, e.g. because it leaves gossip explicitly.

This is my analysis of what happens in https://github.com/deltachat/deltachat-core-rust/pull/5595

link2xt:
This is a result of debugging a problem in Delta Chat when two peers try to establish connection to each other by subscribing to the same topic and supplying each other peer address to `.join()` roughly at the same time.

Fixes #2307

### DIFF
--- a/iroh-gossip/src/net.rs
+++ b/iroh-gossip/src/net.rs
@@ -433,10 +433,22 @@ impl Actor {
 
                 // Spawn a task for this connection
                 let in_event_tx = self.in_event_tx.clone();
+
+                // Clone send_tx and give it to the connection loop
+                // to prevent the channel from closing.
+                // Otherwise connection loop will exit as soon as we drop send_tx
+                // from conn_send_tx map.
+                //
+                // We don't want to exit connection loop
+                // and close the connection just because we don't want to
+                // use the connection for sending anymore, because the other side
+                // may still want to use the connection to send data to us.
+                let send_tx_clone = send_tx.clone();
+
                 tokio::spawn(
                     async move {
                         debug!("connection established");
-                        match connection_loop(peer_id, conn, origin, send_rx, &in_event_tx).await {
+                        match connection_loop(peer_id, conn, origin, send_tx_clone, send_rx, &in_event_tx).await {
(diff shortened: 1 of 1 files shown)


hpk42 (commented):
Wow, so many hours of debugging (dig, l, me, Sebi yesterday) and then it's effectively a 3 line change that verifiably fixes our long standing "it doesn't reliably work" issue. Great stuff. Network programing is hard :)

PR #2310, opened 2024-05-19
fix(iroh-gossip): do not drop existing peer connection when we get incoming one

link2xt:
Otherwise if two peers try to connect to each other at the same time, they may end up creating two connection and then closing both of them. Better keep two one-directional connections.

deltachat-core-rust will have a downstream test that reliably triggers the problem at https://github.com/deltachat/deltachat-core-rust/pull/5595
I did not manage to reproduce the problem in Rust, it seems to depend on timing and is triggered in the test which sends tickets around using SMTP.

Fixes https://github.com/n0-computer/iroh/issues/2307

Replaces https://github.com/n0-computer/iroh/pull/2308

### DIFF
--- a/iroh-gossip/src/net.rs
+++ b/iroh-gossip/src/net.rs
@@ -619,11 +619,13 @@ async fn connection_loop(
     loop {
         tokio::select! {
             biased;
-            msg = send_rx.recv() => {
-                match msg {
-                    None => break,
-                    Some(msg) =>  write_message(&mut send, &mut send_buf, &msg).await?,
-                }
+            // If `send_rx` is closed,
+            // stop selecting it but don't quit.
+            // We are not going to use connection for sending anymore,
+            // but the other side may still want to use it to
+            // send data to us.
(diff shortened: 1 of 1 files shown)


iroh-gossip/src/net.rs
```diff
@@ -616,12 +616,20 @@ async fn connection_loop(
     };
     let mut send_buf = BytesMut::new();
     let mut recv_buf = BytesMut::new();
+    let mut tx_open = true;
     loop {
         tokio::select! {
             biased;
-            msg = send_rx.recv() => {
+            msg = send_rx.recv(), if tx_open => {
```
divagant-martian, line 623: this can be done with `!send_rx.is_closed`

dignifiedquire:
`maint-0.16` exist because there were issues when cutting the release, given that we are cutting a release every 2-3 weeks currently I would much more prefer to merge this into `main` and DC just upgrading to that commit. And you can switch over to the released version in a week. 

The breaking changes on `main` are mostly renames which are easy to adapt to.

PR #2313, opened 2024-05-20
feat(iroh-net)!: improve dns behaviour by staggering requests

divagant-martian:
## Description

Improves dns behaviour by staggering concurrent queries. Other changes include:
- Remove unnecessary trait bounds from `ResolverExt` functions.
- Unify dns capabilities under `ResolverExt`.
- Add helper fn `stagger_call` with a minimal test.
- Add staggering versions of the `ResolverExt` functions.
- Use staggered options and turn on tests

## Breaking Changes

- `iroh_net::dns::node_info::lookup_by_domain` moved to `iroh_net::ResolverExt::lookup_by_domain`

## Notes & open questions

Whether test are less prone to fail randomly is something I can't test locally (no windows) so I'm trusting ci here.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh-net/src/dns.rs
```diff
@@ -97,10 +100,69 @@ pub trait ResolverExt {
...
+
+    /// Perform an ipv4 lookup with a timeout in a staggered fashion.
+    ///
+    /// The first call is done immediately, with added calls according to the staggering strategy.
+    fn staggered_lookup_ipv4<N: IntoName + Clone>(
+        &self,
+        host: N,
+        timeout: Duration,
+        delays_ms: &[u64],
+    ) -> impl Future<Output = Result<impl Iterator<Item = IpAddr>>>;
```
dignifiedquire, line 122: we can use `async fn` in traits now I think :=

divagant-martian: this generates a warning, let me paste it

divagant-martian:
```
warning: use of `async fn` in public traits is discouraged as auto trait bounds cannot be specified
  --> iroh-net/src/dns.rs:84:5
   |
84 |     async fn lookup_ipv4<N: IntoName>(
   |     ^^^^^
   |
   = note: you can suppress this lint if you plan to use the trait only in your own code, or do not care about auto traits like `Send` on the `Future`
```

dignifiedquire: ah okay, I didn't realize this was the case

divagant-martian: the implementation itself does use async, the declaration doesn't, unless we suppress the warning . What do you think?

iroh-net/src/dns.rs
```diff
@@ -97,10 +100,69 @@ pub trait ResolverExt {
         host: N,
         timeout: Duration,
     ) -> impl Future<Output = Result<impl Iterator<Item = IpAddr>>>;
+
+    /// Looks up node info by DNS name.
+    fn lookup_by_domain(&self, name: &str) -> impl Future<Output = Result<NodeAddr>>;
```
flub, line 105: fwiw I find `by_domain` a bit odd here, since it is a hostname you're looking up.  `by_name` would read a little more logical here.  but feel free to disagree and leave it as is.

iroh-net/src/dns.rs
```diff
@@ -97,10 +100,69 @@ pub trait ResolverExt {
...
+    /// Looks up node info by [`NodeId`] and origin domain name.
+    fn lookup_by_id(
+        &self,
+        node_id: &NodeId,
+        origin: &str,
+    ) -> impl Future<Output = Result<NodeAddr>>;
+
+    /// Perform an ipv4 lookup with a timeout in a staggered fashion.
+    ///
+    /// The first call is done immediately, with added calls according to the staggering strategy.
```
flub, line 116: Describing `delays_ms` a little more would be helpful.  E.g. are they all  offsets from the time this function is called, or are they offsets from the last request?

iroh-net/src/dns.rs
flub: Additionally it caught me out that this returned `NodeAddr`.  I only noticed when seeing the impl (which also contains a more extensive doc comment, maybe copy that here?).  Since the `lookup_` prefix is already used by hickory to return `IpAddr`, maybe we could use another prefix?  Like maybe `node_by_`?

iroh-net/src/dns.rs
```diff
@@ -97,10 +100,69 @@ pub trait ResolverExt {
...
+    fn staggered_lookup_ipv4<N: IntoName + Clone>(
```
flub, line 117: Since I'm in a naming nitpicking mood, maybe `lookup_ipv4_staggered` is the right thing here as it leaves the `lookup_` as prefix?
(I'm less sure about `lookup_ipv4_staggered` vs `lookup_staggered_ipv4`)

iroh-net/src/dns/node_info.rs
flub, line 1: Another bit of API pondering.  I wonder if these could be a separate extension trait defined in this module?  It would keep the methods close to the module-level description here which is otherwise a bit lost.

iroh-net/src/netcheck/reportgen.rs
```diff
@@ -1316,7 +1325,6 @@ mod tests {
     //
     // TODO: Not sure what about IPv6 pings using sysctl.
     #[tokio::test]
-    #[cfg_attr(target_os = "windows", ignore = "flaky")]
```
flub, line 1319: :tada: :cold_sweat:

iroh-net/src/netcheck/reportgen.rs
```diff
@@ -192,7 +195,7 @@ struct Actor {
     ///
     /// This is essentially the summary of all the work the [`Actor`] is doing.
     outstanding_tasks: OutstandingTasks,
-    /// The DNS resolver to use for probes that need to resolve DNS records
+    /// The DNS resolver to use for probes that need to resolve DNS records.
```
flub, line 198: :heart:

flub (approved):
All the code looks good, basically I'm mostly bikeshedding nothing I want to block things for as it's subjective anyway.

iroh-net/src/dns.rs
divagant-martian: `by_name` feels more intuitive to me, `lookup` doesn't bother me because it's still a lookup and it still returns ip addresses, just conveniently bundled up with the rest of the required info, so I'd leave it at that. If you feel strongly about it we can discuss it further ofc. In the meantime I've done just the first change in 9be4c1c

iroh-net/src/dns.rs
divagant-martian: I know you are right but I'm struggling to convey that these are not offsets in a way that is not terribly verbose. Any suggestions?

iroh-net/src/dns.rs
divagant-martian: I like the reasoning for `lookup_`, I think it makes it easier to find if you just have the lsp suggesting code. Done in 373e172

iroh-net/src/dns/node_info.rs
divagant-martian: I made the change exactly for this reason. Not in favor of `dns::node_info` but in favor of `dns` as a whole. IMO before this the module did not present a unified front. This much is supported by the fact that the functions defined in the `node_info` submodule were missed in prev refactors. 

One that I think would be a better alternative is to instead mention both the trait and top module docs of `node_info` in the module docs for `dns`. My reasoning is that at the end of the day, for an user of `dns` what it's of importance is how to query nodes by id (since this was the whole reason we added this) much more than the exact details of the record structure, which is already clearly explained in `node_info`.

I've done this in 118c390

iroh-net/src/netcheck/reportgen.rs
divagant-martian: I made the windows all-features, stable workflow run three times and it passed. I have no idea how many runs we feel to be comfortable with this again, but hey it's something? new commits will be the fourth one. Let's see how it goes

flub (approved):
Nice, I think the trait method names are now nicely consistent.  And their first lines of the doc comments descriptive.

PR #2329, opened 2024-05-27
docs(iroh-net): Update toplevel module documentation

flub:
## Description

This updates the top-level docs for iroh-net to be a general
overview of what the crate provides, since this is effectively the
docs.rs landing page.

## Breaking Changes

None

## Notes & open questions

This should include an examples section as well.  However... the API
is a bit pants and does not lend itself to a nice echo example here.
Let's tackle this separately.

This does not change the README, that is already fairly decent and
seems compatible with this content.

## Change checklist

- [x] Self-review.
- ~~[ ] Tests if relevant.~~

### DIFF
--- a/iroh-net/src/lib.rs
+++ b/iroh-net/src/lib.rs
@@ -1,11 +1,102 @@
-//! iroh-net provides connectivity for iroh.
+//! Peer-to-peer connectivity based on QUIC.
 //!
-//! This crate is a collection of tools to establish connectivity between peers.  At
-//! the high level [`Endpoint`] is used to establish a QUIC connection with
-//! authenticated peers, relaying and holepunching support.
+//! iroh-net is a library to establish direct connectivity between peers.  It exposes an
+//! interface to [QUIC] connections and streams to the user, while implementing direct
+//! connectivity using [hole punching] complemented by relay servers under the hood.
 //!
-//! The "relay-only" feature forces all traffic to send over the relays. We still
-//! receive traffic over udp and relay. This feature should only be used for testing.
+//!
+//! # Connection Establishment
+//!
+//! An iroh-net connection between two iroh-net nodes is usually established with the help
+//! of a Relay server.  When creating the [`Endpoint`] it connects to the closest Relay
+//! server and designates this has the *home relay*.  When other nodes want to connect they
+//! first establish connection via this home relay.  As soon as connection between the two
+//! nodes is established they will attempt to create a direct connection, using [hole
+//! punching] if needed.  Once the direct connection is established the relay server is no
+//! longer involved in the connection.
+//!
+//! If one of the iroh-net nodes can be reached directly connectivity can also be
+//! established without involving a Relay server.  The [`NodeAddr`] can also contain a
+//! number of [socket addresses] on which a node is reachable and will be used to establish a
+//! connection.
+//!
+//!
+//! # Encryption
+//!
+//! The connection is encrypted using TLS, like standard QUIC connections.  Unlike standard
+//! QUIC there is no client, server or server TLS key.  Instead each iroh-net node has a
+//! unique [`SecretKey`] used to authenticate and encrypt the connection.  When an iroh-net
+//! node connects it uses the corresponding [`PublicKey`] to ensure the connection is only
+//! established with the intended peer.
+//!
+//! Since the [`PublicKey`] is also used to identify the iroh-net node it is also known as
+//! the [`NodeId`].  As encryption is an integral part of TLS as used in QUIC this
+//! [`NodeId`] is always a required parameter to establish a connection.
+//!
+//! When accepting connections the peer's [`NodeId`] is authenticated.  However it is up to
+//! the application to decide if a particular peer is allowed to connect or not.
+//!
+//!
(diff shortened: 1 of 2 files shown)


iroh-net/src/lib.rs
```diff
@@ -1,11 +1,102 @@
...
+//! the application to decide if a particular peer is allowed to connect or not.
+//!
+//!
+//! # Relay Servers
+//!
+//! Relay servers exist to ensure all iroh-net nodes are always reachable.  They accept
+//! **encrypted** traffic for iroh-net nodes which are connected to them, forwarding it to
+//! the correct destination based on the [`NodeId`] only.  Since nodes only send encrypted
+//! traffic, the Relay servers can not decode any traffic for other iroh-net nodes and only
+//! forwards it.
```
divagant-martian, line 46:
```suggestion
//! forward it.
```

iroh-net/src/lib.rs
```diff
@@ -1,11 +1,102 @@
...
+//!
+//! # Connections and Streams
+//!
+//! Connections are managed using the [`Endpoint`].  To establish a connection to an
+//! iroh-net node you need to know three pieces of information:
+//!
+//! - The [`NodeId`] of the peer to connect to.
+//! - Some addressing information:
+//!   - Usually the [`RelayUrl`] identifying the Relay server.
+//!   - Sometimes, or usually additionally, any direct addresses which might be known.
```
divagant-martian, line 69: Here is worth mentioning dns since we technically might now not need _any_ addressing information beside the node id itself

iroh-net/src/lib.rs
```diff
@@ -1,11 +1,102 @@
...
+//! [`HTTP/3`], but more commonly will be a custom identifier for the application.
+//!
+//! Once connected the API exposes QUIC streams.  These are very cheap to create so can be
+//! created at any time and can be used to create very many short-lived stream as well as
+//! long-lived streams.  There are two stream types to choose from:
+//!
+//! - **Uni-directional** which only allows the peer which initiated the stream to send
+//!   data.
+//!
+//! - **Bi-directional** which allows both peers to send and receive data.  However the
```
divagant-martian, line 83:
```suggestion
//! - **Bi-directional** which allows both peers to send and receive data.  However, the
```

iroh-net/src/lib.rs
```diff
@@ -1,11 +1,102 @@
...
+//!   initiator of this stream has to send data before the peer will be aware of this
+//!   stream.
+//!
+//! Additionally to be extremely light-weight, streams can be interleaved and will not block
```
divagant-martian, line 87:
```suggestion
//! Additionally, to be extremely light-weight, streams can be interleaved and will not block
```

divagant-martian (changes requested):
A lot better.

My two comments:
1. Mention of dns.
2. There are a few commas missing, and I most likely didn't point them all. Risking being even more pedantic, here there is a small pdf about commas in english from some us university https://www.svsu.edu/media/writingcenter/activedocs/Using%20Commas%20Within%20Sentences.pdf in particular about subordinating conjunctions like when, as, if etc. These felt more in plsce to me so I, as a non-native english speaker, wanted to be sure. Please consider it an education resource for both more than me being a prick :sweat_smile:

iroh-net/src/lib.rs
flub: I'm not sure this comma works really.  Maybe it's better to avoid this construction entirely, e.g.:
```suggestion
//! As well as streams being extremely light-weight, they can be interleaved and will not block
```
Would that be better?

iroh-net/src/lib.rs
```diff
@@ -1,11 +1,102 @@
...
+//! If one of the iroh-net nodes can be reached directly, connectivity can also be
+//! established without involving a Relay server.  The [`NodeAddr`] can also contain a
+//! number of [socket addresses] on which a node is reachable and will be used to establish a
+//! connection.
+//!
+//!
+//! # Encryption
+//!
+//! The connection is encrypted using TLS, like standard QUIC connections.  Unlike standard
+//! QUIC there is no client, server or server TLS key.  Instead each iroh-net node has a
```
flub, line 27: Would this change work?
```suggestion
//! QUIC there is no client, server or server TLS key and certificate chain.  Instead each iroh-net node has a
```

iroh-net/src/lib.rs
flub: Good catch that I missed this, I added a discovery section below.  Do you think that's a good way of addressing this?

flub:
>     2. There are a few commas missing, and I most likely didn't point them all. Risking being even more pedantic, here there is a small pdf about commas in english from some us university https://www.svsu.edu/media/writingcenter/activedocs/Using%20Commas%20Within%20Sentences.pdf in particular about subordinating conjunctions like when, as, if etc. These felt more in plsce to me so I, as a non-native english speaker, wanted to be sure. Please consider it an education resource for both more than me being a prick 😅

That's totally fine!  Good resource.

I also liked https://developers.google.com/tech-writing which I'm trying to keep in mind, at least the parts I remember.  But I probably need a refresher on it :)

Issue #2327, opened 2024-05-25
More delays when the network interface is down on Windows .


zh522130:
This is a bug in `netdev`, and I have submitted a Pull Request, awaiting its merge.

https://github.com/shellrow/netdev/pull/70

Issue #2328, opened 2024-05-27
Optimize data transfer delay


zh522130:
I noticed that when the connection type is `ConnectionType::Mixed`, the data transfer delay seems to include the delay from `netcheck`, and under the network environment mentioned at https://github.com/n0-computer/iroh/issues/2327, the transfer delay can be further amplified. The log only captures instances where the transmission delay was very high. I'm not certain if other connection types include this part of the delay. However, from my test results, it seems that iroh's average delay is somewhat higher than that of Tailscale, possibly slower by about 1ms in a local area network condition (this is not very precise, as the router itself can also have fluctuations).

**Code:**
```rust
loop {
        let now = std::time::Instant::now();
        println!("====================> start send");
        let (mut send, mut recv) = conn.open_bi().await?;
        send.write_all(message).await?;

        // Call `finish` to close the send side of the connection gracefully.
        send.finish().await?;
        let message = recv.read_to_end(4096).await?;
        let tk_ms = now.elapsed().as_millis();
        if tk_ms > 2000 {
            println!("xxxxxxxxxxxxxxxxx=> took: {}ms\n\n\n", now.elapsed().as_millis());
        } else {
            println!("took: {}ms", now.elapsed().as_millis());
        }
        let message = String::from_utf8(message)?;
        println!("received: {message}");

        let conninfo = endpoint.connection_info(remote_node_id);
        if let Some(con_info) = conninfo {
            println!("connect type:{:?}", con_info.conn_type);
        }
        println!("====================> end send");
        sleep(Duration::from_millis(500)).await;
    }
```

**Log:**
```
====================> start send
2024-05-27T01:35:39.001511Z  INFO poll_send{me=27evu2sogrbp5m6q}:get_send_addrs{node=w6fj3gwbwqpowxjr}: iroh_net::magicsock::node_map::node_state: new connection type typ=mixed
2024-05-27T01:35:39.001819Z DEBUG poll_send{me=27evu2sogrbp5m6q}:get_send_addrs{node=w6fj3gwbwqpowxjr}:want_call_me_maybe: iroh_net::magicsock::node_map::node_state: best addr not set: need full ping
[...]

PR #2330, opened 2024-05-28
fix(iroh-net): fix extra delay

zh522130:
## Description

Fix for https://github.com/n0-computer/iroh/issues/2328: Enumerating network devices is slow
on windows and causes delays in the magicsock actor.  This makes sure that this operation
does not block the remainder of the actor.

Also bumps netdev to v0.27.0 , to fix https://github.com/n0-computer/iroh/issues/2327.

### DIFF
--- a/iroh-net/src/magicsock.rs
+++ b/iroh-net/src/magicsock.rs
@@ -2001,13 +2001,13 @@ impl Actor {
             .unwrap_or(false);
 
         let msock = self.msock.clone();
-        
-        tokio::spawn( async move {
+
+        tokio::spawn(async move {
             let LocalAddresses {
                 regular: mut ips,
                 loopback,
             } = LocalAddresses::new();
-    
+
             if is_unspecified_v4 || is_unspecified_v6 {
                 if ips.is_empty() && eps.is_empty() {
                     // Only include loopback addresses if we have no
@@ -2024,7 +2024,7 @@ impl Actor {
                         None
                     }
                 });
-    
+
                 let v6_port = local_addr_v6.and_then(|addr| {
                     if addr.ip().is_unspecified() {
                         Some(addr.port())
@@ -2032,7 +2032,7 @@ impl Actor {
                         None
                     }
                 });
-    
+
                 for ip in ips {
                     match ip {
                         IpAddr::V4(_) => {
@@ -2058,7 +2058,7 @@ impl Actor {
                     }
                 }
             }
-    
+
             if !is_unspecified_v4 {
                 if let Some(addr) = local_addr_v4 {
                     // Our local endpoint is bound to a particular address.
@@ -2066,7 +2066,7 @@ impl Actor {
                     add_addr!(already, eps, addr, config::EndpointType::Local);
                 }
             }
-    
+
             if !is_unspecified_v6 {
                 if let Some(addr) = local_addr_v6 {
                     // Our local endpoint is bound to a particular address.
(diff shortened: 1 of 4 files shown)


zh522130:
@flub I have fixed formatting issues. Could you please help to approve the CI workflow again? Thank you very much!

iroh-net/src/magicsock.rs
```diff
@@ -2000,102 +2000,105 @@ impl Actor {
...
-                    Some(addr.port())
-                } else {
-                    None
+        let msock = self.msock.clone();
+
+        tokio::spawn(async move {
+            let LocalAddresses {
+                regular: mut ips,
+                loopback,
+            } = LocalAddresses::new();
```
flub, line 2009: This can still block the executor this is being spawned on.  Could you wrap the new() call into `tokio::task::spawn_blocking()`?  And also leave a comment that on windows this introduces a 10-20ms of delay so that we know why.

flub (commented):
This is a fairly reasonable way to go about this I guess.  The structure of the magicsocket is not great here, but this is just living with that and improving it is a whole other project.

flub:
See https://github.com/n0-computer/iroh/pull/2332 for the flaky test this found :(

iroh-net/src/magicsock.rs
zh522130: I have made the modifications as per your suggestion. I specifically tested the time consumed by the latest version of `netdev::get_interfaces()` under various conditions. The results are as follows: 

```
windows（Lan + wifi off）：min: 9ms, max: 28ms
windows（Lan + wifi connects to mobile hotspot ）：min: 24ms, max: 1174ms

macos (Hackintosh X86): min: 1ms, max: 6ms

linux (7840hs + lan): min: 100us, max: 502us
linux (7840hs + 1lan connect + 3lan disconnect): min: 100us, max: 1092us
linux (2core cpu server + lan): min: 100us, max: 960us
```

The time taken for the function call may be influenced by the platform, the number of network interfaces, and CPU performance (such as on embedded devices or server). Therefore, in the comments, I did not limit my annotations to just Windows. The test results show that the Windows platform indeed takes the most time. Analyzing the source code of `netdev` for the Windows platform confirms that the outcome aligns with expectations.

iroh-net/src/magicsock.rs
```diff
@@ -2000,102 +2000,109 @@ impl Actor {
...
-            let v4_port = local_addr_v4.and_then(|addr| {
-                if addr.ip().is_unspecified() {
-                    Some(addr.port())
-                } else {
-                    None
+        let msock = self.msock.clone();
+
+        tokio::spawn(async move {
+            // Spawn the blocking call in a separate thread to prevent blocking the executor,
+            // and causing additional delay.
```
flub, line 2007:
```suggestion
            // Depending on the OS and network interfaces attached and their state enumerating
            // the local interfaces can take a long time.  Especially Windows is very slow.
```

flub (commented):
clippy is unhappy, otherwise looks good

zh522130:
> clippy is unhappy, otherwise looks good

Thank you for your patient guidance.

iroh-net/src/magicsock.rs
flub: Thanks for these detailed numbers, it's great to see these written down!

flub (approved):

flub:
> Thank you for your patient guidance.

Thanks for your contributions!  Finding and fixing where those delays are from is great!

PR #2333, opened 2024-05-28
fix(iroh-net): prevent adding addressing info that points back to us

divagant-martian:
## Description

Failed disco boxes for which the sender is our own node is are because we tried to contact another node for which the addressing information we have points back to us. We are the sender (this explains the source of the disco box being ourselves) but we are not the meant recipient (thus the failed decryption - we should always be able to decrypt a message meant to ourselves - ).

This is of course undesirable. This Pr should mostly mitigate this issue, but a failed disco box to ourselves can happen in other less likely ways.

## Breaking Changes

n/a

## Notes & open questions

n/a

## Change checklist

- [x] Self-review.
- [ ] ~Tests if relevant.~


iroh-net/src/endpoint.rs
```diff
@@ -640,17 +640,33 @@ impl Endpoint {
     /// present in the netmap.
     ///
     /// # Errors
-    /// Will return an error if we attempt to add our own [`PublicKey`] to the node map.
+    /// Will return an error if we attempt to add our own [`PublicKey`] to the node map, or if the
+    /// node's direct addresses are a subset of our local endpoints.
```
flub, line 644: OMG.  of course  :facepalm:

dignifiedquire:
@divagant-martian what's missing from this to be ready?

iroh-net/src/endpoint.rs
dignifiedquire: where is this actually enforced? the local endpoints part

iroh-net/src/endpoint.rs
```diff
@@ -535,17 +535,40 @@ impl Endpoint {
...
+    ///
+    /// This updates the local state for the remote node.  If the provided [`NodeAddr`] contains a
+    /// [`RelayUrl`] this will be used as the new relay server for this node.  If it contains any
+    /// new IP endpoints they will also be stored and tried when next connecting to this node. The
+    /// source is used for logging exclusively and will not be stored.
+    ///
+    /// # Errors
+    ///
+    /// Will return an error if we attempt to add our own [`PublicKey`] to the node map or if the
+    /// direct addresses are a subset of ours.
```
flub, line 554: When i first read this i assumed it would error as soon as one of the added addresses is also one of ours.  But the implementation silently removes those instead.  I know the wording here is technically correct, but maybe there's a different way of wording this.

```
or if the direct addresses contains only addresses known to belong to
this [`Endpoint`] already.
```
maybe.

Maybe the whole pruning should be described a bit as well, e.g.:
```
It is common that nodes might have the same direct addresses which are
only valid in their local network, while the nodes are not in the same
network.  Because of this direct addresses known to belonging to this
node will be silently removed from the addresses added.
```

nit: I don't think we should use a style like "ours" in the docs, though I often find myself writing like that as well.

iroh-net/src/magicsock/node_map.rs
```diff
@@ -333,6 +362,34 @@ impl NodeMapInner {
         }
     }
 
+    /// Prunes nodes that claim to share a address we know points to us,
```
flub, line 365:
```suggestion
    /// Prunes direct addresses from nodes that claim to share an address we know points to us.
```

iroh-net/src/magicsock/node_map.rs
```diff
@@ -333,6 +362,34 @@ impl NodeMapInner {
...
+    pub(super) fn on_direct_addr_discovered(
+        &mut self,
+        discovered: impl Iterator<Item = impl Into<IpPort>>,
+    ) {
+        for addr in discovered {
+            self.remove_by_ipp(addr.into(), ClearReason::MatchesOurLocalAddr)
+        }
+    }
+
+    /// Removes a node by its IpPort
```
flub, line 375:
```suggestion
    /// Removes a direct address from a node.
```

iroh-net/src/magicsock/node_map/node_state.rs
```diff
@@ -309,7 +311,24 @@ impl NodeState {
         (best_addr, relay_url)
     }
 
-    /// Fixup best_addr from candidates.
+    /// Removes a direct address for this node.
+    ///
+    /// If this is also de best address, it will be cleared as well.
```
flub, line 316:
```suggestion
    /// If this is also the best address, it will be cleared as well.
```

flub (approved):
Looks good!  I have mostly minor nits, now that I care more about public api docs :)

iroh-net/src/endpoint.rs
```diff
@@ -535,17 +535,40 @@ impl Endpoint {
...
     pub fn add_node_addr(&self, node_addr: NodeAddr) -> Result<()> {
+        self.add_node_addr_inner(node_addr, magicsock::Source::App)
+    }
+
+    /// Informs this [`Endpoint`] about addresses of the iroh-net node, noting the source.
+    ///
+    /// This updates the local state for the remote node.  If the provided [`NodeAddr`] contains a
+    /// [`RelayUrl`] this will be used as the new relay server for this node.  If it contains any
+    /// new IP endpoints they will also be stored and tried when next connecting to this node. The
+    /// source is used for logging exclusively and will not be stored.
```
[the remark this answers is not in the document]
divagant-martian: done in c3c95c1d

iroh-net/src/endpoint.rs
```diff
@@ -535,17 +535,40 @@ impl Endpoint {
     ///
     /// # Errors
     ///
-    /// Will return an error if we attempt to add our own [`PublicKey`] to the node map.
+    /// Will return an error if we attempt to add our own [`PublicKey`] to the node map or if the
+    /// direct addresses are a subset of ours.
```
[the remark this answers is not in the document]
divagant-martian: done in c3c95c1d

iroh-net/src/magicsock.rs
```diff
@@ -367,8 +368,37 @@ impl MagicSock {
 
     /// Add addresses for a node to the magic socket's addresbook.
     #[instrument(skip_all, fields(me = %self.me))]
-    pub fn add_node_addr(&self, addr: NodeAddr) {
-        self.node_map.add_node_addr(addr);
+    pub fn add_node_addr(&self, mut addr: NodeAddr, source: node_map::Source) -> Result<()> {
+        let my_addresses = self.endpoints.get().last_endpoints;
```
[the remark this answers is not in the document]
divagant-martian: I left it out. It feels out of scope since I'm not changing the field, simply using it here

iroh-net/src/magicsock.rs
```diff
@@ -367,8 +368,37 @@ impl MagicSock {
...
+            if addr.info.direct_addresses.remove(&my_addr) {
+                warn!(node_id=addr.node_id.fmt_short(), %my_addr, %source, "not adding our addr for node");
+                pruned += 1;
+            }
+        }
+        if !addr.info.is_empty() {
+            self.node_map.add_node_addr(addr, source);
+            Ok(())
+        } else if pruned != 0 {
+            anyhow::bail!("empty addressing info, {pruned} direct addresses have been pruned")
```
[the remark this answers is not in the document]
divagant-martian: done in c3c95c1d

iroh-net/src/magicsock.rs
```diff
@@ -81,6 +81,7 @@ pub use self::node_map::{
     ConnectionType, ConnectionTypeStream, ControlMsg, DirectAddrInfo, NodeInfo as ConnectionInfo,
 };
 pub(super) use self::timer::Timer;
+pub(super) use node_map::Source;
```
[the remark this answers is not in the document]
divagant-martian: done in c3c95c1d

iroh-net/src/endpoint.rs
divagant-martian: done in c3c95c1d

I'll avoid the ours in the future. In the meantime this is simply following the style of docs around the place

iroh-net/src/magicsock/node_map.rs
divagant-martian: done in c3c95c1d

iroh-net/src/magicsock/node_map/node_state.rs
divagant-martian: done in c3c95c1d

iroh-net/src/magicsock/node_map.rs
divagant-martian: done in c3c95c1d

PR #2338, opened 2024-06-03
refactor(iroh-blobs)!: make TempTag non-Clone

rklaehn:
## Description

refactor(iroh-blobs): Make TempTag non-Clone

I found that in most places it is not needed, and if somebody needs them to be clone they can always wrap them in an Arc.

Also, this complicates extending the concept of temp tags via the rpc boundary.

With this change if will be possible to use the same TempTag type both in the low level blobs store API and in the higher level blobs API of iroh.

## Breaking Changes

Changes the signature of TempTag::new to take a `Weak<dyn TagDrop>` instead
of an Arc<dyn LivenessTracker>. This will affect projects that write their own store
impl, so none that I am aware of. Nevertheless it is a breaking change.

The high level rpc API is not affected, since it does not know temp tags (yet).

## Notes & open questions

Note: this is just part 1/x of changes needed to extend temp tags to the rpc api. On its own it does not provide much value.

Note2: you could be even more radical and give a TempTag a lifetime and a reference to its store. But I am pretty sure that this would not play well with creating bindings and might even be inconvenient from pure rust.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh-blobs/src/store.rs
```diff
@@ -66,7 +66,9 @@ impl TempCounterMap {
 
     fn dec(&mut self, value: &HashAndFormat) {
         let HashAndFormat { hash, format } = value;
-        let counters = self.0.get_mut(hash).unwrap();
+        let Some(counters) = self.0.get_mut(hash) else {
+            return;
```
dignifiedquire, line 70: when would this happen?

iroh-blobs/src/store/mem.rs
```diff
@@ -43,20 +43,22 @@ pub struct Store {
...
 struct StoreInner(RwLock<StateInner>);
 
-impl LivenessTracker for StoreInner {
-    fn on_clone(&self, inner: &HashAndFormat) {
-        tracing::trace!("temp tagging: {:?}", inner);
-        let mut state = self.0.write().unwrap();
-        state.temp.inc(inner);
-    }
-
+impl TagDrop for StoreInner {
```
dignifiedquire, line 46: why did you make two traits, is there a use case for only implementing one of the two?

iroh-blobs/src/util.rs
```diff
@@ -187,11 +199,21 @@ pub enum SetTagOption {
...
-    fn on_clone(&self, inner: &HashAndFormat);
-    /// Called on drop
-    fn on_drop(&self, inner: &HashAndFormat);
+pub trait TagCounter: TagDrop + Sized {
+    /// Called on creation of a temp tag
+    fn on_create(&self, inner: &HashAndFormat);
+
+    /// Get this as a weak reference for use in temp tags
+    fn as_weak(self: &Arc<Self>) -> Weak<dyn TagDrop> {
+        let on_drop: Arc<dyn TagDrop> = self.clone();
```
dignifiedquire, line 208: no need to clone? https://doc.rust-lang.org/std/sync/struct.Arc.html#method.downgrade takes a reference

iroh-blobs/src/store/mem.rs
rklaehn: The only trait that most people will interact with is TagDrop. The other one is just for store implementers.

I don't want to give the ability to increase the reference count to temp tags. In case of the rpc there isn't even message for that...

iroh-blobs/src/store.rs
rklaehn: If you decrement something that has not been incremented before. It should never happen, but if it happens it's better to just ignore it than to panic. Given that messages that trigger a dec go via the rpc boundary I don't think it is safe to write unreachable here.

Maybe a warning?

dignifiedquire: given that we don't expect this happening, putting in a warning would be good, so we know if it starts happening

dignifiedquire (approved):

PR #2339, opened 2024-06-03
feat(iroh-bytes): Batch blob api

rklaehn:
## Description

This adds a new API for creating blobs.

You create a batch. Within a batch you got all the usual operations to add stuff, like add_bytes, add_stream, add_from_path etc. Notable differences to the existing api:

- All operations work on individual blobs, so no way to add entire subdirectories with add_from_path
- All operations return a TempTag instead of having a set tag option

The way to use the API is to just perform a complex operation within a batch, and then at the end assign a (non temporary) tag to the root(s) of the created data before dropping the batch.

It is possible to scan a directory and create a collection purely on the client side, so the code to traverse a directory can be removed from the node.

To allow the workflow described above, the tags client has been extended to allow manually setting a tag.

Ideally this API would entirely replace the current blobs API, so all read ops would always happen within the context of a batch.

## Breaking Changes

At this point mostly adding stuff, but changing the RPC api for setting tags. I might decide to remove the entire non batch mutation API in this PR.

How it works is to leave a streaming RPC call open for each batch, then do operations in the context of an unique identifier for this RPC call.

## Notes & open questions

Note: if things work out every add operation refers to a single blob, and the aggregation of many blobs can be driven from the client. That means that a lot of the complexity of the progress events like ids etc. can be removed from the rpc. This still needs to exist, but can be confined to the client.

## Todo

- [ ] ~~Add back fine grained progress~~
  fine grained progress is not needed for all ops, but definitely for add_file and add_dir. Possibly for add_bytes and add_reader. Not sure if it is OK to have it in all cases despite it typically not being used.
- [ ] ~~Purge all tag setting stuff from the blobs API and the downloader.~~

I would propose that we merge this initially as an addition, and do the stripdown of the other APIs in a subsequent PR. Also, if this is an addition we can do the fine grained progress for add_dir in a subsequent PR as well.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


rklaehn:
Here is the preliminary API for batches:

```rust
pub async fn add_bytes(&self, bytes: impl Into<Bytes>, format: BlobFormat) -> Result<TempTag> {
pub async fn add_file(&self, path: PathBuf, import_mode: ImportMode, format: BlobFormat) -> Result<(TempTag, u64)> {
pub async fn add_dir(&self, root: PathBuf, import_mode: ImportMode, wrap: WrapOption) -> Result<TempTag> {
pub async fn add_collection(&self, collection: Collection) -> Result<TempTag> {
pub async fn add_stream(&self, mut input: impl Stream<Item = io::Result<Bytes>> + Send + Unpin + 'static, format: BlobFormat) -> Result<TempTag> {
pub async fn add_blob_seq(&self, iter: impl Iterator<Item = Bytes>) -> Result<TempTag> {
pub async fn temp_tag(&self, content: HashAndFormat) -> Result<TempTag> {
```

Basically very similar to the normal blobs api, but there are no options to create tags. Instead every fn returns a temp tag for the thing that has been created, that the user can then later assign to a permanent tag (or not).

The tags API has been extended to allow creating a tag given a hash and format.

dignifiedquire:
should there be the equivalent `delete` versions as well`

rklaehn:
> should there be the equivalent `delete` versions as well`

WDYM? You delete stuff by ensuring that it is no longer tagged in some way, then GC will take care of it. There is blob delete, but that is really a low level function that you should rarely use directly.

iroh/src/client/blobs.rs
```diff
@@ -55,6 +65,18 @@ impl<C> Client<C>
 where
     C: ServiceConnection<RpcService>,
 {
+    /// Create a new batch for adding data.
+    pub async fn batch(&self) -> Result<Batch<C>> {
```
dignifiedquire, line 69: bikeshed: should this be called `transaction`?

iroh/src/client/blobs.rs
```diff
@@ -368,6 +390,263 @@ where
     }
 }
 
+/// A scope in which blobs can be added.
```
dignifiedquire, line 393: with this much new code, lets put this into a new file

iroh/src/client/blobs.rs
```diff
@@ -368,6 +390,263 @@ where
...
+        let input = futures_lite::stream::once(Ok(bytes.into()));
+        self.add_stream(input, format).await
+    }
+
+    /// Import a blob from a filesystem path.
+    ///
+    /// `path` should be an absolute path valid for the file system on which
+    /// the node runs, which refers to a file.
+    ///
+    /// If `import_mode` is TryReference, Iroh will assume that the data will not
```
dignifiedquire, line 433: maybe write "If you set`ImporteMode::TryReference` ... ", makes it a bit easier (for me at least) to understand that you mean the enum variant

iroh/src/client/blobs.rs
```diff
@@ -368,6 +390,263 @@ where
...
+}
+
+/// A batch for write operations.
+///
+/// This serves mostly as a scope for temporary tags.
+///
+/// It is not a transaction, so things in a batch are not atomic. Also, there is
+/// no isolation between batches.
+#[derive(derive_more::Debug)]
+pub struct Batch<C: ServiceConnection<RpcService>>(Arc<BatchInner<C>>);
```
dignifiedquire, line 412: Is there no `commit` step or at least `close` step necessary?

iroh/src/client/blobs.rs
dignifiedquire: nvm, found the notes about not being a transaction down below

rklaehn: Yeah, it is not a transaction.

iroh/src/client/blobs.rs
rklaehn: No. A batch is not a transaction. So as long as the batch lives, the temp tags are alive. Closing is just on drop. Note that this is a bit sloppy, so in some cases like network error it will take some time for the store to notice that a batch is gone. But that is fine in the context of temp tags, it just means that the data will hang around a bit longer before being purged by the next gc.

dignifiedquire: I think it would be good to add this to the comments in the code

rklaehn: There is a comment on the Batch struct. Maybe move it to the fn batch()?

iroh-blobs/src/store/traits.rs
```diff
@@ -700,7 +703,7 @@ pub enum ImportProgress {
 /// does not make any sense. E.g. an in memory implementation will always have
 /// to copy the file into memory. Also, a disk based implementation might choose
 /// to copy small files even if the mode is `Reference`.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, Serialize, Deserialize)]
 pub enum ImportMode {
```
rklaehn, line 707: I reexport this from the blobs api since I don't want to newtype it...

iroh/src/client/blobs.rs
```diff
@@ -55,6 +61,38 @@ impl<C> Client<C>
...
+    /// the node.
+    pub async fn status(&self, hash: Hash) -> Result<BlobStatus> {
+        let status = self.rpc.rpc(BlobStatusRequest { hash }).await??;
+        Ok(status.0)
+    }
+
+    /// Check if a blob is completely stored on the node.
+    ///
+    /// This is just a convenience wrapper around `status` that returns a boolean.
+    pub async fn has(&self, hash: Hash) -> Result<bool> {
```
rklaehn, line 76: not sure about this, but it sure is convenient...

iroh/src/client/blobs.rs
```diff
@@ -55,6 +61,38 @@ impl<C> Client<C>
...
+    /// Create a new batch for adding data.
+    ///
+    /// A batch is a context in which temp tags are created and data is added to the node. Temp tags
+    /// are automatically deleted when the batch is dropped, leading to the data being garbage collected
+    /// unless a permanent tag is created for it.
+    pub async fn batch(&self) -> Result<Batch<C>> {
+        let (updates, mut stream) = self.rpc.bidi(BatchCreateRequest).await?;
+        let BatchCreateResponse::Id(batch) = stream.next().await.context("expected scope id")??;
+        let rpc = self.rpc.clone();
+        Ok(Batch::new(batch, rpc, updates, 1024))
```
rklaehn, line 93: if the 1024 does not work for everybody, we can do batch_with_opts where this can be configured...

iroh/src/rpc_protocol.rs
```diff
@@ -177,6 +208,41 @@ impl ServerStreamingMsg<RpcService> for BlobValidateRequest {
     type Response = ValidateProgress;
 }
 
+/// Get the status of a blob
```
rklaehn, line 211: this is mostly unrelated to the PR. Just a way to check if you have a thing without downloading it all via the rpc.

iroh-blobs/src/store/traits.rs
```diff
@@ -356,6 +356,9 @@ pub trait Store: ReadableStore + MapMut {
     /// Create a temporary pin for this store
     fn temp_tag(&self, value: HashAndFormat) -> TempTag;
 
+    /// Handle to use to drop tags
+    fn tag_drop(&self) -> Option<&dyn TagDrop>;
```
dignifiedquire, line 360: maybe document what it means if this returns `None`

iroh/src/client/blobs.rs
dignifiedquire: love it

iroh/src/node/rpc.rs
```diff
@@ -842,6 +937,120 @@ impl<D: BaoStore> Handler<D> {
...
+                            &content,
+                            self.inner.db.tag_drop(),
+                        );
+                    }
+                    BatchUpdate::Ping => {}
+                }
+            }
+            self.inner
+                .blob_batches
+                .lock()
```
dignifiedquire, line 963: if this is mostly used in an async context, should be an async lockd

iroh/src/rpc_protocol.rs
```diff
@@ -177,6 +208,41 @@ impl ServerStreamingMsg<RpcService> for BlobValidateRequest {
...
+    pub hash: Hash,
+}
+
+/// The response to a status request
+#[derive(Debug, Serialize, Deserialize, derive_more::From, derive_more::Into)]
+pub struct BlobStatusResponse(pub BlobStatus);
+
+/// Status information about a blob.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum BlobStatus {
```
dignifiedquire, line 224: lets put this into the `blobs.rs` file, given this is part of the public api

iroh/src/node/rpc.rs
rklaehn: Using a sync lock in an async context is fine provided you don't lock over an await point and the lock is held for a very short time. Both is the case here.

dignifiedquire: I know, but in the context of the RPC, this is still risky, because instead of yielding, you block the thread, and there is really no benefit of using a standard mutex in an async context unless you are in a super hot path

rklaehn: I don't see the issue.

Most operations (create and store) are extremely cheap and constant time. remove_one is also cheap, it just decrements a counter in a map. The only op that is even somewhat expensive is remove, and that just decrements a bunch of counters.

So if for some reason you have an absolutely giant number of temp tags in a batch, and then you call remove for that batch, the lock will be taken during that time. And if you then do any op that also relates to batches, you will block an executor for a millisecond.

But ok, in this case it is not super painful to use the async lock, so I will give it a try.

iroh/src/rpc_protocol.rs
rklaehn: My idea was to put it into the rpc protocol and re-export in the blobs.rs, but this also works. I guess it is better when people click on source when browsing docs.rs.

iroh/src/client/blobs.rs
```diff
@@ -360,18 +367,37 @@ where
...
+        Ok(Gen::new(|co| async move {
+            if let Err(cause) = this.list_collections_impl(&co).await {
+                co.yield_(Err(cause)).await;
+            }
+        }))
+    }
+
+    async fn list_collections_impl(&self, co: &Co<Result<CollectionInfo>>) -> Result<()> {
+        let tags = self.tags_client();
+        let mut tags = tags.list().await?;
```
dignifiedquire, line 385: I wonder if we want to add a `tags.list(BlobFormat)`, given with a large db this would be a lot of queries

rklaehn: We could, but you should not have that many tags in the first place. All the "short lived" tag use should move to temp tags.

In any case I moved this stuff in a separate PR

dignifiedquire:
@rklaehn what's the state of this?

rklaehn:
> @rklaehn what's the state of this?

Just merged with main. I want to do another self-review, but currently trying to keep all the stuff up to date with main so it does not bitrot...

iroh/src/client/blobs.rs
```diff
@@ -887,11 +899,30 @@ pub enum DownloadMode {
...
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum BlobStatus {
+    /// The blob is not stored on the node.
+    NotFound,
+    /// The blob is only stored partially.
+    Partial {
+        /// The size of the currently stored partial blob.
+        ///
+        /// This can be either a verified size if the last chunk was received,
+        /// or an unverified size if the last chunk was not yet received.
```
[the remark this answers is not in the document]
rklaehn: Yes, for example if you get *just* the large chunk, the blob status will be partial because everything else is missing. But it will be verified.

dignifiedquire:
closing in favor of #2545

PR #2340, opened 2024-06-03
feat(iroh-gossip): configure the max message size

### LINKED ISSUE #2312: feat: make iroh-gossip message size configurable
There is a TODO here:
https://github.com/n0-computer/iroh/blob/370075c6d5689ad4349664eb9b51ec0a5e7e4e81/iroh-gossip/src/net.rs#L30-L31

I ran into this problem, after trying to send a large message over gossip it stops working (in Delta Chat). Maybe it is our error handling that is wrong, have not investigated further yet.

I wanted to send PCM audio, in 4096 bytes with a single channel of 32-bit floats at 8kHz I can only encode 128ms of audio (8 * 4096 / 32 /  8000) per message. Will split into chunks of 3k bytes now just to make sure, but would be nice to have less overhead and just avoid the problem of the channel silently failing because the message is too large.

ramfox:
## Description

Add configuration option for `max_message_size` for `iroh-gossip::proto::Config`.

This `Config` gets used in `iroh-gossip::Gossip::from_endpoint`.

`iroh-docs` still uses the default 4096 bytes. The `max_message_size` configuration is useful for folks using `iroh-gossip::Gossip` as its own library.

closes #2312

## Breaking Changes
Adds:
  `iroh-gossip::Gossip::max_message_size` - that reports the configured maximum message size for the gossip actor.

Changes:
  `iroh_gossip::net::util::read_message` now takes a `max_message_size: usize` parameter
  `iroh_gossip::net::util::write_message` now takes a `max_message_size: usize` parameter
  `iroh_gossip::net::util::read_lp` now takes a `max_message_size: usize` parameter

Removes:
  `iroh-gossip::proto:: MAX_MESSAGE_SIZE` const

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh-gossip/src/proto/state.rs
+++ b/iroh-gossip/src/proto/state.rs
@@ -196,7 +196,7 @@ impl<PI: PeerIdentity, R: Rng + Clone> State<PI, R> {
             .unwrap_or(false)
     }
 
-    /// Returns the maximum message size configured in the gossip protocoli.
+    /// Returns the maximum message size configured in the gossip protocol.
     pub fn max_message_size(&self) -> usize {
         self.config.max_message_size
     }
--- a/iroh-gossip/src/proto/state.rs
+++ b/iroh-gossip/src/proto/state.rs
@@ -196,6 +196,11 @@ impl<PI: PeerIdentity, R: Rng + Clone> State<PI, R> {
             .unwrap_or(false)
     }
 
+    /// Returns the maximum message size configured in the gossip protocoli.
+    pub fn max_message_size(&self) -> usize {
+        self.config.max_message_size
+    }
+
     /// Handle an [`InEvent`]
     ///
     /// This returns an iterator of [`OutEvent`]s that must be processed.
--- a/iroh-gossip/src/proto/topic.rs
+++ b/iroh-gossip/src/proto/topic.rs
@@ -171,12 +171,26 @@ impl<PI: Clone> IO<PI> for VecDeque<OutEvent<PI>> {
     }
 }
 /// Protocol configuration
-#[derive(Clone, Default, Debug)]
+#[derive(Clone, Debug)]
 pub struct Config {
     /// Configuration for the swarm membership layer
     pub membership: hyparview::Config,
     /// Configuration for the gossip broadcast layer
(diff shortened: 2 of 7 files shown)


iroh-gossip/src/proto/state.rs
```diff
@@ -196,6 +196,11 @@ impl<PI: PeerIdentity, R: Rng + Clone> State<PI, R> {
             .unwrap_or(false)
     }
 
+    /// Returns the maximum message size configured in the gossip protocoli.
```
divagant-martian, line 199:
```suggestion
    /// Returns the maximum message size configured in the gossip protocol.
```
:broccoli: :broccoli: :broccoli:

iroh-gossip/src/proto/topic.rs
```diff
@@ -171,12 +171,26 @@ impl<PI: Clone> IO<PI> for VecDeque<OutEvent<PI>> {
...
 }
 /// Protocol configuration
-#[derive(Clone, Default, Debug)]
+#[derive(Clone, Debug)]
 pub struct Config {
     /// Configuration for the swarm membership layer
     pub membership: hyparview::Config,
     /// Configuration for the gossip broadcast layer
     pub broadcast: plumtree::Config,
+    /// Max message size in bytes
```
divagant-martian, line 180:
```suggestion
    /// Max message size in bytes.
```

just to make Floris happy

iroh-gossip/src/proto/topic.rs
```diff
@@ -171,12 +171,26 @@ impl<PI: Clone> IO<PI> for VecDeque<OutEvent<PI>> {
...
+    ///
+    /// Default is 4096 bytes.
+    pub max_message_size: usize,
```
divagant-martian, line 183: I'd suggest moving this to a constant so that if we ever want to change the default we don't need to remember to update the docs

divagant-martian (commented):
LGTM, maybe add some comments about a couple of caveats.

For example:
- This should be a network-wide configuration parameter to ensure all nodes can transmit and read large messages.
- This should be large enough to allow for gossip control messages to be sent.
- Any other we can think of

Frando (approved):

Issue #2179, opened 2024-04-12
iroh-relay: run behind reverse proxy


link2xt:
Is it possible to run iroh-relay behind reverse proxy such as nginx?

I am currently getting certificates using `acmetool` and run iroh-relay with manual certificate mode:
https://github.com/n0-computer/iroh/blob/b07547b68eb771e789474ad4f1344e02b2223f95/iroh-net/src/bin/iroh-relay.rs#L209

But to give iroh-relay access to the key I have to run it as root. To avoid this I would like to run iroh-relay as HTTP and STUN node and have nginx which has access to TLS key proxy all requests to iroh-relay. For STUN server it is also possible by defining a `stream` section in nginx config: https://github.com/coturn/coturn/issues/702

flub:
Why can't you make the certificates readable by the user that runs the relay?

dignifiedquire:
@link2xt what's the difference here between this issue and #2177 ?

flub:
Fixed by #2341.  If not feel free to reopen

Issue #2178, opened 2024-04-11
iroh-relay: make it possible to disable captive portal


link2xt:
There is an option `captive_portal_port` but in my case HTTP port is already taken by `acmetool redirector` service:
 https://github.com/n0-computer/iroh/blob/b07547b68eb771e789474ad4f1344e02b2223f95/iroh-net/src/bin/iroh-relay.rs#L228

Having captive portal on any port other than 80 does not really make sense if I understand correctly its purpose (maybe to put it behind reverse proxy, but then I can configure nginx or whatever reverse proxy to respond however I want myself), so would be better if there was an option to just disable it. As a workaround I have moved it to an arbitrary port.

flub:
Fixed by https://github.com/n0-computer/iroh/pull/2341. If not feel free to reopen

PR #2341, opened 2024-06-04
refactor(iroh-net)!: Rework relay-server binary, more configurable, reverse-proxy support

### LINKED ISSUE #2177: iroh-relay expects a particular setup of certificate directory
I am trying to setup iroh-relay with this configuration:
```
addr = "[::]:443"
stun_port = 3478
hostname = "iroh.example.org"
enable_stun = true
enable_relay = true

[tls]
cert_mode = "Manual"
prod_tls = true
contact = "root@example.org"
cert_dir = "/var/lib/acme/live/iroh.example.org/"
```

The server is using [acmetool](
https://github.com/hlandau/acmetool) which creates directory with the following files:
```
$ ls /var/lib/acme/live/iroh.example.org/
account
cert
chain
fullchain
privkey
url
```

`iroh` expects `.key` and `.crt` files instead:
https://github.com/n0-computer/iroh/blob/b07547b68eb771e789474ad4f1344e02b2223f95/iroh-net/src/bin/iroh-relay.rs#L108-L109

I can create a folder with symlinks that links `iroh.example.org.crt` to `fullchain` file and `iroh.example.org.key` to `privkey` file, but it would have been easier if instead of `cert_dir` configuration accepted two explicit paths for the key and for certificate chain.

flub:
## Description

This re-architects the relay-server binary.  There is now a struct with detailed configuration which runs the entire server and aborts the server on drop.  This simplifies running the server in various situations, including tests.  The configuration is now done using a declarative struct, which supports more control over how it runs so it can be more easily used behind a reverse proxy, without TLS etc.

This is aiming to fix #2177, #2179 and #2178.

## Breaking Changes

The configuration file format has changed, deployments will need to updated.  For the full format see `struct Config` in `iroh-net/src/bin/iroh-relay.rs`.  Here a summary:

- The 3 parts of the server now have an independent enable setting: `enable_relay`, `enable_stun` and `enable_metrics`.  If omitted they default to `true`.
- The way to specify which addresses the server listens on has changed: `http_bind_addr` is for the relay server, `stun_bind_addr` for the STUN server, `metrics_bind_addr` is for the optional metrics server and `tls.https_bind_addr` is for when TLS is enabled.  Note these are now all full socket addresses.  All have sensible defaults if omitted.
- There are new options in `tls.cert_path` and `tls.key_path` which allow more control over where the manual TLS keys are to be read from.
- `iroh_net::defaults::DEFAULT_RELAY_STUN_PORT` has been renamed to `iroh_net::defaults::DEFAULT_STUN_PORT`.

TBD: some APIs changed as well.  Why are they not all private?

## Notes & open questions

* The `iroh_net::relay::iroh_relay` crate name is a bit weird.  But `iroh_net::relay::server` is already taken.  Maybe `iroh_net::relay::bin` could work, but that would be weird when using it from code in other places.
* The `ServerConfig` struct is a declarative way of controlling the new server interface.  It's kind of nice to use.  Bu it is a public API that will be a breaking change every time it changes, and it will change.  Maybe it's worth creating a builder API for this.  But maybe that's something to only tackle when it is a real demand.  I feel like the `iroh_net::relay::server` builders are an attempt at doing this earlier than needed.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh-net/src/relay/iroh_relay.rs
```diff
@@ -0,0 +1,826 @@
...
+use crate::relay;
+use crate::relay::http::{ServerBuilder as DerpServerBuilder, TlsAcceptor};
+use crate::stun;
+use crate::util::AbortingJoinHandle;
+
+// Module defined in this file.
+use metrics::StunMetrics;
+
+const NO_CONTENT_CHALLENGE_HEADER: &str = "X-Tailscale-Challenge";
+const NO_CONTENT_RESPONSE_HEADER: &str = "X-Tailscale-Response";
```
dignifiedquire, line 32: note for the future, we should really update these and call them `X-Iroh`

iroh-net/src/relay/iroh_relay.rs
```diff
@@ -0,0 +1,826 @@
...
+const NOTFOUND: &[u8] = b"Not Found";
+const RELAY_DISABLED: &[u8] = b"derp server disabled";
+const ROBOTS_TXT: &[u8] = b"User-agent: *\nDisallow: /\n";
+const INDEX: &[u8] = br#"<html><body>
+<h1>DERP</h1>
+<p>
+  This is an
+  <a href="https://iroh.computer/">Iroh</a> DERP
```
dignifiedquire, line 40: we should update this to `Relay`

iroh-net/src/relay/iroh_relay.rs
flub: https://github.com/n0-computer/iroh/issues/2344

flub:
@Arqu probably time for you to give some feedback on the config file changes I'm proposing.  Check src/bin/iroh-relay.rs' `Config` which should have sufficient docs, the changes are highlighted in the pr description

Arqu:
> @Arqu probably time for you to give some feedback on the config file changes I'm proposing. Check src/bin/iroh-relay.rs' `Config` which should have sufficient docs, the changes are highlighted in the pr description

Just went over this, the changes are nice and I'm fine with this moving forward. Once this lands I'll update the configuration files for the deployments.

iroh-net/src/bin/iroh-relay.rs
```diff
@@ -164,96 +79,210 @@ fn load_secret_key(filename: impl AsRef<Path>) -> Result<rustls::PrivateKey> {
...
-    contact: String,
-    /// Directory to store LetsEncrypt certs or read certificates from, if TLS is used.
-    cert_dir: Option<PathBuf>,
-    /// The port on which to serve a response for the captive portal probe over HTTP.
     ///
-    /// The listener is bound to the same IP as specified in the `addr` field. Defaults to 80.
-    /// This field is only read in we are serving the relay server over HTTPS. In that case, we must listen for requests for the `/generate_204` over a non-TLS connection.
-    captive_portal_port: Option<u16>,
+    /// Used when `cert_mode` is `LetsEncrypt`.
+    contact: String,
```
dignifiedquire, line 252: - maybe mention this is an email? 
- should we make this optional?

iroh-net/src/bin/iroh-relay.rs
```diff
@@ -164,96 +79,210 @@ fn load_secret_key(filename: impl AsRef<Path>) -> Result<rustls::PrivateKey> {
...
+    fn cert_path(&self) -> PathBuf {
+        self.manual_cert_path
+            .clone()
+            .unwrap_or_else(|| self.cert_dir().join("default.crt"))
+    }
+
+    fn key_path(&self) -> PathBuf {
+        self.manual_key_path
+            .clone()
+            .unwrap_or_else(|| self.cert_dir().join("default.key"))
```
dignifiedquire, line 274: should these file names be constants?

iroh-net/src/bin/iroh-relay.rs
```diff
@@ -274,12 +303,12 @@ impl Config {
 
     async fn read_from_file(path: impl AsRef<Path>) -> Result<Self> {
         if !path.as_ref().is_file() {
-            bail!("config-path must be a valid toml file");
+            bail!("config-path must be a valid file");
```
dignifiedquire, line 306: remove the "valid"?

iroh-net/src/relay/http/server.rs
```diff
@@ -108,13 +146,15 @@ pub struct TlsConfig {
     pub acceptor: TlsAcceptor,
 }
 
-/// Build a Relay Server that communicates over HTTP or HTTPS, on a given address.
+/// Builder for the Relay HTTP Server.
 ///
-/// Defaults to handling relay requests on the "/derp" endpoint.
+/// Defaults to handling relay requests on the "/derp" endpoint.  Other HTTP endpoints can
```
dignifiedquire, line 151: we should transition this name as well, @flub can you make an issue to migrate this to sth like `/relay`?

iroh-net/src/relay/http/server.rs
```diff
@@ -320,13 +378,17 @@ impl ServerState {
                     }
                 }
             }
+            if let Some(server) = server {
+                // TODO: if the task this is running in is aborted this server is not shut
```
dignifiedquire, line 382: what about this todo?

iroh-net/src/relay/iroh_relay.rs
```diff
@@ -0,0 +1,909 @@
...
+use crate::stun;
+use crate::util::AbortingJoinHandle;
+
+// Module defined in this file.
+use metrics::StunMetrics;
+
+const NO_CONTENT_CHALLENGE_HEADER: &str = "X-Tailscale-Challenge";
+const NO_CONTENT_RESPONSE_HEADER: &str = "X-Tailscale-Response";
+const NOTFOUND: &[u8] = b"Not Found";
+const RELAY_DISABLED: &[u8] = b"derp server disabled";
```
dignifiedquire, line 40: lets change this to "relay server disabled"

iroh-net/src/relay/iroh_relay.rs
```diff
@@ -0,0 +1,909 @@
...
+    http_body_util::Full::new(hyper::body::Bytes::new())
+}
+
+/// Configuration for the full Relay & STUN server.
+///
+/// Be aware the generic parameters are for when using the Let's Encrypt TLS configuration.
+/// If not used dummy ones need to be provided, e.g. `ServerConfig::<(), ()>::default()`.
+#[derive(Debug, Default)]
+pub struct ServerConfig<EC: fmt::Debug, EA: fmt::Debug = EC> {
+    /// Configuration for the DERP server, disabled if `None`.
```
dignifiedquire, line 68: derp -> relay

iroh-net/src/relay/iroh_relay.rs
```diff
@@ -0,0 +1,909 @@
...
+    pub secret_key: SecretKey,
+    /// The socket address on which the Relay HTTP server should bind.
+    ///
+    /// Normally you'd choose port `80`.  The bind address for the HTTPS server is
+    /// configured in [`RelayConfig::tls`].
+    ///
+    /// If [`RelayConfig::tls`] is `None` then this serves all the HTTP services without
+    /// TLS.
+    pub http_bind_addr: SocketAddr,
+    /// TLS configuration for the HTTPS servier.
```
dignifiedquire, line 93: typo

iroh-net/src/relay/iroh_relay.rs
```diff
@@ -0,0 +1,909 @@
...
+                        let http_addr = http_listener.local_addr()?;
+                        tasks.spawn(
+                            run_captive_portal_service(http_listener)
+                                .instrument(info_span!("http-service", addr = %http_addr)),
+                        );
+                        Some(http_addr)
+                    }
+                    None => {
+                        // If running DERP without TLS add the plain HTTP server directly to
+                        // the DERP server.
```
dignifiedquire, line 297: derp -> relay

dignifiedquire (approved):
some small notes, otherwise lgtm

iroh-net/src/bin/iroh-relay.rs
flub: It does say it is an email.  Yes, making it Optional is better.

iroh-net/src/bin/iroh-relay.rs
flub: They used to use the `hostname` field.  But it seemed weird to require the hostname for manual TLS, it was only used to pick up a default name here.  I think a constant default is fine, I honestly expect this to be customised most of the time.

iroh-net/src/relay/http/server.rs
flub: https://github.com/n0-computer/iroh/issues/2378

iroh-net/src/relay/http/server.rs
flub: I decided to leave that in because I felt like if I had to turn all of the relay server actors into structured concurrency the rabbit hole was too deep.  This is basically the status quo so not a regression.  It would be nice to do this, but as a separate PR probably.

Plus, would need to decide on how much I like tokio-graceful-shutdown possibly :)

PR #2343, opened 2024-06-04
refactor: move docs engine into iroh-docs

Frando:
## Description

Moves `iroh::docs_engine` to `iroh_docs::engine`. Most changes are changed imports, dependencies, and visibility changes (plus docs where they are now required). Everything is moved over, only the RPC implementation stays in iroh for now.

The `iroh_docs::engine::Engine` is wrapped in a newtype `DocsEngine` in iroh that derefs to the `Engine` so that we can implement the RPC methods directly.

## Breaking Changes

* `iroh::client::docs::{Origin, SyncReason, SyncEvent}` are now reexports of `iroh_docs::engine::{Origin, SyncReason, SyncEvent}`. There is no need to change anything in your code that uses these structs via `iroh::client::docs`.

## Change checklist

- [x] Self-review.
- ~~[ ] Tests if relevant.~~


iroh/src/node/docs.rs
```diff
@@ -0,0 +1,126 @@
+// use std::{path::PathBuf, sync::RwLock};
```
dignifiedquire, line 1: ?

Frando: removed

dignifiedquire (approved):

Issue #2342, opened 2024-06-04
Add support for DNS resolution on non-port 53 to DnsDiscovery


zh522130:
I do have this need because my port 53 is required for other DNS services. Currently, I have modified the code locally and validated it. If others have a similar need, I can submit a PR.

zh522130:
This is not an issue; I just want to confirm whether I should submit this PR.

dignifiedquire:
which part of the code are you talking about, `iroh-dns-server`?

zh522130:
> which part of the code are you talking about, `iroh-dns-server`?

yes

dignifiedquire:
sure, a PR would be appreciated

zh522130:
There was an issue with my previous description. The actual situation is that the `iroh-dns-server` can run on non-port 53, but `DnsDiscovery` does not support resolution on non-port 53.

flub:
I think this is already possible with `iroh_net::endpoint::Builder::dns_resolver`: it allows you to construct your own resolver client that the endpoint will use, including the DnsDiscovery feature.   You'd use something like `hickory_resolver::config::ResolverConfig::add_name_server` to set custom connection details for your DNS server.

zh522130:
> I think this is already possible with `iroh_net::endpoint::Builder::dns_resolver`: it allows you to construct your own resolver client that the endpoint will use, including the DnsDiscovery feature. You'd use something like `hickory_resolver::config::ResolverConfig::add_name_server` to set custom connection details for your DNS server.

I have previously tested the `dns_resolver,` and the main issue is that the relay also uses `dns_resolver,` which cannot resolve the relay address. It's possible that I'm using it incorrectly, I will continue to research this. Once the tests pass, I'll close this issue.

zh522130:
@flub Thank you. I've implemented custom port resolution through `dns_resolver`. 

Should anyone need it, you can refer to the `create_default_resolver` function in `dns.rs`. You simply need to add a DNS service with a custom port using `config.add_name_server()`.

PR #2347, opened 2024-06-05
feat(iroh-net):Add custom port resolution to DnsDiscovery.

zh522130:
## Description
https://github.com/n0-computer/iroh/issues/2342

## Change checklist

- [x] Self-review.
-  ~~Tests if relevant.~~
-  ~~All breaking changes documented.~~

### DIFF
--- a/iroh-net/src/discovery/dns.rs
+++ b/iroh-net/src/discovery/dns.rs
@@ -1,11 +1,17 @@
 //! DNS node discovery for iroh-net
 
-use anyhow::Result;
+use std::{net::SocketAddr, time::Duration};
+
+use anyhow::{anyhow, Result};
 use futures_lite::stream::Boxed as BoxStream;
+use hickory_resolver::{
+    config::{NameServerConfig, Protocol, ResolverConfig},
+    AsyncResolver,
+};
 
 use crate::{
     discovery::{Discovery, DiscoveryItem},
-    dns::ResolverExt,
+    dns::{DnsResolver, ResolverExt},
     Endpoint, NodeId,
 };
 
@@ -36,12 +42,37 @@ const DNS_STAGGERING_MS: &[u64] = &[200, 300];
 #[derive(Debug)]
 pub struct DnsDiscovery {
     origin_domain: String,
+    port: Option<u16>,
+    use_ipv6: bool,
 }
 
 impl DnsDiscovery {
     /// Create a new DNS discovery.
     pub fn new(origin_domain: String) -> Self {
-        Self { origin_domain }
+        Self {
+            origin_domain,
+            port: None,
+            use_ipv6: false,
+        }
+    }
+
+    /// Create a new DNS discovery with a custom port
+    ///
+    /// This method allows specifying a custom port for DNS queries and whether to use IPv4 or IPv6.
+    ///
+    /// # Arguments
+    ///
+    /// * `origin_domain` - The domain name of the DNS server.
+    /// * `port` - The port the DNS server is bound to.
+    /// * `use_ipv6` - Use IPv6 for connecting to the DNS server.
+    ///
+    /// Note: If you need to support both IPv4 and IPv6, consider using ConcurrentDiscovery.
+    pub fn with_port(origin_domain: String, port: u16, use_ipv6: bool) -> Self {
+        Self {
+            origin_domain,
+            port: Some(port),
+            use_ipv6,
+        }
     }
 
     /// Create a new DNS discovery which uses the [`N0_DNS_NODE_ORIGIN`] origin domain.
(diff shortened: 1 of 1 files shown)


iroh-net/src/discovery/dns.rs
```diff
@@ -36,12 +42,41 @@ const DNS_STAGGERING_MS: &[u64] = &[200, 300];
...
     pub fn new(origin_domain: String) -> Self {
-        Self { origin_domain }
+        Self {
+            origin_domain,
+            port: None,
+            use_ipv6: false,
+        }
+    }
+
+    /// Create a new DNS discovery with a custom port and IP protocol version.
```
flub, line 59:
```suggestion
    /// Creates a new DNS discovery with a custom port and IP protocol version.
```

We should try and follow https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#appendix-a-full-conventions-text for doc comments (though we're not consistently there yet, so there are plenty of bad examples)

flub (commented):
IIUC this is doing something rather unusual: it assumes the origin domain itself resolves to the nameserver to be queried.  Normally this would be done by looking up the `NS` record type for the origin domain I think.

It then queries this nameserver directly for the actual record lookup that needs to be resolved.

The DnsDiscovery feature was designed with a normal DNS auhtoritative nameservers setup: since the nameserver is setup as the authoritative nameserver of your origin domain any DNS server which does not have records for your domain (i.e. any other server) will make a recursive query to your origin DNS server and then forward the result to the client.  So clients don't have to do this fragile dance themselves.  This is how DNS was designed to work and what makes it a resilient and reliable global system.

To be able to use a DNS server that is listening on a different port I think you would use `iroh_net::endpoint::Builder::dns_resolver` to customise the resolver to point to your port.  I'm not sure if anything else would need to be done.

iroh-net/src/discovery/dns.rs
```diff
@@ -36,12 +42,41 @@ const DNS_STAGGERING_MS: &[u64] = &[200, 300];
...
+    ///
+    /// * `origin_domain` - The domain name of the DNS server.
+    /// * `port` - The port number the DNS server is listening on.
+    /// * `use_ipv6` - If true, the query will use IPv6 to connect to the DNS server;
+    ///   if false, it will use IPv4.
+    ///
+    /// Note: To handle both IPv4 and IPv6 connections simultaneously, consider using
+    /// [`ConcurrentDiscovery`](crate::discovery::ConcurrentDiscovery) to combine the queries
+    /// from both versions.
+    pub fn with_port(origin_domain: String, port: u16, use_ipv6: bool) -> Self {
```
rklaehn, line 74: We have embraced a pattern like this:

fn foo(x) // simple function

fn foo_with_opts(x, opts) // more complex function with lots of settings

Maybe we can do the same here.

fn new(origin_domain) // this would call new_with_opts with the default opts.

fn new_with_opts(origin_domain, opts) // opts contains port, use_ipv6 and possibly more in the future

zh522130:
> IIUC this is doing something rather unusual: it assumes the origin domain itself resolves to the nameserver to be queried. Normally this would be done by looking up the `NS` record type for the origin domain I think.
> 
> It then queries this nameserver directly for the actual record lookup that needs to be resolved.
> 
> The DnsDiscovery feature was designed with a normal DNS auhtoritative nameservers setup: since the nameserver is setup as the authoritative nameserver of your origin domain any DNS server which does not have records for your domain (i.e. any other server) will make a recursive query to your origin DNS server and then forward the result to the client. So clients don't have to do this fragile dance themselves. This is how DNS was designed to work and what makes it a resilient and reliable global system.
> 

Through your professional answers, I have learned a lot, but I still need to further study the relevant knowledge. Perhaps I should close this PR for now until I fully understand it.

> To be able to use a DNS server that is listening on a different port I think you would use `iroh_net::endpoint::Builder::dns_resolver` to customise the resolver to point to your port. I'm not sure if anything else would need to be done.

Through your professional answers, I have learned a lot, yet I still need to further study the related knowledge. Perhaps I should close this PR before I have fully understood it, since this portion of the code has not been fully thought through

iroh-net/src/discovery/dns.rs
zh522130: Indeed, this is better.

iroh-net/src/discovery/dns.rs
zh522130: I'm indeed not good at documenting comments, and my English is not very good either; I will learn more about both.

zh522130: Thank you, I think your comments are very well written. I will carefully read the content of this link to improve myself

PR #2348, opened 2024-06-05
refactor(iroh)!: remove tags from downloader

rklaehn:
## Description

Remove tags from downloader. Adding temp or persistent tags is now the responsibility of the caller.

## Breaking Changes

- iroh_blobs::downloader::DownloadRequest
    - field `tag` is removed
    - fn `untagged` is removed
    - fn `tag` is removed

Docs are changed to indicate that tagging is now responibility of the caller...

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-blobs/src/downloader.rs
+++ b/iroh-blobs/src/downloader.rs
@@ -193,7 +193,12 @@ pub struct DownloadRequest {
 impl DownloadRequest {
     /// Create a new download request.
     ///
-    /// The blob will be auto-tagged after the download to prevent it from being garbage collected.
+    /// It is the responsibility of the caller to ensure that the data is tagged either with a
+    /// temp tag or with a persistent tag to make sure the data is not garbage collected during
+    /// the download.
+    ///
+    /// If this is not done, there download will proceed as normal, but there is no guarantee
+    /// that the data is still available when the download is complete.
     pub fn new(
         resource: impl Into<DownloadKind>,
         nodes: impl IntoIterator<Item = impl Into<NodeAddr>>,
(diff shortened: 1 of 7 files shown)


dignifiedquire:
definitely looks like a nice simplification, but @Frando will have to answer if this is missing something crucial

Frando:
Aha, interesting. I somehow did not realize that we can create tags before we have the data, I thought only temptags could do that. So I think this is fine. We should document this somewhere on the downloader though (it's public API) that users need to create tags themselves otherwise their newly downloaded blobs will be garbage collected at any time.

Frando:
One advantage of having it in the downloader is that if queuing the same downloader multiple times from the client API with SetTagOption::Auto you used to have only one auto tag, and now you'd have 2.

rklaehn:
> One advantage of having it in the downloader is that if queuing the same downloader multiple times from the client API with SetTagOption::Auto you used to have only one auto tag, and now you'd get multiple auto tag. I think this was the original reason I moved the tagging into the downloader and added the TagSet.

Not sure that warrants the additional complexity. also, handing the same auto tag to 2 places is a bit dangerous, given that I think currently we are using auto tags as a workaround for the fact that temp tags are not available on the rpc client.

E.g. you got one place that does download with auto, then some validation, then removes the auto tag and replaces it with something permanent/named. The other side will not notice this, so they will have an auto tag that is not actually tagging anything anymore.

In general I think most uses for SetTagOption::Auto will be replaced by temp tags once we have them in the rpc client.

dignifiedquire:
please document the breaking changes

dignifiedquire (approved):

PR #2349, opened 2024-06-06
refactor(iroh-blobs)!: implement some collection related things on the client side

rklaehn:
## Description

A collection is just one particular way to use a hashseq, so it feels a bit weird to have it baked in to the iroh node. With this we can move some of it into the client.

This is a part of https://github.com/n0-computer/iroh/pull/2272 . We can make more similar changes once we have the batch API https://github.com/n0-computer/iroh/pull/2339 .

## Notes & open questions

Note: I closed #2272 because half of the changes in that PR are here, the other half will be part of the batch PR, and moving collections into iroh I am not convinced of yet...

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/client/blobs.rs
+++ b/iroh/src/client/blobs.rs
@@ -13,10 +13,11 @@ use anyhow::{anyhow, Result};
 use bytes::Bytes;
 use futures_lite::{Stream, StreamExt};
 use futures_util::SinkExt;
+use genawaiter::sync::{Co, Gen};
 use iroh_base::{node_addr::AddrInfoOptions, ticket::BlobTicket};
 use iroh_blobs::{
     export::ExportProgress as BytesExportProgress,
-    format::collection::Collection,
+    format::collection::{Collection, SimpleStore},
     get::db::DownloadProgress as BytesDownloadProgress,
     store::{ConsistencyCheckProgress, ExportFormat, ExportMode, ValidateProgress},
     BlobFormat, Hash, Tag,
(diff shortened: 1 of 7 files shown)


iroh/src/client/blobs.rs
```diff
@@ -322,18 +322,37 @@ where
...
+        Ok(Gen::new(|co| async move {
+            if let Err(cause) = this.list_collections_impl(&co).await {
+                co.yield_(Err(cause)).await;
+            }
+        }))
+    }
+
+    async fn list_collections_impl(&self, co: &Co<Result<CollectionInfo>>) -> Result<()> {
+        let tags = self.tags_client();
+        let mut tags = tags.list().await?;
```
dignifiedquire, line 340: what about the filtering I mentioned on the other PR?

rklaehn: My opinion is that we don't need it since you should not have that many tags. But we can add it.

dignifiedquire: don't you end up with a lot of tags through documents?

dignifiedquire: > they just inject a set of hashes to be left alone.

wait, how are they marking them as such then? 😅

dignifiedquire (approved):

iroh/src/client/blobs.rs
rklaehn: there is just a hook where you can feed it hashes...

dignifiedquire: 😢

PR #2350, opened 2024-06-06
refactor(iroh)!: replace public fields in iroh client with accessors and use ref-cast to eliminate them entirely

rklaehn:
## Description

With 4 different clients, the current approach might be OK. But we are going to have more. E.g. gossip, see https://github.com/n0-computer/iroh/pull/2258 .

And in any case it feels weird to store the same thing multiple times.

So this replaces public fields in the iroh client with accessors and use ref-cast to eliminate them entirely. There is now only one rpc field, and you can switch from that to the different subsystem client views without runtime overhead, not even an arc clone.

## Breaking Changes

Everything that uses the iroh client will have to switch from field accesses to accessor fns.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-cli/src/commands/author.rs
+++ b/iroh-cli/src/commands/author.rs
@@ -48,7 +48,7 @@ impl AuthorCommands {
                 println!("Active author is now {}", fmt_short(author.as_bytes()));
             }
             Self::List => {
-                let mut stream = iroh.authors.list().await?;
+                let mut stream = iroh.authors().list().await?;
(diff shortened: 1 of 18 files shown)


dignifiedquire:
lets keep the fields and deprecate them in this PR, and drop them only in the next release

rklaehn:
I asked on the public discord. For what it's worth, people seem to be in favour of this change:
image

dignifiedquire (approved):

PR #2353, opened 2024-06-07
refactor(iroh): use boxed client to get rid of the C type parameter

rklaehn:
## Description

I implemented a boxed connection and a boxed service endpoint in quic-rpc.

With this we can get rid of the `<C: ServiceConnection<RpcService>` type parameter and make the quinn and mem client/server side the same type.

The nice thing about this approach is that it will not lead to additonal boxing on the mem path, and for the quinn or whatever io path the boxing will probably not matter that much compared to all the other things going on.

## Breaking Changes

Breaking changes affect solely the iroh client.

The iroh client as well as all the subsystem clients no longer need a type parameter C to distinguish between an in memory connection and a remote connection.

- Code that directly uses the clients should be unaffected. E.g. `iroh.blobs().read(hash)` will still compile.

- Code that takes a client as an argument will have to be modified to remove the type parameter. E.g.

```rust
fn download<C>(client: blobs::Client<C>) where C: ...
```

will become just

```rust
fn download(client: blobs::Client)
```

The type aliases `iroh::client::MemIroh` and `iroh::client::QuicIroh` for an iroh client specialized for memory or remote use have been retained, but will be removed in one of the next releases.

In detail:

- iroh::client::Iroh loses the `C` type parameter
- iroh::client::blobs::Client loses the `C` type parameter
- iroh::client::tags::Client loses the `C` type parameter
- iroh::client::authors::Client loses the `C` type parameter
- iroh::client::docs::Client loses the `C` type parameter

## Notes & open questions

Note: I marked the old type aliases MemIroh, QuicIroh etc as deprecated. That does not seem to actually do anything, but just serves as a reminder to remove them in the near future.

## Change checklist

- [x] Self-review.
~~- [x] Tests if relevant.~~


iroh/src/client.rs
```diff
@@ -1,18 +1,24 @@
...
 
-pub use self::mem::{Doc as MemDoc, Iroh as MemIroh, RpcClient as MemRpcClient};
+// #[deprecated]
+// pub use self::docs::Doc as MemDoc;
+// #[deprecated]
+// pub use self::Iroh as MemIroh;
+// #[deprecated]
+// pub use self::docs::Doc as QuicDoc;
+// #[deprecated]
+// pub use self::Iroh as QuicIroh;
```
flub, line 19: did you mean to remove these or leave them in?

flub (approved):
This is lovely

iroh/src/client.rs
rklaehn: I wanted to leave them in. I just commented them out to remove all useages because rustc did not scream at me due to the deprecation warnings. I think they don't work for reexports.

iroh/src/client/authors.rs
```diff
@@ -3,27 +3,24 @@
...
-use super::flatten;
+use super::{flatten, RpcService};
 
 /// Iroh authors client.
 #[derive(Debug, Clone, RefCast)]
 #[repr(transparent)]
-pub struct Client<C> {
-    pub(super) rpc: RpcClient<RpcService, C>,
+pub struct Client {
+    pub(super) rpc: RpcClient<RpcService, quic_rpc::transport::boxed::Connection<RpcService>>,
```
Frando, line 20: Use `IrohRpcClient`

iroh/src/client.rs
```diff
@@ -1,18 +1,24 @@
...
 use ref_cast::RefCast;
 
 #[doc(inline)]
 pub use crate::rpc_protocol::RpcService;
 
-mod mem;
 mod quic;
 
-pub use self::mem::{Doc as MemDoc, Iroh as MemIroh, RpcClient as MemRpcClient};
+#[deprecated]
```
Frando, line 12: maybe add a comment when we deprecated this to not forget to remove in the next release.

iroh/src/client/blobs.rs
```diff
@@ -35,28 +35,25 @@ use crate::rpc_protocol::{
...
 #[repr(transparent)]
-pub struct Client<C> {
-    pub(super) rpc: RpcClient<RpcService, C>,
+pub struct Client {
+    pub(super) rpc: IrohRpcClient,
 }
 
-impl<'a, C: ServiceConnection<RpcService>> From<&'a Iroh<C>> for &'a RpcClient<RpcService, C> {
-    fn from(client: &'a Iroh<C>) -> &'a RpcClient<RpcService, C> {
+impl<'a> From<&'a Iroh> for &'a IrohRpcClient {
```
Frando, line 50: Seems to belong in `client.rs` not `blobs.rs`?

iroh/src/client/docs.rs
```diff
@@ -36,21 +36,18 @@ use crate::rpc_protocol::{
...
-use super::{blobs, flatten};
+use super::{blobs, flatten, IrohRpcClient};
 
 /// Iroh docs client.
 #[derive(Debug, Clone, RefCast)]
 #[repr(transparent)]
-pub struct Client<C> {
-    pub(super) rpc: RpcClient<RpcService, C>,
+pub struct Client {
+    pub(super) rpc: RpcClient<RpcService, quic_rpc::transport::boxed::Connection<RpcService>>,
```
Frando, line 45: use `IrohRpcClient`

iroh/src/client/tags.rs
```diff
@@ -3,23 +3,21 @@
...
+use super::RpcService;
+use crate::rpc_protocol::{DeleteTagRequest, ListTagsRequest};
 
 /// Iroh tags client.
 #[derive(Debug, Clone, RefCast)]
 #[repr(transparent)]
-pub struct Client<C> {
-    pub(super) rpc: RpcClient<RpcService, C>,
+pub struct Client {
+    pub(super) rpc: RpcClient<RpcService, quic_rpc::transport::boxed::Connection<RpcService>>,
```
Frando, line 17: use `IrohRpcClient`

iroh/src/node/builder.rs
```diff
@@ -483,6 +483,7 @@ where
 
         // Initialize the internal RPC connection.
         let (internal_rpc, controller) = quic_rpc::transport::flume::connection(1);
+        let controller = quic_rpc::transport::boxed::Connection::new(controller);
```
Frando, line 486: So this still boxes the internal RPC connection? You said somewhere that the internal RPC connection circumvents the boxing. If this is the case, maybe add a comment here why/how this works?

iroh/src/client/blobs.rs
rklaehn: I think it is not needed anymore because we do ref_cast anyway...

dignifiedquire:
@rklaehn please add some more details on the breaking changes, otherwise I have to go through the whole PR when writing the changelog

rklaehn:
> @rklaehn please add some more details on the breaking changes, otherwise I have to go through the whole PR when writing the changelog

Added some more text. Is that sufficient?

dignifiedquire:
thanks, looks great

PR #2357, opened 2024-06-11
feat(iroh): Accept custom protocols

Frando:
## Description

Allow to accept custom protocols in an iroh node.

See the included example which uses both a custom protocol and core iroh protocols.

supersedes #2284

## Breaking Changes

todo

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/examples/custom-protocol.rs
+++ b/iroh/examples/custom-protocol.rs
@@ -31,7 +31,7 @@ async fn main() -> Result<()> {
     let args = Cli::parse();
     // create a new node
     let node = iroh::node::Node::memory()
-        .accept(ExampleProto::ALPN, |node| ExampleProto::build(node))
+        .accept(EXAMPLE_ALPN, |node| ExampleProto::build(node))
         .spawn()
         .await?;
 
@@ -54,6 +54,8 @@ async fn main() -> Result<()> {
     Ok(())
 }
 
+const EXAMPLE_ALPN: &'static [u8] = b"example-proto/0";
+
 #[derive(Debug)]
 struct ExampleProto<S> {
     node: Node<S>,
@@ -70,8 +72,6 @@ impl<S: Store + fmt::Debug> Protocol for ExampleProto<S> {
 }
 
 impl<S: Store + fmt::Debug> ExampleProto<S> {
-    const ALPN: &'static [u8] = b"example-proto/0";
-
     fn build(node: Node<S>) -> Arc<Self> {
         Arc::new(Self { node })
     }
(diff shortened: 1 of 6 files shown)


iroh/examples/custom-protocol.rs
```diff
@@ -0,0 +1,135 @@
...
+struct ExampleProto<S> {
+    node: Node<S>,
+}
+
+impl<S: Store + fmt::Debug> Protocol for ExampleProto<S> {
+    fn as_arc_any(self: Arc<Self>) -> Arc<dyn Any + Send + Sync> {
+        self
+    }
+
+    fn accept(self: Arc<Self>, conn: quinn::Connection) -> Boxed<Result<()>> {
```
flub, line 69: `Boxed` is a bit hard to read, would it make sense to import this as `BoxedFuture`?

flub, line 69: I think `quinn::Connection` is too late, you can't do things like use 0-rtt anymore.  This should probably be `quinn::Connecting` in quinn 0.10 and likely even `quinn::Incoming` in 0.11 (not sure how to ensure we don't miss this in the upgrade since `Connecting` keeps existing in 0.11)

iroh/examples/custom-protocol.rs
```diff
@@ -0,0 +1,135 @@
...
+
+const EXAMPLE_ALPN: &'static [u8] = b"example-proto/0";
+
+#[derive(Debug)]
+struct ExampleProto<S> {
+    node: Node<S>,
+}
+
+impl<S: Store + fmt::Debug> Protocol for ExampleProto<S> {
+    fn as_arc_any(self: Arc<Self>) -> Arc<dyn Any + Send + Sync> {
```
flub, line 65: `into_arc_any` is more correct i guess?  `as` seems like a naming violation

flub: why does this have to exist at all?  why not do this directly on the argument the node builder method accepts?

iroh/src/node.rs
```diff
@@ -150,6 +153,15 @@ impl<D: BaoStore> Node<D> {
         self.inner.endpoint.my_relay()
     }
 
+    /// Returns the protocol handler for a alpn.
+    pub fn get_protocol<P: Protocol>(&self, alpn: &[u8]) -> Option<Arc<P>> {
+        let protocols = self.protocols.read().unwrap();
+        let protocol: Arc<dyn Protocol> = protocols.get(alpn)?.clone();
+        let protocol_any: Arc<dyn Any + Send + Sync> = protocol.as_arc_any();
```
flub, line 160: why stored as `Any` if it is then downcast anyway?

iroh/examples/custom-protocol.rs
Frando: True. Changed to `Connecting`.

iroh/examples/custom-protocol.rs
Frando: I don't think there is a way to cast from `dyn Protocol` to `dyn Any` apart from having this as a trait method implemented on the concrete type.
```rust
let x: Arc<dyn Protocol> = protocols.get(alpn);
// this does not compile even if Protocol: Any
let x: Arc<dyn Any> = x;
```

However what we can do, and what I did in the most recent commit, is move this into a *separate* trait which is auto-impled for all `T: Send + Sync + 'static`. I like this more, now users don't have to care about this at all and it is purely an implementation detail.

Frando:
This was refined further in #2358 after discovering more needs when porting the core iroh protocols to this. Closing this, let's finish it in #2358.

PR #2358, opened 2024-06-12
refactor(iroh): Allow to register custom protocols

Frando:
## Description

* feat(iroh-net): Allow to set the list of accepted ALPN protocols at runtime
* refactor(iroh): Spawning a node can now be performed in two stages: First `Builder::build()` is called, which returns a Future that resolves to a new type `ProtocolBuilder`. The `ProtocolBuilder` is then spawned into the actual, running `Node`. If the intermediate step is not needed, `Builder::spawn` performs both in one call, therefore this change is not breaking.
* feat(iroh): Allow to accept custom ALPN protocols in an Iroh node. Introduce a `ProtocolHandler` trait for accepting incoming connections and adds `ProtocolBuilder::accept` to register these handlers per ALPN.
* refactor(iroh): Move towards more structured concurrency by spawning tasks into a `JoinSet`
* refactor(iroh): Improve shutdown flow and perform more things concurently.

originally based on #2357 but now directly to main

## Breaking Changes

* `iroh_net::endpoint::make_server_config` takes `Arc<quinn::TransportConfig>` instead of `Option<quinn::TransportConfig>`. If you used the `None` case, replace with `quinn::TransportConfig::default()`.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.


iroh/src/node.rs
```diff
@@ -173,8 +171,11 @@ impl<D: BaoStore> Node<D> {
     pub async fn shutdown(self) -> Result<()> {
         self.inner.cancel_token.cancel();
 
-        if let Ok(task) = Arc::try_unwrap(self.task) {
-            task.await?;
+        if let Ok(mut task) = Arc::try_unwrap(self.task) {
+            task.take().expect("cannot be empty").await?;
+        }
+        if let Some(mut tasks) = self.inner.tasks.lock().unwrap().take() {
```
dignifiedquire, line 177: you could skip the option I believe

Frando: did that and also moved into `NodeInner`

rklaehn:
I am still not completely sold on the way we build custom protocols. I am sorry that I was not more active when we discussed this.

Basically we have a single stage builder that we feed callbacks to create protocol handlers. These callbacks will then be called at a later time, when the endpoint already exists. What if we do a two stage builder instead (that can be shortcut in the default case).

```rust
let unspawned = Node::builder().secret_key(key).whatever().build()?;

let h1 = Handler1::new(unspawned.endpoint());
let h2 = Handler2::new(unspawned.endpoint(), unspawned.blob_store(), h1);

let spawned = unspawned.add_handler(b"alpn1", h1).add_handler(b"alpn2", h2).spawn()?;
```

It seems that the handler map that we now have in the node is just a somewhat limited replacement to a local scope.

Frando:
A two-stage builder is certainly an option. The question to me is if the intermediate node, the one you get from `Builder::build()` in your example, will be `Node` or another type like `UnspawnedNode`.
The former has the advantage that we do not have to add another type to the public API and protocol handler builders can just use the `Node`. The downside would be that you wouldn't have type-level knowledge if a node is spawned or not. On the unspawned node, RPC calls would time out because `NodeInner::run` isn't running yet.
This however is the same as in the current code: At the time the protocol handler builder closures are called, the node is readily built on the type level, but its `run` function is not yet spawned, so RPC calls will time out.
I think this is fine if we document it - the reasoning for having the `Node` available in the protocol handler builders is so that they can pass the node or the RPC client into a task if they need to or act on it later (like the `custom_protocols` example).

rklaehn:
So this is how the two stage builder spawn fn looks.

You have to provide the set of ALPNs on build, get an unspawned (or rather listening but non-accepting) node, then have to add all the handlers that you promised, and then spawn (the accept loop).

If you add a handler for a non-registered ALPN or miss one handler, the final spawn will fail.

```rust
    pub async fn spawn(self) -> Result<Node<D>> {
        let unspawned = self.build(&[GOSSIP_ALPN, BLOBS_ALPN, DOCS_ALPN]).await?;
        let gossip = unspawned.gossip.clone();
        let blobs = BlobsProtocol::new(unspawned.blobs_db().clone(), unspawned.local_rt().clone());
        let docs = unspawned.inner.sync.clone();
        unspawned
            .add_handler(GOSSIP_ALPN, Arc::new(gossip))?
            .add_handler(BLOBS_ALPN, Arc::new(blobs))?
            .add_handler(DOCS_ALPN, Arc::new(docs))?
            .spawn()
            .await
    }
```

iroh/src/node/builder.rs
```diff
@@ -404,7 +419,7 @@ where
         let endpoint = Endpoint::builder()
             .secret_key(self.secret_key.clone())
             .proxy_from_env()
-            .alpns(PROTOCOLS.iter().map(|p| p.to_vec()).collect())
+            .alpns(vec![])
```
dignifiedquire, line 422: can you skip calling it?

iroh/src/node/builder.rs
```diff
@@ -630,15 +595,25 @@ where
...
+
+        // Shutdown sync and blobs.
+        inner.sync.shutdown().await.ok();
+        inner.db.shutdown().await;
+
+        // Shutdown protocol handlers.
+        protocols.shutdown().await;
+
+        // Abort remaining tasks.
+        join_set.shutdown().await;
```
dignifiedquire, line 611: should the above all be raced, or do you want to wait on each serially?

Frando: I changed it to racing all shutdowns, apart from the join set this comes after, as potentially the shutdown might want to terminate tasks running there more gracefully (the latter is not the case at the moment though).

iroh-blobs/src/store/traits.rs
```diff
@@ -295,7 +295,7 @@ pub trait ReadableStore: Map {
 }
 
 /// The mutable part of a Bao store.
-pub trait Store: ReadableStore + MapMut {
+pub trait Store: ReadableStore + MapMut + std::fmt::Debug {
```
rklaehn, line 298: 👍

iroh-net/src/endpoint.rs
```diff
@@ -296,17 +293,41 @@ impl Builder {
...
+struct StaticConfig {
+    secret_key: SecretKey,
+    transport_config: Arc<quinn::TransportConfig>,
+    keylog: bool,
+    concurrent_connections: Option<u32>,
+}
+
+impl StaticConfig {
+    /// Build a [`quinn::ServerConfig`] with the specified ALPN protocols.
+    fn build(&self, alpn_protocols: Vec<Vec<u8>>) -> Result<quinn::ServerConfig> {
```
rklaehn, line 307: I don't like the name. It sounds like it builds a StaticConfig, but it does build a final ServerConfig.

iroh/src/node/builder.rs
```diff
@@ -715,6 +690,229 @@ where
...
+    client: crate::client::MemIroh,
+    internal_rpc: FlumeServerEndpoint<RpcService>,
+    rpc_endpoint: E,
+    protocols: ProtocolMap,
+    #[debug("callback")]
+    gc_done_callback: Option<Box<dyn Fn() + Send>>,
+    gc_policy: GcPolicy,
+}
+
+impl<D: iroh_blobs::store::Store, E: ServiceEndpoint<RpcService>> UnspawnedNode<D, E> {
```
rklaehn, line 713: Should we have a get_protocol on this one as well?

E.g. you are building a complex iroh node, and don't want to pass copies of protocol x to the builder fn for protocol y that uses x?

Not strictly necessary as it was before, but since the info is available...

iroh/src/node/builder.rs
```diff
@@ -715,6 +690,229 @@ where
...
+    ///
+    pub fn accept(mut self, alpn: &'static [u8], handler: Arc<dyn Protocol>) -> Self {
+        self.protocols.insert(alpn, handler);
+        self
+    }
+
+    /// Return a client to control this node over an in-memory channel.
+    ///
+    /// Note that the client can only be used after spawning the node,
+    /// until then all RPC calls will time out.
```
rklaehn, line 770: If you grab the client, then do a rpc call, then immediately finish the build, a RPC call might work, right?

So it's not so much that they will time out but that they will hang until the rpc accept loop goes live.

iroh/src/node.rs
```diff
@@ -152,27 +153,33 @@ impl<D: BaoStore> Node<D> {
...
         Ok(())
     }
 
     /// Returns a token that can be used to cancel the node.
     pub fn cancel_token(&self) -> CancellationToken {
         self.inner.cancel_token.clone()
     }
+
+    /// Get a protocol handler.
+    pub fn get_protocol<P: Protocol>(&self, alpn: &[u8]) -> Option<Arc<P>> {
```
rklaehn, line 180: Not for this PR, but I was wondering if we should have a generic "place to put stuff" on the node that also works for non protocols. E.g. the blobs db or the docs db or the downloader, stuff that you might want to use from multiple protocols but that is not a protocol.

iroh/src/node/builder.rs
```diff
@@ -715,6 +690,229 @@ where
...
+/// A node that is initialized but not yet spawned.
+///
+/// This is returned from [`Builder::build`] and may be used to register custom protocols with
+/// [`Self::accept`]. It provides access to the services which are already started, the node's
+/// endpoint and a client to the node.
+///
+/// Note that the client returned from [`Self::client`] can only be used after spawning the node,
+/// until then all RPC calls will time out.
+#[derive(derive_more::Debug)]
+pub struct UnspawnedNode<D, E> {
```
rklaehn, line 702: Maybe call this CustomProtocolBuilder since you will use it only when you want to register custom protocols? Naming 🤷

iroh/src/node/protocol.rs
```diff
@@ -0,0 +1,118 @@
...
+    }
+}
+
+impl<S: iroh_blobs::store::Store> Protocol for BlobsProtocol<S> {
+    fn accept(self: Arc<Self>, conn: Connecting) -> BoxedFuture<Result<()>> {
+        Box::pin(async move {
+            iroh_blobs::provider::handle_connection(
+                conn.await?,
+                self.store.clone(),
+                MockEventSender,
```
rklaehn, line 90: What's the deal with this MockEventSender? If this thing is only ever used with a mock sender, maybe we can remove event sending entirely? Or is it used in iroh.network?

rklaehn (commented):
Added a bunch of comments. Overall looks great.

iroh-net/src/endpoint.rs
dignifiedquire: what if we renamed `StaticConfig` to `ServerConfigBuilder`?

iroh/src/node/builder.rs
Frando: Added this.

iroh/src/node/builder.rs
Frando: Changed the docs.

iroh/src/node/protocol.rs
Frando: I think we ripped the events out of `iroh::Node` but kept support for events in `iroh_blobs`. This is unrelated to this PR though, it just moved around.

iroh-net/src/endpoint.rs
Frando: I went with `StaticConfig::create_server_config` for now. I'm fine to change further, don't mind much (and it's not pubic API, internal only).

Frando:
Addressed most review comments. And:

* Renamed `UnspawnedNode` to `ProtocolBuilder`

Other public API namings we could bikeshed a bit:

* `trait Protocol`: Fine as is? Or better `ProtocolHandler`?
* `ProtocolBuilder::accept(&self, alpn: &'static [u8], protocol: Arc<dyn Protocol>`: Fine as is? Or better `register_protocol`?

I think I'm fine with the current namings, but also very open to changing if others find other names more intuitive.

iroh/src/node/builder.rs
Frando: Went with `ProtocolBuilder` for now. Open to changing further :)

iroh/src/node.rs
Frando: Yup, let's discuss this separately though. Right now the unspawned node / protocol builder gives access to *more* stuff than the final node. But we also want to keep the API of the node small and streamlined, and I think generally advise to work with the client if possible to not have to increase the public API surface too much. Protocol handlers are the kind of "exception" IMO so I think it's fine to have more things on the unspawned node than on the final node (and everything is clone I think).

iroh/src/node/protocol.rs
rklaehn: Yeah, I was just confused the first time I saw this because I thought you had introduced the MockEventSender... Anyway, unrelated to this PR.

rklaehn:
> trait Protocol: Fine as is? Or better ProtocolHandler?

I don't hate trait Protocol, but like ProtocolHandler a bit more. For me a _protocol_ is a rust module containing a bunch of request and response types, usually in some enums, and some help text describing how to use them (e.g. serialize them with postcard), not something active.

rklaehn:
Not sure if we want to address this in this PR or the next, but in Builder::build it now always calls register_iroh_protocols. So there is no way to opt out of them, just to add extra protocols.

I wasn't quite sure how it should look if you want to have a "naked" node. Also, it seems that we are not really ready to create a naked node. So maybe let's do it in the next PR that will allow disabling docs.

rklaehn (approved):

Frando:
I renamed `Protocol` to `ProtocolHandler`. Agree with @rklaehn that this looks better.

> Not sure if we want to address this in this PR or the next, but in Builder::build it now always calls register_iroh_protocols. So there is no way to opt out of them, just to add extra protocols.

Yup, let's defer this to a followup.

iroh/src/node/builder.rs
```diff
@@ -715,6 +690,237 @@ where
...
+
+    /// Returns a protocol handler for an ALPN.
+    ///
+    /// This downcasts to the concrete type and returns `None` if the handler registered for `alpn`
+    /// does not match the passed type.
+    pub fn get_protocol<P: ProtocolHandler>(&self, alpn: &[u8]) -> Option<Arc<P>> {
+        self.protocols.get_typed(alpn)
+    }
+
+    /// Register the core iroh protocols (blobs, gossip, docs).
```
flub, line 808:
```suggestion
    /// Registers the core iroh protocols (blobs, gossip, docs).
```

iroh/src/node/builder.rs
```diff
@@ -715,6 +690,237 @@ where
...
+        self = self.accept(GOSSIP_ALPN, Arc::new(gossip));
+
+        // Register docs.
+        let docs = self.inner.sync.clone();
+        self = self.accept(DOCS_ALPN, Arc::new(docs));
+
+        self
+    }
+
+    /// Spawn the node and start accepting connections.
```
flub, line 826:
```suggestion
    /// Spawns the node and start accepting connections.
```

iroh/src/node/protocol.rs
```diff
@@ -0,0 +1,127 @@
...
+        self
+    }
+}
+
+#[derive(Debug, Clone, Default)]
+pub(super) struct ProtocolMap(BTreeMap<&'static [u8], Arc<dyn ProtocolHandler>>);
+
+impl ProtocolMap {
+    /// Returns the registered protocol handler for an ALPN as a concrete type.
+    pub fn get_typed<P: ProtocolHandler>(&self, alpn: &[u8]) -> Option<Arc<P>> {
```
flub, line 50: Could we have the same visibility marker here as on the struct, so `pub(super)`?  I find that a lot easier to read.  Plus there's fewer surprises: if the super module decides to export this than suddenly these functions are really pub, and that might have been unintended.

iroh/src/node/protocol.rs
```diff
@@ -0,0 +1,127 @@
...
+        let protocol_ref = Arc::downcast(protocol_any).ok()?;
+        Some(protocol_ref)
+    }
+
+    /// Returns the registered protocol handler for an ALPN as a [`Arc<dyn ProtocolHandler>`].
+    pub fn get(&self, alpn: &[u8]) -> Option<Arc<dyn ProtocolHandler>> {
+        self.0.get(alpn).cloned()
+    }
+
+    /// Insert a protocol handler.
```
flub, line 62:
```suggestion
    /// Inserts a protocol handler.
```

iroh/src/node/protocol.rs
```diff
@@ -0,0 +1,127 @@
...
+    pub fn insert(&mut self, alpn: &'static [u8], handler: Arc<dyn ProtocolHandler>) {
+        self.0.insert(alpn, handler);
+    }
+
+    /// Returns an iterator of all registered ALPN protocol identifiers.
+    pub fn alpns(&self) -> impl Iterator<Item = &&[u8]> {
+        self.0.keys()
+    }
+
+    /// Shutdown all protocol handlers.
```
flub, line 72:
```suggestion
    /// Shuts down all protocol handlers.
```

flub (approved):
I like this approach of structured concurrency, essentially the same as I did in iroh-relay.  Also TIL about SharedAbortingJoinHandle.  I'm a little unsure if there's any downsides to having many clones of the JoinHandle, but let's see.  It makes it a lot easier to use.

PR #2366, opened 2024-06-17
refactor(iroh-net)!: Rename Endpoint::local_addr to bound_sockets

flub:
## Description

This renames Endpoint::local_addr to Endpoint::bound_sockets.  We have
a lot of different meanings for "addr" and also don't really have a
notion of "local_" prefix, since the endpoint is not 1:1 connected
with a peer.  The "local_" naming comes from connected sockets which
have local_addr and peer_addr, which is not applicable for the
Endpoint.

## Breaking Changes

- Endpoint::local_addr -> Endpoint::bound_sockets

## Change checklist

- [x] Self-review.
- ~~[ ] Tests if relevant.~~

### DIFF
--- a/iroh-net/bench/src/iroh.rs
+++ b/iroh-net/bench/src/iroh.rs
@@ -29,7 +29,7 @@ pub fn server_endpoint(rt: &tokio::runtime::Runtime, opt: &Opt) -> (NodeAddr, En
             .bind(0)
             .await
             .unwrap();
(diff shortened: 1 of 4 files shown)


matheus23 (approved):
Really like the rename. Bound sockets is a lot clearer to me than local addr.
(Local could e.g. also imply loopback/localhost)

divagant-martian (approved):

PR #2367, opened 2024-06-17
Experiment: two stage builder approach for custom protocols

rklaehn:
## Description

This is like https://github.com/n0-computer/iroh/pull/2358, but instead of having a factory function to create a protocol, it uses a two stage builder approach where you first build a non-accepting node, then add all the promised handlers, and then turn it into a full (accepting, rpc active) node.

Note that I have also changed the signature of the protocol handlers, but that is unrelated to the main difference and a matter of taste I guess. In one case you leave it up to the handlers to make a future long lived, in the other case you force the client to arc the handlers...

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-blobs/src/store/traits.rs
+++ b/iroh-blobs/src/store/traits.rs
@@ -6,6 +6,7 @@ use bao_tree::{
     BaoTree, ChunkRanges,
 };
 use bytes::Bytes;
+use derive_more::Debug;
 use futures_lite::{Stream, StreamExt};
 use genawaiter::rc::{Co, Gen};
 use iroh_base::rpc::RpcError;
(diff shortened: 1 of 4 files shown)


rklaehn:
The custom_protocols example could be made to look like this:

```rust
#[tokio::main]
    setup_logging();
    let args = Cli::parse();
    // create a new node
    let node = iroh::node::Node::memory()
        .accept(EXAMPLE_ALPN)
        .build();
    let proto = ExampleProto::build(&node);
    node
        .add_handler(EXAMPLE_ALPN, proto))
        .spawn()
        .await?;
```

Due to the two stage nature you need to use the `EXAMPLE_ALPN` twice, once before building the unspawned node and once as you add the handler.

rklaehn:
Looks like we can change the ALPN set once the endpoint is created. With that, we get:

```rust
#[tokio::main]
    setup_logging();
    let args = Cli::parse();
    let node = iroh::node::Node::memory()
        .build();
    let proto = ExampleProto::build(&node);
    node
        .add_handler(EXAMPLE_ALPN, proto))
        .spawn()
        .await?;
```

and can remove the duplicate use of the ALPN parameter.

rklaehn:
Folded into https://github.com/n0-computer/iroh/pull/2358

PR #2369, opened 2024-06-17
Flub/endpoint no endpoint

flub:
## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


iroh-net/src/magicsock.rs
```diff
@@ -2592,6 +2586,127 @@ fn disco_message_sent(msg: &disco::Message) {
...
+    /// Whether we have an existing portmap open (UPnP, PMP, or PCP).
+    have_port_map: bool,
+
+    /// Probe indicating the presence of port mapping protocols on the LAN.
+    portmap_probe: Option<portmapper::ProbeOutput>,
+
+    /// This node's preferred relay server for incoming traffic. The node might be be temporarily
+    /// connected to multiple relay servers (to send to other nodes)
+    /// but PreferredRelay is the instance number that the node
+    /// subscribes to traffic at. Zero means disconnected or unknown.
```
divagant-martian, line 2676: This talks about sentinel values, which at least in rust are considered an anti pattern and we don't use.. so I guess this comes from tailscale and the go world. Could you update this part to better reflect the current state?

iroh-net/src/magicsock.rs
```diff
@@ -2592,6 +2586,127 @@ fn disco_message_sent(msg: &disco::Message) {
...
+
+    /// The fastest recent time to reach various relay STUN servers, in seconds.
+    ///
+    /// This should only be updated rarely, or when there's a
+    /// material change, as any change here also gets uploaded to the control plane.
+    relay_latency: BTreeMap<String, f64>,
+}
+
+impl NetInfo {
+    /// reports whether `self` and `other` are basically equal, ignoring changes in relay ServerLatency & RelayLatency.
```
divagant-martian, line 2687: this also looks like it was dragged without much thought from Tailscale, I'm not even sure we have a "server latency"? Worth updating even if just by making them more vague. Better vague than wrong

divagant-martian (approved):
I find this really easier to understand overall. I left some comments about moved docs that _could_ be improved because the look like go docs and are, in fact, inaccurate or even wrong.

If you consider doing those changes out of scope it's fine by me

dignifiedquire (approved):

iroh-net/src/magicsock.rs
flub: [510f865](https://github.com/n0-computer/iroh/pull/2369/commits/510f865b140721ad8eb6bd9434eebc46a71af9c1)

iroh-net/src/magicsock.rs
flub: [510f865](https://github.com/n0-computer/iroh/pull/2369/commits/510f865b140721ad8eb6bd9434eebc46a71af9c1)

PR #2376, opened 2024-06-18
feat(iroh-net): local swarm discovery

### LINKED ISSUE #2354: Add optional LAN discovery service
The current discovery service lacks a Local Area Network (LAN) discovery mechanism. I would like an optional LAN discovery feature that, if enabled, would allow the system to fall back on LAN discovery whenever it is unable to connect to the discovery server.

It should be noted that not all LANs support broadcast or multicast, hence mDNS might not be entirely reliable. Previously, I addressed this issue by performing parallel IP scans via a specified UDP port, but there might be a better solution.

ramfox:
## Description

Using [swarm-discovery](https://github.com/rkuhn/swarm-discovery), this PR implements a discovery service for `LocalSwarmDiscovery`, allowing nodes on the same local network to discover each other without using the outside internet.

closes #2354

## Notes & open questions

1) Is `LocalNodeDiscovery` too much of a mouthful? `swarm-discovery` says, of itself, "This library offers a lightweight discovery service based on mDNS." Soooo, how comfortable are we just calling this `mdns`?
edit: calling this `LocalSwarmDiscovery` & not `mdns`

2) There are some const to bikeshed:
```rust
/// The n0 local node discovery name
const N0_MDNS_SWARM: &str = "iroh.local.node.discovery";
edit: now N0_LOCAL_SWARM: &str = "iroh.loca.swarm"

/// Provenance string
const PROVENANCE: &str = "local.node.discovery";
edit: now "local.swarm.discovery"

/// How long we will wait before we stop sending discovery items
const DISCOVERY_DURATION: Duration = Duration::from_secs(10);
```

3) For the `local_swarm_discovery` example, I pulled some code from `iroh-cli` into a new `iroh-progress`, so that I could re-use some progress terminal output it in the example. `iroh-progress` might not be the right name, but what do we think of a separate crate for putting in reusable UI-ish pieces. If we aren't on board with this, I can just copy and paste the logic in for now & remove the crate.
edit: Removed this in favor of copy and pasting the code

4) I've added `LocalSwarmDiscovery` as a default discovery service, I'm assuming that's what we are going for here?

## Change checklist

- [x] Self-review.
- [x] tests


zh522130:
It would be great if there could be an automatic fallback to local discovery when the relay service is not reachable, rather than being forced to choose between the two.

matheus23:
@zh522130 AFAIK this will be the case - you can enable/disable mdns and using a relay independently of each other! And when you enable both, it'll use whatever is available to try to establish a connection :)

matheus23:
> 1. Is `LocalNodeDiscovery` too much of a mouthful? `swarm-discovery` says, of itself, "This library offers a lightweight discovery service based on mDNS." Soooo, how comfortable are we just calling this `mdns`?

IIUC, there's a different way to do mDNS discovery. I'd say the two types are "swarm discovery" and "query based discovery".
Where swarm discovery tries to not explode in traffic if you have a lot of devices in your network, and query based discovery is a lot dumber: It just asks mdns anytime you're looking for something. This means potentially faster answers and no traffic if no node is looking for anything, but also potentially an explosion of traffic with lots of nodes that want to form a swarm.

I think both approaches have ups and downs - so I'd prefer not to squat the mDNS name - at least not "just" as `MdnsDiscovery`, but maybe `MdnsSwarmDiscovery` :sweat_smile: (Not a lot shorter than `LocalSwarmDiscovery`, but at least it mentions mdns?)

matheus23:
> 3\. For the `local-node-discovery` example, I pulled some code from `iroh-cli` into a new `iroh-progress`, so that I could re-use some progress terminal output it in the example. `iroh-progress` might not be the right name, but what do we think of a separate crate for putting in reusable UI-ish pieces. If we aren't on board with this, I can just copy and paste the logic in for now & remove the crate.

`iroh-cli-utils`? Maybe expanding the scope from only doing progress might be a good idea.

zh522130:
Very good, the previous version was reporting errors during testing, but I tested the latest code today and it is now functioning properly.

iroh-net/src/discovery/local_node_discovery.rs
```diff
@@ -0,0 +1,250 @@
...
+                    }
+                    Message::Timeout(node_id) => {
+                        tracing::trace!(?node_id, "LocalNodeDiscovery Message::Timeout");
+                        senders.remove(&node_id);
+                    }
+                    Message::ChangeLocalAddrs(addrs) => {
+                        tracing::trace!(?addrs, "LocalNodeDiscovery Message::ChangeLocalAddrs");
+                        let callback_send = task_sender.clone();
+                        let g = guard.take();
+                        drop(g);
```
rkuhn, line 158: One of the next things I wanted to do is to add an API for changing the locally advertised addresses without stopping the service; this has the advantage of not losing the swarm size estimate and I just realised that this current approach means that expiration isn’t signalled reliably.

In any case, it is good to see this example before working on that feature. Will let you know when I have implemented it.

iroh-net/src/discovery/local_node_discovery.rs
```diff
@@ -0,0 +1,250 @@
...
+use tokio::task::{JoinHandle, JoinSet};
+
+use crate::{
+    discovery::{Discovery, DiscoveryItem},
+    AddrInfo, Endpoint, NodeId,
+};
+
+/// The n0 local node discovery name
+// TODO(ramfox): bikeshed
+const N0_MDNS_SWARM: &str = "iroh.local.node.discovery";
```
rkuhn, line 23: Since this is used many times within each mDNS packet and size is limited, I’d recommend to use the shortest name you can think of.

I half-read about a compression feature in the DNS spec that would help, but at least I haven’t done anything specifically to implement it—will have to look into hickory to find out.

iroh-net/src/discovery/local_node_discovery.rs
```diff
@@ -0,0 +1,250 @@
...
+                                return;
+                            }
+                        };
+                    }
+                }
+            }
+        });
+        Self {
+            handle,
+            sender: send.clone(),
```
rkuhn, line 179: extraneous clone?

iroh-net/src/discovery/local_node_discovery.rs
```diff
@@ -0,0 +1,250 @@
...
+                        node_addrs.insert(discovered_node_id, peer_info);
+                    }
+                    Message::SendAddrs((node_id, sender)) => {
+                        tracing::trace!(?node_id, "LocalNodeDiscovery Message::SendAddrs");
+                        if let Some(peer_info) = node_addrs.get(&node_id) {
+                            let item: DiscoveryItem = peer_info.into();
+                            tracing::debug!(?item, "sending DiscoveryItem");
+                            sender.send_async(Ok(item)).await.ok();
+                        }
+                        senders.insert(node_id, sender);
```
rkuhn, line 139: should this only happen in case no address is known yet?

iroh-net/src/discovery/local_node_discovery.rs
```diff
@@ -0,0 +1,250 @@
...
+            addrs
+                .entry(socketaddr.port())
+                .and_modify(|a| a.push(socketaddr.ip()))
+                .or_insert(vec![socketaddr.ip()]);
+        }
+
+        let mut discoverer =
+            Discoverer::new_interactive(N0_MDNS_SWARM.to_string(), node_id.to_string())
+                .with_callback(callback);
+        if !addrs.is_empty() {
```
divagant-martian, line 211: you can probable remove this check

iroh-net/src/discovery/local_node_discovery.rs
```diff
@@ -0,0 +1,250 @@
...
+            ) {
+                Ok(guard) => Some(guard),
+                Err(e) => {
+                    tracing::error!("LocalNodeDiscovery error creating discovery service: {e}");
+                    return;
+                }
+            };
+            let mut node_addrs: HashMap<PublicKey, Peer> = HashMap::default();
+            let mut senders: HashMap<PublicKey, Sender<Result<DiscoveryItem>>> = HashMap::default();
+            let mut timeouts = JoinSet::new();
```
divagant-martian, line 77: These seem to be serving conflicting purposes. 

On one hand we store a single `Sender` per node, but this is overwritten (and thus the previous one being dropped) for any new call for the same node. In particular this means that concurrent requests (`Discovery::resolve`) to the same node when no address is known will result in the first one receiving nothing, because the second one effectively cancels it. 

In the same fashion, the timeouts of any previous request take over the timeouts of new requests. So, if we do a request at `t=0` a new timeout task is spawned, then another request at `t=5` the first one will have a timeout of 10secs, the second one of 5 seconds because there is nothing cancelling, or otherwise differentiating, the first timeout.

If we want to reuse senders for all concurrent requests of the same peer, `senders` needs to be moved out of the run loop an be shared between the "service" and "handler". If we want timeouts to be applied per request, these need to be moved to the `Discovery::resolve" impl itself

divagant-martian (changes requested):

iroh-net/src/discovery/local_node_discovery.rs
ramfox: Oh yes this would be a great API change 🙌 thanks!

iroh-cli/Cargo.toml
```diff
@@ -40,6 +40,7 @@ hex = "0.4.3"
 human-time = "0.1.6"
 indicatif = { version = "0.17", features = ["tokio"] }
 iroh = { version = "0.18.0", path = "../iroh", features = ["metrics"] }
+iroh-progress = { version = "0.18.0", path = "../iroh-progress" }
```
dignifiedquire, line 43: why the introduction of a new crate?

iroh-net/src/discovery/local_node_discovery.rs
```diff
@@ -0,0 +1,255 @@
...
+impl Drop for LocalNodeDiscovery {
+    fn drop(&mut self) {
+        self.handle.abort();
+    }
+}
+
+#[derive(Debug)]
+enum Message {
+    Discovery((String, Peer)),
+    SendAddrs((NodeId, Sender<Result<DiscoveryItem>>)),
```
dignifiedquire, line 48: no need for the double parens

iroh-net/src/discovery/local_node_discovery.rs
```diff
@@ -0,0 +1,255 @@
...
+
+impl Drop for LocalNodeDiscovery {
+    fn drop(&mut self) {
+        self.handle.abort();
+    }
+}
+
+#[derive(Debug)]
+enum Message {
+    Discovery((String, Peer)),
```
dignifiedquire, line 47: no need for the double parens

iroh-net/src/discovery/local_node_discovery.rs
```diff
@@ -0,0 +1,255 @@
...
+            let mut guard = match LocalNodeDiscovery::spawn_discoverer(
+                node_id,
+                task_sender.clone(),
+                BTreeSet::new(),
+                &rt,
+            ) {
+                Ok(guard) => Some(guard),
+                Err(e) => {
+                    tracing::error!("LocalNodeDiscovery error creating discovery service: {e}");
+                    return;
```
dignifiedquire, line 72: why is the method not returning an error, instead of swalling it?

dignifiedquire: eg, you could start the discoverer, and then sapwn

iroh-net/src/discovery/local_node_discovery.rs
```diff
@@ -0,0 +1,255 @@
...
+            };
+            let mut node_addrs: HashMap<PublicKey, Peer> = HashMap::default();
+            let mut senders: HashMap<PublicKey, Sender<Result<DiscoveryItem>>> = HashMap::default();
+            let mut timeouts = JoinSet::new();
+            loop {
+                tracing::trace!(?node_addrs, "LocalNodeDiscovery Service loop tick");
+                let msg = match recv.recv_async().await {
+                    Err(err) => {
+                        tracing::error!("LocalNodeDiscovery service error: {err:?}");
+                        tracing::error!("closing LocalNodeDiscovery");
```
dignifiedquire, line 83: please import the tracing macros

iroh-net/src/discovery/local_node_discovery.rs
```diff
@@ -0,0 +1,255 @@
...
+                        timeouts.abort_all();
+                        return;
+                    }
+                    Ok(msg) => msg,
+                };
+                match msg {
```
dignifiedquire, line 89: could be a single match statement

```rust
match msg {
  Err(err) => {}
  Ok(variant..) => {}
  ...
}
```

iroh-net/src/discovery/local_node_discovery.rs
```diff
@@ -0,0 +1,255 @@
...
+        self.sender.send(Message::SendAddrs((node_id, send))).ok();
+        Some(recv.into_stream().boxed())
+    }
+
+    fn publish(&self, info: &AddrInfo) {
+        self.sender
+            .send(Message::ChangeLocalAddrs(info.clone()))
+            .ok();
+    }
+}
```
dignifiedquire, line 255: can we have a simple unit like test that at least tests this on the same machine?

iroh-cli/Cargo.toml
matheus23: From the PR description:
> 3\. For the `local-node-discovery` example, I pulled some code from `iroh-cli` into a new `iroh-progress`, so that I could re-use some progress terminal output it in the example. `iroh-progress` might not be the right name, but what do we think of a separate crate for putting in reusable UI-ish pieces. If we aren't on board with this, I can just copy and paste the logic in for now & remove the crate.

dignifiedquire: reading is hard..

dignifiedquire: I am not the biggest fan of making a crate this small tbh for this, I think I would prefer copy and paste for now, given the example is not being published

iroh-net/src/discovery/local_node_discovery.rs
```diff
@@ -0,0 +1,255 @@
...
+                direct_addresses,
+            },
+        }
+    }
+}
+
+impl Discovery for LocalNodeDiscovery {
+    fn resolve(&self, _ep: Endpoint, node_id: NodeId) -> Option<BoxStream<Result<DiscoveryItem>>> {
+        let (send, recv) = flume::bounded(20);
+        self.sender.send(Message::SendAddrs((node_id, send))).ok();
```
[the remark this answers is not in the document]
Frando: The solution is likely, I guess, either to change `resolve` to return a `Future`, or to spawn a tokio task here that sends the message with `send_async`.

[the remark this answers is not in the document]
rklaehn: When would this block? If the consumer of the stream is not sufficiently fast in consuming the events, right?

It is a stream of results. So we could check the buffer capacity. If it is less than 2, send an error. If it is less than 1 (would block), discard the event.

That way a slow receiver will get a stream

```
event
event
error(lagged)
event
```

... so they know about the issue and we are guaranteed not to block!

iroh-net/src/discovery/local_node_discovery.rs
```diff
@@ -0,0 +1,250 @@
...
+                        senders.insert(node_id, sender);
+                        let timeout_sender = task_sender.clone();
+                        timeouts.spawn(async move {
+                            tokio::time::sleep(DISCOVERY_DURATION).await;
+                            tracing::trace!(?node_id, "discovery timeout");
+                            timeout_sender
+                                .send_async(Message::Timeout(node_id))
+                                .await
+                                .ok();
+                        });
```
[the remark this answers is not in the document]
ramfox: I personally didn't want to add another `select` statement, since we seem to always have trouble with them. That was my logic for doing it this way. I can refactor tho, if folks have a differing opinion.

iroh-net/src/discovery/local_node_discovery.rs
ramfox: i've added a unique id to each request, so we can store multiple senders for each node id & added a test to make sure that we can handle multiple streams for the same node id.

iroh-net/src/discovery/local_node_discovery.rs
ramfox: I left this here in case the discovery service wanted to listen for the full 10s before it times out, in case of address changes. Sometimes discovery just pulls the first address it gets off the stream, but it can also add any subsequent addresses it gets to the underlying node map that we have in our endpoint, if it doesn't cut off the stream early.

iroh-net/src/discovery/local_node_discovery.rs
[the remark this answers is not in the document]
ramfox: It makes more sense to me to spawn a task to wait for the `send_async` to complete. Returning an error in the stream would mean the discovery service would need to start handling an error about lag (which, to me, ultimately means "wait until the next event" instead of just waiting for the next event).

iroh-net/src/discovery/local_swarm_discovery.rs
```diff
@@ -0,0 +1,342 @@
...
+    fn resolve(&self, _ep: Endpoint, node_id: NodeId) -> Option<BoxStream<Result<DiscoveryItem>>> {
+        let (send, recv) = flume::bounded(20);
+        let discovery_sender = self.sender.clone();
+        tokio::spawn(async move {
+            discovery_sender
+                .send_async(Message::SendAddrs(node_id, send))
+                .await
+                .ok();
+        });
+        Some(recv.into_stream().boxed())
```
matheus23, line 281: I would like the `tokio::spawn` side effect to be captured by the return value.

The idea being: If you call `.resolve()`, but then drop the returned stream, I would expect it to never actually kick off any work.

```suggestion
        let stream = async move {
            discovery_sender
                .send_async(Message::SendAddrs(node_id, send))
                .await?;
            Ok(recv.into_stream())
        };
        Some(stream.try_flatten_stream().boxed())
```

This way, the `send_async` is run within the returned stream. Due to the way that `try_flatten_stream` works, this also means it will surface any errors that happen from `send_async`.

ramfox: this is very elegant i'm into it!

ramfox: ahhh, except it relies on a trait from the `futures` crate, and we have been purging our codebase of that dependency.

matheus23 (approved):

divagant-martian (approved):

PR #2386, opened 2024-06-19
refactor(iroh): move code from builder to node and make things nicer

Frando:
## Description

This is a cleanup of the node and builder code.

* Move `run` and `gc_loop` from `builder.rs` to `node.rs` - it is not about building but about running.
* Improve the code flow and naming all around the builder, spawn and run functions
* Make sure we shutdown the node correctly while spawning

## Breaking Changes

* Removed `Node::controller`. Use `Node::client` instead. The former was mostly unusable anyway because we made the RPC structs private.

## Change checklist

- [x] Self-review.
- [ ] ~~Tests if relevant.~~

### DIFF
--- a/iroh/src/node/rpc.rs
+++ b/iroh/src/node/rpc.rs
@@ -29,8 +29,9 @@ use quic_rpc::{
     server::{RpcChannel, RpcServerError},
     ServiceEndpoint,
 };
+use tokio::task::JoinSet;
 use tokio_util::task::LocalPoolHandle;
-use tracing::{debug, info};
+use tracing::{debug, info, warn};
 
 use crate::client::blobs::{BlobInfo, DownloadMode, IncompleteBlobInfo, WrapOption};
 use crate::client::tags::TagInfo;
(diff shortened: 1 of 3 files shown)


dignifiedquire (approved):

rklaehn:
Very much in favour of moving core node logic into node. It always felt weird to have half the logic in the builder!

rklaehn (approved):

iroh/src/node/rpc.rs
```diff
@@ -65,234 +66,235 @@ pub(crate) struct Handler<D> {
...
-                        handler.inner.sync.doc_get_exact(req).await
-                    })
-                    .await
-                }
-                DocStartSync(msg) => {
-                    chan.rpc(msg, handler, |handler, req| async move {
-                        handler.inner.sync.doc_start_sync(req).await
-                    })
+        let handler = Self::new(inner);
+        join_set.spawn(async move {
```
rklaehn, line 83: We are adding lots of tasks to the joinset. Presumably this is so they can all be properly shut down on shutdown.

But is it OK to add lots of tasks without ever calling join_next? won't you aggregate a lot of handles to long-finished short lived tasks?

Frando: We do call join_next in the main loop, and break the main loop if a task panics - which is the main reason for the join set IMO (shutdown as well, but it won't matter much in practice because the tasks will also abort once the join set is dropped).

Issue #2378, opened 2024-06-18
Move relay HTTP handler to be served on /relay


flub:
Currently the relay HTTP handler is served on `/derp`.  We should transition that to `/relay` before we stabilise the wire protocol. But we want to coordinate this as it is a breaking wire-level change.

matheus23:
We can start supporting both options and use `/relay` in the client immediately and then later coordinate the breaking change.

DAOCUONG:
Can you explain why we remove derp option we are facing problem today our peer can not connect because iroh server is shutting down

flub:
> Can you explain why we remove derp option we are facing problem today our peer can not connect because iroh server is shutting down

Could you please open a new issue for your problem and describe exactly what happened and maybe provide logs to show how it looked as well?

We're not removing anything in this issue, we are renaming things from "derp" to "relay" and the URL endpoint is one of the things we want to rename.

PR #2387, opened 2024-06-19
feat(iroh-net): Implement `websocket` protocol upgrade in iroh-relay

### LINKED ISSUE #2370: Add websocket HTTP upgrade/framing support to iroh-relay
This is one of the pieces of work that enables iroh in the browser.

At the moment, relay clients connect via HTTPS and upgrade using the upgrade protocol `iroh derp http`.
This issue is solved once we add support for `websocket` as an HTTP upgrade and then talk websocket framing.

Ideally, this upgrade will be backwards-compatible. Clients connecting with `iroh derp http` should still be able to talk to updated relays, and websocket-based clients should be able to talk to `iroh derp http`-based clients.

Previous work happened on this branch: https://github.com/n0-computer/iroh/compare/main...feat-websockets

matheus23:
Implements the `websocket` protocol upgrade in iroh-relay and changes `iroh-net` clients to connect with `websocket` instead of `iroh derp http`.

State of this is:
- `cargo test -p iroh-net` runs green
- websocket on the relay is supported
- backwards compatibility is tested
- the client also supports connecting using websockets
  - (but we're forced to use tungstenites connection establishment, no custom `MaybeTlsStream`.)
  - It decides this via the URL scheme (`ws(s)://` vs. `http(s)://`) 

TODO:
- [x] Enumify the client so it's possible to keep using the old relay protocol
- [x] Cleanup!
- [x] Some perf TODOs in the server, less copying.
- [x] Update documentation (e.g. `local_addr` having another case being `None`)
- [x] Add metrics for websocket-accept & derp-accept counts
- [x] Snapshot tests (using `parse_hexdump`) for "new" wire format

## Description

- Supports clients connecting to iroh-relay with websockets & running the derp protocol over websocket `Binary` msgs.
- Adds support for the relay answering `Upgrade: websocket` headers instead of `Upgrade: iroh derp http` appropriately.
- Adds support for `ClientBuilder` to dial via websockets, if the relay URL is set with a `ws`/`wss` URL scheme.

## Breaking Changes

- Not in the protocol: The old HTTP upgrade & protocol should still be supported. There's tests to ensure this.
- `Client::local_addr` will now also return `None`, if one connected using websockets.
- `iroh_net::relay::http::ClientError`: Added a variant `WebsocketError` for errors when establishing a connection using websockets.

## Notes & open questions

Closes #2370 

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1507,6 +1507,7 @@ checksum = "9fc0510504f03c51ada170672ac806f1f105a88aa97a5281117e1ddc3368e51a"
 [[package]]
 name = "fastwebsockets"
 version = "0.7.2"
+source = "git+https://github.com/denoland/fastwebsockets#efc0788ec512a2634253bf5d5d41005fa76413cd"
 dependencies = [
  "base64 0.21.7",
  "bytes",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2902,12 +2902,13 @@ dependencies = [
  "tokio",
  "tokio-rustls 0.24.1",
  "tokio-rustls-acme",
- "tokio-tungstenite",
+ "tokio-tungstenite 0.23.1",
+ "tokio-tungstenite-wasm",
  "tokio-util",
  "toml",
  "tracing",
  "tracing-subscriber",
- "tungstenite",
+ "tungstenite 0.23.0",
  "url",
  "watchable",
  "webpki-roots 0.25.4",
@@ -5893,6 +5894,18 @@ dependencies = [
 
 [[package]]
 name = "tokio-tungstenite"
+version = "0.21.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c83b561d025642014097b66e6c1bb422783339e0909e4429cde4749d1990bc38"
+dependencies = [
+ "futures-util",
+ "log",
+ "tokio",
+ "tungstenite 0.21.0",
+]
+
+[[package]]
+name = "tokio-tungstenite"
 version = "0.23.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c6989540ced10490aaf14e6bad2e3d33728a2813310a0c71d1574304c49631cd"
@@ -5900,7 +5913,25 @@ dependencies = [
  "futures-util",
  "log",
  "tokio",
- "tungstenite",
+ "tungstenite 0.23.0",
+]
+
+[[package]]
+name = "tokio-tungstenite-wasm"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8e57a65894797a018b28345fa298a00c450a574aa9671e50b18218a6292a55ac"
+dependencies = [
+ "futures-channel",
+ "futures-util",
+ "http 1.1.0",
+ "httparse",
+ "js-sys",
+ "thiserror",
+ "tokio",
+ "tokio-tungstenite 0.21.0",
+ "wasm-bindgen",
+ "web-sys",
 ]
 
 [[package]]
@@ -6124,6 +6155,25 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4189890526f0168710b6ee65ceaedf1460c48a14318ceec933cb26baa492096a"
 dependencies = [
  "linked-hash-map",
+]
+
+[[package]]
+name = "tungstenite"
+version = "0.21.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9ef1a641ea34f399a848dea702823bbecfb4c486f911735368f1f137cb8257e1"
+dependencies = [
+ "byteorder",
+ "bytes",
+ "data-encoding",
+ "http 1.1.0",
+ "httparse",
+ "log",
+ "rand",
+ "sha1",
+ "thiserror",
+ "url",
+ "utf-8",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2426,13 +2426,13 @@ dependencies = [
  "hex",
  "indicatif",
  "iroh",
- "iroh-base 0.18.0",
+ "iroh-base",
  "iroh-blobs",
  "iroh-docs",
  "iroh-gossip",
  "iroh-io",
- "iroh-metrics 0.18.0",
- "iroh-net 0.18.0",
+ "iroh-metrics",
+ "iroh-net",
  "iroh-quinn",
  "iroh-test",
  "num_cpus",
@@ -2491,34 +2491,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "iroh-base"
-version = "0.18.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c13afc5b57e7204f74df49d1ce73c67d0e9d2a1b588a2985247af52c047ca964"
-dependencies = [
- "aead",
- "anyhow",
- "crypto_box",
- "data-encoding",
- "derive_more",
- "ed25519-dalek",
- "getrandom",
- "hex",
- "iroh-blake3",
- "once_cell",
- "postcard",
- "rand",
- "rand_core",
- "serde",
- "serde-error",
- "ssh-key",
- "thiserror",
- "ttl_cache",
- "url",
- "zeroize",
-]
-
-[[package]]
 name = "iroh-blake3"
 version = "1.4.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2548,11 +2520,11 @@ dependencies = [
  "hashlink",
  "hex",
  "http-body 0.4.6",
- "iroh-base 0.18.0",
+ "iroh-base",
  "iroh-blobs",
  "iroh-io",
- "iroh-metrics 0.18.0",
- "iroh-net 0.18.0",
+ "iroh-metrics",
+ "iroh-net",
  "iroh-quinn",
  "iroh-test",
  "num_cpus",
@@ -2603,7 +2575,7 @@ dependencies = [
  "human-time",
  "indicatif",
  "iroh",
- "iroh-metrics 0.18.0",
+ "iroh-metrics",
  "nix 0.27.1",
  "parking_lot",
  "pkarr",
@@ -2651,8 +2623,8 @@ dependencies = [
  "hickory-resolver",
  "hickory-server",
  "http 1.1.0",
- "iroh-metrics 0.18.0",
- "iroh-net 0.18.0",
+ "iroh-metrics",
+ "iroh-net",
  "iroh-test",
  "lru",
  "mainline",
@@ -2694,12 +2666,12 @@ dependencies = [
  "futures-lite 2.3.0",
  "futures-util",
  "hex",
- "iroh-base 0.18.0",
+ "iroh-base",
  "iroh-blake3",
  "iroh-blobs",
  "iroh-gossip",
- "iroh-metrics 0.18.0",
- "iroh-net 0.18.0",
+ "iroh-metrics",
+ "iroh-net",
  "iroh-test",
  "lru",
  "num_enum",
@@ -2734,10 +2706,10 @@ dependencies = [
  "futures-lite 2.3.0",
  "genawaiter",
  "indexmap 2.2.6",
- "iroh-base 0.18.0",
+ "iroh-base",
  "iroh-blake3",
- "iroh-metrics 0.18.0",
- "iroh-net 0.18.0",
+ "iroh-metrics",
+ "iroh-net",
  "iroh-test",
  "postcard",
  "rand",
@@ -2767,27 +2739,6 @@ dependencies = [
 [[package]]
 name = "iroh-metrics"
 version = "0.18.0"
-dependencies = [
- "anyhow",
- "erased_set",
- "http-body-util",
- "hyper 1.3.1",
- "hyper-util",
- "once_cell",
- "prometheus-client",
- "reqwest 0.12.4",
- "serde",
- "struct_iterable",
- "time",
- "tokio",
- "tracing",
-]
-
-[[package]]
-name = "iroh-metrics"
-version = "0.18.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e767734a4d6b92e0ad0c74afd38240b82a37bdfeb0bb3141245c2a707efa4097"
 dependencies = [
  "anyhow",
  "erased_set",
@@ -2836,9 +2787,8 @@ dependencies = [
  "hyper 1.3.1",
  "hyper-util",
  "igd-next",
- "iroh-base 0.18.0",
- "iroh-metrics 0.18.0",
- "iroh-net 0.18.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "iroh-base",
+ "iroh-metrics",
  "iroh-quinn",
  "iroh-quinn-proto",
  "iroh-quinn-udp",
@@ -2899,82 +2849,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "iroh-net"
-version = "0.18.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "76677befbb2d205510c9c18f5509157cb8d84cb345420e595f6612c9a1a6f88c"
-dependencies = [
- "aead",
- "anyhow",
- "backoff",
- "base64 0.22.1",
- "bytes",
- "der",
- "derive_more",
- "duct",
- "flume",
- "futures-buffered",
- "futures-concurrency",
- "futures-lite 2.3.0",
- "futures-sink",
- "futures-util",
- "governor",
- "hex",
- "hickory-proto",
- "hickory-resolver",
- "hostname",
- "http 1.1.0",
- "http-body-util",
- "hyper 1.3.1",
- "hyper-util",
- "igd-next",
- "iroh-base 0.18.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "iroh-metrics 0.18.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "iroh-quinn",
- "iroh-quinn-proto",
- "iroh-quinn-udp",
- "libc",
- "netdev",
- "netlink-packet-core",
- "netlink-packet-route",
- "netlink-sys",
- "num_enum",
- "once_cell",
- "parking_lot",
- "pin-project",
- "pkarr",
- "postcard",
- "rand",
- "rand_core",
- "rcgen 0.12.1",
- "reqwest 0.12.4",
- "ring 0.17.8",
- "rtnetlink",
- "rustls 0.21.12",
- "rustls-webpki 0.101.7",
- "serde",
- "smallvec",
- "socket2",
- "strum 0.26.2",
- "stun-rs",
- "surge-ping",
- "thiserror",
- "time",
- "tokio",
- "tokio-rustls 0.24.1",
- "tokio-rustls-acme",
- "tokio-util",
- "tracing",
- "url",
- "watchable",
- "webpki-roots 0.25.4",
- "windows 0.51.1",
- "wmi",
- "x509-parser 0.15.1",
- "z32",
-]
-
-[[package]]
 name = "iroh-net-bench"
 version = "0.18.0"
 dependencies = [
@@ -2982,7 +2856,7 @@ dependencies = [
  "bytes",
  "clap",
  "hdrhistogram",
- "iroh-net 0.18.0",
+ "iroh-net",
  "quinn",
  "rcgen 0.11.3",
  "rustls 0.21.12",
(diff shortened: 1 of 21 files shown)


iroh-net/src/relay/client.rs
```diff
@@ -1,17 +1,21 @@
...
 use anyhow::{anyhow, bail, ensure, Result};
 use bytes::{Bytes, BytesMut};
 use fastwebsockets::{WebSocketError, WebSocketRead, WebSocketWrite};
-use futures_lite::StreamExt;
 use futures_sink::Sink;
 use futures_util::sink::SinkExt;
-use futures_util::TryFutureExt;
-use tokio::io::AsyncWrite;
+use futures_util::stream::{SplitSink, SplitStream};
+use futures_util::{Stream, StreamExt, TryFutureExt};
```
dignifiedquire: we try to avoid these if we can, so please use `futures_lite` as much as possible

iroh-net/src/relay/client.rs
```diff
@@ -253,17 +253,87 @@ impl<W: AsyncWrite + Unpin + Send + 'static> ClientWriter<W> {
...
+fn tung_to_io_err(e: tokio_tungstenite_wasm::Error) -> std::io::Error {
+    std::io::Error::new(std::io::ErrorKind::Other, e.to_string())
+}
+
+impl Stream for RelayConnReader {
+    type Item = Result<Frame>;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        match *self {
+            Self::Relay(ref mut ws) => ws.poll_next_unpin(cx),
```
dignifiedquire, line 287: please use `Pin::new(ws).poll_next(cx)` for these, to keep in line with the rest of the code

iroh-blobs/src/downloader/test.rs
```diff
@@ -10,7 +10,7 @@ use iroh_net::key::SecretKey;
 
 use crate::{
     get::{db::BlobId, progress::TransferState},
-    util::progress::{FlumeProgressSender, IdGenerator, ProgressSender},
```
matheus23, line 13: A bunch of these things were marked as unused imports for me.
I think they're not actually unused, because the `mod tests { use super::*; ... }` below actually use them, but I think it's not a bad idea to move these imports into the test module.

matheus23:
FWIW, I'm still hacking on stuff over here. There's some module visibility changes that are probably not necessary. I'll do a pass over the diff at the end - making sure it's minimal.
Just working in the open - that's why the PRs here (and thanks for the comments already - that's why I opened it :) ).

iroh-net/src/relay/client.rs
matheus23: `futures_lite` re-exposes the same `Stream`, so I can use that, but other than that, there's no alternative to `futures_util::StreamExt`, because I need the `split` function from `futures_util`'s `StreamExt`, `futures_lite::StreamExt` doesn't have that :upside_down_face:

dignifiedquire: yeah, anything related to `Sink` is `futures` only unfortunately

dignifiedquire:
> Also: There's the cargo semver check failure. How do I fix that? Do I bump the version within the PR?

please make it a separate PR

Cargo.lock
```diff
@@ -5737,6 +5740,48 @@ dependencies = [
...
+dependencies = [
+ "futures-util",
+ "log",
+ "tokio",
+ "tungstenite 0.21.0",
+]
+
+[[package]]
+name = "tokio-tungstenite"
+version = "0.23.1"
```
dignifiedquire, line 5757: uhm, can we avoid using two different versions please?

iroh-blobs/src/downloader/test.rs
dignifiedquire: maybe you used a different rust version, surprised this wasn't shown in CI before

iroh-net/src/relay/codec.rs
```diff
@@ -264,6 +264,57 @@ impl Frame {
         }
     }
 
+    fn from_ws_vec(vec: Vec<u8>) -> anyhow::Result<Self> {
+        if vec.is_empty() {
+            bail!("error parsing relay::codec::Frame: too few bytes (0)");
+        }
+        let bytes = Bytes::from(vec);
+        let typ = FrameType::from(bytes[0]);
+        let frame = Self::from_bytes(typ, bytes.slice(1..))?;
```
dignifiedquire, line 273: could skip allocating the first byte entirely

```rust
let typ = vec[0];
let bytes = Bytes::from(&vec[1..]);
let frame = Self::from_bytes(typ, bytes.)?;
```

iroh-net/src/relay/codec.rs
```diff
@@ -264,6 +264,57 @@ impl Frame {
...
+        Ok(frame)
+    }
+
+    fn into_ws_vec(self) -> Vec<u8> {
+        let mut bytes = BytesMut::new();
+        bytes.put_u8(self.typ().into());
+        self.write_to(&mut bytes);
+        bytes.to_vec()
```
dignifiedquire, line 281: you can make this a bit nicer, by changing `fn write_to` to take `impl BufMut` instead of `BytesMut` as writer, that way you can pass in a `Vec`

iroh-net/src/relay/codec.rs
```diff
@@ -264,6 +264,57 @@ impl Frame {
...
+    }
+
+    pub fn into_ws_message(self) -> std::io::Result<tungstenite::Message> {
```
dignifiedquire, line 284: please add doc comments to public methods

iroh-net/src/relay/codec.rs
```diff
@@ -264,6 +264,57 @@ impl Frame {
...
+        Ok(tokio_tungstenite_wasm::Message::binary(self.into_ws_vec()))
+    }
+
+    pub fn from_ws_message(
+        msg: tungstenite::Result<tungstenite::Message>,
+    ) -> Option<anyhow::Result<Self>> {
+        match msg {
+            Ok(tungstenite::Message::Binary(vec)) => Some(Self::from_ws_vec(vec)),
+            Ok(msg) => {
+                tracing::warn!(?msg, "Got msg of unsupported type, skipping.");
```
dignifiedquire, line 298: maybe "websocket message"

iroh-net/src/relay/http/client.rs
```diff
@@ -580,6 +588,53 @@ impl Actor {
...
+            return Err(ClientError::Send);
+        }
+
+        trace!("connect_0 done");
+        Ok((relay_client, receiver))
+    }
+
+    async fn connect_ws(&self) -> Result<(RelayConnReader, RelayConnWriter), ClientError> {
+        let mut dial_url = (*self.url).clone();
+        dial_url.set_path("/derp");
```
dignifiedquire, line 623: should we use a different url here?

iroh-net/src/relay/http/server.rs
```diff
@@ -410,13 +461,48 @@ impl Service<Request<Incoming>> for ClientConnHandler {
...
+
+                    let Some(version) = req.headers().get("Sec-WebSocket-Version").cloned() else {
+                        warn!("missing header Sec-WebSocket-Version for websocket relay protocol");
+                        return Ok(builder
+                            .status(StatusCode::BAD_REQUEST)
+                            .body(body_empty())
+                            .expect("valid body"));
+                    };
+
+                    if version.as_bytes() != b"13" {
```
dignifiedquire, line 490: can we make this a constant

iroh-net/src/relay/iroh_relay.rs
```diff
@@ -876,6 +876,164 @@ mod tests {
...
+        if let ReceivedMessage::ReceivedPacket { source, data } = res {
+            assert_eq!(b_key, source);
+            assert_eq!(msg, data);
+        } else {
+            panic!("client_a received unexpected message {res:?}");
+        }
+    }
+
+    #[tokio::test]
+    async fn test_relay_clients_websocket_and_derp() {
```
dignifiedquire, line 957: we should also have tests for the tls variants of these two tests

matheus23: This is not easy to do - mainly because `tokio_tungstenite_wasm::WebSocketStream` cannot be configured to trust self-signed certificates (in the browser you have no control over that!).

Eventually I want to create two code paths on the client side: One that uses `tokio_tungstenite_wasm` in the browser and one that uses `fasterwebsockets` or `tokio_tungstenite` directly with us having more control over it client-side. (Or what that basically means is vendoring `tokio_tungstenite_wasm` and making the stream behind it configurable & exposing a different way to establish connections more manually).

I could do this in this PR still, but it could totally be a separate PR. LMK what you prefer.
The PR would essentially enable (1) proxy support and (2) testing WSS locally and (3) turn more tests that currently only test the derp protocol into tests that test both.

iroh-net/src/relay/http/client.rs
matheus23: I think we should look at relay paths again, and perhaps solve #2378 at the same time.
At the moment the setup is weird:
- There's a config option for `iroh-relay` to use a different path than `/derp`
- There's no way to configure the client to connect to somewhere else than `/derp`...

So at the moment the `iroh-relay` config option is available but de-facto must be set to the default value.
I think we should either make it hard-coded in the relay, or perhaps make the client configurable, or even make `RelayUrl` the source of truth for the path.

---

That all doesn't answer the Q :upside_down_face: 

I think, eventually we should just switch to `/relay` and run both the derp & ws protocols over that.
I don't see the advantage to using separate paths for them. Perhaps it makes blocking one protocol type easier or something if people have a reverse proxy configured? Is that worth it?

iroh-net/src/relay/codec.rs
matheus23: Are we sure that `Bytes::from(&vec[1..])` doesn't copy the vec contents?
I don't see how it can own the underlying bytes, if what's passed in is some slice with a lifetime. :eyes:

iroh-net/src/relay/iroh_relay.rs
dignifiedquire: lets make it a follow up then

iroh-net/src/relay/codec.rs
dignifiedquire: from vec already copies unfortunately

matheus23: I'm pretty sure it doesn't copy...
This is the implementation: https://docs.rs/bytes/latest/src/bytes/bytes.rs.html#830-862
Here's a playground where I'm comparing/printing the pointers: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a2b54d45c3e50b4e96e77ecfaf73a746

dignifiedquire: ah right `Bytes` doesn't allocate, but `BytesMut` does

dignifiedquire: (I think)

iroh-net/src/relay/iroh_relay.rs
matheus23: Filed this: https://github.com/n0-computer/iroh/issues/2418

iroh-net/src/relay/http/client.rs
dignifiedquire: > Today the actual route used in iroh-relay is governed by a config variable. 

We should probably drop this..no client works with this anyway..

iroh-net/src/relay/codec.rs
matheus23: Yep! (from a short peek at [the source](https://docs.rs/bytes/latest/src/bytes/bytes_mut.rs.html#1615-1648))

iroh-net/src/relay/http/client.rs
matheus23: Yes, agreed, that was one of the options I was trying to convey with my first reply. I failed at being concise or focused though :sweat_smile: 

I'll create another PR on top of this that does exactly that!

iroh-net/src/relay/client.rs
```diff
@@ -244,25 +249,92 @@ impl<W: AsyncWrite + Unpin + Send + 'static> ClientWriter<W> {
...
 
-/// The Builder returns a [`Client`] starts a [`ClientWriter`] run task.
+/// The Builder returns a [`Client`] and a started [`ClientWriter`] run task.
 pub struct ClientBuilder {
     secret_key: SecretKey,
-    reader: RelayReader,
-    writer: FramedWrite<MaybeTlsStreamWriter, DerpCodec>,
-    local_addr: SocketAddr,
+    reader: RelayConnReader,
+    writer: RelayConnWriter,
```
divagant-martian, line 256: Opinions on removing the `Relay` prefixes?

iroh-net/src/relay/client.rs
```diff
@@ -244,25 +249,92 @@ impl<W: AsyncWrite + Unpin + Send + 'static> ClientWriter<W> {
...
+    Ws(SplitStream<WebSocketStream>),
+}
+
+pub(crate) enum RelayConnWriter {
+    Derp(FramedWrite<MaybeTlsStreamWriter, DerpCodec>),
+    Ws(SplitSink<WebSocketStream, tokio_tungstenite_wasm::Message>),
+}
+
+fn tung_wasm_to_io_err(e: tokio_tungstenite_wasm::Error) -> std::io::Error {
+    std::io::Error::new(std::io::ErrorKind::Other, e.to_string())
```
divagant-martian, line 271: `tokio_tungstenite_wasm::Error` already has a `Io(io::Error)` variant that would be good to avoid re-wrapping as `Other`

iroh-net/src/relay/codec.rs
```diff
@@ -264,8 +264,67 @@ impl Frame {
...
+            Err(e) => Some(Err(e.into())),
+        }
+    }
+
+    /// Unwraps any binary messages received via websockets (in wasm) and parses them into `Frame`s.
+    ///
+    /// Ignores any non-binary websocket messages with a warning.
+    pub(crate) fn from_wasm_ws_message(
+        msg: tokio_tungstenite_wasm::Result<tokio_tungstenite_wasm::Message>,
+    ) -> Option<anyhow::Result<Self>> {
```
divagant-martian, line 315: Why are these four not `TryFrom` impls?

iroh-net/src/relay/codec.rs
```diff
@@ -264,8 +264,67 @@ impl Frame {
...
+                tracing::warn!(?msg, "Got websocket message of unsupported type, skipping.");
+                None
+            }
+            Err(e) => Some(Err(e.into())),
+        }
+    }
+
     /// Writes it self to the given buffer.
-    fn write_to(&self, dst: &mut BytesMut) {
+    fn write_to(&self, mut dst: impl BufMut) {
```
divagant-martian, line 327: While this work I think it's more correct to use
```suggestion
    fn write_to(&self, dst: &mut impl BufMut) {
```

iroh-net/src/relay/http/client.rs
```diff
@@ -784,10 +828,12 @@ impl Actor {
 
     fn use_https(&self) -> bool {
         // only disable https if we are explicitly dialing a http url
-        if self.url.scheme() == "http" {
-            return false;
+        #[allow(clippy::match_like_matches_macro)]
```
divagant-martian, line 831: why the preference over the suggested form?

iroh-net/src/relay/http/server.rs
```diff
@@ -54,20 +59,68 @@ fn downcast_upgrade(upgraded: Upgraded) -> Result<(MaybeTlsStream, Bytes)> {
...
-        read_buf.is_empty(),
-        "can not deal with buffered data yet: {:?}",
-        read_buf
-    );
-
-    conn_handler.accept(io).await
+/// The HTTP upgrade protocol used for relaying.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum Protocol {
+    /// Relays over the custom relaying protocol with a custom HTTP upgrade header
```
divagant-martian, line 65:
```suggestion
    /// Relays over the custom relaying protocol with a custom HTTP upgrade header.
```

iroh-net/src/relay/http/server.rs
```diff
@@ -54,20 +59,68 @@ fn downcast_upgrade(upgraded: Upgraded) -> Result<(MaybeTlsStream, Bytes)> {
...
+    Relay,
+    /// Relays over websockets
```
divagant-martian, line 67:
```suggestion
    /// Relays over websockets.
```

divagant-martian (commented):
We have been trying to move away from the `DERP` naming in general, but I guess coming up with a better name than "custom relaying protocol", which quite correctly is used here to describe what it is, requires some bikeshedding

iroh-net/src/relay/http/client.rs
matheus23: Clippy suggests this:
```rust
!matches!(self.url.scheme(), "http" | "ws")
```

I personally find it harder to figure out which value this is going to take in which case. I also skip over the first negation really easily. Would you disagree?

matheus23: Orthogonal: Actually reading over this - this reminds me I should rename this function to `use_tls`

iroh-net/src/relay/client.rs
matheus23: Hmmm. I wanted to have something in the name related to `Relay`s.
The old code had a type alias (that effectively had the same job), but it was called `RelayReader`. Are you looking for something shorter/would that be better?

iroh-net/src/relay/codec.rs
matheus23: I'm used to just not do `*From` traits except for errors.
I agree this is one of the cases where using `TryFrom` really is appropriate :+1:

matheus23: Looking at this again - the `from_ws_message` and `from_wasm_ws_message` functions actually return an `Option<Result<...>>`, so you can't implement `TryFrom` for them. We *could* write `From` functions, but IMO that's now getting really far away from how `From` and `TryFrom` are used usually.

Perhaps it makes sense to bikeshed the function names?

iroh-net/src/relay/codec.rs
matheus23: Hmm. I've thought about it a bunch and am on the fence. But I agree that it's slightly less odd to see `&mut` here, and it's a little better because you don't have to care that much about giving away ownership on the inside of the function accidentally :)

iroh-net/src/relay/client.rs
matheus23: Given nothing else is prefixed `Relay*` in this file, I agree I'm going ahead with dropping the prefix :)

iroh-net/src/relay/codec.rs
dignifiedquire: given they are fallible, maybe call them `try_from_ws_message`

matheus23: Meh - ended up moving these helpers out of `Frame` and just inline their logic into the `Stream` and `Sink` implementations.
That also uncovered some weird, unnecessary `?` from the `into_ws_message` call-site. I think the code just belongs closer to the `Stream` implementation, the logic was quite intertwined anyways (effectively some nested matches, this is now flattened).

Instead, it's only two new functions on `Frame` remaining: `Frame::encode_for_ws_msg` and `Frame::decode_from_ws_msg`.

Feels way better IMHO.

iroh-net/src/relay/codec.rs
```diff
@@ -561,6 +584,94 @@ mod tests {
...
+                85 32 c6 ab c8 38 fa cd 5e a7 89 be 0c 76 b2 92
+                03 34 03 9b fa 8b 3d 36 8d 61 88 e7 7b 22 f2 92
+                ab 37 43 5d a8 de 0b c8 cb 84 e2 88 f4 e7 3b 35
+                82 a5 27 31 e9 ff 98 65 46 5c 87 e0 5e 8d 42 7d
+                f4 22 bb 6e 85 e1 c0 5f 6f 74 98 37 ba a4 a5 c7
+                eb a3 23 0d 77 56 99 10 43 0e 03",
+            ),
+            (
+                Frame::Health {
+                    problem: "Hello? Yes this is dog.".into(),
```
ramfox, line 614: 😂

ramfox (approved):
This looks great. Wooo!

My only note, which diva mentioned, is I don't love doubling down on the name "derp", since we have been trying to step away from calling the protocol "derp" since we've technically broken away from it multiple times now 😂 

However, I think this is a problem best left for our "wire protocol" review in anticipation of our mythical "one final breaking change" plan.  Gotta love that we still ping the `/derp` endpoint and use 'iroh derp http' in our headers 🤣

So I say merge without worrying about the derp naming, and leave that to a more general "relay protocol" follow up later.

iroh-net/src/relay/codec.rs
matheus23: :grin:

dignifiedquire (approved):

PR #2390, opened 2024-06-20
feat(iroh): allow to disable docs engine completely

Frando:
## Description

Make the docs engine optional on the iroh node.

* Add `Builder::disable_docs()` to disable the docs engine completely
* If called, the docs engine will not be spawned and the docs protocol will not be registered. Incoming docs connnections will be dropped, and all docs-related RPC calls will return an error "docs are disabled".

## Breaking Changes

* `iroh::node::Builder::with_db_and_store` now takes a `DocsStorage` enum instead of a `iroh_docs::store::Store`.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/node/builder.rs
+++ b/iroh/src/node/builder.rs
@@ -94,7 +94,7 @@ where
     gc_policy: GcPolicy,
     dns_resolver: Option<DnsResolver>,
     node_discovery: DiscoveryConfig,
-    docs_store: Option<DocsStorage>,
+    docs_storage: Option<DocsStorage>,
     #[cfg(any(test, feature = "test-utils"))]
     insecure_skip_relay_cert_verify: bool,
     /// Callback to register when a gc loop is done
@@ -155,7 +155,7 @@ impl Default for Builder<iroh_blobs::store::mem::Store> {
             dns_resolver: None,
             rpc_endpoint: Default::default(),
             gc_policy: GcPolicy::Disabled,
-            docs_store: Some(DocsStorage::Memory),
+            docs_storage: Some(DocsStorage::Memory),
             node_discovery: Default::default(),
             #[cfg(any(test, feature = "test-utils"))]
             insecure_skip_relay_cert_verify: false,
@@ -168,7 +168,7 @@ impl<D: Map> Builder<D> {
     /// Creates a new builder for [`Node`] using the given databases.
     pub fn with_db_and_store(
         blobs_store: D,
-        docs_store: DocsStorage,
+        docs_storage: DocsStorage,
         storage: StorageConfig,
     ) -> Self {
         Self {
@@ -181,7 +181,7 @@ impl<D: Map> Builder<D> {
             dns_resolver: None,
             rpc_endpoint: Default::default(),
             gc_policy: GcPolicy::Disabled,
-            docs_store: Some(docs_store),
+            docs_storage: Some(docs_storage),
             node_discovery: Default::default(),
             #[cfg(any(test, feature = "test-utils"))]
             insecure_skip_relay_cert_verify: false,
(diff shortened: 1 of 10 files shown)


iroh/src/node/builder.rs
```diff
@@ -85,7 +94,7 @@ where
     gc_policy: GcPolicy,
     dns_resolver: Option<DnsResolver>,
     node_discovery: DiscoveryConfig,
-    docs_store: iroh_docs::store::Store,
+    docs_store: Option<DocsStorage>,
```
rklaehn, line 97: This just says where the docs are stored, it is no longer a store.

iroh/src/node/rpc.rs
```diff
@@ -131,92 +151,115 @@ impl<D: BaoStore> Handler<D> {
             BlobReadAt(msg) => chan.server_streaming(msg, self, Self::blob_read_at).await,
             BlobAddStream(msg) => chan.bidi_streaming(msg, self, Self::blob_add_stream).await,
             BlobAddStreamUpdate(_msg) => Err(RpcServerError::UnexpectedUpdateMessage),
+
             AuthorList(msg) => {
                 chan.server_streaming(msg, self, |handler, req| {
-                    handler.inner.docs.author_list(req)
+                    let (tx, rx) = flume::bounded(ITER_CHANNEL_CAP);
```
rklaehn, line 157: So this used to return a stream, and now we give it a flume channel?

Can't we avoid this additional flume buffer by having something like

```rust
if docs.enabled {
  docs.author_list(...)
} else {
  stream::single(Err(anyhow!("docs are disabled")))
}
```

Somewhere in the various futures utils there used to be .left_stream() / .right_stream() to unify the types. Or box them in the worst case. Still better than another flume channel I think.

iroh/src/node/rpc.rs
```diff
@@ -131,92 +151,115 @@ impl<D: BaoStore> Handler<D> {
...
+                    handler.with_docs(|docs| async move { docs.doc_status(req).await })
                 })
                 .await
             }
             DocList(msg) => {
-                chan.server_streaming(msg, self, |handler, req| handler.inner.docs.doc_list(req))
-                    .await
+                chan.server_streaming(msg, self, |handler, req| {
+                    let (tx, rx) = flume::bounded(ITER_CHANNEL_CAP);
+                    if let Some(docs) = handler.docs() {
```
rklaehn, line 228: see above

iroh/src/node/rpc/docs.rs
```diff
@@ -249,9 +246,9 @@ impl DocsEngine {
     pub fn doc_get_many(
         &self,
         req: DocGetManyRequest,
-    ) -> impl Stream<Item = RpcResult<DocGetManyResponse>> {
+        tx: flume::Sender<anyhow::Result<SignedEntry>>,
+    ) {
         let DocGetManyRequest { doc_id, query } = req;
-        let (tx, rx) = flume::bounded(ITER_CHANNEL_CAP);
```
rklaehn, line 254: Ah, you already had the flume channel and just moved it outside?

It seems that the signature of the quic-rpc handler is not ideal, or at least not always.

Maybe it would be better if instead of returning a stream you would get a sink. Would that help?

iroh/src/node/rpc.rs
Frando: We can't unless we box the streams, because otherwise the two branches have different types.
Maybe the boxing is fine. I am never sure when to try to avoid boxing and when not...?

iroh/src/node/builder.rs
dignifiedquire, line 97: why not just add a `Disabled` variant to the enum, this way you don't need to interpret the `Option`

iroh/src/node/rpc.rs
dignifiedquire: https://docs.rs/tokio-util/latest/tokio_util/either/enum.Either.html

Frando: :+1: switched to this, much better

iroh/src/node/builder.rs
Frando: changed to that!

iroh/src/node/rpc/docs.rs
Frando: Switched to using `Either`

iroh/src/node/builder.rs
Frando: changed!

dignifiedquire (approved):

PR #2391, opened 2024-06-20
feat(iroh): allow setting the logging directory via config file

divagant-martian:
## Description

Adds the options to set the directory where logs will be created in the config file.

**Example:**
```toml
[file_logs]
rust_log = "iroh=info"
rotation = "daily"
max_files = 1
dir = "/home/me/my_logs"
```

It is recommended that the path is absolute, otherwise it will be interpreted as relative _to the execution dir_. If this option is not set, the default `<IROH_DATA_DIR>/logs` will be used.

## Breaking Changes

N/a

## Notes & open questions

N/a

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


rklaehn (approved):
Looks pretty straightforward.

We should have some tests for config file parsing where we could add this, but maybe not in this PR. Would be good if this makes it into the next release since this was a user request.

flub (approved):

PR #2398, opened 2024-06-23
feat(iroh)!: expand ability to connect to RPC

dignifiedquire:
Allows for easier connectivity of RPC, when not having access to the folder, e.g. in a docker setup

## Breaking Changes

- renamed `iroh::client::Iroh::connect` -> `iroh::client::Iroh::connect_path` 
- added `iroh::client::Iroh::connect_addr`
- added `rpc_port` field to `iroh::client::NodeStatus` 

## Notes & open questions

There might be nicer ways, but this works for now

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/node.rs
+++ b/iroh/src/node.rs
@@ -58,8 +58,6 @@ pub struct Node<D> {
 #[derive(derive_more::Debug)]
 struct NodeInner<D> {
     db: D,
-    #[allow(dead_code)]
-    storage: StorageConfig,
     rpc_port: Option<u16>,
     docs: Option<DocsEngine>,
     endpoint: Endpoint,
--- a/iroh/src/node.rs
+++ b/iroh/src/node.rs
@@ -58,6 +58,9 @@ pub struct Node<D> {
 #[derive(derive_more::Debug)]
 struct NodeInner<D> {
     db: D,
+    #[allow(dead_code)]
+    storage: StorageConfig,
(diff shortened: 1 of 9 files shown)


iroh/src/node/builder.rs
```diff
@@ -256,7 +260,11 @@ where
     }
 
     /// Configure rpc endpoint, changing the type of the builder to the new endpoint type.
-    pub fn rpc_endpoint<E2: ServiceEndpoint<RpcService>>(self, value: E2) -> Builder<D, E2> {
+    pub fn rpc_endpoint<E2: ServiceEndpoint<RpcService>>(
```
rklaehn, line 263: Not for this PR probably, but I guess we don't need this whole type parameter anymore on the builder due to having the ability to box the endpoint.

So far we don't make use of this on the node side but just on the client side, but maybe something to do for the next release...

We need to get rid of type parameters to make room for new type parameters... 😁

iroh/src/node.rs
```diff
@@ -58,6 +58,9 @@ pub struct Node<D> {
 #[derive(derive_more::Debug)]
 struct NodeInner<D> {
     db: D,
+    #[allow(dead_code)]
+    storage: StorageConfig,
```
rklaehn, line 62: What's this for if it is not used?

dignifiedquire: removed

PR #2400, opened 2024-06-24
fix(iroh): do not panic on blobs IO error

Frando:
## Description

* We `unwrap`ed an `io::Result` in the RPC handler for `BlobsReadAtRequest`. Changed to return an error instead.
* In the blobs client, report the error cases that can occur for reading a blob differently, to help debugging.

## Change checklist

- [x] Self-review.
- [x] ~~Tests if relevant.~~

### DIFF
--- a/iroh/src/client/blobs.rs
+++ b/iroh/src/client/blobs.rs
@@ -798,7 +798,8 @@ impl Reader {
         let (size, is_complete) = match stream.next().await {
             Some(Ok(BlobReadAtResponse::Entry { size, is_complete })) => (size, is_complete),
             Some(Err(err)) => return Err(err),
-            None | Some(Ok(_)) => return Err(anyhow!("Expected header frame")),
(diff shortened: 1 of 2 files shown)


iroh/src/client/blobs.rs
```diff
@@ -798,7 +798,8 @@ impl Reader {
         let (size, is_complete) = match stream.next().await {
             Some(Ok(BlobReadAtResponse::Entry { size, is_complete })) => (size, is_complete),
             Some(Err(err)) => return Err(err),
-            None | Some(Ok(_)) => return Err(anyhow!("Expected header frame")),
+            Some(Ok(_)) => return Err(anyhow!("Expected header frame, but got data frame")),
```
dignifiedquire, line 801: maybe even log the frame?

Frando: there's only 2, that's why I inlined it in the error

dignifiedquire (approved):

PR #2403, opened 2024-06-24
chore: better debug logs for read_at failure

Frando:
## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


rklaehn:
We probably want to close this now that we found the bug.

PR #2404, opened 2024-06-24
feat: docker images for iroh

Arqu:
## Description

- Enable binding of the RPC endpoint to arbitrary addresses
- Add Dockerfile for iroh

## Breaking Changes

- changed `iroh::node::Node::my_rpc_port`  is now `my_rpc_addr` and returns `Option<SocketAddr>`
- `iroh::node::Builder::rpc_endpoint` takes `Option<SocketAddr> instead of `Option<u16>`
- added `iroh::node::Builder::enable_rpc_with_addr` allowing to specify on which address + port to bind to

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.


Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+COPY --from=rust_builder /iroh/target/release/iroh /iroh
+
+RUN chmod +x /iroh
+
+WORKDIR /
+
+# Use nonroot (unprivileged) user
+# USER nonroot
+# expose the default ports
+EXPOSE 4433 8000 9090
```
Arqu, line 37: Adjust this line for the ports you need.

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+ENTRYPOINT ["./iroh start"]
```
Arqu, line 38: Adjust this line for alternate configs or start cmds.

Arqu (commented):
Also to make this easy to connect to other services on the local host, best bet is to run in network mode host.

Dockerfile
dignifiedquire: needs `4919` which is the default RPC port, but we also need to make sure we allow to bind to more than just `127.0.0.1` for rpc, otherwise it won't work

dignifiedquire:
@Arqu once we merge this, can we automate it, to publish this to dockerhub on release?

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+WORKDIR /iroh
+
+# copy entire workspace
+COPY . .
+
+RUN cargo clean
+RUN cargo build --bin iroh --release --all-features
+
+### Final Image
+FROM ubuntu:latest as iroh
```
dignifiedquire, line 20: can we switch this to alpine?

dignifiedquire:
@Arqu we should have one minimal test in CI running the docker image and calling `iroh node status` or sth, to make sure the basic interactions always work

Arqu:
> @Arqu we should have one minimal test in CI running the docker image and calling `iroh node status` or sth, to make sure the basic interactions always work

We got a node stats test now too.

iroh-cli/src/commands.rs
```diff
@@ -40,6 +41,12 @@ pub(crate) struct Cli {
     /// Port to serve metrics on. Disabled by default.
     #[clap(long)]
     pub(crate) metrics_port: Option<MetricsPort>,
+
+    /// Address to serve RPC on.
+    ///
+    /// This overrides the address found in the IPFS_PATH.
```
rklaehn, line 47: What `IPFS_PATH`?

iroh/src/node/builder.rs
```diff
@@ -316,8 +320,10 @@ where
     }
 
     /// Configure the default iroh rpc endpoint.
-    pub async fn enable_rpc(self) -> Result<Builder<D>> {
-        let (ep, actual_rpc_port) = make_rpc_endpoint(&self.secret_key, DEFAULT_RPC_PORT)?;
+    pub async fn enable_rpc(self, rpc_addr: Option<SocketAddr>) -> Result<Builder<D>> {
```
rklaehn, line 323: Can we keep the default builder API simple? I would guess that 99% of all users will want to enable rpc with the default value here.

So either `.enable_rpc()` or `.enable_custom_rpc(AddrOpt)` or maybe `enable_rpc_with_opts(RpcOpts)` in case we expect more options in the future?

iroh/src/node/builder.rs
```diff
@@ -804,20 +812,18 @@ fn make_rpc_endpoint(
...
-    let rpc_quinn_endpoint = quinn::Endpoint::server(server_config.clone(), rpc_addr.into());
+    let rpc_quinn_endpoint = quinn::Endpoint::server(server_config.clone(), rpc_addr);
     let rpc_quinn_endpoint = match rpc_quinn_endpoint {
         Ok(ep) => ep,
         Err(err) => {
             if err.kind() == std::io::ErrorKind::AddrInUse {
                 tracing::warn!(
                     "RPC port {} already in use, switching to random port",
-                    rpc_port
+                    rpc_addr.port(),
```
rklaehn, line 822: We should probably log the entire addr.

iroh-cli/src/commands.rs
dignifiedquire: great question..

iroh/src/node/builder.rs
dignifiedquire: fixed

iroh/src/node/builder.rs
dignifiedquire: fixed

build_docker.sh
```diff
@@ -0,0 +1,2 @@
+docker buildx build -f docker/Dockerfile --target iroh --platform linux/arm64/v8 --tag n0computer/iroh:latest .
+# docker buildx build -f docker/Dockerfile --target iroh --platform linux/amd64 --tag n0computer/iroh:latest .
```
dignifiedquire, line 2: what about this?

docker/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+A set of docker images provided to easily run iroh in a containerized environment.
+Features `iroh`, `iroh-relay` and `iroh-dns-server`.
+
+The provided `Docker` files are intended for CI use but can be also manually built.
+
+## Building
+
+- All commands are run from the root folder
+- If you're on macOS run `docker buildx build -f docker/Dockerfile --target iroh --platform linux/arm64/v8 --tag n0computer/iroh:latest .`
```
dignifiedquire, line 13: what does `buildx` mean?

build_docker.sh
Arqu: We should drop this, stuff is already there in the readme.

docker/README.md
Arqu: Buildx is an official extension by docker for the build system for images. Gives you the ability to specify platforms and some other nifty stuff.

PR #2405, opened 2024-06-25
refactor(iroh)!: Eliminate the type parameter for the rpc service type

rklaehn:
## Description

Eliminate the type parameter for the rpc service type in the builder.

We are going to need more type parameters if we want to make rpc generic and extendable, so we need to make some room... :-)

~~todo: needs new minor version of quic-rpc to make boxed(...) public~~

## Breaking Changes

- Builder loses the `E` type parameter
- ProtocolBuilder loses the `E` type parameter
- rpc_endpoint takes a `boxed::ServerEndpoint`

## Notes & open questions

Note: we now require the user to box the rpc_endpoint if they want to do a custom one. But since that is most likely a non-mem version it should not have a performance impact. If you want a mem rpc controller, we already have that...

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/node/builder.rs
+++ b/iroh/src/node/builder.rs
@@ -53,6 +53,11 @@ const DEFAULT_GC_INTERVAL: Duration = Duration::from_secs(60 * 5);
 const MAX_CONNECTIONS: u32 = 1024;
 const MAX_STREAMS: u64 = 10;
 
+type BoxedServerEndpoint = quic_rpc::transport::boxed::ServerEndpoint<
+    crate::rpc_protocol::Request,
+    crate::rpc_protocol::Response,
+>;
+
 /// Storage backend for documents.
 #[derive(Debug, Clone)]
 pub enum DocsStorage {
@@ -85,10 +90,7 @@ where
     storage: StorageConfig,
     bind_port: Option<u16>,
     secret_key: SecretKey,
-    rpc_endpoint: quic_rpc::transport::boxed::ServerEndpoint<
-        crate::rpc_protocol::Request,
-        crate::rpc_protocol::Response,
-    >,
+    rpc_endpoint: BoxedServerEndpoint,
     blobs_store: D,
     keylog: bool,
     relay_mode: RelayMode,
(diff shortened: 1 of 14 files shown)


iroh/src/node/builder.rs
```diff
@@ -56,6 +53,11 @@ const DEFAULT_GC_INTERVAL: Duration = Duration::from_secs(60 * 5);
 const MAX_CONNECTIONS: u32 = 1024;
 const MAX_STREAMS: u64 = 10;
 
+type BoxedServerEndpoint = quic_rpc::transport::boxed::ServerEndpoint<
```
rklaehn, line 56: Maybe it should be ServiceEndpoint<S> instead of ServerEndpoint<In, Out>. But that would require a new version of quic-rpc, so not today...

dignifiedquire: didn't @Frando PR that at some point?

iroh/src/node/builder.rs
```diff
@@ -145,6 +146,40 @@ impl From<Box<ConcurrentDiscovery>> for DiscoveryConfig {
...
+        &self,
+    ) -> quic_rpc::transport::boxed::AcceptFuture<
+        crate::rpc_protocol::Request,
+        crate::rpc_protocol::Response,
+    > {
+        quic_rpc::transport::boxed::AcceptFuture::boxed(futures_lite::future::pending())
+    }
+
+    fn local_addr(&self) -> &[quic_rpc::transport::LocalAddr] {
+        &[]
```
dignifiedquire, line 175: Oo, that's problematic, no?

rklaehn: It's just the dummy endpoint. What else would you report here?

iroh/src/node/builder.rs
rklaehn: We got a ton of places in quic-rpc where we switched from Channel<In, Out> to ServiceChannel<S>, but not all places.

iroh/src/node/builder.rs
dignifiedquire: oh lol, didn't realize this was the dummy 😅

iroh/src/node/builder.rs
```diff
@@ -146,6 +147,40 @@ impl From<Box<ConcurrentDiscovery>> for DiscoveryConfig {
     }
 }
 
+/// A server endpoint that does nothing. Accept will never resolve.
```
dignifiedquire, line 150: why not use an `Option<BoxedServerEndpoint>` instead?

rklaehn: It complicates the select! statement where you accept on it. This is basically just a boxed future::pending that does no harm whatsoever when disabled.

dignifiedquire (approved):

PR #2408, opened 2024-06-25
fix(iroh): use tokio transport for quic rpc

dignifiedquire:
Depends on https://github.com/n0-computer/quic-rpc/pull/86


rklaehn:
Turns out that cancel safety was the issue, but it was cancel safety of quic-rpc accept, not flume. So this is not needed.

PR #2409, opened 2024-06-25
fix(iroh): increase buffer size for rpc controller

dignifiedquire:


rklaehn:
cherry-picked into https://github.com/n0-computer/iroh/pull/2416

PR #2417, opened 2024-06-26
feat(iroh-net)!: Add PkarrResolver and publish direct addresses in PkarrPublisher when relay is disabled

### LINKED ISSUE #2412: n0 dns server doesn't return relay records
records can be resolved with `https://app.pkarr.org` when pointing it to `https://dns.iroh.link/pkarr`. a bit surprising is that the relayer doesn't publish to the dht, so you can't resolve it with a different relayer. Seems like this is intended after discussing with @rklaehn . however the DnsResolver (the only one currently provided by default in iroh-net) fails to return any records. trying to query it manually seems to show that the problem is with the n0 dns server.

```
dig _iroh.7qy7is9kco9oygshi4ocmbw5ibogm4jme659bdgkrbthkwzxy5yy.dns.iroh.link. txt

; <<>> DiG 9.18.27 <<>> _iroh.7qy7is9kco9oygshi4ocmbw5ibogm4jme659bdgkrbthkwzxy5yy.dns.iroh.link. txt
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 42232
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 65494
;; QUESTION SECTION:
;_iroh.7qy7is9kco9oygshi4ocmbw5ibogm4jme659bdgkrbthkwzxy5yy.dns.iroh.link. IN TXT

;; AUTHORITY SECTION:
dns.iroh.link.		7007	IN	SOA	testdns.iroh.link. hostmaster.iroh.link. 1 7200 900 1209600 86400

;; Query time: 3 msec
;; SERVER: 127.0.0.53#53(127.0.0.53) (UDP)
;; WHEN: Wed Jun 26 12:30:40 CEST 2024
;; MSG SIZE  rcvd: 156
```

### LINKED ISSUE #2413: PkarrPublisher doesn't publish the direct addresses when relaying is disabled
If you're running a server with a publicly reachable IP, disabling relaying (`RelayMode::Disabled`) is a sensible thing to do. For discovery to work, the direct addresses should be published in that case.

dvc94ch:
Tested it, works fine and fixes all my issues.

Closes #2412 and closes #2413 

### Features
- *(iroh-net)* `PkarrResolver` adds an alternative discovery mechanism to the existing `DnsResolver`.

### Bug Fixes
- *(iroh-net)* Direct addresses are published when relay mode is set to disabled.

### Breaking changes
- *(iroh-net)* pkarr_publish module was renamed to pkarr.

### DIFF
--- a/iroh-dns-server/Cargo.toml
+++ b/iroh-dns-server/Cargo.toml
@@ -27,7 +27,7 @@ http = "1.0.0"
 iroh-metrics = { version = "0.19.0", path = "../iroh-metrics" }
 lru = "0.12.3"
 parking_lot = "0.12.1"
-pkarr = { version = "2.0.0", features = [ "async", "rand", "relay", "dht"], default-features = false }
+pkarr = { version = "2.0.0", features = [ "async", "relay", "dht"], default-features = false }
 rcgen = "0.12.1"
 redb = "2.0.0"
 regex = "1.10.3"
--- a/iroh-dns-server/Cargo.toml
+++ b/iroh-dns-server/Cargo.toml
@@ -26,6 +26,7 @@ hickory-server = { version = "0.24.0", features = ["dns-over-rustls"] }
 http = "1.0.0"
 iroh-metrics = { version = "0.19.0", path = "../iroh-metrics" }
 lru = "0.12.3"
+mainline = "2.0.1"
 parking_lot = "0.12.1"
 pkarr = { version = "2.0.0", features = [ "async", "relay", "dht"], default-features = false }
 rcgen = "0.12.1"
@@ -54,7 +55,7 @@ z32 = "1.1.1"
 hickory-resolver = "0.24.0"
 iroh-net = { version = "0.19.0", path = "../iroh-net" }
 iroh-test = { path = "../iroh-test" }
-mainline = "<1.5.0"
+pkarr = { version = "2.0.0", features = ["rand"] }
 
 [package.metadata.docs.rs]
 all-features = true
--- a/iroh-dns-server/Cargo.toml
+++ b/iroh-dns-server/Cargo.toml
@@ -27,7 +27,7 @@ http = "1.0.0"
 iroh-metrics = { version = "0.19.0", path = "../iroh-metrics" }
 lru = "0.12.3"
 parking_lot = "0.12.1"
-pkarr = { version = "1.1.4", features = [ "async", "relay", "dht"], default-features = false }
+pkarr = { version = "2.0.0", features = [ "async", "rand", "relay", "dht"], default-features = false }
 rcgen = "0.12.1"
 redb = "2.0.0"
 regex = "1.10.3"
--- a/iroh-net/Cargo.toml
+++ b/iroh-net/Cargo.toml
@@ -47,7 +47,7 @@ num_enum = "0.7"
 once_cell = "1.18.0"
 parking_lot = "0.12.1"
 pin-project = "1"
-pkarr = { version = "1.1.4", default-features = false, features = ["async", "relay"] }
+pkarr = { version = "2.0.0", default-features = false, features = ["async", "rand", "relay"] }
 postcard = { version = "1", default-features = false, features = ["alloc", "use-std", "experimental-derive"] }
 quinn = { package = "iroh-quinn", version = "0.10.5" }
 quinn-proto = { package = "iroh-quinn-proto", version = "0.10.8" }
--- a/iroh-net/Cargo.toml
+++ b/iroh-net/Cargo.toml
@@ -47,7 +47,7 @@ num_enum = "0.7"
 once_cell = "1.18.0"
 parking_lot = "0.12.1"
 pin-project = "1"
-pkarr = { version = "2.0.0", default-features = false, features = ["async", "rand", "relay"] }
+pkarr = { version = "2.0.0", default-features = false, features = ["async", "relay"] }
 postcard = { version = "1", default-features = false, features = ["alloc", "use-std", "experimental-derive"] }
 quinn = { package = "iroh-quinn", version = "0.10.5" }
 quinn-proto = { package = "iroh-quinn-proto", version = "0.10.8" }
--- a/iroh-net/src/discovery/pkarr.rs
+++ b/iroh-net/src/discovery/pkarr.rs
@@ -1,8 +1,4 @@
-//! A discovery service which publishes node information to a [Pkarr] relay.
-//!
-//! This service only implements the [`Discovery::publish`] method and does not provide discovery.
-//! It encodes the node information into a DNS packet in the format resolvable by the
-//! [`super::dns::DnsDiscovery`].
+//! A discovery service which publishes and resolves node information to a [pkarr] relay.
 //!
 //! [pkarr]: https://pkarr.org
 
@@ -29,13 +25,19 @@ use crate::{
 /// The pkarr relay run by n0.
 pub const N0_DNS_PKARR_RELAY: &str = "https://dns.iroh.link/pkarr";
 
-/// Default TTL for the records in the pkarr signed packet
+/// Default TTL for the records in the pkarr signed packet. TTL tells DNS caches
+/// how long to store a record. It is ignored by the iroh-dns-server as the home
+/// server keeps the records for the domain. When using the pkarr relay no DNS is
+/// involved and the setting is ignored.
 pub const DEFAULT_PKARR_TTL: u32 = 30;
 
 /// Interval in which we will republish our node info even if unchanged: 5 minutes.
 pub const DEFAULT_REPUBLISH_INTERVAL: Duration = Duration::from_secs(60 * 5);
 
 /// Publish node info to a pkarr relay.
+///
+/// Publishes either the relay url if the relay is enabled or the direct addresses
+/// if the relay is disabled.
 #[derive(derive_more::Debug, Clone)]
 pub struct PkarrPublisher {
     node_id: NodeId,
--- a/iroh-net/src/discovery/pkarr.rs
+++ b/iroh-net/src/discovery/pkarr.rs
@@ -9,6 +9,7 @@
 use std::sync::Arc;
 
 use anyhow::{anyhow, bail, Result};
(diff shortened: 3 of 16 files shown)


iroh-net/Cargo.toml
```diff
@@ -47,7 +47,7 @@ num_enum = "0.7"
 once_cell = "1.18.0"
 parking_lot = "0.12.1"
 pin-project = "1"
-pkarr = { version = "1.1.4", default-features = false, features = ["async", "relay"] }
+pkarr = { version = "2.0.0", default-features = false, features = ["async", "rand", "relay"] }
```
dvc94ch, line 50: updating was necessary to convert from [u8; 32] to PublicKey. adding "rand" was necessary because the crate doesn't build otherwise

rklaehn: Can you open an issue at the pkarr repo? We should not have to fix the crate by adding features...

dvc94ch: opened a PR: https://github.com/Nuhvi/pkarr/pull/66

iroh-net/src/discovery/pkarr.rs
```diff
@@ -174,6 +181,48 @@ impl PublisherService {
...
+        Self::new(pkarr_relay)
+    }
+}
+
+impl Discovery for PkarrResolver {
+    fn resolve(
+        &self,
+        _ep: Endpoint,
+        node_id: NodeId,
+    ) -> Option<BoxStream<'static, Result<DiscoveryItem>>> {
```
dvc94ch, line 210: why I had to add 'static here explicitly, I have no idea:
```
cargo check
    Blocking waiting for file lock on build directory
    Checking iroh-net v0.18.0 (/home/dvc/iroh/iroh-net)
error: `impl` item signature doesn't match `trait` item signature
   --> iroh-net/src/discovery/pkarr.rs:206:5
    |
206 |       fn resolve(&self, _ep: Endpoint, node_id: NodeId) -> Option<BoxStream<Result<DiscoveryItem>>> {
    |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ found `fn(&'1 PkarrResolver, endpoint::Endpoint, iroh_base::key::PublicKey) -> std::option::Option<Pin<Box<(dyn futures_lite::Stream<Item = Result<DiscoveryItem, anyhow::Error>> + std::marker::Send + '1)>>>`
    |
   ::: iroh-net/src/discovery.rs:43:5
    |
43  | /     fn resolve(
44  | |         &self,
45  | |         _endpoint: Endpoint,
46  | |         _node_id: NodeId,
47  | |     ) -> Option<BoxStream<Result<DiscoveryItem>>> {
    | |_________________________________________________- expected `fn(&'1 PkarrResolver, endpoint::Endpoint, iroh_base::key::PublicKey) -> std::option::Option<Pin<Box<(dyn futures_lite::Stream<Item = Result<DiscoveryItem, anyhow::Error>> + std::marker::Send + 'static)>>>`
    |
    = note: expected signature `fn(&'1 PkarrResolver, endpoint::Endpoint, iroh_base::key::PublicKey) -> std::option::Option<Pin<Box<(dyn futures_lite::Stream<Item = Result<DiscoveryItem, anyhow::Error>> + std::marker::Send + 'static)>>>`
               found signature `fn(&'1 PkarrResolver, endpoint::Endpoint, iroh_base::key::PublicKey) -> std::option::Option<Pin<Box<(dyn futures_lite::Stream<Item = Result<DiscoveryItem, anyhow::Error>> + std::marker::Send + '1)>>>`
    = help: the lifetime requirements from the `impl` do not correspond to the requirements in the `trait`
    = help: verify the lifetime relationships in the `trait` and `impl` between the `self` argument, the other inputs and its output

error: could not compile `iroh-net` (lib) due to 1 previous error
```
```
rustc --version
rustc 1.78.0 (9b00956e5 2024-04-29)
```

iroh-net/src/discovery/pkarr.rs
```diff
@@ -174,6 +181,48 @@ impl PublisherService {
     }
 }
 
+/// Resolve node info using a pkarr relay.
+#[derive(derive_more::Debug, Clone)]
+pub struct PkarrResolver {
```
rklaehn, line 186: So this is for when you want to resolve directly via the relay? Yeah, makes sense to have this.

iroh-net/src/discovery/pkarr.rs
rklaehn: 🤷

iroh-dns-server/Cargo.toml
```diff
@@ -54,7 +55,7 @@ z32 = "1.1.1"
 hickory-resolver = "0.24.0"
 iroh-net = { version = "0.18.0", path = "../iroh-net" }
 iroh-test = { path = "../iroh-test" }
-mainline = "<1.5.0"
+pkarr = { version = "2.0.0", features = ["rand"] }
```
rklaehn, line 58: I think @nuhvi has published a new version that does not need this anymore.

iroh-net/src/discovery/pkarr.rs
```diff
@@ -174,6 +181,48 @@ impl PublisherService {
     }
 }
 
+/// Resolve node info using a pkarr relay.
```
Frando, line 184: Would be good to expand the docs a little here, maybe with a link to pkarr or somewhere that explains what a pkarr relay is.

Frando (commented):
LGTM to add this! Would be great to expand the docs a little,, ideally detailing when you'd want to use this versus the DNS resolution.

In a future PR, we could then expand the PkarrPublisher and PkarrResolver to support mainline dircectly, optionally. If pkarr v2 fixed the bug that it would always bind a DHT socket even if unconfigured, not sure if that is the case, would have to check.

iroh-dns-server/Cargo.toml
dvc94ch: No, you need this in the tests, since you generate a secret key using pkarr

Nuhvi:
> If pkarr v2 fixed the bug that it would always bind a DHT socket even if unconfigured, not sure if that is the case, would have to check.

If you are only using the relay client, you won't bind any UDP socket https://github.com/Nuhvi/pkarr/blob/main/pkarr/src/relay_client.rs

rklaehn:
Ok, I have taken a quick look at the pkarr crate. It is using a sync http client (ureq), but it is spawning a thread for each request, so it does not block the tokio runtime.

So while this is probably less than optimal, it is probably fine.

Please update pkarr to 2.0.2.

rklaehn (approved):

rklaehn:
Nevermind. We have a dependency on ureq, but we are not actually using it. So other than the unused dependency it's all good. You can remove the async and relay features from the iroh-net pkarr dependency...

iroh-net/src/discovery/pkarr.rs
```diff
@@ -91,11 +97,12 @@ impl PkarrPublisher {
     ///
     /// This is a nonblocking function, the actual update is performed in the background.
     pub fn update_addr_info(&self, info: &AddrInfo) {
-        let info = NodeInfo::new(
-            self.node_id,
-            info.relay_url.clone().map(Into::into),
-            Default::default(),
-        );
+        let (relay_url, direct_addresses) = if let Some(relay_url) = info.relay_url.as_ref() {
```
Frando, line 100: This behavior should be documented: it publishes direct addresses if no relay URL is set, and the relay URL otherwise, but never both. Thinking more - we might want to make it an option to the `PkarrPublisher` which info to publish, and fill in the default of `RelayUrlOrDirectAdresses` or so in the iroh node builder. The latter could be a followup PR also I guess.

dvc94ch: can you explain the use case?

dignifiedquire:
Needs rebase

iroh-net/src/discovery/pkarr.rs
```diff
@@ -174,6 +183,54 @@ impl PublisherService {
     }
 }
 
+/// Resolve node info using a pkarr relay.
+///
+/// Pkarr stores signed DNS records in the mainline dht. These can be queried directly
+/// via the pkarr relay HTTP api or alternatively via a dns server that provides the
+/// pkarr records using `DnsDiscovery`. The main difference is that `DnsDiscovery` makes
+/// use of the system dns resolver and caching which can return stale records, while the
+/// `PkarrResolver` always gets recent data.
```
flub, line 192: so a relay does a lookup to the DHT on every request?

Nuhvi: + LRU caching (available for all consumers of that relay and even PkarrClient that uses that domain as [resolver](pkarr.org/resolvers), and it also returns `cache-control` header, in case your HTTP client can use that.

Nuhvi: Maybe the documentation should say instead that the difference is that the relay returns a `SignedPacket` so it is trustless, unlike DNS.

dvc94ch: Good points. I don't think the ttl is used by iroh-net, seems like it should then. But ttl is a bit weird, if you have multiple DNS caches then the actual ttl is number of caches * ttl no?
The main reason for using this was what turned out to be a DNS misconfiguration (ttl for cache misses was set to 24h)

Nuhvi: PkarrClient uses the _smallest_ ttl in the packet, but it allows you to "clamp" that to a minimum or a maximum. So if your application prefers Zero TTL, it can do that, if it prefers at least 24h TTL it will cache packets for that long too.

But you can't control what clamp values the resolver/relay you are querying is using, only your client.

https://github.com/Nuhvi/pkarr/blob/main/pkarr/src/relay_client.rs#L29

PR #2419, opened 2024-06-26
refactor(iroh-relay)!: Remove `relay_endpoint` config option & rename `/derp` route to `/relay`

### LINKED ISSUE #2378: Move relay HTTP handler to be served on /relay

matheus23:
## Description

- Removes `relay_endpoint` config option from `ServerBuilder`
- Makes `iroh-relay` accept both `/derp` and `/relay` as paths for relaying
- Makes `/relay` the default path used in the iroh-net client

TODO:
- [x] Test both URLs

## Breaking Changes

- `iroh_net::relay::http::ServerBuilder`: Removed `relay_endpoint` function

## Notes & open questions

Closes #2378 

:white_check_mark: Merge this after #2387 

## Change checklist

- [X] Self-review.
- [X] Tests if relevant.

### DIFF
--- a/iroh-net/src/relay/http.rs
+++ b/iroh-net/src/relay/http.rs
@@ -12,6 +12,10 @@ pub(crate) const HTTP_UPGRADE_PROTOCOL: &str = "iroh derp http";
 pub(crate) const WEBSOCKET_UPGRADE_PROTOCOL: &str = "websocket";
 pub(crate) const SUPPORTED_WEBSOCKET_VERSION: &str = "13";
 
+/// The path under which the relay accepts relaying connections
+/// (over websockets and a custom upgrade protocol).
+pub const RELAY_HTTP_PATH: &str = "/relay";
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/iroh-net/src/relay/http.rs
+++ b/iroh-net/src/relay/http.rs
@@ -15,6 +15,8 @@ pub(crate) const SUPPORTED_WEBSOCKET_VERSION: &str = "13";
 /// The path under which the relay accepts relaying connections
 /// (over websockets and a custom upgrade protocol).
 pub const RELAY_HTTP_PATH: &str = "/relay";
+/// The path under which the relay allows doing latency queries for testing.
+pub const RELAY_PROBE_HTTP_PATH: &str = "/relay/probe";
 
 #[cfg(test)]
 mod tests {
--- a/iroh-net/src/relay/http.rs
+++ b/iroh-net/src/relay/http.rs
@@ -12,11 +12,17 @@ pub(crate) const HTTP_UPGRADE_PROTOCOL: &str = "iroh derp http";
 pub(crate) const WEBSOCKET_UPGRADE_PROTOCOL: &str = "websocket";
 pub(crate) const SUPPORTED_WEBSOCKET_VERSION: &str = "13";
 
-/// The path under which the relay accepts relaying connections
+/// The HTTP path under which the relay accepts relaying connections
 /// (over websockets and a custom upgrade protocol).
-pub const RELAY_HTTP_PATH: &str = "/relay";
-/// The path under which the relay allows doing latency queries for testing.
-pub const RELAY_PROBE_HTTP_PATH: &str = "/relay/probe";
+pub const RELAY_PATH: &str = "/relay";
+/// The HTTP path under which the relay allows doing latency queries for testing.
+pub const RELAY_PROBE_PATH: &str = "/relay/probe";
+/// The legacy HTTP path under which the relay used to accept relaying connections.
+/// We keep this for backwards compatibility.
+pub(crate) const LEGACY_RELAY_PATH: &str = "/derp";
+/// The legacy HTTP path under which the relay used to allow latency queries.
+/// We keep this for backwards compatibility.
+pub(crate) const LEGACY_RELAY_PROBE_PATH: &str = "/derp/probe";
 
 #[cfg(test)]
 mod tests {
--- a/iroh-net/src/relay/http/client.rs
+++ b/iroh-net/src/relay/http/client.rs
@@ -178,6 +178,7 @@ struct Actor {
     ping_tasks: JoinSet<()>,
     dns_resolver: DnsResolver,
     proxy_url: Option<Url>,
+    relay_path: &'static str,
 }
 
 #[derive(Default, Debug)]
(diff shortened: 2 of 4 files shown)


dignifiedquire:
any chance we can have a small test, to make sure both urls work?

matheus23:
Yes! Had that planned but forgot about it?

I remember looking into it, probably means I'll need to make the URL it dials configurable in the code - possibly passing it in at the `ClientBuilder` level or so.

Which also reminds me: Would it make sense to include the `/relay` bit in the `RelayUrl`? I.e. a relay URL would be e.g. something like `ws://relay.iroh.net/relay` instead of `ws://relay.iroh.net`.

dignifiedquire:
> Which also reminds me: Would it make sense to include the `/relay` bit in the `RelayUrl`? I.e. a relay URL would be e.g. something like `ws://relay.iroh.net/relay` instead of `ws://relay.iroh.net`.

I think this would actually be really good, as it would allow more interesting deployments setups

dignifiedquire:
actually though, I am not sure, as we hit other endpoints on a relay server, so that URL would be somewhat confusing

cc @ramfox @flub for some thoughts

matheus23:
Let's continue the discussion here.
However, if it turns out we'll want to do this - I'll create an issue :v:

iroh-net/src/relay/iroh_relay.rs
```diff
@@ -800,21 +793,27 @@ mod tests {
...
+
+        // set up client a
+        let secret_key = SecretKey::generate();
+        let resolver = crate::dns::default_resolver().clone();
+        let (client, _) = ClientBuilder::new(relay_url)
+            .relay_path("/derp") // Try the legacy relay path for backwards compatibility
+            .build(secret_key, resolver);
+
+        client.ping().await.unwrap();
+    }
```
matheus23, line 811: Setup this fairly simple smoke test. Does that cover what you were thinking? @dignifiedquire 
It's not doing the whole "two clients connected" thing, but it does exercise the relay to make sure it actually expects to receive `Frame`s and responds.

flub:
> actually though, I am not sure, as we hit other endpoints on a relay server, so that URL would be somewhat confusing
> 
> cc @ramfox @flub for some thoughts

Yeah, I think `RelayUrl` should be a "root" under which we expect to find several paths.  E.g. the client code does `relay_url.join("derp")`, `relay_url.join("generate_204")` etc.  So if someone is using `https://example.com/some/path/` I expect them to have done some reverse-proxy stuff to host it all under a path prefix.  Does that make sense?

So I would not put the `/relay` or `/derp` into the `RelayUrl`.

iroh-net/src/relay/iroh_relay.rs
flub: You could argue that you only care that the `/derp` endpoint responds with the right HTTP response for the test.  That would allow you to write a test with a few request calls instead of checking the entire protocol works on `/derp`.  The reasoning is that we already test the implementation on the `/relay` endpoint and the handler is the same anyway.  It would mean you don't need the `.relay_path` on the builder which is a bit unfortunate I think, as it would encourage people to change in on their setups.

flub: Because once you start providing that option, you open the door to requests for the other paths like `generate_204` etc.  Which just seems messy for very little value.  (maybe someone comes up with a really good usecase and I'll change my mind.  but let's wait for that)

matheus23: Agreed. I want to avoid the `relay_path` as much as possible. I was worried I'm not testing the `/derp` endpoint enough.

I'll change the PR to remove the `relay_path` change & only test the HTTP upgrade on `/derp`.

flub (commented):
LGTM but I wonder if this should be split up in two PRs: the server-side and switching the client.  We need to be pretty sure that we have deployed the new relay server before we release with the client switched.  I'd even do them in two subsequent releases.  /cc @Arqu

matheus23:
> I wonder if this should be split up in two PRs: the server-side and switching the client

Good idea. Probably not hard to do.
That'd also test that the "new" relay works with the old code (at least it'd test that for a week).

matheus23:
I split out #2441

iroh-net/src/relay/http/client.rs
```diff
@@ -620,7 +613,7 @@ impl Actor {
 
     async fn connect_ws(&self) -> Result<(ConnReader, ConnWriter), ClientError> {
         let mut dial_url = (*self.url).clone();
-        dial_url.set_path("/derp");
+        dial_url.set_path(RELAY_HTTP_PATH);
```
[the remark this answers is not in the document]
matheus23: I'm... not sure how the link you've posted relates to your comment. Can you say a little more? :sweat_smile: 

Noting for posterity: The `.join` calculation would also need to happen in `start_upgrade`, should we want to support deploying relays at other path prefixes.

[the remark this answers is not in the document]
flub: lol, that's the wrong link. sorry.
https://github.com/n0-computer/iroh/issues/2440

iroh-net/src/relay/http.rs
```diff
@@ -12,6 +12,18 @@ pub(crate) const HTTP_UPGRADE_PROTOCOL: &str = "iroh derp http";
 pub(crate) const WEBSOCKET_UPGRADE_PROTOCOL: &str = "websocket";
 pub(crate) const SUPPORTED_WEBSOCKET_VERSION: &str = "13";
 
+/// The HTTP path under which the relay accepts relaying connections
+/// (over websockets and a custom upgrade protocol).
+pub const RELAY_PATH: &str = "/relay";
+/// The HTTP path under which the relay allows doing latency queries for testing.
+pub const RELAY_PROBE_PATH: &str = "/relay/probe";
```
flub, line 19: So this is just bikesheeding, but nesting `/relay/probe` is a bit odd, given the other handlers are all on the root (I appreciate this was copied from tailscale).  I'd move this to `/probe` so it is on the same level as `/generate_204` etc.  We can definitely do this for the new route.  We can even remove `/derp/probe` if we want as it's not yet used (hi https-probe that still lives in a half-dead branch :sweat: ).

flub (approved):

iroh-net/src/relay/http.rs
matheus23: I'll keep it in this PR as a refactor instead of a change - let's talk about moving it when we're making use of it :)

PR #2422, opened 2024-06-27
feat(iroh-willow): Private Area Intersection

Frando:
## Description

Implementation of [Private Area Intersection](https://willowprotocol.org/specs/pai/index.html#private_area_intersection) as defined in the [WGPS spec](https://willowprotocol.org/specs/sync/index.html#sync_pai_messages).

The implementation of the `PaiFinder` is ported from the TypeScript implementation in `willow-js`.

Cryptographic scheme is the same as in `earthstar` (I think):

* The `PsiGroup` uses `RistrettoPoint` from `curve25519_dalek`, and `hash_to_group` uses `sha512` for hashing to a curve point. Went with `sha512` because `RistrettoPoint::hash_from_bytes` needs a digest with 64 bytes length.
*  The `PsiScalar` uses the `Scalar` type from `curve25519_dalek`

The smoke test passes :fire:

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


matheus23:
> The `PsiGroup` uses `RistrettoPoint` from `curve25519_dalek`, and `hash_to_group` uses `sha512` for hashing to a curve point. Went with `sha512` because `RistrettoPoint::hash_from_bytes` needs a digest with 64 bytes length.

SHA512 is not a bad choice, but FWIW, `blake3` supports extended output functions - so you can make blake3 give you 64 byte hashes.

That said, the most important factor in choosing this stuff (in my opinion) would be checking with what other people are planning on using. Are the earthstar folks planning to go with SHA-512, too?

Frando:
Earthstar uses SHA512 too, I think:

https://willowprotocol.org/earthstar/spec/index.html#es6_wgps_pai

> The hash_into_group function encodes a fragment using the encode_fragment function that we define below, then uses the encoding as input to edwards25519_XMD:SHA-512_ELL2_RO_ with the ascii encoding of the string earthstar6i as the domain separation tag.

AljoschaMeyer:
Jup, we went with the recommendation in RFC9830. Note that Earthstar uses Edwards25519 as its curve though, *not* ristretto255.

Frando:
> Jup, we went with the recommendation in RFC9830. Note that Earthstar uses Edwards25519 as its curve though, _not_ ristretto255.

Aha, interesting. Will read more, I don't know enough about the differences between these primitives.

I used ristretto  because `curve25519_dalek` has [`RistrettoPoint::from_hash`](https://docs.rs/curve25519-dalek/latest/curve25519_dalek/ristretto/struct.RistrettoPoint.html#method.from_hash) but there's no such method on [`EdwardsPoint`](https://docs.rs/curve25519-dalek/latest/curve25519_dalek/edwards/struct.EdwardsPoint.html).

matheus23:
This is the RFC @AljoschaMeyer is referring to: https://www.rfc-editor.org/rfc/rfc9380.html

matheus23:
> > FWIW, blake3 supports extended output functions - so you can make blake3 give you 64 byte hashes.
> 
> Cool, I didn't know. However the rustcrypto `Digest` trait is impl'd in blake3 with size 32, so we'd have to write some wrapper type that impls `Digest` with size 64 ourselves (if we wanted to use blake3 over sha512 here).

Not that we should do it, this is how you'd do it: use the `blake3` crate (ignore the rustcrypto `Digest` trait):
```rust
let mut hasher = blake3::Hasher::new();
hasher.update(b"Hello, world!");
let mut xof = hasher.finalize_xof();
let mut output = [0u8; 64];
xof.fill(&mut output);
```

iroh-willow/src/session/pai.rs
```diff
@@ -150,7 +156,7 @@ impl PaiFinder {
         Ok(())
     }
 
-    async fn submit_autorisation(&mut self, authorisation: ReadAuthorisation) -> Result<(), Error> {
+    async fn submit_autorisation(&mut self, authorisation: ReadAuthorisation) {
```
matheus23, line 159:
```suggestion
    async fn submit_authorisation(&mut self, authorisation: ReadAuthorisation) {
```

Frando:
I am going to merge this.
Left TODOs in the code for switching to ed25519 with RFC9380 once the crypto is available (or we wrote it).

PR #2423, opened 2024-06-28
refactor(iroh): log inner errors

rklaehn:
## Description

we currently abort the loop on outer errors (errors from spawn). But we do not look at inner errors (task returns properly but returns an error) at all.

This logs the inner errors as debug! and coninues.

Also checks if the outer error is actually a panic, and continues the loop otherwise.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/node.rs
+++ b/iroh/src/node.rs
@@ -308,9 +308,15 @@ impl<D: iroh_blobs::store::Store> NodeInner<D> {
                 },
                 // handle task terminations and quit on panics.
                 res = join_set.join_next(), if !join_set.is_empty() => {
-                    if let Some(Err(err)) = res {
-                        error!("Task failed: {err:?}");
-                        break;
+                    match res {
+                        Some(Err(outer)) => {
+                            error!("Task failed: {outer:?}");
+                            break;
+                        }
+                        Some(Ok(Err(inner))) => {
+                            debug!("Task errored: {inner:?}");
+                        }
+                        _ => {}
                     }
                 },
                 else => break,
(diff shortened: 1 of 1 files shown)


iroh/src/node.rs
```diff
@@ -308,9 +308,22 @@ impl<D: iroh_blobs::store::Store> NodeInner<D> {
...
+                            if outer.is_panic() {
+                                error!("Task panicked: {outer:?}");
+                                break;
+                            } else if outer.is_cancelled() {
+                                debug!("Task cancelled: {outer:?}");
+                            } else {
+                                error!("Task failed: {outer:?}");
+                                break;
+                            }
+                        }
```
matheus23, line 322: Do we need to print cancellations?
And IIUC, the third case is impossible. It's either a cancel or a panic: https://docs.rs/tokio/latest/src/tokio/runtime/task/error.rs.html#15-18

```suggestion
                        Some(Err(outer)) if outer.is_panic() => {
                            error!("Task panicked: {outer:?}");
                        }
```

matheus23 (approved):
I have a minor nit, but I'd say this fixes something important and would also be fine as-is :+1:

iroh/src/node.rs
rklaehn: > Do we need to print cancellations? And IIUC, the third case is impossible. It's either a cancel or a panic: https://docs.rs/tokio/latest/src/tokio/runtime/task/error.rs.html#15-18

Yeah, I thought so as well, but then why don't they expose an enum? The way the JoinError API is written it seems that they want to keep the possibility open to have a third case.

rklaehn: > Do we need to print cancellations?

Not sure if they would ever happen at all. But if it happens, I would like to know.

rklaehn:
> What do you think of renaming this PR `fix(iroh)`? IMO it _is_ a fix.

Not sure. It does not change the behaviour of the program except for logging, so calling it a fix seems wrong.

PR #2435, opened 2024-07-01
chore(ci): separate out android builds & disable netbsd

Arqu:
## Description

Disables the netbsd build due to sudden issues with cross and linking. 
Android separated out for the same reason but now builds directly without cross.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,16 +82,12 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v2
 
-    - name: Install rust stable
-      uses: actions-rs/toolchain@v1
+    - name: Set up Rust
+      uses: dtolnay/rust-toolchain@stable
       with:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -68,6 +69,51 @@ jobs:
...
+      with:
+        ndk-version: r23
+        add-to-path: true
+
+    - name: Build
+      env:
+        ANDROID_NDK_HOME: ${{ steps.setup-ndk.outputs.ndk-path }}
+      run: | 
+        cargo install cargo-ndk
+        cargo ndk --target aarch64-linux-android build
```
dignifiedquire, line 115: I think we can do the other android targets in this setup as well, would be more consistent

Arqu: lemme try

PR #2441, opened 2024-07-02
refactor(iroh-net): Switch to new iroh-relay route `/relay` instead of `/derp`

matheus23:
Split out from #2419 

Splitting this out means we can deploy the relay changes one release in advance to the changes in the client.

Otherwise, we'd need to update the relays ASAP when we release 0.20, as clients will expect to be able to connect to the `/relay` route (which may not be deployed immediately after the release).

## Description

- Switches the client to connect to `/relay` instead of `/derp`.

## Breaking Changes

None

## Notes & open questions

Also a PR related to #2378 

I allowed myself to add in another `derive_more::Debug` instance instead of a custom, outdated `impl Debug`, if that's okay with you :grin: 

## Change checklist

- [X] Self-review.
- [X] Tests if relevant.

### DIFF
--- a/iroh-net/src/relay/http/client.rs
+++ b/iroh-net/src/relay/http/client.rs
@@ -31,6 +31,7 @@ use crate::key::{PublicKey, SecretKey};
 use crate::relay::client::{ConnReader, ConnWriter};
 use crate::relay::codec::DerpCodec;
 use crate::relay::http::streams::{downcast_upgrade, MaybeTlsStream};
+use crate::relay::http::RELAY_PATH;
 use crate::relay::RelayUrl;
 use crate::relay::{
     client::Client as RelayClient, client::ClientBuilder as RelayClientBuilder,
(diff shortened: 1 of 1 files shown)


flub (approved):

matheus23:
**DO NOT MERGE YET**

Or rather - merge after the 0.20 release was published.

Read [this comment](https://github.com/n0-computer/iroh/pull/2419#pullrequestreview-2151330877) for the reasoning.
TLDR: We first publish changes to the relay to support the new `/relay` route with backwards compatibility in the 0.20 release, *then* we publish the changes to the client to make use of this new endpoint with the 0.21 release.

matheus23:
According to @Arqu the relays are now updated to 0.20.
We can merge this, unless we want to wait a bit with breaking changes due to the new release.

If nothing else comes up, I'll merge tomorrow (if I end up with enough time at a computer in Brussels).

flub:
Sounds good, let's merge!

matheus23:
I don't really understand why the base branch didn't switch automatically here :/

Issue #2221, opened 2024-04-22
flaky test: iroh-net discovery::test_dns_pkarr::pkarr_publish_dns_discover 


dignifiedquire:
https://github.com/n0-computer/iroh/actions/runs/8783456856/job/24099637605

```
failures:
    discovery::test_dns_pkarr::pkarr_publish_dns_discover

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 88 filtered out; finished in 2.59s


--- STDERR:              iroh-net discovery::test_dns_pkarr::pkarr_publish_dns_discover ---
Error: timeout

Stack backtrace:
   0: std::backtrace_rs::backtrace::dbghelp::trace
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library\std\src\..\..\backtrace\src\backtrace\dbghelp.rs:131
   1: std::backtrace_rs::backtrace::trace_unsynchronized
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library\std\src\..\..\backtrace\src\backtrace\mod.rs:66
   2: std::backtrace::Backtrace::create
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library\std\src\backtrace.rs:[331](https://github.com/n0-computer/iroh/actions/runs/8783456856/job/24099637605#step:10:332)
   3: std::backtrace::Backtrace::capture
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library\std\src\backtrace.rs:296
   4: anyhow::Error::msg<ref$<str$> >
             at C:\Users\Administrator\.cargo\registry\src\index.crates.io-6f17d22bba15001f\anyhow-1.0.82\src\error.rs:83
   5: anyhow::__private::format_err
             at C:\Users\Administrator\.cargo\registry\src\index.crates.io-6f17d22bba15001f\anyhow-1.0.82\src\lib.rs:688
   6: iroh_net::discovery::test_dns_pkarr::state::impl$0::on_node::async_fn$0
             at .\src\discovery.rs:745
   7: iroh_net::discovery::test_dns_pkarr::pkarr_publish_dns_discover::async_block$0
             at .\src\discovery.rs:655
[...]

PR #2445, opened 2024-07-02
fix(iroh-net): delays of non-stun probes for subsequent relays

### LINKED ISSUE #2444: Netcheck https & icmp probes are scheduled too late for subsequent relay servers
When you have a `RelayMap` with multiple relay servers the ICMP and HTTPS probes for the 2nd or higher servers are scheduled too late:

https://github.com/n0-computer/iroh/blob/main/iroh-net/src/netcheck/reportgen/probes.rs#L237

For the 2nd, 3rd etc relay server this line results in probes being further out in the future than they need to be.

flub:
## Description

The idea of the probe plan is that first the STUN probes happen, if
those don't work we add the other probes.  However when there are
multiple relay servers we accidentally started the subsequent probes
for all but the first relay server too late.

This makes sure to globally record when the last STUN probe was sent
and re-uses this value for all the relay servers which are probed.

## Notes & open questions

Fixes #2444

See #2221 for flaky tests.

## Change checklist

- [x] Self-review.
- ~~[ ] Documentation updates if relevant.~~
- [x] Tests if relevant.


ramfox (approved):
sweet!

Issue #2442, opened 2024-07-02
Make websocket/not websocket choice client-based instead of advertising `wss://` `RelayUrl`s


matheus23:
@flub correctly pointed out [on discord](https://discord.com/channels/949724860232392765/960896793195974708/1257613031609729024):
- Websocket transport is selected based on `wss://` vs. `https://` scheme in the `RelayUrl`
- This means clients would advertise whether they want to use websockets or not in their tickets, even though clients should decide that for themselves only!
- We achieve better compatibility if we announce a single URL
- It's weird to have the `wss://` prefix, given there's also other endpoints like `https://relay.example.org/generate_204`

I think the fix is:
- Make wss or https selection based on a client-side flag (probably in the `ClientBuilder`)
- Always advertise http(s) relay urls.

dignifiedquire:
I think the proposed solution is the right one

matheus23:
I'll put up a fix today, so it'll hopefully make it in before the release on Monday.

dignifiedquire:
Fixed in #2446

PR #2446, opened 2024-07-02
feat(iroh-net)!: Make relay protocol configurable on `ClientBuilder` instead of defined by the relay url scheme

matheus23:
## Description

Details in #2442 

## Breaking Changes

- `iroh_net::relay::http::ClientBuilder`: Added `.protocol()` function to the builder for choosing whether to connect to the relay via websockets or not

## Change checklist

- [X] Self-review.
- [X] Tests if relevant.

### DIFF
--- a/iroh-net/src/relay/http/client.rs
+++ b/iroh-net/src/relay/http/client.rs
@@ -171,6 +171,7 @@ struct Actor {
     address_family_selector: Option<Box<dyn Fn() -> BoxFuture<bool> + Send + Sync + 'static>>,
     conn_gen: usize,
     url: RelayUrl,
+    protocol: Protocol,
     #[debug("TlsConnector")]
     tls_connector: tokio_rustls::TlsConnector,
     pings: PingTracker,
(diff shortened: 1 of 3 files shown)


dignifiedquire (approved):

iroh-net/src/relay/http/client.rs
```diff
@@ -254,6 +258,13 @@ impl ClientBuilder {
         self
     }
 
+    /// Sets whether to connect to the relay via websockets or not.
+    /// Set to use non-websocket, normal relaying by default.
```
flub, line 262: I think it would be great if we could give some guidance here to the user what this means.  Why would a user choose websockets?  What happens when the server doesn't support it?

As written this is tells you something if you know this feature, but you need to explain the feature to the user because this is the first time they're encountering this and they won't have a clue what to do with this.  It's perfectly fine if that means repeating lots of stuff available elsewhere.

matheus23: You're right. At the moment I'd say this API is mostly for testing, or *maybe* if you have an http proxy that blocks http upgrades except for websockets. (Is something like that a thing?)

I'll update the docs :v:

PR #2454, opened 2024-07-03
refactor(iroh): Modularize protocol

rklaehn:
## Description

Use a newly created macro crate https://github.com/n0-computer/nested-enum-utils to allow for from/to conversions for deeply nested enums.

This allows us to split the protocol enum into multiple parts without changing anything about the basic concept of how things are handled.

## Breaking Changes

Weirdly, I think there should be none since the protocol is private.

## Notes & open questions

Obvious next steps would be
- [x] define the subsystem specific messages in submodules
- [x] publish the macro crate
- [x] (maybe) split up the handling of the different subsystems
- [x] (maybe) remove the subsystem prefixes from the messages

Not so obvious next steps (maybe next PR)
- Allow moving the subsystem protocols to different crates. This is difficult because you can not set the target when generating the conversions.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh/src/rpc_protocol/authors.rs
```diff
@@ -0,0 +1,146 @@
...
+use quic_rpc::message::{Msg, RpcMsg, ServerStreaming, ServerStreamingMsg};
+use serde::{Deserialize, Serialize};
+
+use super::RpcService;
+
+#[allow(missing_docs)]
+#[derive(strum::Display, Debug, Serialize, Deserialize)]
+#[nested_enum_utils::enum_conversions(super::Request)]
+pub enum Request {
+    List(AuthorListRequest),
```
dignifiedquire, line 12: given this is scoped now, I would remove the `Author` prefix

rklaehn: Did it for all messages. There are occasionally some ambiguities now, requiring import as..., but they will go away once the handlers are also separated...

rklaehn:
There is a second thing we may or may not want to do: we might want to make the individual subsystem protocols self-contained so that they reexport all the things they need. But I would prefer to do this in a subsequent PR since this one is big enough as it is.

iroh/src/rpc_protocol/authors.rs
```diff
@@ -0,0 +1,146 @@
...
+    GetDefault(RpcResult<GetDefaultResponse>),
+    SetDefault(RpcResult<SetDefaultResponse>),
+    Import(RpcResult<ImportResponse>),
+    Export(RpcResult<ExportResponse>),
+    Delete(RpcResult<DeleteResponse>),
+}
+
+/// List document authors for which we have a secret key.
+#[derive(Serialize, Deserialize, Debug)]
+pub struct ListRequest {}
```
dignifiedquire, line 36: I wonder if it would read nicer/be a better way to organize these like this:

```rust
pub mod request {
  pub struct List {}
  pub struct Create {}
}

pub mod response {
  pub struct List {}
  pub struct Create {}
}
```

or alternatively 

```rust
pub mod list {
  pub struct Request {}
  pub struct Response {}
}

pub mod create {
  pub struct Request {}
  pub struct Response {}
}
```

rklaehn: Nah, I have briefly considered it, but I think you would very frequently have collisions. At the extreme of this approach you don't have any meaning attached to type names at all, and can only glance the meaning by having the complete or at least partial path...

Even removing the subsystem prefixes (NodeFoo to Foo) caused some unfortunate collisions, but my hope is that those will go away once we properly separate the rpc handlers. But for within one subsystem you really want the types to be meaningful by just looking at the local name...

iroh/src/node/rpc.rs
```diff
@@ -121,213 +133,254 @@ impl<D: BaoStore> Handler<D> {
...
 
-            AuthorList(msg) => {
-                chan.server_streaming(msg, self, |handler, req| {
+    async fn handle_tags_request<E: ServiceEndpoint<RpcService>>(
+        self,
+        msg: tags::Request,
+        chan: RpcChannel<RpcService, E>,
+    ) -> Result<(), RpcServerError<E>> {
+        use tags::Request::*;
+        debug!("handling tags request: {msg}");
```
dignifiedquire, line 198: remove?

dignifiedquire (approved):

iroh/src/node/rpc.rs
rklaehn: lol yes I guess so...

PR #2461, opened 2024-07-05
refactor(iroh)!: Remove server channel type parameter

rklaehn:
## Description

This is a followup to the modularize services PR. I did not want to do it in that one since it was big enough already.

## Breaking Changes

- A `#[must_use]` annotation was added to AbortingJoinHandle

```
     Checked [   0.095s] 75 checks; 74 passed, 1 failed, 0 unnecessary

--- failure struct_must_use_added: struct #[must_use] added ---

Description:
A struct is now #[must_use]. Downstream crates that did not use its value will get a compiler lint.
        ref: https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.32.0/src/lints/struct_must_use_added.ron

Failed in:
  struct AbortingJoinHandle in /home/runner/work/iroh/iroh/iroh-net/src/util.rs:18
     Summary semver requires new minor version: 0 major and 1 minor checks failed
```

## Change checklist

- [x] Self-review.
- [x] ~~Tests if relevant.~~

### DIFF
--- a/iroh/src/node.rs
+++ b/iroh/src/node.rs
@@ -18,16 +18,14 @@ use iroh_gossip::net::Gossip;
 use iroh_net::key::SecretKey;
 use iroh_net::Endpoint;
 use iroh_net::{endpoint::DirectAddrsStream, util::SharedAbortingJoinHandle};
-use quic_rpc::{RpcServer, ServiceEndpoint};
+use quic_rpc::transport::ServerEndpoint as _;
+use quic_rpc::RpcServer;
 use tokio::task::JoinSet;
 use tokio_util::sync::CancellationToken;
 use tokio_util::task::LocalPoolHandle;
 use tracing::{debug, error, info, warn};
 
-use crate::{
-    client::RpcService,
-    node::{docs::DocsEngine, protocol::ProtocolMap},
-};
+use crate::node::{docs::DocsEngine, protocol::ProtocolMap};
 
 mod builder;
 mod docs;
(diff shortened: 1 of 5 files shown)


iroh/src/node.rs
```diff
@@ -39,6 +37,11 @@ pub use self::builder::{Builder, DiscoveryConfig, DocsStorage, GcPolicy, Storage
 pub use self::rpc_status::RpcStatus;
 pub use protocol::ProtocolHandler;
 
+type IrohServerEndpoint = quic_rpc::transport::boxed::ServerEndpoint<
```
dignifiedquire, line 40: this appears in a public api I believe, so it should be public

iroh/src/node/builder.rs
```diff
@@ -309,7 +302,7 @@ where
     }
 
     /// Configure rpc endpoint, changing the type of the builder to the new endpoint type.
```
dignifiedquire, line 304: This comment is not correct anymore, as the endpoint type doesn't change. Also I think we don't need to reconstruct the builder anymore here, it can just mutate itself, as the type stays

rklaehn: We are using self and not &mut self everywhere in this builder, why change it here?

dignifiedquire: no need to change the function signature: 

```rust
fn foo(mut self) -> Self {
  self.foo = true;
  self
}
```

rklaehn: Ah yes, I ended up using `Self { self.foo=bla, ..self }`, old FP person that I am.

dignifiedquire (approved):

PR #2463, opened 2024-07-05
docs(iroh): Expand module level documentation in iroh

rklaehn:
## Description

- Actually export the second stage buider.
- Expand the module level docs for iroh and the various client modules.

## Breaking Changes

None, I am going to do those in a separate PR

## Change checklist

- [x] Self-review.
- [x] ~~Tests if relevant.~~

### DIFF
--- a/iroh/src/client/blobs.rs
+++ b/iroh/src/client/blobs.rs
@@ -9,15 +9,15 @@
 //! There are several ways to import data into the local blob store:
 //!
 //! - [`add_bytes`](Client::add_bytes)
-//!   allows importing in memory data.
+//!   imports in memory data.
 //! - [`add_stream`](Client::add_stream)
-//!   allows importing data from a stream of bytes.
+//!   imports data from a stream of bytes.
 //! - [`add_reader`](Client::add_reader)
-//!   allows importing data from an async reader.
+//!   imports data from an [async reader](tokio::io::AsyncRead).
 //! - [`add_from_path`](Client::add_from_path)
-//!   allows importing data from a file.
+//!   imports data from a file.
 //!
-//! The last method allows importing data from a file on the local filesystem.
+//! The last method imports data from a file on the local filesystem.
 //! This is the most efficient way to import large amounts of data.
 //!
 //! ### Exporting data
@@ -25,15 +25,14 @@
 //! There are several ways to export data from the local blob store:
 //!
 //! - [`read_to_bytes`](Client::read_to_bytes) reads data into memory.
-//! - [`read`](Client::read) creates a reader to read data from.
-//! - [`export`](Client::export)
-//!   allows exporting data to a file on the local filesystem.
+//! - [`read`](Client::read) creates a [reader](Reader) to read data from.
+//! - [`export`](Client::export) eports data to a file on the local filesystem.
 //!
 //! ## Interacting with remote nodes
 //!
 //! - [`download`](Client::download) downloads data from a remote node.
-//! - [`share`](Client::share)
-//!   allows creating a ticket to share data with a remote node.
+//! - [`share`](Client::share) allows creating a ticket to share data with a
+//!   remote node.
 //!
 //! ## Interacting with the blob store itself
 //!
--- a/iroh/src/client/blobs.rs
+++ b/iroh/src/client/blobs.rs
@@ -2,12 +2,6 @@
 //!
 //! The main entry point is the [`Client`].
 //!
-//! It can be used to interact with the local blob store, e.g. to import
-//! and export data.
-//!
-//! It can also be used to interact with remote nodes, e.g. to download
-//! data from them.
-//!
 //! ## Interacting with the local blob store
 //!
 //! ### Importing data
--- a/iroh/src/client/blobs.rs
+++ b/iroh/src/client/blobs.rs
@@ -1,5 +1,56 @@
 //! API for blobs management.
-
+//!
+//! The main entry point is the [`Client`].
+//!
+//! It can be used to interact with the local blob store, e.g. to import
+//! and export data.
+//!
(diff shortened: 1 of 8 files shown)


iroh/src/client/blobs.rs
```diff
@@ -1,5 +1,50 @@
...
+//!
+//! The main entry point is the [`Client`].
+//!
+//! ## Interacting with the local blob store
+//!
+//! ### Importing data
+//!
+//! There are several ways to import data into the local blob store:
+//!
+//! - [`add_bytes`](Client::add_bytes)
```
Frando, line 11: Stylewise IMO "imports in-memory data" reads nicer than "allows importing" (this applies to a few more sentences as well)

iroh/src/lib.rs
```diff
@@ -1,5 +1,79 @@
...
+//! Node also implements [Deref](std::ops::Deref) to the client, so you can call
+//! client methods directly on the node.
+//!
+//! ## Subsystems
+//!
+//! The client provides access to various subsystems:
+//! - [node](crate::client::node):
+//!   information and control of the iroh node itself
+//! - [blobs](crate::client::blobs):
+//!   content-addressed blobs
```
Frando, line 49: "manage and share content-addressed blobs of data"

iroh/src/lib.rs
```diff
@@ -1,5 +1,79 @@
...
+//!
+//! ## Remote nodes
+//!
+//! To obtain a client to a remote node, you can use
+//! [connect](crate::client::Iroh::connect_path) to connect to a node running on
+//! the same machine, using the given data directory, or
+//! [connect_addr](crate::client::Iroh::connect_addr) to connect to a node at a
+//! known address.
+//!
+//! **Important**: the protocol to a remote node is not stable will frequently
```
Frando, line 74:
```suggestion
//! **Important**: the protocol to a remote node is not stable and will frequently
```

iroh/src/lib.rs
```diff
@@ -1,5 +1,79 @@
...
+//! change. So the client and server must be running the same version of iroh.
```
Frando, line 75:
```suggestion
//! change. So the client and server must be running the same major version of iroh.
```

iroh/src/node.rs
```diff
@@ -1,6 +1,34 @@
...
+//!
+//! To implement a custom protocol, implement the [`ProtocolHandler`] trait
+//! and use [`ProtocolBuilder::accept`] to add it to the node.
+//!
+//! # Using a node
+//!
+//! Once created, a node offers a small number of methods to interact with it,
+//! most notably the iroh-net [endpoint](Node::endpoint) it is bound to.
+//!
+//! But the main way to interact with a node is through the
```
Frando, line 30:
```suggestion
//! The main way to interact with a node is through the
```
and add "The `Node` implements `Deref<Client>`, which means that methods defined on `Client` can be called on `Node` as well, without going through `Node::client()`.

Frando (approved):
This is great, thank you @rklaehn!

iroh/src/lib.rs
```diff
@@ -1,5 +1,79 @@
...
+//!
+//! ## Example
+//!
+//! Create a new node and add some data to the blobs store. This data will be
+//! available over the network.
+//!
+//! ```rust
+//! # async fn run() -> anyhow::Result<()> {
+//! let node = iroh::node::Node::memory().spawn().await?;
+//! let client = node.client();
```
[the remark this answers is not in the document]
rklaehn: I want to keep it in to make it more clear that the client is a separate entity. Otherwise the explanation section below would read weirdly. Client? what client?

iroh/src/client/blobs.rs
```diff
@@ -1,5 +1,49 @@
...
+//!
+//! The last method imports data from a file on the local filesystem.
+//! This is the most efficient way to import large amounts of data.
+//!
+//! ### Exporting data
+//!
+//! There are several ways to export data from the local blob store:
+//!
+//! - [`read_to_bytes`](Client::read_to_bytes) reads data into memory.
+//! - [`read`](Client::read) creates a [reader](Reader) to read data from.
```
flub, line 28:
```suggestion
//! - [`read`](Client::read) creates a [`Reader`] to read data from.
```
This would make more clear that you're linking directly to a type which I usually prefer.  That's just my personal preference though, no written rule.

iroh/src/client/tags.rs
```diff
@@ -1,5 +1,14 @@
 //! API for tag management.
-
+//!
+//! The purpose of tags is to mark information as important. Currently this is
+//! used for blobs.
```
flub, line 4: Is it described anywhere *how* this is used for blobs?  Why not give a short description here as an example that gives a general idea of how tags are used?

iroh/src/lib.rs
```diff
@@ -1,5 +1,79 @@
...
 //!
+//! # Getting started
+//!
+//! ## Example
+//!
+//! Create a new node and add some data to the blobs store. This data will be
+//! available over the network.
+//!
+//! ```rust
+//! # async fn run() -> anyhow::Result<()> {
```
flub, line 11: wait, what?  We can just write `async fn` in doc tests and it'll work?  How did I now know that?

iroh/src/lib.rs
```diff
@@ -1,5 +1,79 @@
...
+//! println!("hash: {}", hash);
+//! # Ok(())
+//! # }
+//! ```
+//!
+//! ## Explanation
+//!
+//! ### Iroh node
+//!
+//! To create an iroh [Node](crate::node::Node), you use the
```
flub, line 24: style-guide prefers reference-style links rather than inline

PR #2464, opened 2024-07-05
refactor!: metrics

Arqu:
## Description

- Cleans up stale counters
- Introduces unique daily node counts on relays
- Introduces a metrics dumper so we can create dumps of metrics and also view them in doctor plot
- Introduces new trace level counters

## Breaking Changes

- Introduces `metrics_dump_path` on the top level CLI parameters, which if set will make sure metrics are collected at regular intervals and written to the provided path in CSV format.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-blobs/Cargo.toml
+++ b/iroh-blobs/Cargo.toml
@@ -29,7 +29,7 @@ hashlink = { version = "0.9.0", optional = true }
 hex = "0.4.3"
 iroh-base = { version = "0.20.0", features = ["redb"], path = "../iroh-base" }
 iroh-io = { version = "0.6.0", features = ["stats"] }
-iroh-metrics = { version = "0.20.0", path = "../iroh-metrics", optional = true }
+iroh-metrics = { version = "0.20.0", path = "../iroh-metrics", default-features = false }
 iroh-net = { version = "0.20.0", path = "../iroh-net" }
 num_cpus = "1.15.0"
 parking_lot = { version = "0.12.1", optional = true }
@@ -74,10 +74,10 @@ package = "iroh-quinn"
 version = "0.10"
 
 [features]
-default = ["fs-store", "metrics"]
+default = ["fs-store"]
 downloader = ["dep:parking_lot", "tokio-util/time", "dep:hashlink"]
 fs-store = ["dep:reflink-copy", "redb", "dep:redb_v1", "dep:tempfile"]
-metrics = ["dep:iroh-metrics"]
+metrics = ["iroh-metrics/metrics"]
 redb = ["dep:redb"]
 
 [package.metadata.docs.rs]
--- a/iroh-blobs/Cargo.toml
+++ b/iroh-blobs/Cargo.toml
@@ -74,7 +74,7 @@ package = "iroh-quinn"
 version = "0.10"
 
 [features]
-default = ["fs-store"]
+default = ["fs-store", "metrics"]
 downloader = ["dep:parking_lot", "tokio-util/time", "dep:hashlink"]
 fs-store = ["dep:reflink-copy", "redb", "dep:redb_v1", "dep:tempfile"]
 metrics = ["dep:iroh-metrics"]
--- a/iroh-dns-server/src/metrics.rs
+++ b/iroh-dns-server/src/metrics.rs
@@ -9,7 +9,7 @@ use struct_iterable::Iterable;
 pub struct Metrics {
     pub pkarr_publish_update: Counter,
     pub pkarr_publish_noop: Counter,
-    pub pkarr_publish_error: Counter,
+    // pub pkarr_publish_error: Counter,
     pub dns_requests: Counter,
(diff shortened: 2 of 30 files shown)


iroh-blobs/Cargo.toml
```diff
@@ -72,7 +72,7 @@ package = "iroh-quinn"
 version = "0.10"
 
 [features]
-default = ["fs-store"]
+default = ["fs-store", "metrics"]
```
b5, line 75: so this means metrics collection is on by default in blobs now, or just that the feature is available?

iroh-dns-server/src/metrics.rs
```diff
@@ -9,7 +9,7 @@ use struct_iterable::Iterable;
 pub struct Metrics {
     pub pkarr_publish_update: Counter,
     pub pkarr_publish_noop: Counter,
-    pub pkarr_publish_error: Counter,
+    // pub pkarr_publish_error: Counter,
```
b5, line 12: stray comment, plz remove

iroh-net/src/magicsock/metrics.rs
```diff
@@ -7,19 +7,19 @@ use iroh_metrics::{
...
+    // pub send_relay_queued: Counter,
+    // pub send_relay_error_chan: Counter,
+    // pub send_relay_error_closed: Counter,
+    // pub send_relay_error_queue: Counter,
     pub send_ipv4: Counter,
-    pub send_ipv4_error: Counter,
+    // pub send_ipv4_error: Counter,
     pub send_ipv6: Counter,
-    pub send_ipv6_error: Counter,
+    // pub send_ipv6_error: Counter,
```
b5, line 22: remove commented-out code

iroh-net/src/magicsock/metrics.rs
```diff
@@ -40,7 +40,7 @@ pub struct Metrics {
     pub sent_disco_ping: Counter,
     pub sent_disco_pong: Counter,
     pub sent_disco_call_me_maybe: Counter,
-    pub recv_disco_bad_peer: Counter,
+    // pub recv_disco_bad_peer: Counter,
```
b5, line 43: remove commented-out code

iroh-net/src/magicsock/metrics.rs
```diff
@@ -49,7 +49,7 @@ pub struct Metrics {
     pub recv_disco_ping: Counter,
     pub recv_disco_pong: Counter,
     pub recv_disco_call_me_maybe: Counter,
-    pub recv_disco_call_me_maybe_bad_node: Counter,
+    // pub recv_disco_call_me_maybe_bad_node: Counter,
```
b5, line 52: remove

iroh-dns-server/src/metrics.rs
flub, line 12: this is obviously guessing at usage.  but i'm surprised you want to remove this.  i'm used to seeing most counters in at least a pair of `_total` and `_error` so you can do error ratios.

iroh-net/src/magicsock.rs
```diff
@@ -1762,25 +1762,30 @@ impl Actor {
         };
 
         loop {
+            inc!(Metrics, actor_tick_main);
```
flub, line 1765: all these `_tick_` metrics are basically trace-level metrics.  is this really a good idea?  We have trace-level logging for this.  What do you expect these to be used for?

dignifiedquire: we used these quite a lot in the past to detect the health of tokio loops

iroh-blobs/Cargo.toml
Arqu: Reworked so that dep is always available but metrics collection is not on by default.

iroh-dns-server/src/metrics.rs
Arqu: The point of this PR was for the team to claim which ones they would like to keep as they are unused. I'm going to nuke all the ones people didn't vote to keep.

Arqu: I agree, however these are unused counters and I think those can be misleading as you expect to see something in errors yet they never increment even if there was an error.

iroh-net/src/magicsock.rs
Arqu: So 2 reasons for this:
- In general I'd like to move away from trace logging and shift towards trace metrics. Much easier to reason with it on a chart. Though logs do have their place where explicit order is really important.
- We actually used these in development MUCH more and we're more useful to us than "product level health" metrics which you look at when deployed.

I could follow up on this and maybe introduce a split for "trace_metrics" vs "metrics" if people find these too much.

Arqu: The general idea is to use metrics in your dev workflow too.

dignifiedquire (approved):

PR #2466, opened 2024-07-05
fix(iroh-net): use staging URL for pkarr publish in dev mode

Frando:
## Description

We switched DNS discovery to a staging server when run with `cfg(test)` but did NOT switch pkarr publish. This means node discovery is not working on main once `cfg(any(test, feature = "test_utils"))` triggers.

This fixes this. It also means that there is no test for this atm (we always run local? which would be good too).

## Breaking Changes

* Renamed `iroh_net::discovery::pkarr::N0_DNS_PKARR_RELAY` to `N0_DNS_PKARR_RELAY_PROD`

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-dns-server/examples/publish.rs
+++ b/iroh-dns-server/examples/publish.rs
@@ -4,8 +4,8 @@ use anyhow::{bail, Result};
 use clap::{Parser, ValueEnum};
 use iroh_net::{
     discovery::{
-        dns::N0_DNS_NODE_ORIGIN_PROD,
-        pkarr::{PkarrRelayClient, N0_DNS_PKARR_RELAY},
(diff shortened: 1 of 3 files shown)


Arqu (approved):
👍 nice catch

dignifiedquire (approved):
thanks, I missed that

Frando:
OK fixed the examples too.

By the way, I think they are quite useful (`--example publish` and `--example resolve`), I used them for example just now to debug the discovery. Should we add at least the resolve example to the doctor? Might be nice.

Anyway, merging this once its green.

dignifiedquire (approved):

PR #2468, opened 2024-07-05
example(iroh): Improve and document `custom-protocol` example

Frando:
## Description

* Changes the `custom-protocol` example to do something a bit more akin to something useful while still being concise
* Add a lot of comments everywhere to really explain what's going on

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh/examples/custom-protocol.rs
+++ b/iroh/examples/custom-protocol.rs
@@ -1,10 +1,49 @@
-use std::sync::Arc;
+//! Example for adding a custom protocol to a iroh node.
+//!
+//! We are building a very simple custom protocol here, and make our iroh nodes speak this protocol
+//! in addition to the built-in protocols (blobs, gossip, docs).
+//!
+//! Our custom protocol allows querying the blob store of other nodes for text matches. For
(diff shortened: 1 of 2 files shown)


iroh/examples/custom-protocol.rs
```diff
@@ -55,66 +128,157 @@ async fn main() -> Result<()> {
...
+            self.add_to_index(hash).await?;
+            out.push(hash);
+        }
+        Ok(out)
+    }
+
+    /// Query the local database.
+    ///
+    /// Returns the list of hashes of blobs which contain `query` literally.
+    pub async fn query_local(&self, query: &str) -> Result<Vec<Hash>> {
```
matheus23, line 236:
```suggestion
    pub fn query_local(&self, query: &str) -> Vec<Hash> {
```

Needed to do this to fix some type error.

matheus23 (approved):
I trust you to fix the type error.
Other than that I think this is great :)
:tada:

iroh/examples/custom-protocol.rs
Frando: uh yea thanks, this was a refactor leftover

PR #2469, opened 2024-07-05
feat: add Asia Pacific relay url to the default relay url list in production

ramfox:
## Description

Adds a new relay server to the default production list. `https://aps1-1.relay.iroh.network/` is the address for our newest server, located in Mumbai.

This does not affect any tests, as we do not have a corresponding Asia Pacific staging server, and so an additional server has not been added to our test-only list of default relay servers

## Change checklist

- [x] Self-review.
- [ ] ~Tests if relevant.~

### DIFF
--- a/iroh-cli/src/config.rs
+++ b/iroh-cli/src/config.rs
@@ -10,7 +10,7 @@ use std::{
 
 use anyhow::{anyhow, bail, Context, Result};
 use iroh::net::{
-    defaults,
+    defaults::{self},
     relay::{RelayMap, RelayNode},
 };
 use iroh::node::GcPolicy;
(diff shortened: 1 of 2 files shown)


dignifiedquire (approved):

iroh-cli/src/config.rs
```diff
@@ -10,7 +10,7 @@ use std::{
 
 use anyhow::{anyhow, bail, Context, Result};
 use iroh::net::{
-    defaults,
+    defaults::{self},
```
dignifiedquire, line 13: this looks, odd

ramfox: yea, weird. auto import or something. I'll fix it.

PR #2470, opened 2024-07-08
fix(iroh-docs): do not dial invalid peers

dignifiedquire:
## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


Frando (approved):

Frando:
The fix was incorrect, pushed two commits:

* we need to join gossip also if no peers are set. 
* we may not skip node addrs that have empty address info.

For the last one - We maybe should change iroh_net so that `Endpoint::add_node_addr` does not error if called with a `NodeAddr` with emtpy `AddrInfo` (only NodeId set) because it seems cumbersome to check at the user level if the addr contains addr info or not, usually you get these from a ticket or some protocol and just want to add them in case they contain useful info and don't care if not because node discovery is enabled. But that can be a followup.

PR #2471, opened 2024-07-08
fix(iroh-docs): Add `flush_store` and use it to make sure the default author is persisted

### LINKED ISSUE #2462: bug: "The default author is missing from the docs store, to recover delete the file [...]"
We get this bug in a bunch of places.
Here's a report from @azzamsa in discord:

> Hi, 👋🏻 
>
> 
> If I run this code over and over, I don't get any errors.
> 
> ```rust
>     let storage_path = Path::new(env!("HOME"))
>         .join(".local")
>         .join("share")
>         .join("todos-iroh");
>     let config_path = storage_path.join("todos.cfg");
> 
>     // Initialize node
>     let node = Node::persistent(&storage_path).await?.spawn().await?;
>     let author = node.authors().default().await?;
>     dbg!(&author);
> ```
> 
> but the time I add a logic to load previously created document, I always get `Error: The default author is missing from the docs store. To recover, delete the file /home/user/.local/share/todos-iroh/default-author. Then iroh will create a new default author.`
> 
> ```rust
> async fn init_repo() -> Result<(), crate::Error> {
>     let storage_path = Path::new(env!("HOME"))
>         .join(".local")
>         .join("share")
>         .join("todos-iroh");
>     let config_path = storage_path.join("todos.cfg");
> 
>     // Initialize node
>     let node = Node::persistent(&storage_path).await?.spawn().await?;
>     let author = node.authors().default().await?;
>     dbg!(&author);
> 
>     // Get previous document id
>     let document = match config_path.exists() {
>         true => {
>             println!("Using previous config.");
>             let document_id = fs::read_to_string(config_path)?;
>             let document = node
>                 .client()
>                 .docs()
>                 .open(NamespaceId::from_str(&document_id)?)
>                 .await?;
>             document.expect("can't find document")
>         }
>         false => {
>             println!("Creating new config.");
>             let document = node.client().docs().create().await?;
>             let document_id = document.id().to_string();
>             fs::write(config_path, document_id)?;
>             document
>         }
>     };
> 
>     Ok(())
> }
> ```

There are a bunch of failures in our [Flaky CI tests](https://github.com/n0-computer/iroh/actions/runs/9804155922/job/27071450440) that reproduce this:

> ```
> --- STDOUT:              iroh-cli::cli cli_rpc_lock_restart ---
> 
> running 1 test
> start
> |Iroh is running
> |Node ID: soubiinn7jdejlljcmled4i3s4n7znm4krpgoztjzo3hy4l7jb5q
> killing process
> restart
> |Error: The default author is missing from the docs store. To recover, delete the file `/root/actions-runner/_work/iroh/iroh/target/testdir-0/cli/cli_rpc_lock_restart/data-dir/default-author`. Then iroh will create a new default author.
> Expected 1 matches for regex: Iroh is running
> test cli_rpc_lock_restart ... FAILED
> 
> failures:
> 
> failures:
>     cli_rpc_lock_restart
> 
> test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 10 filtered out; finished in 4.67s
> ```

matheus23:
## Description

- Add a `flush_store` function to the docs actor/handle
- Use `docs_store.flush_store().await?` before persisting the default author ID file

There's two pieces of data related to the default author:
- The default author ID file
- The actual default author (the secret key) in redb

The code to store these two pieces of data looked like this:
```rust
docs_store.import_author(author).await?;
self.persist(author_id).await?;
```
So it would *look* like we make sure the author is stored before we store the file (which is basically just a reference to the author in the DB).

However, `import_author` is somewhat lazy, because it batches operations together in a bigger transaction.
The store would *actually* only get flushed in e.g. `shutdown`. So most of the time, the actual persisting of the author would happen *after* the call to `self.persist`.

If you wait for `shutdown` every time, it's all good, because that'll call `flush`. If not, then there will be some issues.

---

The fix in this PR is to introduce a `flush_store` function so we can use it after `import_author` to make sure it's actually flushed before we `self.persist(author_id)`.

## Breaking Changes

None. Only an addition: `iroh-docs::actor::SyncHandle::flush_store`.

## Notes & open questions

There might still be something wrong with `shutdown` not running correctly, but IMO this is an improvement regardless.

Closes #2462 

## Change checklist

- [X] Self-review.
- ~~[ ] Tests if relevant.~~

### DIFF
--- a/iroh-docs/src/actor.rs
+++ b/iroh-docs/src/actor.rs
@@ -71,6 +71,11 @@ enum Action {
         #[debug("reply")]
         reply: oneshot::Sender<Result<ContentHashesIterator>>,
     },
+    #[display("FlushStore")]
+    FlushStore {
+        #[debug("reply")]
+        reply: oneshot::Sender<Result<()>>,
+    },
     #[display("Replica({}, {})", _0.fmt_short(), _1)]
     Replica(NamespaceId, ReplicaAction),
     #[display("Shutdown")]
@@ -542,6 +547,12 @@ impl SyncHandle {
         rx.await?
     }
 
+    pub async fn flush_store(&self) -> Result<()> {
+        let (reply, rx) = oneshot::channel();
+        self.send(Action::FlushStore { reply }).await?;
+        rx.await?
+    }
+
     async fn send(&self, action: Action) -> Result<()> {
         self.tx
             .send_async(action)
@@ -675,6 +686,7 @@ impl Actor {
             Action::ContentHashes { reply } => {
                 send_reply_with(reply, self, |this| this.store.content_hashes())
             }
+            Action::FlushStore { reply } => send_reply(reply, self.store.flush()),
             Action::Replica(namespace, action) => self.on_replica_action(namespace, action),
         }
     }
--- a/iroh-docs/src/actor.rs
+++ b/iroh-docs/src/actor.rs
@@ -547,6 +547,15 @@ impl SyncHandle {
         rx.await?
     }
 
+    /// Makes sure that all pending database operations are persisted to disk.
+    ///
+    /// Otherwise, database operations are batched into bigger transactions for speed.
+    /// Use this if you need to make sure something is written to the database
+    /// before another operation, e.g. to make sure sudden process exits don't corrupt
+    /// your application state.
+    ///
+    /// It's not necessary to call this function before shutdown, as `shutdown` will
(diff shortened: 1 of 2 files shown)


matheus23:
Before this PR, the `cli_rpc_lock_restart` test would fail "reliably".
After this PR, it runs fine locally.
It may still be flaky. I don't know. But at least it's not "broken for sure".

PR #2472, opened 2024-07-08
docs(iroh-net): Update discovery and dialing docs, signatures

flub:
## Description

This updates the documentationfor the dialing and discovery modules.
It also fixes up some signatures to be more consistent with the rest
of iroh-net:

- NodeId is used to identify an iroh-net node that is communicated
  with.  PublicKey is only used in contexts where
  encryption/decryption happens

- NodeId is Copy and we pass it by value normally.  Not by reference.

Some more changes:

- DialFuture is entirely unused, probably left over from a refactor.
  Removed.

- Use pub(super) explicitly, helps clarity of intention as well as
  dead-code detection.

## Breaking Changes

The following APIs now take NodeId by value rather than by reference:
- iroh_blobs::downloader::Dialer::is_pending
- iroh_net::dialer::Dialer::abort_dial
- iroh_net::dialer::Dialer::is_pending
- iroh_net::Endpoint::connect_by_node_id

Other changes:
- iroh_blobs::downloader::ProviderMap::get_candidates yields NodeId rather than &NodeId

## Change checklist

- [x] Self-review.
- ~~[ ] Tests if relevant.~~

### DIFF
--- a/iroh-net/src/discovery.rs
+++ b/iroh-net/src/discovery.rs
@@ -4,9 +4,9 @@
 //! [`RelayUrl`] or one or more *direct addresses*.  However it is often more desirable to
 //! be able to connect with only the [`NodeId`], as [`Endpoint::connect_by_node_id`] does.
 //!
-//! For connecting by [`NodeId`] to work however the endpoint has to get the addressing
+//! For connecting by [`NodeId`] to work however, the endpoint has to get the addressing
 //! information by other means.  This can be done by manually calling
-//! [`Endpoint::add_node_addr`] but that still requires knowing the other addressing
+//! [`Endpoint::add_node_addr`], but that still requires knowing the other addressing
 //! information.
 //!
 //! Node discovery is an automated system for an [`Endpoint`] to retrieve this addressing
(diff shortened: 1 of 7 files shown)


iroh-net/src/discovery.rs
```diff
@@ -1,4 +1,43 @@
-//! Trait and utils for the node discovery mechanism.
+//! Node address discovery.
+//!
+//! To connect to an iroh-net node a [`NodeAddr`] is needed, which needs to contain either a
+//! [`RelayUrl`] or one or more *direct addresses*.  However it is often more desirable to
+//! be able to connect with only the [`NodeId`], as [`Endpoint::connect_by_node_id`] does.
+//!
+//! For connecting by [`NodeId`] to work however the endpoint has to get the addressing
```
matheus23, line 7:
```suggestion
//! For connecting by [`NodeId`] to work however, the endpoint has to get the addressing
```

iroh-net/src/discovery.rs
```diff
@@ -1,4 +1,43 @@
...
+//! information by other means.  This can be done by manually calling
+//! [`Endpoint::add_node_addr`] but that still requires knowing the other addressing
```
matheus23, line 9:
```suggestion
//! [`Endpoint::add_node_addr`], but that still requires knowing the other addressing
```

matheus23 (approved):

flub:
I really need to work on my commas... thanks!

PR #2474, opened 2024-07-08
fix(iroh-docs): ensure docs db write txn gets closed regularly under all circumstances

rklaehn:
## Description

The docs db uses batching to reduce the number of write transactions. Before this, under certain circumstances (lots of activity, exclusively writes) the transaction would remain open for a long time.

This adds logic to modify to make sure that transactions will be committed when they get too old in any case.

## Breaking Changes

None

## Change checklist

- [x] Self-review.
- [x] ~~Tests if relevant.~~

### DIFF
--- a/iroh-docs/src/actor.rs
+++ b/iroh-docs/src/actor.rs
@@ -28,7 +28,7 @@ use crate::{
 };
 
 const ACTION_CAP: usize = 1024;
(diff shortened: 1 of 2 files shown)


matheus23 (approved):
This looks right!

PR #2476, opened 2024-07-08
fix(iroh-net): unexpected `cfg` condition values / possible fix on netbsd

matheus23:
## Description

We had some CLI tests that had `#[cfg(feature = "fs-store")]`, but that feature doesn't exist in `iroh-cli`... so they were never compiled (and some related code needed for these tests was deleted. I reintroduced it).

This also fixes what is likely a bug, where `#[cfg(target_arch = "netbsd")]` was used, but there is no "architecture" netbsd. Only an *OS* netbsd.

## Breaking Changes

None

## Notes & open questions

--

## Change checklist

- [X] Self-review.
- ~~[ ] Tests if relevant.~~

### DIFF
--- a/iroh-cli/tests/cli.rs
+++ b/iroh-cli/tests/cli.rs
@@ -108,11 +108,10 @@ fn cli_provide_tree() -> Result<()> {
     test_provide_get_loop(Input::Path(path), Output::Path)
 }
 
-#[cfg(feature = "fs-store")]
 #[test]
 #[ignore = "flaky"]
 fn cli_provide_tree_resume() -> Result<()> {
-    use iroh_blobs::store::file::test_support::{make_partial, MakePartialResult};
+    use iroh::blobs::store::fs::test_support::{make_partial, MakePartialResult};
 
     /// Get all matches for match group 1 (an explicitly defined match group)
     fn explicit_matches(matches: Vec<(usize, Vec<String>)>) -> Vec<String> {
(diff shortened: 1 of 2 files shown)


dignifiedquire (approved):

iroh-net/src/net/interfaces/bsd.rs
```diff
@@ -374,9 +374,9 @@ impl WireFormat {
                     return Err(RouteError::InvalidMessage);
                 }
 
-                #[cfg(target_arch = "netbsd")]
+                #[cfg(target_os = "netbsd")]
                 let index = u16_from_ne_range(data, 16..18)?;
-                #[cfg(not(target_arch = "netbsd"))]
+                #[cfg(not(target_os = "netbsd"))]
                 let index = u16_from_ne_range(data, 12..14)?;
```
matheus23, line 380: Looks like this was [introduced](https://github.com/n0-computer/iroh/pull/1472/files#diff-75f83c1e913d60def929bd1986bb190036c81a21cbf37b384c11b5251347503fR290-R293) in https://github.com/n0-computer/iroh/pull/1472 by @dignifiedquire.

Given we've been running with this for a while now, do you know if "fixing" this cfg will actually make things better, not worse? @dignifiedquire

dignifiedquire: I suspect we just never hit this branch in practice yet, but the cfg is the only part that's wrong asfatict

iroh-cli/tests/cli.rs
```diff
@@ -1034,3 +1032,51 @@ where
...
+fn copy_dir_all(src: impl AsRef<Path>, dst: impl AsRef<Path>) -> anyhow::Result<usize> {
+    let src = src.as_ref();
+    let dst = dst.as_ref();
+    std::fs::create_dir_all(dst)?;
+    let mut len = 0;
+    for entry in std::fs::read_dir(src)? {
+        let entry = entry.with_context(|| {
+            format!(
+                "failed to read directory entry in `{}`",
+                src.to_string_lossy()
```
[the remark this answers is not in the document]
matheus23: Yeah you're completely right.
To be fair I mostly resurrected this code from history without reviewing it much myself :sweat_smile:

PR #2482, opened 2024-07-09
fix(iroh-gossip): gossip dispatcher: reliable events on join, allow dropping sinks or streams

Frando:
## Description

This has 2 fixes to the gossip dispatcher:

* When a topic is joined, it can currently happen that the first few events on the newly joined topic are missed, because the events can start to arrive at the dispatcher's global subscription before the dispatcher's join task finished and changed the protocol state. If this happens, we now change the topic state to live before processing the incoming event. This should make the existing smoke test non-flaky.

* Currently a topic subscription is quitted when either the command stream (from client to gossip) or the event stream (from gossip to client) is dropped. Now we only quit the subscription if both the command and the event channels are dropped. This should fix the test in iroh-ffi.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/tests/client.rs
+++ b/iroh/tests/client.rs
@@ -58,7 +58,6 @@ fn await_messages(
 }
 
 #[tokio::test]
-#[ignore = "flaky"]
 async fn gossip_smoke() -> TestResult {
     let _ = tracing_subscriber::fmt::try_init();
     let (addr1, node1) = spawn_node();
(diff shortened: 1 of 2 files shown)


dignifiedquire (approved):

iroh/tests/client.rs
```diff
@@ -58,7 +58,6 @@ fn await_messages(
 }
 
 #[tokio::test]
-#[ignore = "flaky"]
```
matheus23, line 61: Wooohoo! :tada:

PR #2485, opened 2024-07-10
docs(iroh-net): fix broken HTTP/3 link

link2xt:
Also make it non-monospace as it is not a crate
name or source code identifier.


matheus23 (approved):
Thank you :)

PR #2487, opened 2024-07-10
fix(iroh-net): Stable select of direct address guess

### LINKED ISSUE #2480: iroh-net: When guessing at a direct address we do not always send to the same one
When we have a relay connection and some direct addresses, but have not yet figured out any information about those direct addresses - like whether they work- we try to pick a random direct address from the list and send to relay + this guess at the same time.

The idea is that maybe we picked a usable address and it just works and we figure this out and start using this path for real.

However we do not stably select the direct address path, so each time we send a packet a new direct address path is being selected.  That's rather useless.

This is the bug: https://github.com/n0-computer/iroh/blob/4640f4d0049e582cc89d81e976b59961043efb3f/iroh-net/src/magicsock/node_map/node_state.rs#L295

flub:
## Description

When there are direct addresses but no best address a random one is
chosen just in case it happens to work (this could be because it was
added manually and is the only one).  However it is important to
select the same destination each time an address is needed in this
situation, so that all packets are sent to the same address instead of
only a fraction of the packets being delivered there.

The previous implementation did not result in a stable selection.
This version does work stably.  It stores the selection as NodeState
so that it is cheaper to find it again, compared to invoking an RNG,
since this is on the hot path of poll_send.  However the selection
still needs to be random since otherwise it could be used to direct
phoney traffic.

The state is never cleared, only updated when the number of direct
addresses change.  This is sufficient to ensure a reasonable
unpredictable address is selected.

## Breaking Changes

none

## Notes & open questions

Fixes #2480.

I don't particularly like the extra state/fields on NodeState for
this.  It feels like the abstraction should be moved to another
level as there are already plenty of things on NodeState that do
not need to interact with each other.  But that refactor would
likely involve BestAddr so I shied away from this.  Maybe that's
a bit lazy.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/src/magicsock/node_map/node_state.rs
+++ b/iroh-net/src/magicsock/node_map/node_state.rs
@@ -2004,7 +2004,7 @@ mod tests {
                     break;
                 }
                 // Didn't select a different one, that's possible since we still use a
-                // random one out of the possiblities.  Let's try again.
+                // random one out of the possibilities.  Let's try again.
             }
         }
     }
--- a/iroh-net/src/magicsock/node_map/node_state.rs
+++ b/iroh-net/src/magicsock/node_map/node_state.rs
@@ -6,7 +6,7 @@ use std::{
 };
 
 use iroh_metrics::inc;
-use rand::seq::IteratorRandom;
+use rand::Rng;
 use serde::{Deserialize, Serialize};
 use tokio::sync::mpsc;
 use tracing::{debug, info, instrument, trace, warn};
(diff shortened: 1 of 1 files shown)


iroh-net/src/magicsock/node_map/node_state.rs
```diff
@@ -311,6 +315,41 @@ impl NodeState {
...
+    /// When we pick a random address however, we want this to remain stable and always pick
+    /// the same one as long the candidate addresses did not change.
+    fn stable_random_candidate_addr(&mut self, have_ipv6: bool) -> Option<SocketAddr> {
+        let candidate_iter = || {
+            self.direct_addr_state.keys().filter(|ipp| match ipp.ip() {
+                IpAddr::V4(_) => true,
+                IpAddr::V6(_) => have_ipv6,
+            })
+        };
+        let candidate_len = dbg!(candidate_iter().count());
```
flub, line 333: All these `dbg!` calls still need to be removed :)

matheus23:
>  It feels like the abstraction should be moved to another level as there are already plenty of things on NodeState that do not need to interact with each other.

Yeah - came here to write about a different way to refactor this :sweat_smile: 

I think we should probably
- store the result of `addr_for_send` as a whole
- see what data dependencies `addr_for_send` has that make it change result value
- look at all the places these dependent values are changed, and make sure that we recompute the send addr at that point in time.

Given that it's probably all internal code, we should be able to find the places where data dependent values for this function are modified exhaustively.

---

I'm not super familiar with the code - so I'm not super confident that I'm correct, so I'd like to know what you think :v:

That said, this PR is probably already a huge improvement. If it turns out my proposal for a refactor is too much work, perhaps we merge this in the mean time.

flub:
Superseded by #2509

PR #2489, opened 2024-07-11
refactor(iroh-net): Switch to new iroh-relay route `/relay` instead of `/derp`

matheus23:
Split out from #2419, copy of #2441, because I forgot to switch the base branch :upside_down_face: 

Splitting this out means we can deploy the relay changes one release in advance to the changes in the client.

Otherwise, we'd need to update the relays ASAP when we release 0.20, as clients will expect to be able to connect to the `/relay` route (which may not be deployed immediately after the release).

## Description

- Switches the client to connect to `/relay` instead of `/derp`.

## Breaking Changes

None

## Notes & open questions

Also a PR related to #2378 

I allowed myself to add in another `derive_more::Debug` instance instead of a custom, outdated `impl Debug`, if that's okay with you :grin: 

## Change checklist

- [X] Self-review.
- [X] Tests if relevant.


matheus23:
Base branch didn't switch automatically in the original PR... Will need to address conflicts today :v:

matheus23:
This PR contains the same +5 -12 lines of code change that #2441 does (I verified this manually).

dignifiedquire:
one issue with this I see is that relays that are not updated will not work anymore, so we might want to wait even one more release to merge this

matheus23:
@dignifiedquire one release more seems somewhat arbitrary.
Can we tell partners that run their own relay that they have to update their relays before they update their clients?
This will be true regardless of when we release this change.

matheus23:
@dignifiedquire looks like you've solved the social problem, given this is now assigned the 0.21 milestone.
Can I have a quick :+1: on this PR? It's actually the same code as [another PR that was already accepted](https://github.com/n0-computer/iroh/pull/2441).

dignifiedquire (approved):

PR #2495, opened 2024-07-12
feat(iroh-net): Add holepunching events

flub:
## Description

The aim here is to make users' reports about holepunching problems
actionable.  For this it creates a curated set of events.  This is
much more efficient than trawling through the entire logfile.

The events are emitted for logical reasons.  This means they might not
be emitted in a place where they are guaranteed to have happened, but
rather are emitted where it logically makes sense and all the information
is available.  This is fine since they are not meant to debug the code
itself.  E.g. after a `call-me-maybe.sent` event has been emitted the
packet could still be dropped in the queue of the relay client.  From
the event point of view this is fine, the event happened but the peer
may not have seen the result of this event, which will be visible in the
peer's own events.

Events have all their info in the fiels (and the target itself) and
do not rely on spans geneerally.  However the exception is the `NodeId`
of the endpoint emitting the event which has to be supplied somewhere in
a span.

See also the README for some design behind the events.

Because emitting useful events reveals bugs, this also fixes some:

- `store_endpoints_update` spawned without attaching a span, it now
  preserves the span.

- Since pings are sent at `HEARTBEAT_INTERVAL` they can often be received
  a little later.  This was immediately interpreted as a missed ping,
  which was too eager and resulted in the `PathState` flip-flopping too
  much.  The time window to receive heartbeat pings has been extended a
  little.

Other changes:

- `PingRole::Reactivate` is called `Activate` for now.  This is most of
  the time more accurate.  If needed we can add `Reactivate` as an
  additional role, but I'm not yet sure this is helpful.

- `PathState` now keeps track of the `NodeId` and path (`SendAddr`) it
  applies to.  This makes it easier to reason about things, however
  duplicates this information from where it is stored in the parent struct.

- `DiscoPingPurpose::PingBack` has been explicitly added to make the
  events a bit more meaningful.

## Breaking Changes

none

## Notes & open questions

It is arguable whether this becomes just a different log file source
over time.  I hope that with some effort we can avoid this.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/src/magicsock.rs
+++ b/iroh-net/src/magicsock.rs
@@ -1200,29 +1200,29 @@ impl MagicSock {
         }
     }
 
-    fn send_or_queue_call_me_maybe(&self, url: &RelayUrl, node_id: NodeId) {
+    fn send_or_queue_call_me_maybe(&self, url: &RelayUrl, dst_node: NodeId) {
         let endpoints = self.endpoints.read();
         if endpoints.fresh_enough() {
             let addrs: Vec<_> = endpoints.iter().collect();
             event!(
                 target: "events.net.call-me-maybe.sent",
                 Level::DEBUG,
-                dst_node = %node_id.fmt_short(),
+                dst_node = %dst_node.fmt_short(),
                 via = ?url,
                 ?addrs,
             );
             let msg = endpoints.to_call_me_maybe_message();
             let msg = disco::Message::CallMeMaybe(msg);
-            if !self.send_disco_message_relay(url, node_id, msg) {
-                warn!(dstkey = %node_id.fmt_short(), relayurl = ?url,
+            if !self.send_disco_message_relay(url, dst_node, msg) {
+                warn!(dstkey = %dst_node.fmt_short(), relayurl = ?url,
                       "relay channel full, dropping call-me-maybe");
             } else {
-                debug!(dstkey = %node_id.fmt_short(), relayurl = ?url, "call-me-maybe sent");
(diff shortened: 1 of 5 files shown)


iroh-net/src/magicsock.rs
```diff
@@ -1181,21 +1200,29 @@ impl MagicSock {
         }
     }
 
-    fn send_or_queue_call_me_maybe(&self, url: &RelayUrl, dst_key: PublicKey) {
+    fn send_or_queue_call_me_maybe(&self, url: &RelayUrl, node_id: NodeId) {
```
dignifiedquire, line 1203: maybe keep the destination part, to make it clear this is not our own `node_id`? eg `dst_node_id`

iroh-net/src/magicsock/node_map/node_state.rs
```diff
@@ -1203,7 +1290,7 @@ impl PathState {
     ///
     /// This is the most recent instant between:
     /// - when last pong was received.
-    /// - when the last CallMeMaybe was received.
+    /// - when the last CallMeMaybe was received.  WRONG
```
dignifiedquire, line 1293: WRONG?

iroh-net/src/relay/http/client.rs
```diff
@@ -623,6 +623,14 @@ impl Actor {
             return Err(ClientError::Send);
         }
 
+        event!(
+            target: "events.net.relay.connected",
+            // parent: &span!(Level::TRACE, "events"),
```
dignifiedquire, line 628: ?

PR #2496, opened 2024-07-12
ci: add a flaky tests failure report to our discord notification

divagant-martian:
## Description

Modifies the `flaky` workflow to generate a report with all the tests that failed per matrix combo, so that we can receive it on discord. This requires to modify the `tests` workflow as well to generate `libtest` json reports.

Reports are uploaded as job artifacts with an unique name, and retained for a day.

From my last commit at the time of writing, this is the report we get:
```md
Flaky tests failure:

- **ubuntu-latest all stable**
iroh-net::iroh_net::discovery::local_swarm_discovery::tests::test_local_swarm_discovery
ubuntu-latest default stable
iroh-cli::cli::cli_bao_store_migration
- **windows-latest all stable**
iroh-cli::cli::cli_provide_file_resume
iroh-cli::cli::cli_provide_tree_resume
- **windows-latest default stable**
iroh-cli::cli::cli_provide_file_resume
iroh-cli::cli::cli_provide_tree_resume
- **windows-latest none stable**
iroh-cli::cli::cli_provide_tree_resume
iroh-cli::cli::cli_provide_file_resume

See https://github.com/n0-computer/iroh/actions/workflows/flaky.yaml
```
which reads clear enough in discord imo

## Breaking Changes
n/a

## Notes & open questions

- there is another test report which is more widely used but for which there is not a single decent parser I could find. The format is Jenkins' junit xml format. From my searches, everyone knows how to produce these files, but not how to read them. Since this is xml, which is not exactly compatible with serde, reading these kind of files was a task I considered not worth doing for what we actually want to achieve, which is simply more visibility over flaky tests.
- That being said, `libtest` format is not stable, and the `nextest` feature to obtain it is unstable as well. It does not seem to change quickly or drastically at all, so I think we will be fine for some time. Being json the underlying format, it should be easy to adapt if necessary.

## Change checklist

- [x] Self-review.
- [ ] ~~Tests if relevant.~~

### DIFF
--- a/.github/workflows/flaky.yaml
+++ b/.github/workflows/flaky.yaml
@@ -64,7 +64,7 @@ jobs:
       - name: read results
         run: |
           ls nextest-results
-          for result in nextest-results/*.json; do
+          for result in nextest-results/libtest_run_${{ github.run_number }}-${{ github.run_attempt }}*.json; do
             echo "$result"
             jq --slurp '.[] | select(.type == "test" and .event == "failed" ) | .name' $result
           done
--- a/.github/workflows/flaky.yaml
+++ b/.github/workflows/flaky.yaml
@@ -79,6 +79,8 @@ jobs:
           severity: warn
           details: |
             Flaky tests failure:
+
             ${{ steps.make_summary.outputs.summary }}
+
             See https://github.com/n0-computer/iroh/actions/workflows/flaky.yaml
           webhookUrl: ${{ secrets.DISCORD_N0_GITHUB_CHANNEL_WEBHOOK_URL }}
--- a/.github/workflows/flaky.yaml
+++ b/.github/workflows/flaky.yaml
@@ -61,6 +61,7 @@ jobs:
           path: nextest-results
           pattern: libtest_run_${{ github.run_number }}-${{ github.run_attempt }}-*
           merge-multiple: true
+      - name: read results
         run: |
           ls nextest-results
           for result in nextest-results/*.json; do
--- a/.github/workflows/flaky.yaml
+++ b/.github/workflows/flaky.yaml
@@ -66,7 +66,7 @@ jobs:
           for report in nextest-results/*.json; do
             echo $report
             cat $report
-            jq '.[] | select(.type == "test" and .event == "failed" ) | .name' $report
+            jq '.[]' $report
           done
       - name: Notify discord on failure
         uses: n0-computer/discord-webhook-notify@v1
--- a/.github/workflows/flaky.yaml
+++ b/.github/workflows/flaky.yaml
@@ -71,7 +71,7 @@ jobs:
             name=$(echo ${report:16:-5} | tr _ ' ')
             echo $name
             echo "- **$name**" >> $GITHUB_OUTPUT
-            failure=$(jq --slurp '.[] | select(.["type"] == "test" and .["event"] == "failed" ) | ("`" + .["name"]) + "`"' $report | sed -e 's/"//g' -e 's/\$/::/')
+            failure=$(jq --slurp '.[] | select(.["type"] == "test" and .["event"] == "failed" ) | ("`" + .["name"]) + "`\n"' $report | sed -e 's/"//g' -e 's/\$/::/')
             echo $failure
             echo $failure >> $GITHUB_OUTPUT
           done
--- a/.github/workflows/flaky.yaml
+++ b/.github/workflows/flaky.yaml
@@ -68,10 +68,10 @@ jobs:
           EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
           echo "summary<<$EOF" >> $GITHUB_OUTPUT
           for report in nextest-results/*.json; do
-            name=$(echo ${report:35:-5} | tr _ ' ')
+            name=$(echo ${report:17:-5} | tr _ ' ')
             echo $name
             echo "- **$name**" >> $GITHUB_OUTPUT
-            failure=$(jq --slurp '.[] | select(.["type"] == "test" and .["event"] == "failed" ) | ("\t- " + .["name"])' $report | sed -e 's/"//g' -e 's/\$/::/')
+            failure=$(jq --slurp '.[] | select(.["type"] == "test" and .["event"] == "failed" ) | ("`" + .["name"]) + "`"' $report | sed -e 's/"//g' -e 's/\$/::/')
             echo $failure
             echo $failure >> $GITHUB_OUTPUT
           done
--- a/.github/workflows/flaky.yaml
+++ b/.github/workflows/flaky.yaml
@@ -66,8 +66,9 @@ jobs:
         run: |
           # prevent the glob expression in the loop to match on itself when the dir is empty
           shopt -s nullglob
-          # to deal with multiline outputs it's recommened to use a random EOF
-          EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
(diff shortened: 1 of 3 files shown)


.github/workflows/flaky.yaml
```diff
@@ -55,12 +55,36 @@ jobs:
...
+        uses: actions/download-artifact@v4
+        with:
+          pattern: libtest_run_${{ github.run_number }}-${{ github.run_attempt }}-*
+          merge-multiple: true
+          path: nextest-results
+      - name: create summary report
+        id: make_summary
+        run: |
+          shopt -s nullglob
+          EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
```
flub, line 68: why not pick something fixed that's not valid json?  in theory this could produce valid json i guess?

divagant-martian: this is just a string of characters, I commented the line since getting a new one every time seems unnecessary, but the line tells us how we got it. The syntax 

```
{name}<<{delimiter}
{value}
{delimiter}
```

source: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings

dignifiedquire (approved):
the results are great, for github action yml review code, I delegate to others 😅

.github/workflows/tests.yaml
```diff
@@ -112,9 +112,20 @@ jobs:
...
       run: |
-        cargo nextest run --workspace ${{ env.FEATURES }} --lib --bins --tests --run-ignored ${{ inputs.flaky && 'all' || 'default' }} --no-fail-fast
+        mkdir -p output
+        cargo nextest run --workspace ${{ env.FEATURES }} --lib --bins --tests --run-ignored ${{ inputs.flaky && 'all' || 'default' }} --no-fail-fast --message-format ${{ inputs.flaky && 'libtest-json' || 'human' }} > output/${{ matrix.name }}_${{ matrix.features }}_${{ matrix.rust }}.json
       env:
         RUST_LOG: ${{ runner.debug && 'TRACE' || 'DEBUG'}}
+        NEXTEST_EXPERIMENTAL_LIBTEST_JSON: 1
+
+    - name: upload results
+      if: ${{ failure() && inputs.flaky }}
```
flub, line 122: why only upload on failure and not always upload?  If it way always uploaded we could also use this info to show statistics of how flaky a test is if we ever felt like doing so.  I think there's very little harm to always upload.  AFAIK github expires artifacts after some time anyway, and I doubt our storage use is a problem right now.

.github/workflows/tests.yaml
```diff
@@ -112,9 +112,20 @@ jobs:
...
+      uses: actions/upload-artifact@v4
+      with:
+        name: libtest_run_${{ github.run_number }}-${{ github.run_attempt }}-${{ matrix.name }}_${{ matrix.features }}_${{ matrix.rust }}.json
+        path: output
+        retention-days: 1
```
flub, line 127: why only keep them one day?  why not keep them for like 60 days?  Do we have storage limitation issues?

.github/workflows/flaky.yaml
```diff
@@ -55,12 +55,42 @@ jobs:
...
+          merge-multiple: true
+          path: nextest-results
+      - name: create summary report
+        id: make_summary
+        run: |
+          # prevent the glob expression in the loop to match on itself when the dir is empty
+          shopt -s nullglob
+          # to deal with multiline outputs it's recommended to use a random EOF
+          # EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
+          EOF=aP51VriWCxNJ1JjvmO9i
```
flub, line 71: any string that's not valid json should do no?  is leaving the note about random useful?

flub (approved):
Very nice!  Love the result

.github/workflows/flaky.yaml
flub: ah, whoops.  made the same comment twice :)

.github/workflows/tests.yaml
divagant-martian: since we only use them for failures.. it makes sense to just upload on failure

.github/workflows/tests.yaml
divagant-martian: not really (or more precisely, I have no idea), but to me it makes to keep them the least amount of time necessary to achieve what we want, which is an "immediate report"

.github/workflows/flaky.yaml
divagant-martian: If you want I can remove it. It's a bit esoteric, so I prefer some comment - any comment - over nothing. 

I am not sure any non json is valid. Maybe a better comment would be a reference to the syntax, which explains the why, rather than command, which explains the how

divagant-martian:
@flub from your comments I see you want to take this further by running statistics. I don't have a problem with that but I hope we don't actually ever _need_ it. Our focus should be into reducing the flakes rather than improving the reports. Of course, given history is more than reasonable to be skeptical. 

That being said, changing this would also require changes to the `flaky` workflow since the assumptions is that if we download a report, then it means it contains at least one failure. Uploading always would produce a report with matrix names but no content (no test failures).

So I suggest we increase the retention period, provided @Arqu gives us a good-to-go on a reasonable number of days and deal with statistics in a later PR

.github/workflows/tests.yaml
divagant-martian: closing to continue in main thread :)

.github/workflows/flaky.yaml
divagant-martian: done in d94b7b67987f874435dcebc95f5f37043645d426

Arqu (approved):
🔥 

As for the stats/report retention. I don't think it's a piece of work worth our time currently but probably a good addition somewhere down the line.

Think the max retention is 90 days. Given these are super light I would stick it to something like 45 days. That's plenty of history IMHO, and wont really clog up much of our storage usage.

Arqu:
😅

.github/workflows/tests.yaml
```diff
@@ -200,6 +211,17 @@ jobs:
...
         RUST_LOG: ${{ runner.debug && 'TRACE' || 'DEBUG'}}
+        NEXTEST_EXPERIMENTAL_LIBTEST_JSON: 1
+  
+    - name: upload results
+      if: ${{ failure() && inputs.flaky }}
+      uses: actions/upload-artifact@v4
+      with:
+        name: libtest_run_${{ github.run_number }}-${{ github.run_attempt }}-${{ matrix.name }}_${{ matrix.features }}_${{ matrix.rust }}.json
+        path: output
+        retention-days: 1
```
flub, line 226: missed this one :face_with_diagonal_mouth:

flub:
> @flub from your comments I see you want to take this further by running statistics. I don't have a problem with that but I hope we don't actually ever _need_ it. Our focus should be into reducing the flakes rather than improving the reports. Of course, given history is more than reasonable to be skeptical.
> 
> That being said, changing this would also require changes to the `flaky` workflow since the assumptions is that if we download a report, then it means it contains at least one failure. Uploading always would produce a report with matrix names but no content (no test failures).
> 
> So I suggest we increase the retention period, provided @Arqu gives us a good-to-go on a reasonable number of days and deal with statistics in a later PR

I think the approach you took of keeping things a bit longer just in case we want to do anything with them is fine for now.  This is a great improvement and I'm all for small improvements as we need them.  So we can do statistics when we want to handle statistics.

But also your point of we actually strive for 0 is very good.

Thanks for doing all this!

Issue #1859, opened 2023-12-03
doc get <key> not returning hash for large (1.5Mb) document import


ntangldev:
I am running the iroh CLI 0.11.0 on Mac M1 Sonoma 14.0

I installed iroh using 

`curl -fsSL https://sh.iroh.computer/install.sh | sh
`

I have imported a 1.5Mb document

When I try to get the document I was expecting to get the hash returned, but with either the default mode or explicitly requesting hash mode no response is returned.

`author:szozlqglrywojk35 doc:owcr5mvpvzhf4ics
> doc import ~/projects/testblob
Preparing import...
Import 1 files totaling 1.52 MiB? yes
Success! (0 seconds)

author:szozlqglrywojk35 doc:owcr5mvpvzhf4ics
> doc keys
@szozlqglrywojk35: testblob = lsoaqmjoqh4bkc44 (1.52 MiB)

author:szozlqglrywojk35 doc:owcr5mvpvzhf4ics
> doc get testblob

author:szozlqglrywojk35 doc:owcr5mvpvzhf4ics
> doc get --mode hash testblob

`

For small (3char) values the key/value is returned

ramfox:
Closing because docs have been moved out of iroh and we no longer have an `iroh` CLI

Issue #1619, opened 2023-10-11
appending `\0` aka the null byte character to keys to prevent unintended prefix deletion


ramfox:
An entry w/ the key `foobar`, will get overridden by an entry with the key `foo`
However an entry w/ the key `foobar` will not get overridden by an entry with the key `foo\0`

to prevent our APIs from doing unexpected and unwanted behaviour we can either append the null byte to a key:
- at the top level API in the iroh cli / console, append the null byte as expected
- one layer down: have all api calls append and remove the null byte so that it is part of the iroh API and it is expected that we take care of this as a library
- at the document level, and append the null byte for all `set`-ing and `get`-ing needs

Regardless of where this gets implemented, we need to add some kind of `prefix` flag on all `set` and `get`-like actions that will inform us that the user *intends* to do batch work on all entries with keys underneath that prefix.

rklaehn:
Hm, what about not doing this and instead telling users about how prefix deletion works?

Prefix deletion/replacement is a *very* useful feature if you know how to use it, and always appending a 0 byte would close this down to the user.

If we do this at all, I would have it **just** as an option of the cli / console that you can opt out of with an extra arg --raw or whatever. So for me option 1 makes most sense.

matheus23:
Personally I think using an enum for rust APIs:
```rust
pub enum Key<'a> {
    Full(&'a [u8]),
    Prefix(&'a [u8]),
}
```

and a `--prefix` flag for CLI commands: `iroh doc set <KEY> --prefix`.

It looks like `iroh doc get` already has a `--prefix` flag, but `set` apparently doesn't.

ramfox:
We have `path_to_key` and `key_to_path` methods that take care of this now on the CLI

PR #2497, opened 2024-07-13
Fix: Can't export from CLI because of the null byte in the key last byte

israelyago:
## Description
- Fix a bug that the user can't export a file because the key wouldn't match, because internally it was being added a null byte at the end of the key. Possibly, it is the bug presented in #1859
E.g: 
`doc import path/to/file/name_here`
`doc export name_here -o target/dir` would have searched for `name_here`, but would not match the internal `name_here\0` (Null byte at the end), therefore, never exporting the file.

- Relates to #1619

## Breaking Changes

- Possible breaking change: Removes the `\0` (Null byte) at the end of a key in file imported from the CLI

## Notes & open questions
- If the `\0` (Null byte) has some actual definition and purpose in the key, it is needed to also apply this byte to the CLI command  `doc set foo bar`, which does not seem to add the Null byte

- I may have missed something, but the original commit that introduced the char insertion did not make it clear what purpose it serves (https://github.com/n0-computer/iroh/pull/1793/files#diff-6cb6932fd260f19df9c3f69652d5115449234fb6e0f3931e64d356962f47e01bR286)

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/util/fs.rs
+++ b/iroh/src/util/fs.rs
@@ -215,19 +215,14 @@ fn path_content_info0(path: impl AsRef<Path>) -> anyhow::Result<PathContent> {
 ///
 /// If `prefix` exists, it will be stripped before converting back to a path
 /// If `root` exists, will add the root as a parent to the created path
-/// Removes any null byte that has been appended to the key
 pub fn key_to_path(
     key: impl AsRef<[u8]>,
     prefix: Option<String>,
     root: Option<PathBuf>,
 ) -> anyhow::Result<PathBuf> {
-    let mut key = key.as_ref();
+    let key = key.as_ref();
     if key.is_empty() {
         return Ok(PathBuf::new());
-    }
-    // if the last element is the null byte, remove it
-    if b'\0' == key[key.len() - 1] {
-        key = &key[..key.len() - 1]
     }
 
     let key = if let Some(prefix) = prefix {
@@ -265,8 +260,6 @@ pub fn key_to_path(
 }
 
 /// Helper function that creates a document key from a canonicalized path, removing the `root` and adding the `prefix`, if they exist
-///
-/// Appends the null byte to the end of the key.
 pub fn path_to_key(
     path: impl AsRef<Path>,
     prefix: Option<String>,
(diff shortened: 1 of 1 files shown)


matheus23:
Thank you so much for contributing! I don't think we can merge this exactly as-is. Perhaps it's best if I answer some of your questions first:

> the original commit that introduced the char insertion did not make it clear what purpose [the null byte] serves

It's pretty technical.
Deletions in [willow] (the specification that iroh documents are based on) work by overwriting a key with a tombstone.
Think of deleting a value at `path/to/value` by overwriting `path/to/value` with something indicating a deletion.

These paths are hierarchical, though, so that means that you can also delete whole "trees" of keys by writing something that indicates deletion to e.g. `path/to`, and then `path/to/value` and any other `path/to/x` will be gone.

The problem is that sometimes your intention is to delete a *specific key* not a whole tree. Willow by itself doesn't distinguish these, but it's easy to modify the key structure to support this distinction: That's where the trailing null byte comes in.
To indicate that a key is supposed to represent a single key, not a whole tree, you add a trailing null byte. This way, if you delete `path/to/value\0`, that's not going to be a prefix of `path/to/value/child\0`, so overwriting it won't delete the child.
But you still retain the option to delete whole trees by leaving out the trailing null byte on an operation.

> * If the `\0` (Null byte) has some actual definition and purpose in the key, it is needed to also apply this byte to the CLI command  `doc set foo bar`, which does not seem to add the Null byte

Yes. My personal opinion is that the `iroh doc set` command should work like `iroh doc get` with respect to the key handling. The `doc get` command supports a `--prefix` option which skips adding the null byte suffix. But I'll let the other's from the iroh team weigh in on that.

> Possible breaking change: Removes the `\0` (Null byte) at the end of a key in file imported from the CLI

We should keep the null byte suffix for the prefix deletion reasons I explained above.

[willow]: https://willowprotocol.org

dignifiedquire:
> Yes. My personal opinion is that the `iroh doc set` command should work like `iroh doc get` with respect to the key handling. The `doc get` command supports a `--prefix` option which skips adding the null byte suffix. But I'll let the other's from the iroh team weigh in on that.

yes, agreed

dignifiedquire:
Closing as for the above described reasons. Lets keep the conversation in the mentioned issues going.

PR #2508, opened 2024-07-16
refactor(iroh): make use of quic-rpc-derive macros to prettify the rpc declarations

rklaehn:
## Description

Make use of quic-rpc-derive macros to make the rpc declarations nicer to read. With this, you can clearly see the kind of interaction and the response type for each rpc message type in one place. Under the hood, this does the exact same thing as before.

## Breaking Changes

None

## Change checklist

- [x] Self-review.
- [x] ~~Tests if relevant.~~

### DIFF
--- a/iroh/src/rpc_protocol/node.rs
+++ b/iroh/src/rpc_protocol/node.rs
@@ -3,6 +3,7 @@ use std::collections::BTreeMap;
 use iroh_base::rpc::RpcResult;
 use iroh_net::{endpoint::ConnectionInfo, key::PublicKey, relay::RelayUrl, NodeAddr, NodeId};
 use quic_rpc::message::{Msg, RpcMsg, ServerStreaming, ServerStreamingMsg};
+use quic_rpc_derive::rpc_requests;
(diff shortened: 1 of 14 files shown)


iroh/src/rpc_protocol/node.rs
```diff
@@ -11,12 +13,18 @@ use super::RpcService;
...
     Status(StatusRequest),
+    #[rpc(response = RpcResult<NodeId>)]
     Id(IdRequest),
+    #[rpc(response = RpcResult<NodeAddr>)]
     Addr(AddrRequest),
+    #[rpc(response = RpcResult<()>)]
     AddAddr(AddAddrRequest),
+    #[rpc(response = RpcResult<Option<RelayUrl>>)]
     Relay(RelayRequest),
     Stats(StatsRequest),
```
dignifiedquire, line 29: what about these?

rklaehn: oops

flub (approved):

PR #2509, opened 2024-07-16
refactor(iroh-net): remove random choice of direct addr

### LINKED ISSUE #2480: iroh-net: When guessing at a direct address we do not always send to the same one

flub:
## Description

When we had any direct addresses but not yet a best address we would
randomly try to pick one and send to it on the off chance it worked.
This required to always pick the same direct addr to send to which was
not happening and a bug.

However the scenarios in which this would speed up holepunching were
very unlikely: only really if a single direct addr known to work was
added manually would this improve things.

Simply removing this feature is rather nice:

- The direct connection still happens very quickly in the best-case
  scenario: a single ping-pong roundtrip is all it costs.  No relay
  server or discovery mechanism needed so no functionality change.

- In all other scenarios there is no change.  Connection establishment
  behaves exactly the same as before.

- There are no more confusing connection type changes at startup.
  This would go relay -> mixed (-> mixed a few times more due to the
  bug) -> direct.  Now it just goes relay -> direct which is intuitively
  what people would expect.

The simplest way to test this is by running doctor accept and connect
with direct addrs only and observe the holepunching events emitted.

## Breaking Changes

none

## Notes & open questions

This is an alternative to #2487.  And my preferred solution.

Fixes #2480.

See #2512 for the flaky test.

test_two_devices_roundtrip_network_change managed to hit a timeout on
cross CI.  Give it some more time.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/src/magicsock/node_map/node_state.rs
+++ b/iroh-net/src/magicsock/node_map/node_state.rs
@@ -6,7 +6,6 @@ use std::{
 };
 
 use iroh_metrics::inc;
-use rand::seq::IteratorRandom;
 use serde::{Deserialize, Serialize};
 use tokio::sync::mpsc;
 use tracing::{debug, event, info, instrument, trace, warn, Level};
@@ -259,11 +258,7 @@ impl NodeState {
     /// Returns the address(es) that should be used for sending the next packet.
     ///
     /// Any or all of the UDP and relay addrs may be non-zero.
-    fn addr_for_send(
-        &mut self,
-        now: &Instant,
-        have_ipv6: bool,
-    ) -> (Option<SocketAddr>, Option<RelayUrl>) {
+    fn addr_for_send(&mut self, now: &Instant) -> (Option<SocketAddr>, Option<RelayUrl>) {
         if relay_only_mode() {
             debug!("in `DEV_relay_ONLY` mode, giving the relay address as the only viable address for this endpoint");
             return (None, self.relay_url());
@@ -287,20 +282,8 @@ impl NodeState {
                 (Some(best_addr.addr), self.relay_url())
             }
             best_addr::State::Empty => {
-                // No direct connection has been used before.  If we know of any possible
-                // candidate addresses, randomly try to use one while also sending via relay
-                // at the same time.
-                let addr = self
-                    .direct_addr_state
-                    .keys()
-                    .filter(|ipp| match ipp.ip() {
-                        IpAddr::V4(_) => true,
-                        IpAddr::V6(_) => have_ipv6,
-                    })
-                    .choose_stable(&mut rand::thread_rng())
-                    .map(|ipp| SocketAddr::from(*ipp));
-                trace!(udp_addr = ?addr, "best_addr is unset, use candidate addr and relay");
-                (addr, self.relay_url())
+                trace!("best_addr is unset, use relay");
+                (None, self.relay_url())
             }
         };
         let typ = match (best_addr, relay_url.clone()) {
@@ -1103,11 +1086,10 @@ impl NodeState {
     #[instrument("get_send_addrs", skip_all, fields(node = %self.node_id.fmt_short()))]
     pub(crate) fn get_send_addrs(
         &mut self,
-        have_ipv6: bool,
     ) -> (Option<SocketAddr>, Option<RelayUrl>, Vec<PingAction>) {
         let now = Instant::now();
         self.last_used.replace(now);
-        let (udp_addr, relay_url) = self.addr_for_send(&now, have_ipv6);
+        let (udp_addr, relay_url) = self.addr_for_send(&now);
         let mut ping_msgs = Vec::new();
 
         if self.want_call_me_maybe(&now) {
--- a/iroh-net/src/magicsock/node_map/node_state.rs
+++ b/iroh-net/src/magicsock/node_map/node_state.rs
@@ -257,7 +257,7 @@ impl NodeState {
 
     /// Returns the address(es) that should be used for sending the next packet.
     ///
-    /// Any or all of the UDP and relay addrs may be non-zero.
+    /// This may return to send on one, both or no paths.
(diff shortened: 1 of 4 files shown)


dignifiedquire:
In case no relay is available (eg offline network), will this still work? My worry is that if we can't use the relay, that we will error out in this case, even though we have potential available direct addresses

iroh-net/src/magicsock/node_map/node_state.rs
```diff
@@ -287,20 +282,8 @@ impl NodeState {
...
-                    .filter(|ipp| match ipp.ip() {
-                        IpAddr::V4(_) => true,
-                        IpAddr::V6(_) => have_ipv6,
-                    })
-                    .choose_stable(&mut rand::thread_rng())
-                    .map(|ipp| SocketAddr::from(*ipp));
-                trace!(udp_addr = ?addr, "best_addr is unset, use candidate addr and relay");
-                (addr, self.relay_url())
+                trace!("best_addr is unset, use relay");
+                (None, self.relay_url())
```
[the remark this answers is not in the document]
matheus23: Some more info, this is the code path that we end up in, if we return `(None, None)` here:
https://github.com/n0-computer/iroh/blob/04df203e8b7da7d8807a86c90c8d830785370e1c/iroh-net/src/magicsock.rs#L567-L582

flub:
> In case no relay is available (eg offline network), will this still work? My worry is that if we can't use the relay, that we will error out in this case, even though we have potential available direct addresses

In case there is no relay available this will do a ping to the direct address and use it as soon as there's a pong.  I tested this to make sure using iroh doctor and all looks good.

This delays using this scenario by 1rtt but for the simplicity we get in return that's a nice tradeoff to me.

iroh-net/src/magicsock/node_map/node_state.rs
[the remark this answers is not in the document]
dignifiedquire: relay_url can be `some` without the relay being reachable asfaik, but also we might not have a relay url, eg in a pure mdns setting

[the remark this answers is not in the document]
flub: Yes, some initial packets will be dropped.  A short while later we start having a destination that works and retries of the initial packets will succeed.  Having to wait one extra RTT before a packets gets through is fine for all packets that need to be acknowledged IIUC.

flub:
It should be pointed out that this is no observable behaviour change.  Before an arbitrary number of packets, but not all of them, **might** have made it to a reachable node.  So essentially you get a significant amount of packet loss.  Then the ping-pong would get through and then all packets make it.  Now you get a bit more predictability in that all packets will be lost until the ping-pong makes it through.

dignifiedquire:
if it works I am happy

matheus23 (approved):
This looks good from my perspective, so I'm approving, but would like the others to chime in, still, as I'm super new to these parts :v:

iroh-net/src/magicsock/node_map/node_state.rs
```diff
@@ -259,11 +258,7 @@ impl NodeState {
     /// Returns the address(es) that should be used for sending the next packet.
     ///
     /// Any or all of the UDP and relay addrs may be non-zero.
```
divagant-martian, line 260: unrelated I guess, but does this even mean? an address being "non zero"? more go baggage?

iroh-net/src/magicsock/node_map/node_state.rs
```diff
@@ -287,20 +282,8 @@ impl NodeState {
...
-                    .keys()
-                    .filter(|ipp| match ipp.ip() {
-                        IpAddr::V4(_) => true,
-                        IpAddr::V6(_) => have_ipv6,
-                    })
-                    .choose_stable(&mut rand::thread_rng())
-                    .map(|ipp| SocketAddr::from(*ipp));
-                trace!(udp_addr = ?addr, "best_addr is unset, use candidate addr and relay");
-                (addr, self.relay_url())
+                trace!("best_addr is unset, use relay");
```
divagant-martian, line 285: will we get here from somewhere above the node for which a relay only attempt will be made? Or should we have here `self.me` (or whatever the name of that field around here is)

divagant-martian (commented):
Just want to be sure I'm following: 

Since `assign_best_addr_from_candidates_if_empty` (damn that's long) picks only from addresses with recent pongs, and in case the best address is not set (so, we have no recent pongs, or no direct addresses at all) we go with relay, we will not use for app-level packages any direct address unless it's first pong-verified. Is this correct? If so then I guess this makes sense and the `(None, None`) case for best address and relay is actually recoverable.

iroh-net/src/magicsock/node_map/node_state.rs
flub: `2024-07-18T11:31:49.239813Z TRACE poll_send{me=xamcuaerdkwkwwjj}:get_send_addrs{node=dgjnkpaczxaekzxf}: iroh_net::magicsock::node_map::node_state: 285: best_addr is unset, use relay`

is how this shows up.  so i think this information is already suitably captured in the spans.

iroh-net/src/magicsock/node_map/node_state.rs
flub: I read this as "may be `Some`".  Probably the wording has survived from go.

flub: i re-worded this a bit.

PR #2510, opened 2024-07-16
feat(iroh-net)!: remove fs based peers storage

dignifiedquire:
## Description

Instead introduces

- `endpoint::Builder::known_nodes` to provide existing `NodeAddr`s
- you can use `endpoint::Endpoint::connection_infos` to retrieve the current list of known nodes that can be persisted

## Notes

- `prune_inactive` is not called anymore when storing, only on the heartbeat, and so there is a chance of storing uninteresting addresses
- the logic of filtering nodes has changed a little bit, to simplify the code paths

## Breaking Changes

- `endpoint::Builder::peers_path` is removed
- no automatic storage of known peers in `iroh-net`

## TODO

- [x] Recreate automatic storage and loading of peers inside of `iroh`


iroh-net/src/endpoint.rs
```diff
@@ -76,8 +75,8 @@ pub struct Builder {
     keylog: bool,
     discovery: Option<Box<dyn Discovery>>,
     proxy_url: Option<Url>,
-    /// Path for known peers. See [`Builder::peers_data_path`].
-    peers_path: Option<PathBuf>,
+    /// List of known nodes. See [`Builder::know_nodes`].
```
matheus23, line 78:
```suggestion
    /// List of known nodes. See [`Builder::known_nodes`].
```

matheus23 (commented):
In principle I like this.

The important part is getting back the saving functionality that used to be in iroh-net.
Ideally we bring back the same "save on tick" functionality where the peer map is saved.

Perhaps we can even save it in redb instead? That could be nicer for something that may be a little more corruption resistant.

iroh-base/src/node_addr.rs
```diff
@@ -46,6 +46,18 @@ pub struct NodeAddr {
     pub info: AddrInfo,
 }
 
+impl PartialOrd for NodeAddr {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        self.node_id.partial_cmp(&other.node_id)
+    }
+}
+
+impl Ord for NodeAddr {
```
rklaehn, line 55: I don't like this. This means that you can construct two NodeAddrs such that `a.cmp(b) == Ordering::Equal` yet `a != b`. I think this is forbidden or at least heavily discouraged. And if it is not - it should be...

iroh-net/src/endpoint.rs
```diff
@@ -771,6 +767,11 @@ impl Endpoint {
         self.msock.conn_type_stream(node_id)
     }
 
+    /// Get the known node addresses which should be persisted.
+    pub fn node_addresses_for_storage(&self) -> Vec<NodeAddr> {
+        self.msock.node_addresses_for_storage()
```
flub, line 772: This information can already be constructed from the `Endpoint::connection_infos` response.  Maybe it makes more sense to add a `From<ConnectionInfo> for NodeAddr` impl?

flub:
> The important part is getting back the saving functionality that used to be in iroh-net. Ideally we bring back the same "save on tick" functionality where the peer map is saved.
> 
> Perhaps we can even save it in redb instead? That could be nicer for something that may be a little more corruption resistant.

This sounds like `Endpoint::connection_infos` (or something else) should return a stream, yielding a new item every time the internal node map is updated.

iroh-base/src/node_addr.rs
dignifiedquire: fixed

dignifiedquire:
I implemented a timer based version of storage again, which I think works well enough for now

iroh-net/src/endpoint.rs
dignifiedquire: good idea, done

iroh/src/node.rs
```diff
@@ -299,6 +305,41 @@ impl<D: iroh_blobs::store::Store> NodeInner<D> {
...
+                        }
+                        _ = save_timer.tick() => {
+                            trace!("save known node addresses tick");
+                            let addrs = node_addresses_for_storage(&ep);
+                            if let Err(err) = store_node_addrs(&nodes_data_path, &addrs).await {
+                                warn!("failed to store knonw node addresses: {:?}", err);
+                            }
+                        }
+                    }
+                }
```
flub, line 337: Please always attach a tracing span when spawning a new tokio task.  Otherwise all the span information is gone.  If a new span is not suitable than `.instrument(Span::current())` propagates the current span, but here a new span makes sense: `.instrument(info_span!("known-addrs"))` or so.

flub:
lgtm i think

iroh-net/src/endpoint.rs
divagant-martian: I'm not sure this is the same. The previous code saved only the verified paths of used nodes and all paths of unverified nodes. Now, as far as i can tell, we will store again paths not active/used for nodes with known verified paths

dignifiedquire: @divagant-martian should be fixed, the filtering is happening now at the node layer

iroh-net/src/magicsock/node_map/node_state.rs
```diff
@@ -1478,6 +1465,8 @@ pub struct DirectAddrInfo {
     pub last_control: Option<(Duration, ControlMsg)>,
     /// How long ago was the last payload message for this node.
     pub last_payload: Option<Duration>,
+    /// When was this connection last alive, if ever.
+    pub last_alive: Option<Duration>,
```
divagant-martian, line 1469: This might be redundant, tldr this is the max between the two above. But it's harmless anyway

divagant-martian (approved):
:+1: lgtm

matheus23 (approved):

PR #2515, opened 2024-07-16
refactor(iroh-net): More renaming of endpoint to direct address

flub:
## Description

The name "endpoint" is overloaded and we decided a while ago to not
use it for things other than `iroh_net::Endpoint`.  This further
renames more places that were called endpoint with the new name we
already introduced for them: direct address.  The direct address
naming is already used consistently on public APIs, this is only for
internal consistency.

## Breaking Changes

metric changed: `magicsock::metrics::Metrics::update_endpoints` -> `Metrics::update_direct_addrs`.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/src/magicsock.rs
+++ b/iroh-net/src/magicsock.rs
@@ -229,15 +229,15 @@ pub(crate) struct MagicSock {
     /// Optional discovery service
     discovery: Option<Box<dyn Discovery>>,
 
-    /// Our discovered endpoints
-    endpoints: Watchable<DiscoveredEndpoints>,
+    /// Our discovered direct addresses
+    direct_addrs: Watchable<DiscoveredDirectAddrs>,
 
     /// List of CallMeMaybe disco messages that should be sent out after the next endpoint update
     /// completes
     pending_call_me_maybes: parking_lot::Mutex<HashMap<PublicKey, RelayUrl>>,
 
-    /// Indicates the update endpoint state.
-    endpoints_update_state: EndpointUpdateState,
+    /// Indicates the direct addr update state.
+    direct_addr_update_state: DirectAddrUpdateState,
 
     /// Skip verification of SSL certificates from relay servers
     ///
@@ -321,8 +321,8 @@ impl MagicSock {
     /// received.
     pub(crate) fn direct_addresses(&self) -> DirectAddrsStream {
         DirectAddrsStream {
-            initial: Some(self.endpoints.get()),
-            inner: self.endpoints.watch().into_stream(),
+            initial: Some(self.direct_addrs.get()),
+            inner: self.direct_addrs.watch().into_stream(),
         }
     }
 
@@ -370,7 +370,7 @@ impl MagicSock {
     /// Add addresses for a node to the magic socket's addresbook.
     #[instrument(skip_all, fields(me = %self.me))]
     pub fn add_node_addr(&self, mut addr: NodeAddr, source: node_map::Source) -> Result<()> {
-        let my_addresses = self.endpoints.get().last_endpoints;
+        let my_addresses = self.direct_addrs.get().addrs;
         let mut pruned = 0;
         for my_addr in my_addresses.into_iter().map(|ep| ep.addr) {
             if addr.info.direct_addresses.remove(&my_addr) {
@@ -394,9 +394,12 @@ impl MagicSock {
     ///
     /// On a successful update, our address is published to discovery.
     pub(super) fn update_direct_addresses(&self, eps: Vec<DirectAddr>) {
-        let updated = self.endpoints.update(DiscoveredEndpoints::new(eps)).is_ok();
+        let updated = self
+            .direct_addrs
+            .update(DiscoveredDirectAddrs::new(eps))
+            .is_ok();
         if updated {
-            let eps = self.endpoints.read();
+            let eps = self.direct_addrs.read();
             eps.log_endpoint_change();
             self.node_map
                 .on_direct_addr_discovered(eps.iter().map(|ep| ep.addr));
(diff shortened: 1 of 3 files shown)


iroh-net/src/magicsock.rs
```diff
@@ -394,9 +394,12 @@ impl MagicSock {
...
     pub(super) fn update_direct_addresses(&self, eps: Vec<DirectAddr>) {
-        let updated = self.endpoints.update(DiscoveredEndpoints::new(eps)).is_ok();
+        let updated = self
+            .direct_addrs
+            .update(DiscoveredDirectAddrs::new(eps))
+            .is_ok();
         if updated {
-            let eps = self.endpoints.read();
+            let eps = self.direct_addrs.read();
             eps.log_endpoint_change();
```
matheus23, line 403:
```suggestion
            let eps = self.direct_addrs.read();
            eps.log_direct_addrs_change();
```

iroh-net/src/magicsock.rs
```diff
@@ -1236,15 +1239,15 @@ impl MagicSock {
...
+        self.direct_addr_update_state.schedule_run(why);
     }
 
     /// Publishes our address to a discovery service, if configured.
     ///
     /// Called whenever our addresses or home relay node changes.
     fn publish_my_addr(&self) {
         if let Some(ref discovery) = self.discovery {
-            let eps = self.endpoints.read();
+            let eps = self.direct_addrs.read();
```
matheus23, line 1250:
```suggestion
            let direct_addrs = self.direct_addrs.read();
```

iroh-net/src/magicsock.rs
```diff
@@ -1236,15 +1239,15 @@ impl MagicSock {
...
             let relay_url = self.my_relay();
             let direct_addresses = eps.iter().map(|ep| ep.addr).collect();
             let info = AddrInfo {
```
matheus23, line 1253:
```suggestion
            let direct_addrs = self.direct_addrs.read();
            let relay_url = self.my_relay();
            let addrs = direct_addrs.iter().map(|da| da.addr).collect();
            let info = AddrInfo {
```

iroh-net/src/magicsock.rs
```diff
@@ -2000,39 +2006,39 @@ impl Actor {
...
-                    $already.insert($ipp, $et);
-                    $eps.push(DirectAddr {
+                    $already.insert($ipp, $typ);
+                    $addrs.push(DirectAddr {
                         addr: $ipp,
-                        typ: $et,
+                        typ: $typ,
                     });
                 }
             };
```
matheus23, line 2044: Not relevant to this PR, but why is this a macro, not a function?

matheus23 (approved):
Only small nits :v:

iroh-net/src/magicsock.rs
dignifiedquire: why not :P

matheus23: I see you half-joking, but I'll still explain myself: It's mostly about complexity. I'd argue macros are more complex and that leads to a bunch of inefficiencies. Tooling gets confused/disabled/slow more often, people are more likely to shy away from that code, etc. Same arguments as in any other case with more complexity, e.g. when there's too many abstraction layers.

dignifiedquire: I never joke ;) 

The answer is when I wrote this originally, it was easier to use a macro, I don't remember why though

iroh-net/src/magicsock.rs
```diff
@@ -229,15 +229,15 @@ pub(crate) struct MagicSock {
     /// Optional discovery service
     discovery: Option<Box<dyn Discovery>>,
 
-    /// Our discovered endpoints
-    endpoints: Watchable<DiscoveredEndpoints>,
+    /// Our discovered direct addresses
```
divagant-martian, line 232:
```suggestion
    /// Our discovered direct addresses.
```

iroh-net/src/magicsock.rs
divagant-martian: I would generally avoid non obvious abbreviations. `direct_addrs` is fine by me

iroh-net/src/magicsock.rs
```diff
@@ -1295,25 +1298,24 @@ impl DiscoMessageSource {
...
 /// - if an update is scheduled while another one is running, remember that
 ///   and start a new one when the current one has finished
 #[derive(Debug)]
-struct EndpointUpdateState {
+struct DirectAddrUpdateState {
     /// If running, set to the reason for the currently the update.
     running: sync::watch::Sender<Option<&'static str>>,
-    /// If set, this means we will start a new endpoint update state as soon as the current one
-    /// is finished.
+    /// If set start a new update as soon as the current one is finished.
```
divagant-martian, line 1311:
```suggestion
    /// If set, start a new update as soon as the current one is finished.
```

iroh-net/src/magicsock.rs
```diff
@@ -2461,59 +2467,63 @@ fn bind(port: u16) -> Result<(UdpConn, Option<UdpConn>)> {
     Ok((pconn4, pconn6))
 }
 
+/// The discovered direct addresses of this [`MagicSock`].
+///
+/// These are all the [`DirectAddr`]s that this [`MagicSock`] is aware of exist for itself.
```
divagant-martian, line 2472:
```suggestion
/// These are all the [`DirectAddr`]s that this [`MagicSock`] is aware of for itself.
```

another alternative: `These are all the addresses that this endpoint knows exist for itself.`

not sure if a typo when changing the idea mid-write, which happens to all, or intended. However, as is, this is grammatically not correct

divagant-martian (approved):
minor things. LGTM and I find it easier to read

iroh-net/src/magicsock.rs
flub: yeah, I share @matheus23's opinion here.  I tried to re-write this once as a function and it was impossible due to borrowing rules.  It probably is still possible with a bit more refactoring though, I just recall abandoning it for some reason.

PR #2517, opened 2024-07-16
fix(iroh-blobs): properly handle Drop in local pool during shutdown

rklaehn:
## Description

The tokio_util LocalPoolHandle does not properly handle Drop during shutdown. Its threads are just spawned as detached. So any drop impl that runs in a local pool thread will be stopped as soon as the process terminates. This can have some bad consequences if that drop operation performs IO, like closing files and committing database transactions.

Here is where the threads get spawned. The `std::thread::JoinHandle`s are just dropped.
https://docs.rs/tokio-util/latest/src/tokio_util/task/spawn_pinned.rs.html#381

Here is some discussion of the observed effects:
https://discord.com/channels/949724860232392765/1260571544414064670

LocalPoolHandle also, of course, is using an unbounded channel:
https://docs.rs/tokio-util/latest/src/tokio_util/task/spawn_pinned.rs.html#372

## Breaking Changes

Public interfaces using tokio_util::task::LocalPoolHandle will now use our own LocalPool/LocalPoolHandle.

## Notes & open questions

Should we use an unbounded channel like tokio::spawn or LocalPoolHandle::spawn_pinned? Seems like a big footgun. But if not, we need to somehow handle when the queue is full.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh/src/node/builder.rs
```diff
@@ -454,7 +455,10 @@ where
 
     async fn build_inner(self) -> Result<ProtocolBuilder<D>> {
         trace!("building node");
-        let lp = LocalPoolHandle::new(num_cpus::get());
+        let lp = LocalPool::new(local_pool::Config {
+            panic_mode: PanicMode::LogAndContinue,
```
rklaehn, line 459: Womp womp, this is because of https://github.com/tokio-rs/tracing/issues/2870, mostly

iroh-blobs/src/util/local_pool.rs
```diff
@@ -0,0 +1,698 @@
...
+}
+
+impl<T> Future for Run<T> {
+    type Output = std::result::Result<T, SpawnError>;
+
+    fn poll(
+        mut self: Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Self::Output> {
+        self.0.poll(cx).map_err(|_| SpawnError::Shutdown)
```
dignifiedquire, line 367: should we log this error?

rklaehn: It's not really an error. It is just what happens if you await a future that runs on a pool that has been shut down.

iroh-blobs/src/util/local_pool.rs
```diff
@@ -0,0 +1,698 @@
...
+        self.try_spawn(item)?;
+        Ok(Run(recv_res))
+    }
+
+    /// Run a task in the pool.
+    ///
+    /// The task will be run detached. This can be useful if
+    /// you are not interested in the result or in in cancellation or
+    /// you provide your own result handling and cancellation mechanism.
+    pub fn try_spawn<F, Fut>(&self, gen: F) -> SpawnResult<()>
```
dignifiedquire, line 448: maybe `try_spawn_detached`?

iroh/src/node.rs
```diff
@@ -377,6 +375,9 @@ impl<D: iroh_blobs::store::Store> NodeInner<D> {
 
         // Abort remaining tasks.
         join_set.shutdown().await;
+
+        // Abort remaining local tasks.
+        // local_pool.shutdown().await;
```
dignifiedquire, line 380: why is this commented out?

iroh-blobs/src/util/local_pool.rs
rklaehn: doesn't spawn imply detached? At least that is how it is usually used in tokio. I was going for "run": you get back a handle, and dropping it will cancel the task, like our AbortableJoinHandle, and "spawn" for: you don't care about the result at this location.

iroh/src/node.rs
rklaehn: It is an experiment if this will fix the CI issues. Next will be leak 🙄.

This will not remain in the PR when we merge it...

iroh-blobs/src/util/local_pool.rs
rklaehn: I originally had this just return a oneshot receiver, but having my own wrapper allows implementing Into<io::Error> for it, which has some advantages (JoinError does the same).

dignifiedquire: maybe add a comment, so it is clear for future readers

iroh-blobs/src/util/local_pool.rs
rklaehn: Ok, renamed it

(try_)run -> (try_)spawn
(try_)spawn -> (try_)spawn_detached

dignifiedquire (approved):

PR #2525, opened 2024-07-22
refactor(iroh): Single runtime

rklaehn:
## Description

Make sure the local pool threads use the main tokio runtime instead of a current_thread runtime per local pool thread.

This is mostly relevant for call to spawn_blocking and spawn from inside local tasks. Before they would go to the single threaded runtime of that thread, now they go to the blocking pool of the main runtime.

This means that the local futures are more tightly integrated with the main runtime. Everything you can do from a spawned future, you should be able to also do from a local future.

## Breaking Changes

Surprisingly, none

## Notes & open questions

Still not sure if this is OK to do, but given that even tokio::runtime::Handle has a block_on fn, it seems intended usage.

## Change checklist

- [x] Self-review.
- [x] ~~Tests if relevant.~~


dignifiedquire (approved):
if it works, ship it

PR #2527, opened 2024-07-22
refactor(iroh)!: directly use gossip net dispatcher

### LINKED ISSUE #2460: Unify gossip dispatchers / flaky test: `gossip_smoke`
We got 2 dispatchers in the public iroh gossip client API. The one that directly drives the gossip client, and then another one inside iroh-gossip.

We should unify the two somehow so we don't have gossip messages going through 2 queues on the way to the consumer.

> So the remaining TODO is to merge the two dispatchers (in dispatcher.rs and net.rs) so that messages flow only over a single channel and not two. We should do this, but IMO it can also happen in a followup.

_Originally posted by @Frando in https://github.com/n0-computer/iroh/issues/2258#issuecomment-2210299055_

dignifiedquire:
<TODO>

Closes #2460


iroh/src/node/rpc.rs
```diff
@@ -209,22 +210,47 @@ impl<D: BaoStore> Handler<D> {
...
+            Broadcast(msg) => chan.rpc(msg, self, Self::gossip_broadcast).await,
+            BroadcastNeighbours(msg) => {
+                chan.rpc(msg, self, Self::gossip_neighbours_broadcast).await
+            }
+            Quit(msg) => chan.rpc(msg, self, Self::gossip_quit).await,
         }
     }
 
+    async fn gossip_broadcast(self, msg: gossip::BroadcastRequest) -> RpcResult<()> {
+        self.inner.gossip.broadcast(msg.topic, msg.message).await?;
```
rklaehn, line 234: Does this even work if I am not subscribed to the topic?

The usage of the client API allows calling broadcast without subscribing to a topic, as far as I can see.

rklaehn (commented):
I am not sure what happens if you use broadcast or broadcast_neighbours for a topic you are not subscribed to.

Also, why is there subscribe and quit as separate calls? And what happens if you quit a topic that is subscribed in another part of the code base?

PR #2532, opened 2024-07-23
fix(iroh-cli)!: Improve cli and configuration file

flub:
## Description

The configuration file behaviour is improved:

- Unknown fields will now cause an error rather than be silently ignored.

- Make it possible to express GcPolicy in the TOML config file.  This
  was not possible before.  It was possible to disable it before, but
  it was disabled by default so that was rather moot.

The --help output is improved:

- Update header of iroh.
- Directly point to config file docs.
- Use consistent style.

## Breaking Changes

- Unknown fields in the configuration file will now cause an error.

- Configuring the GC Policy in the configuration file has changed.

## Change checklist

- [ ] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2641,6 +2641,7 @@ dependencies = [
  "tracing",
  "tracing-appender",
  "tracing-subscriber",
+ "url",
  "walkdir",
 ]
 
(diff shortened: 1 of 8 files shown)


matheus23:
I'm confused by the PR title. Is it missing a `feat` or `fix` prefix?

flub:
> I'm confused by the PR title. Is it missing a `feat` or `fix` prefix?

lol, i forgot how to write PR titles

dignifiedquire (approved):

PR #2533, opened 2024-07-23
fix(iroh-blobs): use async_channel instead of flume for local_pool

rklaehn:
## Description

During soft shutdown of the local pool, a Finish message is sent to all threads. On main, this occasionally hangs. Further investigation showed that this is a message that is being sent but not received despite being in the channel. Adding a simple timeout to the select! so the flume recv call is executed again fixes it.

So it seems that there is a bug in flume that occasionally leads to notifications being dropped.

This PR just does a 1:1 replacement of flume with async_channel.

Before this change, I can get the test_shutdown test to fail easily by running it 1000 times:

```
for i in $(seq 1 1000); do cargo test --release -p iroh-blobs test_shutdown -- --nocapture >> log.txt; done
```

Result:
```
...
sending shutdown message
sending shutdown message
sending shutdown message
sending shutdown message
test util::local_pool::tests::test_shutdown has been running for over 60 seconds
```

After this change, I can not get test_finish (renamed because it tests finish, not shutdown) to fail at all even after several 1000 tests.

## Breaking Changes

None

## Notes & open questions

Can somebody talk me out of this? I would prefer to keep flume, but the evidence above seems conclusive...

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-blobs/src/util/local_pool.rs
+++ b/iroh-blobs/src/util/local_pool.rs
@@ -247,7 +247,7 @@ impl LocalPool {
 
     /// Immediately stop polling all tasks and wait for all threads to finish.
     ///
-    /// This is like droo, but waits for thread completion asynchronously.
+    /// This is like drop, but waits for thread completion asynchronously.
     ///
     /// If there was a panic on any of the threads, it will be re-thrown here.
     pub async fn shutdown(self) {
@@ -270,7 +270,6 @@ impl LocalPool {
         // we assume that there are exactly as many threads as there are handles.
         // also, we assume that the threads are still running.
         for _ in 0..self.threads_u32() {
-            println!("sending shutdown message");
             // send the shutdown message
             // sending will fail if all threads are already finished, but
             // in that case we don't need to do anything.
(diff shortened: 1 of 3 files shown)


dignifiedquire (approved):
if it helps, lets do it

iroh-blobs/src/util/local_pool.rs
```diff
@@ -270,14 +270,13 @@ impl LocalPool {
...
         for _ in 0..self.threads_u32() {
-            println!("sending shutdown message");
             // send the shutdown message
             // sending will fail if all threads are already finished, but
             // in that case we don't need to do anything.
             //
             // Threads will add a permit in any case, so await_thread_completion
             // will then immediately return.
-            self.send.send(Message::Finish).ok();
+            self.send.send_blocking(Message::Finish).ok();
```
dignifiedquire, line 279: why is this not using an `async` send?

divagant-martian:
since it's shutdown, could this be related to flume's lack of cancel-safety?

rklaehn:
> since it's shutdown, could this be related to flume's lack of cancel-safety?

Could be. But what I am seeing is not a message getting lost during recv, but just a message being "stuck". Adding a tokio::time::sleep(...) to the select! statement for testing "fixes" it by waking up the future, then the message gets delivered.

iroh-blobs/src/util/local_pool.rs
rklaehn: I guess that was back when this was used in Drop... It should be async, really...

PR #2536, opened 2024-07-23
fix(iroh): get rid of flume channels

rklaehn:
## Description

Get rid of flume channels

**I am doing this as multiple PRs**

Flume channels have a bug?(*) regarding notifications when the futures returned by recv_async are cancelled. Reproducer here: https://github.com/rklaehn/flume-bug-reproducer . Note that this does not lead to loss of messages, but loss of *notifications*.

(it is unclear what guarantees regarding cancel safety flume gives at all... See https://github.com/zesterer/flume/issues/104 )

If you have another future like a regular timer that resolves frequently, you will never see the issue. But if you have something like an actor where the flume channel is the only source of regular events, you might be "stuck" until somebody sends another message.

So in every single location where we use select! with one of the cases being recv_async, there is a small potential to lose notifications.

Given that we use select! a lot, I think we can't afford to carefully study every single case. Replacing the select! with code that merges streams also does not seem to be a viable option.

So the only remaining option is to change to an async mpmc channel that is experimentally shown to be safe against cancellation. The only one I found is async_channel. When using async_channel in the reproducer above, the bug immediately goes away.

## Breaking Changes

None

## Notes & open questions

Other options?
Benchmarks?
Remove the unsafe same_channel workaround or keep it for now?

## Change checklist

- [ ] Self-review.
- [ ] ~~Tests if relevant.~~


iroh-docs/src/actor.rs
```diff
@@ -581,7 +581,10 @@ impl Drop for SyncHandle {
     fn drop(&mut self) {
         // this means we're dropping the last reference
         if let Some(handle) = Arc::get_mut(&mut self.join_handle) {
-            self.tx.send(Action::Shutdown { reply: None }).ok();
+            // this call is the reason tx can not be a tokio mpsc channel.
+            // we have no control about where drop is called, yet tokio send_blocking panics
+            // when called from inside a tokio runtime.
+            self.tx.send_blocking(Action::Shutdown { reply: None }).ok();
```
dignifiedquire, line 587: couldn't we detect the current runtime, and if one exists use `block_on` + `send` and otherwise use `send_blocking`?

rklaehn: I don't think so. block_on panics if you call it from inside a runtime... You could use spawn(send), but really I just don't want to do anything runtime related during a Drop...

```
Panics
This function panics if the provided future panics, if called within an asynchronous execution context, or if a timer future is executed on a runtime that has been shut down.
```

dignifiedquire:
Now all usage of flume in the repo are gone. The only missing thing is a blocking `recv_timeout`

rklaehn:
We still have the dependency via quic-rpc, but other than that we are flume-free. So I am going to close this.

PR #2539, opened 2024-07-24
refactor(iroh-net): replace flume in iroh-net with async_channel

rklaehn:
## Description

refactor(iroh-net): replace flume in iroh-net with async_channel

Rationale: see https://github.com/n0-computer/iroh/pull/2536

This is the first in a series of PRs that will replace flume with async_channel. In this case it is very close to a drop in replacement without any need for changes.

Only noteable changes:

send_async becomes send
send becomes send_blocking
Receiver implements Stream, so no need for .into_stream()
Receiver is not Unpin, so to make it unpin you need to do `Box::pin(recv)` or use `.boxed()`

## Breaking Changes

None

## Notes & open questions

- Anything specific from flume we were relying on here?

## Change checklist

- [x] Self-review.
- [x] ~~Tests if relevant.~~

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2836,6 +2836,7 @@ name = "iroh-net"
 version = "0.21.0"
 dependencies = [
  "anyhow",
+ "async-channel",
  "axum",
  "backoff",
  "base64 0.22.1",
@@ -2846,7 +2847,6 @@ dependencies = [
  "der",
  "derive_more",
  "duct",
- "flume",
  "futures-buffered",
  "futures-concurrency",
  "futures-lite 2.3.0",
--- a/iroh-net/src/discovery/local_swarm_discovery.rs
+++ b/iroh-net/src/discovery/local_swarm_discovery.rs
@@ -14,7 +14,7 @@ use derive_more::FromStr;
 use futures_lite::{stream::Boxed as BoxStream, StreamExt};
 use tracing::{debug, error, trace, warn};
 
-use flume::Sender;
+use async_channel::Sender;
 use iroh_base::key::PublicKey;
 use swarm_discovery::{Discoverer, DropGuard, IpClass, Peer};
 use tokio::task::JoinSet;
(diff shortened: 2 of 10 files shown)


Cargo.lock
```diff
@@ -2846,7 +2847,6 @@ dependencies = [
  "der",
  "derive_more",
  "duct",
- "flume",
```
flub, line 2849: notice that this doesn't bad flume from our codebase entirely...

iroh-net/src/discovery/local_swarm_discovery.rs
```diff
@@ -210,7 +210,7 @@ impl LocalSwarmDiscovery {
             );
 
             sender
-                .send(Message::Discovery(node_id.to_string(), peer.clone()))
+                .send_blocking(Message::Discovery(node_id.to_string(), peer.clone()))
```
flub, line 213: Reading the docs this is pretty sad.  They say you should not call this from async contexts, but there is nothing we can do to make sure this function is not called from async contexts.  In fact this **is** called from an async context because it spawns a new task.

Maybe you can convert this function into async without too many repercussions?

iroh-net/src/discovery/local_swarm_discovery.rs
```diff
@@ -267,23 +267,23 @@ impl From<&Peer> for DiscoveryItem {
...
-        let (send, recv) = flume::bounded(20);
+        let (send, recv) = async_channel::bounded(20);
         let discovery_sender = self.sender.clone();
         tokio::spawn(async move {
             discovery_sender
-                .send_async(Message::SendAddrs(node_id, send))
+                .send(Message::SendAddrs(node_id, send))
                 .await
                 .ok();
         });
```
flub, line 277: does... this even make sense?  our actors are so messy!  but nothing this PR introduced...

flub (approved):
the problem with a pr like this is that you see all the crimes that are being committed with channels.  but the change is basically equivalent so fine, just close my eyes.

Cargo.lock
rklaehn: I am very much aware. There will be more PRs... I decided to do this one by one since some of them are more complex than others.

The biggest will be quic-rpc...

iroh-net/src/discovery/local_swarm_discovery.rs
rklaehn: I would prefer to keep this PR solely a 1:1 flume -> async_channel replacement. Should I make an issue?

iroh-net/src/discovery/local_swarm_discovery.rs
rklaehn: 🤷‍♂️ I am just the search and replace machine at this point. I don't understand any of this...

matheus23: Yeah I mentioned the spawn thing in code review as well: https://github.com/n0-computer/iroh/pull/2376#discussion_r1668183933

We agreed to handle this in a followup: https://discord.com/channels/949724860232392765/955909916487463032/1259908995376877609

PR #2540, opened 2024-07-24
refactor(iroh-docs): Replace flume with async_channel in docs

rklaehn:
## Description

This is mostly a 1:1 replacement, except for the fact that the same_channel api is missing from async_channel. So I replaced it with some ugly code that uses the fact that a async_channel Sender or Receiver is just an Arc<Channel>.

To be removed if/when https://github.com/smol-rs/async-channel/pull/98 is merged, but until then I think it is fine.

## Breaking Changes

None

## Notes & open questions

Note: we can not use tokio::sync::mpsc::Channel for the actor because we can't control from which thread Drop is called.
Note: some streams were Unpin before, but it was not explicit. Now I added Unpin explicitly (and boxed the stream to make it true). Not sure if the version check would catch this, pretty sure that not. But taking away Unpin would have been a breaking change.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


iroh/src/node/rpc/docs.rs
```diff
@@ -111,18 +110,21 @@ impl DocsEngine {
         Ok(DropResponse {})
     }
 
-    pub fn doc_list(&self, _req: DocListRequest) -> impl Stream<Item = RpcResult<DocListResponse>> {
-        let (tx, rx) = flume::bounded(ITER_CHANNEL_CAP);
+    pub fn doc_list(
+        &self,
+        _req: DocListRequest,
+    ) -> impl Stream<Item = RpcResult<DocListResponse>> + Unpin {
```
rklaehn, line 116: Added Unpin here. it was unpin before - implicitly assumed, but it is better to be explicit. Maybe we should remove this later, but for now let's keep it as close as possible to before. Otherwise I think this would be a breaking change.

iroh-docs/src/sync.rs
```diff
@@ -108,17 +108,34 @@ pub struct SyncOutcome {
     pub num_sent: usize,
 }
 
+fn get_as_ptr<T>(value: &T) -> Option<usize> {
+    use std::mem;
+    if mem::size_of::<T>() == std::mem::size_of::<usize>()
+        && mem::align_of::<T>() == mem::align_of::<usize>()
+    {
+        // Safe only if size and alignment requirements are met
+        unsafe { Some(mem::transmute_copy(value)) }
```
Frando, line 117: Is this really the simplest way to check for pointer equality?

I'd much prefer https://github.com/smol-rs/async-channel/pull/98 to get merged soon.

Frando (commented):
LGTM. Not sure about the same_channel unsafe, would have to read rust mem docs to validate first. But if you did this, fine to approve.

iroh-docs/src/sync.rs
rklaehn: Yeah, me too. But what can you do...

This is fine as long as an async_channel remains an Arc<whatever>...

Frando (approved):

PR #2545, opened 2024-07-25
feat(iroh)!: Blob batch PR, attempt 3

rklaehn:
## Description

This is the third attempt to add a batch API for adding blobs. Previous one was https://github.com/n0-computer/iroh/pull/2339

The basic idea is the following: all changes to the store happen in the context of a _batch_. All write operations within a batch produce temp tags. These temp tags are scoped to the batch and keep the data alive as long as the batch exists.

At some point, the API user has to upgrade one or more temp tags to permanent tags.

All non-batch operations would long term be implemented in terms of batch operations.

In a second step, the following rpc calls would be replaced by their batch equivalent.
- AddStream
- AddPath
- CreateCollection

The third one is very nice, since it means that the notion of a collection (as in a special kind of hashseq) no longer has to even exist in the node code.

## Breaking Changes

- iroh::client::blobs::BlobStatus has a new case NotFound
- iroh::client::blobs::BlobStatus::Partial: size is now a BaoBlobSize instead of a u64

All other public changes are adding of new APIs.

## Notes & open questions

Note: in the previous version I had an optimisation to avoid storing TempTags in the case where there are multiple TempTags with the same hash. I removed this to keep things simple. We can add it back later.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


matheus23:
> I would be curious what you think of the API.

I admit, after looking at it a little more, I'm somewhat surprised by the `TempTag` API combined with the `Batch`. I guess I had a more limited API that's more tailored to creating `Collection`s or `HashSeq`es in mind, something like this:
```rust
pub struct Batch;

pub async fn iroh::client::blobs::Client::batch() -> Result<Batch>;

pub async fn iroh::client::blobs::Batch::push_bytes(&self, name: String, bytes: impl Into<Bytes>) -> Result<()>;
pub async fn iroh::client::blobs::Batch::push_file(&self, name: String, path: PathBuf) -> Result<u64>
pub async fn iroh::client::blobs::Batch::push_dir(&self, name: String, root: PathBuf) -> Result<()>
// [...]
pub async fn iroh::client::blobs::push_dir_with_opts(
    &self,
    root: PathBuf,
    opts: AddDirOpts,
) -> Result<()>;
// [...]
pub async fn iroh::client::blobs::Batch::commit(&self) -> Result<Tag>; // maybe even taking an owned `self`?
pub async fn iroh::client::blobs::Batch::commit_to(&self, tag: &Tag) -> Result<()>;
```

So this would `commit` the batch to be a `Collection`, and you wouldn't surface the concept of `TempTag`s to the user at all.

But it's also more limited, so a classic complexity vs. expressiveness issue.
Thoughts?

Perhaps it makes sense to investigate some use-cases to see if the added expressiveness is worth it. "Importing the linux kernel" is definitely one.

rklaehn:
> > I would be curious what you think of the API.
> 
> I admit, after looking at it a little more, I'm somewhat surprised by the `TempTag` API combined with the `Batch`. I guess I had a more limited API that's more tailored to creating `Collection`s or `HashSeq`es in mind, something like this:
> 
> ```rust
> impl iroh::client::blobs::Client {
>   pub async fn batch() -> Result<Batch>;
> }
> 
> impl iroh::client::blobs::Batch {
>   pub async fn push_bytes(&self, name: String, bytes: impl Into<Bytes>) -> Result<()>;
>   pub async fn push_file(&self, name: String, path: PathBuf) -> Result<u64>;
>   pub async fn push_dir(&self, name: String, root: PathBuf) -> Result<()>;
>   // [...]
>   pub async fn push_dir_with_opts(
>     &self,
>     root: PathBuf,
>     opts: AddDirOpts,
>   ) -> Result<()>;
>   // [...]
>   pub async fn commit(&self) -> Result<Tag>; // maybe even taking an owned `self`?
>   pub async fn commit_to(&self, tag: &Tag) -> Result<()>;
> }
> ```
> 
> So this would `commit` the batch to be a `Collection`, and you wouldn't surface the concept of `TempTag`s to the user at all.
> 
> But it's also more limited, so a classic complexity vs. expressiveness issue. Thoughts?
> 
> Perhaps it makes sense to investigate some use-cases to see if the added expressiveness is worth it. "Importing the linux kernel" is definitely one.

(I would prefer the concept of a collection to be just a tiny sliver in the client)

The above proposal assumes that you always want to create exactly 1 collection or HashSeq, but e.g. it is possible that you want to create multiple individually tagged blobs or hashseqs in a batch.

The long term idea is that even interactions with the network that modify the local store will happen in the context of a batch, so you will be able to use temp tags.

Imagine traversing a dag. You get a blob, parse it for links, get another blob, parse it for links, finally find what you have been looking for, then assign a permanent tag. This would be easily expressible using the API using temp tags.

matheus23:
Okay. I'm convinced the additional expressiveness is worth it.
One thing I'd really like to change is the naming of `Batch::upgrade`, though. Can we switch it to something like `make_permanent` or `commit`? Or something else if there are other suggestions.
Upgrade only makes sense once I've read the docs for `Tag` and `TempTag`, looked at the `Batch::upgrade` signature and compared the two items in my head.
And if I'm looking for ways to make my temporary changes permanent, I end up skipping a function name like `upgrade` in my mind.

rklaehn:
> Okay. I'm convinced the additional expressiveness is worth it.

Glad I convinced you.

> One thing I'd really like to change is the naming of Batch::upgrade, though. Can we switch it to something like make_permanent or commit? Or something else if there are other suggestions.

Agree that upgrade is not great. Maybe persist? I mean, a temp tag is purely in memory while a named tag is persistent in the database.

matheus23:
Yes to `persist` in favor of `upgrade`

rklaehn:
Ok, renamed. So here is the batch API: https://n0-computer.github.io/iroh/pr/2545/docs/iroh/client/blobs/struct.Batch.html

iroh/src/client/blobs/batch.rs
```diff
@@ -0,0 +1,460 @@
...
+        input: impl Stream<Item = io::Result<Bytes>> + Send + Unpin + 'static,
+    ) -> Result<TempTag> {
+        self.add_stream_with_opts(input, Default::default()).await
+    }
+
+    /// Create a temp tag to protect some content (blob or hashseq) from being deleted.
+    ///
+    /// A typical use case is that you are downloading some data and want to protect it
+    /// from deletion while the download is ongoing, but don't want to protect it permanently
+    /// until the download is completed.
```
matheus23, line 179:
```suggestion
    /// Creates a temp tag to protect some content (blob or hashseq) from being deleted.
    ///
    /// This is a lower-level API. The other functions in [`Batch`] already create [`TempTag`]s automatically.
    ///
    /// [`TempTag`]s allow you to protect some data from deletion while a download is ongoing,
    /// even if you don't want to protect it permanently.
```

Not super sure what I'd need the `temp_tag` function for? Tried my best to improve the docs so that people don't expect to *need* to call this function to persist their stuff.

iroh/src/client/blobs/batch.rs
```diff
@@ -0,0 +1,460 @@
...
+                content,
+            })
+            .await??;
+        // Only after success of the above call, we can create the corresponding local temp tag
+        Ok(self.local_temp_tag(content, None))
+    }
+
+    /// Write a blob by passing an async reader.
+    ///
+    /// This produces a stream from the reader with a hardcoded buffer size of 64KB.
```
matheus23, line 195:
```suggestion
    /// This consumes the stream in chunks using `opts.chunk_size`. A good default is 64KB.
```

iroh/src/client/blobs/batch.rs
```diff
@@ -0,0 +1,460 @@
...
+    }
+
+    /// Add a directory as a hashseq in iroh collection format
+    pub async fn add_dir(&self, root: PathBuf) -> Result<TempTag> {
+        self.add_dir_with_opts(root, Default::default()).await
+    }
+
+    /// Write a blob by passing an async reader.
+    ///
+    /// This will use a default chunk size of 64KB, and a format of [BlobFormat::Raw].
```
matheus23, line 159:
```suggestion
    /// This will consume the stream in 64KB chunks, and use a format of [BlobFormat::Raw].
    ///
    /// For more options, see [`add_dir_with_opts`].
```

iroh/src/client/blobs/batch.rs
```diff
@@ -0,0 +1,460 @@
...
+    /// Import a blob from a filesystem path.
+    ///
+    /// `path` should be an absolute path valid for the file system on which
+    /// the node runs, which refers to a file.
+    ///
+    /// If you use [ImportMode::TryReference], Iroh will assume that the data will not
+    /// change and will share it in place without copying to the Iroh data directory
+    /// if appropriate. However, for tiny files, Iroh will copy the data.
+    ///
+    /// If you use [ImportMode::Copy], Iroh will always copy the data.
```
matheus23, line 225:
```suggestion
    /// If you use [`ImportMode::TryReference`], Iroh will assume that the data will not
    /// change and will share it in place without copying to the Iroh data directory
    /// if appropriate. However, for tiny files, Iroh will copy the data.
    ///
    /// If you use [`ImportMode::Copy`], Iroh will always copy the data.
```

I'm just a little bit of a fan of this kind of formatting :grin:

iroh/src/client/blobs/batch.rs
```diff
@@ -0,0 +1,460 @@
...
+                    res = Some(hash);
+                }
+                _ => {}
+            }
+        }
+        let hash = res.context("Missing answer")?;
+        Ok(self.local_temp_tag(HashAndFormat { hash, format }, Some(size)))
+    }
+
+    /// Add a collection
```
matheus23, line 380:
```suggestion
    /// Add a collection.
```

iroh/src/client/blobs/batch.rs
```diff
@@ -0,0 +1,460 @@
...
+    ) -> Self {
+        let updates = updates.buffer(buffer_size);
+        Self(Arc::new(BatchInner {
+            batch,
+            rpc,
+            updates: updates.into(),
+        }))
+    }
+
+    /// Write a blob by passing bytes.
```
matheus23, line 140:
```suggestion
    /// Writes a blob from given bytes.
```

iroh/src/client/blobs/batch.rs
```diff
@@ -0,0 +1,460 @@
...
+    pub async fn add_bytes(&self, bytes: impl Into<Bytes>) -> Result<TempTag> {
+        self.add_bytes_with_opts(bytes, Default::default()).await
+    }
+
+    /// Import a blob from a filesystem path, using the default options.
```
matheus23, line 145:
```suggestion
    /// Imports a blob from a filesystem path, using the default options.
```

iroh/src/client/blobs/batch.rs
```diff
@@ -0,0 +1,460 @@
...
+    ///
+    /// For more control, use [`Self::add_file_with_opts`].
+    pub async fn add_file(&self, path: PathBuf) -> Result<(TempTag, u64)> {
+        self.add_file_with_opts(path, AddFileOpts::default()).await
+    }
+
+    /// Add a directory as a hashseq in iroh collection format
```
matheus23, line 152:
```suggestion
    /// Adds a directory as a hashseq in iroh collection format.
```

iroh/src/client/blobs/batch.rs
```diff
@@ -0,0 +1,460 @@
...
+    pub async fn add_dir(&self, root: PathBuf) -> Result<TempTag> {
+        self.add_dir_with_opts(root, Default::default()).await
+    }
+
+    /// Write a blob by passing an async reader.
```
matheus23, line 157:
```suggestion
    /// Writes a blob by reading given async reader.
```

iroh/src/client/blobs/batch.rs
```diff
@@ -0,0 +1,460 @@
...
+    ///
+    /// This will use a default chunk size of 64KB, and a format of [BlobFormat::Raw].
+    pub async fn add_reader(
+        &self,
+        reader: impl AsyncRead + Unpin + Send + 'static,
+    ) -> anyhow::Result<TempTag> {
+        self.add_reader_with_opts(reader, Default::default()).await
+    }
+
+    /// Write a blob by passing a stream of bytes.
```
matheus23, line 167:
```suggestion
    /// Writes a blob by reading given a stream of bytes.
```

iroh/src/client/blobs/batch.rs
```diff
@@ -0,0 +1,460 @@
...
+            .rpc(BatchCreateTempTagRequest {
+                batch: self.0.batch,
+                content,
+            })
+            .await??;
+        // Only after success of the above call, we can create the corresponding local temp tag
+        Ok(self.local_temp_tag(content, None))
+    }
+
+    /// Write a blob by passing an async reader.
```
matheus23, line 193:
```suggestion
    /// Writes a blob by passing an async reader.
```

iroh/src/client/blobs/batch.rs
```diff
@@ -0,0 +1,460 @@
...
+        &self,
+        reader: impl AsyncRead + Unpin + Send + 'static,
+        opts: AddReaderOpts,
+    ) -> anyhow::Result<TempTag> {
+        let AddReaderOpts { format, chunk_size } = opts;
+        let input = ReaderStream::with_capacity(reader, chunk_size);
+        self.add_stream_with_opts(input, format).await
+    }
+
+    /// Write a blob by passing bytes.
```
matheus23, line 206:
```suggestion
    /// Writes a blob from given bytes.
    ///
    /// You can specify which [`BlobFormat`] they should have (hash sequence or raw blob).
    /// This matters for the garbage collector and for transfers with the blobs protocol.
```

iroh/src/client/blobs/batch.rs
```diff
@@ -0,0 +1,460 @@
...
+    pub async fn add_bytes_with_opts(
+        &self,
+        bytes: impl Into<Bytes>,
+        format: BlobFormat,
+    ) -> Result<TempTag> {
+        let input = futures_lite::stream::once(Ok(bytes.into()));
+        self.add_stream_with_opts(input, format).await
+    }
+
+    /// Import a blob from a filesystem path.
```
matheus23, line 216:
```suggestion
    /// Imports a blob from a filesystem path.
```

iroh/src/client/blobs/batch.rs
```diff
@@ -0,0 +1,460 @@
...
+        Ok((
+            self.local_temp_tag(HashAndFormat { hash, format }, Some(size)),
+            size,
+        ))
+    }
+
+    /// Add a directory as a hashseq in iroh collection format
+    ///
+    /// This can also be used to add a single file as a collection, if
+    /// wrap is set to [WrapOption::Wrap].
```
matheus23, line 280:
```suggestion
    /// Adds a directory as a hashseq in iroh collection format.
    ///
    /// This can also be used to add a single file as a collection, if
    /// wrap is set to [`WrapOption::Wrap`].
```

iroh/src/client/blobs/batch.rs
```diff
@@ -0,0 +1,460 @@
...
+            .into_iter()
+            .map(|(name, hash, _, tag)| ((name, hash), tag))
+            .unzip();
+
+        let tag = self.add_collection(collection).await?;
+        drop(child_tags);
+        Ok(tag)
+    }
+
+    /// Write a blob by passing a stream of bytes.
```
matheus23, line 327:
```suggestion
    /// Writes a blob from given stream of bytes.
```

iroh/src/client/blobs/batch.rs
```diff
@@ -0,0 +1,460 @@
...
+    /// the hash sequence.
+    ///
+    /// Note that this does not guarantee that the data that the collection refers to
+    /// actually exists. It will just create 2 blobs, the metadata and the hash sequence
+    /// itself.
+    pub async fn add_collection(&self, collection: Collection) -> Result<TempTag> {
+        self.add_blob_seq(collection.to_blobs()).await
+    }
+
+    /// Add a sequence of blobs, where the last is a hash sequence.
```
matheus23, line 393:
```suggestion
    /// Adds a sequence of blobs, where the last is a hash sequence.
```

iroh/src/client/blobs/batch.rs
```diff
@@ -0,0 +1,460 @@
...
+            let blob = blobs.next().context("Failed to get next blob")?;
+            if blobs.peek().is_none() {
+                return self.add_bytes_with_opts(blob, BlobFormat::HashSeq).await;
+            } else {
+                tags.push(self.add_bytes(blob).await?);
+            }
+        }
+    }
+
+    /// Upgrade a temp tag to a persistent tag.
```
matheus23, line 412:
```suggestion
    /// Upgrades a temp tag to a persistent tag.
```

iroh/src/client/blobs/batch.rs
```diff
@@ -0,0 +1,460 @@
...
+            .rpc(tags::CreateRequest {
+                value: tt.hash_and_format(),
+                batch: Some(self.0.batch),
+                sync: SyncMode::Full,
+            })
+            .await??;
+        Ok(tag)
+    }
+
+    /// Upgrade a temp tag to a persistent tag with a specific name.
```
matheus23, line 426:
```suggestion
    /// Upgrades a temp tag to a persistent tag with a specific name.
```

iroh/src/client/blobs/batch.rs
```diff
@@ -0,0 +1,460 @@
...
+                value: Some(tt.hash_and_format()),
+                batch: Some(self.0.batch),
+                sync: SyncMode::Full,
+            })
+            .await??;
+        Ok(())
+    }
+
+    /// Upgrade a temp tag to a persistent tag with either a specific name or
+    /// an automatically generated name.
```
matheus23, line 441:
```suggestion
    /// Upgrades a temp tag to a persistent tag with either a specific name or
    /// an automatically generated name.
```

iroh/src/node/rpc.rs
```diff
@@ -921,6 +990,124 @@ impl<D: BaoStore> Handler<D> {
         })
     }
 
+    #[allow(clippy::unused_async)]
```
matheus23, line 993:
```suggestion
```

iroh/tests/batch.rs
```diff
@@ -0,0 +1,239 @@
...
+    iroh::node::Node::memory()
+        .gc_policy(GcPolicy::Interval(Duration::from_millis(10)))
+        .spawn()
+        .await
+}
+
+async fn wait_for_gc() {
+    // wait for multiple gc cycles to ensure that the data is actually gone
+    tokio::time::sleep(Duration::from_millis(50)).await;
+}
```
matheus23, line 21: Is there no other way of getting notified when GC has run or perhaps even triggering GC ourselves?
I really don't want yet another suite of tests that runs 50+ seconds :(

matheus23 (changes requested):
All in all I'm pretty happy with this. I'd love to now be able to turn all *other* non-batch blob APIs into ones that always call `sync()`, but I guess before we can really do that, we'd have needed to make sure that the `docs` API also has a batch API without immediate `sync()`s after every operation.
Anyhow, that's for another PR, but just repeating that that's why I'm excited for this.

As for the review:
- Lots of small documentation nits/suggestions
- I'm not really happy with the `wait_for_gc` function simply sleeping for 50s. We *really* shouldn't have tests like that.

iroh/tests/batch.rs
rklaehn: It's 50 milliseconds. Sure, I can hook up some callbacks, but I think 50ms is fine...

matheus23: We have [`iroh::node::Builder::register_gc_done_cb`](https://n0-computer.github.io/iroh/pr/2545/docs/iroh/node/struct.Builder.html#method.register_gc_done_cb) can we use this here instead?
I think this may be more robust against slow machines, bad CI days and such.

matheus23 (approved):

PR #2547, opened 2024-07-25
ci: Generate docs for each PR

matheus23:
## Description

Added a `docs.yaml` workflow that makes the github action bot reply to PRs with a link to built docs to your PRs.
It only does so once, and then just updates the comment (although actually it doesn't necessarily need to, the link is always the same, but it's still nice in case someone changes the workflow).

This is using the nightly toolchain, because we're using [`#[feature(doc_cfg)]`](https://doc.rust-lang.org/unstable-book/language-features/doc-cfg.html).

## Motivation

I personally wanted this. Every now and then we get a PR with the description saying "the best way to review this is to start by looking at the generated docs". But then I need to checkout the PR, build the docs and open them. Having a link directly to the docs would be *amazing* IMO.

I *also* think that having easy access to docs on every PR will make people check out the rendered docs on PRs more often. My hope here is that doc quality will thus improve.

## Breaking Changes

None of course. This is only CI.

## Notes & open questions

Wdyt?

## Change checklist

- [x] Self-review.
- ~~[ ] Tests if relevant.~~

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -12,6 +12,7 @@ jobs:
     env:
       RUSTC_WRAPPER: "sccache"
       SCCACHE_GHA_ENABLED: "on"
+      SCCACHE_CACHE_SIZE: "50G"
       PREVIEW_PATH: pr/${{ github.event.pull_request.number }}/docs
 
     steps:
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -49,7 +49,7 @@ jobs:
         issue-number: ${{ github.event.pull_request.number }}
         comment-id: ${{ steps.fc.outputs.comment-id }}
         body: |
-          Documentation for this PR has been generated and is available at: https://${{ github.repository_owner }}.github.io/${{ github.event.repository.name }}/${{ env.PREVIEW_PATH }}/
+          Documentation for this PR has been generated and is available at: https://${{ github.repository_owner }}.github.io/${{ github.event.repository.name }}/${{ env.PREVIEW_PATH }}/iroh/
           
           Last updated: ${{ github.event.pull_request.updated_at }}
         edit-mode: replace
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -27,18 +27,13 @@ jobs:
       env:
         RUSTDOCFLAGS: --cfg docsrs
 
-    - name: Prepare Docs for Deployment
-      run: |
-        mkdir -p ${{ env.PREVIEW_PATH }}
-        cp -R target/doc/* ${{ env.PREVIEW_PATH }}/
-
     - name: Deploy Docs to Preview Branch
       uses: peaceiris/actions-gh-pages@v4
       with:
         github_token: ${{ secrets.GITHUB_TOKEN }}
-        publish_dir: .
+        publish_dir: ./target/doc/
+        destination_dir: ${{ env.PREVIEW_PATH }}
         publish_branch: generated-docs-preview
-        keep_files: true
 
     - name: Find Docs Comment
       uses: peter-evans/find-comment@v3
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -24,7 +24,7 @@ jobs:
       uses: mozilla-actions/sccache-action@v0.0.5
 
     - name: Generate Docs
-      run: cargo doc --workspace --all-features --no-deps --document-private-items
+      run: cargo doc --workspace --all-features --no-deps
       env:
         RUSTDOCFLAGS: --cfg docsrs
 
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -12,6 +12,8 @@ jobs:
     env:
       RUSTC_WRAPPER: "sccache"
       SCCACHE_GHA_ENABLED: "on"
+      PREVIEW_PATH: pr/${{ github.event.pull_request.number }}/docs
+
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


matheus23:
I think this isn't far from working, but I think someone with more access rights than I have needs to setup github pages in the settings for this to work.
Whoever has access, please ping me if you'd like to make this happen :v:

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,52 @@
...
+    if: "github.event_name == 'pull_request'"
+    runs-on: ubuntu-latest
+    env:
+      RUSTC_WRAPPER: "sccache"
+      SCCACHE_GHA_ENABLED: "on"
+    steps:
+    - uses: actions/checkout@v4
+    - uses: dtolnay/rust-toolchain@master
+      with:
+        toolchain: nightly-2024-05-02
```
flub, line 19: why nightly?

matheus23: Hm. This is copied from the `check_docs` job in `ci.yaml`. I don't know?

flub:
> Documentation for this PR has been generated and is available at: https://n0-computer.github.io/iroh/2547/
> 
> Last updated: 2024-07-25T13:16:26Z

The link is a 404 now, is that expected?  How long do they stay valid?

matheus23:
@flub see my msg above: https://github.com/n0-computer/iroh/pull/2547#issuecomment-2250345711

This doesn't work yet :( Need an admin to jump in.

.github/workflows/docs.yaml
flub: try changing back to stable, i can't see any reason in the commit that switched to nightly and in general we only want to be on nightly until the specific issue is fixed in stable.  nightly get too outdated otherwise.

flub:
> @flub see my msg above: [#2547 (comment)](https://github.com/n0-computer/iroh/pull/2547#issuecomment-2250345711)
> 
> This doesn't work yet :( Need an admin to jump in.

oops, missed that.

but do they stay valid forever?  will we ever run out of space?

matheus23:
> but do they stay valid forever? will we ever run out of space?

IIUC, the github action will push to a branch (that's `gh-pages` by default, but we can customize, I'd suggest using `docs-preview`).

My plan was for the github action to publish different PRs to different directories on that branch, indexed by PR number. We'd have one directory for each PR.

We can go in there by hand and remove old PRs. From reading the [github documentation](https://docs.github.com/en/repositories/creating-and-managing-repositories/repository-limits), I don't see any limits to repository size, though.

flub:
I guess if it's just a repo than we have unlimited storage and can as well leave it there.  it's nicer that it doesn't expire.  just surprises me they are that generous :)

.github/workflows/docs.yaml
matheus23: Now we know why:
```
error[E0554]: `#![feature]` may not be used on the stable release channel
  --> iroh/src/lib.rs:90:21
   |
90 | #![cfg_attr(docsrs, feature(doc_cfg))]
   |                     ^^^^^^^^^^^^^^^^
```

flub: Great!  Can you make sure this is described in the commit message?  (the PR description becomes the commit message)

matheus23:
Forgot to mention this now just works (thus I flipped it to ready for review). The comment above is a demo of it working: https://github.com/n0-computer/iroh/pull/2547#issuecomment-2250305738

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,55 @@
...
+jobs:
+  preview_docs:
+    timeout-minutes: 30
+    name: Docs preview
+    if: "github.event_name == 'pull_request'"
+    runs-on: ubuntu-latest
+    env:
+      RUSTC_WRAPPER: "sccache"
+      SCCACHE_GHA_ENABLED: "on"
+      PREVIEW_PATH: pr/${{ github.event.pull_request.number }}/docs
```
flub, line 15: You also need to set the cache size.  If we have one action that uses sccache with a different size it will blow away half of the cache of the runner every time it runs.  The default size is a lot smaller:
https://github.com/n0-computer/iroh/blob/0fc37942be3d68399fbe45401ba7d67be43a83a6/.github/workflows/ci.yml#L20

flub (approved):
LGTM but please follow up on the two comments before merging

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,55 @@
...
+    steps:
+    - uses: actions/checkout@v4
+    - uses: dtolnay/rust-toolchain@master
+      with:
+        toolchain: nightly-2024-05-02
+    - name: Install sccache
+      uses: mozilla-actions/sccache-action@v0.0.5
+
+    - name: Generate Docs
+      run: cargo doc --workspace --all-features --no-deps --document-private-items
```
matheus23, line 26: One open question I still have is: Should we use `--document-private-items`?
This is a good option for the `check_docs` job in the other workflow, but may not be good for the goals of this PR, as this PR is way more about a "simulation of what the actual docs would look like". (Which don't contain private items)

flub: Hum, yes.  Good point.  Let's remove `--document-private-items` as well

.github/workflows/docs.yaml
matheus23: Added something to the PR description :v:

.github/workflows/docs.yaml
matheus23: Oh shit. Thanks! Good call.

.github/workflows/docs.yaml
matheus23: Alright. I switched this setting off now, I think that's better. So no more `--document-private-items`.

PR #2551, opened 2024-07-25
feat: override to staging relays

Arqu:
## Description

While I'm not a huge fan of this piece of code, I don't really see much of an alternative as we want to be able to force the override to use staging relays even in "prod" environments or when it's deeply integrated such as the FFI library cases. 

This should allow us to override the default relay maps for all of our CI setups in all of our repos.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-cli/src/config.rs
+++ b/iroh-cli/src/config.rs
@@ -27,8 +27,6 @@ const ENV_AUTHOR: &str = "IROH_AUTHOR";
 const ENV_DOC: &str = "IROH_DOC";
 const ENV_CONFIG_DIR: &str = "IROH_CONFIG_DIR";
 const ENV_FILE_RUST_LOG: &str = "IROH_FILE_RUST_LOG";
-#[allow(dead_code)]
-const ENV_FORCE_STAGING_RELAYS: &str = "IROH_FORCE_STAGING_RELAYS";
 
 /// CONFIG_FILE_NAME is the name of the optional config file located in the iroh home directory
 pub(crate) const CONFIG_FILE_NAME: &str = "iroh.config.toml";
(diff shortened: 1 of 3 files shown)


iroh-net/src/endpoint.rs
```diff
@@ -59,6 +59,9 @@ pub use iroh_base::node_addr::{AddrInfo, NodeAddr};
 /// is still no connection the configured [`Discovery`] will be used however.
 const DISCOVERY_WAIT_PERIOD: Duration = Duration::from_millis(500);
 
+/// Environment variable to force the use of staging relays.
+pub const ENV_FORCE_STAGING_RELAYS: &str = "IROH_FORCE_STAGING_RELAYS";
```
dignifiedquire, line 63: if you put the env var in here, I would also put the check in here and make it public function

iroh-cli/src/config.rs
```diff
@@ -68,24 +68,34 @@ pub(crate) struct NodeConfig {
...
+            } else {
+                use defaults::prod::{
+                    default_ap_relay_node, default_eu_relay_node, default_na_relay_node,
+                };
+                [
+                    default_na_relay_node(),
+                    default_eu_relay_node(),
+                    default_ap_relay_node(),
+                ]
+                .into()
```
dignifiedquire, line 94: this should just be a function from iroh_net, to avoid this duplication

iroh-net/src/relay/map.rs
```diff
@@ -133,6 +145,17 @@ pub struct RelayNode {
     pub stun_port: u16,
 }
 
+impl RelayNode {
+    /// Creates a new relay node from itself.
+    pub fn copied(&self) -> Self {
```
Arqu, line 150: I don't know what's the expected naming or way to do this properly. I just needed to get it out of the arc and this was cheap and easy enough.

iroh-net/src/endpoint.rs
dignifiedquire: does this still need `pub`?

iroh-net/src/relay/map.rs
dignifiedquire: no need for this method, just use this to get it out of the arc: https://doc.rust-lang.org/std/sync/struct.Arc.html#method.unwrap_or_clone

Arqu: Figured there was a saner way to do this.

dignifiedquire (approved):

PR #2553, opened 2024-07-26
feat(iroh): Improve documentation and canonicalize docs in `iroh::client`

matheus23:
## Description

- More pointers between individual modules to help people understand how they go from `A` to `B` (e.g. `Node` -> `Iroh`, `Iroh` -> `blobs::Client`, etc.)
- Documentation on how to initialize `client::Iroh`
- Generally make methods adhere to our documentation style.
- Expose the type alias `iroh::client::RpcClient`, so it doesn't show up as `RpcClient<RpcService, Connection<RpcService>>` everywhere it's used in the docs.

## Breaking Changes

None, but noteworthy:
- Expose the type alias `iroh::client::RpcClient`.

## Change checklist

- [X] Self-review.
- ~~[ ] Tests if relevant.~~

### DIFF
--- a/iroh/src/client/docs.rs
+++ b/iroh/src/client/docs.rs
@@ -312,7 +312,7 @@ impl Doc {
         Ok(res.entry.map(|entry| entry.into()))
     }
 
-    /// Returns entries.
+    /// Returns all entries matching the query.
     pub async fn get_many(
         &self,
         query: impl Into<Query>,
--- a/iroh/src/client/docs.rs
+++ b/iroh/src/client/docs.rs
@@ -51,14 +51,14 @@ pub struct Client {
 }
 
 impl Client {
-    /// Create a new document.
+    /// Creates a new document.
     pub async fn create(&self) -> Result<Doc> {
         let res = self.rpc.rpc(CreateRequest {}).await??;
         let doc = Doc::new(self.rpc.clone(), res.id);
         Ok(doc)
     }
 
-    /// Delete a document from the local node.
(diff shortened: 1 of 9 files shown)


iroh/src/client/docs.rs
```diff
@@ -306,7 +312,7 @@ impl Doc {
         Ok(res.entry.map(|entry| entry.into()))
     }
 
-    /// Get entries.
+    /// Returns entries.
```
flub, line 315:
```suggestion
    /// Returns all entries matching the query.
```
?

iroh/src/client/docs.rs
```diff
@@ -378,14 +384,14 @@ impl Doc {
...
-    /// Get status info for this document
+    /// Returns status info for this document
     pub async fn status(&self) -> anyhow::Result<OpenState> {
         self.ensure_open()?;
         let res = self.rpc(StatusRequest { doc_id: self.id() }).await??;
         Ok(res.status)
     }
 
-    /// Set the download policy for this document
+    /// Sets the download policy for this document
```
flub, line 394: Things like this should probably explain a bit what download policies are and what is available and how they affect things.

But I'll take any improvement, so like this is also fine.

flub (approved):
**adding** a public api is not a breaking change is it?

iroh/src/client/docs.rs
matheus23: Agreed. Happy to pick this up on another docs day :)

PR #2556, opened 2024-07-26
[ignore] chore(ci): ci warnings stuff

divagant-martian:
## Description

this is commit `980b53d: refactor(iroh-bytes): Rewrite the blob store to use redb (#2051)`
with two commits on top: `git restore .github --source main`
and `cargo update -p time` because of some regression in type inference or something

This generates some warnings:

Now quickly about the trait Io is never used and others, I kinda ignored them at the point because that's not what I wanted to check but those should ofc make ci fail as well.

About the cfg. This not unification or anything like that, not related to features of dependencies, etc. There is a line in that commit (two lines actually) that put some tests behind said feature with
```
warning: unexpected `cfg` condition value: `file-db`
   --> iroh-cli/tests/cli.rs:216:7
    |
216 | #[cfg(feature = "file-db")]
```

this is an issue because those tests have not been compiled at all in the last 4-5 months since that line was added. So, I wanted to understand how is that we didn't realize

My question here is **why are the test workflows not failing now if I'm using the latest github folder**

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+# Keep GitHub Actions up to date with GitHub's Dependabot...
+# https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot
+# https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem
+version: 2
+updates:
+  - package-ecosystem: github-actions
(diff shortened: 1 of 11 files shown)


divagant-martian:
Updates to the last commit before the issue was fixed and it's reproducible that tests don't fail. However, I can't really figure out what's going on and latest code seems to work _even if this already uses latest .github folder_. Closing this since I've run out of ideas for the time being

PR #2558, opened 2024-07-29
fix: disable docs preview on forks

Arqu:
## Description

We disable docs previews on forks because otherwise external contributors cant get a ✅  on CI. The reason for it being is that those forks don't get write permissions to the branch where docs are hosted and the alternative would be to manually juggle a lot of tokens for each user.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -7,7 +7,7 @@ jobs:
   preview_docs:
     timeout-minutes: 30
     name: Docs preview
-    if: "github.event_name == 'pull_request'"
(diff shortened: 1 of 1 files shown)


flub (approved):
Please write a description (why is it disabled?  in the future i won't know).  And use the checkboxes :)

matheus23 (approved):
Thank you!

Found this discussion where this option was mentioned/discussed: https://github.com/orgs/community/discussions/25217#discussioncomment-3246904

PR #2559, opened 2024-07-29
test(iroh-blobs): comment out ignored test (that is not a flaky test)

matheus23:
## Description

There's this test that was introduced in [the `local_pool` PR](https://github.com/n0-computer/iroh/pull/2517).
It was ignored via `#[ignore = "todo"]`. Notably, it's *not flaky*, it always fails.
Our flaky tests are run with `cargo nextest run --run-ignored all [...]`. We can't be more specific with the `ignore`d tests. The only options are `default`, `ignored-only` and `all`.
This kind of test is really hard to write. IIUC, `#[should_panic]` can only test for the panic happening in the thread that the test is initiated in, it doesn't detect panics that are thrown in threads spawned from the test. I assume this is the reason writing this test was abandoned.

Keeping this test with the `#[ignore = "todo"]` on it means we're always running it in our flaky test suite, which is confusing. We thought this test was flaky, but it's not. IMO it's better to comment it out/remove it than to pollute our flaky test results.

## Breaking Changes

None

## Notes & open questions

In this PR I'm commenting this test. Should we remove it instead?
Or do people have ideas on how to make this test work?
Do we have an idea what we're *expecting* of our `LocalPool` implementation? Should a panic on one of the threads cause a panic in the `finish()` function?

## Change checklist

- [X] Self-review.
- ~~[ ] Tests if relevant.~~

### DIFF
--- a/iroh-blobs/src/util/local_pool.rs
+++ b/iroh-blobs/src/util/local_pool.rs
@@ -634,20 +634,20 @@ mod tests {
         assert_eq!(c2.load(std::sync::atomic::Ordering::SeqCst), 0);
     }
 
-    #[tokio::test]
-    #[should_panic]
-    #[ignore = "todo"]
-    async fn test_panic() {
-        let _ = tracing_subscriber::fmt::try_init();
-        let pool = LocalPool::new(Config {
-            threads: 2,
-            ..Config::default()
-        });
-        pool.spawn_detached(|| async {
-            panic!("test panic");
-        });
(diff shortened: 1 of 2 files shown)


divagant-martian (approved):
personally agree. 

Having a test that _always_ fails makes it harder to identify real flakes vs outright broken stuff failing 100% of the time. As @matheus23 said: alarm fatigue

matheus23:
The irony of a new flaky test keeping this PR from being merged :upside_down_face:

PR #2563, opened 2024-07-29
test(iroh-cli): reduce flakyness of cli_provide_file_resume

divagant-martian:
## Description

In windows there is no way to copy a file being accessed by another process in an accessible way. This is what makes this test fail since it attempt to copy the blobs.db folder a couple of times while the iroh instance that handles is running.

The change is simple: shutdown the provider, copy the files, re-start the provider. From my perspective, this does not affect what the test is attempting to assert.

Now, since this includes re-starting the iroh instance that provides the files several times, and we match on output, sometimes we can get weird logs related to shutdown. One of those (and the only one I have seen so far) is for a netcheck report and didn't finish on time. Instead of logging this in the reportgen actor, the error is bubbled up to be handled by the netcheck actor (which will have shutdown by then) thus reducing noise and allowing for better error handling in the future

## Breaking Changes

n/a

## Notes & open questions

n/a

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/src/netcheck.rs
+++ b/iroh-net/src/netcheck.rs
@@ -327,7 +327,7 @@ pub(crate) enum Message {
     /// A report produced by the [`reportgen`] actor.
     ReportReady { report: Box<Report> },
     /// The [`reportgen`] actor failed to produce a report.
-    ReportAborted,
+    ReportAborted { err: anyhow::Error },
     /// An incoming STUN packet to parse.
     StunPacket {
         /// The raw UDP payload.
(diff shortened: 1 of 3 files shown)


iroh-net/src/netcheck.rs
```diff
@@ -547,10 +547,10 @@ impl Actor {
         }
     }
 
-    fn handle_report_aborted(&mut self) {
+    fn handle_report_aborted(&mut self, err: anyhow::Error) {
         self.in_flight_stun_requests.clear();
         if let Some(ReportRun { report_tx, .. }) = self.current_report_run.take() {
-            report_tx.send(Err(anyhow!("report aborted"))).ok();
+            report_tx.send(Err(anyhow!("report aborted: {err}"))).ok();
```
dignifiedquire, line 553: nit, I think this could be `err.context("report aborted")` instead

dignifiedquire (approved):
less flaky tests make me happy, one small nit

divagant-martian:
while I had to re-add the flaky mark since one of the windows worflows failed, I think it's still an improvement: the test is now flaky instead of failing 100% of the time. Addressed your need and if it's ok I'll merge and we can keep looking into the new kinds of failures in later antiflaky attempts

PR #2564, opened 2024-07-29
test(iroh-cli): make cli resumption tests not flaky

divagant-martian:
## Description

Takes a different approach to tests by doing the setup for them all first
before running the tests, to prevent windows from whining about a locked
blobs.db used by the provider. This still requires an ephemeral iroh to obtain
the blobs.db but does not require much re-starting, making it less prone to
weird shutdown issues. 

## Breaking Changes

n/a

## Notes & open questions

Note that there is sleep as a synchronization point because otherwise the
folder for the tests will not have the correct contents. Unexpected but have 
seen it happen a lot. We also use a different folder for the iroh instance
that actually is providing during the tests, in case shutdown is not done
for the instance that produces the blobs.db, when the one that provides is 
started.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


matheus23:
We were duplicating a bit of work here, I was going through exactly the same motions yesterday :upside_down_face: 

The [last issue I've hit](https://github.com/n0-computer/iroh/actions/runs/10146348415/job/28054192852) was port binding not working when the iroh is rapidly stopped & started again.
> |Error: iroh is already running on port 4919

I think in rare cases port binding can be racy. Have you hit this as well?

iroh-cli/tests/cli.rs
```diff
@@ -142,74 +142,98 @@ fn cli_provide_tree_resume() -> Result<()> {
...
+            name: "missing blobs produces partial transfer",
+            get_folder: missing_blobs_data_dir,
+            transfer_size: "98.04 KiB",
+        },
+        TestCase {
+            name: "partial blobs produces partial transfer",
+            get_folder: partial_data_dir,
+            transfer_size: "65.98 KiB",
+        },
+    ];
```
matheus23, line 212: Do you have thoughts on using a `struct TestCase` vs. writing a small `fn test_case(name: &'static str, get_folder: PathBuf, transfer_size: &'static str)` that is called four times?

Could also make it not a `fn`, but a closure instead. That way it can capture references from the environment, thus not needing extra arguments beyond the three.

matheus23 (approved):
Really happy to take this improvement. I've played with these tests before. Glad you found a somewhat consistent way to prevent the lock on `blobs.db`, even if it means our tests sleep for 3 seconds.

iroh-cli/tests/cli.rs
divagant-martian: I would also prefer something less verbose, but I would much rather have a struct that very explicitly lets us know what is that we are testing, than go back to the previous state where, at least imo, it was not clear what the test is intended to assert. I'll give it a try to maybe a fn instead of a loop, and yeah if it's indeed equally clear but less verbose let's go with it

divagant-martian: ok, this required more refactoring and renaming to make cargofmt fit each test on one line and actually reduce line count, but at the end it's indeed less verbose :)

iroh-cli/tests/cli.rs
```diff
@@ -226,58 +222,55 @@ fn cli_provide_file_resume() -> Result<()> {
...
+            assert_eq!(matches, vec![transfer_size], "{name}: wrong transfer size");
+            let current_hash = Hash::new(std::fs::read(&tgt)?);
+            assert_eq!(current_hash, hash, "{name}: wrong blob contents");
+            std::fs::remove_file(&tgt)?;
+            Ok(())
+        };
+
+    run_test("no data needs full transfer", empty_data_dir, "98.04 KiB")?;
+    run_test("full folder needs no transfer", full_data_dir, "0 B")?;
+    run_test("partial data needs transfer", partial_data_dir, "65.98 KiB")?;
```
dignifiedquire, line 273: much more readable <3

PR #2566, opened 2024-07-30
refactor(iroh-net): Move more server code behind `iroh-relay` feature flag

matheus23:
## Description

This moves more relay server code behind the `iroh-relay` feature flag.

This is a continuation of #2516 

Unfortunately I didn't quite catch "all server things".

I'm fairly confident I do have all covered now, because this is extracted from a branch where server-side code wouldn't compile (to Wasm).

Also:
- Makes use of `MaybeTlsStream` instead of the `MaybeTlsStreamServer` alias. Using the original definition instead of the alias confuses rust-analyzer a lot less. And the aliasing back-and-forth was silly.

## Breaking Changes

None

## Change checklist

- [x] Self-review.
- [x] Documentation updates following the [style guide](https://rust-lang.github.io/rfcs/1574-more-api-documentation-conventions.html#appendix-a-full-conventions-text), if relevant.
- ~~[ ] Tests if relevant.~~
- [x] All breaking changes documented.

PR #2568, opened 2024-07-30
refactor(iroh-net)!: Improve server modules structure & rename structs

matheus23:
## Description

This is moving around the relay modules a bit. Mostly so that all server-specific code is under `iroh-net/src/relay/server/`.
This means there's only a handful of `#[cfg(feature = "iroh-relay")]`s in `relay.rs` and a bunch of other places due to unused warnings otherwise.

I've also renamed what was previously `iroh::relay::client::{Client, ClientBuilder, ClientReceiver, ...}` to `iroh::relay::client::conn::{Conn, ConnBuilder, ConnReceiver, ...}`, because otherwise the `Client*` names were duplicated in two modules that were in a parent-child relationship, which is super confusing.
And really, the inner `Client`s are more about individual connections.

I also resolved a TODO around moving to an `AbortingJoinHandle` in favor of a custom `RelayHttpServerGuard`.

## Breaking Changes

- Refactored the module structure for users of the `iroh-relay` feature in iroh-net.

## Notes & open questions

I realize the diff here is quite substantial. Unfortunately the diff is mostly just bad because I ended up "swapping places" for modules in two cases, which ended up creating really bad diffs.

I know the diff is quite sad, so it's fair if we say "this is unreviewable" and I scrap this piece of work. I'm not blocked on this merging, and I only fairly quickly knocked this out today.
As Mark Twain famously said
> I apologize for such a long PR - I didn't have time to write a short one.

Or sth like that.

## Change checklist

- [X] Self-review.
- ~~[ ] Tests if relevant.~~
- [X] All breaking changes documented. Yeah? Kinda?


iroh-net/src/relay/client.rs
```diff
@@ -1,511 +1,1112 @@
-//! based on tailscale/derp/derp_client.go
-use std::net::SocketAddr;
-use std::pin::Pin;
+//! Based on tailscale/derp/derphttp/derphttp_client.go
```
matheus23, line 1: As you can see, the diff here is fucked :grimacing: 

I moved `relay/http/client.rs` to here, and moved the old `relay/client.rs` to `relay/client/conn.rs`.
I also renamed a bunch of the `Client*` things to `Conn*`, which probably made this diff look bad.

iroh-net/src/relay/server.rs
```diff
@@ -1,903 +1,1042 @@
-//! based on tailscale/derp/derp_server.go
-use std::collections::HashMap;
+//! A full-fledged iroh-relay server.
```
matheus23, line 1: Again, the diff here ended up bad. I think it may actually be due to the fact that I've moved this module somewhere else while simultaneously moving something else to here :/

iroh-net/src/relay/server/http_server.rs
```diff
@@ -682,3 +681,235 @@ impl std::ops::DerefMut for Handlers {
         &mut self.0
     }
 }
+
+#[cfg(test)]
+mod tests {
```
matheus23, line 686: I moved these tests to here from `relay/http.rs`, as that module mainly included types and constants, not the code.

matheus23:
There's some small unused things remaining. Will update the PR tomorrow morning :v:

In the mean time - I've written up some notes in the PR description.

dignifiedquire:
seems mostly fine from my side, @ramfox is the master of this code though, so she will have to deliver final verdict

ramfox (approved):
Okay, went through this. Looks good. I especially like the rename from `Client` to `Conn`. I think `client::Conn` and `server::ClientConn` mirror each other better now. It makes it a bit more clear that the `server::ClientConn` is the server-side driver of the connection to the client.

Approved! Nice work! Happy to re-review on the last bits of clean up when you need it.

matheus23:
Current state of this is:
I need https://github.com/n0-computer/iroh/pull/2566 to be reviewed and then merged first, because this depends on that.

flub (approved):
nice restructure!

looking forward to merging this into the quinn11 branch :upside_down_face:

PR #2569, opened 2024-07-31
refactor(iroh-net): Switch to (now stable) `IpAddr::to_canonical`

matheus23:
## Description

Very small refactor. Just switching away from something vendored.
- `IpAddr::to_canonical` is [stable since version 1.75](https://doc.rust-lang.org/std/net/enum.IpAddr.html#method.to_canonical).
- We've set MSRV to 1.76 in lots of CI jobs (and 1.66 in some remaining ones?)
- I want to cfg-out the `net/ip.rs` module in the Wasm case
- `to_canonical` is used in a bunch of places, so I can only cfg it out, if that function is kept. Thus the switch to the stable version

## Breaking Changes

- Removed `iroh::net::net::ip::to_canonical`. Use `std::net::IpAddr::to_canonical` instead.

## Change checklist

- [X] Self-review.
- ~~[ ] Tests if relevant.~~


flub (approved):

dignifiedquire:
> * (and 1.66 in some remaining ones?)

that was likely just an oversight

dignifiedquire (approved):

PR #2570, opened 2024-07-31
feat(iroh-gossip)!: dispatch gossip events and updates by topic

Frando:
## Description

This PR changes the main public `iroh_gossip` to keep track of client-side gossip subscriptions. The `net::Gossip` struct now keeps track of client-side subscribers per topic, which are made up of a pair of two streams/channels: from the client to the actor a stream of updates (outgoing messages) and from the actor to the client a stream of events (incoming messages). Once all client streams&sinks for a topic are dropped, the topic is being quit.

This builds on the client API added in #2258, but completely removes the `dispatcher` module, integrating its features directly into the gossip actor. See below for a short list of the API changes. The new API can be browsed [here](https://n0-computer.github.io/iroh/pr/2570/docs/iroh/gossip/net/index.html).

The refactor turned out bigger than initially intended, sorry for that, but I did not see a good way to reduce the scope.

What's still missing (can also be follow-ups)?:

- [ ] Review the new public API
- [ ] Align the client API to the iroh_gossip API. The `GossipTopic` can be made to work on both the client and the native API, as it only deals with streams and sinks.

## Breaking Changes

* `iroh_gossip::dispatcher` is removed with everything that was in it. use the new API from `iroh_gossip::net::Gossip` instead (see below).
* `iroh_gossip::net::Gossip` methods changed:
  * changed: `join` now returns a `GossipTopic`
  * removed: `broadcast`, `broadcast_neighbors`, `subscribe`, `subscribe_all`, `quit`.
      * for `subscribe` use `join` instead, which returns a `GossipTopic`
      * for `broadcast` and `broadcast_neighbors` use the respective methods on `GossipTopic` . 
      * `quit` is obsolete now, the topic will be quitted once all `GossipTopic` handles are dropped.
      * `subscribe_all` is no longer available
* `iroh_gossip::net::JoinTopicFut` is removed (is now obsolete)

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.


matheus23:
Direct link to the gossip docs: https://n0-computer.github.io/iroh/pr/2570/docs/iroh/gossip/net/index.html :)

Frando:
Tentatively marking ready for review. Still has TODOs (see description) but that could also be a followup.

iroh-docs/src/engine/live.rs
```diff
@@ -29,8 +28,9 @@ use crate::{
     },
     AuthorHeads, ContentStatus, NamespaceId, SignedEntry,
 };
+use crate::{engine::gossip::GossipState, metrics::Metrics};
 
-use super::gossip::{GossipActor, ToGossipActor};
+// use super::gossip::{GossipActor, ToGossipActor};
```
dignifiedquire, line 33: delete

iroh-docs/src/engine/live.rs
```diff
@@ -379,11 +370,6 @@ impl<B: iroh_blobs::store::Store> LiveActor<B> {
     async fn shutdown(&mut self) -> anyhow::Result<()> {
         // cancel all subscriptions
         self.subscribers.clear();
-        // shutdown gossip actor
-        self.gossip_actor_tx
-            .send(ToGossipActor::Shutdown)
-            .await
-            .ok();
```
dignifiedquire, line 386: can't we wait for the gossip actor exit anymore?

iroh-gossip/src/net/handles.rs
```diff
@@ -0,0 +1,254 @@
...
+
+    /// Lists our current direct neighbors.
+    pub fn neighbors(&self) -> impl Iterator<Item = NodeId> + '_ {
+        self.neighbors.iter().copied()
+    }
+
+    /// Waits until we are connected to at least one node.
+    pub async fn joined(&mut self) -> Result<()> {
+        while self.neighbors.is_empty() {
+            let _ = self.try_next().await?;
```
dignifiedquire, line 133: shouldn't this be explicit about the type of event?

iroh-docs/src/engine/live.rs
Frando: there's no gossip actor anymore in the live engine! the `GossipState` is not a loop of its own, it just contains state and abortable handles to the topic loop tasks.

the main `Gossip` is passed into the live engine, so should be shutdown outside it as well.

iroh-gossip/src/net/handles.rs
Frando: the `Joined` event is handled in `try_next`. But we could be more explict and return an error if any other event than `Joined` is returned, yes.

iroh-docs/src/engine/live.rs
Frando: Pushed a commit that properly joins the topic task loops in the sync engine to catch panics. Also added a TODO to catch panics when aborthing other joinsets in the sync engine too.

dignifiedquire (approved):

PR #2572, opened 2024-08-01
fix(iroh-gossip): connection loop misuses `tokio::select!` leading to read errors

Frando:
## Description

The connection loop of iroh-gossip misused tokio-select by selecting over a future that is not cancellation safe. This means that if the timings are bad, the message reading future would be aborted midway in a message, and then restart by reading a length, which would then yield some random number because it would be reading some random bytes in the middle of a message. This means it would lead to random connection drops.

## Breaking Changes

Backport from #2570 

## Change checklist

- [x] Self-review.
- [ ] ~~Tests if relevant.~~

### DIFF
--- a/iroh-gossip/src/net.rs
+++ b/iroh-gossip/src/net.rs
@@ -652,27 +652,26 @@ async fn connection_loop(
     };
     let mut send_buf = BytesMut::new();
     let mut recv_buf = BytesMut::new();
-    loop {
-        tokio::select! {
-            biased;
-            // If `send_rx` is closed,
-            // stop selecting it but don't quit.
-            // We are not going to use connection for sending anymore,
-            // but the other side may still want to use it to
-            // send data to us.
-            Some(msg) = send_rx.recv(), if !send_rx.is_closed() => {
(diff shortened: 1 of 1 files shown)


dignifiedquire (approved):
I wonder if we have this bug in other parts of the stack as well, iroh-net relay handshakes comes to mind

matheus23:
We use `tokio_util::codec::Framed` everywhere else we do something like this. Perhaps we should do so here, too?
`Framed` implements `Sink` and `Stream` both in a cancel-safe manner, [according to its docs](https://docs.rs/tokio-util/latest/tokio_util/codec/struct.Framed.html#cancellation-safety).

> I wonder if we have this bug in other parts of the stack as well, iroh-net relay handshakes comes to mind

Erm. Now I do, too. What exactly are you referring to, though?
We have `iroh-net::relay::codec::read_frame`, which is only used in `recv_client_key`, which is (IIUC), never used in a `tokio::select!`.
Is it something else you're thinking about?

Frando:
Yeah, using framed makes sense here likely.

PR #2574, opened 2024-08-01
test(iroh-net): increase timeout for local swarm discovery test

ramfox:
## Description
The `local_swarm_discovery` test was reliably timing out when run on the linux CI. After checking the logs and NOT finding any evidence that anything goes wrong, I increased the timeout and can no longer trigger the failure.

## Change checklist

- [x] Self-review.


iroh-net/src/discovery/local_swarm_discovery.rs
```diff
@@ -296,8 +296,16 @@ mod tests {
...
 
     #[tokio::test]
-    #[ignore = "flaky"]
+    // #[ignore = "flaky"]
     async fn test_local_swarm_discovery() -> TestResult {
+        let _guard = iroh_test::logging::setup();
+        for _i in 0..10 {
+            local_swarm_discovery().await?;
+        }
+        Ok(())
```
ramfox, line 305: should i keep this loop to catch any future flaky-ness sooner?

matheus23: Hm. It seems weird and somewhat arbitrary to me.
In the worst case this is going to be a test that runs nearly 100 seconds as part of the normal test suite in iroh-net. This could be annoying when developing.

matheus23: Hm. It seems weird and somewhat arbitrary to me.
In the worst case this is going to be a test that runs nearly 100 seconds as part of the normal test suite in iroh-net. This could be annoying when developing.

ramfox: extremely good point

matheus23 (approved):

PR #2576, opened 2024-08-02
fix(iroh): don't hit network for complete blobs&collections

### LINKED ISSUE #2575: Queued blob download from oneself fails
```
> iroh blobs add foo.txt
<HASH>

# Direct succeeds
> iroh blobs get --node <NODE_ID> <HASH>

# Queued fails
> iroh blobs get --node <NODE_ID> --queued <HASH>
 Error: download aborted: Failed to complete download
```
when `NODE_ID` is the node this is running on.

dignifiedquire:
## Description

Two changes to the downloader:

* Never try to download from ourselves. If the only provider node added is our own node, fail with error "no providers".
* Check if we have the blob already available locally *before* starting to open any connections

The second point turned out to be a bit more complicated:
* For hashseqs (collections) we have to check that we have *all* children completely available to not start a download
* We should still emit the `FoundLocal` events so that the state tracking at the client works the same no matter if an actual download was performed or not.

Both these things are now implemented, through a new function `check_local_with_progress_if_complete` in iroh_blobs that is called before starting the download. It checks if the blob is fully available locally (and all children too for hashseqs) and if so returns `true`, and emits the  `FoundLocal` events. No events are emitted if the blob is not completely available, as in that case we start the actual download machinery, which will emit the `FoundLocal` events too on the go.

Adds three tests:

* Downloading a missing blob from the local node fails without trying to connect to ourselves
* Downloading an existing blob succeeds without trying to download
* Downloading an existing collection succeeds without trying to download

Closes #2575

## Notes and open questions

The API around the new function and the exisitng `get_to_db` could be improved, and we could skip the double-checking, but that would be a more involved change to a public API that I didn't want to make.

## Breaking changes

None, only an API addition to the public API of  iroh_blobs: `iroh_blobs::get::check_local_with_progress_if_complete`

### DIFF
--- a/iroh-blobs/src/get/db.rs
+++ b/iroh-blobs/src/get/db.rs
@@ -59,6 +59,73 @@ pub async fn get_to_db<
     }
 }
 
+/// Checks if a blob or collection exists fully in the local store.
+///
+/// Returns `true` if the blob is complete (and for hashseqs if all children are complete too).
+///
+/// Emits the same sequence of progress events as if we were downloading the blob, but only if the
+/// return value is `true`, i.e. if the blob (and all children for hashseqs) are fully available in
+/// the store. If the return value is `false`, no events will be emitted at all.
+pub async fn check_local_with_progress_if_complete<D: BaoStore>(
+    db: &D,
+    hash_and_format: &HashAndFormat,
+    progress: Option<impl ProgressSender<Msg = DownloadProgress> + IdGenerator>,
+) -> anyhow::Result<bool> {
+    // We collect all progress events that should be emitted if the blob/hashseq is complete.
+    // We do not emit them right away because we don't want to emit any events in case the
+    // blob/hashseq is not complete.
+    let mut progress_events = vec![];
+
+    // Check if the root blob is fully available.
+    let HashAndFormat { hash, format } = *hash_and_format;
+    let entry = match db.get(&hash).await? {
+        Some(entry) if entry.is_complete() => entry,
+        _ => return Ok(false),
+    };
+    progress_events.push(DownloadProgress::FoundLocal {
+        child: BlobId::Root,
+        hash,
+        size: entry.size(),
+        valid_ranges: RangeSpec::all(),
+    });
+
+    match format {
+        BlobFormat::Raw => {
+            // For a raw blob, we're done.
+        }
+        BlobFormat::HashSeq => {
+            // For a hashseq, check if all children are complete.
+            let reader = entry.data_reader().await?;
+            let (mut hash_seq, children) = parse_hash_seq(reader).await.map_err(|err| {
+                GetError::NoncompliantNode(anyhow!("Failed to parse downloaded HashSeq: {err}"))
+            })?;
+            progress_events.push(DownloadProgress::FoundHashSeq { hash, children });
+
+            let mut children: Vec<Hash> = vec![];
+            while let Some(hash) = hash_seq.next().await? {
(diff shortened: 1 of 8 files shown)


dignifiedquire:
lgtm, can't approve my own pr 😅

iroh/src/client/blobs.rs
```diff
@@ -1248,4 +1248,91 @@ mod tests {
 
         Ok(())
     }
+
+    /// Download a blob from oneself
+    #[tokio::test]
+    async fn test_blob_get_self() -> Result<()> {
```
matheus23, line 1254: FWIW, instead of all of the `.context()` calls, you can also just use `TestResult` instead of `anyhow::Result`.
This way you'll get the exact line that fails in the error that's thrown.
(It essentially behaves the same as if you did `.unwrap()` everywhere)

matheus23 (approved):
LGTM, but I have limited experience in this part of the code.

One note is that we should document the additions as breaking changes in the `Dialer` and `Getter` traits.

Frando:
As discussed in Discord, the previous fix was incorrect as it would report a succeeding download for collections if the root blob is complete even if we had no or only some of the children available locally.

I added two commits that fix this:
* We now check if a blob is complete (with all children for hashseqs) before starting a download. We still emit the expected FoundLocal events in that case.
* A test for downloading a collection which exists locally, succeeding without hitting the network

Also updated the PR description.

Please re-review :)

iroh/src/client/blobs.rs
Frando: Changed the new tests to that. Could also do a big pass over everything, but not today..

iroh-blobs/src/get/db.rs
```diff
@@ -59,6 +59,73 @@ pub async fn get_to_db<
...
+            // For a hashseq, check if all children are complete.
+            let reader = entry.data_reader().await?;
+            let (mut hash_seq, children) = parse_hash_seq(reader).await.map_err(|err| {
+                GetError::NoncompliantNode(anyhow!("Failed to parse downloaded HashSeq: {err}"))
+            })?;
+            progress_events.push(DownloadProgress::FoundHashSeq { hash, children });
+
+            let mut children: Vec<Hash> = vec![];
+            while let Some(hash) = hash_seq.next().await? {
+                children.push(hash);
```
dignifiedquire, line 106: why collect first, instead of directly doing the db checks?

iroh/src/client/blobs.rs
```diff
@@ -1248,4 +1251,185 @@ mod tests {
...
+                hash,
+                DownloadOptions {
+                    format: BlobFormat::HashSeq,
+                    nodes: vec![node_id.into()],
+                    tag: SetTagOption::Auto,
+                    mode: DownloadMode::Queued,
+                },
+            )
+            .await?
+            .await
```
dignifiedquire, line 1427: we should probably check the events in these tests, to make sure they look like we expect them, especially in the collection case

iroh-blobs/src/get/db.rs
Frando: Because with the PR as-is, we may only emit events here if everything is complete. Otherwise the same events would be emitted again when actually starting the download.

This is suboptimal for sure. See #2586 for a revised approach that does not need to collect events by using a generator and turning `get_to_db` into a two-step process.

dignifiedquire:
closing in favor of #2586

Issue #2546, opened 2024-07-25
NodeAddr with a direct address incurs an extra RTT


flub:
In #2509 we added an extra DISCO Ping-Pong exchange before we start using a direct address in all situations.  The reasoning was that this was a niche usecase which would not impact many real-life scenarios.

However it makes an iroh Endpoint no longer as good as a direct socket or QUIC endpoint.  If we have been given exactly one direct address we should start using it as well as using the relay path like before.  Otherwise we are significantly worse than a normal QUIC endpoint for this scenario and we do not want that.

If we do this, we may as well revert the entire decision and do the stable select again if there are multiple.  It seems a bit odd to now special-case this to just one direct address.

divagant-martian:
I find the explanation reasonable enough and clear enough to be ok with a special case

dignifiedquire:
you usually don't just get a single direct addr, most of the time it will be at least an IPv4 and IPv6 addr, at which point the special case would already fail to fix the issue though.

Frando:
If I see things correctly, #2509 made connection establishment with direct addresses *much* slower. The cost is not an additional RTT in practice, but ~1s in waiting time until quinn re-sends the initial packets, which are lost because no validated address is available yet.

Before 2509:
```rust
// addr is a NodeAddr with only direct addresses
endpoint.add_node_addr(addr);
endpoint.connect_by_node_id(addr.node_id).await?;
// works and is fast!
```
On current main

// addr is a NodeAddr with only direct addresses
endpoint.add_node_addr(addr);
endpoint.connect_by_node_id(addr.node_id).await?;
```
```
 2024-07-31T22:08:25.774202Z ERROR poll_send{me=bbg6osx53doxyits}: iroh_net::magicsock: failed to send: no UDP or relay addr node=mwy7btpljbjy2lux
2024-07-31T22:08:25.775316Z  INFO poll_recv{me=bbg6osx53doxyits}:disco_in{node=mwy7btpljbjy2lux src=Udp(192.168.178.142:42563)}:handle_pong{m=Pong { tx_id: TransactionId(0x07
2E6E08F0C626C7B908F8E1), ping_observed_addr: Udp(192.168.178.142:11204) } src=Udp(192.168.178.142:42563)}: iroh_net::magicsock::node_map::best_addr: selecting new direct path
 for node addr=192.168.178.142:42563 latency=1.548481ms trust_for=6.499989536s
[...]

PR #2579, opened 2024-08-02
ref(iroh-net): Remove need for relay info in best_addr

flub:
## Description

This moves metrics for when switches between direct connections and
relayed connections to the NodeState.  This is after all where this
decision is made.  BestAddr only knows about the best UDP address, it
has no business keeping track of relays.

Cleaning this up enables further refactoring of BestAddr and PathState
which need to improve because they depend on each other, but are
hindered by the relay state sneaking into there.

## Breaking Changes

None

## Notes & open questions

These metrics ignore the fact that we have mixed as a possibility so
over-simplify the situation.  This makes the logic to increment them
complex, which in turn makes it no one will really know what they
mean.

I need this to move on with #2546 so don't really want to get into
designing our metrics though.  I believe this way the keep the same
behaviour.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/src/magicsock/node_map/node_state.rs
+++ b/iroh-net/src/magicsock/node_map/node_state.rs
@@ -295,8 +295,8 @@ impl NodeState {
             (None, Some(relay_url)) => ConnectionType::Relay(relay_url),
             (None, None) => ConnectionType::None,
         };
-        if self.conn_type.update(typ).is_ok() {
-            let typ = self.conn_type.get();
+        if let Ok(prev_typ) = self.conn_type.update(typ.clone()) {
+            // The connection type has changed.
             event!(
                 target: "events.net.conn_type.changed",
                 Level::DEBUG,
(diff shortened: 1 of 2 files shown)


Arqu (approved):
Nice little state machine you got there :D

iroh-net/src/magicsock/node_map/node_state.rs
```diff
@@ -295,15 +295,39 @@ impl NodeState {
...
+                inc!(MagicsockMetrics, num_relay_conns_added);
+            }
+            if matches!(typ, ConnectionType::Direct(_) | ConnectionType::None)
+                && matches!(
+                    prev_typ,
+                    ConnectionType::Relay(_) | ConnectionType::Mixed(_, _)
+                )
+            {
+                inc!(MagicsockMetrics, num_relay_conns_removed);
+            }
```
matheus23, line 330: I wouldn't mind if this were refactored into a `match (prev_typ, typ)` expression.

flub: Be careful what you wish for...  This version is huge and contains a whole bunch of states that are unreachable.  Admittedly the `if` version has to be read backwards, where you think about what conditions are needed for a certain state.  So I guess the big match is less of a head-twister.

matheus23: I'd didn't necessarily want it to be an exhaustive match 😅 so yeah I guess I *have* to be careful in what I wish for haha!
Basically, having a last `_ => {}` case would've been fine by me :)

PR #2580, opened 2024-08-02
fix(iroh-net): Make a single direct address in NodeAddr instant

### LINKED ISSUE #2546: NodeAddr with a direct address incurs an extra RTT

flub:
## Description

In #2509 (c1c3539c0a07d8659979ffacdb5bd1fc23d6939c) we removed chosing
a random unconfirmed direct address and through both it and the relay
if we did not yet have any confirmed direct address.  This required
any direct address passed in via a NodeAddr to do a DISCO Ping-Pong
round trip before it would be used.

It turns out this is used a lot in the sense that a common socket
address is used: when you know an IP is reachable.  Requiring a ping
would mean the first QUIC packets would be dropped, Quinn would wait
to detect this before retrying and connections are delayed by an
observable second by the end of it all.  Not great for this scenario,
even if it can be argued that outside of tests this is not a common
scenario.

So this brings back the random selection of unconfirmed addresses.

It tries to do this without adding more complexity to the NodeState.
Instead it consolidates the path information and best address into the
NodeUdpState struct.  But **without any logic changes.**  It then adds
just this extra caching of the chosen address to this struct.

This is a huge simplification of a much larger refactor I was trying
to do first.  I believe that was the right track but resulted in too
many changes to do in one large PR.  So the long-term goal here is to
continue removing state manipulation outside of NodeUdpState struct so
that it gains more control of how to manage the UDP paths.  This will
help the logic of how UDP paths are chosen to be based entirely on the
PathState - which is the right model for this.

## Breaking Changes

None

## Notes & open questions

Fixed #2546

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/src/magicsock/node_map/node_state.rs
+++ b/iroh-net/src/magicsock/node_map/node_state.rs
@@ -270,9 +270,6 @@ impl NodeState {
             debug!("in `DEV_relay_ONLY` mode, giving the relay address as the only viable address for this endpoint");
             return (None, self.relay_url());
         }
-        // Update our best addr from candidate addresses (only if it is empty and if we have
-        // recent pongs).
-        self.assign_best_addr_from_candidates_if_empty();
         let (best_addr, relay_url) = match self.udp_paths.send_addr(*now, have_ipv6) {
             super::udp_paths::UdpSendAddr::Valid(addr) => {
                 // If we have a valid address we use it.
@@ -376,54 +373,6 @@ impl NodeState {
             reason,
             self.relay_url.is_some(),
         );
-    }
-
-    /// Fixup best_adrr from candidates.
-    ///
-    /// If somehow we end up in a state where we failed to set a best_addr, while we do have
-    /// valid candidates, this will chose a candidate and set best_addr again.  Most likely
-    /// this is a bug elsewhere though.
-    fn assign_best_addr_from_candidates_if_empty(&mut self) {
-        if !self.udp_paths.best_addr.is_empty() {
-            return;
-        }
-
-        // The highest acceptable latency for an endpoint path.  If the latency is higher
-        // then this the path will be ignored.
-        const MAX_LATENCY: Duration = Duration::from_secs(60 * 60);
-        let best_pong = self
-            .udp_paths
-            .paths
-            .iter()
-            .fold(None, |best_pong, (ipp, state)| {
-                let best_latency = best_pong
-                    .map(|p: &PongReply| p.latency)
-                    .unwrap_or(MAX_LATENCY);
-                match state.recent_pong() {
-                    // This pong is better if it has a lower latency, or if it has the same
-                    // latency but on an IPv6 path.
-                    Some(pong)
-                        if pong.latency < best_latency
-                            || (pong.latency == best_latency && ipp.ip().is_ipv6()) =>
-                    {
-                        Some(pong)
-                    }
-                    _ => best_pong,
-                }
-            });
-
-        // If we found a candidate, set to best addr
-        if let Some(pong) = best_pong {
-            if let SendAddr::Udp(addr) = pong.from {
-                warn!(%addr, "No best_addr was set, choose candidate with lowest latency");
-                self.udp_paths.best_addr.insert_if_better_or_reconfirm(
-                    addr,
-                    pong.latency,
-                    best_addr::Source::BestCandidate,
-                    pong.pong_at,
-                )
-            }
-        }
     }
 
     /// Whether we need to send another call-me-maybe to the endpoint.
--- a/iroh-net/src/magicsock/node_map/node_state.rs
+++ b/iroh-net/src/magicsock/node_map/node_state.rs
@@ -1,9 +1,7 @@
-use std::{
-    collections::{btree_map::Entry, BTreeMap, BTreeSet, HashMap},
-    hash::Hash,
-    net::{IpAddr, SocketAddr},
-    time::{Duration, Instant},
-};
+use std::collections::{btree_map::Entry, BTreeMap, BTreeSet, HashMap};
+use std::hash::Hash;
+use std::net::{IpAddr, SocketAddr};
+use std::time::{Duration, Instant};
 
 use iroh_metrics::inc;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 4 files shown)


dignifiedquire:
doc comments are sad

iroh-net/src/magicsock/node_map/node_state.rs
```diff
@@ -261,31 +261,34 @@ impl NodeState {
...
             debug!("in `DEV_relay_ONLY` mode, giving the relay address as the only viable address for this endpoint");
             return (None, self.relay_url());
         }
-        // Update our best addr from candidate addresses (only if it is empty and if we have
-        // recent pongs).
-        self.assign_best_addr_from_candidates_if_empty();
-        let (best_addr, relay_url) = match self.best_addr.state(*now) {
-            best_addr::State::Valid(best_addr) => {
+        let (best_addr, relay_url) = match self.udp_paths.send_addr(*now, have_ipv6) {
+            super::udp_paths::UdpSendAddr::Valid(addr) => {
```
dignifiedquire, line 274: probably nicer if you import this

iroh-net/src/magicsock/node_map/udp_paths.rs
```diff
@@ -0,0 +1,148 @@
...
+use tracing::warn;
+
+use crate::disco::SendAddr;
+
+use super::best_addr::{self, BestAddr};
+use super::node_state::{PathState, PongReply};
+use super::IpPort;
+
+#[derive(Debug)]
+pub(super) enum UdpSendAddr {
```
dignifiedquire, line 21: would be good to have some docs on this enum

iroh-net/src/magicsock/node_map/udp_paths.rs
```diff
@@ -0,0 +1,148 @@
...
+    /// it should be `&self` and not `&mut self`.  This is only possible once the state from
+    /// [`NodeUdpPaths`] is no longer modified from outside.
+    pub(super) fn send_addr(&mut self, now: Instant, have_ipv6: bool) -> UdpSendAddr {
+        self.assign_best_addr_from_candidates_if_empty();
+        match self.best_addr.state(now) {
+            best_addr::State::Valid(addr) => UdpSendAddr::Valid(addr.addr),
+            best_addr::State::Outdated(addr) => UdpSendAddr::Outdated(addr.addr),
+            best_addr::State::Empty => {
+                // No direct connection has been used before.  If we know of any possible
+                // candidate addresses, randomlly try to use one.  This path is most
```
dignifiedquire, line 75: typo

iroh-net/src/magicsock/node_map/udp_paths.rs
```diff
@@ -0,0 +1,148 @@
...
+                        // Look for a new candidate in all the known paths.  This may look
+                        // like a RNG use on the hot-path but this is normally invoked at
+                        // most most once at startup.
+                        let addr = self
+                            .paths
+                            .values()
+                            .filter_map(|path| path.udp_addr())
+                            .filter(|addr| addr.is_ipv4() || have_ipv6)
+                            .choose(&mut rand::thread_rng());
+                        self.chosen_candidate = addr.as_ref().map(|addr| IpPort::from(*addr));
```
dignifiedquire, line 94: I think `addr.map(IpPort::from)`  should work

iroh-net/src/magicsock/node_map/udp_paths.rs
```diff
@@ -0,0 +1,148 @@
...
+                    });
+                match addr {
+                    Some(addr) => UdpSendAddr::Unconfirmed(addr),
+                    None => UdpSendAddr::None,
+                }
+            }
+        }
+    }
+
+    /// Fixup best_adrr from candidates.
```
dignifiedquire, line 105: typo

iroh-net/src/magicsock/node_map/udp_paths.rs
```diff
@@ -0,0 +1,148 @@
...
+    ///
+    /// If somehow we end up in a state where we failed to set a best_addr, while we do have
+    /// valid candidates, this will chose a candidate and set best_addr again.  Most likely
+    /// this is a bug elsewhere though.
```
dignifiedquire, line 109: if this is a bug, shouldn't we log that we are doing this, to figure out why this happens?

iroh-net/src/magicsock/node_map/udp_paths.rs
```diff
@@ -0,0 +1,148 @@
...
+    fn assign_best_addr_from_candidates_if_empty(&mut self) {
+        if !self.best_addr.is_empty() {
+            return;
+        }
+
+        // The highest acceptable latency for an endpoint path.  If the latency is higher
+        // then this the path will be ignored.
+        const MAX_LATENCY: Duration = Duration::from_secs(60 * 60);
```
dignifiedquire, line 117: this seems rather pointless.. latency of 1 hour?

iroh-net/src/magicsock/node_map/udp_paths.rs
flub: It is already logged on `warn` for this purpose.

iroh-net/src/magicsock/node_map/udp_paths.rs
flub: Yes, I don't really know how this is useful either.  Though this code is copy-pasted and I'd really like to avoid doing unrelated changes here.  I do want to carry on fixing up things here though, but in small PRs that each target one thing because my attempt to do a bunch of this together became messy and un-reviewable.

flub:
fwiw test that rely on this functionality now run in 200ms again instead of being 1.2s

dignifiedquire (approved):

PR #2583, opened 2024-08-02
feat: allow custom blob providing event handling

dignifiedquire:
## Description

This adds the ability back to track blob provide events from an iroh node.

## Breaking Changes

- Modifies `Event` enum:
  - removes CustomGetRequestReceived
  - adds TransferProgress
- send_blob now takes an EventSender
- trait `EventSender` has been renamed `CustomEventSender`
- the concrete (boxed) event sender is now called just `EventSender`

## Notes & open questions

Notes:

I added progress events that get sent every single 16 KiB chunk. These events contain the usual data (connection id and request id) plus offset and hash. Adding the hash makes the event larger, but makes it very much self-contained.

The event is fired after the read, but that does not mean that the data has been successfully transferred. Should be good enough to show progress anyway.

These will be a lot of events, so they are sent using try_send in order not to slow down the transfer. That means that you might miss some events, but since they only contain the end_offset that should be fine.

Open questions:

I called the trait CustomEventSender and the concrete handle for an optional CustomEventSender just `EventSender`. I can revert this if somebody can think of a better name for the trait or the handle...

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh/src/node/builder.rs
```diff
@@ -239,19 +244,52 @@ impl<D: Map> Builder<D> {
...
             gc_done_callback: None,
+            blob_events: MockEventSender,
+        }
+    }
+}
+
+impl<D, E> Builder<D, E>
+where
+    D: BaoStore,
+    E: iroh_blobs::provider::EventSender,
```
rklaehn, line 255: I think the EventSender ist not important enough to warrant its own type parameter. Maybe we can make this an `Arc<dyn` or `Box<dyn` thing.

dignifiedquire (commented):
lgtm

rklaehn (approved):

PR #2585, opened 2024-08-05
ref(iroh-net): Don't write the match as fully exhaustive

flub:
## Description

This makes it a bit earier to read.  It's harder to figure out if thre
are changes to the ConnectionType though, but let's assume that
doesn't happen too often.

## Breaking Changes

None

## Notes & open questions

I'm open to other ways of structuring this code.  Not entirely sure
what the best shape would be.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/src/magicsock/node_map/node_state.rs
+++ b/iroh-net/src/magicsock/node_map/node_state.rs
@@ -307,46 +307,31 @@ impl NodeState {
 
             // Update some metrics
             match (prev_typ, typ) {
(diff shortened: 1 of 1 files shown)


Arqu (approved):

matheus23 (approved):
Yes, exactly what I was thinking of originally.
The main goal being: the exact effects of a transfer from connection type A to B on the metrics is easy to read off, similar to looking it up in a table.
I never intended the goal to be exhaustiveness checking & causing a compiler error in case `ConnectionType` gets modified.

Issue #2577, opened 2024-08-02
iroh-blobs logs are being swalloed


dignifiedquire:
there is no logging output anymore even on `TRACE` level from `iroh-blobs`

flub:
`RUST_LOG=iroh_blobs=trace cargo test -p iroh --lib -- test_blob_get_self --show-output` produces no logs on [fix-download-self](https://github.com/n0-computer/iroh/tree/fix-download-self) branch

PR #2589, opened 2024-08-05
bug: `LocalPool` swallowing logs in `iroh-blobs` tests

ramfox:
## Description

While investigating the flaky CI failure of the `localpool::test_drop` test, I found an example of `LocalPool` eating logging.

The logs inside the `test_drop` print, but any logs written in `LocalPool::spawn_detached` are lost.

run `cargo test --package iroh-blobs test_drop -- --nocapture` to view.

I also had to change from using `tracing_subscriber::fmt::try_init()` to `iroh_test::logging::setup()` in the test to get ANY logging to show.

Related to #2577

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-blobs/src/util/local_pool.rs
+++ b/iroh-blobs/src/util/local_pool.rs
@@ -14,13 +14,15 @@ use tokio::{
     sync::{Notify, Semaphore},
     task::{JoinError, JoinSet, LocalSet},
 };
+use tracing::trace;
 
 type BoxedFut<T = ()> = Pin<Box<dyn Future<Output = T>>>;
 type SpawnFn<T = ()> = Box<dyn FnOnce() -> BoxedFut<T> + Send + 'static>;
 
+#[derive(derive_more::Debug)]
 enum Message {
     /// Create a new task and execute it locally
-    Execute(SpawnFn),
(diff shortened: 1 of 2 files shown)


iroh-blobs/src/util/local_pool.rs
```diff
@@ -568,6 +572,7 @@ mod tests {
         tokio::time::sleep(Duration::from_millis(100)).await;
         // drop x at the end. we will never get here when the future is
         // no longer polled, but drop should still be called
+        tracing::debug!("Dropping {x:?}");
```
ramfox, line 575: @dignifiedquire should this line still be caught in logging? It's not currently.

flub:
Still looking into this.  FWIW the `iroh_test::logging::setup` has never worked with tests using multiple threads, so that's not really any different with the LocalPool.

The problem with using a global subscriber is that all tests use the same subscriber and IIRC your tests output starts flowing into the wrong tests due to how test isolation and output capturing works.

flub:
Oops, I thought this was an issue and it's a PR...  I replaced this in #2735

flub:
The other PR did get merged

Issue #2273, opened 2024-05-08
upgrade to quinn@0.11 & rustls@0.23


dignifiedquire:
WIP branch: https://github.com/n0-computer/iroh/tree/dig/quinn11

- [x] The `dig/quinn11` branch needs to work and pass all tests
  - [x] On patched dependencies
  - [x] On released dependencies
- [x] We need an **iroh-quinn** version released for 0.11
- [x] We need a new **quic-rpc** version released
- [x] We need a new **axum-server** version released
- [x] We need a new **tokio-rustls-acme** release
- [x] We should upgrade **request** @0.12.5 so it uses rustls@0.23

Once the basics work, we need to make it good:
- [x] get net performance numbers
  - we have declared per to be ok

Nice to have:
- [x] hickory-resolver upgrade to rustls@0.23
  - We are using `=0.25.0-alpha.2` 
  - https://github.com/hickory-dns/hickory-dns/issues/2206

Outstanding issues:
- [ ] When sending to the relay all packets are copied.
  - AsyncUdpSocket @0.10 gave us `Bytes`, but @0.11 gives us a `&[u8]` and we end up copying this to send it along channels to the relay actors.  That's a severe performance regression
- [x] Review all newly added TODO items

matheus23:
Can I help move this along somehow?
For browser work, it'd be great to have `rustls` upgraded to version 0.22 or above, because at that version it doesn't depend on `std::time::Instant::now` anymore, which panics in Wasm.

flub:
> Can I help move this along somehow? For browser work, it'd be great to have `rustls` upgraded to version 0.22 or above, because at that version it doesn't depend on `std::time::Instant::now` anymore, which panics in Wasm.

It's currently in the mode of making the test suite pass.  All "changes" should be done and it is "just" debugging.  So the work right now is: run test suite, pick most promising failing test, debug it, push fix, run test suite again to see if more failures are magically fixed, pick next test.

But most test failures have just been fixes in the test suite, so progress has been slow.

ramfox:
Closed by #2595

PR #2595, opened 2024-08-06
feat(iroh-net)!: upgrade to Quinn 0.11 and Rustls 0.23

flub:
## Description

Upgrades the Quinn and Rustls dependencies.  This touches a lot of dependencies and affects a lot of the internal API due to the changes in Quinn.

## Breaking Changes

- `iroh::net::endpoint::Endpoint::accept` now returns `Incoming` instead of `Connecting`. This allows rejecting incoming connections earlier, notify the initiating side to retry later and more. To get back `Connecting`, simply use `Incoming::accept()`. If all you did with `Connecting` before was to `.await` it, you can do the same with `Incoming`, too, and get back a `Connection`.
- Removed `iroh::net::endpoint::Builder::concurrent_connections`. The amount of concurrent connections is now controlled by either calling `Incoming::accept` or `Incoming::refuse` after `Endpoint::accept`.

## Notes & open questions

See #2273 for the tracking issue coordinating all this work.  This is the last piece of the puzzle.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


matheus23:
When I try to run `sendme send ...` with a dependency on this branch, I get:
```
thread 'main' panicked at /home/philipp/program/work/iroh-quinn11/iroh-net/src/tls.rs:77:6:
Server cert key DER is valid; qed: InvalidCertificate(Other(OtherError(UnsupportedCriticalExtension)))
```

Seems to be related to the libp2p extension perhaps? I'm investigating.

matheus23:
## Information on the `windows` test failures

### Quinn `UdpSocket` updates on windows

Part of the updates between quinn 0.10 and quinn 0.11 are a bunch of new features to their socket impl: https://github.com/quinn-rs/quinn/pull/1701
Previously, the `quinn::Transmit::src_ip: Option<IpAddr>` field was simply ignored, now it's actually set via `IP_PKTINFO` socket options.
Now, in windows it's apparently an error when `src_ip: Some(0.0.0.0)`, or any other unspecified IP address, since that doesn't help windows figure out which IP interface to use. It's also an error when the `src_ip` is Ipv6, but the `destination` is an Ipv4 socket address (and I assume the other way around, too? not sure). In the error msg above you can see both problems.

### So, where does the `src_ip` come from?

It's actually only set on the *server side*, or in our case, whatever the side ends up being that calls `accept`, instead of `connect`.
The mechanism here is, once you `accept`, you've received a packet, and that packet *usually* contains the IP address that it arrived at (the `dst_ip`), and from that quinn determines "oh that's now the interface we should use to answer", which makes sense.
However, in `magicsock`, we *do* overwrite the `dst_ip` when we receive packets, namely with `self.normalized_local_addr`, which unfortunately resolves to `::` or `0.0.0.0`.
This `dst_ip` then gets passed back to quinn, wihch uses it in the `src_ip` for transmits & it infects the `try_send` calls quinn makes, causing us to call the socket badly.

### What do we do about it?

Not sure. I'd love people to tell me ideas!
The only thing that I think reliably made the test suite pass so far is always setting `dst_ip: None`.
What *hasn't* worked, and I find that quite surprising is not overwriting the `dst_ip` anymore, and set it to `None` when the packet is "received" from the relay. (Perhaps worth diagnosing this further?)
Other ideas we had:
- Set `src_ip` in the `Transmit`s to results from netcheck
- Use a "fake IP address" in relay receives, so that, conceptually, the "relay" is just another "IP interface" that we recieve on (just a fake one).

### This PR

For now, this PR does this:
```rust
#[cfg(not(windows))]
let dst_ip = self.normalized_local_addr().ok().map(|addr| addr.ip());
#[cfg(windows)]
let dst_ip = None;
```
Which seems to pass on Linux, Mac and Windows.

matheus23:
Information on the recent `gossip_net_smoke` flakyness: https://github.com/quinn-rs/quinn/issues/1974

iroh-net/src/endpoint.rs
```diff
@@ -291,21 +280,17 @@ struct StaticConfig {
...
     transport_config: Arc<quinn::TransportConfig>,
     keylog: bool,
-    concurrent_connections: Option<u32>,
 }
 
 impl StaticConfig {
     /// Create a [`quinn::ServerConfig`] with the specified ALPN protocols.
     fn create_server_config(&self, alpn_protocols: Vec<Vec<u8>>) -> Result<quinn::ServerConfig> {
-        let mut server_config = make_server_config(
+        let server_config = make_server_config(
```
matheus23, line 288: `create_server_config` and `make_server_config` are now practically equivalent, but I want to keep `make_server_config`, so it doesn't change the public API, and I want to keep `StaticConfig::create_server_config` because it's the better internal API.

iroh-net/src/tls/certificate.rs
```diff
@@ -26,6 +28,45 @@ const P2P_SIGNING_PREFIX: [u8; 21] = *b"libp2p-tls-handshake:";
 // Similarly, hash functions with an output length less than 256 bits MUST NOT be used.
 static P2P_SIGNATURE_ALGORITHM: &rcgen::SignatureAlgorithm = &rcgen::PKCS_ECDSA_P256_SHA256;
 
+#[derive(Debug)]
+pub(crate) struct AlwaysResolvesCert(Arc<rustls::sign::CertifiedKey>);
```
matheus23, line 32: This stuff is copied from https://github.com/libp2p/rust-libp2p/pull/5498 (for anyone wondering about this in code review)

iroh-net/src/magicsock.rs
```diff
@@ -993,24 +994,30 @@ impl MagicSock {
...
+    ///
+    /// Note that on failure the (remaining) ping actions are simply dropped.  That's bad!
+    /// The Endpoint will think a full ping was done and not request a new full-ping for a
+    /// while.  We should probably be buffering the pings.
+    fn try_send_ping_actions(&self, msgs: Vec<PingAction>) -> io::Result<()> {
+        for msg in msgs {
+            // Abort sending as soon as we know we are shutting down.
+            if self.is_closing() || self.is_closed() {
+                return Ok(());
+            }
```
matheus23, line 1007: Is this here to prevent cases where we `try_send_ping` even though `self.is_closing()`?
Are these cases bad?
And if this is here to prevent that, then I don't think it's doing that, as the endpoint closing & this running could still race.

Is this perhaps meant to be outside the `for` loop?
Or can `try_send_ping` cause `self.is_closing()`?

In any case I'm a little confused by this.

iroh-net/src/magicsock.rs
```diff
@@ -1690,6 +1693,69 @@ impl AsyncUdpSocket for Handle {
...
+        }
+    }
+
+    fn max_receive_segments(&self) -> usize {
+        if let Some(pconn6) = self.pconn6.as_ref() {
+            // TODO(matheus23): This is somewhat incorrect. Needs updates to `MagicSock::poll_recv` to fix this though.
+            std::cmp::min(
+                pconn6.max_receive_segments(),
+                self.pconn4.max_receive_segments(),
+            )
```
matheus23, line 1714: Ah, my bad. I didn't push a fix to this yet, I think it should be like this:
```suggestion
            // `max_receive_segments` controls the size of the `RecvMeta` buffer
            // that quinn creates. Having buffers slightly bigger than necessary
            // isn't terrible, and makes sure a single socket can read the maximum
            // amount with a single poll. We considered adding these numbers instead,
            // but we never get data from both sockets at the same time in `poll_recv`
            // and it's impossible and unnecessary to be refactored that way.
            std::cmp::max(
                pconn6.max_receive_segments(),
                self.pconn4.max_receive_segments(),
            )
```

b5:
don't we need a `!` on the PR description?

iroh-net/src/magicsock.rs
```diff
@@ -1690,6 +1693,69 @@ impl AsyncUdpSocket for Handle {
             (_, Some(ipv6)) => Ok(*ipv6),
         }
     }
+
+    fn max_transmit_segments(&self) -> usize {
+        if let Some(pconn6) = self.pconn6.as_ref() {
+            std::cmp::max(
+                pconn6.max_transmit_segments(),
+                self.pconn4.max_transmit_segments(),
+            )
```
matheus23, line 1702: Somewhere, some change I wanted to make got lost. I think in *theory* this should be `min`, not `max`.
In the sense of "this is how many transmit segments at a time the socket can handle". And if we use `max` here, we may bombard a socket with more segments that it can handle.

That said, in practice the value for `max_transmit_segments` doesn't change between IPV4 and IPV6, and it's either 512 or 1 on windows, and either 64 or 1 on unix.

```suggestion
            std::cmp::min(
                pconn6.max_transmit_segments(),
                self.pconn4.max_transmit_segments(),
            )
```

iroh-net/src/magicsock.rs
```diff
@@ -3544,4 +3608,276 @@ mod tests {
...
+            magicsock_connect(&ep_1, secret_key_1.clone(), addr, node_id_2),
+        )
+        .await
+        .expect("timeout while connecting");
+
+        // aka assert!(res.is_ok()) but with nicer error reporting.
+        res.unwrap();
+
+        // TODO: Now check if we can connect to a repaired ep_3, but we can't modify that
+        // much internal state for now.
```
matheus23, line 3776: Wdym "repaired `ep_3`"?
Isn't `ep_2` non-broken?
I guess when the bug existed, this would've made sense, but now that the test is fixed, maybe we can just remove this TODO?

matheus23 (commented):
Sorry, I guess at some point I didn't commit some changes I wanted to make (re: `max_transmit_segments` and `max_receive_segments`).

iroh-net/src/magicsock.rs
flub: oh, this was meant to repair what is behind `bad_addr`.  But that doesn't even have a NodeId right now.   Yeah, let's remove this.

iroh-net/src/magicsock.rs
flub: this is here because... history?  maybe from the go code still?  Maybe it was a bug fix early on?  I actually suspect the last one.

I don't care too much about this kind of check, it appears in a few places.  Ultimately I think the architecture should be robust enough to deal without this.  But as long as we have a "closed" bool in the state we'll have this kind of thing.  I think it's fine to leave in.

iroh-net/src/endpoint.rs
flub: fwiw this `make_server_config` is used by the `Node` builder as well to make the RPC endpoint.  I'm not sure you can completely remove it.

flub: oh, that is actually the reason this is pub!  because it's another crate :(

iroh-net/src/magicsock.rs
matheus23: Okay, let's leave it :)

matheus23 (approved):

PR #2597, opened 2024-08-06
fix(iroh-gossip): clarify docs and semantics of gossip joined event

Frando:
## Description

Improves documentation around the `GossipEvent::Joined` event: It is only emitted once at the beginning of the stream, and the event will not be emitted when awaiting `GossipReceiver::joined`.

Also makes sure that the event is actually only emitted once per intent (it potentially could have been emitted multiple times before if the neighbor count first got down to 0 and then up again for `GossipTopics` subscribing inbetween).

## Notes & open questions

Inspired by the discussion in https://github.com/deltachat/deltachat-core-rust/pull/5860

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-gossip/src/net/handles.rs
+++ b/iroh-gossip/src/net/handles.rs
@@ -8,7 +8,7 @@ use std::{
     task::{Context, Poll},
 };
 
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, Context as _, Result};
 use bytes::Bytes;
 use futures_lite::{Stream, StreamExt};
 use iroh_net::NodeId;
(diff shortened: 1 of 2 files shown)


iroh-gossip/src/net/handles.rs
```diff
@@ -128,9 +130,22 @@ impl GossipReceiver {
...
     pub async fn joined(&mut self) -> Result<()> {
-        while self.neighbors.is_empty() {
-            let _ = self.try_next().await?;
+        if !self.joined {
+            match self
+                .try_next()
+                .await?
+                .context("Gossip receiver closed before Joined event was received.")?
+            {
+                Event::Gossip(GossipEvent::Joined(_)) => {}
```
dignifiedquire, line 146: why are you throwing the initial neighbours here away?

Frando: They are added to `self.neighbors` in `poll_next` already.

dignifiedquire (approved):

PR #2603, opened 2024-08-08
DO NOT MERGE: Demonstration of how to badly close a QUIC connection

flub:
This is an attempt at demonstrating how relying on the ACKs from QUIC does not sufficiently ensure all application data is delivered.

Notice that because we wait until stopped most of the time this happens to work with Quinn.

### DIFF
--- a/iroh-net/src/endpoint.rs
+++ b/iroh-net/src/endpoint.rs
@@ -1683,4 +1683,81 @@ mod tests {
         r1.expect("ep1 timeout").unwrap();
         r2.expect("ep2 timeout").unwrap();
     }
+
+    #[tokio::test]
+    async fn test_proto_bad_close() {
+        let _logging_guard = iroh_test::logging::setup();
+        let ep1 = Endpoint::builder()
+            .alpns(vec![TEST_ALPN.to_vec()])
+            .relay_mode(RelayMode::Disabled)
+            .bind(0)
+            .await
+            .unwrap();
+        let ep2 = Endpoint::builder()
+            .alpns(vec![TEST_ALPN.to_vec()])
+            .relay_mode(RelayMode::Disabled)
+            .bind(0)
+            .await
+            .unwrap();
+        let ep1_nodeaddr = ep1.node_addr().await.unwrap();
+        let ep2_nodeaddr = ep2.node_addr().await.unwrap();
+        ep1.add_node_addr(ep2_nodeaddr.clone()).unwrap();
+        ep2.add_node_addr(ep1_nodeaddr.clone()).unwrap();
+        let ep1_nodeid = ep1.node_id();
+        let ep2_nodeid = ep2.node_id();
+        eprintln!("node id 1 {ep1_nodeid}");
+        eprintln!("node id 2 {ep2_nodeid}");
+        let start = Instant::now();
+
+        async fn connect_hello(ep: Endpoint, dst: NodeAddr, start: Instant) {
+            let conn = ep.connect(dst, TEST_ALPN).await.unwrap();
+            let (mut send, mut recv) = conn.open_bi().await.unwrap();
+            send.write_all(b"hello").await.unwrap();
+            send.finish().unwrap();
+
+            let mut buf = [0u8; 5];
+            recv.read_exact(&mut buf).await.unwrap();
+
+            // Waiting here means we make sure we have received the CONNECTION_CLOSE frame
+            // before continuing to read.
+            tokio::time::sleep(Duration::from_secs(3)).await;
+
+            // Reading now we get the ApplicationClosed error instead of the FIN.
+            eprintln!("{:?} 2. reading to end", start.elapsed());
+            recv.read_to_end(0).await.unwrap();
+            conn.close(1u8.into(), b"done"); // never reached
+        }
+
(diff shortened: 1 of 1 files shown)


iroh-net/src/endpoint.rs
```diff
@@ -1683,4 +1683,81 @@ mod tests {
...
+            send.write_all(b"world").await.unwrap();
+            send.finish().unwrap();
+
+            // Using this line makes it work.  This is because Quinn is less strict than
+            // RFC9000 requires: it will still deliver acknowledged stream data to the
+            // application after a CONNECTION_CLOSE frame has been received.  Only returning
+            // the close error once all this data has been read.  This is why all our
+            // current protocols seem to work in practice.
+
+            // send.stopped().await.unwrap();
```
matheus23, line 1746: Ah okay, so to verify my understanding: The theoretical problem is that the `CONNECTION_CLOSE` is a totally asynchronous event that isn't "in-order" with the streams at all, it sits on another level, interrupting all streams that happen currently.

flub: indeed

iroh-net/src/endpoint.rs
```diff
@@ -1683,4 +1683,81 @@ mod tests {
...
+
+            // This was sent and delivered ages ago, so is readily available and does not
+            // slow us down.
+            recv.read_to_end(0).await.unwrap();
+
+            // We close this before the FIN has been read by the peer.
+            eprintln!("{:?} 1. closing connection", start.elapsed());
+            conn.close(2u8.into(), b"boom");
+        }
```
matheus23, line 1755: This is awesome! Thank you so much.
So re-reading your blog post together with this test case really helped me I think.
When I was reworking this example to fix it - sort of as an exercise - I came up with a verison that's really close to the one you actually describe in the blog post, except for the `wait_idle()`, which is actually a really good point.

So this doesn't rely on protocol-level closing frames or anything like that. Also no `.finish()` or `.stopped().await` needed at all. Because it's a simple request-response, we just *know* what role in the protocol we have & when we're done, so we can just `.close()` and wait for `closed()` in the right places, right?

```suggestion
        async fn connect_hello(ep: Endpoint, dst: NodeAddr, start: Instant) {
            let conn = ep.connect(dst, TEST_ALPN).await.unwrap();
            let (mut send, mut recv) = conn.open_bi().await.unwrap();
            send.write_all(b"hello").await.unwrap();

            let mut buf = [0u8; 5];
            recv.read_exact(&mut buf).await.unwrap();

            // We close this
            eprintln!("{:?} 1. closing connection", start.elapsed());
            conn.close(1u8.into(), b"done");
            // We make sure the close frame is resent in case it got lost
            ep.endpoint().wait_idle().await;
        }

        async fn accept_world(ep: Endpoint, start: Instant) {
            let incoming = ep.accept().await.unwrap();
            let conn = incoming.await.unwrap();
            let (mut send, mut recv) = conn.accept_bi().await.unwrap();
            let mut buf = [0u8; 5];
            recv.read_exact(&mut buf).await.unwrap();
            send.write_all(b"world").await.unwrap();

            let close = conn.closed().await;
            eprintln!("{:?} closed: {close:?}", start.elapsed());
        }
```

matheus23: Ugh, this diff sucks...

matheus23: And if I want the same behaviour like my `.close(); ep.wait_idle().await` and `.closed().await`, but on the stream-level (maybe because I want to keep the connection open & support multiple request-response things on the same connection), then I can use `.stop()` and `.stopped()`, right? I just need to make sure I have my protocol roles in check.

flub: > So this doesn't rely on protocol-level closing frames or anything like that. Also no `.finish()` or `.stopped().await` needed at all. Because it's a simple request-response, we just _know_ what role in the protocol we have & when we're done, so we can just `.close()` and wait for `closed()` in the right places, right?

indeed.

> And if I want the same behaviour like my .close(); ep.wait_idle().await and .closed().await, but on the stream-level (maybe because I want to keep the connection open & support multiple request-response things on the same connection), then I can use .stop() and .stopped(), right? I just need to make sure I have my protocol roles in check.

You can care about a single stream, but ultimately your entire protocol has to coordinate connection-level shut down at some point, so needs to know your streams are all done.

`RecvStream::stop()` is mainly used when you are no longer interested in a stream.  Say you're streaming a large file with BAO from two sources, some kind of simple bittorrent-like downloading, one source completes so you stop the other one.

`SendSTream::stopped()` is not that useful I think.  Usually you'll get the stopped error from a write operation.  And knowing when your sent data have been ACKed by the peer turns out not to add much to the party.

In Quinn dropping a `SendStream` does an implicit `SendStream::finish()`, dropping a `RecvStream` does an implicit `RecvStream::stop(0)`.  Both of these are mostly what you expect.  So when you're done with those streams a task can just exit and not care too much about dropping.  It is only dropping the `Connection` that has surprising side-effects of maybe losing in-transit data.

matheus23: This is great info, thanks!

matheus23:
We can probably close this, right?

PR #2605, opened 2024-08-08
feat(iroh-net): upgrade to new `swarm-discovery` api

ramfox:
## Description

The new `swarm-discovery` api allows you to remove and add addresses you want to publish, without needing to restart the swarm-discovery service.

There is a chance that this process (restarting the swarm-discovery service) caused flaky-ness in our `test_local_swarm_discovery` test.

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh-net/src/discovery/local_swarm_discovery.rs
+++ b/iroh-net/src/discovery/local_swarm_discovery.rs
@@ -278,10 +278,10 @@ mod tests {
     #[tokio::test]
     async fn test_local_swarm_discovery() -> TestResult {
         let _guard = iroh_test::logging::setup();
-        let (node_id_a, discovery_a) = make_discoverer()?;
-        let (_, discovery_b) = make_discoverer()?;
+        let (_, discovery_a) = make_discoverer()?;
+        let (node_id_b, discovery_b) = make_discoverer()?;
 
-        // make addr info for discoverer a
+        // make addr info for discoverer b
         let addr_info = AddrInfo {
             relay_url: None,
             direct_addresses: BTreeSet::from(["0.0.0.0:11111".parse()?]),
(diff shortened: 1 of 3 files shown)


divagant-martian:
looking into failing logs, we do get a restarted service, so it's likely this is the reason. From what I can tell, restarting means the sent request timeouts because it's actually lost (the previous instance had it)

iroh-net/src/discovery/local_swarm_discovery.rs
```diff
@@ -213,38 +204,27 @@ impl LocalSwarmDiscovery {
...
-            (false, true) => IpClass::V6Only,
-            // this case indicates no ip addresses were supplied, in which case, default to ipv4
-            (false, false) => IpClass::V4Only,
-        };
+        let addrs = LocalSwarmDiscovery::socketaddrs_to_addrs(socketaddrs);
         let mut discoverer =
             Discoverer::new_interactive(N0_LOCAL_SWARM.to_string(), node_id.to_string())
                 .with_callback(callback)
-                .with_ip_class(ip_class);
+                .with_ip_class(IpClass::Auto);
```
divagant-martian, line 211: not sure I like this change too much, since what this does underneath is try to bind to a port which is later dropped. Sometimes this ends up with absurd race conditions about the port being taken. I'm going to go ahead with trusting this, but just leaving this here for the record I guess

divagant-martian (approved):
LGTM

iroh-net/src/discovery/local_swarm_discovery.rs
rkuhn: Not sure whether there may be a misunderstanding: the code tries binding to both ports and uses whatever worked — no successes are dropped. Is there a race condition in this that I’m unaware of?

PR #2607, opened 2024-08-09
test(iroh-cli): Replace `cli_provide_one_file_large` with a faster test

matheus23:
## Description

The `cli_provide_one_file_large` test takes multiple minutes to complete.

We talked about removing it altogether in the discord.
The reasoning for removing it is that we already exercise big file provide & get loops via netsim. So there is no need to run the same code in debug mode taking >3 minutes in CI.

However, I ended up not removing the test but instead changing it, because the other tests test really small file sizes, e.g. `1000` bytes. At those sizes, outboards aren't even generated and files are put inline into the database.
So instead I worked out the file size required for exercising the whole external-outboard (and external-file) machinery of iroh-blobs.
The cutoff point is at ~4.1MB, which is really far away from the 1GB size and should transfer within seconds.

## Breaking Changes

None

## Notes & open questions

Should we just delete the test altogether?

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-cli/tests/cli.rs
+++ b/iroh-cli/tests/cli.rs
@@ -50,10 +50,20 @@ fn cli_provide_one_file_basic() -> Result<()> {
 }
 
 #[test]
-fn cli_provide_one_file_large() -> Result<()> {
+fn cli_provide_one_file_external_outboard() -> Result<()> {
     let dir = testdir!();
     let path = dir.join("foo");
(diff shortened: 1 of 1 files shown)


Arqu (approved):
Love it, keeps the spirit of the test without the CI time suck.

flub (approved):
good catch!  nice to still have this tested

PR #2610, opened 2024-08-09
refactor(iroh,iroh-net)!: prefer `remote` to `connection` in api

flub:
## Description

Moves some functions and types to use `remote` to describe the information we know about a node,
instead of connections. This, based on the fact that iroh-net does not itself handle connections,
making the wording misleading at best.

### iroh
- `client::node::Client::connection_info` -> `client::node::Client::remote_info`
- `client::node::Client::connections` -> `client::node::Client::remote_infos_iter`

### iroh-net
- `endpoint::ConnectionInfo` -> `endpoint::RemoteInfo`
- `endpoint::ConnectionInfo::id` is removed since it's internal information.
- `endpoint::ConnectionInfo::last_alive_relay` is deprecated. Use `.relay_url.last_alive`
- `endpoint::Endpoint::connection_info` -> `endpoint::Endpoint::remote_info`
- `endpoint::Endpoint::connection_infos` -> `endpoint::Endpoint::remote_infos_iter`

### iroh-cli (bin)
- `node connection-info` -> `node remote-info`
- `node connections` -> `node remote-list`

## Notes & open questions

1. We decided to also move many of the (currently) node commands that come directly from iroh-net to
   a new `net` top level api. Doing so we should aim as well to move the cli to the more gracious
   `remote <node-id>` and `remote list`. In summary:
  - we have `iroh node remote <node-id>` and we want `iroh net remote <remote-id>`
  - we have `iroh node remote-list` and we want `iroh net remote list`
  Issue #2639 is created for this
2. This adds a deprecation that needs to be removed later. Issue #2640 is created for this.

## Change checklist

- [x] Self-review.
- [ ] ~~Tests if relevant.~~


flub:
@dignifiedquire while this doesn't work yet could I get some feedback on using deprecations?

Also on the rename itself from `Endpoint::connection_info[s]` to `Endpoint::node_info[_iter]`.  /cc @ramfox you added this originally.

dignifiedquire:
given we haven't used deprecations for the last rounds of api changes, I don't think they are necessary currently

iroh-cli/src/commands/node.rs
```diff
@@ -196,6 +195,7 @@ fn fmt_connection(info: ConnectionInfo) -> String {
 }
 
 fn direct_addr_row(info: DirectAddrInfo) -> comfy_table::Row {
+    #[allow(deprecated)]
```
dignifiedquire, line 198: still needed?

iroh-net/src/magicsock.rs
```diff
@@ -286,18 +288,18 @@ impl MagicSock {
...
+        self.node_info(node_key)
             .map(|info| info.has_send_address())
             .unwrap_or(false)
     }
 
-    /// Retrieve connection information about nodes in the network.
-    pub(crate) fn connection_infos(&self) -> Vec<ConnectionInfo> {
+    /// Return all the [`NodeInfo`]s of all nodes in the node map.
+    pub(crate) fn node_info_all_nodes(&self) -> Vec<NodeInfo> {
         self.node_map.node_infos(Instant::now())
```
dignifiedquire, line 298: it's a bit confusing that we use three names now for this

- `node_infos` in the `node_map`
- `node_info_all_nodes` here
- and `node_info_iter`

iroh-net/src/magicsock/node_map/node_state.rs
```diff
@@ -1227,64 +1271,85 @@ impl From<RelayUrlInfo> for RelayUrl {
...
+    /// payload or control messages.  Note that sending to the remote node does not imply
+    /// the remote node received anything.
     pub last_used: Option<Duration>,
 }
 
 impl NodeInfo {
     /// Get the duration since the last activity we received from this endpoint
     /// on any of its direct addresses.
+    // TODO: This does not contain what it claims to contain!  It can not be fixed easily.
+    #[deprecated(since = "0.23.0", note = "this is broken")]
```
dignifiedquire, line 1309: I am really against this, we shouldn't deprecate methods without having replacements, just because we think/know they are broken

iroh/src/node/rpc.rs
```diff
@@ -1054,31 +1051,30 @@ impl<D: BaoStore> Handler<D> {
         rx
     }
 
-    fn node_connections(
+    fn all_node_info(
```
dignifiedquire, line 1054: oh no, yet another name for almost the same thing :/

ramfox:
I think this is a great name change: the original name is from when this data was ephemeral. I agree, it makes no sense now.

iroh-net/src/magicsock/node_map/node_state.rs
```diff
@@ -1227,64 +1271,85 @@ impl From<RelayUrlInfo> for RelayUrl {
     }
 }
 
-/// Details about an iroh node which is known to this node.
+/// Details about a remote iroh-net node which is known to this node.
+///
+/// Having details of a node does not mean it can be connected to.  Nor that it has ever
```
ramfox, line 1276:
```suggestion
/// Having details of a node does not mean it can be connected to, nor that it has ever
```

iroh-net/src/magicsock/node_map/node_state.rs
```diff
@@ -1227,64 +1271,85 @@ impl From<RelayUrlInfo> for RelayUrl {
...
-    /// Relay connection over relay
+    /// A connection via a relay server.
     #[display("relay")]
     Relay(RelayUrl),
     /// Both a UDP and a relay connection are used.
     ///
-    /// This is the case if we do have a UDP address, but are missing a recent confirmation that
-    /// the address works.
+    /// This is the case if there is a known UDP address, but no recent confirmation that
+    /// the address works.  Normally on a healthy network path this should not occur and the
```
ramfox, line 1348:
```suggestion
    /// the address works.  Normally, on a healthy network path, this should not occur and the
```

iroh-net/src/magicsock/node_map/path_state.rs
```diff
@@ -160,6 +162,15 @@ impl PathState {
             .copied()
     }
 
+    /// The last control or DISCO message **about** this path.
+    ///
+    /// This is a combination of both when this was last advertised via a call-me-maybe
+    /// message as well as any direct ping-pong communication over this path.
```
ramfox, line 168:
```suggestion
    /// This is a combination of both:
    /// - the last time this path was advertised via a call-me-maybe message
    /// - the last time there was direct ping-pong communication over this path.
```

iroh/src/rpc_protocol/node.rs
```diff
@@ -46,38 +46,39 @@ pub enum Response {
...
 }
 
-/// List connection information about all the nodes we know about
+/// List network path information about all the remote nodes know by this node.
 ///
-/// These can be nodes that we have explicitly connected to or nodes
-/// that have initiated connections to us.
+/// There may never have been connections to these nodes, and connections may not even be
+/// possible.  As well due to connections nodes can become known due to discovery mechanisms
+/// or be added manually.
```
ramfox, line 59:
```suggestion
/// possible.  Nodes can become known due to discovery mechanisms or be added manually.
```

iroh-net/src/magicsock.rs
ramfox: What about unifying to `all_node_infos`? I assume `node_infos` is too close to `node_info` to be a memorable method name?

Then I wouldn't mind `all_node_infos_iter` as the one exception

iroh-net/src/magicsock/node_map/node_state.rs
divagant-martian: reverted in e6cb86313862fbb9cf9e9c6cb096436f9b3e3163

iroh-cli/src/commands/node.rs
divagant-martian: reverted in f521dbb

dignifiedquire (approved):

iroh-net/src/endpoint.rs
```diff
@@ -708,26 +708,33 @@ impl Endpoint {
...
-    /// known direct addresses, when there was last any contact with this node and what kind
-    /// of connection this was.
-    pub fn connection_info(&self, node_id: NodeId) -> Option<ConnectionInfo> {
-        self.msock.connection_info(node_id)
+    /// The [`Endpoint`] keeps some information about remote iroh-net nodes, which it uses to find
+    /// the best path to a node. Having information on a remote node, however, does not mean it has
+    /// ever been connected or even whether a connection is even possible. The information about a
+    /// remote node will change over time, as the [`Endpoint`] learns more about the node. Future
+    /// calls may return different information. Furthermore, node information may even be
+    /// completely evicted as it becomes stale.
```
ramfox, line 718:
```suggestion
    /// the best path to a node. Having information on a remote node, however, does not mean we have
    /// ever connected to it to or even whether a connection is even possible. The information about a
    /// remote node will change over time, as the [`Endpoint`] learns more about the node. Future
    /// calls may return different information. Furthermore, node information may even be
    /// completely evicted as it becomes stale.
```

iroh-net/src/magicsock/node_map/node_state.rs
```diff
@@ -1185,29 +1188,58 @@ pub enum ControlMsg {
...
+    /// Elapsed time since this network path was known to exist.
+    ///
+    /// A network path be considered to exist only because the remote node advertised it.
+    /// It may not mean the path is usable.  However if there was any communication with the
+    /// remote node over this network path it also means the path exists.
+    ///
+    /// The elapsed time since *any* confirmation of the path's existence was received is
+    /// returned.  If the remote node moved network and no longer has this path, this could
+    /// be a long duration.  If the path was added via [`Endpoint::add_node_addr`] or some
+    /// node discovery the path may never have been known to exist.
```
ramfox, line 1229:
```suggestion
    /// returned.  If the remote node moved networks and no longer has this path, this could
    /// be a long duration.  If the path was added via [`Endpoint::add_node_addr`] or some
    /// node discovery the path may never have been known to exist.
```

iroh-net/src/magicsock/node_map/node_state.rs
```diff
@@ -1227,27 +1259,38 @@ impl From<RelayUrlInfo> for RelayUrl {
...
+    /// Some of these addresses might only be valid for networks we are not part of, but the remote
+    /// node might be a part of.
     pub addrs: Vec<DirectAddrInfo>,
     /// The type of connection we have to the node, either direct or over relay.
     pub conn_type: ConnectionType,
-    /// The latency of the `conn_type`.
+    /// The latency of the current network path to the remote node.
     pub latency: Option<Duration>,
-    /// Duration since the last time this node was used.
+    /// Time elapsed time since last sending to, or received from the node.
```
ramfox, line 1285:
```suggestion
    /// Time elapsed time since last we have sent to or received from the node.
```

iroh-net/src/magicsock/node_map/node_state.rs
```diff
@@ -1227,27 +1259,38 @@ impl From<RelayUrlInfo> for RelayUrl {
...
+    ///
+    /// This is the duration since *any* data was sent or receive from the remote node,
+    /// payload or control messages.  Note that sending to the remote node does not imply
```
ramfox, line 1288:
```suggestion
    /// This is the duration since *any* data (payload or control messages) was sent or receive 
    /// from the remote node. Note that sending to the remote node does not imply
```

iroh-net/src/magicsock/node_map/path_state.rs
```diff
@@ -160,6 +162,14 @@ impl PathState {
             .copied()
     }
 
+    /// The last control or DISCO message **about** this path.
+    ///
+    /// This is the most recent instant between:
```
ramfox, line 167:
```suggestion
    /// This is the most recent instant among:
```

PR #2611, opened 2024-08-10
gen api for iroh.computer

divagant-martian:
## Description

little utility to generate what iroh.computer expects as input for the api descriptions. The generated output still needs checks since not everything in the console is relevant for the api


dignifiedquire:
@divagant-martian are you planning to get this merge ready, or is this just an experiment?

divagant-martian:
Not sure. I have used this a couple times already and has been quite useful. In a scenario where the CLI continues to exist and we consider it relevant enough to keep updating the docs, this is useful. That being said, this is not a user-facing api (so it's actually a hidden command) and it's seems to be the CLI's future is not certain? This is the reason why I have kept it as a draft and updated it when needed.

To answer your question, I don't really have plans to make this merge-ready. At least not until we decide what we want to do with the CLI. If we keep it maybe this can be feature-gated under a `dev` flag or something similar. 

Thoughts?

Issue #2601, opened 2024-08-08
feat: add `locally_discovered_nodes` API


ramfox:
Add a method on the `Endpoint` that gives you a list of the nodes that have been discovered using mdns/swarm-discovery so far. If local discovery is not configured for that `Endpoint` it returns an empty list.

andrewdavidmackenzie:
I'd consider returning a `Result`
- `Ok<Vec<Node>>` (Vec could be empty, which is valid, assuming it will filter out "self")
- `Err(e)` if discovery is not configured, or some other error

The entry in the list should at least have the information necessary to connect to it (`NodeId`?)

flub:
What is the usecase for this?

Mostly I'm wondering why mDNS nodes are different here than other nodes and whether this is trying to expose a very specialised sub-view of the NodeState we have.  And thus whether this might be better as a more generic view into the NodeState.

The property of mDNS discovered nodes in NodeState terms is probably:

- We have a UDP path for the node which we know the node itself currently thinks it is reachable on.  This is currently the `PathState::last_alive` condition but sadly that's a very confusing name.

- We know we are on the same network as this path.

I don't think we have this 2nd bit of information in the `PathState` currently.  Maybe that's what should be added to enable this.

You can already approximate it by guessing based on the network, but we can both be on `192.168.0.0/24` without being on the same network.

---

We already have `Endpoint::connection_infos` which gives you a view of the `NodeState`.  Maybe this is sufficient and the returned `ConnectionInfo` could have a new fields or methods to figure out if you think this node is on the same network as you are?

matheus23:
> What is the usecase for this?

From discord:

> I'd like to have a method in my app that finds compatible instances of the app (peers) in the local network and lists them in the UI, for selection to connect to.

---

> We already have `Endpoint::connection_infos` which gives you a view of the `NodeState`. Maybe this is sufficient and the returned `ConnectionInfo` could have a new fields or methods to figure out if you think this node is on the same network as you are?

AFAIU `connection_infos` only lists you nodes that you've connected to or tried to connect to at least once. But what our users are asking for is a way to "discover" node IDs that they likely want to connect to.

flub:
> > What is the usecase for this?
> 
> From discord:
[...]

Ah, thanks.  I did browse discord quickly but did not find this.  Anyway, good to include it in the issue.


[...]

PR #2612, opened 2024-08-12
feat: expose locally discovered node id and addresses

ramfox:
## Description

We've had multiple requests for user to get a list of nodes that have been discovered on the local network when using `local-swarm-discovery`. This is my attempt.

## Breaking Changes

Adds `locally_discovered_nodes` method to `Discovery` trait

Adds `locally_discovered_nodes` method to the `Endpoint`

Adds rpc structs and a `locally_discovered_nodes` method on the client. 

## Notes & open questions

This may be a tragic addition to the `Discovery` trait, but wanted to get feedback. 

Essentially, this adds  a `locally_discovered_nodes` method to the `Discovery` trait, that returns `None`, unless your discovery mechanism discovers nodes locally. 

I'm not sure how else we would expose this, without a large `Discovery` refactor.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/src/discovery.rs
+++ b/iroh-net/src/discovery.rs
@@ -136,6 +136,11 @@ impl ConcurrentDiscovery {
     pub fn add(&mut self, service: impl Discovery + 'static) {
         self.services.push(Box::new(service));
     }
+
+    /// Return a slice of the [`Discovery`] services we have available
+    pub fn services(&self) -> &[Box<dyn Discovery>] {
+        &self.services
+    }
 }
 
 impl<T> From<T> for ConcurrentDiscovery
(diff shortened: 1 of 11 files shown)


rklaehn:
Not sure if having a separate fn in Discovery is the way to go. We already have a `provenance` field in DiscoveryItem that is as yet mostly unused but can be used to distinguish between different discovery mechanisms.

Why not lean on that? People can just call discovery and filter out all items that don't have the right provenance. And for the address book or whatever, we should track provenance.

Maybe provenance should be an open enum instead of just a 'static str for more type safety, but that's another question.

```rust
enum Provenance {
  Local,
  Dns,
  Pkarr,
  Ticket,
   Other(&'static str),
}
```

or something...

matheus23:
> I'm not sure how else we would expose this, without a large `Discovery` refactor.

Thoughts on putting this information into `Endpoint::connection_info`? I.e. reusing `Endpoint::add_node_addr` in the `LocalSwarmDiscovery` task.
(See [this comment](https://github.com/n0-computer/iroh/issues/2601#issuecomment-2277657919))

ramfox:
replaced by #2656

PR #2614, opened 2024-08-12
experiment: try to only use tokio::mpsc in iroh-net

dignifiedquire:
I am failing to implement the `send_blocking`, if anyone has ideas, I would love to hear them

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2839,7 +2839,6 @@ name = "iroh-net"
 version = "0.22.0"
 dependencies = [
  "anyhow",
- "async-channel",
  "axum",
  "backoff",
  "base64 0.22.1",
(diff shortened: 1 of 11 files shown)


rklaehn:
Pushed a fix. But this isn't really send_blocking, more like send_blocking without runtime and send_detached or whatever in case a runtime is found.

It might be that this is OK in most cases. But then the fn should be called send_sync or something and return just unit to signal that it is fire and forget / best effort.

rklaehn:
As you mention, the calls to send_blocking can be replaced by a normal send call just by making some fns async. So should we just do that?

send_blocking works now, except that it is not really blocking in the case where a runtime is found. But we don't need it at all in iroh-net.

dignifiedquire:
> As you mention, the calls to send_blocking can be replaced by a normal send call just by making some fns async. So should we just do that?

yes, we should try and do that

dignifiedquire:
closing in favor of #2620

PR #2617, opened 2024-08-12
chore(ci): use nextests groups to isolate some tests

divagant-martian:
## Description

After a lot of debugging we realized the integrations tests (`gc` tests, for
example) run with swarm discovery on since they are not technically using test
iroh, so they bypass the cfg that prevents it from being turned on. This causes
the `test_local_swarm_discovery` test to be flooded with nodes from other
tests, making it flaky. Isolating the test should make it no longer flaky
anymore

## Breaking Changes

n/a

## Notes & open questions

We should still create a test that creates a bunch of nodes and makes sure they
all can find each other within reasonable time. There is still reason to
believe swarm discovery is slow, particularly in linux, but we should address
it in a test under our countrol, not via the non-deterministic nature of
concurrent tests in ci.

## Change checklist

- [x] Self-review.
- [ ] ~~Tests if relevant.~~

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,6 @@
+[test-groups]
+run-in-isolation = { max-threads = 1 } # this are tests that must not run with other tests concurrently
+
+[[profile.default.overrides]]
+filter = 'test(::run_in_isolation::)'
+test-group = 'run-in-isolation'
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,6 +1,10 @@
 [test-groups]
-run-in-isolation = { max-threads = 1 } # this are tests that must not run with other tests concurrently
+run-in-isolation = { max-threads = 16 } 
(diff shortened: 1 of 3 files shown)


Arqu (approved):
Agree on the tests to be added.

This looks to work as intended. Ran `cargo nextest show-config test-groups` to confirm and got what was expected:
```
 Finished `test` profile [unoptimized + debuginfo] target(s) in 1m 35s
group: run-in-isolation (max threads = 1)
  * override for default profile with filter 'test(::run_in_isolation::)':
      iroh-net:
          discovery::local_swarm_discovery::tests::run_in_isolation::test_local_swarm_discovery
          ```

.config/nextest.toml
flub, line 1: oh, this cool.  maybe there are other things we could use this for.  A bit of a shame the only kind of marker you have is the name, but better than nothing.

PR #2618, opened 2024-08-13
fix(iroh-willow): simplify graceful connection termination

Frando:
## Description

It appeared to me that we can very much simplify the graceful connection termination logic. We do not need the information which peer started the termination. We can simply ..
* send a goodbye byte
* wait until we either a) receive the goodbye byte, or b) the peer closes the connection with our  graceful error code
* only now close the connection with the graceful error code

On both sides, we only close  the connection after either receiving the goodbye byte or a graceful connection termination. This makes sure that neither peer ever closes the connection gracefully prematurely.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-willow/src/net.rs
+++ b/iroh-willow/src/net.rs
@@ -4,7 +4,8 @@ use anyhow::{anyhow, ensure, Context as _, Result};
 use futures_concurrency::future::TryJoin;
 use futures_util::future::TryFutureExt;
 use iroh_base::key::NodeId;
-use iroh_net::endpoint::{Connection, ConnectionError, RecvStream, SendStream, VarInt};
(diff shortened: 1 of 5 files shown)


iroh-willow/src/net.rs
```diff
@@ -30,6 +31,9 @@ pub const CHANNEL_CAP: usize = 1024 * 64;
 /// The ALPN protocol name for iroh-willow.
 pub const ALPN: &[u8] = b"iroh-willow/0";
 
+/// QUIC application error code for closing with failure.
+pub const ERROR_CODE_FAIL: VarInt = VarInt::from_u32(0);
```
flub, line 35: i usually reserve `0` for "accidentally dropped" because it used by default if you drop in quinn

PR #2619, opened 2024-08-13
test(iroh): Disable DNS discovery altogether in tests

matheus23:
## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


matheus23:
Situation sucks IMHO. But this is not the solution. Will need to think about this & talk to flub&others about this. We can reopen if we want.

PR #2620, opened 2024-08-13
refactor(iroh-net)!: remove async channel

rklaehn:
## Description

Removes async-channel from iroh-net and replaces it with the tokio mpsc channel.

This is a mergeable version of https://github.com/n0-computer/iroh/pull/2614

## Breaking Changes

LocalSwarmDiscovery is no longer UnwindSafe

## Notes & open questions

Open question: I am using blocking_send from inside the windows RouteMonitor. This depends on that the callback thread is not a tokio thread. I have no idea if this is a reasonable assumption to make. Otherwise we would have to bring back the runtime check condition or just capture a runtime handle and spawn a task.

Note: what even is this? The only message there is is NetworkMessage::Change, and there is no timestamp or anything. So could I just use try_send? If the queue is full, there is already a Change in it, so it is as good as the new one...

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/Cargo.toml
+++ b/iroh-net/Cargo.toml
@@ -17,7 +17,6 @@ workspace = true
 
 [dependencies]
 anyhow = { version = "1" }
-async-channel = "2.3.1"
 base64 = "0.22.1"
 backoff = "0.4.0"
 bytes = "1"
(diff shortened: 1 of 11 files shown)


iroh-net/Cargo.toml
```diff
@@ -127,7 +130,7 @@ harness = false
 duct = "0.13.6"
 
 [features]
-default = ["metrics"]
+default = ["metrics", "local_swarm_discovery", "iroh-relay"]
```
dignifiedquire, line 133: I only added this for testing, please undo

dignifiedquire (approved):

PR #2622, opened 2024-08-13
refactor(iroh-bytes): remove flume dependency

rklaehn:
## Description

This is mostly a pretty mechanical replacement of flume with async_channel in the rebd blob store.

One not so tiny change is that to support closing a read or write transaction after some delay, and because async_channel does not have recv_timeout, we switch the entire actor loop to run *async* but on a dedicated thread of the main runtime.

That way we can use `tokio::select! {..., timeout}`

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


rklaehn:
@Frando this is similar to what you did with the docs store, I guess. The only difference is that I am using the main runtime to block on so you get the main blocking pool when calling spawn_blocking instead of the (single?) io thread of a single threaded runtime...

dignifiedquire (approved):

PR #2624, opened 2024-08-14
refactor(iroh-blobs): use oneshot channel from oneshot crate

rklaehn:
## Description

In several places we need a oneshot channel to communicate between sync and async code. In particular, there is one place where we intentionally block inside async code to be able to implement a sync callback (in the docs code, will go away at some point hopefully).

So far we have used flume and now async_channel channels with capacity 1 for this, which is quite wasteful.

This replaces all oneshot use in the fs store with the oneshot channel from the oneshot crate, which is pretty minimal and does not panic when calling blocking_recv from a tokio context.

## Breaking Changes

None

## Notes & open questions

Note: this was part of an attempt to get rid of async_channel in iroh-blobs. We still need it for sync interactions with the blob store from inside tokio tasks, for the current docs impl. But hopefully we should be able to replace the channel for the redb actor with a mpsc channel once that is sorted out.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


dignifiedquire (approved):
seems fine

PR #2627, opened 2024-08-14
fix(ci): report flaky outupt only on success and failure

divagant-martian:
## Description

in #2623 the condition for `notify` is not being cancelled, but a job can also be skipped. This is the output when a PR is not even running the tests. So, we get a good report in cases where there was no output. This fixes that issue

## Breaking Changes
n/a

## Notes & open questions

n/a

## Change checklist

- [x] Self-review.
- [ ] ~~Tests if relevant.~~


Arqu (approved):
Ah yes, this is a known gotcha with actions :D

PR #2628, opened 2024-08-14
feat(iroh-net): Add PkarrNodeDiscovery to iroh-net

rklaehn:
## Description

Deprecate iroh-pkarr-node-discovery and add the full dht based node discovery to iroh itself.

## Breaking Changes

None

## Notes & open questions

PkarrNodeDiscovery does both resolution and publishing in one thing, which is different to the two relay based versions.
Should we keep it this way? I think yes, given that we are instantiating a full dht client and we don't want 2 of them...

Should we keep in the relay options, or remove them and make this one a pure DHT one?

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/Cargo.toml
+++ b/iroh-net/Cargo.toml
@@ -93,6 +93,7 @@ strum = { version = "0.26.2", features = ["derive"] }
 # local_swarm_discovery
 swarm-discovery = { version = "0.2.1", optional = true }
 tokio-stream = { version = "0.1.15", optional = true }
+genawaiter = "0.99.1"
 
 
 [target.'cfg(any(target_os = "linux", target_os = "android"))'.dependencies]
--- a/iroh-net/src/discovery/pkarr.rs
+++ b/iroh-net/src/discovery/pkarr.rs
@@ -21,6 +21,7 @@ use crate::{
     key::SecretKey,
     AddrInfo, Endpoint, NodeId,
 };
+pub mod dht;
 
 /// The pkarr relay run by n0, for production.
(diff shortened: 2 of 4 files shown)


iroh-net/src/discovery/pkarr.rs
```diff
@@ -21,6 +21,7 @@ use crate::{
     key::SecretKey,
     AddrInfo, Endpoint, NodeId,
 };
+pub mod dht;
```
dignifiedquire, line 24: lets put this behind a feature flag, disabled by default

iroh-net/Cargo.toml
```diff
@@ -93,7 +93,7 @@ strum = { version = "0.26.2", features = ["derive"] }
 # local_swarm_discovery
 swarm-discovery = { version = "0.2.1", optional = true }
 tokio-stream = { version = "0.1.15", optional = true }
-
+genawaiter = { version = "0.99.1" }
```
dignifiedquire, line 96: shouldn't this be optional?

iroh-net/src/discovery/pkarr/dht.rs
```diff
@@ -0,0 +1,365 @@
+//! # Pkarr based node discovery for iroh-net
```
dignifiedquire, line 1: this has no tests, would be good to have some basic test to make sure this works

rklaehn: I will see what I can do. The mainline crate has a testnet feature, otherwise I would have to assume that the global DHT is available from tests...

dignifiedquire (approved):

PR #2631, opened 2024-08-15
refactor(iroh-cli)!: metrics-addr cli arg, metrics off by default

b5:
## Description

When iroh is running in a docker container, not being able to specify the full address is an issue. This aligns with the rpc-addr flag.

at the same time–but for different reasons–I'd also like to disable surfacing metrics by default. My primary reason for this is an experience I've had numerous times using iroh:
1. Start an initial node
2. Start a second node, on the same machine (usually while debugging or following a tutorial)
3. See a "can't bind to port" error from the iroh metrics crate
4. Wonder to myself what metrics do, and if they're being sent anywhere (I know they by default are not)

I think metrics should be treated as an advanced, opt-in feature that power users discover and are delighted to find. As tribute to this work I offer to write a docs page about enabling metrics to get this merged.

## Breaking Changes

* `--metrics-port` flag on the iroh cli is now `--metrics-addr`, eg: specify `127.0.0.1:9090` instead of `9090`

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


dignifiedquire:
In general I like this, disabling metrics by default seems right to me

rklaehn:
I also found it quite annoying to deal with metrics when running multiple local irohs...

Arqu:
Codewise fine, but this brings us to a weird spot now.
Couple things:
- note that this doesn't ship any metrics anywhere, just exposes them on :9090 or whatever port
- the idea from a couple months ago was to go default on, mainly to support the flow with `iroh doctor plot`
- it was also shortened to only `port` for ergonomics as usually you just want it to be closed off to local host. Only hosted environments would want to bind to `0.0.0.0` or any other alternative and should be fine setting a config file anyways.

That said, and it going against what we wanted just a while ago, is that for docker this actually makes a ton of sense. I imagine most people don't run multiple nodes on a single box but dev flows do. Also for docker you generally would want to be able to bind to other addrs cause the current alternative is to mount a config file as a volume.

I'll give it a 👍 but am generally divided on the topic.

Arqu (approved):

b5:
Yeah, I hear you loud & clear @Arqu that this is a reversal on our prior rationale. I think we've mainly learned a bunch from actually using iroh in this setup, which is a good thing. it will hinder `iroh doctor plot` a bit, but I do personally plan to fix that by turning on metrics through config in my local dev repo

Arqu:
All good!

PR #2634, opened 2024-08-15
refactor(iroh-net)!: Rename the local-swarm-discovery feature to discovery-local-network

rklaehn:
## Description

Rename the local-swarm-discovery feature to discovery-local-network to be consistent with discovery-pkarr-dht

## Breaking Changes

renames the `local-swarm-discovery` feature flag in iroh-net to `discovery-local-network`

## Notes & open questions

Naming is hard. Pretty sure the ff should start with `discovery-`, but should it be `discovery-local-network` or `discovery-local-swarm` or just -local...?

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/Cargo.toml
+++ b/iroh-net/Cargo.toml
@@ -146,7 +146,7 @@ iroh-relay = [
 ]
 metrics = ["iroh-metrics/metrics"]
 test-utils = ["iroh-relay"]
-local-swarm-discovery = ["dep:swarm-discovery", "dep:tokio-stream"]
+discovery-local-network = ["dep:swarm-discovery", "dep:tokio-stream"]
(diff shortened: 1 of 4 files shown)


matheus23:
My vote goes for `discovery-local-swarm`.
There's a different version of local discovery that doesn't try to establish a swarm but instead tries to establish connections very selectively and simply makes use of multicast in a more broadcast-request/single-response way.
That one would also be local, and also discovery, and also network, but not swarm.

flub:
Agree that a `discovery-` prefix is good, but indeed the name of it is tricky.  I'm still trying to understand how this discovery works...

dignifiedquire (approved):

Issue #2621, opened 2024-08-13
Node discovery uses wrong detection of when new paths are needed


flub:
`DiscoveryTask::needs_discovery` uses some APIs which provides the wrong information.  We need to re-evaluate how exactly it decides this.

flub:
This is totally different for multipath anyway

PR #2636, opened 2024-08-15
[IGNORE] flub/connection-info-wtf bkp

divagant-martian:
**copy-pasted from original**
## Description

This migrates Endpoint::connection_info and Endpoint::connection_infos to Endpoint::node_info and Endpoint::node_info_iter.  This returns information about nodes in the node map.  This has not that much to do with connections.  Calling this connections is rather misleading.

It documents all types below it properly, deprecating some problematic fields and functions.

### iroh-net

- `Endpoint::connection_info` -> `Endpoint::node_info`
- `Endpoint::connection_infos` -> `Endpoint::node_info_iter` (returns `Iterator` instead of `Vec`.
- `ConnectionInfo` -> `NodeInfo`
- `DirectAddrInfo::last_control` is deprecated, it contains misleading information.
- `NodeInfo::id` is removed, it was internal information.
- `NodeInfo::last_received` is deprecated as this information is wrong and can not currently be provided.
- `NodeInfo::last_alive_relay` is deprecated as it is a simple accessor for an already public field.

### iroh

- `Client::connections` -> `Client::node_info_iter`
- `Client::connection_info` -> `Client::node_info`

## Notes & open questions

This started as a documentation PR, so really concentrates on doing the
documentation right.  It could not avoid doing some fixes to the API though.

The `node_info_iter()` API is potentially a weird naming and API choice.  It
is a `Vec` returned as an iter.  I think this is ok for several reasons:
- The API name is much better than e.g. a single-letter difference: `node_infos`.  This would be a lot more prone to typos.
- Returning things as an iterator is kind of rust-like.
- If some place collects it into a `Vec` again the compiler will optimise it away.

The node discovery uses some deprecated APIs.  This is the reason I have left
those around as deprecated instead of entirely removing them.  The information
provided by these APIs is not what it claims to be and finding the right way to
fix the usage is better done as a follow-up: #2621.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-cli/src/commands/doctor.rs
+++ b/iroh-cli/src/commands/doctor.rs
@@ -30,7 +30,7 @@ use iroh::{
         defaults::DEFAULT_STUN_PORT,
         discovery::{dns::DnsDiscovery, pkarr::PkarrPublisher, ConcurrentDiscovery, Discovery},
         dns::default_resolver,
-        endpoint::{self, Connection, ConnectionTypeStream, RecvStream, SendStream},
+        endpoint::{self, Connection, ConnectionTypeStream, NodeInfo, RecvStream, SendStream},
         key::{PublicKey, SecretKey},
         netcheck, portmapper,
         relay::{RelayMap, RelayMode, RelayUrl},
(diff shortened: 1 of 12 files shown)


divagant-martian:
everyone seems happy with the state of things so I'm going to close this

PR #2641, opened 2024-08-19
refactor(iroh-net): move all timeouts into one file

### LINKED ISSUE #2602: refactor: move ALL timeouts into one file
first step in centralizing our iroh-net timeouts to be able to give users a lever to adjust timeouts the way they want.

edit:
This is specifically about moving all of the existing timeouts into one location, so it's easier to work through #2337 and understand how all of the timeouts relate to each other.

ramfox:
## Description

Co-locates all timeout defaults that have to do with making connections or running netcheck.

closes #2602

## Notes & open questions

I nested some of the timeouts in another module (specifically the ones pertaining to the relay) but I could do this more for other sections, ie, the portmapper timeouts could be nested in a `portmapper` module.

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh-net/src/defaults.rs
+++ b/iroh-net/src/defaults.rs
@@ -125,3 +125,77 @@ pub mod staging {
         }
     }
 }
+
+/// Contains all timeouts that we use in `iroh-net`.
+pub mod timeouts {
+    use std::time::Duration;
+
+    // Timeouts for netcheck
+
+    /// Maximum duration to wait for a netcheck report.
+    pub(crate) const NETCHECK_REPORT_TIMEOUT: Duration = Duration::from_secs(10);
+
+    /// The maximum amount of time netcheck will spend gathering a single report.
+    pub(crate) const OVERALL_REPORT_TIMEOUT: Duration = Duration::from_secs(5);
+
+    /// The total time we wait for all the probes.
+    ///
+    /// This includes the STUN, ICMP and HTTPS probes, which will all
+    /// start at different times based on the [`ProbePlan`].
+    pub(crate) const PROBES_TIMEOUT: Duration = Duration::from_secs(3);
+
+    /// How long to await for a captive-portal result.
+    ///
+    /// This delay is chosen so it starts after good-working STUN probes
+    /// would have finished, but not too long so the delay is bearable if
+    /// STUN is blocked.
+    pub(crate) const CAPTIVE_PORTAL_DELAY: Duration = Duration::from_millis(200);
+
+    /// Timeout for captive portal checks
+    ///
+    /// Must be lower than [`OVERALL_REPORT_TIMEOUT`] minus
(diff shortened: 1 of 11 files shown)


iroh-net/src/netcheck/reportgen.rs
```diff
@@ -46,32 +46,13 @@ mod probes;
...
+    CAPTIVE_PORTAL_DELAY, CAPTIVE_PORTAL_TIMEOUT, DNS_TIMEOUT, OVERALL_REPORT_TIMEOUT,
+    PROBES_TIMEOUT,
+};
 
 const ENOUGH_NODES: usize = 3;
 
-const DNS_TIMEOUT: Duration = Duration::from_secs(3);
-
 /// Delay used to perform staggered dns queries.
 const DNS_STAGGERING_MS: &[u64] = &[200, 300];
```
matheus23, line 57: Not strictly a timeout, but still time-related. Should we move this in the timeout module?

matheus23: I guess there's more timing stuff like this, but it makes sense to keep it at timeouts only :)

iroh-net/src/defaults.rs
```diff
@@ -125,3 +125,77 @@ pub mod staging {
         }
     }
 }
+
+/// Contains all timeouts that we use in `iroh-net`.
+pub mod timeouts {
```
flub, line 130: `pub(crate)` I think?

I've come to the conclusion it's a lot easier to always be explicit about the visibility of things rather than rely on things like "this is pub but it's inside a pub(crate)" or similar.

flub (commented):
I'm kind of -1 on this.  While it is nice and ordered right now I think it will be hard to maintain this.  New timeouts will be added and not make it here because there's nothing naturally pointing here when you're in new code.  `const` items are now also treated differently whether they're a timeout or something else which is also pretty arbitrary.

I agree it is nice to see everything in one place and with docs.  Perhaps that's something that should be possible with a bit of grep though.  (`rg -B 3 ': Duration = Duration::from'` is probably a decent start).

So I"m mainly worried this will result in a much messier code structure in the long term.

dignifiedquire:
> I'm kind of -1 on this. While it is nice and ordered right now I think it will be hard to maintain this.

Well this is the first round,  eventually these will be configurable, making this more sensible I believe

dignifiedquire:
@ramfox I don't think you are updating the DNS resolver configuration, or the quinn connection configuration yet

ramfox:
@dignifiedquire we currently just use the defaults for quinn (10s) and the dns resolver (5s), so there are no `const`s to move around. The only places that we adjust the `max_idle_timeout` for quinn is in tests. And we have 2 consts for dns timeouts, one for netcheck and one for the client relay, but those are timeouts that wrap the dns resolver, so they can only shorten the timeout, not lengthen it.

dignifiedquire (approved):

PR #2645, opened 2024-08-19
refactor(iroh-net,iroh)!: rename to remote_info_iter, fixup some docs

flub:
## Description

This renames remote_infos_iter to remote_info_iter.  This returns an
iterator of remote_info items.  Using plural for "info" adds a double
plural together with "iter", creating something of a pleonasm.

## Breaking Changes

None right now, this is unreleased.

But `Endpoint::remote_infos_iter` -> `Endpoint::remote_info_iter` from
previous breaking changes.

## Change checklist

- [x] Self-review.
- ~~Tests if relevant.~~

### DIFF
--- a/iroh-cli/src/commands/node.rs
+++ b/iroh-cli/src/commands/node.rs
@@ -49,7 +49,7 @@ impl NodeCommands {
     pub async fn run(self, iroh: &Iroh) -> Result<()> {
         match self {
             Self::RemoteList => {
-                let connections = iroh.remote_infos_iter().await?;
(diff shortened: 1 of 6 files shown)


dignifiedquire (approved):

divagant-martian:
I find `remote_infos_iter` fine since we iterate many of them instead of one internally, giving something like addresses for example, but fine regardless I think

ramfox (approved):
Ah, interesting. This definitely didn't bother me, but this is more semantically correct.

PR #2656, opened 2024-08-21
 feat(iroh-net)!: add `subscribe` method to `Discovery` trait

ramfox:
## Description

Adds a `subscribe` method to the `Discovery` trait that returns an `Option<BoxStream>`. The `subscribe` method will send `DiscoveryItems` each time the discovery service discovers a remote node.

The magicsock is now subscribed to the discovery service and updates the internal address book each time it receives a `DiscoveryItem`. The source is marked as `Source::Discovery{ service: String }` and the `Instant` that the magicsock received the information.

Users can now filter their list of `RemoteInfo`s for sources.

## Breaking Changes

- struct `RemoteInfo` now has field `sources`, which is a `Vec` of `(iroh::net::endpoint::Source, Duration)`. The `Source` is how we heard about the remote, and the `Duration` is how long ago we heard about it. The `sources` field is ordered from the first time we heard about the remote node to the most recent time we learned about the remote node, in this session.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh-net/src/discovery.rs
```diff
@@ -95,6 +95,125 @@ pub trait Discovery: std::fmt::Debug + Send + Sync {
...
+        self.0.handle.abort();
+    }
+}
+
+impl DiscoveryEvents {
+    pub fn new() -> Self {
+        let (sender, mut recv) = channel(20);
+        let event_loop = async move {
+            loop {
+                let mut subscribers: HashMap<u64, Sender<DiscoveryItem>> = HashMap::default();
```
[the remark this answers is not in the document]
ramfox: yepppp

iroh-net/src/discovery.rs
```diff
@@ -95,6 +95,125 @@ pub trait Discovery: std::fmt::Debug + Send + Sync {
...
+    /// Returns a stream of `DiscoveryItems` and an optional unsubscribe
+    /// function.
+    ///
+    /// If a stream is dropped, the `DiscoveryEvent` struct will clean
+    /// up the dangling subscription, even if you do not unsubscribe.
+    pub fn subscribe(
+        &self,
+    ) -> (
+        BoxStream<DiscoveryItem>,
+        impl std::future::Future<Output = ()>,
```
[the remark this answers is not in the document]
ramfox: love it, appreciate the strong opinion & i agree.

iroh-net/src/discovery.rs
```diff
@@ -95,6 +95,125 @@ pub trait Discovery: std::fmt::Debug + Send + Sync {
...
+            .0
+            .last_id
+            .fetch_add(1, std::sync::atomic::Ordering::SeqCst);
+        let subscribe_sender = self.0.sender.clone();
+        tokio::spawn(async move {
+            subscribe_sender
+                .send(Message::Subscribe((id, send)))
+                .await
+                .ok();
+        });
```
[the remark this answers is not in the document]
ramfox: Okay, so you are saying that because we create a task for every time `subscribe` is called, we could be creating an unbounded number of tasks, each of which would be waiting for `subscribe_sender` to have space? So it's better to make `subscribe_sender` unbounded & use `try_send`, so we don't need to `await` the `subscribe_sender.send()` & no longer need to spawn a task?

Just making sure I follow.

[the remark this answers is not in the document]
matheus23: I'd prefer if this was using `futures_util`'s [`flatten_stream`](https://docs.rs/futures-util/latest/futures_util/future/trait.FutureExt.html#method.flatten_stream) and thus do the `send` as part of the first action in the stream and avoid `tokio::spawn` and `try_send` altogether.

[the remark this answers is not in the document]
ramfox: This is much prettier looking, at minimum 😂 

To ensure I understand: this basically would defer sending the `Message::Subscribe` event until the stream is `await`d, correct?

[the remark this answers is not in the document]
matheus23: Yeah.

iroh-net/src/magicsock/node_map/node_state.rs
```diff
@@ -128,6 +128,9 @@ pub(super) struct NodeState {
     last_call_me_maybe: Option<Instant>,
     /// The type of connection we have to the node, either direct, relay, mixed, or none.
     conn_type: Watchable<ConnectionType>,
+    /// How we learned about this node. We can have learned about this node from multiple stores.
+    /// Sources later in the list were sources we learned from more recently.
+    sources: Vec<(super::Source, Instant)>,
```
divagant-martian, line 133: I'm a bit worried this can only grow, also from checking the output in the console I think I might want is the source of the particular addresses rather than the full node, not sure what are other's thoughts

divagant-martian (commented):
about the use of channels:

I generally prefer bounded channels, since they provide back pressure, but by spawning a task that does the send every time the back pressure is lost, making an unbounded channel preferable. In other channel, when sending the event back to the subscriber, there isn't this per-send spawn so it does provide back pressure, but I would argue here it's not what we want. Discovery should not be held up by an outsider reader not making room for another event. Also, a slow subscriber should not be able to held up other subscribers. I _think_ this scenario is somewhat being negated by some other `spawn` somewhere up in the stack, but the point remains: If we do intend to be held a slow reader (this would make sense for example if we schedule something such that we only care about the last result), or if we are able to drop data when full, then a bounded channel is the best choice. If not, then unbounded channels are the way to go.

I another note, I checked out the cli output and was a bit confused to get the source for the node and no the address. I would think the "source of the node" is the source of the best address we use to communicate. Maybe it's just me, I'm open to leave it like this and understand it in another way, but would like to know what others think

flub:
wrt to the bounded/unbounded channel.  I only wanted to point out that spawning a task to send on a bounded channel is not a good pattern and is essentially worse than using an unbounded channel.  It does always make sense to think about what backpressure is needed and whether something is possible or better without unbounded channels.  @divagant-martian and @matheus23 were much more constructive in that regard :)

iroh-net/src/discovery.rs
```diff
@@ -95,6 +100,92 @@ pub trait Discovery: std::fmt::Debug + Send + Sync {
...
+            .sender
+            .send(Message::Event((node_id, discovery_item)))
+            .await
+            .ok();
+    }
+
+    /// Allows you to subscribe to messages from the `DiscoveryEvent`
+    ///
+    /// Returns a stream of `DiscoveryItems` and an optional unsubscribe
+    /// function.
```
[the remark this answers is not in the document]
ramfox: yes, nice catch! thank you

iroh-net/src/magicsock/node_map/node_state.rs
ramfox: Having sources on a per-address basis is interesting, and as we talk about beefing up the node map / writing a node address book, this could be something we want to look into.

Right now, the only thing users have requested was a way to determine how a node was discovered, not about particular addresses.

I agree it's concerning to have this only grow. Maybe we only keep the last X number?

ramfox:
After a discussion with dig last week, he suggested I just make the `Discovery` trait async, to avoid these gotchas.

iroh-net/src/magicsock/node_map/node_state.rs
```diff
@@ -1288,6 +1297,10 @@ pub struct RemoteInfo {
     /// from the remote node. Note that sending to the remote node does not imply
     /// the remote node received anything.
     pub last_used: Option<Duration>,
+    /// List of places we have learned about this remote as well as [`Duration`] since we learned about it.
+    ///
+    /// Sources are sorted from least recent to most recent.
+    pub sources: Vec<(super::Source, Duration)>,
```
flub, line 1303: When I was first thinking of how to add this to the node map I though of adding a property to the `PathState` (i.e. `DirectAddrInfo` in this struct) that indicates the path is known to be on a directly attached network.  That is essentially the property that is being asked for by the users I think.  It is also slightly more general to me, and depends less on a particular naming of a source.  Having to filter for a source name seems a bit backwards to me here, because you're not asking for a property of the discovered network.

Furthermore, and to weigh in on https://github.com/n0-computer/iroh/pull/2656/files#r1741278810, putting it on `PathState` means there is already a notion of "liveness" in several properties on the `DirectAddrInfo` (probably `last_alive` is fine for this usecase).  They may need some cleaning up, but I think this is vastly preferable to adding another notion of "liveness" that needs to be coordinated, "liveness" is already is sufficiently messy and adding yet another uncoordinated version of it with a vague meaning does not appeal to me.

So my proposal would be to keep `Source` internal for now (and hopefully forever?) and add a `PathState::is_local_network: bool`, and expose that as `DirectAddrInfo::is_local_network`.  At some point in the future `PathState::is_local_network` could become cleverder and actually store the network of the interface on which this was discovered which can then be compared with the currently attached networks.  But that feels like pushing too many features onto this PR :).  For now I think `is_local_network` with `last_alive` is fine and gives users what they want I think.

iroh-net/src/discovery.rs
```diff
@@ -95,6 +100,96 @@ pub trait Discovery: std::fmt::Debug + Send + Sync {
     ) -> Option<BoxStream<Result<DiscoveryItem>>> {
         None
     }
+
+    /// Subscribe to all addresses that get discovered.
```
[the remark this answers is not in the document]
flub: FWIW: at the end of this review I think I'm fine with keeping the subscribe method, but I do think it should only carry unsolicited `DiscoveryItem`s, so that most discovery mechanisms can just return `None`.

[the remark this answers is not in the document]
ramfox: Yes, this is a great question and not one I have a straight forward answer to.

I'm actually intrigued by your suggestion: only worry about adding events on a `subscribe` stream that were not explicitly asked for by calling `resolve`. I agree, this would make things much simpler for nodes that only discover "on command".

My question is: is this too weird? What I mean is, let's say you are writing a new discovery service, does this flexibility to report discovery in multiple ways make it confusing, or does this enable different "flavours" of discovery in a good way?

The more I think about it, the more I'm thinking it could be good if we give enough explanations and guidance, which I would try to do in the docs.

iroh-net/src/magicsock/node_map/node_state.rs
ramfox: First, let me understand what you mean for `is_local_network`. Does 
 this return "hey this is an address that is within the private network iP address range"? Or is this "we learned about this address via local discovery"? 

I ask because, if a node is in our local network, it might have advertised a public IP address. And, on the flip side, a node may have advertised a private network address, but it we might have discovered it some other way, so we don't know if this node is really dial-able from that address. 

My main reaction tho is that I disagree about hiding the `Source` & tucking this information down into the address level. I think there can be valid reasons you want to know where you learned about a node and how long ago. It's possible that folks might want to know how we learned about an *address* and how long ago, but I actually don't think that's the user's business 😂 , I think it's ours.

This is a strong opinion weakly held 😂 but I'm going to try it on for size right now:

**I think users should only care about node ids and we care about everything underneath.**

What I mean by that is, there is no useful reason (beyond debugging) for a user to know that a node has a private addr vs a public addr or that a specific address was discovered in a specific way. They should just trust that we pick the best one when they ask us to connect to another node.

In contrast to that, for the specific use case that prompted this PR, a user wanted a list of **node ids** that were discovered locally, and expected to use that to dial & we take care of the rest.

The only other thing I stand behind an application wanting to know is the connection type, if they are sending over a relay for a long chunk of time & that it might cost them more money than they want to spend.

That's my initial reaction to wanting to expose `Source` for the node id, rather than keep this information at the address level.

Thoughts?

iroh-net/src/endpoint.rs
```diff
@@ -753,6 +758,21 @@ impl Endpoint {
         self.msock.list_remote_infos().into_iter()
     }
 
+    /// Return all nodes discovered in the last `duration`.
+    pub fn recently_discovered_nodes(
```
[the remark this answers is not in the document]
ramfox: We can, this was just for convenience, since it surfaced what users were asking for in clean way.

iroh-net/src/discovery.rs
```diff
@@ -148,6 +239,7 @@ where
     }
 }
 
+#[async_trait]
```
dignifiedquire, line 242: I don't think we need this, we can just do `fn foo() -> impl Future` I think

iroh-net/src/discovery.rs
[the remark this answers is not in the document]
flub: The original design was that `resolve` would give you a stream of possible answers, once you established a connection you'd drop the stream to signal that you don't need any more answers.  I think part of the reasoning here was that maybe one discovery mechanism would yield a working result and then the others can stop expensive work.  

The new requirement is to support discovery mechanisms that just unilaterally push node information to the endpoint.  I think I'm in favour of clearly making them for different purposes.  If we clearly document that nodes discovered for `resolve` SHOULD not be pushed to the subscribers then this should not be that weird.  I think it will make the reason to have both these mechanisms much clearer.

So yes, even 24h later I'm still in favour of this.

iroh-net/src/endpoint.rs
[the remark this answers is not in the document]
flub: I'm kind of cautious about increasing the API surface.  I'd rather document this clearly in the local_swarm_discovery module docs that you can do this via `remote_info_iter().filter()` than add a new API for every thing a user asked.

It doesn't sit right with me to add a top-level API so prominently for something that's only available if you have configured this one optional component in the right way (which would have to be documented clearly here as well).

iroh-net/src/magicsock/node_map/node_state.rs
flub: > First, let me understand what you mean for `is_local_network`. Does
> this return "hey this is an address that is within the private network iP address range"? Or is this "we learned about this address via local discovery"?

I'm afraid neither!  It means that we **think** the direct address which we have for the peer is on a network to which we are also directly attached.  That could also happen on a non-private address range.  It could also happen without discovery.

> I ask because, if a node is in our local network, it might have advertised a public IP address. And, on the flip side, a node may have advertised a private network address, but it we might have discovered it some other way, so we don't know if this node is really dial-able from that address.

Good catch!  I remember thinking about this earlier and thought that it is an attribute that should be added to the `DiscoveryItem` and `PathState` to indicate that we really think this is a local network rather than something that may just be on a similar network elsewhere.  The one on `PathState` is what we're discussing here, but it needs to be passed through by `DiscoveryItem`.

> **I think users should only care about node ids and we care about everything underneath.**

Oh, I like this opinion!  I was just accepting the status quo that the users see all the addresses and working with that.  But I **love** giving the users less knowledge and API surface!  So I'd gladly take away all the `DirectAddress`es we give them currently.

So I don't think these two opinions of ours are incompatible here.  You think users should only care about NodeIds, so let's only give them this on the level of the NodeId.  I think they should not care about the particular source something was discovered with, but care about the property of "this node is a local neighbour of mine".  So let's expose it in that way (clearly my first naming attempt of `is_local_network` does not yet express this).

iroh-net/src/discovery.rs
ramfox: LMAO YES TY

iroh-net/src/magicsock/node_map/node_state.rs
ramfox: Would this look more like having a method on remoteinfo that returns `true` if we think this node is a neighbor?

`is_local_node`?

I don't fully understand how we know this, though. I was basing my logic on "this was discovered locally within X amount of time, so it's likely this is local"

Am I missing something?

iroh-net/src/discovery.rs
```diff
@@ -74,6 +75,7 @@ const SOURCE_NAME: &str = "discovery";
 /// refresh, it should start its own task.
 ///
 /// [`RelayUrl`]: crate::relay::RelayUrl
+#[async_trait]
```
rklaehn, line 78: Can we just use

```rust
fn foo() -> impl Future<...>
```

in the trait (this is allowed now)

and do
```rust
async fn foo() {
```

in the implementation?

I don't see much need for the async_trait macro since RPITIT (return position impl trait in trait) has been stabilized.

iroh-net/src/discovery.rs
```diff
@@ -95,6 +97,95 @@ pub trait Discovery: std::fmt::Debug + Send + Sync {
     ) -> Option<BoxStream<Result<DiscoveryItem>>> {
         None
     }
+
+    /// Subscribe to all addresses that get discovered.
+    async fn subscribe(&self) -> Option<BoxStream<(NodeId, DiscoveryItem)>> {
```
rklaehn, line 102: This would be just

```rust
fn subscribe(&self) -> impl Future<Output = Option<BoxStream<(NodeId, DiscoveryItem)>>>
```

I think you can use async even without async_trait now. But the downside is that you can't specify additional bonds like `Send`. If you use the verbose impl Future... syntax you *can* specify additional bonds.

iroh-net/src/discovery.rs
ramfox: Okay, I've tried a few iterations of this.

When converting `subscribe` to use `fn subscribe(&self) -> impl Future`
I get an error saying that we cannot convert the discovery trait into an Object

And when I tried instead to use `BoxFuture` in the signature, I was getting errors attempting to return `None`. I tried returning `Box::pin(FutureOption::None)`, but that also gave me errors.

Can I get a hand here?

Appreciate any help.

rklaehn: Why wasn't this a problem when using the async_trait macro? Because async_trait always just uses Box<dyn Future<Output = T>> under the hood. That is one of the main criticisms of the async_trait crate and why people were so happy about RPITIT.

About the next compile error: the fn now needs to return a Future<Option<Stream<T>>>, so the return type would have to be something like `Box::pin(future::ready(None))`.

But this signature is starting to become a bit annoying. I think the reason we returned Option<Stream> instead of Stream was that it would be easy and cheap to implement the trait with resolve disabled, by just returning None. No boxing, no nothing. Just None.

If you need to do something async when creating the stream, can't you already do this when constructing the stream? TBH I don't see the reason why this fn needs to be async.

rklaehn: OK, to make this more clear: the most obvious signature for resolve would have been just this:

```rust
fn resolve(&self, node_id: NodeId) -> BoxStream<Result<DiscoveryItem>>
```

But then if you don't want to implement resolve, like a publish-only discovery impl, you would have to return `Box::pin(futures::stream::empty())` which seemed like an unnecessary box. That is why it returns an option. None just stands for the empty stream.

If you want to do some async setup before creating the stream, you can either use this pattern:

```rust
let res = async move {
  let foo = setup().await;
  make_stream(foo) // returns an impl Stream<Result<DiscoveryItem>>
}.flatten_stream()
Some(Box::pin(res))
```

or use genawaiter. So I am not sure what a signature of BoxFuture<Option<BoxStream>> buys you over just BoxStream or Option<BoxStream>.

iroh-net/src/discovery.rs
```diff
@@ -165,7 +261,19 @@ impl Discovery for ConcurrentDiscovery {
...
+
+    async fn subscribe(&self) -> Option<BoxStream<(NodeId, DiscoveryItem)>> {
+        let mut streams = vec![];
+        for service in self.services.iter() {
+            if let Some(stream) = service.subscribe().await {
+                streams.push(stream)
+            }
+        }
+
+        let streams = futures_buffered::MergeBounded::from_iter(streams);
```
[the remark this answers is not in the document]
matheus23: `MergeBounded` is bounded in the amount of concurrent streams it contains, yeah.
There is [a vague comment about this in `MergeUnbounded`](https://docs.rs/futures-buffered/latest/futures_buffered/struct.MergeUnbounded.html).

`MergeBounded` is just a small wrapper around `FuturesUnorderedBounded`, and is the thing with the limited capacity.
We're using `MergeBounded`'s `FromIterator` implementation here, which uses `FuturesUnorderedBounded::from_iter`, which [sets the capacity to the size of the iterator](https://docs.rs/futures-buffered/latest/src/futures_buffered/futures_unordered_bounded.rs.html#315).
I think this means that if you were to `streams.push` here immediately before polling the stream, that'd panic.
But because we don't do this we're fine. And if we were polling a bit before pushing (making some streams "finish" first), we'd be fine to push more work into it.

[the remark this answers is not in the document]
matheus23: All of this said, this is not an actual PR change. `Merge` was used before this PR, and that simply got deprecated and replaced by `MergeBounded`.

iroh-net/Cargo.toml
```diff
@@ -16,6 +16,7 @@ rust-version = "1.76"
 workspace = true
 
 [dependencies]
+async-trait = "0.1.77"
```
matheus23, line 19: I think we don't need this anymore, so let's not forget removing it :)

matheus23 (changes requested):

iroh-net/Cargo.toml
ramfox: bless

matheus23, line 19: It's still here :eyes:

matheus23 (changes requested):

iroh-net/Cargo.toml
```diff
@@ -171,7 +170,3 @@ name = "connect-unreliable"
 [[example]]
 name = "dht_discovery"
 required-features = ["discovery-pkarr-dht"]
-
-[[example]]
-name = "locally-discovered-nodes"
-required-features = ["discovery-local-network"]
```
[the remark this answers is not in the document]
ramfox: that shouldn't have been removed, i'm not sure what happened there.

andrewdavidmackenzie:
You could consider using cargo udep in CI to detect unused dependencies, like the async-trait exchange above

matheus23 (changes requested):
Can you update the `futures-buffered` bounds in `iroh-net` to version `0.2.8`, please?
I'm sorry for the many nits :sweat_smile: 

As far as I can tell, that's the first version that has the `MergeBounded`/`MergeUnbounded` types.
If one would add iroh to a project that has `futures-buffered` at version 0.2.6, then cargo won't install a newer version, and instead iroh-net will fail to compile with:
```rs
error[E0433]: failed to resolve: could not find `MergeBounded` in `futures_buffered`
   --> iroh-net/src/discovery.rs:186:41
    |
186 |         let streams = futures_buffered::MergeBounded::from_iter(streams);
    |                                         ^^^^^^^^^^^^ could not find `MergeBounded` in `futures_buffered`

error[E0433]: failed to resolve: could not find `MergeBounded` in `futures_buffered`
   --> iroh-net/src/discovery.rs:198:41
    |
198 |         let streams = futures_buffered::MergeBounded::from_iter(streams);
    |                                         ^^^^^^^^^^^^ could not find `MergeBounded` in `futures_buffered`

For more information about this error, try `rustc --explain E0433`.
error: could not compile `iroh-net` (lib) due to 2 previous errors
```

We don't hit this error in this PR, because we have `futures-buffered` version 0.2.8 in our `Cargo.lock` in this repo.

iroh-net/src/magicsock/node_map.rs
```diff
@@ -85,11 +86,9 @@ enum NodeStateKey {
 }
 
 /// Source for a new node.
-///
-/// This is used for debugging purposes.
-#[derive(strum::Display, Debug)]
+#[derive(Serialize, Deserialize, strum::Display, Debug, Clone, Eq, PartialEq, Hash)]
 #[strum(serialize_all = "kebab-case")]
-pub(crate) enum Source {
+pub enum Source {
```
flub, line 91: Now that this is pub it definitely needs more docs.

* Is this a source for a node or a source for a network path to a node?  (Genuine question, I'm not actually sure and would have to carefully re-read this whole PR!  Maybe it's both!)
* What is "new" about this node? 
* What is considered a source?  Sounds silly but a sentence or two saying that is how we know about the existence of a node is appropriate to me.  Most docs should try and stand on themselves a bit without having to understand the entire system.
* Where might I encounter this?  Seems like a reference to (custom discovery mechanisms as well as `Endpoint::add_node_addr_with_source` would not be amiss.

I also chased down what `Saved` means as this was surprising to me - iroh-net does not interact with the filesystem.  It turns out it is used for any `NodeAddr` added using `iroh_net::endpoint::Builder::known_nodes`.  I don't know what to suggest here, but "saved" seems hardly accurate.  Maybe `App` would be more suitable, but that leads to...

Let's chase down why we have `App` and `NamedApp`!  `App` seems to come directly from `Endpoint::add_node_addr`, that's a pretty prominent API.  `NamedApp` comes from `Endpoint::add_node_addr_with_source`.  I think this relation should be described in the items here because otherwise someone looking at the sources here is just as confused about it as I am.

I despair at myself for caring about these things.  The rabbit holes won't end.

divagant-martian: `Saved` is a remnant from when persistence was inside iroh-net, and should def be removed. About `App`, I see your concerns but I'm not sure how to best solve it. This is akin to `Unknown`. except that it's not entirely unknown, it was manually added via the api, in contrast to automatically added via discovery, for example. Maybe based on this we can come up with a better name

iroh-net/src/magicsock/node_map/node_state.rs
```diff
@@ -222,17 +224,26 @@ impl NodeState {
...
-                last_alive: path_state
-                    .last_alive()
-                    .map(|instant| now.duration_since(instant)),
+            .map(|(addr, path_state)| {
+                let mut sources: Vec<(Source, Duration)> = path_state
+                    .sources
+                    .iter()
+                    .map(|(source, instant)| (source.clone(), now.duration_since(*instant)))
+                    .collect();
+                sources.sort_by(|a, b| b.1.cmp(&a.1));
```
divagant-martian, line 233: this is such a small thing to complain, but here we go:

I assume the sources are sorted because it's nicer to show them that way in the cli. However, this line of code belongs to the hot path, so I argue this should not happen here. The information to do this is already returned and the caller can sort them if they want, in their most-likely not-hot path. In general this order is not used, since it's used for example for filtering. 

I won't die on this hill, but I'd like to hear opinions on the matter

iroh-net/src/magicsock/node_map/node_state.rs
```diff
@@ -1233,6 +1281,10 @@ pub struct DirectAddrInfo {
     ///
     /// [`Endpoint::add_node_addr`]: crate::endpoint::Endpoint::add_node_addr
     pub last_alive: Option<Duration>,
+    /// Sources is a list of [`Source`] and [`Duration`]s, keeping track of all the ways we have learned about this path
+    /// We keep track of only the latest [`Duration`] for each [`Source`], keeping the size of the list of sources down to one entry per type of source.
```
divagant-martian, line 1285: Since this is not a struct we keep track of, but generate instantaneously on call, I think this wording needs to be adjusted. For example

`Duration is calculated as the time passed from when this source was last recorded and the instant this object was created`

This should also help with clarity as requested by @flub

iroh-net/src/magicsock/node_map/node_state.rs
```diff
@@ -1233,6 +1281,10 @@ pub struct DirectAddrInfo {
...
+    /// The sources are sorted from oldest to most recent.
```
divagant-martian, line 1286: comment above about why I would prefer to not do this

iroh-net/src/magicsock/node_map/node_state.rs
```diff
@@ -1310,6 +1362,24 @@ impl RemoteInfo {
     pub fn has_send_address(&self) -> bool {
         self.relay_url.is_some() || !self.addrs.is_empty()
     }
+
+    /// Merges and deduplicates the [`Source`] and the [`Duration`] since they were discovered for each direct address, returning a list of the most recent [`Source`]s.
```
divagant-martian, line 1366: Lots of these comments are full line width, this one alone sits at 168 column width. In the past I have noticed @flub has a 70 or 80 (can't remember) column width setup and I have a 100 one. Both these mean (as far as I remember) a 69/79/99 setup. 

I digress. My point is: we don't have a standard for comment width and enforce one is probably a low priority, high effort, low return endeavor. But even then, most of this comments fall under "extremely long" category for me, and would appreciate having them be 70/80/90/100 width at most

iroh-net/src/discovery/local_swarm_discovery.rs
```diff
@@ -2,7 +2,36 @@
 //!
 //! This allows you to use an mdns-like swarm discovery service to find address information about nodes that are on your local network, no relay or outside internet needed.
 //! See the [`swarm-discovery`](https://crates.io/crates/swarm-discovery) crate for more details.
-
+//!
+//! If you have [`LocalSwarmDiscovery`] enabled, you can get a list of the locally discovered nodes by filtering a list of `RemoteInfo`s.
```
divagant-martian, line 6: Since this comment might be relevant for future code as well, I remember @flub having a preference for avoiding the `we` and `you` phrasing in code. An alternative could simply be

```suggestion
//! When [`LocalSwarmDiscovery`] is enabled, it's possible to get a list of the locally discovered nodes by filtering a list of `RemoteInfo`s.
```

iroh-net/src/magicsock/node_map/node_state.rs
divagant-martian: authoritative argument for 100 width
https://github.com/rust-lang/rust/tree/HEAD/src/doc/style-guide/src

flub: heh.  I must admit I never tweaked how comments are formatted by my editor.  I realise they're shorter than most people, since most people use the 100-char limit.  But I also admit that I secretly often have only space for 92 characters in my editor so I kind of like the shorter comments formatting...

I agree that enforcing a standard on comment width is too much effort unless someone writes it into rustfmt.  I do think the 100-char max line-width should apply though.

As for doc comments, I usually refer to the docs style guide which says the first line should be a short summary.  It doesn't give any limits but again I take that to mean the code formatting rules of a single line, so max 100 characters including indentation.

iroh-net/src/discovery/local_swarm_discovery.rs
flub: Backing this up with an actual style-guide... style guide says "you" is fine :smile:, but "we", "our" or "us" is not:  https://developers.google.com/style/person

But it's true that I also try and avoid "you".  Maybe that is over the top though.

iroh-net/src/magicsock/node_map.rs
flub: I think it's fine to keep `App`, but acknowledge in it's docs that `NamedApp` also exists and vice-versa.

iroh-net/src/discovery/local_swarm_discovery.rs
```diff
@@ -264,25 +331,29 @@ impl Discovery for LocalSwarmDiscovery {
...
-        });
-        let stream = tokio_stream::wrappers::ReceiverStream::new(recv);
-        Some(Box::pin(stream))
+            tokio_stream::wrappers::ReceiverStream::new(recv)
+        };
+        Some(Box::pin(stream.flatten_stream()))
     }
 
     fn publish(&self, info: &AddrInfo) {
+        tracing::info!("PUBLISHING aka replacing `addrs`");
```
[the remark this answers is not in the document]
ramfox: this was added for testing and i never removed it

thanks for catching

iroh-net/src/discovery/local_swarm_discovery.rs
```diff
@@ -81,7 +116,22 @@ impl LocalSwarmDiscovery {
             let mut timeouts = JoinSet::new();
             loop {
                 trace!(?node_addrs, "LocalSwarmDiscovery Service loop tick");
-                let msg = match recv.recv().await {
+                let msg = tokio::select! {
+                    msg = recv.recv() => {
+                        msg
+                    }
+                    Ok(Some(addrs))= addrs_change.next_value_async() => {
+                        tracing::info!(?addrs, "LocalSwarmDiscovery address changed");
```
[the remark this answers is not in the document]
ramfox: thanks for catching, i changed this to find more easily when looking at logs but it should be `trace`

iroh-net/src/discovery/local_swarm_discovery.rs
```diff
@@ -28,7 +59,7 @@ use crate::{
 const N0_LOCAL_SWARM: &str = "iroh.local.swarm";
 
 /// Provenance string
-const PROVENANCE: &str = "local.swarm.discovery";
+const PROVENANCE: &str = "local";
```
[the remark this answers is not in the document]
ramfox: After reviewing the other discovery services they all had short broad names like "relay" "mainline", "pkarr" etc. But, I don't hate "local.swarm.discovery", since "local" is arguably more broad than the others. I'll change it back.

I initially wanted to `pub` this and do the same for each discovery service. Namely, have a pub const called something like `SOURCE_NAME` so that people could do this trick for each discovery service and so that it would be clearer for a user how to search for each discovery source BUT the dht discovery service has two different sources, so I couldn't use `SOURCE_NAME` for each. Because it required more thought to make sure I wouldn't be breaking the public API a bunch of times, I decided not to do it rn & to try to come up with another solution.

But, I don't hate the idea of just doing it with just this source, since it has, so far, been the only one that users have asked for. I'll change it to `SOURCE_NAME`, add documentation, and make it public. If others complain to us, then we can adjust it in the future.

What do you think?

iroh-net/examples/locally-discovered-nodes.rs
```diff
@@ -0,0 +1,66 @@
...
+use iroh_net::{
+    discovery::local_swarm_discovery::LocalSwarmDiscovery, endpoint::Source, key::SecretKey,
+    Endpoint,
+};
+use std::time::Duration;
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    tracing_subscriber::fmt::init();
+    println!("locally discovered nodes example!\n");
```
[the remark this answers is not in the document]
ramfox: I understand the sentiment, but I don't think it makes sense to expect someone running the example to know they need to set `RUST_LOG` in order to get output.

iroh-net/src/discovery/local_swarm_discovery.rs
```diff
@@ -297,6 +368,13 @@ mod tests {
         use testresult::TestResult;
 
         #[tokio::test]
+        async fn local_swarm_discovery_smoke() -> TestResult {
+            // need to ensure that these tests run one after the other, otherwise
+            // they interfere with each other
+            test_local_swarm_discovery().await?;
+            test_subscribe().await
```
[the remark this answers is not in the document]
ramfox: ahhh, okay well, tbh these tests should not longer conflict at all regardless, due to the changes in `subscribe` since I initially wrote the tests.

passes fine locally to have these separated, so lets see if CI is mad about it for some reason.

iroh-net/src/magicsock/node_map/node_state.rs
```diff
@@ -1233,6 +1281,10 @@ pub struct DirectAddrInfo {
     ///
     /// [`Endpoint::add_node_addr`]: crate::endpoint::Endpoint::add_node_addr
     pub last_alive: Option<Duration>,
+    /// Sources is a list of [`Source`] and [`Duration`]s, keeping track of all the ways we have learned about this path
+    /// We keep track of only the latest [`Duration`] for each [`Source`], keeping the size of the list of sources down to one entry per type of source.
+    /// The sources are sorted from oldest to most recent.
+    pub sources: Vec<(Source, Duration)>,
```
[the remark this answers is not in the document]
ramfox: Yea...vec was left over from a previous iteration. It's now a `HashMap<Source, Duration>`, mirroring what it is on the `PathState`

iroh-net/src/magicsock/node_map/node_state.rs
ramfox: removed!

iroh-net/examples/locally-discovered-nodes.rs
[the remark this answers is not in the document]
divagant-martian: oh yeah that's fair. We could always set the debug level directly, but it's a good argument to keep it like this

iroh-net/src/discovery/local_swarm_discovery.rs
```diff
@@ -81,6 +86,36 @@ enum Message {
...
+        let mut clean_up = vec![];
+        for (i, subscriber) in self.0.iter().enumerate() {
+            // assume subscriber was dropped
+            if (subscriber.send((node_id, item.clone())).await).is_err() {
+                clean_up.push(i);
+            }
+        }
+        for i in clean_up {
+            self.0.swap_remove(i);
+        }
```
divagant-martian, line 115: this will panic if any of the removed elements is among the last ones and more than their index is removed.

[playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021)

divagant-martian: I think that for this to work you would need to iterate the indexes in reverse

ramfox: nice catch, thank you.

reversing will work. in the worse case scenario, where we remove every single element in reverse order, every element will just pop off the end of the list. ty ty

iroh-net/src/magicsock/node_map.rs
```diff
@@ -84,22 +85,57 @@ enum NodeStateKey {
...
-    NamedApp { name: &'static str },
+    NamedApp {
+        /// The name of the application that added the node
+        name: String,
+    },
+}
+
+impl Source {
+    /// Returns true if the [`Source::NamedApp`] is the same as the given string.
+    pub fn is_named(&self, named_app: &str) -> bool {
```
flub, line 133: nothing uses this api or am i missing something?  Also, this is kind of a really weird API...

I must admit I spent more than 5 minutes in the example of the local-swarm-discovery trying to figure out if you can write that `if let` statement nicer.

You can do slightly nicer with a match by putting the `if` condition on the match arm, but not much.  It was rather disappointing.  If you make the `name` field a `&'static str` you can do a pretty nice match and if-let, but with the giant caveat you can't use the constant for the service name anymore because that's not a literal.

Anyway, that's a lot of words just to say that I don't think this method is the answer.  Can we remove it?

matheus23 (approved):

matheus23:
:white_check_mark: as in "happy from my perspective". But ofc would welcome more :white_check_mark:s from diva and flub :)

iroh-net/src/magicsock/node_map/node_state.rs
```diff
@@ -1233,6 +1277,13 @@ pub struct DirectAddrInfo {
     ///
     /// [`Endpoint::add_node_addr`]: crate::endpoint::Endpoint::add_node_addr
     pub last_alive: Option<Duration>,
+    /// A [`HashMap`] of [`Source`]s to [`Duration`]s.
+    ///
+    /// Duration is calculated as the time passed from when this source was last recorded and
+    /// the instant this object was created.
```
flub, line 1283:
```suggestion
    /// The [`Duration`] indicates the elapsed time since this source last
    /// recorded this address.
```
Somehow I find that easier to parse. The other descriptions also do not explicitly say this is elapsed time from when this object is returned.

iroh-net/src/magicsock/node_map/node_state.rs
```diff
@@ -1233,6 +1277,13 @@ pub struct DirectAddrInfo {
...
+    ///
+    /// Only the smallest [`Duration`] for each [`Source`] is kept.
```
flub, line 1285:
```suggestion
    /// The [`Duration`] will always indicate the most recent time the source
    /// recorded this address.
```

iroh-net/src/magicsock/node_map/node_state.rs
```diff
@@ -1310,6 +1361,27 @@ impl RemoteInfo {
     pub fn has_send_address(&self) -> bool {
         self.relay_url.is_some() || !self.addrs.is_empty()
     }
+
+    /// Returns a deduplicated list of [`Source`]s merged from all address in the [`RemoteInfo`].
+    ///
+    /// Deduplication is on the (`Source`, `Duration`) tuple, so you will get multiple [`Source`]s
+    /// for each `Source` variant, if those addresses were discovered from the same [`Source`]
+    /// at different times.
```
flub, line 1369:
```suggestion
    /// Deduplication is on the (`Source`, `Duration`) tuple, so you will get multiple [`Source`]s
    /// for each `Source` variant, if different addresses were discovered from the same [`Source`]
    /// at different times.
```

iroh-net/src/magicsock/node_map/node_state.rs
```diff
@@ -1310,6 +1361,27 @@ impl RemoteInfo {
...
+    ///
+    /// The list is sorted from least to most recent [`Source`].
```
flub, line 1371: It is sorted only by duration and not by source at all?  Or am I reading that wrong?

flub (commented):
lgtm

iroh-net/src/discovery/local_swarm_discovery.rs
```diff
@@ -39,14 +74,46 @@ pub struct LocalSwarmDiscovery {
...
     Discovery(String, Peer),
-    SendAddrs(NodeId, mpsc::Sender<Result<DiscoveryItem>>),
-    ChangeLocalAddrs(AddrInfo),
+    Resolve(NodeId, mpsc::Sender<Result<DiscoveryItem>>),
     Timeout(NodeId, usize),
+    Subscribe(mpsc::Sender<(NodeId, DiscoveryItem)>),
+}
+
+/// Manages the list of subscribers that are subscribed to this discovery service.
+struct Subscribers(Vec<mpsc::Sender<(NodeId, DiscoveryItem)>>);
```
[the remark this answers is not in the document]
ramfox: its only required if the struct is pub, but yes i will be adding that ty!

iroh-net/src/magicsock/node_map/node_state.rs
ramfox: yes, only by duration.

iroh-net/src/discovery.rs
```diff
@@ -95,6 +91,28 @@ pub trait Discovery: std::fmt::Debug + Send + Sync {
...
         None
     }
+
+    /// Subscribe to all addresses that get *passively* discovered.
+    ///
+    /// An implementation may choose to defer subscribing to the stream until the
+    /// stream is actually polled. To avoid missing discovered nodes, poll the
+    /// stream as soon as possible.
+    ///
+    /// Any discovery systems that only discover when explicitly resolving a
```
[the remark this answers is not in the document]
ramfox: original grammar is right "systems discover" vs "systems discovers"

It's confusing cause "discovery" seems singular, but that's actually just the adjective for the noun "systems".

iroh-net/src/discovery.rs
```diff
@@ -95,6 +91,28 @@ pub trait Discovery: std::fmt::Debug + Send + Sync {
...
+    /// addresses that are discovered by calling `resolve` should NOT be added
+    /// to the `subscribe` stream.
+    ///
+    /// Discovery systems that are capable of receiving information about [`NodeId`]s
+    /// and their [`AddrInfo`]s without explicitly calling `resolve`, i.e.,
+    /// systems that do "passive" discovery, should implement this method.
+    ///
+    /// The [`crate::endpoint::Endpoint`] will `subscribe` to the discovery system
+    /// and add the discovered addresses to the internal address book as they arrive
+    /// on this stream.
```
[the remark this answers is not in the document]
ramfox: good point. all items should show up on all streams, and that's how it's currently implemented in local swarm discovery. adding a note.

iroh-net/src/discovery.rs
```diff
@@ -95,6 +91,28 @@ pub trait Discovery: std::fmt::Debug + Send + Sync {
...
+    fn subscribe(&self) -> Option<BoxStream<(NodeId, DiscoveryItem)>> {
```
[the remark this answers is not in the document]
ramfox: done

divagant-martian (approved):
LGTM :tada:

PR #2659, opened 2024-08-21
fix: docker CI performance & release builds

Arqu:
## Description

- We now build the docker test CI jobs in a way that more closely resembles the release process. It should compile the binaries and build the image significantly faster now. (down from 8+ min to 2 min)
- Secrets are properly passed in to the reusable workflow so [release docker builds should no longer fail](https://github.com/n0-computer/iroh/actions/runs/10473026816/job/29005092378)

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -451,6 +451,7 @@ jobs:
         with:
           context: .
           push: false
+          load: true
           tags: n0computer/iroh-test:latest
           target: iroh
           platforms: linux/amd64
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -429,7 +429,7 @@ jobs:
         uses: mozilla-actions/sccache-action@v0.0.5
 
       - name: Build iroh
-        run: cargo build --release --workspace --all-features
+        run: cargo build --release --all-features
 
       - name: Prep bins
         run: |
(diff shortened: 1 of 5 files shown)


dignifiedquire:
looks good, no idea if this actually does the thing 😅

Arqu:
> looks good, no idea if this actually does the thing 😅

As far as I've tested, should be good. Tagged in @flub jic.

.github/workflows/ci.yml
```diff
@@ -445,12 +456,14 @@ jobs:
...
           platforms: linux/amd64
-          file: docker/Dockerfile
+          file: docker/Dockerfile.ci
 
       - name: Run Docker image & stats test
         run: |
           docker run -p 9090:9090 -p 4919:4919/udp -Pd n0computer/iroh-test:latest --rpc-addr 0.0.0.0:4919 start
-          cargo run --bin iroh -- --rpc-addr 127.0.0.1:4919 stats
+          # Give the server time to start
+          sleep 3
```
flub, line 465: I predict a future flake :(

Not sure what you can do better though, try to connect to the rpc addr in a loop with a short sleep and overall timeout of 30s?  your connect would have to fail quickly in an identifiable way for that to work though.

flub (approved):
would be nice if that 3s sleep could be done better, but i'm not going to block this because of it

Arqu:
I agree, but it would require either horrible parsing hacks that listen to the docker logs or actually implementing a /healthz route on the iroh cli.

.github/workflows/ci.yml
Arqu: If this gets flakey we probably should deal with it asap as a 3 sec startup is really not good.
Also the "old" setup just relied on `cargo run` to take it's sweet time compiling.

PR #2663, opened 2024-08-23
test(iroh-gossip): Wait for the relay to make `gossip_net_smoke` faster.

matheus23:
## Description

Just a tiny change that improves performance from ~1s -> ~0.4s on my machine.

## Breaking Changes

None

## Notes & open questions

One could argue we shouldn't make this change, because it doesn't cause that much erratic behavior. We should rather write tests that fail on "more bugs".
I'm not sure. Perhaps the gossip smoke test shouldn't be the test that we use to stress network edge cases.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-gossip/src/net.rs
+++ b/iroh-gossip/src/net.rs
@@ -840,13 +840,16 @@ mod test {
         rng: &mut rand_chacha::ChaCha12Rng,
         relay_map: RelayMap,
     ) -> anyhow::Result<Endpoint> {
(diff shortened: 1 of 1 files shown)


flub (approved):
> Perhaps the gossip smoke test shouldn't be the test that we use to stress network edge cases.

Totally, I think tests should be fast and target just the thing they care about.

PR #2664, opened 2024-08-23
feat(bench): Add `--with-relay` option to allow testing relay throughput

matheus23:
## Description

I did these changes to allow me to test relay throughput differences between the `dig/quinn11` and `main` branches.

The way I test relay throughput now is:
```sh
$ cd iroh-net/bench
$ DEV_RELAY_ONLY=true cargo run --release -- iroh --with-relay
```

Otherwise it simply allows testing iroh with a relay enabled, too, which is nice I think.

## Breaking Changes

None

## Notes & open questions

I think it's pretty obvious one could expand this to enable testing throughput not only with local relays, but also with any relay one wants to test with (e.g. testing euw1 relay for fun). I worked on this for a couple of minutes, then deemed it out of scope.

## Change checklist

- [x] Self-review.
- ~~[ ] Tests if relevant.~~

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2937,6 +2937,7 @@ dependencies = [
  "anyhow",
  "bytes",
  "clap",
+ "futures-lite 2.3.0",
  "hdrhistogram",
  "iroh-net",
  "quinn 0.10.2",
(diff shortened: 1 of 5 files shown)


Arqu (approved):
Great addition!
Think having it contained to only local relays is plenty as it's purely a technical benchmark and not a real world sim. Doing this against any hosted relay would probably make little sense.

PR #2665, opened 2024-08-23
Start adding tests for some odd paths in try_send

flub:
My NodeMap is completely empty when calling `msock_2.get_mapping_addr()` and I am stumped.

### DIFF
--- a/iroh-net/src/endpoint.rs
+++ b/iroh-net/src/endpoint.rs
@@ -294,6 +294,7 @@ impl StaticConfig {
 }
 
 /// Creates a [`quinn::ServerConfig`] with the given secret key and limits.
+// TODO: this return type can not longer be used anywhere in our public API.  Remove this.
 pub fn make_server_config(
     secret_key: &SecretKey,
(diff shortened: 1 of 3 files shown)


iroh-net/src/endpoint.rs
```diff
@@ -294,6 +294,7 @@ impl StaticConfig {
 }
 
 /// Creates a [`quinn::ServerConfig`] with the given secret key and limits.
+// TODO: this return type can not longer be used anywhere in our public API.  Remove this.
```
matheus23, line 297: Ah. Good to know. I wanted move this function into `StaticConfig::create_server_config`!
I can do so now with good conscience.

iroh-net/src/endpoint.rs
```diff
@@ -294,6 +294,7 @@ impl StaticConfig {
...
 pub fn make_server_config(
```
matheus23, line 298: Ah. Good to know. I wanted move this function into `StaticConfig::create_server_config`!
I can do so now with good conscience.

PR #2674, opened 2024-08-27
fix(iroh-willow): Remove `flume` dependency

matheus23:
## Description

Purging `flume` as a dependency from `iroh-willow`.

## Notes & open questions

Well, there's the whole "make a last send in `impl Drop`" thing.

## Change checklist

- [x] Self-review.
- ~~[ ] Tests if relevant.~~

### DIFF
--- a/iroh-willow/src/engine/actor.rs
+++ b/iroh-willow/src/engine/actor.rs
@@ -7,6 +7,7 @@ use tokio::{
     sync::{mpsc, oneshot},
     task::JoinSet,
 };
+use tokio_stream::wrappers::ReceiverStream;
 use tracing::{debug, error, error_span, trace, warn, Instrument};
 
 use crate::{
@@ -125,14 +126,14 @@ impl ActorHandle {
         namespace: NamespaceId,
         range: Range3d,
     ) -> Result<impl Stream<Item = anyhow::Result<Entry>>> {
-        let (tx, rx) = flume::bounded(1024);
+        let (tx, rx) = tokio::sync::mpsc::channel(1024);
         self.send(Input::GetEntries {
             namespace,
             reply: tx,
             range,
         })
         .await?;
-        Ok(rx.into_stream())
+        Ok(ReceiverStream::new(rx))
     }
 
     pub(crate) async fn init_session(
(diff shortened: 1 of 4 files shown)


iroh-willow/src/engine/actor.rs
```diff
@@ -201,7 +202,21 @@ impl Drop for ActorHandle {
...
+                    let inbox_tx = std::mem::replace(&mut self.inbox_tx, dumb);
+                    runtime
+                        .spawn(async move { inbox_tx.send(Input::Shutdown { reply: None }).await });
+                }
+                Err(_) => {
+                    self.inbox_tx
+                        .blocking_send(Input::Shutdown { reply: None })
+                        .ok();
+                }
+            }
```
matheus23, line 218: @Frando obviously this is quite a workaround for the fact that `blocking_send` panics if there's a runtime.

We should either move this into some central utility, or just switch to `async_channel` instead (like most other places, tbh).
But kind of good to know that an MPSC is technically good enough for us.

Frando:
Ah, sorry, I force-pushed the willow-rpc branch. 
Your commits apply cleanly on top of the force-pushed changes though:
https://github.com/n0-computer/iroh/compare/willow-rpc...Frando/willow-flume-purge?expand=1
so you can just push Frando/willow-flume-purge to your branch (didn't want to mess with your branch further).

My current working with the various willow branches was:
* No rebases/force pushes on the `willow` branch. Merge main into it regularily.
* Occasionally rebase feature branches against latest willow branch
* 
Maybe should stop the latter and instead merge willow into the feature branches, now that I'm not the only one working on it :D

iroh-willow/src/engine/actor.rs
Frando: Yeah. I don't have a better solution OTOH. @rklaehn fought this hard as well.

matheus23:
> Ah, sorry, I force-pushed the willow-rpc branch.

Don't worry about it! I was just lazy and based the branch on top of `willow-rpc` instead of `willow`, which would've been the sane choice.

I changed the base to `willow` now and cherry-picked the changes. It all applies cleanly as far as I can tell.

Frando (approved):

PR #2677, opened 2024-08-28
Simplify handling of no sending

flub:
We had two paths that handled not sending a transmit at all.  This meant the 2nd path was never triggered.  This instead removes the early-returns making it easier to reason about the paths taken.  Now only an error is logged if there was nothing sent.

### DIFF
--- a/iroh-net/src/magicsock.rs
+++ b/iroh-net/src/magicsock.rs
@@ -553,9 +553,8 @@ impl MagicSock {
                             send_relay = ?relay_url,
                             "sent transmit",
                         );
-                    }
-                    if !relay_sent && !udp_sent && !pings_sent {
-                        // Returning Ok here means we let QUIC handle a timeout for a lots
+                    } else if !pings_sent {
+                        // Returning Ok here means we let QUIC handle a timeout for a lost
                         // packet, same would happen if we returned any errors.  The
                         // philosophy of quinn-udp is that a UDP connection could come back
                         // at any time so these errors should be treated as transient and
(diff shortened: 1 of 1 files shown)


flub:
spellcheck error is in upstream branch. ignoring it here

iroh-net/src/magicsock.rs
```diff
@@ -563,26 +546,30 @@ impl MagicSock {
...
+                } else {
+                    if relay_sent || udp_sent {
+                        trace!(
+                            node = %node_id.fmt_short(),
+                            send_udp = ?udp_addr,
+                            send_relay = ?relay_url,
+                            "sent transmit",
+                        );
+                    }
+                    if !relay_sent && !udp_sent && !pings_sent {
```
matheus23, line 559:
```suggestion
                    } else if !pings_sent {
```
Reads a little clearer to me

iroh-net/src/magicsock.rs
```diff
@@ -563,26 +546,30 @@ impl MagicSock {
...
+                        // Returning Ok here means we let QUIC handle a timeout for a lots
```
matheus23, line 560:
```suggestion
                        // Returning Ok here means we let QUIC handle a timeout for a lost
```

matheus23 (approved):
Okay, so after staring at this code for a bit, it looks like it doesn't actually return any different value (`Ok` vs. `Err(WouldBlock)`) for different inputs, is that right?
So it's a refactor, right?
Also it changes the logs a bit, and those changes I really appreciate.

Also have two small nits.

PR #2678, opened 2024-08-28
deps(iroh-gossip): Do not depend directly on Quinn

flub:
## Description

iroh-gossip has a direct dependency on Quinn but does not need it. Remove this.

## Breaking Changes

none

## Notes & open questions

Please do not merge before #2595 is merged.  I'm trying to not modify that PR any further.

## Change checklist

- [x] Self-review.
- ~~[ ] Tests if relevant.~~

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2795,7 +2795,6 @@ dependencies = [
  "iroh-blake3",
  "iroh-metrics",
  "iroh-net",
- "iroh-quinn",
  "iroh-test",
  "postcard",
  "rand",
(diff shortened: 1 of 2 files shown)


matheus23 (approved):
Yeah - I'm really confused.
I swear this is another change I did during self-review at some point, but got lost somehow.
The reason the dependency here used to exist is because not enough quinn types were exposed in `iroh-net`, specifically error types.
We added those re-exports in iroh-net, now iroh-gossip doesn't have to depend on quinn itself anymore.

I think this may be a merge thing... because in this PR you didn't have to change the iroh-net & iroh-gossip code (which I had to do the last time).

Anyhow.

PR #2681, opened 2024-08-29
feat(iroh-willow): Event subscriptions

Frando:
## Change checklist

- [ ] Self-review.
- [ ] Documentation updates following the [style guide](https://rust-lang.github.io/rfcs/1574-more-api-documentation-conventions.html#appendix-a-full-conventions-text), if relevant.
- [ ] Tests if relevant.
- [ ] All breaking changes documented.

PR #2682, opened 2024-08-29
feat(iroh-willow): Event subscriptions

Frando:
## Description

* Moves event subscription from a wrapper of the store into the `EntryStorage` trait
* Adds persistent subscriptions, with interface - mostly - from https://github.com/earthstar-project/willow-rs/pull/21/
* Exposes subscriptions over RPC
* Adds client methods for subscriptions
* Adds a smoke test

## Notes & open questions

Should look into remaining differences between our trait and willow-rs with regard to event subscriptions, and see if anything from here should be proposed for change in the willow-rs trait. Notably:
* `subscribe_area`  takes `SubscribeOpts` not `QueryOpts`, with option to ignore ingested event for some session id, and to receive *only* Ingested event. This lets me replace the previous impl easily in the WGPS impl, and it makes sense IMO.

Has a few TODOs on optimizing things.

Will need much more tests.

Replaces #2681 (had wrong branch)

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


matheus23:
I think I've played with this enough in tauri-todos to confirm the subscription bits are working. Still need to take a look at the actual code, but honestly, that'll take a bit until I warm up to the codebase generally.

iroh-willow/src/engine/actor.rs
```diff
@@ -480,6 +531,44 @@ impl<S: Storage> Actor<S> {
...
+            } => {
+                let store = self.store.clone();
+                self.tasks.spawn_local(async move {
+                    let mut stream = store.entries().subscribe_area(namespace, area, params);
+                    while let Some(event) = stream.next().await {
+                        if sender.send(event).await.is_err() {
+                            break;
+                        }
+                    }
+                });
```
matheus23, line 548: This kind of forwarding of events seems slightly weird to me. Although I guess it's either this or more invasive changes to the `trait Storage`/`trait EntryStorage` (allowing you to just give it a sender instead of it returning a stream).

iroh-willow/src/proto/grouping.rs
```diff
@@ -93,6 +95,7 @@ impl AreaExt for Area {
 /// A single point in the 3D range space.
 ///
 /// I.e. an entry.
+// TODO: Upstream to willow-rs.
```
matheus23, line 98: Yes these are really good TODOs. Also need this on `PathExt` and `EntryExt` in the `data_model.rs` file.

iroh/tests/spaces.rs
```diff
@@ -132,3 +136,87 @@ async fn spaces_smoke() -> Result<()> {
 
     Ok(())
 }
+
```
matheus23, line 139: This test fails on this branch, but works in the `willow` branch.
It's waiting on `syncs.complete_all().await`, which never completes (neither with `SessionMode::ReconcileOnce` nor `SessionMode::Continuous`).
The test succeeds if we share the ticket using `RestrictArea::None` instead of `RestrictArea::Restrict(...)`.
```suggestion
#[tokio::test]
async fn test_regression_restricted_area_sync() -> testresult::TestResult {
    iroh_test::logging::setup_multithreaded();

    const TIMEOUT: std::time::Duration = std::time::Duration::from_secs(2);

    let (alfie_addr, alfie) = spawn_node().await;
    let (betty_addr, betty) = spawn_node().await;
    info!("alfie is {}", alfie_addr.node_id.fmt_short());
    info!("betty is {}", betty_addr.node_id.fmt_short());

    let alfie_user = alfie.spaces().create_user().await?;
    let betty_user = betty.spaces().create_user().await?;

    let alfie_space = alfie
        .spaces()
        .create(NamespaceKind::Owned, alfie_user)
        .await?;

    let space_ticket = alfie_space
        .share(
            betty_user,
            AccessMode::Write,
            // RestrictArea::None, // succeeds with this
            RestrictArea::Restrict(Area::new_subspace(betty_user)),
        )
        .await?;

    let (betty_space, syncs) = betty
        .spaces()
        .import_and_sync(space_ticket, SessionMode::ReconcileOnce)
        .await?;

    let completion = tokio::time::timeout(TIMEOUT, syncs.complete_all()).await?;
    println!("Completed syncs: {completion:#?}");

    let stream = betty_space.get_many(Range3d::new_full()).await?;
    let entries: Vec<_> = stream.try_collect().await?;
    println!("{entries:#?}");

    Ok(())
}

```

matheus23: I've also seen `Space::sync_once` just hang when two nodes try to sync that have never exchanged capabilities.

Frando: I think the test failed to pass because this branch was missing a commit from main: #2683 
Or at least the test passes just fine for me after rebasing this branch on willow!

Frando: > I've also seen Space::sync_once just hang when two nodes try to sync that have never exchanged capabilities.

I think this is currently expected behavior: PAI waits forever for matching capabilities to be submitted. We need to close our side after we have submitted everything I think. Aljoscha made some recent modifications to WGPS to support this properly (I still have to check them out)

iroh-willow/src/engine/actor.rs
Frando: I agree that this is not ideal. We do similar things already currently in the RPC boundary for old iroh-docs.
Maybe it makes sense for the subscribe methods on the store trait to take in a sender instead of returning a stream.
This brings up the question though which trait it should be generic over: Sink? ufotofu::Consumer? a custom trait that has only `async fn send(item: Event)`?

Frando: Added a comment for now.

iroh/tests/spaces.rs
Frando: The test passes reliably for me.

PR #2688, opened 2024-09-02
fix(iroh-blobs): Timeout based on correct `max_write_duration` option

matheus23:
## Description

This fixes a small copy and paste bug that was introduced in #2622, here: https://github.com/n0-computer/iroh/pull/2622/files?diff=unified&w=0#diff-bbe46ca272a808bb1c94619ab3264586a47a6ec6e0f7cd5a63dba5e99099241bL1521-R1526 (You need to load the diff for the `fs.rs` file) 

## Breaking Changes

None

## Notes & open questions

Tests would be nice, but I'm looking at other things right now, so I'd prefer to merge this first :S

## Change checklist

- [x] Self-review.
- ~~[ ] Tests if relevant.~~


rklaehn (approved):
Yeah, this seems correct 😊

PR #2690, opened 2024-09-02
fix(iroh-blobs): Remove debugging logs & more cleanup

matheus23:
## Description

More stuff to clean up I came across while trying to debug an issue.
Also, this makes it so that errors from `iroh_blobs::get::db::get_blob` that happen *before* the `get_conn` step get printed.

## Breaking Changes

None.

## Notes & open questions

I decided to just remove the tracing logs in `downloader/progress.rs`. They never really helped us surface the bug, they mostly polluted the logs, to be honest.
The only thing they surfaced was that we didn't properly clean up `TransferState::progress_id_to_blob` when there's an error in `iroh_blobs::get::db::get_blob` that causes us to not send the `DownloadProgress::Done` event.

## Change checklist

- [x] Self-review.
- ~~[ ] Tests if relevant.~~

### DIFF
--- a/iroh-blobs/src/downloader.rs
+++ b/iroh-blobs/src/downloader.rs
@@ -717,7 +717,6 @@ impl<G: Getter<Connection = D::Connection>, D: Dialer> Service<G, D> {
                 entry.get_mut().intents.insert(intent_id, intent_handlers);
             }
             hash_map::Entry::Vacant(entry) => {
-                tracing::warn!("is new, queue");
                 let progress_sender = self.progress_tracker.track(
                     kind,
                     intent_handlers
--- a/iroh-blobs/src/downloader.rs
+++ b/iroh-blobs/src/downloader.rs
@@ -727,7 +727,9 @@ impl<G: Getter<Connection = D::Connection>, D: Dialer> Service<G, D> {
                 );
 
                 let get_state = match self.getter.get(kind, progress_sender.clone()).await {
-                    Err(_err) => {
+                    Err(err) => {
+                        // This prints a "FailureAction" which is somewhat weird, but that's all we get here.
+                        tracing::error!(?err, "failed queuing new download");
                         self.finalize_download(
                             kind,
                             [(intent_id, intent_handlers)].into(),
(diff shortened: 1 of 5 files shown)


iroh-blobs/src/downloader.rs
```diff
@@ -728,7 +727,9 @@ impl<G: Getter<Connection = D::Connection>, D: Dialer> Service<G, D> {
                 );
 
                 let get_state = match self.getter.get(kind, progress_sender.clone()).await {
-                    Err(_err) => {
+                    Err(err) => {
+                        // This prints a "FailureAction" which is somewhat weird, but that's all we get here.
+                        tracing::error!(?err, "failed queuing new download");
```
divagant-martian, line 732: we already return an error `DownloadError::DownloadFailed` which admittedly does not say a lot, but the error case is handled, so I honestly do not see what do we gain from this addition

iroh-blobs/src/downloader.rs
```diff
@@ -717,7 +717,6 @@ impl<G: Getter<Connection = D::Connection>, D: Dialer> Service<G, D> {
                 entry.get_mut().intents.insert(intent_id, intent_handlers);
             }
             hash_map::Entry::Vacant(entry) => {
-                tracing::warn!("is new, queue");
```
divagant-martian, line 720: uhhh this is ugly

iroh-blobs/src/downloader.rs
matheus23: The idea is to log the error cause.

matheus23:
@divagant-martian did you intentionally enable auto-merge? You didn't approve the PR :sweat_smile:

divagant-martian:
I did. I also wanted to know what were you looking for by logging that error. It's still better than before so let's go ahead

divagant-martian (approved):

matheus23:
> what were you looking for by logging that error

There's a [thread in discord](https://discord.com/channels/949724860232392765/1279395918952534086) that has some details.

More specifically, any errors in `iroh_blobs::get::db::get_blob` *before* its call to `co.get_conn().await` get swallowed and just reported as `DownloadFailed`.
I suspected that Pasha's issues might have occurred in that part of the code, so wanted to log the actual errors.

I hope I remember the part of the code correctly - following the whole call structure took some time. But as far as I remember failures end up as this `FailureAction` error type returned from `self.getter.get` (in the place in this PR's diff).

PR #2694, opened 2024-09-04
feat: allow to bind to a specific address

### LINKED ISSUE #2565: Feature request: set bind address
Currently I can set bind port in builder. And it will auto bind to all interfaces. But it will be better if i can manually set one/some interfaces to bind. For example, I will run this service on a machine connected to 2 networks, one is public and one is private. I don't want to expose the port to public network.

dignifiedquire:
## Description

Allows to specify the IPv4 and IPv6 addresses where iroh should bind to.

Closes #2565

## Breaking Changes

- changed 
  - `iroh_net::endpoint::Endpoint::bind` now takes  no arguments
- removed
  -  `iroh::node::Builder::bind_port`
- added
  -  `iroh_net::endpoint::Builder::bind_addr_v4`
  -  `iroh_net::endpoint::Builder::bind_addr_v6`
  -  `iroh::node::Builder::bind_addr_v4`
  -  `iroh::node::Builder::bind_addr_v6`
  -  `iroh::node::Builder::bind_random_port`

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/examples/connect.rs
+++ b/iroh-net/examples/connect.rs
@@ -50,7 +50,7 @@ async fn main() -> anyhow::Result<()> {
         // If you want to experiment with relaying using your own relay server, you must pass in the same custom relay url to both the `listen` code AND the `connect` code
         .relay_mode(RelayMode::Default)
         // You can choose an address to bind to, but passing in `None` will bind the socket to a random available port
-        .bind(None, None)
+        .bind()
         .await?;
 
     let me = endpoint.node_id();
--- a/iroh-net/examples/connect.rs
+++ b/iroh-net/examples/connect.rs
@@ -49,8 +49,8 @@ async fn main() -> anyhow::Result<()> {
         // Use `RelayMode::Disable` to disable holepunching and relaying over HTTPS
         // If you want to experiment with relaying using your own relay server, you must pass in the same custom relay url to both the `listen` code AND the `connect` code
         .relay_mode(RelayMode::Default)
-        // You can choose a port to bind to, but passing in `0` will bind the socket to a random available port
-        .bind(0)
+        // You can choose an addr to bind to, but passing in `None` will bind the socket to a random available port
+        .bind(None, None)
         .await?;
 
     let me = endpoint.node_id();
--- a/iroh-net/examples/connect.rs
+++ b/iroh-net/examples/connect.rs
@@ -49,7 +49,7 @@ async fn main() -> anyhow::Result<()> {
         // Use `RelayMode::Disable` to disable holepunching and relaying over HTTPS
         // If you want to experiment with relaying using your own relay server, you must pass in the same custom relay url to both the `listen` code AND the `connect` code
         .relay_mode(RelayMode::Default)
-        // You can choose an addr to bind to, but passing in `None` will bind the socket to a random available port
+        // You can choose an address to bind to, but passing in `None` will bind the socket to a random available port
         .bind(None, None)
         .await?;
 
--- a/iroh-net/src/endpoint.rs
+++ b/iroh-net/src/endpoint.rs
@@ -710,7 +710,7 @@ impl Endpoint {
     ///
     /// # let rt = tokio::runtime::Builder::new_current_thread().enable_all().build().unwrap();
     /// # rt.block_on(async move {
-    /// let mep =  Endpoint::builder().bind(None, None).await.unwrap();
+    /// let mep =  Endpoint::builder().bind().await.unwrap();
     /// let _addrs = mep.direct_addresses().next().await;
     /// # });
     /// ```
--- a/iroh-net/src/endpoint.rs
+++ b/iroh-net/src/endpoint.rs
@@ -111,7 +111,7 @@ impl Builder {
 
     // # The final constructor that everyone needs.
 
-    /// Binds the magic endpoint on the specified socket address.
+    /// Binds the magic endpoint.
     pub async fn bind(self) -> Result<Endpoint> {
         let relay_map = self.relay_mode.relay_map();
         let secret_key = self.secret_key.unwrap_or_else(SecretKey::generate);
--- a/iroh-net/src/endpoint.rs
+++ b/iroh-net/src/endpoint.rs
@@ -119,7 +119,7 @@ impl Builder {
     ///
     /// The *addr_v6* is the address that should be bound locally on IPv6.
     ///
-    /// You can pass `None` to let the operating system choose a free port for you, and bind to `::1`.
+    /// You can pass `None` to let the operating system choose a free port for you, and bind to `[::]`.
     pub async fn bind(self) -> Result<Endpoint> {
         let relay_map = self.relay_mode.relay_map();
         let secret_key = self.secret_key.unwrap_or_else(SecretKey::generate);
@@ -163,7 +163,7 @@ impl Builder {
     ///
     /// Setting the port to `0` will use a random port.
     ///
-    /// By default will use `::1:0` to bind to.
+    /// By default will use `[::]:0` to bind to.
     pub fn bind_addr_v6(mut self, addr: SocketAddrV6) -> Self {
         self.addr_v6.replace(addr);
         self
--- a/iroh-net/src/endpoint.rs
+++ b/iroh-net/src/endpoint.rs
@@ -690,7 +690,7 @@ impl Endpoint {
     ///
     /// # let rt = tokio::runtime::Builder::new_current_thread().enable_all().build().unwrap();
     /// # rt.block_on(async move {
-    /// let mep =  Endpoint::builder().bind(0).await.unwrap();
+    /// let mep =  Endpoint::builder().bind(None, None).await.unwrap();
     /// let _addrs = mep.direct_addresses().next().await;
     /// # });
     /// ```
--- a/iroh-net/src/endpoint.rs
+++ b/iroh-net/src/endpoint.rs
@@ -112,14 +112,6 @@ impl Builder {
     // # The final constructor that everyone needs.
 
     /// Binds the magic endpoint on the specified socket address.
-    ///
-    /// The *addr_v4* is the address that should be bound locally on IPv4.
-    ///
-    /// You can pass `None` to let the operating system choose a free port for you, and bind to `0.0.0.0`.
-    ///
-    /// The *addr_v6* is the address that should be bound locally on IPv6.
-    ///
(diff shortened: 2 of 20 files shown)


iroh-net/examples/connect.rs
```diff
@@ -49,8 +49,8 @@ async fn main() -> anyhow::Result<()> {
         // Use `RelayMode::Disable` to disable holepunching and relaying over HTTPS
         // If you want to experiment with relaying using your own relay server, you must pass in the same custom relay url to both the `listen` code AND the `connect` code
         .relay_mode(RelayMode::Default)
-        // You can choose a port to bind to, but passing in `0` will bind the socket to a random available port
-        .bind(0)
+        // You can choose an addr to bind to, but passing in `None` will bind the socket to a random available port
```
ramfox, line 52: nit: makes it the same as in the other connect example.

```suggestion
        // You can choose an address to bind to, but passing in `None` will bind the socket to a random available port
```

ramfox (approved):
Looks good! one small comment nit

iroh-net/src/endpoint.rs
```diff
@@ -109,12 +109,18 @@ impl Builder {
...
+    /// You can pass `None` to let the operating system choose a free port for you, and bind to `0.0.0.0`.
+    ///
+    /// The *addr_v6* is the address that should be bound locally on IPv6.
+    ///
+    /// You can pass `None` to let the operating system choose a free port for you, and bind to `::1`.
+    pub async fn bind(
+        self,
+        addr_v4: Option<SocketAddrV4>,
+        addr_v6: Option<SocketAddrV6>,
+    ) -> Result<Endpoint> {
```
flub, line 123: Maybe an `Option` here is a bit odd?  E.g. I can't request to only bind on IPv4 right now because `None` does mean "os chooses" which is a bit surprising.

We also have the slightly odd thing that we will iterate over ports until one is selected, we should at least document that for now.

I'm wondering if a nicer API could be to add new builder methods and make the bind call not take any argument:
```
builder
    .bind_ipv4(Option<SocketAddrV4>)
    .bind_ipv6(Option<SocketAddrV6>)
    .build()
```
This way `None` can mean, don't bind (though we don't currently support not binding IPv4, I think we can just error for now and leave that for the future).  And the default can be the current default but without having to write it out in full.

I would be tempted to also fail if we can't bind the specific port requested if one is used here.  The only question is then how to handle the way the `iroh::node::Node` currently binds: it requests a specific port but is happy to fall back to `0` if it is not available.  Maybe a `.fallback_port(u16)` to the builder?  Not great.  Maybe allow passing in bound sockets directly?  But that's not great as we need a specifically set-up socket.

iroh-net/src/endpoint.rs
```diff
@@ -109,12 +113,14 @@ impl Builder {
...
-    /// You can pass `0` to let the operating system choose a free port for you.
+    /// The *addr_v4* is the address that should be bound locally on IPv4.
     ///
-    /// NOTE: This will be improved soon to add support for binding on specific addresses.
-    pub async fn bind(self, bind_port: u16) -> Result<Endpoint> {
+    /// You can pass `None` to let the operating system choose a free port for you, and bind to `0.0.0.0`.
+    ///
+    /// The *addr_v6* is the address that should be bound locally on IPv6.
+    ///
+    /// You can pass `None` to let the operating system choose a free port for you, and bind to `::1`.
```
flub, line 122: this whole doc comment still needs to be updated.

iroh-net/src/endpoint.rs
```diff
@@ -142,6 +149,26 @@ impl Builder {
...
+    pub fn bind_addr_v4(mut self, addr: SocketAddrV4) -> Self {
+        self.addr_v4.replace(addr);
+        self
+    }
+
+    /// Set the IPv6 bind address.
+    ///
+    /// Setting the port to `0` will use a random port.
+    ///
+    /// By default will use `::1:0` to bind to.
```
flub, line 166: `::1` can not and should not be the default, it should be `::`.  `::1` is the loopback address, `::` is the unspecified address.

btw, the way to spell this is `[::]:0`

iroh-net/src/magicsock.rs
```diff
@@ -94,9 +94,14 @@ const HEARTBEAT_INTERVAL: Duration = Duration::from_secs(5);
...
-    /// The port to listen on.
-    /// Zero means to pick one automatically.
-    pub(crate) port: u16,
+    /// The IPv4 address to listen on.
+    ///
+    /// If set to `None` it will choose a random port and listen on `0.0.0.0`.
+    pub(crate) addr_v4: Option<SocketAddrV4>,
+    /// The IPv6 address to listen on.
+    ///
+    /// If set to `None` it will choose a random port and listen on `::1`.
```
flub, line 103:
```suggestion
    /// If set to `None` it will choose a random port and listen on `[::]:0`.
```

iroh-net/src/magicsock.rs
```diff
@@ -94,9 +94,14 @@ const HEARTBEAT_INTERVAL: Duration = Duration::from_secs(5);
 /// Contains options for `MagicSock::listen`.
 #[derive(derive_more::Debug)]
 pub(crate) struct Options {
-    /// The port to listen on.
-    /// Zero means to pick one automatically.
-    pub(crate) port: u16,
+    /// The IPv4 address to listen on.
+    ///
+    /// If set to `None` it will choose a random port and listen on `0.0.0.0`.
```
flub, line 99:
```suggestion
    /// If set to `None` it will choose a random port and listen on `0.0.0.0:0`.
```

iroh-net/src/endpoint.rs
```diff
@@ -142,6 +149,26 @@ impl Builder {
 
     // # The very common methods everyone basically needs.
 
+    /// Set the IPv4 bind address.
+    ///
+    /// Setting the port to `0` will use a random port.
+    ///
+    /// By default will use `0.0.0.0:0` to bind to.
```
flub, line 156: Could we add a line describing that we fall back to port 0 if the specified port is already in use?  Because this is kind of surprisingly different from classical behaviour.  Same for IPv6.

iroh/src/node/builder.rs
```diff
@@ -422,11 +426,22 @@ where
...
+        self
+    }
+
+    /// Binds the node service to a specific socket IPv6 address.
+    pub fn bind_addr_v6(mut self, addr: SocketAddrV6) -> Self {
+        self.addr_v6 = addr;
+        self
+    }
+
+    /// Use a random port for both IPv4 and IPv6.
```
flub, line 441:
```suggestion
    /// Use a random port for both IPv4 and IPv6.
    ///
    /// This is a convenience function useful when you do not need a specific port 
    /// and want to avoid conflicts when running multiple instances, e.g. in tests.
    ///
    /// This overrides the ports of the socket addresses provided by [`Builder::bind_addr_v4`]
    /// and [`Builder::bind_addr_v6`].  By default both of those bind to the
    /// unspecified address, which would result in `0.0.0.0:0` and `[::]:0` as bind
    /// addresses unless they are changed.
```

But to be honest, given the behaviour of the normal port binding.  Is there much value to having this?  Because in tests we will already just fall back to port `0` if another test is already using it.

iroh/src/node/builder.rs
```diff
@@ -422,11 +426,22 @@ where
         self
     }
 
-    /// Binds the node service to a different socket.
-    ///
-    /// By default it binds to `127.0.0.1:11204`.
-    pub fn bind_port(mut self, port: u16) -> Self {
-        self.bind_port.replace(port);
+    /// Binds the node service to a specific socket IPv4 address.
```
flub, line 429:
```suggestion
    /// Binds the node service to a specific socket IPv4 address.
    ///
    /// If the port used in the bind address is not free, a random different
    /// port will be used.
```

flub (commented):
Agree that these builder methods make sense as they are.

iroh-net/src/endpoint.rs
dignifiedquire: ah right, I can never remember these

iroh/src/node/builder.rs
dignifiedquire: the iroh port is used by default, not `0` for `iroh` itself

dignifiedquire: updated docs

iroh-net/src/endpoint.rs
dignifiedquire: updated docs some more

iroh-net/src/endpoint.rs
```diff
@@ -109,12 +113,14 @@ impl Builder {
...
-    /// You can pass `0` to let the operating system choose a free port for you.
+    /// The *addr_v4* is the address that should be bound locally on IPv4.
     ///
-    /// NOTE: This will be improved soon to add support for binding on specific addresses.
-    pub async fn bind(self, bind_port: u16) -> Result<Endpoint> {
+    /// You can pass `None` to let the operating system choose a free port for you, and bind to `0.0.0.0`.
+    ///
+    /// The *addr_v6* is the address that should be bound locally on IPv6.
+    ///
+    /// You can pass `None` to let the operating system choose a free port for you, and bind to `[::]`.
```
flub, line 122: Can't this whole block now be `Binds the magic endpoint.`?  All the details are from an earlier version still?

iroh-net/src/endpoint.rs
```diff
@@ -142,6 +149,28 @@ impl Builder {
 
     // # The very common methods everyone basically needs.
 
+    /// Set the IPv4 bind address.
```
flub, line 152:
```suggestion
    /// Sets the IPv4 bind address.
```

iroh-net/src/endpoint.rs
```diff
@@ -142,6 +149,28 @@ impl Builder {
...
+    /// Setting the port to `0` will use a random port.
+    /// If the port specified is already in use, it will fallback to choosing a random port.
+    ///
+    /// By default will use `0.0.0.0:0` to bind to.
+    pub fn bind_addr_v4(mut self, addr: SocketAddrV4) -> Self {
+        self.addr_v4.replace(addr);
+        self
+    }
+
+    /// Set the IPv6 bind address.
```
flub, line 163:
```suggestion
    /// Sets the IPv6 bind address.
```

flub (approved):

iroh-net/src/endpoint.rs
dignifiedquire: yes, removed

Issue #2660, opened 2024-08-21
make sure everything in the cli is documented


divagant-martian:
not sure if we can deny missing docs on a specific struct, but this would be the way to go. Docs become help text in the console and it's missing in a couple of places

divagant-martian:
This is a really good issue, since it might allow you to explore the public API without writing a project on top of iroh, for example. The CLI structure is defined [here](https://github.com/n0-computer/iroh/blob/a029d89a2b9cb1d04eba054960114801b755b582/iroh-cli/src/commands.rs#L29) and everything below this structure (the types of the fields, recursively) should have docs. This does not need to happen in a single big PR, small but complete PRs are encouraged and welcome. Most likely it would look like one PR per file in https://github.com/n0-computer/iroh/tree/main/iroh-cli/src/commands so feel free to work on any (or all) if you want

divagant-martian:
Thanks a lot for your contributions!. I did not review the prs being on holidays but being now all merged I trust this is thoroughly addressed. Closing

PR #2710, opened 2024-09-05
docs(iroh-cli): udpate `doctor` command documentation

palozano:
## Description

Update documentation to `doctor` command submodule.

This partially addresses #2660, and belongs to a series of PRs to complete it.

## Notes & open questions

Imports have been formated to be grouped.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-cli/src/commands/doctor.rs
+++ b/iroh-cli/src/commands/doctor.rs
@@ -1,24 +1,19 @@
 //! Tool to get information about the current network environment of a node,
 //! and to test connectivity to specific other nodes.
-use std::{
-    collections::HashMap,
-    io,
-    net::SocketAddr,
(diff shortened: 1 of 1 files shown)


matheus23 (changes requested):
Thanks for this, can you use e.g. "Update**s** the progress bar" instead of "Update the progress bar" in general though? We've only recently adopted a new documentation style guide, so we're not super consistent at this yet, but we should try moving towards using it more.

> The summary line should be written in third person singular present indicative form. Basically, this means write ‘Returns’ instead of ‘Return’.

I know it's easy to skip this sentence in the guide (I had to look for it a bit just now, too), but if you could update the PRs to use this style, it'd be great :)

PR #2715, opened 2024-09-06
refactor(iroh): Remove custom impl of `SharedAbortingJoinHandle`

matheus23:
## Description

Removes our custom implementation of `SharedAbortingJoinHandle` in favor of a combination of
- `futures_util::future::Shared`
- `futures_util::future::MapErr` and
- `tokio_util::task::AbortOnDropHandle`

## Breaking Changes

Should all be internal, so none.

## Notes & open questions

We need to turn the `JoinError` to a `String`, because `JoinError` is `!Send`, but `String` is.
This is exactly the same behavior as what our custom `SharedAbortingJoinHandle` was doing before.

## Change checklist

- [x] Self-review.
- ~~[ ] Tests if relevant.~~

### DIFF
--- a/iroh/src/node.rs
+++ b/iroh/src/node.rs
@@ -104,6 +104,12 @@ pub type IrohServerEndpoint = quic_rpc::transport::boxed::ServerEndpoint<
 #[derive(Debug, Clone)]
 pub struct Node<D> {
     inner: Arc<NodeInner<D>>,
+    // `Node` needs to be `Clone + Send`, and we need to `task.await` in its `shutdown()` impl.
+    // So we need
+    // - `Shared` so we can `task.await` from all `Node` clones
+    // - `MapErr` to map the `JoinError` to a `String`, because `JoinError` is `!Send`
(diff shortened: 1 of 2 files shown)


flub (approved):

iroh/src/node.rs
```diff
@@ -106,10 +104,12 @@ pub type IrohServerEndpoint = quic_rpc::transport::boxed::ServerEndpoint<
 #[derive(Debug, Clone)]
 pub struct Node<D> {
     inner: Arc<NodeInner<D>>,
-    task: SharedAbortingJoinHandle<()>,
+    task: Shared<MapErr<AbortOnDropHandle<()>, JoinErrToStr>>,
```
flub, line 107: Maybe its worth adding a comment here why this construction exists and what magic is exactly performs.  Right now I know, but I can imagine being baffled sometime in the future.

matheus23: Agreed, will add that :)

Issue #2647, opened 2024-08-19
iroh-net: netcheck actor dropped stun response channel


dignifiedquire:
Have some reports of this happening quite often, inside docker (iroh@0.22)

```
ERROR iroh_net::netcheck::reportgen::hairpin: Hairpin actor failed: netcheck actor dropped stun response channel
at <..>/.cargo/registry/src/index.crates.io-6f17d22bba15001f/iroh-net-0.22.0/src/netcheck/reportgen/hairpin.rs:97
```

flub:
I should summarise what I know from having looked at this:

- This happens when the netcheck actor is no longer running but the hairpin actor is not running anymore.
- This itself should not happen.
- **Maybe** it could happen at shutdown
  - All the actors involved (`netcheck -> reportgen -> hairpin`) are all supposed to abort on drop.
  - I think drop order might be fine but an not entirely sure.  If say the hairpin actor is dropped while it happens to be running what happens?  It first waits until it is pending again before it is dropped?
- If this happens during normal runtime that means the netcheck actor died.
  - This should also not happen!
  - Sounds a little similar to https://github.com/fishfolk/bones/issues/428 where possibly an actor dies which should not.
  - So I'm thinking of generically improving the actors supervision structure in MagicSock, but have currently no clue either here or at the https://github.com/fishfolk/bones/issues/42 issue whether that will really help.  It might expose the actual problem better though.

flub:
Does it still happen to the original reporter?  (I assume you are a proxy)

dignifiedquire:
I don’t know, will try to find out

PR #2716, opened 2024-09-06
refactor(iroh-net): Make netcheck::Client !Clone

flub:
## Description

netcheck::Client owns the actor task and when dropped it will abort
the actor task.  Making a struct owning a task Clone means it is easy
to lose track of who should be owning a task like this.  I now believe
each task should have a clear supervisor/owner in charge of it.

This cleans up the multiple-ownership of the netcheck::Client, which
is just a small step into this direction.  Later on, for e.g. #2647,
more supervision will be added.  But small changes are good.

## Breaking Changes

None, this is all internal.

## Change checklist

- [ ] Self-review.
- ~~[ ] Tests if relevant.~~

### DIFF
--- a/iroh-net/src/netcheck.rs
+++ b/iroh-net/src/netcheck.rs
@@ -166,7 +166,7 @@ impl RelayLatencies {
 ///
 /// While running the netcheck actor expects to be passed all received stun packets using
 /// [`Client::receive_stun_packet`].
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub struct Client {
     /// Channel to send message to the [`Actor`].
     ///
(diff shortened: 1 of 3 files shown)


iroh-net/src/netcheck.rs
```diff
@@ -217,27 +217,12 @@ impl Client {
...
-    /// consuming them.
-    pub fn receive_stun_packet(&self, payload: Bytes, src: SocketAddr) {
-        if let Err(mpsc::error::TrySendError::Full(_)) = self.addr.try_send(Message::StunPacket {
-            payload,
-            from_addr: src,
-        }) {
-            inc!(NetcheckMetrics, stun_packets_dropped);
-            warn!("dropping stun packet from {}", src);
-        }
+    /// Unlike the client itself this does own the actor task, only allows sending messages
```
dignifiedquire, line 222: this reads odd

dignifiedquire (approved):

PR #2717, opened 2024-09-06
feat(iroh-base): Implement `From` & `Into` between `NodeAddr` and `NodeTicket`

matheus23:
## Description

Previously, when all you had was a `NodeTicket`, all you could get was a reference `&NodeAddr` from it to use with `Endpoint::connect`, which requires an *owned* `NodeAddr`.

This adds `From` impls for converting between `NodeTicket` and `NodeAddr`, so you can call `endpoint.connect(ticket.into())`.

## Breaking Changes

None - only additions.

## Notes & open questions

When I originally raised this, I thought of making the `node_addr` argument `impl Into<NodeAddr>`, but we have an `instrument(remote = %node_addr.node_id.fmt_sort)` annotation on that function. Due to ownership, it's hard to make that work with `impl Into<NodeAddr>`. As far as I can see there's three options:
- Make it `node_addr: impl Into<NodeAddr> + Clone`
- Split `Endpoint::connect` into two functions, a public, outer one with `node_addr: impl Into<NodeAddr>` and without an `instrument` annotation, and an inner private one with `node_addr: NodeAddr`, but with an `instrument` annotation.
- We just keep it as `node_addr: NodeAddr` with the `instrument` annotation and require users to call `.into()`.

I found the third option to be the simplest. @flub please let me know if that sounds good.

## Change checklist

- [x] Self-review.
- ~~[ ] Tests if relevant.~~ I honestly don't think we need tests for this.

### DIFF
--- a/iroh-base/src/ticket/node.rs
+++ b/iroh-base/src/ticket/node.rs
@@ -10,12 +10,17 @@ use crate::{
     ticket::{self, Ticket},
 };
 
-/// A token containing information for dialing a node.
+/// A token containing information for establishing a connection to a node.
 ///
 /// Contains
-/// - The [`NodeId`] of the node to connect to (a 32-byte ed25519 public key)
+/// - The [`NodeId`] of the node to connect to (a 32-byte ed25519 public key).
+/// - If used, the ['RelayUrl`] of on which the node can be reached.
+/// - Any *direct addresses* on which the node might be reachable.
 ///
-/// It is a single item which can be easily serialized and deserialized.
+/// This allows establishing a connection to the node in most circumstances where it is
+/// possible to do so.
+///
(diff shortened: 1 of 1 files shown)


iroh-base/src/ticket/node.rs
```diff
@@ -10,9 +10,14 @@ use crate::{
     ticket::{self, Ticket},
 };
 
-/// A token containing everything to get a file from the provider.
+/// A token containing information for dialing a node.
+///
+/// Contains
+/// - The [`NodeId`] of the node to connect to (a 32-byte ed25519 public key)
```
flub, line 16: I love adding docs, though this is a list with one item.  Did you mean to add some more things?

flub (approved):

dignifiedquire:
@flub can you take this PR over pleaes?

PR #2722, opened 2024-09-09
docs(iroh-net): Improve pkarr discovery docs

flub:
## Description

This tries to improve the pkarr discovery docs by trying to describe
enough about how it works that users do not need to go looking
elsewhere.

## Breaking Changes

none

## Notes & open questions

There are so many code overlaps and naming inconsistencies here.  But this was
part of docs friday so I'm just documenting things for now.

- Discovery mechanisms are not named particularly consistent:
- `PkarrPublisher` is probably a `PkarrRelayPublisher`
- `PkarrResolver` is also a `PkarrRelayResolver`
- `DnsDiscovery` only does resolving -> `DnsResolver`
- `DhtDiscovery` can also use relay server, but at least it does both publish and resolve
- The `DhtDiscovery` is pretty generic, it can probably be made to do both replace `PkarrResolver` and `PkarrPublisher`, maybe with custom constructors.

## Change checklist

- [x] Self-review.
- ~~[ ] Tests if relevant.~~

### DIFF
--- a/iroh-net/src/discovery/pkarr.rs
+++ b/iroh-net/src/discovery/pkarr.rs
@@ -1,10 +1,10 @@
 //! A discovery service which publishes and resolves node information using a [pkarr] relay.
 //!
-//! The Public-Key Addressable Resource Records, [pkarr], is a system which allows
-//! publishing [DNS Resource Records] owned by a particular [`SecretKey`] under a name
-//! derived from its corresponding [`PublicKey`], also known as the [`NodeId`].
-//! Additionally this pkarr Resource Record is signed using the same [`SecretKey`], ensuring
-//! authenticity of the record.
+//! Public-Key Addressable Resource Records, [pkarr], is a system which allows publishing
+//! [DNS Resource Records] owned by a particular [`SecretKey`] under a name derived from its
+//! corresponding [`PublicKey`], also known as the [`NodeId`].  Additionally this pkarr
+//! Resource Record is signed using the same [`SecretKey`], ensuring authenticity of the
+//! record.
 //!
 //! Pkarr normally stores these records on the [Mainline DHT], but also provides two bridges
 //! that do not require clients to directly interact with the DHT:
--- a/iroh-net/src/discovery/pkarr.rs
+++ b/iroh-net/src/discovery/pkarr.rs
@@ -74,12 +74,12 @@ pub mod dht;
 /// central service.  It is a reliable service to use for node discovery.
 ///
 /// [number 0]: https://n0.computer
-/// [crate documentation]: crate::discovery::pkarr
+/// [module documentation]: crate::discovery::pkarr
 pub const N0_DNS_PKARR_RELAY_PROD: &str = "https://dns.iroh.link/pkarr";
 /// The testing pkarr relay run by [number 0].
 ///
 /// This server operates similarly to [`N0_DNS_PKARR_RELAY_PROD`] but is not as reliable.
-/// It is meant for more experimental usages.
+/// It is meant for more experimental use and testing purposes.
 ///
 /// [number 0]: https://n0.computer
 pub const N0_DNS_PKARR_RELAY_STAGING: &str = "https://staging-dns.iroh.link/pkarr";
--- a/iroh-net/src/discovery/pkarr.rs
+++ b/iroh-net/src/discovery/pkarr.rs
@@ -1,6 +1,48 @@
-//! A discovery service which publishes and resolves node information to a [pkarr] relay.
+//! A discovery service which publishes and resolves node information using a [pkarr] relay.
+//!
+//! The Public-Key Addressable Resource Records, [pkarr], is a system which allows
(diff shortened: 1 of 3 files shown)


flub:
@Frando @rklaehn ping

iroh-net/src/discovery/pkarr.rs
```diff
@@ -21,27 +63,56 @@ use crate::{
...
 };
+
 #[cfg(feature = "discovery-pkarr-dht")]
 pub mod dht;
 
-/// The pkarr relay run by n0, for production.
+/// The production pkarr relay run by [number 0].
+///
+/// This server is both a pkarr relay server as well as a DNS resolver, see the [module
+/// documentation].  However it does not interact with the Mainline DHT, so is a more
```
Frando, line 73: `iroh-dns-server` *can* interact with the mainline DHT for resolving node ids not found in the local database, see [here](https://github.com/n0-computer/iroh/blob/main/iroh-dns-server/src/store.rs#L66). It does not (yet?) support publishing records to mainline DHT.

 The [config template for production](https://github.com/n0-computer/iroh/blob/main/iroh-dns-server/config.prod.toml#L16) does *not* enable the mainline support by defualt. IIRC we wanted to enable it on `dns.iroh.link` though, I do not know OTOH if that is the case. I would guess @Arqu can verify.

iroh-net/src/discovery/pkarr.rs
```diff
@@ -21,27 +63,56 @@ use crate::{
...
-/// how long to store a record. It is ignored by the iroh-dns-server as the home
-/// server keeps the records for the domain. When using the pkarr relay no DNS is
-/// involved and the setting is ignored.
+/// Default TTL for the records in the pkarr signed packet.
+///
+/// The Time To Live (TTL) tells DNS caches how long to store a record. It is ignored by the
+/// `iroh-dns-server`, e.g. as running on [`N0_DNS_PKARR_RELAY_PROD`], as the home server
+/// keeps the records for the domain. When using the pkarr relay no DNS is involved and the
+/// setting is ignored.
+// TODO(flub): huh?
```
Frando, line 93: I also cannot say what the last sentence is supposed to mean.

This paragraph should likely instead explain that this is the timeframe in which, after publishing new node addr info, other nodes might still receive the old outdated info when resolving via DNS.

iroh-net/src/discovery/pkarr.rs
```diff
@@ -21,27 +63,56 @@ use crate::{
...
 pub const DEFAULT_PKARR_TTL: u32 = 30;
 
-/// Interval in which we will republish our node info even if unchanged: 5 minutes.
+/// Interval in which to republish the node info even if unchanged: 5 minutes.
 pub const DEFAULT_REPUBLISH_INTERVAL: Duration = Duration::from_secs(60 * 5);
 
-/// Publish node info to a pkarr relay.
+/// Publisher of node discovery information to a [pkarr] relay.
+///
+/// This publisher uses HTTP to publish to node discovery information to a pkarr relay
```
Frando, line 101:
```suggestion
/// This publisher uses HTTP to publish node discovery information to a pkarr relay
```

iroh-net/src/discovery/pkarr.rs
```diff
@@ -199,28 +283,41 @@ impl PublisherService {
...
     }
 
-    /// Create a pkarr resolver which uses the [`N0_DNS_PKARR_RELAY_PROD`] pkarr relay and in testing
-    /// uses [`N0_DNS_PKARR_RELAY_STAGING`].
+    /// Creates a pkarr resolver which uses the [number 0] pkarr relay server.
+    ///
+    /// This uses the pkarr relay server operated by [number 0] at
+    /// [`N0_DNS_PKARR_RELAY_PROD`].
+    ///
+    /// When compiling for tests, i.e. when using `cfg(test)` is true, or when the
```
Frando, line 316:
```suggestion
    /// When compiling for tests, i.e. when `cfg(test)` is true, or when the
```

Frando (approved):
Thank you @flub for writing this sorely missing piece of documentation.
A couple of comments inline.

Also: I agree to your suggested renames above. And if we can combine it all into a single `PkarrDiscovery` trait with feature flags and a builder that might be even better IMO.

The current situation is very much the result of us adding piece after piece, and not giving things another high-level view I think. So refactoring this into a more coherent picture would be good.

iroh-net/src/discovery/pkarr.rs
Arqu: Im not sure, lets chat about this on the meeting today.

dignifiedquire:
@flub what's missing here to merge this?

iroh-net/src/discovery/pkarr.rs
flub: If it is currently not enabled let's leave it documented like this.  I'm not trying to change anything here, merely documenting the current state.

PR #2725, opened 2024-09-09
feat!(iroh): Vec<u8> instead of static bytes for ALPNs

dignifiedquire:
Otherwise it is not possible to dynamically assign ALPN names.

## Breaking Changes

- `ProtocolBuilder::accept` uses a `Vec<u8>` instead of `&'static [u8]`


Frando:
LGTM yes, we already cast to vec anyway once we add the ALPNs to the endpoint, so this only adds flexibility.

Frando (approved):

PR #2731, opened 2024-09-16
refactor(iroh-gossip): make use of Endpoint::direct_addresses in iroh_gossip::net

Frando:
## Description

`iroh_gossip::net::Gossip` had a method to update the direct addresses of the iroh_net::Endpoint. IIRC that dates back when the Endpoint only exposed a single, non-cloneable stream of endpoint events, which was handled in the iroh Node and thus had to be forwarded to consumers like Gossip. This was long ago and by now the Endpoint provides a capable API for this. Therefore there is no need to expose this as a method (because Gossip always contains an Endpoint).

## Breaking Changes

`Gossip::update_direct_addresses` is removed. Updating the direct addresses is now handled by `Gossip` automatically.

## Notes & open questions

To react to all updates to the endpoint's `NodeAddr`, I am watching both `Endpoint::direct_addresses` and `Endpoint::watch_home_relay`. A simpler API for this usecase would be `Endpoint::watch_node_addr`.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-gossip/src/net.rs
+++ b/iroh-gossip/src/net.rs
@@ -11,7 +11,7 @@ use futures_util::TryFutureExt;
 use iroh_metrics::inc;
 use iroh_net::{
     dialer::Dialer,
-    endpoint::{get_remote_node_id, Connection},
+    endpoint::{get_remote_node_id, Connection, DirectAddr},
     key::PublicKey,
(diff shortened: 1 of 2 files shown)


dignifiedquire:
nice

PR #2732, opened 2024-09-16
feat(iroh-net): add Endpoint::watch_node_addr

Frando:
## Description

This adds `Endpoint::watch_node_addr` which returns a stream that yields `NodeAddr`.

We have `Endpoint::direct_addresses` and `Endpoint::watch_home_relay`. In #2731 what I needed, and what I suspect apps will oftenly need, is the combination of both: Give me a stream that yields the endpoint's NodeAddr whenever it changes, either due to a change in our home relay, or due to a change in our discovered addresses.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.


flub:
Assigning to myself as @Frando is now OOO

dignifiedquire:
Closing this, as we now have the watcher based API

matheus23:
FWIW, we *do* have the watcher-based API, but it's only in #3045 where I actually introduce an API that allows waiting for changes to the whole `NodeAddr`, instead of having to consolidate changes in `direct_addresses` and `home_relay` yourself.

PR #2736, opened 2024-09-17
feat(deps): use derive_more@1.0.0 (no more beta!)

ukstv:
Apparently, derive_more is out of beta.

Please, update your deps.


dignifiedquire (approved):
thanks

PR #2744, opened 2024-09-23
refactor(iroh-net): Various logging improvements

flub:
## Description

- Attach the spans when spawning the local-swarm-discovery.  We should
  always use .instrument() when spawning.

- Clearly log the start and end of connection establishment.

- In the events.net.* events always use remote_node when showing the
  NodeId of the remote node.  These were introduced before we settled
  on the "remote node" terminology, now it makes more sense to double
  down on this as it makes the logs clearer.

## Breaking Changes

None, events.net is not considered part of the public API.

## Change checklist

- [x] Self-review.
- ~~[ ] Tests if relevant.~~

### DIFF
--- a/iroh-net/src/discovery/local_swarm_discovery.rs
+++ b/iroh-net/src/discovery/local_swarm_discovery.rs
@@ -12,7 +12,7 @@ use std::{
 use anyhow::Result;
 use derive_more::FromStr;
 use futures_lite::stream::Boxed as BoxStream;
-use tracing::{debug, error, trace, warn};
+use tracing::{debug, error, info_span, trace, warn, Instrument};
 
 use iroh_base::key::PublicKey;
 use swarm_discovery::{Discoverer, DropGuard, IpClass, Peer};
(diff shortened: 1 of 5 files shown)


iroh-net/src/discovery/local_swarm_discovery.rs
```diff
@@ -71,126 +71,130 @@ impl LocalSwarmDiscovery {
...
+                                LocalSwarmDiscovery::socketaddrs_to_addrs(addrs.direct_addresses);
+                            for addr in addrs {
+                                discovery.add(addr.0, addr.1)
+                            }
                         }
                     }
                 }
             }
-        });
+            .instrument(info_span!("swarm-discovery.actor")),
```
flub, line 196: Only this `.instrument()` was added here.  All the other code is unchanged - despite what the diff may make you wonder.

divagant-martian (approved):
lgtm. For the sake of getting such a simple change in quickly, I modified this to reduce the large confusing diff. Better to have a clean git history

flub:
> lgtm. For the sake of getting such a simple change in quickly, I modified this to reduce the large confusing diff. Better to have a clean git history

oh yes, clever move.

PR #2745, opened 2024-09-23
refactor(iroh-net): Remove PathState::recent_pong()

flub:
## Description

The field itself is also directly accessible with the same visibility
as this function.  Having two ways to get to this value is a bit
confusing.  And directly using the field makes it easier when checking
users by following the references from rust-analyzer.

## Breaking Changes

None

## Notes & open questions

None

## Change checklist

- [x] Self-review.
- ~~[ ] Tests if relevant.~~


divagant-martian (approved):
lgtm. I wonder why was added in the first place

PR #2746, opened 2024-09-24
feat(iroh-net): Log the crate version number

flub:
## Description

This adds a DEBUG-level log logging the crate version number.

## Breaking Changes

None

## Notes & open questions

This is not the first log message appearing, the MagicSock and a few
other pieces already have been logging by this time.  Also I'm not
particularly bound to this wording.  Maybe there are other things that
should be added to this line?

## Change checklist

- [x] Self-review.
- ~~[ ] Tests if relevant.~~

### DIFF
--- a/iroh-net/src/endpoint.rs
+++ b/iroh-net/src/endpoint.rs
@@ -405,7 +405,7 @@ impl Endpoint {
             Arc::new(quinn::TokioRuntime),
         )?;
         trace!("created quinn endpoint");
-
+        debug!(version = env!("CARGO_PKG_VERSION"), "iroh Endpoint created");
         Ok(Self {
             msock,
             endpoint,


iroh-net/src/endpoint.rs
```diff
@@ -405,7 +405,7 @@ impl Endpoint {
             Arc::new(quinn::TokioRuntime),
         )?;
         trace!("created quinn endpoint");
-
+        debug!(version = env!("CARGO_PKG_VERSION"), "iroh Endpoint created");
```
matheus23, line 408: Open Q: Do we still need the trace log?

matheus23 (approved):

iroh-net/src/endpoint.rs
dignifiedquire: what about commit version?

matheus23: Also I'm not sure if my linked answer would work with libraries. :thinking:

flub: > Open Q: Do we still need the trace log?

the trace logs distinguish magicsock creation from quinn endpoint creation.  i don't know right now if that's useful or not.  but it's trace so left it alone.

flub: git commit sounds useful, but I don't know how to do it - as shown by @matheus23 's responses.  happy to add it if there's a simple recipe though.

flub:
> Did you test whether the environment variable reports the version of the crate, even if it's a deep transitive dependency? I.e. I'd like to avoid a case where it reports version "0.1.0", if you compile "my-crate v 0.1.0" which depends on "iroh-net v 0.26.0"

Pretty sure that's fine: https://doc.rust-lang.org/cargo/reference/environment-variables.html

PR #2751, opened 2024-09-27
feat(iroh-cli): improve ergonomics of `iroh gossip subscribe` CLI cmd

### LINKED ISSUE #2698: gossip cli commands are unusable
subscribe is not really usable

```
> gossip subscribe test
error: invalid value 'test' for '[BOOTSTRAP]...': decoding: base32: non-zero trailing bits at 3
```

sending a message seems not to be available at all

matheus23:
## Description

Previously the gossip command could be confusing:
```sh
$ iroh gossip subscribe test --start
error: invalid value 'test' for '[BOOTSTRAP]...': decoding: base32: non-zero trailing bits at 3
```

Now, the output is more helpful:
```sh
$ iroh gossip subscribe test --start
error: the following required arguments were not provided:
  <--topic <TOPIC>|--raw-topic <RAW_TOPIC>>

Usage: iroh gossip subscribe --start <--topic <TOPIC>|--raw-topic <RAW_TOPIC>> <BOOTSTRAP>...

For more information, try '--help'.
```

And the help text also contains examples & explanation on how the command works.

## Breaking Changes

None. The command still works the same way.

## Notes & open questions

Closes #2698

Should we rename the command? When I read `gossip subscribe` initially, I thought a `gossip publish` command may be missing. But actually `gossip subscribe` does both.
Perhaps `gossip join` may be better?
But also: We want to keep the CLI names in sync with the names in the code API. Perhaps we should rename the API call to be `join` instead of `subscribe`? Or perhaps something like `open_topic`?
Anyhow - perhaps that's food for thought & another PR.

## Change checklist

- [x] Self-review.
- ~~[ ] Tests if relevant.~~


dignifiedquire (approved):
thanks, at least now one can figure out how to use it 😅

PR #2757, opened 2024-09-30
fix(ci): make netsim work on forks

Arqu:
## Description

Simply skips the troublesome step if it's not configured properly (due to missing org secrets).

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/.github/workflows/netsim_runner.yaml
+++ b/.github/workflows/netsim_runner.yaml
@@ -158,7 +158,7 @@ jobs:
         python3 reports_csv.py --metro --commit ${{ env.LAST_COMMIT_SHA }} > report_metro.txt
         python3 reports_csv.py --metro --integration --commit ${{ env.LAST_COMMIT_SHA }} > report_metro_integration.txt
     
-    - name: Dump report
+    - name: Upload report
       run: |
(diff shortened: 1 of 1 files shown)


divagant-martian (commented):
github is not allowing me to review unchanged parts, which is fair I guess, so here will have to do:

won't `curl -X POST -H "Content-Type: application/json" -H "Authorization: Bearer ${{secrets.METRO_TOKEN}}" --data "$metro_data" ${{secrets.METRO_ENDPOINT}}` fail as well?

or are the other parts that use `secrets` already handled?

Arqu:
> github is not allowing me to review unchanged parts, which is fair I guess, so here will have to do:
> 
> won't `curl -X POST -H "Content-Type: application/json" -H "Authorization: Bearer ${{secrets.METRO_TOKEN}}" --data "$metro_data" ${{secrets.METRO_ENDPOINT}}` fail as well?
> 
> or are the other parts that use `secrets` already handled?

No because only merge to main actually runs with publish stats. I guess running manually from your work with the intent to publish stats would fail, but that's not a case you should run into.

divagant-martian (approved):

PR #2758, opened 2024-09-30
fix(ci): add cleanup workflow to retain `generated-docs-preview` for only the last 25 PRs

Arqu:
## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -37,7 +37,7 @@ jobs:
         github_token: ${{ secrets.GITHUB_TOKEN }}
         publish_dir: ./target/doc/
         destination_dir: ${{ env.PREVIEW_PATH }}
-        publish_branch: generated-docs-preview
+        publish_branch: generated-docs-preview-${{ github.event.pull_request.number }}
 
     - name: Find Docs Comment
       uses: peter-evans/find-comment@v3
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -3,6 +3,10 @@ name: Docs Preview
 on:
   pull_request:
 
+# ensure job runs sequentially so pushing to the preview branch doesn't conflict
+concurrency:
+  group: ci-docs-preview
+
 env:
   IROH_FORCE_STAGING_RELAYS: "1"
 
@@ -37,7 +41,7 @@ jobs:
         github_token: ${{ secrets.GITHUB_TOKEN }}
         publish_dir: ./target/doc/
         destination_dir: ${{ env.PREVIEW_PATH }}
-        publish_branch: generated-docs-preview-${{ github.event.pull_request.number }}
+        publish_branch: generated-docs-preview
 
     - name: Find Docs Comment
       uses: peter-evans/find-comment@v3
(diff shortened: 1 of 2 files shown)


dignifiedquire (approved):

matheus23 (changes requested):
Ehm, I'm not sure if this will work. The `generated-docs-preview` branch is a "special branch" as far as github is concerned I think.

matheus23:
What exactly is the problem you're trying to solve? Some explanation in the PR description would help me review this!

Arqu:
Thanks for jumping in, I was hoping it works for any branch you set it to (dynamically).
Anyways, the crux of it is this https://github.com/n0-computer/iroh/actions/runs/11105394334/job/30851460274?pr=2756
It seems mostly to be a race between PRs to publish their own. Maybe do a softer version of and do like up to 5 retries with back off.

dignifiedquire:
- what if we switch this to a command instead of auto generating it? eg `/generate_docs`
- alternatively, maybe we just push this onto one of our CI machines and run nginx there just hosting these ourselves per branch

matheus23:
Seems to be a common issue with the action we're using to update the branch: https://github.com/peaceiris/actions-gh-pages/issues/1078

matheus23:
This was recommended as a fix to the concurrency issue: https://github.com/peaceiris/actions-gh-pages/issues/998#issuecomment-1742383037

I don't think our problem is the concurrency though. It seems the action has stopped at some point in the last month or so.

matheus23:
> Thanks for jumping in, I was hoping it works for any branch you set it to (dynamically). Anyways, the crux of it is this https://github.com/n0-computer/iroh/actions/runs/11105394334/job/30851460274?pr=2756 It seems mostly to be a race between PRs to publish their own. Maybe do a softer version of and do like up to 5 retries with back off.

Yeah actually I think this is a separate issue. If pushing to the generated-docs-preview branch were the problem, then the stuff that's *in the branch* would still be served correctly, but e.g. the 2760 PR [exists on the branch](https://github.com/n0-computer/iroh/tree/generated-docs-preview/pr/2760/docs), but [isn't served by github](https://n0-computer.github.io/iroh/pr/2760/docs/iroh/index.html) for some reason.

Arqu:
Ok, think I've figured it out. Seems like we're hitting some file size limits judging by the action outputs. I'll manually prune some of the old PRs and see if that makes things work again.

Arqu:
Yup that worked just fine now. I'll add a step to clean up the branch every now and then.

Arqu:
@matheus23 This should now also trigger a weekly job to clean up the docs branch. It does risk to remove the preview for PRs that are old but still open, but I guess if you still need or want the docs preview you can easily re-run the job to generate them.

.github/workflows/docs.yaml
```diff
@@ -3,6 +3,10 @@ name: Docs Preview
 on:
   pull_request:
```
matheus23, line 4: Can you add `workflow_dispatch` here, too? So we can trigger it manually?

matheus23 (approved):
Thank you for figuring it out! :)
Just a small nit.

Arqu:
Think this should do it, but cant be 100% sure till we merge and test it out due to the nature of workflows and permissions on GHA.

Arqu:
> Why does netsim keep adding comments? 😅 (Is that intentional?)

It's supposed to update them but something is messed up, that's what I'm going to fix next.

PR #2759, opened 2024-09-30
docs(iroh-net): Document cargo features needed for APIs

flub:
## Description

This uses the feature of docs.rs to enable showing which cargo
features need to be enabled for APIs.  This is much friendlier for
users.

## Breaking Changes

None

## Notes & open questions

I'm trying to do this for the entire workspace but for some reason am
not getting this to work on other crates yet.  This is all such a
mystery and mess.  Just get this in stable rust already and do this by
default without needing extra annotations.

I'll try and do the rest of the workspace still in a followup PR, but
at least this might make the release so we get to see if it works on
actual docs.rs as well.

## Change checklist

- [x] Self-review.
- ~~[ ] Tests if relevant.~~

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,5 +34,15 @@ incremental = false
 [workspace.lints.rust]
 missing_debug_implementations = "warn"
 
+# We use this --cfg for documenting the cargo features on which an API
+# is available.  To preview this locally use: RUSTFLAGS="--cfg
+# iroh_docsrs cargo +nightly doc --all-features".  We use our own
+# iroh_docsrs instead of the common docsrs to avoid also enabling this
+# feature in any dependencies, because some indirect dependencies
+# require a feature enabled when using `--cfg docsrs` which we can not
+# do.  To enable for a crate set `#![cfg_attr(iroh_docsrs,
+# feature(doc_cfg))]` in the crate.
+unexpected_cfgs = { level = "warn", check-cfg = ["cfg(iroh_docsrs)"] }
+
 [workspace.lints.clippy]
 unused-async = "warn"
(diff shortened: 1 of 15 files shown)


matheus23:
I wish the docs action would work :cry: 

Not sure what happened. :/ (Am looking into this)

Cargo.toml
```diff
@@ -34,5 +34,15 @@ incremental = false
...
 
+# We use this --cfg for documenting the cargo features on which an API
+# is available.  To preview this locally use: RUSTFLAGS="--cfg
+# iroh_docsrs cargo +nightly doc --all-features".  We use our own
+# iroh_docsrs instead of the common docsrs to avoid also enabling this
+# feature in any dependencies, because some indirect dependencies
+# require a feature enabled when using `--cfg docsrs` which we can not
+# do.  To enable for a crate set `#![cfg_attr(iroh_docsrs,
+# feature(doc_cfg))]` in the crate.
+unexpected_cfgs = { level = "warn", check-cfg = ["cfg(iroh_docsrs)"] }
```
dignifiedquire, line 45: don't we need to specify this feature in the individual `Cargo.toml`s?

flub:
> I wish the docs action would work 😢
> 
> Not sure what happened. :/ (Am looking into this)

Oh, this action will also need to be fixed for this so it builds with `RUSTFLAGS=--cfg iroh_docsrs`.  I guess we'll do that separately.

Cargo.toml
flub: no, it's not even allowed: they all have `lints.workspace = true` so pick it up from here.

dignifiedquire: I see, that's complicated

dignifiedquire (approved):

Issue #2337, opened 2024-06-01
Netcheck timeouts on high latency networks


link2xt:
I have been testing Delta Chat with iroh-gossip realtime channels on a network with high latency (>10s ping) and while Delta Chat worked for sending messages, p2p channels failed to connect.

Apparently netcheck has a very low timeout with no backoff:
https://github.com/n0-computer/iroh/blob/4dd69f44d62e3b671339ce586a2f7e97a47559ff/iroh-net/src/netcheck/reportgen.rs#L50-L51

Is it true that there is no way to use iroh-net when network has >5s RTT?

dignifiedquire:
The short answer is, yes. In most scenarios we are currently deploying to, latency larger than this is an indicator that the device is offline/things are broken.

dignifiedquire:
There are quite a few things that will break on networks with this kind of latency, for example most hole punching strategies will fail, because routers often keep things open for only around 5 seconds.

dignifiedquire:
this is about the port mappings inside the router, when no connection has been established yet, which are often times much shorter

re relay connection, sure in theory that can be made to work, but it will take some thinking anf work to do so, without making things worse for folks on lower latency connections, as failing fast is often times very desirable

PR #2764, opened 2024-10-01
[test] Empty PR for testing netsim

matheus23:
## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


matheus23:
Tried testing iroh with netsim configured to 5s latency, in order to test use cases like #2337
Logs are here: https://github.com/n0-computer/iroh/actions/runs/11122485757/artifacts/1999863925

Seems like it ran through fine.
The throughput was still set fairly high (1gbps), which would be unrealistic for a mobile network on a train, and 2% packet loss may or may not be super accurate, but shouldn't be *too* far off.

Arqu:
Seems like double NAT + relay works fine too https://github.com/n0-computer/iroh/actions/runs/11122850370

matheus23:
Judging from the throughput numbers, in the NAT case, did it perhaps only use the relay?

Arqu:
No, it seems fine, just that test used a 1MB transfer which doesn't leave much time for it to ramp up the throughput and is mostly time dominated by the setup of the connection.

PR #2766, opened 2024-10-01
fix(ci): netsim commenting fixes

Arqu:
## Description

Adjusts permissions, uses proper input vars & the search & replace for comments is now aligned.
Also adds examples to the list of possible executables for the job.

Note: Docs still seem to fail in this PR with the set permissions. However write-all in a public repo is very constrained so it should be perfectly fine. The GITHUB_TOKEN has it by default, however I think since we're running sub jobs etc it's not propagating correctly.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -292,6 +292,7 @@ jobs:
           command-arguments: "-Dwarnings"
 
   netsim-integration-tests:
+    permissions: write-all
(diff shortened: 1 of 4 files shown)


matheus23 (approved):
`permissions: write-all` makes me a little uneasy, but I trust your assessment of

> However write-all in a public repo is very constrained so it should be perfectly fine.

Adding the examples in seems fine, too.

PR #2772, opened 2024-10-02
feat(iroh-dns-server)!: Make http rate limit configurable

HeavenVolkoff:
Hello,

We are currently testing some new cloud features at [Spacedrive](https://github.com/spacedriveapp/spacedrive), and our implementation relies heavily on iroh. As part of this, we are deploying our own iroh-dns-server. However, since all of our backend services operate behind a reverse proxy, we noticed that the iroh-dns-server was frequently hitting its rate limit because it wasn’t aware of the proxy setup. To address this, I decided to implement a configurable rate limit for the iroh-dns-server.

## Description

This PR adds a new entry to the `iroh-dns-server` TOML file for configuring the HTTP rate limit. The new configuration allows for disabling the rate limit and also supports configuring it to use the [SmartIPKeyExtract](https://github.com/benwis/tower-governor/blob/v0.4.2/src/key_extractor.rs#L85-L119), making it compatible with reverse proxies.

## Breaking Changes

- `iroh-dns-server`'s configuration structure now has a new field allowing to choose the rate limiting algorithms.

## Notes & open questions

:)

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-dns-server/config.dev.toml
+++ b/iroh-dns-server/config.dev.toml
@@ -1,7 +1,7 @@
 [http]
 port = 8080
 bind_addr = "127.0.0.1"
-rate_limit = false
+rate_limit = "disabled"
 
 [https]
 port = 8443
--- a/iroh-dns-server/config.dev.toml
+++ b/iroh-dns-server/config.dev.toml
@@ -1,6 +1,7 @@
 [http]
 port = 8080
 bind_addr = "127.0.0.1"
+rate_limit = false
 
 [https]
 port = 8443
--- a/iroh-dns-server/config.dev.toml
+++ b/iroh-dns-server/config.dev.toml
@@ -1,7 +1,8 @@
+pkarr_put_rate_limit = "disabled"
+
 [http]
 port = 8080
 bind_addr = "127.0.0.1"
-rate_limit = "disabled"
 
 [https]
 port = 8443
--- a/iroh-dns-server/config.prod.toml
+++ b/iroh-dns-server/config.prod.toml
@@ -1,3 +1,6 @@
+[http]
+rate_limit = true
+
 [https]
 port = 443
 domains = ["irohdns.example.org"]
--- a/iroh-dns-server/config.prod.toml
+++ b/iroh-dns-server/config.prod.toml
@@ -1,5 +1,5 @@
 [http]
-rate_limit = true
+rate_limit = "smart"
 
 [https]
 port = 443
--- a/iroh-dns-server/config.prod.toml
+++ b/iroh-dns-server/config.prod.toml
@@ -1,10 +1,8 @@
-[http]
-rate_limit = "smart"
-
 [https]
 port = 443
 domains = ["irohdns.example.org"]
 cert_mode = "lets_encrypt"
+rate_limit = "smart"
 letsencrypt_prod = true
 
 [dns]
--- a/iroh-dns-server/config.prod.toml
+++ b/iroh-dns-server/config.prod.toml
@@ -1,8 +1,9 @@
+pkarr_put_rate_limit = "smart"
+
 [https]
 port = 443
 domains = ["irohdns.example.org"]
 cert_mode = "lets_encrypt"
-rate_limit = "smart"
 letsencrypt_prod = true
 
 [dns]
--- a/iroh-dns-server/src/http.rs
+++ b/iroh-dns-server/src/http.rs
@@ -1,7 +1,6 @@
 //! HTTP server part of iroh-dns-server
 
 use std::{
-    fmt,
     net::{IpAddr, Ipv4Addr, SocketAddr},
     time::Instant,
 };
@@ -17,7 +16,7 @@ use axum::{
     Router,
 };
 use iroh_metrics::{inc, inc_by};
-use serde::{de, Deserialize, Deserializer, Serialize};
+use serde::{Deserialize, Serialize};
 use tokio::{net::TcpListener, task::JoinSet};
 use tower_http::{
     cors::{self, CorsLayer},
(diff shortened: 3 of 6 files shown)


iroh-dns-server/src/http.rs
```diff
@@ -35,13 +36,61 @@ use crate::{config::Config, metrics::Metrics};
 
 pub use self::tls::CertMode;
 
+/// Config for http rate limit
+#[derive(Debug, Serialize, Clone)]
+pub enum RateLimitConfig {
+    /// Enable or disable rate limiting
+    Boolean(bool),
```
rklaehn, line 43: Can we just have two named cases for the two boolean values?

So
```rust
pub enum RateLimitConfig {
  Disabled,
  Enabled, // non-smart? what's the best way to describe the defaullt mode?
  Smart
}
```

divagant-martian: Right now `Some(Boolean(true))` and `None` have the same meaning/effect. We shouldn't create this duplicated behaviour. 

If `Boolean(true)` is kept, or _ideally_ changed to `Simple`/`Conventional`/etc then the options should go away and we should be setting the default with serde. If the option is kept then the enum should only be `Disabled` and `Smart`, but not both.

iroh-dns-server/src/http.rs
```diff
@@ -35,13 +36,61 @@ use crate::{config::Config, metrics::Metrics};
 
 pub use self::tls::CertMode;
 
+/// Config for http rate limit
+#[derive(Debug, Serialize, Clone)]
```
divagant-martian, line 40: Since serialization is not custom made, I think the serialization-deserialization roundrip might not work

iroh-dns-server/src/http.rs
```diff
@@ -35,13 +36,61 @@ use crate::{config::Config, metrics::Metrics};
 
 pub use self::tls::CertMode;
 
+/// Config for http rate limit
```
divagant-martian, line 39:
```suggestion
/// Config for http rate limit.
```

divagant-martian (changes requested):

iroh-dns-server/src/http.rs
HeavenVolkoff: Decided to drop Boolean and go with Disabled/Simple/Smart and leave the toml setting value as a string

HeavenVolkoff:
Hi, I am not sure why the CI is failing. Is this expected? Cargo deny is complaining even though I didn't add any new dependencies, so I am not sure what is going on there. Netsim test is failing making a request to a GitHub API for some reason. The check_semver I imagine is failing because my changes require me to bump a minor version, should I do that?

divagant-martian:
yes, this is not your fault! we fixed this recently, could you update your branch?

iroh-dns-server/src/http/rate_limiting.rs
```diff
@@ -1,21 +1,60 @@
...
-        // .use_headers()
-        .per_second(4)
-        .burst_size(2)
+    let mut governor_conf_builder = GovernorConfigBuilder::default();
+    // governor_conf_builder.use_headers()
+    governor_conf_builder.per_second(4);
+    governor_conf_builder.burst_size(2);
+
+    if use_smart_extractor {
+        tracing::info!("Rate limiting using smart extractor");
```
divagant-martian, line 53: could we change these two info lines for just one that includes the kind of rate limiting?
as in:
```
tracing::info!("Rate limiting enabled ({rate_limit_config})");
```

divagant-martian (approved):
looks generally good from my point of view. Waiting for @Arqu's opinion. Small changes suggested too

HeavenVolkoff:
So, I had to add a `rate_limit` property to `[https]`, because the program crashed with only `rate_limit` defined in `[http]`, as it requires a `port` property because it assumes that if you have a `[http]` block it means you want to enable the http server. So I ended up just implementing an optional `rate_limit` in `[https]` which has preference over what is configured in `[http]` in case both are defined.

iroh-dns-server/config.dev.toml
```diff
@@ -1,6 +1,7 @@
 [http]
 port = 8080
 bind_addr = "127.0.0.1"
+rate_limit = "disabled"
```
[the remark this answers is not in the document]
Arqu: No, looks pretty good. We did these type of enum based settings before, worked nice.

iroh-dns-server/config.prod.toml
```diff
@@ -1,3 +1,6 @@
+[http]
+rate_limit = "smart"
```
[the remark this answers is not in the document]
Arqu: Yeah, I'll have to follow up on our end but we generally don't use these as is in deployments.

rklaehn:
Sorry about the conflicts. We have unified the way our imports are done, should be easy to fix.

HeavenVolkoff:
> Hey, thanks for following up on this PR. After the latest changes, it seems clearer the rate limiting applies to the whole dns server and making it present in the http config AND https config might be easily misunderstood in the case where both are defined. You have mentioned a strategy but we should strive to make the meaning of options as obvious as possible.

Yeah, originally I had placed the `rate_limit` configuration at the top level, but it felt odd since it was the only option without a block. The name also started getting confusing, as it only applies to HTTP, and placing it as a global option might lead users to assume it also applies to the DNS part of the server. I agree that having `rate_limit` under both `[http]` and `[https]` isn’t ideal, though. Maybe a better alternative might be to create a new `[rate_limit]` block with an `http = '<disabled|simple|smart>'` option. This would make the configuration clearer to users, avoid repetition, and allow for future expandability if we ever need to support rate limiting for the DNS side as well.

divagant-martian:
@matheus23 suggested `pkarr_put_rate_limit` in the top level config. I find this a good option. It's verbose but extremely clear since this is the only place rate limiting applies.

divagant-martian (approved):
Very clean and nice. Ty! LGTM

PR #2774, opened 2024-10-02
feat(iroh-net)!: Allow using a NodeId directly in connect.

rklaehn:
## Description

Allow using anything can be converted to a NodeAddr when connecting. This means that we can directly connect with NodeIds, and connect_by_node_id is no longer needed.

Also deprecate connect_by_node_id since it is no longer needed.

## Breaking Changes

- iroh-net: Endpoint::connect now takes an `impl Into<NodeAddr>` instead of a `NodeAddr`
- iroh-net: Endpoint::connect_by_node_id is deprecated (to be removed next release)

## Notes & open questions

Note: Doing this means that the instrumentation of connect no longer contains the remote, which is bad I guess. Not sure how to work around this. One way would be to have connect and connect_impl - connect_impl is like the current connect and can be instrumented as before. Any other way to do this? 

Note: since there is a From<NodeTicket> for NodeAddr, you can also directly dial using a ticket, which is kinda neat.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/src/endpoint.rs
+++ b/iroh-net/src/endpoint.rs
@@ -459,7 +459,7 @@ impl Endpoint {
         alpn: &[u8],
     ) -> Result<quinn::Connection> {
         let node_addr = node_addr.into();
-        tracing::Span::current().record("remote", &node_addr.node_id.fmt_short());
+        tracing::Span::current().record("remote", node_addr.node_id.fmt_short());
         // Connecting to ourselves is not supported.
         if node_addr.node_id == self.node_id() {
             bail!(
--- a/iroh-net/src/endpoint.rs
+++ b/iroh-net/src/endpoint.rs
@@ -512,7 +512,7 @@ impl Endpoint {
     /// uses the discovery service to establish a connection to a remote node.
     #[deprecated(
         since = "0.27.0",
-        note = "Please use `connect` directly with a `NodeAddr` or `NodeId`. This feature will be removed in 0.28.0."
+        note = "Please use `connect` directly with a NodeId. This fn will be removed in 0.28.0."
     )]
     pub async fn connect_by_node_id(
         &self,
--- a/iroh-net/src/endpoint.rs
+++ b/iroh-net/src/endpoint.rs
@@ -450,8 +450,13 @@ impl Endpoint {
     /// The `alpn`, or application-level protocol identifier, is also required. The remote
     /// endpoint must support this `alpn`, otherwise the connection attempt will fail with
     /// an error.
-    #[instrument(skip_all, fields(me = %self.node_id().fmt_short(), remote = %node_addr.node_id.fmt_short(), alpn = ?String::from_utf8_lossy(alpn)))]
-    pub async fn connect(&self, node_addr: NodeAddr, alpn: &[u8]) -> Result<quinn::Connection> {
+    #[instrument(skip_all, fields(me = %self.node_id().fmt_short(), alpn = ?String::from_utf8_lossy(alpn)))]
+    pub async fn connect(
+        &self,
+        node_addr: impl Into<NodeAddr>,
+        alpn: &[u8],
+    ) -> Result<quinn::Connection> {
+        let node_addr = node_addr.into();
         // Connecting to ourselves is not supported.
         if node_addr.node_id == self.node_id() {
             bail!(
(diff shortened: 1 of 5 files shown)


dignifiedquire (approved):
nice, I like it

iroh-net/src/endpoint.rs
```diff
@@ -502,6 +507,10 @@ impl Endpoint {
     /// information being provided by either the discovery service or using
     /// [`Endpoint::add_node_addr`].  See [`Endpoint::connect`] for the details of how it
     /// uses the discovery service to establish a connection to a remote node.
+    #[deprecated(
+        since = "0.27.0",
+        note = "Please use `connect` directly with a `NodeAddr` or `NodeId`"
```
flub, line 512: can we add when we'll be removing this?

when will we be removing it?  :grin:

iroh-net/src/endpoint.rs
```diff
@@ -450,8 +450,13 @@ impl Endpoint {
     /// The `alpn`, or application-level protocol identifier, is also required. The remote
     /// endpoint must support this `alpn`, otherwise the connection attempt will fail with
     /// an error.
-    #[instrument(skip_all, fields(me = %self.node_id().fmt_short(), remote = %node_addr.node_id.fmt_short(), alpn = ?String::from_utf8_lossy(alpn)))]
-    pub async fn connect(&self, node_addr: NodeAddr, alpn: &[u8]) -> Result<quinn::Connection> {
+    #[instrument(skip_all, fields(me = %self.node_id().fmt_short(), alpn = ?String::from_utf8_lossy(alpn)))]
```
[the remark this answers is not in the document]
rklaehn: Yeah, it is clear you want it. Otherwise you wouldn't have added it in the first place. But the only way I can see to keep it in is to do as described and do a private _impl fn where the addr is already converted to a NodeAddr.

Will do that if you don't have a better idea.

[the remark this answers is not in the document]
rklaehn: Pushed an update. You can just add something to the current span it seems...

iroh-net/src/endpoint.rs
```diff
@@ -450,8 +450,13 @@ impl Endpoint {
...
+    pub async fn connect(
```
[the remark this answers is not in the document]
rklaehn: Updated the docs.

iroh-net/src/endpoint.rs
rklaehn: Added 0.28.0. If we remove it one later nobody will be super unhappy I guess...

iroh-net/src/endpoint.rs
```diff
@@ -432,10 +432,12 @@ impl Endpoint {
...
     /// Connects to a remote [`Endpoint`].
     ///
-    /// A [`NodeAddr`] is required. It must contain the [`NodeId`] to dial and may also
-    /// contain a [`RelayUrl`] and direct addresses. If direct addresses are provided, they
-    /// will be used to try and establish a direct connection without involving a relay
-    /// server.
+    /// A value that can be converted into a [`NodeAddr`] is required. This can be either a
+    /// [`NodeAddr`], a [`NodeId`] or a [`iroh_base::ticket::NodeTicket`].
+    ///
+    /// The [`NodeAddr`] must contain the [`NodeId`] to dial and may also contain a [`RelayUrl`]
```
flub, line 438: We don't use `dial` in any of the API names of `iroh-net::Endpoint` so I prefer not to use that in documentation either as that is probably confusing to users.  I usually stick to "connect", "establish a connection" or something like that.

(yeah, there's the somewhat weird dialer module which is a bit of an odd duck...)

iroh-net/src/endpoint.rs
```diff
@@ -450,8 +452,14 @@ impl Endpoint {
...
-    #[instrument(skip_all, fields(me = %self.node_id().fmt_short(), remote = %node_addr.node_id.fmt_short(), alpn = ?String::from_utf8_lossy(alpn)))]
-    pub async fn connect(&self, node_addr: NodeAddr, alpn: &[u8]) -> Result<quinn::Connection> {
+    #[instrument(skip_all, fields(me = %self.node_id().fmt_short(), alpn = ?String::from_utf8_lossy(alpn)))]
+    pub async fn connect(
+        &self,
+        node_addr: impl Into<NodeAddr>,
+        alpn: &[u8],
+    ) -> Result<quinn::Connection> {
+        let node_addr = node_addr.into();
+        tracing::Span::current().record("remote", node_addr.node_id.fmt_short());
```
flub, line 462: that's cool!  TIL

PR #2778, opened 2024-10-02
fix(cfg)!: make sure we use correct relays

Arqu:
## Description

This takes us back to only relying on the env var for forcing staging relays. However for tests it should still remain active with the test/feature combo by default.

## Breaking Changes

- `TEST_DNS_NODE_ORIGIN` is removed
- iroh now defaults to using prod relays unless `IROH_FORCE_STAGING_RELAYS` is set to a non empty value

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/src/discovery/dns.rs
+++ b/iroh-net/src/discovery/dns.rs
@@ -6,6 +6,7 @@ use futures_lite::stream::Boxed as BoxStream;
 use crate::{
     discovery::{Discovery, DiscoveryItem},
     dns::ResolverExt,
+    relay::force_staging_infra,
     Endpoint, NodeId,
 };
 
@@ -13,10 +14,6 @@ use crate::{
 pub const N0_DNS_NODE_ORIGIN_PROD: &str = "dns.iroh.link";
 /// The n0 testing DNS node origin, for testing.
 pub const N0_DNS_NODE_ORIGIN_STAGING: &str = "staging-dns.iroh.link";
-/// Testing DNS node origin, must run server from [`crate::test_utils::DnsPkarrServer`].
-#[cfg(any(test, doc))]
-#[cfg_attr(iroh_docsrs, doc(cfg(any(test, doc))))]
-pub const TEST_DNS_NODE_ORIGIN: &str = "dns.iroh.test";
 
 const DNS_STAGGERING_MS: &[u64] = &[200, 300];
 
@@ -57,25 +54,14 @@ impl DnsDiscovery {
     ///
     /// # Usage during tests
     ///
-    /// When `cfg(test)` is enabled the
-    /// [`crate::discovery::dns::TEST_DNS_NODE_ORIGIN`] is used.
-    ///
-    /// Note that the `iroh.test` domain is not integrated with the global DNS network and
-    /// thus node discovery is effectively disabled.  To use node discovery in a test use
-    /// the [`crate::test_utils::DnsPkarrServer`] in the test and configure it as a
-    /// custom discovery mechanism.
-    ///
-    /// For testing it is also possible to use the [`N0_DNS_NODE_ORIGIN_STAGING`] domain
-    /// with [`DnsDiscovery::new`].  This would then use a hosted discovery service again,
-    /// but for testing purposes.
+    /// For testing it is possible to use the [`N0_DNS_NODE_ORIGIN_STAGING`] domain
+    /// with [`DnsDiscovery::new`].  This would then use a hosted staging discovery
+    /// service for testing purposes.
     pub fn n0_dns() -> Self {
-        #[cfg(not(test))]
-        match std::env::var(crate::relay::ENV_FORCE_STAGING_RELAYS) {
-            Ok(value) if !value.is_empty() => Self::new(N0_DNS_NODE_ORIGIN_STAGING.to_string()),
-            _ => Self::new(N0_DNS_NODE_ORIGIN_PROD.to_string()),
+        match force_staging_infra() {
+            true => Self::new(N0_DNS_NODE_ORIGIN_STAGING.to_string()),
+            false => Self::new(N0_DNS_NODE_ORIGIN_PROD.to_string()),
         }
-        #[cfg(test)]
-        Self::new(TEST_DNS_NODE_ORIGIN.to_string())
     }
 }
 
--- a/iroh-net/src/discovery/dns.rs
+++ b/iroh-net/src/discovery/dns.rs
@@ -14,11 +14,11 @@ pub const N0_DNS_NODE_ORIGIN_PROD: &str = "dns.iroh.link";
 /// The n0 testing DNS node origin, for testing.
 pub const N0_DNS_NODE_ORIGIN_STAGING: &str = "staging-dns.iroh.link";
 /// Testing DNS node origin, must run server from [`crate::test_utils::DnsPkarrServer`].
-#[cfg(all(test, feature = "test-utils"))]
-#[cfg_attr(iroh_docsrs, doc(cfg(any(test, feature = "test-utils"))))]
+#[cfg(test)]
 pub const TEST_DNS_NODE_ORIGIN: &str = "dns.iroh.test";
 
 /// Environment variable to force the use of staging relays.
+#[cfg(not(test))]
 #[cfg_attr(iroh_docsrs, doc(cfg(not(any(test, feature = "test-utils")))))]
 const ENV_FORCE_STAGING_RELAYS: &str = "IROH_FORCE_STAGING_RELAYS";
 
@@ -73,18 +73,19 @@ impl DnsDiscovery {
     /// with [`DnsDiscovery::new`].  This would then use a hosted discovery service again,
     /// but for testing purposes.
     pub fn n0_dns() -> Self {
-        let force_staging_relays = match std::env::var(ENV_FORCE_STAGING_RELAYS) {
-            Ok(value) => value == "1",
-            Err(_) => false,
-        };
-        #[cfg(not(all(test, feature = "test-utils")))]
-        match force_staging_relays {
-            true => Self::new(N0_DNS_NODE_ORIGIN_STAGING.to_string()),
-            false => Self::new(N0_DNS_NODE_ORIGIN_PROD.to_string()),
+        #[cfg(not(test))]
+        {
+            let force_staging_relays = match std::env::var(ENV_FORCE_STAGING_RELAYS) {
+                Ok(value) => value == "1",
+                Err(_) => false,
+            };
+            match force_staging_relays {
+                true => Self::new(N0_DNS_NODE_ORIGIN_STAGING.to_string()),
+                false => Self::new(N0_DNS_NODE_ORIGIN_PROD.to_string()),
+            }
         }
-
-        #[cfg(all(test, feature = "test-utils"))]
-        Self::new(N0_DNS_NODE_ORIGIN_STAGING.to_string())
+        #[cfg(test)]
+        Self::new(TEST_DNS_NODE_ORIGIN.to_string())
     }
 }
 
--- a/iroh-net/src/discovery/dns.rs
+++ b/iroh-net/src/discovery/dns.rs
@@ -15,7 +15,7 @@ pub const N0_DNS_NODE_ORIGIN_PROD: &str = "dns.iroh.link";
 pub const N0_DNS_NODE_ORIGIN_STAGING: &str = "staging-dns.iroh.link";
 /// Testing DNS node origin, must run server from [`crate::test_utils::DnsPkarrServer`].
 #[cfg(test)]
-#[cfg_attr(iroh_docsrs, doc(cfg(not(test))))]
+#[cfg_attr(iroh_docsrs, doc(cfg(test)))]
 pub const TEST_DNS_NODE_ORIGIN: &str = "dns.iroh.test";
 
 const DNS_STAGGERING_MS: &[u64] = &[200, 300];
@@ -57,8 +57,8 @@ impl DnsDiscovery {
     ///
     /// # Usage during tests
     ///
-    /// When `cfg(test)` is enabled or when using the `test-utils` cargo feature the
-    /// [`TEST_DNS_NODE_ORIGIN`] is used.
+    /// When `cfg(test)` is enabled the
+    /// [`crate::discovery::dns::TEST_DNS_NODE_ORIGIN`] is used.
     ///
     /// Note that the `iroh.test` domain is not integrated with the global DNS network and
     /// thus node discovery is effectively disabled.  To use node discovery in a test use
--- a/iroh-net/src/discovery/dns.rs
+++ b/iroh-net/src/discovery/dns.rs
@@ -15,7 +15,7 @@ pub const N0_DNS_NODE_ORIGIN_PROD: &str = "dns.iroh.link";
 pub const N0_DNS_NODE_ORIGIN_STAGING: &str = "staging-dns.iroh.link";
 /// Testing DNS node origin, must run server from [`crate::test_utils::DnsPkarrServer`].
 #[cfg(test)]
-#[cfg_attr(iroh_docsrs, doc(cfg(test)))]
+#[cfg_attr(iroh_docsrs, doc(cfg(not(test))))]
 pub const TEST_DNS_NODE_ORIGIN: &str = "dns.iroh.test";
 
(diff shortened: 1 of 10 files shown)


iroh-net/src/discovery/dns.rs
```diff
@@ -14,10 +14,14 @@ pub const N0_DNS_NODE_ORIGIN_PROD: &str = "dns.iroh.link";
...
 /// Testing DNS node origin, must run server from [`crate::test_utils::DnsPkarrServer`].
-#[cfg(any(test, feature = "test-utils"))]
-#[cfg_attr(iroh_docsrs, doc(cfg(any(test, feature = "test-utils"))))]
+#[cfg(test)]
 pub const TEST_DNS_NODE_ORIGIN: &str = "dns.iroh.test";
 
+/// Environment variable to force the use of staging relays.
+#[cfg(not(test))]
+#[cfg_attr(iroh_docsrs, doc(cfg(not(any(test, feature = "test-utils")))))]
+const ENV_FORCE_STAGING_RELAYS: &str = "IROH_FORCE_STAGING_RELAYS";
```
flub, line 23: You're adding this to more than one file.  Let's add it to `src/relays.rs` and make it pub so we can document this as well.

iroh-net/src/discovery/pkarr.rs
```diff
@@ -64,6 +64,9 @@ use crate::{
     AddrInfo, Endpoint, NodeId,
 };
 
+/// Environment variable to force the use of staging relays.
+const ENV_FORCE_STAGING_RELAYS: &str = "IROH_FORCE_STAGING_RELAYS";
```
flub, line 68: let's move this to `src/relay.rs` as suggested in the other comment.

iroh-net/src/discovery/pkarr.rs
```diff
@@ -177,15 +180,20 @@ impl PkarrPublisher {
...
+    /// server is used instead.
     ///
     /// [number 0]: https://n0.computer
     pub fn n0_dns(secret_key: SecretKey) -> Self {
-        #[cfg(not(any(test, feature = "test-utils")))]
-        let pkarr_relay = N0_DNS_PKARR_RELAY_PROD;
-        #[cfg(any(test, feature = "test-utils"))]
-        let pkarr_relay = N0_DNS_PKARR_RELAY_STAGING;
+        let force_staging_relay = match std::env::var(ENV_FORCE_STAGING_RELAYS) {
+            Ok(value) => value == "1",
```
flub, line 190: should also just be for any non-nil value.

iroh-net/src/discovery/pkarr.rs
```diff
@@ -321,9 +329,9 @@ impl PkarrResolver {
     ///
     /// [number 0]: https://n0.computer
     pub fn n0_dns() -> Self {
-        #[cfg(not(any(test, feature = "test-utils")))]
+        #[cfg(not(all(test, feature = "test-utils")))]
```
flub, line 332: didn't @matheus23 say this doesn't work for something?

iroh-net/src/endpoint.rs
```diff
@@ -58,8 +58,6 @@ pub use iroh_base::node_addr::{AddrInfo, NodeAddr};
 const DISCOVERY_WAIT_PERIOD: Duration = Duration::from_millis(500);
 
 /// Environment variable to force the use of staging relays.
-#[cfg(not(any(test, feature = "test-utils")))]
-#[cfg_attr(iroh_docsrs, doc(cfg(not(any(test, feature = "test-utils")))))]
 const ENV_FORCE_STAGING_RELAYS: &str = "IROH_FORCE_STAGING_RELAYS";
```
flub, line 61: let's move this to one place as discussed.

iroh-net/src/endpoint.rs
```diff
@@ -1247,14 +1245,10 @@ fn proxy_url_from_env() -> Option<Url> {
 /// Otherwise, it will return `RelayMode::Default`.
 pub fn default_relay_mode() -> RelayMode {
     // Use staging in testing
-    #[cfg(not(any(test, feature = "test-utils")))]
     let force_staging_relays = match std::env::var(ENV_FORCE_STAGING_RELAYS) {
         Ok(value) => value == "1",
```
flub, line 1249: should be any non-nil value.  Maybe we can add a `pub(crate)` function to read this variable and also put it in `src/relay.rs`.  Otherwise this behaviour gets to easily different between different call sites.

iroh-net/src/endpoint.rs
```diff
@@ -1247,14 +1245,10 @@ fn proxy_url_from_env() -> Option<Url> {
 /// Otherwise, it will return `RelayMode::Default`.
 pub fn default_relay_mode() -> RelayMode {
     // Use staging in testing
-    #[cfg(not(any(test, feature = "test-utils")))]
```
flub, line 1250: this one does not have to stay at `cfg(not(test))`?

iroh/src/node/builder.rs
```diff
@@ -231,9 +231,9 @@ fn mk_external_rpc() -> IrohServerEndpoint {
 impl Default for Builder<iroh_blobs::store::mem::Store> {
     fn default() -> Self {
         // Use staging in testing
-        #[cfg(not(any(test, feature = "test-utils")))]
+        #[cfg(not(all(test, feature = "test-utils")))]
```
flub, line 234: This should probably be just `cfg(not(test))`?  If someone is requesting `test-utils` they can still enable `test` themselves.  Unless this breaks one of the benchmarks again...

matheus23 (changes requested):
> This takes us back to only relying on the env var for forcing staging relays. However for tests it should still remain active with the test/feature combo by default.

I don't think this is how it should be.

I [agree with dig](https://discord.com/channels/949724860232392765/982366224749064302/1291127779152367738), we should drop the `#[cfg(..)]` completely, and solely rely on `std::env::var(FORCE_STAGING_RELAYS).is_ok()`.

There's two permutations of this idea that I'm on the fence about:
- we could require this env variable to be present at compile time, i.e. use `std::env!("IROH_FORCE_STAGING_RELAYS")`. This might be annoying if we try to test with, let's say, a pre-built iroh-doctor binary.
- we could match on `cfg!(test) || std::env::var(FORCE_STAGING_RELAYS).is_ok()`. This way we make sure that *any* testing code *always* uses staging relays. This shouldn't be *too* bad. But also I still don't think this would be right to do. I want it to be possible to run the test env as similar to prod as possible, *if I want to*.

About `insecure_skip_relay_cert_verify`: This should just be gated on `#[cfg(feature = "test-utils")]` IMO.
But whether or not it's exposed is actually a concern outside of the scope of this PR, so this PR shouldn't change when this function actually is exposed.

iroh-net/src/discovery/dns.rs
```diff
@@ -14,8 +14,8 @@ pub const N0_DNS_NODE_ORIGIN_PROD: &str = "dns.iroh.link";
 /// The n0 testing DNS node origin, for testing.
 pub const N0_DNS_NODE_ORIGIN_STAGING: &str = "staging-dns.iroh.link";
 /// Testing DNS node origin, must run server from [`crate::test_utils::DnsPkarrServer`].
-#[cfg(any(test, feature = "test-utils"))]
-#[cfg_attr(iroh_docsrs, doc(cfg(any(test, feature = "test-utils"))))]
+#[cfg(test)]
+#[cfg_attr(iroh_docsrs, doc(cfg(not(test))))]
```
matheus23, line 18:
```suggestion
#[cfg_attr(iroh_docsrs, doc(cfg(test)))]
```

iroh-net/src/discovery/pkarr.rs
```diff
@@ -177,15 +178,16 @@ impl PkarrPublisher {
     /// This uses the pkarr relay server operated by [number 0], at
     /// [`N0_DNS_PKARR_RELAY_PROD`].
     ///
-    /// When compiling for tests, i.e. when `cfg(test)` is true, or when the `test-utils`
-    /// crate feature is enabled the [`N0_DNS_PKARR_RELAY_STAGING`] server is used instead.
+    /// When running with the environment variable
+    /// `IROH_FORCE_STAGING_RELAYS` set to `1` the [`N0_DNS_PKARR_RELAY_STAGING`]
```
matheus23, line 182:
```suggestion
    /// `IROH_FORCE_STAGING_RELAYS` set, the [`N0_DNS_PKARR_RELAY_STAGING`]
```

matheus23 (commented):
You match on `Ok(value) = std::env::var(...) if !value.is_empty()`, but in the docs you write "if the ... env variable is set".

I think you should simply use `std::env::var(ENV_FORCE_STAGING_RELAYS).is_ok()`.
If the env variable is not set, then `std::env::var` returns `Err(NotPresent)`, so checking for `is_ok()` should be fine IMO.

With your current setup, this would *not* force staging relays:
```sh
$ IROH_FORCE_STAGING_RELAYS="" cargo run # ...
```

iroh-net/src/discovery/pkarr.rs
```diff
@@ -177,15 +178,16 @@ impl PkarrPublisher {
...
     /// [number 0]: https://n0.computer
     pub fn n0_dns(secret_key: SecretKey) -> Self {
-        #[cfg(not(any(test, feature = "test-utils")))]
-        let pkarr_relay = N0_DNS_PKARR_RELAY_PROD;
-        #[cfg(any(test, feature = "test-utils"))]
-        let pkarr_relay = N0_DNS_PKARR_RELAY_STAGING;
+        let pkarr_relay = match std::env::var(ENV_FORCE_STAGING_RELAYS) {
+            Ok(value) if !value.is_empty() => N0_DNS_PKARR_RELAY_STAGING,
+            _ => N0_DNS_PKARR_RELAY_PROD,
+        };
```
matheus23, line 190: This should probably also do the `cfg(test)` dance as the other things, do, right?

iroh-net/src/discovery/pkarr.rs
```diff
@@ -321,9 +323,9 @@ impl PkarrResolver {
     ///
     /// [number 0]: https://n0.computer
     pub fn n0_dns() -> Self {
-        #[cfg(not(any(test, feature = "test-utils")))]
+        #[cfg(not(test))]
         let pkarr_relay = N0_DNS_PKARR_RELAY_PROD;
-        #[cfg(any(test, feature = "test-utils"))]
+        #[cfg(test)]
         let pkarr_relay = N0_DNS_PKARR_RELAY_STAGING;
```
matheus23, line 329: We should probably also match on the env var here, right?

iroh-net/src/endpoint.rs
```diff
@@ -1245,21 +1240,13 @@ fn proxy_url_from_env() -> Option<Url> {
...
-    };
-    #[cfg(any(test, feature = "test-utils"))]
-    let force_staging_relays = true;
-
-    match force_staging_relays {
-        true => RelayMode::Staging,
-        false => RelayMode::Default,
+    match std::env::var(ENV_FORCE_STAGING_RELAYS) {
+        Ok(value) if !value.is_empty() => RelayMode::Staging,
+        _ => RelayMode::Default,
```
matheus23, line 1249: Err. I'm a little confused.
Do we want to match on *both* env vars *and* `cfg(test)` now, or not?
Usage now seems somewhat inconsistent.

Perhaps we should just have a single `fn use_staging_infra() -> bool` somewhere instead of the `ENV_FORCE_STAGING_RELAYS` constant.
Then we can make sure our use is consistent.

iroh/src/node/builder.rs
```diff
@@ -229,9 +228,9 @@ fn mk_external_rpc() -> IrohServerEndpoint {
 impl Default for Builder<iroh_blobs::store::mem::Store> {
     fn default() -> Self {
         // Use staging in testing
-        #[cfg(not(any(test, feature = "test-utils")))]
+        #[cfg(not(test))]
         let relay_mode = RelayMode::Default;
-        #[cfg(any(test, feature = "test-utils"))]
+        #[cfg(test)]
         let relay_mode = RelayMode::Staging;
```
matheus23, line 234: IMO This should additionally test for the env variable here, too.

iroh/src/node/builder.rs
```diff
@@ -265,9 +264,9 @@ impl<D: Map> Builder<D> {
         storage: StorageConfig,
     ) -> Self {
         // Use staging in testing
-        #[cfg(not(any(test, feature = "test-utils")))]
+        #[cfg(not(test))]
         let relay_mode = RelayMode::Default;
-        #[cfg(any(test, feature = "test-utils"))]
+        #[cfg(test)]
         let relay_mode = RelayMode::Staging;
```
matheus23, line 270: (same here)

iroh/src/node/builder.rs
```diff
@@ -507,15 +506,13 @@ where
...
-    #[cfg_attr(iroh_docsrs, doc(cfg(any(test, feature = "test-utils"))))]
     pub fn insecure_skip_relay_cert_verify(mut self, skip_verify: bool) -> Self {
         self.insecure_skip_relay_cert_verify = skip_verify;
         self
     }
 
     /// Register a callback for when GC is done.
     #[cfg(any(test, feature = "test-utils"))]
-    #[cfg_attr(iroh_docsrs, doc(cfg(any(test, feature = "test-utils"))))]
     pub fn register_gc_done_cb(mut self, cb: Box<dyn Fn() + Send>) -> Self {
```
matheus23, line 516: Accidentally deleted?

iroh-net/src/discovery/pkarr.rs
matheus23: Hmm no, I think `TEST_DNS_NODE_ORIGIN` is a special case: It uses the `.test` TLD which has special meaning in DNS resolvers!

We don't have any equivalent for any of the other infra identifiers.

flub: So from my recollection `TEST_DNS_NODE_ORIGIN` was added so that it uses a special URL which will not be resolved by any normal resolvers and will only be resolved if you start a local resolver in the test using a helper function.  This stops tests from hitting any external DNS server.

Arqu:
Revamped the thing again. This is the "slim" version.
- No fluffy matching in each file, single generic function
- No `cfg(test)` shenanigans
- No TEST_urls
- Works only by env var

iroh-net/src/discovery/pkarr.rs
Arqu: IIRC we basically didn't have any requests and this was more a precaution when we were dealing with the relays as well. (Though in spirit I do agree with it)

iroh-net/src/discovery/dns.rs
```diff
@@ -57,25 +54,13 @@ impl DnsDiscovery {
...
-        #[cfg(not(any(test, feature = "test-utils")))]
-        {
-            Self::new(N0_DNS_NODE_ORIGIN_PROD.to_string())
-        }
-        #[cfg(any(test, feature = "test-utils"))]
-        {
-            Self::new(TEST_DNS_NODE_ORIGIN.to_string())
+        match force_staging_infra() {
+            true => Self::new(N0_DNS_NODE_ORIGIN_STAGING.to_string()),
+            false => Self::new(N0_DNS_NODE_ORIGIN_PROD.to_string()),
```
matheus23, line 63:
```suggestion
        if force_staging_infra() {
            Self::new(N0_DNS_NODE_ORIGIN_STAGING.to_string())
        } else {
            Self::new(N0_DNS_NODE_ORIGIN_PROD.to_string())
```

matheus23: I guess it'd be unusual to see a match on booleans. But honestly, it's shorter, and I'd almost say the match reads better in this case? I'm somewhat undecided :D

matheus23 (approved):
I'm happy with this :)

(Btw: I didn't know we *already* set the `IROH_FORCE_STAGING_RELAYS` env var in testing CI! :D)

iroh-net/src/discovery/dns.rs
Arqu: Yeah, I just followed like the others (where I `let x = match ...`) but technically here it is a simple if else. Updated to that 🤷

matheus23: lol, I saw the other uses and was like "ah well, it's probably fine", but now we have it inconsistent :see_no_evil: 
Anyhow, we can update another time.

PR #2779, opened 2024-10-03
refactor(iroh-net): Keep connection name, remove connection count

flub:
## Description

These are two cleanups in the relay client:

- The `relay::Client` hands out a connection object when asked to
  connect.  This `Conn` was imported with rename to `RelayClient`
  which was a bit confusing as this was already the relay client.  It
  is now left at it's original name which makes a lot more sense.  The
  related builder struct etc are renamed to match.

- The `relay::Client` had a counter for the number of connections made
  to the relay.  That seems fun, but was entirely unused.  If this is
  a useful thing to have it should probably be a counter metric
  instead but let's not add anything that no one is using.  Removing
  this makes a lot of APIs a bit simpler and removes some state
  tracking.

## Breaking Changes

None hopefully, please let this all  be internal APIs.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/src/relay/client.rs
+++ b/iroh-net/src/relay/client.rs
@@ -28,8 +28,8 @@ use tracing::{debug, error, event, info_span, trace, warn, Instrument, Level};
 use url::Url;
 
 use conn::{
-    Conn as RelayClient, ConnBuilder as RelayClientBuilder, ConnReader,
-    ConnReceiver as RelayClientReceiver, ConnWriter, ReceivedMessage,
+    Conn as RelayConn, ConnBuilder as RelayConnBuilder, ConnReader,
+    ConnReceiver as RelayConnReceiver, ConnWriter, ReceivedMessage,
 };
 use streams::{downcast_upgrade, MaybeTlsStream, ProxyStream};
 
@@ -140,7 +140,7 @@ pub struct Client {
 
 #[derive(Debug)]
 enum ActorMessage {
-    Connect(oneshot::Sender<Result<(RelayClient, usize), ClientError>>),
+    Connect(oneshot::Sender<Result<RelayConn, ClientError>>),
     NotePreferred(bool),
     LocalAddr(oneshot::Sender<Result<Option<SocketAddr>, ClientError>>),
     Ping(oneshot::Sender<Result<Duration, ClientError>>),
@@ -154,7 +154,7 @@ enum ActorMessage {
 /// Receiving end of a [`Client`].
 #[derive(Debug)]
 pub struct ClientReceiver {
-    msg_receiver: mpsc::Receiver<Result<(ReceivedMessage, usize), ClientError>>,
+    msg_receiver: mpsc::Receiver<Result<ReceivedMessage, ClientError>>,
 }
 
 #[derive(derive_more::Debug)]
@@ -162,11 +162,10 @@ struct Actor {
     secret_key: SecretKey,
     can_ack_pings: bool,
     is_preferred: bool,
-    relay_client: Option<(RelayClient, RelayClientReceiver)>,
+    relay_conn: Option<(RelayConn, RelayConnReceiver)>,
     is_closed: bool,
     #[debug("address family selector callback")]
     address_family_selector: Option<Box<dyn Fn() -> BoxFuture<bool> + Send + Sync + 'static>>,
-    conn_gen: usize,
     url: RelayUrl,
     protocol: Protocol,
     #[debug("TlsConnector")]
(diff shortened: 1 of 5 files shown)


iroh-net/src/relay/client.rs
```diff
@@ -28,8 +28,8 @@ use tracing::{debug, error, event, info_span, trace, warn, Instrument, Level};
 use url::Url;
 
 use conn::{
-    Conn as RelayClient, ConnBuilder as RelayClientBuilder, ConnReader,
-    ConnReceiver as RelayClientReceiver, ConnWriter, ReceivedMessage,
+    Conn as RelayConn, ConnBuilder as RelayConnBuilder, ConnReader,
+    ConnReceiver as RelayConnReceiver, ConnWriter, ReceivedMessage,
```
divagant-martian, line 32: Wild suggestion, but what if we import `conn:: self as relay` and just use `relay::ConnReceiver`, `relay::Conn` and `relay::ConnBuilder`. 

It's not shorter, but maybe it's a bit more aligned with this idea of stop using prefixes and prefer module references.

What are your thoughts?

iroh-net/src/relay/client.rs
```diff
@@ -399,13 +396,13 @@ impl Client {
         }
     }
 
-    /// Connect to a relay Server and returns the underlying relay Client.
+    /// Connect to a relay Server and returns the underlying relay connection.
```
divagant-martian, line 399: now that we are here and the line is flagged as changed, can we pick one from imperative and third person? we have here both `Connect to` and `returns`, otherwise the change is an improvement imo

iroh-net/src/relay/client.rs
```diff
@@ -549,46 +546,52 @@ impl Actor {
         }
     }
 
+    /// Returns a connection to the relay.
+    ///
+    /// If the client is currently connected the existing connection is returned, otherwise
```
divagant-martian, line 551:
```suggestion
    /// If the client is currently connected, the existing connection is returned; otherwise,
```

iroh-net/src/relay/client.rs
```diff
@@ -549,46 +546,52 @@ impl Actor {
...
+    /// a new connection is made.
+    ///
+    /// Returns:
+    /// - A clonable connection object which can send to DISCO messages to the relay.
```
divagant-martian, line 555:
```suggestion
    /// - A clonable connection object which can send DISCO messages to the relay.
```

iroh-net/src/relay/client.rs
```diff
@@ -1012,7 +1005,7 @@ impl Actor {
     /// requires a connection, it will call `connect`.
     async fn close_for_reconnect(&mut self) {
         debug!("close for reconnect");
-        if let Some((client, _)) = self.relay_client.take() {
+        if let Some((client, _)) = self.relay_conn.take() {
```
divagant-martian, line 1008:
```suggestion
        if let Some((conn, _)) = self.relay_conn.take() {
```

iroh-net/src/relay/client.rs
```diff
@@ -987,13 +981,12 @@ impl Actor {
         }
     }
 
-    async fn recv_detail(&mut self) -> Result<(ReceivedMessage, usize), ClientError> {
-        if let Some((_client, client_receiver)) = self.relay_client.as_mut() {
+    async fn recv_detail(&mut self) -> Result<ReceivedMessage, ClientError> {
+        if let Some((_client, client_receiver)) = self.relay_conn.as_mut() {
```
divagant-martian, line 985:
```suggestion
        if let Some((_conn, client_receiver)) = self.relay_conn.as_mut() {
```

maybe? not sure about the receiver part

iroh-net/src/relay/client.rs
flub: I originally thought this was a great idea, but actually doing it having `conn as relay` inside `iroh_net::relay::client` was just a bit too weird.

But why bother renaming at all?  Looking at it again I think it is at least as clear to use `Conn` as `RelayConn` given that we're in `iroh_net::relay::client`.  So that's what I did together with improving the docs on `Conn`, `ConnBuilder` and `ConnReceiver`.

What do you think?

divagant-martian (approved):
LGTM!

Don't want to get nitpicky so do as you wish: There are still many variables named client when doing `relay_conn.method()` like `take`, `as_ref` etc. I find it understandable enough to leave it like that, just slightly cleaner to change it

flub:
> LGTM!
> 
> Don't want to get nitpicky so do as you wish: There are still many variables named client when doing `relay_conn.method()` like `take`, `as_ref` etc. I find it understandable enough to leave it like that, just slightly cleaner to change it

Your nitpicking is totally right, might as well try and do this more consistent because it does make it more readable.  But yeah, I was moving on as I wanted to get to the actual bugfix I was aiming for at the end of this chain of PRs.

PTAL

divagant-martian (approved):

PR #2781, opened 2024-10-03
refactor(iroh-net): Optimise present nodes in ActiveRelay

flub:
## Description

The ActiveRelay actor keeps track of which remote nodes are present on the relay connection so that we can optimise relay connections to remote nodes.  This does two main optimisations:

- There were two sets of these nodes kept, they could easily be unified.

- The set is best stored in a BTreeSet since they are simple NodeIds stored in them.

- Bonus: rename peer to node to match our naming convention.

- Bonus: identify nodes by NodeId since this is a routing key here.

## Breaking Changes

Still none if all is well.

## Notes & open questions

This targets #2779 as base.

## Change checklist

- [x] Self-review.
- ~~[ ] Tests if relevant.~~

### DIFF
--- a/iroh-net/src/magicsock/relay_actor.rs
+++ b/iroh-net/src/magicsock/relay_actor.rs
@@ -165,10 +165,7 @@ impl ActiveRelay {
         Ok(())
     }
 
-    async fn handle_relay_msg(
-        &mut self,
-        msg: Result<(ReceivedMessage, usize), ClientError>,
-    ) -> ReadResult {
+    async fn handle_relay_msg(&mut self, msg: Result<ReceivedMessage, ClientError>) -> ReadResult {
         match msg {
             Err(err) => {
                 warn!("recv error {:?}", err);
@@ -200,7 +197,7 @@ impl ActiveRelay {
                     None => ReadResult::Break,
                 }
             }
-            Ok((msg, _conn_gen)) => {
+            Ok(msg) => {
                 // reset
                 self.backoff.reset();
                 let now = Instant::now();
--- a/iroh-net/src/magicsock/relay_actor.rs
+++ b/iroh-net/src/magicsock/relay_actor.rs
@@ -56,8 +56,8 @@ struct ActiveRelay {
     relay_client_receiver: relay::client::ClientReceiver,
     /// The set of remote nodes we know are present on this relay server.
     ///
-    /// If we receive messages from a remote node via this server it is added to this set,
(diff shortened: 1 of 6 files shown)


iroh-net/src/magicsock/relay_actor.rs
```diff
@@ -51,21 +51,17 @@ struct ActiveRelay {
...
     url: RelayUrl,
     relay_client: relay::client::Client,
     relay_client_receiver: relay::client::ClientReceiver,
-    /// The set of senders we know are present on this connection, based on
-    /// messages we've received from the server.
-    peer_present: HashSet<PublicKey>,
+    /// The set of remote nodes we know are present on this relay server.
+    ///
+    /// If we receive messages from a remote node via this server it is added to this set,
+    /// if the server notifies us this node is gone it is removed from this set.
```
divagant-martian, line 60:
```suggestion
    /// If we receive messages from a remote node via, this server it is added to this set.
    /// If the server notifies us this node is gone, it is removed from this set.
```

iroh-net/src/magicsock/relay_actor.rs
```diff
@@ -414,9 +410,9 @@ impl RelayActor {
     async fn connect_relay(
         &mut self,
         url: &RelayUrl,
-        peer: Option<&PublicKey>,
+        remote_node: Option<&NodeId>,
     ) -> relay::client::Client {
-        debug!("connect relay {} for peer {:?}", url, peer);
+        debug!("connect relay {} for peer {:?}", url, remote_node);
```
divagant-martian, line 415:
```suggestion
        debug!(%url, ?remote_node, "connect relay");
```

iroh-net/src/magicsock/relay_actor.rs
```diff
@@ -1,5 +1,5 @@
 use std::{
-    collections::{BTreeMap, HashSet},
+    collections::{BTreeMap, BTreeSet},
```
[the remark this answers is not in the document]
flub: The aim is to save the cost of hashing.  Since these are just 32 bytes I think BTreeSet is very suitable, not sure fnv would be any better looking at the graphs linked from the fnv docs.  Maybe if the inputs were < 15 bytes it might be worth it to try fnv.

divagant-martian (approved):
I checked the original code and while the two sets were not exactly the same, I see no reason why this should work differently to the previous one. From my perspective this lgtm!

PR #2785, opened 2024-10-04
docs(iroh-net): Add examples to the module docs

flub:
## Description

This gives at least some quick indication of how to make things work.

## Change checklist

- [x] Self-review.
- ~~[ ] Tests if relevant.~~

### DIFF
--- a/iroh-net/src/lib.rs
+++ b/iroh-net/src/lib.rs
@@ -1,4 +1,4 @@
-//! Peer-to-peer connectivity QUIC connections.
+//! Peer-to-peer QUIC connections.
 //!
 //! iroh-net is a library to establish direct connectivity between peers.  It exposes an
 //! interface to [QUIC] connections and streams to the user, while implementing direct
--- a/iroh-net/src/lib.rs
+++ b/iroh-net/src/lib.rs
@@ -1,4 +1,4 @@
-//! Peer-to-peer connectivity based on QUIC.
+//! Peer-to-peer connectivity QUIC connections.
 //!
 //! iroh-net is a library to establish direct connectivity between peers.  It exposes an
 //! interface to [QUIC] connections and streams to the user, while implementing direct
@@ -88,9 +88,18 @@
 //!   initiator of this stream has to send data before the peer will be aware of this
 //!   stream.
 //!
-//! Additionally to being extremely light-weight, streams can be interleaved and will not block
-//! each other.  Allowing many streams to co-exist, regardless of how long they last.
+//! Additionally to being extremely light-weight, streams can be interleaved and will not
+//! block each other.  Allowing many streams to co-exist, regardless of how long they last.
 //!
+//! <div class="warning">
+//!
+//! To keep streams cheap, they are lazily created on the network: only once a sender starts
+//! sending data on the stream will the receiver become aware of a stream.  This means only
+//! calling [`Connection::open_bi`] is not sufficient for the corresponding call to
+//! [`Connection::accept_bi`] to return.  The sender **must** send data on the stream before
+//! the receiver's [`Connection::accept_bi`] call will return.
+//!
+//! </div>
 //!
 //! ## Node Discovery
 //!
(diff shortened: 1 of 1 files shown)


iroh-net/src/lib.rs
```diff
@@ -103,8 +112,72 @@
...
 //! connecting using only the [`NodeId`] which is often more convenient and resilient.
 //!
+//! See [the discovery module] for more details.
+//!
+//!
+//! # Examples
+//!
+//! The central struct is the [`Endpoint`], which allows you to connect to other nodes:
+//!
+//! ```no_run
```
dignifiedquire, line 122: does `no_run` still check if this compiles? otherwise I am very worried about these getting out of date

flub: Yep, `no_run` checks it compiles but does not try to run it.  Seemed like the right compromise here, since building a fully workable example was so much extra noise.  I guess I could hide all that noise to still show it like this, but the hiding is a pretty horrible experience to write as well...

iroh-net/src/lib.rs
```diff
@@ -103,8 +112,72 @@
...
+//!     // Establish a QUIC connection, open a bi-directional stream, exchange messages.
+//!     let conn = ep.connect(addr, b"hello-world").await?;
+//!     let (mut send_stream, mut recv_stream) = conn.open_bi().await?;
+//!     send_stream.write_all(b"hello").await?;
+//!     send_stream.finish()?;
+//!     let msg = recv_stream.read_to_end(10).await?;
+//!     println!("Server message: {}", String::from_utf8_lossy(&msg));
+//!
+//!     // Gracefully close the connection and endpoint.
+//!     conn.close(1u8.into(), b"done");
```
flub, line 139: I could delete this line as well.  I'm undecided.

iroh-net/src/lib.rs
```diff
@@ -103,8 +112,72 @@
...
+//! ```no_run
+//! use anyhow::Result;
+//! use iroh_net::{Endpoint, NodeAddr};
+//!
+//! async fn connect(addr: NodeAddr) -> Result<()> {
+//!     // The Endpoint is the central object that manages an iroh-net node.
+//!     let ep = Endpoint::builder().bind().await?;
+//!
+//!     // Establish a QUIC connection, open a bi-directional stream, exchange messages.
+//!     let conn = ep.connect(addr, b"hello-world").await?;
```
rklaehn, line 131: I think putting "hello-world" into a const ALPN would make this clearer.

iroh-net/src/lib.rs
rklaehn: Don't you need this to make sure the conn.close has time to be sent before the entire program exits?

If anything I would have thought that the conn.close is the one that can be removed.

flub: Indeed, I meant to put that comment on the `conn.close()` line.

flub:
> Very nice. Not sure about this, but maybe even move the example further up.

Yeah, initially I was trying to have them very high up.  But unfortunately they're quite long if you want to do them completely, and now they take up a whole page basically which I think is too much to lead with.

Maybe some incomplete version that just connects and sends a single message could be added to the start.

iroh-net/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-//! Peer-to-peer connectivity based on QUIC.
+//! Peer-to-peer connectivity QUIC connections.
```
matheus23, line 1:
```suggestion
//! Peer-to-peer QUIC connections.
```

Not sure about my suggestion.
But the current state basically reads like two subjects back-to-back in a sentence.

iroh-net/src/lib.rs
matheus23: No, you can leave out this `.close` call here.
I think Rüdiger was mixing this call up with the call to `ep.close` in the line below. That's the line that's *actually* waiting for the connection to close gracefully (because it calls `Endpoint::wait_idle` internally).

matheus23: I'm getting trolled by github.
This is when I'm in the "Conversation" tab:
image

This is when I'm in the "Files" tab:
image

matheus23: Anyways. I'm in favor of having a `conn.close` call here.
In the future I'd like the code to look like this instead:
```rs
conn.close(0u8.into(), b"done");
ep.all_closed().await?;
```
But that's irrelevant for *this* PR in particular.

iroh-net/src/lib.rs
```diff
@@ -103,8 +112,72 @@
...
+//!     println!("Client message: {}", String::from_utf8_lossy(&msg));
+//!     send_stream.write_all(b"world").await?;
+//!     send_stream.finish()?;
+//!
+//!     // Wait for the client to close the connection and gracefully close the endpoint.
+//!     conn.closed().await;
+//!     ep.close(0u8.into(), b"ep closing").await?;
+//!     Ok(())
+//! }
+//! ```
```
matheus23, line 174: If we *really* want to, we can slim down the example like this:

```suggestion
//! ```no_run
//! # use anyhow::Result;
//! # use iroh_net::{Endpoint, NodeAddr};
//! #
//! # async fn connect(addr: NodeAddr) -> Result<()> {
//! // The Endpoint is the central object that manages an iroh-net node.
//! let ep = Endpoint::builder().bind().await?;
//!
//! // Establish a QUIC connection, open a bi-directional stream, exchange messages.
//! let conn = ep.connect(addr, b"hello-world").await?;
//! let (mut send_stream, mut recv_stream) = conn.open_bi().await?;
//! send_stream.write_all(b"hello").await?;
//! send_stream.finish()?;
//! let msg = recv_stream.read_to_end(10).await?;
//!
//! // Gracefully close the connection and endpoint.
//! conn.close(1u8.into(), b"done");
//! ep.close(0u8.into(), b"ep closing").await?;
//! # Ok(())
//! # }
//! ```
//!
//! Every [`Endpoint`] can also accept connections:
//!
//! ```no_run
//! # use anyhow::{Context, Result};
//! # use futures_lite::StreamExt;
//! # use iroh_net::ticket::NodeTicket;
//! # use iroh_net::{Endpoint, NodeAddr};
//! #
//! const ALPN: &[u8] = b"hello-world"; 
//! #
//! # async fn accept() -> Result<()> {
//! // To accept connections at least one ALPN must be configured.
//! let ep = Endpoint::builder().alpns(vec![ALPN.to_vec()]).bind().await?;
//!
//! // Accept a QUIC connection, accept a bi-directional stream, exchange messages.
//! let conn = ep.accept().await.context("no incoming connection")?.await?;
//! let (mut send_stream, mut recv_stream) = conn.accept_bi().await?;
//! let msg = recv_stream.read_to_end(10).await?;
//! send_stream.write_all(b"world").await?;
//! send_stream.finish()?;
//!
//! // Wait for the client to close the connection and gracefully close the endpoint.
//! conn.closed().await;
//! ep.close(0u8.into(), b"ep closing").await?;
//! # Ok(())
//! # }
//! ```
```

I'm stripping out the printlns and hiding the imports and function scaffolding here.
This might be fine:
1. Even if you can copy-paste these examples, you can't immediately run them, because there's no `main` function.
2. Because you can't run them, you don't need to have `println`s here, I think.
3. I hope that people don't need to see imports here, as they are already *in the docs* where they can search for everything.

I can see people disagreeing with me here, even I'm not sure, but being more elaborate is the easier option.

matheus23 (approved):
I think this is really nice. I agree with Rüdiger, I think we can move the example up.

Feel free to ignore my idea for stripping down the example so it occupies less space :P

matheus23:
Eh - too late anyways.

flub:
oh lol.  apologies, i forgot i enabled auto-merge 2 weeks ago :joy: 

i'll  follow up on this

PR #2787, opened 2024-10-07
feat(iroh-net): Add counters for holepunching events

matheus23:
## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/src/magicsock.rs
+++ b/iroh-net/src/magicsock.rs
@@ -1232,6 +1232,7 @@ impl MagicSock {
                 via = ?url,
                 ?addrs,
             );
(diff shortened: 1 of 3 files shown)


matheus23:
Okay, I think `sent_disco_relay` basically counts the holepunching attempts, and `holepunched` with this PR counts successful holepunching events.
`sent_disco_relay - holepunched` should give us holepunching failures.

matheus23:
I don't think this is the right way to measure this.
We don't really care about individual holepunch successes. We care about how many network configurations can switch to direct connections.
We can get closer to that metric via #2795 instead.

This might be nice to track additionally eventually, but we don't need this number *now*.

PR #2788, opened 2024-10-07
feat(metrics): add option to export telemetry

Arqu:
## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


iroh-metrics/src/lib.rs
```diff
@@ -7,8 +7,9 @@ pub mod metrics;
 pub mod core;
 
 /// Expose iroh metrics
+/// todo: remove pub export
```
divagant-martian, line 10: what's the alternative here? is this because we are considering it temporary™?

rklaehn: Export just the one fn we need, not the entire module?

Arqu: Yeah I dropped this in favor of keeping everything in line as it was.

PR #2806, opened 2024-10-15
feat(iroh-net): Add a Watchable struct for use in the Endpoint API

flub:
## Description

This implements a `Watchable` struct and a `Watcher` which provides access to the watched value in several ways, including some streams.

## Breaking Changes

- `iroh::Endpoint::conn_type_stream` is renamed to `iroh::Endpoint::conn_type` and returns `Result<Watcher<ConnectionType>>` instead of `Result<ConnectionTypeStream>`
  To migrate, use `endpoint.conn_type()?.stream()` instead of `endpoint.conn_type_stream()?`.
- `iroh::Endpoint::home_relay` now returns `Watcher<Option<RelayUrl>>` instead of `Option<RelayUrl>`.
  To migrate, use `endpoint.home_relay().get()?` instead of `endpoint.home_relay()`.
- removed `iroh::Endpoint::watch_home_relay`
  To migrate, use `endpoint.home_relay().initialized().await?` instead of `endpoint.watch_home_relay().next().await` and
  use `endpoint.home_relay().stream()` instead of `endpoint.watch_home_relay().next().await`.
- removed `DirectAddrsStream` and `ConnTypeStream`.
  Use `iroh::watchable::WatcherStream` for as named types instead.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


flub:
@mepi262 the current implementation is racy and needs fixing.

What makes you interested in this?

mepi262:
The reason in which I'm interested is followings.

- I'm very much looking forward to the iroh v1.0.0.
- This pull request is related to the issue, which is contained milestone v1.0.0
- I'm worry about whether this pull request is merged, because this pull request has not been active since last month.
  - People forget memory
  - Merging pull request is difficult when time has gone, because of risk of conflict.

I wish you good health, good luck and success.

iroh-net/src/util/watchable.rs
```diff
@@ -0,0 +1,401 @@
...
+        //    separate:
+        //    - thread 1 runs poll_next
+        //    - thread 2 runs set
+        //    1. thread 1: load epoch
+        //    2. thread 2: lock value, replace value, unlock value
+        //    3. thread 2: store epoch
+        //    4. thread 2: lock wakers, drain wakers, unlock wakers
+        //    5. thread 1: lock wakers, install waker, unlock wakers
+        //
+        //    I believe the epoch and wakers need to be stored in the same RwLock.
```
matheus23, line 72: You're completely right about the wakers possibly missing a wakeup once a value is available.
I've tested this implementation of Watchable with [`loom`](https://docs.rs/loom/latest/loom/), and can confirm that it gets stuck waiting for `watcher.initialized()` in some cases, even though there's a `watchable.set()` call.
(Interestingly, reproducing it practically *requires* loom usage, since it's very timing-sensitive and *doesn't* reproduce easily with running the same test case in a tight loop a couple thousand times.)

You're incorrect about the fact that the epoch and wakers needing to be stored on the same RwLock.

Similar to [this example](https://docs.rs/futures/latest/futures/task/struct.AtomicWaker.html#examples), after writing the waker, you can re-check the epoch to see if there was a write in-between.

Adding this second check, it passes loom.

matheus23:
Made some changes:
- Added loom to `watchable.rs` (conditionally on the `iroh_loom` cfg)
- Added a loom-based test that was failing in exactly the way @flub predicted
- Fixed the race condition
- Changed `wakers` from using a `RwLock` to using a `Mutex`, because we were only using `RwLock::write`
- Removed `Either` by making `initialized` simpler

For anyone who's interested in running the loom test, you get some fun output like this:
```
LOOM_LOG=trace RUSTFLAGS="--cfg iroh_loom" cargo test --package iroh-net --lib -- util::watchable::tests::test_initialized_always_resolves --exact --nocapture
```

iroh-net/src/util/watchable.rs
```diff
@@ -0,0 +1,467 @@
...
+        WatchNextFut { watcher: self }
+    }
+
+    /// Returns a stream which will yield an items for the most recent value.
+    ///
+    /// The first item of the stream is the current value, so that this stream can be easily
+    /// used to operate on the most recent value.  If the stream is not yet initialized the
+    /// first item of the stream will not be readily available.
+    ///
+    /// Note however that only the last item is stored.  If the stream is not polled when an
```
flub, line 246:
```suggestion
    /// Note however, that only the last item is stored.  If the stream is not polled when an
```

iroh-net/src/util/watchable.rs
```diff
@@ -0,0 +1,467 @@
...
+        self.epoch = INITIAL_EPOCH;
+        WatchNextFut { watcher: self }
+    }
+
+    /// Returns a future completing once a new value is set.
+    pub(crate) fn updated(&mut self) -> WatchNextFut<'_, T> {
+        WatchNextFut { watcher: self }
+    }
+
+    /// Returns a stream which will yield an items for the most recent value.
```
flub, line 240:
```suggestion
    /// Returns a stream which will yield items containing the most recent value.
```

iroh-net/src/util/watchable.rs
```diff
@@ -0,0 +1,467 @@
...
+    ///
+    /// Note however that only the last item is stored.  If the stream is not polled when an
+    /// item is available it can be replaced with another item by the time it is polled.
+    pub fn stream(mut self) -> impl Stream<Item = T> {
+        debug_assert!(self.epoch > 0);
+        self.epoch -= 1;
+        WatchNextStream { watcher: self }
+    }
+
+    /// Returns a stream which will yield an item for changes to the watched value.
```
flub, line 254:
```suggestion
    /// Returns a stream which will yield items for changes to the watched value.
```

iroh-net/src/util/watchable.rs
```diff
@@ -0,0 +1,467 @@
...
+    ///
+    /// Note however that only the last item is stored.  If the stream is not polled when an
+    /// item is available it can be replaced with another item by the time it is polled.
+    pub fn stream_updates_only(self) -> impl Stream<Item = T> {
+        WatchNextStream { watcher: self }
+    }
+}
+
+#[derive(Debug)]
+#[repr(transparent)]
```
flub, line 267: Why?  I thought this was a thing only needed for C APIs.

iroh-net/src/util/watchable.rs
```diff
@@ -0,0 +1,467 @@
...
+pub struct WatchNextFut<'a, T> {
```
flub, line 268: Needs a doc comment

flub (commented):
Looks good!  Any reason this is still draft?

flub:
> Should we (well, I) maybe address [#2860 (comment)](https://github.com/n0-computer/iroh/discussions/2860#discussioncomment-11239318) in here too?
> 
> > What's the relationship between [Endpoint::conn_type_stream](https://docs.rs/iroh-net/0.28.1/iroh_net/endpoint/struct.Endpoint.html#method.conn_type_stream) and [Connection::remote_address](https://docs.rs/iroh-net/0.28.1/iroh_net/endpoint/struct.Connection.html#method.remote_address)? It seems like remote_address should be deprecated and replaced with a method that returns a ConnectionType based on the current state of the connection.

I think this PR just introduces the watchable, follow-up PRs can start using it.

matheus23:
> Looks good! Any reason this is still draft?

> I think this PR just introduces the watchable, follow-up PRs can start using it.

Ah okay. Well my plan was to learn from the refactor of making iroh-net *use* this new watchable.

And I'm already finding things. E.g. do we want the Watchable to have `shutdown` functionality.

So, I'm still learning!
In any case, I'll split up the PR into watchable introduction & the refactor anyways.

iroh-net/src/util/watchable.rs
```diff
@@ -0,0 +1,467 @@
...
+
+const INITIAL_EPOCH: u64 = 1;
+
+type Result<T> = std::result::Result<T, Error>;
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("Watch lost connection to underlying Watchable, it was dropped")]
+    WatchableClosed,
+}
```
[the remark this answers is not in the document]
matheus23: Yeah thought about that, too. Thanks for chiming in. I'll call it `Disconnected` similar to what it's called in the `watchable` crate.

iroh-net/src/util/watchable.rs
```diff
@@ -0,0 +1,467 @@
...
+#[derive(Debug, Clone)]
+pub struct Watcher<T> {
+    epoch: u64,
+    shared: Weak<Shared<T>>,
+}
+
+impl<T: Clone + Eq> Watcher<T> {
+    /// Returns the currently held value.
+    ///
+    /// Returns `None` if the value was not set yet.
```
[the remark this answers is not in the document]
flub: lol, it's so hard!

Cargo.toml
```diff
@@ -31,13 +31,14 @@ missing_debug_implementations = "warn"
 
 # We use this --cfg for documenting the cargo features on which an API
 # is available.  To preview this locally use: RUSTFLAGS="--cfg
-# iroh_docsrs cargo +nightly doc --all-features".  We use our own
+# iroh_docsrs" cargo +nightly doc --all-features.  We use our own
```
dignifiedquire, line 34: ?

matheus23: that's... the right way to write this command?
`RUSTFLAGS="--cfg iroh_docsrs cargo +nightly doc --all-features"` vs.
`RUSTFLAGS="--cfg iroh_docsrs" cargo +nightly doc --all-features`

iroh-net/src/util/watchable.rs
```diff
@@ -0,0 +1,467 @@
...
+
+        {
+            let state = self.state.read().unwrap();
+            let epoch = state.epoch;
+
+            if last_epoch < epoch {
+                if let Some(value) = state.value.clone() {
+                    // Once initialized our Option is never set back to None, but nevertheless
+                    // this code is safer without relying on that invariant.
+                    return Poll::Ready((epoch, value));
```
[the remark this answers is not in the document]
matheus23: Yeah - spurious wakeup. The alternative is to lock the wakers *again*, and go through all of them, filtering any waker that would wake this task.
My worry is that's not worth it. You're blocking a bunch of other tasks in O(n) time for very little gain.

iroh/src/discovery.rs
```diff
@@ -833,27 +834,6 @@ mod test_dns_pkarr {
         Ok(())
     }
 
-    #[tokio::test]
```
dignifiedquire, line 836: why is this removed?

iroh/src/discovery/pkarr.rs
```diff
@@ -255,9 +258,9 @@ impl PublisherService {
             }
             // Wait until either the retry/republish timeout is reached, or the node info changed.
             tokio::select! {
-                res = self.watcher.watch_async() => match res {
-                    Ok(()) => debug!("Publish node info to pkarr (info changed)"),
-                    Err(_disconnected) => break,
+                res = self.watcher.updated() => match res {
```
dignifiedquire, line 261: I hope this is cancel safe ;)

iroh/src/discovery.rs
matheus23: That test is identical to the test above.

iroh/src/discovery/pkarr.rs
matheus23: Yes. I've added comments & a rough test.
Really the only thing that can go wrong here is that the watcher ends up in a broken state after cancellation.
That is not the case.

Missing values is expected of the interface anyways.

iroh/src/watchable.rs
```diff
@@ -0,0 +1,488 @@
...
+    sync::{Arc, Weak},
+    task::{self, Poll, Waker},
+};
+
+use futures_lite::stream::Stream;
+#[cfg(iroh_loom)]
+use loom::sync;
+use sync::{Mutex, RwLock};
+
+/// A value who's changes over time can be observed.
```
dignifiedquire, line 26: this reads strange

matheus23: Agreed. lol

iroh/src/discovery/pkarr.rs
```diff
@@ -150,7 +150,7 @@ impl PkarrPublisher {
         debug!("creating pkarr publisher that publishes to {pkarr_relay}");
         let node_id = secret_key.public();
         let pkarr_client = PkarrRelayClient::new(pkarr_relay);
-        let watchable = Watchable::default();
+        let watchable = Watchable::new(None);
```
dignifiedquire, line 153: would be nice to implement `Default for Watchable<T: Default>`

matheus23: It already does, I should use it again, though.

iroh/src/endpoint.rs
```diff
@@ -883,38 +879,45 @@ impl Endpoint {
...
+    /// It is possible for the connection type to be [`ConnectionType::None`] if you've
+    /// recently connected to this node id but previous methods of reaching the node have
+    /// become inaccessible.
+    ///
     /// # Errors
     ///
     /// Will error if we do not have any address information for the given `node_id`.
-    pub fn conn_type_stream(&self, node_id: NodeId) -> Result<ConnectionTypeStream> {
-        self.msock.conn_type_stream(node_id)
+    pub fn conn_type(&self, node_id: NodeId) -> Result<Watcher<ConnectionType>> {
```
flub, line 907: Probably out of scope for this PR.  But what would folks think of changing `fn conn_type(node_id)` into `fn network_path(node_id)`?  And Also rename `ConnectionType` into `NetworkPath`.

This name has always bothered me a bit, I'm not sure why.  Maybe because this doesn't actually have much to do with a `Connection`, it is on a different level as shown by the fact this is an `Endpoint` API.  But this is subtle and not very clear.

flub (commented):
LGTM

iroh/src/endpoint.rs
matheus23: I like the idea :)
I have a bunch of follow-up PR ideas brewing anyways, maybe I'll just add them to the follow-up.

matheus23 (approved):
Turning @flub's "LGTM" into a github-approved LGTM :upside_down_face:

PR #2808, opened 2024-10-16
fix(iroh-net)!: DiscoveredDirectAddrs need to update the timestamp

flub:
## Description

DiscoverdDirectAddrs has a timestamp of when it was last changed, which is used to decide if the direct addr information is recent enough to use in call-me-maybe messages.  However when it was updated this timestamp was not included in the comparison to see if there were changes, so that watchers would not get updates when the direct addresses did not change.  Unfortunately this meant the timestamp also never updated and the DiscoveredDirectAddrs was always considered out of date (unless there were changes in the discovered addresses).

This fixes this by putting only the DirectAddrs in the watcher.

It does a few other improvements:

- Moves all logic about DiscoveredDirectAddrs into one place.  This could be split out to a module now if desired.

- Tidies up the code which creates the DirectAddrs from the portmapper, netcheck and the local addresses.  Again this is just a first pass in cleaning up this code, there is so much more that could be done.

## Breaking Changes

* `iroh_net::Endpoint::direct_addresses` now returns a stream yielding Items of `BTreeSet<DirectAddr>` instead of `Vec<DirectAddr>`.
* `iroh_base::node_addr::NodeAddr::from_parts` now takes a `impl IntoIterator<Item = DirectAddr>` instead of `Vec<DirectAddr>`.  But since `Vec` implements `IntoIterator` that is not really a breaking change.

## Notes & open questions

The crux of this fix is moving from `Watchable<DiscoveredDirectAddresses>` to `DiscoveredDirectAddresses::addrs` being `Watchable<Vec<DirectAddr>>` instead of the whole struct being watchable.

I could split the refactoring of store_direct_addr_update and the renaming of it's related functions into a separate PR in front of this bugfix if desired.  I had to change that code because of the format change from `Vec<DirectAddrs>` to `BTreeSet<DirectAddr>` and it was just way clearer doing it with the refactoring.

As indicated in the (already existing) comments that whole direct address updating logic can still be more refactored and improved.  But many small steps is how we improve.

I went very opinionated on the imports... apologies.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-base/src/node_addr.rs
+++ b/iroh-base/src/node_addr.rs
@@ -74,13 +74,13 @@ impl NodeAddr {
     pub fn from_parts(
         node_id: PublicKey,
         relay_url: Option<RelayUrl>,
-        direct_addresses: Vec<SocketAddr>,
+        direct_addresses: BTreeSet<SocketAddr>,
     ) -> Self {
         Self {
             node_id,
             info: AddrInfo {
                 relay_url,
-                direct_addresses: direct_addresses.into_iter().collect(),
+                direct_addresses,
             },
         }
     }
--- a/iroh-base/src/node_addr.rs
+++ b/iroh-base/src/node_addr.rs
@@ -192,11 +192,11 @@ pub enum AddrInfoOptions {
     /// This usually means that iroh-dns discovery is used to find address information.
     #[default]
     Id,
-    /// Includes both the relay URL and the direct addresses.
+    /// Includes the Node ID and both the relay URL, and the direct addresses.
     RelayAndAddresses,
-    /// Only includes the relay URL.
+    /// Includes the Node ID and the relay URL.
     Relay,
-    /// Only includes the direct addresses.
+    /// Includes the Node ID and the direct addresses.
     Addresses,
 }
 
--- a/iroh-base/src/node_addr.rs
+++ b/iroh-base/src/node_addr.rs
@@ -74,13 +74,13 @@ impl NodeAddr {
     pub fn from_parts(
         node_id: PublicKey,
         relay_url: Option<RelayUrl>,
-        direct_addresses: BTreeSet<SocketAddr>,
+        direct_addresses: impl IntoIterator<Item = SocketAddr>,
     ) -> Self {
         Self {
             node_id,
             info: AddrInfo {
                 relay_url,
-                direct_addresses,
+                direct_addresses: direct_addresses.into_iter().collect(),
             },
         }
     }
--- a/iroh-net/src/magicsock.rs
+++ b/iroh-net/src/magicsock.rs
@@ -2478,13 +2478,13 @@ pub struct DirectAddrsStream {
 }
 
 impl Stream for DirectAddrsStream {
-    type Item = Vec<DirectAddr>;
+    type Item = BTreeSet<DirectAddr>;
 
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         let this = &mut *self;
         if let Some(addrs) = this.initial.take() {
             if !addrs.is_empty() {
-                return Poll::Ready(Some(addrs.iter().cloned().collect()));
+                return Poll::Ready(Some(addrs));
             }
         }
         loop {
@@ -2501,7 +2501,7 @@ impl Stream for DirectAddrsStream {
                         // returning Poll::Pending immediately here.
                         continue;
                     } else {
-                        break Poll::Ready(Some(addrs.iter().cloned().collect()));
+                        break Poll::Ready(Some(addrs));
                     }
                 }
                 Poll::Ready(None) => break Poll::Ready(None),
@@ -2845,7 +2845,11 @@ mod tests {
     #[instrument(skip_all)]
     async fn mesh_stacks(stacks: Vec<MagicStack>) -> Result<CallOnDrop> {
         /// Registers endpoint addresses of a node to all other nodes.
-        fn update_direct_addrs(stacks: &[MagicStack], my_idx: usize, new_addrs: Vec<DirectAddr>) {
+        fn update_direct_addrs(
+            stacks: &[MagicStack],
+            my_idx: usize,
+            new_addrs: BTreeSet<DirectAddr>,
+        ) {
             let me = &stacks[my_idx];
             for (i, m) in stacks.iter().enumerate() {
                 if i == my_idx {
@@ -3542,14 +3546,12 @@ mod tests {
         let ms = Handle::new(Default::default()).await.unwrap();
 
         // See if we can get endpoints.
-        let mut eps0 = ms.direct_addresses().next().await.unwrap();
-        eps0.sort();
+        let eps0 = ms.direct_addresses().next().await.unwrap();
         println!("{eps0:?}");
         assert!(!eps0.is_empty());
 
         // Getting the endpoints again immediately should give the same results.
-        let mut eps1 = ms.direct_addresses().next().await.unwrap();
-        eps1.sort();
+        let eps1 = ms.direct_addresses().next().await.unwrap();
         println!("{eps1:?}");
         assert_eq!(eps0, eps1);
     }
--- a/iroh-net/src/magicsock.rs
+++ b/iroh-net/src/magicsock.rs
@@ -1998,26 +1998,20 @@ impl Actor {
         // First add PortMapper provided addresses.
         let maybe_port_mapped = *portmap_watcher.borrow();
         if let Some(portmap_ext) = maybe_port_mapped.map(SocketAddr::V4) {
-            addrs.insert(
-                portmap_ext,
-                DirectAddr {
-                    addr: portmap_ext,
-                    typ: DirectAddrType::Portmapped,
-                },
-            );
+            addrs.entry(portmap_ext).or_insert(DirectAddr {
+                addr: portmap_ext,
+                typ: DirectAddrType::Portmapped,
+            });
             self.set_net_info_have_port_map();
         }
 
         // Next add STUN addresses from the netcheck report.
         if let Some(netcheck_report) = netcheck_report {
             if let Some(global_v4) = netcheck_report.global_v4 {
-                addrs.insert(
-                    global_v4.into(),
-                    DirectAddr {
-                        addr: global_v4.into(),
-                        typ: DirectAddrType::Stun,
-                    },
-                );
+                addrs.entry(global_v4.into()).or_insert(DirectAddr {
+                    addr: global_v4.into(),
+                    typ: DirectAddrType::Stun,
+                });
 
                 // If they're behind a hard NAT and are using a fixed
                 // port locally, assume they might've added a static
@@ -2031,23 +2025,17 @@ impl Actor {
                 {
                     let mut addr = global_v4;
                     addr.set_port(port);
-                    addrs.insert(
-                        addr.into(),
-                        DirectAddr {
-                            addr: addr.into(),
-                            typ: DirectAddrType::Stun4LocalPort,
-                        },
-                    );
+                    addrs.entry(addr.into()).or_insert(DirectAddr {
+                        addr: addr.into(),
+                        typ: DirectAddrType::Stun4LocalPort,
+                    });
                 }
             }
             if let Some(global_v6) = netcheck_report.global_v6 {
-                addrs.insert(
-                    global_v6.into(),
-                    DirectAddr {
-                        addr: global_v6.into(),
-                        typ: DirectAddrType::Stun,
-                    },
-                );
+                addrs.entry(global_v6.into()).or_insert(DirectAddr {
+                    addr: global_v6.into(),
+                    typ: DirectAddrType::Stun,
+                });
             }
         }
 
@@ -2096,13 +2084,10 @@ impl Actor {
                         };
                         if let Some(port) = port_if_unspecified {
                             let addr = SocketAddr::new(ip, port);
-                            addrs.insert(
+                            addrs.entry(addr).or_insert(DirectAddr {
                                 addr,
-                                DirectAddr {
-                                    addr,
-                                    typ: DirectAddrType::Local,
-                                },
-                            );
+                                typ: DirectAddrType::Local,
+                            });
                         }
                     }
                 }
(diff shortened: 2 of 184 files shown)


iroh-net/src/magicsock.rs
```diff
@@ -1548,46 +1537,6 @@ impl Handle {
     }
 }
 
-/// Stream returning local endpoints as they change.
```
flub, line 1551: This code is only moved, not changed.

iroh-net/src/magicsock.rs
```diff
@@ -1638,34 +1587,6 @@ enum DiscoBoxError {
 
 type RelayRecvResult = Result<(PublicKey, quinn_udp::RecvMeta, Bytes), io::Error>;
 
-/// Reports whether x and y represent the same set of endpoints. The order doesn't matter.
```
flub, line 1641: This code is removed.  This custom eq was needed to make the Watchable behave correctly, but also caused the bug that the updated_at field of the DiscoveredDirectAddrs was never updated.

iroh-net/src/magicsock.rs
```diff
@@ -2051,62 +1972,73 @@ impl Actor {
...
-                    $addrs.push(DirectAddr {
-                        addr: $ipp,
-                        typ: $typ,
-                    });
-                }
-            };
-        }
+        // We only want to have one DirectAddr for each SocketAddr we have.  So we store
+        // this as a map:
+        let mut addrs: BTreeMap<SocketAddr, DirectAddr> = BTreeMap::new();
```
ramfox, line 1996: nitpick: since `DirectAddr` is just a `SocketAddr` and `DirectAddrType`, couldn't `addrs` be a `BTreeMap<SocketAddr, DirectAddrType>` & then you can map & collect it when you call `msock.store_direct_addresses`?

iroh-net/src/magicsock.rs
```diff
@@ -2518,56 +2420,97 @@ fn bind(
...
+
+/// Stream returning local endpoints as they change.
+#[derive(Debug)]
+pub struct DirectAddrsStream {
+    initial: Option<BTreeSet<DirectAddr>>,
+    inner: watchable::WatcherStream<BTreeSet<DirectAddr>>,
+}
+
+impl Stream for DirectAddrsStream {
+    type Item = Vec<DirectAddr>;
```
ramfox, line 2485: `type Item = BTreeSet<DirectAddr>;`?

ramfox (approved):
looks good to me!

There are some weird cross comp failing tests, but that doesn't look like it started with this PR.

iroh-net/src/magicsock.rs
flub: this is an excellent suggestion!

iroh-net/src/magicsock.rs
flub: Yeah, this is a fair question.  I left this alone as it changes the public facing API, but it also isn't a big deal to break this I guess.

I really don't know what the right thing is to do here.  Generally direct addresses are going to be few, let's say <10, and in Rust this means a Vec is probably the most efficient data structure for this if you were trying to eek out every bit of perf.

However **logically** it is a set, which is why internally I changed it since it made it easier to reason about and is not on a hot path anyway.  So you're probably right that the same is true for the public API and that conveying that this is a set does make the API more expressive.

iroh-net/src/magicsock.rs
```diff
@@ -2518,56 +2420,97 @@ fn bind(
...
+    fn sockaddrs(&self) -> BTreeSet<SocketAddr> {
+        self.addrs.read().iter().map(|da| da.addr).collect()
     }
 
     fn fresh_enough(&self) -> bool {
-        match self.updated_at.as_ref() {
+        match *self.updated_at.read().expect("poisoned") {
             None => false,
             Some(time) => time.elapsed() <= ENDPOINTS_FRESH_ENOUGH_DURATION,
         }
```
matheus23, line 2457:
```suggestion
        self.last_refresh_ago().map_or(false, |dur| dur <= ENDPOINTS_FRESH_ENOUGH_DURATION)
```

iroh-net/src/magicsock.rs
```diff
@@ -2051,62 +1972,73 @@ impl Actor {
...
-                #[allow(clippy::map_entry)]
-                if !$already.contains_key(&$ipp) {
-                    $already.insert($ipp, $typ);
-                    $addrs.push(DirectAddr {
-                        addr: $ipp,
-                        typ: $typ,
-                    });
-                }
-            };
-        }
```
matheus23, line 2082: omg I'm happy this is gone finally.

matheus23 (approved):
About imports:
We could/perhaps should add a `.rustfmt.toml` file configured to your personal import style. (Let's also talk about it with the team first ;) )
Specifically, there's the [`imports_granularity`](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#imports_granularity) and [`group_imports`](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#group_imports) settings.
```toml
edition = "2021"
imports_granularity = "Module"
group_imports = "StdExternalCrate"
```

iroh-net/src/magicsock.rs
matheus23: Personally I think I'm in favor of a set here.

We could even keep the ordering info by having a custom `Ord` impl on `DirectAddr` that first checks the `DirectAddrType`, then the `SocketAddr`. (We might have to only swap the field order for that to happen)

rklaehn: A BTreeSet will basically use a vec internally anyway for anything below some threshold. So a BTreeSet with 2 or 3 elements is effectively a vec (single allocation IIRC). It is not something pointer-heavy like a red/black tree.

In any case I think the efficiency is fine in both cases, so using a set to express that duplicates don't make sense here is best in my opinion.

flub: Ok, did this now.  Of course this is now a breaking change.  I didn't propagate this change into the `iroh` crate.

iroh-net/src/magicsock.rs
flub: TIL about `Option::map_or`.  I guess it's shorter than `.map().unwrap_or()`, not yet convinced I like it.

Anyway, this is subjective.  I don't have anything against this suggestion but also don't particularly like it.  Probably because I'm not really happy with the `last_refresh_ago` name...

flub:
> About imports: We could/perhaps should add a `.rustfmt.toml` file configured to your personal import style. (Let's also talk about it with the team first ;) ) Specifically, there's the [`imports_granularity`](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#imports_granularity) and [`group_imports`](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#group_imports) settings.
> 
> ```toml
> edition = "2021"
> imports_granularity = "Module"
> group_imports = "StdExternalCrate"
> ```

I would love this! But they're not on stable rust.

iroh-base/src/node_addr.rs
```diff
@@ -74,13 +74,13 @@ impl NodeAddr {
     pub fn from_parts(
         node_id: PublicKey,
         relay_url: Option<RelayUrl>,
-        direct_addresses: Vec<SocketAddr>,
+        direct_addresses: BTreeSet<SocketAddr>,
```
divagant-martian, line 77: Could we receive iterators instead? All calls already are performed over iterators. 
```suggestion
        direct_addresses: impl IntoIterator<Item = SocketAddr>,
```
along with a call to `BTreeSet::from_iter`

With this most calls become a bit cleaner.
For example:
```diff
-            node: NodeAddr::from_parts(peer, relay_url, BTreeSet::from_iter([addr])),
+            node: NodeAddr::from_parts(peer, relay_url, [addr]),
```

iroh-net/src/magicsock.rs
ramfox: it's barely english, so it's understandable that you aren't happy with it 😂

iroh-net/src/magicsock.rs
```diff
@@ -51,25 +45,22 @@ use tracing::{
...
+use crate::disco::{self, CallMeMaybe, SendAddr};
+use crate::discovery::{Discovery, DiscoveryItem};
+use crate::dns::DnsResolver;
+use crate::endpoint::NodeAddr;
+use crate::key::{PublicKey, SecretKey, SharedSecret};
+use crate::net::ip::LocalAddresses;
+use crate::net::{interfaces, netmon};
+use crate::relay::{RelayMap, RelayUrl};
+use crate::{netcheck, portmapper};
+use crate::{stun, AddrInfo};
```
divagant-martian, line 58: uhh do we prefer the import granularity of `module`? I'm sure `crate` is far more common

iroh-net/src/magicsock.rs
```diff
@@ -38,11 +34,9 @@ use iroh_metrics::{inc, inc_by};
 use quinn::AsyncUdpSocket;
 use rand::{seq::SliceRandom, Rng, SeedableRng};
```
divagant-martian, line 35: note that this is inconsistent with the rest

iroh-net/src/magicsock.rs
```diff
@@ -1225,10 +1213,14 @@ impl MagicSock {
         }
     }
 
+    /// Sends the call-me-maybe DISCO message, queuing if addresses are too stale.
+    ///
+    /// To send the call-me-maybe message we need to know our current direct addresses.  If
+    /// this information is too stale the call-me-maybe is queued while a netcheck run is
+    /// scheduled.  Once this run finishes the call-me-maybe will be sent.
```
divagant-martian, line 1220:
```suggestion
    /// To send the call-me-maybe message, we need to know our current direct addresses.  If
    /// this information is too stale, the call-me-maybe is queued while a netcheck run is
    /// scheduled.  Once this run finishes, the call-me-maybe will be sent.
```

iroh-net/src/magicsock.rs
```diff
@@ -1225,10 +1213,14 @@ impl MagicSock {
...
     fn send_or_queue_call_me_maybe(&self, url: &RelayUrl, dst_node: NodeId) {
-        let direct_addrs = self.direct_addrs.read();
-        if direct_addrs.fresh_enough() {
-            let msg = direct_addrs.to_call_me_maybe_message();
+        if self.direct_addrs.fresh_enough() {
```
divagant-martian, line 1222: I think this changes the semantics. Before, this method worked on one copy of the direct addresses, now there is a read access per call which can potentially lead to different values. So, tldr I think this is racy. Or at least seen only from the perspective of this function. If there is some magic above preventing multiple accesses to this, I'm not sure

divagant-martian (commented):
To accelerate the process, sending an incomplete review so that you can act on this.

Also, sorry to be so pedantic, but a good chunk of the changes are unrelated to the pr title and have diverted into conversations that should not keep these changes from going through, like import formatting. Would you mind splitting this into its own PR?

iroh-net/src/magicsock.rs
flub: moved to https://github.com/n0-computer/iroh/pull/2812

iroh-net/src/magicsock.rs
```diff
@@ -15,19 +15,15 @@
...
-};
+use std::collections::{BTreeMap, BTreeSet, HashMap};
+use std::fmt::Display;
+use std::io;
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6};
+use std::pin::Pin;
+use std::sync::atomic::{AtomicBool, AtomicU16, AtomicU64, Ordering};
+use std::sync::{Arc, RwLock};
+use std::task::{Context, Poll, Waker};
+use std::time::{Duration, Instant};
```
divagant-martian: changes still here

iroh-net/src/magicsock.rs
flub: no longer true, the new code yields `BTreeSet<DirectAddr>` instead of `Vec<DirectAddr>`

iroh-base/src/node_addr.rs
flub: Yes, this works out well and is a much nicer API.  Great idea!

iroh-net/src/magicsock.rs
flub: formatter now takes care of this :tada:

iroh-net/src/magicsock.rs
flub: formatter now takes care of this :tada:

iroh-net/src/magicsock.rs
flub: Firstly, I don't think this should rely on magic outside of the function so I'm with you with ignoring what happens in the callers.

Looking at the code however I think the only thing which is racy is the debug log of `last_refresh_ago`.  If someone updates the direct addresses after the `fresh_enough` call says it's fresh enough... it still is fresh enough.  So I don't think calling `to_call_me_maybe_messages` separate from calling `fresh_enough` is a problem and am tempted to leave that.

flub: I changed the API of `fresh_enough` to return `Result<(), Duration>` to fix this.  Also gets rid of the horrible `last_fresh_ago` method :smile: 

I appreciate this is not the best code.  I my current thinking to do this thoroughly is to split this whole direct address management into an actor and have that actor just keep it's state (splitting off all that state from the magicsock will already be grand!) and refresh this stuff.  Whenever is has new direct addresses it should send them to the magicsock so it can store those for access on the hot path.  that actor itself can basically supervise itself: either it continues to run and provides addresses or it panics and brings down the process - i dont' think there needs to be an in-between (bear in mind "fresh enough" is 27 seconds).

(I also think that the interfaces information should be in an actor that behaves somewhat similar, but that's very tangential to this.)

iroh-net/src/magicsock.rs
```diff
@@ -2117,98 +2058,67 @@ impl Actor {
...
+                            _ => None,
+                        };
+                        if let Some(port) = port_if_unspecified {
+                            let addr = SocketAddr::new(ip, port);
+                            addrs.entry(addr).or_insert(DirectAddrType::Local);
                         }
                     }
                 }
 
+                // If a socket is bound to a specific address, add it.
```
divagant-martian, line 2094: nice comma

divagant-martian (approved):
LGTM

PR #2809, opened 2024-10-16
feat: persistent store impl using `willow_store` and `redb`

matheus23:
At the moment, this replaces the memory store in `willow` with the tree in `willow-store` + its `MemStore`.

Next steps:
- [x] Clean up
  - [x] Move `willow-store`/`willow-rs` glue into one sane place:
    - [x] conversions: `BlobSeq` <-> `Path`, `AuthorisedEntry` <-> `StoredAuthorizedEntry`, `QueryRange3d` <-> `Range3d`
    - [x] definitions: `StoredAuthorizedEntry`, `WillowParams`
  - [x] Fix spelling (Authori**s**ed vs. Authori**z**ed is inconsitent at the moment)
  - [x] Make use of `NamespaceStore` again
- [x] Create a store based on `willow_store::RedbBlobStore` instead (this way it's persistent)
- [x] Think about undoing changes to the iroh-willow memory store? Or perhaps make it use redb in memory mode? TBD.
- [x] Subscriptions and subscription resumption
  - ~~Need access to lower-level details of willow-store to be able to use the trick of using `NodeId`s for `progress_id`s.~~
    Went with the old "live subscriptions only" thing again


rklaehn:
I guess the willow PR and this PR need to go against a separate iroh-willow repo now?

matheus23:
Yeah - I'll finish this work and then move this to a separate repository.

matheus23:
I guess part of this is I have to look at how exactly y'all will do the RPC client thing with iroh-docs. Then I'll mirror that with iroh-willow's `spaces`.

PR #2812, opened 2024-10-17
chore: Format imports using rustfmt

flub:
## Description

This enables a strict style in rustfmt to format imports statements.  It needs rust nightly for rustfmt to be able to do this.

To not make the `cargo fmt` from stable rust really noisy it does not add a rustfmt.toml file, instead it adds a `cargo-make` task to automate running this locally.  This same task is used in CI to run the formatter.

The `Crate`m formatting will be preserved by the default configuration of rust-analyzer, which makes external contributions easier as well, most folks shouldn't be running into this formatter.

## Notes & open questions

I guess we need buy-in from everyone on the team for this.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-blobs/src/store/bao_file.rs
+++ b/iroh-blobs/src/store/bao_file.rs
@@ -9,25 +9,33 @@
 //! the file system for the data, outboard, and sizes file. There is also a
 //! combined implementation that starts in memory and switches to file when
 //! the memory limit is reached.
-use std::fs::{File, OpenOptions};
-use std::io;
-use std::ops::{Deref, DerefMut};
-use std::path::{Path, PathBuf};
-use std::sync::{Arc, RwLock, Weak};
+use std::{
+    fs::{File, OpenOptions},
+    io,
+    ops::{Deref, DerefMut},
+    path::{Path, PathBuf},
+    sync::{Arc, RwLock, Weak},
+};
 
-use bao_tree::io::fsm::BaoContentItem;
-use bao_tree::io::outboard::PreOrderOutboard;
-use bao_tree::io::sync::{ReadAt, WriteAt};
-use bao_tree::BaoTree;
+use bao_tree::{
+    io::{
+        fsm::BaoContentItem,
+        outboard::PreOrderOutboard,
+        sync::{ReadAt, WriteAt},
+    },
+    BaoTree,
+};
 use bytes::{Bytes, BytesMut};
 use derive_more::Debug;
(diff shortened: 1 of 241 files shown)


iroh-blobs/src/store/bao_file.rs
```diff
@@ -9,34 +9,25 @@
...
+use std::fs::{File, OpenOptions};
+use std::io;
+use std::ops::{Deref, DerefMut};
+use std::path::{Path, PathBuf};
+use std::sync::{Arc, RwLock, Weak};
+
+use bao_tree::io::fsm::BaoContentItem;
+use bao_tree::io::outboard::PreOrderOutboard;
+use bao_tree::io::sync::{ReadAt, WriteAt};
+use bao_tree::BaoTree;
```
divagant-martian, line 21: do you know what ordering is going on here? it is not alphabetic

matheus23:
image
Lol

iroh-blobs/src/store/bao_file.rs
flub: It puts direct items always below sub-imports i think.

I don't really mind, as long as it's consistent in what it does :)

flub:
FWIW, personally I like `module` style.  But I'd also be really happy if `crate` style manages to be consistently formatted.  So I'm totally open to adopt that if it's preferred.  I don't want to die on a hill about what formatting is used.  I only want to die on the hill of consistency.

divagant-martian (approved):
I updated the pr to not depend on nightly. 

Also pushed changes that are incoming on the nightly channel but are accepted by check.

Also changed the or to `Crate`, which a couple of us seem to prefer.. so obviously LGTM there.

Also did a couple tries of removing imports and asking rust-analyzer do add them back without further configuration and it behaves how we want. Also tested that `cargo make format` produces no diff locally.

Happy we are improving our code quality and standards

iroh-blobs/src/store/bao_file.rs
divagant-martian: I tried changing the order to some other stuff and it revert to whatever this is, so it seems to be consistent even if I don't fully get the order definition

flub (commented):
LGTM, let's merge this already before it gets out of date!

(can't approve my own PR)

dignifiedquire (approved):

divagant-martian:
Merging, as internal discussions point to an (at least by majority) consensus

PR #2825, opened 2024-10-23
feat(iroh-net): Add StaticDiscovery to provide static info to endpoints

rklaehn:
## Description

Add StaticProvider to provide static info to endpoints.

The long term plan is to reduce ways to interact directly with the endpoint node map and instead provide this.

## Breaking Changes

None, this just adds stuff. Long term we want to remove endpoint.add_node_info with this.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


iroh-net/src/discovery/static_discovery.rs
```diff
@@ -0,0 +1,126 @@
...
+
+    /// Get node info for the given node id.
+    pub fn get_node_addr(&self, node_id: NodeId) -> Option<NodeAddr> {
+        let guard = self.nodes.read().unwrap();
+        let info = guard.get(&node_id).cloned()?;
+        Some(NodeAddr { node_id, info })
+    }
+
+    /// Remove node info for the given node id.
+    pub fn remove_node_addr(&self, node_id: NodeId) {
```
dignifiedquire, line 97: return the existing data?

iroh-net/src/discovery/static_discovery.rs
```diff
@@ -0,0 +1,126 @@
...
+        node_id: NodeId,
+    ) -> Option<futures_lite::stream::Boxed<anyhow::Result<super::DiscoveryItem>>> {
+        let guard = self.nodes.read().unwrap();
+        let info = guard.get(&node_id);
+        match info {
+            Some(addr_info) => {
+                let item = DiscoveryItem {
+                    node_id,
+                    provenance: Self::PROVENANCE,
+                    last_updated: None,
```
dignifiedquire, line 118: you could store the timestamp of insertion, and use it here

iroh-net/src/discovery/static_discovery.rs
```diff
@@ -0,0 +1,126 @@
...
+    key::NodeId,
+    node_addr::{AddrInfo, NodeAddr},
+};
+
+use super::{Discovery, DiscoveryItem};
+
+/// A static discovery implementation that allows adding info for nodes manually.
+#[derive(Debug, Default)]
+#[repr(transparent)]
+pub struct StaticDiscovery {
```
flub, line 18: bikeshedding: at some point we complained that all the discovery services have confusing names and that they should be consistently named something like `ThingPublisher` or `ThingProvider`.  Calling them `ThingDiscovery` is redundant as it already is in the discovery module.

So `StaticProvider`?

iroh-net/src/discovery/static_discovery.rs
rklaehn: Not sure about that. I think we don't have a single discovery service that sets this field, and also it is unused in code. Maybe just remove it?

dignifiedquire: @ramfox @flub thoughts on this field?

flub: perhaps it should be used?  quoting ramfox's code comment:

> It is only truly relevant when comparing `relay_urls`, since we can attempt to dial any number of socket addresses, but expect each node to have one "home relay" that we will attempt to contact them on. This means we would need some way to determine which relay url to choose between, if more than one relay url is reported

So I would tentatively leave this in I think.  I can imagine you adding one RelayUrl statically, later the node publishing to DNS and providing a different one.  I think that's a situation we should probably handle even if we don't right now.

rklaehn: I added filling out this number, despite nobody using it as of now...

iroh-net/src/discovery/static_provider.rs
```diff
@@ -0,0 +1,162 @@
+//! A static discovery implementation that allows adding info for nodes manually.
```
flub, line 1: This should really explain **why** this thing exists.  In what situations do you want to use this?  Maybe this should include an example using a Ticket even.  It could suggest that you may want to remove items when you know they're stale for some reason.  etc.

All the same things also need to be on `StaticProvider` directly, so it's fine to only write this down once and refer to the other place for the other.

iroh-net/src/discovery/static_provider.rs
```diff
@@ -0,0 +1,162 @@
...
+}
+
+#[derive(Debug)]
+struct NodeInfo {
+    info: AddrInfo,
+    last_updated: SystemTime,
+}
+
+impl StaticProvider {
+    /// The provenance string for this discovery implementation.
```
flub, line 30: This should say what it is for.  Like this it is completely lost on the user.

e.g. local swarm discovery uses this text:
```
/// Used as the `provenance` field in [`DiscoveryItem`]s.
///
/// Used in the [`crate::endpoint::Source::Discovery`] enum variant as the `name`.
```

I like having it as an associated constant though :)

iroh-net/src/discovery/static_provider.rs
```diff
@@ -0,0 +1,162 @@
...
+    /// ```
+    pub fn from_node_addrs(infos: impl IntoIterator<Item = impl Into<NodeAddr>>) -> Self {
+        let res = Self::default();
+        for info in infos {
+            res.add_node_addr(info);
+        }
+        res
+    }
+
+    /// Add node info for the given node id.
```
flub, line 69: `Adds`

iroh-net/src/discovery/static_provider.rs
```diff
@@ -0,0 +1,162 @@
...
+                last_updated,
+            },
+        );
+        previous.map(|x| NodeAddr {
+            node_id: info.node_id,
+            info: x.info,
+        })
+    }
+
+    /// Add node info for the given node id, combining it with any existing info.
```
flub, line 89: `Adds`

iroh-net/src/discovery/static_provider.rs
```diff
@@ -0,0 +1,162 @@
...
+            Entry::Vacant(entry) => {
+                entry.insert(NodeInfo {
+                    info: info.info,
+                    last_updated,
+                });
+            }
+        }
+    }
+
+    /// Get node info for the given node id.
```
flub, line 115: `Returns node info`

iroh-net/src/discovery/static_provider.rs
```diff
@@ -0,0 +1,162 @@
...
+    pub fn get_node_addr(&self, node_id: NodeId) -> Option<NodeAddr> {
+        let guard = self.nodes.read().unwrap();
+        let info = guard.get(&node_id).map(|x| x.info.clone())?;
+        Some(NodeAddr { node_id, info })
+    }
+
+    /// Remove node info for the given node id.
```
flub, line 122: `Removes node info`

flub (commented):
Oh, this went from draft to merged very quickly... I missed the "open" window.  In draft I'm only looking for how the design fits, but not really paying much attention to details.

I think we need to be better on docs for stuff landing in main... don't want to be keeping playing catchup with docs fridays forever.

PR #2830, opened 2024-10-23
refactor: move iroh-docs to external repo

dignifiedquire:
the source now lives in https://github.com/n0-computer/iroh-docs


rklaehn (approved):
I can confirm that this removes iroh-docs and updates ci. Did not check that the new repo actually contains the right stuff though...

PR #2831, opened 2024-10-23
refactor(iroh): move protocol relevant impls into node/protocols

dignifiedquire:
Some internal refactorings


rklaehn:
I think we should use a rpc protocol that is relatively isolated (like gossip, not docs (huge) and not blobs (mixed with tags)) and see if we can move the rpc protocol definition to the iroh-gossip crate and the client handler under a feature flag.

PR #2837, opened 2024-10-25
split network utilities into their own crates

divagant-martian:
## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


net-tools/net/Cargo.toml
flub, line 1: I imagine we'll further split this up broadly as:
- interfaces & LocalAddresses (which is a view of iterfaces really)
- udp socket

And then somehow figure out where the remaining small bits go.

But leaving this as one "net" is fine for now I think.  Small steps are easier.

divagant-martian: yeah it's not great, and the name is not great either, but it's the best way to progress on this for the time being

dignifiedquire: we will need sth else for next week, otherwise we can’t release it (I don’t think net is free on crates.io)

dignifiedquire:
@divagant-martian can this be closedß

PR #2838, opened 2024-10-25
feat(iroh-net): More Quinn re-exports

flub:
## Description

These are some more public types that we need to re-export.  I needed these while getting h3 to work and then added a more comprehensive review adding all types.

## Notes & open questions

The types are for now simply exported, they expose too much functionality that we'll still want to reduce as we approach 1.0.  But having them re-exported gives a good starting point that shows us the surface area we currently expose.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


dignifiedquire:
> * quinn:rustls

I would try to avoid this, given users of `iroh-net` should not touch rustls directly

dignifiedquire (approved):

PR #2873, opened 2024-10-31
feat!: add iroh-relay crate

divagant-martian:
## Description

Adds a new crate `iroh-relay`. It's structured as follows:
- protos: protocols the server is able to handle, this includes the relay protocol, stun, and partially disco.
- server: the server code, as it was already in the og mod
- client: the client code, as it was already in the og mod
- server binary (bin `iroh-relay`) as it was previously in iroh-net's bin.

This moves the code around in the least disruptive possible way to give us a starting point over which to improve this. Check the notes and open questions for future work

## Breaking Changes

- `iroh_net::relay` is removed. `RelayUrl`, `RelayMode`, `RelayNode` and `RelayMap` are moved to the top (`iroh_net`). All other members of this module are now moved to the new crate `iroh-relay`

## Notes & open questions

- Disco includes a transaction id defined in terms of stun's transaction id. Every use of this is now directly from `stun_rs` to make it clear this is not defined in terms of the relay's stun server. The relay can't read the contents of disco packages so I consider this the right thing to do, even if the types are actually the same.
- The relay needs to identify Disco packages to queue them separately. Disco is not yet its own crate because we do not know its future. Based on this I chose to duplicate the code that does the identification. But just this signals we should start the work on deciding what disco looks like in a post ts world to de-duplicate this code and use it from a single source. The code as it stands is highly unlikely to change; however, any change to the protocol's wrapper would now need to be changed in two places, which is ofc not ideal.
- The relay's "protocol" includes quite a lot of non protocol code, that probably belongs more to the server. I was really tempted to try to move this code to better places but in reality this is not the time. The purpose of this PR is among many others, to allow for future work like this.
- What used to be called the relay module had a lot of relay-related code without a clear objective or offering. This includes for example, all the relay map related code. The relay as a protocol, server, and client, has absolutely no need for these types. Relay topology is entirely a `iroh`/`iroh-net` topic. On the same page is the infra-related code. Nothing in the protocol, server or client is different depending on the infra values. This is entirely a `iroh`/`iroh-net` topic. Based on these arguments, these types, functions, etc that were part of the relay module have actually stayed in `iroh-net` instead.
- I did a couple passes of `udeps`, but it's messy. Optional dependencies are their own feature so they aren't flagged as unused. There are (in general, not in the relay) dependencies that differ in targets, and the features combinations also affect this as well. So the deps that remain are a "reasonable best effort" in accordance with our CI checks. In reality, - and in particular when checking feature combinations with `fc` - there's a lot of work left to do on that front.
- The crate's version is 0.29 because there's no point in setting the version to 0.28, which will never be released, to then later be changed to 0.29. This might be unexpected but it's the option that reduces senseless future work.
- dns code is duplicated. This is unfortunate but since it's only used in tests to configure the client there's really not much to do here. Most of what this does are fixes we should try to upstream

## Change checklist

- [x] Self-review.
- [ ] ~~Tests if relevant.~~


iroh-net/Cargo.toml
```diff
@@ -40,6 +40,7 @@ hyper = { version = "1", features = ["server", "client", "http1"] }
 hyper-util = "0.1.1"
 igd-next = { version = "0.15.1", features = ["aio_tokio"] }
 iroh-base = { version = "0.28.0", features = ["key"] }
+iroh-relay = { path = "../iroh-relay" }
```
dignifiedquire, line 43: missing version

iroh-net/Cargo.toml
```diff
@@ -134,7 +135,7 @@ harness = false
 duct = "0.13.6"
 
 [features]
-default = ["metrics", "discovery-pkarr-dht"]
+default = ["metrics", "discovery-pkarr-dht", "iroh-relay"]
 iroh-relay = [
```
dignifiedquire, line 139: missing `dep:iroh-relay`

iroh-relay/Cargo.toml
```diff
@@ -0,0 +1,96 @@
+[package]
+name = "iroh-relay"
+version = "0.1.0"
+edition = "2021"
+
```
dignifiedquire, line 5: missing license and other fields

iroh-relay/Cargo.toml
```diff
@@ -0,0 +1,96 @@
+[package]
+name = "iroh-relay"
+version = "0.1.0"
```
dignifiedquire, line 3: we probably want to match this version to the other iroh crates, eg `0.28.0`

dignifiedquire:
did a first read through, and I think structure wise this all looks good to me 👍

iroh-net/src/lib.rs
```diff
@@ -243,14 +243,16 @@ mod magicsock;
...
+mod relay_map;
 pub mod ticket;
 pub mod tls;
+
 pub(crate) mod util;
 
 pub use endpoint::{AddrInfo, Endpoint, NodeAddr};
 pub use iroh_base::{key, key::NodeId};
+pub use iroh_relay as relay;
+pub use relay_map::{RelayMap, RelayMode, RelayNode, RelayUrl};
```
matheus23, line 255: Ah I see now.

Initially I thought "why do `mod relay_map` & `pub use relay_map::*` instead of just keeping a `pub mod relay`?", but now I see the `pub use iroh_relay as relay` re-export below.

:thinking: that said, I think it should be possible to write a `relay.rs` module that re-exports all of `iroh_relay` at its own level & also has `RelayMap`, `RelayMode`, `RelayNode` & `RelayUrl`.
That would be slightly less breakage for users. May or may not be worth it?

iroh-net/src/netcheck.rs
```diff
@@ -781,18 +781,150 @@ mod tests {
...
+        /// After dropping the test infrastructure will asynchronously shutdown and release its
+        /// resources.
+        // Nightly sees the sender as dead code currently, but we only rely on Drop of the
+        // sender.
+        #[derive(Debug)]
+        pub struct CleanupDropGuard {
+            _guard: oneshot::Sender<()>,
+        }
+
+        // (read_ipv4, read_ipv5)
```
matheus23, line 815: While we're here
```suggestion
        // (read_ipv4, read_ipv6)
```

iroh-relay/src/lib.rs
```diff
@@ -0,0 +1,30 @@
+//! Package `relay` implements a revised version of the Designated Encrypted Relay for Packets
+//! (DERP) protocol written by Tailscale.
```
matheus23, line 2: This is now the first thing that people see when they open the docs for the relay package.

Given that this is way more prominent now, perhaps it's worth adjusting this to the wording used in the README, which personally I found more fitting?

iroh-relay/src/client.rs
```diff
@@ -2,6 +2,7 @@
```
matheus23, line 2: (I'm lacking a good place to comment)

Same reasoning as in the other comment: The module comment is now much more prominent (just a click away from the top-level iroh-relay crate docs), I think even just something like "Code for establishing connections to a relay server" is more useful.

matheus23 (approved):
Some (perhaps optional) nits, but I'm happy with this overall :)

`cargo deny` is not happy, but not much we can do about it right now.

dignifiedquire:
please make the version 0.28, because when I run the release with cargo-release, bumps all happen automatically and if it is 0.29 I need to manually exclude the crate

flub (approved):

flub:
> please make the version 0.28, because when I run the release with cargo-release, bumps all happen automatically and if it is 0.29 I need to manually exclude the crate

I'm surprised this needs to be in sync?  I thought the reason to split these crates out was so they could stay <1.0 while the main crates get 1.0.  Short term of course this doesn't matter.

iroh-net/src/netcheck.rs
```diff
@@ -812,7 +812,7 @@ mod tests {
             _guard: oneshot::Sender<()>,
         }
 
-        // (read_ipv4, read_ipv5)
+        // (read_ipv4, read_ipv6)
```
dignifiedquire, line 815: 💙

iroh-relay/src/client.rs
```diff
@@ -1,3 +1,5 @@
+//! Exposes [`Client`], which allows to establish connections to a relay server.
+//!
 //! Based on tailscale/derp/derphttp/derphttp_client.go
```
matheus23, line 3: Sorry, can we make this at least a non-documentation comment? :sweat_smile:

divagant-martian: I would rather not remove it but I get why people would prefer to not have it as crate docs.. in order to make this merge ready I did the change

.github/workflows/tests.yaml
```diff
@@ -23,7 +23,7 @@ env:
   RUSTFLAGS: -Dwarnings
   RUSTDOCFLAGS: -Dwarnings
   SCCACHE_CACHE_SIZE: "50G"
-  CRATES_LIST: "iroh,iroh-metrics,iroh-net,iroh-net-bench,iroh-test,iroh-cli,iroh-dns-server,iroh-router,netwatch,portmapper"
+  CRATES_LIST: "iroh,iroh-metrics,iroh-net,iroh-net-bench,iroh-test,iroh-cli,iroh-dns-server,iroh-router,netwatch,portmapper,iroh-relay"
```
dignifiedquire, line 26: this also needs updating `ci.yaml`

iroh-relay/Cargo.toml
```diff
@@ -0,0 +1,102 @@
...
+name = "iroh-relay"
+version = "0.28.0"
+edition = "2021"
+readme = "README.md"
+description = "Iroh's relay server and client"
+license = "MIT OR Apache-2.0"
+authors = ["n0 team"]
+repository = "https://github.com/n0-computer/iroh"
+keywords = ["networking", "holepunching", "p2p"]
+
```
dignifiedquire, line 11: missing rust version

dignifiedquire (approved):

PR #2874, opened 2024-10-31
refactor(iroh)!: Move blobs and tags rpc client and server to iroh-blobs

rklaehn:
## Description

Move blobs and tags rpc client and server to iroh-blobs.

Depends on https://github.com/n0-computer/iroh-blobs/pull/7

Todo:

- [x] merge https://github.com/n0-computer/iroh-blobs/pull/7

## Breaking Changes

- Lots of types in client::blobs and client::tags become reexports
- blob share goes away, since it requires the node client

## Notes & open questions

I want to keep the client::blobs and client::tags modules self-contained, so the idea is that these will reexport all the things from iroh-blobs::rpc::client that a user will need (except I have probably forgotten something). Maybe I should use wildcard exports here even though people dislike them... ?

~~The client::blobs::Client itself is *not* a reexport but a newtype to hide the ugly type parameters. Same for client::tags::Client.~~

With the changes in quic-rpc, these are now just module reexports!

The Blobs protocol handler now takes an Endpoint, since that was needed to implement one of the functions.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh-cli/src/commands/blobs.rs
```diff
@@ -28,8 +29,7 @@ use iroh::{
     },
     client::{
         blobs::{
-            BlobInfo, BlobStatus, CollectionInfo, DownloadMode, DownloadOptions,
-            IncompleteBlobInfo, WrapOption,
+            BlobInfo, BlobStatus, CollectionInfo, DownloadOptions, IncompleteBlobInfo, WrapOption,
```
rklaehn, line 32: This will probably go away if docs are removed as well.

iroh/src/node/rpc.rs
```diff
@@ -186,61 +155,81 @@ impl<D: BaoStore> Handler<D> {
...
-            ListTags(msg) => chan.server_streaming(msg, self, Self::blob_list_tags).await,
-            DeleteTag(msg) => chan.rpc(msg, self, Self::blob_delete_tag).await,
-            Create(msg) => chan.rpc(msg, self, Self::tags_create).await,
-            Set(msg) => chan.rpc(msg, self, Self::tags_set).await,
-        }
-    }
+        self.blobs().handle_rpc_request(msg, chan).await
+    }
+
+    // async fn handle_blobs_request(
```
dignifiedquire, line 170: delete?

iroh/src/rpc_protocol/net.rs
```diff
@@ -37,6 +37,7 @@ pub enum Response {
     RemoteInfosIter(RpcResult<RemoteInfosIterResponse>),
     RemoteInfo(RpcResult<RemoteInfoResponse>),
     Watch(WatchResponse),
+    Unit(RpcResult<()>),
```
dignifiedquire, line 40: what's this?

rklaehn:
> what about removing the fs utils you added to `iroh-blobs`?

I think some of them are also used from docs, so I can't remove them all until docs are also gone... But will check.

iroh/src/rpc_protocol/net.rs
rklaehn: Some rpc calls return RpcResult<()>. We have to have a way to get the result back.

rklaehn:
> > what about removing the fs utils you added to `iroh-blobs`?
> 
> I think some of them are also used from docs, so I can't remove them all until docs are also gone... But will check.

Removed several, including stuff that needs walkdir. But there is still something left that is needed from docs.

dignifiedquire (approved):

PR #2881, opened 2024-11-01
feat(iroh-net): Send HTTP/1.1 `HOST` header on requests to relay

### LINKED ISSUE #2870: iroh-net does not send the `Host:` header to the relay
I am trying to put iroh-relay behind nginx: https://github.com/deltachat/chatmail/issues/445
Current problem is that nginx rejects the requests from iroh-net 0.25.0 with error 400.
According to the logs nginx gets `GET /relay HTTP/1.1`, then header `upgrade: iroh derp http`, then rejects because `client sent HTTP/1.1 request without "Host" header`.
I cannot make it proxy the request to iroh-relay.

HTTP/1.1 requires the `Host:` header and it seems to be impossible to make nginx proxy requests that don't have one.

I will try to upgrade iroh-net and see if it resolves the problem. EDIT: upgraded to 0.26.0 so far, and the problem remained. Trying to upgrade to 0.27.0, but it has its own problems: https://github.com/deltachat/deltachat-core-rust/pull/6133

matheus23:
## Description

Reverse proxies that map multiple virtual hosts need some help to know where to forward requests to.
That's what the [`HOST` header](https://datatracker.ietf.org/doc/html/rfc2616#section-14.23) is for, and it's mandatory for HTTP/1.1 requests.

Closes #2870 

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/src/relay/client.rs
+++ b/iroh-net/src/relay/client.rs
@@ -709,15 +709,7 @@ impl Actor {
     where
         T: AsyncRead + AsyncWrite + Send + Unpin + 'static,
     {
-        let relay_url_host = relay_url
-            .host_str()
-            .ok_or_else(|| ClientError::InvalidUrl(relay_url.to_string()))?;
-        let mut host_header_value = String::with_capacity(relay_url_host.len() + 6); // 6 chars for a ":" and max port
-        host_header_value += relay_url_host;
-        if let Some(port) = relay_url.port() {
-            host_header_value += ":";
-            host_header_value += &port.to_string();
-        }
+        let host_header_value = host_header_value(relay_url)?;
 
         let io = hyper_util::rt::TokioIo::new(io);
         let (mut request_sender, connection) = hyper::client::conn::http1::Builder::new()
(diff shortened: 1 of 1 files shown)


matheus23:
@link2xt please try out this PR and let us know if it helps :v:

matheus23:
I can see that reading strange to you.
Not all proxies need this HOST header to help them. Only the ones that map multiple virtual hosts.
And if I'm reading the history here correctly, they added mandatory support for HOST headers to make all client applications send them by default, so any virtual-host-mapping proxies out there just work.

iroh-net/src/relay/client.rs
```diff
@@ -696,10 +702,23 @@ impl Actor {
...
     where
         T: AsyncRead + AsyncWrite + Send + Unpin + 'static,
     {
+        let relay_url_host = relay_url
+            .host_str()
+            .ok_or_else(|| ClientError::InvalidUrl(relay_url.to_string()))?;
+        let mut host_header_value = String::with_capacity(relay_url_host.len() + 6); // 6 chars for a ":" and max port
+        host_header_value += relay_url_host;
+        if let Some(port) = relay_url.port() {
+            host_header_value += ":";
```
dignifiedquire, line 718: can you put this in a function and add a sanity test, please

dignifiedquire (approved):

PR #2882, opened 2024-11-03
ref(iroh-net)!: Make Endpoint::close not consume self

iacore:
## Description

The endpoint can be freely cloned, so consuming `self` on shutdown is somewhat fruitless.  This changes it to take `&self` like a normal method.

## Breaking Changes

- `Endpoint::close` now takes `&self` rather than `self`.  This can in some situations mean an existing (clone of an) endpoint might be dropped too early as a temporary variable.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh-net/src/endpoint.rs
```diff
@@ -853,14 +853,12 @@ impl Endpoint {
             endpoint,
             cancel_token,
             ..
-        } = self;
+        } = &self;
```
dignifiedquire, line 856: to actually change the behaviour of this function, this needs to take `&self` as the argument, so the function signature should become

```rust
pub async fn close(&self, error_code: VarInt, reason: &[u8]) -> Result<()>
 ```

Comment (iacore) [reply iroh-net/src/endpoint.rs:856]: Thanks for the review!

Comment (flub) [reply iroh-net/src/endpoint.rs:856]:
```suggestion
        } = self;
```
I don't think this change is needed

iroh/src/node.rs
```diff
@@ -465,7 +465,6 @@ impl<D: iroh_blobs::store::Store> NodeInner<D> {
             // connections: Operations will immediately fail with ConnectionError::LocallyClosed.
             // All streams are interrupted, this is not graceful.
             self.endpoint
-                .clone()
```
flub, line 468: lol, frankly this is probably sufficient reason to not make this move the endpoint.

flub (commented):
This needs to get the changes in main, and then check if the new doc comment needs tweaking because we need to make it clear how this works.  But in general I think this is probably a good change.

iroh-net/src/endpoint.rs
```diff
@@ -858,9 +858,7 @@ impl Endpoint {
         tracing::debug!("Closing connections");
         endpoint.close(error_code, reason);
         endpoint.wait_idle().await;
-        // In case this is the last clone of `Endpoint`, dropping the `quinn::Endpoint` will
-        // make it more likely that the underlying socket is not polled by quinn anymore after this
-        drop(endpoint);
```
flub, line 863: @divagant-martian does it make sense to store the `quinn::Endpoint` in an `Option` so that it can be taken out and dropped?  That would also mean that regardless how many clones there are of `iroh_net::Endpoint` they will all lose the `quinn::Endpoint` when closed, which can be used as a very good indicator that the endpoint is closed.

iacore: `Arc<Option<EndPoint>>`?

flub: having to put that into a mutex would not be great.  it already has mutexes inside it.

flub: ok, we should not try and drop the quinn endpoint here.

flub (approved):
LGTM to me, but would love @divagant-martian to lgtm too (or not)

divagant-martian:
this was in part because of how quinn used to work. Closing the magic socket before dropping the endpoint would basically always emit an error in logs breaking all our CI tests. The changes in the endpoint loop make this less likely to happen in this scenario. This is not to say the previous state was exactly what we needed: As the comment mentions, it's there to make it _less likely_ that the socket is pulled. In general closing here is messy and kinda gross. When we wrote this code (not sure now) the socket was the last resource the (quinn) endpoint dropped. This was later that the drop of the endpoint itself given the many cloned Arcs and spawned tasks.

All this long rant is to say: there was a reason this looked this way. To get a best effort imperfect clone because both how quinn works and the lack of async drop (and I thought the rant was over) but after we updated quinn I think the error is no longer an issue and we are fine with dropping the endpoint after the socket in most cases

flub:
Thanks @iacore for pushing this change!

PR #2895, opened 2024-11-05
docs: Format code in doc comments

flub:
## Description

This enables the unstable rustfmt option to format the code in docs.
Should make our examples a lot more consistent.

## Notes & open questions

This is unstable, so there might be some churn or bad results.  But I
don't think the churn should be much of an issue for us.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


dignifiedquire (approved):
nice

PR #2896, opened 2024-11-05
chore: Break long strings in the formatter

flub:
## Description

This configures rustfmt to break long strings.

## Notes & open questions

This one is probably more open to opinions.  I don't feel very
strongly about this, but thought the result does look a little nicer
so thought I'd submit it anyway.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-base/src/ticket/blob.rs
+++ b/iroh-base/src/ticket/blob.rs
@@ -168,14 +168,19 @@ mod tests {
             hash,
         };
         let base32 = base32::parse_vec(ticket.to_string().strip_prefix("blob").unwrap()).unwrap();
-        let expected = parse_hexdump("
+        let expected = parse_hexdump(
+            "
             00 # discriminator for variant 0
-            ae58ff8833241ac82d6ff7611046ed67b5072d142c588d0063e942d9a75502b6 # node id, 32 bytes, see above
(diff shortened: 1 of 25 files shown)


iroh-base/src/ticket/blob.rs
```diff
@@ -168,14 +168,19 @@ mod tests {
...
             00 # relay url
             00 # number of addresses (0)
             00 # format (raw)
-            0b84d358e4c8be6c38626b2182ff575818ba6bd3f4b90464994be14cb354a072 # hash, 32 bytes, see above
-        ").unwrap();
+            0b84d358e4c8be6c38626b2182ff575818ba6bd3f4b90464994be14cb354a072 # hash, 32 bytes, see \
+             above
+        ",
+        )
+        .unwrap();
```
flub, line 183: these changes are probably worse.  i don't think there's an opt-out for a block so best option is to re-word these to look nicer after formatting

dignifiedquire:
I am about -10 on this one (even though i regularly use quite narrow editor windows) I think the multiline notations in rust strings are just not great to read, and overall I dislike this much tweaking of `cargo fmt`

flub:
sure, one person being that many minus points is enough not to bother with this.

PR #2900, opened 2024-11-05
ci: Kill tests after 60s using nextest

flub:
## Description

By using nextest to kill slow tests we get logs when a test is killed. This is better than the github timeout kicking in as then we get no feedback on what was stuck.

Tests are now marked as slow after 20s and killed after 60s.

## Notes & open questions

discovery::tests::endpoint_discovery_combined_wrong_only seems like it is a slow tests needing about 30s to complete.  To be fair I think it's fine to call that out as a slow test, so I'm fine with that.

It is possible that this 60s timeout will introduce some new flakyness to the slow CI machines.  We'll have to see and potentially tweak this a bit.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -8,3 +8,6 @@ run-in-isolation = { max-threads = 32 }
 filter = 'test(::run_in_isolation::)'
 test-group = 'run-in-isolation'
 threads-required = 32
+
+[profile.default]
+slow-timeout = { period = "20s", terminate-after = 3 }
--- a/iroh-net/src/endpoint.rs
+++ b/iroh-net/src/endpoint.rs
@@ -1426,7 +1426,6 @@ mod tests {
         assert!(res.is_err());
         let err = res.err().unwrap();
         assert!(err.to_string().starts_with("Adding our own address"));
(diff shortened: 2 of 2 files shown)


Arqu:
60 seconds might be a bit too tight, some tests in cross can run between 60 and 120s. 
Also the cross tests you want to debug with this are not running with nextest.
Which brings me to another point, we had to specifically block some tests running in parallel before because of interference with each other and nextest allowed us to do this which might be a thing happening now?

flub:
ahahahaha. lol.  cross does not use nextest :sob:

Arqu:
> ahahahaha. lol. cross does not use nextest 😭

Still a great addition, we've talked about this before. No sane single unit test should run >1 min anyways. In general I'd hope <5 sec.

dignifiedquire:
@flub what's the state here?

flub:
Let's merge it, we want this.  and cross will have to be solved separately

flub:
`iroh::provide::sizes` seems to have timed out a few times...

flub:
seems like this might have been triggered by bad staging relays though, not going to investigate more for now

Issue #2901, opened 2024-11-05
iroh: provide a way to supply custom TLS connection code for relay


link2xt:
In Delta Chat we have custom code to setup TLS connections that supports built-in DNS cache and various types of proxies: SOCKS5, HTTP(S) proxies and Shadowsocks. This code is used for all IMAP, SMTP and HTTPS connections except for the connection to iroh relay. Because of this relay connection is less reliable, e.g. it will not work if DNS is down and it cannot work over Shadowsocks which is not supported by iroh-net.

If's fine for STUN and P2P QUIC connection to be outside of control for iroh-net library user, but for relay we need some way to provide custom code to establish TLS connections that can reuse existing DNS caches, certificate checking code and so on. iroh can then run HTTP stream on top of generic stream.

flub:
This would presumably also need to be enabled for the various HTTPS probes in netcheck?

- captive portal check
- https probe (ping) check

dignifiedquire:
also what about general DNS resolution of the relay urls, and in the discovery service?

matheus23:
> In Delta Chat we have custom code to setup TLS connections that supports built-in DNS cache and various types of proxies: SOCKS5, HTTP(S) proxies and Shadowsocks. This code is used for all IMAP, SMTP and HTTPS connections except for the connection to iroh relay.

I assume the reason you went for custom TLS setup code in favor of, say, using reqwest (given it supports SOCKS5, HTTP(s) proxies and DNS resolution customization), is that you want to also use this for IMAP and SMPT, right?

I'm thinking through what it'd mean if we double-down on reqwest in iroh-net, so we get SOCKS5, HTTP(s) proxies, etc. all out of the box for *all* requests we make in iroh-net, and have them be configurable in *one* way.

matheus23:
> Does iroh resolve relay URLs without the intention to establish a connection to the relay?

Yes. At the time that dig wrote this, we used to probe the relay IPs using ICMP and STUN. We don't do that now anymore, but instead we QAD-probe them ("QUIC address discovery" - i.e. uses UDP). This gives us the closest relay plus our public IP addresses that the relays see. (If we can't QAD, we use normal HTTPS to probe the relays and find the closest one.)

> I don't know what captive portal check is for. If it is about public Wi-Fi captive portal, then why does iroh need to know about it? User has to open it in a browser anyway.

It informs iroh about whether connectivity to relays should be expected or not (I don't think it's terribly important at the moment).

---

[...]

PR #2903, opened 2024-11-06
feat(iroh-net): Implement the https probe

flub:
## Description

This implements the https probe.

There does not need to be separate IPv4 or IPv6 HTTPS probes, because we just go with whatever the HTTPS connection works over.  This is what we need to establish a relay connection after all, this also does not care.  The only reason for this is to get some kind of latency to be able to chose a relay server.

Luckily the https probes have been integrated in the probe plan a long time so nothing to do there.

## Breaking Changes

The URLs served by the relay changed:
- `/relay/probe` has moved to `/ping`
- `/derp/probe` has been removed.

Unless you were manually using those URLs you will not notice these changes, nothing in the iroh codebase ever used the changed URLs.

## Notes & open questions

Currently this uses reqwest with it's default DNS resolution.  This is not great as it is not the same DNS resolution used everywhere else. It is however the same as done in the captive portal check.  This is worth at least looking at before merging this.

## Change checklist

- [x] Use common DNS resolver.
- [x] Update URL
- [x] Write docs for URLs of relay server.
- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/src/netcheck/reportgen.rs
+++ b/iroh-net/src/netcheck/reportgen.rs
@@ -467,6 +467,7 @@ impl Actor {
                 .as_ref()
                 .and_then(|l| l.preferred_relay.clone());
 
+            let dns_resolver = self.dns_resolver.clone();
             let dm = self.relay_map.clone();
             self.outstanding_tasks.captive_task = true;
             MaybeFuture {
@@ -475,7 +476,7 @@ impl Actor {
                     debug!("Captive portal check started after {CAPTIVE_PORTAL_DELAY:?}");
                     let captive_portal_check = tokio::time::timeout(
                         CAPTIVE_PORTAL_TIMEOUT,
-                        check_captive_portal(&dm, preferred_relay)
+                        check_captive_portal(&dns_resolver, &dm, preferred_relay)
                             .instrument(debug_span!("captive-portal")),
                     );
                     match captive_portal_check.await {
@@ -854,7 +855,11 @@ async fn run_stun_probe(
 /// return a "204 No Content" response and checking if that's what we get.
 ///
 /// The boolean return is whether we think we have a captive portal.
-async fn check_captive_portal(dm: &RelayMap, preferred_relay: Option<RelayUrl>) -> Result<bool> {
+async fn check_captive_portal(
+    dns_resolver: &DnsResolver,
+    dm: &RelayMap,
+    preferred_relay: Option<RelayUrl>,
+) -> Result<bool> {
     // If we have a preferred relay node and we can use it for non-STUN requests, try that;
     // otherwise, pick a random one suitable for non-STUN requests.
     let preferred_relay = preferred_relay.and_then(|url| match dm.get_node(&url) {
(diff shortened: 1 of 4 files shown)


iroh-net/src/test_utils.rs
```diff
@@ -67,7 +69,7 @@ pub async fn run_relay_server_with(
         metrics_addr: None,
     };
     let server = Server::spawn(config).await.unwrap();
-    let url: RelayUrl = format!("https://localhost:{}", server.https_addr().unwrap().port())
+    let url: RelayUrl = format!("https://{}", server.https_addr().expect("configured"))
```
flub, line 72: Notice that this changes the relay URL for tests to be the IP address.  This is needed to avoid the DNS lookups for `localhost`: depending on the host configuration this can result in IPv4 or IPv6 addresses but we want to have just IPv4 because that's what the server listens on.

iroh-net/src/netcheck/reportgen.rs
```diff
@@ -881,9 +887,23 @@ async fn check_captive_portal(dm: &RelayMap, preferred_relay: Option<RelayUrl>)
...
+        // ignore the port, extracting it from the URL instead.  We supply a dummy port.
+        //
+        // Ideally we would try to resolve **both** IPv4 and IPv6 rather than purely race
+        // them.  But our resolver doesn't support that yet.
+        let addrs: Vec<_> = dns_resolver
+            .lookup_ipv4_ipv6_staggered(domain, DNS_TIMEOUT, DNS_STAGGERING_MS)
+            .await?
+            .map(|ipaddr| SocketAddr::new(ipaddr, 80))
+            .collect();
+        builder = builder.resolve_to_addrs(domain, &addrs);
```
flub, line 904: Another option is to implement reqwests' `Resolver` trait for our resolver.  But we don't own the type so that's quite a bit of changes.  This keeps our changes targeted to this feature and is not really any worse.

dignifiedquire (approved):
nice and clean

iroh-net/src/netcheck/reportgen.rs
matheus23: I really like the idea of being able to keep using `reqwest` and implementing our custom `Resolver`, perhaps allowing users to provide their own `Resolver` if they want to, perhaps under an unstable feature flag.

Is it feasible to lean in to reqwest even more heavily? It seems like [it supports HTTP upgrades](https://docs.rs/reqwest/latest/reqwest/struct.Response.html#method.upgrade) and thus we could use it in even more places.

If we manage to replace all usage of direct hyper with reqwest, we might be able to have one universal way of configuring custom DNS resolvers or HTTPS proxying.

iroh-net/src/netcheck/reportgen.rs
```diff
@@ -881,9 +887,23 @@ async fn check_captive_portal(dm: &RelayMap, preferred_relay: Option<RelayUrl>)
...
+        //
+        // For some reason reqwest wants SocketAddr rather than IpAddr but then proceeds to
+        // ignore the port, extracting it from the URL instead.  We supply a dummy port.
+        //
+        // Ideally we would try to resolve **both** IPv4 and IPv6 rather than purely race
+        // them.  But our resolver doesn't support that yet.
+        let addrs: Vec<_> = dns_resolver
+            .lookup_ipv4_ipv6_staggered(domain, DNS_TIMEOUT, DNS_STAGGERING_MS)
+            .await?
+            .map(|ipaddr| SocketAddr::new(ipaddr, 80))
```
matheus23, line 902: @flub I think you should probably set these ports to 0, if I'm understanding [this issue](https://github.com/seanmonstar/reqwest/issues/2413#issuecomment-2405545067) correctly.

matheus23: The docs have seemingly not been updated yet. I'll look into the source if they're actually unused or not now.

flub: Good catch!  Also, bad reqwest!  https://github.com/n0-computer/iroh/pull/2906

Nuhvi: Sorry... the change was in `hyper_util` not in reqwest so that is why the documentation stayed outdated. https://github.com/seanmonstar/reqwest/pull/2473

Nuhvi: In my defense, this change in `hyper_util` is really useful to leverage HTTPS records in custom resolvers. Even more work needed on Reqwest side to decide when to switch to HTTP2/3 according to the ALPN property in HTTPS records.

iroh-net/src/netcheck/reportgen.rs
flub: https://docs.rs/iroh-net/latest/iroh_net/endpoint/struct.Builder.html#method.dns_resolver is the current way to configure the DNS resolver for users.  That's what this is picking up.

I don't have a lot of opinion on reqwest use, I've seen people remove it in favour of more low-level control and to reduce the dependency count.  We also need to take into account https://github.com/n0-computer/iroh/issues/2901.

But if we can do everything with reqwest and there's benefits to it then I don't have anything against it.

iroh-net/src/netcheck/reportgen.rs
matheus23: All good. IMO it's more weird to pass in bogus info that's unused. This is what made me skeptical in the first place.

matheus23: Also ugh I didn't quite follow the data-flow in reqwest for how these domain overrides are used. It goes through some `Resolver` and `tower` traits and just gets complicated quickly. So I'm glad to see someone who knows just update the docs.

PR #2913, opened 2024-11-08
fix(netwatch): BSD rebind socket on errors

### LINKED ISSUE #2909: BSD RouteMonitor does not handle AF_ROUTE socket disconnection well
On a few different iPhones, the AF_ROUTE socket in the `RouteMonitor` task has disconnected which resulted in [this log](https://github.com/n0-computer/iroh/blob/23b874c0d52dab340ce7fe03263d3e26e605e772/net-tools/netwatch/src/netmon/bsd.rs#L54) `AF_ROUTE: error reading: Os { code: 57, kind: NotConnected, message: \"Socket is not connected\" }` displaying every millisecond until the app was killed by the os.

I believe this always happened while the app was in the background.

I'm not sure what the significance of that socket disconnecting is, so I'm unsure whether recreating it would work or if the task should just exit at that point

dignifiedquire:
also uses `AbortOnDropHandle` to better cleanup tasks

Closes #2909

### DIFF
--- a/net-tools/netwatch/src/netmon.rs
+++ b/net-tools/netwatch/src/netmon.rs
@@ -4,7 +4,6 @@ use anyhow::Result;
 use futures_lite::future::Boxed as BoxFuture;
 use tokio::sync::{mpsc, oneshot};
 use tokio_util::task::AbortOnDropHandle;
-use tracing::warn;
 
 mod actor;
 #[cfg(target_os = "android")]
@@ -29,8 +28,7 @@ use self::actor::{Actor, ActorMessage};
 #[derive(Debug)]
 pub struct Monitor {
     /// Task handle for the monitor task.
-    #[allow(dead_code)]
-    handle: AbortOnDropHandle<()>,
+    _handle: AbortOnDropHandle<()>,
     actor_tx: mpsc::Sender<ActorMessage>,
 }
 
@@ -41,13 +39,11 @@ impl Monitor {
         let actor_tx = actor.subscribe();
 
         let handle = tokio::task::spawn(async move {
-            if let Err(err) = actor.run().await {
-                warn!("netmon actor died: {:?}", err);
-            }
+            actor.run().await;
         });
 
         Ok(Monitor {
-            handle: AbortOnDropHandle::new(handle),
+            _handle: AbortOnDropHandle::new(handle),
             actor_tx,
         })
     }
--- a/net-tools/netwatch/src/netmon.rs
+++ b/net-tools/netwatch/src/netmon.rs
@@ -2,10 +2,9 @@
 
 use anyhow::Result;
 use futures_lite::future::Boxed as BoxFuture;
-use tokio::{
-    sync::{mpsc, oneshot},
-    task::JoinHandle,
-};
+use tokio::sync::{mpsc, oneshot};
+use tokio_util::task::AbortOnDropHandle;
+use tracing::warn;
 
 mod actor;
 #[cfg(target_os = "android")]
@@ -30,14 +29,9 @@ use self::actor::{Actor, ActorMessage};
 #[derive(Debug)]
 pub struct Monitor {
     /// Task handle for the monitor task.
-    handle: JoinHandle<()>,
+    #[allow(dead_code)]
+    handle: AbortOnDropHandle<()>,
     actor_tx: mpsc::Sender<ActorMessage>,
-}
-
-impl Drop for Monitor {
-    fn drop(&mut self) {
-        self.handle.abort();
-    }
 }
 
 impl Monitor {
@@ -47,10 +41,15 @@ impl Monitor {
         let actor_tx = actor.subscribe();
 
         let handle = tokio::task::spawn(async move {
-            actor.run().await;
+            if let Err(err) = actor.run().await {
+                warn!("netmon actor died: {:?}", err);
+            }
         });
 
-        Ok(Monitor { handle, actor_tx })
+        Ok(Monitor {
+            handle: AbortOnDropHandle::new(handle),
+            actor_tx,
+        })
     }
 
     /// Subscribe to network changes.
--- a/net-tools/netwatch/src/netmon/actor.rs
+++ b/net-tools/netwatch/src/netmon/actor.rs
@@ -33,7 +33,6 @@ use crate::{
 #[derive(Debug, Copy, Clone)]
 pub(super) enum NetworkMessage {
     /// A change was detected.
-    #[allow(dead_code)]
     Change,
 }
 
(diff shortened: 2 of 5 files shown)


net-tools/netwatch/src/netmon.rs
```diff
@@ -30,27 +29,27 @@ use self::actor::{Actor, ActorMessage};
 #[derive(Debug)]
 pub struct Monitor {
     /// Task handle for the monitor task.
-    handle: JoinHandle<()>,
+    #[allow(dead_code)]
+    handle: AbortOnDropHandle<()>,
```
flub, line 33:
```suggestion
    _handle: AbortOnDropHandle<()>,
```
?

net-tools/netwatch/src/netmon.rs
```diff
@@ -30,27 +29,27 @@ use self::actor::{Actor, ActorMessage};
...
     /// Create a new monitor.
     pub async fn new() -> Result<Self> {
         let actor = Actor::new().await?;
         let actor_tx = actor.subscribe();
 
         let handle = tokio::task::spawn(async move {
-            actor.run().await;
+            if let Err(err) = actor.run().await {
+                warn!("netmon actor died: {:?}", err);
+            }
```
flub, line 46: This is essentially a supervisor.  The problem with this version is that you get a single WARN in the logfile and things just carry on.  I wonder if maybe it might as well restart the actor - maybe with a small delay.  That is still not proper supervision - if it keeps dying we should bring down the entire Endpoint - but it's a start and we'll flood the logs with WARN messages which could be a bit more obvious.

net-tools/netwatch/src/netmon/actor.rs
```diff
@@ -127,10 +126,14 @@ impl Actor {
                         debounce_interval.reset_immediately();
                     }
                 }
-                Some(_event) = self.mon_receiver.recv() => {
-                    trace!("network activity detected");
-                    last_event.replace(false);
-                    debounce_interval.reset_immediately();
+                Some(event) = self.mon_receiver.recv() => {
+                    match event {
+                        NetworkMessage::Change => {
```
flub, line 131: `Some(NetworkMessage::Change) = self.mon_receiver.recv()` might be a nicer way to write this?

I also noticed that `NetworkMessage::Change` is marked with `#[allow(dead_code)]`.  This seems suspicious, can we remove that marker as well?

net-tools/netwatch/src/netmon/actor.rs
```diff
@@ -102,7 +101,7 @@ impl Actor {
         self.actor_sender.clone()
     }
 
-    pub(super) async fn run(mut self) {
+    pub(super) async fn run(mut self) -> Result<()> {
```
flub, line 104: I'm confused, the signature is changed but there's not a single error return?

net-tools/netwatch/src/netmon/bsd.rs
```diff
@@ -11,22 +12,24 @@ use crate::{interfaces::bsd::WireMessage, ip::is_link_local};
 
 #[derive(Debug)]
 pub(super) struct RouteMonitor {
-    handle: JoinHandle<()>,
+    #[allow(dead_code)]
+    handle: AbortOnDropHandle<()>,
```
flub, line 16:
```suggestion
    _handle: AbortOnDropHandle<()>,
```
?

I'm not a fan of `#[allow(dead_code)]` as it easily stays around too long.  The `_` is much easier to keep track of as the compiler will tell you.

net-tools/netwatch/src/netmon/bsd.rs
```diff
@@ -52,12 +55,25 @@ impl RouteMonitor {
                     }
                     Err(err) => {
                         warn!("AF_ROUTE: error reading: {:?}", err);
+                        // recreate socket, as it is likely in an invalid state
```
flub, line 58: Does this need a small delay in case this is hot-looping?

net-tools/netwatch/src/netmon.rs
dignifiedquire: it's actually used, which is why I don't want to use `_handle`

net-tools/netwatch/src/netmon.rs
dignifiedquire: generally yes, but currently there is no error case where this happens (and it is not super easy to restart the actor currently)

net-tools/netwatch/src/netmon/actor.rs
dignifiedquire: > This seems suspicious, can we remove that marker as well?

no because this message is not used on android, and I didn't want to start adding cfgs to lint hints

dignifiedquire: > `Some(NetworkMessage::Change) = self.mon_receiver.recv()` might be a nicer way to write this?

maybe, we usually always put the match inside, so I wrote it like that for consistency

net-tools/netwatch/src/netmon.rs
flub: then it doesn't need `#[allow(dead_code)]`?

dignifiedquire: but rust is too stupid..it gets constructed and it has drop semantics.. but the compiler doesn't understand that as usage

flub: Or are we having a philosophical disagreement on what unused means?

My interpretation is the opposite in that case, to me it is definitely not dead code.  `_thing` does not mean something is unused, `let _guard = thing();` is an entirely common pattern, this is just the same but in a struct.

net-tools/netwatch/src/netmon/actor.rs
flub: yeah, let's revert as then the supervisor discussion goes away as well :)

flub (approved):

net-tools/netwatch/src/netmon.rs
dignifiedquire: you are right, will change

net-tools/netwatch/src/netmon/bsd.rs
dignifiedquire: fixed

net-tools/netwatch/src/netmon/actor.rs
dignifiedquire: see issues with tokio::select, I cleaned this up to not use an `else`branch

net-tools/netwatch/src/netmon/actor.rs
dignifiedquire: done

net-tools/netwatch/src/netmon.rs
dignifiedquire: removed

net-tools/netwatch/src/netmon.rs
dignifiedquire: fixed

PR #2914, opened 2024-11-08
fix(iroh-net): Make sure the rtt-actor is shutdown correctly

### LINKED ISSUE #2911: `RttActor` isn't cleaned up on `Endpoint` drop
I have a pretty simple repro [here](https://github.com/PaulOlteanu/iroh/blob/987df2175e18d2efd92ec446fbae7fe5c9a8659b/iroh-net/examples/listen.rs) based on the listen example (and I guess the [connect example](https://github.com/PaulOlteanu/iroh/blob/987df2175e18d2efd92ec446fbae7fe5c9a8659b/iroh-net/examples/connect.rs) for the other end).

A quick fix would be to just [abort the tasks in the Drop impl](https://github.com/PaulOlteanu/iroh/commit/0d9a7632029524acb31992e7f1963c566f8819a6), but probably the better way would be with `JoinSet`s

flub:
## Description

When the endpoint is dropped the rtt-actor should be shut down.  This
makes sure this actually happens.

Fixes #2911

## Notes & open questions

Also make sure the task is aborted on drop.  We should always do this anyway.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/src/endpoint/rtt_actor.rs
+++ b/iroh-net/src/endpoint/rtt_actor.rs
@@ -8,10 +8,10 @@ use iroh_base::key::NodeId;
 use iroh_metrics::inc;
 use tokio::{
     sync::{mpsc, Notify},
-    task::JoinHandle,
     time::Duration,
 };
-use tracing::{debug, error, info_span, trace, warn, Instrument};
+use tokio_util::task::AbortOnDropHandle;
+use tracing::{debug, error, info_span, trace, Instrument};
 
 use crate::{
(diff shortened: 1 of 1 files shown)


PaulOlteanu:
I believe the `else` branch will only execute if none of the other branch's patterns match. Since the `_ = cleanup_interval.tick()` pattern can never not match, the else branch won't ever execute.

dignifiedquire (approved):

flub:
github actions network issues it seems like?  try again

PR #2915, opened 2024-11-08
fix(iroh-net): fix memory leaks in the iroh-relay server

dignifiedquire:
We observed continous memory increase on the relays, this fixes the main offenders we have found. Mainly the issue was not cleaning up tasks that were spawned into `JoinSet`s.

### DIFF
--- a/iroh-relay/src/server.rs
+++ b/iroh-relay/src/server.rs
@@ -456,7 +456,14 @@ async fn server_stun_listener(sock: UdpSocket) -> Result<()> {
     loop {
         tokio::select! {
             biased;
-            _ = tasks.join_next(), if !tasks.is_empty() => (),
+
+            Some(res) = tasks.join_next(), if !tasks.is_empty() => {
+                if let Err(err) = res {
+                    if err.is_panic() {
+                        panic!("task panicked: {:#?}", err);
+                    }
+                }
+            }
             res = sock.recv_from(&mut buffer) => {
                 match res {
                     Ok((n, src_addr)) => {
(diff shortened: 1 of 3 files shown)


Arqu (approved):
A bunch of testing and more profiling later, can confirm this seems to fix the issue. Prod is the real test :P

iroh-net/src/relay/server/http_server.rs
```diff
@@ -542,10 +550,12 @@ impl RelayService {
...
     {
-        hyper::server::conn::http1::Builder::new()
+        let conn = hyper::server::conn::http1::Builder::new()
             .serve_connection(hyper_util::rt::TokioIo::new(io), self)
-            .with_upgrades()
-            .await?;
+            .with_upgrades();
+
+        conn.await?;
+
```
matheus23, line 558: A little bit of diff noise

iroh-net/src/relay/server/http_server.rs
```diff
@@ -253,11 +253,19 @@ impl ServerState {
...
+                                panic!("task panicked: {:#?}", err);
+                            }
+                        }
+                    }
                     res = listener.accept() => match res {
                         Ok((stream, peer_addr)) => {
                             debug!("[{http_str}] relay: Connection opened from {peer_addr}");
                             let tls_config = tls_config.clone();
                             let service = service.clone();
+
```
matheus23, line 268: Diff noise

matheus23 (approved):
I'm excited by this PR (& the idea of a 8.5MB memory-at-idle relay).

Usually, I wouldn't care much about diff noise, but given there will likely be conflicts with #2873, we should reduce diff noise as much as possible.

rklaehn (approved):

flub (approved):

iroh-net/src/relay/server/http_server.rs
dignifiedquire: fixed

iroh-net/src/relay/server/http_server.rs
dignifiedquire: fixed

PR #2916, opened 2024-11-11
fix: rework client counters on the relay

Arqu:
## Description

This substitutes the current "basic" logic in the relay unique node counters with a more durable (redis) store and provides a globally unique setup. Otherwise it should have all the same characteristics as before.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/src/relay/server/metrics.rs
+++ b/iroh-net/src/relay/server/metrics.rs
@@ -1,7 +1,12 @@
+use std::{collections::HashSet, sync::Arc, time::Duration};
+
+use iroh_base::key::PublicKey;
 use iroh_metrics::{
     core::{Counter, Metric},
     struct_iterable::Iterable,
 };
+use redis::AsyncCommands;
+use tokio::sync::Mutex;
 
 /// Metrics tracked for the relay server
 #[allow(missing_docs)]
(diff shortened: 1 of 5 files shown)


dignifiedquire:
this should be fully optional, we don’t want to depend on redis by default

iroh-net/src/relay/server/metrics.rs
```diff
@@ -117,3 +133,161 @@ impl Metric for Metrics {
...
+
+            loop {
+                interval.tick().await;
+                let batch = Arc::clone(&batch);
+                let rclient_clone = rclient_clone.clone();
+                let relay_key = relay_key.clone();
+                let global_key = global_key.clone();
+                tokio::time::timeout(Duration::from_secs(60), async move {
+                    {
+                        let mut batch = batch.lock().await;
```
flub, line 243: I'm not sure why you go for this complicated architecture of having a local hashset which you then fully iterate over every time to send to the relay server.  Most of these will be duplicates since iroh-net nodes tend to be connected to a relay server for a long time.  And the way this takes the lock ends up with the messages being buffered in a channel while the lock is held here (which is being held even while a connection to the redis server is being checked out - way to long a critical section).


Anyway, what is the benefit of batching?  I suspect this is counter-benficial.

I'd suggest an architecture where there's an actor that runs for the lifetime of the magicsock.  The actor receives messages with NodeIds.  The actor also maintains a connection to the redis server and each time it receives an update it sends it to the redis server.  It should do this in a way so that the sender of the messages is never blocked.

I think the logic flow will be a lot easier to follow, all  state is on the redis server so less memory needed locally.

flub (commented):
Are we fine with centrally storing all `NodeId`s we ever see?  Could we do something fancy with bloom filters intead?

matheus23:
If I may paraphrase the two problems this PR intends to solve:
False "positives" on unique nodes because of the uniqueness only being calculated 
- per-relay, not in one memory space across all relays
- in-memory, so not persisted between restarts.

Arqu:
During some side channel discussion we decided to shelf this for a later point in time.

PR #2919, opened 2024-11-12
test(iroh-cli): increase wait on windows resumption tests

divagant-martian:
## Description

These tests fail when windows is too slow writing and copying the files needed. To avoid ignoring the tests, this PR increases the time we give it to be ready on windows runs to double the original value.

## Breaking Changes

n/a

## Notes & open questions

It would still be good to try to identify if anything changed in our code to make iroh slower on windows. This is however very low priority so this offers a workaround.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


dignifiedquire:
will this break in combination with https://github.com/n0-computer/iroh/pull/2900 ? cc @flub

dignifiedquire (approved):

PR #2921, opened 2024-11-13
feat(iroh-base, iroh-net-report)!: intro net-report as a crate

divagant-martian:
## Description

Introduces net-report as its own crate. For this, introduces a new feature flag `relay` in `iroh-base` for the relay topology types.

- in `iroh-base`, the `RelayUrl` is moved to its own file. It can still be used as before, this is not a breaking change.
- `RelayMap`, `RelayNode` are moved to `iroh-base` under the `relay` feature flag.
- `RelayMode` is moved to `endpoint`. See the NOTES section about this.
- net-report now has metrics as a feature flag. I'm not sure how is that this was not feature flagged before but still worked.
- ping is moved without changes from `iroh-net` to `iroh-net-report`. This is the only place where it's used so it should have probably been part of `netcheck` as a module from the beginig.
- Adds missing license files

## Breaking Changes

- `iroh-net`'s `NetcheckMetrics` are now called `NetReportMetrics`

## Notes & open questions

- `RelayMode` remains in `iroh-net` instead of moving to `iroh-base` with other relay related because this is coupled with the staging and prod configuration of relay urls. Since this is more configuration than concept I think it's best to keep it in iroh-net.
- The crate is called `iroh-net-report` because -at least currently- it depends on relay concepts. For example, the tests depend heavily on the internal order of the `RelayMap`. Can we change this? yes. Should we change this? we can discuss it. Is this PR the place for any of that? No.
- The api needs some love. Some types that are supposed to be internal are used by dependents (in this case `iroh-net`) It's not terrible but it could be improved.

## Change checklist

- [x] Self-review.
- [ ] ~Tests if relevant.~


iroh-netcheck/Cargo.toml
```diff
@@ -0,0 +1,54 @@
+[package]
+name = "iroh-netcheck"
+version = "0.1.0"
```
dignifiedquire, line 3: lets use the community version number `0.28` for this, given it has the `iroh-` prefix

dignifiedquire:
make sure to also add this to the two ci files that need a list of crates

iroh-net-report/src/reportgen/hairpin.rs
```diff
@@ -163,7 +163,7 @@ impl Actor {
 
         socket
             .send_to(
-                b"tailscale netcheck; see https://github.com/tailscale/tailscale/issues/188",
+                b"tailscale net_report; see https://github.com/tailscale/tailscale/issues/188",
```
dignifiedquire, line 166: I am not sure we can change this, needs checking

dignifiedquire: sorry 😅 I was too eager

divagant-martian: we can change this and allows us to make this more generic

iroh-net-report/src/lib.rs
```diff
@@ -3,7 +3,7 @@
 //! Netcheck is responsible for finding out the network conditions of the current host, like
 //! whether it is connected to the internet via IPv4 and/or IPv6, what the NAT situation is
 //! etc and reachability to the configured relays.
-// Based on <https://github.com/tailscale/tailscale/blob/main/net/netcheck/netcheck.go>
+// Based on <https://github.com/tailscale/tailscale/blob/main/net/net_report/net_report.go>
```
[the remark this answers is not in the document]
divagant-martian: reverted

iroh-net/src/metrics.rs
```diff
@@ -2,6 +2,7 @@
 #[cfg(feature = "test-utils")]
 #[cfg_attr(iroh_docsrs, doc(cfg(feature = "test-utils")))]
 pub use iroh_relay::server::Metrics as RelayMetrics;
+pub use net_report::Metrics as NetcheckMetrics;
```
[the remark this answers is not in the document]
divagant-martian: yep you are right I missed all these!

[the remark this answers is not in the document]
divagant-martian: ok let's try one more time 6f24fa7d8

iroh-net/src/defaults.rs
```diff
@@ -130,38 +130,8 @@ pub mod staging {
 pub(crate) mod timeouts {
     use std::time::Duration;
 
-    // Timeouts for netcheck
+    // Timeouts for net_report
 
-    /// Maximum duration to wait for a netcheck report.
+    /// Maximum duration to wait for a net_report report.
     pub(crate) const NETCHECK_REPORT_TIMEOUT: Duration = Duration::from_secs(10);
```
dignifiedquire, line 136: the constant should be renamed as well?

divagant-martian:
@dignifiedquire fixed. There's a bunch of `net_report report` that I don't love but I.. think it will have to do for now? or maybe we can rename `Report` to `Check` or something? not looking forwards to more renaming but it's something I noticed. Thoughts? These are all in strings (docs, comments, etc). Code that would have been a `NetReportReport(Report)` looks like `NetReport(Report)` at least

dignifiedquire (approved):

dignifiedquire:
> There's a bunch of `net_report report` that I don't love but I..

I think we can clean this up in follow ups :) lets get the structure in

PR #2923, opened 2024-11-13
fix(netwatch): hold on to netmon sender reference in android

PaulOlteanu:
## Description

With #2913, the netmon actor properly exits when the channel sender is dropped, but the android implementation never kept the sender around. The magicsock actor would then error when trying to subscribe to the network monitor, and the `Endpoint` would be unusable.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.


dignifiedquire (approved):
thank you for catching this!

PR #2928, opened 2024-11-14
refactor!: remove default protocols and iroh-cli

dignifiedquire:
## Description

As previously announced and planned, this removes all protocol implementations from the `iroh` crate, as well as removes the `iroh-cli` crate.

The core functionality of those protocols, including RPC & CLI now is in the individual crates.

## Remove from `iroh`

- [x] `iroh-docs`
- [x] `iroh-blobs`
- [x] `iroh-gossip`
 
## Examples

- [x] Move examples to the appropriate repos: https://github.com/n0-computer/iroh-blobs/pull/15

## Tests

- [x] Move the remaining tests to the appropriate repos

CLI
- [x] Remove 

## Breaking Changes

- `iroh-cli` removed
  - the crate itself
  - the ability to run `iroh` itself in a docker container, as there is no binary anymore
- removed 
  - `iroh::blobs` use `iroh_blobs`
  - `iroh::docs` use `iroh_docs`
  - `iroh::gossip` use `iroh_gossip`
  - `iroh::client::blobs` use `iroh_blobs::rpc::client`. a memory client is available on Blobs
  - `iroh::client::tags` use `iroh_blobs::rpc::client`. a memory client is available on Blobs
  - `iroh::client::gossip`  use `iroh_gossip::rpc::client`. a memory client is available on Gossip
  - `iroh::client::docs`  use `iroh_docs::rpc::client`. a memory client is available on Docs
  - `iroh::client::authors`  use `iroh_docs::rpc::client`. a memory client is available on Docs
  - `iroh::node::MemNode`, use `Node` directly
  - `iroh::node::FsNode`, use `Node` directly
  - `iroh::node::Node::local_pool_handle`
  - `iroh::node::builder::DocsStorage`
  - `iroh::node::builder::Builder::enable_gc_policy`
  - `iroh::node::builder::Builder::enable_docs`
  - `iroh::node::builder::Builder::register_cb_done`
  - `iroh::node::builder::ProtocolBuilder::local_pool_handle`
  - `iroh::node::builder::GcPolicy`
  - `iroh::util::progress`
  - `iroh::util::path::IrohPaths::BaoStoreDir`
  - `iroh::util::path::IrohPaths::DocsDatabase `
  - `iroh::util::path::IrohPaths::Console`
  - `iroh::util::path::IrohPaths::DefaultAuthor`
- `iroh::node::Builder` has no more generic parameters anymore
- `iroh::node::Node` has no more generic parameters anymore

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


Arqu:
So what's the plan now. Ideally we will set up an `iroh-net` based example which just tests the connectivity & pipes and we plug that back into netsim and sort out the blobs story on its own.

Also the docker files should not be removed as we still build images for `iroh-relay` and `iroh-dns-server` which are arguably much more important anyways. I just need to adapt the images & CI.

iroh/examples/rpc.rs
```diff
@@ -21,10 +20,7 @@ pub fn setup_logging() {
         .ok();
 }
 
-async fn run<S>(builder: iroh::node::Builder<S>) -> anyhow::Result<()>
-where
-    S: Store,
-{
+async fn run(builder: iroh::node::Builder) -> anyhow::Result<()> {
     let node = builder
         .enable_rpc()
```
rklaehn, line 25: I was a bit confused at first what node does now. I thought the idea was that iroh is a pure reexport crate and whenever you build a node like thing you DIY it using the router API. E.g. this enable_rpc enables rpc, but only for the node api, which is almost never what people want if they want rpc.

Is the plan to change this at some point?

iroh/examples/transfer.rs
```diff
@@ -0,0 +1,324 @@
+use std::{
```
rklaehn, line 1: This is not really an example but a bench, right?

iroh/src/node/builder.rs
```diff
@@ -125,29 +63,19 @@ async fn spawn_docs<S: iroh_blobs::store::Store>(
...
     rpc_addr: Option<SocketAddr>,
-    blobs_store: D,
     keylog: bool,
     relay_mode: RelayMode,
-    gc_policy: GcPolicy,
     dns_resolver: Option<DnsResolver>,
     node_discovery: DiscoveryConfig,
-    docs_storage: DocsStorage,
     #[cfg(any(test, feature = "test-utils"))]
     insecure_skip_relay_cert_verify: bool,
```
rklaehn, line 78: This mirrors all the options of the endpoint API. I found that somewhat tedious. Have you considered starting this builder with an endpoint builder?

```rust
let ep_builder = Endpoint::builder().insecure_skip_relay_cert_verify(true); // do not not call build yet
Node::builder(ep_builder).node_specific_stuff().build().await?;
```

iroh/examples/rpc.rs
dignifiedquire: yes, that `Node` will go away, as a follow up to this PR

rklaehn (approved):
All my reservations are about code that is going to be removed anyway. Also, there seems to be a lot of reformatting of yaml files.

So let's get this over with...

dignifiedquire:
> Also, there seems to be a lot of reformatting of yaml files.

yeah sorry, fighting my temporary editor..

iroh/examples/transfer.rs
Arqu: The intent is for it to be example that we also happen to use for netsim. It benches nothing on it's own though.

Arqu (approved):

PR #2940, opened 2024-11-16
fix:  remove problematic usage of `else` branches in `tokio::select`s

PaulOlteanu:
## Description

Fixed some more cases of `else` branches in `tokio::select!`s that would never actually run. I noticed because the relay actor would start logging `recv error Receive(shut down)` when I dropped the `Endpoint`.

## Notes & open questions

There's 2 cases where the `else` branch still would never run but I wasn't sure of what the intended behaviour was, so I left a todo comment.

In the `RelayActor` it seems like it would make sense to exit in both cases - the `ping_tasks` `JoinSet` being empty and the `receiver` being closed.

The magicsock actor I'm much less sure of - I suspect the `else` branch could just be removed?

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net/src/magicsock.rs
+++ b/iroh-net/src/magicsock.rs
@@ -1821,6 +1821,7 @@ impl Actor {
                         warn!(?node_addr, "unable to add discovered node address to the node map: {e:?}");
                     }
                 }
+                // This case will never hit. TODO: Figure out what the intention was here
                 else => {
                     trace!("tick: other");
                     inc!(Metrics, actor_tick_other);
--- a/iroh-net/src/magicsock.rs
+++ b/iroh-net/src/magicsock.rs
@@ -1767,10 +1767,22 @@ impl Actor {
                 discovery_events = events;
             }
         }
+
+        let mut receiver_closed = false;
+        let mut portmap_watcher_closed = false;
+        let mut link_change_closed = false;
         loop {
             inc!(Metrics, actor_tick_main);
             tokio::select! {
-                Some(msg) = self.msg_receiver.recv() => {
+                msg = self.msg_receiver.recv(), if !receiver_closed => {
+                    let Some(msg) = msg else {
+                        trace!("tick: magicsock receiver closed");
+                        inc!(Metrics, actor_tick_other);
+
+                        receiver_closed = true;
+                        continue;
+                    };
+
                     trace!(?msg, "tick: msg");
                     inc!(Metrics, actor_tick_msg);
                     if self.handle_actor_message(msg).await {
@@ -1782,7 +1794,15 @@ impl Actor {
                     inc!(Metrics, actor_tick_re_stun);
                     self.msock.re_stun("periodic");
                 }
-                Ok(()) = portmap_watcher.changed() => {
+                change = portmap_watcher.changed(), if !portmap_watcher_closed => {
+                    if change.is_err() {
+                        trace!("tick: portmap watcher closed");
+                        inc!(Metrics, actor_tick_other);
+
+                        portmap_watcher_closed = true;
+                        continue;
+                    }
+
                     trace!("tick: portmap changed");
                     inc!(Metrics, actor_tick_portmap_changed);
                     let new_external_address = *portmap_watcher.borrow();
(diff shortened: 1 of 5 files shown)


iroh-net/src/magicsock.rs
```diff
@@ -1821,6 +1821,7 @@ impl Actor {
                         warn!(?node_addr, "unable to add discovered node address to the node map: {e:?}");
                     }
                 }
+                // This case will never hit. TODO: Figure out what the intention was here
```
dignifiedquire, line 1824: I think the intention was to capture when the branches above would respond with `None`/`Err`, and at least log it to notice it

iroh-net/src/magicsock/relay_actor.rs
```diff
@@ -316,6 +318,7 @@ impl RelayActor {
                     trace!("tick: cleanup");
                     with_cancel(self.cancel_token.child_token(), self.clean_stale_relay()).await;
                 }
+                // This case will never hit. TODO: Figure out what the intention was here
```
dignifiedquire, line 321: deal with `join_next` returning `None` or `receiver.recv()` returning `None`

PaulOlteanu: I left a comment on the `join_next` case - it returning `None` isn't actually a special case so we don't really need to handle that differently

iroh-net/src/magicsock.rs
PaulOlteanu: Presumably we don't want the actor to exit in those cases so we need to add a bunch of bools to prevent tokio from polling those branches once they return `None`. Otherwise we'll basically end up with a busy wait of getting `None` -> logging and `continue`ing -> repeat.

iroh-net/src/magicsock.rs
```diff
@@ -1767,10 +1767,22 @@ impl Actor {
...
-                Some(msg) = self.msg_receiver.recv() => {
+                msg = self.msg_receiver.recv(), if !receiver_closed => {
+                    let Some(msg) = msg else {
+                        trace!("tick: magicsock receiver closed");
+                        inc!(Metrics, actor_tick_other);
+
+                        receiver_closed = true;
+                        continue;
+                    };
+
```
flub, line 1785: This means there are no senders anymore.  So either this is fatal or new senders can be created I think.  If new senders can be created this needs to be polled again, so the simple `receiver_closed` boolean is not sufficient because it always disables this branch.

iroh-net/src/magicsock.rs
```diff
@@ -1782,7 +1794,15 @@ impl Actor {
                     inc!(Metrics, actor_tick_re_stun);
                     self.msock.re_stun("periodic");
                 }
-                Ok(()) = portmap_watcher.changed() => {
+                change = portmap_watcher.changed(), if !portmap_watcher_closed => {
+                    if change.is_err() {
+                        trace!("tick: portmap watcher closed");
+                        inc!(Metrics, actor_tick_other);
+
+                        portmap_watcher_closed = true;
```
flub, line 1802: Is the portmapper critical?
- yes: then it should probably be restarted and a new watcher installed.
- no: then this should perhaps log on error and carry on like done here.

I'm kind of tempted to go with "no" for now, but "yes" long term.  The magicsock can work without a portmapper, just won't be as good.  But this is almost certainly a bug in our code.

iroh-net/src/magicsock.rs
```diff
@@ -1809,22 +1829,29 @@ impl Actor {
...
                     }
                 }
-                Some(is_major) = link_change_r.recv() => {
+                is_major = link_change_r.recv(), if !link_change_closed => {
+                    let Some(is_major) = is_major else {
+                        trace!("tick: link change receiver closed");
+                        inc!(Metrics, actor_tick_other);
+
+                        link_change_closed = true;
+                        continue;
```
flub, line 1838: Here I also think this is a critical error.  Clearly it doesn't happen much, so maybe let's just log this on error level for now and carry on.  Long term the link monitor (or whatever) can make sure it is supervised and doesn't die on us like that.

dignifiedquire (approved):
looks good, thanks a lot

flub (approved):

iroh-net/src/magicsock/relay_actor.rs
```diff
@@ -301,25 +303,37 @@ impl RelayActor {
...
                 }
-                Some(Ok((url, ping_success))) = self.ping_tasks.join_next() => {
-                    if !ping_success {
-                        with_cancel(
-                            self.cancel_token.child_token(),
-                            self.close_or_reconnect_relay(&url, "rebind-ping-fail")
-                        ).await;
+                // `ping_tasks` being empty is a normal situation - in fact it starts empty
+                // until a `MaybeCloseRelaysOnRebind` message is received.
+                Some(task_result) = self.ping_tasks.join_next() => {
```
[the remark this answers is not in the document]
flub: If the JoinSet is empty, isn't `.join_next()` always ready?  I think this needs a `!self.ping_tasks.is_empty()` guard, no?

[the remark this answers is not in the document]
PaulOlteanu: Yeah the guard makes more sense than polling and discarding the result of that poll every loop

iroh-net/src/magicsock.rs
PaulOlteanu: No new senders can be created at that point (it's an mpsc channel so there'd have to be an existing sender to clone) so yeah I guess it'd be fatal

PaulOlteanu:
seems like this still wasn't quite enough to fix the `recv error Receive(shut down)` logs I was seeing - the sender must not be getting dropped. I'll see if I can look into it some more and also add some of the changes suggested after this pr got merged

iroh-net/src/magicsock.rs
PaulOlteanu: I think technically things can continue fine if the network monitor were to close as long as the network doesn't change, or even if it does as long as  `Endpoint::network_change` is used.

But I do agree that it should probably be logged at a higher level - I'll change it to `warn`

PaulOlteanu:
Ah, the `relay_actor_sender` gets put into an `Arc<MagicSock>`, which is then cloned into the relay actor itself. So the `RelayActor` keeps itself alive forever unless cancelled with its `cancel_token`.

iroh-net/src/magicsock.rs
flub: Thanks for checking!

PR #2946, opened 2024-11-18
[WIP] feat(iroh-net): allow the underlying UdpSockets to be rebound

### LINKED ISSUE #2939: Handle socket failures
related: https://github.com/quinn-rs/quinn/issues/2011

If the underlying socket for an `Endpoint` fails, the `Endpoint` doesn't recover, with errors including:  
`probe failed: Failed to send STUN request: StunIpv4: Broken pipe (os error 32)`

`tick: probes timed out`

`no successful probes in ProbeSet`

`sendmsg error: Os { code: 32, kind: BrokenPipe, message: \"Broken pipe\" }, Transmit: { destination: 155.47.212.61:52827, src_ip: None, ecn: None, len: 124, segment_size: None }`

This was reproducible on iOS by suspending and resuming the app after creating an `Endpoint`

dignifiedquire:
## Description

In order to fix #2939 we need to rebind our UDP sockets when they break. This is the beginning of allowing this to happen.

## Breaking Changes

There will be some in the `netmon::UdpSocket` API

## Notes & open questions

I have tried putting this logic higher in the stack, but unfortunately that did not work out.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


net-tools/netwatch/src/udp.rs
```diff
@@ -51,97 +65,432 @@ impl UdpSocket {
...
-        .context("socket create")?;
-
-        if let Err(err) = socket.set_recv_buffer_size(SOCKET_BUFFER_SIZE) {
-            warn!(
-                "failed to set recv_buffer_size to {}: {:?}",
-                SOCKET_BUFFER_SIZE, err
-            );
+        let mut addr = addr.into();
+        let socket = inner_bind(addr)?;
+        if addr.port() == 0 {
```
flub, line 105: why even make this conditional?

net-tools/netwatch/src/udp.rs
```diff
@@ -51,97 +65,429 @@ impl UdpSocket {
         Self::bind_raw(addr)
     }
 
+    /// Is the socket broken and needs a rebind?
+    pub fn is_broken(&self) -> bool {
+        self.is_broken.load(std::sync::atomic::Ordering::SeqCst)
+    }
+
+    /// Marks this socket as needing a rebind
+    pub fn mark_broken(&self) {
```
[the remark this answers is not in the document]
dignifiedquire: this is all in progress, don't worry about the api too much for now

Arqu:
| iroh | 1_to_1 | 0.65 | 1.05 |
| iroh | 1_to_3 | 1.46 | 2.14 |
| iroh | 1_to_5 | 2.88 | 4.96 |
| iroh | 1_to_10 | 4.31 | 5.62 |
| iroh | 2_to_2 | 1.05 | 1.61 |
| iroh | 2_to_4 | 2.03 | 3.16 |
| iroh | 2_to_6 | 3.19 | 4.86 |
| iroh | 2_to_10 | 4.58 | 6.48 |

net-tools/netwatch/src/udp.rs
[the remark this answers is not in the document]
divagant-martian: I would also prefer to go with setting the `Option` to `None` since it makes this less vulnerable to toctou races. 

Maybe there's something that prevents setting `is_broken` in a non synchronized way with `rebind` (which is called after `is_broken()`) in the higher level calls, but since this is exposed as part of the public i think it would be better to have rebinding (the use part) and is_broken (the check part) as a single locked entity.

We currently use the option only for drop but we could simply do the fancy drop routine when there's actually a socket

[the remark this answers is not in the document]
divagant-martian: Going a bit further we could have a enum like
```rust
enum InnerSock {
	Active(tokio::net::UddSocket),
	Rebind(std::net::SocketAddr),
}
```

should use a teeny tiny less memory, does not require storing the address when not needed, no need to think about races about whether being broken and actually rebinding are being updated in some incorrect way and uses a single lock

net-tools/netwatch/src/udp.rs
```diff
@@ -51,97 +65,617 @@ impl UdpSocket {
...
+        debug!("rebinding {}", self.addr);
+        // Remove old socket
+        let mut guard = self.socket.write().unwrap();
+        {
+            let socket = guard.take().expect("not yet dropped");
+            drop(socket);
+        }
+
+        // Prepare new socket
+        let new_socket = inner_bind(self.addr)?;
```
divagant-martian, line 90: note that if rebind fails, the inner socket will be `None` when not dropped and a bunch of `expect`s will scream

net-tools/netwatch/src/udp.rs
```diff
@@ -51,97 +65,617 @@ impl UdpSocket {
...
+    }
+
+    /// Rebind the underlying socket.
+    pub fn rebind(&self) -> Result<()> {
+        debug!("rebinding {}", self.addr);
+        // Remove old socket
+        let mut guard = self.socket.write().unwrap();
+        {
+            let socket = guard.take().expect("not yet dropped");
+            drop(socket);
```
divagant-martian, line 86: I think we pay here the penalty of the slow drop of the previous socket?

divagant-martian (commented):
reviewed a small piece (what seemed more critical I think?) in any case I think this is the right approach. What we had before "replacing" some `Arc` was totally wrong

net-tools/netwatch/src/udp.rs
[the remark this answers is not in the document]
dignifiedquire: > Maybe there's something that prevents setting `is_broken` in a non synchronized way with `rebind` (which is called after `is_broken()`) in the higher level calls, but since this is exposed as part of the public i think it would be better to have rebinding (the use part) and is_broken (the check part) as a single locked entity.

I understand the desire, but `mark_broken` needs to work without taking a write lock, so it needs to be seperated into an atomic, or its own lock

net-tools/netwatch/src/udp.rs
dignifiedquire: fixed

iroh-net/src/magicsock/udp_conn.rs
```diff
@@ -29,41 +28,88 @@ impl UdpConn {
...
     }
 
+    pub(super) fn rebind(&self) -> anyhow::Result<()> {
+        // Rebind underlying socket
+        self.io.rebind()?;
+
+        // update socket state
+        let new_state = self.io.with_socket(|socket| {
+            quinn_udp::UdpSocketState::new(quinn_udp::UdpSockRef::from(socket))
+        })??;
```
flub, line 48: If this fails the UdpConn is left in an unknown and unusable state right?  But this takes `&self` so there is no real indicator that this is broken afterwards.

iroh-net/src/magicsock/udp_conn.rs
```diff
@@ -73,21 +119,63 @@ impl AsyncUdpSocket for UdpConn {
...
+            if self.io.is_broken() {
+                match self.rebind() {
+                    Ok(()) => {
+                        // all good
+                    }
+                    Err(err) => {
+                        warn!("failed to rebind socket: {:?}", err);
+                        // TODO: improve error
+                        let err =
+                            std::io::Error::new(std::io::ErrorKind::NotConnected, err.to_string());
```
flub, line 131: I fear that quinn will swallow these errors and try to carry on.

flub: Perhaps rebind should directly return `std::io::Error`?  This branch returns `NotConnected` while in `try_send` it returns `BrokenPipe`.  Should these not be consistent?

iroh-net/src/magicsock/udp_conn.rs
```diff
@@ -73,21 +119,63 @@ impl AsyncUdpSocket for UdpConn {
...
+            match res {
+                Ok(Ok(count)) => {
+                    for meta in meta.iter().take(count) {
+                        trace!(
+                            src = %meta.addr,
+                            len = meta.len,
+                            count = meta.len / meta.stride,
+                            dst = %meta.dst_ip.map(|x| x.to_string()).unwrap_or_default(),
+                            "UDP recv"
+                        );
```
flub, line 162: top marks for logging call! :)

net-tools/netwatch/src/udp.rs
```diff
@@ -1,17 +1,31 @@
...
+    socket: RwLock<Option<tokio::net::UdpSocket>>,
+    /// The addr we are binding to.
+    addr: SocketAddr,
+    /// Set to true, when an error occurred, that means we need to rebind the socket.
+    is_broken: AtomicBool,
+}
 
 /// UDP socket read/write buffer size (7MB). The value of 7MB is chosen as it
-/// is the max supported by a default configuration of macOS. Some platforms will silently clamp the value.
+/// is the ma supported by a default configuration of macOS. Some platforms will silently clamp the value.
```
flub, line 28:
```suggestion
/// is the max supported by a default configuration of macOS. Some platforms will silently clamp the value.
```

net-tools/netwatch/src/udp.rs
```diff
@@ -51,97 +65,668 @@ impl UdpSocket {
...
+
+    /// Use the socket
+    pub fn with_socket<F, T>(&self, f: F) -> std::io::Result<T>
+    where
+        F: FnOnce(&tokio::net::UdpSocket) -> T,
+    {
+        let guard = self.socket.read().unwrap();
+        let Some(socket) = guard.as_ref() else {
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::BrokenPipe,
```
flub, line 125: how was this error chosen?  i mean, why is this one the right error?  might be worth documenting in the code.

iroh-net/src/magicsock/udp_conn.rs
dignifiedquire: there is? all the methods will return an error if this happens

net-tools/netwatch/src/udp.rs
dignifiedquire: it was the closest to me that indicates that the connection is broken, would love to hear better suggestions.

net-tools/netwatch/src/udp.rs
```diff
@@ -51,97 +68,810 @@ impl UdpSocket {
...
+    }
+
+    /// Rebind the underlying socket.
+    pub fn rebind(&self) -> Result<()> {
+        debug!("rebinding {}", self.addr);
+        // Remove old socket
+        let mut guard = self.socket.write().unwrap();
+        {
+            let Some(socket) = guard.take() else {
+                bail!("cannot rebind closed socket");
```
flub, line 89: should state the reason: socket already closed or no socket to rebind or something like that.

net-tools/netwatch/src/udp.rs
```diff
@@ -51,97 +68,810 @@ impl UdpSocket {
...
+    let network = IpFamily::from(addr.ip());
+    let socket = socket2::Socket::new(
+        network.into(),
+        socket2::Type::DGRAM,
+        Some(socket2::Protocol::UDP),
+    )
+    .context("socket create")?;
+
+    if let Err(err) = socket.set_recv_buffer_size(SOCKET_BUFFER_SIZE) {
+        warn!(
```
flub, line 760: We need to downgrade these to debug.  We've had reports that this fails on iOS.  (Sadly not filed as an issue.)  And these logs don't just appear once but re-appear on every rebind.

net-tools/netwatch/src/udp.rs
```diff
@@ -51,97 +68,810 @@ impl UdpSocket {
...
+        loop {
+            // check if the socket needs a rebind
+            if self.is_broken() {
+                match self.rebind() {
+                    Ok(()) => {
+                        // all good
+                    }
+                    Err(err) => {
+                        warn!("failed to rebind socket: {:?}", err);
+                        // TODO: improve error
```
flub, line 288: What still needs to be added to the error?  I think this should be resolved before merging as otherwise it'll be here forever.

net-tools/netwatch/src/udp.rs
```diff
@@ -51,97 +68,810 @@ impl UdpSocket {
...
+                }
+            }
+            let guard = self.socket.read().unwrap();
+            let Some((socket, state)) = guard.as_ref() else {
+                warn!("socket closed");
+                return Err(std::io::Error::new(
+                    std::io::ErrorKind::BrokenPipe,
+                    "socket closed",
+                ));
+            };
```
flub, line 346: I wonder if you can de-duplicate this block somehow?  Would it be worth it?

iroh-net/src/magicsock/udp_conn.rs
flub, line 1: What value does this still provide now?  Would it make more sense to impl AsyncUdpSocket on our UdpSocket directly now that the UdpSocketState is kept there?  I think this is now just an Arc'd UdpSocket?

net-tools/netwatch/src/udp.rs
```diff
@@ -51,97 +68,810 @@ impl UdpSocket {
...
+        Ok(UdpSocket {
+            socket: RwLock::new(Some(socket)),
+            recv_waker: AtomicWaker::default(),
+            send_waker: AtomicWaker::default(),
+            addr,
+            is_broken: AtomicBool::new(false),
+        })
+    }
+
+    /// TODO
```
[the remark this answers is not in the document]
dignifiedquire: to copy paste docs..

net-tools/netwatch/src/udp.rs
```diff
@@ -51,97 +68,810 @@ impl UdpSocket {
...
+    pub fn send_to<'a, 'b>(&'b self, buffer: &'a [u8], to: SocketAddr) -> SendToFut<'a, 'b> {
+        SendToFut {
+            socket: self,
+            buffer,
+            to,
+        }
+    }
+
+    /// TODO
+    pub fn connect(&self, addr: SocketAddr) -> std::io::Result<()> {
```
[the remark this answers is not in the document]
dignifiedquire: yes, portmapper does

iroh-net/src/magicsock/udp_conn.rs
dignifiedquire: it implements the quinn abstractions which I still wanted to keep out of upd.rs as much as possible

net-tools/netwatch/src/udp.rs
dignifiedquire: done

net-tools/netwatch/src/udp.rs
dignifiedquire: agreed, I am just not sure what the best way here is

net-tools/netwatch/src/udp.rs
dignifiedquire: done

net-tools/netwatch/src/udp.rs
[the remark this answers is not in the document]
dignifiedquire: done

net-tools/netwatch/src/udp.rs
```diff
@@ -51,97 +68,810 @@ impl UdpSocket {
...
+
+        // Prepare new socket
+        let new_socket = inner_bind(self.addr)?;
+
+        // Insert new socket
+        guard.replace(new_socket);
+
+        // Clear errors
+        self.is_broken
+            .store(false, std::sync::atomic::Ordering::SeqCst);
```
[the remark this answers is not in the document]
flub: `SeqCst` will not give you anything wrong, but is overkill.  When working with simple loads and stores on bools the docs on when to use `Release` and `Acquire` are pretty straight forward: https://doc.rust-lang.org/stable/std/sync/atomic/enum.Ordering.html#variant.Release

I can't find the direct quote, but the argument by Mara Bos (who wrote Rust Atomics and Locks) is something like: SeqCst is a code-smell because it means there has been no thought about the concurrency involved.  It suggests the program needs a total order.  Or something like that.

net-tools/netwatch/src/udp.rs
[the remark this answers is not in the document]
flub: TIL

net-tools/netwatch/src/udp.rs
```diff
@@ -51,97 +68,810 @@ impl UdpSocket {
         Self::bind_raw(addr)
     }
 
+    /// Is the socket broken and needs a rebind?
+    pub fn is_broken(&self) -> bool {
+        self.is_broken.load(std::sync::atomic::Ordering::SeqCst)
```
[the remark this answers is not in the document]
matheus23: Yes agreed.

net-tools/netwatch/src/udp.rs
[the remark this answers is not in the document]
matheus23: Agreed.
I love the "happens before relationship" mental model that mara bos explains really well in her [atomics and locks book](https://marabos.nl/atomics/memory-ordering.html).
With that mental model - what you care about is that other threads see the `is_broken` bool flip *after* the atomics in the socket lock flip.
From the perspective of the thread we're looking at, this happens-before relationship is true.
Using `Acquire` and `Release` orderings makes sure this is also the case from the perspective of other threads.

It's not *wrong* to use `SeqCst`, but mara writes something about that in the final words in the chapter:

> Myth: Sequentially consistent memory ordering is a great default and is always correct.
> [...]
> It is advisable to see SeqCst as a warning sign. Seeing it in the wild often means that either something complicated is going on, or simply that the author did not take the time to analyze their memory ordering related assumptions, both of which are reasons for extra scrutiny.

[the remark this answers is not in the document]
dignifiedquire: fair, I use it to make sure I don't introduce a race condition by accident..

net-tools/netwatch/src/udp.rs
```diff
@@ -51,8 +66,616 @@ impl UdpSocket {
...
+                    }
+                }
+            }
+        }
+    }
+
+    /// Whether transmitted datagrams might get fragmented by the IP layer
+    ///
+    /// Returns `false` on targets which employ e.g. the `IPV6_DONTFRAG` socket option.
+    pub fn may_fragment(&self) -> std::io::Result<bool> {
```
flub, line 419: I wonder for these 3 functions whether you could store them at construction and rebind time.  That way they could keep the signature without `Result`, which matches the signature higher up and avoids the odd `.unwrap_or_default()` there.

It is a lot of effort in order to give out real numbers for an effectively closed socket though.  So understand if you don't find it worth it or improves things much.

flub (approved):
There are still unresolved comments, but essentially this looks fine I think.

iroh-net/src/magicsock/udp_conn.rs
```diff
@@ -72,40 +60,23 @@ impl AsyncUdpSocket for UdpConn {
...
 
     fn max_transmit_segments(&self) -> usize {
-        self.inner.max_gso_segments()
+        self.io.max_gso_segments().unwrap_or_default()
     }
 
     fn max_receive_segments(&self) -> usize {
-        self.inner.gro_segments()
+        self.io.gro_segments().unwrap_or_default()
     }
```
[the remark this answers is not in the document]
dignifiedquire: I disagree, there is no reason to panic here, the socket is closed, but we can still respond with values, given that quinn is not able to handle errors here. The closed socket state will be discovered on the next read or write already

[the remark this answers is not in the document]
flub: At this point the socket is essentially closed, so this is a bit an academic complaint.  But I share the complaint :)  (I wrote a comment about making the lower level infallible again as a possible workaround though)

net-tools/netwatch/src/udp.rs
```diff
@@ -89,9 +712,11 @@ impl UdpSocket {
 
         // Convert into tokio UdpSocket
         let socket = tokio::net::UdpSocket::from_std(socket).context("conversion to tokio")?;
+        let socket_ref = quinn_udp::UdpSockRef::from(&socket);
+        let socket_state = quinn_udp::UdpSocketState::new(socket_ref)?;
 
+        let local_addr = socket.local_addr().context("local addr")?;
         if addr.port() != 0 {
-            let local_addr = socket.local_addr().context("local addr")?;
             ensure!(
```
matheus23, line 720: Eh - I think this was me who moved the `local_addr` call accidentally (I used it when tracing stuff, but removed only the tracing call).

I think it can move back into the `if`.

net-tools/netwatch/src/udp.rs
dignifiedquire: it's not a bad idea, let me look at the way they actually get generated

net-tools/netwatch/src/udp.rs
dignifiedquire: no I moved this on purpose, we store the `local_addr` here now

net-tools/netwatch/src/udp.rs
```diff
@@ -51,8 +66,616 @@ impl UdpSocket {
...
+#[derive(Debug)]
+enum SocketState {
+    Connected {
+        socket: tokio::net::UdpSocket,
+        state: quinn_udp::UdpSocketState,
+        /// The addr we are binding to.
+        addr: SocketAddr,
+    },
+    Closed,
+}
```
matheus23, line 656: :white_heart:

net-tools/netwatch/src/udp.rs
matheus23: I swear I can read code. :woozy_face:

net-tools/netwatch/src/udp.rs
```diff
@@ -100,48 +725,152 @@ impl UdpSocket {
...
+                    // this will be empty if `close` was called before
+                    let std_sock = socket.into_std();
+                    handle.spawn_blocking(move || {
+                        // Calls libc::close, which can block
+                        drop(std_sock);
+                    });
+                }
+            }
+            None => {}
+        }
```
matheus23, line 779: I agree with clippy here. This can be an `if-let`:
```suggestion
        if let Some((socket, _)) = self.socket.write().unwrap().close() {
            if let Ok(handle) = tokio::runtime::Handle::try_current() {
                // No wakeup after dropping write lock here, since we're getting dropped.
                // this will be empty if `close` was called before
                let std_sock = socket.into_std();
                handle.spawn_blocking(move || {
                    // Calls libc::close, which can block
                    drop(std_sock);
                });
            }
        }
```

net-tools/netwatch/src/udp.rs
dignifiedquire: done

net-tools/netwatch/src/udp.rs
dignifiedquire: I hate if let.. it always makes me feel like I forgot sth

net-tools/netwatch/src/udp.rs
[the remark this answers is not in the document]
dignifiedquire: fixed

net-tools/netwatch/src/udp.rs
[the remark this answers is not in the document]
dignifiedquire: done

iroh-net/src/magicsock/udp_conn.rs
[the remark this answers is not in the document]
matheus23: I like the idea of remembering the last values.

matheus23 (approved):
Yay this is nice :shipit: 
(Also make the formatter happy though: `cargo make format`)

flub (approved):

Arqu:
Current branch last run:
| iroh | 1_to_1 | 1.37 | 1.37 |
| iroh | 1_to_3 | 4.42 | 4.42 |
| iroh | 1_to_5 | 7.20 | 7.21 |
| iroh | 1_to_10 | 11.86 | 11.88 |
| iroh | 2_to_2 | 3.03 | 3.04 |
| iroh | 2_to_4 | 5.89 | 5.90 |
| iroh | 2_to_6 | 8.38 | 8.39 |
| iroh | 2_to_10 | 13.55 | 13.57 |

Current main:
| iroh | 1_to_1 | 1.36 | 1.36 |
| iroh | 1_to_3 | 4.42 | 4.43 |
| iroh | 1_to_5 | 6.89 | 6.90 |
| iroh | 1_to_10 | 11.33 | 11.35 |
| iroh | 2_to_2 | 2.72 | 2.72 |
| iroh | 2_to_4 | 5.54 | 5.55 |
| iroh | 2_to_6 | 7.87 | 7.88 |
| iroh | 2_to_10 | 13.32 | 13.35 |


Seems like its about the same or slight improvement.

PR #2956, opened 2024-11-21
fix(ci): try to reuse msys2 dep

Arqu:
## Description

According to https://github.com/msys2/setup-msys2?tab=readme-ov-file#release this should try to reuse an existing msys2 installation (though seems to reference a default fallback to the one installed on the GH hosted runners, unsure if it works for self hosted)

Edit: seems to be working

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


divagant-martian (approved):
seems appropriate according to the docs

PR #2959, opened 2024-11-22
fix: make netsim run from examples

Arqu:
## Description

Moves our netsim tests to run from an example and only test the lib part of the code, not the full CLI & blobs.

Things left to do:
- [x] rename the example from `new` to anything that makes more sense `netsim` runs it under `iroh-transfer`
- [x] clean up the example code
- [x] add the option to provide either a relay config or pass in at least the relay url as an argument
- [x] continue the CI adjustment and move all invocations of the netsim runner to run from `iroh_v2` and `integration_v2` sims (we want a less abrupt netsim switchover)
- [x] convert the remaining sims in `chuck/netsim/sims` to the new format
- [ ] after some time flip back the CI invocations to be regular `iroh` and `integration` sims which includes doing the same on `netsim` and removing the old ones

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -295,7 +295,7 @@ jobs:
       branch: ${{ github.ref }}
       max_workers: 4
       netsim_branch: "main"
-      sim_paths: "sims/iroh/iroh_v2.json,sims/integration_v2"
+      sim_paths: "sims/iroh_v2/iroh.json,sims/integration_v2"
       pr_number: ${{ github.event.pull_request.number || '' }}
 
   codespell:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -295,7 +295,7 @@ jobs:
       branch: ${{ github.ref }}
       max_workers: 4
       netsim_branch: "main"
-      sim_paths: "sims/iroh/iroh.json,sims/integration"
+      sim_paths: "sims/iroh/iroh_v2.json,sims/integration_v2"
       pr_number: ${{ github.event.pull_request.number || '' }}
 
   codespell:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -246,13 +246,13 @@ jobs:
       # TODO: We have a bunch of platform-dependent code so should
       #    probably run this job on the full platform matrix
       - name: clippy check (all features)
-        run: cargo clippy --workspace --all-features --all-targets --bins --tests --benches
+        run: cargo clippy --workspace --all-features --all-targets --bins --tests --benches --examples
 
       - name: clippy check (no features)
-        run: cargo clippy --workspace --no-default-features --lib --bins --tests
+        run: cargo clippy --workspace --no-default-features --lib --bins --tests --examples
 
       - name: clippy check (default features)
-        run: cargo clippy --workspace --all-targets
+        run: cargo clippy --workspace --all-targets --examples
 
   msrv:
     if: "github.event_name != 'pull_request' || ! contains(github.event.pull_request.labels.*.name, 'flaky-test')"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -246,13 +246,13 @@ jobs:
       # TODO: We have a bunch of platform-dependent code so should
       #    probably run this job on the full platform matrix
       - name: clippy check (all features)
-        run: cargo clippy --workspace --all-features --all-targets --bins --tests --benches --examples
+        run: cargo clippy --workspace --all-features --all-targets --lib --bins --tests --benches --examples
 
       - name: clippy check (no features)
-        run: cargo clippy --workspace --no-default-features --lib --bins --tests --examples
+        run: cargo clippy --workspace --no-default-features --all-targets --lib --bins --tests --benches --examples
 
       - name: clippy check (default features)
-        run: cargo clippy --workspace --all-targets --examples
+        run: cargo clippy --workspace --all-targets --lib --bins --tests --benches --examples
 
   msrv:
     if: "github.event_name != 'pull_request' || ! contains(github.event.pull_request.labels.*.name, 'flaky-test')"
--- a/iroh/examples/transfer.rs
+++ b/iroh/examples/transfer.rs
@@ -211,7 +211,10 @@ async fn fetch(ticket: &str, relay_url: Option<String>) -> anyhow::Result<()> {
     let duration = start.elapsed();
     println!(
         "Received {} B in {:.4}s with ttfb {}s in {} chunks",
-        HumanBytes(len as u64), duration.as_secs_f64(), ttfb.as_secs_f64(), chnk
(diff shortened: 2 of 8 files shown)


Arqu:
| iroh | 1_to_1 | 1.32 | 1.33 |
| iroh | 1_to_3 | 4.28 | 4.28 |
| iroh | 1_to_5 | 6.89 | 6.90 |
| iroh | 1_to_10 | 11.65 | 11.67 |
| iroh | 2_to_2 | 2.77 | 2.78 |
| iroh | 2_to_4 | 5.42 | 5.43 |
| iroh | 2_to_6 | 8.22 | 8.23 |
| iroh | 2_to_10 | 13.55 | 13.58 |

.github/workflows/ci.yml
```diff
@@ -246,13 +246,13 @@ jobs:
...
-        run: cargo clippy --workspace --all-features --all-targets --bins --tests --benches
+        run: cargo clippy --workspace --all-features --all-targets --bins --tests --benches --examples
 
       - name: clippy check (no features)
-        run: cargo clippy --workspace --no-default-features --lib --bins --tests
+        run: cargo clippy --workspace --no-default-features --lib --bins --tests --examples
 
       - name: clippy check (default features)
-        run: cargo clippy --workspace --all-targets
+        run: cargo clippy --workspace --all-targets --examples
```
dignifiedquire, line 255: shouldn't we also do `lib` `bins` and `tests` for this variant?

dignifiedquire (approved):
looks good overall

.github/workflows/ci.yml
Arqu: Now that you mention it, made all of them uniform

Arqu:
Waiting for CI and will merge.

iroh/examples/transfer.rs
```diff
@@ -0,0 +1,313 @@
...
+
+        // spawn a task to handle reading and writing off of the connection
+        tokio::spawn(async move {
+            // accept a bi-directional QUIC connection
+            // use the `quinn` APIs to send and recv content
+            let (mut send, mut recv) = conn.accept_bi().await?;
+            tracing::debug!("accepted bi stream, waiting for data...");
+            let message = recv.read_to_end(100).await?;
+            let message = String::from_utf8(message)?;
+            println!("received: {message}");
```
flub, line 128: This seems expensive for something that's meant to measure the transfer speed.  Printing to terminals (or whatever output) is notoriously slow.  Checking valid UTF-8 is also slow.

iroh/examples/transfer.rs
```diff
@@ -0,0 +1,313 @@
...
+        .await?;
+    info!("connected");
+
+    // Use the Quinn API to send and recv content.
+    let (mut send, mut recv) = conn.open_bi().await?;
+
+    let message = format!("{me} is saying 'hello!'");
+    send.write_all(message.as_bytes()).await?;
+
+    // Call `finish` to close the send side of the connection gracefully.
```
flub, line 202:
```suggestion
    // Call `finish` to signal no more data will be sent on this stream.
```
This doesn't really have anything to do with gracefully closing anything.  It only signals that no more data will be sent on this stream.  (Which is what `read_to_end` on the other side relies on.)

iroh/examples/transfer.rs
```diff
@@ -0,0 +1,313 @@
...
+    let (len, ttfb, chnk) = drain_stream(&mut recv, false).await?;
+
+    // We received the last message: close all connections and allow for the close
+    // message to be sent.
+    endpoint.close(0u8.into(), b"bye").await?;
+
+    let duration = start.elapsed();
+    println!(
+        "Received {} B in {:.4}s with ttfb {}s in {} chunks",
+        HumanBytes(len as u64),
```
flub, line 214: Did you mean to use `HumanBytes` here?  the `B` in the format string suggests you just want an integer number of bytes that are easy to parse?  The next line seems to do the human-friendly printing.

iroh/examples/transfer.rs
```diff
@@ -0,0 +1,313 @@
...
+
+async fn drain_stream(
+    stream: &mut iroh_net::endpoint::RecvStream,
+    read_unordered: bool,
+) -> Result<(usize, Duration, u64)> {
+    let mut read = 0;
+
+    let download_start = Instant::now();
+    let mut first_byte = true;
+    let mut ttfb = download_start.elapsed();
```
flub, line 237: FWIW I have no idea what `ttfb` stands for...

iroh/examples/transfer.rs
```diff
@@ -0,0 +1,313 @@
...
+        ttfb.as_secs_f64(),
+        chnk
+    );
+    println!(
+        "Transferred {} in {:.4}, {}/s",
+        HumanBytes(len as u64),
+        duration.as_secs_f64(),
+        HumanBytes((len as f64 / duration.as_secs_f64()) as u64)
+    );
+
```
flub, line 225: You now need to explicitly call `endpoint.close().await` to make a best effort to let the other side know you closed the connection.  If you don't do this and the datagram carrying the close was lost the other side will trigger a timeout.  Awaiting the `endpoint.close` makes sure to wait until the datagram carrying the close is acknowledged (or times out on the acknowledgement).

Currently you probably also want to put this `endpoint.close().await` wrapped in a application-level timeout like you do on the other side because our API WRT timeouts is a bit crap right now...

iroh/examples/transfer.rs
dignifiedquire: time to first byte probably

iroh/examples/transfer.rs
Arqu: The example itself is slightly verbose so that you know things are moving around when other folks play with it. It does not print any of the simulated data transfer. It's a cost Im willing to accept for a better user experience.

iroh/examples/transfer.rs
Arqu: Yes, mostly this is a precursor for moving the current netsim parsing to a different format. This just shows we have more data now.

Arqu: Did fix the extra B though.

iroh/examples/transfer.rs
Arqu: time to first byte :). I'll make sure the vars are more readable. TY

iroh/examples/transfer.rs
Arqu: Something like now?

iroh/examples/transfer.rs
flub: Yeah, fair.  Given the amount of data this doesn't actually matter as much, I didn't really realise this yet when I wrote this.

iroh/examples/transfer.rs
flub: Something like this here:

```suggestion
tokio::time::timeout(Duration::from_secs(3), async move {
    endpoint.close().await
})??;
```

PR #2961, opened 2024-11-22
feat(iroh-relay): Rate-limit client connections

flub:
## Description

Add a rate limit to the incoming data from client connections.

## Breaking Changes

If not configured there is now a default rate limit for incoming data from client connections: 4KiB/s steady-stream and 16MiB burst capacity.

## Notes & open questions

- The choice here is made to rate-limit the incoming bytes, regardless of what they are.  The benefit is that the incoming stream is slowed down, pushing back to the client over the TCP connection.  The downside is that someone who is rate-limited will get a fairly bad experience since all DISCO traffic is also delayed.

- Only rate-limiting non-disco traffic is an option, but it would not push back on the TCP stream, which is worse as then you'd still have to swallow all the incoming traffic.  Also it would be open to abuse fairly easy as the detection of disco packets is based on a magic number which could easily be spoofed.

- Maybe the `RateLimitedRelayedStream` should live in `stream.rs` next to the `RelayedStream`?  Not really sure.

### TODO

- [x] Allow rate-limit configuration in the config file.
- [x] Test config file loading.
- [x] Set a sensible default rate-limit.
- [x] Improve tests to more fully test the rate limiting.
- [x] Metrics when rate limits are hit.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh-relay/src/server/client_conn.rs
```diff
@@ -317,10 +324,153 @@ impl Actor {
...
+
+    #[instrument(name = "rate_limited_relayed_stream", skip_all)]
+    fn poll_next(
+        mut self: Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> Poll<Option<Self::Item>> {
+        loop {
+            // If we have a delay installed, we need to await it.
+            if let Some(ref mut wait_fut) = self.delay {
+                tokio::pin!(wait_fut);
```
dignifiedquire, line 364: I think you can just write `Pin::new(wait_fut)`, which is a bit easier to understand avoids the macro

iroh-relay/src/server/client_conn.rs
```diff
@@ -317,10 +324,153 @@ impl Actor {
...
+                match wait_fut.poll(cx) {
+                    Poll::Ready(_) => {
+                        self.delay.take();
```
dignifiedquire, line 367: you don't seem to be putting a new `delay` in there, ever, is that on purpose?

flub: Yes, we're allowed to receive frames again, so we take out the delay future and enter the loop anew.  On this next loop it will skip past polling the delay function and check if there's a buffered frame to yield or not.  That branch will put the delay future back in when if it is needed again.

flub: So line 411 puts the delay future in if that was what you missed.

iroh-relay/src/server/client_conn.rs
```diff
@@ -317,10 +327,148 @@ impl Actor {
...
+                        }
+                        Poll::Ready(None) => return Poll::Ready(None),
+                        Poll::Pending => return Poll::Pending,
+                    }
+                }
+                State::Blocked { delay, .. } => {
+                    match delay.poll(cx) {
+                        Poll::Ready(_) => {
+                            match std::mem::replace(&mut self.state, State::Ready) {
+                                State::Ready => continue, // unreachable
```
dignifiedquire, line 421: why not mark as `unreachable!`?

iroh-relay/src/server.rs
```diff
@@ -140,12 +140,44 @@ pub struct TlsConfig<EC: fmt::Debug, EA: fmt::Debug = EC> {
...
+    pub(super) const MAX: ClientConnRateLimit = ClientConnRateLimit {
+        bytes_per_second: NonZeroU32::MAX,
+        max_burst_bytes: NonZeroU32::MAX,
+    };
+}
+
+impl Default for ClientConnRateLimit {
+    fn default() -> Self {
+        Self {
+            bytes_per_second: NonZeroU32::try_from(1024 * 4).expect("nonzero"),
```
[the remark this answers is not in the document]
flub: Setting the defaults to u32::MAX now, which is essentially unlimited.  If you want rate limiting you need to configure it.

iroh-relay/src/protos/relay.rs
```diff
@@ -200,9 +203,14 @@ pub(crate) async fn recv_client_key<S: Stream<Item = anyhow::Result<Frame>> + Un
     }
 }
 
+/// The protocol for the relay server.
+///
+/// This is a framed protocol, using [`tokio_util::codec`] to turn the streams of bytes into
+/// [`Frame`]s.
 #[derive(Debug, Default, Clone)]
 pub(crate) struct DerpCodec;
```
Arqu, line 211: Shouldn't we at some point rename this?

iroh-relay/src/server.rs
```diff
@@ -140,12 +140,41 @@ pub struct TlsConfig<EC: fmt::Debug, EA: fmt::Debug = EC> {
...
+    pub bytes_per_second: NonZeroU32,
+    /// Max number of bytes to read in a single burst.
+    ///
+    /// Defaults to [`NonZeroU32::MAX`], effectively unlimited.
+    pub max_burst_bytes: NonZeroU32,
+}
+
+impl ClientConnRateLimit {
+    pub(super) const MAX: ClientConnRateLimit = ClientConnRateLimit {
+        bytes_per_second: NonZeroU32::MAX,
```
Arqu, line 169: To be fair this limits it to 4GB/sec, while not low, definitely not unlimited.

Arqu (commented):
Aside from sad CI, code looks good.
I did add another comment about defaults though :D

iroh-relay/src/protos/relay.rs
```diff
@@ -18,6 +18,9 @@ use tokio_util::codec::{Decoder, Encoder};
 /// including its on-wire framing overhead)
 pub const MAX_PACKET_SIZE: usize = 64 * 1024;
 
+/// The maximum frame size.
+///
+/// This is also the minimum burst size that a rate-limiter has to accept.
 const MAX_FRAME_SIZE: usize = 1024 * 1024;
```
dignifiedquire, line 24: note to @flub and future self, we should reevaluate if this is a good number

iroh-relay/src/protos/relay.rs
dignifiedquire: we should, but in a new PR

iroh-relay/src/server/client_conn.rs
```diff
@@ -66,9 +69,16 @@ impl ClientConn {
             stream: io,
             write_timeout,
             channel_capacity,
+            rate_limit: rate_limit_config,
             server_channel,
         } = config;
 
+        let quota = governor::Quota::per_second(rate_limit_config.bytes_per_second)
+            .allow_burst(rate_limit_config.max_burst_bytes);
+        // TODO: Allow creating this with mocked time for tests?
```
dignifiedquire, line 78: you can use this: https://docs.rs/tokio/latest/tokio/time/fn.advance.html for testing tokio time

iroh-relay/src/server/client_conn.rs
```diff
@@ -317,10 +327,166 @@ impl Actor {
...
+                                            // Item is rate-limited.
+                                            self.record_rate_limited();
+                                            let limiter = self.limiter.clone();
+                                            let delay = Box::pin(async move {
+                                                limiter.until_n_ready(frame_len).await.ok();
+                                            });
+                                            self.state = State::Blocked { delay, item };
+                                            continue;
+                                        }
+                                        Err(_insufficient_capacity) => {
```
dignifiedquire, line 418: how could this happen?

dignifiedquire (approved):

iroh-relay/src/protos/relay.rs
flub: Probably maybe.  But not today :)

iroh-relay/src/server.rs
flub: Yeah, I did this for simplicity.  Do you think this is unreasonable and we should do truly unlimited?

iroh-relay/src/server/client_conn.rs
flub: Oh, I should have removed that TODO.

We can't because `governor::RateLimiter` has its own time source.  It does parametrise on that clock, which I started plugging in to get this tests better... only to discover they have a `ReasonablyRealtime` trait which must be implemented for the `until_n_ready` method we use and the fake clock source does not implement that.  I get that this is a bit difficult to do... so decided a test of 1.2s is the best we can do.  I'm hoping it isn't going to turn out to be flaky though.

iroh-relay/src/server/client_conn.rs
flub: If someone configures a max_burst_bytes smaller than the MAX_FRAME_SIZE.  So this is effectively a mis-configuration, hence I opted to keep the relay functional and leave the frame through but log an error.

I've improved the error message to explain this a bit more.

flub:
@Arqu demanded that the rate limiter could be disabled.  So please review the new option spaghetti.

iroh-relay/src/server/client_conn.rs
```diff
@@ -384,6 +388,10 @@ impl Stream for RateLimitedRelayedStream {
         mut self: Pin<&mut Self>,
         cx: &mut std::task::Context<'_>,
     ) -> Poll<Option<Self::Item>> {
+        let Some(limiter) = self.limiter.clone() else {
```
dignifiedquire, line 391: shouldn't the clone happen only on the branch that needs it?

iroh-relay/src/server/client_conn.rs
```diff
@@ -354,10 +358,10 @@ enum State {
 }
 
 impl RateLimitedRelayedStream {
-    fn new(inner: RelayedStream, limiter: governor::DefaultDirectRateLimiter) -> Self {
+    fn new(inner: RelayedStream, limiter: Option<governor::DefaultDirectRateLimiter>) -> Self {
```
dignifiedquire, line 361: maybe two apis, `RateLimitedRelayedStream::unlimited(RelayedStream)` and `RateLimitedRelayedStream::new(RelayedStream, RateLimiter)`?

iroh-relay/src/server/client_conn.rs
flub: Good catch!  fixed.

iroh-relay/src/server/client_conn.rs
flub: also a good call.  done

iroh-relay/src/server.rs
flub: So unfortunately the rate limiter doesn't support u64 and I didn't look into why.  I assume they have a reason.

And using 0 for unlimited is both a type-level sin and also a problem for max_burst_bytes since 0 is actually a meaningful value and would be ambiguous and very confusing to users.

PR #2963, opened 2024-11-25
refactor(iroh-router)!: change accept to take an AsRef<[u8]>

rklaehn:
## Description

Change RouterBuilder::accept to take an AsRef<[u8]> instead of a Vec<u8>. Also change iroh node accept.

## Breaking Changes

- RouterBuilder::accept takes impl AsRef<[u8]>. Existing code should still work!
- Node::accept takes impl AsRef<[u8]>. Existing code should still work!

## Notes & open questions

Question: since we are making the API more pleasant at the expense of a bit of inefficiency, should we also change the second parameter to take a &Arc instead of an Arc?

So
```rust
fn accept(alpn: impl AsRef<[u8]>, proto: &Arc<dyn ProtocolHandler>)
```

In the various examples I found that you very frequently need the protocol handler again after accept-ing it. So this saves one .clone() in many cases, at the expense of an arc clone, which we basically treat as free anyway.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


flub (approved):
Nice, I like this.

flub:
> Question: since we are making the API more pleasant at the expense of a bit of inefficiency, should we also change the second parameter to take a &Arc instead of an Arc?
> 
> So
> 
> ```rust
> fn accept(alpn: impl AsRef<[u8]>, proto: &Arc<dyn ProtocolHandler>)
> ```
> 
> In the various examples I found that you very frequently need the protocol handler again after accept-ing it. So this saves one .clone() in many cases, at the expense of an arc clone, which we basically treat as free anyway.

I find `&Arc<dyn ProtocolHandler>` a very weird signature though.  That's purely an emotional response and not a technical one though.

dignifiedquire (approved):

PR #2965, opened 2024-11-25
feat(iroh-relay)!: add a QUIC server for QUIC address discovery to the iroh relay.

### LINKED ISSUE #2964: feat(iroh-relay): add QUIC endpoint for QUIC Address Discovery
Also, need to add configuration to `RelayNode` in `iroh-net`, that is enabled by default, that allows you to configure a custom QUIC port for that relay.

ramfox:
## Description

This PR adds a QUIC endpoint to the relay server that can do QUIC address discovery. It also contains structs/functions for properly doing the Client side interaction for this process.

Also, this adjust the `RelayNode` to include configuration on how to speak to the QUIC endpoint on the relay server.

QUIC is disabled by default and requires a `TlsConfig` to be configured in order to work.

closes #2964 

## Breaking Changes

- `iroh_base::relay_map::RelayNode` now has field `quic` that takes a `Option<iroh_base::relay_map::QuicConfig>`
- `iroh::test_utils::run_relay_server_with(stun: Option<StunConfig>)` => `iroh::test_utils::run_relay_server_with(stun: Option<StunConfig>, quic: bool)`
    - when `quic` is `true`, it will start a quic server for QUIC address discovery, that has self signed tls certs for testing.
- `iroh_relay::server::ServerConfig` has field `quic` that takes a `Option<iroh_relay::server::QuicConfig>`
- `iroh_relay::server::TlsConfig.quic_bind_addr` is a new field that takes a `SocketAddr`
- `iroh_relay::server::TlsConfig.server_config` is a new field that takes a `rustls::ServerConfig`
- field `config` has been removed from variant `iroh_relay::server::CertConfig::LetsEncrypt`
- variant `iroh_relay::server::CertConfig::LetsEncrypt` has a new field `state` that takes a `tokio_rustls_acme::AcmeState<EC, EA>`
- variant `iroh_relay::server::CertConfig::Manual` no longer has field `private_key`

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/example.config.toml
+++ b/example.config.toml
@@ -1,4 +1,4 @@
-[[derp_nodes]]
+[[relay_nodes]]
 url = "https://foo.bar"
 stun_only = false
 stun_port = 1244
--- a/iroh-base/src/relay_map.rs
+++ b/iroh-base/src/relay_map.rs
@@ -16,7 +16,7 @@ pub const DEFAULT_STUN_PORT: u16 = 3478;
 /// for QUIC address discovery
 ///
 /// The port is "QUIC" typed on a phone keypad.
-pub const DEFAULT_QUIC_PORT: u16 = 7842;
+pub const DEFAULT_RELAY_QUIC_PORT: u16 = 7842;
 
 /// Configuration of all the relay servers that can be used.
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -68,7 +68,7 @@ impl RelayMap {
     /// Allows to set a custom STUN port and different IP addresses for IPv4 and IPv6.
     /// If IP addresses are provided, no DNS lookup will be performed.
     ///
-    /// Sets the port to the default [`DEFAULT_QUIC_PORT`].
+    /// Sets the port to the default [`DEFAULT_RELAY_QUIC_PORT`].
     pub fn default_from_node(url: RelayUrl, stun_port: u16) -> Self {
         let mut nodes = BTreeMap::new();
         nodes.insert(
@@ -155,7 +155,7 @@ pub struct QuicConfig {
 impl Default for QuicConfig {
     fn default() -> Self {
         Self {
-            port: DEFAULT_QUIC_PORT,
+            port: DEFAULT_RELAY_QUIC_PORT,
         }
     }
 }
--- a/iroh-base/src/relay_map.rs
+++ b/iroh-base/src/relay_map.rs
@@ -141,7 +141,7 @@ pub struct RelayNode {
 }
 
 fn quic_port() -> u16 {
-    0
+    DEFAULT_QUIC_PORT
 }
 
 impl fmt::Display for RelayNode {
--- a/iroh-base/src/relay_map.rs
+++ b/iroh-base/src/relay_map.rs
@@ -77,7 +77,7 @@ impl RelayMap {
                 url,
                 stun_only: false,
                 stun_port,
-                quic_port: DEFAULT_QUIC_PORT,
+                quic: Some(QuicConfig::default()),
             }
             .into(),
         );
@@ -133,15 +133,31 @@ pub struct RelayNode {
     ///
     /// Setting this to `0` means the default STUN port is used.
     pub stun_port: u16,
-    /// The QUIC endpoint port of the relay server.
+    /// Configuration to speak to the QUIC endpoint on the relay server.
     ///
-    /// Setting this to `0` means the default QUIC port is used.
-    #[serde(default = "quic_port")]
-    pub quic_port: u16,
+    /// When `None`, we will not attempt to do QUIC address discovery
+    /// with this relay server.
+    #[serde(default = "quic_config")]
+    pub quic: Option<QuicConfig>,
 }
 
-fn quic_port() -> u16 {
-    DEFAULT_QUIC_PORT
+fn quic_config() -> Option<QuicConfig> {
+    Some(QuicConfig::default())
+}
+
+/// Configuration for speaking to the QUIC endpoint on the relay
+/// server to do QUIC address discovery.
+#[derive(Debug, Deserialize, Serialize, Clone, Eq, PartialEq, PartialOrd, Ord)]
+pub struct QuicConfig {
+    pub port: u16,
+}
+
+impl Default for QuicConfig {
+    fn default() -> Self {
+        Self {
+            port: DEFAULT_QUIC_PORT,
+        }
+    }
 }
 
 impl fmt::Display for RelayNode {
--- a/iroh-base/src/relay_map.rs
+++ b/iroh-base/src/relay_map.rs
@@ -134,7 +134,12 @@ pub struct RelayNode {
     /// The QUIC endpoint port of the relay server.
     ///
     /// Setting this to `0` means the default QUIC port is used.
+    #[serde(default = "quic_port")]
     pub quic_port: u16,
+}
+
+fn quic_port() -> u16 {
+    0
 }
 
 impl fmt::Display for RelayNode {
--- a/iroh-base/src/relay_map.rs
+++ b/iroh-base/src/relay_map.rs
@@ -67,7 +67,9 @@ impl RelayMap {
     ///
     /// Allows to set a custom STUN port and different IP addresses for IPv4 and IPv6.
     /// If IP addresses are provided, no DNS lookup will be performed.
-    pub fn default_from_node(url: RelayUrl, stun_port: u16, quic_port: u16) -> Self {
+    ///
+    /// Sets the port to the default [`DEFAULT_QUIC_PORT`].
+    pub fn default_from_node(url: RelayUrl, stun_port: u16) -> Self {
         let mut nodes = BTreeMap::new();
         nodes.insert(
             url.clone(),
@@ -75,7 +77,7 @@ impl RelayMap {
                 url,
                 stun_only: false,
                 stun_port,
-                quic_port,
+                quic_port: DEFAULT_QUIC_PORT,
             }
             .into(),
         );
@@ -92,7 +94,7 @@ impl RelayMap {
     /// resolved from the URL's host name via DNS.
     /// relay nodes are specified at <../../docs/relay_nodes.md>
     pub fn from_url(url: RelayUrl) -> Self {
-        Self::default_from_node(url, DEFAULT_STUN_PORT, DEFAULT_QUIC_PORT)
+        Self::default_from_node(url, DEFAULT_STUN_PORT)
     }
 
     /// Constructs the [`RelayMap`] from an iterator of [`RelayNode`]s.
--- a/iroh-base/src/relay_map.rs
+++ b/iroh-base/src/relay_map.rs
@@ -10,7 +10,13 @@ pub use crate::relay_url::RelayUrl;
 /// The default STUN port used by the Relay server.
 ///
 /// The STUN port as defined by [RFC 8489](<https://www.rfc-editor.org/rfc/rfc8489#section-18.6>)
-const DEFAULT_STUN_PORT: u16 = 3478;
+pub const DEFAULT_STUN_PORT: u16 = 3478;
+
+/// The default QUIC port used by the Relay server to accept QUIC connections
+/// for QUIC address discovery
+///
+/// The port is "QUIC" typed on a phone keypad.
+pub const DEFAULT_QUIC_PORT: u16 = 7842;
 
 /// Configuration of all the relay servers that can be used.
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -61,7 +67,7 @@ impl RelayMap {
     ///
     /// Allows to set a custom STUN port and different IP addresses for IPv4 and IPv6.
     /// If IP addresses are provided, no DNS lookup will be performed.
-    pub fn default_from_node(url: RelayUrl, stun_port: u16) -> Self {
+    pub fn default_from_node(url: RelayUrl, stun_port: u16, quic_port: u16) -> Self {
         let mut nodes = BTreeMap::new();
         nodes.insert(
             url.clone(),
@@ -69,6 +75,7 @@ impl RelayMap {
                 url,
                 stun_only: false,
                 stun_port,
+                quic_port,
             }
             .into(),
         );
@@ -80,10 +87,12 @@ impl RelayMap {
 
     /// Returns a [`RelayMap`] from a [`RelayUrl`].
     ///
-    /// This will use the default STUN port and IP addresses resolved from the URL's host name via DNS.
+    /// This will use the default STUN port, the default QUIC port
+    /// (as defined by the `iroh-relay` crate) and IP addresses
+    /// resolved from the URL's host name via DNS.
     /// relay nodes are specified at <../../docs/relay_nodes.md>
     pub fn from_url(url: RelayUrl) -> Self {
-        Self::default_from_node(url, DEFAULT_STUN_PORT)
+        Self::default_from_node(url, DEFAULT_STUN_PORT, DEFAULT_QUIC_PORT)
     }
 
     /// Constructs the [`RelayMap`] from an iterator of [`RelayNode`]s.
@@ -122,6 +131,10 @@ pub struct RelayNode {
     ///
     /// Setting this to `0` means the default STUN port is used.
     pub stun_port: u16,
+    /// The QUIC endpoint port of the relay server.
+    ///
+    /// Setting this to `0` means the default QUIC port is used.
+    pub quic_port: u16,
 }
 
 impl fmt::Display for RelayNode {
--- a/iroh-relay/Cargo.toml
+++ b/iroh-relay/Cargo.toml
@@ -128,7 +128,6 @@ server = [
 ]
 metrics = ["iroh-metrics/metrics", "server"]
 test-utils = []
-dangerous-certs = []
 
 [[bin]]
 name = "iroh-relay"
--- a/iroh-relay/Cargo.toml
+++ b/iroh-relay/Cargo.toml
@@ -133,7 +133,7 @@ dangerous-certs = []
 [[bin]]
 name = "iroh-relay"
 path = "src/main.rs"
-required-features = ["server", "dangerous-certs"]
+required-features = ["server"]
 
 [package.metadata.docs.rs]
 all-features = true
--- a/iroh-relay/Cargo.toml
+++ b/iroh-relay/Cargo.toml
@@ -50,6 +50,8 @@ postcard = { version = "1", default-features = false, features = [
     "use-std",
     "experimental-derive",
 ] }
+quinn = { package = "iroh-quinn", version = "0.12.0" }
+quinn-proto = { package = "iroh-quinn-proto", version = "0.12.0" }
 rand = "0.8"
 rcgen = { version = "0.13", optional = true }
 regex = { version = "1.7.1", optional = true }
@@ -126,11 +128,12 @@ server = [
 ]
 metrics = ["iroh-metrics/metrics", "server"]
 test-utils = []
+dangerous-certs = []
 
 [[bin]]
 name = "iroh-relay"
 path = "src/main.rs"
-required-features = ["server"]
+required-features = ["server", "dangerous-certs"]
 
 [package.metadata.docs.rs]
 all-features = true
--- a/iroh-relay/src/client.rs
+++ b/iroh-relay/src/client.rs
@@ -378,27 +378,21 @@ impl ClientBuilder {
 }
 
 #[cfg(any(test, feature = "test-utils", feature = "dangerous-certs"))]
-/// Creates a client config that trusts relay servers without verification
+/// Creates a client config that trusts any servers without verifying their TLS certificate.
 ///
 /// Should be used for testing local relay setups only.
-pub fn make_dangerous_client_config() -> quinn::crypto::rustls::QuicClientConfig {
-    let roots = rustls::RootCertStore {
-        roots: webpki_roots::TLS_SERVER_ROOTS.to_vec(),
-    };
-    let mut config = rustls::client::ClientConfig::builder_with_provider(Arc::new(
+pub fn make_dangerous_client_config() -> rustls::ClientConfig {
+    warn!(
+        "Insecure config: SSL certificates from relay servers will be trusted without verification"
+    );
+    rustls::client::ClientConfig::builder_with_provider(Arc::new(
         rustls::crypto::ring::default_provider(),
     ))
     .with_protocol_versions(&[&rustls::version::TLS13])
     .expect("protocols supported by ring")
-    .with_root_certificates(roots)
-    .with_no_client_auth();
-    warn!(
-        "Insecure config: SSL certificates from relay servers will be trusted without verification"
-    );
-    config
-        .dangerous()
-        .set_certificate_verifier(Arc::new(NoCertVerifier));
-    quinn::crypto::rustls::QuicClientConfig::try_from(config).expect("Using tls1.3")
+    .dangerous()
+    .with_custom_certificate_verifier(Arc::new(NoCertVerifier))
+    .with_no_client_auth()
 }
 
 impl ClientReceiver {
@@ -1134,12 +1128,15 @@ impl DnsExt for DnsResolver {
 }
 
 /// Used to allow self signed certificates in tests
-#[cfg(any(test, feature = "test-utils"))]
-#[cfg_attr(iroh_docsrs, doc(cfg(any(test, feature = "test-utils"))))]
+#[cfg(any(test, feature = "test-utils", feature = "dangerous-certs"))]
+#[cfg_attr(
+    iroh_docsrs,
+    doc(cfg(any(test, feature = "test-utils", feature = "dangerous-certs")))
+)]
 #[derive(Debug)]
-pub(crate) struct NoCertVerifier;
+struct NoCertVerifier;
 
-#[cfg(any(test, feature = "test-utils"))]
+#[cfg(any(test, feature = "test-utils", feature = "dangerous-certs"))]
 impl rustls::client::danger::ServerCertVerifier for NoCertVerifier {
     fn verify_server_cert(
         &self,
--- a/iroh-relay/src/client.rs
+++ b/iroh-relay/src/client.rs
@@ -377,11 +377,11 @@ impl ClientBuilder {
     }
 }
 
-#[cfg(any(test, feature = "test-utils", feature = "dangerous-certs"))]
+#[cfg(test)]
 /// Creates a client config that trusts any servers without verifying their TLS certificate.
 ///
 /// Should be used for testing local relay setups only.
-pub fn make_dangerous_client_config() -> rustls::ClientConfig {
+pub(crate) fn make_dangerous_client_config() -> rustls::ClientConfig {
     warn!(
         "Insecure config: SSL certificates from relay servers will be trusted without verification"
     );
@@ -1128,15 +1128,12 @@ impl DnsExt for DnsResolver {
 }
 
 /// Used to allow self signed certificates in tests
-#[cfg(any(test, feature = "test-utils", feature = "dangerous-certs"))]
-#[cfg_attr(
-    iroh_docsrs,
-    doc(cfg(any(test, feature = "test-utils", feature = "dangerous-certs")))
-)]
+#[cfg(any(test, feature = "test-utils"))]
+#[cfg_attr(iroh_docsrs, doc(cfg(any(test, feature = "test-utils"))))]
 #[derive(Debug)]
 struct NoCertVerifier;
 
-#[cfg(any(test, feature = "test-utils", feature = "dangerous-certs"))]
+#[cfg(any(test, feature = "test-utils"))]
 impl rustls::client::danger::ServerCertVerifier for NoCertVerifier {
     fn verify_server_cert(
         &self,
--- a/iroh-relay/src/client.rs
+++ b/iroh-relay/src/client.rs
@@ -377,6 +377,30 @@ impl ClientBuilder {
     }
 }
 
+#[cfg(any(test, feature = "test-utils", feature = "dangerous-certs"))]
+/// Creates a client config that trusts relay servers without verification
+///
+/// Should be used for testing local relay setups only.
+pub fn make_dangerous_client_config() -> quinn::crypto::rustls::QuicClientConfig {
+    let roots = rustls::RootCertStore {
+        roots: webpki_roots::TLS_SERVER_ROOTS.to_vec(),
+    };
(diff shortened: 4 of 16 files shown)


iroh-base/src/relay_map.rs
```diff
@@ -122,6 +131,15 @@ pub struct RelayNode {
...
     pub stun_port: u16,
+    /// The QUIC endpoint port of the relay server.
+    ///
+    /// Setting this to `0` means the default QUIC port is used.
+    #[serde(default = "quic_port")]
+    pub quic_port: u16,
+}
+
+fn quic_port() -> u16 {
+    0
```
dignifiedquire, line 142: why not use the default port?

iroh-net/src/defaults.rs
```diff
@@ -1,14 +1,19 @@
...
-use url::Url;
-
-use crate::{RelayMap, RelayNode};
-
+use iroh_base::relay_map::QuicConfig;
+/// The default QUIC port used by the Relay server to accept QUIC connections
+/// for QUIC address discovery
+///
+/// The port is "QUIC" typed on a phone keypad.
+pub use iroh_base::relay_map::DEFAULT_QUIC_PORT;
```
dignifiedquire, line 8: if we reexport this, maybe as sth like `DEFAULT_RELAY_QUIC_PORT`?

iroh-relay/src/client.rs
```diff
@@ -377,6 +377,30 @@ impl ClientBuilder {
     }
 }
 
+#[cfg(any(test, feature = "test-utils", feature = "dangerous-certs"))]
+/// Creates a client config that trusts relay servers without verification
```
dignifiedquire, line 381: more specifically: trusts any server without verifying their TLS certificate

iroh-relay/src/main.rs
```diff
@@ -132,6 +132,27 @@ struct Config {
...
+    enable_quic_addr_discovery: bool,
+    /// Whether to use self-signed local certificates for the QUIC server.
+    ///
+    /// Only considered when `enable_quic_addr_discovery` is `true` and should
+    /// only be used when the relay is run locally in dev mode.
+    ///
+    /// Will ignore any other tls certificates passed to the relay server.
+    ///
+    /// Binds the QUIC server to "127.0.0.1".
+    enable_quic_local_cert: bool,
```
dignifiedquire, line 148: maybe we should prefix this with sth like `dangerous_`?

iroh-relay/src/main.rs
```diff
@@ -132,6 +132,27 @@ struct Config {
...
+    /// The port to bind the QUIC server on.
+    ///
+    /// The socket address will use the ip address from the tls config.
+    ///
+    /// If `None`, then the [`DEFAULT_QUIC_PORT`] will be used. If 0,
+    /// then the socket will bind to a random unused port.
+    quic_bind_port: Option<u16>,
```
dignifiedquire, line 155: just `quic_port`?

iroh-relay/src/main.rs
```diff
@@ -362,46 +391,83 @@ async fn main() -> Result<()> {
...
+/// Convert the TOML-loaded config to the [`relay::RelayConfig`] format.
+async fn build_relay_config(cfg: Config) -> Result<relay::ServerConfig<std::io::Error>> {
+    let relay_tls = maybe_load_tls(&cfg).await?;
+
+    let mut quic_config = None;
+    if cfg.enable_quic_addr_discovery {
+        if let Some(ref tls) = relay_tls {
+            quic_config = Some(QuicConfig::new(
+                tls.server_config.clone(),
+                tls.https_bind_addr.ip(),
```
dignifiedquire, line 456: should we rename this config value if we change its meaning?

iroh-relay/src/main.rs
```diff
@@ -362,46 +391,83 @@ async fn main() -> Result<()> {
...
+                cfg.quic_bind_port,
+            )?);
+        } else if cfg.enable_quic_local_cert {
+            let (_, server_config) = iroh_relay::server::self_signed_tls_certs_and_config();
+            quic_config = Some(QuicConfig::new(
+                server_config,
+                Ipv6Addr::UNSPECIFIED.into(),
```
dignifiedquire, line 463: why ipv6 by default?

iroh-relay/src/quic.rs
```diff
@@ -0,0 +1,301 @@
...
+            let endpoint = quinn::Endpoint::server(server_config, quic_config.bind_addr)?;
+            let bind_addr = endpoint.local_addr()?;
+
+            info!("QUIC server bound on {bind_addr:?}");
+
+            let cancel = CancellationToken::new();
+            let cancel_accept_loop = cancel.clone();
+
+            let task = tokio::task::spawn(async move {
+            let mut set = JoinSet::new();
```
dignifiedquire, line 81: missing indent?

iroh-relay/src/quic.rs
```diff
@@ -0,0 +1,301 @@
...
+                bind_addr,
+                cancel,
+                handle: AbortOnDropHandle::new(task),
+            })
+        }
+
+        /// Closes the underlying QUIC endpoint and the tasks running the
+        /// QUIC connections.
+        pub fn shutdown(&self) {
+            self.cancel.cancel();
```
dignifiedquire, line 131: `async` and await the task?

iroh-relay/src/server.rs
```diff
@@ -122,6 +138,33 @@ pub struct StunConfig {
     pub bind_addr: SocketAddr,
 }
 
+/// Configuration for the QUIC server.
+// TODO(ramfox): limits? Limiting connections or accepted # of connections from a single IP? from a single node ID?
```
dignifiedquire, line 142: we also need to likely adjust the stream/connection limits for a server here

iroh-relay/src/main.rs
ramfox: quinn specific: if you bind to `Ipv6Addr::UNSPECIFIED` it will attempt to bind to both `Ipv6Addr::UNSPECIFIED` and `Ipv4Addr::UNSPECIFIED` if you can do dual-stack sockets on that machine.

dignifiedquire: I see, can you add a comment about this?

iroh-relay/src/quic.rs
```diff
@@ -0,0 +1,301 @@
...
+
+    impl ServerHandle {
+        /// Gracefully shut down the quic endpoint.
+        pub fn shutdown(&self) {
+            self.cancel_token.cancel()
+        }
+    }
+
+    async fn handle_connection(conn: quinn::Incoming) -> Result<()> {
+        let connection = conn.await?;
```
flub, line 151: This will trigger all the garbage that gets sent to this socket and log them as a warning by the caller.  (Quinn itself doesn't document this very well, we have at least some explanation hidden somewhere so we can say "told you so": https://docs.rs/iroh-net/latest/iroh_net/endpoint/struct.Incoming.html#method.accept).

So perhaps we can create something like:
```rust
#[thiserror::Error]
enum ConnError {
    Incoming(quinn::ConnectionError),
    Real(#[from] anyhow::Error),
}
```
so that the caller can distinguish maybe only debug-log the incoming errors, and then map the error from the incoming.

iroh-relay/src/quic.rs
```diff
@@ -0,0 +1,301 @@
...
+        info!("established");
```
flub, line 152: That's too high for prod, should be debug.

iroh-relay/src/quic.rs
```diff
@@ -0,0 +1,301 @@
...
+        /// # Panics
+        /// If there is a panic during a connection, it will be propagated
+        /// up here. Any other errors in a connection will be logged as a
+        ///  warning.
+        pub(crate) fn spawn(quic_config: QuicConfig) -> Result<Self> {
+            let mut server_config =
+                quinn::ServerConfig::with_crypto(Arc::new(quic_config.server_config));
+            let transport_config = Arc::get_mut(&mut server_config.transport).unwrap();
+            transport_config
+                .max_concurrent_uni_streams(0_u8.into())
```
flub, line 68: Should we also disable `max_concurrent_bidi_streams`?  That would avoid another set of buffers that can be filled by a malicious remote.  And my understanding from the protocol is that no data is ever sent.

iroh-relay/src/quic.rs
```diff
@@ -0,0 +1,301 @@
...
+        pub fn shutdown(&self) {
+            self.cancel_token.cancel()
+        }
+    }
+
+    async fn handle_connection(conn: quinn::Incoming) -> Result<()> {
+        let connection = conn.await?;
+        info!("established");
+        // wait for the client to close the connection
+        let connection_err = connection.closed().await;
```
flub, line 154: Do we need a timeout on this?  Or do we allow this to be connected forever?  I guess forever is fine and allows for more interesting use in the future.

iroh-relay/src/quic.rs
```diff
@@ -0,0 +1,301 @@
...
+        match connection_err {
+            quinn::ConnectionError::ApplicationClosed(ApplicationClose { error_code, .. })
+                if error_code == QUIC_ADDR_DISC_CLOSE_CODE =>
+            {
+                return Ok(());
+            }
+            _ => {
+                warn!(
```
flub, line 162: I'd turn down this level to debug.  It is something we as the server have no control over.  While it is interesting to get this info, it is not a problem with the server.

iroh-relay/src/quic.rs
```diff
@@ -0,0 +1,301 @@
...
+                    "{} - error closing connection {connection_err:?}",
+                    connection.remote_address()
```
flub, line 164:
```suggestion
                    error_code, reason, "error closing connection {connection_err:#}"
```
The address is already part of the span.

You'll need to expose reason to add it as a field.

flub (commented):
Don't let the extensive commenting put you off, it looks good!

The config plumbing is just painful.

example.config.toml
```diff
@@ -1,4 +1,4 @@
-[[derp_nodes]]
+[[relay_nodes]]
```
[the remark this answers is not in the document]
flub: Looking at this afterwards maybe this change is by mistake?  I don't think I've seen the code change for it anywhere.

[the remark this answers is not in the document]
ramfox: this is just an example config file that we've had floating in our repo and it's been wrong for quite some time now...

iroh-net/src/test_utils.rs
```diff
@@ -66,10 +92,14 @@ pub async fn run_relay_server_with(
     let url: RelayUrl = format!("https://{}", server.https_addr().expect("configured"))
         .parse()
         .unwrap();
+    let quic = Some(iroh_base::relay_map::QuicConfig {
+        port: server.quic_addr().map_or(DEFAULT_QUIC_PORT, |s| s.port()),
```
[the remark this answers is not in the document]
ramfox: oh yea duh ty

iroh-relay/src/quic.rs
```diff
@@ -0,0 +1,301 @@
+//! Create a QUIC server that accepts connections
+//! for QUIC address discovery.
+use std::{net::SocketAddr, sync::Arc};
+
+use anyhow::Result;
+use quinn::VarInt;
+
+/// ALPN for our quic addr discovery
+pub const ALPN_QUIC_ADDR_DISC: &[u8] = b"quic";
```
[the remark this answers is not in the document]
ramfox: oh! i need to remove this, this is unused! We don't negotiate any alpns because QAD doesn't have anything to do with what protocol you are using and is negotiated based on the quic transport parameters. We don't even send any data over, we just handshake to establish the connection and then wait for the server to send over the observed addresses.

iroh-relay/src/quic.rs
```diff
@@ -0,0 +1,301 @@
...
+/// Endpoint close error code
+pub const QUIC_ADDR_DISC_CLOSE_CODE: VarInt = VarInt::from_u32(0);
```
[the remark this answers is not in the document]
ramfox: changing to 1 sounds good to me

example.config.toml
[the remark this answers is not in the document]
flub: oh lol, you're fixing a broken example.  :heart:

iroh-relay/src/quic.rs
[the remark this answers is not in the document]
flub: I would argue there is some kind of protocol here for qad:  namely that clients nor servers are allowed to create any streams and can only connect and disconnect and keep the connection alive using normal quic transport config things.  That's a protocol.

iroh-relay/src/server.rs
```diff
@@ -122,6 +138,33 @@ pub struct StunConfig {
...
+    ///
+    /// Normally you'd chose port `7842`, see [`crate::defaults::DEFAULT_QUIC_PORT`].
+    pub bind_addr: SocketAddr,
+    /// The TLS server configuration for the QUIC server.
+    #[debug(skip)]
+    pub server_config: QuicServerConfig,
+}
+
+impl QuicConfig {
+    /// Create a new [`QuicConfig`] from a [`TlsConfig`].
```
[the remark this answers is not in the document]
ramfox: ahh yes, thank you. this used to be MUCH more complicated, but as I understood what was going on better, it slimmed down to this! I'll see what removing it looks like 😂

iroh-relay/src/quic.rs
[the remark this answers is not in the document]
ramfox: fair. okay, i'll add in the ALPN to the endpoints.

iroh-net/src/defaults.rs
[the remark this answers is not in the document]
ramfox: For now, net-report just ignores the `quic` field, so this shouldn't be a problem.

iroh-relay/src/quic.rs
```diff
@@ -0,0 +1,301 @@
...
+                        Some(conn) => {
+                             debug!("accepting connection from {:?}", conn.remote_address())       ;
+                             set.spawn(async move {
+                                 let remote_addr = conn.remote_address();
+                                 let res = handle_connection(conn).await;
+                                 if let Err(ref err) = res {
+                                     warn!(remote_address = ?remote_addr, "error handling connection {err:?}")
+                                 }
+                                 res
+                             });
```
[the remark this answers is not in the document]
ramfox: ty for pointing this out. it isn't fully natural to me yet, but soooo obviously better when looking through logs 😂

iroh-relay/src/quic.rs
```diff
@@ -0,0 +1,301 @@
...
+}
+
+impl QuicClient {
+    /// Create a new QuicClient to handle the client side of QUIC
+    /// address discovery.
+    pub fn new(ep: quinn::Endpoint, mut client_config: quinn::ClientConfig) -> Self {
+        let mut transport = quinn_proto::TransportConfig::default();
+        // enable address discovery
+        transport
+            .send_observed_address_reports(true)
```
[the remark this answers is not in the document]
ramfox: I had some logic for keeping it, but I don't remember what it was, which tells me it probably wasn't very logical in the first place 😂 

removing it.

iroh-relay/src/server.rs
```diff
@@ -153,14 +198,12 @@ pub struct Limits {
...
-        /// Configuration for Let's Encrypt certificates.
+        /// State for Let's Encrypt certificates.
         #[debug("AcmeConfig")]
-        config: tokio_rustls_acme::AcmeConfig<EC, EA>,
+        state: tokio_rustls_acme::AcmeState<EC, EA>,
     },
     /// Use a static TLS key and certificate chain.
     Manual {
-        /// The TLS private key.
-        private_key: rustls::pki_types::PrivateKeyDer<'static>,
```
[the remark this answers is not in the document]
ramfox: We used to generate the server config much later in the process. Once you have the certs & server config generated, you don't need the private keys anymore. the `private_key` field became unused.

iroh-relay/src/server.rs
ramfox: https://github.com/n0-computer/iroh/issues/2967

iroh-relay/src/quic.rs
ramfox: not all connection errors have error_codes or reasons, but I see what you are saying!

iroh-relay/src/quic.rs
ramfox: yea, i had this same question and went for the easiest path right now. Trying to understand how this could be abused, but since we don't accept any streams, it seems like the worst someone can do is open up a ton of connections and just hang around? But then they are also hanging around with a bunch of connection opens 😂 Anyway, if you don't have a strong opinion about this right now, i'd like to leave it as is.

iroh-relay/src/main.rs
ramfox: yea i'm into the `tls.quic_bind_addr` config option

iroh-relay/src/server.rs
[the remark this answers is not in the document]
ramfox: Okay, so now, with the `quic_bind_addr` in the TlsConfig, it actually makes sense to keep this and pass in a TlsConfig so that the QuicConfig is supplied with all the correct pieces.

iroh-relay/src/main.rs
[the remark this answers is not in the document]
ramfox: it's now moved into the TlsConfig!

Need to do a bit more auditing to make sure that we are still binding to the addrs we want, by default, but the config is looking cleaner!

iroh-relay/src/server.rs
flub: I took the liberty to edit that issue.

iroh-relay/src/quic.rs
flub: Agreed that opening a ton of connections is the biggest malicious use right now.  And that's covered by the generic "do we need connection limits" issue.  So agree that we can leave this like this for now.

iroh-relay/src/quic.rs
ramfox: ooo okay while implementing this, I saw that we end up debug-logging all other errors that happen in this method anyway. So I'm just debug-logging all errors in this method and not returning anything.

iroh-relay/src/main.rs
```diff
@@ -362,46 +391,83 @@ async fn main() -> Result<()> {
...
-                        .cache_option(Some(DirCache::new(tls.cert_dir())))
-                        .directory_lets_encrypt(tls.prod_tls);
-                    relay::CertConfig::LetsEncrypt { config }
-                }
-            };
-            Some(relay::TlsConfig {
-                https_bind_addr: tls.https_bind_addr(&cfg),
-                cert: cert_config,
+async fn maybe_load_tls(
+    cfg: &Config,
```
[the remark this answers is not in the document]
ramfox: hmmm, okay so I want to do this, but we use the `cfg: &Config` to generate the `https_bind_port` and `quic_bind_port`.

I think the params may have to stay as-is

iroh-relay/src/quic.rs
ramfox: to circle back to this, I added back `Return<()>` to the `handle_connection` method, but now we:

- log all errors in this method at the "debug" level (no more `warn` for `Incoming` errors)
- when errors are no longer handled at the call site, we just pass the `Result` along to the join set
- at the `set.join_next()` site, we only take action if the error is a `panic`, in which case we call `panic!`, otherwise the error has already been logged & we ignore it.

ramfox:
@flub can you take a look at the config options for the relay server again.

I removed the option to generate self-signed certs automatically and added info to the readme about how to pass those in.

However, the `--dev` flag removes all tls config in the code and I wanted to keep the "simple" dev mode available for those who don't care about QUIC addr disc. SO there are now 2 dev-like flags, `--dev` that runs the relay server locally over http with QAD disabled and `--dev-quic` that expects tls to be configured for QAD, but still runs the server over http. 

Is this the worst? I tried to keep it as simple as possible 😂  but could use another set of eyes.

ramfox:
actually, @flub 

I'm wondering if instead, `enabled_quic_addr_discovery` should actually default to `false` and if we should add a `--quic` flag on the CLI that explicitly enables it instead of doing this weird `--dev-quic` thing.

So if you want to run QAD locally you would run:
`cargo run --bin iroh-relay -- --config-path=PATH --dev --quic`

flub:
> That feels cleaner? thoughts?

Hum, it is awkward that TLS certs are basically becoming mandatory even for a local dev relay server.  But it is inevitable and we'll only grow more dependence over a QUIC transport over time.

But, the good news is that we don't have to get this right from the start.  The level of stability for `--dev` really doesn't matter that much.  And as you say, right now qad is just an optional extra rather than an integral part.

So I'm fine with both of these options.  And yes, the latter one, disable qad by default, is the simpler one so why not.

iroh-relay/src/quic.rs
```diff
@@ -0,0 +1,325 @@
...
+                                    }
+                                }
+                            }
+                            res = endpoint.accept() => match res {
+                                Some(conn) => {
+                                     debug!("accepting connection");
+                                     let remote_addr = conn.remote_address();
+                                     set.spawn(async move {
+                                         handle_connection(conn).await
+                                     }.instrument(info_span!("qad-conn", %remote_addr)));
```
flub, line 113: I think you can write this without wrapping it in another future.  Something like:
```suggestion
                                     set.spawn(
                                         handle_connection(conn)
                                         .instrument(info_span!("qad-conn", %remote_addr))
                                     );
```

iroh-relay/src/quic.rs
```diff
@@ -0,0 +1,325 @@
...
+        // wait for the client to close the connection
+        let connection_err = connection.closed().await;
+        match connection_err {
+            quinn::ConnectionError::ApplicationClosed(ApplicationClose { error_code, .. })
+                if error_code == QUIC_ADDR_DISC_CLOSE_CODE =>
+            {
+                Ok(())
+            }
+            _ => {
+                debug!("error closing connection {connection_err:#?}",);
```
flub, line 193: "connection closed with error"?

This side is not doing the active closing I guess.

iroh-relay/src/quic.rs
```diff
@@ -0,0 +1,325 @@
...
+        pub fn shutdown(&self) {
+            self.cancel_token.cancel()
+        }
+    }
+
+    /// Handle the connection from the client.
+    ///
+    /// Any errors that happen during this connection do not need to be handled,
+    /// and will be logged at the debug level in this function.
+    async fn handle_connection(incoming: quinn::Incoming) -> Result<()> {
```
flub, line 175: I find it a bit odd that this returns `Result` but then nothing ever does anything with it.  This is not needed for the panic case you already handle anyway.  Would it not be clearer to return nil here?  (i.e. return nothing as nil return is implicit)

iroh-relay/src/quic.rs
```diff
@@ -0,0 +1,325 @@
...
+        let res = match external_addresses.wait_for(|addr| addr.is_some()).await {
+            Ok(res) => res,
+            Err(err) => {
+                // attempt to gracefully close the connections
+                conn.close(QUIC_ADDR_DISC_CLOSE_CODE, QUIC_ADDR_DISC_CLOSE_REASON);
+                return Err(err.into());
+            }
+        };
+        let mut observed_addr = res.expect("checked");
+        // if we've sent an to an ipv4 address, but
```
flub, line 268:
```suggestion
        // if we've sent to an ipv4 address, but
```

iroh-relay/src/server.rs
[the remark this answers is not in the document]
flub: Ah yes, makes sense.  Thanks

iroh-relay/src/server.rs
```diff
@@ -438,19 +478,50 @@ impl Server {
...
-        (Some(relay), _) => {
+    let res = match (
+        relay_http_server.as_mut(),
+        quic_server.as_mut(),
+        tasks.len(),
+    ) {
+        (None, None, 0) => Ok(Err(anyhow!("Nothing to supervise"))),
+        (None, None, _) => tasks.join_next().await.expect("checked"),
+        (None, Some(quic_ep), 0) => quic_ep.task_handle().await.map(anyhow::Ok),
+        (None, Some(quic_ep), _) => {
```
flub, line 491: This match explosion here is a bit unpleasant I guess.  I had to try this a few times but I think this might work to remove the match and do just a single `select!` statement for this entire supervisor:

```rust
    let quic_enabled = quic_server.is_some();
    let mut quic_fut = match quic_server {
        Some(ref mut server) => futures_util::future::Either::Left(server.task_handle()),
        None => futures_util::future::Either::Right(futures_lite::future::pending()),
    };
    tokio::select! {
        biased;
        ret = tasks.join_next(), if tasks.len() > 0 => todo!(),
        ret = &mut quic_fut, if quic_enabled => todo!(),
    }
```

The explanation:

- You can disable branches in the select using the `if`.
- A disabled branch won't poll the future from that branch.
- But even if disabled all futures are constructed so need to be valid.
- So you need a future that exists regardless of whether a server is present.
- Because of borrowing rules you need to do the `.is_some()` stuff up front.

flub (approved):
Looks great!  Don't think there's anything blocking but still a few suggestions that are hopefully useful.

iroh-net/src/defaults.rs
```diff
@@ -1,14 +1,19 @@
...
-use url::Url;
-
-use crate::{RelayMap, RelayNode};
-
+use iroh_base::relay_map::QuicConfig;
+/// The default QUIC port used by the Relay server to accept QUIC connections
+/// for QUIC address discovery
+///
+/// The port is "QUIC" typed on a phone keypad.
+pub use iroh_base::relay_map::DEFAULT_QUIC_PORT as DEFAULT_RELAY_QUIC_PORT;
```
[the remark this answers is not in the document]
ramfox: 😂 dig asked for the rename, since it would no longer have the `relay_map` prefix

we can also...not have this in the defaults & just import it from iroh_base where needed.

iroh-relay/src/client.rs
```diff
@@ -377,6 +377,24 @@ impl ClientBuilder {
     }
 }
 
+#[cfg(any(test, feature = "test-utils", feature = "dangerous-certs"))]
+/// Creates a client config that trusts any servers without verifying their TLS certificate.
+///
+/// Should be used for testing local relay setups only.
+pub fn make_dangerous_client_config() -> rustls::ClientConfig {
```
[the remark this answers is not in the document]
ramfox: This was `pub` for the `doctor report` command for the client part in the next PR. It's where the `dangerous_certs` feature originally came from. I'm going to keep it private for now, and see what is actually needed, when the next PR settles down.

iroh-relay/Cargo.toml
```diff
@@ -129,11 +131,12 @@ server = [
 ]
 metrics = ["iroh-metrics/metrics", "server"]
 test-utils = []
+dangerous-certs = []
 
 [[bin]]
 name = "iroh-relay"
 path = "src/main.rs"
-required-features = ["server"]
+required-features = ["server", "dangerous-certs"]
```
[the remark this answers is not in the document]
ramfox: removed

iroh-relay/src/client.rs
```diff
@@ -377,6 +377,24 @@ impl ClientBuilder {
     }
 }
 
+#[cfg(any(test, feature = "test-utils", feature = "dangerous-certs"))]
```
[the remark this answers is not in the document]
ramfox: only used in tests now, so I've removed the other features. I'll remember to add it if we end up needing the `dangerous-certs` feature in the next PR

iroh-relay/src/main.rs
```diff
@@ -115,13 +124,20 @@ struct Config {
     /// `https_bind_addr` of the `tls` configuration section and only the captive portal
     /// will be served from the HTTP socket.
     http_bind_addr: Option<SocketAddr>,
+    /// When `true`, will only bind the relay to http.
+    /// Default is `false`.
+    ///
+    /// When the `--dev` flag is used on the CLI, it will set `http_only` to `true`.
+    #[serde(default = "cfg_defaults::http_only")]
+    http_only: bool,
```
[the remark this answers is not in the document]
ramfox: did both. moved it TlsConfig, called it `dangerous_http_only`, and added comments explaining.

iroh-relay/src/main.rs
```diff
@@ -132,6 +148,13 @@ struct Config {
     ///
     /// Defaults to using the `http_bind_addr` with the port set to [`DEFAULT_STUN_PORT`].
     stun_bind_addr: Option<SocketAddr>,
+    /// Whether to allow QUIC connections for QUIC address discovery
+    ///
+    /// If no `tls` is set, this will error.
+    ///
+    /// Defaults to `true`
```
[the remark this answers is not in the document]
ramfox: Yea, that's why I thinking QUIC should be `false` on default. If someone wants to use QUIC they need to add a config file anyway, so enabling QUIC via the config at that point is easy.

iroh-relay/src/main.rs
```diff
@@ -366,45 +418,74 @@ async fn main() -> Result<()> {
     relay.shutdown().await
 }
 
+async fn maybe_load_tls(
+    cfg: &Config,
+) -> Result<Option<relay::TlsConfig<std::io::Error, std::io::Error>>> {
+    let Some(ref tls) = cfg.tls else {
+        println!("no tls, returning none");
```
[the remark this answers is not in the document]
ramfox: oops! should not have made it past manual testing...😅

iroh-relay/src/main.rs
```diff
@@ -513,6 +597,8 @@ mod tests {
     #[tokio::test]
     async fn test_rate_limit_config() -> TestResult {
         let config = "
+            enable_quic_addr_discovery = false
```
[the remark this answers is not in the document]
ramfox: no this was good, it made me realize that we should handle enabling quic and the tls config differently.

iroh-relay/src/quic.rs
ramfox: tru

iroh-relay/src/quic.rs
ramfox: good point. this was confusing; I had a hard time figuring out where the errors were handled. Switched it. Took out the debugs here and log the error in the same place we check for panics.

iroh-net/src/defaults.rs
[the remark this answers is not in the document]
ramfox: okay actually, I just renamed them all to DEFAULT_RELAY_QUIC_PORT. This isn't like STUN that has a default port, this our personal QUIC port, so it feels like addition of `RELAY` is justified.

iroh-relay/src/server.rs
ramfox: damn, nice.

PR #2966, opened 2024-11-26
Not even WIP - protocol and protocolhandler split

rklaehn:
## Description

I find it quite annoying that the natural form of ProtocolHandlers is Arced. Very frequently there are good reasons to have a split between an outer type and an inner type. Protocol structs are no exception.

So you will often have a situation where the protocol *itself* will consist of an Arc<ProtocolInner> and be cheaply cloneable, but also you need to wrap it in an Arc to make it work with the router accept. The alternative is having all methods in the protocol handler, even those that are unrelated to the protocol handling, take a `self: Arc<Self>`. The latter is super ugly and also requires lots of .clone()

So the idea here is:

a protocol *has* a protocol handler instead of *being* a protocol handler. RouterBuilder accept just takes a `&impl Protocol`, and calls a fn to get the handler. For the reverse direction, there is a fn on Protocol to get back from an Arc<dyn ProtocolHandler> to the *outer* type by downcasting to the inner type and then newtype wrapping.

```rust
pub fn accept(mut self, alpn: impl AsRef<[u8]>, protocol: &impl Protocol) -> Self
```

The result is that you don't have all these Arc<Self> in the public api of the protocol. It is just a struct that has an Arc<Inner> and is therefore cheaply cloneable if desired.

```rust
struct Ping(Arc<PingInner>);

impl Protocol for Ping { ...
```

The downside is that you have 2 concepts instead of one (Protocol and ProtocolHandler), and that you need to impl Protocol to do the downcast stuff (you could get rid of the latter somehow though).

```rust
pub fn get_protocol<P: Protocol>(&self, alpn: &[u8]) -> Option<P> // no arc!
```

Two more points to motivate why this ^ is good despite being a bit more complex.

1. Sometimes it is useful to have a Foo/FooInner split via a Box<dyn ...> or Arc<dyn ...> for *within* the protocol handler, e.g. to get rid of type parameters, e.g. go from Blobs<S> to just Blobs. By letting a protocol *have* a protocol handler instead of being a protocol handler, you allow handing out this inner arc instead of having a double dyn.

2. I think ergonomics and flexibility favours letting Protocols be cheaply cloneable newtypes around Arc<ProtocolInner> instead of wrapping in Arc by default. See https://github.com/n0-computer/iroh.computer/pull/232/files

---

As a second thought: I am not sure how useful the ability to get a concrete protocol struct by alpn and type is at all. In the situations where I have had protocol dependencies (docs building on gossip and blobs, most notably), you can look up the default blobs and gossip impl under the default ALPN, but I think there are many reasons for why you might want to have blobs and/or gossip for docs under a non-standard ALPN. And what then? Pass in the ALPNs to use? In that case you might as well pass in the protocol structs themselves.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/examples/ping.rs
+++ b/iroh/examples/ping.rs
@@ -1,7 +1,7 @@
 use std::{str::FromStr, sync::Arc};
 
 use futures_lite::future;
-use iroh_net::{endpoint, Endpoint, NodeId};
+use iroh_net::{Endpoint, NodeId};
 use iroh_router::{Protocol, ProtocolHandler};
 
 #[derive(Debug)]
@@ -33,7 +33,10 @@ impl Ping {
 }
 
 impl ProtocolHandler for PingInner {
-    fn accept(self: Arc<Self>, conn: iroh_net::endpoint::Connecting) -> future::Boxed<anyhow::Result<()>> {
+    fn accept(
+        self: Arc<Self>,
+        conn: iroh_net::endpoint::Connecting,
+    ) -> future::Boxed<anyhow::Result<()>> {
         Box::pin(async move {
             let conn = conn.await?;
             let (mut send, mut recv) = conn.accept_bi().await?;
@@ -52,10 +55,6 @@ impl ProtocolHandler for PingInner {
 impl Protocol for Ping {
     fn protocol_handler(&self) -> Arc<dyn iroh_router::ProtocolHandler> {
         self.0.clone()
-    }
-
-    fn from_protocol_handler(handler: Arc<dyn iroh_router::ProtocolHandler>) -> Option<Self> {
-        Self::downcast_via::<PingInner>(handler)
     }
 }
 
@@ -79,7 +78,6 @@ async fn main() -> anyhow::Result<()> {
         let ping = Ping::new(endpoint);
         let builder = builder.accept(ALPN, &ping);
         let router = builder.spawn().await?;
-        let t = router.get_protocol::<Ping>(&ALPN).unwrap();
         println!("Listening for pings on {}", router.endpoint().node_id());
         tokio::signal::ctrl_c().await?;
         router.shutdown().await?;
--- a/iroh/examples/ping.rs
+++ b/iroh/examples/ping.rs
@@ -1,0 +1,82 @@
+use std::{str::FromStr, sync::Arc};
+
+use futures_lite::future;
+use iroh_net::{endpoint, Endpoint, NodeId};
+use iroh_router::{Protocol, ProtocolHandler};
+
+#[derive(Debug)]
+struct Ping(Arc<PingInner>);
+
+const ALPN: &[u8] = b"ping";
+
+impl Ping {
+    fn new(endpoint: Endpoint) -> Self {
+        Self(Arc::new(PingInner { endpoint }))
+    }
+
+    async fn ping(&self, target: NodeId) -> anyhow::Result<()> {
+        let conn = self.0.endpoint.connect(target, ALPN).await?;
+        let (mut send, mut recv) = conn.open_bi().await?;
+        let request = b"Ping".to_vec();
+        send.write_all(&request).await?;
+        send.finish()?;
+        let response = recv.read_to_end(1024).await?;
+        println!("got response: {:?}", String::from_utf8_lossy(&response));
+        Ok(())
+    }
+}
+
+impl ProtocolHandler for PingInner {
+    fn accept(self: Arc<Self>, conn: iroh_net::endpoint::Connecting) -> future::Boxed<anyhow::Result<()>> {
+        Box::pin(async move {
+            let conn = conn.await?;
+            let (mut send, mut recv) = conn.accept_bi().await?;
+            let request = recv.read_to_end(1024).await?;
+            println!("got request: {:?}", String::from_utf8_lossy(&request));
+            let response = b"Pong".to_vec();
+            send.write_all(&response).await?;
+            send.finish()?;
+            conn.closed().await;
+            Ok(())
+        })
+    }
+}
+
+// todo: figure out a way to derive this?
+impl Protocol for Ping {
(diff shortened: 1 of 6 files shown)


iroh/examples/ping.rs
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// todo: figure out a way to derive this?
+impl Protocol for Ping {
+    fn protocol_handler(&self) -> Arc<dyn iroh_router::ProtocolHandler> {
+        self.0.clone()
+    }
+
+    fn from_protocol_handler(handler: Arc<dyn iroh_router::ProtocolHandler>) -> Option<Self> {
+        Some(Self(handler.into_arc_any().downcast::<PingInner>().ok()?))
```
dignifiedquire, line 52: yeah..having users write this is not great, but you could auto implement this I think

rklaehn: You need to know the intermediate type. If you have that you can do it, but it would be quite some machinery.

```rust
trait ProtocolIfIKnowTheIntermediateType {
  type Inner;
}

impl<T: ProtocolIfIKnowTheIntermediateType> Protocol for T {
  // basically the thing from above
}
```

then

```rust
impl ProtocolIfIKnowTheIntermediateType for Ping {
  type Inner = PingInner
}
```

or a derive macro, which feels a bit like overkill for this one LOC.

rklaehn:
The reason for this ^.

1. Here is one way to get rid of the `S` type parameter in Blobs<S>. You make a trait for all the things you want to do with a Blobs<S> that does not involve knowing the type of S. And then you have something like struct Blobs(Box<dyn BlobsInner>) where Blobs<S> implements BlobsInner. So now I got a double indirection, one for the Arc<dyn ProtocolHandler> thing and one to get rid of the type parameter.

2. Sometimes it is really convenient if the thing you define an async fn on is Clone so you don't end up in lifetime hell. But with the protocol handlers our idea is that the handler is not Clone, but we by default wrap it in an Arc. So now you end up with an Arc<dyn Arc<dyn<...>>

iroh/examples/ping.rs
rklaehn: Ok, how about this to simplify it a bit:

```rust
pub trait Protocol: Sized {
    fn protocol_handler(&self) -> Arc<dyn ProtocolHandler>;
    fn from_protocol_handler(handler: Arc<dyn ProtocolHandler>) -> Option<Self>;
    fn downcast_via<T: Any + Send + Sync>(handler: Arc<dyn ProtocolHandler>) -> Option<Self>
    where
        Self: From<Arc<T>>
    {
        Some(Self::from(handler.into_arc_any().downcast::<T>().ok()?))
    }
}
```

and then:
```rust
impl Protocol for Ping {
    fn protocol_handler(&self) -> Arc<dyn iroh_router::ProtocolHandler> {
        self.0.clone()
    }

    fn from_protocol_handler(handler: Arc<dyn iroh_router::ProtocolHandler>) -> Option<Self> {
        Self::downcast_via::<PingInner>(handler)
    }
}
```

matheus23:
> I find it quite annoying that the natural form of ProtocolHandlers is Arced. Very frequently there are good reasons to have a split between an outer type and an inner type. Protocol structs are no exception.

I agree it's annoying, but I'm not sure if the complexity of introducing another trait is worth it...

The way I see it there's no way around the `Arc` as long as we use object-safe traits, right?
We *could* byte the bullet and add a type parameter to `Router` to make it accumulate a typed list of protocols instead.
This would also mean you don't add lots of pointer indirection, all of your protocol types can live next to each other in memory.
But it *is* somewhat confusing for users... We'd have to make sure that inference works well.

rklaehn:
> The way I see it there's no way around the `Arc` as long as we use object-safe traits, right?

Yeah, there has to be an Arc or Box or triomphe::Arc somewhere if you want dyn. All I am saying is that it does not have to be so in your face.

You often have a need for an Arc *inside* your protocol handler, and so now you got double indirection. Also using the thing from the outside just feels a bit weird. It's not quite as bad as futures with their `self: Pin<&mut Self>` but almost.

matheus23:
> So what do you think about the usefulness of get_protocol in general? At least for docs I found that just giving the other protocols or clients to the docs constructor or builder works just fine.

Yeah - in trying to design alternative proposals, I've found that this is a really core question!
I *did* grep for usage of `get_protocol` all over our code-bases, and so far it's used in iroh-ffi (and we [don't like its use there](https://github.com/n0-computer/iroh-ffi/pull/209#discussion_r1852241217)), and I've recently written an example where I've made use of that (but it would totally also work without doing that. It's just one more `Arc::clone`, the `Blobs` I need is in scope anyways).

So I'd say I don't think it's super useful and it's blocking some potential improvements I think. But @dignifiedquire is the one who added it I think? Maybe there's some vision that just hasn't manifested itself yet. Would be good to have your input there, dig.

iroh/examples/ping.rs
rklaehn: If we get rid of the ability to downcast the whole problem/boilerplate goes away...

dignifiedquire:
made an alternative PR: https://github.com/n0-computer/iroh/pull/3010

matheus23:
(As someone requested for review here)
I'd be in favor of the linked PR over this one. Much simpler. Getting rid of get_protocol makes things easier, then going for `Box<dyn ProtocolHandler>` internally is the logical next step :)

rklaehn:
> (As someone requested for review here) I'd be in favor of the linked PR over this one. Much simpler. Getting rid of get_protocol makes things easier, then going for `Box<dyn ProtocolHandler>` internally is the logical next step :)

Yeah, the linked PR is really all I could wish for.

rklaehn:
Closing in favour of https://github.com/n0-computer/iroh/pull/3010

PR #2975, opened 2024-11-28
feat(iroh): make all important iroh_base types available

dignifiedquire:
## Description

Otherwise one has to pull in some things from `iroh-base` and some from here

## Breaking Changes

None, just additions

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -43,7 +43,7 @@ use crate::{
 mod rtt_actor;
 
 pub use bytes::Bytes;
-pub use iroh_base::node_addr::{AddrInfo, NodeAddr};
+pub use iroh_base::node_addr::{AddrInfo, AddrInfoOptions, NodeAddr};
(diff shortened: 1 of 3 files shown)


matheus23 (approved):
Well so this adds `iroh::hash` and `iroh::AddrInfoOptions` (also available under `endpoint`).

The only other thing in `iroh_base` that is not yet exported is `iroh_base::base32`.

rklaehn (approved):
I assume that this does not mean that things like AddrInfoOptions will remain until 1.0. Personally I think it is not that useful.

dignifiedquire:
> I assume that this does not mean that things like AddrInfoOptions

it is actively used right now in code, which is why I need to expose it, but would love to see a PR improving the API

dignifiedquire:
> will remain until 1.0.

`iroh-base` is part of hitting `1.0`, so time to clean up any APIs you don't think should stay

PR #2977, opened 2024-11-28
ref(iroh): Rename target for events

flub:
## Description

Turns out people will configure logs with `RUST_LOG=iroh=debug` and then the `events` target is completely lost.  That's bad!

This changes the events to be in the main crate target, so users don't have to separately enable them.

It also changes the separators to use the `::`.  The driver for this is compatible with the `EnvFliter` syntax, allowing you do do e.g. `RUST_LOG=iroh::events=debug` to enable just events.

The main downside is that it is harder to enable events across different crates.  But then we only use this in one crate so far, so perhaps that's not an issue yet.

Another downside is that `iroh::events` could quite easily become a module name, in which case the targets would conflict.

## Breaking Changes

- Events are emitted on different tracing targets: `iroh::events` instead of `events.net`.
- Event names have changed to use `::` as separator, previously this was `.`.

## Notes & open questions

Perhaps there is a better name than `events` that is less likely to conflict with a real module?  Maybe we could use an illegal character?  Experimenting with `iroh::-events::` as prefix does seem to work: it is accepted by the tracing and the EnvFilter is fine with it.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-relay/src/client.rs
+++ b/iroh-relay/src/client.rs
@@ -620,7 +620,7 @@ impl Actor {
         }
 
         event!(
-            target: "iroh::events::relay::connected",
+            target: "iroh::_events::relay::connected",
             Level::DEBUG,
             home = self.is_preferred,
             url = %self.url,
--- a/iroh-relay/src/client.rs
+++ b/iroh-relay/src/client.rs
@@ -620,7 +620,7 @@ impl Actor {
         }
 
         event!(
-            target: "events.net.relay.connected",
+            target: "iroh::events::relay::connected",
             Level::DEBUG,
             home = self.is_preferred,
             url = %self.url,
--- a/iroh/README.md
+++ b/iroh/README.md
@@ -20,9 +20,10 @@ The library uses [tracing](https://docs.rs/tracing) to for logging as
 well as for **structured events**.  Events are different from normal
 logging by convention:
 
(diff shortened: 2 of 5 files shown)


rklaehn (approved):
If this works for you, go for it. However, I find it a bit weird that these things look like module paths but don't actually match the modules the events are emitted in.

Given the latter, maybe having a special "module name" for the events pseudomodule would be nice to distinguish these from normal module paths.

flub:
> If this works for you, go for it. However, I find it a bit weird that these things look like module paths but don't actually match the modules the events are emitted in.

The tracing metadata also has a [file](https://docs.rs/tracing/latest/tracing/struct.Metadata.html#method.file) field which records the file.  This is however not shown by the default fmt subscriber.  Also, a file name is not 1:1 mapped to a module name, so yeah.  So far I have been trying to compromise between practicality and enough distinguishing semantics that you can process them automatically.

> Given the latter, maybe having a special "module name" for the events pseudomodule would be nice to distinguish these from normal module paths.

matheus23 (approved):
This makes sense to me.

> Perhaps there is a better name than `events` that is less likely to conflict with a real module?

I suggested `_events` in the discord. This both avoids a conflict with a potential module (we probably will never add an _events module), and also doesn't use a special character, so wouldn't confuse some tool that only expects valid module names.

PR #2978, opened 2024-11-29
Feat improve shutdown

dignifiedquire:
## Description

- `shutdown` now takes `&self`
- `shutdown` is idempotent, and will immediately return if already shutdown
- there is now a `is_shutdown` method to check if the router is already closed
- spawned tasks are cancelled through the cancel token as well, to improve shutdown speed
- 
## Breaking Changes

- `iroh::protocol::Router::shutdown` takes `&self` instead of `self`

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


iroh/src/protocol.rs
```diff
@@ -201,16 +201,27 @@ impl Router {
...
+        self.cancel_token.is_cancelled()
+    }
+
     /// Shuts down the accept loop cleanly.
     ///
     /// If some [`ProtocolHandler`] panicked in the accept loop, this will propagate
     /// that panic into the result here.
-    pub async fn shutdown(self) -> Result<()> {
+    ///
+    /// If already shutdown, it just returns `Ok`.
```
flub, line 203:
```suggestion
    /// If already shutdown, it returns `Ok`.
```
Text is almost always better off without words like "just" or "only" etc.

flub: This should probably say something about what it is waiting on.  Is it fully shut down when it returns?

iroh/src/protocol.rs
```diff
@@ -300,12 +302,28 @@ impl RouterBuilder {
                                     break;
                                 }
                             }
-                            Some(Ok(Err(inner))) => {
-                                debug!("Task errored: {inner:?}");
+                            Some(Ok(Some(()))) => {
+                                debug!("Task finished");
```
flub, line 297: What is the granularity of tasks here?  Basically I'm wondering whether this is too nosy to be on debug level and should rather be on trace level.

iroh/src/protocol.rs
```diff
@@ -300,12 +302,28 @@ impl RouterBuilder {
...
+                            }
+                            Some(Ok(None)) => {
+                                debug!("Task cancelled");
```
matheus23, line 300: I'm not sure if we should log this... I mean, it doesn't cost *much*, but I do think there's potential for needless log pollution.

iroh/src/protocol.rs
```diff
@@ -300,12 +302,28 @@ impl RouterBuilder {
...
+                    // handle incoming p2p connections.
+                    incoming = endpoint.accept() => {
+                        let Some(incoming) = incoming else {
+                            break;
+                        };
+
+                        let protocols = protocols.clone();
+                        let token = cancel_token.child_token();
+                        join_set.spawn(async move {
+                            token.run_until_cancelled(handle_connection(incoming, protocols)).await
```
flub, line 315: TIL about this function, I would have built this myself!

flub (approved):

iroh/src/protocol.rs
dignifiedquire: One task per incoming connection

iroh/src/protocol.rs
```diff
@@ -300,12 +302,28 @@ impl RouterBuilder {
...
+                        });
```
matheus23, line 316:
```suggestion
                        join_set.spawn(token.run_until_cancelled(handle_connection(incoming, protocols)));
```
I'm a sucker for pointfree style. That said, this might not compile. And it might not actually read better to others.

matheus23 (approved):
Looks good. I thought about the change to `Mutex`, but yeah I agree that's fine, probably even better.

I'm not sure where I stand w.r.t. tests. Always better to have them :) Maybe it's possible to dig the old `Node` tests up and see what can be adapted to here.

dignifiedquire:
> I'm not sure where I stand w.r.t. tests. Always better to have them :) Maybe it's possible to dig the old `Node` tests up and see what can be adapted to here.

they all need protocols, which is why I haven't done it yet

iroh/src/protocol.rs
dignifiedquire: done

matheus23:
You could copy the `Echo` protocol from the README (or the `echo.rs` example)

iroh/src/protocol.rs
dignifiedquire: doesn't compile..

dignifiedquire:
lets to the test adding in a different PR, as they don't directly test shutdown

iroh/src/protocol.rs
dignifiedquire: switched to `trace`

PR #2980, opened 2024-11-29
feat(iroh): improve shutdown interactions

dignifiedquire:
## Description

- ties the cancellation tokens between `Router` and `Endpoint` closer together
- closing `Endpoint` is now idempotent

## Breaking Changes

- added `iroh::Endpoint::close` that takes no arguments anymore, it default to using code `0` and an empty message
- added `iroh::Endpoint::is_closed`

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/examples/echo.rs
+++ b/iroh/examples/echo.rs
@@ -53,9 +53,6 @@ async fn connect_side(addr: NodeAddr) -> Result<()> {
     let response = recv.read_to_end(1000).await?;
     assert_eq!(&response, b"Hello, world!");
 
-    // Close the endpoint (and all its connections) in one:
-    endpoint.close(0u32.into(), b"bye!").await?;
-
     Ok(())
 }
 
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -26,7 +26,7 @@ use derive_more::Debug;
 use futures_lite::{Stream, StreamExt};
 use iroh_base::relay_map::RelayMap;
 use pin_project::pin_project;
-use tokio_util::sync::{CancellationToken, WaitForCancellationFuture};
+use tokio_util::sync::CancellationToken;
 use tracing::{debug, instrument, trace, warn};
 use url::Url;
 
@@ -965,27 +965,30 @@ impl Endpoint {
     /// Returns an error if closing the magic socket failed.
     /// TODO: Document error cases.
     pub async fn close(&self, error_code: VarInt, reason: &[u8]) -> Result<()> {
-        let Endpoint {
-            msock,
-            endpoint,
-            cancel_token,
-            ..
-        } = self;
-        cancel_token.cancel();
+        if self.is_closed() {
+            return Ok(());
+        }
+
+        self.cancel_token.cancel();
         tracing::debug!("Closing connections");
-        endpoint.close(error_code, reason);
-        endpoint.wait_idle().await;
+        self.endpoint.close(error_code, reason);
+        self.endpoint.wait_idle().await;
 
         tracing::debug!("Connections closed");
-
-        msock.close().await?;
+        self.msock.close().await?;
         Ok(())
+    }
+
+    /// Check if this endpoint is still alive, or already closed.
+    pub fn is_closed(&self) -> bool {
+        self.cancel_token.is_cancelled() && self.msock.is_closed()
     }
 
     // # Remaining private methods
 
-    pub(crate) fn cancelled(&self) -> WaitForCancellationFuture<'_> {
-        self.cancel_token.cancelled()
+    /// Expose the internal [`CancellationToken`] to link shutdowns.
+    pub(crate) fn cancel_token(&self) -> &CancellationToken {
+        &self.cancel_token
     }
 
     /// Return the quic mapped address for this `node_id` and possibly start discovery
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -954,6 +954,22 @@ impl Endpoint {
 
     /// Closes the QUIC endpoint and the magic socket.
     ///
+    /// This will close all open QUIC connections.
+    ///
+    /// It will then wait for all connections to actually be shutdown, and afterwards close
+    /// the magic socket.  Be aware however that the underlying UDP sockets are only closed
+    /// on [`Drop`], bearing in mind the [`Endpoint`] is only dropped once all the clones
+    /// are dropped.
+    ///
+    /// Returns an error if closing the magic socket failed.
+    /// TODO: Document error cases.
+    pub async fn close(&self) -> Result<()> {
+        self.close_with_code(1u16.into(), b"shutting down").await?;
+        Ok(())
+    }
+
+    /// Closes the QUIC endpoint and the magic socket.
+    ///
     /// This will close all open QUIC connections with the provided error_code and
     /// reason. See [`quinn::Connection`] for details on how these are interpreted.
     ///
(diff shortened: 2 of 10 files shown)


iroh/src/protocol.rs
```diff
@@ -268,7 +268,10 @@ impl RouterBuilder {
         let mut join_set = JoinSet::new();
         let endpoint = self.endpoint.clone();
         let protos = protocols.clone();
-        let cancel = CancellationToken::new();
+
+        // We use a child token of the endpoint, to ensure that this is shutdown
+        // when the endpoint is shutdown, but that we can shutdown ourselves independently.
+        let cancel = endpoint.cancel_token().child_token();
```
rklaehn, line 274: So you can shutdown a router without shutting down the endpoint. Makes sense I guess.

Nobody is stopping you from running multiple routers on the same endpoint, but that would probably be a pretty bad idea...

rklaehn (approved):
It would be nice to have some very basic tests of this.

Like:

endpoint.close();
assert!(router.is_closed());

just to make sure we don't break this.

iroh/src/endpoint.rs
```diff
@@ -965,27 +965,30 @@ impl Endpoint {
...
         tracing::debug!("Connections closed");
-
-        msock.close().await?;
+        self.msock.close().await?;
         Ok(())
     }
 
+    /// Check if this endpoint is still alive, or already closed.
+    pub fn is_closed(&self) -> bool {
+        self.cancel_token.is_cancelled() && self.msock.is_closed()
```
flub, line 984: `||`?

iroh/src/endpoint.rs
```diff
@@ -965,27 +965,30 @@ impl Endpoint {
     /// Returns an error if closing the magic socket failed.
     /// TODO: Document error cases.
     pub async fn close(&self, error_code: VarInt, reason: &[u8]) -> Result<()> {
```
flub, line 967: This is a separate topic, and ties in with @matheus23 's closing ideas.  But I think it is really un-ergonomic that we make this take any arguments.  We should provide some other method to close all connections and make this close connections with a default error code and reason if there are any left.

flub (commented):
seems reasonable

iroh/src/protocol.rs
dignifiedquire: we currently shutdown the endpoint as well from the router, but we could change that

iroh/src/endpoint.rs
```diff
@@ -952,6 +952,22 @@ impl Endpoint {
...
+    ///
+    /// It will then wait for all connections to actually be shutdown, and afterwards close
+    /// the magic socket.  Be aware however that the underlying UDP sockets are only closed
+    /// on [`Drop`], bearing in mind the [`Endpoint`] is only dropped once all the clones
+    /// are dropped.
+    ///
+    /// Returns an error if closing the magic socket failed.
+    /// TODO: Document error cases.
+    pub async fn close(&self) -> Result<()> {
+        self.close_with_code(1u16.into(), b"shutting down").await?;
```
dignifiedquire, line 967: suggestions welcome here for code and and message

iroh/src/endpoint.rs
dignifiedquire: but I am not fully happy about either

dignifiedquire:
> It would be nice to have some very basic tests of this.

added a minimal test

iroh/src/endpoint.rs
```diff
@@ -964,28 +980,31 @@ impl Endpoint {
...
     /// TODO: Document error cases.
-    pub async fn close(&self, error_code: VarInt, reason: &[u8]) -> Result<()> {
-        let Endpoint {
-            msock,
-            endpoint,
-            cancel_token,
-            ..
-        } = self;
-        cancel_token.cancel();
+    pub async fn close_with_code(&self, error_code: VarInt, reason: &[u8]) -> Result<()> {
```
flub, line 983: I think this should be `close_all_connections(&self, error_code: impl Into<VarInt>, reason: &[u8])` and not actually close all endpoint itself.  That's probably an easier separation to understand.

@matheus23 I can't find your design on this in notion?

iroh/src/endpoint.rs
flub: I would suggest `0` and `b""`.  That is the same behaviour you get when dropping the endpoint or dropping a connection.

iroh/src/endpoint.rs
flub: don't you want this to be showing as closed if the mosck died?

iroh/src/endpoint.rs
```diff
@@ -964,28 +980,31 @@ impl Endpoint {
...
+    /// Check if this endpoint is still alive, or already closed.
+    pub fn is_closed(&self) -> bool {
+        self.cancel_token.is_cancelled() && self.msock.is_closed()
+    }
+
     // # Remaining private methods
 
-    pub(crate) fn cancelled(&self) -> WaitForCancellationFuture<'_> {
-        self.cancel_token.cancelled()
+    /// Expose the internal [`CancellationToken`] to link shutdowns.
```
flub, line 1005: Is it worth leaving a TODO about the fact we don't own this type?

iroh/examples/echo.rs
```diff
@@ -53,9 +53,6 @@ async fn connect_side(addr: NodeAddr) -> Result<()> {
     let response = recv.read_to_end(1000).await?;
     assert_eq!(&response, b"Hello, world!");
 
-    // Close the endpoint (and all its connections) in one:
-    endpoint.close(0u32.into(), b"bye!").await?;
```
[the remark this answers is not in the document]
dignifiedquire: because we call `router` shutdown above, which calls this as well

iroh/src/lib.rs
```diff
@@ -173,7 +173,7 @@
 //!
 //!     // Gracefully close the connection and endpoint.
 //!     conn.close(1u8.into(), b"done");
-//!     ep.close(0u8.into(), b"ep closing").await?;
+//!     ep.close().await?;
```
flub, line 176: this is so much nicer now!

iroh/src/endpoint.rs
dignifiedquire: but it does the same as `.close` in that regard?

iroh/src/endpoint.rs
dignifiedquire: it's a private API, so I am not worried about it

dignifiedquire:
> don't you want this to be showing as closed if the mosck died?

well I use this in the `close` call, so I still want to cancel the token if this happens

iroh/src/endpoint.rs
matheus23: They're here: https://www.notion.so/number-zero/RFC-draft-Improving-connection-closing-APIs-11d5df1306fb8050b09cf49e521ca741

flub: @matheus23 thanks!
@dignifiedquire not sure I understand your question.

I think we need:

- A **sync** `fn close_all_connection(error_code, reason)` that just sends close on all the connections and doesn't wait at all.
- An `async fn close()` as you already implement that closes the endpoint gracefully.

So a recommended way of force-closing everything would be:
```rust
ep.close_all_connection(1, b"shutting down");
ep.close().await?;
```

And of course omitting `close_all_connections` would still work and be well behaved, you only gave up control over the error code and message.

dignifiedquire: Oh, that is a quite different to what I implemented.

I don't understand how this would work though, `close` would still call `magisock::close(0, b"")` so you end up calling close twice in your example above.

flub: So I guess we remove this method for now?  Until someone really can't live without?

iroh/src/endpoint.rs
flub: oh right, i missed that :D

iroh/src/endpoint.rs
dignifiedquire: okay will do

dignifiedquire: done

flub (approved):

flub:
The breaking changes section needs updating.

PR #2986, opened 2024-12-02
fix(iroh, iroh-relay)!: Bypass magicsock::Actor for datagrams from the relay

### LINKED ISSUE #2982: Bypass MagicSock Actor for datagrams received from the ActiveRelay actor
The `ActiveRelay` actor is connected to a relay server and receives the messages from it.  Any QUIC datagrams it receives are currently sent to the `magicsock::Actor`.  This actor moves these messages, with a little bit of processing, from it's inbox into a channel from which `AsyncUdpSocket::poll_recv` reads from.

Problem is that this 2nd channel is half the size of the inbox (128 vs 256) and the `magicsock::Actor` does a blocking send on it.  Effectively blocking the entire actor.

There is no need for these messages to be sent via the `magicsock::Actor`.  All the processing it does could be done by the `ActiveRelay` actor, which would only need a few more static details know at the time it is created to do this processing.  The `ActiveRelay` actor could directly send to the channel from which `magicsock::Actor::poll_recv` reads and also wake its `Waker`.

----

This was uncovered by #2951.

flub:
## Description

Datagrams received via the relay used to be sent to the
`magicsock::Actor` for processing (handling DISCO packets and adding
RecvMeta).  After processing they were put on a channel from which
`AsyncUdpSocket::poll_recv` would read them.

The major problem was that the relay actor did not handle the
backpressure well: the channel it was placing the processed datagrams
on is smaller than its inbox and it did not want to drop
packets (which is good, dropping packets at this point is not
helpful).  So the entire actor blocked, which has negative effects all
around.

This change connects a channel directly between the `ConnectedRelay`
actor which receives the packets and the `AsyncUdpSocket::poll_recv`.
The processing which was happening in the `magicsock::Actor` now
happens inside `poll_recv`.  This is fine because:

- The processing needs access to the `MagicSock` inners, like the
  `NodeMap`.  This required accessing the same mutexes that
  `AsyncUdpSocket::poll_recv` was already accessing anyway.  So this
  path is now slowed down.

- The alternative was moving the processing to the `ConnectedRelay`
  actor, which would have added potential more conention on the
  `MagicSock` mutexted-inners.

To support this there are a few additional  changes:

- The channel between the `AsyncUdpSocket::poll_recv(_relay)` and the
  `ConnectedRelay` actor has been abstracted into the
  `RelayRecvChannel`.  This takes care of the wakers, which now do not
  need to be separate fields on the `MagicSock`, reducing the
  `MagicSock` state and risk of using the wakers wrongly.

- The `ActiveRelay` actor is renamed to `ConnectedRelay`.  This
  improves readability.

- The `PacketSplitIter` is moved into the `relay_actor` module, this
  is where the corresponding `PacketizeIter` lives and makes the two
  much better logically linked.

- ReceviedMessage::ReceivedPacket::source is renamed into
  remote_node_id.

### iroh-relay

- `ReceivedMessage::ReceivedPacket::source` is renamed to `ReceivedMessage::ReceivedPacket::remote_node_id`.

## Notes & open questions

Closes #2982.

I'd be happy to separate out some more changes from this:

- PacketSplitIter move can easily be its own independent PR.

- ActiveRelay -> ConnectedRelay rename can easily be independent too.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/magicsock.rs
+++ b/iroh/src/magicsock.rs
@@ -22,7 +22,7 @@ use std::{
     net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6},
     pin::Pin,
     sync::{
-        atomic::{AtomicBool, AtomicU16, AtomicU64, Ordering},
+        atomic::{AtomicBool, AtomicU16, AtomicU64, AtomicUsize, Ordering},
         Arc, RwLock,
     },
     task::{Context, Poll, Waker},
@@ -188,6 +188,8 @@ pub(crate) struct MagicSock {
     relay_recv_channel: RelayRecvReceiver,
 
     network_send_wakers: Arc<parking_lot::Mutex<Option<Waker>>>,
+    /// Counter for ordering of [`MagicSock::poll_recv`] polling order.
+    poll_recv_counter: AtomicUsize,
 
     /// The DNS resolver to be used in this magicsock.
     dns_resolver: DnsResolver,
@@ -652,27 +654,86 @@ impl MagicSock {
         bufs: &mut [io::IoSliceMut<'_>],
         metas: &mut [quinn_udp::RecvMeta],
     ) -> Poll<io::Result<usize>> {
-        // FIXME: currently ipv4 load results in ipv6 traffic being ignored
         debug_assert_eq!(bufs.len(), metas.len(), "non matching bufs & metas");
         if self.is_closed() {
             return Poll::Pending;
         }
 
-        // order of polling is: UDPv4, UDPv6, relay
-        let (msgs, from_ipv4) = match self.pconn4.poll_recv(cx, bufs, metas)? {
-            Poll::Pending | Poll::Ready(0) => match &self.pconn6 {
-                Some(conn) => match conn.poll_recv(cx, bufs, metas)? {
-                    Poll::Pending | Poll::Ready(0) => {
-                        return self.poll_recv_relay(cx, bufs, metas);
+        // Three macros to help polling: they return if they get a result, execution
+        // continues if they were Pending and we need to poll others (or finally return
+        // Pending).
+        macro_rules! poll_ipv4 {
+            () => {
+                match self.pconn4.poll_recv(cx, bufs, metas)? {
+                    Poll::Pending | Poll::Ready(0) => {}
+                    Poll::Ready(n) => {
+                        self.process_udp_datagrams(true, &mut bufs[..n], &mut metas[..n]);
+                        return Poll::Ready(Ok(n));
                     }
-                    Poll::Ready(n) => (n, false),
-                },
-                None => {
-                    return self.poll_recv_relay(cx, bufs, metas);
                 }
-            },
-            Poll::Ready(n) => (n, true),
-        };
+            };
+        }
+        macro_rules! poll_ipv6 {
+            () => {
+                if let Some(ref pconn) = self.pconn6 {
+                    match pconn.poll_recv(cx, bufs, metas)? {
+                        Poll::Pending | Poll::Ready(0) => {}
+                        Poll::Ready(n) => {
+                            self.process_udp_datagrams(false, &mut bufs[..n], &mut metas[..n]);
+                            return Poll::Ready(Ok(n));
+                        }
+                    }
+                }
+            };
+        }
+        macro_rules! poll_relay {
+            () => {
+                match self.poll_recv_relay(cx, bufs, metas) {
+                    Poll::Pending => {}
+                    Poll::Ready(n) => return Poll::Ready(n),
+                }
+            };
+        }
+
+        let counter = self.poll_recv_counter.fetch_add(1, Ordering::Relaxed);
+        match counter % 3 {
+            0 => {
+                // order of polling: UDPv4, UDPv6, relay
+                poll_ipv4!();
+                poll_ipv6!();
+                poll_relay!();
+                Poll::Pending
+            }
+            1 => {
+                // order of polling: UDPv6, relay, UDPv4
+                poll_ipv6!();
+                poll_relay!();
+                poll_ipv4!();
+                Poll::Pending
+            }
+            _ => {
+                // order of polling: relay, UDPv4, UDPv6
+                poll_relay!();
+                poll_ipv4!();
+                poll_ipv6!();
+                Poll::Pending
+            }
+        }
+    }
+
+    /// Process datagrams received from UDP sockets.
+    ///
+    /// All the `bufs` and `metas` should have initialized packets in them.
+    ///
+    /// This fixes up the datagrams to use the correct [`QuicMappedAddr`] and extracts DISCO
+    /// packets, processing them inside the magic socket.
+    fn process_udp_datagrams(
+        &self,
+        from_ipv4: bool,
+        bufs: &mut [io::IoSliceMut<'_>],
+        metas: &mut [quinn_udp::RecvMeta],
+    ) {
+        debug_assert_eq!(bufs.len(), metas.len(), "non matching bufs & metas");
 
         // Adding the IP address we received something on results in Quinn using this
         // address on the send path to send from.  However we let Quinn use a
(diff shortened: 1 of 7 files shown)


dignifiedquire:
> So... Is this a breaking change now because iroh-relay is its own crate? I don't think we wanted iroh-relay to be a 1.0 crate? Do I need to list this as breaking changes?

If you break any public api in a crate you touch, yes it is a breaking change, independent of our 1.0 plans

iroh/src/magicsock.rs
```diff
@@ -1561,7 +1633,50 @@ enum DiscoBoxError {
...
-type RelayRecvResult = Result<(PublicKey, quinn_udp::RecvMeta, Bytes), io::Error>;
+/// Channel for [`MagicSock::poll_recv_relay`] to receive QUIC datagrams from relays.
+///
+/// This channel contains both the send and receive sides.  The receive side implements the
+/// poll interface with required wakers.
+#[derive(Debug)]
+struct RelayRecvChannel {
+    sender: mpsc::Sender<RelayReadResult>,
+    receiver: parking_lot::Mutex<mpsc::Receiver<RelayReadResult>>,
+    waker: parking_lot::Mutex<Option<Waker>>,
```
dignifiedquire, line 1644: we should just use `AtomicWaker` for this

iroh/src/magicsock.rs
```diff
@@ -1561,7 +1633,50 @@ enum DiscoBoxError {
...
 
-type RelayRecvResult = Result<(PublicKey, quinn_udp::RecvMeta, Bytes), io::Error>;
+/// Channel for [`MagicSock::poll_recv_relay`] to receive QUIC datagrams from relays.
+///
+/// This channel contains both the send and receive sides.  The receive side implements the
+/// poll interface with required wakers.
+#[derive(Debug)]
+struct RelayRecvChannel {
+    sender: mpsc::Sender<RelayReadResult>,
+    receiver: parking_lot::Mutex<mpsc::Receiver<RelayReadResult>>,
```
dignifiedquire, line 1643: could this just be `Mutex<VecDeque<RelayReadResult>>>`?

iroh/src/magicsock/relay_actor.rs
```diff
@@ -266,20 +283,20 @@ impl ActiveRelay {
...
-    active_relay: BTreeMap<RelayUrl, (mpsc::Sender<ActiveRelayMessage>, JoinHandle<()>)>,
-    msg_sender: mpsc::Sender<ActorMessage>,
+    connected_relays: BTreeMap<RelayUrl, (mpsc::Sender<ConnectedRelayMessage>, JoinHandle<()>)>,
     ping_tasks: JoinSet<(RelayUrl, bool)>,
     cancel_token: CancellationToken,
 }
 
 impl RelayActor {
-    pub(super) fn new(msock: Arc<MagicSock>, msg_sender: mpsc::Sender<ActorMessage>) -> Self {
+    pub(super) fn new(msock: Arc<MagicSock>, recv_channel: Arc<RelayRecvChannel>) -> Self {
```
dignifiedquire, line 294: do we ever use  `RelayRecvChannel` not wrapped in an `Arc` if so just internalize it and make it `Clone`?

divagant-martian: Since this has 3 fields inside, this would make us have another structure a la `Inner`. I find this clearer. Since this is a private struct I would think it's fine?

iroh/src/magicsock/relay_actor.rs
```diff
@@ -482,11 +510,10 @@ impl RelayActor {
 
         let my_relay = self.msock.my_relay();
         let ipv6_reported = self.msock.ipv6_reported.clone();
-        let url = url.clone();
-        let url1 = url.clone();
```
divagant-martian, line 513: wut was this

iroh/src/magicsock/relay_actor.rs
```diff
@@ -750,6 +788,51 @@ where
     }
 }
 
+/// Splits a single [`ReceivedMessage::ReceivedPacket`] frame into datagrams.
+///
+/// This is splits packets joined by [`PacketizeIter`] back into individual datagrams.  See
```
divagant-martian, line 793:
```suggestion
/// This splits packets joined by [`PacketizeIter`] back into individual datagrams.  See
```

iroh/src/magicsock.rs
```diff
@@ -1561,7 +1633,50 @@ enum DiscoBoxError {
...
+            waker: parking_lot::Mutex::new(None),
+        }
+    }
+
+    fn poll_recv(&self, cx: &mut Context) -> Poll<RelayReadResult> {
+        match self.receiver.lock().try_recv() {
+            Ok(item) => Poll::Ready(item),
+            Err(mpsc::error::TryRecvError::Empty) => {
+                self.waker.lock().replace(cx.waker().clone());
+                Poll::Pending
```
matheus23, line 1662: Somebody who knows more about rust syntax/semantics than I do:
Does this mean that both locks `receiver` and `waker` are held at the same time, or does this mean that `receiver` is locked, then dropped, and then `waker` is locked, then dropped?

If the latter, then there's a potential for a race. Not a biggie, just re-check `receiver` after registering the waker, or make sure to keep around the `receiver` lock, but still.

iroh/src/magicsock.rs
```diff
@@ -785,29 +787,33 @@ impl MagicSock {
...
+                            node = %node_id.fmt_short(),
+                            count = meta.len / meta.stride,
+                            len = meta.len,
+                            "recv quic packets from relay",
+                        );
+                        buf_out[..buf.len()].copy_from_slice(&buf);
+                        *meta_out = meta;
+                        num_msgs += 1;
+                        break;
+                    }
```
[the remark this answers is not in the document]
flub: The comment is already there:
> // Received a DISCO packet that was handled internally

This is augmented with the doc comment on `process_relay_read_result`.

The reason this wasn't happening before is that a different task (magicsock::Actor) was doing this processing and only putting the QUIC datagrams on the channel read here.  But now we need to filter out the DISCO and STUN datagrams here.

Do you still feel this needs more description?

iroh/src/magicsock.rs
flub: WRT to all comments on the choice of channel and waker: these did cross my mind.  There are probably also channel implementations that might give us the desired wakers behaviour out of the box (I think both flume and async_channle have stream APIs that are probably usable).

However I explicitly chose to keep these the same: I was targetting this change to the architectural change and didn't feel like also changing the implementation here.

Anyway, if folks would like to see this change at the same time I'm happy to do so.

So, onto this particular suggestion: using a `VecDeque` would make this an unbounded queue I think which is probably a bad idea.  I would expect a fixed-size channel to be more efficiently implemented than a variable-size one.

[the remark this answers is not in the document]
flub: Heh, that did cross my mind but it was so much simpler to put it in a single struct :smile:.  I was not expecting such aversion to this.

The nice thing about version is that a class of errors go away because there is always a sender.  But sure, if the sender starts calling recv and the other way round things would be... odd.  But also I didn't see that as a particular big risk.

Neither side of this need to be clonable.  I kind of like this as a newtype regardless of how we want to change the implementation.  But anyway, see my comment above on avoiding changing the implementation so far.

iroh/src/magicsock.rs
flub: My expectation is that the `receiver` is only locked while executing the `try_recv` call and unlocked immediately after, before the match is executed.  My understanding is that the guard is a temporary variable that gets dropped at the end of the expression.

Same with the `waker` lock, I expect it to be released at the end of that expression.

https://smallcultfollowing.com/babysteps/blog/2023/03/15/temporary-lifetimes/
and
https://blog.m-ou.se/super-let/
are good discussions of this topic.

**But**!  That doesn't mean I got it wrong.  If we're unsure I could re-write this to be explicit about where the guards are dropped?

iroh/src/magicsock/relay_actor.rs
flub: Am I the only one who likes external `Arc`s?  It's nice and explicit, there's nothing wrong with carrying it around: it takes a little extra space but at the same time communicates exactly what this.

I find all this internal Arcing so bothersome, always have to look at the implementation to understand what is going on.  (And you have to deal with all this `Inner` naming! :stuck_out_tongue: ).

The inner fields still need their individual locks anyway, so it's not like it simplifies anything.

iroh/src/magicsock.rs
matheus23: FWIW, if you made all of this `ConcurrentQueue` (from `concurrent-queue`), and keep the waker, then you essentially are 80% towards an implementation of `async-channel` :upside_down_face: 

But no seriously, I think this might be a good case for `ConcurrentQueue` (for an eventual follow-up PR).

iroh/src/magicsock.rs
flub: Postponed for followup.  Left in comment.

iroh/src/magicsock.rs
flub: Postponed to followup, left a todo message for it.

iroh/src/magicsock/relay_actor.rs
flub: Split the sender and receiver and now there's no external arc left. :smile:

iroh/src/magicsock.rs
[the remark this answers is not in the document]
divagant-martian: this is just a comment explaining the `None` value, which is itself clear. So yeah, some comments about the function body itself would be appreciated

iroh/src/magicsock.rs
flub: Good catch!  I believe this bug was there already in the old implementation too.  :scream:

iroh/src/magicsock.rs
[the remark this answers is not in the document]
flub: Added another comment.  If that's not what you had in mind I would appreciate a suggested comment.

iroh/src/magicsock.rs
flub: eeeh.  This commit broke things :(

matheus23 (approved):
This looks good to me :)

The added comment makes it clear to me why there's a new loop introduced - the *filtering* of QUIC datagrams from the relay moved from somewhere else to `poll_recv`, so we need to poll until we find a QUIC datagram.

iroh/src/magicsock.rs
```diff
@@ -1561,7 +1646,67 @@ enum DiscoBoxError {
...
+            Ok(item) => {
+                self.waker.lock().take();
+                Poll::Ready(Ok(item))
+            }
+            Err(mpsc::error::TryRecvError::Empty) => Poll::Pending,
+            Err(mpsc::error::TryRecvError::Disconnected) => {
+                Poll::Ready(Err(anyhow!("All RelayRecvSenders disconnected")))
+            }
+        }
+    }
```
flub, line 1708: Could folks, especially @matheus23 as he spotted this, review this again?  The previous fix for this was horrendously broken.  This I think is more correct.

matheus23: You're totally right. There was a potential for a race but it looks different than what I thought :upside_down_face: 

Let's not waste a lot of time on this and instead just fix it by getting rid of it in a follow-up that replaces this with an async-channel or concurrent queue.

matheus23: I wonder why you think the previous logic was horrendously broken. Did it fail tests?

flub: A broken test run: https://github.com/n0-computer/iroh/actions/runs/12143158923/job/33859383227?pr=2986

PR #2987, opened 2024-12-02
feat(benchmark): iroh-dns-server write benchmark

Arqu:
## Description

Looking at the code paths I think this should give a decent benchmark for publishing dns records. 
On my M1 Max I get about 3k IOPS and 25-30 MB/s written in all the test runs and sit at ~165-170 publishes per second, which is far from ideal. This holds true even without the fixes in the parent branch.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-dns-server/src/lib.rs
+++ b/iroh-dns-server/src/lib.rs
@@ -8,7 +8,7 @@ pub mod http;
 pub mod metrics;
 pub mod server;
 pub mod state;
-mod store;
+pub mod store;
 mod util;
 
 #[cfg(test)]
--- a/iroh-dns-server/src/lib.rs
+++ b/iroh-dns-server/src/lib.rs
@@ -8,8 +8,11 @@ pub mod http;
 pub mod metrics;
 pub mod server;
 pub mod state;
-pub mod store;
+mod store;
 mod util;
+
+// Re-export to be able to construct your own dns-server
+pub use store::ZoneStore;
 
 #[cfg(test)]
 mod tests {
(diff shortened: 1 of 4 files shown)


rklaehn:
Ooh, nice.

I am very confident that we can improve this by several orders of magnitude. The issue is that we do a write transaction for every single update. A write transaction in redb is always followed by a sync. So after every single write transaction we write all buffers to disk. You could put this on a 128 core monster machine and it would not get much faster.

PaulOlteanu:
haha I tried the same thing last night to learn how to use criterion https://github.com/n0-computer/iroh/compare/main...PaulOlteanu:iroh:dns-bench.

Unsurprisingly there was no benefit to parallelization for the persistent store, and the in-memory store was at least 8x faster if I remember correctly.

Arqu:
> haha I tried the same thing last night to learn how to use criterion [main...PaulOlteanu:iroh:dns-bench](https://github.com/n0-computer/iroh/compare/main...PaulOlteanu:iroh:dns-bench).
> 
> Unsurprisingly there was no benefit to parallelization for the persistent store, and the in-memory store was at least 8x faster if I remember correctly.

Sometimes I wonder how many people write the literal same code for the same reasons and just never know about it :D

Yeah redb runs on it's own thread and is detached from the tokio thread pool doing its own (blocking) execution.

iroh-dns-server/src/lib.rs
```diff
@@ -11,6 +11,9 @@ pub mod state;
 mod store;
 mod util;
 
+// Re-export to be able to construct your own dns-server
+pub use store::ZoneStore;
```
rklaehn, line 15: I think it is fine. Without this Server::spawn is pretty useless...

PR #2988, opened 2024-12-02
chore(ci): reduce sccache size

Arqu:
## Description

Im not convinced this is actually helping much on our self hosted runners. 
Also we've started bumping into disk size issues on some machines because of our recent proliferation of repos.

I didn't remove it entirely so we can observe it for a bit, but would potentially like to follow up and drop it unless we want to rework this with a large central cache server for our CI boxes.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


rklaehn (approved):
10G ought to be enough for everybody

Issue #2912, opened 2024-11-08
iroh-dns-server should have some LRU or other eviction policy for persistant data


Arqu:
The current implementation accumulates data on disk infinitely. We should have some size or time based eviction policy from disk as well.

matheus23:
https://github.com/n0-computer/iroh/pull/2997 closed this

PR #2994, opened 2024-12-03
feat(iroh-dns-server): Add evictable dns store

PaulOlteanu:
## Description

Sort of some progress towards #2912, although for an in-memory store only. I don't think these changes can really be extended to a persistent one so it might be better to come up with a strategy that works for persistent and in-memory stores instead of this.

## Notes & open questions

To be honest I'm not sure why I refactored the store to be a trait - I think it's unlikely this dns server will actually be extended by others.  
I have also not tested this change at all, and I'm also not sure how the `StoreType` actually looks in a config file (but this is definitely something that should be configurable - a user can't use the existing in-memory store without modifying the hardcoded value and recompiling the server).

I'm not even particularly advocating for the `timedmap` crate here, I think we could use any implementation of a hashmap with eviction (https://crates.io/crates/delay_map, our own, etc.).  
Like previously mentioned, I'm more interested in trying to figure out if we think it's worth adding something of this kind now and figuring out a solution for cleaning up the persistent store later, or if we want to figure out a solution entirely built on `redb` to keep the differences in the implementation small

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-dns-server/Cargo.toml
+++ b/iroh-dns-server/Cargo.toml
@@ -32,7 +32,11 @@ http = "1.0.0"
 iroh-metrics = "0.29"
 lru = "0.12.3"
 parking_lot = "0.12.1"
-pkarr = { version = "2.2.0", features = [ "async", "relay", "dht"], default-features = false }
+pkarr = { version = "2.2.0", features = [
+    "async",
+    "relay",
+    "dht",
+], default-features = false }
 rcgen = "0.13"
 redb = "2.0.0"
 regex = "1.10.3"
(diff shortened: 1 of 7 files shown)


iroh-dns-server/Cargo.toml
```diff
@@ -41,6 +45,7 @@ rustls-pemfile = { version = "2.1" }
 serde = { version = "1", features = ["derive"] }
 struct_iterable = "0.1.1"
 strum = { version = "0.26", features = ["derive"] }
+timedmap = "1.0"
```
matheus23, line 48: We already have a dependency on `ttl_cache` which has an almost identical API.

matheus23: (Just for reference, I know this doesn't matter right now)

rklaehn:
I am not fond of adding a trait. Traits cause a lot of friction and box you in to a particular kind of implementation.

If I understand it correctly the use case for this is to be able to run the in-memory store indefinitely, in part to solve performance problems with the persistent store. There are some very low hanging fruits to improve the persistent store, we should exhaust those before we give up on persistence. See https://github.com/n0-computer/iroh/pull/2995

Also, we should probably just add eviction to the persistent store.

PR #2995, opened 2024-12-03
refactor(iroh-dns-server): Move db ops into an actor and implement write batching

rklaehn:
## Description

Batch to make write go brrr....

- move all db interactions into an actor
- inside the actor, have two nested loops, inner loop keeps write txn open
- txn is closed when a number of msgs have been written or some time has elapsed, whatever comes first

## Breaking Changes

None

## Benches

This branch:

```
Benchmarking dns_server_writes/1000: Collecting 10 samples in estimated 9.4698 s
dns_server_writes/1000  time:   [171.28 ms 174.05 ms 176.58 ms]
                        thrpt:  [5.6631 Kelem/s 5.7455 Kelem/s 5.8384 Kelem/s]
                 change:
                        time:   [-6.8425% -3.3112% -0.0384%] (p = 0.10 > 0.05)
                        thrpt:  [+0.0384% +3.4246% +7.3451%]
                        No change in performance detected.
```

arqu/dns-bench:

```
Benchmarking dns_server_writes/1000: Warming up for 3.0000 s
dns_server_writes/1000  time:   [6.2530 s 6.3920 s 6.5351 s]
                        thrpt:  [153.02  elem/s 156.45  elem/s 159.92  elem/s]
                 change:
                        time:   [+3881.0% +3974.8% +4080.8%] (p = 0.00 < 0.05)
                        thrpt:  [-97.608% -97.546% -97.488%]
                        Performance has regressed.
```

## Downsides

The downside of write batching is that in case of a crash or hard program termination we lose the last MAX_BATCH_TIME seconds and MAX_BATCH_SIZE writes. I think that this is acceptable since discovery information is republished.


iroh-dns-server/src/store/signed_packets.rs
```diff
@@ -14,7 +16,121 @@ const SIGNED_PACKETS_TABLE: TableDefinition<&SignedPacketsKey, &[u8]> =
...
+            tokio::pin!(timeout);
+            loop {
+                for _ in 0..self.max_batch_size {
+                    tokio::select! {
+                        _ = self.cancel.cancelled() => {
+                            drop(tables);
+                            transaction.commit()?;
+                            return Ok(());
+                        }
+                        _ = &mut timeout => break,
```
matheus23, line 77: Wouldn't this mean you're dropping the current `transaction`, and thus not committing it?
Shouldn't you instead commit the transaction here, then break?

rklaehn: Yeah, forgot that. Also, there is a loop in a loop in a loop. It should be only 2 levels... 😊

Arqu:
Nice

iroh-dns-server/src/store/signed_packets.rs
```diff
@@ -1,20 +1,145 @@
...
+        match self.run0().await {
+            Ok(()) => {}
+            Err(e) => {
+                self.cancel.cancel();
+                tracing::error!("packet store actor failed: {:?}", e);
+            }
+        }
+    }
+
+    async fn run0(&mut self) -> anyhow::Result<()> {
```
matheus23, line 72: The `run0` reads somewhat weird to me.
How about this:
- Move the outer `loop` from `run0` into `run`
- Rename `run0` to `run_one_batch`

iroh-dns-server/src/store/signed_packets.rs
```diff
@@ -1,20 +1,145 @@
...
+    fn drop(&mut self) {
+        // cancel the actor
+        self.cancel.cancel();
+        // join the thread. This is important so that Drop implementations that
+        // are called from the actor thread can complete before we return.
+        if let Some(thread) = self.thread.take() {
+            let _ = thread.join();
+        }
+    }
+}
```
matheus23, line 36: Hm. I worry that this blocking `Drop` implementation will wreak havoc when `SignedPacketStore` is dropped from within a async fn running in tokio.
Can somebody remind me again what the risks there are? Is there no real risk to that?

iroh-dns-server/src/store/signed_packets.rs
```diff
@@ -42,73 +167,53 @@ impl SignedPacketStore {
...
+            max_batch_time: MAX_BATCH_TIME,
+        };
+        // start an io thread and donate it to the tokio runtime so we can do blocking IO
+        // inside the thread despite being in a tokio runtime
+        let handle = tokio::runtime::Handle::try_current()?;
+        let thread = std::thread::Builder::new()
+            .name("packet-store-actor".into())
+            .spawn(move || {
+                handle.block_on(actor.run());
+            })?;
```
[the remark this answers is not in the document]
rklaehn: I think it will. The docs for Handle::block_on say:

"Runs a future to completion on this Handle’s associated Runtime.

This runs the given future on the *current thread*, blocking until it is complete, and yielding its resolved result. Any tasks or timers which the future spawns internally will be executed on the runtime."

Also, as far as I can see the `F` type parameter in block_on does not require Send, so tokio isn't even allowed to move the actual execution to one of the main tokio thread pool threads.

The reason I like this more than creating a separate current_thread or local runtime is that if you use spawn_blocking or spawn inside the fn, this is perfectly fine and will use the normal blocking or non-blocking thread pool of the main runtime.

iroh-dns-server/src/store/signed_packets.rs
rklaehn: I often have this pattern. It is basically just to make sure that every unexpected error shuts down the entire loop with a nice error logging, while still being able to use just ? syntax.

Not quite sure why there isn't a try!{ } macro that allows you to do a block of code where you can use ? syntax inside... That would be nicer.

iroh-dns-server/src/store/signed_packets.rs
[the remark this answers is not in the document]
rklaehn: Here is a rust playground: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d35c78b6fa8dddd7d314067354d2de40

You can move an `Rc<...>` into the block_on, so it must be polled on the same thread.

iroh-dns-server/src/store/signed_packets.rs
rklaehn: Ah yes, we are basically performing blocking io on the tokio thread, if that happens. But I think blocking the tokio runtime for some time in an emergency is still vastly preferable to not running redb::Database::drop and leaving the db files in an inconsistent state that requires a costly rebuild on restart.

We can add an async shutdown function if we want to give people the option to cleanly shut down the store, but I wanted to keep this minimal.

matheus23: If the worst thing that happens is that we're blocking the tokio runtime for a little bit (and it's just slower during that time), then this is totally fine for me.

PR #2996, opened 2024-12-03
fix(iroh): Poll all AsyncUdpSocket sources fairly

### LINKED ISSUE #2981: AsyncUdpSocket::poll_recv fairness
`AsyncUdpSocket::poll_recv` multiplexes receiving datagrams from 3 different sources:

- IPv4 UDP socket
- IPv6 UDP socket
- Relay channel

Currently they are polled in this specific order: first IPv4, only if that returns `Pending` poll IPv6 and only if also that returns `Pending` does it poll the relay channel.  If one of the polls returns `Ready` the remaining ones are not polled.

This unfairly penalises the relay channel.

----

A possible solution is for the MagicSock to have buffers for datagrams that are ready.  Whenever there is something in the buffers return it as `Poll::Ready` without polling any of the sources.  If the buffers are empty always poll **all** the sources and store any `Ready` results that are not immediately returned in a buffer.

----

This is one of the issues found from #2951.

flub:
## Description

The existing AsyncUdpSocket::poll_recv polled the UDP IPv4 socket
first, only if that has no datagrams would it poll UDP IPv6 and only
also that had no datagrams to deliver was the relay socket polled.

This highly favoured IPv4 relay traffic, potentially starving the
relay traffic.

This change makes it prefer a different source on each poll.  Meaning
if multiple sockets have data to deliver a fairer delivery happens.

Closes #2981.

## Notes & open questions

- Using Ordering::Relaxed is probably fine.  I believe normally
  this polling would only happen from a single task (the Quinn
  endpoint driver) and even if it was polled concurrently from
  different threads it would not be that bad and still an improvement
  on the current order.
  
- I'm not particularly fond of the implementation, but the macros is
  the best I could come up with.  Maybe there's something cleverer
  which can be done.  

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/magicsock.rs
+++ b/iroh/src/magicsock.rs
@@ -186,6 +186,8 @@ pub(crate) struct MagicSock {
     /// Stores wakers, to be called when relay_recv_ch receives new data.
     network_recv_wakers: parking_lot::Mutex<Option<Waker>>,
     network_send_wakers: Arc<parking_lot::Mutex<Option<Waker>>>,
+    /// Counter for ordering of [`MagicSock::poll_recv`] polling order.
+    poll_recv_counter: AtomicU64,
 
     /// The DNS resolver to be used in this magicsock.
     dns_resolver: DnsResolver,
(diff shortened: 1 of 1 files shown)


iroh/src/magicsock.rs
```diff
@@ -650,27 +652,86 @@ impl MagicSock {
...
+        macro_rules! poll_relay {
+            () => {
+                match self.poll_recv_relay(cx, bufs, metas) {
+                    Poll::Pending => {}
+                    Poll::Ready(n) => return Poll::Ready(n),
+                }
+            };
+        }
+
+        let counter = self.poll_recv_counter.fetch_add(1, Ordering::Relaxed);
```
dignifiedquire, line 696: is this an overflowing add? otherwise this could become an issue for long running sessions

matheus23: Yep, atomics wrap on overflow: https://doc.rust-lang.org/std/sync/atomic/struct.AtomicU64.html#method.fetch_add

When I talked to @flub about this the last time, I wanted to go even a step further and be intentional about this: Make this an `AtomicU8`.

iroh/src/magicsock.rs
```diff
@@ -650,27 +652,86 @@ impl MagicSock {
...
+                poll_ipv4!();
+                poll_ipv6!();
+                Poll::Pending
+            }
+        }
+    }
+
+    /// Process datagrams received from UDP sockets.
+    ///
+    /// All the `bufs` and `metas` should have initialised packets in them.
```
matheus23, line 724:
```suggestion
    /// All the `bufs` and `metas` should have initialized packets in them.
```
Freedom spelling

iroh/src/magicsock.rs
```diff
@@ -186,6 +186,8 @@ pub(crate) struct MagicSock {
     /// Stores wakers, to be called when relay_recv_ch receives new data.
     network_recv_wakers: parking_lot::Mutex<Option<Waker>>,
     network_send_wakers: Arc<parking_lot::Mutex<Option<Waker>>>,
+    /// Counter for ordering of [`MagicSock::poll_recv`] polling order.
+    poll_recv_counter: AtomicU64,
```
matheus23, line 190:
```suggestion
    /// Counter for ordering of [`MagicSock::poll_recv`] polling order.
    poll_recv_counter: AtomicU8,
```

iroh/src/magicsock.rs
flub: :sob: 

but... but... it's not even public docs!

iroh/src/magicsock.rs
flub: So should it be U8?  Can we expect U8 to be more performant?

Choosing U8 means you get more bias though, since 16 isn't devisable by 3?

iroh/src/magicsock.rs
```diff
@@ -692,77 +753,83 @@ impl MagicSock {
...
-                if packet_is_quic {
-                    quic_packets_count += 1;
-                    is_quic = true;
-                } else {
-                    // overwrite the first byte of the packets with zero.
-                    // this makes quinn reliably and quickly ignore the packet as long as
-                    // [`quinn::EndpointConfig::grease_quic_bit`] is set to `false`
-                    // (which we always do in Endpoint::bind).
-                    packet[0] = 0u8;
-                }
```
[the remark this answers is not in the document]
flub: I guess this is subjective.  I think the way I wrote it has less cognitive overhead since you save a variable which you have to store on your mental stack of things to keep track off and link back to the source.  There is also something about assignment to a variable from 3 different places with like a dozen lines of code between them that I find hard to visually keep track off to map back to the variable.

And since the thing you're doing for each branch is only a single line I prefer this version and don't find the extra cognitive load of a variable worthwhile  :shrug:

iroh/src/magicsock.rs
dignifiedquire: if you want optimial performance, use `usize`, that is usually going to be most efficient

matheus23: > Choosing U8 means you get more bias though, since 16 isn't devisable by 3?

Yeah seems like you'll slightly bias towards 0: 255 % 3 = 0 and 0 % 3 = 0, too. All in all a slightly less than 1% bias though, I don't think it should matter for this case, right?

(If updated to `AtomicUsize` then the bias is much much smaller)

flub: Switched to AtomicUsize

iroh/src/magicsock.rs
flub: switched to AtomicUsize

PR #2997, opened 2024-12-03
feat(iroh-dns-server)!: eviction of stale zonestore entries

rklaehn:
## Description

Configurable eviction of stale zonestore entries.

This works by taking a snapshot of the database in regular intervals and checking if a record is possibly expired. For all possibly expired records a fire and forget message will be sent to the io write actor to check again. The actor will check the expiry again (entry could have been renewed since the last snapshot) and then deletes it if the final check confirms expiry.

Between expiry checks there is a configurable delay, so the thread is not constantly spinning checking for expiry.

We use a second thread since we don't want to block writing new entries by sifting through old entries.

## Breaking Changes

- struct config::Config has a new field zone_store
- struct metrics::Metrics has a new field store_packets_expired

## Notes & open questions

Note: there are two ways to do eviction. One is to carefully keep track of the time for each entry by having a second table that has (timestamp, key) as the key and () as the value. Then you could just evict without doing a full scan by sorting by time ascending.

The downside of course is that you need an entire new table, and you need to update this table every time you update an entry (delete (old time, id), insert (new time, id).

~~So I decided to just do a full scan instead for simplicity. We can change it if it becomes a problem.~~

~~Hm, maybe we should avoid the full scan after all. I can imagine the thing being a bit less responsive than usual while the scan is ongoing. Another idea would be to have an "event log" where you just store (time, id) -> () and then use that to look for eviction *candidates*. Don't bother cleaning up this event log on every update.~~

I have now implemented a second table. It is a multimap table from timestamp to id. This gets updated on every write (slight perf downside here), and can be used to scan for evictions without having to do a full scan.

There is quite a lot of code just to expose these config options. We could also omit this and just use reasonable defaults.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


iroh-dns-server/src/store/signed_packets.rs
```diff
@@ -110,40 +163,79 @@ impl Actor {
...
 pub(super) struct Tables<'a> {
     pub signed_packets: redb::Table<'a, &'static SignedPacketsKey, &'static [u8]>,
+    pub update_time: redb::MultimapTable<'a, [u8; 8], SignedPacketsKey>,
 }
 
 impl<'txn> Tables<'txn> {
     pub fn new(tx: &'txn redb::WriteTransaction) -> result::Result<Self, redb::TableError> {
         Ok(Self {
             signed_packets: tx.open_table(SIGNED_PACKETS_TABLE)?,
+            update_time: tx.open_multimap_table(UPDATE_TIME_TABLE)?,
```
matheus23, line 216: So, `redb` docs tell me this will create the table, if it doesn't exist already.
No need to wipe the DB & recreate, right? The upgrade should be seamless, right?

iroh-dns-server/src/store/signed_packets.rs
```diff
@@ -110,40 +163,79 @@ impl Actor {
...
                                 }
                                 res.send(updated).ok();
                             }
+                            Message::Snapshot { res } => {
+                                trace!("snapshot");
+                                res.send(Snapshot::new(&self.db)?).ok();
+                            }
+                            Message::CheckExpired { key, time } => {
+                                trace!("check expired {} at {}", key, u64::from_be_bytes(time));
+                                tables.update_time.remove_all(&time)?;
```
matheus23, line 186: ... Should we make this `remove_all` conditional on `time < expired`? 
Also, should we be more precise about which `update_time` record we're removing?

I worry that we're nuking *more* records here than necessary, and as a result, the eviction thread doesn't ask the actor to clean up some keys that were scheduled to be evicted here.

Perhaps ideally this call becomes a `tables.update_time.remove(&time, &key)` call, and it moves into the `if packet.timestamp() < expired` condition.

matheus23: Following up on that even more: Maybe we should abstract out two functions `insert_packet` and `remove_packet` and primarily use those, so we avoid ever forgetting handling both (potentially breaking invariants).

Arqu (approved):
LGTM

Arqu:
Aside from the messed up git history I guess :D

matheus23 (commented):
> There is quite a lot of code just to expose these config options. We could also omit this and just use reasonable defaults.

I think the options are neat, but I'll let @Arqu comment more on these preferences.

Some slight worry about keeping invariants, other than that, I think this looks good :)

iroh-dns-server/src/store/signed_packets.rs
rklaehn: yes, it is only an issue if you change the schema.

iroh-dns-server/src/store/signed_packets.rs
matheus23: > this would only happen for packets that have the exact same time

Yes, that was my thinking, but I don't think this is entirely impossible, and could even happen regularly. It depends on how much traffic there is, the particular pattern that the packets are ingested (bursts?) and potential OS-dependent timer precision.

iroh-dns-server/src/store/signed_packets.rs
```diff
@@ -86,22 +133,28 @@ impl Actor {
...
-                                if let Some(existing) = get_packet(&tables.signed_packets, &key)? {
+                                trace!("upsert {}", key);
+                                let replaced = if let Some(existing) = get_packet(&tables.signed_packets, &key)? {
                                     if existing.more_recent_than(&packet) {
                                         res.send(false).ok();
                                         continue;
                                     } else {
-                                        replaced = true;
+                                        // remove the packet from the update time index
+                                        tables.update_time.remove(&packet.timestamp().to_be_bytes(), key.as_bytes())?;
```
PaulOlteanu, line 149: should this be `tables.update_time.remove(&existing.timestamp().to_be_bytes(), key.as_bytes())?;` (`existing` instead of `packet`)?

PR #2998, opened 2024-12-03
feat: Implement `RelayDatagramsQueue`

matheus23:
Based on #2986 

## Description

Replaces `RelayRecvReceiver` and `RelayRecvSender` with the (clonable & shared) `RelayDatagramsQueue`.
This queue contains a `ConcurrentQueue` (from the smol library `concurrent-queue`) and an `AtomicWaker`.
It should only be polled from one task.
If polled from multiple tasks, then tasks will overwrite each other's wakers.

Unfortunately we can't make it use `&mut Self` in `poll_recv` because `quinn` expects the `AsyncUdpSocket`s `poll_recv` interface to be `&self`.

This (un)fortunately doesn't have an effect on performance for me.

(The benchmark is completely broken half the time for some reason, but when it runs it produces normal numbers:)
```sh
$ DEV_RELAY_ONLY=true cargo run -p iroh-net-bench --release --features=local-relay -- iroh --with-relay --download-size=100M
      │  Throughput   │ Duration 
──────┼───────────────┼──────────
 AVG  │   55.05 MiB/s │     1.82s
 P0   │   55.03 MiB/s │     1.82s
 P10  │   55.06 MiB/s │     1.82s
 P50  │   55.06 MiB/s │     1.82s
 P90  │   55.06 MiB/s │     1.82s
 P100 │   55.06 MiB/s │     1.82s
 ```
 
 And basically exactly the same times for the PR this is based on.

## Todo

- [x] Rename variables to e.g. `relay_datagrams_queue` instead of `relay_recv_sender` or `relay_recv_channel` etc.
- [x] Add documentation about multiple tasks polling, etc.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/magicsock.rs
+++ b/iroh/src/magicsock.rs
@@ -1713,6 +1713,15 @@ enum DiscoBoxError {
     Parse(anyhow::Error),
 }
 
+/// A queue holding [`RelayRecvDatagram`]s that can be polled in async
+/// contexts, and wakes up tasks when something adds items using [`try_send`].
+///
+/// This is used to transfer relay datagrams between the [`RelayActor`]
+/// and [`MagicSock`].
+///
+/// [`try_send`]: Self::try_send
+/// [`RelayActor`]: crate::magicsock::RelayActor
+/// [`MagicSock`]: crate::magicsock::MagicSock
 #[derive(Debug)]
 struct RelayDatagramsQueue {
     queue: ConcurrentQueue<RelayRecvDatagram>,
@@ -1720,6 +1729,7 @@ struct RelayDatagramsQueue {
 }
 
 impl RelayDatagramsQueue {
+    /// Creates a new, empty queue with a fixed size bound of 128 items.
     fn new() -> Self {
         Self {
             queue: ConcurrentQueue::bounded(128),
@@ -1727,6 +1737,10 @@ impl RelayDatagramsQueue {
         }
     }
 
+    /// Sends an item into this queue and wakes a potential task
+    /// that's registered its waker with a [`poll_recv`] call.
+    ///
+    /// [`poll_recv`]: Self::poll_recv
     fn try_send(
         &self,
         item: RelayRecvDatagram,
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: iroh/src/magicsock.rs [hunk 1]
```diff
@@ -1712,64 +1713,48 @@ enum DiscoBoxError {
...
-            Err(mpsc::error::TryRecvError::Disconnected) => {
-                Poll::Ready(Err(anyhow!("All RelayRecvSenders disconnected")))
+        match self.queue.pop() {
+            Ok(value) => Poll::Ready(Ok(value)),
+            Err(concurrent_queue::PopError::Empty) => {
+                self.waker.register(cx.waker());
+
+                match self.queue.pop() {
+                    Ok(value) => {
+                        self.waker.take();
```
flub, line 1747: FWIW the Flag examples in `AtomicWaker` don't bother with removing this waker if it is no longer needed.  I really don't know anymore how much this matters.  It depends on what suprious wakeups might happen or how concurrently this is called I guess.

This *should* be reasonably not-concurrently called IIUC, but then I don't understand why I had things not waking up in the broken version.

matheus23: Well, yeah, not removing the waker here means you're generating another spurious wakeup AFAIU. Not terrible, but also not great.

flub: Of course custom poll functions could mean some other poll decides to still return pending for other reasons.  In that case you'll get spurious wakeups I guess.

matheus23: Yes it's exactly that case you describe. If another poll fn decides to return pending for other reasons, then there's no need to wake up the future if there's another item in the queue. The only reason to wake up the future is for the other poll fn's reasons.

flub (approved):
I'm a bit surprised that you joined the sender & receiver back into a single object, after complaining on my PR that they were like this.  But... I don't mind.

Good naming though!

matheus23:
:joy: Sorry about that.

It's not that I want the sender & receiver to be separate necessarily.
It's just that sticking both a `tokio::mpsc::Sender` and `tokio::mpsc::Receiver` into the same object reads weird to me.

flub:
> Good naming though!

I take this part back.  This is only the `RelayRecvDatagramsQueue`.  We'll probably wan a send version too.  No big deal though.

matheus23:
> > Good naming though!
> 
> I take this part back. This is only the `RelayRecvDatagramsQueue`. We'll probably wan a send version too. No big deal though.

What if we re-use the same implementation for the send side :eyes: (and only rename the *variable* names to `relay_recv_queue`, not the struct name)?

Issue #1108, opened 2023-06-14
relay: no support for wildcard certs


Arqu:
The current implementation of derper does not support wildcard certs.
[This](https://letsencrypt.org/docs/challenge-types/) describes the possible challenge methods and which are compatible with `*.domain.tld` certificates.

dignifiedquire:
Manual certificate mode is available as workaround for now.

Arqu:
This is still active, we only do the `TLS-ALPN-01` challenge

Arqu:
Making sure the issue is up to date:
This basically boils down to it not being worth supporting the DNS challenge directly and writing whatever number of integrations with DNS providers. The solution is to write a custom cert resolver that just hot-reloads the cert every day and putting certbot in front with the right DNS config to handle the renewals.

PR #2999, opened 2024-12-03
feat(relay): reloading certificate resolver

Arqu:
## Description

This sets us up so we can have manually/externally managed certs that will reload on a daily basis. This should close https://github.com/n0-computer/iroh/issues/1108 and we should follow up on the ops side to utilize this where needed.

Ok, this turned out to be much grosser than expected. 
- I don't like the new dependencies we introduce
- I don't like the loader setup
- I'm unsure whether to keep the reloading resolver in the `iroh` repo or move it into `tokio-rustls-acme` (think it should move over, but for the sake of this discussion, included it here)
- Should I do anything with the `_handle` for the reloader in regards to shutting down?

Most of this still feels necessary. I could maybe feature flag it?
I had a really hard time trying to extract the loader creation into a standalone func which also aligns all the traits and error types so the thing compiles.

Suggestions are welcome :)

Example config:
```
enable_relay = true
http_bind_addr = "[::]:80"
enable_stun = true
stun_bind_addr = "[::]:3478"
enable_metrics = true
metrics_addr = "127.0.0.1:9090"

[tls]
manual_cert_path="certificate.der"
manual_key_path="private_key.der"
cert_mode = "Reloading"
```

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-relay/src/server/resolver.rs
+++ b/iroh-relay/src/server/resolver.rs
@@ -1,0 +1,65 @@
+use std::{sync::Arc, time::Duration};
+
+use anyhow::{anyhow, Result};
+use reloadable_state::Reloadable;
+use rustls::{
+    server::{ClientHello, ResolvesServerCert},
+    sign::CertifiedKey,
+};
+use tokio::task::JoinHandle;
+
+/// A Certificate resolver that reloads the certificate every interval
+#[derive(Debug)]
+pub struct ReloadingResolver<Loader: Send + 'static> {
+    /// The inner reloadable value.
+    reloadable: Arc<Reloadable<CertifiedKey, Loader>>,
+    /// The handle to the task that reloads the certificate.
+    _handle: JoinHandle<()>,
+}
+
+impl<Loader> ReloadingResolver<Loader>
+where
+    Loader: Send + reloadable_state::core::Loader<Value = CertifiedKey> + 'static,
+{
+    /// Perform the initial load and construct the [`ReloadableResolver`].
+    pub async fn init(loader: Loader, interval: Duration) -> Result<Self> {
+        let (reloadable, _) = Reloadable::init_load(loader)
+            .await
+            .map_err(|_| anyhow!("Failed to load the certificate"))?;
+        let reloadable = Arc::new(reloadable);
+
+        // Spwan a task to reload the certificate every interval.
+        let _reloadable = reloadable.clone();
+        let _handle = tokio::spawn(async move {
+            let mut interval = tokio::time::interval(interval);
+            loop {
+                interval.tick().await;
+                let _ = _reloadable.reload().await;
+            }
+        });
+
+        Ok(Self {
+            reloadable,
+            _handle,
+        })
+    }
+}
+
+impl<Loader> ResolvesServerCert for ReloadingResolver<Loader>
+where
+    Loader: reloadable_state::core::Loader<Value = CertifiedKey>,
+    Loader: Send,
+    Loader: std::fmt::Debug,
+{
+    fn resolve(&self, _client_hello: ClientHello) -> Option<Arc<CertifiedKey>> {
+        Some(self.reloadable.get())
+    }
+}
+
+impl<Loader: Send> std::ops::Deref for ReloadingResolver<Loader> {
+    type Target = Reloadable<CertifiedKey, Loader>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.reloadable
+    }
+}
--- a/iroh-relay/src/server/resolver.rs
+++ b/iroh-relay/src/server/resolver.rs
@@ -6,7 +6,7 @@ use rustls::{
     server::{ClientHello, ResolvesServerCert},
     sign::CertifiedKey,
 };
-use tokio::task::JoinHandle;
+use tokio_util::task::AbortOnDropHandle;
 
 /// A Certificate resolver that reloads the certificate every interval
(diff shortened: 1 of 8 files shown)


Arqu:
From @flub in discord
> i have some comments on the config structure, code (and bikeshedding 😉 ), but the principle of it doesn't seem that gross?

> IIUC this reads certs from a file at given intervals.  which seems fine
Correct
> i'm also not sure why you think this might belong in tokio-acme-rustls, it doesn't seem to have anything acme about it?
I guess what I wanted it to be is mostly it's own separate thing. It's both generally useful outside of iroh and also makes the relay codebase less cluttered. I also expect this to crop up with other services we're running, I'd rather not copy it over.
> wrt the comments, my biggest structural one is probably why this is separate from Manual as it seems at some level this is > just manual + an interval?
I agree, the `CertMode` handling should be separate, but bellow that it's the same as `Manual` anyways. 

Just moving the discussion over here to make things a bit easier.

Thanks for the input!

flub:
> > i'm also not sure why you think this might belong in tokio-acme-rustls, it doesn't seem to have anything acme about it?
> 
> I guess what I wanted it to be is mostly it's own separate thing. It's both generally useful outside of iroh and also makes the relay codebase less cluttered. I also expect this to crop up with other services we're running, I'd rather not copy it over.

This is not that much code and I think it fits just fine inside the relay server for now.  Once we have to start copy-pasting it we can re-visit that I guess.

iroh-relay/src/server/resolver.rs
```diff
@@ -0,0 +1,65 @@
...
+    /// Perform the initial load and construct the [`ReloadingResolver`].
+    pub async fn init(loader: Loader, interval: Duration) -> Result<Self> {
+        let (reloadable, _) = Reloadable::init_load(loader)
+            .await
+            .map_err(|_| anyhow!("Failed to load the certificate"))?;
+        let reloadable = Arc::new(reloadable);
+
+        // Spwan a task to reload the certificate every interval.
+        let _reloadable = reloadable.clone();
+        let _handle = tokio::spawn(async move {
```
dignifiedquire, line 33: should have a way to cancel/shutdown?

Arqu:
Looking to get back to this tonight

iroh-relay/src/server/resolver.rs
```diff
@@ -0,0 +1,90 @@
...
+
+/// A Certificate resolver that reloads the certificate every interval
+#[derive(Debug)]
+pub struct ReloadingResolver<Loader: Send + 'static> {
+    /// The inner reloadable value.
+    reloadable: Arc<Reloadable<CertifiedKey, Loader>>,
+    /// The handle to the task that reloads the certificate.
+    _handle: AbortOnDropHandle<()>,
+    /// Shutdown signal sender
+    _shutdown_tx: tokio::sync::oneshot::Sender<()>,
```
Arqu, line 19: Should I bubble this up into the relay so we call it on relay.shutdown() or is the AbortOnDropHandle enough for this usecase?

Arqu:
Do we have to bump the MSRV because redb got bumped to 2.3?

dignifiedquire:
> Do we have to bump the MSRV because redb got bumped to 2.3?

what's their new MSRV? we probably want to bump it to normalize cargo.lock to v4

Arqu:
> > Do we have to bump the MSRV because redb got bumped to 2.3?
> 
> what's their new MSRV? we probably want to bump it to normalize cargo.lock to v4

[1.81](https://github.com/cberner/redb/commit/bb4415413906bb01341c7fb852d2b0af17046d9b)

dignifiedquire:
lets bump MSRV to 1.81 then, this will make live easier for us

dignifiedquire:
but in a new PR

dignifiedquire:
https://github.com/n0-computer/iroh/issues/3031

Arqu:
Rest of the PR good now?

iroh-relay/src/server/resolver.rs
```diff
@@ -0,0 +1,90 @@
...
+            let mut interval = tokio::time::interval(interval);
+            tokio::select! {
+                _ = async {
+                    loop {
+                        interval.tick().await;
+                        let _ = _reloadable.reload().await;
+                        tracing::info!("Reloaded the certificate");
+                    }
+                } => {},
+                _ = shutdown_rx => {
```
dignifiedquire, line 48: we usually use cancellation token for this

dignifiedquire:
MSRV fix is in now

dignifiedquire (approved):

iroh-relay/src/server/resolver.rs
```diff
@@ -0,0 +1,93 @@
...
+use anyhow::{anyhow, Result};
+use reloadable_state::Reloadable;
+use rustls::{
+    server::{ClientHello, ResolvesServerCert},
+    sign::CertifiedKey,
+};
+use tokio_util::{sync::CancellationToken, task::AbortOnDropHandle};
+
+/// The default certificate reload interval.
+pub const DEFAULT_CERT_RELOAD_INTERVAL: u64 = 60 * 60 * 24;
```
flub, line 12: No units!  Maybe make it a `Duration`?  Or is that still not `const`?  Should definitely document the units though.

flub:
lol, i miss the review window again :smile:

Arqu:
I'll do a follow up, just for you :D

PR #3010, opened 2024-12-04
feat(iroh): remove `Arc` requirements from `ProtocolHandler`

dignifiedquire:
## Breaking Changes

- `iroh::protocol::ProtocolHandler::accept` now takes `&self` instead of `Arc<Self>`
- `iroh::protocol::ProtocolHandler::shutdown` now takes `&self` instead of `Arc<Self>`
- `iroh::protocol::RouterBuilder::accept` now takes `T: ProtocolHandler` instead of `Arc<dyn ProtocolHandler>`
- `iroh::protocol::ProtocolMap` is now private

### DIFF
--- a/iroh/examples/search.rs
+++ b/iroh/examples/search.rs
@@ -127,7 +127,8 @@ impl ProtocolHandler for BlobSearch {
     ///
     /// The returned future runs on a newly spawned tokio task, so it can run as long as
     /// the connection lasts.
-    fn accept(self: Arc<Self>, connecting: Connecting) -> BoxedFuture<Result<()>> {
+    fn accept(&self, connecting: Connecting) -> BoxedFuture<Result<()>> {
+        let this = self.clone();
         // We have to return a boxed future from the handler.
         Box::pin(async move {
             // Wait for the connection to be fully established.
@@ -145,7 +146,7 @@ impl ProtocolHandler for BlobSearch {
 
             // Now, we can perform the actual query on our local database.
             let query = String::from_utf8(query_bytes)?;
-            let num_matches = self.query_local(&query).await;
+            let num_matches = this.query_local(&query).await;
 
             // We want to return a list of hashes. We do the simplest thing possible, and just send
             // one hash after the other. Because the hashes have a fixed size of 32 bytes, this is
@@ -167,11 +168,11 @@ impl ProtocolHandler for BlobSearch {
 
 impl BlobSearch {
     /// Create a new protocol handler.
-    pub fn new(endpoint: Endpoint) -> Arc<Self> {
-        Arc::new(Self {
+    pub fn new(endpoint: Endpoint) -> Self {
+        Self {
             endpoint,
             blobs: Default::default(),
-        })
+        }
     }
 
     /// Query a remote node, download all matching blobs and print the results.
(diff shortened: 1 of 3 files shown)


iroh/examples/search.rs
```diff
@@ -127,7 +127,8 @@ impl ProtocolHandler for BlobSearch {
     ///
     /// The returned future runs on a newly spawned tokio task, so it can run as long as
     /// the connection lasts.
-    fn accept(self: Arc<Self>, connecting: Connecting) -> BoxedFuture<Result<()>> {
+    fn accept(&self, connecting: Connecting) -> BoxedFuture<Result<()>> {
+        let this = self.clone();
```
flub, line 131: This clone is not needed by the router at all and only helpful for the implementations themselves, right?  So this is entirely an application-level concern.  But I guess it is a common pattern?

I wonder how to document this nicely so folks aren't lost searching for this for hours.

flub (approved):
looks nice!

rklaehn (approved):
I like this!

Two things:

1. I think we might want to provide a number of simple forwarder impls for ProtocolHandler. Something like below.

2. maybe we should widen the accept even more and say `impl Into<impl ProtcolHandler>` or impl Into<BoxedProtocolHandler> or something like that. I think you might sometimes have the situation where the object that comes out of the builder is something you want to keep, so you can not move it by value into the router. E.g. Blobs. If blobs provides access to the store and the downloader and the local pool, you want to put it somewhere. But you also don't want to make it Clone, so you can reason about its lifetime. So now you can't say `router.accept(iroh_blobs::ALPN, blobs)` anymore, but have to have an indirection: `router.accept(iroh_blobs::ALPN, blobs.protocol_handler())` (so you can keep the blobs to store somewhere or e.g. give a ref to them to the docs builder...

```rust

impl<T: ProtocolHandler> ProtocolHandler for Arc<T> {
    fn accept(&self, conn: Connecting) -> BoxedFuture<Result<()>> {
        self.as_ref().accept(conn)
    }

    fn shutdown(&self) -> BoxedFuture<()> {
        self.as_ref().shutdown()
    }
}

impl<T: ProtocolHandler> ProtocolHandler for Box<T> {
    fn accept(&self, conn: Connecting) -> BoxedFuture<Result<()>> {
        self.as_ref().accept(conn)
    }

    fn shutdown(&self) -> BoxedFuture<()> {
        self.as_ref().shutdown()
    }
}
```

dignifiedquire:
>  `impl Into<impl ProtcolHandler>`

this doesn't work asfaict, because we need a concrete type at some point, as we `Box` it

matheus23 (approved):
I like this. I like the forwarder impls, too.

Change the PR title, and I'm even happier :)

rklaehn:
> > `impl Into<impl ProtcolHandler>`
> 
> this doesn't work asfaict, because we need a concrete type at some point, as we `Box` it

I don't know. I tried it and it seems to work. Nested impl traits are not allowed though, so the thing looks a bit ugly:

```rust

    /// Configures the router to accept the [`ProtocolHandler`] when receiving a connection
    /// with this `alpn`.
    pub fn accept<ALPN, IH, H>(mut self, alpn: ALPN, handler: IH) -> Self
        where
            ALPN: AsRef<[u8]>,
            IH: Into<H>,
            H: ProtocolHandler,
    {
        let handler = handler.into();
        self.protocols.insert(alpn.as_ref().to_vec(), handler);
        self
    }
    ```

PR #3016, opened 2024-12-06
chore: Don't log expected errors

matheus23:
## Description

I'm frequently seeing stuff like this in logs when running tests:
```
2024-12-06T11:33:18.547916Z ERROR iroh::magicsock: 515: failed to send udp: operation would block node=m4suaktjwmn62wiz dst=172.17.0.1:60377
2024-12-06T11:33:18.548017Z ERROR iroh::magicsock: 560: no UDP or relay paths available for node node=m4suaktjwmn62wiz
```

These two logs are totally non-fatal and actually expected.
The first one *especially* since `try_send_upd` can simply return `io::ErrorKind::WouldBlock` and that's totally normal.

So we avoid printing an error in that case.

I've also made the "no UDP or relay paths available for node" and another similar log debug level, as that's a state that we routinely encounter, is expected, and handled.

## Notes & open questions

Just quickly noting that this log was in fact *so* annoying, I kept seeing the `WouldBlock` stuff a *lot* when running `iroh-doctor` with you on a train ride back then. It would mess up the whole terminal rendering :face_with_spiral_eyes: 

## Change checklist

- [x] Self-review.
- ~~[ ] Tests if relevant.~~


iroh/src/magicsock.rs
```diff
@@ -518,8 +518,11 @@ impl MagicSock {
                             udp_sent = true;
                         }
                         Err(err) => {
-                            error!(node = %node_id.fmt_short(), dst = %addr,
-                                   "failed to send udp: {err:#}");
+                            // No need to print "WouldBlock" errors to the console
+                            if err.kind() != io::ErrorKind::WouldBlock {
+                                error!(node = %node_id.fmt_short(), dst = %addr,
```
flub, line 523: I'd even turn this down to WARN.

iroh/src/magicsock.rs
```diff
@@ -563,16 +566,16 @@ impl MagicSock {
...
-                            "no UDP or relay paths available for node",
+                            "no UDP or relay paths available for node, voiding transmit",
                         );
                     }
                     Ok(())
                 }
             }
             None => {
-                error!(%dest, "no NodeState for mapped address");
+                debug!(%dest, "no NodeState for mapped address, voiding transmit");
```
flub, line 578: How are we sending to something we have no NodeState for?  AFAIK that's genuinely bad and a bug we should fix.  We should have a QuicMappedAddr with NodeId for everything that makes it this far should we not?

iroh/src/magicsock.rs
```diff
@@ -563,16 +566,16 @@ impl MagicSock {
                         // at any time so these errors should be treated as transient and
                         // are just timeouts.  Hence we opt for returning Ok.  See
                         // test_try_send_no_udp_addr_or_relay_url to explore this further.
-                        error!(
+                        debug!(
```
flub, line 569: This one I find tricky to judge.  I'm not entirely sure if this is a benign condition.

iroh/src/magicsock.rs
matheus23: Hm, I don't mind seeing error logs for non-fatal errors from the UDP socket.

iroh/src/magicsock.rs
matheus23: I don't think I've actually hit that in practice yet. So I'm fine with making this `error` again, or even panicking or sth like that.

That said, there's a comment below this talking about how this is fine, so I thought it's fine?

iroh/src/magicsock.rs
matheus23: It happens every time I run a test.

I think the condition this happens in is for example: I create a new `Endpoint` with relays disabled.
Then I send data with the udp socket send eventually returning `pending`.

iroh/src/magicsock.rs
flub: Yeah, the socket and quic might be able to do more useful work, probably with other nodes that the buggy one.  But that still leaves it a bug.

iroh/src/magicsock.rs
matheus23: If you look at the code above:
- `udp_pending` ends up being `true`
- `relay_pending` is false, because `relay_url` is `None`
- `relay_sent` and `udp_sent` are both `false`
- `pings_sent` ends up being `false` too, maybe because there's just no pings scheduled to send.

flub: Huh, in that case is it easy to create an isolated reproducer for this?

flub: > If you look at the code above:
> 
>     * `udp_pending` ends up being `true`
> 
>     * `relay_pending` is false, because `relay_url` is `None`
> 
>     * `relay_sent` and `udp_sent` are both `false`
> 
>     * `pings_sent` ends up being `false` too, maybe because there's just no pings scheduled to send.

Nice!

So this means this **is** a bug: the socket should return `WouldBlock` in this case.  Right?

matheus23: > So this means this **is** a bug: the socket should return `WouldBlock` in this case. Right?

Not sure. There's another comment close by that says:

> Returning WouldBlock triggers a hot loop

flub: To summarise an out-of-band conversation: this can not return WouldBlock because the next datagram being sent may have a destination with a RelayUrl so if you would go into blocking here you'd be wrongly blocking.

I've been sufficiently convinced now that this really is a normal situation and that we can turn down the log level here.

Could you add a comment explaining when this happens though?  Basically some version of https://github.com/n0-computer/iroh/pull/3016#discussion_r1873229444 maybe with the bit that you can't block because you can't predict the dest of the next datagram.

iroh/src/magicsock.rs
flub: I'm still not convinced this is acceptable...

flub: By which I mean, that this is a normal condition that is not a bug.

iroh/src/magicsock.rs
flub: I think ERROR is something that you essentially can consider the `Endpoint` to become unusable.  It is reasonable to expect now that you should destroy it and create a new one.  This is clearly not an error because the socket could recover and continue.

iroh/src/magicsock.rs
matheus23: Alright. I'll revert this to `error!`.

iroh/src/magicsock.rs
matheus23: I added more comments. Lmk what you think :)

flub (approved):
Looks good!

PR #3018, opened 2024-12-06
refactor(iroh)!: make iroh::tls private

dignifiedquire:
## Description

- makes `iroh::tls` private
- match quinns tls setup for benchmarking

## Breaking Changes

- `iroh::tls` is removed from the public API

## Notes & open questions

The goal is to reduce the API surface of `iroh`, to what is really needed. I couldn't find any critical usage of this in our dependencies.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.


.github/workflows/ci.yml
```diff
@@ -189,7 +189,7 @@ jobs:
         # uses: obi1kenobi/cargo-semver-checks-action@v2
         uses: n0-computer/cargo-semver-checks-action@feat-baseline
         with:
-          package: iroh, iroh-base, iroh-dns-server, iroh-net-bench, iroh-relay, iroh-net-report
+          package: iroh, iroh-base, iroh-dns-server, iroh-bench, iroh-relay, iroh-net-report
```
rklaehn, line 192: so iroh-net-report does not get renamed to iroh-report for now?

rklaehn (approved):
The only non-bench change is making tls private. I took a quick look at the bench changes and found nothing obviously wrong.

.github/workflows/ci.yml
dignifiedquire: don't know if we want to do that, but that's seperate from this PR

PR #3020, opened 2024-12-09
Very WIP Experiment: don't require Box in protocol handler accept

rklaehn:
## Description

Add an UnboxedProtocolHandler trait and try it out in the examples.

## Breaking Changes

The blanket impls for Box and Arc conflict with this, so I had to comment them out

## Notes & open questions

Is this worth the effort in simplifying writing protocols?
Should ProtocolHandler always be like this, and we do the transformation to something boxable internally?

Maybe the following would be good:

- rename UnboxedProtocolHandler to just ProtocolHandler (or Protocol if we want to go this way)
- keep ProtocolHandler as BoxedProtocolHandler but hide it away a bit

That way by default people get the simple interface, but if somebody has special requirements such as
- they have an already boxed future and want to avoid boxing it again
- they don't want owned self because they want to clone not all of self but only a part
they can still do it by implementing BoxedProtocolHandler manually. All ProtocolHandler impls get a default BoxedProtocolHandler impl.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/examples/echo.rs
+++ b/iroh/examples/echo.rs
@@ -7,10 +7,9 @@
 //!     cargo run --example echo --features=examples
 
 use anyhow::Result;
-use futures_lite::future::Boxed as BoxedFuture;
 use iroh::{
     endpoint::Connecting,
-    protocol::{ProtocolHandler, Router},
+    protocol::{Router, UnboxedProtocolHandler},
     Endpoint, NodeAddr,
 };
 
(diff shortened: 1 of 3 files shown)


rklaehn:
Ok, let's see how our handlers are implemented:

# Gossip

```rust
impl ProtocolHandler for Gossip {
    fn accept(&self, conn: Connecting) -> BoxedFuture<Result<()>> {
        let this = self.clone();
        Box::pin(async move { this.handle_connection(conn.await?).await })
    }
}
```

# Blobs

```rust
impl<S: crate::store::Store> ProtocolHandler for Blobs<S> {
    fn accept(&self, conn: Connecting) -> BoxedFuture<Result<()>> {
        let db = self.store().clone();
        let events = self.events().clone();
        let rt = self.rt().clone();

        Box::pin(async move {
            crate::provider::handle_connection(conn.await?, db, events, rt).await;
            Ok(())
        })
    }

    fn shutdown(&self) -> BoxedFuture<()> {
        let store = self.store().clone();
        Box::pin(async move {
            store.shutdown().await;
        })
    }
}
```

# Docs

```
impl<S: iroh_blobs::store::Store> ProtocolHandler for Docs<S> {
    fn accept(&self, conn: Connecting) -> BoxedFuture<Result<()>> {
        let this = self.engine.clone();
        Box::pin(async move { this.handle_connection(conn).await })
    }

    fn shutdown(&self) -> BoxedFuture<()> {
        let this = self.engine.clone();
        Box::pin(async move {
            if let Err(err) = this.shutdown().await {
                tracing::warn!("shutdown error: {:?}", err);
            }
        })
    }
}
```

# Gossip

```rust
impl ProtocolHandler for Gossip {
    fn accept(&self, conn: Connecting) -> BoxedFuture<Result<()>> {
        let this = self.clone();
        Box::pin(async move { this.handle_connection(conn.await?).await })
    }
}
```

Clones self, but arguably it should only clone inner.

So while in all examples we clone self, in all actual impls we try to clone as little as possible to move it into the future.

matheus23:
This has gone a little stale since #3010 
But it contains another possibly handy idea which is not requiring to return `BoxedFuture`, and instead allow implementing the protocol handler using an `async fn`.
Should we close? Should we attempt to change iroh to allow non-boxed futures?

dignifiedquire:
lets close it, and make a new PR if we can make it work/see the need for it

PR #3021, opened 2024-12-09
chore: update to hickory =0.25.0-alpha.4

rklaehn:
## Description

Update hickory to 0.25.0-alpha.4 and try to get the code to compile again. Hopefully the API won't change that much from this to the next non-alpha release 0.25 https://github.com/hickory-dns/hickory-dns/issues/2206#issuecomment-2511154019 .

## Breaking Changes

None

## Notes & open questions

Note: we still trigger cargo-deny because of the swarm-discovery crate: https://github.com/rkuhn/swarm-discovery/blob/master/Cargo.toml#L20

I guess even once we have merged this we still have to add an exemption to the deny.toml

Is the usage of the new API correct, specifically regarding the new LookupControlFlow enum.

It seems from the doc that I should just take the existing code and wrap it in LookupControlFlow::Continue?

```rust

/// Result of a Lookup in the Catalog and Authority
///
/// * **All authorities should default to using LookupControlFlow::Continue to wrap their responses.**
///   These responses may be passed to other authorities for analysis or requery purposes.
/// * Authorities may use LookupControlFlow::Break to indicate the response must be returned
///   immediately to the client, without consulting any other authorities.  For example, if the
///   the user configures a blocklist authority, it would not be appropriate to pass the query to
///   any additional authorities to try to resolve, as that might be used to leak information to a
///   hostile party, and so a blocklist (or similar) authority should wrap responses for any
///   blocklist hits in LookupControlFlow::Break.
/// * Authorities may use LookupControlFlow::Skip to indicate the authority did not attempt to
///   process a particular query.  This might be used, for example, in a block list authority for
///   any queries that **did not** match the blocklist, to allow the recursor or forwarder to
///   resolve the query. Skip must not be used to represent an empty lookup; (use
///   Continue(EmptyLookup) or Break(EmptyLookup) for that.)
pub enum LookupControlFlow<T, E = LookupError> {
    /// A lookup response that may be passed to one or more additional authorities before
    /// being returned to the client.
    Continue(Result<T, E>),
    /// A lookup response that must be immediately returned to the client without consulting
    /// any other authorities.
    Break(Result<T, E>),
    /// The authority did not answer the query and the next authority in the chain should
    /// be consulted.
    Skip,
}
```

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


matheus23:
FWIW I created https://github.com/rkuhn/swarm-discovery/pull/10

dignifiedquire (approved):
looks reasonable

PR #3023, opened 2024-12-09
refactor!: improve reexport structure

dignifiedquire:
## Description

Restructures both `iroh_base` and `iroh` reexports and type placements, to simplify and clarify the overall structures.

## Breaking Changes

- `iroh-base`
    - `key` -> `iroh_base::NodeId` etc
    - `hash` -> `iroh_base::Hash` etc
    - `node_addr` -> `iroh_base::NodeAddr`
    - `relay_url`  -> `iroh_base::RelayUrl`
    - `relay_map` -> use `iroh_relay::relay_map`

- `iroh`
  - `hash` ->  `iroh_base`
  - `ticket` -> `iroh_base::ticket`)
  - `relay` -> removed, use `iroh_relay` directly if needed
  - `key` -> `NodeId`, etc top level now
  - `endpoint::Bytes` removed use `bytes::Bytes`
  - `endpoint::NodeAddr`, use `iroh::NodeAddr`

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


iroh-base/src/lib.rs
```diff
@@ -1,23 +1,45 @@
...
-#[cfg_attr(iroh_docsrs, doc(cfg(feature = "base32")))]
-pub mod ticket;
+
+#[cfg(feature = "relay")]
+#[cfg_attr(iroh_docsrs, doc(cfg(feature = "relay")))]
+pub use self::relay_map::{
+    QuicConfig as RelayQuicConfig, RelayMap, RelayNode, DEFAULT_RELAY_QUIC_PORT, DEFAULT_STUN_PORT,
+};
+#[cfg(any(feature = "relay", feature = "key"))]
+#[cfg_attr(iroh_docsrs, doc(cfg(any(feature = "relay", feature = "key"))))]
```
rklaehn, line 31: You should try this out, but I think if you reexport something from a module that has the `#[cfg_attr(iroh_docsrs, doc(cfg(any(feature = "relay", feature = "key"))))]` thing, all reexports will also have the proper tags in the docs.

So you could simplify this by just having this statement on the (non-pub) module.

iroh-base/src/lib.rs
```diff
@@ -1,23 +1,45 @@
...
 #[cfg(any(feature = "relay", feature = "key"))]
 mod relay_url;
-#[cfg(feature = "base32")]
-#[cfg_attr(iroh_docsrs, doc(cfg(feature = "base32")))]
-pub mod ticket;
+
+#[cfg(feature = "relay")]
+#[cfg_attr(iroh_docsrs, doc(cfg(feature = "relay")))]
+pub use self::relay_map::{
+    QuicConfig as RelayQuicConfig, RelayMap, RelayNode, DEFAULT_RELAY_QUIC_PORT, DEFAULT_STUN_PORT,
```
rklaehn, line 28: Not sure I like having these constants at top level. It is probably very rare that somebody needs them, so having them hidden away in some public mod is not that bad.

iroh-base/src/lib.rs
rklaehn: Ah, never mind. This is just iroh-base...

rklaehn: Why do we need a const for the size of an ed key in the first place?

iroh/src/lib.rs
```diff
@@ -233,27 +233,22 @@
...
-pub use endpoint::{AddrInfo, AddrInfoOptions, Endpoint, NodeAddr, RelayMode};
-pub use iroh_base::{
-    hash, key,
-    key::NodeId,
-    relay_map::{RelayMap, RelayNode, RelayUrl},
-    ticket,
-};
-pub use iroh_relay as relay;
+pub use endpoint::{Endpoint, RelayMode};
+pub use iroh_base::{hash, key, node_addr, relay_map, ticket};
```
flub, line 251: - do we need `hash`?
- I'd like to keep `NodeId` at least at the toplevel.
- If we're removing `RelayUrl` from here then why keep `RelayMode`?  Yes it's used in the builder but if you want anything else then disabled or default you'll need a whole bunch more relay types anyway.

dignifiedquire: because RelayMode is defined in iroh, not in base

iroh-base/src/lib.rs
dignifiedquire: > Why do we need a const for the size of an ed key in the first place?

we can drop that probably, need to check

iroh-base/src/lib.rs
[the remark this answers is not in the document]
dignifiedquire: sure

iroh-base/src/lib.rs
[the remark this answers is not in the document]
dignifiedquire: > because my hope is still that this crate will go away entirely...

🤣 sweet summer child

iroh/src/lib.rs
dignifiedquire: > * I'd like to keep `NodeId` at least at the toplevel.

that's fair

dignifiedquire: > * do we need `hash`?

interesting question, we actually don't

dignifiedquire: `hash` is gone now

iroh-base/src/lib.rs
dignifiedquire: > Why do we need a const for the size of an ed key in the first place?

it is used in a few places, sorry

iroh-base/src/lib.rs
dignifiedquire: I didn't find a better way, this is only for `iroh-base` though, in `iroh` it is much cleaner now

iroh-base/src/lib.rs
rklaehn: Just write 32 or have a const per crate. It is not going to change.

iroh/src/magicsock/node_map/path_state.rs
```diff
@@ -6,7 +6,7 @@ use std::{
     time::{Duration, Instant},
 };
 
-use iroh_base::key::NodeId;
+use iroh_base::NodeId;
```
flub, line 9: I find it a little odd that the imports don't come from `crate::NodeId`... but I don't think this is a problem.  Probably rust-analzyer will keep inserting this anyway.

flub (approved):
seems like a decent improvement

PR #3024, opened 2024-12-09
refactor: remove `AddrInfo`

dignifiedquire:
## Description

Simplify type structures

TODO

- [x] fix tests
- [x] resolve ticket breakage

## Breaking Changes

- remove `iroh_base::node_addr::AddrInfo`
- remove `iroh_base::node_addr::AddrInfoOptions`
- introduce `ticket` feature for `iroh_base`, to use `iroh_base::ticket`

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-base/src/node_addr.rs
+++ b/iroh-base/src/node_addr.rs
@@ -85,6 +85,11 @@ impl NodeAddr {
         }
     }
 
+    /// Returns true, if only a [`NodeId`] is present.
+    pub fn is_empty(&self) -> bool {
+        self.relay_url.is_none() && self.direct_addresses.is_empty()
+    }
+
     /// Applies the options to `self`.
     ///
     /// This is used to more tightly control the information stored in a [`NodeAddr`]
--- a/iroh-base/src/node_addr.rs
+++ b/iroh-base/src/node_addr.rs
@@ -41,31 +41,34 @@ pub use crate::relay_url::RelayUrl;
 pub struct NodeAddr {
     /// The node's identifier.
     pub node_id: NodeId,
-    /// Addressing information to connect to [`Self::node_id`].
-    pub info: AddrInfo,
+    /// The node's home relay url.
+    pub relay_url: Option<RelayUrl>,
+    /// Socket addresses where the peer might be reached directly.
+    pub direct_addresses: BTreeSet<SocketAddr>,
 }
 
 impl NodeAddr {
-    /// Creates a new [`NodeAddr`] with empty [`AddrInfo`].
+    /// Creates a new [`NodeAddr`] with no `relay_url` and no `direct_addresses`.
     pub fn new(node_id: PublicKey) -> Self {
         NodeAddr {
             node_id,
-            info: Default::default(),
+            relay_url: None,
+            direct_addresses: Default::default(),
         }
     }
 
-    /// Adds a relay url to the node's [`AddrInfo`].
+    /// Adds a relay url.
     pub fn with_relay_url(mut self, relay_url: RelayUrl) -> Self {
-        self.info.relay_url = Some(relay_url);
+        self.relay_url = Some(relay_url);
         self
     }
 
-    /// Adds the given direct addresses to the peer's [`AddrInfo`].
+    /// Adds the given direct addresses.
     pub fn with_direct_addresses(
         mut self,
         addresses: impl IntoIterator<Item = SocketAddr>,
     ) -> Self {
-        self.info.direct_addresses = addresses.into_iter().collect();
+        self.direct_addresses = addresses.into_iter().collect();
         self
     }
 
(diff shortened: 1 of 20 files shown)


iroh/src/discovery.rs
```diff
@@ -147,14 +148,14 @@ pub mod static_provider;
...
-    /// Publishes the given [`AddrInfo`] to the discovery mechanism.
+    /// Publishes the given [`RelayUrl`] and direct addreesses to the discovery mechanism.
     ///
     /// This is fire and forget, since the [`Endpoint`] can not wait for successful
     /// publishing. If publishing is async, the implementation should start it's own task.
     ///
     /// This will be called from a tokio task, so it is safe to spawn new tasks.
     /// These tasks will be run on the runtime of the [`super::Endpoint`].
-    fn publish(&self, _info: &AddrInfo) {}
+    fn publish(&self, _url: Option<&RelayUrl>, _addrs: &BTreeSet<SocketAddr>) {}
```
rklaehn, line 158: Can we keep something like AddrInfo just for discovery? I agree that having AddrInfo in the main API was a bit unpleasant, but I think this change here is not an improvement.

rklaehn (commented):
As I have discussed with @flub some time ago, I am not a fan of getting rid of AddrInfo entirely. I think keeping AddrInfo just for discovery would give you the same benefits for the most common use cases while still making implementing and working with discovery more pleasant.

iroh/src/discovery.rs
rklaehn: In the context of discovery these two things (direct addresses and relay url) are **always** used together. In my opinion it is a very clear case for having a struct. If this PR was new my review comment would be that it's good code but it would be nice to group the arguments for publish in a struct. 🤷‍♂️

iroh/Cargo.toml
```diff
@@ -177,7 +177,7 @@ name = "key"
 harness = false
 
 [features]
-default = ["metrics", "discovery-pkarr-dht"]
+default = ["metrics", "discovery-pkarr-dht", "discovery-local-network"]
```
flub, line 180: Did you mean to change this default?

iroh/src/discovery.rs
flub: Not sure I agree, I agonised and stalled  over this when I wrote my version of this a while back.  But now that DiscoveryItem can just contain NodeAddr (NodeId used to not be part of it before) this option of having two args for publish is very reasonable.

The changes to the discovery services also don't indicate AddrInfo is being replicated.

iroh-base/src/node_addr.rs
```diff
@@ -92,17 +93,31 @@ impl NodeAddr {
     ///
     /// [discovery]: https://docs.rs/iroh/*/iroh/index.html#node-discovery
     pub fn apply_options(&mut self, opts: AddrInfoOptions) {
-        self.info.apply_options(opts);
+        match opts {
+            AddrInfoOptions::Id => {
```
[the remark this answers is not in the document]
dignifiedquire: yeah, I will look at the fate of that at then

iroh/src/discovery.rs
dignifiedquire: I think this is a better api than a struct with only two fields, if we were expecting more arguments, that might be different, but I think having the two explicitly in there is actually good

iroh/Cargo.toml
dignifiedquire: no this is just for easier debugging

iroh/src/discovery.rs
rklaehn: 🤷‍♂️ I still don't like it very much but am not extremely opposed to this

 However, didn't we have the idea that publishing could in the future contain an optional ALPN in addition to relay url and direct addresses? At least we discussed this during the retreat.

dignifiedquire: I expect the API to change in some form, but I guess that needs some more design work in general

dignifiedquire:
@flub @rklaehn I have implemented internal legacy structs to keep the serialization stable

iroh-base/src/ticket/blob.rs
```diff
@@ -31,21 +32,47 @@ pub struct BlobTicket {
 /// postcard to add a discriminator.
 #[derive(Serialize, Deserialize)]
 enum TicketWireFormat {
-    Variant0(BlobTicket),
+    Variant0(Variant0BlobTicket),
```
flub, line 35: Oh, very nice how you don't even change the format of the tickets at all!

flub (commented):
I guess remaining is:

- doing something about the `AddrInfoOptions`.
- reverting the cargo features defaults

iroh-base/src/node_addr.rs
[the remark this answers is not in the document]
dignifiedquire: removed, this was really not pulling its wait anymore, can be easily achieved through other means now

flub (approved):

iroh-base/src/node_addr.rs
```diff
@@ -41,31 +41,34 @@ pub use crate::relay_url::RelayUrl;
 pub struct NodeAddr {
     /// The node's identifier.
     pub node_id: NodeId,
-    /// Addressing information to connect to [`Self::node_id`].
-    pub info: AddrInfo,
```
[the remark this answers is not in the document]
flub: Ugh, sorry about that.  We handled this for users who were serialising tickets: those are backwards compatible.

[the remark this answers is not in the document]
matheus23: Should we add backwards compatibility via a custom `Deserialize` implementation in a future release?

PR #3025, opened 2024-12-10
chore: (Deps) Update swarm-discovery to avoid idna <= 0.5 dep

matheus23:
## Description

- Updates to latest `swarm-discovery` alpha, so that `cargo deny` is happy again

## Breaking Changes

None, swarm-discovery isn't exposed in the API.

## Notes & open questions

It might be questionable if we want to depend on *more* alphas, but IMHO, this case is fine to avoid a security vulnerability.
Hopefully this is the last alpha. The next version of hickory we depend on should be 0.25.

## Change checklist

- [x] Self-review.
- ~~[ ] Tests if relevant.~~

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1627,30 +1627,6 @@ checksum = "7f24254aa9a54b5c858eaee2f5bccdb46aaf0e486a595ed5fd8f86ba55232a70"
 
 [[package]]
 name = "hickory-proto"
-version = "0.24.1"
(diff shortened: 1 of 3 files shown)


rklaehn (approved):
At least for swarm discovery - @rkuhn is quite busy with other stuff, so we are unlikely to see a major api change despite it being an alpha. So I think this is better than the deny.toml.

PR #3026, opened 2024-12-10
refactor(iroh): Add send queue between relay actor and relays

flub:
## Description

This introduces a new packet send queue between the RelayActor and the various ConnectedRelayActors.  Each ConnectedRelayActor will not block while sending them to the actual relay client.

This is important for the relay client to become nothing more than a stream and sink directly connected to the network stream in future refactoring.

It does not yet change the communication between
AsyncUdpSocket::try_send and the RelayActor itself to keep this change smaller.  This also should become a separate channel so that the inbox can still handle messages when the relay channel is full.

This is a small incremental change that does a few not very nice things, like having some .expect() calls.  This is deliberate to keep the changes smaller to review and reason about.

## Notes & open questions

Pay most attention to the ConnectedRelayActor and the way it does create a send future without blocking.  The exact mechanism will change once it is sending to a stream, but in the mean time this needs to be somewhat reasonable.

There are still tons of bugs in here.  But I don't think there are any regressions.  All the bugs existed already.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh/src/magicsock/relay_actor.rs
```diff
@@ -104,59 +110,37 @@ impl ConnectedRelayActor {
...
-                        }
+                    if self.handle_actor_msg(msg).await {
+                        break;
                     }
                 }
+                _ = &mut relay_send_fut, if matches!(relay_send_fut, Either::Right(_)) => {
+                    relay_send_fut = Either::Left(future::pending());
+                }
+                Some(msg) = self.relay_datagrams_send.recv(), if matches!(relay_send_fut, Either::Left(_)) => {
+                    relay_send_fut = Either::Right(Box::pin(relay_client.send(msg.0, msg.1)));
```
dignifiedquire, line 140: why not just spawn these? or sth..this construction is very hard to follow..

iroh/src/magicsock/relay_actor.rs
```diff
@@ -173,6 +157,41 @@ impl ConnectedRelayActor {
         Ok(())
     }
 
+    async fn handle_actor_msg(&mut self, msg: ConnectedRelayMessage) -> bool {
+        trace!("tick: inbox: {:?}", msg);
+        match msg {
```
dignifiedquire, line 162: are you sure all these should be preferred over regular writes? it seems some of these will trigger additional sending, eg `Ping` and `NotePreferred`, or amy I missing sth?

iroh/src/magicsock/relay_actor.rs
flub: I really do not want to spawn a task here.  The plan is for `relay_client` to be directly a `AsyncWriter`, probably wrapped in a `Sink` (wrapped in some different client of Client struct).  So this actor's loop has to learn how to send things to the relay server while not deadlocking itself.

My approach here is to start doing this for just sending packets.  The others which send to the relay are `Ping` and `NotePreferred` (and kind of `GetLocalAddr` because it queues up on the same actor inbox right now, but doesn't go all the way to the relay).

If you look at the points at which `Ping` and `GetLocalAddr` are being used: this is only when there network has changed and the `RelayActor` wants to be sure this connection still works.  So at that time there's already a potential hiccup happening.  I totally want to fix those too, but plan to do that by moving those things into the `ConnectedRelayActor` instead.

iroh/src/magicsock/relay_actor.rs
flub: They are not being preferred, the `tokio::select!` is not biased.  The ones that do trigger a send are indeed blocking.  I don't like it, but it is not a regression so in the interest of taking small steps I was only improving the SendPacket path.

iroh/src/magicsock/relay_actor.rs
dignifiedquire: could you use `MaybeFuture` to make this a bit simpler?

flub: I made some changes to `MaybeFuture`, agreed that this is nicer.  PTAL

iroh/src/util.rs
```diff
@@ -6,17 +6,44 @@ use std::{
...
+/// The [`Default`] impl will create a [`MaybeFuture`] without an inner.
 #[derive(Debug)]
 pub(crate) struct MaybeFuture<T> {
     /// Future to be polled.
     pub inner: Option<T>,
 }
 
+impl<T> MaybeFuture<T> {
+    /// Creates a [`MaybeFuture`] without an inner future.
+    pub(crate) fn none() -> Self {
```
flub, line 23: I've leant into the `Option` sematics and gone with `::none()` constructor and `.is_some()` and `.is_none()`.  This seems to match the naming which is almost `OptionalFuture`.

The alternative is to lean into collection APIs: `::empty()` as constructor and `.is_empty()` to check (I guess `.len()` would be it's counterpart but the clippy convention is to use `!m.is_empty()`).

dignifiedquire (approved):

iroh/src/util.rs
dignifiedquire: I like the Option version

iroh/src/magicsock/relay_actor.rs
dignifiedquire: much easier to understand (at least for me) now, thanks

PR #3028, opened 2024-12-11
feat(iroh-net-report)!: add QUIC address discovery probes

ramfox:
## Description

Adds probes for QUIC Address Discovery.

## Breaking Changes

- `iroh-net-report`
    - changed
        - `iroh_net_report::Client::get_report` takes new parameter `quic_config: net_report::QuicConfig`
        - `iroh_net_report::Client::get_report_channel` takes new parameter `quic_config: net_report::QuicConfig`

## Notes & open questions

Next PR is properly adding this to iroh (& a new PR for iroh-doctor). `magicsock.rs` was adjusted so that `iroh` can still depend on the repo version of `net-report`, but does not enable QAD.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh-net-report/src/reportgen.rs
```diff
@@ -851,6 +883,39 @@ async fn run_stun_probe(
     }
 }
 
+/// Run a QUIC address discovery probe.
+// TODO(ramfox): if this probe is aborted, then the connection will never be
+// properly closed, possibly causing errors on the server.
```
ramfox, line 888: highlighting this, not sure how worth it it is to clean this properly. I haven't noticed any warnings when running against a local server or in tests

iroh-net-report/src/reportgen.rs
```diff
@@ -851,6 +883,39 @@ async fn run_stun_probe(
...
+    url: RelayUrl,
+    relay_addr: SocketAddr,
+    probe: Probe,
+) -> Result<ProbeReport, ProbeError> {
+    match probe.proto() {
+        ProbeProto::QuicIpv4 => debug_assert!(relay_addr.is_ipv4()),
+        ProbeProto::QuicIpv6 => debug_assert!(relay_addr.is_ipv6()),
+        _ => debug_assert!(false, "wrong probe"),
+    }
+    // TODO(ramfox): what to put here if no host is given?
```
ramfox, line 900: also highlighting this. Anything better?

iroh-net-report/src/reportgen.rs
flub: Why do you think it won't be closed properly?  The `Connection` itself is closed on drop.  The `Endpoint` is but a clone of the endpoint inside the magic socket so will stay open as long as needed and be sending the retries of the CONNECTION_CLOSE frame sent.  So as long as the `iroh::Endpoint::close` is called and awaited this will always close the connection "properly".  And if someone doesn't do that, then well - we'll need to rely on timeouts which is fine.  The server can tweak its timeouts if that becomes a problem, at the end of the day you can't stop boxes from just going down either.

iroh-net-report/src/reportgen/probes.rs
```diff
@@ -88,6 +92,16 @@ pub(super) enum Probe {
         delay: Duration,
         node: Arc<RelayNode>,
     },
+    #[display("QUIC Address Discovery Ivp4 after {delay:?} to {node}")]
```
flub, line 95: What would you think of calling this `QAD IPv4 after...`?  (also - typo on IPv4)

The way these are logged i think it would be nicer to have a bit shorter and I don't mind that you need to know what QAD means here.  These logs are for us.

flub (approved):
Looks great!

iroh-net-report/src/reportgen.rs
ramfox: oh yea, thank you. I was so far in the "oh shit we have to close correctly" brain space i forgot that the connections will just close on drop just fine as long as the endpoint is handled correctly.

🙏

iroh-net-report/src/reportgen.rs
```diff
@@ -898,8 +896,15 @@ async fn run_quic_probe(
...
-    let host = url.host_str().unwrap_or("localhost");
+    let host = match url.host_str() {
+        Some(host) => host,
+        None => {
+            return Err(ProbeError::Error(
+                anyhow!("URL must have 'host' to use QUIC address discovery probes"),
+                probe.clone(),
+            ));
+        }
+    };
```
flub, line 907: FYI this is a prime place for using the fancy new-ish syntax of `let Some(host) = url.host_str() else { return ... };`  Not important enough to follow up on though.

iroh-net-report/src/reportgen.rs
```diff
@@ -898,8 +896,15 @@ async fn run_quic_probe(
         ProbeProto::QuicIpv6 => debug_assert!(relay_addr.is_ipv6()),
         _ => debug_assert!(false, "wrong probe"),
     }
-    // TODO(ramfox): what to put here if no host is given?
-    let host = url.host_str().unwrap_or("localhost");
+    let host = match url.host_str() {
+        Some(host) => host,
+        None => {
+            return Err(ProbeError::Error(
```
flub, line 902: Hum, I believe this should be `ProbeError::AbortSet`?  Why did you choose just this one?  If the RelayUrl has no host none of the retries after a small delay will work either.

PR #3029, opened 2024-12-11
docs(*): Use doc-auto-cfg feature

flub:
## Description

This switches from the doc-cfg feature which needs an explicit
annotation to the doc-auto-cfg feature which does not need this.  We
already had several places where the docs and actual features were out
of sync.  So hopefully this will work better.

doc-auto-cfg does not seem perfect and may miss some things.  E.g. one
difference I noticed is that it does not seem to surface the "test"
feature when we do `any(test, feature = "something")`.  But I'm
thinking that might be a reasonable price to pay here.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,13 +30,13 @@ incremental = false
 missing_debug_implementations = "warn"
 
 # We use this --cfg for documenting the cargo features on which an API
-# is available.  To preview this locally use: RUSTFLAGS="--cfg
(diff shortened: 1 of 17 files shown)


dignifiedquire:
beautiful

flub:
> beautiful

Beautiful enough to approve?

dignifiedquire (approved):

PR #3030, opened 2024-12-11
refactor(iroh): Newtype the packet sent over relay servers

flub:
## Description

There is a difference between the datagrams that the MagicSock sends and the packets that contain datagrams that gets sent to relay servers.  But this was all just Bytes.

This newtypes the packets explicitly and makes it a sibling of the existing RelayRecvDatagram.  Now the PacketizeIter at least does not both consume and produce the same type, because they're not the same things.

## Notes & open questions

This clones the RelayUrl (usually) twice per packet sent now, instead of (usually) once.  That's the only thing I don't like about this really.

I still wonder whether RelayUrl should not be some sort of interned and cloned thing.

Watch out, the base of this PR is `flub/active-relay-again`.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/magicsock/relay_actor.rs
+++ b/iroh/src/magicsock/relay_actor.rs
@@ -797,7 +797,10 @@ struct ActiveRelayHandle {
 
 /// A packet to send over the relay.
 ///
-/// This is nothing but a newtype, it should be constructed using [`PacketizeIter`].
+/// This is nothing but a newtype, it should be constructed using [`PacketizeIter`].  This
+/// is a packet of one or more datagrams, each prefixed with a u16-be length.  This is what
+/// the `Frame::SendPacket` of the `DerpCodec` transports and is produced by
+/// [`PacketizeIter`] and transformed back into datagrams using [`PacketSplitIter`].
 #[derive(Debug, PartialEq, Eq)]
 struct RelaySendPacket {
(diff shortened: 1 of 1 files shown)


matheus23 (approved):
This looks like a good change, but I need someone to explain to me why this is a `RelaySend`*`Packet`*, not a `RelaySend`*`Datagram`* and why the `Recv` one is the other way around.

flub:
> This looks like a good change, but I need someone to explain to me why this is a `RelaySend`_`Packet`_, not a `RelaySend`_`Datagram`_ and why the `Recv` one is the other way around.

The explanation is in the `PacketizeIter`: it prefixes every datagram with a u16be length and potentially concatenates several datagrams together.

Whether that is a good idea I'm going to leave alone.  I... don't think it is??  It basically only happens when Quinn manages to send a GSO datagram (and I'm not sure if that is really hooked up properly currently so that it would, I think it is though).  But the point of GSO datagrams is to have less work but we now do another copy of **every** packet because we need to prefix it with 2 bytes.  So we really are hurting because of this instead of improving.  The relay protocol should probably have a separate frame/packet type for GSO datagrams with the stride length encoded in it.

Actually... let me write this as an issue...

(I'll also improve the docs in this PR a little)

PR #3032, opened 2024-12-12
feat(net-report)!: add `net_report::Options` to specify which probes you want to run

ramfox:
## Description

Breaking change that allows you to specify which probe protocols you would like to run.

Really written so it is easier to test QUIC address discovery on `iroh doctor`, but this also allows us to test specific protocols in isolation.

## Breaking Changes

- `iroh-net-report`
    - added
        - `net_report::Client::get_report_with_options`
    - changed
        - `net_report::Client::get_report_channel` now takes an `opts: net_report::Options`
        - `net_report::Client` will no longer bind `UdpSocket`s when one is not provided for both STUN over IPv4 or STUN over IPv6.

## Notes & open questions

I'm open to adjustments on the `Options`. The whole "we build a UDP socket for you if one isn't provided" bit means that I couldn't rely on, for example, `stun_sock_v4 == None` to know whether the user wanted to run a stun ipv4 probe, so I needed the `protocols: BTreeSet<ProbeProtocol>` field. I think this is a decent compromise between annoying and useful, but would love feedback.

Edit:
Adjusted this so that the Options DO correlate to the protocols we want to run. This means that we will no longer bind a UdpSocket for you if one is not provided.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net-report/src/lib.rs
+++ b/iroh-net-report/src/lib.rs
@@ -16,19 +16,19 @@ use std::{
     sync::Arc,
 };
 
-use anyhow::{anyhow, Context as _, Result};
+use anyhow::{anyhow, Result};
 use bytes::Bytes;
 use hickory_resolver::TokioResolver as DnsResolver;
 use iroh_base::RelayUrl;
 #[cfg(feature = "metrics")]
 use iroh_metrics::inc;
 use iroh_relay::{protos::stun, RelayMap};
-use netwatch::{IpFamily, UdpSocket};
+use netwatch::UdpSocket;
 use tokio::{
     sync::{self, mpsc, oneshot},
     time::{Duration, Instant},
 };
-use tokio_util::{sync::CancellationToken, task::AbortOnDropHandle};
+use tokio_util::task::AbortOnDropHandle;
 use tracing::{debug, error, info_span, trace, warn, Instrument};
 
 mod defaults;
@@ -40,6 +40,8 @@ mod reportgen;
 pub use metrics::Metrics;
 use reportgen::ProbeProto;
 pub use reportgen::QuicConfig;
+#[cfg(feature = "stun-utils")]
+pub use stun_utils::bind_local_stun_socket;
 
 const FULL_REPORT_INTERVAL: Duration = Duration::from_secs(5 * 60);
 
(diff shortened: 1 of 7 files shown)


iroh-net-report/src/reportgen/probes.rs
```diff
@@ -261,19 +265,30 @@ impl ProbePlan {
...
+                    quic_ipv6_probes,
+                ],
+            );
 
             // The HTTP and ICMP probes only start after the STUN probes have had a chance.
             let mut https_probes = ProbeSet::new(ProbeProto::Https);
             let mut icmp_probes_ipv4 = ProbeSet::new(ProbeProto::IcmpV4);
             let mut icmp_probes_ipv6 = ProbeSet::new(ProbeProto::IcmpV6);
+
+            let has_priority_probes = plan.has_priority_probes();
```
flub, line 283: little point in another variable is there?

iroh-net-report/src/reportgen/probes.rs
```diff
@@ -298,9 +313,11 @@ impl ProbePlan {
...
                 }
             }
-            plan.add(https_probes);
-            plan.add(icmp_probes_ipv4);
-            plan.add(icmp_probes_ipv6);
+
+            plan.add_probes(
+                &protocols,
+                vec![https_probes, icmp_probes_ipv4, icmp_probes_ipv6],
+            );
```
flub, line 320: I wonder if you can add the `protocols` to the plan itself and `plan.add()` would then be able to check if it actually needs adding.  That way this code doesn't have to change (I don't know why, but the old version with 3 different calls to plan.add() is more readable to me).  Maybe rename it to `plan.add_if_enabled(probes)`.

iroh-net-report/src/reportgen/probes.rs
```diff
@@ -261,19 +265,30 @@ impl ProbePlan {
...
+
+            let has_priority_probes = plan.has_priority_probes();
             for attempt in 0..3 {
-                let start = *max_stun_delay.get_or_insert_with(|| plan.max_delay())
-                    + DEFAULT_INITIAL_RETRANSMIT;
-                let delay = start + DEFAULT_INITIAL_RETRANSMIT * attempt as u32;
+                let delay = if !has_priority_probes {
+                    DEFAULT_INITIAL_RETRANSMIT * attempt as u32
+                } else {
+                    let start = *max_stun_delay.get_or_insert_with(|| plan.max_delay())
```
flub, line 288: `max_stun_delay` is no longer a correct variable name I guess?  `max_high_prio_delay`?

iroh-net-report/src/reportgen/probes.rs
```diff
@@ -261,19 +265,30 @@ impl ProbePlan {
...
             let mut https_probes = ProbeSet::new(ProbeProto::Https);
             let mut icmp_probes_ipv4 = ProbeSet::new(ProbeProto::IcmpV4);
             let mut icmp_probes_ipv6 = ProbeSet::new(ProbeProto::IcmpV6);
+
+            let has_priority_probes = plan.has_priority_probes();
             for attempt in 0..3 {
-                let start = *max_stun_delay.get_or_insert_with(|| plan.max_delay())
-                    + DEFAULT_INITIAL_RETRANSMIT;
-                let delay = start + DEFAULT_INITIAL_RETRANSMIT * attempt as u32;
+                let delay = if !has_priority_probes {
```
flub, line 285: I'm not entirely sure why this needs changing.  The old version uses `let start = *max_stun_delay.get_or_insert_with(|| plan.max_delay())`.  I would expect `plan.max_delay()` to return 0 if there were no high-prio probes.  So then `start` should also be 0?

Admittedly I'm just reading code and didn't try this out.

flub:
> > The whole "we build a UDP socket for you if one isn't provided"
> 
> I wonder if we should change this

Yes, I agree this is worth trying out.  I forgot what the reason for this is, but I think it might have just been for ease of testing, which you're changing here anyway by being able to not have stun probes when testing.  And in the MagicSock we'll never need this functionality.

divagant-martian:
I tried to create an alternative that would look like:
`Options::run_stun(probe_ipv4: bool, probe_ipv6: socket: None)` and so on for each "type" of protocol to keep the functionality and make it consistent. 

But I wonder if we should then just remove it completely. Creating a socket is not that big of a hassle to justify essentially equivalent options to behave so differently/inconsistently among them, and make it consistent across protocols is also (comparatively) a lot more work.

iroh-net-report/src/reportgen/probes.rs
ramfox: agreed

iroh-net-report/src/reportgen/probes.rs
ramfox: changed!

iroh-net-report/src/reportgen/probes.rs
ramfox: okay, I definitely changed this in the most confusing way, so refactored and added comments.

Basically, if there are no high priority probes, then we don't need `start` to also add the `DEFAULT_INITIAL_RETRANSMIT`, but if it DOES than we want to add that buffer.

Hopefully this is clearer. The issue was that if you don't have any high prio probes in the list of probes you are attempting, then all the probes added in this section started with a 100ms delay.

ramfox:
Okay here is what it looks like with removing the `UdpSocket` and adjusting the `Options` accordingly.

iroh-net-report/src/lib.rs
```diff
@@ -252,39 +387,56 @@ impl Client {
...
+    ///
+    /// This will attempt to use *all* probe protocols.
     pub async fn get_report(
         &mut self,
-        dm: RelayMap,
-        stun_conn4: Option<Arc<UdpSocket>>,
-        stun_conn6: Option<Arc<UdpSocket>>,
+        relay_map: RelayMap,
+        stun_sock_v4: Option<Arc<UdpSocket>>,
+        stun_sock_v6: Option<Arc<UdpSocket>>,
```
flub, line 396: :green_heart:

iroh-net-report/src/lib.rs
```diff
@@ -206,6 +207,140 @@ impl Default for Reports {
...
+    }
+
+    /// Enable quic probes
+    pub fn quic_config(mut self, quic_config: Option<QuicConfig>) -> Self {
+        self.quic_config = quic_config;
+        self
+    }
+
+    /// Enable icmp_v4 probe
+    pub fn enable_icmp_v4(mut self) -> Self {
```
flub, line 297: It is slightly odd that the `stun_v4`, `stun_v6` and `quic_config` methods all allow both enabling and disabling but then you have this instead of `icmp_v4(mut self, enable: bool)`.

But it doesn't matter much.  This meets our needs anyway.

iroh-net-report/src/reportgen/probes.rs
```diff
@@ -263,46 +273,52 @@ impl ProbePlan {
...
+            let mut icmp_v6_probes = ProbeSet::new(ProbeProto::IcmpV6);
+
             for attempt in 0..3 {
-                let start = *max_stun_delay.get_or_insert_with(|| plan.max_delay())
-                    + DEFAULT_INITIAL_RETRANSMIT;
+                let mut start = *max_high_prio_delay.get_or_insert_with(|| plan.max_delay());
+                // if there are high priority probes, ensure there is a buffer between
+                // the highest probe delay and the next probes we create
+                // if there are no high priority probes, we don't need a buffer
+                if plan.has_priority_probes() {
```
flub, line 291: This is indeed much more readable, thanks!

Can this condition now be `if !start.is_zero()` which would avoid this extra function and categorizing probes?

flub (approved):

iroh-net-report/src/lib.rs
```diff
@@ -206,6 +207,140 @@ impl Default for Reports {
...
+            icmp_v4: true,
+            icmp_v6: true,
+            https: true,
+        }
+    }
+}
+
+impl Options {
+    /// Create an empty Options that enables no probes
+    pub fn empty() -> Self {
```
dignifiedquire, line 267: bikeshed: `disabled`?

iroh-net-report/src/lib.rs
```diff
@@ -740,7 +863,7 @@ struct ReportRun {
 /// If successful this returns the bound socket and will forward STUN responses to the
 /// provided *actor_addr*.  The *cancel_token* serves to stop the packet forwarding when the
 /// socket is no longer needed.
-fn bind_local_stun_socket(
+pub fn bind_local_stun_socket(
```
dignifiedquire, line 866: do we have to make this public?

dignifiedquire: maybe just under a test_utils flag?

iroh-net-report/src/reportgen/probes.rs
ramfox: I thought about that, but was trying to future-proof:

basically, if we do some refactor that means that it's possible for only 1 probe to be attempted, that probe will be at delay == 0, which would make `plan.max_delay` == 0

iroh-net-report/src/lib.rs
ramfox: Yea good point. It's useful for `iroh doctor` but anyone else attempting a net report for production use would likely do something similar to what we currently do in iroh & pass in the specific socket they are using.

iroh-net-report/src/reportgen/probes.rs
flub: Oh, right.  Didn't think of that case, yes your approach here makes the most sense then.

iroh-net-report/src/lib.rs
ramfox: nah ur right it's weird

PR #3034, opened 2024-12-12
refactor: remove parking-lot dependency

dignifiedquire:
## Description

Currently there was an inconsistent usage of `std::sync::Mutex` and `parking_lot::Mutex`. This normalizes the usage to always use `std::sync::Mutex` and remove the external dependency.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh/src/discovery.rs
```diff
@@ -500,6 +499,7 @@ mod tests {
             self.shared
                 .nodes
                 .lock()
+                .unwrap()
```
flub, line 502: I would prefer if we used `.expect("poisoned")` as that way we can more easily treat `.unwrap()` as "un-audited".  (Having done this whole thing about auditing all our unwraps once before :wink: )

flub (commented):
I'm all for this but would prefer `.expect("poisoned")` over `.unwrap()`.

dignifiedquire:
> I'm all for this but would prefer `.expect("poisoned")` over `.unwrap()`.

we have never done this, and that a lot of unwraps like this that don't have that.. we can do it in one go as a follow up if you really want to

dignifiedquire:
seems we don't have many `unwrap`s.., okay I will change them

iroh/src/discovery.rs
dignifiedquire: fixed

flub:
> > I'm all for this but would prefer `.expect("poisoned")` over `.unwrap()`.
> 
> we have never done this, and that a lot of unwraps like this that don't have that.. we can do it in one go as a follow up if you really want to

For completeness, we have done this:
https://github.com/n0-computer/iroh/issues/1528
https://github.com/n0-computer/iroh/pull/2046

dignifiedquire:
yep, we just missed adding the lints, which is why they can creep back in

iroh/src/magicsock/node_map.rs
```diff
@@ -762,7 +799,7 @@ mod tests {
             // add address
             node_map.add_test_addr(node_addr);
             // make it active
-            node_map.inner.lock().receive_udp(addr);
+            node_map.inner.lock().unwrap().receive_udp(addr);
```
flub, line 802: still unwrap

iroh/src/magicsock/node_map.rs
```diff
@@ -772,7 +809,7 @@ mod tests {
             node_map.add_test_addr(node_addr);
         }
 
-        let mut node_map_inner = node_map.inner.lock();
+        let mut node_map_inner = node_map.inner.lock().unwrap();
```
flub, line 812: still unwrap

iroh/src/magicsock/node_map.rs
```diff
@@ -824,6 +866,7 @@ mod tests {
         node_map
             .inner
             .lock()
+            .unwrap()
```
flub, line 869: still unwrap

iroh/src/discovery.rs
```diff
@@ -508,7 +508,7 @@ mod tests {
             node_id: NodeId,
         ) -> Option<BoxStream<Result<DiscoveryItem>>> {
             let addr_info = match self.resolve_wrong {
-                false => self.shared.nodes.lock().get(&node_id).cloned(),
+                false => self.shared.nodes.lock().unwrap().get(&node_id).cloned(),
```
flub, line 511: still unwrap

iroh/src/discovery.rs
flub: except this one?

iroh/src/magicsock/node_map.rs
dignifiedquire: that's a test, that's fine

flub (approved):

PR #3041, opened 2024-12-13
refactor(iroh): Remove ActiveRelayMessage::GetClient

flub:
## Description

The headline feature is the removal of ActiveRelayMessage::GetClient.
This now means the relay client lives entirely in the ActiveRelayActor
and can start being refactored into a stream.

But there's a whole bunch more!

- ActiveRelayMessage::GetNodeRoute also no longer sends the client
  back.

- The creation of the relay::client::Client has been moved over to the
  ActiveRelayClient.  This will make it easier for this actor to be
  in charge of this and be able to re-create the connection to the
  relay.

- The creation of ActiveRelayActors is now independent of finding
  routes to nodes and staring.  Resulting in smaller functions with
  fewer concerns.

- The RelayActor now notices when an ActiveRelayActor task terminates
  and cleans up.

- close_or_reconnect_relay now always reconnects.  We already require
  relays staying connected when the ActiveRelayActor exists so this was
  a path in which this was not honoured.

- The RelayActor does some more concurrent futures work when sending
  messages to several ActiveRelayActors at once.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh/src/magicsock/relay_actor.rs
```diff
@@ -98,7 +127,33 @@ impl ActiveRelayActor {
...
+            dns_resolver,
+            proxy_url,
+            prefer_ipv6,
+            #[cfg(any(test, feature = "test-utils"))]
+            insecure_skip_cert_verify,
+        } = opts;
+        let mut builder =
+            relay::client::ClientBuilder::new(url).address_family_selector(move || {
+                let prefer_ipv6 = prefer_ipv6.clone();
+                Box::pin(async move { prefer_ipv6.load(Ordering::Relaxed) })
```
dignifiedquire, line 145: do we still need the `address_family_selector` to accept a future?

flub: Not sure, I was not really aiming to touch that bit of the code.  I hope when I change `relay::client::Client` to not be an actor that this would disappear, but that's for a later PR.

dignifiedquire: I just checked, this is the only usage, I'll do a PR to change it

dignifiedquire: https://github.com/n0-computer/iroh/pull/3044

iroh/src/magicsock/relay_actor.rs
```diff
@@ -346,6 +405,12 @@ impl RelayActor {
                     trace!("shutting down");
                     break;
                 }
+                Some(Err(err)) = self.active_relay_tasks.join_next() => {
```
dignifiedquire, line 408: needs to check for `!.is_empty()`

iroh/src/magicsock/relay_actor.rs
```diff
@@ -346,6 +405,12 @@ impl RelayActor {
...
+                    if !err.is_cancelled() {
```
dignifiedquire, line 409: needs panic check

iroh/src/magicsock/relay_actor.rs
```diff
@@ -455,171 +522,110 @@ impl RelayActor {
...
+                if maybe_url.is_some() {
+                    found_relay = maybe_url;
+                    break;
                 }
             }
         }
-        self.active_relay_handle(url).await
+        match found_relay {
+            Some(url) => self.active_relay_handle(&url).await,
+            None => self.active_relay_handle(url).await,
```
dignifiedquire, line 574: might read easier as

```rust
let active_url = found_relay.as_ref().unwrap_or(url);
self.active_relay_handle(url).await;
```

iroh/src/magicsock/relay_actor.rs
```diff
@@ -455,171 +522,110 @@ impl RelayActor {
...
+        let (inbox_tx, inbox_rx) = mpsc::channel(64);
+        let span = info_span!("active-relay", %url);
+        let opts = ActiveRelayActorOptions {
+            url,
+            relay_datagrams_send: send_datagram_rx,
+            relay_datagrams_recv: self.relay_datagram_recv_queue.clone(),
+            connection_opts,
+        };
+        let actor = ActiveRelayActor::new(opts);
+        self.active_relay_tasks.spawn(
```
dignifiedquire, line 614: 💙

iroh/src/magicsock/relay_actor.rs
flub: It (surprisingly) doesn't!  Which I also only realised once I thought I found a bug that this should cause a hot-loop but it didn't cause a hot-loop.

Because this `tokio::select!` call does not have an `else` branch the behaviour is as follows:
- the `self.active_relay_tasks.join_next()` future will will resolve immediately upon calling `tokio::select!`.
- The future output will not match the pattern `Some(Err(_))`.  **This disables the branch.**
- Because there are other enabled branches the select will continue waiting for the next future to resolve.
- Because there is not `else` branch, when all branches are disabled this will finally loop.
- But in any case, the last thing will never happen because one of the awaited futures is the cancellation token.

You could argue as to why rely on this behaviour?  But the branch disabling and `else` behaviour is exactly the same whether this branch is disabled by not matching the pattern or whether is is disabled by an if-clause.  So this doesn't change what you have to think about.  So it ends up with this only being a syntactic difference, not a difference in behaviour or cognitive load.  And the pattern-maching-disabling is less code and easier to read I think - once you understand it.

I've come to treat `tokio::select!` as the better enemy that I know.  So I am happy to rely fully on it for logic like this.

iroh/src/magicsock/relay_actor.rs
flub: Does it?  I explicitly only check for anything **not** cancelled so it includes panics.  My first version indeed had `if err.is_panic()` but the error handling in this case was exactly the same: because I don't think propagating a panic up here will help much.

Or do you think panics should **always** be propagated upwards?

dignifiedquire: > Or do you think panics should **always** be propagated upwards?

this is currently what we do in all other cases, so for consistency I think that is the best way to do so for now

iroh/src/magicsock/relay_actor.rs
dignifiedquire: in that case we should probably update the other places we do it like this, having this incosistent in the code base is going to confuse folks even more

flub: #3052

dignifiedquire (approved):

PR #3043, opened 2024-12-13
refactor!: Make PUBLIC_KEY_LENGTH a const that is on PublicKey

rklaehn:
## Description

Make PUBLIC_KEY_LENGTH a const that is on PublicKey

that way it is not so in your face

## Breaking Changes

- PUBLIC_KEY_LENGTH is moved from a top level constant to PublicKey::LENGTH

### DIFF
--- a/iroh-base/src/key.rs
+++ b/iroh-base/src/key.rs
@@ -10,7 +10,7 @@ use std::{
     time::Duration,
 };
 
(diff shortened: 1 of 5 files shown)


divagant-martian (approved):
personally very much in favor of this

dignifiedquire (approved):

PR #3046, opened 2024-12-13
refactor(iroh-base)!: reduce dependencies

dignifiedquire:
## Description

- ensure `zeroize` is used for `SecretKey`
- remove unused dependencies and features from `iroh-base`

## Breaking Changes

- `iroh_base::SecretKey::generate` now takes an rng
- removed `iroh_base::SecretKey::generate_with_rng`, use `generate` directly
- removed `iroh_base::SecretKey::to_openssh`
- removed `iroh_base::SecretKey::from_openssh`
- `anyhow::Error` is replaced with explicit errors for `RelayUrl::from_str`
- `anyhow::Error` is replaced with explicit errors for `SharedSecret::open`

## Notes & open questions

As a follow up, all usage of `SecretKey::generate` in tests should be changed to use a seeded rng, like chacha8, instead of `thread_rng`

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-base/src/key.rs
+++ b/iroh-base/src/key.rs
@@ -335,6 +335,12 @@ impl SecretKey {
     }
 
     /// Generate a new [`SecretKey`] with a randomness generator.
+    ///
+    /// ```rust
+    /// // use the OsRng option for OS depedndent most secure RNG.
+    /// let mut rng = rand::rngs::OsRng;
+    /// let _key = iroh_base::SecretKey::generate(&mut rng);
+    /// ```
     pub fn generate<R: CryptoRngCore>(mut csprng: R) -> Self {
         let secret = SigningKey::generate(&mut csprng);
 
--- a/iroh-base/src/key.rs
+++ b/iroh-base/src/key.rs
@@ -6,20 +6,18 @@ use std::{
     fmt::{Debug, Display},
     hash::Hash,
     str::FromStr,
-    sync::Mutex,
+    sync::{Mutex, OnceLock},
     time::Duration,
 };
 
 pub use ed25519_dalek::Signature;
 use ed25519_dalek::{SignatureError, SigningKey, VerifyingKey};
-use once_cell::sync::OnceCell;
 use rand_core::CryptoRngCore;
 use serde::{Deserialize, Serialize};
-use ssh_key::LineEnding;
 use ttl_cache::TtlCache;
 
-pub use self::encryption::SharedSecret;
 use self::encryption::{public_ed_box, secret_ed_box};
+pub use self::encryption::{DecryptionError, SharedSecret};
 
 #[derive(Debug)]
 struct CryptoKeys {
@@ -48,7 +46,7 @@ const KEY_CACHE_TTL: Duration = Duration::from_secs(60);
 ///
 /// So that is about 4MB of max memory for the cache.
 const KEY_CACHE_CAPACITY: usize = 1024 * 16;
-static KEY_CACHE: OnceCell<Mutex<TtlCache<[u8; 32], CryptoKeys>>> = OnceCell::new();
+static KEY_CACHE: OnceLock<Mutex<TtlCache<[u8; 32], CryptoKeys>>> = OnceLock::new();
 
 fn lock_key_cache() -> std::sync::MutexGuard<'static, TtlCache<[u8; 32], CryptoKeys>> {
     let mutex = KEY_CACHE.get_or_init(|| Mutex::new(TtlCache::new(KEY_CACHE_CAPACITY)));
@@ -181,6 +179,7 @@ impl PublicKey {
         data_encoding::HEXLOWER.encode(&self.as_bytes()[..5])
     }
 
+    /// The length of an ed25519 `PublicKey`, in bytes.
     pub const LENGTH: usize = ed25519_dalek::PUBLIC_KEY_LENGTH;
 }
 
@@ -259,6 +258,7 @@ pub enum KeyParsingError {
     /// Error when decoding the public key.
     #[error("key: {0}")]
     Key(#[from] ed25519_dalek::SignatureError),
+    /// The encoded information had the wrong length.
     #[error("invalid length")]
     DecodeInvalidLength,
 }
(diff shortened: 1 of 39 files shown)


matheus23:
Hm. I would argue the `SecretKey::generate` function change may not be justified. This increases the burden on the library user. We need to *at least* have a doctest that shows them how to use the right rng with `Secret::generate` (I'd say for most cases that's likely `rand::thread_rng()`).
I suspect that removing the `rand` dependency from `iroh-base` only makes `iroh-base` lighter, not `iroh` itself, does it?

iroh-base/src/key.rs
```diff
@@ -280,7 +280,7 @@ impl FromStr for PublicKey {
 #[derive(Clone)]
 pub struct SecretKey {
     secret: SigningKey,
-    secret_crypto_box: OnceCell<crypto_box::SecretKey>,
+    secret_crypto_box: OnceLock<crypto_box::SecretKey>,
```
rklaehn, line 283: This wasn't stable back when we wrote this initially, right?

iroh-base/src/key.rs
```diff
@@ -334,40 +334,13 @@ impl SecretKey {
...
+        let secret = SigningKey::generate(&mut csprng);
 
         Self {
             secret,
-            secret_crypto_box: OnceCell::default(),
-        }
-    }
-
-    /// Serialise this key to OpenSSH format.
-    pub fn to_openssh(&self) -> ssh_key::Result<zeroize::Zeroizing<String>> {
```
rklaehn, line 354: These are useful. Can we have this code somewhere, maybe in examples?

matheus23 (approved):
Alright. I'll defer to your experience in that case then, but I *do* demand a `no_run` doctest with `rand::thread_rng()` example usage in the `SecretKey::generate` docs before you merge :P

iroh-base/src/key.rs
```diff
@@ -485,7 +450,7 @@ mod tests {
 
     #[test]
     fn test_display_from_str() {
-        let key = SecretKey::generate();
```
rklaehn, line 488: "As a follow up, all usage of SecretKey::generate in tests should be changed to use a seeded rng, like chacha8, instead of thread_rng"

So what's wrong with using thread_rng in tests? Nondeterminism, or something else. I agree that nondeterminism in tests is not great, but we have not been religious about this in the past.

dignifiedquire:
add an actually running doctest

rklaehn (approved):
Not having generate without providing a RNG is great.

iroh-base/src/key.rs
matheus23: I think it's worth taking steps towards determinism.

iroh-base/src/key.rs
```diff
@@ -334,40 +334,19 @@ impl SecretKey {
...
     /// Generate a new [`SecretKey`] with a randomness generator.
-    pub fn generate_with_rng<R: CryptoRngCore + ?Sized>(csprng: &mut R) -> Self {
-        let secret = SigningKey::generate(csprng);
+    ///
+    /// ```rust
+    /// // use the OsRng option for OS depedndent most secure RNG.
+    /// let mut rng = rand::rngs::OsRng;
+    /// let _key = iroh_base::SecretKey::generate(&mut rng);
+    /// ```
+    pub fn generate<R: CryptoRngCore>(mut csprng: R) -> Self {
```
flub, line 344: The rand crate is on version 0.8.  Is it really a good idea to require it on the public API?

iroh-base/src/relay_url.rs
```diff
@@ -36,15 +35,20 @@ impl From<Url> for RelayUrl {
     }
 }
 
+/// Can occur when parsing a string into a [`RelayUrl`].
+#[derive(Debug, thiserror::Error)]
+#[error("Failed to parse: {0}")]
+pub struct RelayUrlParsingError(#[from] url::ParseError);
```
flub, line 41: I realise this is bikeshedding, but it's somewhat surprising that this is "Pars**ing**", rather than `RelayUrlParseError`

iroh-base/src/key.rs
```diff
@@ -334,40 +334,19 @@ impl SecretKey {
...
-    }
-
     /// Generate a new [`SecretKey`] with a randomness generator.
-    pub fn generate_with_rng<R: CryptoRngCore + ?Sized>(csprng: &mut R) -> Self {
-        let secret = SigningKey::generate(csprng);
+    ///
+    /// ```rust
+    /// // use the OsRng option for OS depedndent most secure RNG.
+    /// let mut rng = rand::rngs::OsRng;
+    /// let _key = iroh_base::SecretKey::generate(&mut rng);
```
flub, line 342: From your later changes seems like you can directly call `SecretKey::generate(rand::rngs::OsRng)` which might be a little simpler as recipe?

iroh-base/src/key.rs
dignifiedquire: I think so

iroh-base/src/key.rs
dignifiedquire: I would put this into the `load_key` function, which uses this, when we update iroh-blobs

iroh-base/src/key.rs
dignifiedquire: it's only about nondeterminism why I want to do this, we already have a lot of randomness from the network, so reducing it from factors like this is usually very helpful

iroh-base/src/key.rs
dignifiedquire: all of the rust crypto world uses `rand` like this in their public interfaces.. I don't think there is much point in trying to do sth differently

iroh-base/src/relay_url.rs
dignifiedquire: I don't care at all 🤷

iroh-base/src/key.rs
flub: I kind off like the `SecretKey::from_bytes(&[1u8; 32])` approach for deterministic keys in tests.  But I agree that if you use only one key in a test I'm not sure why bother with it?

iroh-base/src/key.rs
dignifiedquire: you can, but that's not what most code will end up doing, which is why I made it two explicit steps

iroh-base/src/relay_url.rs
dignifiedquire: changed

PR #3047, opened 2024-12-13
refactor: unify hex encoding with data-encoding

dignifiedquire:
## Description

We already use `data-encoding` for `base32` and `hex` encoding, this replaces our usage of the `hex` crate with this.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


matheus23 (approved):
This is okay I think. I guess I'll be missing how the easy `hex::encode` reads, but the data-encoding version isn't too far off.
Two very minor things:
1. Can we make sure to consistently use `HEXLOWER.encode` with a `use data_encoding::HEXLOWER` import?
2. Would it make sense to add `hex` to `deny.toml`, with a comment saying "use data-encoding::HEXLOWER instead"?

rklaehn (approved):
If we didn't have to have data_encoding anyway because of the base32 stuff I would prefer to keep hex. But since we do, makes sense...

PR #3050, opened 2024-12-16
ci(*): Try use external beta workflow

flub:
## Description

Experiment with using a shared beta workflow.

## Notes & open questions

This isn't very nice to be honest.  There currently seems a rather
tight integration between the beta workflow and the tests it runs.

Maybe it's possible to replace only the notify action somehow?
E.g. if in the beta.yaml you could write:

``` yaml
jobs:
  tests:
    # as before
    uses: './.github/workflows/tests.yaml'
    with:
      rust-version: beta
  notify:
    needs: tests
    if: $${{ always() }}
    runs-on: ubuntu-latest
    steps:
      # this is new
      - uses: n0-computer/action-notify-beta@main
```

I'm not sure how realistic this is or how much work it would be.

I'm also not even sure how to test this without merging it?

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/.github/workflows/beta.yaml
+++ b/.github/workflows/beta.yaml
@@ -16,28 +16,8 @@ env:
   IROH_FORCE_STAGING_RELAYS: "1"
 
 jobs:
-  tests:
-    uses: './.github/workflows/tests.yaml'
+  beta:
+    uses: 'n0-computer/workflows/.github/workflows/beta.yaml@main'
     with:
-      rust-version: beta
-  notify:
-    needs: tests
-    if: ${{ always() }}
-    runs-on: ubuntu-latest
-    steps:
-      - name: Extract test results
-        run: |
-          printf '${{ toJSON(needs) }}\n'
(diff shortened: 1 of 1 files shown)


Arqu:
Hmmm, this is the odd ball territory of GHA.
Im not sure this will even work, maybe?
To test this out without merging you have to temporarily adjust this to run on `pull_request:` and then manually iterate on the PR. Once happy remove that so it only runs on a schedule. (Obviously before testing you need to either merge on `n0computer/workflows` or you need to adjust the branch (and push the branch too) from which it picks up the workflow. 

The benefit for the `beta` workflow specifically might be pretty limited but I'm more than happy to start expanding on this practice. We could also join the docs cleanup and beta workflows into a single scheduled run to reduce the surface area.

flub:
Meh, I'm not convinced this will lead anywhere decent and don't really have the time to experiment lots with it.  I'll leave this for now.  This means the n0-computer/workflows repo isn't used yet.  Probably it will be at some point though so I'm not going to delete it just yet.  Might push something to the README though.

PR #3051, opened 2024-12-16
refactor(iroh-base): remove automatic key caching

dignifiedquire:
## Description

- remove automatic caching of public key material
- move `crypto_box` based shared secret into `iroh` behind a private API

## Breaking Changes

- remove `iroh_base::SharedSecret`
- remove `iroh_base::DecryptionError`, `iroh::DecryptionError`
- remove `iroh_base::SecretKey::shared`

## Notes

I removed the benchmark for this, because it would require keeping this API public.
This needs a follow up PR adding an explicit key cache for the relay protocol, which needs to check keys on a per packet basis.


iroh-base/src/key.rs
```diff
@@ -162,8 +110,9 @@ impl PublicKey {
     /// a valid `ed25519_dalek` curve point. Will never fail for bytes return from [`Self::as_bytes`].
     /// See [`VerifyingKey::from_bytes`] for details.
     pub fn from_bytes(bytes: &[u8; 32]) -> Result<Self, SignatureError> {
-        get_or_create_crypto_keys(bytes, |item| item.verifying_key)?;
-        Ok(Self(*bytes))
+        let key = VerifyingKey::from_bytes(bytes)?;
```
rklaehn, line 113: Does it still make sense for the error to be VerifyingKey::SignatureError? That way we have a _public_ dependency on ed25519_dalek.

If we have this public dependency anyway, it would be nice if we had a way to get at the inner `CompressedEdwardsY` - in general provide simple conversions from our primitives to ed25519_dalek. If not, we could return our own error here.

rklaehn (approved):
There is the question whether we should publicly depend on `ed25519_dalek`, but that is unrelated to this PR.

iroh-base/src/key.rs
rklaehn: Ah, never mind. It seems we already have conversions. We didn't last time I checked this, which was some time ago.

matheus23 (approved):
Looks like moving stuff around mostly.
Checked some things to make sure the CompressedEdwardsY stuff is correct.
I'm happy to see the shared secret stuff move into `iroh`. Ideally fewer things try to use this "trick".

Arqu:
Current branch:
| iroh | 1_to_1 | 1.35 | 1.35 |
| iroh | 1_to_3 | 4.47 | 4.48 |
| iroh | 1_to_5 | 7.08 | 7.09 |
| iroh | 1_to_10 | 12.02 | 12.04 |
| iroh | 2_to_2 | 2.82 | 2.82 |
| iroh | 2_to_4 | 5.52 | 5.53 |
| iroh | 2_to_6 | 8.45 | 8.46 |
| iroh | 2_to_10 | 13.60 | 13.67 |

main:
| iroh | 1_to_1 | 1.35 | 1.35 |
| iroh | 1_to_3 | 4.23 | 4.24 |
| iroh | 1_to_5 | 7.05 | 7.06 |
| iroh | 1_to_10 | 11.64 | 11.66 |
| iroh | 2_to_2 | 2.78 | 2.79 |
| iroh | 2_to_4 | 5.46 | 5.47 |
| iroh | 2_to_6 | 8.22 | 8.23 |
| iroh | 2_to_10 | 13.57 | 13.60 |

PR #3053, opened 2024-12-16
feat(iroh-relay)!: use explicit key cache

rklaehn:
## Description

This wires up an explicit key cache to replace the implicit one that was removed in #3051.

~~The default for a key cache is Disabled. A disabled key cache has a size of 1 pointer and otherwise zero performance overhead.~~

I have removed the Default instance for both KeyCache and DerpProtocol so you don't accidentally pass the default despite having a cache available.

We use the lru crate for the cache for now. Please comment if it should be something else.

Benchmarks have shown that conversion from a [u8;32] to a VerifyingKey is relatively cheap, so the purpose of the cache is solely to validate incoming public keys.

We add a Borrow instance to PublicKey so we can use it as a cache key.

Some performance measurements:

```
Benchmarking validate valid ed25519 key: 3.7674 µs
Benchmarking validate invalid ed25519 key: 3.6637 µs
Benchmarking validate valid iroh ed25519 key: 67.089 ns
Benchmarking validate invalid iroh ed25519 key: 64.004 ns
```

So just from validating incoming keys, without cache you would be limited to ~250 000 msgs/s per thread. At a message size of 1KiB that would be 250MB/s, which is not great.

With the cache deserialization can do 14 000 000 msgs/s, which means that this is no longer a bottleneck.

## Breaking Changes

- RelayConfig has new public field key_cache_capacity

## Notes & open questions

- Size of the cache
- source_and_box has a PublicKey::try_from. Is that perf critical?

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-dns-server/src/store.rs
+++ b/iroh-dns-server/src/store.rs
@@ -34,7 +34,7 @@ pub enum PacketSource {
 
 /// A store for pkarr signed packets.
 ///
-/// Packets are stored in the persistent [`SignedPacketStore`], and cached on-demand in an in-memory LRU
+/// Packets are stored in the persistent `SignedPacketStore`, and cached on-demand in an in-memory LRU
 /// cache used for resolving DNS queries.
 #[derive(Debug, Clone)]
 pub struct ZoneStore {
--- a/iroh-net-report/Cargo.toml
+++ b/iroh-net-report/Cargo.toml
@@ -49,6 +49,7 @@ tokio = { version = "1", default-features = false, features = ["test-util"] }
 [features]
 default = ["metrics"]
 metrics = ["iroh-metrics/metrics", "portmapper/metrics"]
+stun-utils = []
 
 [package.metadata.docs.rs]
 all-features = true
--- a/iroh-net-report/src/lib.rs
+++ b/iroh-net-report/src/lib.rs
@@ -10,25 +10,25 @@
 #![cfg_attr(not(test), deny(clippy::unwrap_used))]
 
 use std::{
-    collections::{BTreeMap, HashMap},
+    collections::{BTreeMap, BTreeSet, HashMap},
     fmt::{self, Debug},
     net::{SocketAddr, SocketAddrV4, SocketAddrV6},
     sync::Arc,
 };
 
-use anyhow::{anyhow, Context as _, Result};
+use anyhow::{anyhow, Result};
 use bytes::Bytes;
 use hickory_resolver::TokioResolver as DnsResolver;
 use iroh_base::RelayUrl;
 #[cfg(feature = "metrics")]
 use iroh_metrics::inc;
 use iroh_relay::{protos::stun, RelayMap};
-use netwatch::{IpFamily, UdpSocket};
+use netwatch::UdpSocket;
 use tokio::{
     sync::{self, mpsc, oneshot},
     time::{Duration, Instant},
 };
-use tokio_util::{sync::CancellationToken, task::AbortOnDropHandle};
+use tokio_util::task::AbortOnDropHandle;
 use tracing::{debug, error, info_span, trace, warn, Instrument};
 
 mod defaults;
@@ -38,7 +38,10 @@ mod ping;
 mod reportgen;
 
 pub use metrics::Metrics;
+use reportgen::ProbeProto;
 pub use reportgen::QuicConfig;
+#[cfg(feature = "stun-utils")]
+pub use stun_utils::bind_local_stun_socket;
 
 const FULL_REPORT_INTERVAL: Duration = Duration::from_secs(5 * 60);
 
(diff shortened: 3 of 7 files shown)


iroh-relay/Cargo.toml
```diff
@@ -96,6 +96,7 @@ url = { version = "2.5", features = ["serde"] }
 webpki = { package = "rustls-webpki", version = "0.102" }
 webpki-roots = "0.26"
 data-encoding = "2.6.0"
+lru = "0.12.5"
```
flub, line 99: Purely a drive-by comment.  But I've had another reason to add an LRU cache to iroh-relay very recently.  So I think this dependency would make it here at some point.

dignifiedquire:
I think a default of maybe 128 or so would be good, as it is very small still but will improve perf for most basic endpoint usages

rklaehn:
> I think a default of maybe 128 or so would be good, as it is very small still but will improve perf for most basic endpoint usages

I made 128 the default for client and 1024 * 1024 the default for a http server. Presumably for the server 32 megabytes (or 56 megabytes if you add the 2 ptrs that LruEntry adds and the key) is no big deal, and it will work up to 1m connections.

iroh-relay/src/server/http_server.rs
```diff
@@ -172,6 +175,7 @@ impl ServerBuilder {
             handlers: Default::default(),
             headers: HeaderMap::new(),
             client_rx_ratelimit: None,
+            key_cache_capacity: 1024 * 1024,
```
dignifiedquire, line 178: maybe add a comment why this number was choosen

iroh-relay/src/key_cache.rs
```diff
@@ -0,0 +1,65 @@
...
+    Disabled,
+    /// The key cache is enabled with a fixed capacity. It is shared between
+    /// multiple threads.
+    Shared(Arc<Mutex<lru::LruCache<PublicKey, ()>>>),
+}
+
+impl KeyCache {
+    /// Key cache to be used in tests.
+    #[cfg(test)]
+    pub fn test() -> Self {
```
dignifiedquire, line 31: this is the same as `KeyCache::default()`, why is this needed?

iroh-relay/src/key_cache.rs
```diff
@@ -0,0 +1,65 @@
...
+    /// Get a key from a slice of bytes.
+    pub fn key_from_slice(&self, slice: &[u8]) -> Result<PublicKey, SignatureError> {
+        let Self::Shared(cache) = self else {
+            return PublicKey::try_from(slice);
+        };
+        let Ok(bytes) = PublicKeyBytes::try_from(slice) else {
+            // if the size is wrong, use PublicKey::try_from to fail with a
+            // SignatureError.
+            PublicKey::try_from(slice)?;
+            unreachable!();
```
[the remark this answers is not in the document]
rklaehn: PublicKey::try_from(slice)?; is just a fancy way to return an error. It is guaranteed to fail when slice is not 32 bytes long.

iroh-relay/src/key_cache.rs
rklaehn: I was looking through the code for places where KeyCache::default is used to make sure I got them all. So I renamed this to test so it is clear that it is only for tests.

iroh-relay/src/key_cache.rs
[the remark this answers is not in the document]
dignifiedquire: this reads a bit nicer
```rust
return Err(PublicKey::try_from(slice).expect_err("invalid length"));
```

iroh-relay/src/key_cache.rs
rklaehn: I have now done the same for DerpCodec and removed the Default instance. I think with Default it is too easy to accidentally have a place where the cache is not used at all.

iroh-relay/src/protos/relay.rs
```diff
@@ -204,8 +206,23 @@ pub(crate) async fn recv_client_key<S: Stream<Item = anyhow::Result<Frame>> + Un
...
-#[derive(Debug, Default, Clone)]
-pub(crate) struct DerpCodec;
+#[derive(Debug, Clone)]
+pub(crate) struct DerpCodec {
+    cache: KeyCache,
+}
+
+impl DerpCodec {
+    #[cfg(test)]
+    pub fn test() -> Self {
```
dignifiedquire, line 216: `pub(crate)`?

dignifiedquire:
I am having a hard time finding the place this is configured for a non server endpoint, can you point me to that?

rklaehn:
> I am having a hard time finding the place this is configured for a non server endpoint, can you point me to that?

It is configurable using the ClientBuilder, with a default of 128 as you suggested.

```rust
  /// Create a new [`ClientBuilder`]
  pub fn new(url: impl Into<RelayUrl>) -> Self {
      ClientBuilder {
          is_preferred: false,
          address_family_selector: None,
          is_prober: false,
          server_public_key: None,
          url: url.into(),
          protocol: Protocol::Relay,
          insecure_skip_cert_verify: false,
          proxy_url: None,
          key_cache_capacity: 128,
      }
  }
...


    /// Set the capacity of the cache for public keys.
    pub fn key_cache_capacity(mut self, capacity: usize) -> Self {
        self.key_cache_capacity = capacity;
        self
    }
  ```

dignifiedquire (approved):

dignifiedquire:
> Is that not the right place?

It is, I was just being blind

PR #3056, opened 2024-12-17
feat(relay)!: relay only mode now configurable

Arqu:
## Description

This will allow us to configure relay only mode at runtime. ~~Useful for easy testing without having to rebuild with the `DEV_RELAY_ONLY` environment variable set.~~

The `DEV_RELAY_ONLY` compile time env var has been completely dropped and a `relay_only` option has been threaded through the entire stack when `test-utils` is enabled. An example of it can be followed with the `iroh/examples/transfer.rs` where you can set up a provide and fetch node with `--relay-only`. 

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/bench/src/lib.rs
+++ b/iroh/bench/src/lib.rs
@@ -71,6 +71,7 @@ pub struct Opt {
     /// Whether to run a local relay and have the server and clients connect to that.
     /// This will force all traffic over the relay and can be used to test
     /// throughput for relay-only traffic.
+    #[cfg(feature = "local-relay")]
     #[clap(long, default_value_t = false)]
     pub only_relay: bool,
 }
--- a/iroh/bench/src/lib.rs
+++ b/iroh/bench/src/lib.rs
@@ -70,9 +70,8 @@ pub struct Opt {
     pub initial_mtu: u16,
     /// Whether to run a local relay and have the server and clients connect to that.
     ///
-    /// Can be combined with the `DEV_RELAY_ONLY` environment variable (at compile time)
+    /// Can be combined with the `relay_only` configuration option on the endpoint
     /// to test throughput for relay-only traffic locally.
-    /// (e.g. `DEV_RELAY_ONLY=true cargo run --release -- iroh --with-relay`)
     #[clap(long, default_value_t = false)]
     pub with_relay: bool,
 }
--- a/iroh/bench/src/lib.rs
+++ b/iroh/bench/src/lib.rs
@@ -69,11 +69,10 @@ pub struct Opt {
     #[clap(long, default_value = "1200")]
     pub initial_mtu: u16,
     /// Whether to run a local relay and have the server and clients connect to that.
-    ///
-    /// Can be combined with the `relay_only` configuration option on the endpoint
-    /// to test throughput for relay-only traffic locally.
+    /// This will force all traffic over the relay and can be used to test
+    /// throughput for relay-only traffic.
     #[clap(long, default_value_t = false)]
-    pub with_relay: bool,
+    pub only_relay: bool,
 }
 
 pub enum EndpointSelector {
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -76,6 +76,7 @@ type DiscoveryBuilder = Box<dyn FnOnce(&SecretKey) -> Option<Box<dyn Discovery>>
 
 /// Defines the mode of path selection for all traffic flowing through
 /// the endpoint.
+#[cfg(any(test, feature = "test-utils"))]
 #[derive(Debug, Default, Copy, Clone, PartialEq, Eq)]
 pub enum PathSelection {
     /// Uses all available paths
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -164,8 +164,6 @@ impl Builder {
             dns_resolver,
             #[cfg(any(test, feature = "test-utils"))]
             insecure_skip_relay_cert_verify: self.insecure_skip_relay_cert_verify,
-            #[cfg(any(test, feature = "test-utils"))]
-            relay_only: self.relay_only,
         };
         Endpoint::bind(static_config, msock_opts, self.alpn_protocols).await
     }
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -97,6 +97,8 @@ pub struct Builder {
     insecure_skip_relay_cert_verify: bool,
     addr_v4: Option<SocketAddrV4>,
     addr_v6: Option<SocketAddrV6>,
+    #[cfg(any(test, feature = "test-utils"))]
+    relay_only: bool,
 }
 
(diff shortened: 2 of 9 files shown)


dignifiedquire:
please also remove the old env variable, to simplify things

flub (commented):
I think it is a good move that this is still `cfg(any(test, feature = "test-utils"))`.  I think this generally looks fine?

iroh/bench/src/lib.rs
```diff
@@ -70,9 +70,8 @@ pub struct Opt {
     pub initial_mtu: u16,
     /// Whether to run a local relay and have the server and clients connect to that.
     ///
-    /// Can be combined with the `DEV_RELAY_ONLY` environment variable (at compile time)
+    /// Can be combined with the `relay_only` configuration option on the endpoint
     /// to test throughput for relay-only traffic locally.
-    /// (e.g. `DEV_RELAY_ONLY=true cargo run --release -- iroh --with-relay`)
     #[clap(long, default_value_t = false)]
     pub with_relay: bool,
```
matheus23, line 76: Thanks for adjusting the docs here, but these docs are actually the CLI help text.
So saying "hey you can change the endpoint config" doesn't really help when you're reading this as CLI text :grimacing: 

I'd appreciate if there'd be two possible solutions:
1. Add another `pub only_relay: bool` option, so you can run `cargo run --release -- iroh --only-relay` (and `only-relay` should imply `with-relay`) or
2. Rename `with_relay` to `only_relay` and make that also set `.relay_only(true)` on the endpoint.

Arqu: Went with option 2 but am not happy about it as it depends on the test/test-utils feature to be active. I'll revert it to option 1.

matheus23 (approved):

iroh/bench/src/lib.rs
matheus23: Nice!
Only other thing that might be nice is documenting that you need the `local-relay` feature to be active (and telling the user how to run this thing, so `cargo run --release --features=local-relay -- iroh --only-relay`).

Arqu:
FWIW this PR is nowhere ready to be merged, the pieces checking and utilizing the relay_only flag are disconnected from the endpoint and never get passed in. This was circumvented by just dropping in that env var, but had some random loose bits. I managed to clean up the confusing bits, but need to thread the needle a bit more.

iroh/src/endpoint.rs
```diff
@@ -417,6 +423,14 @@ impl Builder {
         self.insecure_skip_relay_cert_verify = skip_verify;
         self
     }
+
+    /// This implies we only use the relay to communicate
+    /// and do not attempt to do any hole punching.
+    #[cfg(any(test, feature = "test-utils"))]
+    pub fn relay_only(mut self, relay_only: bool) -> Self {
```
rklaehn, line 430: It is not the end of the world since this is a debug feature, but I would probably favour an enum to configure which path to use instead of a bool. E.g. we might in the future want to have more complex logic to select which path to take (IPV4/IPV6/Relay)

Arqu: I guess that will make more sense to implement once we start dealing with multipath as we currently don't really have an idea of everything we will need to pass in.

rklaehn: Make it an enum with 2 cases, then it is still easier to extend than a bool

rklaehn:
Any particular reason we need the feature flag dance?

```rust
```

I think in none of the places where this is used a flag check is going to be a performance problem. So let's make the path use (for now just RelayOnly and All) an enum, set it to All by default, and have this enabled even without feature flags.

Arqu:
> Any particular reason we need the feature flag dance?
> 
> ```rust
> #[cfg(any(test, feature = "test-utils"))]
> ```
> 
> I think in none of the places where this is used a flag check is going to be a performance problem. So let's make the path use (for now just RelayOnly and All) an enum, set it to All by default, and have this enabled even without feature flags.

We generally wanted to avoid exposing this fluff to end users without specific reason as you generally should never want to do relay only. Im fine with doing that, simplifies the code a nice bit. @dignifiedquire  you wanted to have it feature flagged initially, anything against it?

rklaehn:
> We generally wanted to avoid exposing this fluff to end users without specific reason as you generally should never want to do relay only. Im fine with doing that, simplifies the code a nice bit. @dignifiedquire you wanted to have it feature flagged initially, anything against it?

I can see why we don't want users to mess with this, but having lots of code gated by feature flags is not nice.

So either mark the builder fn to set this with docs(hidden), or have _just the public fn to set this_ feature gated?

dignifiedquire:
> or have _just the public fn to set this_ feature gated?

that usually leads to a bunch of dead code warnings

Arqu:
This is now at a middle ground using and enum for path selection. 
What's the verdict at the end, to feature flag or not?

matheus23 (approved):
Needs some CI massage.

1. Keeping this behind the `test_utils` feature flag sets expectations. IMO important ones. Generally we want people to avoid enabling relay-only mode.
2. If someone *really* wants relay-only mode, it's just a feature flag away. It's not dangerous to enable `test_utils`. I'll totally steal this for e.g. iroh-doctor for testing stuff on mobile.

iroh/src/endpoint.rs
```diff
@@ -74,6 +74,17 @@ const DISCOVERY_WAIT_PERIOD: Duration = Duration::from_millis(500);
 
 type DiscoveryBuilder = Box<dyn FnOnce(&SecretKey) -> Option<Box<dyn Discovery>> + Send + Sync>;
 
+/// Defines the mode of path selection for all traffic flowing through
+/// the endpoint.
+#[derive(Debug, Default, Copy, Clone, PartialEq, Eq)]
+pub enum PathSelection {
```
dignifiedquire, line 80: shouldn't this be cfged as well?

iroh/src/magicsock/node_map.rs
```diff
@@ -123,11 +127,18 @@ pub enum Source {
...
 impl NodeMap {
+    #[cfg(not(any(test, feature = "test-utils")))]
     /// Create a new [`NodeMap`] from a list of [`NodeAddr`]s.
     pub(super) fn load_from_vec(nodes: Vec<NodeAddr>) -> Self {
         Self::from_inner(NodeMapInner::load_from_vec(nodes))
     }
 
+    #[cfg(any(test, feature = "test-utils"))]
+    /// Create a new [`NodeMap`] from a list of [`NodeAddr`]s.
+    pub(super) fn load_from_vec(nodes: Vec<NodeAddr>, path_selection: PathSelection) -> Self {
```
dignifiedquire, line 138: I would prefer to make this a seperate method, like `set_path_selection`, instead of changing the signature

Arqu: Issue is, the inner requires the flag and it needs to be provided at load time. Should I keep it or move forward and unpack the whole thing?

PR #3058, opened 2024-12-17
refactor(iroh, iroh-relay)!: Do reconnection in ActiveRelayActor

flub:
## Description

This moves the reconnection logic ingo ActiveRelayActor instead of
doing this in the RelayActor.  The logic itself is not really changed.
This does simplify things a fair bit though, it is a lot easier to do
this in one place where all the state needs to be.

In the grand scheme of things this moves "logical" control over the
relay Client entirely into the ActiveRelayActor.

Some small side effects:

- The ActiveRelayMessage::Ping and ActiveRelayMessage::GetLocalAddr
  messages are no longer needed.  Though I decided to keep the latter
  for testing purposes.

- ActiveRelayMessage::NotePreferred has been rename to SetHomeRelay.
  Since the ActiveRelayActor now reconnects to relays on its own it
  needs to keep track of whether it is the home relay itself as well.

- The duplicate ping task tracking in the RelayActor is removed.  This
  is still not ideal in the ActiveRelayActor because it still blocks
  this actor.  But that's no regression and to be addressed when the
  relay Client is no longer an actor.

- RelayActor::close_and_reconnect_relay and
  RelayActor::send_to_active_relay are no longer needed.  Because we
  await all the ActiveRelayActor tasks already we do not need to
  manually track them in case they need to be closed.

- The ClientBuilder was no longer using is_preferred, so remove this.

### iroh-relay

- `ClientBuilder::is_preferred` no longer exists.  Set the preference once the client exists instead.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/magicsock/relay_actor.rs
+++ b/iroh/src/magicsock/relay_actor.rs
@@ -1086,7 +1086,8 @@ mod tests {
         let recv = future::poll_fn(|cx| datagram_recv_queue.poll_recv(cx)).await?;
         assert_eq!(recv.buf.as_ref(), b"hello");
 
-        // Now ask to check the connection, this will reconnect.
+        // Now ask to check the connection, this will reconnect without pinging because we
+        // do not supply any "valid" local IP addresses.
         info!("check connection");
         inbox_tx
             .send(ActiveRelayMessage::CheckConnection(Vec::new()))
--- a/iroh/src/magicsock/relay_actor.rs
+++ b/iroh/src/magicsock/relay_actor.rs
@@ -82,7 +82,7 @@ struct ActiveRelayActor {
 enum ActiveRelayMessage {
     GetLastWrite(oneshot::Sender<Instant>),
     /// Returns whether or not this relay can reach the NodeId.
-    GetNodeRoute(NodeId, oneshot::Sender<bool>),
+    HasNodeRoute(NodeId, oneshot::Sender<bool>),
     /// Triggers a connection check to the relay server.
     ///
     /// Sometimes it is known the local network interfaces have changed in which case it
@@ -243,7 +243,7 @@ impl ActiveRelayActor {
                 self.is_home_relay = is_preferred;
                 self.relay_client.note_preferred(is_preferred).await;
             }
-            ActiveRelayMessage::GetNodeRoute(peer, r) => {
+            ActiveRelayMessage::HasNodeRoute(peer, r) => {
                 let has_peer = self.node_present.contains(&peer);
                 r.send(has_peer).ok();
             }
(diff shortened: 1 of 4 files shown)


iroh/src/magicsock/relay_actor.rs
```diff
@@ -72,11 +81,22 @@ struct ActiveRelayActor {
 #[allow(clippy::large_enum_variant)]
 enum ActiveRelayMessage {
     GetLastWrite(oneshot::Sender<Instant>),
-    Ping(oneshot::Sender<Result<Duration, ClientError>>),
-    GetLocalAddr(oneshot::Sender<Option<SocketAddr>>),
+    /// Returns whether or not this relay can reach the NodeId.
     GetNodeRoute(NodeId, oneshot::Sender<bool>),
```
dignifiedquire, line 85: should this be renamed to `HasNodeRoute`?

iroh/src/magicsock/relay_actor.rs
```diff
@@ -964,4 +951,163 @@ mod tests {
...
+        // Echo should still work.
+        info!("second echo");
+        let packet = PacketizeIter::<_, MAX_PAYLOAD_SIZE>::new(peer_node, [b"hello"])
+            .next()
+            .context("no packet")?;
+        send_datagram_tx.send(packet).await?;
+        let recv = future::poll_fn(|cx| datagram_recv_queue.poll_recv(cx)).await?;
+        assert_eq!(recv.buf.as_ref(), b"hello");
+
+        // Now ask to check the connection, this will reconnect.
```
dignifiedquire, line 1089: why does this reconnect? when was the connection dropped?

iroh/src/magicsock/relay_actor.rs
flub: Thanks, I needed to do this in the last PR but didn't find a good name.  But that suggestion is great.

iroh/src/magicsock/relay_actor.rs
flub: Because the `CheckConnection` message does not contain any local IPs the ActiveRelayActor will consider the local addr of the connection to be no longer available on the host so automatically reconnect without first pinging.

The ActiveRelayActor itself now handles the reconnect (in the new `reconnect` method) and simply drops the previous client.

What actually happens is that the drop might not happen soon enough, depending on async runtime scheduling, and the relay server sees two connections and closes the older one.  But sometimes the drop happens first and the relay connection sees the connection being closed (abruptly).

dignifiedquire: maybe add a short comment so
we don’t forget

dignifiedquire (approved):

PR #3060, opened 2024-12-18
refactor(iroh): magical pin projections

dignifiedquire:
## Description

Avoids the `Box::pin` inside of the loop

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/magicsock/relay_actor.rs
+++ b/iroh/src/magicsock/relay_actor.rs
@@ -182,8 +182,7 @@ impl ActiveRelayActor {
         // When this future has an inner, it is a future which is currently sending
         // something to the relay server.  Nothing else can be sent to the relay server at
         // the same time.
-        let mut relay_send_fut = MaybeFuture::none();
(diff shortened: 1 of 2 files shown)


iroh/src/magicsock/relay_actor.rs
```diff
@@ -182,7 +182,7 @@ impl ActiveRelayActor {
         // When this future has an inner, it is a future which is currently sending
         // something to the relay server.  Nothing else can be sent to the relay server at
         // the same time.
-        let mut relay_send_fut = MaybeFuture::none();
+        let mut relay_send_fut = std::pin::pin!(MaybeFuture::default());
```
flub, line 185: FWIW I do prefer the `::none()` constructor over `::default()`.  It feels more coherent with `.set_none()`, `.is_none()` etc.  But I can live with `::default()` if you strongly prefer it.

flub (approved):

PR #3061, opened 2024-12-18
refactor(iroh): ActiveRelayActor terminates itself

flub:
## Description

This moves the logic of terminating an ActiveRelayActor from the
RelayActor to the ActiveRelayActor itself.  This is nice for two
reasons:

- The interactions between ActiveRelayActor and RelayActor are even
  simpler.  RelayActor logic is now easy to reason about.

- It will allow ActiveRelayActor to manage reconnections with
  exponential backoff better.  Currently this behaviour is not changed
  but the RelayActor getting involved meant the backoff behaviour was
  influcenced in two places.

Now that the ActiveRelayActor lifecyle is much more straight forward
the RelayActor's terminating is tidied up as well.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/magicsock/relay_actor.rs
+++ b/iroh/src/magicsock/relay_actor.rs
@@ -12,7 +12,6 @@ use std::{
         atomic::{AtomicBool, Ordering},
         Arc,
     },
-    time::{Duration, Instant},
 };
 
 use anyhow::Context;
@@ -26,7 +25,7 @@ use iroh_relay::{self as relay, client::ClientError, ReceivedMessage, MAX_PACKET
 use tokio::{
     sync::{mpsc, oneshot},
     task::JoinSet,
-    time,
+    time::{self, Duration, Instant},
 };
 use tokio_util::sync::CancellationToken;
 use tracing::{debug, error, info, info_span, trace, warn, Instrument};
@@ -41,9 +40,6 @@ use crate::{
 /// How long a non-home relay connection needs to be idle (last written to) before we close it.
 const RELAY_INACTIVE_CLEANUP_TIME: Duration = Duration::from_secs(60);
 
-/// How often `clean_stale_relay` runs when there are potentially-stale relay connections to close.
-const RELAY_CLEAN_STALE_INTERVAL: Duration = Duration::from_secs(15);
-
 /// Maximum size a datagram payload is allowed to be.
 const MAX_PAYLOAD_SIZE: usize = MAX_PACKET_SIZE - PublicKey::LENGTH;
 
@@ -53,9 +49,6 @@ const MAX_PAYLOAD_SIZE: usize = MAX_PACKET_SIZE - PublicKey::LENGTH;
 /// communication with it.
 #[derive(Debug)]
 struct ActiveRelayActor {
-    /// The time of the last request for its write
-    /// channel (currently even if there was no write).
-    last_write: Instant,
     /// Queue to send received relay datagrams on.
     relay_datagrams_recv: Arc<RelayDatagramsQueue>,
     /// Channel on which we receive packets to send to the relay.
@@ -80,7 +73,6 @@ struct ActiveRelayActor {
 #[derive(Debug)]
 #[allow(clippy::large_enum_variant)]
 enum ActiveRelayMessage {
-    GetLastWrite(oneshot::Sender<Instant>),
     /// Returns whether or not this relay can reach the NodeId.
     HasNodeRoute(NodeId, oneshot::Sender<bool>),
     /// Triggers a connection check to the relay server.
@@ -131,7 +123,6 @@ impl ActiveRelayActor {
             Self::create_relay_client(url.clone(), connection_opts.clone());
 
         ActiveRelayActor {
-            last_write: Instant::now(),
             relay_datagrams_recv,
             relay_datagrams_send,
             url,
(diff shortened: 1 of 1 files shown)


iroh/src/magicsock/relay_actor.rs
```diff
@@ -600,19 +592,21 @@ impl RelayActor {
...
-                handle
+                if let Err(err) = handle
                     .inbox_addr
-                    .send(ActiveRelayMessage::SetHomeRelay(true))
-                    .await
-                    .ok();
+                    .try_send(ActiveRelayMessage::SetHomeRelay(true))
+                {
+                    error!("Send to newly started active relay failed: {err:#}.");
+                    warn!("Home relay not set");
```
dignifiedquire, line 608: why to log messages?

flub: Not sure...
- One is to explain what the bug is.
- The other explains what the impact is.

I was very close to calling `.expect()` on this to let it panic.  I think it probably is a bug if you can't send to the inbox right after creating it (and it was nicer to have this method not be async).  But I went with very noisy logging instead.

iroh/src/magicsock/relay_actor.rs
```diff
@@ -213,7 +209,7 @@ impl ActiveRelayActor {
                         relay_client.send(msg.node_id, msg.packet).await
                     };
                     relay_send_fut.as_mut().set_future(fut);
-                    self.last_write = Instant::now();
+                    inactive_timeout.as_mut().reset(Instant::now() + RELAY_INACTIVE_CLEANUP_TIME);
```
dignifiedquire, line 212: not sure, but you could use `Interval` and call `reset()` here

iroh/src/magicsock/relay_actor.rs
```diff
@@ -600,19 +592,20 @@ impl RelayActor {
...
         }
         let url = found_relay.as_ref().unwrap_or(url);
-        self.active_relay_handle(url).await
+        self.active_relay_handle(url)
     }
 
     /// Returns the handle of the [`ActiveRelayActor`].
-    async fn active_relay_handle(&mut self, url: &RelayUrl) -> &ActiveRelayHandle {
+    fn active_relay_handle(&mut self, url: &RelayUrl) -> &ActiveRelayHandle {
         if !self.active_relays.contains_key(url) {
```
dignifiedquire, line 600: this could probably be written nicer and more efficiently using the `.entry()` API

iroh/src/magicsock/relay_actor.rs
```diff
@@ -676,75 +669,41 @@ impl RelayActor {
...
-                .await
-                .ok();
-            match rx.await {
-                Ok(last_write) if last_write.duration_since(now) <= RELAY_INACTIVE_CLEANUP_TIME => {
-                    None
-                }
-                _ => Some(url.clone()),
+    /// Cleans up [`ActiveRelayActor`]s which have stopped running.
+    fn reap_active_relays(&mut self) {
+        let mut stopped_actors = Vec::with_capacity(self.active_relays.len());
```
dignifiedquire, line 674: you don't have any `await`s in here, so I think you could just use `self.active_relays.retain`

iroh/src/magicsock/relay_actor.rs
flub: Problem is that `self.start_active_relay()` still takes `&mut self` because it does `self.active_relay_tasks.spawn()` and that `spawn` needs a mutable reference to the `JoinSet`.  Even moving that JoinSet to an argument of `start_active_relay` won't help.

iroh/src/magicsock/relay_actor.rs
flub: whoa, retain is nice!

iroh/src/magicsock/relay_actor.rs
flub: Yeah ok, it's a close call because interval is a little weird here because you only ever use the first tick.  But the code does look simpler when using interval so let's go with it.

iroh/src/magicsock/relay_actor.rs
flub: I guess you could put the `JoinSet` in an `Arc<Mutex<JoinSet<_>>>` and that mutex would be un-contended anyway because this is an actor and we get the joy of `&mut self`-exclusivity.  But that is a lot of kludge and moving complexity to the actor struct itself.  Here at least the kludge stays local.

flub:
oops, seems the interval change didn't work...

flub:
> oops, seems the interval change didn't work...

Fixed.  It kind of shows why I didn't use this the first time round: a lot more complex.

dignifiedquire:
> Fixed. It kind of shows why I didn't use this the first time round: a lot more complex.

feel free to go back if you think it is simpler

flub:
> > Fixed. It kind of shows why I didn't use this the first time round: a lot more complex.
> 
> feel free to go back if you think it is simpler

It's ok: tests caught it and it still leaves the code clearer to read.  I don't think this should be a maintenance burden as this doesn't interact with anything else the actor does.

iroh/src/magicsock/relay_actor.rs
```diff
@@ -184,6 +175,12 @@ impl ActiveRelayActor {
         // the same time.
         let mut relay_send_fut = std::pin::pin!(MaybeFuture::none());
 
+        // If inactive for one tick the actor should exit.  Inactivity is only tracked on
+        // the last datagrams sent to the relay, received datagrams will trigger ACKs which
+        // is sufficient to keep active connections open.
+        let mut inactive_timeout = tokio::time::interval(RELAY_INACTIVE_CLEANUP_TIME);
+        inactive_timeout.reset(); // skip immediate tick
```
dignifiedquire, line 182: wouldn't using `interval_at` be cleaner? (honest question, not sure it does what you need here)

flub: probably not, gets very verbose again and then i'd prefer the original sleep probably.  going to leave this as-is.

PR #3062, opened 2024-12-18
fix(iroh, iroh-relay)!: Optimise the relay datagram path through the MagicSock

### LINKED ISSUE #2951: iroh-net: regression: blob downloads freeze at a low percentage completed
My iroh network started to see downloads not complete after 0.27.
we bisected the problematic change down to https://github.com/n0-computer/iroh/pull/2782 
this PR caused about 1/16 of our blob downloads to freeze & fail, reliably.
I've sent trace logs over to @b5 via `sendme`  - they're a few gigs in size.

0.28 did not fix this issue, though https://github.com/n0-computer/iroh/issues/2876 was marked as potential fixer.
This issue https://github.com/n0-computer/iroh/issues/2852 might be related, since it was opened around the same time.

### LINKED ISSUE #2971: relay-related issues connecting many client `Endpoint`s to one server `Endpoint`
Not 100% sure what to title this, I don't know if the issue was on the `Endpoint` side or the relay server side.

I had about 10,000 clients trying to connect to a server at a rate of ~300/min, and then sending <1kb every 2 sec. On my server `Endpoint` side I was seeing a lot of `try_send: iroh_net::magicsock: send relay: message dropped, channel to actor is full`. When I restarted the server, which would result in all the clients trying to connect at once, which resulted in a wall of that log:

```
2024-11-25T22:05:57.709301Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: send relay: message dropped, channel to actor is full node=z4wj5ubvyixfn6jw relay_url=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.709688Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: send relay: message dropped, channel to actor is full node=z4wj5ubvyixfn6jw relay_url=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.709717Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: relay channel full, dropping call-me-maybe dstkey=z4wj5ubvyixfn6jw relayurl=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.709760Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: send relay: message dropped, channel to actor is full node=unzs4fhhvzsxvaq5 relay_url=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.710196Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: send relay: message dropped, channel to actor is full node=unzs4fhhvzsxvaq5 relay_url=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.710228Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: relay channel full, dropping call-me-maybe dstkey=unzs4fhhvzsxvaq5 relayurl=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.710266Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: send relay: message dropped, channel to actor is full node=p6yde6zwtrpwerqw relay_url=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.710839Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: send relay: message dropped, channel to actor is full node=p6yde6zwtrpwerqw relay_url=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.710867Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: relay channel full, dropping call-me-maybe dstkey=p6yde6zwtrpwerqw relayurl=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.710903Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: send relay: message dropped, channel to actor is full node=fstlzagjalkbnymo relay_url=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.711368Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: send relay: message dropped, channel to actor is full node=fstlzagjalkbnymo relay_url=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.711420Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: relay channel full, dropping call-me-maybe dstkey=fstlzagjalkbnymo relayurl=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.711457Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: send relay: message dropped, channel to actor is full node=4poqvd4kppki6dao relay_url=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.711959Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: send relay: message dropped, channel to actor is full node=4poqvd4kppki6dao relay_url=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.711990Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: relay channel full, dropping call-me-maybe dstkey=4poqvd4kppki6dao relayurl=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.712028Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: send relay: message dropped, channel to actor is full node=yzewz7orsbjmwlpx relay_url=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.712434Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: send relay: message dropped, channel to actor is full node=yzewz7orsbjmwlpx relay_url=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.712463Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: relay channel full, dropping call-me-maybe dstkey=yzewz7orsbjmwlpx relayurl=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.712499Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: send relay: message dropped, channel to actor is full node=5t44ivrezsrhw4es relay_url=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.713075Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: send relay: message dropped, channel to actor is full node=5t44ivrezsrhw4es relay_url=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.713104Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: relay channel full, dropping call-me-maybe dstkey=5t44ivrezsrhw4es relayurl=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.713247Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: send relay: message dropped, channel to actor is full node=ikm6sqdfrrxlrntl relay_url=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.713541Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: send relay: message dropped, channel to actor is full node=ikm6sqdfrrxlrntl relay_url=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.713596Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: relay channel full, dropping call-me-maybe dstkey=ikm6sqdfrrxlrntl relayurl=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.713633Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: send relay: message dropped, channel to actor is full node=2g6jdbviav6t3vxv relay_url=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.713976Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: send relay: message dropped, channel to actor is full node=2g6jdbviav6t3vxv relay_url=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.714006Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: relay channel full, dropping call-me-maybe dstkey=2g6jdbviav6t3vxv relayurl=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.714041Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: send relay: message dropped, channel to actor is full node=sygycyckomhlhyvo relay_url=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.714280Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: send relay: message dropped, channel to actor is full node=sygycyckomhlhyvo relay_url=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.714307Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: relay channel full, dropping call-me-maybe dstkey=sygycyckomhlhyvo relayurl=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.714344Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: send relay: message dropped, channel to actor is full node=6lb527ubwgqk6ykn relay_url=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.714676Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: send relay: message dropped, channel to actor is full node=6lb527ubwgqk6ykn relay_url=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.714704Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: relay channel full, dropping call-me-maybe dstkey=6lb527ubwgqk6ykn relayurl=xxx me=vr6nx5o36gts5kqw
2024-11-25T22:05:57.714730Z  WARN ep:magicsock:actor:handle_ping_actions: iroh_net::magicsock: send relay: message dropped, channel to actor is full node=bv7ti723wdm7bgck relay_url=xxx me=vr6nx5o36gts5kqw
etc...
```

on the relay server side, there were a lot of logs of
```
2024-11-25T21:40:49.528804Z  WARN iroh_relay::server::actor: no way to reach client, dropped packet dst=PublicKey(ysbnyw5olxxtxuzt)
2024-11-25T21:40:49.529038Z  WARN iroh_relay::server::actor: no way to reach client, dropped packet dst=PublicKey(ysbnyw5olxxtxuzt)
2024-11-25T21:40:49.529549Z  WARN iroh_relay::server::actor: no way to reach client, dropped packet dst=PublicKey(ysbnyw5olxxtxuzt)
2024-11-25T21:40:49.529561Z  WARN iroh_relay::server::actor: no way to reach client, dropped packet dst=PublicKey(ysbnyw5olxxtxuzt)
2024-11-25T21:40:49.529565Z  WARN iroh_relay::server::actor: no way to reach client, dropped packet dst=PublicKey(ysbnyw5olxxtxuzt)
2024-11-25T21:40:49.529570Z  WARN iroh_relay::server::actor: no way to reach client, dropped packet dst=PublicKey(ysbnyw5olxxtxuzt)
2024-11-25T21:40:49.529573Z  WARN iroh_relay::server::actor: no way to reach client, dropped packet dst=PublicKey(ysbnyw5olxxtxuzt)
```

with the `PublicKey` being that of my server-`Endpoint` at the time.

There were some clients able to connect and get requests through successfully, but most failed trying to connect

### LINKED ISSUE #3008: The relay send path of an `Endpoint` is too easily congested.
I'm occasionally seeing `Endpoint`s get into a state where they're unable to connect to any other nodes, and the logs start showing this:

```
2024-12-04 04:31:18 -0500 EST debug "send relay: message dropped, channel to actor is full"
2024-12-04 04:31:18.002 -0500 EST debug "relay channel full, dropping call-me-maybe"
2024-12-04 04:31:19.926 -0500 EST debug "send relay: message dropped, channel to actor is full"
2024-12-04 04:31:19.928 -0500 EST debug "send relay: message dropped, channel to actor is full"
2024-12-04 04:31:19.929 -0500 EST debug "relay channel full, dropping call-me-maybe"
2024-12-04 04:31:19.931 -0500 EST debug "send relay: message dropped, channel to actor is full"
2024-12-04 04:31:20.926 -0500 EST debug "send relay: message dropped, channel to actor is full"
2024-12-04 04:31:20.929 -0500 EST debug "no UDP or relay paths available for node"
2024-12-04 04:31:20.93 -0500 EST debug "send relay: message dropped, channel to actor is full"
2024-12-04 04:31:20.932 -0500 EST debug "no UDP or relay paths available for node"
2024-12-04 04:31:22.924 -0500 EST debug "send relay: message dropped, channel to actor is full"
2024-12-04 04:31:22.925 -0500 EST debug "no UDP or relay paths available for node"
2024-12-04 04:31:22.926 -0500 EST debug "send relay: message dropped, channel to actor is full"
2024-12-04 04:31:22.927 -0500 EST debug "no UDP or relay paths available for node"
2024-12-04 04:31:22.999 -0500 EST debug "send relay: message dropped, channel to actor is full"
2024-12-04 04:31:23.001 -0500 EST debug "send relay: message dropped, channel to actor is full"
2024-12-04 04:31:23.003 -0500 EST debug "relay channel full, dropping call-me-maybe"
2024-12-04 04:31:23.004 -0500 EST debug "send relay: message dropped, channel to actor is full"
2024-12-04 04:31:23.006 -0500 EST debug "send relay: message dropped, channel to actor is full"
2024-12-04 04:31:23.007 -0500 EST debug "relay channel full, dropping call-me-maybe"
```

This continues for a few minutes and never seems to recover until the `Endpoint` is dropped and a new one is created, at which point things work fine again.

In this particular case the client was in India, connecting to a relay in the US (because we're only running one relay). Not sure if latency to the relay could affect this. I have seen this happen with a client from Canada so I don't think it would be exclusively due to high latency

flub:
## Description

This refactors how datagrams flow from the MagicSock (AsyncUdpSocket) to relay server and back.  It also vastly simplifies the actors involved in communicating with a relay server.

- The `RelayActor` managed all connections to relay servers.
  - It starts a new `ActiveRelayActor` for each relay server needed.
  - The `ActiveRelayActor` will exit when unused.
    - Unless it is for the home relay, this one never exits.
  - Each `ActiveRelayActor` uses a relay `Client`.
    - The relay `Client` is now a `Stream` and `Sink` directly connected to the `TcpStream` connected to the relay server.  This eliminates several actors previously used here in the `Client` and `Conn`.
  - Each `ActiveRelayActor` will try and maintain a connection with the relay server.
    - If connections fail, exponential backoff is used for reconnections.
- When `AsyncUdpSocket` needs to send datagrams:
  - It (now) puts them on a queue to the `RelayActor`.
  - The `RelayActor` ensures the correct `ActiveRelayActor` is running and forwards datagrams to it.
  - The `ActiveRelayActor` sends datagrams directly to the relay server.
- The relay receive path is now:
  - Whenever `ActiveRelayActor` is connected it reads from the underlying `TcpStream`.
  - Received datagrams are placed on an mpsc channel that now bypasses the `RelayActor` and goes straight to the `AsyncUpdSocket` interface.

Along the way many bugs are fixed.  Some of them:

- The relay datagrams send and receive queue now behave more correctly when they are full.  So the `AsyncUdpSocket` behaves better.
  - Though there still is a bug with the send queue not waking up all the tasks that might be waiting to send.  This needs a followup: #3067.
- The `RelayActor` now avoids blocking.  This means it can still react to events when the datagrams queues are full and reconnect relay servers etc as needed to unblock.
- The `ActiveRelayActor` also avoids blocking.  Allowing it to react to connection breakage and the need to reconnect at any time.
- The `ActiveRelayActor` now correctly handles connection errors and retries with backoff.
- The `ActiveRleayActor` will no longer queue unsent datagrams forever, but flush them every 400ms.
  - This also stops the send queue into the `RelayActor` from completely blocking.

### iroh-relay

- `Conn` is no longer public.
- The `Client` is completely changed.  See module docs.

## Notes & open questions

- Potentially the relay `Conn` and `Client` don't need to be two separate things now? Though Client is convenient as it only implements one Sink interface, while Conn is also a Frame sink. This means on Conn you often have to use very annoying syntax when calling things like .flush() or .close() etc.
- Maybe a few items from the `ActiveRelayActor` can be moved back into the relay `Client`, though that would probably require some gymnastics.  The current structure of `ActiveRelayActor` is fairly reasonable and handles things correctly.  Though it does have a lot of client knowledge baked in.  Being able to reason about the client as a stream + sink is what enabled me to write the good `ActiveRelayActor` though, so I'm fairly happy that this code makes sense as it is.

If all goes well this should:
Closes #3008 
Closes #2971 
Closes #2951

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-relay/src/client.rs
+++ b/iroh-relay/src/client.rs
@@ -68,12 +68,12 @@ pub struct ClientBuilder {
     insecure_skip_cert_verify: bool,
     /// HTTP Proxy
     proxy_url: Option<Url>,
-    /// Capacity of the key cache
-    key_cache_capacity: usize,
     /// The secret key of this client.
     secret_key: SecretKey,
     /// The DNS resolver to use.
     dns_resolver: DnsResolver,
+    /// Cache for public keys of remote nodes.
+    key_cache: KeyCache,
 }
 
 impl ClientBuilder {
@@ -87,9 +87,9 @@ impl ClientBuilder {
             #[cfg(any(test, feature = "test-utils"))]
             insecure_skip_cert_verify: false,
             proxy_url: None,
-            key_cache_capacity: 128,
             secret_key,
             dns_resolver,
+            key_cache: KeyCache::new(128),
         }
     }
 
@@ -137,7 +137,7 @@ impl ClientBuilder {
 
     /// Set the capacity of the cache for public keys.
     pub fn key_cache_capacity(mut self, capacity: usize) -> Self {
-        self.key_cache_capacity = capacity;
+        self.key_cache = KeyCache::new(capacity);
         self
     }
 
@@ -160,24 +160,293 @@ impl ClientBuilder {
                 .dangerous()
                 .set_certificate_verifier(Arc::new(NoCertVerifier));
         }
-
         config.resumption = Resumption::default();
-
         let tls_connector: tokio_rustls::TlsConnector = Arc::new(config).into();
 
-        let builder = ConnectionBuilder {
-            secret_key: self.secret_key.clone(),
-            address_family_selector: self.address_family_selector.clone(),
-            url: self.url.clone(),
-            protocol: self.protocol,
-            tls_connector,
-            dns_resolver: self.dns_resolver.clone(),
-            proxy_url: self.proxy_url.clone(),
-            key_cache: KeyCache::new(self.key_cache_capacity),
-        };
-        let (conn, local_addr) = builder.connect_0().await?;
+        let (conn, local_addr) = self.connect_0(tls_connector).await?;
 
         Ok(Client { conn, local_addr })
+    }
+
+    async fn connect_0(
+        &self,
+        tls_connector: tokio_rustls::TlsConnector,
+    ) -> Result<(Conn, Option<SocketAddr>)> {
+        let (conn, local_addr) = match self.protocol {
+            Protocol::Websocket => {
+                let conn = self.connect_ws().await?;
+                let local_addr = None;
+                (conn, local_addr)
+            }
+            Protocol::Relay => {
+                let (conn, local_addr) = self.connect_relay(tls_connector).await?;
+                (conn, Some(local_addr))
+            }
+        };
+
+        event!(
+            target: "events.net.relay.connected",
+            Level::DEBUG,
+            url = %self.url,
+            protocol = ?self.protocol,
+        );
+
+        trace!("connect_0 done");
+        Ok((conn, local_addr))
+    }
+
+    async fn connect_ws(&self) -> Result<Conn> {
+        let mut dial_url = (*self.url).clone();
+        dial_url.set_path(RELAY_PATH);
+        // The relay URL is exchanged with the http(s) scheme in tickets and similar.
+        // We need to use the ws:// or wss:// schemes when connecting with websockets, though.
+        dial_url
+            .set_scheme(if self.use_tls() { "wss" } else { "ws" })
+            .map_err(|()| anyhow!("Invalid URL"))?;
+
+        debug!(%dial_url, "Dialing relay by websocket");
+
+        let conn = tokio_tungstenite_wasm::connect(dial_url).await?;
+        let conn = Conn::new_ws(conn, self.key_cache.clone(), &self.secret_key).await?;
+        Ok(conn)
+    }
+
+    async fn connect_relay(
+        &self,
+        tls_connector: tokio_rustls::TlsConnector,
+    ) -> Result<(Conn, SocketAddr)> {
+        let url = self.url.clone();
+        let tcp_stream = self.dial_url(&tls_connector).await?;
+
+        let local_addr = tcp_stream
+            .local_addr()
+            .context("No local addr for TCP stream")?;
+
+        debug!(server_addr = ?tcp_stream.peer_addr(), %local_addr, "TCP stream connected");
+
+        let response = if self.use_tls() {
+            debug!("Starting TLS handshake");
+            let hostname = self
+                .tls_servername()
+                .ok_or_else(|| anyhow!("No tls servername"))?;
+            let hostname = hostname.to_owned();
+            let tls_stream = tls_connector.connect(hostname, tcp_stream).await?;
+            debug!("tls_connector connect success");
+            Self::start_upgrade(tls_stream, url).await?
+        } else {
+            debug!("Starting handshake");
+            Self::start_upgrade(tcp_stream, url).await?
+        };
+
+        if response.status() != hyper::StatusCode::SWITCHING_PROTOCOLS {
+            bail!(
+                "Unexpected status code: expected {}, actual: {}",
+                hyper::StatusCode::SWITCHING_PROTOCOLS,
+                response.status(),
+            );
+        }
+
+        debug!("starting upgrade");
+        let upgraded = hyper::upgrade::on(response)
+            .await
+            .context("Upgrade failed")?;
+
+        debug!("connection upgraded");
+        let conn = downcast_upgrade(upgraded)?;
+
+        let conn = Conn::new_relay(conn, self.key_cache.clone(), &self.secret_key).await?;
+
+        Ok((conn, local_addr))
+    }
+
+    /// Sends the HTTP upgrade request to the relay server.
+    async fn start_upgrade<T>(io: T, relay_url: RelayUrl) -> Result<hyper::Response<Incoming>>
+    where
+        T: AsyncRead + AsyncWrite + Send + Unpin + 'static,
+    {
+        let host_header_value = host_header_value(relay_url)?;
+
+        let io = hyper_util::rt::TokioIo::new(io);
+        let (mut request_sender, connection) = hyper::client::conn::http1::Builder::new()
+            .handshake(io)
+            .await?;
+        tokio::spawn(
+            // This task drives the HTTP exchange, completes once connection is upgraded.
+            async move {
+                debug!("HTTP upgrade driver started");
+                if let Err(err) = connection.with_upgrades().await {
+                    error!("HTTP upgrade error: {err:#}");
+                }
+                debug!("HTTP upgrade driver finished");
+            }
+            .instrument(info_span!("http-driver")),
+        );
+        debug!("Sending upgrade request");
+        let req = Request::builder()
+            .uri(RELAY_PATH)
+            .header(UPGRADE, Protocol::Relay.upgrade_header())
+            // https://datatracker.ietf.org/doc/html/rfc2616#section-14.23
+            // > A client MUST include a Host header field in all HTTP/1.1 request messages.
+            // This header value helps reverse proxies identify how to forward requests.
+            .header(HOST, host_header_value)
+            .body(http_body_util::Empty::<hyper::body::Bytes>::new())?;
+        request_sender.send_request(req).await.map_err(From::from)
+    }
+
+    fn tls_servername(&self) -> Option<rustls::pki_types::ServerName> {
+        self.url
+            .host_str()
+            .and_then(|s| rustls::pki_types::ServerName::try_from(s).ok())
+    }
+
+    fn use_tls(&self) -> bool {
+        // only disable tls if we are explicitly dialing a http url
+        #[allow(clippy::match_like_matches_macro)]
+        match self.url.scheme() {
+            "http" => false,
+            "ws" => false,
+            _ => true,
+        }
+    }
+
+    async fn dial_url(&self, tls_connector: &tokio_rustls::TlsConnector) -> Result<ProxyStream> {
+        if let Some(ref proxy) = self.proxy_url {
+            let stream = self.dial_url_proxy(proxy.clone(), tls_connector).await?;
+            Ok(ProxyStream::Proxied(stream))
+        } else {
+            let stream = self.dial_url_direct().await?;
+            Ok(ProxyStream::Raw(stream))
+        }
+    }
+
+    async fn dial_url_direct(&self) -> Result<TcpStream> {
+        debug!(%self.url, "dial url");
+        let prefer_ipv6 = self.prefer_ipv6();
+        let dst_ip = self
+            .dns_resolver
+            .resolve_host(&self.url, prefer_ipv6)
+            .await?;
+
+        let port = url_port(&self.url).ok_or_else(|| anyhow!("Missing URL port"))?;
+        let addr = SocketAddr::new(dst_ip, port);
+
+        debug!("connecting to {}", addr);
+        let tcp_stream =
+            tokio::time::timeout(
+                DIAL_NODE_TIMEOUT,
+                async move { TcpStream::connect(addr).await },
+            )
+            .await
+            .context("Timeout connecting")?
+            .context("Failed connecting")?;
+        tcp_stream.set_nodelay(true)?;
+
+        Ok(tcp_stream)
+    }
+
+    async fn dial_url_proxy(
+        &self,
+        proxy_url: Url,
+        tls_connector: &tokio_rustls::TlsConnector,
+    ) -> Result<util::Chain<std::io::Cursor<Bytes>, MaybeTlsStream>> {
+        debug!(%self.url, %proxy_url, "dial url via proxy");
+
+        // Resolve proxy DNS
+        let prefer_ipv6 = self.prefer_ipv6();
+        let proxy_ip = self
+            .dns_resolver
+            .resolve_host(&proxy_url, prefer_ipv6)
+            .await?;
+
+        let proxy_port = url_port(&proxy_url).ok_or_else(|| anyhow!("Missing proxy url port"))?;
+        let proxy_addr = SocketAddr::new(proxy_ip, proxy_port);
+
+        debug!(%proxy_addr, "connecting to proxy");
+
+        let tcp_stream = tokio::time::timeout(DIAL_NODE_TIMEOUT, async move {
+            TcpStream::connect(proxy_addr).await
+        })
+        .await
+        .context("Timeout connecting")?
+        .context("Connecting")?;
+
+        tcp_stream.set_nodelay(true)?;
+
+        // Setup TLS if necessary
+        let io = if proxy_url.scheme() == "http" {
+            MaybeTlsStream::Raw(tcp_stream)
+        } else {
+            let hostname = proxy_url.host_str().context("No hostname in proxy URL")?;
+            let hostname = rustls::pki_types::ServerName::try_from(hostname.to_string())?;
+            let tls_stream = tls_connector.connect(hostname, tcp_stream).await?;
+            MaybeTlsStream::Tls(tls_stream)
+        };
+        let io = TokioIo::new(io);
+
+        let target_host = self
+            .url
+            .host_str()
+            .ok_or_else(|| anyhow!("Missing proxy host"))?;
+
+        let port = url_port(&self.url).ok_or_else(|| anyhow!("invalid target port"))?;
+
+        // Establish Proxy Tunnel
+        let mut req_builder = Request::builder()
+            .uri(format!("{}:{}", target_host, port))
+            .method("CONNECT")
+            .header("Host", target_host)
+            .header("Proxy-Connection", "Keep-Alive");
+        if !proxy_url.username().is_empty() {
+            // Passthrough authorization
+            // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Proxy-Authorization
+            debug!(
+                "setting proxy-authorization: username={}",
+                proxy_url.username()
+            );
+            let to_encode = format!(
+                "{}:{}",
+                proxy_url.username(),
+                proxy_url.password().unwrap_or_default()
+            );
+            let encoded = BASE64URL.encode(to_encode.as_bytes());
+            req_builder = req_builder.header("Proxy-Authorization", format!("Basic {}", encoded));
+        }
+        let req = req_builder.body(Empty::<Bytes>::new())?;
+
+        debug!("Sending proxy request: {:?}", req);
+
+        let (mut sender, conn) = hyper::client::conn::http1::handshake(io).await?;
+        tokio::task::spawn(async move {
+            if let Err(err) = conn.with_upgrades().await {
+                error!("Proxy connection failed: {:?}", err);
+            }
+        });
+
+        let res = sender.send_request(req).await?;
+        if !res.status().is_success() {
+            bail!("Failed to connect to proxy: {}", res.status());
+        }
+
+        let upgraded = hyper::upgrade::on(res).await?;
+        let Ok(Parts { io, read_buf, .. }) = upgraded.downcast::<TokioIo<MaybeTlsStream>>() else {
+            bail!("Invalid upgrade");
+        };
+
+        let res = util::chain(std::io::Cursor::new(read_buf), io.into_inner());
+
+        Ok(res)
+    }
+
+    /// Reports whether IPv4 dials should be slightly
+    /// delayed to give IPv6 a better chance of winning dial races.
+    /// Implementations should only return true if IPv6 is expected
+    /// to succeed. (otherwise delaying IPv4 will delay the connection
+    /// overall)
+    fn prefer_ipv6(&self) -> bool {
+        match self.address_family_selector {
+            Some(ref selector) => selector(),
+            None => false,
+        }
     }
 }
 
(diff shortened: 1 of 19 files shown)


flub:
@matheus23 this is still draft but you might already have opinions on the channel created (`RelayDatagramSendChannelSender` and `RelayDatagramSendChannelReceiver`), which -pending review- final already.

iroh/src/magicsock.rs
```diff
@@ -1729,6 +1727,71 @@ enum DiscoBoxError {
...
+
+/// Sender to send datagrams to the [`RelayActor`].
+///
+/// This includes the waker coordination required to support [`AsyncUdpSocket::try_send`]
+/// and [`quinn::UdpPoller::poll_writable`].
+#[derive(Debug, Clone)]
+struct RelayDatagramSendChannelSender {
+    sender: mpsc::Sender<RelaySendItem>,
+    waker: Arc<AtomicWaker>,
+}
```
matheus23, line 1756: I guess this works, assuming there's never more than one `RelayDatagramSendChannelSender` registering for wakeup concurrently (and I think that's the case. Technically MagicSock is available via `&self`, but quinn is just lying with its type signature in `AsyncUdpSocket` there, and it's practically `&mut self`).

Any reason for doing this instead of trying to use [`PollSender`](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.PollSender.html)?
I guess there's a chance that `PollSender::send_item` panics, if the `poll_reserve` doesn't work out as planned, but otherwise?

iroh/src/magicsock.rs
```diff
@@ -1729,6 +1727,71 @@ enum DiscoBoxError {
...
+        self.sender.try_send(item)
+    }
+
+    fn poll_writable(&self, cx: &mut Context) -> Poll<io::Result<()>> {
+        match self.sender.capacity() {
+            0 => {
+                self.waker.register(cx.waker());
+                Poll::Pending
+            }
+            _ => Poll::Ready(Ok(())),
```
dignifiedquire, line 1772: I really wish this would use the `try_reserve` API instead, and store the `Permit` for sending later..

dignifiedquire: https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Sender.html#method.try_reserve

flub: Bear in mind that `try_send(&self, item)` and `poll_writeable(&mut self, cx)`.

This becomes pretty unwieldy to make this happen.  The poller could be it's own struct, but it would still have a clone of the sender and waker, plus you'd probably need a channel to send an `mpsc::OwnedPermit` back to the `Sender`.  At this point I'm not sure that version still looks appealing.

iroh/src/magicsock.rs
flub: I didn't know about `PollSender`...  But I don't think it helps much: all it's methods are all `&mut self` and the sequencing of `poll_reserve` & `send_item` is kind of tricky when spread across two methods.  Sending a permit around seems favourable over maybe panicking.

iroh/src/magicsock.rs
dignifiedquire: > try_send(&self, item)

I hate this function signature so much..

iroh/src/magicsock.rs
matheus23: Regarding the first issue again: You're sharing a single `AtomicWaker` for all tasks that might want to be woken up when there's space again. There *are* multiple tasks involved, because quinn will `try_send` on the same socket for multiple connections concurrently.
So, I don't think this code is correct.

matheus23: This is different from the `RelayDatagramRecvQueue`, because that's only accessed from a single task IIUC: The endpoint driver.

matheus23: As [discussed in discord](https://discord.com/channels/949724860232392765/1293936403733807165/1319348835952099420): It's not *this PR* that's introducing this bug, this PR is mostly refactoring the way it worked before, so let's not have my concern block this PR from moving on.

iroh/src/magicsock/relay_actor.rs
```diff
@@ -152,265 +224,450 @@ impl ActiveRelayActor {
...
+                }
+                msg = self.inbox.recv() => {
+                    let Some(msg) = msg else {
+                        warn!("Inbox closed, shutdown.");
+                        break Ok(());
+                    };
+                    match msg {
+                        ActiveRelayMessage::SetHomeRelay(is_preferred) => {
+                            let fut = client_sink.send(SendMessage::NotePreferred(is_preferred));
+                            self.run_sending(fut, &mut state, &mut client_stream).await?;
```
dignifiedquire, line 448: this is missing setting `is_home_relay` on the state?

iroh-relay/src/server/http_server.rs
```diff
@@ -998,15 +981,20 @@ mod tests {
...
+        let _res = client_a
+            .send(SendMessage::SendPacket(
+                public_key_b,
+                Bytes::from_static(b"try to send"),
+            ))
             .await;
-        assert!(res.is_err());
-        assert!(client_receiver_b.recv().await.is_err());
+        // TODO: this send seems to succeed currently.
+        // assert!(res.is_err());
```
dignifiedquire, line 996: this should be failing now

iroh/src/magicsock.rs
```diff
@@ -184,12 +184,9 @@ pub(crate) struct MagicSock {
     /// Relay datagrams received by relays are put into this queue and consumed by
     /// [`AsyncUdpSocket`].  This queue takes care of the wakers needed by
     /// [`AsyncUdpSocket::poll_recv`].
-    relay_datagrams_queue: Arc<RelayDatagramsQueue>,
-    /// Waker to wake the [`AsyncUdpSocket`] when more data can be sent to the relay server.
-    ///
-    /// This waker is used by [`IoPoller`] and the [`RelayActor`] to signal when more
-    /// datagrams can be sent to the relays.
-    relay_send_waker: Arc<std::sync::Mutex<Option<Waker>>>,
+    relay_datagram_recv_queue: Arc<RelayDatagramRecvQueue>,
```
dignifiedquire, line 187: seems inconsistent, one neds an `Arc` and one doesn't?

iroh/src/magicsock.rs
```diff
@@ -1739,16 +1812,16 @@ enum DiscoBoxError {
...
     waker: AtomicWaker,
 }
 
-impl RelayDatagramsQueue {
+impl RelayDatagramRecvQueue {
     /// Creates a new, empty queue with a fixed size bound of 128 items.
     fn new() -> Self {
         Self {
-            queue: ConcurrentQueue::bounded(128),
+            queue: ConcurrentQueue::bounded(512),
```
dignifiedquire, line 1824: why this number? also the comment is now outdated

iroh-relay/src/client.rs
```diff
@@ -903,19 +525,17 @@ impl Actor {
             TcpStream::connect(proxy_addr).await
         })
         .await
-        .map_err(|_| ClientError::ConnectTimeout)?
-        .map_err(ClientError::DialIO)?;
+        .context("Timeout connecting")?
+        .context("Error connecting")?;
```
dignifiedquire, line 529: no need to write `error` in these contexts, these are already shown as errors

iroh/src/magicsock/relay_actor.rs
```diff
@@ -43,38 +51,91 @@ const RELAY_INACTIVE_CLEANUP_TIME: Duration = Duration::from_secs(60);
...
+///   server at this time.  However it is actively sending data to the server so can not
+///   consume any further items from inboxes which will result in sending more data to the
+///   server until the actor goes back to the `connected` state.
+///
+/// All these are driven from the top-level [`ActiveRelayActor::run`] loop.
 #[derive(Debug)]
 struct ActiveRelayActor {
+    // The inboxes and channels this actor communicates over.
+    /// Inbox for messages which should be handled without any blocking.
+    fast_inbox: mpsc::Receiver<ActiveRelayFastMessage>,
```
dignifiedquire, line 109: nit: `high_priority_inbox` would make more sense imo

iroh/src/magicsock/relay_actor.rs
```diff
@@ -152,265 +224,455 @@ impl ActiveRelayActor {
...
-        // When this future has an inner, it is a future which is currently sending
-        // something to the relay server.  Nothing else can be sent to the relay server at
-        // the same time.
-        let mut relay_send_fut = std::pin::pin!(MaybeFuture::none());
+        loop {
+            let Some(client) = self.run_dialing().instrument(info_span!("dialing")).await else {
+                break;
+            };
+            match self
+                .run_connected(client)
```
dignifiedquire, line 248: I don't know if it makes this code more complex, but I am wondering if it would be better to model the state like this

```rust
enum ActiveRelayConnector {
  Dialing {
    // ..
  },
  Connected {
    // .. 
  }
}
```

iroh-relay/src/server/http_server.rs
dignifiedquire: nvm, this is only fixed in my PR

iroh/src/magicsock.rs
flub: Well, one is a `Sender` which is `Clone`.  The other is a queue.  I don't think this is a big deal, just an artifact of how the channels/queues are built internally.

iroh/src/magicsock.rs
flub: This is based on the previous per client queues which were this size IIRC.

At 1500 bytes/packet this is 750KiB of buffer.  I think ideally the datagram queues would all have a capacity based on the number of datagram **bytes** queued rather than number of items.  I'm using 1500 bytes/packet to estimate, but with GSO etc it could also be much larger as the max packet size is 64KiB, but being that big is probably not common.

But I think doing buffers capacity on buffered bytes should be a separate effort, and it isn't very high priority right now.

iroh-relay/src/client.rs
flub: Fair, I probably did that because of the "timeout" clarification in the other one.

iroh/src/magicsock/relay_actor.rs
flub: Yeah, I did this in a rush and didn't really think about the naming.  Renamed to `prio_inbox` everywhere.

iroh/src/magicsock/relay_actor.rs
flub: Currently the state in the `ActiveRelayActor` itself is used in all 3 states the actor can be in.  The `ConnectedRelayState` is only shared between sending and connected states.  I don't feel like trying to put this state in the actor itself will help much, you'd still have `run_sending` implemented in the disconnected state, but this is already alleviated by it needing the `ConnectedRelayState`.  What might work is create `ConnectedRelay` and `DisconnectedRelay` and impl the methods on there.  They could take a ref to the data in the `ActiveRelayActor` probably.

iroh-relay/src/client.rs
```diff
@@ -306,356 +164,230 @@ impl ClientBuilder {
...
-        let writer = ConnWriter::Ws(writer);
-
-        Ok((reader, writer))
+        let conn = tokio_tungstenite_wasm::connect(dial_url).await?;
+        let conn = Conn::new_ws(conn, self.key_cache.clone(), &self.secret_key).await?;
+        Ok(conn)
     }
 
-    async fn connect_derp(&self) -> Result<(ConnReader, ConnWriter, SocketAddr), ClientError> {
+    async fn connect_relay(&self) -> Result<(Conn, SocketAddr)> {
```
matheus23, line 376: I've stumbled upon this a bunch.

```suggestion
    async fn connect_derp(&self) -> Result<(Conn, SocketAddr)> {
```

Calling this `connect_relay` as opposed to `connect_ws` is weird. `connect_ws` *also* connects to a relay. And also via the "relay protocol".

The only difference between the two is that `connect_relay` uses the `http derp ...` HTTP upgrade and continues with "derp framing", whereas `_ws` uses `websockets` as HTTP upgrade and uses derp framing.

My opinion is that `_derp` should be kept until the "http derp ..." HTTP upgrade string is removed.

iroh-relay/src/client/conn.rs
```diff
@@ -3,280 +3,105 @@
...
-            Ok(ReceivedMessage::ServerRestarting {
-                reconnect_in,
-                try_for,
-            })
-        }
-        _ => bail!("unexpected packet: {:?}", frame.typ()),
-    }
-}
+    /// Constructs a new websocket connection, including the initial server handshake.
+    pub(crate) async fn new_relay(
```
matheus23, line 78: Same thing with the derp vs. relay naming here:
```suggestion
    pub(crate) async fn new_derp(
```

iroh-relay/src/client/conn.rs
```diff
@@ -3,280 +3,105 @@
...
-    /// Is `None` in tests or when using websockets (because we don't control connection establishment in browsers).
-    local_addr: Option<SocketAddr>,
-    /// Channel on which to communicate to the server. The associated [`mpsc::Receiver`] will close
-    /// if there is ever an error writing to the server.
-    writer_channel: mpsc::Sender<ConnWriterMessage>,
-    /// JoinHandle for the [`ConnWriter`] task
-    writer_task: AbortOnDropHandle<Result<()>>,
-    reader_task: AbortOnDropHandle<()>,
+pub(crate) enum Conn {
+    Relay {
```
matheus23, line 51: Same thing with the naming here
```suggestion
    Derp {
```

iroh-relay/src/client/conn.rs
```diff
@@ -308,140 +146,113 @@ impl Stream for ConnReader {
...
 
     fn poll_ready(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
         match *self {
-            Self::Derp(ref mut ws) => Pin::new(ws).poll_ready(cx),
-            Self::Ws(ref mut ws) => Pin::new(ws).poll_ready(cx).map_err(tung_wasm_to_io_err),
+            Self::Relay { ref mut conn } => Pin::new(conn).poll_ready(cx).map_err(Into::into),
+            Self::Ws { ref mut conn, .. } => Pin::new(conn)
+                .poll_ready(cx)
+                .map_err(tung_wasm_to_io_err)
+                .map_err(Into::into),
```
matheus23, line 158: You're using this a bunch.
At this point I think it makes sense to change `tung_wasm_to_io_err` into `tung_wasm_to_conn_send_err` and make it return `ConnSendError`.

Or perhaps - now that we own `ConnSendError` - we could implement `From<tungstenite_wasm::Error> for ConnSendError`.

iroh/src/magicsock/relay_actor.rs
flub: https://github.com/n0-computer/iroh/pull/3095

iroh-relay/src/server.rs
```diff
@@ -886,7 +907,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_relay_clients_both_derp() {
+    async fn test_relay_clients_both_relay() -> Result<()> {
```
matheus23, line 910: I'd recommend `TestResult` instead.
The difference is that `TestResult` immediately panics on every `?`.
This helps locate the actual line that the error failed on quickly.
```suggestion
    async fn test_relay_clients_both_relay() -> TestResult {
```
(Same with the other tests below)

iroh-relay/src/server/metrics.rs
```diff
@@ -61,7 +61,7 @@ pub struct Metrics {
     /// Number of accepted websocket connections
     pub websocket_accepts: Counter,
     /// Number of accepted 'iroh derp http' connection upgrades
-    pub derp_accepts: Counter,
+    pub relay_accepts: Counter,
```
matheus23, line 64: Another place where I'd like to keep the "derp" naming :sweat_smile:

iroh/src/magicsock/relay_actor.rs
```diff
@@ -43,38 +51,91 @@ const RELAY_INACTIVE_CLEANUP_TIME: Duration = Duration::from_secs(60);
...
+/// longer than this timeout datagrams will be dropped.
+const UNDELIVERABLE_DATAGRAM_TIMEOUT: Duration = Duration::from_millis(400);
+
 /// An actor which handles the connection to a single relay server.
 ///
 /// It is responsible for maintaining the connection to the relay server and handling all
 /// communication with it.
+///
+/// The actor shuts down itself on inactivity: inactivity is determined when no more
+/// datagrams are being received to send.
```
matheus23, line 83: "received to send" reads a little confusingly. Perhaps
```suggestion
/// datagrams are being queued to send.
```

iroh/src/magicsock/relay_actor.rs
```diff
@@ -43,38 +51,91 @@ const RELAY_INACTIVE_CLEANUP_TIME: Duration = Duration::from_secs(60);
...
 #[derive(Debug)]
 struct ActiveRelayActor {
+    // The inboxes and channels this actor communicates over.
+    /// Inbox for messages which should be handled without any blocking.
+    prio_inbox: mpsc::Receiver<ActiveRelayPrioMessage>,
+    /// Inbox for messages which involve sending to the relay server.
+    inbox: mpsc::Receiver<ActiveRelayMessage>,
     /// Queue to send received relay datagrams on.
-    relay_datagrams_recv: Arc<RelayDatagramsQueue>,
+    relay_datagrams_recv: Arc<RelayDatagramRecvQueue>,
```
matheus23, line 113: Maybe "Queue for received relay datagrams"?

iroh/src/magicsock/relay_actor.rs
```diff
@@ -43,38 +51,91 @@ const RELAY_INACTIVE_CLEANUP_TIME: Duration = Duration::from_secs(60);
...
     /// Channel on which we receive packets to send to the relay.
-    relay_datagrams_send: mpsc::Receiver<RelaySendPacket>,
+    relay_datagrams_send: mpsc::Receiver<RelaySendItem>,
```
matheus23, line 115: Could this be "Channel on which we queue packets to send to the relay"?

matheus23 (commented):
Didn't take a super close look at `relay_actor.rs` yet, given there's another stacked PR open about it right now.

I disagree with the `Conn::Derp` -> `Conn::Relay` rename and related. I've commented on that.

Otherwise only small stuff, some comments, etc. I had a hard time trying to figure out if this is a true refactor, because of the changes, so I gave up on that and mostly looked at what the code looks like now, which tbh is so much better.

iroh-relay/src/client.rs
dignifiedquire: Well at that point likely `connect_ws` will be the only version.. naming is weird I agree, not sure which state I prefer at this point

iroh-relay/src/server.rs
dignifiedquire: we can do this later, but we haven't been using asfaik in here

flub: Done, but /cc @dignifiedquire as he might have written this.  (not actually sure)

iroh-relay/src/client.rs
matheus23: Yeah that's okay. I don't hate the name `connect_ws`.

iroh-relay/src/client.rs
```diff
@@ -306,356 +164,230 @@ impl ClientBuilder {
...
+    tls_connector: tokio_rustls::TlsConnector,
+    dns_resolver: DnsResolver,
+    proxy_url: Option<Url>,
+    key_cache: KeyCache,
+}
 
-    async fn connect_0(&self) -> Result<(Conn, ConnReceiver), ClientError> {
-        let (reader, writer, local_addr) = match self.protocol {
+impl ConnectionBuilder {
+    async fn connect_0(&self) -> Result<(Conn, Option<SocketAddr>)> {
```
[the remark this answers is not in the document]
flub: Good call!  Done.

That was one of those bits of code that I just didn't want to think about and just had to keep working without getting in my way.

flub:
> Didn't take a super close look at `relay_actor.rs` yet, given there's another stacked PR open about it right now.

That's just a draft to explore some possible futures, I don't really intend to merge that in here if folks are happy enough with the current code structure (I am).

> I disagree with the `Conn::Derp` -> `Conn::Relay` rename and related. I've commented on that.

I'm pretty agnostic to that.  Not what I care about here.  Can we leave these for now or do you fear it will introduce too much churn?

> Otherwise only small stuff, some comments, etc. I had a hard time trying to figure out if this is a true refactor, because of the changes, so I gave up on that and mostly looked at what the code looks like now, which tbh is so much better.

What do you mean "true refactor"?  For iroh this is all internal and not visible.  For iroh-relay this involes public API changes.  But still is a refactor and not a feature I'd say.

matheus23 (approved):
LGTM. We can leave the derp -> relay changes.

> What do you mean "true refactor"?

Ah right. I categorize changes into "refactor"/"not refactor" somewhat strictly in the sense that everything that doesn't change program behavior observably, but changes code structure is a ("true") refactor.
Something that's supposed to be a refactor, but actually changes behavior I'd say is *not* a "true refactor".

flub:
> LGTM. We can leave the derp -> relay changes.
> 
> > What do you mean "true refactor"?
> 
> Ah right. I categorize changes into "refactor"/"not refactor" somewhat strictly in the sense that everything that doesn't change program behavior observably, but changes code structure is a ("true") refactor. Something that's supposed to be a refactor, but actually changes behavior I'd say is _not_ a "true refactor".

Yeah ok.  Probably not a true refactor in that case.  Another question is whether this should really be classified as a fix?  It should close #3008, #2971 and #2951?  Oh well, I'll rename this fix :)

iroh-relay/src/server.rs
dignifiedquire: I don't mind

dignifiedquire (approved):

PR #3063, opened 2024-12-19
ci(*): Run flaky tests with --verbose and use ci profile

flub:
## Description

Also fix building of tests to use --profile ci which is used to run
the tests.  The test run did not trigger a re-build.  So mosly likely
the profile is not encoded in the build artifacts and we were running
tests on the normal dev profile rather than the ci profile.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -102,7 +102,7 @@ jobs:
 
       - name: build tests
         run: |
-          cargo nextest run --workspace ${{ env.FEATURES }} --lib --bins --tests --profile ci --no-run
+          cargo nextest run --workspace ${{ env.FEATURES }} --lib --bins --tests --no-run
 
       - name: list ignored tests
         run: |
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -102,7 +102,7 @@ jobs:
 
       - name: build tests
         run: |
-          cargo nextest run --workspace ${{ env.FEATURES }} --lib --bins --tests --no-run
+          cargo nextest run --workspace ${{ env.FEATURES }} --lib --bins --tests --profile ci --no-run
 
       - name: list ignored tests
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -102,7 +102,7 @@ jobs:
 
       - name: build tests
         run: |
-          cargo nextest run --workspace ${{ env.FEATURES }} --lib --bins --tests --no-run
+          cargo nextest run --workspace ${{ env.FEATURES }} --lib --bins --tests --profile ci --no-run
```
Arqu, line 105: I don't think this works the way you expected. IIRC this shifts the nextest profile which is a different thing from the cargo run build profiles. 

If you look at `.config/nextest.toml` you will see that this specifies an execution setup but is not related to the build artifacts in any way. Also the reason running didn't trigger a rebuild.

flub: eh.  TIL.  Reverted

Arqu (approved):

.github/workflows/tests.yaml
```diff
@@ -111,7 +111,7 @@ jobs:
       - name: run tests
         run: |
           mkdir -p output
-          cargo nextest run --workspace ${{ env.FEATURES }} --lib --bins --tests --profile ci --run-ignored ${{ inputs.flaky && 'all' || 'default' }} --no-fail-fast --message-format ${{ inputs.flaky && 'libtest-json' || 'human' }} > output/${{ matrix.name }}_${{ matrix.features }}_${{ matrix.rust }}.json
+          cargo nextest run --workspace ${{ env.FEATURES }} --lib --bins --tests --profile ci --run-ignored ${{ inputs.flaky && 'all' || 'default' }} ${{ inputs.flaky && '--verbose' }} --no-fail-fast --message-format ${{ inputs.flaky && 'libtest-json' || 'human' }} > output/${{ matrix.name }}_${{ matrix.features }}_${{ matrix.rust }}.json
```
Arqu, line 114: Fyi in the test suite this renders as "false" for the CLI command, thing you need to add `|| ''` so it returns an empty string when evaluating to `false`.

PR #3066, opened 2024-12-20
tests(iroh): Add some context to test errors

flub:
## Description

The flaky tests are being super weird and making no sense.  This is
clutching at straws for any hint.

## Notes & open questions

/cc @ramfox 

Two slightly odd things I spotted while looking at this:

- The use of 240.0.0.1 as "bad" IP address returned in discovery.  Maybe one from the test nets in RFC 5737 would be better?
- The endpoints accept a connection and immediately drop it.  This could potentially send the CONNECTION_CLOSE before the ACK from the connection resulting in the connect call returning an error (BUT THAT WOULD BE A NORMAL READABLE TEST FAILURE).  This probably doesn't happen because it's all local and very fast and Quinn has already ACKed the connection before the code manages to drop it.  Still, essentially racy code.

Fine, I added a fix for the 2nd.  But it's not the issue.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/discovery.rs
+++ b/iroh/src/discovery.rs
@@ -712,10 +712,13 @@ mod tests {
         let handle = tokio::spawn({
             let ep = ep.clone();
             async move {
+                // Keep connections alive until the task is dropped.
+                let mut connections = Vec::new();
                 // we skip accept() errors, they can be caused by retransmits
                 while let Some(connecting) = ep.accept().await.and_then(|inc| inc.accept().ok()) {
-                    let _conn = connecting.await?;
                     // Just accept incoming connections, but don't do anything with them.
+                    let conn = connecting.await?;
+                    connections.push(conn);
                 }
 
                 anyhow::Ok(())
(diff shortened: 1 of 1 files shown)


flub:
@ramfox ping

matheus23:
I think the code you're modifying `git blames` back to me, actually. So I'll just go ahead and comment on this :v:

> The flaky tests are being super weird and making no sense. This is
> clutching at straws for any hint.

I'd love to know what specific errors you're trying to fix, so if you could link to the failing run you were looking at, that'd be great.

matheus23 (approved):
Okay, so I think what you're trying is to be better prepared, should the flake occur again, do I understand that correctly?

If so, I think this can just be merged.

flub:
> I think the code you're modifying `git blames` back to me, actually. So I'll just go ahead and comment on this ✌️
> 
> > The flaky tests are being super weird and making no sense. This is
> > clutching at straws for any hint.
> 
> I'd love to know what specific errors you're trying to fix, so if you could link to the failing run you were looking at, that'd be great.

Yeah, classic mistake of assuming it's obvious and now it's lost somewhere... but still, if it comes back this should make understanding it a bit easier.

PR #3068, opened 2024-12-20
refactor(iroh-relay): Make the ConnReceiver actor a ConnMessageStream

flub:
## Description

This removes an actor by making it a stream.

The main (temporary) downside is that a read error no longer shuts down the Conn WriterTasks.  This will not be an issue for that long as the WriterTasks are going away next and then the Client can manage this.

The RelayDatagramRecvQueue is grown to 512 datagrams.  We used to keep this many frames in the per-relay stream.  Though that's potentially an awful lot.  For datagrams we can assume that they will at max settle close to 1500 bytes each, so this buffer will end up being 750KiB max.  That seems somewhat reasonable, though we could probably double it still.

There will effectively no longer be a per-relay buffer - other than inside the relay's TCP stream.

## Notes & open questions

This does some other renaming, e.g. `ConnReader` is now a `ConnFrameStream` which is a bit more coherent.  It also moves a bunch of code around in the `conn.rs` file to give it some more stucture.  This makes the diff harder to read then it needs to be.  `process_incoming_frame` has not changed and the main other change is that the reader task is no longer created and instead is moved into the `Stream` impl of the new `ConnMessageStream`.

This is open for review.  Though whether or not we want to merge it is another matter.  I'll see for how the next few PRs on top of this look to see what the real impact of the WriterTasks not being shut down is.  Maybe we'll end up merging several PRs together before merging to main.  But logically this is a nice dividing point to review.

## Change checklist

- [X] Self-review.
- [X] Tests if relevant.

### DIFF
--- a/iroh/src/magicsock.rs
+++ b/iroh/src/magicsock.rs
@@ -1821,7 +1821,7 @@ impl RelayDatagramRecvQueue {
     /// Creates a new, empty queue with a fixed size bound of 128 items.
     fn new() -> Self {
         Self {
-            queue: ConcurrentQueue::bounded(128),
+            queue: ConcurrentQueue::bounded(512),
             waker: AtomicWaker::new(),
         }
     }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -329,9 +329,12 @@ version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b62ddb9cb1ec0a098ad4bbf9344d0713fa193ae1a80af55febcff2627b6a00c1"
 dependencies = [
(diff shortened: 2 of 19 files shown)


iroh/src/magicsock.rs
```diff
@@ -1821,7 +1821,7 @@ impl RelayDatagramRecvQueue {
     /// Creates a new, empty queue with a fixed size bound of 128 items.
```
matheus23, line 1821: This comment is outdated

matheus23: This comment is still outdated.

Let's fix it if we ever walk past it in another PR again, cc @flub :)

PR #3070, opened 2024-12-22
fix!: cleanup dead and unused dependencies

dignifiedquire:
## Breaking Changes

- removed `regex` feature (as the dependency is not used anymore)

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


iroh/Cargo.toml
```diff
@@ -42,36 +41,26 @@ futures-sink = "0.3"
...
-    "alloc",
-    "use-std",
-    "experimental-derive",
-] }
 quinn = { package = "iroh-quinn", version = "0.12.0" }
 quinn-proto = { package = "iroh-quinn-proto", version = "0.12.0" }
 quinn-udp = { package = "iroh-quinn-udp", version = "0.5.5" }
 rand = "0.8"
 rcgen = "0.13"
-regex = { version = "1.7.1", optional = true }
```
flub, line 74: The semvers check correctly points out that this is a breaking change because every optional dependency has an implicit feature of its own name.  So you'll have to list this in the breaking changes section and fixup the title.

flub (approved):
Needs the breaking change documented, otherwise looks great.

PR #3088, opened 2025-01-03
ci: Pin an older nextest version

flub:
## Description

We get very weird errors from nextest, I suspect this is a regression.
This version is 3 months old, which is probably something nice to try.

This does not yet do this everywhere, e.g. windows still installs the
latest.  But this is enough to see if it helps.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -66,7 +66,7 @@ jobs:
       - name: Install cargo-nextest
         uses: taiki-e/install-action@v2
         with:
-          tool: nextest
+          tool: nextest@0.9.80
 
       - name: Install sccache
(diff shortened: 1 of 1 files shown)


Arqu (approved):
Do you have a link / example of the regression?

Arqu:
Ah ok, I wasn't sure which problem you were referring to. Yeah we need more info there.

matheus23:
:sob: And now let's copy this PR into all other repos :smiling_face_with_tear:

Arqu:
Wait are we happy with this solution, did it work?

matheus23:
Hard to know for sure. But so far this solution *hasn't failed* (succeeded 3 times).

PR #3090, opened 2025-01-03
fix-or-feat(iroh): Set MaybeFuture to None on Poll::Ready

flub:
## Description

This removes a footgun where after polling the future it is easy to
forget to call MaybeFuture::set_none, which can result in a panic if
the inner future is not fused.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/util.rs
+++ b/iroh/src/util.rs
@@ -53,11 +53,18 @@ impl<T> MaybeFuture<T> {
 impl<T: Future> Future for MaybeFuture<T> {
     type Output = T::Output;
 
-    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
-        let this = self.project();
-        match this {
-            MaybeFutureProj::Some(t) => t.poll(cx),
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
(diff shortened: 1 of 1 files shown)


dignifiedquire (approved):

iroh/src/util.rs
```diff
@@ -53,11 +53,18 @@ impl<T> MaybeFuture<T> {
...
-            MaybeFutureProj::Some(t) => t.poll(cx),
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let mut this = self.as_mut().project();
+        let poll_res = match this {
+            MaybeFutureProj::Some(ref mut t) => t.as_mut().poll(cx),
             MaybeFutureProj::None => Poll::Pending,
+        };
+        match poll_res {
+            Poll::Ready(val) => {
+                self.as_mut().project_replace(Self::None);
```
rklaehn, line 64: project_replace... how are you supposed to keep these APIs in your head...

flub: You don't: you post a challenge in the rust channel on discord and @dignifiedquire will figures it out :stuck_out_tongue: 

Alternatively you try to use pin-project on an enum, discover this blows up and go read the docs.

PR #3095, opened 2025-01-03
experiment with splitting this actor state a bit more

flub:
## Description

Experiments with splitting the `ActiveRelayActor` a bit more.  Inspired by https://github.com/n0-computer/iroh/pull/3062#discussion_r1901728617

## Notes & open questions

This doesn't work yet because there's a nasty `Send` issue and I'm not determined enough to fix this yet.  But it's a nice place to discuss the architecture a bit.

What **is** nice from this is that the methods on `ConnectedRelay` become, well, methods rather than taking this `&mut state` so that part feels more natural.

What would be nice to achieve on top of this is force the `run_sending` state into the type system as well since currently calling that is rather by convention.  That would be the thing that would make this approach really appealing.

----

A totally different approach is to split the actual actor in 3 again:
- The current actor which manages:
- A send-only actor
- A receive-only-actor.

The main benefit I would hope to achieve is that the concurrency of the send/receive is easier.  In the current version keeping receive going while sending is a bit awkward.  What it does not simplify is the state change between dialing and connected that the current actor has to juggle.

The main cost is that there would be another channel between the `ActiveRelayActor` and the sending sub-actor.  The nice thing about this is that the `ActiveRelayActor` can work with permits, which could be super nice!  The hard thing is deciding on the sizing of this channel.  But maybe a single-item channel wouldn't be so bad.

I do believe this version is worth exploring more, despite the fact it introduces two more actors again (but one one more queue on the send side already discussed).  But also I'm keenly aware I'm time-limited and the current version is not bad.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -329,9 +329,12 @@ version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b62ddb9cb1ec0a098ad4bbf9344d0713fa193ae1a80af55febcff2627b6a00c1"
 dependencies = [
+ "futures-core",
  "getrandom",
  "instant",
+ "pin-project-lite",
  "rand",
+ "tokio",
 ]
 
 [[package]]
--- a/iroh-relay/src/client.rs
+++ b/iroh-relay/src/client.rs
@@ -526,7 +526,7 @@ impl ConnectionBuilder {
         })
         .await
         .context("Timeout connecting")?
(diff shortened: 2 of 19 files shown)


flub:
No one felt like having architecture discussions :grin:

dignifiedquire:
> No one felt like having architecture discussions 😁

sorry, it got lost..

PR #3099, opened 2025-01-07
fix(iroh-relay): removes deadlock in `Clients`

ramfox:
## Description

Remove the deadlock that was a bit hidden due to `DashMap`.

Added `client` parameter to `unregister` to explicitly drop before attempting to call into the `clients` `DashMap` again.

## Notes & open questions

The `warn` log for stream termination seems a little fear-mongering, but I'm not sure the best way to "downgrade" this, as we seem to rely on this error in tests like `endpoint_relay_connect_loop`. Ended up leaving it as is.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-relay/src/server/clients.rs
+++ b/iroh-relay/src/server/clients.rs
@@ -56,7 +56,14 @@ impl Clients {
     /// Removes the client from the map of clients, & sends a notification
     /// to each client that peers has sent data to, to let them know that
     /// peer is gone from the network.
-    async fn unregister(&self, node_id: NodeId) {
+    ///
+    /// Explicitly drops the reference to the client to avoid deadlock.
+    async fn unregister<'a>(
+        &self,
+        client: dashmap::mapref::one::Ref<'a, iroh_base::PublicKey, Client>,
+        node_id: NodeId,
+    ) {
+        drop(client); // avoid deadlock
         trace!(node_id = node_id.fmt_short(), "unregistering client");
 
         if let Some((_, client)) = self.0.clients.remove(&node_id) {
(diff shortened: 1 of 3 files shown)


Arqu:
Can confirm this fixes the issue for me.

iroh-relay/src/server/clients.rs
```diff
@@ -85,8 +85,13 @@ impl Clients {
...
     /// Attempt to send a packet to client with [`NodeId`] `dst`
     pub(super) async fn send_packet(&self, dst: NodeId, data: Bytes, src: NodeId) -> Result<()> {
+        let mut res = None;
         if let Some(client) = self.0.clients.get(&dst) {
-            let res = client.try_send_packet(src, data);
+            res = Some(client.try_send_packet(src, data));
+        }
+        if let Some(res) = res {
+            // `process_result` may call `unregister`, which itself calls `self.0.clients`
+            // so we can't nest this under the above call to `self.0.clients`
```
flub, line 94: Oof, the size of this footgun!

I guess `&mut self` is a tool to fix that, but I assume that's difficult to do here because `Clients` needs to be `Clone` I assume, I didn't check how dearly that is needed but presumably that is what makes the routing scale across cores.  I don't have a useful suggestion without studying this deeper.

flub (approved):
LGTM but let's wait for @dignifiedquire as he wrote this I think

ramfox:
Okay, added a `client` param to `unregister`.

The thing that really sold it for me was that it is now easy for this deadlock to show up in a test if there are future refactors. It will be difficult to explicitly remove the `drop` and change the `test_clients` test without realizing that you are removing important protections.

dignifiedquire (approved):

PR #3103, opened 2025-01-08
WIP Have relay actor remove itself from clients

PaulOlteanu:
## Description

Still WIP.

It seems to me that if a relay client actor errors almost anywhere ([for example when reading from the connection to the client](https://github.com/n0-computer/iroh/blob/c650ea83dae8e25165c9eb9b502d58113c7febc5/iroh-relay/src/server/client.rs#L219), which happens all the time) then that `Client` never gets removed from the `Clients` unless someone else tries sending a message to that node.

The `ConnectionId` thing is a pattern I've found useful for handling connection replacements. When the new connection gets put into the `Clients` list and then the old one gets shutdown, if we didn't check the connection id then the old connection would remove the new one from the list.

Also I haven't tested these changes at all, just interested in getting thoughts on them

## Breaking Changes

None?

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2318,6 +2318,7 @@ dependencies = [
  "tracing",
  "tracing-subscriber",
  "url",
+ "uuid",
  "webpki-roots",
 ]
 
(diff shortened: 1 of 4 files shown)


dignifiedquire:
created https://github.com/n0-computer/iroh/pull/3105 as a follow up

PR #3104, opened 2025-01-08
fix(iroh): Parse DNS answer with multiple records into a proper `NodeAddr`

matheus23:
## Description

Running `iroh-doctor` this morning with flub I wasn't able to connect. Flub was on iroh 0.30, I was on iroh main (commit c650ea83).
I noticed this in the logs:
```
2025-01-08T08:09:59.939435Z DEBUG connect{me=cabd2a9e0b alpn="n0/drderp/1"}:discovery{me=cabd2a9e0b node=1992d53c02}: hickory_proto::error: response: ; header 65474:RESPONSE:RD,RA:NoError:QUERY:3/0/0
; query
;; _iroh.dgjpkxyn3zyrk3zfads5duwdgbqpkwbjxfj4yt7rezidr3fijccy.dns.iroh.link. IN TXT
; answers 3
_iroh.dgjpkxyn3zyrk3zfads5duwdgbqpkwbjxfj4yt7rezidr3fijccy.dns.iroh.link. 30 IN TXT addr=192.168.96.145:60165
_iroh.dgjpkxyn3zyrk3zfads5duwdgbqpkwbjxfj4yt7rezidr3fijccy.dns.iroh.link. 30 IN TXT addr=213.208.157.87:60165
_iroh.dgjpkxyn3zyrk3zfads5duwdgbqpkwbjxfj4yt7rezidr3fijccy.dns.iroh.link. 30 IN TXT relay=https://euw1-1.relay.iroh.network./
; nameservers 0
; additionals 0

2025-01-08T08:09:59.939617Z DEBUG connect{me=cabd2a9e0b alpn="n0/drderp/1"}:discovery{me=cabd2a9e0b node=1992d53c02}: iroh::discovery: discovery: new address found provenance=dns addr=NodeAddr { node_id: PublicKey(1992d53c02cdc04566e5c0edb1ce83305cd550297953a047a445ea3264b54b18), relay_url: None, direct_addresses: {192.168.96.145:60165} }
[...]
2025-01-08T08:10:09.545100Z DEBUG ep{me=cabd2a9e0b}:magicsock:actor:stayin_alive{node=1992d53c02}: iroh::magicsock::node_map::node_state: can not send call-me-maybe, no relay URL
```

"no relay URL"? Even though the DNS record clearly contains one.

From the logs it showed that the `DiscoveryItem` didn't contain all information from the DNS answer.

This PR adds a test that initially failed - it only parsed the first TXT record out of the DNS answer, because of a call to `records.all`, which drained the iterator of all remaining items.

It also fixes this bug by avoiding a call to `.all`.

## Breaking Changes

None.

## Notes & open questions

I tried to avoid functional changes besides the bugfix. But, all of this parsing code makes me think of [Postel's Law / the robustness principle](https://en.wikipedia.org/wiki/Robustness_principle): Instead of erroring out when we see conflicting information - should we filter out non-matching answer records?
There seems to be some difference between `from_pkarr_signed_packet` and `from_hickory_records`: The former is very lenient in what it parses, the latter is very strict and opts to error out when something's amiss.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/dns/node_info.rs
+++ b/iroh/src/dns/node_info.rs
@@ -405,9 +405,14 @@ fn node_domain(node_id: &NodeId, origin: &str) -> Result<Name> {
 
 #[cfg(test)]
 mod tests {
-    use std::str::FromStr;
+    use std::{collections::BTreeSet, str::FromStr};
 
-    use iroh_base::SecretKey;
+    use hickory_resolver::{
+        proto::rr::{rdata::TXT, RData, Record},
+        Name,
+    };
+    use iroh_base::{NodeId, SecretKey};
+    use testresult::TestResult;
 
     use super::NodeInfo;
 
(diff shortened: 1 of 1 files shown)


dignifiedquire:
> The former is very lenient in what it parses, the latter is very strict and opts to error out when something's amiss.

the reasons is likely, that we "know" what we expect to publish here, but I agree we probably want to be more relaxed here, and just log invalid entries

iroh/src/dns/node_info.rs
```diff
@@ -311,24 +311,19 @@ impl<T: FromStr + Display + Hash + Ord> TxtAttrs<T> {
...
-                node_id_from_hickory_name(rr.name()).map(|node_id| (node_id, txt))
+    /// Parses a TXT records lookup.
+    pub fn from_hickory_lookup(lookup: hickory_resolver::lookup::TxtLookup) -> Result<Self> {
+        let node_id = node_id_from_hickory_name(lookup.query().name())
+            .ok_or_else(|| anyhow!("invalid DNS answer: not a query for _iroh.z32encodedpubkey"))?;
+
+        let strings = lookup.as_lookup().record_iter().filter_map(|record| {
+            match node_id_from_hickory_name(record.name()) {
+                // Filter out only TXT record answers that match the node_id we searched for.
+                Some(n) if n == node_id => record.data().as_txt().map(|txt| txt.to_string()),
```
dignifiedquire, line 322: maybe print a log line about invalid entries?

iroh/src/dns/node_info.rs
```diff
@@ -438,4 +442,55 @@ mod tests {
...
+    /// establishment.
+    ///
+    /// The reason was that only the first address was parsed (e.g. 192.168.96.145 in
+    /// this example), which could be a local, unreachable address.
+    #[test]
+    fn test_dns_answer_multiple_txt_records() -> TestResult {
+        let name = Name::from_utf8(
+            "_iroh.dgjpkxyn3zyrk3zfads5duwdgbqpkwbjxfj4yt7rezidr3fijccy.dns.iroh.link.",
+        )?;
+        let query = Query::query(name.clone(), RecordType::TXT);
```
dignifiedquire, line 457: also add an invalid one, to make sure the filtering works?

iroh/src/dns/node_info.rs
```diff
@@ -311,24 +311,19 @@ impl<T: FromStr + Display + Hash + Ord> TxtAttrs<T> {
...
-        let mut records = records.iter().filter_map(|rr| match rr.data() {
-            rr::RData::TXT(txt) => {
-                node_id_from_hickory_name(rr.name()).map(|node_id| (node_id, txt))
+    /// Parses a TXT records lookup.
+    pub fn from_hickory_lookup(lookup: hickory_resolver::lookup::TxtLookup) -> Result<Self> {
+        let node_id = node_id_from_hickory_name(lookup.query().name())
+            .ok_or_else(|| anyhow!("invalid DNS answer: not a query for _iroh.z32encodedpubkey"))?;
+
+        let strings = lookup.as_lookup().record_iter().filter_map(|record| {
+            match node_id_from_hickory_name(record.name()) {
```
dignifiedquire, line 320: you can just use `?` instead of the match statement, to return on `None`

matheus23: I also check `if n == node_id`.

matheus23:
I verified that this fixes the issue connecting between me and flub using `iroh-doctor` again.

PR #3107, opened 2025-01-08
feat(iroh): implement Discovery for Arc'ed Discovery types

flub:
## Description

Nothing in the trait stops it from working for discovery structs which
are Arc'ed.  Not all discovery types are in control of the users so
this is helpful.

I stumbled upon this because StaticProvider is not Clone.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/discovery.rs
+++ b/iroh/src/discovery.rs
@@ -447,7 +447,6 @@ mod tests {
 
     use iroh_base::SecretKey;
     use rand::Rng;
-    use testresult::TestResult;
     use tokio_util::task::AbortOnDropHandle;
 
     use super::*;
--- a/iroh/src/discovery.rs
+++ b/iroh/src/discovery.rs
@@ -111,7 +111,7 @@
     doc = "[`LocalSwarmDiscovery`]: local_swarm_discovery::LocalSwarmDiscovery"
 )]
 
-use std::{collections::BTreeSet, net::SocketAddr, time::Duration};
+use std::{collections::BTreeSet, net::SocketAddr, sync::Arc, time::Duration};
 
 use anyhow::{anyhow, ensure, Result};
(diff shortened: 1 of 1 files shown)


iroh/src/discovery.rs
```diff
@@ -725,6 +728,21 @@ mod tests {
             .expect("time drift")
             .as_micros() as u64
     }
+
+    #[tokio::test]
+    async fn test_arc_discovery() -> TestResult {
```
dignifiedquire, line 733: maybe use something like static_assert instead?

flub: I'm unsure what you have in mind.  Could you elaborate a bit more?

dignifiedquire: using https://crates.io/crates/static_assertions you can just check for trait baounds, instead of doing this manual adding

flub: I guess that has [`assert_impl_all`](https://docs.rs/static_assertions/1.1.0/static_assertions/macro.assert_impl_all.html) which I could use.  But then, how?  Adding it at build-time would just put it right below the impl... which seems a bit odd, because it's implemented right there.  Plus it would add a new build-time dependency.

I can use it in the test I guess, but again a new dependency seems like a lot.

I added this test mostly to assure myself I did the right thing.  But talking about it here seems to have little value?  Maybe it should just be removed instead?

dignifiedquire: > Maybe it should just be removed instead?

maybe, not sure testing this makes some sense, but also it's a static thing, so I don't think there is much value in the test

flub: removed, PTAL

dignifiedquire (approved):

PR #3111, opened 2025-01-09
feat(iroh): Allow customising the TransportConfig for connections

### LINKED ISSUE #2872: Iroh Connection cannot adjust TransportConfig for outgoing connections
I was advised to create an issue from https://github.com/n0-computer/iroh/discussions/2860#discussioncomment-11100552. Iroh-net's [Endpoint::connect](https://docs.rs/iroh-net/latest/iroh_net/endpoint/struct.Endpoint.html#method.connect) is [hard-coded](https://github.com/n0-computer/iroh/blob/fad3e24b3f2698ce6c1fa3fdad54201bec668298/iroh-net/src/endpoint.rs#L660) to use a TransportConfig of every-second keepalives with 30-second timeouts (latter default from Quinn), regardless of any TransportConfig that was passed to [Builder::transport_config](https://docs.rs/iroh-net/latest/iroh_net/endpoint/struct.Builder.html#method.transport_config) (which is only used for incoming connections). This effectively means that you can only increase the keepalive interval and reduce the timeout from these defaults.

I can't think of a great reason that the incoming and outgoing TransportConfig values should ever be different, so I think the default should be to use the Endpoint's configured values for outgoing connections. However, if there is a use case for this (perhaps different ALPNs should have different timeouts), a method like `connect_with(node_addr, alpn, transport_config)` could work well.

flub:
## Description

Previously the TransportConfig was hardcoded for the
client (client/server in QUIC terminology: client initiates the
connection, server accepts the connection).  This rendered some
customisations on the server-side, as supported by the Builder,
impossible since the client would negotiate different limits.

Specifically the QUIC keep-alives were at a 1s or faster interval
while the connection timeout was at 30s or less.

This adds a new API marked as experimental which allows customising
the transport config for clients.

Secondly it also respects any custom TransportConfig set in the
Builder for clients as well as servers.  This is not a behaviour
change since the parameters set by default are now moved to the
Builder and due to how these are negotiated it does not matter that
the server will now also have the same values set by default.

Closes #2872.

### iroh

If Builder::transport_config is used it will now also affect initiated
connections, while before it only affected the accepted connections.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -380,6 +380,9 @@ impl Builder {
     /// internet applications. Applications protocols which forbid remotely-initiated
     /// streams should set `max_concurrent_bidi_streams` and `max_concurrent_uni_streams` to
     /// zero.
+    ///
+    /// Please be aware that changing some settings may have adverse effects on establishing
+    /// and maintaining direct connections.
     pub fn transport_config(mut self, transport_config: quinn::TransportConfig) -> Self {
         self.transport_config = transport_config;
         self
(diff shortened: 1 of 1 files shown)


matheus23 (approved):
Generally not opposed to this - but I *do* want to note that this goes in the other direction in some aspect of where we're moving the codebase currently: This is exposing *more* of quinn in our top-level API, not less. So eventually I guess we'll need to wrap the transport config, too?

PR #3112, opened 2025-01-09
feat: Make `Endpoint::close` infallible

matheus23:
## Description

The only source for a `Result::Err` in `Endpoint::close` was a failed magic sock actor `Shutdown` message send.

This send should basically shutdown the actor, and sending only fails when the actor dropped the receiver, which only happens when the actor already shut down, gracefully or otherwise. In any case, it can't react to that anyways and it's not worth surfacing that error to the user.

## Breaking Changes

- `iroh::Endpoint::close`'s future is now infallible, instead of returning a `Result`.

## Change checklist

- [x] Self-review.
- ~~[ ] Tests if relevant.~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ let response = recv.read_to_end(1000).await?;
 assert_eq!(&response, b"Hello, world!");
 
 // Close the endpoint and all its connections
-endpoint.close().await?;
+endpoint.close().await;
 ```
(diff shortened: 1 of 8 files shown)


dignifiedquire (approved):
love it

flub (approved):
More of this!

PR #3113, opened 2025-01-09
feat(iroh-relay): send regular pings to check the connection

dignifiedquire:
## Description

Because of issues with TCP, we now send regular ping-pongs from client to server and vice versa.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-relay/src/ping_tracker.rs
+++ b/iroh-relay/src/ping_tracker.rs
@@ -2,6 +2,9 @@ use std::time::{Duration, Instant};
 
 use tracing::debug;
 
+/// Maximum time for a ping response in the relay protocol.
+pub const PING_TIMEOUT: Duration = Duration::from_secs(5);
+
 /// Tracks pings on a single relay connection.
 ///
 /// Only the last ping needs is useful, any previously sent ping is forgotten and ignored.
@@ -17,6 +20,12 @@ struct PingInner {
     deadline: Instant,
 }
 
+impl Default for PingTracker {
+    fn default() -> Self {
+        Self::new(PING_TIMEOUT)
+    }
+}
+
 impl PingTracker {
     /// Creates a new ping tracker, setting the ping timeout for pings.
     pub fn new(default_timeout: Duration) -> Self {
@@ -24,6 +33,11 @@ impl PingTracker {
             inner: None,
             default_timeout,
         }
+    }
+
+    /// Returns the current timeout set for pings.
+    pub fn default_timeout(&self) -> Duration {
+        self.default_timeout
     }
 
     /// Starts a new ping.
--- a/iroh-relay/src/ping_tracker.rs
+++ b/iroh-relay/src/ping_tracker.rs
@@ -1,0 +1,64 @@
+use std::time::{Duration, Instant};
+
+use tracing::debug;
+
+/// Tracks pings on a single relay connection.
+///
+/// Only the last ping needs is useful, any previously sent ping is forgotten and ignored.
+#[derive(Debug)]
+pub struct PingTracker {
+    inner: Option<PingInner>,
+    default_timeout: Duration,
+}
+
+#[derive(Debug)]
+struct PingInner {
+    data: [u8; 8],
+    deadline: Instant,
+}
+
+impl PingTracker {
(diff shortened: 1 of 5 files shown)


iroh/src/magicsock/relay_actor.rs
```diff
@@ -623,13 +641,19 @@ impl ActiveRelayActor {
         state: &mut ConnectedRelayState,
         client_stream: &mut iroh_relay::client::ClientStream,
     ) -> Result<()> {
+        const SEND_TIMEOUT: Duration = Duration::from_secs(10); // TODO: what should this be?
```
dignifiedquire, line 644: what should this be?

iroh-relay/src/ping_tracker.rs
```diff
@@ -0,0 +1,64 @@
...
+
+#[derive(Debug)]
+struct PingInner {
+    data: [u8; 8],
+    deadline: Instant,
+}
+
+impl PingTracker {
+    /// Creates a new ping tracker, setting the ping timeout for pings.
+    pub fn new(default_timeout: Duration) -> Self {
```
flub, line 22: Why make the timeout configurable if both users use the same timeout?

Also both users define their own constant for the timeout right now.  I'd be tempted to put the constant in this module and hardcode it again.  But I don't feel very strongly about this.

iroh-relay/src/protos/relay.rs
```diff
@@ -41,8 +41,17 @@ const MAX_FRAME_SIZE: usize = 1024 * 1024;
 /// The Relay magic number, sent in the FrameType::ClientInfo frame upon initial connection.
 const MAGIC: &str = "RELAY🔑";
 
+/// Maximum time for a relay server to respond to a relay protocol ping.
 #[cfg(feature = "server")]
-pub(crate) const KEEP_ALIVE: Duration = Duration::from_secs(60);
+pub(crate) const PING_TIMEOUT: Duration = Duration::from_secs(5);
+
+/// Interval in which we ping the relay server to ensure the connection is alive.
```
flub, line 48: This is used by the client to ping the server.  But maybe just make it pub so the magicsock can import this too and it's only defined once?

iroh/src/magicsock/relay_actor.rs
flub: contraints are:
- MAX_PACKET_SIZE is set to 64 KiB
- The client sends up to 20 packets at once (if this is too annoying we could take out that batching)
- so worst-case a single send is 200 KiB
- Practically I don't think 64 KiB packets will be sent
- 1500 bytes/packet is more likely
- so ~30 KiB is a more typical max send size

I'm not sure what lower limit of bandwidth we should assume here.

flub (commented):
Mostly nits, only thing I'm not sure on yet is what is a reasonable send timeout for the magicsock

iroh-relay/src/ping_tracker.rs
dignifiedquire: I am not against that, I just wasn't sure we'd always want to have the same numbers 🤷

flub:
One feature we haven't considered here yet is to defer sending the pings if there's real traffic.  That is, every time you receive some data you can reset the ping interval and postpone sending a ping.  This is approximately what QUIC does.

dignifiedquire:
> One feature we haven't considered here yet is to defer sending the pings if there's real traffic. That is, every time you receive some data you can reset the ping interval and postpone sending a ping. This is approximately what QUIC does.

Yes, I think we should do that 👍 this makes the overhead non existent during activity

iroh/src/magicsock/relay_actor.rs
dignifiedquire: set to  `PING_INTERVAL` for now, as discussed

iroh-relay/src/protos/relay.rs
dignifiedquire: fixed

flub (approved):

temoxtrading:
He

temoxtrading:
https://github.com/user-attachments/assets/8bfad2a0-7909-404c-9e4a-7a74a67b6f78

PR #3116, opened 2025-01-10
feat(iroh-net): add deprecation notice, bump to 0.28.2

### LINKED ISSUE #3075: iroh-net: Docs should point to iroh
If you look at https://crates.io/crates/iroh-net or https://docs.rs/iroh-net/latest/iroh_net/ there is no indication that the crate is discontinued and that the `iroh` crate should be used.  We should probably do a point release with a warning in the README and toplevel crate docs.

flub:
## Description

Too many users are finding this crate and trying to use it.  Make it obvious where they should be looking instead.

Closes #3075.

## Notes & open questions

- This targets the new `maint-0.28` branch which starts off the `v0.28.1` tag.
- Only the iroh-net crate has a new version number, I believe only publishing this crate makes the most sense.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net/src/lib.rs
+++ b/iroh-net/src/lib.rs
@@ -1,5 +1,12 @@
 //! Peer-to-peer QUIC connections.
 //!
+//! <div class="warning">
+//!
+//! The `iroh-net` crate has been renamed to `iroh`.  Please use the
+//! `iroh` crate instead.
+//!
+//! </div>
+//!
 //! iroh-net is a library to establish direct connectivity between peers.  It exposes an
 //! interface to [QUIC] connections and streams to the user, while implementing direct
 //! connectivity using [hole punching] complemented by relay servers under the hood.
--- a/iroh-net/src/lib.rs
+++ b/iroh-net/src/lib.rs
@@ -237,6 +237,10 @@
 //! [`Connection::open_bi`]: crate::endpoint::Connection::open_bi
 //! [`Connection::accept_bi`]: crate::endpoint::Connection::accept_bi
 
+#![deprecated(
+    note = "This crate has been renamed from 'iroh-net' to 'iroh', please use the new crate"
+)]
+#![allow(deprecated)]
 #![recursion_limit = "256"]
 #![deny(missing_docs, rustdoc::broken_intra_doc_links)]
 #![cfg_attr(iroh_docsrs, feature(doc_cfg))]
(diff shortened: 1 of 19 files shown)


flub:
`cargo publish -p iroh-net --dry-run` works on this branch.

matheus23:
Should we perhaps put a deprecation on ... all of the API? So if they use it, their editor tells them to use the new crate?

I tested a basic
```rust
#![deprecated(
    note = "this crate has been renamed from 'iroh-net' to 'iroh', please use that crate instead"
)]
```
in `iroh-net/src/lib.rs`. Seems to work.

flub:
> Should we perhaps put a deprecation on ... all of the API? So if they use it, their editor tells them to use the new crate?

I think that's a bit harsh.  There is nothing wrong with this crate and there are legitimate reasons why someone might not have upgraded yet.

matheus23:
Well, to be fair, people would only see that when the install the *latest* version of `iroh-net`, not if they just keep depending on `0.28.1`, which wouldn't have these deprecation warnings.

iroh-net/src/lib.rs
```diff
@@ -1,5 +1,12 @@
 //! Peer-to-peer QUIC connections.
 //!
+//! <div class="warning">
+//!
+//! The `iroh-net` crate has been renamed to `iroh`.  Please use the
+//! `iroh` crate instead.
+//!
+//! </div>
+//!
```
dignifiedquire, line 9: isn't there sth like `#![deprecated]` that we can do?

matheus23: Flub answered this question here: https://github.com/n0-computer/iroh/pull/3116#issuecomment-2582743069

dignifiedquire: I think we should do it, otherwise I think this will be very easily missed by downstream users, you can always use `[allow(deprectated)]`

matheus23 (approved):

dignifiedquire (approved):

matheus23:
@flub can we merge this? And can we get someone who has release rights (@dignifiedquire @ramfox ?) to cut a release?

flub:
Does it work though?  The adding of deprecations had all sorts of weird errors that I didn't dig into, so I'm not sure this is actually releasable as-is.

flub:
Ok, ran tests, built the docs, did a publish dry-run and all is fine and dandy.  I did all this with just `-p iroh-net` not for all crates.  So I'm going to merge this.

PR #3117, opened 2025-01-10
fix(iroh): `AsyncUdpSocket` backpressure when connection type is *not* mixed

matheus23:
## Description

With the previous logic, when `relay_url` is `None`, then `relay_pending` is `false`.
The same is true with `udp_addr` and `udp_pending`.
However, we only ever report `WouldBlock` if `udp_pending` *and* `relay_pending`.
So this means we only ever report `WouldBlock` if `relay_url` isn't `None` and `udp_addr` isn't `None`, so when we're in mixed connection mode.
Because in e.g. direct connection mode, `relay_url` is `None`, thus `relay_pending` is `false` and we never return `Err(WouldBlock)`.

## Notes & open questions

~~I'm curious about netsim numbers.~~ No changes

Maybe this should have a test. EDIT: Hm this is really hard to test. I can't find a way to 'induce' `WouldBlock` in the relay and/or socket without fairly invasive changes that would allow replacing the socket & relay implementations.

## Change checklist

- [x] Self-review.
- ~~[ ] Tests if relevant.~~ Very hard to test this.

### DIFF
--- a/iroh/src/magicsock.rs
+++ b/iroh/src/magicsock.rs
@@ -545,13 +545,16 @@ impl MagicSock {
                     .map(|err| err.kind() == io::ErrorKind::WouldBlock)
                     .unwrap_or_default();
 
-                let udp_path = udp_addr.is_some();
-                let relay_path = relay_url.is_some();
-                let has_path = udp_path || relay_path;
+                let would_block = match (&udp_addr, &relay_url) {
+                    (Some(_udp), Some(_relay)) => udp_pending && relay_pending,
+                    (Some(_udp), None) => udp_pending,
+                    (None, Some(_relay)) => relay_pending,
+                    (None, None) => false,
+                };
 
-                if (udp_pending || !udp_path) && (relay_pending || !relay_path) && has_path {
+                if would_block {
                     // Handle backpressure.
-                    inc!(MagicsockMetrics, send_pending);
+                    inc!(MagicsockMetrics, send_would_block);
                     Err(io::Error::new(io::ErrorKind::WouldBlock, "pending"))
                 } else {
                     if relay_sent || udp_sent {
--- a/iroh/src/magicsock.rs
+++ b/iroh/src/magicsock.rs
@@ -551,6 +551,7 @@ impl MagicSock {
 
                 if udp_pending && relay_pending && has_path {
                     // Handle backpressure.
+                    inc!(MagicsockMetrics, send_pending);
                     Err(io::Error::new(io::ErrorKind::WouldBlock, "pending"))
                 } else {
                     if relay_sent || udp_sent {
--- a/iroh/src/magicsock.rs
+++ b/iroh/src/magicsock.rs
@@ -535,15 +535,21 @@ impl MagicSock {
                     }
                 }
 
-                let udp_pending = udp_error
-                    .as_ref()
-                    .map(|err| err.kind() == io::ErrorKind::WouldBlock)
-                    .unwrap_or_default();
-                let relay_pending = relay_error
-                    .as_ref()
-                    .map(|err| err.kind() == io::ErrorKind::WouldBlock)
-                    .unwrap_or_default();
-                if udp_pending && relay_pending {
+                let udp_pending = udp_addr.is_some()
+                    && udp_error
+                        .as_ref()
+                        .map(|err| err.kind() == io::ErrorKind::WouldBlock)
+                        .unwrap_or_default();
+
+                let relay_pending = relay_url.is_some()
+                    && relay_error
+                        .as_ref()
+                        .map(|err| err.kind() == io::ErrorKind::WouldBlock)
+                        .unwrap_or_default();
+
+                let has_path = udp_addr.is_some() || relay_url.is_some();
+
+                if udp_pending && relay_pending && has_path {
                     // Handle backpressure.
                     Err(io::Error::new(io::ErrorKind::WouldBlock, "pending"))
                 } else {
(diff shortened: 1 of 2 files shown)


matheus23:
Netsim numbers:

| iroh | 1_to_1 | 1.38 | 1.38 |
| iroh | 1_to_3 | 4.52 | 4.52 |
| iroh | 1_to_5 | 6.90 | 6.91 |
| iroh | 1_to_10 | 11.89 | 11.91 |
| iroh | 2_to_2 | 2.77 | 2.78 |
| iroh | 2_to_4 | 5.57 | 5.58 |
| iroh | 2_to_6 | 8.26 | 8.27 |
| iroh | 2_to_10 | 13.56 | 13.59 |

matheus23:
I haven't managed to get the `send_pending` metric to increase yet.

This means we practically (?) never run into the `udp_pending && relay_pending` case.

matheus23:
> Looking at the perf numbers I assume this has no impact and is all just noise?

I have never actually managed to hit the case where we return `WouldBlock` from `try_send`. Neither with this PR nor on `main`.

So yes, it's all very theoretical and all just noise :smiling_face_with_tear:

iroh/src/magicsock.rs
```diff
@@ -535,16 +535,23 @@ impl MagicSock {
...
+
+                let relay_pending = relay_url.is_some()
+                    && relay_error
+                        .as_ref()
+                        .map(|err| err.kind() == io::ErrorKind::WouldBlock)
+                        .unwrap_or_default();
+
+                let has_path = udp_addr.is_some() || relay_url.is_some();
+
+                if udp_pending && relay_pending && has_path {
```
flub, line 552: Is this the right correct?

- assume relay-only path
- udp_addr is none
- so udp_pending is false
- relay_url is some
- relay_pending is true (assuming it blocked)
- has_path is true
- udp_pending (false) && .... well, doesn't matter anymore.  this is false now and it won't block?

matheus23: > - udp_addr is none
> - so udp_pending is false

Yeah - Ugh, I force-pushed a change that made this implication true (originally I had the logic correct).

Will fix.

matheus23: Updated.

matheus23:
> You might have more luck seeing the blocked metric go up if you test relay-only since then you have a single path. You can even go further and setup a relay server which rate-limits client connections so it is easier to fill up the magicsock pipes.

Tried relay-only with rate limits. No effect :/ `send_pending == 0`.

matheus23:
> I had previously held off this because I was worried that you do not know the next packet that is going to be sent. But I think my logic was wrong: you do know the next packet: it will be the same one. So I think it's interesting to try this out.

But then why return `Err(WouldBlock)` at all? And why do it for *only* the case where the connection type is mixed?

iroh/src/magicsock.rs
```diff
@@ -539,12 +539,19 @@ impl MagicSock {
...
                     .as_ref()
                     .map(|err| err.kind() == io::ErrorKind::WouldBlock)
                     .unwrap_or_default();
-                if udp_pending && relay_pending {
+
+                let udp_path = udp_addr.is_some();
+                let relay_path = relay_url.is_some();
+                let has_path = udp_path || relay_path;
+
+                if (udp_pending || !udp_path) && (relay_pending || !relay_path) && has_path {
```
flub, line 552: Maybe I'm not that good at reading boolean logic, but I don't find this the most readable.  Would this be the same:
```rust
let would_block = match (&udp_addr, &relay_url) {
    (Some(_udp), Some(_relay)) => udp_pending && relay_pending,
    (Some(_udp), None) => udp_pending,
    (None, Some(_relay)) => relay_pending,
    (None, None) => false,
};
if would_block {
    // Handle backpressure.
    todo!()
}
```
And it is a more readable way to write this?  There might well be even further simplifications.  I tried one variant with not even making the `udp_pending` and `relay_pending` variables but that didn't end up clearer the way I tried it.

iroh/src/magicsock/metrics.rs
```diff
@@ -20,6 +20,7 @@ pub struct Metrics {
     // Data packets (non-disco)
     pub send_data: Counter,
     pub send_data_network_down: Counter,
+    pub send_pending: Counter,
```
flub, line 23: I know existing metric names do not do this, but could we follow https://prometheus.io/docs/practices/naming/ for new metric names?

I think this would be `send_would_block_total`, the prefix is already done by the framework here.

/cc @Arqu

Arqu: The current setup automatically adds _total to all counters, unless you define a unit on the counter which we currently don't do anywhere. So technically we are using the correct metrics naming convention, just not utilizing it fully.

flub (approved):

flub:
> > I had previously held off this because I was worried that you do not know the next packet that is going to be sent. But I think my logic was wrong: you do know the next packet: it will be the same one. So I think it's interesting to try this out.
> 
> But then why return `Err(WouldBlock)` at all? And why do it for _only_ the case where the connection type is mixed?

Good questions.

I think my above statement is a bit wrong.  The reason we didn't do this **is** because we do not know where the next packet is going to be sent.  If you have a direct path to one node, but relay paths another node you do not want to be blocking the direct path because the relay path is congested.

Which still makes this change somewhat debatable (and I'm not sure I should have approved).  We should certainly do a lot more testing before we commit to a change like this.

matheus23:
I agree with flub: This PR not having any actual effect (and the code path not being hit) is a sign I don't fully understand what's happening here. On top of this - this PR doesn't fix a concrete issue we're having. I'll keep this in mind, but don't have the time to look through the code paths from quinn to iroh AsyncUdpSocket to understand what exactly would need to be done.

PR #3118, opened 2025-01-10
fix: Try improving `relay_datagram_send_channel()`

### LINKED ISSUE #3067: Relay send path sets wakers incorrectly and might report `Poll::Pending` more than necessary
https://github.com/n0-computer/iroh/pull/3062/commits/20016168449c6df3ca3905553905ac728355e274

This likely doesn't turn into a pressing problem in practice, but could become an issue eventually/could help improve performance.

A lot of the difficulty in fixing this is working around the API bottleneck that is the `&self` reference in the `AsyncUdpSocket::try_send` API.

matheus23:
## Description

Closes #3067 

Since multiple `Connection`s can have multiple `AsyncUdpSocket` `IpPollers`, it's incorrect to assume that a single `AtomicWaker` can wake up these tasks correctly.
Instead, if there's only one `AtomicWaker` for all of them, only the last registered waker will be woken when there's capacity again.

I'm changing this such that `poll_writable` will actually add the current task's waker to a list, if it hasn't been added yet.
And successfully `recv`ing an item will wake all tasks.

A second issue was that *between* the call to `self.sender.capacity()` (it being 0) and `self.waker.register`, another thread might have successfully `recv`d an item *and* called `self.waker.wake()`.
This means that the first thread could've registered a waker, while the capacity is actually non-zero. Not terrible (it's very likely that it'll be woken up when the capacity jumps to 2 this time), but also not great.
The fix is to re-check the capacity after registering the waker.
The downside is that we keep the waker and thus potentially get a spurious wakeup, but that's fine.

## Notes & open questions

This... "fixes" the concerns, but I'd actually like to write a loom test for the concerns eventually.

It's just... a *lot* of work, and I'd rather prioritize other things, but also improve this part of the code at the same time.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/magicsock.rs
+++ b/iroh/src/magicsock.rs
@@ -25,7 +25,7 @@ use std::{
         atomic::{AtomicBool, AtomicU16, AtomicU64, AtomicUsize, Ordering},
         Arc, RwLock,
     },
-    task::{Context, Poll},
+    task::{Context, Poll, Waker},
     time::{Duration, Instant},
 };
 
@@ -1548,7 +1548,7 @@ impl Handle {
 
         let (actor_sender, actor_receiver) = mpsc::channel(256);
         let (relay_actor_sender, relay_actor_receiver) = mpsc::channel(256);
-        let (relay_datagram_send_tx, relay_datagram_send_rx) = relay_datagram_sender();
+        let (relay_datagram_send_tx, relay_datagram_send_rx) = relay_datagram_send_channel();
         let (udp_disco_sender, mut udp_disco_receiver) = mpsc::channel(256);
 
         // load the node data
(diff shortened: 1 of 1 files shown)


matheus23:
Netsim results:

| iroh | 1_to_1 | 1.41 | 1.41 |
| iroh | 1_to_3 | 4.26 | 4.27 |
| iroh | 1_to_5 | 7.20 | 7.21 |
| iroh | 1_to_10 | 12.25 | 12.27 |
| iroh | 2_to_2 | 3.03 | 3.03 |
| iroh | 2_to_4 | 5.57 | 5.57 |
| iroh | 2_to_6 | 9.03 | 9.05 |
| iroh | 2_to_10 | 13.69 | 13.74 |

versus main:

| iroh | 1_to_1 | 1.38 | 1.38 |
| iroh | 1_to_3 | 4.27 | 4.28 |
| iroh | 1_to_5 | 6.99 | 7.00 |
| iroh | 1_to_10 | 12.01 | 12.03 |
| iroh | 2_to_2 | 3.06 | 3.07 |
| iroh | 2_to_4 | 5.86 | 5.87 |
| iroh | 2_to_6 | 9.20 | 9.22 |
| iroh | 2_to_10 | 13.62 | 13.65 |

No noticable difference.

Relay-only performance might be impacted though? Don't know if we're tracking that somewhere.

flub (approved):
nice!

PR #3119, opened 2025-01-10
feat(iroh-relay): Make the client side of `iroh-relay` compile & run in browsers

matheus23:
## Description

Allows compiling and running `iroh-relay` in browsers (as long as the `server` feature is disabled).

By default uses websockets to connect to relays when compiling for the `wasm32-unknown-unknown` target.
If configured to `Protocol::Relay` specifically, it'll error out.

Test it yourself: `cargo rustc -p iroh-relay --no-default-features --target=wasm32-unknown-unknown --crate-type=cdylib`

## Breaking Changes

- iroh: `iroh-quinn` re-exports are now updated from version 0.12 to version 0.13

## Notes & open questions

**Note:** This updates *all* crate's versions to `iroh-quinn` 0.13!
(Because we need them to be consistent between crates in this workspace.)

## TODO

- [x] Add CI check to make sure the built .wasm is `(import "env" ...)`-free
- [x] Need a new release of `iroh-quinn` with my browser PR in
- [x]  ~~**Blocking this from merging**: Need a new `netwatch` release with the `quinn-udp = "0.5.5"` version bound relaxed~~ it's already relaxed

## Change checklist

- [x] Self-review.


iroh-relay/src/client.rs
```diff
@@ -143,45 +141,21 @@ impl ClientBuilder {
...
-        .with_no_client_auth();
-        #[cfg(any(test, feature = "test-utils"))]
-        if self.insecure_skip_cert_verify {
-            warn!("Insecure config: SSL certificates from relay servers not verified");
-            config
-                .dangerous()
-                .set_certificate_verifier(Arc::new(NoCertVerifier));
-        }
-        config.resumption = Resumption::default();
-        let tls_connector: tokio_rustls::TlsConnector = Arc::new(config).into();
```
matheus23, line 164: Instead of immediately constructing the `tls_connector` in `connect`, and then either using it or discarding it depending on whether we use `connect_ws` or `connect_relay`, I've moved it into `connect_relay`.

iroh-relay/src/client.rs
```diff
@@ -211,94 +185,6 @@ impl ClientBuilder {
         Ok(conn)
     }
 
-    async fn connect_relay(
```
matheus23, line 214: `connect_relay` and all of its call graph dependencies are now moved into a submodule `iroh_relay::client::connect_relay`, this makes cfg-ing out much easier.

iroh-relay/Cargo.toml
```diff
@@ -28,69 +27,81 @@ futures-buffered = "0.2.9"
 futures-lite = "2.5"
 futures-sink = "0.3"
 futures-util = "0.3"
```
dignifiedquire, line 29: these `futures-*` deps should be dropped in favor of `n0-future`?

iroh-relay/src/client.rs
```diff
@@ -71,23 +57,36 @@ pub struct ClientBuilder {
...
+    ) -> Self {
         ClientBuilder {
             address_family_selector: None,
             is_prober: false,
             url: url.into(),
+
+            // always prefer websockets in browser contexts
+            #[cfg(wasm_browser)]
+            protocol: Protocol::Websocket,
+            #[cfg(not(wasm_browser))]
```
dignifiedquire, line 81: maybe it would be nicer to `impl Default` with cfgs for `Protocol`?

iroh-relay/src/client.rs
```diff
@@ -143,45 +142,21 @@ impl ClientBuilder {
...
             }
+            #[cfg(not(wasm_browser))]
             Protocol::Relay => {
-                let (conn, local_addr) = self.connect_relay(tls_connector).await?;
+                let (conn, local_addr) = self.connect_relay().await?;
                 (conn, Some(local_addr))
             }
+            #[cfg(wasm_browser)]
+            Protocol::Relay => {
+                bail!("Can only connect to relay using websockets in browsers.");
```
dignifiedquire, line 158: this error mesage seems wrong

dignifiedquire:
other than my comments this is looking pretty good

iroh-relay/src/client/conn.rs
```diff
@@ -219,13 +227,15 @@ impl Sink<SendMessage> for Conn {
...
         match *self {
+            #[cfg(not(wasm_browser))]
             Self::Relay { ref mut conn } => Pin::new(conn).poll_flush(cx).map_err(Into::into),
             Self::Ws { ref mut conn, .. } => Pin::new(conn).poll_flush(cx).map_err(Into::into),
         }
     }
 
     fn poll_close(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
         match *self {
+            #[cfg(not(wasm_browser))]
```
dignifiedquire, line 238: just noting this down: it's really unfortunate we have to do this on all the branches, but I don't have a better idea

dignifiedquire (approved):

iroh-relay/src/client/conn.rs
matheus23: Yeah I think we can remove them, once/if we go websocket connections only.

PR #3121, opened 2025-01-13
feat: allow dns discovery in transfer example

Arqu:
## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/examples/transfer.rs
+++ b/iroh/examples/transfer.rs
@@ -8,6 +8,7 @@ use bytes::Bytes;
 use clap::{Parser, Subcommand};
 use indicatif::HumanBytes;
 use iroh::{
+    discovery::pkarr::PkarrPublisher,
     endpoint::{ConnectionError, PathSelection},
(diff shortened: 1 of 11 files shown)


iroh/examples/transfer.rs
```diff
@@ -32,6 +33,10 @@ enum Commands {
         relay_url: Option<String>,
         #[clap(long, default_value = "false")]
         relay_only: bool,
+        #[clap(long)]
+        pkarr_relay_url: Option<String>,
+        #[clap(long)]
+        dns_url: Option<String>,
```
flub, line 39: Having this named `dns_url` is very confusing.  I had to look up the `DnsDiscovery` docs to understand what was going on.  Why not call this `dns_origin_domain`?  (And of course everywhere else too)

flub (approved):
LGTM, though I do think you shouldn't call that a `dns_url`.  There's nothing URL about it.

PR #3124, opened 2025-01-13
chore: Remove checkboxes from PR templates

flub:
## Description

A significant number of our own PRs don't bother with the checkboxes.
This makes the checkboxes themselves have low value, the number of
completed tasks in the PR list has no bearing on the state of PRs.

It also makes it hard to enforce (I've tried!) and does not give us
anything to ask from external contributors either.

I'm also often torn what to do with non-applicable items, I've tried
all of:
- Remove them.  Weird.
- Strike them out, not so weird, but so much effort.
- Tick off even if not-applicable (my current approach), also weird.

So instead of bothering the humans I'm adapting to their behaviour.

## Notes & open questions

An alternative I could live with would be enforce this by machines:
Have a required CI check that needs the checkboxes ticked off so that
you can not merge a PR while ignoring this.  Maybe with some more
guidance on what to do if not applicable.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -12,7 +12,9 @@
 
 ## Change checklist
 
-- [ ] Self-review.
-- [ ] Documentation updates following the [style guide](https://rust-lang.github.io/rfcs/1574-more-api-documentation-conventions.html#appendix-a-full-conventions-text), if relevant.
-- [ ] Tests if relevant.
-- [ ] All breaking changes documented.
+Please make sure to consider all of these:
+
+- Self-review.
+- Documentation updates following the [style guide](https://rust-lang.github.io/rfcs/1574-more-api-documentation-conventions.html#appendix-a-full-conventions-text).
+- Tests.
+- Document breaking changes.


.github/pull_request_template.md
```diff
@@ -12,7 +12,9 @@
...
-- [ ] Self-review.
-- [ ] Documentation updates following the [style guide](https://rust-lang.github.io/rfcs/1574-more-api-documentation-conventions.html#appendix-a-full-conventions-text), if relevant.
-- [ ] Tests if relevant.
-- [ ] All breaking changes documented.
+Please make sure to consider all of these:
+
+- Self-review.
+- Documentation updates following the [style guide](https://rust-lang.github.io/rfcs/1574-more-api-documentation-conventions.html#appendix-a-full-conventions-text).
+- Tests.
+- Document breaking changes.
```
dignifiedquire, line 20: maybe put them behind html comments, so they don't show up in the actual issue?

flub: Is that desired though?  Then folks will just not see this at all.

I get that it could be nice to hide them in the commit messages, but the html comments do not achieve that I think.

matheus23:
FWIW, I *at least* like the self-review checkbox, and regularly leave that unchecked in draft PRs so I remember to go through PRs once more when I undraft.

ramfox (commented):
Hmmm, to be honest, I really like having these as checkboxes for myself.

If one doesn't fit, I just delete the line & I think it's fine for a developer to decide that one of the todos isn't relevant for their PR.

I also don't mind having these in each merge commit, since it gives a general overview of "hey no tests were added or fixed" or "no breaking changes in this commit"

flub:
So FWIW my personal opinion is that I like these.  But I review so many PRs where they are ignored and that inconsistency is what bothers me as I never know how to handle it.  So this PR is my attempt at accepting the status quo.

I'd be fine with requiring these checkboxes, but if that's the case it should probably be done by automation rather than by reviewers.  So I'd be totally on board with the alternative of enforcing them in CI like we enforce the PR title already.

Or maybe the checkboxes are just optional to use in case the author likes to use them?


FWIW, I asked a review from everyone individually because I do think something like this needs broad acceptance.  I don't want to force this change at all.

flub:
> Perhaps being more clear about "the rules" here might be helpful. At the moment these checkboxes are here, but ... if you don't check them, then what? I'd propose adding some text along the lines of:
> 
> > These checkboxes are _for you_, to help ensure you're submitting complete pull requests that don't need to go through unnecessary rounds of review, and to make sure you don't forget things. Feel free to delete if these checkboxes are inappropriate for your PR.

I could totally live with this text.

> And then in terms of rules:
> 
>     1. Make the check-for-cc action ensure that all checkboxes (that are left in) are checked off.
> 
>     2. Allow deleting checkboxes entirely.

2 should certainly allowed.  1 is kind of optional with the suggested text, but we could do if it's simple to do.

flub:
So I learned that the presence of checkboxes does not imply the need to use checkboxes.  People consider them to be no more than a decoration and hint and I need to get over my OCD that every checkbox needs to be ticked.  I'll let checkboxes be themselves.

PR #3130, opened 2025-01-14
fix(iroh): Return error if disco send via relay fails

flub:
## Description

The signature of this function expects an error if the send fails, but
the relay part did not do this.  The impact of this is that you might
silently drop a packet, which makes debugging a bit harder.  Returning
an error should propagate this correctly and the failures will be
noticed and logged.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/magicsock.rs
+++ b/iroh/src/magicsock.rs
@@ -1214,7 +1214,9 @@ impl MagicSock {
                 self.try_send_disco_message_udp(addr, dst_key, &msg)?;
             }
             SendAddr::Relay(ref url) => {
-                self.send_disco_message_relay(url, dst_key, msg);
+                if !self.send_disco_message_relay(url, dst_key, msg) {
+                    return Err(io::Error::new(io::ErrorKind::Other, "Relay channel full"));
+                }
             }
         }
         Ok(())


matheus23:
Looking through this part of the code I'm realizing some unrelated things that could be fixed:
- It looks like there exists `MagicSock::handle_ping_actions` and it's apparently unused (but... rustc doesn't realize apparently due to the `#[instrument]` macro)
- Instead of returning a bool from `send_disco_message_relay` we should probably just return the error we get from `try_send_relay`, and later `.map_err`, so we don't forget such stuff in the future. Even `Result<(), ()>` > `bool`.

flub:
> Looking through this part of the code I'm realizing some unrelated things that could be fixed:
> 
>     * It looks like there exists `MagicSock::handle_ping_actions` and it's apparently unused (but... rustc doesn't realize apparently due to the `#[instrument]` macro)
> 
>     * Instead of returning a bool from `send_disco_message_relay` we should probably just return the error we get from `try_send_relay`, and later `.map_err`, so we don't forget such stuff in the future. Even `Result<(), ()>` > `bool`.

Yeah... this code hasn't yet seen a proper cleanup.

Also, sad that `#[instrument]` does that, but I probably shouldn't be surprised.

PR #3134, opened 2025-01-16
fix(iroh-relay): fix client actors not closing

PaulOlteanu:
## Description

When a client actor closes, it awaits `Clients::unregister`, which then awaits for the actor task to close, so almost every actor task was being leaked. Since the only place `Clients::unregister` is called is from when the actor is already closing, I chose to break the cycle by not have it call `Client::shutdown`.

I also added a timeout to the tls handshake because that also gets stuck forever if the client goes away. I didn't add it to the `LetsEncrypt` acceptor because I wasn't sure where exactly it belonged, but that should probably have one too. I can move this to a separate PR if preferred. Maybe the timeout should be configurable as well?

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


flub (approved):
Apologies for the long wait, guess this slipped through the cracks...

That's a reasonable way to break that loop.

I haven't looked much at the server side of the relay connections, but I suspect that this entire thing should have been turned on it's head so that the actor task exiting automatically unregisters it from the clients, like the task should be owned by the clients or something.  But let's leave that for another day.

Could you file an issue for the still missing timeouts you mention?

PR #3136, opened 2025-01-17
chore: new project syncing setup

Arqu:
## Description
Revamps how we track PRs and Issues to add onto repos. Mainly because this produced issues with forks but also did all the work attribution to me. 

The downside of this approach is that it runs on a schedule (set to every 6 hours but we can change). Also if GH api craps out on us we have to wait till the next trigger.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


.github/workflows/project_sync.yaml
```diff
@@ -0,0 +1,46 @@
+name: Add PRs and Issues to the Project Board
+
+on:
+  schedule:
+    - cron: '0 0,6,12,18 * * *'
```
Arqu, line 5: Do we want a different schedule?

flub (approved):

.github/workflows/project_sync.yaml
ramfox: I agree with every hour unless there are rate limits.

And also, if possible a manual trigger.

.github/workflows/project_sync.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+env:
+  REPOS_LIST: "iroh-doctor-server quinn"
```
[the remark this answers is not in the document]
ramfox: The ones we did for the last iteration should be right!

.github/workflows/project_sync.yaml
Arqu: Rate limits are per hour anyways but are also fairly generous with the Github App setup, should be fine.

Arqu: Also will add manual triggering.

PR #3139, opened 2025-01-20
feat(iroh-net-report): Support wasm32 building & running

matheus23:
## Description

This makes iroh-net-report compile to Wasm. We make sure it's likely to work by grepping the generated Wasm file for "env" imports (which would make it not work in browsers).
Mostly this will disable functionality in iroh-net-report for the browser target - specifically, it disables STUN, QAD, ICMP and all other probes besides HTTPS to relays.
I've experimentally verified this code works with an integration test, it's really hard to come up with a good strategy for automating this testing due to requiring running cross-languages.
I'd like to do that as a follow-up once #3145 is merged.

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh-net-report/src/lib.rs
+++ b/iroh-net-report/src/lib.rs
@@ -8,6 +8,7 @@
 #![cfg_attr(iroh_docsrs, feature(doc_auto_cfg))]
 #![deny(missing_docs, rustdoc::broken_intra_doc_links)]
 #![cfg_attr(not(test), deny(clippy::unwrap_used))]
+#![cfg_attr(wasm_browser, allow(unused))]
 
 use std::{
     collections::{BTreeMap, HashMap},
--- a/iroh-net-report/src/lib.rs
+++ b/iroh-net-report/src/lib.rs
@@ -244,6 +244,7 @@ impl Default for Reports {
 ///
 /// Use [`Options::stun_v4`], [`Options::stun_v6`], and [`Options::quic_config`]
 /// to enable STUN over IPv4, STUN over IPv6, and QUIC address discovery.
+#[cfg(not(wasm_browser))]
 #[derive(Debug, Clone)]
 pub struct Options {
     /// Socket to send IPv4 STUN probes from.
@@ -253,7 +254,6 @@ pub struct Options {
     /// other packets from in the magicsocket (`MagicSock`).
     ///
     /// If not provided, STUN probes will not be sent over IPv4.
-    #[cfg(not(wasm_browser))]
     stun_sock_v4: Option<Arc<UdpSocket>>,
     /// Socket to send IPv6 STUN probes from.
     ///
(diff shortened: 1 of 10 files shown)


matheus23:
The only thing that iroh-net-report would be doing in Wasm is running HTTPS probes.
Turns out we don't actually use them in iroh.

I have a browser version of iroh working that doesn't depend on iroh-net-report at all. This makes stuff a lot simpler.

So, I'm going to close this.

matheus23:
Hm, what I wrote above is incorrect. We do need the relay latencies to decide which relay to connect to.

iroh-net-report/src/lib.rs
```diff
@@ -86,6 +90,7 @@ pub struct Report {
     /// public IP address (on IPv4).
     pub hair_pinning: Option<bool>,
     /// Probe indicating the presence of port mapping protocols on the LAN.
+    #[cfg(not(wasm_browser))]
```
dignifiedquire, line 93: given how many things are cfged out, maybe it makes sense to split up the report in a couple of scoped structs

dignifiedquire: Actually, I just realized, this is mostly just options, so this is fine.

iroh-net-report/src/lib.rs
```diff
@@ -232,18 +238,22 @@ pub struct Options {
     /// other packets from in the magicsocket (`MagicSock`).
     ///
     /// If not provided, STUN probes will not be sent over IPv6.
+    #[cfg(not(wasm_browser))]
     stun_sock_v6: Option<Arc<UdpSocket>>,
```
dignifiedquire, line 242: maybe create `StunConfig` and `IcmpConfig`, like there is `QuicConfig` to group things a little nicer

iroh-net-report/src/reportgen.rs
```diff
@@ -87,12 +88,12 @@ impl Client {
...
+        #[cfg(not(wasm_browser))] port_mapper: Option<portmapper::Client>,
         relay_map: RelayMap,
-        stun_sock4: Option<Arc<UdpSocket>>,
-        stun_sock6: Option<Arc<UdpSocket>>,
-        quic_config: Option<QuicConfig>,
-        dns_resolver: DnsResolver,
+        #[cfg(not(wasm_browser))] stun_sock4: Option<Arc<UdpSocket>>,
+        #[cfg(not(wasm_browser))] stun_sock6: Option<Arc<UdpSocket>>,
+        #[cfg(not(wasm_browser))] quic_config: Option<QuicConfig>,
+        #[cfg(not(wasm_browser))] dns_resolver: DnsResolver,
```
dignifiedquire, line 96: this sucks, but this method already is problematic because it takes too many arguments

matheus23: Yeah this whole pr is just in the "getting it working" stage not yet in the "make it nice" one.
I think separating some of these things will be possible.

iroh-net-report/src/lib.rs
matheus23: Yeah - I've opted to keep the cfgs in the `Options` struct now.

iroh-net-report/src/reportgen.rs
matheus23: I've taken these and grouped them together in a `SocketState` struct. This helped reduce arguments a bunch.

iroh-net-report/src/lib.rs
matheus23: I've simplified lots of these cfgs by plucking out a wasm-only version of `Options`, instead of doing a bunch of cfgs on the fields & methods.

iroh-net-report/src/lib.rs
```diff
@@ -344,17 +375,67 @@ impl Options {
...
+
+#[cfg(wasm_browser)]
+impl Default for Options {
+    fn default() -> Self {
+        Self { https: true }
+    }
+}
+
+#[cfg(wasm_browser)]
+impl Options {
```
dignifiedquire, line 399: maybe have a submodule for `options` one for wasm and one for the browser?

```rust

pub use options::Options;

#[cfg(wasm_browser)]
mod options {
  pub struct Options
}

#[cfg(not(wasm_browser))]
mod options {
  pub struct Options
}
```rust

dignifiedquire (approved):

PR #3142, opened 2025-01-21
refactor(iroh-net-report): Remove dead code around `check_captive_portal`

matheus23:
## Description

**DRAFT** because I'm not sure if this short-term refactor is the right way to go about this.

When looking at this code for Wasm, I noticed some obviously dead code.
I think the main problem is likely that the captive portal check is just unused and thus implemented incorrectly.
Perhaps we shouldn't merge this, but just remove the check entirely and resurrect it once we actually fix captive portal detection again.

## Notes & open questions

What this refactor shows is: Effectively the captive portal check will always do the check with *a random relay* from the relay map.
The reason for that is that it *only runs*, when there's no net report yet, which seems wrong, too.

Should we perhaps file an issue for a captive portal check investigation at some point?

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-net-report/src/reportgen.rs
+++ b/iroh-net-report/src/reportgen.rs
@@ -476,15 +476,8 @@ impl Actor {
         // delay by a bit to wait for UDP STUN to finish, to avoid the probe if
         // it's unnecessary.
         if self.last_report.is_none() {
-            // Even if we're doing a non-incremental update, we may want to try our
-            // preferred relay for captive portal detection.
-            let preferred_relay = self
-                .last_report
-                .as_ref()
-                .and_then(|l| l.preferred_relay.clone());
-
             let dns_resolver = self.dns_resolver.clone();
-            let dm = self.relay_map.clone();
+            let relay_map = self.relay_map.clone();
             self.outstanding_tasks.captive_task = true;
             MaybeFuture {
                 inner: Some(Box::pin(async move {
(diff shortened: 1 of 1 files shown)


iroh-net-report/src/reportgen.rs
```diff
@@ -476,23 +476,16 @@ impl Actor {
         // delay by a bit to wait for UDP STUN to finish, to avoid the probe if
         // it's unnecessary.
         if self.last_report.is_none() {
-            // Even if we're doing a non-incremental update, we may want to try our
-            // preferred relay for captive portal detection.
-            let preferred_relay = self
-                .last_report
-                .as_ref()
-                .and_then(|l| l.preferred_relay.clone());
```
matheus23, line 484: Because we're in the `self.last_report.is_none()` branch, it follows that `self.last_report.and_then(...)` will also be `None`.

The rest of the code changes all follow from this.

flub:
> What this refactor shows is: Effectively the captive portal check will always do the check with _a random relay_ from the relay map. The reason for that is that it _only runs_, when there's no net report yet, which seems wrong, too.

This second part is definitely **not** wrong and by design.  We only need to check the captive portal if we are not getting any connectivity.  This is why it is scheduled with a small delay, in most setups it never runs and that is good.

If that means it should use a random relay?  Probably, maybe.  I mean, it doesn't seem the worst choice.

matheus23:
@flub no connectivity != no report yet, I'd argue.

The idea being, I think you want this to work:
- you run iroh in working WiFi
- you switch WiFi networks to a hotel WiFi with a captive portal (or the hotel wifi resets you into captive mode)
- you want to run a captive portal check with the previously preferred relay, not a random one

flub:
I don't see any harm with using a previously known preferred relay.  But it won't drastically change the result, using a close relay only makes the check work a little bit faster.  So I think it's a minor annoyance, not a worrying bug.

matheus23:
This PR might be nice for starting a discussion around net-report & captive portal checks.
That said, I think the discussion needs to be much bigger. We need to actually understand what we use/want to use net-report (and captive portal checks) for, and it's bigger than this PR. This PR is not worth thinking about in the mean time.

Issue #3123, opened 2025-01-13
Connection::remote_node_id should perhaps be infallible


flub:
The `Connection::remote_node_id` method currently returns a `Result`.  The kind of errors it throws are *currently* all essentially bugs in the code.  I am not sure how future proof this is however:

- Maybe #2937 which introduces raw public keys would change things?
- Maybe some other change will no longer mean the number of certificates will always be 1 and then having this check in old versions would at least mean they don't panic.

It **is** however an eyesore that this API is fallible right now.  Is there a way we can safely ensure this API is actually infallible?  Maybe it involves changing how the handshake data gets stored so we can directly store a NodeId?  This would also involve making the `peer_identity` no longer optional.

matheus23:
> Maybe some other change will no longer mean the number of certificates will always be 1 and then having this check in old versions would at least mean they don't panic.

I still strongly believe `Connection::remote_node_id` should be infallible. At the point at which you have a `Connection`, you should know your peer's `NodeId`. Otherwise how would it be an end-to-end encrypted channel with specifically *that* `NodeId`.

> Maybe raw public keys would change things?

No - we still have client authentication. And I don't think one should have access to a `Connection` without client auth. And if you have client auth, then the remote node ID is essentially what you "are authenticated to".

flub:
I think you're right.  But it'd be so much nicer if it didn't involve `.unwrap()`.  This should be possible, should it not?

matheus23:
I don't think so, because (1) the `quinn::Connection` is more general than our use case and includes use cases without client authentication and (2) the type of certificate is not a type parameter - so it's impossible to codify that fact at compile time.

matheus23:
Okay, I looked at this again/thought about this a little more.
I think there's a way of doing this "without unwrap", and that is: In `Connecting::poll`, after we get `Poll::Ready(Ok(connection))` where `connection: quinn::Connection`, we check that the `quinn::Connection` really is the type of connection we'd expect, by calling `get_remote_node_id`. If not, we error out. And if it succeeds, we can either store that `NodeId` directly on the `Connection` struct (no unwrap!), or we just repeat that call in `Connection::remote_node_id`, and use unwrap there.

rklaehn:
I tried just assuming that you always have a remote node id, here: https://github.com/n0-computer/iroh/pull/3455

But it turns out that in the case of 0-rtt you *don't* have the remote node id initially, you will only have it once a full handshake has completed. So it is not as easy.

[...]

PR #3147, opened 2025-01-23
feat(iroh)!: Make retrieving NodeID and ALPN infallible

flub:
## Description

Our crypto setup makese some guarantees here.  But we do need to be
careful not to unwrap something that can be triggered by a modified or
custom remote.

## Breaking Changes

TODO

### iroh

- Removes `Connecting::handshake_data`.
- Removes `Connection::handshake_data`.
- Removes `Connection::peer_identity`.
- Error type for `Connecting::alpn` changed to ConnectionError.
- `Connection::alpn` is now infallible.
- `Connection::remote_node_id` is now infallible.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -12,7 +12,6 @@
 //! [module docs]: crate
 
 use std::{
-    any::Any,
     collections::BTreeSet,
     future::{Future, IntoFuture},
     net::{IpAddr, SocketAddr, SocketAddrV4, SocketAddrV6},
@@ -1277,11 +1276,6 @@ impl Connecting {
         }
     }
 
-    /// Parameters negotiated during the handshake
-    pub async fn handshake_data(&mut self) -> Result<Box<dyn Any>, ConnectionError> {
-        self.inner.handshake_data().await
-    }
-
     /// The local IP address which was used when the peer established the connection.
     pub fn local_ip(&self) -> Option<IpAddr> {
         self.inner.local_ip()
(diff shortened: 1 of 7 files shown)


iroh/src/endpoint.rs
```diff
@@ -1295,15 +1289,12 @@ impl Connecting {
...
-            },
-            Err(_) => bail!("unknown handshake type"),
-        }
+    pub async fn alpn(&mut self) -> Result<Vec<u8>, ConnectionError> {
+        let data = self.inner.handshake_data().await?;
+        let data = data
+            .downcast::<quinn::crypto::rustls::HandshakeData>()
+            .expect("crypto setup for iroh");
+        Ok(data.protocol.expect("ALPN required by iroh crypto setup"))
     }
```
matheus23, line 1298:
```suggestion
    pub async fn alpn(&mut self) -> Result<Vec<u8>, ConnectionError> {
        let data = self.inner.handshake_data().await?;
        let data = data
            .downcast::<quinn::crypto::rustls::HandshakeData>()
            .expect("invariant checked by iroh's custom rustls verifier");
        Ok(data.protocol.expect("invariant upheld by iroh's custom rustls verifier"))
    }
```
Ideally insert the function names of whatever is *actually* verifying these invariants here.

iroh/src/endpoint.rs
```diff
@@ -1581,25 +1547,17 @@ impl Connection {
...
+        let data = self
+            .inner
+            .peer_identity()
+            .expect("required by iroh crypto setup");
+        let certs = data
+            .downcast::<Vec<rustls::pki_types::CertificateDer>>()
+            .expect("iroh crypto setup");
+        debug_assert_eq!(certs.len(), 1);
+        let cert = tls::certificate::parse(&certs[0]).expect("TODO: is this guaranteed by now?");
+        cert.peer_id()
```
matheus23, line 1560: I'd like better expect messages here, too. "required by iroh crypto setup" doesn't suffice IMO.
The message should mention the place that the `peer_identity` field is populated.
I tried figuring this out, but yeah - it's probably not that easy (flows through quinn, right?).

iroh/src/endpoint.rs
```diff
@@ -1539,38 +1530,13 @@ impl Connection {
...
-    /// [`downcast`]: Box::downcast
-    #[inline]
-    pub fn peer_identity(&self) -> Option<Box<dyn Any>> {
-        self.inner.peer_identity()
+    pub fn alpn(&self) -> Vec<u8> {
+        let data = self.inner.handshake_data().expect("iroh crypto setup");
+        let data = data
+            .downcast::<quinn::crypto::rustls::HandshakeData>()
+            .expect("iroh crypto setup");
+        data.protocol.expect("ALPN required by iroh crypto setup")
```
dignifiedquire, line 1539: > Guaranteed to be set if a nonempty list of protocols was specified for this connection.

Are we sure this will always be the case from the public api usage?

flub:
Newer attempts at this are tracked in #3123.  Needs work on 0-RTT

PR #3148, opened 2025-01-23
feat(iroh)!: Remove access to local and remote IP addresses

flub:
## Description

The Connecting and Connection structs exposed IP addresses, these are
probably the NodeIdMappedAddresses rather than real ones.  We address
based on NodeId, so let's remove this stuff.

### iroh

- `Connecting::local_ip` is removed.
- `Connecting::remote_address` is removed.
- `Connection::local_ip` is removed.
- `Connection::remote_address` is removed.

## Notes & open questions

I didn't think long about this.  Getting this PR out so someone else
can think carefully about this.  :)

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/examples/listen-unreliable.rs
+++ b/iroh/examples/listen-unreliable.rs
@@ -71,9 +71,8 @@ async fn main() -> anyhow::Result<()> {
         let conn = connecting.await?;
         let node_id = conn.remote_node_id()?;
         info!(
-            "new (unreliable) connection from {node_id} with ALPN {} (coming from {})",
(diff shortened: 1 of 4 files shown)


flub:
> Agreed we should remove the current version.
> 
> At some point it might be nice to add a version back that allows denylisting incoming connections by IP addr, if they misbehave often enough. For that, we'd need the non-mapped addresses though. And also would need more design (unsure about how to realize this with relays).

Yes, note that `Incoming` is the first struct you get on the accept side and this still has the local and remote IP methods.  Though they are probably also the wrong IPs and the thing does need more design.

PR #3155, opened 2025-01-28
fix: handle invalid input length when parsing a node id

### LINKED ISSUE #3153: Node::from_str() can panic
**Describe the bug**
- I accept a NodeID (as String input in a text field) in a dialog in my app, to then connect to another node via iroh.
- I create a NodeId from the string, checking for Ok/Err.
- I have tests that check this works, and that invalid NodeIds are caught and the user informed
- When updating from iroh-net to iroh, this test fails with a panic.
 
**Relevant Logs**

**Expected behavior**
- NodeId::from_str() returns an error when an invalid string is passed in

**Platform(s)**
Desktop:
 - OS: macos
 - Version 15.3

**Additional Context / Screenshots / GIFs**
This code
```
impl FromStr for PublicKey {
    type Err = KeyParsingError;

        let bytes = decode_base32_hex(s)?;

        Ok(Self::from_bytes(&bytes)?)
    }
}
```

panics when you pass it a non-bas32 string (e.g. "invalid_node_id") and the parse error is not caught.

```assert_eq!(Ok(output.len()), self.decode_len(input.len()));```

in lib.rs:1453
I think that needs to be changed to return an Error, not panic.

dignifiedquire:
This avoids a panic when calling `decode_mut` with an invalid input length.

Closes #3153

### DIFF
--- a/iroh-base/src/key.rs
+++ b/iroh-base/src/key.rs
@@ -330,7 +330,12 @@ fn decode_base32_hex(s: &str) -> Result<[u8; 32], KeyParsingError> {
         // hex
         data_encoding::HEXLOWER.decode_mut(s.as_bytes(), &mut bytes)
     } else {
-        data_encoding::BASE32_NOPAD.decode_mut(s.to_ascii_uppercase().as_bytes(), &mut bytes)
+        let input = s.to_ascii_uppercase();
+        let input = input.as_bytes();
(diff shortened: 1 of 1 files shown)


matheus23 (approved):
Very annoying that `decode_mut` panics... seems like a footgun.

andrewdavidmackenzie:
Thanks!

PR #3156, opened 2025-01-29
refactor: Use `n0-future` in favor of `futures-*` libraries and `tokio::{spawn,task,time}`

matheus23:
## Description

- replaces all imports of `futures_lite`, `futures_util`, `futures_sink` and `futures_buffered` with re-exports from `n0_future`.
- replaces all imports of `std::time` with `n0_future::time`
- replaces all occurrences of `tokio::{spawn,task,time}` with `n0_future::{spawn,task,time}`
- **does not affect tests**: we use a bunch of special things there, like `tokio::time::advance` that we don't have n0-future yet. It's also not strictly necessary for shipping e.g. browser stuff.

The goal of this PR is
- to get us closer to browser compatibility (I've verified that I've replaced enough of all task spawning, etc. such that mostly cfg-ing out and other unrelated changes remain browser support)
- to make the diffs of the browser-support PRs easier to follow

Essentially this is work extracted out of the browser PRs.

## Breaking Changes

I don't think there are any breaking changes from this.
It's possible that `JoinSet` or `JoinHandle` types leak from the interfaces somewhere, but they're identical to the old ones (just re-exports) outside of `wasm*-*-unknown` targets.

## Notes & open questions

I've tried setting up semgrep, but (1) IIUC, setting it up for CI requires that you have an account with them, only running locally is free and (2) I haven't found a way to ignore e.g. `std::time` imports when you're inside a `*::tests` module, which means we'd either need to convert all our tests to `n0-future` or ignore warnings and I think both of these options might suck.

## Change checklist

- [x] Self-review.


iroh/src/discovery/pkarr.rs
```diff
@@ -236,8 +227,8 @@ struct PublisherService {
 impl PublisherService {
     async fn run(mut self) {
         let mut failed_attempts = 0;
-        let republish = tokio::time::sleep(Duration::MAX);
-        tokio::pin!(republish);
+        let republish = time::sleep(Duration::MAX);
+        n0_future::pin!(republish);
```
dignifiedquire, line 231: nit: the pin macro exists in the std lib

dignifiedquire (approved):

iroh/src/discovery/pkarr.rs
matheus23: TIL

matheus23: Seems like it doesn't work though. If I use `std::pin::pin!(republish)` it doesn't compile for me :thinking:

PR #3157, opened 2025-01-31
feat(iroh)!: allow for limiting incoming connections on the router

dignifiedquire:
## Description

Changes the `ProtocolHandler` trait to allow for intercepting `connecting` and `connection` states explicitly

## Breaking Changes

- changed: `iroh::protocol::ProtocolHandler::accept` now takes `Connection` instead of `Connecting`

## Notes & open questions

The new limiter could also just be part of the test code/an example, instead of giving users an API, unclear to me 

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,9 +94,8 @@ let router = Router::builder(endpoint)
 struct Echo;
 
 impl ProtocolHandler for Echo {
-    fn accept(self: Arc<Self>, connecting: Connecting) -> BoxedFuture<Result<()>> {
+    fn accept(&self, connection: Connection) -> BoxedFuture<Result<()>> {
         Box::pin(async move {
-            let connection = connecting.await?;
             let (mut send, mut recv) = connection.accept_bi().await?;
 
             // Echo any bytes received back directly.
--- a/iroh/src/protocol.rs
+++ b/iroh/src/protocol.rs
@@ -41,6 +41,7 @@
 use std::{collections::BTreeMap, sync::Arc};
 
 use anyhow::Result;
+use iroh_base::NodeId;
 use n0_future::{
     boxed::BoxFuture,
     join_all,
@@ -50,7 +51,10 @@ use tokio::sync::Mutex;
 use tokio_util::sync::CancellationToken;
 use tracing::{error, info_span, trace, warn, Instrument};
 
-use crate::{endpoint::Connecting, Endpoint};
+use crate::{
+    endpoint::{Connecting, Connection},
+    Endpoint,
+};
 
 /// The built router.
 ///
@@ -109,10 +113,18 @@ pub struct RouterBuilder {
 /// The protocol handler must then be registered on the node for an ALPN protocol with
 /// [`crate::protocol::RouterBuilder::accept`].
 pub trait ProtocolHandler: Send + Sync + std::fmt::Debug + 'static {
+    /// Optional interception point to handle the `Connecting` state.
+    fn on_connecting(&self, conn: Connecting) -> BoxFuture<Result<Connection>> {
+        Box::pin(async move {
+            let conn = conn.await?;
+            Ok(conn)
+        })
+    }
+
     /// Handle an incoming connection.
     ///
     /// This runs on a freshly spawned tokio task so this can be long-running.
-    fn accept(&self, conn: Connecting) -> BoxFuture<Result<()>>;
+    fn accept(&self, conn: Connection) -> BoxFuture<Result<()>>;
 
     /// Called when the node shuts down.
     fn shutdown(&self) -> BoxFuture<()> {
@@ -121,7 +133,10 @@ pub trait ProtocolHandler: Send + Sync + std::fmt::Debug + 'static {
 }
 
 impl<T: ProtocolHandler> ProtocolHandler for Arc<T> {
-    fn accept(&self, conn: Connecting) -> BoxFuture<Result<()>> {
+    fn on_connecting(&self, conn: Connecting) -> BoxFuture<Result<Connection>> {
+        self.as_ref().on_connecting(conn)
+    }
+    fn accept(&self, conn: Connection) -> BoxFuture<Result<()>> {
         self.as_ref().accept(conn)
     }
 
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -94,9 +94,8 @@ let router = Router::builder(endpoint)
 struct Echo;
 
 impl ProtocolHandler for Echo {
-    fn accept(self: Arc<Self>, connecting: Connecting) -> BoxedFuture<Result<()>> {
+    fn accept(&self, connection: Connection) -> BoxedFuture<Result<()>> {
```
rklaehn, line 97: Why was this Arc<Self> before? Was the readme not up to date?

iroh/src/protocol.rs
```diff
@@ -109,10 +113,18 @@ pub struct RouterBuilder {
 /// The protocol handler must then be registered on the node for an ALPN protocol with
 /// [`crate::protocol::RouterBuilder::accept`].
 pub trait ProtocolHandler: Send + Sync + std::fmt::Debug + 'static {
+    /// Optional interception point to handle the `Connecting` state.
+    fn on_connecting(&self, conn: Connecting) -> BoxFuture<Result<Connection>> {
```
rklaehn, line 117: This is not needed anywhere. So why do we need it? For future use?

rklaehn (commented):
I'm on the way to the airport, so just a bunch of drive by comments.

README.md
dignifiedquire: > Why was this Arc before? Was the readme not up to date?

indeed

iroh/src/protocol.rs
dignifiedquire: currently this is the thing we expose, which is why I added this hook for now. don't know if it is really needed in practice

README.md
matheus23: Hm. How do we make sure the readme's code is doc-tested?

flub: `include_str!` it into a doc comment on a private dummy item and then `cargo doc` might pick it up?  (not tested if a private item is enough for this).

iroh/src/protocol.rs
rklaehn: One thing I dislike about this is that if you want to allow protocol handlers to be composed, like done in this PR, it means that you add boxing at every composition layer.

The way I usually deal with this is to have a `trait ProtocolHandler` which is not *dyn*able but has concrete return types for the futures, and then have a second `trait BoxedProtocolHandler` which has the return type fixed to BoxFuture. That way you box only once when you convert into a dyn. But that is a lot of boilerplate to write and difficult to explain to beginners.

So if we don't yet have a concrete use case for `on_connecting` I would just leave it out for now. But 🤷 ...

dignifiedquire: > So if we don't yet have a concrete use case for `on_connecting` I would just leave it out for now. But 🤷 ...

we have now, with https://github.com/n0-computer/iroh/pull/3163

README.md
ramfox: Can confirm that flub's suggestion works. I had a few of these trigger errors in our protocol repos that contain the readme as part of the documentation, when APIs changed and no longer compiled.

dignifiedquire:
@matheus23 can check if this change is good enough to still be able to use 0-rtt from `ProtocolHandler` impl please?

matheus23:
Yeah - this change makes using 0-RTT/0.5-RTT on the accepting side possible *at all*, since otherwise you wouldn't be able to call `Connecting::into_0rtt`.

And all you need to do on the accepting side is call `Connecting::into_0rtt`, then if you get `Ok`, you take the connection, otherwise if you get an `Err`, you await.
That's it.
(You should also make sure you know the security implications of what you're doing - but from an "enabling 0-RTT" perspective that is indeed it.)

iroh/src/protocol.rs
```diff
@@ -109,10 +112,18 @@ pub struct RouterBuilder {
 /// The protocol handler must then be registered on the node for an ALPN protocol with
 /// [`crate::protocol::RouterBuilder::accept`].
 pub trait ProtocolHandler: Send + Sync + std::fmt::Debug + 'static {
+    /// Optional interception point to handle the `Connecting` state.
```
flub, line 115: I'm always a fan of being extremely verbose on the docs.  I would add a paragraph suggesting when you might to use this:
```
This enables accepting 0-RTT data from clients, among other things.
```

iroh/src/protocol.rs
```diff
@@ -350,8 +369,64 @@ async fn handle_connection(incoming: crate::endpoint::Incoming, protocols: Arc<P
...
+    }
+}
+
+/// Wraps an existing protocol, limiting its access,
+/// based on the provided function.
+#[derive(derive_more::Debug, Clone)]
+pub struct AccessLimit<P: ProtocolHandler + Clone> {
+    proto: P,
+    #[debug("limiter")]
+    limiter: Arc<dyn Fn(NodeId) -> bool + Send + Sync + 'static>,
```
flub, line 390: Not sure if this was discussed before: will we get folks upset that they can't do async stuff here?  I think this itself runs in an async context.

flub: Of course we/they can just make an `AsyncAcessLimit` which is probably nicer than forcing everyone to use async.

iroh/src/protocol.rs
```diff
@@ -350,8 +369,64 @@ async fn handle_connection(incoming: crate::endpoint::Incoming, protocols: Arc<P
...
+        self.proto.on_connecting(conn)
+    }
+
+    fn accept(&self, conn: Connection) -> BoxFuture<Result<()>> {
+        let this = self.clone();
+        Box::pin(async move {
+            let remote = conn.remote_node_id()?;
+            let is_allowed = (this.limiter)(remote);
+            if !is_allowed {
+                conn.close(0u32.into(), b"not allowed");
```
flub, line 420: This needs to be documented.  Probably on the struct itself.  Something like:

```
Any refused connection will be closed with an error code of `0` and reason `not allowed`.
```

(I can also imagine folks wanting this configurable.  But I think that's possible in a backwards-compatible way so should be fine to add later.  I can also imagine a world where the function returns those values.)

flub (approved):
I can also imagine a need for `on_incoming` or something like that in the future to be able to reject etc even further.  But the nice thing about this design is that you can add this later without breaking backwards compatibility.

I have a minor bikeshedding concern that `AccessLimit` sounds too much like rate-limiting.  But I don't know if there is a commonly used term that's better.

iroh/src/protocol.rs
dignifiedquire: > Of course we/they can just make an AsyncAcessLimit which is probably nicer than forcing everyone to use async.

yeah, this is more of an example for me

iroh/src/protocol.rs
flub: Sure, why not.  Don't think there'd be an objection to this.

Issue #2741, opened 2024-09-22
switch to using custom error types, rather than generic anyhow::Error


matthiasbeyer:
[anyhow](https://docs.rs/anyhow/latest/anyhow/) is for application error handling, please replace it with dedicated error types in your library crates, so users of these crates can do sane error handling.

flub:
Could you be more specific to which error return values you would like to match on for different failure cases?

In general you are right, and as iroh moves towards 1.0 we probably also should migrate to more specific error types.  On the other hand we really need this to be guided by application needs rather than enumerate all the current failure types, as that would lead to breaking the API on every release.  If we were to use a custom error type right now things would probably be a single opaque error type, which is just like anyhow, but worse.

So far in our own uses we have not needed to distinguish error types, but this doesn't mean there isn't a need!  This is why to move this forward we really need specific APIs with specific usecases called out for needing a custom error type.  We'd really appreciate your input on this.

b5:
Hey @matthiasbeyer, great to see you back after [our last encounter](https://github.com/n0-computer/iroh/pull/495), 

Last we spoke, the team came to the working conclusion that we'd stick with anyhow pre-1.0, and transition to `thiserror` as part of a 1.0 release ([reference](https://github.com/n0-computer/iroh/pull/495#issuecomment-1318677449)), thankfully, we've made a bunch of progress on iroh in the last year, and are now starting to think seriously about a 1.0 roadmap. From our latest [blog post](https://iroh.computer/blog/smaller-is-better):

> We're aiming to publish a 1.0 roadmap in 2025.

Which means (unless others, especially @dignifiedquire disagree): we need to figure out when to transition over to `thiserror`, and at least have a plan in place in 2025, so maybe we use this issue & your baseline set of errors as a starting point for planning?

iacore:
If the rewrite can happen in a day, would it happen earlier? What's blocking it from happening now?

matheus23:
> Are there any other ways where one can craft a unique `anyhow::Error`?

Yes. There is this implementation: `impl<E> From<E> for Error where E: StdError + Send + Sync + 'static` 

So everywhere `?` in any function that returns `anyhow::Result`, there might be a conversion from something that impls `StdError` (that includes `anyhow::Error` itself) we may need to add either
- an additional `impl From<...> for IrohError`
- a new error enum variant with a `#[from]` annotation or
- a call to `.map_err` before the `?`.

[...]

PR #3161, opened 2025-02-03
feat(iroh-relay)!: use concret errors for the client part of iroh-relay

dignifiedquire:
## Description

Another step towards https://github.com/n0-computer/iroh/issues/2741

## Breaking Changes

- `RelayMap::from_nodes` does not return an error anymore
- errors in client apis of `iroh_relay` are not `anyhow::Error` but rather concrete error types

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-relay/src/client.rs
+++ b/iroh-relay/src/client.rs
@@ -38,18 +38,45 @@ pub(crate) mod streams;
 mod util;
 
 /// Client related errors
+#[allow(missing_docs)]
 #[derive(Debug, thiserror::Error)]
+#[non_exhaustive]
 pub enum Error {
     #[error("Invliad target port")]
     InvalidTargetPort,
     #[error("Invalid URL {0}")]
     InvalidUrl(Url),
+    #[error("Invalid Proxy URL {0}")]
+    InvalidProxyUrl(Url),
     #[error("Invalid URL for websocket {0}")]
     InvalidWebsocketUrl(Url),
     #[error(transparent)]
     Websocket(#[from] tokio_tungstenite_wasm::Error),
     #[error(transparent)]
     Dns(#[from] DnsError),
+    #[error(transparent)]
+    Hyper(#[from] hyper::Error),
+    #[error(transparent)]
+    InvalidDnsName(#[from] rustls::pki_types::InvalidDnsNameError),
+    #[error("Unexpected status during upgrade: {0}")]
+    UnexpectedUpgradeStatus(hyper::StatusCode),
+    #[error("Failed proxy connection: {0}")]
+    FailedProxyConnection(hyper::StatusCode),
+    #[error(transparent)]
+    ProtoRelay(#[from] crate::protos::relay::Error),
+    #[error(transparent)]
+    Io(#[from] std::io::Error),
+    #[error("Timeout")]
+    Timeout(#[from] tokio::time::error::Elapsed),
+    #[error(transparent)]
+    Http(#[from] hyper::http::Error),
+    #[error("Unexpected frame received {0}")]
+    UnexpectedFrame(crate::protos::relay::FrameType),
+    #[error(transparent)]
+    Utf8(#[from] std::str::Utf8Error),
+    #[cfg(wasm_browser)]
+    #[error("The relay protocol is not available in browsers")]
+    RelayProtoNotAvailable,
 }
 
 /// Build a Client.
@@ -166,9 +193,7 @@ impl ClientBuilder {
                 (conn, Some(local_addr))
             }
             #[cfg(wasm_browser)]
-            Protocol::Relay => {
-                bail!("Can only connect to relay using websockets in browsers.");
-            }
+            Protocol::Relay => return Err(Error::RelayProtoNotAvailable),
         };
 
         event!(
--- a/iroh-relay/src/client.rs
+++ b/iroh-relay/src/client.rs
@@ -9,8 +9,8 @@ use std::{
     task::{self, Poll},
 };
 
-use anyhow::{anyhow, bail, Result};
 use conn::Conn;
+use connect_relay::DnsError;
 #[cfg(not(wasm_browser))]
 use hickory_resolver::TokioResolver as DnsResolver;
 use iroh_base::{RelayUrl, SecretKey};
@@ -36,6 +36,21 @@ mod connect_relay;
 pub(crate) mod streams;
 #[cfg(not(wasm_browser))]
 mod util;
+
+/// Client related errors
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("Invliad target port")]
+    InvalidTargetPort,
+    #[error("Invalid URL {0}")]
+    InvalidUrl(Url),
+    #[error("Invalid URL for websocket {0}")]
+    InvalidWebsocketUrl(Url),
+    #[error(transparent)]
+    Websocket(#[from] tokio_tungstenite_wasm::Error),
+    #[error(transparent)]
+    Dns(#[from] DnsError),
+}
 
 /// Build a Client.
 #[derive(derive_more::Debug, Clone)]
@@ -138,7 +153,7 @@ impl ClientBuilder {
     }
 
     /// Establishes a new connection to the relay server.
-    pub async fn connect(&self) -> Result<Client> {
+    pub async fn connect(&self) -> Result<Client, Error> {
         let (conn, local_addr) = match self.protocol {
             Protocol::Websocket => {
                 let conn = self.connect_ws().await?;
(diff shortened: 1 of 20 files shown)


iroh-relay/src/client.rs
```diff
@@ -37,6 +37,48 @@ pub(crate) mod streams;
 #[cfg(not(wasm_browser))]
 mod util;
 
+/// Client related errors
+#[allow(missing_docs)]
```
dignifiedquire, line 41: I don't think there is much value in documenting these variants currently, which is why I am skipping them

flub (changes requested):
I think fundamentally I'm not happy with these giant enum errors.  I think at every point an error can be returned, each error variant of the type must be possible.  The approach here just makes every source error a variant without any concern on whether this really helps anything.

iroh-relay/src/client.rs
```diff
@@ -37,6 +38,49 @@ pub(crate) mod streams;
...
+    #[error(transparent)]
+    Http(#[from] hyper::http::Error),
+    #[error("Unexpected frame received {0}")]
+    UnexpectedFrame(crate::protos::relay::FrameType),
+    #[error(transparent)]
+    Utf8(#[from] std::str::Utf8Error),
+    #[cfg(wasm_browser)]
+    #[error("The relay protocol is not available in browsers")]
+    RelayProtoNotAvailable,
+}
```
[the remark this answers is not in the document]
dignifiedquire: that's fair, I guess I have more work to do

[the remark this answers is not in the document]
dignifiedquire: @flub please take another look, while certainly not perfect, I think this is a good improvement over the status quo

iroh-relay/src/client.rs
```diff
@@ -37,6 +36,64 @@ pub(crate) mod streams;
...
+#[allow(missing_docs)]
+#[derive(Debug, thiserror::Error)]
+#[non_exhaustive]
+pub enum ConnectError {
+    #[error("Invalid URL for websocket {0}")]
+    InvalidWebsocketUrl(Url),
+    #[error("Invalid relay URL {0}")]
+    InvalidRelayUrl(Url),
+    #[error(transparent)]
+    Websocket(#[from] tokio_tungstenite_wasm::Error),
```
flub, line 49: I'm not sure this error is self-describing.  Maybe an `#[error("Websocket transport error: {0}")]` would be better?  (If that is the correct interpretation)

Which immediately makes me wonder "but what if I'm not using websocket"?  Why is there no generic IO or Transport error this falls under?

iroh-relay/src/client.rs
```diff
@@ -37,6 +36,64 @@ pub(crate) mod streams;
...
+    Dial(#[from] DialError),
+    #[error("Unexpected status during upgrade: {0}")]
+    UnexpectedUpgradeStatus(hyper::StatusCode),
+    #[error("Failed to upgrade response")]
+    Upgrade(#[source] hyper::Error),
+    #[error("Invalid TLS servername")]
+    InvalidTlsServername,
+    #[error("No local address available")]
+    NoLocalAddr,
+    #[error("tls connection failed")]
```
flub, line 62:
```suggestion
    #[error("TLS connection failed")]
```

iroh-relay/src/client.rs
```diff
@@ -37,6 +36,64 @@ pub(crate) mod streams;
...
+    #[cfg(wasm_browser)]
+    #[error("The relay protocol is not available in browsers")]
+    RelayProtoNotAvailable,
+}
+
+/// Dialing errors
+#[allow(missing_docs)]
+#[derive(Debug, thiserror::Error)]
+#[non_exhaustive]
+pub enum DialError {
```
flub, line 73: Reading just the error types I am rather confused what the difference between `ConnectError` and `DialError` is.  Sounds like the same to me.

It seems `DialError` is a variant of `ConnectError` so maybe it's more of a `ConnectionError`?  But then also `ConnectError` is returned from `Client::connect` so it probably was meant as `ConnectError`  :face_with_spiral_eyes:

iroh-relay/src/client/conn.rs
```diff
@@ -17,22 +16,50 @@ use tokio_tungstenite_wasm::WebSocketStream;
...
+    #[error(transparent)]
+    Io(#[from] io::Error),
+    #[error(transparent)]
+    Protocol(#[from] crate::protos::relay::Error),
+    #[error(transparent)]
+    Websocket(#[from] tokio_tungstenite_wasm::Error),
+    #[error("invalid protocol message encoding")]
+    InvalidProtocolMessageEncoding,
+    #[error("Unexpected frame received {0}")]
+    UnexpectedFrame(crate::protos::relay::FrameType),
```
flub, line 60: `Websocket` seems like part of the `Io` variant?
And `InvalidProtocolMessageEncoding` and `UnexpectedFrame` are part of the `Protocol` variant?

I think the `io/protocol` distinction is pretty decent on a high-level and that was what I was going for in the `ConnSendError`.

Also it might be possible to have `ConnSendError` and `RecvError` be the same type?

flub (commented):
I am struggling to figure out how to be constructive on this.  I still don't particularly like this error style.  I think it does not think enough about the human consumer.  The error variants inside single enums are often of widely different "types".

OTOH I don't feel like particularly holding this up.  Though I am a little worried if it becomes the model we use everywhere.  If I were to properly handle this I'd create a counter-proposal as PR... but that takes time.  Which I guess is part of the question, my approach would be a much more time-intensive I expect as it really asks for a deep understanding of each API.

Anyway, eh, feel free to merge?

dignifiedquire:
Unfortunately I found a massive drawback to using `thiserror` type errors. You loose backtrace information for most errors that are wrapped, including `std` errors. This results in quite an awful debugging experience, especially in a test runner setup. 

The reasons for this is that tldr https://github.com/rust-lang/rust/issues/99301 is not stable yet and error libraries can't propagate backtrace information properly. This works with anyhow because it uses a singular error type, which can poperly propagate backtraces. 

Until I can find a solution for this, using error enums is for now unacceptable.

If you ask, Yes I probably looked at your favourite error handling library, they all fail to do this on STABLE rust, including

- `eyre` and it's various reporters
- `thiserror`
- `snafu`
- `anyhow` combined with `thiserror` or `snafu`
- `testresult`

matheus23:
We should rename this PR now.

iroh-dns-server/src/server.rs
```diff
@@ -29,7 +28,7 @@ pub async fn run_with_config_until_ctrl_c(config: Config) -> Result<()> {
         store = store.with_mainline_fallback(bootstrap);
     };
     let server = Server::spawn(config, store, metrics).await?;
-    tokio::signal::ctrl_c().await?;
+    tokio::signal::ctrl_c().await.context("ctrl-c")?;
```
matheus23, line 31: These `.context`s aren't super fun. Should we change them to `.e()`?

matheus23 (commented):
Posting some of my review comments so far (quite a big PR).

iroh-dns-server/src/dns.rs
```diff
@@ -232,9 +242,10 @@ fn create_static_authority(
         RecordType::SOA,
         config.default_soa.split_ascii_whitespace(),
         None,
-    )?
+    )
+    .context("rdata")?
     .into_soa()
-    .map_err(|_| anyhow!("Couldn't parse SOA: {}", config.default_soa))?;
+    .map_err(|_| format_err!("Couldn't parse SOA: {}", config.default_soa))?;
```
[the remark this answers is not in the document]
dignifiedquire: whatever is equal to anyhow::bail and format_err is equal to anyhow::anyhow i
I think

iroh-dns-server/src/server.rs
dignifiedquire: sure

iroh-dns-server/examples/resolve.rs
```diff
@@ -1,10 +1,10 @@
-use anyhow::Context;
 use clap::{Parser, ValueEnum};
 use iroh::{
     discovery::dns::{N0_DNS_NODE_ORIGIN_PROD, N0_DNS_NODE_ORIGIN_STAGING},
     dns::DnsResolver,
     NodeId,
 };
+use n0_snafu::{TestResult as Result, TestResultExt};
```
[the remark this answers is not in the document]
matheus23: How about just `Result`? :D
Maybe `ReportResult`?

iroh/src/protocol.rs
```diff
@@ -100,6 +99,46 @@ pub struct RouterBuilder {
...
+pub enum ProtocolError {
+    #[snafu(transparent)]
+    Connect {
+        source: crate::endpoint::ConnectionError,
+    },
+    #[snafu(transparent)]
+    User {
+        source: Box<dyn std::error::Error + Send + Sync + 'static>,
+    },
+}
```
ramfox, line 113: @dignifiedquire questions:

The `User` variant is meant for when users need to write custom protocol errors in their protocol implementations, right? If so, thoughts on `Custom` rather than `User`.

We don't have `backtrace` or `span_trace` anywhere on here, even for the `Connect` variant. That makes sense to me for the `User`/`Custom` variant, but is there are reason it's missing from `Connect`?

iroh-dns-server/examples/resolve.rs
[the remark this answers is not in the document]
dignifiedquire: fixed

ramfox (approved):

PR #3162, opened 2025-02-03
refactor!: remove iroh-test crate

dignifiedquire:
## Description

These testutils we are using, actually can be replaced through either simpler code or other crates. See the individual commits for the alternatives

## Breaking Changes

- `iroh-test` crate is gone

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/magicsock.rs
+++ b/iroh/src/magicsock.rs
@@ -3214,7 +3214,7 @@ mod tests {
     ///
     /// When the returned drop guard is dropped, the tasks doing this updating are stopped.
     #[instrument(skip_all)]
-    async fn mesh_stacks(stacks: Vec<MagicStack>) -> Result<Vec<AbortOnDropHandle<()>>> {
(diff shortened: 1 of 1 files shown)


iroh/src/magicsock.rs
```diff
@@ -3238,22 +3238,19 @@ mod tests {
 
         // For each node, start a task which monitors its local endpoints and registers them
         // with the other nodes as local endpoints become known.
-        let mut tasks = JoinSet::new();
+        let mut tasks = Vec::new();
```
flub, line 3241: Wait?  What?  A `JoinSet` aborts all tasks on drop already.  Why are we jumping through all these hoops?

flub (approved):

iroh/src/magicsock.rs
dignifiedquire: bonus fun 🤣 I can go back to joinset I guess

PR #3163, opened 2025-02-03
feat(iroh): Enable applications to establish 0-RTT connections

### LINKED ISSUE #3146: Allow using 0-RTT when connecting to another node
The `Endpoint::connect` method returns `Connection`, skipping the `Connecting` struct.  However without `Connecting` you can not use 0-RTT data.  We should allow 0-RTT data from the client.

On the `Endpoint::accept` path, we do use `Connecting` so 0.5-RTT can be used already.

matheus23:
## Description

Implements necessary APIs to make use of 0-RTT QUIC connections.

0-RTT allows you to skip a round-trip in case you have connected to a known endpoint ahead of time, and stored the given TLS session ticket.
With this PR, we by default will cache up to 8 session tickets per endpoint you connect to, and remember up to 32 endpoints maximum.
This cache only lives in-memory. We might add customization to the `EndpointBuilder` in the future to allow for customizing this cache (allowing you to persist it), but that obviously has security implications, so will need careful consideration.

This PR enables using 0-RTT via the `Endpoint::connect_with_opts` function, which - unlike `Endpoint::connect` - returns a `Connecting`, a state prior to a full `Connection`. By calling `Connecting::into_0rtt` you can attempt to turn this connection into a full 0-RTT connection. However, security caveats apply. See that function's documentation for details.

Migration guide:
```rs
let connection = endpoint.connect_with(node_addr, alpn, transport_config).await?;
```
to
```rs
let connection = endpoint.connect_with_opts(
    node_addr,
    alpn,
    ConnectOptions::new().with_transport_config(transport_config),
)
.await?
.await?; // second await for Connecting -> Connection
```

Closes #3146 

## Breaking Changes

- `iroh::Endpoint::connect_with` was removed, and `iroh::Endpoint::connect_with_opts` was added instead, but returning an `iroh::endpoint::Connecting` instead of an `iroh::endpoint::Connection`, allowing use of QUIC's 0-RTT feature.
- `iroh::endpoint::Connection::into_0rtt` now returns `iroh::endpoint::ZeroRttAccepted` (among other things), instead of `iroh_quinn::ZeroRttAccepted`. This wrapper is equivalent in functionality, but makes sure we're not depending on API-breaking changes in quinn and can keep a discovery task alive for as long as needed, until a connection is established.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


dignifiedquire:
This makes the API worse for basically everyone using it today, so I don't think this is a good way to go about it. I would rather see a new method, like `connect_in_steps -> Connecting` (bikeshed the name todo), than breaking this api

matheus23:
That's fair. Will change it.

iroh/src/endpoint.rs
```diff
@@ -682,21 +694,10 @@ impl Endpoint {
...
-        let connection = connect
-            .await
-            .context("failed connecting to remote endpoint")?;
-
-        let rtt_msg = RttMessage::NewConnection {
-            connection: connection.weak_handle(),
-            conn_type_changes: self.conn_type(node_id)?.stream(),
-            node_id,
-        };
-        if let Err(err) = self.rtt_actor.msg_tx.send(rtt_msg).await {
```
dignifiedquire, line 694: why is this not needed anymore?

iroh/src/endpoint.rs
```diff
@@ -587,24 +587,32 @@ impl Endpoint {
...
+    ///    **Note:** Please be aware that changing transport config settings may have adverse effects on
+    ///    establishing and maintaining direct connections.  Carefully test settings you use and
+    ///    consider this currently as still rather experimental.
+    pub async fn connect_with_opts(
         &self,
         node_addr: impl Into<NodeAddr>,
         alpn: &[u8],
-        transport_config: Arc<TransportConfig>,
-    ) -> Result<Connection> {
+        options: ConnectOptions,
```
flub, line 614: I generally like this approach and will attempt to push this one notch further: There's no need to **require** the ALPN.  If you don't use the router it is perfectly possible to accept a connection without an ALPN.  Why not put the ALPN into the `ConnectOptions` as an Option as well?

iroh/src/endpoint.rs
matheus23: Because by wrapping `quinn::Connecting` with `iroh::endpoint::Connecting`, we run `try_send_rtt_msg` in `impl Future for Connecting`.

iroh/src/endpoint.rs
matheus23: Hm. Interesting idea. I thought we'd aligned on "iroh connections will always have an ALPN" somewhere in discord though. I think it might be a good nudge for users to always implement signaling on *some* level.

matheus23:
Current state:
- I've painted the bikeshed this way now: there's `Endpoint::connect`, which is not modified, and there's `Endpoint::connect_with_opts` which can now *optionally* get a custom `TransportConfig`, and it returns `Connecting`, allowing 0-RTT stuff.
- There's two already existing tests that fail for reasons I don't understand yet (`discovery::tests::endpoint_discovery_combined_with_empty_and_wrong` and `discovery::tests::endpoint_discovery_with_wrong_existing_addr`)
- I couldn't get 0-RTT connections to work in a test for iroh. Not sure why.

iroh/src/endpoint.rs
flub: Ok, totally fair.  Let's not worry about ALPNs now and leave it as this.

matheus23:
Leaving breadcrumbs for myself:
- quinn-proto's `Connection::into_0rtt` doesn't do anything, because `self.early_crypto()` is `None`.
- That's due to quinn's `TlsSession::early_crypto` returning `None`
- Which is due to rustls's `Quic::early_secret` being `None`.
Would need to look into `emit_client_hello_for_retry` in rustls and why its local variable `early_key_schedule` is (likely) `None`. Probably no `tls13_session`? And then need to look into why `ClientHelloInput` doesn't have `resumption` or sth.

dignifiedquire:
@matheus23 this PR might help: https://github.com/quinn-rs/quinn/pull/1826

matheus23:
@dignifiedquire thanks, this helped me a ton. This would've taken me at least 10h longer otherwise.
I got a basic test working, but the current version is just preliminary to get the test to work at all.
E.g. I suspect that at the moment you can only use 0-RTT on your *second consecutive* connection, but not if you do any connection attempts in between.

The reason for that is that we're using the same "server name" for all connections, namely `localhost`. This means we don't know which session tickets belong to which endpoint we're trying to connect to - we throw them all in the same bucket.

I don't know all of the things changing that `localhost` would affect though :/

matheus23:
> E.g. I suspect that at the moment you can only use 0-RTT on your _second consecutive_ connection, but not if you do any connection attempts in between.

My suspicion was correct. I've changed the `"localhost"` `server_name` to be `{node_id_base32}.iroh.invalid` instead.
This means our store for TLS session tickets will put the tickets into the correct buckets for different remote endpoints.

I've also fixed a regression that changing the `connect_with` function caused: The `DiscoveryTask` used to be alive until we've established a `Connection` (and then be dropped/aborted), but by only returning `Connecting`, it never ran as long.

So I've opted to keep the task alive for as long as we haven't heard back from the remote (thus established a "real" connection). This also means wrapping `ZeroRttAccepted` to make sure we also keep the discovery task alive for long enough in case the user used `Connecting::into_0rtt()`.

iroh/src/discovery.rs
```diff
@@ -398,15 +398,15 @@ impl DiscoveryTask {
...
         loop {
             match stream.next().await {
                 Some(Ok(r)) => {
                     if r.node_addr.is_empty() {
-                        debug!(provenance = %r.provenance, "discovery: empty address found");
+                        debug!(provenance = %r.provenance, "empty address found");
                         continue;
                     }
-                    debug!(provenance = %r.provenance, addr = ?r.node_addr, "discovery: new address found");
+                    debug!(provenance = %r.provenance, addr = ?r.node_addr, "new address found");
```
matheus23, line 409: These prefixes are redundant, this function is already called with a "discovery" instrument span.

iroh/src/discovery.rs
```diff
@@ -659,10 +658,19 @@ mod tests {
...
+        let mut config = TransportConfig::default();
+        config.keep_alive_interval(Some(Duration::from_secs(1)));
+        config.max_idle_timeout(Some(IdleTimeout::try_from(Duration::from_secs(3))?));
+        let opts = ConnectOptions::new().with_transport_config(Arc::new(config));
+
+        let res = ep2
+            .connect_with_opts(ep1.node_id(), TEST_ALPN, opts)
+            .await? // -> Connecting works
+            .await; // -> Connection is expected to fail
         assert!(res.is_err());
```
matheus23, line 674: This test always relied on the connection timing out - so it would always run *at least* 30s.
We used to not have a way of customizing the transport config to reduce the idle timeout, but now we have. So now this test succeeds/fails after only about 3 seconds.

iroh/src/endpoint.rs
```diff
@@ -634,68 +668,42 @@ impl Endpoint {
...
-    #[instrument(
-        name = "connect",
-        skip_all,
-        fields(
-            me = %self.node_id().fmt_short(),
-            remote_node = node_id.fmt_short(),
-            alpn = %String::from_utf8_lossy(alpn),
-        )
-    )]
-    async fn connect_quinn(
```
matheus23, line 657: `connect_quinn` used to only be called by `connect_with`.
Instead of separating out the function, I've moved it in with `connect_with`, which was renamed to `connect_with_opts`.

I've tried to preserve the tracing instrumentation though, thus all of the `span.record` calls.

matheus23:
Required CI will be red until #3178 is merged.

iroh/src/endpoint.rs
```diff
@@ -634,68 +668,42 @@ impl Endpoint {
...
-        &self,
-        node_id: NodeId,
-        alpn: &[u8],
-        addr: NodeIdMappedAddr,
-        transport_config: Arc<TransportConfig>,
-    ) -> Result<Connection> {
         debug!("Attempting connection...");
```
flub, line 680: Let's remove this line, there's debug message just above.

iroh/src/endpoint.rs
```diff
@@ -634,68 +668,42 @@ impl Endpoint {
                 )
             })?;
 
-        debug!(
-            "connecting to {}: (via {} - {:?})",
-            node_id, addr, direct_addresses
-        );
+        debug!("connecting via {addr} - {direct_addresses:?}",);
```
flub, line 671: This message has always bothered me.  It is not clear to the user that `addr` in an internal QuicMappedAddr.  It also is not clear that a relay URL is used as well.  I think this is just a very very early debugging message from a different era.

Maybe something like `debug!(mapped_addr=?addr, relay_url=?node_addr.relay_url, ?direct_addresses, "Attempting connection...");`

If you feel like arguing to remove it entirely I probably won't fight you.

iroh/src/endpoint.rs
```diff
@@ -634,68 +668,42 @@ impl Endpoint {
...
+                Some(self.session_store.clone()),
                 self.static_config.keylog,
             )?;
             let mut client_config = quinn::ClientConfig::new(Arc::new(quic_client_config));
             client_config.transport_config(transport_config);
             client_config
         };
 
-        // TODO: We'd eventually want to replace "localhost" with something that makes more sense.
+        let server_name = &format!("{}.iroh.invalid", BASE32_DNSSEC.encode(node_id.as_bytes()));
```
flub, line 695: Could we add a comment that this is only used as a key in the TLS session store and has no other function?  (If that is indeed correct?  At least that is currently the only *reason* for this.)  Might help with future decisions that don't like this naming pattern anymore.

iroh/src/endpoint.rs
```diff
@@ -543,10 +545,17 @@ impl Endpoint {
...
         let ep = Self {
             msock: msock.clone(),
             rtt_actor: Arc::new(rtt_actor::RttHandle::new()),
             static_config: Arc::new(static_config),
+            // 8 tickets per remote endpoint, 32 different endpoints
+            // ~200 bytes per session + certificates (which are ~387 bytes)
+            // So 8 * 32 * (200 + 387) = 150.272 bytes, assuming pointers to certificates
+            // are never aliased pointers (they're Arc'ed).
+            // I think 150KB is an acceptable default upper limit for such a cache.
+            session_store: Arc::new(rustls::client::ClientSessionMemoryCache::new(8 * 32)),
```
dignifiedquire, line 558: lets make this number a constant

iroh/src/endpoint.rs
dignifiedquire, line 695: maybe add a comment why this name is a "good choice"

iroh/src/endpoint.rs
```diff
@@ -2183,4 +2262,110 @@ mod tests {
...
+    async fn test_connect_with_opts_0rtt() -> TestResult {
+        async fn spawn_server() -> Result<Endpoint> {
+            let server = Endpoint::builder()
+                .alpns(vec![TEST_ALPN.to_vec()])
+                .relay_mode(RelayMode::Disabled)
+                .bind()
+                .await?;
+
+            // Gets aborted via the endpoint closing causing an `Err`
+            tokio::spawn({
```
dignifiedquire, line 2277: would be good to test actually sending/receiving some data on the handshake

iroh/src/endpoint.rs
```diff
@@ -1286,13 +1339,38 @@ impl Future for Connecting {
...
                 let conn = Connection { inner };
-                try_send_rtt_msg(&conn, this.ep);
+                try_send_rtt_msg(&conn, this.ep, *this.remote_node_id);
                 Poll::Ready(Ok(conn))
             }
         }
     }
 }
 
+/// Future that completes when a connection is fully established
```
flub, line 1349:
```suggestion
/// Future that completes when a connection is fully established.
```

We follow the rust style guide, not the Quinn style guide.  I know, it's terrible.

iroh/src/endpoint.rs
```diff
@@ -2183,4 +2262,110 @@ mod tests {
...
+                        let mut send = conn.open_uni().await?;
+                        send.write_all(b"thank you").await?;
+                        send.finish()?;
+
+                        // Gracefully close the connection
+                        conn.closed().await;
+                    }
+                    anyhow::Ok(())
+                }
+            });
```
flub, line 2304: `async move { ... }.instrument(info_span!("server"))` does wonders here.

iroh/src/endpoint.rs
```diff
@@ -2183,4 +2262,110 @@ mod tests {
...
+
+        async fn connect_client_expect_0rtt_ok(
+            client: &Endpoint,
+            server_addr: NodeAddr,
+        ) -> Result<()> {
+            let (conn, zero_rtt_accepted) = client
+                .connect_with_opts(server_addr, TEST_ALPN, ConnectOptions::new())
+                .await?
+                .into_0rtt()
+                .map_err(|_| "0rtt failed")
```
flub, line 2337: You could merge these two client functions by adding an extra param to indicate whether 0-rtt is expected to work?  It seems they are otherwise identical so might be easier to maintain.

flub (approved):
Nice!  Thanks for picking this up!

iroh/src/endpoint.rs
```diff
@@ -2183,4 +2262,110 @@ mod tests {
...
+            Ok(())
+        }
+
+        let client = Endpoint::builder()
+            .relay_mode(RelayMode::Disabled)
+            .bind()
+            .await?;
+        let server = spawn_server().await?;
+
+        connect_client_expect_0rtt_err(&client, server.node_addr().await?).await?;
```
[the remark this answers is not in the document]
matheus23: This happens with the `server.node_addr().await?` call on the line you're commenting on.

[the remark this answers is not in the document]
flub: Ahahaha.  lol.  totally missed there was an extra await in there :see_no_evil:

iroh/src/tls.rs
```diff
@@ -91,6 +97,7 @@ pub fn make_server_config(
         warn!("enabling SSLKEYLOGFILE for TLS pre-master keys");
         crypto.key_log = Arc::new(rustls::KeyLogFile::new());
     }
+    crypto.max_early_data_size = u32::MAX; // must be u32::MAX or 0 (the default). Any other value panics with QUIC
```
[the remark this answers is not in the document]
matheus23: :smiling_face_with_tear: 
Yeah.
I should add a [link to the documentation](https://docs.rs/iroh-quinn/latest/iroh_quinn/crypto/rustls/struct.QuicServerConfig.html) or *something* close to that line.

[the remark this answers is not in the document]
matheus23: It comes from specifically this RFC section: https://www.rfc-editor.org/rfc/rfc9001#section-4.6.1
> Servers MUST NOT send the early_data extension with a max_early_data_size field set to any value other than 0xffffffff.

iroh/src/endpoint.rs
```diff
@@ -2183,4 +2262,110 @@ mod tests {
             .unwrap()
             .unwrap();
     }
+
+    #[tokio::test]
+    #[traced_test]
+    async fn test_connect_with_opts_0rtt() -> TestResult {
+        async fn spawn_server() -> Result<Endpoint> {
```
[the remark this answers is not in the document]
matheus23: Agreed. I really contemplated this - I want to split up the test, too - one simple test for 0-RTT and one test with another connection attempt to somewhere else in between.

iroh/src/endpoint.rs
matheus23: Er... what do you mean? This test is actually doing that?

When you get a 0-RTT connection, you get out the 0-RTT data by running `accept_uni` or `accept_bi` after calling `Connecting::into_0rtt()`.

matheus23: Or - are you saying I should make it an echo server, so that it doesn't just go through the motion, but also makes sure that the data flows in/out *correctly*?

iroh/src/endpoint.rs
matheus23: Yep, there's no other reason for it yet, and I haven't seen changes to this break anything anywhere yet.

iroh/src/endpoint.rs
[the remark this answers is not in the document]
matheus23: Split up the test and extracted out these functions.

iroh/src/endpoint.rs
matheus23: I tried merging the functions, but IMO the maintenance effort actually goes up when you realize how weird the function ends up looking. ("What do I do with the `assert!(zero_rtt_accepted.await)`? do I replace that with a dummy value if I don't expect 0-RTT to work? If I replace that with a dummy value, what type does it have? `Either<ZeroRttAccepted, future::Ready<bool>>`? What is the name of that variable, given it's not really checking that 0RTT was accepted?" etc.)
It just gets more confusing. Doing the same mechanical change in two places should e.g. the stream API change is much less confusing.

iroh/src/endpoint.rs
```diff
@@ -1282,7 +1307,44 @@ pub struct Connecting {
...
-    /// Convert into a 0-RTT or 0.5-RTT connection at the cost of weakened security.
+    /// Converts this [`Connecting`] into a 0-RTT or 0.5-RTT connection at the cost of weakened
+    /// security.
+    ///
+    /// Returns `Ok` immediately if the local endpoint is able to attempt sending 0/0.5-RTT data.
+    /// If so, the returned [`Connection`] can be used to send application data without waiting for
+    /// the rest of the handshake to complete, at the cost of weakened cryptographic security
+    /// guarantees. The returned [`ZeroRttAccepted`] future resolves when the handshake does
+    /// complete, at which point subsequently opened streams and written data will have full
+    /// cryptographic protection.
```
flub, line 1318: So the `Connection` object you were already using for 0-rtt will suddenly become 1-rtt?  I think you could make this a little bit more explicit.

iroh/src/endpoint.rs
```diff
@@ -1282,7 +1307,44 @@ pub struct Connecting {
...
+    /// always return `Ok` and the [`ZeroRttAccepted`] will always resolve to true.
+    ///
+    /// ## Security
+    ///
+    /// On outgoing connections, this enables transmission of 0-RTT data, which is vulnerable to
+    /// replay attacks, and should therefore never invoke non-idempotent operations.
+    ///
+    /// On incoming connections, this enables transmission of 0.5-RTT data, which may be sent
+    /// before TLS client authentication has occurred, and should therefore not be used to send
+    /// data for which client authentication is being used.
```
flub, line 1347: Love this new text, it is great!

iroh/src/endpoint.rs
```diff
@@ -1282,7 +1307,44 @@ pub struct Connecting {
...
+    /// local operations on them will return `ZeroRttRejected` errors.
+    ///
+    /// A server may reject 0-RTT data at its discretion, but accepting 0-RTT data requires the
+    /// relevant resumption state to be stored in the server, which servers may limit or lose for
+    /// various reasons including not persisting resumption state across server restarts.
+    ///
+    /// ## Incoming
+    ///
+    /// For incoming connections, conversion to 0.5-RTT will always fully succeed. `into_0rtt` will
+    /// always return `Ok` and the [`ZeroRttAccepted`] will always resolve to true.
```
flub, line 1338: Could be worth pointing to `RecvStream::is_0rtt` as well here?  Because as a server that's about the only real way to know whether you have a 0-rtt stream or not IIUC.

flub (approved):

iroh/src/endpoint.rs
matheus23: Haha thanks - but I mainly copied this from the quinn docs & removed some sections that didn't apply to iroh (because we own the rustls `CryptoConfig`).

iroh/src/endpoint.rs
matheus23: Errr. There's no "becoming 1-rtt". You always need 1 RTT to do a full handshake. Using 0-RTT data simply means you pre-load the handshake with data. Once the handshake finishes, you're back to a normal connection. 0-RTT, 0.5-RTT and 1-RTT only relate to differences in how the connection handshake starts out.

Can you help me come up with docs that would've helped here?

flub: I'm still not sure on the semantics.  "The returned [`ZeroRttAccepted`] future resolves when the handshake does complete, at which point subsequently opened streams and written data will have full cryptographic protection."

- The `Connection` object stays the same: a stream created before `ZeroRttAccepted` completed is 0-rtt, any stream created after a `ZeroRttAccepted` is not 0-rtt.
- "and written data will have full cryptographic protection" is confusing to me.  Do streams opened before `ZeroRttAccepted` suddently become fully protected?  Or does this only refer to data written into streams that were created after `ZeroRttAccepted` completed.

matheus23: >  Do streams opened before ZeroRttAccepted suddently become fully protected?
Yes, but I suspect it's not really something one can depend on - given you don't really know how much of the stream was read in 0-RTT and how much was read after the full handshake. So it's safe to say streams opened before aren't fully protected.

PR #3164, opened 2025-02-03
refactor: cleaning up unnecessary logs

ramfox:
## Description

Cleaning up logs that were only useful when debugging the endpoint closing delay.

## Change checklist

- [x] Self-review.


matheus23 (approved):
These are trace-level logs now, so who cares, but: I don't think we need the `magicsock` prefix, it'll be there with the instrument macro.

PR #3165, opened 2025-02-03
fix(iroh): remove `quinn::Endpoint::wait_idle` from `iroh::Endpoint::close` process

ramfox:
## Description

`quinn::Endpoint::wait_idle` adds a minimum 3 second closing time to `iroh::Endpoint::close` if you have any connections that have not closed gracefully. 

While we *want* users to close connections gracefully, we should also not be forcing users to wait 3 seconds to close the `iroh::Endpoint` if a connection has "gone wrong" if they don't want to.

So instead, we are taking out `quinn::Endpoint::wait_idle` and adding more context for how to close a connection gracefully.

## Notes & open questions

Before 1.0 we will be re-visiting closing to make sure the APIs make sense.

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -991,10 +991,13 @@ impl Endpoint {
     /// of `0` and an empty reason.  Though it is best practice to close those
     /// explicitly before with a custom error code and reason.
     ///
-    /// It will then make a best effort to wait for all close notifications to be
-    /// acknowledged by the peers, re-transmitting them if needed. This ensures the
-    /// peers are aware of the closed connections instead of having to wait for a timeout
-    /// on the connection. Once all connections are closed or timed out, the magic socket is closed.
+    /// This will not wait for the [`quinn::Endpoint`] to drain connections.
+    ///
+    /// To ensure no data is lost, design protocols so that the last *sender*
+    /// of data in the protocol calls [`Connection::closed`], and `await`s until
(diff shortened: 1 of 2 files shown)


dignifiedquire (approved):

iroh/src/endpoint.rs
```diff
@@ -1393,21 +1396,16 @@ impl Connection {
...
-    ///
-    /// [`close`]: Connection::close
+    /// Only the peer last **receiving** application data can be certain that all data is
+    /// delivered.
+    ///
+    /// To communicate to the last **sender** of the application data that all the data was received, we recommend designing protocols that follow this pattern:
+    ///
+    /// 1) The **sender** sends the last data. It then calls [`Connection::closed`]. This will wait until it receives a CONNECTION_CLOSE frame from the other side.
+    /// 2) The **receiver** receives the last data. It then calls [`Connection::close`] and provides an error_code and/or reason.
+    /// 3) The **sender** checks that the error_code is the expected error code.
```
flub, line 1406: This provides no guidance how the receiver of the last data ensures that the close is sent and delivered.  Which possibly is no longer possible right now?

matheus23: Yeah - this seems impossible at the moment, see also this failing test case that's fixed by reverting this PR: https://github.com/n0-computer/iroh/pull/3170/commits/82334ca4e5b41a28270c93f7223081a32058fc9a

PR #3167, opened 2025-02-04
refactor!: use a single DNS resolver

Frando:
## Description

We used to have more-or-less identical code in various places to set a static DNS resolver and add a few extension methods on the `hickory_resolver::TokioResolver`. This PR cleans this up and makes all of iroh use the same resolver, defined in a single place, as a newtype struct around the `TokioResolver` from `hickory_resolver`:

* Adds a newtype struct `DnsResolver` around `hickory_resolver::TokioResolver` in `iroh_relay::dns`. It lives (for now) in `iroh_relay`, because this is a the bottom of our dependency chain: both `iroh` and `iroh-net-report` need to use it as well, and both depend on `iroh_relay`.
* Adds most methods from the previous extension traits to the `DnsResolver` struct
* Changes all imports to use this `DnsResolver` now instead of the previous type alias to TokioResolver
* Removes all now-obsolete duplicate extension traits and type aliases and duplicated tests
* Removes the global `static` DnsResolver that was accessible via `iroh::dns::default_resolver`.  This means that by default multiple endpoints in the same app would not share the DnsResolver. However, I'd say a) it is not too common to use multiple endpoints, and b) if you do and want to share the DnsResolver, you can still do so by creating the DnsResolver yourself and cloning it into the endpoint builder.
* Cleans up and aligns the various methods on `DnsResolver`, and newtype the types from hickory so that this public iroh API does not depend on hickory types (apart from a `From<TokioResolver> for DnsResolver`, which is deferred to a follow-up to investigate)

  The only place outside of tests and examples where a new `DnsResolver` is constructed is in endpoint builder if no custom DnsResolver is set. This means: As long as an app has a single endpoint only, there will only be one DnsResolver.

## Breaking Changes

* `iroh::dns::resolver` and `iroh::dns::default_resolver` are removed. There is no static, global DNS resolver anymore. If you want to share a DNS resolver between endpoints, create the resolver yourself with `iroh::dns::DnsResolver::new` and clone it into the endpoint builders (in `EndpointBuilder::dns_resolver`). If you want to reuse the DNS resolver from an endpoint, you can access it with `Endpoint::dns_resolver` and clone it to wherever you need it.
* `iroh::dns::DnsResolver` used to be a type alias and now is a reexport of `iroh_relay::dns::DnsResolver` struct
* `iroh::dns::node_info` module is now a reexport of `iroh_relay::dns::node_info`
  * `iroh::dns::node_info::{IrohAttr, TxtAttrs, node_id_from_hickory_name}` are no longer public. Use `iroh::dns::DnsResolver::lookup_node_by_id` or `iroh::dns::DnsResolver::lookup_node_by_domain_name` to lookup node info from DNS.
  * `iroh::dns::node_info::{to_z32, from_z32}`are removed. Use the methods on `iroh::dns::node_info::NodeIdExt` trait instead.
* `iroh::discovery::dns::{N0_DNS_NODE_ORIGIN_PROD, N0_DNS_NODE_ORIGIN_STAGING}` are now reexports of `iroh_relay::dns::{N0_DNS_NODE_ORIGIN_PROD, N0_DNS_NODE_ORIGIN_STAGING}`
* `iroh::dns::ResolverExt` is removed. Use the methods on `iroh::dns::DnsResolver` instead.
* The methods in `iroh::dns::DnsResolver` now take an `impl ToString` instead of `impl hickory_proto::rr::domain::IntoName` for their `host` argument

* `iroh::test_utils::create_dns_resolver` is removed, use `iroh::dns::DnsResolver::new` instead

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh-relay/src/dns.rs
+++ b/iroh-relay/src/dns.rs
@@ -7,7 +7,7 @@ use std::{
 };
 
 use anyhow::{bail, Context, Result};
-use hickory_resolver::{IntoName, Resolver, TokioResolver};
+use hickory_resolver::{Resolver, TokioResolver};
 use iroh_base::{NodeAddr, NodeId};
 use n0_future::{
     time::{self, Duration},
@@ -75,27 +75,30 @@ impl DnsResolver {
     }
 
     /// Lookup a TXT record.
-    pub async fn lookup_txt(&self, query: impl IntoName, timeout: Duration) -> Result<TxtLookup> {
-        let res = time::timeout(timeout, self.0.txt_lookup(query)).await??;
+    pub async fn lookup_txt(&self, host: impl ToString, timeout: Duration) -> Result<TxtLookup> {
+        let host = host.to_string();
+        let res = time::timeout(timeout, self.0.txt_lookup(host)).await??;
         Ok(TxtLookup(res))
     }
 
     /// Perform an ipv4 lookup with a timeout.
-    pub async fn lookup_ipv4<N: IntoName>(
+    pub async fn lookup_ipv4(
         &self,
-        host: N,
+        host: impl ToString,
         timeout: Duration,
     ) -> Result<impl Iterator<Item = IpAddr>> {
+        let host = host.to_string();
         let addrs = time::timeout(timeout, self.0.ipv4_lookup(host)).await??;
         Ok(addrs.into_iter().map(|ip| IpAddr::V4(ip.0)))
     }
 
     /// Perform an ipv6 lookup with a timeout.
-    pub async fn lookup_ipv6<N: IntoName>(
+    pub async fn lookup_ipv6(
         &self,
-        host: N,
+        host: impl ToString,
         timeout: Duration,
     ) -> Result<impl Iterator<Item = IpAddr>> {
+        let host = host.to_string();
         let addrs = time::timeout(timeout, self.0.ipv6_lookup(host)).await??;
         Ok(addrs.into_iter().map(|ip| IpAddr::V6(ip.0)))
     }
@@ -105,11 +108,12 @@ impl DnsResolver {
     /// `LookupIpStrategy::Ipv4AndIpv6` will wait for ipv6 resolution timeout, even if it is
     /// not usable on the stack, so we manually query both lookups concurrently and time them out
     /// individually.
-    pub async fn lookup_ipv4_ipv6<N: IntoName + Clone>(
+    pub async fn lookup_ipv4_ipv6(
         &self,
-        host: N,
+        host: impl ToString,
         timeout: Duration,
     ) -> Result<impl Iterator<Item = IpAddr>> {
+        let host = host.to_string();
         let res = tokio::join!(
             self.lookup_ipv4(host.clone(), timeout),
             self.lookup_ipv6(host, timeout)
@@ -161,12 +165,13 @@ impl DnsResolver {
     /// `delays_ms` with the first call being done immediately. `[200ms, 300ms]` results in calls
     /// at T+0ms, T+200ms and T+300ms. The `timeout` is applied to each call individually. The
     /// result of the first successful call is returned, or a summary of all errors otherwise.
-    pub async fn lookup_ipv4_staggered<N: IntoName + Clone>(
+    pub async fn lookup_ipv4_staggered(
         &self,
-        host: N,
+        host: impl ToString,
         timeout: Duration,
         delays_ms: &[u64],
     ) -> Result<impl Iterator<Item = IpAddr>> {
+        let host = host.to_string();
         let f = || self.lookup_ipv4(host.clone(), timeout);
         stagger_call(f, delays_ms).await
     }
@@ -177,12 +182,13 @@ impl DnsResolver {
     /// `delays_ms` with the first call being done immediately. `[200ms, 300ms]` results in calls
     /// at T+0ms, T+200ms and T+300ms. The `timeout` is applied to each call individually. The
     /// result of the first successful call is returned, or a summary of all errors otherwise.
-    pub async fn lookup_ipv6_staggered<N: IntoName + Clone>(
+    pub async fn lookup_ipv6_staggered(
         &self,
-        host: N,
+        host: impl ToString,
         timeout: Duration,
         delays_ms: &[u64],
     ) -> Result<impl Iterator<Item = IpAddr>> {
+        let host = host.to_string();
         let f = || self.lookup_ipv6(host.clone(), timeout);
         stagger_call(f, delays_ms).await
     }
@@ -194,12 +200,13 @@ impl DnsResolver {
     /// at T+0ms, T+200ms and T+300ms. The `timeout` is applied as stated in
     /// [`Self::lookup_ipv4_ipv6`]. The result of the first successful call is returned, or a
     /// summary of all errors otherwise.
-    pub async fn lookup_ipv4_ipv6_staggered<N: IntoName + Clone>(
+    pub async fn lookup_ipv4_ipv6_staggered(
         &self,
-        host: N,
+        host: impl ToString,
         timeout: Duration,
         delays_ms: &[u64],
     ) -> Result<impl Iterator<Item = IpAddr>> {
+        let host = host.to_string();
         let f = || self.lookup_ipv4_ipv6(host.clone(), timeout);
         stagger_call(f, delays_ms).await
     }


iroh-relay/src/dns.rs
```diff
@@ -1,22 +1,205 @@
...
+
+        // see [`DnsResolver::lookup_ipv4_ipv6`] for info on why we avoid `LookupIpStrategy::Ipv4AndIpv6`
+        options.ip_strategy = hickory_resolver::config::LookupIpStrategy::Ipv4thenIpv6;
+
+        let resolver = Resolver::tokio(config, options);
+        Ok(DnsResolver(resolver))
+    }
+
+    /// Create a new `DnsResolver` from a [`TokioResolver`].
+    pub fn from_tokio_resolver(resolver: TokioResolver) -> Self {
```
dignifiedquire, line 66: maybe use `From<TokioResolver>` instead

iroh-relay/src/dns.rs
```diff
@@ -33,33 +216,60 @@ const WINDOWS_BAD_SITE_LOCAL_DNS_SERVERS: [IpAddr; 3] = [
...
     }
-    for name in system_config.search() {
-        config.add_search(name.clone());
+}
+
+/// Staggers calls to the future F with the given delays.
+///
+/// The first call is performed immediately. The first call to succeed generates an Ok result
+/// ignoring any previous error. If all calls fail, an error summarizing all errors is returned.
+// TODO: Should likely not be public, but it is also used in iroh::dns::ResolverExt.
```
dignifiedquire, line 242: this can be fixed now

iroh/src/dns.rs
```diff
@@ -5,108 +5,22 @@
...
-
-    /// Race an ipv4 and ipv6 lookup with a timeout.
-    fn lookup_ipv4_ipv6<N: IntoName + Clone>(
-        &self,
-        host: N,
-        timeout: Duration,
-    ) -> impl Future<Output = Result<impl Iterator<Item = IpAddr>>>;
-
     /// Looks up node info by DNS name.
     fn lookup_by_name(&self, name: &str) -> impl Future<Output = Result<NodeAddr>>;
```
dignifiedquire, line 25: why not move the others as well?

Frando: Went forward with this and moved the node_info module and the lookup methods for nodes to iroh_relay as well.

This means iroh_relay now has a dependency on pkarr (because the functions to parse node info from pkarr packets live on the same structs as parsing from hickory dns records) - without default features though, so should be fine (includes neither the relay nor the dht clients).

iroh-relay/src/dns.rs
Frando: Removed the comment

iroh-relay/src/dns.rs
Frando: :+1:

iroh-relay/src/dns.rs
```diff
@@ -1,22 +1,259 @@
...
-/// It is configured to use the system's DNS configuration.
-pub fn default_resolver() -> &'static DnsResolver {
-    DNS_RESOLVER.get_or_init(|| create_default_resolver().expect("unable to create DNS resolver"))
+impl DnsResolver {
+    /// Create a new DNS resolver with sensible cross-platform defaults.
+    ///
+    /// We first try to read the system's resolver from `/etc/resolv.conf`.
+    /// This does not work at least on some Androids, therefore we fallback
+    /// to the default `ResolverConfig` which uses eg. to google's `8.8.8.8` or `8.8.4.4`.
+    pub fn new_with_defaults() -> Self {
```
flub, line 35: This is an odd name for rust conventions, why not just `new`?  And often `Default::default` is aliased to this as well, clippy even has a lint for when you impl new without args but not Default IIRC.

iroh-relay/src/dns.rs
```diff
@@ -1,22 +1,259 @@
...
+
+        // see [`DnsResolver::lookup_ipv4_ipv6`] for info on why we avoid `LookupIpStrategy::Ipv4AndIpv6`
+        options.ip_strategy = hickory_resolver::config::LookupIpStrategy::Ipv4thenIpv6;
+
+        let resolver = Resolver::tokio(config, options);
+        DnsResolver(resolver)
+    }
+
+    /// Create a new DNS resolver configured with a single UDP DNS nameserver.
+    pub fn new_with_single_nameserver(nameserver: SocketAddr) -> Self {
```
flub, line 62: IIRC the convention is `with_single_nameserver` without the `new` prefix.

iroh-relay/src/dns.rs
```diff
@@ -1,22 +1,259 @@
...
+        DnsResolver(Resolver::tokio(config, Default::default()))
+    }
+
+    /// Removes all entries from the cache.
+    pub fn clear_cache(&self) {
+        self.0.clear_cache();
+    }
+
+    /// Lookup a TXT record.
+    pub async fn lookup_txt(&self, query: impl IntoName) -> Result<TxtLookup, ResolveError> {
```
flub, line 78: Howe come this doesn't have a timeout while the ipv4 & ipv6 looksups do?

I'm also confused by the return type, it's owned by hickory but not `pub use`?  I'm sure I missed something because otherwise that shouldn't work.

Anyway, long way to say that we should own all our types and not re-export new types from external crates.  We can newtype them.

iroh-relay/src/dns/node_info.rs
```diff
@@ -203,9 +205,7 @@ impl NodeInfo {
 /// Takes a [`hickory_resolver::proto::rr::Name`] DNS name and expects the first label to be
 /// [`IROH_TXT_NAME`] and the second label to be a z32 encoded [`NodeId`]. Ignores
 /// subsequent labels.
-pub(crate) fn node_id_from_hickory_name(
-    name: &hickory_resolver::proto::rr::Name,
-) -> Option<NodeId> {
+pub fn node_id_from_hickory_name(name: &hickory_resolver::proto::rr::Name) -> Option<NodeId> {
```
flub, line 208: Unfortunate that this becomes fully `pub`, it sounds very implementation-y

iroh/src/endpoint.rs
```diff
@@ -154,7 +154,7 @@ impl Builder {
         };
         let dns_resolver = self
             .dns_resolver
-            .unwrap_or_else(|| default_resolver().clone());
+            .unwrap_or_else(DnsResolver::new_with_defaults);
```
flub, line 157: This should really be an `.unwrap_or_default()` call.

flub (commented):
I like these changes overall!  Mostly some nitpicking.

I haven't looked at it, but this should take into account #2901 so that whatever it wants will still be possible with this DNS interface.

iroh-relay/src/dns.rs
matheus23: Yeah - I see where you're coming from, but I often see `Default::default()` interpreted as "empty".
And in this case, the `DnsResolver` is sort of a container that "has" nameservers.
So I understand being hesitant implementing `Default::default` or just calling this `new`.

It *might* make sense to implement a `DnsResolver::builder()` that allows you to `.with_nameserver`, or `.with_defaults`, but in the end it might also mostly blow up use sites that now are mostly limited to a single call to `new_with_defaults` or `with_single_nameserver`.

matheus23: Ah - I wrote this up *right* before the latest two commits :upside_down_face: 
Whatever: I like the changed version, too!

Frando: :+1: changed

iroh-relay/src/dns.rs
Frando: :+1: changed

iroh-relay/src/dns.rs
Frando: All these methods were so far left as-is from the existing implementations in the extension traits.

I added a timeout to be in line with the others.

Also tried out adding newtypes for `TxtLookup`. It turned out OK. At the same time I though I realized that we still have quite a few `hickory_resolver` or `hickory_proto` types exposed, e.g. the `IntoName` trait is used in most of the methods on `DnsResolver`. Do we want to newtype everything we expose? I'll do a review of the API next, but many of these hickory types are composed or match with other hickory types, so to cover the full API surface of a Dns resolver that will be quite some newtyping, methods etc.

iroh-relay/src/dns.rs
```diff
@@ -1,22 +1,259 @@
...
+        }
+    }
+
+    /// Resolve a hostname from a URL to an IP address.
+    pub async fn resolve_host(
+        &self,
+        url: &Url,
+        prefer_ipv6: bool,
+        timeout: Duration,
+    ) -> Result<IpAddr> {
```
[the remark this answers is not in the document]
Frando: Taken over from existing usage. Likely we want to change that, yes - it is not used much so will be a simple change to align with the others.

iroh-relay/src/dns/node_info.rs
Frando: Made it `pub(crate)` and added a `node_id_from_domain_name(domain_name: &str)` pub method instead. Small performance penalty if using that while having a hickory `Name` in a crate using this, but in this way the API is not tied to hickory.

iroh-relay/src/dns.rs
```diff
@@ -250,12 +251,55 @@ impl DnsResolver {
...
+
 impl From<TokioResolver> for DnsResolver {
     fn from(resolver: TokioResolver) -> Self {
         DnsResolver(resolver)
     }
 }
 
+/// TXT records returned from [`DnsResolver::lookup_txt`]
+#[derive(Debug, Clone)]
+pub struct TxtLookup(pub(self) hickory_resolver::lookup::TxtLookup);
```
flub, line 268: TIL about `pub(self)` (had to look this up in https://doc.rust-lang.org/reference/visibility-and-privacy.html#r-vis.scoped.self).  I'm confused why this even exists in rust?  Seems like leaving it off is exactly the same, so why do folks want to spell it out sometimes?  What is the motivation here?

iroh-relay/src/dns/node_info.rs
```diff
@@ -200,12 +200,32 @@ impl NodeInfo {
...
 }
 
+/// Parses a [`NodeId`] from a DNS domain name.
+///
+/// Splits the domain name into labels on each dot. Expects the first label to be
+/// [`IROH_TXT_NAME`] and the second label to be a z32 encoded [`NodeId`]. Ignores
+/// subsequent labels.
+///
+/// Returns a [`NodeId`] if parsed successfully, otherwise `None`.
+pub fn node_id_from_domain_name(name: &str) -> Option<NodeId> {
```
flub, line 210: This seems like a pkarr helper method.  Is there any reason to make all this pub?  Maybe this stuff was already pub, but I guess we haven't tried to make the pub API here much smaller yet.  If we don't want to reduce the amount of pub in this PR to keep concerns separate (I often prefer this in my own PRs) I'm also fine to leave this here for now.

flub (approved):
Just more nits, like before nothing serious.  Overall this is good.

iroh-relay/src/dns.rs
```diff
@@ -1,22 +1,303 @@
...
+
+        // see [`DnsResolver::lookup_ipv4_ipv6`] for info on why we avoid `LookupIpStrategy::Ipv4AndIpv6`
+        options.ip_strategy = hickory_resolver::config::LookupIpStrategy::Ipv4thenIpv6;
+
+        let resolver = Resolver::tokio(config, options);
+        DnsResolver(resolver)
+    }
+
+    /// Create a new DNS resolver configured with a single UDP DNS nameserver.
+    pub fn with_single_nameserver(nameserver: SocketAddr) -> Self {
```
dignifiedquire, line 62: nit: I would probably prefer `with_namesever`

iroh-relay/src/dns.rs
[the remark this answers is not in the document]
dignifiedquire: > This is a bit an odd API, why not return impl Iterator<Item = IpAddr> like for the other resolve methods?

the point of this method when I last touched it, is to handle the turning the iterator into a single address, for you

iroh-relay/src/dns/node_info.rs
Frando: It was only used in `iroh::test_utils` - moved it there.

iroh-relay/src/dns.rs
Frando: Yea, dunno, removed it :)

iroh-relay/src/dns.rs
Frando: The remaining usage of `hickory-proto` is now the `IntoName` trait, which efficiently converts `String`, `&str` and `Name`s into the `Name` struct from `hickory_proto`.

iroh-relay/src/dns.rs
Frando: agree, renamed

Frando:
I pushed another round of commits:
* Rename some methods as suggested
* Further reduce public API surface. Move z32 encoding of NodeId into an extension trait instead of public methods.
* Move public methods that were only used in test_utils to test_utils

There's now three remaining uses of `hickory_*` types in the public API surface of the `dns` module:

* `IntoName` trait - we can change this to `ToString` instead, I think (with a perfomance penalty if you have a hickory `Name` already) - see #3173 
* `impl From<hickory_resolver::TokioResolver> for DnsResolver` - we could remove this, it is not used within the iroh repo at the moment. However, this is the current escape hatch for a custom configured `DnsResolver`. Removing that would mean that we'd have to newtype the full, complex builder pattern of hickory_resolver if we wanted to allow custom configuration. Without this `From` impl, we currently only expose the default resolver (using the system's configured nameservers) and a function to create a resolver with a single UDP nameserver.

Frando:
> The From<HickoryResolver> is probably fine as an external type? Can this still get you into trouble if they update it?

We should revisit this before 1.0. I think the impl is no different than a method that takes the TokioResolver, which means we cannot update hickory major version without that being a semver breaking change to iroh...

flub:
> > The From is probably fine as an external type? Can this still get you into trouble if they update it?
> 
> We should revisit this before 1.0. I think the impl is no different than a method that takes the TokioResolver, which means we cannot update hickory major version without that being a semver breaking change to iroh...
> 
> But I'd leave figuring out a solution to a followup PR.

Hum, yes.  You're probably right.  :+1:

PR #3170, opened 2025-02-05
fix(iroh): Allow gracefully closing connections

matheus23:
## Description

This adds a test case that checks that we can call `Connection::close` and actually receive the close code on the other side, indicating we've gracefully closed the connection.
Even though in the test we call `Endpoint::close` and wait for that, with current iroh 0.32, we don't receive the close frame. The only way I can see to make it work with iroh's 0.32 API, is to choose some time to sleep before calling `Endpoint::close`.

I've also attached a revert of #3165 which fixes the test, as that makes `Endpoint::close` wait for the close frames to be acknowledged again (by calling `quinn::Endpoint::wait_idle`).

## Notes & open questions

There might be other ways to fix this, I welcome feedback on that. I do believe this is the right way to fix it, though.

Given that the `Endpoint::wait_idle` call has been introduced and removed *twice* already, we should finally make a choice on whether it belongs in there or not, and then add sufficient code comments to make sure whoever looks at it next is fully informed.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -2513,4 +2513,43 @@ mod tests {
 
         Ok(())
     }
+
+    #[tokio::test]
+    #[traced_test]
+    async fn graceful_close() -> testresult::TestResult {
+        let client = Endpoint::builder().bind().await?;
+        let server = Endpoint::builder()
+            .alpns(vec![TEST_ALPN.to_vec()])
+            .bind()
+            .await?;
+        let server_addr = server.node_addr().await?;
+        let server_task = tokio::spawn(async move {
+            let incoming = server.accept().await.unwrap();
+            let conn = incoming.await?;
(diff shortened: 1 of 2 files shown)


matheus23:
See the PR title as "what this PR aspires to be eventually".

For now, this only contains a very basic test case that shows we write protocols that gracefully close connections right now.

IMO we need to fix this ASAP.

matheus23:
I've pushed a revert of #3165 
That's one way to get this to work again.
Other options:
- Expose calling `quinn::Endpoint::wait_idle` *somehow*
- Come up with some other API, e.g. `Connection::wait_idle` (or something named better)

Essentially all options for "fixing" the test, so we can wait for a proper close again.

Would love input on how people think we should fix this.

ramfox (approved):
Great! And also great comments that will give context to future devs.

PR #3172, opened 2025-02-06
tests(iroh-net-report): Do not ping hosts on the internet

flub:
## description

No matter what we do, pinging hosts on the internet is flaky, packets
do get lost.  We only care that we manage to send and receive pings
correctly, which we fully do by pinging localhost.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-net-report/src/ping.rs
+++ b/iroh-net-report/src/ping.rs
@@ -128,35 +128,6 @@ mod tests {
 
     use super::*;
 
-    #[tokio::test]
-    #[ignore] // Doesn't work in CI
-    #[traced_test]
-    async fn test_ping_google() -> Result<()> {
-        // Public DNS addrs from google based on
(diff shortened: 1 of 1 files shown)


matheus23 (approved):
:joy: thanks.

Also note the `#[ignore] // Doesn't work in CI` line, lol.

As far as I know, this test hasn't caught anything useful for us, soo, I'm happy to see it removed.

flub:
> I disagree, this test is to ensure compliance with std stun servers that are not our own. we can remove that ability but that’s a different statement and change than dropping this test

Huh?  This is an ICMP ping, nothing to do with STUN.

dignifiedquire:
Oh lol, reading is hard, especially early in the morning.

dignifiedquire (approved):

PR #3176, opened 2025-02-10
feat(iroh): publish and resolve user-defined data in discovery

Frando:
## Description

**Based on #3175**

Adds support for publishing and resolving a string of user-defined data through Iroh's discovery services.

This gives applications a way to add additional information to the records published and resolved through discovery, in addition to the addressing information (relay URL and direct addresses). Iroh itself does not parse or use this user-defined data string in any way.

All existing discovery services are updated to support the user data. All DNS-based discovery services encode the user-defined data in a TXT record, in the same way as the DNS/pkarr services already encode the relay URL and direct addresses. Therefore, the length of the `user-data=<user-data>` string is limited to 255 bytes, which is the max length for a single character string in DNS messages. Thus, the length of the actual user-defined data string is limited to 245 bytes (255 minux `user-data=`). This limit is enforced when constructing a `UserData` (which is just a wrapper around a `String` otherwise).

The local swarm discovery uses `swarm-discovery` under the hood, for which I added support for TXT records in https://github.com/rkuhn/swarm-discovery/pull/12. This was released as `swarm-discovery@0.3.0-alpha.2`.

## Notes & open questions

Note that currently the `DiscoveryItem` which contains the user-defined data is only accessible when directly calling the methods on the `Discovery` trait. For discoveries initiated by the endpoint, both active from `Endpoint::connect` and passive from the magicsock's subscription to `Discovery::subscribe`, the `DiscoveryItem`s are currently not exposed in any way, thus there's no way for an app to access the user data of nodes discovered by the endpoint. However, with #3181, there will be a way to watch for all `DiscoveryItem`s found by the endpoint.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-relay/src/dns/node_info.rs
+++ b/iroh-relay/src/dns/node_info.rs
@@ -34,7 +34,7 @@
 
 use std::{
     collections::{BTreeMap, BTreeSet},
-    fmt::Display,
+    fmt::{self, Display},
     hash::Hash,
     net::SocketAddr,
     str::FromStr,
@@ -80,7 +80,9 @@ impl NodeIdExt for NodeId {
 
 /// Data about a node that may be published to and resolved from discovery services.
 ///
-/// This includes an optional [`RelayUrl`] and a set of direct addresses.
+/// This includes an optional [`RelayUrl`], a set of direct addresses, and the optional
+/// [`UserData`], a string that can be set by applications and is not parsed or used by iroh
+/// itself.
 ///
 /// This struct does not include the node's [`NodeId`], only the data *about* a certain
 /// node. See [`NodeInfo`] for a struct that contains a [`NodeId`] with associated [`NodeData`].
@@ -90,6 +92,8 @@ pub struct NodeData {
     relay_url: Option<RelayUrl>,
     /// Direct addresses where this node can be reached.
     direct_addresses: BTreeSet<SocketAddr>,
+    /// Optional user-defined [`UserData`] for this node.
+    user_data: Option<UserData>,
 }
 
 impl NodeData {
@@ -98,6 +102,7 @@ impl NodeData {
         Self {
             relay_url,
             direct_addresses,
+            user_data: None,
         }
     }
 
@@ -113,9 +118,20 @@ impl NodeData {
         self
     }
 
+    /// Sets the user-defined data and returns the updated node data.
+    pub fn with_user_data(mut self, user_data: Option<UserData>) -> Self {
+        self.user_data = user_data;
+        self
+    }
+
     /// Returns the relay URL of the node.
     pub fn relay_url(&self) -> Option<&RelayUrl> {
         self.relay_url.as_ref()
+    }
+
+    /// Returns the optional user-defined data of the node.
+    pub fn user_data(&self) -> Option<&UserData> {
+        self.user_data.as_ref()
     }
 
     /// Returns the direct addresses of the node.
@@ -137,6 +153,11 @@ impl NodeData {
     pub fn set_relay_url(&mut self, relay_url: Option<RelayUrl>) {
         self.relay_url = relay_url
     }
+
+    /// Sets the user-defined data of the node data.
+    pub fn set_user_data(&mut self, user_data: Option<UserData>) {
+        self.user_data = user_data;
+    }
 }
 
 impl From<NodeAddr> for NodeData {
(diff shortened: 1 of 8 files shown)


iroh-relay/src/dns/node_info.rs
```diff
@@ -137,17 +153,79 @@ impl NodeData {
...
     fn from(node_addr: NodeAddr) -> Self {
         Self {
             relay_url: node_addr.relay_url,
             direct_addresses: node_addr.direct_addresses,
+            user_data: None,
+        }
+    }
+}
+
+// User defined data that can be published and resolved through node discovery.
```
ramfox, line 173:
```suggestion
/// User-defined data that can be published and resolved through node discovery.
```

iroh-relay/src/dns/node_info.rs
```diff
@@ -137,17 +153,79 @@ impl NodeData {
...
+///
+/// Under the hood this is a UTF-8 String that is less than or equal to
+/// [`USER_DATA_MAX_LENGTH`] bytes.
+///
+/// Iroh does not keep track of or examine user defined data.
```
ramfox, line 178:
```suggestion
/// Iroh does not keep track of or examine user-defined data.
```

iroh-relay/src/dns/node_info.rs
```diff
@@ -137,17 +153,79 @@ impl NodeData {
...
+#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub struct UserData(String);
+
+/// The max byte length allowed for user defined data.
```
ramfox, line 182:
```suggestion
/// The max byte length allowed for user-defined data.
```

iroh/src/discovery/static_provider.rs
```diff
@@ -121,19 +124,23 @@ impl StaticProvider {
...
     }
 
     /// Sets node addressing information for the given node ID.
     ///
     /// This will completely overwrite any existing info for the node.
-    pub fn set_node_addr(&self, info: impl Into<NodeAddr>) -> Option<NodeAddr> {
+    pub fn set_node_addr(
+        &self,
+        node_addr: impl Into<NodeAddr>,
+        user_data: Option<UserData>,
```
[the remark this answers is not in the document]
Frando: :+1: Fixed in #3175 and thus here too (this PR is rebased on top of 3175)__

iroh/src/discovery/static_provider.rs
```diff
@@ -144,7 +151,7 @@ impl StaticProvider {
     /// The provided addressing information is combined with the existing info in the static
     /// provider.  Any new direct addresses are added to those already present while the
     /// relay URL is overwritten.
-    pub fn add_node_addr(&self, info: impl Into<NodeAddr>) {
+    pub fn add_node_addr(&self, info: impl Into<NodeAddr>, user_data: Option<UserData>) {
```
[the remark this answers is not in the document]
Frando: Also fixed in #3175  (but with `impl Into<NodeInfo>` not `NodeData` because we also need the node id).

iroh-relay/src/dns/node_info.rs
Frando: Fixed

iroh-relay/src/dns/node_info.rs
Frando: Fixed, and also moved the const onto the `UserData` struct

ramfox (approved):
looks good! @rklaehn might be a good second set of eyes for opinions on this and the next PR #3181.

iroh/src/discovery/local_swarm_discovery.rs
```diff
@@ -63,6 +63,8 @@ const N0_LOCAL_SWARM: &str = "iroh.local.swarm";
 /// Used in the [`crate::endpoint::Source::Discovery`] enum variant as the `name`.
 pub const NAME: &str = "local.swarm.discovery";
 
+const USER_DATA_ATTRIBUTE: &str = "user-data";
```
dignifiedquire, line 66: comment for what this is used?

iroh/src/discovery/local_swarm_discovery.rs
```diff
@@ -348,7 +355,16 @@ fn peer_to_discovery_item(peer: &Peer, node_id: &NodeId) -> DiscoveryItem {
         .iter()
         .map(|(ip, port)| SocketAddr::new(*ip, *port))
         .collect();
-    let node_info = NodeInfo::new(*node_id).with_direct_addresses(direct_addresses);
+    // Get the user-defined data from the resolved peer info. We expect an attribute with a value
+    // that parses as `UserData`. Otherwise, omit.
+    let user_data = if let Some(Some(user_data)) = peer.txt_attribute(USER_DATA_ATTRIBUTE) {
+        user_data.parse().ok()
```
dignifiedquire, line 361: I think we should print at least a debug message of the errror here, so that it isn't fully silent if there is invalid data floating around, and you are trying to debug why it can't get resolved

dignifiedquire (approved):

PR #3179, opened 2025-02-11
fix(iroh): Remove wrong tracing::record call

flub:
## Description

This was left over from when the spans were created differently.  Now
the span is created on `connect_quinn` and that record call was adding
a field to the wrong span.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


flub:
@matheus23 is doing this in another PR somewhere.

PR #3180, opened 2025-02-11
chore: add additional todos in the "change checklist"

ramfox:
## Description

In order to be better prepared for releases, we are trying to "front load" more work on understanding the effects of breaking changes in our downstream dependencies. Part of this is to remind folks to take a look at the downstream dependencies when they make a breaking change, and file an issue (or they have capacity), open a PR on how to deal with those breaking changes in that dependency.

The PR template looks like this now:

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.
    - [ ] `quic-rpc`
    - [ ] `iroh-gossip`
    - [ ] `iroh-blobs`
    - [ ] `dumbpipe`
    - [ ] `sendme`

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -16,7 +16,7 @@
 - [ ] Documentation updates following the [style guide](https://rust-lang.github.io/rfcs/1574-more-api-documentation-conventions.html#appendix-a-full-conventions-text), if relevant.
 - [ ] Tests if relevant.
 - [ ] All breaking changes documented.
-  - [ ] Document or create PRs for how these changes will effect downstream dependencies. The major ones are:
+  - [ ] Open an issue or PR on any affected downstream dependency repos. The major ones are:
     - [ ] `quic-rpc`
     - [ ] `iroh-gossip`
     - [ ] `iroh-blobs`
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -16,3 +16,9 @@
 - [ ] Documentation updates following the [style guide](https://rust-lang.github.io/rfcs/1574-more-api-documentation-conventions.html#appendix-a-full-conventions-text), if relevant.
 - [ ] Tests if relevant.
 - [ ] All breaking changes documented.
+  - [ ] Document or create PRs for how these changes will effect downstream dependencies. The major ones are:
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -11,8 +11,14 @@
 <!-- Any notes, remarks or open questions you have to make about the PR. -->
 
 ## Change checklist
-
+<!-- Remove any that are not relevant. -->
```
flub, line 14: Oh, you're daring to add explicit instructions!  I had given up on this :D

flub (approved):

ramfox:
> Also, is this really a subtask of "document all breaking changes?

I think yes! It's literally about documenting breaking changes, including adding instructions about how breaking changes are going to effect our other repos.

divagant-martian:
I tried to think how this work and now I'm a bit _less convinced_, so sorry for holding this up. Maybe at the end you are right anyway.

What's the case for a release that has multiple breaking changes? is there one issue per repo per breaking change then? What should these issues contain as content? I would think they have the "how to update" but it's that not what simply documenting the breaking changes here should achieve? I know these dependants are those we own but I would think our breaking changes description should be good enough to allow _any_ dependant to update correctly

flub:
> So the point of creating the issue is not to mention the breaking change in another place - but to take the architecture of the repository that needs to be updated into account and come up with an upgrade plan - or even propose a concrete upgrade with a PR.

Yeah, I interpret this as: create a draft PR with a patch to the branch with your breaking change.  If it's simple it's fine.  If all things break you'll notice.  At this point you can either try and solve it or describe what needs to be done in an issue I guess.  But at least it'll be more visible.

As part of this I'm also guessing that the `main` branch of those repos will often end up living with a patch pointing to the iroh main branch in between releases.

But I'm making a lot more assumptions then is being described.  I understand the template tries to strike a balance with keeping it light though.

ramfox:
> Yeah, I interpret this as: create a draft PR with a patch to the branch with your breaking change. If it's simple it's fine. If all things break you'll notice. At this point you can either try and solve it or describe what needs to be done in an issue I guess. But at least it'll be more visible.

This is my intension.

dignifiedquire (approved):

PR #3181, opened 2025-02-12
feat(iroh): watch discovery

Frando:
## Description

Implements a subscription stream to watch all items retrieved through discovery services. the stream includes both nodes discovered actively (through `Discovery::resolve`, which is invoked when `Endpoint::connect` is called for nodes we don't have addressing info), and nodes discovered passively (through `Discovery::subscribe`, which is only implemented for the local swarm discovery).

Uses `tokio::sync:.broadcast` internally (but without exposing this in the public API) which fits well here I think: if a consumer is too slow in processing the stream, it will miss events but can still continue.

Currently the capacity of the broadcast channel is hardcoded to 128. We could also expose this as a setting on the endpoint builder, not sure if there is a default that fits everything.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/discovery.rs
+++ b/iroh/src/discovery.rs
@@ -407,10 +407,12 @@ impl DiscoveryTask {
                         continue;
                     }
                     debug!(provenance = %r.provenance, addr = ?r.node_addr, "discovery: new address found");
-                    ep.add_node_addr_with_source(r.node_addr, r.provenance).ok();
+                    ep.add_node_addr_with_source(r.node_addr.clone(), r.provenance)
+                        .ok();
                     if let Some(tx) = on_first_tx.take() {
                         tx.send(Ok(())).ok();
                     }
+                    ep.discovery_subscribers().send(r);
                 }
                 Some(Err(err)) => {
                     warn!(?err, "discovery service produced error");
@@ -429,6 +431,26 @@ impl DiscoveryTask {
 impl Drop for DiscoveryTask {
     fn drop(&mut self) {
         self.task.abort();
+    }
+}
+
+#[derive(Clone, Debug)]
+pub(super) struct DiscoverySubscribers {
+    inner: tokio::sync::broadcast::Sender<DiscoveryItem>,
+}
+
+impl DiscoverySubscribers {
+    pub(crate) fn new() -> Self {
+        Self {
+            inner: tokio::sync::broadcast::Sender::new(1024),
+        }
+    }
+    pub(crate) fn subscribe(&self) -> tokio::sync::broadcast::Receiver<DiscoveryItem> {
+        self.inner.subscribe()
+    }
+
+    pub(crate) fn send(&self, item: DiscoveryItem) {
+        self.inner.send(item).ok();
     }
 }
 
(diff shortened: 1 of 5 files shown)


ramfox (commented):
Oh very cool. So the idea here is not to mess with the `Discovery` trait or impls at all. I think this is a good way to handle it, personally.

iroh/src/magicsock.rs
```diff
@@ -2330,6 +2339,7 @@ impl Actor {
                         }) {
                         warn!(?discovery_item.node_addr, "unable to add discovered node address to the node map: {e:?}");
                     }
+                    self.msock.discovery_subscribers.send(discovery_item);
```
ramfox, line 2342: I'm assuming the "lagging" concept will prevent this from ever locking/hanging correct?

iroh/src/discovery.rs
```diff
@@ -432,6 +435,47 @@ impl Drop for DiscoveryTask {
...
+pub struct Lagged(pub u64);
+
+#[derive(Clone, Debug)]
+pub(super) struct DiscoverySubscribers {
+    inner: tokio::sync::broadcast::Sender<DiscoveryItem>,
+}
+
+impl DiscoverySubscribers {
+    pub(crate) fn new() -> Self {
+        // TODO: Make capacity configurable from the endpoint builder?
```
flub, line 457: I'd rather avoid this if possible.  We don't want too many crazy knobs that no one understands in the endpoint builder.

iroh/src/endpoint.rs
```diff
@@ -919,6 +920,31 @@ impl Endpoint {
...
+    ///
+    /// Whenever a node is discovered via the endpoint's discovery service, the corresponding
+    /// [`DiscoveryItem`] is yielded from this stream. This includes nodes discovered actively
+    /// through [`Discovery::resolve`], which is invoked automatically when calling
+    /// [`Endpoint::connect`] for a [`NodeId`] unknown to the endpoint. It also includes
+    /// nodes that the endpoint discovers passively from discovery services that implement
+    /// [`Discovery::subscribe`], which e.g. [`LocalSwarmDiscovery`] does.
+    ///
+    /// The stream does not yield information about nodes that are added manually to the endpoint's
+    /// addressbook by calling [`Endpoint::add_node_addr`] or by supplying a full [`NodeAddr`] to
```
flub, line 933: What do you think should happen when `StaticProvider::add_node_addr` is called?  Should it show up here?

Frando: Good question. I'm not sure?

Right now they would show up in the discovery stream only once you connect to a node added to the StaticProvider, but not at the time you add them to the StaticProvider.

Frando: Added docs about current behavior. 

We could change this by having the `StaticProvider` impl `Discovery::subscribe` and emit a stream of all items added there, but that could be a followup (then it would also behave more like `add_node_addr` in that it adds the info the magic sock node map right away, not only on `connect` - not sure if that is desired?)

dignifiedquire (approved):
I think the semantics are tricky to get right, but this is a good step forward and we can iterate on it from here

PR #3182, opened 2025-02-13
feat(iroh-relay): Make `Endpoint::close` faster by aborting QAD connections faster

matheus23:
## Description

We reduce the RTT estimate for QAD connections. This means we don't wait as long for close frame acknowledgements when we abort these connection attempts.

Sooo no more 3s `Endpoint::close` time due to QAD probes never closing properly because they're sent into a black hole somewhere in the network path.

## Notes & open questions

The 111ms is arbitrarily chosen to give us a 1s probe timeout. Let me know if I'm missing something important here.

Keep in mind this doesn't prevent us from connecting if the *actual* RTT is longer. The test ensures that.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-relay/src/quic.rs
+++ b/iroh-relay/src/quic.rs
@@ -333,7 +333,7 @@ mod tests {
 
     #[tokio::test]
     #[traced_test]
-    async fn quic_endpoint_closes_fast() -> anyhow::Result<()> {
+    async fn test_qad_client_closes_unresponsive_fast() -> anyhow::Result<()> {
         // create a client-side endpoint
         let client_endpoint =
             quinn::Endpoint::client(SocketAddr::new(Ipv4Addr::LOCALHOST.into(), 0))?;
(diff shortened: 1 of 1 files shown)


iroh-relay/src/quic.rs
```diff
@@ -312,4 +330,133 @@ mod tests {
...
+
+        println!("Closed in {time:?}");
+        assert!(Duration::from_millis(900) < time);
+        assert!(time < Duration::from_millis(1100));
+
+        Ok(())
+    }
+
+    /// Makes sure that, even though the RTT was set to some fairly low value,
+    /// we *do* try to connect longer than the probe timeout (1s), as long as
```
flub, line 378: "longer than 3 * probe timeout" I think.

flub (approved):

flub:
ugh, apologies about the outdated comments.  I wrote those yesterday before I played with it and github suddenly surfaces them now somehow without giving me the option of deleting them...

PR #3185, opened 2025-02-14
Fix test, add test

flub:
Fix the test by not using a fixed port: that's asking for flakyness.

Add test ensuring that the connection still succeeds even after that
1s deadline exceeds.


matheus23 (approved):
Interesting!
So essentially `quic_endpoint_closes_fast` behaves the same as before, but only because we don't make use of ICMP packets.

And `quic_endpoint_unresponsive` makes sure that we *do* have the *eventually connecting* behavior we're looking for, even if the connection takes *more* than 1s.

Sweet, I like it. I think I'll change the test names a bit, but other than that :+1:

PR #3186, opened 2025-02-14
deps: update pkarr to v3

Frando:
## Description

a mostly mechanical stab at updating pkarr to pkarr v3. the pkarr client now supports both relays and DHT, so the DHT discovery code becomes a bit simpler.

let's see what CI says (didn't test yet)

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


Nuhvi:
@Frando The wasm error is what I get for updating to `rand 0.9.0` which demands `getrandom 0.3.0` which  requires some [massaging](https://github.com/rust-random/getrandom?tab=readme-ov-file#opt-in-backends), even worse that sometimes you just have to do the `0.2.0` [massaging](https://docs.rs/getrandom/0.2.15/getrandom/index.html#webassembly-support) as well if you have other dependencies that require it.

Long story short you need to add `getrandom 0.3.0` with `wasm_js` feature _and_ add this too https://github.com/pubky/pkarr/blob/main/pkarr/.cargo/config.toml

Nuhvi:
I am not sure how would iroh-relay run in wasm, but anyway, I think the .cargo/config.toml should go in that crate.

Frando:
Note: While all tests pass this is not yet good to go.  the PkarrClient in `pkarr 3.4` has a bug in that it unconditionally DNS-resolves the default bootstrap URLs even if configured differently (i.e. without DHT or with different bootstrap URLs) because it does the resolving in a `Default` impl where this is not checked.

> even when initializing a `ClientBuilder`with `no_default_networks` or `no_dht`, the `ClientBuilder::default` which is called from `Client::builder`  goes to `Config::default` which unconditionally calls `Dht::builder`, which at least in the most recent mainline release  goes to `Dhtbuilder::default` and then `mainline::rpc::Config::default` which unconditionally calls `bootstrap: to_socket_address(&DEFAULT_BOOTSTRAP_NODES)` which (synchrounsly, blockingly!) DNS-resolves the default bootstrap URLs - even if the DHT is not used because `no_default_networks` is called
> even if using the DHT, a sync-blocking call to `to_socket_address` for ~4 URLs is a bit of a nogo, it will block the tokio event loop for ~2s for me
> but it definitely should not do that if the DHT is disabled in the pkarr client builder
> (frando in discord)

So let's wait for a new release that fixes these issues.

Nuhvi:
@Frando published `pkarr@3.5.0`, and besides fixing the mainline issue, pkarr now re-exports mainline, so you can skip adding mainline as an explicit dependency.

Nuhvi:
`pkarr v3.6.0` have few bug fixes, but also enables setting timeout for requests to relays in wasm (after Reqwest published a new [version](https://github.com/seanmonstar/reqwest/blob/e44e37130538cc2d3c95ffee1215b4db3d71b025/CHANGELOG.md?plain=1#L12)).

Frando:
Alright, I updated this PR:
* Merged `main`
* Update to `pkarr v3.6`
* Remove explicit mainline dependency
* Remove debug leftovers

Let's see what CI says, but I think this should be good to go now.

iroh-dns-server/src/lib.rs
```diff
@@ -101,13 +101,13 @@ mod tests {
                 30,
                 dns::rdata::RData::AAAA(Ipv6Addr::LOCALHOST.into()),
             ));
-            SignedPacket::from_packet(&keypair, &packet)?
+            SignedPacket::new(&keypair, &packet.answers, Timestamp::now())?
```
Nuhvi, line 104: It is not important here, but if there are other places where you have similar verbosity, reminder that the new `SignedPacket::builder()` api is much nicer.

Arqu:
> Alright, I updated this PR:
> 
>     * Merged `main`
> 
>     * Update to `pkarr v3.6`
> 
>     * Remove explicit mainline dependency
> 
>     * Remove debug leftovers
> 
> 
> Let's see what CI says, but I think this should be good to go now.
> 
> Requested a review from @rklaehn as the biggest changes are in the pkarr dht discovery impl. It became quite a bit simpler, as more of the logic to use both relays and the DHT is now handled by pkarr itself.
> 
> There's also some changes in the DNS server. To make sure that we don't break anything: @Arqu, can we testdrive this on staging or in a test environment somehow easily?

Yes, we can set this up on the staging dns instance and do some tests.

iroh/src/discovery/pkarr/dht.rs
```diff
@@ -90,57 +81,29 @@ struct Inner {
...
 
         let maybe_packet = self.pkarr.resolve(&key).await;
         match maybe_packet {
-            Ok(Some(signed_packet)) => match NodeInfo::from_pkarr_signed_packet(&signed_packet) {
+            Some(signed_packet) => match NodeInfo::from_pkarr_signed_packet(&signed_packet) {
                 Ok(node_info) => {
-                    tracing::info!("discovered node info from DHT {:?}", &node_info);
-                    Some(Ok(DiscoveryItem::new(node_info, "mainline", None)))
+                    tracing::info!("discovered node info {:?}", node_info);
+                    Some(Ok(DiscoveryItem::new(node_info, "dht", None)))
```
dignifiedquire, line 96: is this always `dht` or should we rename this to `pkarr`?

Frando: Good point, it supports both relay and DHT, so changed to `pkarr`.

rklaehn (approved):
Changes seem like a straightforward API update.

I tried the dht_discovery example. It still seems to work.

Not sure if pkarr has gotten heavier with this update, but since it is an optional feature I guess we will have to live with it.

Did a quick cargo tree, and didn't see anything unusual. Just the usual rust madness.

PR #3145, opened 2025-01-22
feat(iroh): Make `iroh` compile to Wasm & run in browsers

matheus23:
## Description

So far this enables this demo: https://github.com/matheus23/browser-iroh-net-demo/blob/update/src/lib.rs

## Notes & open questions

Needs lots of love and refactoring, and other PRs landing first.

## Change checklist

- [ ] Self-review.
- [ ] Documentation updates following the [style guide](https://rust-lang.github.io/rfcs/1574-more-api-documentation-conventions.html#appendix-a-full-conventions-text), if relevant.
- [ ] Tests if relevant.
- [ ] All breaking changes documented.

PR #3189, opened 2025-02-19
feat(iroh): Make `iroh` compile to `wasm32-unknown-unknown`

matheus23:
## Description

(This PR is a rebase & some changes from the #3145 PR, so we don't disturb users who are trying out the iroh browser alpha.)

- Adds a bunch of `#[cfg(not(wasm_browser))]` to anything related to socket networking (UDP sockets, DNS, etc.)
- Implements a `WebRuntime` for quinn to use instead of `TokioRuntime` in browsers
- Implements sleeping & timing utilities for `backoff` that work in browsers
- Moves `node_info` out of the `dns` module, so we can disable `dns` as a whole, but keep `node_info` for discovery
- Disables `MappedAddr::Ip` in browsers
- Adjusts dependencies so we don't depend on unused/unsupported things in browsers

## Breaking Changes

- `iroh-relay`: The `iroh_relay::dns::node_info` module was moved to `iroh_relay::node_info`

## Notes & open questions

I'm holding off on refactoring bigger parts for now. I have two main ideas:
- Group a bunch of structs in `magicsock.rs`, i.e. everything that depends on sockets.
  However, to do this, I want to refactor the way we *sometimes* use `Arc<netwatch::UdpSocket>` and *sometimes* use `UdpConn`, which is just a newtype over said type.
  Ideally we store `Arc<netwatch::UdpSocket>` everywhere, since it's more commonly used in APIs, e.g. in iroh-net-report, and only use `UdpConn`, when we need the udp sockets to implement `AsyncUdpSocket` from quinn.
- Refactor the `node_info` module so there's fewer places that depend on `DnsResolver` in it.
  There's a bunch of things in that module that IMO are needless indirections (we go from `hickory::TxtLookup` -> `node_info::TxtAttrs<IrohAttr>` -> `(NodeId, NodeData)` -> `NodeInfo`. And even then, `NodeInfo` is mostly isomorphic to `NodeAttr`).

I think both these refactors are helpful, but don't necessarily make the diff easier to read? Not sure. I'd like to merge this & then work on these refactors next week.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh-relay/src/lib.rs
```diff
@@ -46,6 +46,7 @@ pub(crate) use key_cache::KeyCache;
 
 #[cfg(not(wasm_browser))]
 pub mod dns;
+pub mod node_info;
```
matheus23, line 49: Moving `node_info` from `crate::dns::node_info` to `crate::node_info`, because we need it for discovery in browsers, but it's still nice to disable the whole `dns` module in browsers, too.

iroh/src/discovery.rs
```diff
@@ -496,12 +498,6 @@ impl DiscoveryTask {
     }
 }
 
-impl Drop for DiscoveryTask {
-    fn drop(&mut self) {
-        self.task.abort();
-    }
-}
-
```
matheus23, line 504: `DiscoveryTask::task` is an `AbortOnDropHandle`.

iroh-relay/src/node_info.rs
```diff
@@ -394,23 +401,24 @@ impl<T: FromStr + Display + Hash + Ord> TxtAttrs<T> {
...
 
     /// Returns the node id.
-    pub(super) fn node_id(&self) -> NodeId {
+    pub(crate) fn node_id(&self) -> NodeId {
         self.node_id
     }
 
     /// Parses a [`pkarr::SignedPacket`].
-    pub(super) fn from_pkarr_signed_packet(packet: &pkarr::SignedPacket) -> Result<Self> {
+    pub(crate) fn from_pkarr_signed_packet(packet: &pkarr::SignedPacket) -> Result<Self> {
```
dignifiedquire, line 421: can we do pkarr based lookups in the browser? are we doing that already?

iroh/Cargo.toml
```diff
@@ -12,6 +12,9 @@ keywords = ["quic", "networking", "holepunching", "p2p"]
 # Sadly this also needs to be updated in .github/workflows/ci.yml
 rust-version = "1.81"
 
+[lib]
+crate-type = ["lib", "cdylib"]
```
dignifiedquire, line 16: can you add a comment  about why these two are here

iroh/src/magicsock.rs
```diff
@@ -771,7 +793,9 @@ impl MagicSock {
             };
         }
 
+        #[cfg(not(wasm_browser))]
         let counter = self.poll_recv_counter.fetch_add(1, Ordering::Relaxed);
+        #[cfg(not(wasm_browser))]
```
dignifiedquire, line 798: the two cfgs could be in a single block, making this a bit nicer

iroh/src/magicsock.rs
```diff
@@ -2257,10 +2375,26 @@ impl Actor {
...
+
+            #[cfg(not(wasm_browser))]
+            let direct_addr_heartbeat_timer_tick = direct_addr_heartbeat_timer.tick();
+            #[cfg(wasm_browser)]
+            let direct_addr_heartbeat_timer_tick = n0_future::future::pending();
+
+            #[cfg(not(wasm_browser))]
+            let link_change_r_recv = link_change_r.recv();
+            #[cfg(wasm_browser)]
+            let link_change_r_recv = n0_future::future::pending();
```
dignifiedquire, line 2396: not for this PR, but we should totally implement this, detecting connection changes seems to be sensible, even if we don't do direct connections

iroh-relay/src/node_info.rs
matheus23: Yes we can! There's a basic HTTP-based route on the dns server we have (`/relay`).
See also `discovery_n0` in the builder - and how we replace `DnsDiscovery` with `PkarrResolver` in the browser.

iroh/src/magicsock.rs
matheus23: Yepyep. Filed https://github.com/n0-computer/iroh/issues/3195 for this.

iroh/src/magicsock.rs
matheus23: Thanks for making me look at this again, doing the cfg on the whole function was much much better:
```rs
/// NOTE: Receiving on a [`Self::closed`] socket will return [`Poll::Pending`] indefinitely.
#[instrument(skip_all)]
#[cfg(not(wasm_browser))]
fn poll_recv(
    &self,
    cx: &mut Context,
    bufs: &mut [io::IoSliceMut<'_>],
    metas: &mut [quinn_udp::RecvMeta],
) -> Poll<io::Result<usize>> {
    // ...
}

/// poll_recv in browsers is "just" polling the relay receive path
#[cfg(wasm_browser)]
fn poll_recv(
    &self,
    cx: &mut Context,
    bufs: &mut [io::IoSliceMut<'_>],
    metas: &mut [quinn_udp::RecvMeta],
) -> Poll<io::Result<usize>> {
    self.poll_recv_relay(cx, bufs, metas)
}
```

iroh-relay/src/node_info.rs
dignifiedquire: lovely

PR #3190, opened 2025-02-19
fix: `Endpoint::node_addr` completes with direct addrs or relay ready

Frando:
## Description

Currently, `Endpoint::node_addr` waits for the endpoint's direct addresses to be initialized, and then returns. This is an issue for iroh in the browser, because there we never have direct addresses.

This PR changes the behavior of `Endpoint::node_addr` so that it completes once either the direct addresses or the home relay URL are initialized.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -2317,6 +2317,8 @@ mod tests {
         let ep1_nodeid = ep1.node_id();
         let ep2_nodeid = ep2.node_id();
 
+        // wait for the direct addresses to be initialized.
+        ep1.direct_addresses().initialized().await.unwrap();
         let ep1_nodeaddr = ep1.node_addr().await.unwrap();
         tracing::info!(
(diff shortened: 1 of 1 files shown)


matheus23 (approved):
IMHO we could also make `node_addr` infallible, because all the `?` only fail when the `Watcher`s are disconnected. Which should be impossible, given we have a `&self` reference alive, which holds the respective `Watchable` at the same time.

If there's errors while getting the home relay or direct addresses, these errors aren't propagated to here anyways - instead it'll fail to resolve :/

Frando:
One test in iroh needed fixing for this. I am still not totally convinced that this is the right approach. See #3192  for an alternative.

Frando:
This was implemented, albeit a bit differently, in #3279:  `Endpoint::node_addr` now behaves differently in `wasm_browser`  target, where it resolves with only a relay URL.

PR #3192, opened 2025-02-19
refactor: Make `Endpoint::node_addr` sync and infallible, and add `Endpoint::node_addr_initialized`

Frando:
## Description

This makes `Endpoint::node_addr` a synchronous, infallible function that always returns the current state of the endpoint's node address. Right after binding that node address might be empty, because neither the home relay nor our direct addresses are yet determined.

Therefore, we add a `async fn node_addr_initialized(&self)` that waits for either the home relay or the direct addresses to be initialized.

Alternative to #3190 

## Breaking Changes

* `Endpoint::node_addr` now is sync and infallible. If you want to wait for the address to be initialized, await `Endpoint::node_addr_initialized` before.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/examples/connect-unreliable.rs
+++ b/iroh/examples/connect-unreliable.rs
@@ -50,7 +50,7 @@ async fn main() -> anyhow::Result<()> {
         .bind()
         .await?;
 
-    let node_addr = endpoint.node_addr().await?;
+    let node_addr = endpoint.node_addr();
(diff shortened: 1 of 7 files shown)


Frando:
This was implemented as part of #3279: `Endpoint::node_addr` now returns a `Watcher`.

PR #3194, opened 2025-02-21
fix(iroh-relay): Fix the number of active relay connections

flub:
## Description

The disconnects metric was accidentally removed and this has been
giving us wrong numbers since forever and it is so painful yet so easy
to fix.

## Change checklist

- [x] Self-review.


rklaehn (approved):

dignifiedquire (approved):

Arqu:
Thank you!

PR #3093, opened 2025-01-03
refactor(iroh-relay)!: server actor task is not a task or actor anymore

dignifiedquire:
## Description

- Restructures the server part of the relay to not mainly run an actor, but onyl proxy between a single actor per connection.
- removes notifying the server about preferred/home relay status, to simplify the protocol, as this was only used for metrics

Replaces #3073

## Breaking Changes

- renamed `iroh_relay::server::ClientConnRateLimit` to `ClientRateLimit`
- removed `iroh_relay::server::MaybeTlsStreamServer` from the public API
- removed `iroh_relay::client::SendMessage::NotePreferred`

## Change checklist

- [ ] Self-review.
- [ ] Documentation updates following the [style guide](https://rust-lang.github.io/rfcs/1574-more-api-documentation-conventions.html#appendix-a-full-conventions-text), if relevant.
- [ ] Tests if relevant.
- [ ] All breaking changes documented.

PR #3197, opened 2025-02-24
fix(iroh-relay): bring back unique node counts

Arqu:
## Description

This got lost in https://github.com/n0-computer/iroh/pull/3093 and wasn't covered in https://github.com/n0-computer/iroh/pull/3194

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-relay/src/server/client.rs
+++ b/iroh-relay/src/server/client.rs
@@ -1,6 +1,9 @@
 //! The server-side representation of an ongoing client relaying connection.
 
-use std::{future::Future, num::NonZeroU32, pin::Pin, sync::Arc, task::Poll, time::Duration};
+use std::{
+    collections::HashMap, future::Future, num::NonZeroU32, pin::Pin, sync::Arc, task::Poll,
+    time::Duration,
+};
 
 use anyhow::{bail, Context, Result};
 use bytes::Bytes;
@@ -8,6 +11,7 @@ use iroh_base::NodeId;
 use iroh_metrics::{inc, inc_by};
 use n0_future::{FutureExt, Sink, SinkExt, Stream, StreamExt};
 use rand::Rng;
+use time::{Date, OffsetDateTime};
 use tokio::{
     sync::mpsc::{self, error::TrySendError},
     time::MissedTickBehavior,
@@ -105,6 +109,7 @@ impl Client {
             node_id,
             connection_id,
             clients: clients.clone(),
+            client_counter: ClientCounter::default(),
             ping_tracker: PingTracker::default(),
         };
 
@@ -205,6 +210,8 @@ struct Actor {
     connection_id: u64,
     /// Reference to the other connected clients.
     clients: Clients,
+    /// Statistics about the connected clients
+    client_counter: ClientCounter,
     ping_tracker: PingTracker,
 }
 
@@ -214,6 +221,8 @@ impl Actor {
         // connection is accepted long before this in the HTTP server, but it is clearer to
         // handle the metric here.
         inc!(Metrics, accepts);
+        let nc = self.client_counter.update(self.node_id);
+        inc_by!(Metrics, unique_client_keys, nc);
         match self.run_inner(done).await {
             Err(e) => {
                 warn!("actor errored {e:#?}, exiting");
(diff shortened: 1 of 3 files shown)


iroh-relay/src/server/client.rs
```diff
@@ -557,6 +566,42 @@ impl Sink<Frame> for RateLimitedRelayedStream {
...
+    fn check_and_clear(&mut self) {
+        let today = OffsetDateTime::now_utc().date();
+        if today != self.last_clear_date {
+            self.clients.clear();
+            self.last_clear_date = today;
+        }
+    }
+
+    /// Updates the client counter.
+    fn update(&mut self, client: NodeId) -> u64 {
```
dignifiedquire, line 596: maybe return a bool and do sth like 

```rust
if self.client_counter.update(self.node_id) {
  inc!(Metrics, unique_client_keys, nc);
}
```

iroh-relay/src/server/client.rs
```diff
@@ -557,6 +567,41 @@ impl Sink<Frame> for RateLimitedRelayedStream {
     }
 }
 
+/// Counts how many `NodeId`s seen, how many times.
+/// Gets reset every day.
+#[derive(Debug)]
+struct ClientCounter {
+    clients: HashMap<NodeId, usize>,
```
flub, line 574: So `usize` is how many times this has `NodeId` has been seen?  Is that a useful thing?  But also... you don't ever increase the value of it past `0`?  I'm a bit confused.

dignifiedquire (approved):

iroh-relay/src/server/client.rs
```diff
@@ -557,6 +567,41 @@ impl Sink<Frame> for RateLimitedRelayedStream {
...
+            self.last_clear_date = today;
+        }
+    }
+
+    /// Updates the client counter.
+    fn update(&mut self, client: NodeId) -> bool {
+        self.check_and_clear();
+        let new_conn = !self.clients.contains_key(&client);
+        let _ = self.clients.entry(client).or_insert(0);
+        new_conn
```
flub, line 601: You can probably do this while only accessing the map once:
```suggestion
        let count = self.clients.entry(client).or_insert(0);
        count == 0
```

I think that's the equivalent of what you had.  But as my other comment says, I'm not sure that's what you want or what you meant.

dignifiedquire:
fix for cargo deny is here: https://github.com/n0-computer/iroh/pull/3199

iroh-relay/src/server/client.rs
Arqu: Good point, this is a relic from the old counter. Simplified it to a `HashSet`.

iroh-relay/src/server/client.rs
```diff
@@ -557,6 +567,41 @@ impl Sink<Frame> for RateLimitedRelayedStream {
     }
 }
 
+/// Counts how many `NodeId`s seen, how many times.
+/// Gets reset every day.
```
flub, line 571: Guess this isn't correct anymore :)

iroh-relay/src/server/client.rs
```diff
@@ -557,6 +567,41 @@ impl Sink<Frame> for RateLimitedRelayedStream {
...
+            self.last_clear_date = today;
+        }
+    }
+
+    /// Updates the client counter.
+    fn update(&mut self, client: NodeId) -> bool {
+        self.check_and_clear();
+        let new_conn = !self.clients.contains(&client);
+        self.clients.insert(client);
+        new_conn
```
flub, line 601: You're again doing two set lookups where only one is needed I think:
```suggestion
        self.clients.insert(client);
```

Also, documenting the return value wouldn't do any harm :)

iroh-relay/src/server/client.rs
flub:
```suggestion
/// Tracks how many unique nodes have been seen during the last day.
```

iroh-relay/src/server/client.rs
```diff
@@ -557,6 +567,41 @@ impl Sink<Frame> for RateLimitedRelayedStream {
...
+impl ClientCounter {
+    fn check_and_clear(&mut self) {
+        let today = OffsetDateTime::now_utc().date();
+        if today != self.last_clear_date {
+            self.clients.clear();
+            self.last_clear_date = today;
+        }
+    }
+
+    /// Updates the client counter.
```
flub, line 596:
```suggestion
    /// Marks this node as seen, returns whether it is new today or not.
```

flub (commented):
I'm just adding everything as suggestions to speed this PR up.

flub (approved):
Fixed up the cargo.lock file

PR #3202, opened 2025-02-25
fix(iroh-base): Remove unused `getrandom` dependency

matheus23:
## Description

Removes `getrandom` from `iroh-base`'s dependencies. It's unused.

## Notes & open questions

Will this remove a feature flag called `getrandom` and thus be a breaking change? Not sure.

## Change checklist

- [x] Self-review.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2295,7 +2295,6 @@ dependencies = [
  "data-encoding",
  "derive_more",
  "ed25519-dalek",
- "getrandom 0.2.15",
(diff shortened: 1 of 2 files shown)


flub:
> Will this remove a feature flag called `getrandom` and thus be a breaking change? Not sure.

Lol, yes.  The semvers check confirms this.  That wasn't the best design decision ever...

Just list it as breaking change, probably no one will be harmed by this.

PR #3204, opened 2025-02-25
fix(relay)!: change default cert format from der to pem

Arqu:
## Description

PEM format is probably the more used option for this use case so switching around defaults. ~~Also adding `cert_format` so you can specify `DER` as an alternative.~~

## Breaking Changes

Changes the default cert format for `Reloading` certificate mode from `DER` to `PEM`

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-relay/src/main.rs
+++ b/iroh-relay/src/main.rs
@@ -309,13 +309,6 @@ struct TlsConfig {
     ///
     /// Defaults to the servers' current working directory.
     cert_dir: Option<PathBuf>,
-    /// The format of the certificate file.
-    ///
-    /// Defaults to `PEM`.
-    ///
-    /// Only used when `cert_mode` is `Reloading`.
-    #[cfg(feature = "server")]
-    cert_format: Option<relay::CertFormat>,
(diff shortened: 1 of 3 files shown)


flub (commented):
- Why does reloading need to choose the cert format while manual does not?  Why can't they both use the same format and be done because no one seems to have asked for the other format?  And if it does need to be configurable, why only for one of them?
- The `cfg` is rather suspicious.  All of main.rs is server-only surely.  Surely we can figure out a way to make this work correctly without having to cfg-spaghetti the code.

Arqu:
> * Why does reloading need to choose the cert format while manual does not?  Why can't they both use the same format and be done because no one seems to have asked for the other format?  And if it does need to be configurable, why only for one of them?
> 
>     * The `cfg` is rather suspicious.  All of main.rs is server-only surely.  Surely we can figure out a way to make this work correctly without having to cfg-spaghetti the code.

I guess I just went with providing both since the original was not satisfactory. Moved everything to PEM by default. If the need arises I'll bring back DER and add it to both manual and reloading.

PR #3217, opened 2025-03-06
feat(iroh,iroh-relay): Enable proxying and test-utils support for websockets, allow configuring websockets in `endpoint::Builder`

matheus23:
## Description

- Make connecting to relays via websockets the default code path
- Enable using HTTPS proxies when dialing a relay with websockets
- Abstract away the proxied TLS stream establishment in `MaybeTlsStreamBuilder` (rename `connect_relay.rs` to `tls.rs`)
- Enable reusing `MaybeTlsStream` by making it generic over the plain stream (can be either `TcpStream` or `ProxyStream` now)
- Switch from `tokio-tungstenite` to `tokio-websockets` (has `bytes::Bytes` support, great maintenance & performance)
- Switch from `tokio-tungstenite-wasm` to `ws_stream_wasm` (no need to pay for tungstenite dependency)
- Added `iroh::endpoint::Builder::relay_conn_protocol` that allows setting the relay protocol to websockets
- Re-exporting `iroh_relay::http::Protocol` as `iroh::RelayProtocol`

## Breaking Changes

- iroh-relay: Minor change in the `From` impls for `ConnSendError` due to changing the underlying library

## Notes & open questions

We're not switching the default to websockets yet, the reason being we want to use the opportunity of that switch to update to a new relay protocol.

## Change checklist

- [x] Self-review.


iroh-relay/src/client/conn.rs
```diff
@@ -33,11 +35,20 @@ pub enum ConnSendError {
...
 
-impl From<tokio_tungstenite_wasm::Error> for ConnSendError {
-    fn from(source: tokio_tungstenite_wasm::Error) -> Self {
-        let io_err = match source {
-            tokio_tungstenite_wasm::Error::Io(io_err) => io_err,
-            _ => std::io::Error::new(std::io::ErrorKind::Other, source.to_string()),
+#[cfg(wasm_browser)]
+impl From<ws_stream_wasm::WsErr> for ConnSendError {
+    fn from(source: ws_stream_wasm::WsErr) -> Self {
+        let io_err = std::io::Error::new(std::io::ErrorKind::Other, source.to_string());
```
flub, line 41: hum, before most IO errors were real IO errors, now they're all "Other" errors?  I don't think this is very important, but it is unfortunate.  Unless of course before they were always "Other" as well and the code was just misleading us...

matheus23: Turns out `tokio-tungstenite-wasm`, which we were using before, never actually constructed `tokio_tungstenite_wasm::Error::Io`, so we also didn't have any other `io::ErrorKinds` apart from `Other` in browsers before.

I've pushed a commit that *tries* to map `ws_stream_wasm::WsErr`s to `io::Error`s properly.

PR #3221, opened 2025-03-09
feat(Connecting): add remote_node_id method to Connecting

TotalKrill:
## Description

add remote_node_id method to Connecting to read remote id before Connecting is turned into a Connection, which can be useful if one wants to create a Router or drop connections that are not in a predetermined list of allowed nodes. 

## Breaking Changes

None

### DIFF
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -1510,6 +1510,11 @@ impl Connecting {
             Err(_) => bail!("unknown handshake type"),
         }
     }
+
+    /// The remote_node_id, this should always be available after a successfull call to ['Connecting::handshake_data'] or ['Connecting::alpn']
+    pub fn remote_node_id(&self) -> Option<NodeId> {
+        self.remote_node_id
+    }
 }
 
 impl Future for Connecting {


iroh/src/endpoint.rs
```diff
@@ -1510,6 +1510,11 @@ impl Connecting {
             Err(_) => bail!("unknown handshake type"),
         }
     }
+
+    /// The remote_node_id, this should always be available after a successfull call to ['Connecting::handshake_data'] or ['Connecting::alpn']
```
flub, line 1514: Is this really true?  Looking at the code I think this is populated when the `Connecting` is created via `Endpoint::connect` but not at all when created via `Endpoint::accept`.

Looking at `Connection::remote_node_id` it seems you need the peer identity data to be able to get the remote node id.  I think you have to dig deeper into rustls to get the nodeid out at this stage.  It would be very cool if you could figure out how to do that though!

I think you should also write a unittest for this.

TotalKrill:
Turns out I was a bit hasty, and the application code I was toying around with was kinda, buggy and not running a codepath I thought it was. 

Confidently writing a unit-test and seeing it fail is always such a good reminder on why one should start with that :D

PR #3227, opened 2025-03-12
chore: switch from `backoff` to `backon`

ramfox:
## Description

`backoff` is no longer maintained, upgrade path in cargo-deny indicates switching to `backon`.

## Notes & open questions

There are some options in `backon` for using sleepers in the retry that are Wasm compatible: https://docs.rs/backon/1.4.0/backon/#sleep

I'm curious if we should be taking advantage.

Further questions:
~how do I resolve the `paste` cargo-deny issue?~
I don't: added `paste` and `humantime` advisories to the "allow" list

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh/src/magicsock/relay_actor.rs
+++ b/iroh/src/magicsock/relay_actor.rs
@@ -429,7 +429,7 @@ impl ActiveRelayActor {
             }
         };
 
-        // We implement our own `Sleeper` here, so that we can use the `backoff`
+        // We implement our own `Sleeper` here, so that we can use the `backon`
         // crate with our own implementation of `time::sleep` (from `n0_future`)
         // that works in browsers.
         struct Sleeper;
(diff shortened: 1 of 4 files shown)


iroh/src/magicsock/relay_actor.rs
```diff
@@ -433,25 +434,15 @@ impl ActiveRelayActor {
         // that works in browsers.
         struct Sleeper;
 
-        impl backoff::future::Sleeper for Sleeper {
+        impl backon::Sleeper for Sleeper {
```
matheus23, line 437: Needs a comment update above (backoff -> backon)

matheus23 (approved):
Thanks for this. `backon` even has a better API than `backoff` IMO.

Also this makes me consider moving the retry logic into the `iroh_relay::ClientBuilder`.

flub (approved):
Goodness, why do we have these unmaintained advisories?  Unmaintained crates are great!  They don't update all the time.  Why can't we have finished software?

PR #3230, opened 2025-03-14
fix(iroh-relay): Report round-trip-latency instead of single-trip for QAD

### LINKED ISSUE #3225: bug: STUN and QAD probes do not report the same latency
**Describe the bug**

If you look the logs of an `iroh-doctor` run for `ProbeReport` you'll find latencies to relay servers for different probes and their latencies.  You'll notice that the STUN and QAD probes disagree what the latency is, with QAD probes being about half the time of the corresponding STUN probes. 

**Relevant Logs**

StunV6 to euw: 12ms
QuicV6 to euw: 5ms
StunV4 to euw: 11ms
QuicV4 to euw: 8ms
Stunv4 to use: 99ms
QuicV4 to use: 50ms
StunV6 to aps: 244ms
QuicV6 to aps: 121ms
Stunv4 to aps: 246ms
Quicv4 to aps: 122ms

**Expected behavior**

The latencies should be much closer to each other.

matheus23:
## Description

Switches QAD latency to be round-trip instead of single-trip to match STUN latencies.

## Notes & open questions

Fixes #3225

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh-relay/src/quic.rs
+++ b/iroh-relay/src/quic.rs
@@ -273,7 +273,7 @@ impl QuicClient {
         // if we've sent to an ipv4 address, but received an observed address
         // that is ivp6 then the address is an [IPv4-Mapped IPv6 Addresses](https://doc.rust-lang.org/beta/std/net/struct.Ipv6Addr.html#ipv4-mapped-ipv6-addresses)
         observed_addr = SocketAddr::new(observed_addr.ip().to_canonical(), observed_addr.port());
-        let latency = conn.rtt() / 2;
+        let latency = conn.rtt();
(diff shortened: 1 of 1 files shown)


flub (approved):

matheus23:
We looked into renaming things from `latency` to `rtt`/"round trip time" generally, as latency sometimes refers to single-trip time and sometimes round-trip-time.

However, renaming would end up fairly invasive: E.g. `iroh_net_report`'s `RelayLatencies` would need to be renamed, and all struct fields, and a lot more. I've started that locally, but it's quite a bit of stuff/invasive.

PR #3234, opened 2025-03-17
fix(iroh): Don't cause re-stuns all the time in browsers

### LINKED ISSUE #3232: bug: browser wasm: extreme ping HTTP request spam, 3 req/s
**Describe the bug**

It seems the amount of HTTP ping connections that go out to the relays is proportional to the number of peers you have in the chat demo.

Here is a pic with >2000 requests done in 15min

Image

Here is a reproduction on the chat demo with 500 request in 5min (firefox)

Image

Here is again with Brave/Chrome,  120 requests / min

Image

This can't be good for battery life or server bandwidth...

** Steps to reproduce **

1. Open chat example in Firefox browser
2. Create chat room, copy URL
3. Open 5-10 new browser tabs, all join same chat room
4. Open "Dev Tools" and look at Network tab
5. 100s of PING requests per minute

**Relevant Logs**

https://gist.github.com/johnny-smitherson/eac09f37718957911ad5cb14061069a1

Logs directly from chat demo page.

A lot of "need full ping"

**Expected behavior**

Amount of PING requests per minute stays the same (low number), regardless of how many peers you have in the chat.

**Platform(s)**
Browsers checked: Firefox 136 +  brave 1.68 / windows 11

**Additional Context / Screenshots / GIFs**

There seems to be one ping happening per message traffic. But if you have 5 open browser tabs, then you have a lot of "presence" messages (about 1-2/s), which seems to cause the same amount of pings.

matheus23:
## Description

- Always return `false` in `want_call_me_maybe` in browsers, as we never actually want to send call me maybes, since there's no direct addresses we can put into these messages.
- Fix the shutdown logging in magicsock. Redundant warning were printed on graceful shutdown, and one was skipped in non-graceful shutdown.

## Notes & open questions

Closes #3232

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh/src/magicsock.rs
+++ b/iroh/src/magicsock.rs
@@ -1913,11 +1913,16 @@ impl Handle {
             }
         })
         .await;
-        if shutdown_done.is_ok() {
-            warn!("tasks shutdown complete");
-            // shutdown all tasks
-            warn!("aborting remaining {}/3 tasks", tasks.len());
-            tasks.shutdown().await;
+        match shutdown_done {
+            Ok(_) => trace!("tasks finished in time, shutdown complete"),
+            Err(_elapsed) => {
+                // shutdown all tasks
+                warn!(
+                    "tasks didn't finish in time, aborting remaining {}/3 tasks",
+                    tasks.len()
+                );
+                tasks.shutdown().await;
+            }
         }
         trace!("magicsock closed");
     }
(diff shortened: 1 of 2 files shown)


matheus23:
Confirmed this fixes issues for who submitted #3232 and when testing personally.

flub (approved):

PR #3245, opened 2025-03-26
fix(iroh-relay): write default config if config path is set but does not exist

Frando:
## Description

In the `iroh-relay` binary, the helptext says that the default config will be written if `--config` is set but the specified file does not exist. However, we didn't actually do this. This PR changes this so that the default config is indeed written, which is useful because then this file can be adjusted as needed without having to grab everything from the config struct definition.

## Change checklist

- [x] Sef-review.

### DIFF
--- a/iroh-relay/src/main.rs
+++ b/iroh-relay/src/main.rs
@@ -22,7 +22,7 @@ use iroh_relay::{
 use n0_future::FutureExt;
 use serde::{Deserialize, Serialize};
 use tokio_rustls_acme::{caches::DirCache, AcmeConfig};
-use tracing::debug;
+use tracing::{debug, info};
 use tracing_subscriber::{prelude::*, EnvFilter};
 
 /// The default `http_bind_port` when using `--dev`.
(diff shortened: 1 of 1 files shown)


dignifiedquire (approved):

flub (changes requested):
Please, no.  We removed this on purpose.

flub:
Binaries shouldn't write their own config, it is very bad behaviour to write a file in a hidden location somewhere without telling you.  This e.g. messes around with containerised deploys and setups that use read-only filesystems.  It also makes things much more complicated on upgrades.

Once the relay server removed the need to save its own private key we finally removed this strange behaviour.  Instead the docs should be adjusted to not say the default config is written.

The fact you need to read the config structs to to know the file format is indeed a bit unfortunate.  But again something we should fix with documentation, we just also need to make sure the documentation remains correct by testing the documented config snippets.

Frando:
Closing this in response to @flub's reasoning, makes sense to me. We should fix the documentation still. Linking to issue: https://github.com/n0-computer/iroh/issues/3258

PR #3246, opened 2025-03-26
feat(iroh-relay): allow to authenticate nodes via a HTTP POST request

Frando:
## Description

This adds a new `access` config option to the `iroh-relay` binary: If set to `http`, a `POST` request will be performed for each node that attempts to connect to the relay. The URL to request is set in the config, and a `{node_id}` token in the configured URL will be replaced with the hex-encoded node id to check access for. If and only if the response has a 2xx status code, and a response text of `true`, the connecting node will be granted access. In all other cases, the node will be denied access.

Config example to use this:
```toml
enable_relay = true
enable_stun = true
enable_quic_addr_discovery = false
enable_metrics = false

[access]
http = "http://localhost:8000/api/relays/check-auth?user=frando&node_id={node_id}"
```

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


iroh-relay/src/main.rs
```diff
@@ -215,10 +223,59 @@ impl From<AccessConfig> for iroh_relay::server::AccessConfig {
...
 }
 
+#[tracing::instrument("http-access-check", skip_all, fields(node_id=%node_id.fmt_short()))]
+async fn http_access_check(
+    client: &reqwest::Client,
+    url_template: &str,
+    node_id: NodeId,
+) -> iroh_relay::server::Access {
+    use iroh_relay::server::Access;
+    let url = url_template.replace("{node_id}", &node_id.to_string());
```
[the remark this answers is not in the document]
Frando: Yeah, those would work as well. Also no strong preference on my side.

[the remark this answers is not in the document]
flub: This string interpolation might be a bit brittle?  I kind of like the idea of some new headers for custom info.  It also means you don't have to change the config if we in the future decide to add things like the IP address or whatever as you can just add more headers.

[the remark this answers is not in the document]
Frando: Changed it:
* no more URL string replacement
* The node id is passed in an `X-Iroh-NodeId` header

iroh-relay/src/main.rs
```diff
@@ -215,10 +225,62 @@ impl From<AccessConfig> for iroh_relay::server::AccessConfig {
...
+) -> iroh_relay::server::Access {
+    use iroh_relay::server::Access;
+    debug!(%url, "Check relay access via HTTP POST");
+    let res = match client
+        .post(url)
+        .header("X-Iroh-NodeId", node_id.to_string())
+        .send()
+        .await
+    {
+        Ok(t) => t,
```
flub, line 254: You could use the fany(ish) new(ish) let-else expression:

`let Ok(res) = client.post...await else { warn!(...); return Access::Deny; }`?

iroh-relay/src/main.rs
```diff
@@ -181,6 +183,14 @@ enum AccessConfig {
     Allowlist(Vec<NodeId>),
     /// Allows everyone, except these nodes.
     Denylist(Vec<NodeId>),
+    /// Performs a HTTP POST request to determine access for each node that connects to the relay.
+    ///
+    /// The request will have a header `X-Iroh-Node-Id` set to the hex-encoded node id attempting
+    /// to connect to the relay.
+    ///
+    /// To grant access, the HTTP endpoint must return a `200` response with `true` as the response text.
```
flub, line 191: What's the motivation to need a `true` in the body?

flub (approved):

iroh-relay/src/main.rs
Frando: With `let-else` you can't print the error in the warn message, which I think is helpful, no?

iroh-relay/src/main.rs
Frando: I thought of this as a safeguard, so that the endpoint implementing this would have to be more explicit. Maybe it's not needed, no strong opinion here.

flub:
> > What happens if the authentication server needs to authenticate the clients connecting to it?
> 
> The authentication server has the node id, and can do whatever it likes to authenticate it (i.e. check against an ACL or against a capability token which it previously received over another communication channel).
> 
> Or did I misunderstand your question?

Yeah, sorry.  Didn't ask that very well.  What I was wondering is whether the authentication service wants to protect itself.  You don't want it to become a NodeId oracle but only want it to be used by authorised relay servers, and right now you can only do that by deploying it on a private network.  Or maybe you could put a secret in the URL, but that secret would then have to be in the relay server's config file.

So maybe in the future we'll have to add features like this, and how compatible would that be to the current configuration format is what I was wondering.  But it's probably fine and we can deal with it when that happens.

Frando:
>  What I was wondering is whether the authentication service wants to protect itself. You don't want it to become a NodeId oracle but only want it to be used by authorised relay servers, and right now you can only do that by deploying it on a private network. Or maybe you could put a secret in the URL, but that secret would then have to be in the relay server's config file.

Ah, now I'm understanding. This is a good point. A secret in the URL would work. What other methods would work well? Some ideas:

1) Allow to set an `Authentication` header via either a config value, or an environment variable `IROH_RELAY_HTTP_AUTH_HEADER` or similar
2) Allow to set arbitrary headers via the config file

I think a first safeguard would be to make the `http` config a map instead of a single value, so that more options can be added without breaking the config format. Let me add this.

flub:
> A secret in the URL would work. What other methods would work well? Some ideas:
> 
>     1. Allow to set an `Authentication` header via either a config value, or an environment variable `IROH_RELAY_HTTP_AUTH_HEADER` or similar
> 
>     2. Allow to set arbitrary headers via the config file
> 
> 
> I think a first safeguard would be to make the `http` config a map instead of a single value, so that more options can be added without breaking the config format. Let me add this.

Making `http` a map is a great future-proofing response indeed.  I wouldn't go any further than that for now.

(And 1. above sounds the most appealing to me in the future, but let's see)

Frando:
I added support for bearer tokens too.

iroh-relay/src/main.rs
```diff
@@ -181,6 +187,27 @@ enum AccessConfig {
...
+    /// In all other cases, the node will be denied access.
+    Http(HttpAccessConfig),
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+struct HttpAccessConfig {
```
flub, line 201: I'm always nervous whether serde manages to de-serialise from the config file format correctly.  I've seen cases where serde expressed things that could not be expressed in the target serialisation.  Mind adding a de-serialisation test for this from toml?  (We don't need serialise support on these anymore, they're historic.)

There are a few similar tests already, so should be easy enough to copy.

flub (approved):

dignifiedquire (approved):

PR #3249, opened 2025-03-27
fix(iroh): Reap ActiveRelayActor handles for idle relays

flub:
## Description

The ActiveRelayActor exists when it is idle for a while (60s for now).
When it does the RelayActor is supposed to notice the tasks exists and
clean up the handles it keeps to all the ActiveRelayActors.
Unfortunately it was only doing this for ActiveRelayActors which
errored and not for those that exited normally because they were idle.

This makes sure to also reap active relay handles when the actor
exists normally.

Additionally this stops propagating the panic.  The relay actor itself
is currently spawned without keeping track of the handle.  Panicking
it would make it disappear silently.  Propagating panics is
reasonable, but I don't want to fix the task spawning.  That's for
another time.

## Notes & open questions

I looked at writing a test for this, but testing the RelayActor is
still something that's difficult to do.  So sadly no tests.

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh/src/magicsock/relay_actor.rs
+++ b/iroh/src/magicsock/relay_actor.rs
@@ -807,15 +807,19 @@ impl RelayActor {
             tokio::select! {
                 biased;
                 _ = self.cancel_token.cancelled() => {
-                    trace!("shutting down");
+                    debug!("shutting down");
                     break;
                 }
-                Some(Err(err)) = self.active_relay_tasks.join_next() => {
-                    if err.is_panic() {
-                        panic!("ActiveRelayActor task panicked: {err:#?}");
-                    }
-                    if !err.is_cancelled() {
-                        error!("ActiveRelayActor failed: {err:?}");
+                Some(res) = self.active_relay_tasks.join_next() => {
+                    match res {
+                        Ok(()) => (),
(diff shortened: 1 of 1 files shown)


iroh/src/magicsock/relay_actor.rs
```diff
@@ -807,15 +807,19 @@ impl RelayActor {
...
+                Some(res) = self.active_relay_tasks.join_next() => {
+                    match res {
+                        Ok(()) => (),
+                        Err(err) if err.is_panic() => {
+                            error!("ActiveRelayActor task panicked: {err:#?}");
+                        }
+                        Err(err) if err.is_cancelled() => {
+                            error!("ActiveRelayActor cancelled: {err:#?}");
+                        }
+                        Err(err) => error!("ActiveRelayActor failed: {err:#?}"),
```
flub, line 822: Look, aren't real rust matching semantics nice?  The compiler told me I had a case that wasn't covered (ok, ok.  The compiler is actually wrong because task errors are always either panics or cancelled, tokio doesn't allow anything else.  But still, the compiler is right.)

ramfox (approved):
oooo nice and insidious 🙌

PR #3251, opened 2025-03-31
feat(iroh)!: move `iroh-net-report` back into `iroh`

ramfox:
## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


iroh/src/lib.rs
```diff
@@ -249,6 +249,7 @@ pub mod discovery;
 pub mod dns;
 pub mod endpoint;
 pub mod metrics;
+pub mod net_report;
```
flub, line 252: It'd be nice to hide this behind a `unstable_api` cargo feature (or whatever bikeshed colour you like)?  Apparently @dignifiedquire had opinions on something like this in the past though, I don't know if favour or against such constructs.  Or maybe keep it entirely internal, as the API for doctor will probably become something else now according to the discussion on discourse?

flub (commented):
I think this is fine, but I'd like to not expose this as a public API.  I can't remember if we discussed a mechanism before so that doctor can still use it, I think we did?  I think it was by means of a cargo feature?  But I'm not sure.

iroh/src/lib.rs
dignifiedquire: lets merge this as is, and get the API reduction surface before the next release in as well would be my suggestion

dignifiedquire (approved):

iroh/src/lib.rs
ramfox: After our planning meeting, we decided I'm going to reduce the API surface to just what is necessary for `doctor` and get this merged by Friday.

PR #3255, opened 2025-04-03
feat: fix minimal crates selection

dignifiedquire:
Depends on

- [x] https://github.com/n0-computer/net-tools/pull/19
- [x] https://github.com/n0-computer/n0-future/pull/3


Cargo.toml
```diff
@@ -41,3 +41,8 @@ unexpected_cfgs = { level = "warn", check-cfg = ["cfg(iroh_docsrs)", "cfg(iroh_l
 
 [workspace.lints.clippy]
 unused-async = "warn"
+
+[patch.crates-io]
+n0-future = { git = "https://github.com/n0-computer/n0-future", branch = "minimal-versions" }
+portmapper = { git = "https://github.com/n0-computer/net-tools", branch = "minimal-versions" }
+netwatch = { git = "https://github.com/n0-computer/net-tools", branch = "minimal-versions" }
```
flub, line 48: I know this is blocked on these, but just adding a comment to not forget.

matheus23:
@dignifiedquire the Wasm failure is due to `stun-rs` depending (indirectly) on getrandom version 0.3 starting with version 0.1.9.
In the PR, you've configured 0.1.11 as the stun-rs version, this causes us to suddenly pull in getrandom 0.3 in iroh-relay, which we didn't do before.
Given that rand 0.9 dependencies are only going to get more common in the future (and in our transitive dependencies), I think there's no way around burdening our users with getrandom 0.3.
So we need to add a `getrandom = { version = "0.3.0", features = ["wasm_js"] }` wasm-gated dependency to iroh-relay.

iroh/Cargo.toml
```diff
@@ -59,12 +59,12 @@ reqwest = { version = "0.12", default-features = false, features = [
...
 rustls = { version = "0.23", default-features = false, features = ["ring"] }
-serde = { version = "1", features = ["derive", "rc"] }
+serde = { version = "1.0.219", features = ["derive", "rc"] }
 smallvec = "1.11.1"
 strum = { version = "0.26", features = ["derive"] }
-stun-rs = "0.1.5"
+stun-rs = "0.1.11"
 thiserror = "2"
-tokio = { version = "1", features = [
+tokio = { version = "1.44.1", features = [
```
rklaehn, line 67: That's pretty recent. On the plus side it allows using https://docs.rs/tokio/latest/tokio/sync/oneshot/struct.Receiver.html#method.is_terminated I think...

ramfox (approved):

PR #3262, opened 2025-04-08
feat!: metrics refactor

Frando:
## Description

Depends on https://github.com/n0-computer/net-tools/pull/20

Until now, we were using a superglobal static `iroh_metrics::core::Core` struct to collect metrics into. This allowed us to use macros to track metrics from anywhere in the codebase. However, this also made it impossible to collect metrics *per endpoint*, which is what you want usually as soon as you have more than one endpoint in your app.

This PR builds on n0-computer/iroh-metrics#15, n0-computer/iroh-metrics#22, and n0-computer/iroh-metrics#23. It removes the global metrics collection from all crates in the iroh repository. Instead, we now create and pass metrics collector structs to all places where we need to collect metrics.

This PR disables the `static_core` feature from iroh-metrics, which means the macros for superglobal metrics collection are not available anymore. This is good, because otherwise we could easily miss metrics not tracked onto the proper metrics collector.

This PR also updates iroh-dns-server and iroh-relay to use manual metrics collection.

While this means that we have to pass our metrics structs to more places, it also makes metrics collection more visible, and we can now also split the metrics structs further easily if we want to separate concerns more.

This PR should not change anything apart from metrics collection. Most places are straightforward conversions from the macros to methods on the metrics collectors. At a few places, logic was changed slightly to move metrics collection a layer up to save a few clones.

## Breaking Changes

* All metrics structs (`iroh::metrics::{MagicsockMetrics, PortmapMetrics, NetReportMetrics}`) now implement `MetricsGroup` from the new version `0.34` of `iroh-metrics` and no longer implement traits from `iroh-metrics@0.32`. 
* Metrics are no longer registered onto the static superglobal `Core`. `iroh` does not use `static_core` feature of `iroh-metrics`. Metrics are now exposed from the subsystems that track them, see e.g. `Endpoint::metrics`. 

Several methods now take a `Metrics` argument. You can always pass `Default::default` if you don't want to unify metrics tracking with other sections.

#### `iroh`

* `iroh::metrics::{MagicsockMetrics, NetReportMetrics, PortmapMetrics}` all are now marked `non_exhaustive`, and implement `iroh_metrics::MetricsGroup` from `iroh-metrics@0.34`  and no longer implement `iroh_metrics::Metric` from `iroh-metrics@0.32`. They also no longer implement `Clone` (put them into an `Arc` for cloning instead).

* `iroh::net_report::Client::new` now takes `iroh::net_report::metrics::Metrics` as forth argument

#### `iroh-dns-server`
*  `iroh_dns_server::server::Server::spawn` now takes `Metrics` as third argument
*  `iroh_dns_server::ZoneStore::persistent` now takes  `Metrics` as third argument
*  `iroh_dns_server::ZoneStore::in_memory` now takes  `Metrics` as third argument
*  `iroh_dns_server::ZoneStore::new` now takes  `Metrics` as third argument
* `iroh_dns_server::state::AppState` now has a public `metrics: Metrics` field
*  `iroh_dns_server::dns::DnsHandler::new` now takes  `Metrics` as third argument
* function `iroh_dns_server::metrics::init_metrics` is removed

#### `iroh-relay`

* `iroh_relay::metrics::{StunMetrics, Metrics}` all are now marked `non_exhaustive`, and implement `iroh_metrics::MetricsGroup` from `iroh-metrics@0.34`  and no longer implement `iroh_metrics::Metric` from `iroh-metrics@0.32`. They also no longer implement `Clone` (put them into an `Arc` for cloning instead).

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.
  - [x] Open an issue or PR on any number0 repos that are affected by this breaking change. Give guidance on how the updates should be handled or do the actual updates themselves.
    - [x] [`iroh-gossip`](https://github.com/n0-computer/iroh-gossip)
      - https://github.com/n0-computer/iroh-gossip/pull/58
    - [x] [`iroh-blobs`](https://github.com/n0-computer/iroh-blobs)
      - https://github.com/n0-computer/iroh-blobs/pull/85
    - [x] [`iroh-docs`](https://github.com/n0-computer/iroh-docs)
      - https://github.com/n0-computer/iroh-docs/pull/41

### DIFF
--- a/iroh-relay/src/server/http_server.rs
+++ b/iroh-relay/src/server/http_server.rs
@@ -13,7 +13,6 @@ use hyper::{
     upgrade::Upgraded,
     HeaderMap, Method, Request, Response, StatusCode,
 };
-use iroh_metrics::inc;
 use n0_future::{FutureExt, SinkExt};
 use tokio::net::{TcpListener, TcpStream};
 use tokio_rustls_acme::AcmeAcceptor;
@@ -176,11 +175,12 @@ pub(super) struct ServerBuilder {
     key_cache_capacity: usize,
     /// Access config for nodes.
     access: AccessConfig,
+    metrics: Arc<Metrics>,
 }
 
 impl ServerBuilder {
     /// Creates a new [ServerBuilder].
-    pub(super) fn new(addr: SocketAddr) -> Self {
+    pub(super) fn new(addr: SocketAddr, metrics: Arc<Metrics>) -> Self {
         Self {
             addr,
             tls_config: None,
@@ -189,6 +189,7 @@ impl ServerBuilder {
             client_rx_ratelimit: None,
             key_cache_capacity: DEFAULT_KEY_CACHE_CAPACITY,
             access: AccessConfig::Everyone,
+            metrics,
         }
     }
 
@@ -248,6 +249,7 @@ impl ServerBuilder {
             self.client_rx_ratelimit,
             KeyCache::new(self.key_cache_capacity),
             self.access,
+            self.metrics,
         );
 
         let addr = self.addr;
@@ -327,6 +329,7 @@ struct Inner {
     rate_limit: Option<ClientRateLimit>,
     key_cache: KeyCache,
     access: AccessConfig,
+    metrics: Arc<Metrics>,
 }
 
 impl RelayService {
@@ -517,11 +520,11 @@ impl Inner {
         trace!(?protocol, "accept: start");
         let mut io = match protocol {
             Protocol::Relay => {
-                inc!(Metrics, relay_accepts);
+                self.metrics.relay_accepts.inc();
                 RelayedStream::Relay(Framed::new(io, RelayCodec::new(self.key_cache.clone())))
             }
             Protocol::Websocket => {
-                inc!(Metrics, websocket_accepts);
+                self.metrics.websocket_accepts.inc();
                 // Since we already did the HTTP upgrade in the previous step,
                 // we use tokio-websockets to handle this connection
                 // Create a server builder with default config
@@ -569,7 +572,9 @@ impl Inner {
 
         // build and register client, starting up read & write loops for the client
         // connection
-        self.clients.register(client_conn_builder).await;
+        self.clients
+            .register(client_conn_builder, self.metrics.clone())
+            .await;
         Ok(())
     }
 }
@@ -591,6 +596,7 @@ impl RelayService {
         rate_limit: Option<ClientRateLimit>,
         key_cache: KeyCache,
         access: AccessConfig,
+        metrics: Arc<Metrics>,
     ) -> Self {
         Self(Arc::new(Inner {
             handlers,
@@ -600,6 +606,7 @@ impl RelayService {
             rate_limit,
             key_cache,
             access,
+            metrics,
         }))
     }
 
@@ -761,7 +768,7 @@ mod tests {
         let b_key = SecretKey::generate(rand::thread_rng());
 
         // start server
-        let server = ServerBuilder::new("127.0.0.1:0".parse().unwrap())
+        let server = ServerBuilder::new("127.0.0.1:0".parse().unwrap(), Default::default())
             .spawn()
             .await?;
 
@@ -869,7 +876,7 @@ mod tests {
         let tls_config = make_tls_config();
 
         // start server
-        let mut server = ServerBuilder::new("127.0.0.1:0".parse().unwrap())
+        let mut server = ServerBuilder::new("127.0.0.1:0".parse().unwrap(), Default::default())
             .tls_config(Some(tls_config))
             .spawn()
             .await?;
@@ -952,6 +959,7 @@ mod tests {
             None,
             KeyCache::test(),
             AccessConfig::Everyone,
+            Default::default(),
         );
 
         info!("Create client A and connect it to the server.");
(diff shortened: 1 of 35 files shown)


iroh/src/endpoint/rtt_actor.rs
```diff
@@ -99,10 +105,10 @@ impl Stream for MappedStream {
                     if !self.was_direct_before && matches!(new_conn_type, ConnectionType::Direct(_))
                     {
                         self.was_direct_before = true;
-                        inc!(MagicsockMetrics, connection_became_direct);
+                        became_direct = true
```
flub, line 108: Hum, I don't think you want to change this logic.  You can totally become direct more than once for the same connection.

flub: I think it is best to keep the refactor faithull to the current behaviour, including all the bugs.  and do the bug fixes in another PR(s).  That will make it much easier to review.

matheus23: I don't think this changes the logic.
Nor do I think the logic had bugs. It executes the body of this `if` only once over the course of `was_direct_before`'s lifetime.

Frando: My intention was not do fix any bugs or change existing behavior, only to not clone the the metrics once more into each connection type stream, but instead handle the metrics update in the top level actor. So if I changed anything of substance, I will revert this.

iroh/src/magicsock.rs
```diff
@@ -464,14 +469,20 @@ impl MagicSock {
     /// Implementation for AsyncUdpSocket::try_send
     #[instrument(skip_all)]
     fn try_send(&self, transmit: &quinn_udp::Transmit) -> io::Result<()> {
-        inc_by!(MagicsockMetrics, send_data, transmit.contents.len() as _);
+        self.metrics
+            .magicsock
+            .send_data
+            .inc_by(transmit.contents.len() as _);
```
Frando, line 475: Duplicate line, remove

iroh-relay/src/server/http_server.rs
```diff
@@ -869,7 +876,7 @@ mod tests {
         let tls_config = make_tls_config();
 
         // start server
-        let mut server = ServerBuilder::new("127.0.0.1:0".parse().unwrap())
+        let mut server = ServerBuilder::new("127.0.0.1:0".parse().unwrap(), Default::default())
```
dignifiedquire, line 879: might be nicer to make the metrics be a call 

eg
```rust
// default
ServerBuilder::new("127.0.0.1:0".parse().unwrap())
  .spawn();
// explicit
ServerBuilder::new("127.0.0.1:0".parse().unwrap())
  .metrics(metrics)
  .spawn();
```

iroh/src/metrics.rs
```diff
@@ -1,7 +1,46 @@
...
 #[cfg(not(wasm_browser))]
 pub use portmapper::Metrics as PortmapMetrics;
 
 pub use crate::{magicsock::Metrics as MagicsockMetrics, net_report::Metrics as NetReportMetrics};
+
+/// Metrics collected by an [`crate::endpoint::Endpoint`].
+///
+/// See [`crate::endpoint::Endpoint::metrics`] for details.
+#[derive(Default, Debug, Clone)]
+pub struct EndpointMetrics {
```
dignifiedquire, line 16: we probably should mark all the metrics structs as `non_exhaustive`

iroh/src/magicsock/node_map/node_state.rs
```diff
@@ -160,10 +159,12 @@ impl NodeState {
     pub(super) fn new(id: usize, options: Options) -> Self {
         let quic_mapped_addr = NodeIdMappedAddr::generate();
 
-        if options.relay_url.is_some() {
-            // we potentially have a relay connection to the node
-            inc!(MagicsockMetrics, num_relay_conns_added);
-        }
+        // TODO(frando): I don't think we need to track the `num_relay_conns_added`
```
dignifiedquire, line 162: @matheus23 @flub ?

iroh-relay/src/server/http_server.rs
Frando: This would mean making it optional though, no? Currently metrics are required, which is why I added this on the constructor.
I can make them optional, it'll be `if let Some(metrics) = self.metrics { metrics.foo.inc(); }` then at each place where metrics are used.

iroh/src/magicsock/relay_actor.rs
```diff
@@ -285,7 +288,9 @@ impl ActiveRelayActor {
     ///
     /// Primarily switches between the dialing and connected states.
     async fn run(mut self) -> Result<()> {
-        inc!(MagicsockMetrics, num_relay_conns_added);
+        // TODO(frando): decide what this metric means, it's either wrong here or in node_state.rs.
```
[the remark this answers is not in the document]
Frando: https://github.com/n0-computer/iroh/issues/3270

iroh-relay/src/server/http_server.rs
dignifiedquire: I didn't mean to make them optional, just construct them as `default` if none are passed in

Frando: Changed that to use the builder method instead of Default in the constructor.

iroh/src/metrics.rs
Frando: Done

dignifiedquire (approved):

PR #3267, opened 2025-04-16
chore(iroh): Add `echo-no-router.rs` example

matheus23:
## Description

Adds an example showing how to implement a basic protocol without using a `Router`.
As a bonus, one can compare what `echo.rs` and `echo-no-router.rs` are doing to see the difference.

## Notes & open questions

I think this is slightly useful, but YMMV. I don't think it's a lot of work to maintain though.

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh/examples/echo-no-router.rs
+++ b/iroh/examples/echo-no-router.rs
@@ -1,0 +1,102 @@
+//! Very basic example showing how to implement a basic echo protocol,
+//! without using the `Router` API. (For the router version, check out the echo.rs example.)
+//!
+//! The echo protocol echos any data sent to it in the first stream.
+//!
+//! ## Running the Example
+//!
+//!     cargo run --example echo-no-router --features=examples
+
+use anyhow::Result;
+use iroh::{Endpoint, NodeAddr};
+
+/// Each protocol is identified by its ALPN string.
+///
+/// The ALPN, or application-layer protocol negotiation, is exchanged in the connection handshake,
+/// and the connection is aborted unless both nodes pass the same bytestring.
+const ALPN: &[u8] = b"iroh-example/echo/0";
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    let endpoint = accept_side().await?;
+    let node_addr = endpoint.node_addr().await?;
+
+    connect_side(node_addr).await?;
+
+    endpoint.close().await;
+
+    Ok(())
+}
+
+async fn connect_side(addr: NodeAddr) -> Result<()> {
+    let endpoint = Endpoint::builder().discovery_n0().bind().await?;
+
+    // Open a connection to the accepting node
+    let conn = endpoint.connect(addr, ALPN).await?;
+
+    // Open a bidirectional QUIC stream
+    let (mut send, mut recv) = conn.open_bi().await?;
+
+    // Send some data to be echoed
+    send.write_all(b"Hello, world!").await?;
+
+    // Signal the end of transfer on this particular stream.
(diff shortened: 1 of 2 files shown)


iroh/examples/echo-no-router.rs
```diff
@@ -0,0 +1,102 @@
...
+const ALPN: &[u8] = b"iroh-example/echo/0";
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    let endpoint = accept_side().await?;
+    let node_addr = endpoint.node_addr().await?;
+
+    connect_side(node_addr).await?;
+
+    endpoint.close().await;
```
flub, line 26: maybe explain that you should not skip this?

flub (commented):
I think I would go for a sub-directory called "router" and move the router examples into there.  But that might be more controversial.

Anyway, all comments rather cosmetic and I don't mind if you do something about them or not.  Apart from the connect-side not closing the endpoint.  That's rather critical.

iroh/examples/echo-no-router.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    // Signal the end of transfer on this particular stream.
+    send.finish()?;
+
+    // Receive the echo
+    let response = recv.read_to_end(1000).await?;
+    assert_eq!(&response, b"Hello, world!");
+
+    // Explicitly close the whole connection.
+    conn.close(0u32.into(), b"bye!");
```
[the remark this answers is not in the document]
matheus23: Yepyep you're right.

The other `echo.rs` example has the same issue.

iroh/examples/echo-no-router.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+async fn accept_side() -> Result<Endpoint> {
+    let endpoint = Endpoint::builder()
+        .discovery_n0()
+        .alpns(vec![ALPN.to_vec()])
+        .bind()
+        .await?;
+
+    let ep = endpoint.clone(); // Cloning so we can both return it and use it in the tokio task
```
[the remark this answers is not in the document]
matheus23: Yeah I thought about this for way too long without doing it

iroh/examples/echo-no-router.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    // spawn a task so that `accept_side` returns immediately and we can continue in main().
+    tokio::spawn(async move {
```
dignifiedquire, line 66: maybe make a mention that you should track all these tasks in a more serious environement

matheus23:
> Anyway, all comments rather cosmetic and I don't mind if you do something about them or not. Apart from the connect-side not closing the endpoint. That's rather critical.

Please re-review :)

iroh/examples/echo-no-router.rs
```diff
@@ -0,0 +1,120 @@
...
+    let response = recv.read_to_end(1000).await?;
+    assert_eq!(&response, b"Hello, world!");
+
+    // Explicitly close the whole connection.
+    conn.close(0u32.into(), b"bye!");
+
+    // The above call only queues a close message to be sent (see how it's not async!).
+    // We need to actually call this to make sure this message is sent out.
+    endpoint.close().await;
+    // If we don't call this, but continue using the endpoint, we then the queued
```
flub, line 58:
```suggestion
    // If we don't call this, but continue using the endpoint, then the queued
```

flub (approved):
Oh, good catch on also fixing the original example!

PR #3268, opened 2025-04-16
ci: update to mozilla-actions/sccache-action@v0.0.9

dignifiedquire:
This is needed because of
https://github.blog/changelog/2025-03-20-notification-of-upcoming-breaking-changes-in-github-actions/#decommissioned-cache-service-brownouts


Cargo.lock
flub, line 1: Did you mean to also do this change?

PR #3274, opened 2025-04-21
feat: update to axum@0.8

### LINKED ISSUE #3273: Can iroh dns axum be upgraded to 0.8.X?
Can iroh dns axum be upgraded to 0.8.X?

Now many axum related libraries have been upgraded to 0.8.X. I am using iroh dns to combine with some web frameworks to manage multiple nodes, which leads to axum version conflicts. If possible, consider upgrading to 0.8.X
I tried to solve it, but since I just learned rust not long ago, I encountered many problems.

dignifiedquire:
## Description

Depends on https://github.com/n0-computer/tokio-rustls-acme/pull/25

Closes #3273

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,3 @@ unexpected_cfgs = { level = "warn", check-cfg = ["cfg(iroh_docsrs)", "cfg(iroh_l
 
 [workspace.lints.clippy]
 unused-async = "warn"
-
-[patch.crates-io]
-tokio-rustls-acme =  { git = "https://github.com/n0-computer/tokio-rustls-acme", branch = "feat-axum-8" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,3 +40,6 @@ unexpected_cfgs = { level = "warn", check-cfg = ["cfg(iroh_docsrs)", "cfg(iroh_l
 
 [workspace.lints.clippy]
 unused-async = "warn"
+
+[patch.crates-io]
+tokio-rustls-acme =  { git = "https://github.com/n0-computer/tokio-rustls-acme", branch = "feat-axum-8" }
--- a/iroh-dns-server/Cargo.toml
+++ b/iroh-dns-server/Cargo.toml
@@ -12,7 +12,7 @@ readme = "README.md"
 [dependencies]
 anyhow = "1.0.80"
 async-trait = "0.1.77"
-axum = { version = "0.7", features = ["macros"] }
+axum = { version = "0.8", features = ["macros"] }
 axum-server = { version = "0.7", features = ["tls-rustls-no-provider"] }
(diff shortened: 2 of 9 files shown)


Cargo.toml
```diff
@@ -40,3 +40,6 @@ unexpected_cfgs = { level = "warn", check-cfg = ["cfg(iroh_docsrs)", "cfg(iroh_l
 
 [workspace.lints.clippy]
 unused-async = "warn"
+
+[patch.crates-io]
+tokio-rustls-acme =  { git = "https://github.com/n0-computer/tokio-rustls-acme", branch = "feat-axum-8" }
```
matheus23, line 45: This is now merged. Probably needs a release & then we can depend on that?

iroh-dns-server/Cargo.toml
```diff
@@ -24,7 +24,7 @@ derive_more = { version = "1.0.0", features = [
     "from",
 ] }
 dirs-next = "2.0.0"
-governor = "0.6.3" #needs new release of tower_governor for 0.7.0
+governor = "0.8"
```
matheus23, line 27: Bonus points if we can move `iroh-relay` to govenor `0.8`, too, so we don't depend on govenor in two different versions here.
Although admittedly, this is only for server binaries, so doesn't matter that much.

iroh-dns-server/src/http/doh/extract.rs
```diff
@@ -98,53 +99,63 @@ pub struct DnsRequestQuery(pub(crate) DNSRequest, pub(crate) DnsMimeType);
...
-            Some(content_type) if content_type == "application/dns-message" => {
-                handle_dns_message_query(parts, state, src_addr).await
-            }
-            Some(content_type) if content_type == "application/dns-json" => {
-                handle_dns_json_query(parts, state, src_addr).await
+    #[allow(clippy::manual_async_fn)]
+    fn from_request_parts(
+        parts: &mut Parts,
+        state: &S,
+    ) -> impl Future<Output = Result<Self, Self::Rejection>> + Send {
```
matheus23, line 112: I agree with clippy here, why the manual async fn?

Anyhow, seems like you've thought about it, given you've specifically added the `allow`, so I'm fine keeping it. Mostly curious & making sure it's not like an unintentionally left TODO.

dignifiedquire: the return future seems to need an explicit send bound, and I could not get that to compile, feel free to try and make it so

Cargo.toml
dignifiedquire: done

iroh-dns-server/Cargo.toml
dignifiedquire: done

Issue #3276, opened 2025-04-25
Iroh on QUIC multipath


flub:
We want to move most of the magicsocket and holepunching logic into iroh-quinn using the QUIC Multipath and QUIC NAT Traversal drafts.

- [ ] Add multipath to iroh-quinn(-proto)
- [ ] Add nat-traversal to iroh-quinn(-proto)
- [ ] Use multipath in iroh

ramfox:
closed! We've merged the `feat-multipath` branch to `main` and any remaining issues will be tracked via the milestones.

PR #3279, opened 2025-04-28
feat(iroh)!: introduce transport abstraction

dignifiedquire:
A first step towards the work of #3276 

## Breaking changes

- `iroh::watcher` is now its own crate `n0-watcher`
- `iroh::endpoint::Endpoint::node_addr` now returns `impl Watcher<Value = Option<NodeAddr>>`
- `iroh::endpoint::Endpoint::home_relay` now returns `impl Watcher<Value = Vec<RelayUrl>>`
- `iroh::endpoint::Endpoint::bound_sockets` now returns `Vec<SocketAddr>`
- `iroh-quinn` is updated to `0.14.0`
 

Depends on 
- [x] https://github.com/n0-computer/n0-watcher/pull/1
- [x] https://github.com/n0-computer/net-tools/pull/26
- [x] https://github.com/n0-computer/quinn/pull/67
- [x] https://github.com/n0-computer/n0-watcher/pull/3


matheus23:
This PR:
| iroh | 1_to_1 | 1.39 | 2.70 |
| iroh | 1_to_3 | 4.58 | 9.87 |
| iroh | 1_to_5 | 7.11 | 14.31 |
| iroh | 1_to_10 | 12.47 | 22.29 |
| iroh | 2_to_2 | 3.06 | 6.60 |
| iroh | 2_to_4 | 6.21 | 13.64 |
| iroh | 2_to_6 | 9.05 | 19.39 |
| iroh | 2_to_10 | 13.98 | 27.57 |

`main`:
| iroh | 1_to_1 | 1.41 | 2.78 |
| iroh | 1_to_3 | 3.31 | 5.41 |
| iroh | 1_to_5 | 4.89 | 7.47 |
| iroh | 1_to_10 | 6.37 | 8.22 |
| iroh | 2_to_2 | 2.82 | 5.57 |
| iroh | 2_to_4 | 4.11 | 6.43 |
| iroh | 2_to_6 | 5.79 | 8.79 |
| iroh | 2_to_10 | 8.13 | 11.41 |

matheus23 (commented):
Posting a preliminary review before I go through 2k LoC changes in magicsock.rs :sweat_smile:

iroh/src/endpoint.rs
```diff
@@ -908,25 +885,22 @@ impl Endpoint {
...
                 )),
-                (None, Some(relay)) => Some(NodeAddr::from_parts(
-                    node_id,
-                    Some(relay),
-                    std::iter::empty(),
-                )),
-                (None, None) => None,
+                None => relays.pop().map(|relay_url| {
+                    NodeAddr::from_parts(node_id, Some(relay_url), std::iter::empty())
+                }),
```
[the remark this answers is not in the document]
dignifiedquire: that’s what my code does in the last fixed version, that is the result of map?

iroh/src/discovery.rs
```diff
@@ -1038,10 +1038,16 @@ mod test_dns_pkarr {
...
+        // retry multiple times, as the information that was published, might not be enough
+        for _ in 0..10 {
+            // we connect only by node id!
+            let res = ep2.connect(ep1.node_id(), TEST_ALPN).await;
+            if res.is_ok() {
+                return Ok(());
+            }
+            tokio::time::sleep(Duration::from_secs(1)).await;
+        }
+        panic!("failed to connect after 10 attempts")
```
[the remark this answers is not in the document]
Frando: There's an util that does (nearly) that: https://docs.rs/iroh/latest/iroh/test_utils/struct.DnsPkarrServer.html#method.on_node

Should only need adding checking the published addr for some field I think.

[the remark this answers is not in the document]
dignifiedquire: looks like I fixed the bug why I had to change the test, it's back now

iroh/src/test_utils.rs
```diff
@@ -400,7 +400,13 @@ pub(crate) mod pkarr_dns_state {
         pub async fn on_node(&self, node: &NodeId, timeout: Duration) -> Result<()> {
             let timeout = tokio::time::sleep(timeout);
             tokio::pin!(timeout);
-            while self.get(node, |p| p.is_none()) {
+            while self.get(node, |p| {
+                let node_info = p
+                    .as_ref()
+                    .and_then(|p| NodeInfo::from_pkarr_signed_packet(p).ok());
+                eprintln!("got info {:#?}", node_info);
```
matheus23, line 407: Leftover debug print

iroh/src/endpoint.rs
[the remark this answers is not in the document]
matheus23: Sorry, you're right, I misread the code :sweat_smile:

iroh/src/magicsock/transports.rs
```diff
@@ -0,0 +1,434 @@
...
+    pub(crate) fn poll_recv(
+        &self,
+        cx: &mut Context,
+        bufs: &mut [IoSliceMut<'_>],
+        metas: &mut [quinn_udp::RecvMeta],
+        source_addrs: &mut [Addr],
+    ) -> Poll<io::Result<usize>> {
+        debug_assert_eq!(bufs.len(), metas.len(), "non matching bufs & metas");
+
+        // TODO: randomization
```
matheus23, line 112: Or rather, not randomization, but round-robin.
Do you want to deal with this TODO before merging? Seems somewhat fine without.

iroh/src/magicsock/transports.rs
```diff
@@ -0,0 +1,434 @@
...
+    }
+
+    #[cfg(wasm_browser)]
+    pub(crate) fn local_addrs_watch(&self) -> impl Watcher<Value = Vec<Addr>> + Send + Sync {
+        let relays = self.relay.iter().map(|t| {
+            t.local_addr_watch()
+                .map(move |t| t.map(|(url, id)| Addr::RelayUrl(url, id)))
+                .expect("disconnected")
+        });
+        watcher::JoinOpt::new(relays)
```
matheus23, line 170: Use `.map(|ls| ls.into_iter().flatten().collect())` instead? :upside_down_face: 
I'm still a firm believer in having as few (but powerful) combinators as possible.

iroh/src/magicsock/transports.rs
```diff
@@ -0,0 +1,434 @@
...
+    }
+
+    #[cfg(wasm_browser)]
+    pub(crate) fn may_fragment(&self) -> bool {
+        false
+    }
+
+    /// Check if a transport is writable, aka sendable on, for the given `addr`.
+    pub(crate) fn poll_writable(&self, cx: &mut Context, addr: &Addr) -> Poll<io::Result<()>> {
+        // TODO: what about multiple matches?
```
matheus23, line 226: Leftover TODO?
Yeah not sure. Perhaps worth experimenting with more than one relay transport or more than 2 IP transports, and seeing what happens here.
But tbh, the whole polling/writability logic is deeply flawed currently, so :man_shrugging: 
I really hope it'll get better when we eat more of the quinn code/support multiple sockets in quinn.

iroh/src/watcher.rs
```diff
@@ -304,13 +322,133 @@ impl<S: Watcher, T: Watcher> Watcher for (S, T) {
...
+            Poll::Pending
+        }
+    }
+}
+
+/// Combinator to join two watchers
+#[derive(Debug, Clone)]
+pub struct JoinOpt<T: Clone + Eq, W: Watcher<Value = Option<T>>> {
+    watchers: Vec<W>,
+}
```
matheus23, line 387: Continuation from another comment: I'd be happier if we didn't have this, and instead advise using `.map(|vec| vec.into_iter().flatten().collect())`

matheus23 (commented):
Amazing work! Makes me happy seeing this.
Re: Making use of this for WebRTC: Needs more work. E.g. WebRTC probably needs even more changes to `NodeAddr` (to support sending a certificate hash).

Would be fun to experiment with multiple relays configured now, and see if/where it falls apart (and the same for multiple IP transports).

iroh/src/magicsock.rs
```diff
@@ -1013,146 +814,21 @@ impl MagicSock {
...
+    #[instrument("disco_in", skip_all, fields(node = %sender.fmt_short(), ?src))]
+    fn handle_disco_message(&self, sender: PublicKey, sealed_box: &[u8], src: &transports::Addr) {
         trace!("handle_disco_message start");
         if self.is_closed() {
             return;
         }
 
+        if let transports::Addr::RelayUrl(_, node_id) = src {
+            if node_id != &sender {
+                // TODO: return here?
```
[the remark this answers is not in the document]
dignifiedquire: this is just copied from the previous code, I am still undecided..

iroh/src/magicsock.rs
```diff
@@ -2523,6 +1854,19 @@ impl Actor {
                     self.msock.metrics.magicsock.actor_tick_re_stun.inc();
                     self.msock.re_stun("periodic");
                 }
+                new_addr = local_addr_changes.next() => {
```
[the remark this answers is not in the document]
dignifiedquire: nice, I didn't realize this was possible

iroh/src/magicsock/transports.rs
dignifiedquire: yeah, I am unsure about it

cc @flub who introduced this behaviour I believe

iroh/src/magicsock/transports.rs
dignifiedquire: Given the current setup it can't happen, as each transport supports exactly one `Addr` type, so I was planning on leaving this in for now

iroh/src/magicsock.rs
```diff
@@ -487,12 +449,26 @@ impl MagicSock {
...
-        let addr = if let Some(v6) = v6 { v6 } else { v4 };
-        Ok(addr)
+        let addrs: Vec<_> = self
+            .transports
+            .local_addrs()
+            .into_iter()
+            .filter_map(|addr| {
+                let addr: SocketAddr = addr.try_into().ok()?;
+                Some(addr)
+            })
```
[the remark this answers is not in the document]
dignifiedquire: unfortunately, no because `Result != Option`

iroh/src/magicsock/transports.rs
dignifiedquire: I am failing to implement this in a way that it compiles

iroh/src/magicsock.rs
```diff
@@ -513,16 +489,12 @@ impl MagicSock {
                 "connection closed",
             ));
         }
+
+        let mut available_paths = Vec::new();
```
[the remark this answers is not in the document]
dignifiedquire: renamed

iroh/src/magicsock.rs
```diff
@@ -1364,125 +1032,72 @@ impl MagicSock {
         dst_key: PublicKey,
         msg: disco::Message,
     ) -> bool {
-        match dst {
-            SendAddr::Udp(addr) => self.udp_disco_sender.try_send((addr, dst_key, msg)).is_ok(),
-            SendAddr::Relay(ref url) => self.send_disco_message_relay(url, dst_key, msg),
-        }
+        self.disco_sender.try_send((dst, dst_key, msg)).is_ok()
```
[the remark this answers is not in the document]
dignifiedquire: done

iroh/src/magicsock.rs
```diff
@@ -1509,7 +1124,14 @@ impl MagicSock {
             tx_id,
             node_key: self.public_key(),
         });
-        self.try_send_disco_message(dst.clone(), dst_node, msg)?;
+
+        let addr = match dst {
+            SendAddr::Udp(SocketAddr::V4(v4)) => transports::Addr::Ipv4(*v4.ip(), Some(v4.port())),
+            SendAddr::Udp(SocketAddr::V6(v6)) => transports::Addr::Ipv6(*v6.ip(), Some(v6.port())),
+            SendAddr::Relay(ref url) => transports::Addr::RelayUrl(url.clone(), dst_node),
+        };
```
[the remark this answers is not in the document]
dignifiedquire: I tried, but it duplicates the key in quite a few places, which is quite unfortunate, so I wouldn't do this (yet)

iroh/src/magicsock.rs
```diff
@@ -2227,106 +1653,44 @@ impl AsyncUdpSocket for MagicSock {
         self.poll_recv(cx, bufs, metas)
     }
 
+    #[cfg(not(wasm_browser))]
     fn local_addr(&self) -> io::Result<SocketAddr> {
-        #[cfg(not(wasm_browser))]
-        match &*self.sockets.local_addrs.read().expect("not poisoned") {
-            (ipv4, None) => {
-                // Pretend to be IPv6, because our `MappedAddr`s
-                // need to be IPv6.
```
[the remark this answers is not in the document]
dignifiedquire: done

iroh/src/magicsock/transports.rs
matheus23:
```rs
#[cfg(wasm_browser)]
pub(crate) fn local_addrs_watch(&self) -> impl Watcher<Value = Vec<Addr>> + Send + Sync {
    let relays = self.relay.iter().map(|t| t.local_addr_watch());
    watcher::Join::new(relays)
        .map(|relays| relays.into_iter().flatten().map(Addr::from).collect())
        .expect("disconnected")
}
```

iroh/src/watcher.rs
matheus23: (`.map()` isn't all you need, you also need `Join::new` in advance. I'm totally in favor of keeping `Join`!)

iroh/src/magicsock/transports.rs
dignifiedquire: thanks

iroh/src/magicsock/transports.rs
```diff
@@ -0,0 +1,432 @@
...
+
+#[derive(Debug)]
+pub(crate) struct Transports {
+    #[cfg(not(wasm_browser))]
+    ip: Vec<IpTransport>,
+    relay: Vec<RelayTransport>,
+}
+
+impl Transports {
+    /// Create a new transports structure.
```
flub, line 33: nit, especially since this is internal, but good habits and all that:
> The summary line should be written in third person singular present indicative form. Basically, this means write ‘Returns’ instead of ‘Return’.

https://rust-lang.github.io/rfcs/1574-more-api-documentation-conventions.html#appendix-a-full-conventions-text

iroh/src/magicsock/transports.rs
```diff
@@ -0,0 +1,432 @@
...
+        trace!(?destination, "sending");
+
+        match destination {
+            #[cfg(wasm_browser)]
+            Addr::Ipv4(..) => {
+                return Poll::Ready(Err(io::Error::other("IPv4 is unsupported in browser")))
+            }
+            #[cfg(not(wasm_browser))]
+            Addr::Ipv4(addr, port) => {
+                let addr = SocketAddr::V4(SocketAddrV4::new(*addr, port.unwrap_or_default()));
```
flub, line 62: I'm confused how you can try send to a `0` port.

iroh/src/magicsock/transports.rs
```diff
@@ -0,0 +1,432 @@
...
+        metas: &mut [quinn_udp::RecvMeta],
+        source_addrs: &mut [Addr],
+    ) -> Poll<io::Result<usize>> {
+        debug_assert_eq!(bufs.len(), metas.len(), "non matching bufs & metas");
+
+        // TODO: randomization
+        macro_rules! poll_transport {
+            ($socket:expr) => {
+                match $socket.poll_recv(cx, bufs, metas, source_addrs)? {
+                    Poll::Pending | Poll::Ready(0) => {}
```
flub, line 116: Here I'm also a bit worried about `Ready(0)`.

iroh/src/magicsock/transports.rs
```diff
@@ -0,0 +1,432 @@
...
+        // Note however that the current quinn impl calls UdpPoller::poll_writable()
+        // **before** it calls try_send(), as opposed to how it is documented.  That is a
+        // problem as we would not yet know the path that needs to be polled.  To avoid such
+        // ambiguity the API could be changed to a .poll_send(&self, cx: &mut Context,
+        // io_poller: Pin<&mut dyn UdpPoller>, transmit: &Transmit) -> Poll<io::Result<()>>
+        // instead of the existing .try_send() because then we would have control over this.
+        //
+        // Right now however we have one single poller behaving the same for each
+        // connection.  It checks all paths and returns Poll::Ready as soon as any path is
+        // ready.
```
flub, line 265: I'm not sure any longer how correct this comment is.  Maybe we should stop cargo-culting it around.  The issue as we currently understand it is really that quinn needs to be aware of multiple sockets so that it can understand pending differently for each one.  And less to do with per-NodeId stuff.

iroh/src/magicsock/transports/ip.rs
```diff
@@ -0,0 +1,157 @@
...
+        let res = self.socket.try_send_quinn(&quinn_udp::Transmit {
+            destination,
+            ecn: transmit.ecn,
+            contents: transmit.contents,
+            segment_size: transmit.segment_size,
+            src_ip: transmit
+                .src_ip
+                .clone()
+                .map(|a| a.try_into().expect("invalid src_ip")),
+        });
```
flub, line 61: Is it possible at all to keep panic-generating code out of this bit of the hot path?  I assume this has to deal with a relay addr appearing where an IP one is needed for this socket?

I could also imagine a situation where someone binds two IPv4 addresses (or IPv6, whatever) and they're both on different network interfaces, which with their own subnet.  You wouldn't be able to send with the wrong IP address in that case.  I think you're right now just bumping this into the kernel which would probably just return an error instead of sending it.  So maybe no need to worry about IP here.

Anyway, `quinn_udp::Transmit` says `src_ip: Option<IpAddr>`.  So one option would be to log a warning and turn this into `None` if you got something crazy.

iroh/src/magicsock/transports/relay.rs
```diff
@@ -0,0 +1,472 @@
...
+
+    pub(super) fn poll_recv(
+        &self,
+        cx: &mut Context,
+        bufs: &mut [io::IoSliceMut<'_>],
+        metas: &mut [quinn_udp::RecvMeta],
+        source_addrs: &mut [Addr],
+    ) -> Poll<io::Result<usize>> {
+        let mut num_msgs = 0;
+        'outer: for ((buf_out, meta_out), addr) in bufs
```
flub, line 133: Not sure why this has a label?  It's not a nested loop?  I guess it was before.

iroh/src/endpoint.rs
```diff
@@ -1013,7 +991,7 @@ impl Endpoint {
     /// ```
     ///
     /// [STUN]: https://en.wikipedia.org/wiki/STUN
-    pub fn direct_addresses(&self) -> watcher::Direct<Option<BTreeSet<DirectAddr>>> {
+    pub fn direct_addresses(&self) -> n0_watcher::Direct<Option<BTreeSet<DirectAddr>>> {
```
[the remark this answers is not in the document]
dignifiedquire: I suspect we will want to use more impl traits in the final api, but for now I would leave it as is, until these apis are more settled

iroh/src/magicsock/transports.rs
dignifiedquire: as above, this is what we have on `main`

iroh/src/magicsock/transports.rs
dignifiedquire: so I should remove the comment?

iroh/src/magicsock/transports/relay.rs
dignifiedquire: fixed

iroh/src/magicsock/transports/ip.rs
dignifiedquire: fixed

iroh/src/magicsock/transports.rs
```diff
@@ -0,0 +1,432 @@
...
+pub(crate) struct Transmit<'a> {
+    pub(crate) ecn: Option<quinn_udp::EcnCodepoint>,
+    pub(crate) contents: &'a [u8],
+    pub(crate) segment_size: Option<usize>,
+    pub(crate) src_ip: Option<Addr>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) enum Addr {
+    Ipv4(Ipv4Addr, Option<u16>),
```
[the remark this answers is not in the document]
dignifiedquire: fixed

iroh/src/magicsock/transports.rs
dignifiedquire: fixed

iroh/src/magicsock/transports.rs
```diff
@@ -0,0 +1,432 @@
...
+mod ip;
+mod relay;
+
+#[cfg(not(wasm_browser))]
+pub(crate) use self::ip::IpTransport;
+pub(crate) use self::relay::{RelayActorConfig, RelayTransport};
+use super::NetInfo;
+
+#[derive(Debug)]
+pub(crate) struct Transports {
```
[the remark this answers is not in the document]
dignifiedquire: done

iroh/src/magicsock.rs
```diff
@@ -513,16 +486,12 @@ impl MagicSock {
                 "connection closed",
             ));
         }
+
+        let mut active_paths = Vec::new();
```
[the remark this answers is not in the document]
dignifiedquire: done

iroh/src/magicsock/transports.rs
flub: Yeah, I think so.

dignifiedquire: done

matheus23 (approved):
I think with the most recent changes one of my previous review comments does apply again: You could merge `Addr::Ipv4` and `Addr::Ipv6` into an `Addr::Ip` and store a `SocketAddr`. But that's mostly a nit, and if it makes code look a little worse, let's not do it.

flub (approved):

iroh/src/magicsock/transports.rs
```diff
@@ -0,0 +1,429 @@
...
+    Ipv4(SocketAddrV4),
+    Ipv6(SocketAddrV6),
+    RelayUrl(RelayUrl, NodeId),
+}
+
+impl Default for Addr {
+    fn default() -> Self {
+        Self::Ipv6(SocketAddrV6::new(Ipv6Addr::UNSPECIFIED, 0, 0, 0))
+    }
+}
```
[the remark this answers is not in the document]
flub: Obsolete comment, I found out by now...

matheus23:
> > I think with the most recent changes one of my previous review comments does apply again: You could merge `Addr::Ipv4` and `Addr::Ipv6` into an `Addr::Ip` and store a `SocketAddr`. But that's mostly a nit, and if it makes code look a little worse, let's not do it.
> 
> The issue is not the port, it is the `NodeId`, which creates problems when trying to unify them

I can't quite follow. There's no `NodeId` in `Addr::Ip` or `SocketAddr`. As far as I can tell, `Addr::Ipv4` and `Addr::Ipv6` and `SocketAddr` are isomorphic?

FWIW, I'm *not* talking about the other `Addr::Relay` variant

iroh/src/magicsock/transports.rs
```diff
@@ -0,0 +1,429 @@
...
+    pub(crate) ecn: Option<quinn_udp::EcnCodepoint>,
+    pub(crate) contents: &'a [u8],
+    pub(crate) segment_size: Option<usize>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) enum Addr {
+    Ipv4(SocketAddrV4),
+    Ipv6(SocketAddrV6),
+    RelayUrl(RelayUrl, NodeId),
```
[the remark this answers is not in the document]
dignifiedquire: yep, good call

flub (approved):

iroh/src/magicsock/transports/relay.rs
```diff
@@ -0,0 +1,423 @@
...
+    _actor_handle: AbortOnDropHandle<()>,
+    my_relay: Watchable<Option<RelayUrl>>,
+    my_node_id: NodeId,
+}
+
+impl RelayTransport {
+    pub(crate) fn new(config: RelayActorConfig) -> Self {
+        let (relay_datagram_send_tx, relay_datagram_send_rx) = mpsc::channel(256);
+
+        let (relay_datagram_recv_tx, relay_datagram_recv_rx) = mpsc::channel(512);
```
flub, line 42: I know you're probably copying these values... but how did we end up with the recv queue being twice as long as the send queue?  Anyway, something to clean up some other time.

iroh/src/magicsock/transports/relay.rs
```diff
@@ -0,0 +1,423 @@
...
+                        io::ErrorKind::NotConnected,
+                        "connection closed",
+                    )));
+                }
+                Poll::Pending => {
+                    break;
+                }
+            };
+
+            buf_out[..dm.buf.len()].copy_from_slice(&dm.buf);
```
flub, line 103: Long, long in the future I can now see us polling the TCP socket from the relay server directly during this poll, allowing us to avoid this copy.  Especially once we don't have any non-quic messages left since then only very few times we'll need to send a non-QUIC message over the relay protocol that should not be passed up.

iroh/src/magicsock/transports.rs
```diff
@@ -0,0 +1,622 @@
...
+        trace!(?destination, "sending");
+
+        let mut any_match = false;
+        match destination {
+            #[cfg(wasm_browser)]
+            Addr::Ip(..) => return Err(io::Error::other("IP is unsupported in browser")),
+            #[cfg(not(wasm_browser))]
+            Addr::Ip(addr) => {
+                for sender in &self.ip {
+                    if sender.is_valid_send_addr(addr) {
```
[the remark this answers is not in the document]
dignifiedquire: I actually find that much harder to read, because the `for` loop gets split over multiple lines

iroh/src/magicsock/transports/ip.rs
```diff
@@ -0,0 +1,249 @@
...
+}
+
+impl IpTransport {
+    pub(crate) fn new(
+        bind_addr: SocketAddr,
+        socket: Arc<UdpSocket>,
+        metrics: Arc<MagicsockMetrics>,
+    ) -> Self {
+        // Currently gets updated on manual rebind
+        // TODO: update when UdpSocket under the hood rebinds automatically
```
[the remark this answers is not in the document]
dignifiedquire: it's important todo eventually, but the world doesn't end if it doesn't make it in this PR

Cargo.lock
```diff
@@ -5489,7 +5488,7 @@ version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cf221c93e13a30d793f7645a0e7762c55d169dbb0a49671918a2319d289b10bb"
 dependencies = [
- "windows-sys 0.59.0",
+ "windows-sys 0.48.0",
```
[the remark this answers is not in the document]
dignifiedquire: no idea, this changes back in my other PR, cargo seems to resolve this differently in some cases 🤷

iroh/src/magicsock.rs
```diff
@@ -924,84 +628,118 @@ impl MagicSock {
...
-                                len = meta.len,
-                                "UDP recv quic packets: no node state found, skipping",
-                            );
-                            // If we have no node state for the from addr, set len to 0 to make
-                            // quinn skip the buf completely.
-                            meta.len = 0;
+                enum AddrOrUrl {
+                    Addr(SocketAddr),
+                    Url(RelayUrl, NodeId),
+                }
```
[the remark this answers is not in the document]
dignifiedquire: removed, good catch

iroh/src/magicsock.rs
```diff
@@ -872,12 +573,11 @@ impl MagicSock {
...
-    #[cfg(not(wasm_browser))]
-    fn process_udp_datagrams(
+    fn process_datagrams(
         &self,
-        from_ipv4: bool,
         bufs: &mut [io::IoSliceMut<'_>],
         metas: &mut [quinn_udp::RecvMeta],
+        source_addrs: &[transports::Addr],
     ) {
         debug_assert_eq!(bufs.len(), metas.len(), "non matching bufs & metas");
```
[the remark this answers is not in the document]
dignifiedquire: done

iroh/src/magicsock.rs
```diff
@@ -487,17 +462,31 @@ impl MagicSock {
...
+            return Ok(*addr);
+        }
+
+        Err(io::Error::other("no valid socket available"))
     }
 
-    /// Implementation for AsyncUdpSocket::try_send
     #[instrument(skip_all)]
-    fn try_send(&self, transmit: &quinn_udp::Transmit) -> io::Result<()> {
+    fn prepare_send(
```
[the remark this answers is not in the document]
dignifiedquire: done

iroh/src/magicsock.rs
```diff
@@ -487,17 +462,31 @@ impl MagicSock {
...
-        let (v4, v6) = self.local_addr();
-        let addr = if let Some(v6) = v6 { v6 } else { v4 };
-        Ok(addr)
+        let addrs: Vec<_> = self
+            .local_addrs_watch
+            .get()
+            .expect("disconnected")
+            .into_iter()
+            .filter_map(|addr| SocketAddr::try_from(addr).ok())
+            .collect();
```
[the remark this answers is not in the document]
dignifiedquire: done

PR #3282, opened 2025-04-29
feat(iroh): Allow connecting with "fallback" ALPNs

### LINKED ISSUE #2949: Allow specifying multiple ALPNs in `Endoint::connect`
When implementing a backwards-compatible protocol, one wants to assign different ALPNs to different protocol versions.
Examples:
- `http/0.9`, `http/1.0`, `http/1.1`, `h2` and `h3`
- `iroh-blobs/4`, eventually `iroh-blobs/5`

It's important to indicate which versions you support, and often is the case that you indicate *multiple versions* simultaneously. E.g. sending an HTTP request using `["http/1.1", "h2", "h3"]` as the ALPNs.

We currently don't have a way of specifying multiple ALPNs when doing an `Endpoint::connect`.

On the server side, you usually configure an *ordered* list of ALPNs and it chooses the first matching one. This is described in [RFC 7301](https://www.rfc-editor.org/rfc/rfc7301#section-3.2):
> It is expected that a server will have a list of protocols that it supports, in preference order, and will only select a protocol if the client supports it.  In that case, the server SHOULD select the most highly preferred protocol that it supports and that is also advertised by the client.

This manifests itself in APIs like that of [rustls](https://docs.rs/rustls/latest/rustls/server/struct.ClientHello.html#method.alpn), its docs write:
> The server can specify supported ALPN protocols by setting [`ServerConfig::alpn_protocols`](https://docs.rs/rustls/latest/rustls/server/struct.ServerConfig.html#structfield.alpn_protocols). During the handshake, the server will select the first protocol configured that the client supports.

The docs for `ServerConfig::alpn_protocols` say:
> Protocol names we support, most preferred first.

matheus23:
## Description

Allows connecting with multiple ALPNs so you can actually take advantage of ALPN also on the connecting side to support connections with older accept sides.

## Notes & open questions

This only implements this via the `ConnectOptions`, which I think is fair? I generally like the pattern of `connect` -> `connect_with_opts`, and I don't think this is worth changing `connect` for.

Closes #2949 

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -253,7 +253,7 @@ impl Builder {
     /// Sets the [ALPN] protocols that this endpoint will accept on incoming connections.
     ///
     /// Not setting this will still allow creating connections, but to accept incoming
-    /// connections the [ALPN] must be set.
+    /// connections at least one [ALPN] must be set.
     ///
     /// [ALPN]: https://en.wikipedia.org/wiki/Application-Layer_Protocol_Negotiation
     pub fn alpns(mut self, alpn_protocols: Vec<Vec<u8>>) -> Self {
@@ -1282,7 +1282,23 @@ impl ConnectOptions {
         self
     }
 
-    /// TODO
+    /// Sets more [ALPN] identifiers that should be signaled as supported on connection.
+    ///
+    /// This allows connecting to servers that may only support older versions of your
+    /// protocol. In this case, you would add the older [ALPN] identifiers with this
+    /// function.
+    ///
+    /// You'll know the final negotiated [ALPN] identifier once your connection was
+    /// established using [`Connection::alpn`], or even slightly earlier in the
+    /// handshake by using [`Connecting::alpn`].
+    ///
+    /// The [ALPN] identifier order on the connect side doesn't matter, since it's the
+    /// accept side that determines the protocol.
+    ///
+    /// For setting the supported [ALPN] identifiers on the accept side, see the endpoint
+    /// builder's [`Builder::alpns`] function.
+    ///
+    /// [ALPN]: https://en.wikipedia.org/wiki/Application-Layer_Protocol_Negotiation
     pub fn with_additional_alpns(mut self, alpns: Vec<Vec<u8>>) -> Self {
         self.additional_alpns = alpns;
         self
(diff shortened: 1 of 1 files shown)


iroh/src/endpoint.rs
```diff
@@ -1279,6 +1281,28 @@ impl ConnectOptions {
         self.transport_config = Some(transport_config);
         self
     }
+
+    /// Sets more [ALPN] identifiers that should be signaled as supported on connection.
```
flub, line 1285:
```suggestion
    /// Sets multiple [ALPN] identifiers to negotiate with the server.
```

iroh/src/endpoint.rs
```diff
@@ -2779,4 +2803,102 @@ mod tests {
...
+        let server = Endpoint::builder()
+            .relay_mode(RelayMode::Disabled)
+            .alpns(accept_alpns)
+            .bind()
+            .await?;
+        let server_addr = server.node_addr().await?;
+        let server_task = tokio::spawn({
+            let server = server.clone();
+            async move {
+                let incoming = server.accept().await.unwrap();
```
dignifiedquire, line 2827: assert that this is the correct alpn in here?

iroh/src/endpoint.rs
```diff
@@ -1279,6 +1281,28 @@ impl ConnectOptions {
...
+    /// handshake by using [`Connecting::alpn`].
+    ///
+    /// The [ALPN] identifier order on the connect side doesn't matter, since it's the
+    /// accept side that determines the protocol.
+    ///
+    /// For setting the supported [ALPN] identifiers on the accept side, see the endpoint
+    /// builder's [`Builder::alpns`] function.
+    ///
+    /// [ALPN]: https://en.wikipedia.org/wiki/Application-Layer_Protocol_Negotiation
+    pub fn with_additional_alpns(mut self, alpns: Vec<Vec<u8>>) -> Self {
```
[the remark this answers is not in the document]
flub: Hum, I see.  I don't have any obviously better ideas, so I think your reasoning here is fine.

flub (approved):

iroh/src/endpoint.rs
matheus23: I don't like your suggestion, but it also made me dislike the current wording.
I have some ideas for something better :v:

matheus23: Pushed something that I think is an improvement.

PR #3288, opened 2025-05-02
chore: update deps

dignifiedquire:
Updating our dependencies breaks the `rtt` tests for some reason..


matheus23:
#3290 includes this PR, and we can't merge this, as it's missing the necessary fix, and needs some conflicts resolved.

Should we just close this in favor of #3290?

PR #3290, opened 2025-05-05
fix(iroh)!: Update dependencies & fix 0-RTT with newer rustls by pulling the expected `NodeId` out of the `ServerName` in verifiers

matheus23:
## Description

This is code that will both update deps like #3288 but also makes 0-RTT tests pass again. 

They were failing due to a [change in `rustls`](https://github.com/rustls/rustls/releases/tag/v%2F0.23.24):
> Behavior change: Clients no longer offer resumption between different `ClientConfig`s that share a resumption store but do not share server certificate verification and client authentication credentials. If you share a resumption store between multiple `ClientConfig`s, please ensure their server certificate verification and client authentication credentials are also shared. Please read the new documentation on the `ClientConfig::resumption` item for details.
>
> Additionally, if you share a resumption store or ticketer between multiple `ServerConfig`s, please see the new documentation on `ServerConfig` about this.

Essentially what they're doing is [checking that the certificate verifiers are `Arc::prt_eq` and depending on that allow or disallow 0-RTT](https://github.com/rustls/rustls/pull/2361/files#diff-9d58cb90e1df25604b88544aca7ea999ccd23bc5657384e196e77f407c0f4acfR256).

When running the tests with rustls logging enabled, I confirmed that this is indeed what's happening:
> [TRACE rustls::msgs::persist] resumption not allowed between different ServerCertVerifiers

---

The fix isn't straight-forward.
The reason we're creating new `ServerVerifier`s every time is that we're passing in the expected remote node ID every time.

There used to be no other way of grabbing the other node's ID in the verifier, until, coincidentally #3146 landed, which changed the server name that's set when connecting from always being `localhost` to `<base32 NodeId>.iroh.invalid`.

I'm now pulling in that value from the `ServerName` that's passed to the `ServerCertificateVerifier` in `verify_server_cert`.
As far as I understand that value *should* be a client-set value not a value we get over the network.

The downside of doing this is that we'll not be compatible with iroh versions that don't use the `<base32 NodeId>.iroh.invalid` server names.

---

There's an unrelated change in here that's changing `make_client_config` and `make_server_config` infallible, as was done in #3161, which will definitely also have merge conflicts with this PR.

## Breaking Changes

- `Router::spawn` is now a plain function instead of an `async fn`
- `Router::spawn` is now infallible, instead of returning `anyhow::Result<()>`

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh/src/tls/name.rs
+++ b/iroh/src/tls/name.rs
@@ -1,0 +1,33 @@
+//! Implementation of encoding iroh NodeIds as domain names.
+//!
+//! We used to use a constant "localhost" for the TLS server name - however, that affects
+//! 0-RTT and would put all of the TLS session tickets we receive into the same bucket in
+//! the TLS session ticket cache.
+//! So we choose something that'd dependent on the NodeId.
+//! We cannot use hex to encode the NodeId, as that'd encode to 64 characters, but we only
+//! have 63 maximum per DNS subdomain. Base32 is the next best alternative.
+//! We use the `.invalid` TLD, as that's specified (in RFC 2606) to never actually resolve
+//! "for real", unlike `.localhost` which is allowed to resolve to `127.0.0.1`.
+//! We also add "iroh" as a subdomain, although those 5 bytes might not be necessary.
+//! We *could* decide to remove that indicator in the future likely without breakage.
+
+use data_encoding::BASE32_DNSSEC;
+use iroh_base::NodeId;
+
+pub(crate) fn encode(node_id: NodeId) -> String {
+    format!("{}.iroh.invalid", BASE32_DNSSEC.encode(node_id.as_bytes()))
+}
+
+pub(crate) fn decode(name: &str) -> Option<NodeId> {
+    let [base32_node_id, "iroh", "invalid"] = name.split(".").collect::<Vec<_>>()[..] else {
+        return None;
+    };
+    NodeId::from_bytes(
+        &BASE32_DNSSEC
+            .decode(base32_node_id.as_bytes())
+            .ok()?
+            .try_into()
+            .ok()?,
+    )
+    .ok()
+}
--- a/iroh/src/tls/name.rs
+++ b/iroh/src/tls/name.rs
@@ -31,3 +31,25 @@ pub(crate) fn decode(name: &str) -> Option<NodeId> {
     )
     .ok()
 }
+
+#[cfg(test)]
+mod tests {
+    use iroh_base::SecretKey;
+
+    #[test]
+    fn test_roundtrip() {
+        let key = SecretKey::generate(rand::rngs::OsRng);
+        let node_id = key.public();
+        println!("{}", super::encode(node_id));
+        assert_eq!(Some(node_id), super::decode(&super::encode(node_id)));
+    }
+
+    #[test]
+    fn test_snapshot() {
+        let key = SecretKey::from_bytes(&[0; 32]);
+        assert_eq!(
+            super::encode(key.public()),
+            "7dl2ff6emqi2qol3l382krodedij45bn3nh479hqo14a32qpr8kg.iroh.invalid",
+        );
+    }
+}
--- a/iroh/src/tls/verifier.rs
+++ b/iroh/src/tls/verifier.rs
@@ -95,7 +95,10 @@ impl ServerCertVerifier for ServerCertificateVerifier {
         _ocsp_response: &[u8],
         _now: rustls::pki_types::UnixTime,
     ) -> Result<ServerCertVerified, rustls::Error> {
-        let Some(remote_peer_id) = super::name::decode(server_name.to_str().as_ref()) else {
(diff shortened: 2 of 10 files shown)


iroh/src/tls/verifier.rs
```diff
@@ -75,32 +72,35 @@ pub(super) struct ServerCertificateVerifier {
...
+    remote_key
+}
+
 impl ServerCertVerifier for ServerCertificateVerifier {
     fn verify_server_cert(
         &self,
         end_entity: &Certificate,
         intermediates: &[Certificate],
-        _server_name: &rustls::pki_types::ServerName,
+        server_name: &rustls::pki_types::ServerName,
```
dignifiedquire, line 94: who sets this value?

matheus23: That said - I find this reasoning quite convincing: The only place we use `tls::name` is in `Endpoint::connect`.
What we want to prevent is the server side setting the `server_name`. If that were the case, then there must've been some place in the sever-side code for us where we use `tls::name` to set the `server_name`. If the server name would've been set to something else, then `verify_sever_cert` would fail due to not parsing it correctly.
But there's no other occurrence of `tls::name` or greps for an `iroh.invalid` string.

FWIW

Frando:
> Cannot connect to iroh nodes of versions 0.32 or below due to changes in how the TLS setup works. (Accepting incoming connections should keep working just fine.)

Deltachat only updated to 0.33 ~2 months ago, so I presume that this change would mean that Deltachat cannot update to 0.35 because this would be a wire-level breaking change between all Deltachat versions with android which are older than 2 months. AFAIK deltachat so far intended to update to 0.35 as a last update before iroh 1.0.

matheus23:
Yeah honestly, I think this PR should probably not go into this release.

Regressing 0-RTT in this release vs. what is effectively a soft wire-breaking change makes me choose regressing 0-RTT.

matheus23:
Actually @flub correctly pointed out: It's not actually a wire-breaking change! As I write in the PR review comment, the `server_name` that's checked isn't the one that's set from the remote node, it's the server name that *we* set. Thus, there's no breaking change concerns.

This is easy enough to test, so I pulled out the `echo.rs` example at tag `v0.32.0` and this branch's `echo.rs` example, and let them connect to each other via a small modification so I could enter a `NodeTicket`, and yeah, it works.

I didn't test 0-RTT, since (1) the changes in this PR are actually only tangentially relevant to 0-RTT and are more changes to how we configure rustls' client authentication/server authentication, and (2) because version 0.32 didn't have 0-RTT support.

iroh/src/tls/name.rs
```diff
@@ -0,0 +1,33 @@
...
+    };
+    NodeId::from_bytes(
+        &BASE32_DNSSEC
+            .decode(base32_node_id.as_bytes())
+            .ok()?
+            .try_into()
+            .ok()?,
+    )
+    .ok()
+}
```
dignifiedquire, line 33: add a small test that these roundrip?

iroh/src/tls/verifier.rs
```diff
@@ -126,7 +128,7 @@ impl ServerCertVerifier for ServerCertificateVerifier {
 
                 let end_entity_as_spki = SubjectPublicKeyInfoDer::from(end_entity.as_ref());
 
-                if !self.trusted_spki.contains(&end_entity_as_spki) {
+                if public_key_to_spki(&remote_peer_id) != end_entity_as_spki {
```
dignifiedquire, line 131: I wonder if we should instead decode `end_entity` into a `NodeId`

matheus23: I looked into this a bit now, but `SubjectPublicKeyInfoDer` turns out to just be a wrapper around opaque bytes.
`public_key_to_spki` indirectly uses the [spki](https://docs.rs/spki/latest/spki/index.html) crate to encode. I looked at it a bit, but it's... not fun?

FWIW, for Ed25519 keys, there's not much that can go wrong. It's the Ed25519 public key ASN.1 object identifier + the public bytes encoded as bitstring, so effectively a zero byte (indicating there was zero bits added as padding) and the raw public key bytes.

All in all, we *could* match the first 12 bytes of the SPKI against some constant byte string (they should never change), but I'm not sure we should?
Using a library really wouldn't be fun.

matheus23: Now that I think about this, this is effectively doing the same thing: By using `public_key_to_spki` we prefix the expected public key bytes (`remote_peer_id`) with the object identifier and ASN.1 encoding stuff and then verify that the prefix matches as well as the key itself.

:man_shrugging:

dignifiedquire: I guess we can just leave it

matheus23: I'll add a comment with what I learned

matheus23:
Not updating redb as that'd be MSRV breaking

dignifiedquire (approved):

PR #3292, opened 2025-05-06
feat(iroh-relay)!: adjust APIs to make it easier to create `RelayMap`s from lists of `RelayUrls`

ramfox:
## Description

In our examples and tests, it is enough to build `RelayMap`s from a single `RelayUrl`, but our users will more likely be in the situation where they have multiple relays.

Now we have `From<RelayUrl>`, `From<(RelayUrl, u16)>`, `FromIterator<RelayUrl>` `FromIterator<(RelayUrl, u16)>` impls to create a `RelayMap`.

Also, `From<RelayUrl>` and `From<(RelayUrl, u16)>` to create a `RelayNode`.

## Breaking Changes

- iroh
    - remove
        - `pub fn default_from_node(url: RelayUrl, stun_port: u16) -> RelayMap`
    - change
        - `pub fn from_url(url: RelayUrl) -> RelayMap`, use `From<RelayUrl>` instead
       

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh-relay/src/relay_map.rs
+++ b/iroh-relay/src/relay_map.rs
@@ -97,7 +97,7 @@ impl From<RelayNode> for RelayMap {
     }
 }
 
-/// Create a `RelayMap` from an iterator of `RelayUrl`.
+/// Creates a [`RelayMap`] from an iterator of [`RelayUrl`].
 ///
 /// The [`RelayNode`]s in the [`RelayMap`] will have the default STUN and QUIC address
 /// discovery ports.
--- a/iroh-relay/src/relay_map.rs
+++ b/iroh-relay/src/relay_map.rs
@@ -67,7 +67,7 @@ impl FromIterator<RelayNode> for RelayMap {
 
 /// Creates a [`RelayMap`] from a [`RelayUrl`].
 ///
-/// The `RelayNode`s in the `RelayMap` will have the default stun and quic address
+/// The [`RelayNode`]s in the [`RelayMap`] will have the default STUN and QUIC address
 /// discovery ports.
 impl From<RelayUrl> for RelayMap {
     fn from(value: RelayUrl) -> Self {
--- a/iroh-relay/src/relay_map.rs
+++ b/iroh-relay/src/relay_map.rs
@@ -2,7 +2,6 @@
 
 use std::{collections::BTreeMap, fmt, sync::Arc};
 
-use anyhow::{ensure, Result};
 use iroh_base::RelayUrl;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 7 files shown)


flub:
I think it is great to be sleuthing at these APIs!  They are indeed not very ergonomic.  Would they not be more rusty by leveraging traits more?

- Instead of `from_url` use `From<RelayUrl>`
- Instead of `default_from_nodes` use `FromIterator<(RelayUrl, u16)>`
- Instead of `from_urls` use `FromIterator<RelayUrl>`
- Instead of `from_nodes` use `FromIterator<Arc<RelayNode>>`.  Actually, what is going on with that constructor?  It has an `Into<Arc<RelayNode>>` already.

I'm kind of curious why they grew into what they currently are, so might be missing the mark.

ramfox:
Did everything suggested in @flub your comment, but adjusted the code to remove the need for the `Arc` in `FromIterator<Arc<RelayNode>> for RelayMap`

iroh-relay/src/relay_map.rs
```diff
@@ -52,50 +51,81 @@ impl RelayMap {
...
-    }
-
-    /// Constructs the [`RelayMap`] from an iterator of [`RelayNode`]s.
-    pub fn from_nodes<I: Into<Arc<RelayNode>>>(value: impl IntoIterator<Item = I>) -> Result<Self> {
-        let mut map = BTreeMap::new();
-        for node in value.into_iter() {
-            let node = node.into();
-            ensure!(!map.contains_key(&node.url), "Duplicate node url");
-            map.insert(node.url.clone(), node);
+/// Create a `RelayMap` from a `RelayUrl`.
```
flub, line 68: Let's make these links in the docs:
```suggestion
/// Creates a [`RelayMap`] from a [`RelayUrl`].
```

iroh-relay/src/relay_map.rs
```diff
@@ -52,50 +51,81 @@ impl RelayMap {
...
+///
+/// The `RelayNode`s in the `RelayMap` will have the default stun and quic address
```
flub, line 70:
```suggestion
/// The [`RelayNode`]s in the [`RelayMap`] will have the default STUN and QUIC address
```

iroh-relay/src/relay_map.rs
```diff
@@ -52,50 +51,81 @@ impl RelayMap {
...
+/// discovery ports.
+impl From<RelayUrl> for RelayMap {
+    fn from(value: RelayUrl) -> Self {
+        Self {
+            nodes: Arc::new([(value.clone(), Arc::new(value.into()))].into()),
+        }
+    }
+}
+
+/// Create a `RelayMap` from a `RelayUrl` and a stun port.
```
flub, line 80:
```suggestion
/// Creates a [`RelayMap`] from a [`RelayUrl`] and a stun port.
```

flub: To be honest, I'm also fine with removing this shortcut entirely.  It is a bit weird, because why is there no such shortcut for the QAD port?  And it's easier to add API later than remove it.

iroh-relay/src/relay_map.rs
```diff
@@ -52,50 +51,81 @@ impl RelayMap {
...
+
+impl From<RelayNode> for RelayMap {
+    fn from(value: RelayNode) -> Self {
+        Self {
+            nodes: Arc::new([(value.url.clone(), Arc::new(value))].into()),
+        }
+    }
+}
+
+/// Create a `RelayMap` from an iterator of `RelayUrl`.
```
flub, line 100:
```suggestion
/// Creates a [`RelayMap`] from an iterator of [`RelayUrl`].
```

iroh-relay/src/relay_map.rs
```diff
@@ -52,50 +51,81 @@ impl RelayMap {
...
+///
+/// The `RelayNode`s in the `RelayMap` will have the default stun and quic address
```
flub, line 102:
```suggestion
/// The [`RelayNode`]s in the [`RelayMap`] will have the default STUN and QUIC address
```

iroh-relay/src/relay_map.rs
```diff
@@ -52,50 +51,81 @@ impl RelayMap {
...
+            ),
+        }
+    }
+}
+
+/// Create a `RelayMap` from an iterator of `RelayUrl`s and a stun ports.
+///
+/// The resulting `RelayMap` contains `RelayNode`s with the default quic address
+/// discovery port.
+impl FromIterator<(RelayUrl, u16)> for RelayMap {
```
flub, line 120: Like the above `(url, stun_port)` constructor I'd be fine with removing this if it doesn't cause too much breakage.

iroh-relay/src/relay_map.rs
```diff
@@ -131,6 +161,28 @@ pub struct RelayNode {
...
+        Self {
+            url: value,
+            stun_only: false,
+            stun_port: DEFAULT_STUN_PORT,
+            quic: quic_config(),
+        }
+    }
+}
+
+impl From<(RelayUrl, u16)> for RelayNode {
```
flub, line 175: Again, I'm only so-so convinced this is worth it rather than more confusing.  It made more sense when there only was STUN before, but now with QAD?

iroh-relay/src/relay_map.rs
```diff
@@ -52,50 +51,81 @@ impl RelayMap {
...
+                    .collect(),
+            ),
+        }
+    }
+}
+
+/// Create a `RelayMap` from an iterator of `RelayUrl`s and a stun ports.
+///
+/// The resulting `RelayMap` contains `RelayNode`s with the default quic address
+/// discovery port.
```
matheus23, line 119: I really like these comments!
Quick tip: You can put doc comments on the `fn` *inside* the `impl` block as well, and they'll show up in documentation and IDEs, so you can kind of "overwrite" the doc comment on the trait definition inside the implementation. That'd be even better :)

iroh-relay/src/relay_map.rs
ramfox: agreed on all your points about the `(RelayUrl, u16)` impls, I'm removing them. It looks like the original code was only used in `iroh-doctor` and can easily be adjusted.

iroh-relay/src/relay_map.rs
ramfox: hmmm when I run `cargo doc` it's showing up as a replacement doc comment already. I've tried both above the impl and above the fn and it shows the same thing.

I'll move it though.

flub (approved):
The "Breaking Changes" section in the PR description needs updating, but otherwise looks good!

iroh-relay/src/relay_map.rs
matheus23: T*I*L!

matheus23 (approved):

PR #3293, opened 2025-05-06
feat(iroh): add `net-report` method on the `iroh::Endpoint` that returns a `Watchable<Report>`

ramfox:
## Description

This change allow a user to get updates on the latest `net-report::Report` run.

## Change checklist

- [x] Self-review.


iroh/src/endpoint.rs
```diff
@@ -992,6 +993,39 @@ impl Endpoint {
...
+    /// use futures_lite::StreamExt;
+    /// use iroh::Endpoint;
+    ///
+    /// # let rt = tokio::runtime::Builder::new_current_thread().enable_all().build().unwrap();
+    /// # rt.block_on(async move {
+    /// let ep = Endpoint::builder().bind().await.unwrap();
+    /// let _report = ep.net_report().initialized().await.unwrap();
+    /// # });
+    /// ```
+    pub fn net_report(&self) -> Watcher<Option<Arc<Report>>> {
```
flub, line 1025: It's a great doc comment.  Though I'd like to make this `#[doc(hidden)]` as I would like to avoid having this in the 1.0 API.  I think it is fairly commonly accepted that doc(hidden) items are not part of a public API.  If we need we could also hide it behind an `experimental_watchable_net_report` cargo feature.

If we hide it maybe the example should be a test instead?  Wait, maybe the example should be turned into a test anyway.  Independent of whether it stays as an example or not.

flub (commented):
The implementation is great, I'm only weary of exposing this to users.

matheus23 (commented):
The implementation looks totally fine.

We've thought about moving to a more trickle-like construction for net report at some point, though, so we should have the foresight of not exposing this as part of the "actual" API (e.g. by marking it `#[doc(hidden)]` as flub suggested).

dignifiedquire (approved):

flub (approved):
*Maybe* it's still worth updating the (now hidden) doc comment to say this is not part of the 1.0 API.  In any case I think if we do end up changing this, the existing call may have to stay around to please the linters.  It'll just be pending forever.  So it's probably fine as it is now as well.

PR #3297, opened 2025-05-09
fix(iroh-dns-server): Fixes for packet expiry

Frando:
## Description

Based on #3295 

I noticed in the debug logs of our staging server that some entries in the expiry table were never deleted, and thus they were processed again and again on each expiry tick.

This PR has the following changes:

* Bugfix: When upserting a packet, we were not properly clearing the *old* expiry entry! [This line](https://github.com/n0-computer/iroh/blob/f3c9af35542a7f2e00303090d1ca5273f86bc375/iroh-dns-server/src/store/signed_packets.rs#L148) is wrong: It should be `existing.timestamp()`, because that's the entry we want to remove, not the (not-yet-existing) entry for the timestamp of the *new* packet.
* Improve type handling and debug formatting around expiry timestamps, to make logs better readable and be less error prone to bad type conversions
* When an expiry entry is being processed, and the corresponding packet does not exist anymore, the expiry entry is deleted

## Notes & open questions

We need tests around expiry. This should've been caught earlier. There's currently only a [single test](https://github.com/n0-computer/iroh/blob/f3c9af35542a7f2e00303090d1ca5273f86bc375/iroh-dns-server/src/lib.rs#L190) around eviction, added in the PR that added eviction (#2997). But that test only tests that expiry works at all. There's no test currently that verifies if entries in the expiry table are removed when a packet is upserted, which is the test that would fail without this PR.

## Change checklist

- [x] Self-review.
- Tests if relevant: Would be relevant, but needs to come later.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2335,6 +2335,7 @@ dependencies = [
  "hickory-resolver",
  "hickory-server",
  "http 1.3.1",
+ "humantime",
  "humantime-serde",
  "iroh",
  "iroh-metrics",
--- a/iroh-dns-server/Cargo.toml
+++ b/iroh-dns-server/Cargo.toml
@@ -27,6 +27,7 @@ dirs-next = "2.0.0"
 governor = "0.8"
 hickory-server = { version = "0.25.1", features = ["https-ring"] }
 http = "1.0.0"
+humantime = "2.2.0"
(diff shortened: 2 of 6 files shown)


matheus23 (approved):
Yeah I don't mind the extra allocations for `Timestamp`. I think this is better.

PR #3306, opened 2025-05-13
iroh(deps): Dedupe lru and webpki

webbeef:
## Description

Dedupe a couple of crates, because why not.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2251,7 +2251,8 @@ dependencies = [
  "reqwest",
  "ring",
  "rustls",
- "rustls-webpki 0.102.8",
+ "rustls-pki-types",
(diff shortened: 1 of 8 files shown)


flub (approved):
thanks!

flub (changes requested):
Ugh, seems like the webpki upgrade isn't so simple.  CI is unhappy, can you check this out?

flub:
Updating deps, always such fun... I'm afraid you're not done yet.

flub:
Getting close!  Only minimal crate versions are unhappy.

flub (approved):
Nice!  Thanks you so much for persevering for what you thought would be a quick simple improvement...

PR #3309, opened 2025-05-14
refactor: use a separate queue for inbound disco packets from relays

Frando:
## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


iroh/src/magicsock/relay_actor.rs
```diff
@@ -862,6 +903,17 @@ impl RelayActor {
                     debug!("shutting down");
                     break;
                 }
+                Some(message) = self.relay_disco_recv_rx.recv() => {
+                    self.msock.
+                        handle_disco_message(
```
flub, line 907: I'm a bit worried doing this *in* the RelayActor.  This actor needs to be able to move packets **fast**.  I would like it to avoid doing any processing like decrypting messages and taking a whole bunch of locks (what this call does).

Frando: Agreed. A separate task that's spawned from the magicsock maybe that does just that?

iroh/src/magicsock/relay_actor.rs
```diff
@@ -651,8 +667,27 @@ impl ActiveRelayActor {
...
+                                warn!("Received relay disco message from connection for {}, but with message from {}", remote_node_id.fmt_short(), source.fmt_short());
+                            }
+                            let message = RelayDiscoMessage {
+                                source,
+                                sealed_box,
+                                relay_url: datagram.url.clone(),
+                                relay_remote_node_id: datagram.src,
+                            };
+                            if let Err(err) = self.relay_disco_recv.try_send(message) {
+                                warn!("Dropping received relay disco packet: {err:#}");
```
flub, line 683: I would be tempted to push this `relay_disco_recv` all the way into `ActiveRelayActor` and then it can queue this message and stop reading from the relay channel instead of having to drop it.

Frando: This is already in the ActiveRelayActor! So all it would take is making `ActiveRelayActor::handle_relay_msg` an `async` fn, and using `send(message).await` here I think? It is already only called from async fns, and that should then add backpressure I think.

Frando: Pushed that change. Tests seem fine with it!

iroh/src/magicsock/relay_actor.rs
```diff
@@ -605,7 +618,7 @@ impl ActiveRelayActor {
                     };
                     match msg {
                         Ok(msg) => {
-                            self.handle_relay_msg(msg, &mut state);
+                            self.handle_relay_msg(msg, &mut state).await;
```
flub, line 621: We can not make this async.  Actors can not do async stuff in their main loop.  We violate this in plenty of places, but this actor has finally been cleaned up.  We should not regress this.

Frando: Right. But how would you apply backpressure then?

iroh/src/magicsock/relay_actor.rs
flub: Ah, github was giving me the wrong context and I thought this was the RelayActor...

Conceptually the "queue the message" is the same as my "we can't do async stuf" comment above.  If the `self.relay_disco_recv.try_send` fails somehow the message needs to be stored on the actor and then the main loop should stop polling for further messages from the relay until it was finally able to send this.  The RelayActor and ActiveRelayActor already go to some length to do this and it has worked very well.  Exactly where the state should be stored and how to keep yet another case of this readable and maintainable is another question.  E.g. in the RelayActor stores a `datagram_send_fut` "sending future" (the `MaybeFuture`) and if that's set then don't poll something.  The ActiveRelayActor has various versions of the run loops.  That may not work so well if it also has to do this.

iroh/src/magicsock/relay_actor.rs
Frando: Removed that commit again.

iroh/src/magicsock.rs
```diff
@@ -1826,6 +1801,23 @@ impl Handle {
...
+                    msock.handle_disco_message(
+                        message.source,
+                        &message.sealed_box,
+                        DiscoMessageSource::Relay {
+                            url: message.relay_url,
+                            key: message.relay_remote_node_id,
+                        },
+                    );
+                }
+                debug!("relay-disco-recv actor closed");
```
flub, line 1817: nit: "actor closed" is sufficient since you already have an info span with the actor name.  But I don't mind if you do this either.

iroh/src/magicsock/relay_actor.rs
flub: > Right. But how would you apply backpressure then?

For completeness answering that here as well (it is mentioned above): by keeping the blocked message around in the actor state *somehow* and not polling for new messages from the relay while you have this pending message.

Of course in this case it means you're backpressuring the entire relay TCP stream while you only really want to backpressure disco messages.  But to really do something about this you'd have to make two tcp streams to the relay and... lots of things we don't want to do.

iroh/src/magicsock/relay_actor.rs
```diff
@@ -159,6 +162,20 @@ struct ActiveRelayActor {
     /// Token indicating the [`ActiveRelayActor`] should stop.
     stop_token: CancellationToken,
     metrics: Arc<MagicsockMetrics>,
+    /// Received relay packets that could not yet be forwarded to the magicsocket.
+    pending_received: Option<PendingRecv>,
```
flub, line 166: Shouldn't this live on `ConnectedRelayState`?

iroh/src/magicsock/relay_actor.rs
```diff
@@ -788,6 +789,105 @@ impl ActiveRelayActor {
...
+                return;
+            };
+            Some(permit)
+        }
+    };
+    // We now take the inner value by value. it is cancellation safe here because
+    // no further `await`s occur after here.
+    // The unwrap is guaranteed to be safe because we checked above that it is not none.
+    #[allow(clippy::unwrap_used, reason = "checked above")]
+    let pending = maybe_pending.take().unwrap();
```
flub, line 830: We use `.expect()` for cases like this, I thought `.unwrap()` was denied by some linter even?

Frando:
I pushed two commits:
* A fix to the changes to the RelayDatagramRecvQueue: We need to store a list of send wakers, because this can be sent to from multiple ActiveRelayActors
* Some cleanups to the new code added to the active relay actor to make this a bit nicer

iroh/src/magicsock/relay_actor.rs
Frando: I moved this into a new struct together with the queue senders in the latest commit, but still kept it on the `ActiveRelayActor`. My reasoning: If the connection to a relay breaks, but we still have pending received items queued, we can still try pushing those to the magicsock.

dignifiedquire:
closing, out of date and will be even more so once multipath lands

PR #3314, opened 2025-05-16
[WIP]: reworkt net_report

dignifiedquire:
- use more watchables
- drop STUN, ICMP and hairpin, they are not actually needed anymore
- restructure report generation
- more will come..


iroh/src/net_report/ip_mapped_addrs.rs
flub, line 1: :heart: the visibility qualifiers

iroh-relay/src/main.rs
```diff
@@ -174,7 +164,7 @@ struct Config {
     key_cache_capacity: Option<usize>,
     /// Access control for relaying connections.
     ///
-    /// This controls which nodes are allowed to relay connections, other endpoints, like STUN are not controlled by this.
+    /// This controls which nodes are allowed to relay connections, other endpoints, are not controlled by this.
```
ramfox:
```suggestion
    /// This controls which nodes are allowed to relay connections, other endpoints are not controlled by this.
```

iroh/src/net_report/probes.rs
```diff
@@ -247,73 +132,43 @@ impl fmt::Display for ProbeSet {
...
-            plan.add_if_enabled(quic_ipv4_probes);
-            plan.add_if_enabled(quic_ipv6_probes);
+            plan.add_if_enabled(protocols, quic_ipv4_probes);
+            plan.add_if_enabled(protocols, quic_ipv6_probes);
 
-            // The HTTP and ICMP probes only start after the STUN probes have had a chance.
-            let mut https_probes = ProbeSet::new(ProbeProto::Https);
-            let mut icmp_v4_probes = ProbeSet::new(ProbeProto::IcmpV4);
-            let mut icmp_v6_probes = ProbeSet::new(ProbeProto::IcmpV6);
+            // The HTTP probes only start after the QUAD probes have had a chance.
```
ramfox:
```suggestion
            // The HTTP probes only start after the QAD probes have had a chance.
```

iroh/src/net_report/report.rs
```diff
@@ -0,0 +1,204 @@
...
+    /// an IPv4 packet was able to be sent
+    pub ipv4_can_send: bool,
+    /// could bind a socket to ::1
+    pub os_has_ipv6: bool,
+    /// Whether STUN results depend on which STUN server you're talking to (on IPv4).
+    pub mapping_varies_by_dest_ip: Option<bool>,
+    /// Whether STUN results depend on which STUN server you're talking to (on IPv6).
+    ///
+    /// Note that we don't really expect this to happen and are merely logging this if
+    /// detecting rather than using it.  For now.
```
ramfox:
```suggestion
    /// Whether the reported public address differs when probing different servers (on IPv4).
    pub mapping_varies_by_dest_ip: Option<bool>,
    /// Whether the reported public address differs when probing different servers (on IPv6).
    ///
    /// Note that we don't really expect this to happen and are merely logging this if
    /// detecting rather than using it.  For now.
```

iroh/src/net_report/probes.rs
```diff
@@ -390,23 +216,22 @@ impl ProbePlan {
...
 
-        // The first time we need add probes after the STUN we record this delay, so that
+        // The first time we need to add probes after the STUN we record this delay, so that
         // further relay servers can reuse this delay.
         let mut max_stun_delay: Option<Duration> = None;
 
-        let had_stun_ipv4 = !last_report.relay_v4_latency.is_empty();
-        let had_stun_ipv6 = !last_report.relay_v6_latency.is_empty();
+        let had_stun_ipv4 = !last_report.relay_latency.ipv4().is_empty();
+        let had_stun_ipv6 = !last_report.relay_latency.ipv6().is_empty();
```
ramfox: still referencing stun here

iroh/src/net_report.rs
```diff
@@ -85,116 +74,7 @@ const FULL_REPORT_INTERVAL: Duration = Duration::from_secs(5 * 60);
...
-
-    fn is_empty(&self) -> bool {
-        self.0.is_empty()
-    }
-
-    fn get(&self, url: &RelayUrl) -> Option<Duration> {
-        self.0.get(url).copied()
-    }
-}
+const ENOUGH_NODES: usize = 3;
```
ramfox, line 101:
```suggestion
/// The maximum number of nodes we need to probe successfully, before the report is considered finished.
///
/// If the number of available nodes is less than 3, than we need to hear from all of the available relay servers
/// before the report is considered finished.
const ENOUGH_NODES: usize = 3;
```

Do we refer to these as "nodes" and not "relay servers" here for a specific reason? I'd assume it's because we can potentially probe anything that has these protocols, but it's a bit confusing since we use the word "nodes" in most places in the code to refer to peers in our network.

ramfox:
Okay there are two things that I'm having trouble following.

The first is in `get_report`. I think there is a possibility that the loop in there can be infinite, if we aren't doing a full report.

If we are NOT doing a full report, and our last report had latencies, we return an empty probe plan in `ProbePlan::with_last_report` & we don't run the captive portal probe either. As far as I can tell, this means that we never send any `ProbeFinished` responses on the `probe_tx` receiver in `get_report`.

If we never get any responses in `probe_tx` in the `loop + tokio::select!`, we don't seem to ever run `self.have_enough_reports`. This means that setting the `timeout_fut` never happens. The `timeout_fut` arm of `tokio::select!` is the only arm where we break the`get_report` loop.

Hopefully i'm just mis-reading something! lines of code are here: https://github.com/n0-computer/iroh/blob/303dce9c744f21fa613d211d65edecd59f9af072/iroh/src/net_report.rs#L326-L378

My second question is about the `home_relay` and the `quinn::Connection`s we keep alive for qad. I *think* it's possible (but really unlikely) that the relay we are connected to for qad can be different from the relay that we recommend for the home relay. For the maintained connections, we keep around the first qad conn that comes back with a response. I can't imagine how this wouldn't be the one with the lowest latency, but since we only rely on latency to determine with relay we consider the `preferred_relay`, couldn't it techincally be possible for us to recommend a different relay than the one we are connected to? Should we refer to the `QadConn` relays first when trying to determine the preferred relay, before relaying on latencies if the `QadConn`s don't exist?

Anyways, those were the only two pieces that raised questions for me, everything else looked in enough order that we should probs just merge!

iroh/README.md
```diff
@@ -4,9 +4,9 @@ Iroh is a library to establish direct connectivity between peers.
 It's built on peer-to-peer [QUIC](https://en.wikipedia.org/wiki/QUIC) using both relays and holepunching.
 The main structure for connection is the `Endpoint` entrypoint.
 
-Peer to peer connectivity is established with the help of a _relay server_. The relay server provides Session Traversal Utilities for NAT [(STUN)](https://en.wikipedia.org/wiki/STUN) for the peers. If no direct connection can be established, the connection is relayed via the server.
+Peer to peer connectivity is established with the help of a _relay server_. The relay server provides QUIC NAT traversal (QAD) for the peers. If no direct connection can be established, the connection is relayed via the server.
```
Frando:
```suggestion
Peer to peer connectivity is established with the help of a _relay server_. The relay server provides [QUIC Address Discovery](https://www.ietf.org/archive/id/draft-ietf-quic-address-discovery-00.html) (QAD) and hole-punching assistance for the peers. If no direct connection can be established, the connection is relayed via the server.
```

iroh/src/endpoint.rs
```diff
@@ -1045,7 +1045,7 @@ impl Endpoint {
     /// # });
     /// ```
     ///
-    /// [STUN]: https://en.wikipedia.org/wiki/STUN
+    /// [QAD]: https://www.ietf.org/archive/id/draft-seemann-quic-nat-traversal-02.html
```
```suggestion
    /// [QAD]: https://www.ietf.org/archive/id/draft-ietf-quic-address-discovery-00.html
```
or did you mean to link to the QUIC NAT traversal draft? That one isn't usually abbreviated as QAD though.

iroh/src/lib.rs
```diff
@@ -232,7 +232,7 @@
 //! [bi-directional streams]: crate::endpoint::Connection::open_bi
 //! [hole punching]: https://en.wikipedia.org/wiki/Hole_punching_(networking)
 //! [socket addresses]: https://doc.rust-lang.org/stable/std/net/enum.SocketAddr.html
-//! [STUN]: https://en.wikipedia.org/wiki/STUN
+//! [QAD]: https://www.ietf.org/archive/id/draft-seemann-quic-nat-traversal-02.html
```
Frando: nit: qad link, see above

iroh/src/endpoint.rs
dignifiedquire: lol, I am an idiot, thanks

dignifiedquire:
> If we never get any responses in `probe_tx` in the `loop + tokio::select!`, we don't seem to ever run `self.have_enough_reports`. This means that setting the `timeout_fut` never happens. The `timeout_fut` arm of `tokio::select!` is the only arm where we break the`get_report` loop.

thanks for checking this, the reason we exit is because `reporgen::Actor`  gets dropped, and the receiver returns `None` where we break out of the loop

dignifiedquire:
> For the maintained connections, we keep around the first qad conn that comes back with a response. I can't imagine how this wouldn't be the one with the lowest latency, but since we only rely on latency to determine with relay we consider the `preferred_relay`, couldn't it techincally be possible for us to recommend a different relay than the one we are connected to? Should we refer to the `QadConn` relays first when trying to determine the preferred relay, before relaying on latencies if the `QadConn`s don't exist?


Hmm, it is certainly possible for the two to differ, and I could see us optimizing the differences in the future, but for now I think this should "just work out", as the latencies we will get from these QADs will be the ones returned in every run, and subsequently used for evaluation

PR #3316, opened 2025-05-16
feat: add nix packages and dev shell

elsirion:
This commit adds packages for the iroh-dns-server and iroh-relay binaries as well as a rust dev shell for development. To make the packaging work better (less redundant versions in ni flake) I also made the crate versions a workspace property that is inherited.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


flub:
Hi, for folks who don't use nix at all could you add some context?  Does having this make a big difference to nix users?  Is is impossible to do this out-of-project?  Once concern I have for example is that this looks like it needs to be kept up-to-date, which is not great if the maintainers don't use it.

elsirion:
The benefit for nix users over an out-of-repo version would be that they will see that a packaged version is available and can get started deploying/developing without much friction. Tbf that's a rather minimal improvement since packaging turned out to be quite simple, for more complicated to build projects it can save a lot of headaches.

PR #3319, opened 2025-05-20
feat(iroh): Allow protocols to gracefully shutdown connections

Frando:
## Description

Currently, when `Router::shutdown` is called, we shutdown the endpoint and await the futures returned from `ProtocolHandler::shutdown` concurrently. This means that `ProtocolHandler::shutdown` cannot actually gracefully close their connections, because `Endpoint::close` is already called at the same time, which closes all connections immediately with an error code `0`.

This PR changes the shutdown order such that `ProtocolHandler::shutdown` is called and awaited first, and `Endpoint::close` afterwards.

I also added a test that fails without this change and passes now.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/protocol.rs
+++ b/iroh/src/protocol.rs
@@ -331,7 +331,13 @@ impl RouterBuilder {
             // Finally, we abort the remaining accept tasks. This should be a noop because we already cancelled
             // the futures above.
             tracing::debug!("Shutting down remaining tasks");
-            join_set.shutdown().await;
+            join_set.abort_all();
+            while let Some(res) = join_set.join_next().await {
+                match res {
+                    Err(err) if err.is_panic() => error!("Task panicked: {err:?}"),
+                    _ => {}
+                }
+            }
         };
         let task = task::spawn(run_loop_fut);
         let task = AbortOnDropHandle::new(task);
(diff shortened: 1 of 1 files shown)


matheus23 (approved):

iroh/src/protocol.rs
```diff
@@ -495,4 +497,56 @@ mod tests {
...
+                    tokio::time::sleep(Duration::from_millis(100)).await;
+                    let mut connections = this.connections.lock().expect("poisoned");
+                    for conn in connections.drain(..) {
+                        conn.close(42u32.into(), b"shutdown");
+                    }
+                })
+            }
+        }
+
+        let endpoint = Endpoint::builder().bind().await?;
```
matheus23, line 531: I thought it might make sense for some tests to set `.relay_mode(RelayMode::Disabled)`, if they don't need it, since that should speed up getting a `node_addr` a bit.

Frando: Changed that

Frando:
Pushed another commit to not disregard panics in ProtocolHandler tasks silently.

dignifiedquire (approved):

PR #3320, opened 2025-05-20
feat(iroh)!: make ProtocolHandler use async functions

Frando:
## Description

This changes the `ProtocolHandler` trait so that the functions return `impl Future` instead of `BoxFuture`, which means implementors can simply implement the trait with `async fn`. We don't even need a `'static` bound on these futures, so you can use `&self` freely without cloning and moving.

Internally, we still box the futures return by `accept`, `on_connecting` and `shutdown`. So performance wise this change should not have any effects. But the boxing is hidden from the user and implementing the trait now feels a lot more normal and simpler.

## Breaking Changes

* `iroh::protocol::ProtocolHandler` methods now return `impl Future` instead of `BoxFuture`. You can simply remove the `Box::pin(async move {})` from the implementations and instead implement the methods as `async fn`. See the updated documentation for the `iroh::protocol` module for an example.
* `iroh::protocol::ProtocolHandler` is no longer dyn-compatible. If you need a dyn-compatible version, you need to build your own dyn-compatible wrapper trait. See the (non-public) `DynProtocolHandler` in `iroh::protocol` as an example.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,18 +97,16 @@ let router = Router::builder(endpoint)
 struct Echo;
 
 impl ProtocolHandler for Echo {
-    fn accept(&self, connection: Connection) -> BoxedFuture<Result<()>> {
-        Box::pin(async move {
-            let (mut send, mut recv) = connection.accept_bi().await?;
+    async fn accept(&self, connection: Connection) -> Result<()> {
+        let (mut send, mut recv) = connection.accept_bi().await?;
 
-            // Echo any bytes received back directly.
-            let bytes_sent = tokio::io::copy(&mut recv, &mut send).await?;
+        // Echo any bytes received back directly.
+        let bytes_sent = tokio::io::copy(&mut recv, &mut send).await?;
 
-            send.finish()?;
-            connection.closed().await;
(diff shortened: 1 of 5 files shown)


rklaehn:
Thanks for trying this. My experience explaining `Pin<Box<dyn Future...>>` to very well meaning, almost enthusiastic, rust newbies, was not great. I felt a bit embarrassed for rust.

So the only overhead I can see is that if you for some reason already have a `Pin<Box<dyn Future ...>>`, I will have to return that as a impl Future, and it will be boxed again. But given that this interface is very high level, not the hot part of a low level function, I think it does not matter at all.

So I am very much in favour of this before 1.0.

matheus23 (approved):
This makes me happy. `DynProtocolHandler` is fully internal as far as I can tell, so I think this is good!

If a user ends up double-boxing their accept futures and that indeed becomes a problem for them, well at that point they're such a power-user that switching away from the `Router` shouldn't be a problem for them IMO.

PR #3322, opened 2025-05-23
Fix listen example: empty local_addrs

juchiast:
## Description

I tried running this example and `local_addrs` is always empty.


flub (approved):
Thanks!  Indeed, that original code would be extremely likely to always have empty local addresses.

flub:
Looked like the CI runners had some network hiccups during that run.  Let's try again.

PR #3326, opened 2025-05-27
feat!: Add a `Resolver` trait to abstract over DNS resolvers

Frando:
## Description

This adds a `Resolver` trait to abstract over DNS resolution. It contains methods to resolve IPv4 or IPv6 addresses, and TXT records. Because the trait needs to be dyn-compatible for our usage (we don't want to have a generic for the DNS resolver on the Endpoint), all methods return boxed futures that contain boxed iterators. 

The `DnsResolver` struct is changed to contain an `Arc<dyn Resolver>` instead of being hardcoded to `hickory_resolver::TokioResolver`. Iroh ships an implementation of `Resolver` for `hickory_resolver::TokioResolver`, but does not have any hickory types in the public API anymore. 

Users can implement the `Resolver` trait on whatever struct to use a completely custom DNS resolver.

## Breaking Changes

Changes in `iroh_relay::dns` (reexported from `iroh::dns`):
* `DnsResolver::new` now takes an `impl Resolver`. To create a default resolver, use `DnsResolver::default` or the equivalent `DnsResolver::new_with_system_defaults`
*  `DnsResolver::lookup_txt` now returns `impl Iterator<Item = TxtRecord>`
* `TxtLookup` and `TXT` are removed.

Changes in `iroh_relay::node_info` (reexported from `iroh::discovery`):
* `NodeInfo::from_txt_lookup` now takes `name: String, lookup: impl Iterator<Item = crate::dns::TxtRecord>`

## Notes & open questions

We could now put the implementation of `Resolver` for `hickory_resolver::TokioResolver` behind a (default) feature flag to not have to hard-depend on a specific version of hickory_resolver anymore during iroh 1.0 - I think? Or maybe we'd still have to newtype it. Will need to think about this some more.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-dns-server/examples/resolve.rs
+++ b/iroh-dns-server/examples/resolve.rs
@@ -54,7 +54,7 @@ async fn main() -> anyhow::Result<()> {
         DnsResolver::with_nameserver(addr)
     } else {
         match args.env {
-            Env::Staging | Env::Prod => DnsResolver::new(),
+            Env::Staging | Env::Prod => DnsResolver::default(),
             Env::Dev => {
                 DnsResolver::with_nameserver(DEV_DNS_SERVER.parse().expect("valid address"))
             }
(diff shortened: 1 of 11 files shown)


dignifiedquire:
I am not convinced this amount of boxing is really worth it, I'd rather not have a trait, and simply wrap hickory types for 1.0 stability

Frando:
Closing in favor of #3473

PR #3327, opened 2025-05-27
refactor(iroh)!: Add `IntoDiscovery` trait

Frando:
## Description

Discovery services may want to access the endpoint they are attached to for various reasons. 
* Currently, we pass a reference to the `&Endpoint` in `Discovery::resolve`, but not in `Discovery::publish`. 
* All pkarr-based publishers need access to the endpoint's secret key for signing packets. For this reason, we allow passing a closure to `endpoint::Builder::add_discovery` that gets the secret key as an argument
* The pkarr publisher *should* get hold of the endpoint's DNS resolver to resolve the pkarr relay URL (but doesn't atm, which is a bug - it uses reqwest's default resolver, always)

So this PR changes the construction by adding an `IntoDiscovery` trait with a single method `into_discovery(self, context: &DiscoveryContext) -> impl Discovery`. And `endpoint::Builder::add_discovery` now takes any  `IntoDiscovery`. Through this, discovery service construction is a two-step process where users create a builder and set any options available, and then during endpoint building we call `into_discovery` and pass a discovery context.

We can't pass a full endpoint to discovery services, because if they'd clone it, the Endpoint could never be dropped due to a reference cycle: `Endpoint -> Magicsock -> Box<dyn Discovery> -> Endpoint`. Therefore, this PR adds a `DiscoveryContext` struct that contains the endpoint's secret key and DNS resolver. We can add more things later, if needed.

There's also a blanket impl  of `IntoDiscovery` for any `T: Discovery`, so discovery services that do not need an endpoint can skip the builder and don't have to implement `IntoDiscovery` manually at all, users can just pass the constructed discovery struct to `Endpoint::add_discovery`.

This means *writing* discovery services is a little bit more involved, as you have to create a builder and implement a second trait. However, writing discovery services is a rather rare task, and with good docs the additional work is very minor. On the upside, this creates a clean way for discovery services to access the endpoint they are mounted on, which also opens the door for discovery services using iroh itself.

## Breaking Changes

Changed:
* `iroh::endpoint::Builder::add_discovery` now takes an `impl iroh::discovery::IntoDiscovery` argument instead of a closure that returns a `Discovery`. You can  implement that on a builder struct, and any `T: Discovery`  has an auto-impl of `IntoDiscovery`.
* `iroh::discovery::Discovery::resolve` no longer gets a `&Endpoint` argument. If you need an endpoint in your discovery service, add a builder struct and implement `IntoDiscovery` for it, which gives you an endpoint that you can clone into your service
*  Removed `iroh::discovery::dns::DnsDiscovery::new`, use `DnsDiscovery::builder` instead
*   Removed `iroh::discovery::pkarr::PkarrPublisher::new`, use `PkarrPublisher::builder` instead 
*   Removed `iroh::discovery::pkarr::PkarrPublisher::with_options`, use `PkarrPublisher::builder` instead
*   Removed `iroh::discovery::pkarr::PkarrResolver::new`, use `PkarrResolver::builder` instead 
* `iroh::discovery::pkarr::PkarrPublisher::n0_dns` now takes no arguments and returns a `PkarrPublisherBuilder`. The secret key is set on `PkarrPublisherBuilder::build` instead.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -951,7 +951,7 @@ impl Endpoint {
     /// with a [`NodeAddr`] that only contains a relay URL, but no direct addresses,
     /// as there are no APIs for directly using sockets in browsers.
     #[cfg(wasm_browser)]
-    pub fn node_addr(&self) -> impl n0_watcher::Watcher<Value = Option<NodeAddr>> + '_ {
+    pub fn node_addr(&self) -> impl n0_watcher::Watcher<Value = Option<NodeAddr>> {
         // In browsers, there will never be any direct addresses, so we wait
         // for the home relay instead. This makes the `NodeAddr` have *some* way
         // of connecting to us.
@@ -2460,7 +2460,7 @@ mod tests {
         Ok(())
     }
 
-    #[cfg_attr(target_os = "windows", ignore = "flaky")]
+    #[cfg_attr(windows, ignore = "flaky")]
     #[tokio::test]
     #[traced_test]
     async fn endpoint_relay_connect_loop() -> Result {
--- a/.github/workflows/project_sync.yaml
+++ b/.github/workflows/project_sync.yaml
@@ -13,7 +13,7 @@ jobs:
     name: Add to project board
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/create-github-app-token@v1
+      - uses: actions/create-github-app-token@v2
         id: app-token
         with:
           app-id: ${{ secrets.N0_BOT_APP_ID }}


dignifiedquire:
I really like this approach 👍 this seems like a good solution to the issue

iroh-relay/Cargo.toml
```diff
@@ -19,7 +19,7 @@ crate-type = ["lib", "cdylib"]
 workspace = true
 
 [dependencies]
-anyhow = { version = "1" }
+anyhow = { version = "1.0.98" }
```
Frando, line 22: Needed for https://docs.rs/anyhow/latest/anyhow/struct.Error.html#method.into_boxed_dyn_error

iroh/src/discovery.rs
```diff
@@ -130,6 +132,45 @@ pub mod mdns;
...
 pub mod static_provider;
 
+/// Trait for structs that can be converted into [`Discovery`].
+///
+/// This trait is implemented on builders for discovery services. Any type that implements this
+/// trait can be added as a discovery service in [`Builder::add_discovery`].
+///
+/// Any type that implements [`Discovery`] also implements [`IntoDiscovery`].
+///
+/// [`Builder::add_discovery`]: crate::endpoint::Builder::add_discovery
```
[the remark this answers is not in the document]
Frando: Added the additional lines!

iroh/src/discovery.rs
```diff
@@ -130,6 +132,45 @@ pub mod mdns;
...
+pub trait IntoDiscovery: Send + Sync + 'static {
+    /// Turns this struct into a boxed [`Discovery`].
+    ///
+    /// The discovery service will be used by passed [`Endpoint`]. If a discovery service needs
+    /// an endpoint or information from it like its secret key, it can access or clone the endpoint
+    /// from here.
+    ///
+    /// If an error is returned, building the endpoint will fail with this error.
```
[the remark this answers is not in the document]
Frando: Good point. No more endpoint for discovery services. They get a `DiscoveryContext` now.

Frando:
> I really like this, but I'm a little worried that this might lead to reference counter cycles.
> 
> If you get the `&Endpoint` in `into_discovery`, and decide to clone it and store it, I think you've got a cycle, right? AFAIK, this won't error out in rust, it'll just leak memory.

This is true, and we did not find a good way around this.

So for now discovery services may not use the endpoint. Instead, they now get a `DiscoveryContext` that gives access to the endpoint's node id, secret key, and DNS resolver.

iroh/src/endpoint.rs
```diff
@@ -132,7 +131,7 @@ pub struct Builder {
     transport_config: quinn::TransportConfig,
     keylog: bool,
     #[debug(skip)]
-    discovery: Vec<DiscoveryBuilder>,
+    discovery: Vec<Box<dyn DynIntoDiscovery>>,
```
matheus23, line 109: Sorry for the drive-by comment, but looking at the `#[debug(skip)]` above: We could consider making `DynIntoDiscovery: std::fmt::Debug`, so that we don't have to do this (and get debug output if needed in the future).

Also, it's a bound that's easier to remove than to add in the future.

But that's really just a nit at best.

Frando: Changed that!

PR #3330, opened 2025-06-02
feat(iroh)!: remove deprecated x509 libp2p TLS authentication

dignifiedquire:
## Description

As previously announced, we are fully transitioning to using Raw Public Keys for the TLS authentication, so this old method is now removed, in preparation of cleanup & stabilization in 1.0.

## Breaking Changes

- `x509` authentication mechanism removed


matheus23 (approved):
LGTM, seems fairly straightforward.

Any reason to keep the enum and config option? I'd say we remove these options for 1.0. Should we want to add them back we can do that (and we'll have the commits to refer back to how we did it if we want to).

dignifiedquire:
> Any reason to keep the enum and config option? I'd say we remove these options for 1.0.

I wasn't sure in either direction, you reckon I should just drop it?

matheus23:
yeah

matheus23:
I guess that'd mean we're not only removing a deprecated function, but also a non-deprecated one.
That said, I expect no one actually called that function. And even if they did, then that function call would be unnecessary and can just be removed (and we can say so in the breaking changes, even if slightly confusing).

dignifiedquire:
cleaned up more

matheus23:
Can you expand a bit on the breaking changes? I.e. methods that are removed and what to do if you used to call them before.

DXist:
Are there any documented design decisions about dropping x509 support?

Are iroh users expected to roll protocol-level identification & authorization instead of doing it on TLS layer via x509?

matheus23:
Here's a blog post we wrote about this @DXist: https://www.iroh.computer/blog/iroh-0-34-0-raw-public-keys

All iroh connections stay fully authenticated once they've been established. Nothing about that changes, only the way its implemented.

PR #3331, opened 2025-06-02
feat(iroh-relay)!: Implement new handshake protocol, refactor frame types

matheus23:
## Description

This PR changes the iroh relay handshake and relaying protocol.
The main goals are:
- Actually authenticate the client by verifying a signed challenge (previously only the client info was signed)
  - We also allow using TLS extracted keying material as the subject to sign and sending it in an HTTP header, if that is available. When this works (it does so most of the time), it's 1RTT faster.
- Use a better encoding for messages (we now use QUIC varints for frame type, use websocket framing for the protocol)
- Remove outdated and unused frames (KeepAlive, NotePreferred)

There's also some secondary changes:
- Conceptually, the relay handshake and relaying protocols are now split: There's a `protos::handshake` module for the handshake, and a `protos::relay` for the send/recv protocol.
- I've refactored the streams we have in iroh-relay. We now use this stack:
  - On the client, we establish TLS streams using `MaybeTlsStream<ProxyStream>`. `MaybeTlsStream` gives us the TLS handshake and `ProxyStream` gives us HTTPS proxy support.
  - On the server we establish rate-limited TLS streams using `RateLimit<MaybeTlsStream>`.
  - On top of these two, we stack `WsBytesFramed`, which internally uses `tokio_websockets::WebSocketStream` natively and on the server and `ws_stream_wasm::WsStream` in browsers. This wrapper does the websocket framing and translates opaque `Bytes` messages into websocket bytes frames.
  - This `WsBytesFramed` is then directly used with the `handshake` protocol on the server and client side.
  - For the `send_recv` protocol, we then upgrade the server side to `RelayedStream` and the client side to the `Conn` type.
- Instead of having `Frame`, `ReceivedMessage` and `SendMessage` as types, we now only have `RelayToClientMsg` and `ClientToRelayMsg` for the send/recv protocol.

## Breaking Changes

In iroh:
- Connections to older relays don't work anymore.

In iroh-relay:
- Removed `iroh_relay::client::SendMessage` and `iroh_relay::client::ReceivedMessage` in favor of `ClientToRelayMsg` and `RelayToClientMsg` respectively.
- `impl Stream for Client` now produces `RelayToClientMsg` instead of `ReceivedMessage`
- `Client` now `impl Sink<ClientToRelayMsg>` instead of `impl Sink<SendMessage>` 
- Removed `ClientBuilder::is_prober`
- Moved `protos::relay::FrameType` to `protos::common::FrameType` and adjusted frame types to those of the current set of protocols

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


matheus23:
This is going to fail the integration tests for a while, since it'll need the production relays to be updated to the new version.

iroh/tests/integration.rs
```diff
@@ -37,8 +39,19 @@ async fn simple_node_id_based_connection_transfer() -> Result {
     std::panic::set_hook(Box::new(console_error_panic_hook::hook));
     setup_logging();
 
-    let client = Endpoint::builder().discovery_n0().bind().await?;
+    // TODO(matheus23): Replace this with actual production relays eventually
+    let relay_map = RelayMode::Custom(RelayMap::from_iter([RelayUrl::from_str(
+        "https://philipp.iroh.link.",
+    )
+    .e()?]));
```
matheus23: @Arqu This is where an *actual* set of iterim relays would come in until we cut a release and update the canary relays

iroh-relay/src/server/client.rs
```diff
@@ -537,187 +531,6 @@ impl ForwardPacketError {
     }
 }
 
-/// Rate limiter for reading from a [`RelayedStream`].
```
dignifiedquire: why no rate limiting anymore?

dignifiedquire: nvm, found it

matheus23:
Netsim report numbers for relay-only performance `intg_iroh_relay_only__1_to_1__i_get.json`:
```json
{
    "raw": [
        {
            "data_len": 1073741824,
            "elapsed": 3.365,
            "mbits": 2552.7294478454683,
            "reported_mbits": 2728.730296,
            "reported_time": 3.147960281964048
        }
    ],
    // ..
}
```

Same file on the latest run on main:
```json
{
    "raw": [
        {
            "data_len": 1073741824,
            "elapsed": 3.282,
            "mbits": 2617.2865911029858,
            "reported_mbits": 2802.46616,
            "reported_time": 3.0651341003168437
        }
    ],
    // ..
}
```

iroh-relay/src/client/tls.rs
```diff
@@ -258,7 +243,7 @@ impl MaybeTlsStreamBuilder {
             .context(ProxyConnectSnafu)?;
         task::spawn(async move {
             if let Err(err) = conn.with_upgrades().await {
-                error!("Proxy connection failed: {:?}", err);
+                tracing::error!("Proxy connection failed: {:?}", err);
```
dignifiedquire: we usually import these, instead of using the crate

iroh-relay/src/http.rs
```diff
@@ -10,49 +12,6 @@ pub(crate) const SUPPORTED_WEBSOCKET_VERSION: &str = "13";
...
-            Some(Protocol::Relay)
-        } else if header_bytes == Protocol::Websocket.upgrade_header().as_bytes() {
-            Some(Protocol::Websocket)
-        } else {
-            None
-        }
-    }
-}
+/// The HTTP header name for relay client authentication
+pub const CLIENT_AUTH_HEADER: HeaderName = HeaderName::from_static("x-iroh-relay-client-auth");
```
dignifiedquire: maybe add a version number?

iroh-relay/src/lib.rs
```diff
@@ -54,3 +54,13 @@ pub use self::{
     ping_tracker::PingTracker,
     relay_map::{RelayMap, RelayNode, RelayQuicConfig},
 };
+
+pub(crate) trait ExportKeyingMaterial {
```
dignifiedquire, line 63: please add some docs as to what this trait is for

iroh-relay/src/protos/streams.rs
```diff
@@ -0,0 +1,168 @@
...
+        Pin::new(&mut self.io).poll_close(cx).map_err(Into::into)
+    }
+}
+
+#[cfg(wasm_browser)]
+impl Sink<Bytes> for WsBytesFramed {
+    type Error = StreamError;
+
+    fn start_send(mut self: Pin<&mut Self>, bytes: Bytes) -> Result<(), Self::Error> {
+        let msg = ws_stream_wasm::WsMessage::Binary(bytes.to_vec());
```
dignifiedquire: https://docs.rs/bytes/latest/bytes/struct.Bytes.html#impl-From%3CBytes%3E-for-Vec%3Cu8%3E is more efficient

iroh-relay/src/http.rs
matheus23: Good catch. I wanted to have only one version identifier for the whole relay protocol, and that would've been provided in the webprotocol subprotocols header. But since that arrives at the same time (and may contain multiple subprotocols for the sever to choose from), it can't decide the version used in this header.

So instead this header is now versioned, and also instead of postponing this version negotiation stuff (technically we could've started doing that only with v2), I implemented that and when we'll have a v2 we'll have the option of sending any number of client auth headers, should their format change.

iroh-relay/src/protos/common.rs
```diff
@@ -0,0 +1,86 @@
+//! Common types between the [`super::handshake`] and [`super::relay`] protocols.
+//!
+//! Hosts the [`FrameType`] enum to make sure we're not accidentally reusing frame type
+//! integers for different frames.
+
+use bytes::{BufMut, Bytes};
+use quinn_proto::{coding::Codec, VarInt};
+
+/// Possible frame types during handshaking
```
flub, line 11: Could you explain in the doc comment why some numbers are skipped?

iroh-relay/src/protos/common.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// 32B src pub key + ECN byte + segment size u16 + datagrams contents
+    RelayToClientDatagrams = 11,
+    /// Sent from server to client to signal that a previous sender is no longer connected.
+    ///
+    /// That is, if A sent to B, and then if A disconnects, the server sends `FrameType::PeerGone`
+    /// to B so B can forget that a reverse path exists on that connection to get back to A
+    ///
+    /// 32B pub key of peer that's gone
+    NodeGone = 14,
+    /// Frames 9-11 concern meshing, which we have eliminated from our version of the protocol.
```
flub: 10 and 11 we use?

iroh-relay/src/protos/common.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// Messages with these frames will be ignored.
+    /// 8 byte ping payload, to be echoed back in FrameType::Pong
```
flub, line 38: lol, ok don't act on this.  But a long time ago we discussed how you like 8 byte randomness and I am partial to a sequence number.  And I just found another argument!

If you variable-length encode a sequence number you can use way fewer bytes over a long-living connection.

iroh-relay/src/protos/common.rs
```diff
@@ -0,0 +1,86 @@
...
+    Ping = 15,
+    /// 8 byte payload, the contents of ping being replied to
+    Pong = 16,
+    /// Sent from server to client to tell the client if their connection is
+    /// unhealthy somehow.
```
flub: What's in the payload?  Other frame types seem to mention this.

iroh-relay/src/protos/common.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// and how long to try total.
+    ///
+    /// Handled on the `[relay::Client]`, but currently never sent on the `[relay::Server]`
+    Restarting = 18,
+    /// The frame type was unknown.
+    ///
+    /// This frame is the result of parsing any future frame types that this implementation
+    /// does not yet understand.
+    #[num_enum(default)]
+    Unknown,
```
flub: Should you not make this u32::MAX?  Currently it will be filled in a 19 which is fine until a peer knows about 19 and starts sending it and now you have "Unknown" frames on the wire, which is a bit odd (but probably wouldn't break anything?).

iroh-relay/src/protos/streams.rs
matheus23: Good catch, thanks!

iroh-relay/src/protos/handshake.rs
```diff
@@ -0,0 +1,749 @@
...
+#[cfg(feature = "server")]
+use rand::{CryptoRng, RngCore};
+use snafu::{Backtrace, OptionExt, ResultExt, Snafu};
+
+use super::{common::FrameType, streams::BytesStreamSink};
+use crate::ExportKeyingMaterial;
+
+/// Authentication message from the client.
+#[derive(derive_more::Debug, serde::Serialize)]
+#[cfg_attr(feature = "server", derive(serde::Deserialize))]
```
flub: Is this really worth the complexity (also reading overhead) of `cfg_attr`?  On the client-side this will result in slightly longer compilation time and then will not be included in the final binary since it is unused.

iroh-relay/src/protos/handshake.rs
```diff
@@ -0,0 +1,749 @@
...
+            .build())
+        }
+    }
+}
+
+async fn write_frame<F: serde::Serialize + Frame>(
+    io: &mut impl BytesStreamSink,
+    frame: F,
+) -> Result<(), Error> {
+    let mut bytes = BytesMut::new();
```
flub, line 486: I could be tempted to create this with `BytesMut::with_capacity(8192)`.  The allocator will probably start reusing those allocations soon.  I think 8KiB might be reasonable though we could consider 4KiB.  A single "1500" MTU will fit in both, 8KiB has slightly more space for a GSO batch.

iroh-relay/src/protos/common.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// Writes the frame type to the buffer (as a QUIC-encoded varint).
+    pub(crate) fn write_to<O: BufMut>(&self, mut dst: O) -> O {
+        VarInt::from(*self).encode(&mut dst);
+        dst
+    }
+
+    /// Parses the frame type (as a QUIC-encoded varint) from the first couple of bytes given
+    /// and returns the frame type and the rest.
+    pub(crate) fn from_bytes(bytes: Bytes) -> Option<(Self, Bytes)> {
+        let mut cursor = std::io::Cursor::new(&bytes);
```
flub: I think you're only using this cursor to keep track of the position, because you never seek?  That's a bit odd.

Firstly `Bytes` already keeps track of a cursor internally, there's `Buf::remaining` which you can use before and after reading the `VarInt`.

Secondly there's `VarInt::size` which... oh well, it's not pub, even though `VarInt::MAX_SIZE` is pub :sob: :sob: :sob:

Lastly, I find it strange that there's no `Bytes::split_off_from_buf_cursor` API.  It really seems weird we have to do this manually.

iroh-relay/src/protos/common.rs
```diff
@@ -0,0 +1,86 @@
...
+        let tag = VarInt::decode(&mut cursor).ok()?;
+        let tag_u32 = u32::try_from(u64::from(tag)).ok()?;
+        let frame_type = FrameType::from(tag_u32);
+        let content = bytes.slice(cursor.position() as usize..);
```
flub: I guess I'm used to seeing this as `bytes.split_off(cursor.position())` but... meh, both are documented as O(1) and looking at the implementation there really doesn't seem to be much difference.  Maybe `.slice` involves slightly more instructions, but :shrug:.

iroh-relay/src/protos/relay.rs
```diff
@@ -168,383 +62,240 @@ pub enum RecvError {
...
+}
 
-    pub(crate) fn new(cache: KeyCache) -> Self {
-        Self { cache }
+impl Datagrams {
+    fn write_to<O: BufMut>(&self, mut dst: O) -> O {
+        let ecn = self.ecn.map_or(0, |ecn| ecn as u8);
+        let segment_size = self.segment_size.unwrap_or_default();
+        dst.put_u8(ecn);
+        dst.put_u16(segment_size);
```
flub: If you want to save 2 bytes on the wire for lots of packets, you can have 2 separate frame types for a single datagram and a transmit.  I'm kind of on the fence on whether it is worth it.  I suspect it might affect a lot of datagrams so it might be worth it.  And we're not short on frametypes, we have u32::MAX minus whatever we've used already.  Even in the encodes-to-1-byte-varint range we have space for now.

iroh-relay/src/protos/common.rs
flub: Another comment on the skipped numbers.  Why skip them at all?  This is a completely new protocol anyway, with no backwards compat, isn't it?  The number that encode to a 1-byte varint are 0-63.  It might be worth to conserve them a little bit: start from 0 and make then sequential.

iroh-relay/src/protos/relay.rs
```diff
@@ -568,385 +321,301 @@ impl Frame {
...
-        }
-    }
-}
+    /// Tries to decode a frame received over websockets.
+    ///
+    /// Specifically, bytes received from a binary websocket message frame.
+    #[allow(clippy::result_large_err)]
+    #[cfg(feature = "server")]
+    pub(crate) fn from_bytes(bytes: Bytes, cache: &KeyCache) -> Result<Self, Error> {
+        let (frame_type, content) = FrameType::from_bytes(bytes).context(InvalidFrameSnafu)?;
```
flub: IIUC this invalid (or rather, unknown?) frame gets propagated all the way the the client Stream interface.  So it is clients themselves that decide what to do with this.  In our case the ActiveRelayActor stops when it receives an unknown frame.

However, if we allow skipping over any unknown frames - we could do that right here - then we can gracefully extend the protocol by adding new frame types.  Older clients would ignore them and we'd only need to be careful to not keep sending them if the client ignore them.

It would also simplify a lot of downstream use as the stream just becomes `Item = Option<Frame>` rather than the current `Item = Option<Result<Frame, ()>>`.

iroh-relay/src/server/streams.rs
```diff
@@ -190,4 +217,307 @@ impl AsyncWrite for MaybeTlsStream {
...
+    fn update_state(&mut self) {
+        let now = time::Instant::now();
+        let refill_periods = now.saturating_duration_since(self.last_fill).as_millis() as u32
+            / self.refill_period.as_millis() as u32;
+        if refill_periods == 0 {
+            // Nothing to do - we won't refill yet
+            return;
+        }
+
+        self.fill += refill_periods as i64 * self.refill;
```
flub: can this not overflow/wrap-around?

iroh-relay/src/server/streams.rs
```diff
@@ -190,4 +217,307 @@ impl AsyncWrite for MaybeTlsStream {
...
+        self.fill = std::cmp::min(self.fill, self.max);
+        self.last_fill += self.refill_period * refill_periods;
+    }
+
+    fn consume(&mut self, bytes: i64) -> Result<(), time::Instant> {
+        self.update_state();
+
+        self.fill -= bytes;
```
flub: can this not underflow/wrap-around?

iroh/src/magicsock/transports/relay.rs
```diff
@@ -313,17 +315,16 @@ impl RelaySender {
...
-fn split_packets(transmit: &Transmit<'_>) -> RelayContents {
-    let mut res = SmallVec::with_capacity(1);
-    let contents = transmit.contents;
-    if let Some(segment_size) = transmit.segment_size {
-        for chunk in contents.chunks(segment_size) {
-            res.push(Bytes::from(chunk.to_vec()));
-        }
-    } else {
-        res.push(Bytes::from(contents.to_vec()));
+fn datagrams_from_transmit(transmit: &Transmit<'_>) -> Datagrams {
```
flub: Probably should update the comments

iroh/src/magicsock/transports/relay.rs
```diff
@@ -313,17 +315,16 @@ impl RelaySender {
...
+    Datagrams {
+        ecn: transmit.ecn.map(|ecn| match ecn {
+            quinn_udp::EcnCodepoint::Ect0 => quinn_proto::EcnCodepoint::Ect0,
+            quinn_udp::EcnCodepoint::Ect1 => quinn_proto::EcnCodepoint::Ect1,
+            quinn_udp::EcnCodepoint::Ce => quinn_proto::EcnCodepoint::Ce,
+        }),
+        segment_size: transmit.segment_size.map(|ss| ss as u16),
+        contents: Bytes::copy_from_slice(transmit.contents),
```
flub: We still have to copy?  :sob:
Why?

flub (commented):
Overall this is great.

The old relay protocol had a very cool feature where effectively a black box was passed through from client-to-client.  This is now kind of removed since it's replaced with `Datagrams` which gives more structure.

You can argue that the datagrams can still be a black box however, it is datagrams with a lot of the internet facilities removed: we no longer have the UDP source and destination ports.  So it is not really a UDP datagram transport and you can effectively only transfer a single application protocol between two nodes (QUIC in our case).

I can't help but feel this is a loss.  CONNECT-UDP solves this with "Context Identifiers" I think: https://www.rfc-editor.org/rfc/rfc9298.html#name-context-identifiers.

Note that we don't HAVE to solve this right now, if we allow extending frame types in the future...

iroh/src/magicsock/transports/relay.rs
matheus23: Because `transmit.contents` has type `&[u8]`, i.e. we don't own the buffer contents, so we can't `try_send` them over a task boundary.

iroh-relay/src/protos/handshake.rs
matheus23: The handshake frames are usually very small (<100 bytes), so that'd be a bit of a big size.

You seem to be commenting on other frame types though, since you're speaking of MTU and GSO batches, and yeah *whoops*! I can actually compute the size a frame will encode to in advance so I can choose to allocate exactly the right size of `BytesMut` in advance, which I'm doing now.

iroh-relay/src/protos/common.rs
matheus23: Yeah I think I was way too conservative about changes to `FrameType`. I removed this case now.

iroh-relay/src/protos/common.rs
matheus23: Changed the comment.

iroh-relay/src/protos/common.rs
matheus23: I think that came up in the past discussion already, though?
Anyhow - sequence numbers make your protocol stateful in an annoying way. I'm happy to pay 7 bytes every 2-5 seconds for a simpler protocol :)

iroh-relay/src/protos/common.rs
matheus23: Removed this part of the comment.

iroh-relay/src/protos/common.rs
matheus23: Yeah - I was conservative with changes here for no reason. Changed this to use consecutive numbers.

iroh-relay/src/protos/common.rs
matheus23: Thank you, I wasn't aware that `Bytes` implements `Buf`. For some reason I thought it was fully immutable.

iroh-relay/src/protos/relay.rs
matheus23: Thanks for noting this. I changed this so that there's no more `FrameType::Unknown`.

I'm still keeping the `Result` though. E.g. in the `handshake` protocol it's nice to not skip unknown frames, since any unknown frame is immediately an error.

iroh-relay/src/server/streams.rs
matheus23: Thank you for these comments, I've added some tests & made the integer arithmetic more robust w.r.t. close-to-bounds values for configurable values (`bytes_per_second` being super high or the value passed to `consume` being insanely high).

Also made sure to check that we never div by zero etc.
I've added them to the other PR that I backported the rate limiter to: https://github.com/n0-computer/iroh/pull/3384/commits/46e8cd6a7bd1707ad432ae61be01e750222a89ee

iroh-relay/src/protos/common.rs
```diff
@@ -3,55 +3,65 @@
...
 #[repr(u32)]
-#[derive(Copy, Clone, PartialEq, Eq, Debug, num_enum::IntoPrimitive, num_enum::FromPrimitive)]
+#[derive(
+    Copy, Clone, PartialEq, Eq, Debug, num_enum::IntoPrimitive, num_enum::TryFromPrimitive,
+)]
 // needs to be pub due to being exposed in error types
 pub enum FrameType {
     /// The server frame type for the challenge response
-    ServerChallenge = 2,
+    ServerChallenge = 1,
```
flub: Why not start from 0?  Though it's not the end of the world, we can always use that frame later when we want to.

iroh-relay/src/protos/common.rs
```diff
@@ -69,13 +79,15 @@ impl FrameType {
...
-        let content = bytes.slice(cursor.position() as usize..);
-        Some((frame_type, content))
+    pub(crate) fn from_bytes(buf: &mut impl Buf) -> Result<Self, FrameTypeError> {
+        let tag = VarInt::decode(buf).ok().context(UnexpectedEndSnafu)?;
+        let tag_u32 = u32::try_from(u64::from(tag))
+            .ok()
+            .context(UnknownFrameTypeSnafu { tag })?;
+        let frame_type = FrameType::try_from(tag_u32)
+            .ok()
+            .context(UnknownFrameTypeSnafu { tag })?;
```
flub, line 98: Is there any reason you went with u32 for the frame type?  Seeing these shenanigans here seems needlessly complex?  Why didn't you go with a u64?  

Though also, this is purely a code thing.  Changing this changes nothing on the wire.

iroh-relay/src/protos/common.rs
matheus23: Oof. I don't know. No particular reason.
I'd appreciate not having to do the dance of updating all snapshot tests & the staging relays again though now :sweat_smile:

iroh-relay/src/protos/common.rs
matheus23: That's what the `num_enum::TryFromPrimitive` macro gives me.
And I think the reason for that is that that's what rust does during compilation usually (it represents enums as u32 at runtime).

iroh-relay/src/protos/handshake.rs
matheus23: Fair. I cleaned it up

matheus23:
> The old relay protocol had a very cool feature where effectively a black box was passed through from client-to-client. This is now kind of removed since it's replaced with `Datagrams` which gives more structure.
> 
> You can argue that the datagrams can still be a black box however, it is datagrams with a lot of the internet facilities removed: we no longer have the UDP source and destination ports. So it is not really a UDP datagram transport and you can effectively only transfer a single application protocol between two nodes (QUIC in our case).
> 
> I can't help but feel this is a loss. CONNECT-UDP solves this with "Context Identifiers" I think: https://www.rfc-editor.org/rfc/rfc9298.html#name-context-identifiers.
> 
> Note that we don't HAVE to solve this right now, if we allow extending frame types in the future...

I decided to undo the changes and pull them out into https://github.com/n0-computer/iroh/pull/3389 together with another commit that tries to address the saving 2 bytes when you don't do GSO frames.

iroh-relay/src/protos/common.rs
```diff
@@ -0,0 +1,98 @@
...
+#[derive(Debug, Snafu)]
+#[non_exhaustive]
+pub enum FrameTypeError {
+    #[snafu(display("not enough bytes to parse frame type"))]
+    UnexpectedEnd {},
+    #[snafu(display("frame type unknown"))]
+    UnknownFrameType { tag: VarInt },
+}
+
+impl std::fmt::Display for FrameType {
```
dignifiedquire: why implement display for this type?

iroh-relay/src/protos/handshake.rs
```diff
@@ -0,0 +1,750 @@
+//! Implements the handshake protocol that iroh's relays and iroh clients that connect to it go through.
```
dignifiedquire: this sentence reads odd

iroh-relay/src/protos/handshake.rs
```diff
@@ -0,0 +1,750 @@
...
+
+/// Authentication message from the client.
+#[derive(derive_more::Debug, serde::Serialize, serde::Deserialize)]
+#[cfg_attr(wasm_browser, allow(unused))]
+pub(crate) struct KeyMaterialClientAuth {
+    /// The client's public key
+    pub(crate) public_key: PublicKey,
+    /// A signature of (a hash of) extracted key material.
+    #[serde(with = "serde_bytes")]
+    pub(crate) signature: [u8; 64],
```
dignifiedquire, line 61: encode as hex in debug?

iroh-relay/src/protos/handshake.rs
```diff
@@ -0,0 +1,750 @@
...
+    /// Part of the extracted key material.
+    ///
+    /// Allows making sure we have the same underlying key material.
+    pub(crate) key_material_suffix: [u8; 16],
```
dignifiedquire, line 66: encode as hex in debug?

iroh-relay/src/protos/handshake.rs
```diff
@@ -0,0 +1,750 @@
...
+}
+
+/// A challenge for the client to sign with their secret key for NodeId authentication.
+#[derive(derive_more::Debug, serde::Serialize, serde::Deserialize)]
+pub(crate) struct ServerChallenge {
+    /// The challenge to sign.
+    /// Must be randomly generated with an RNG that is safe to use for crypto.
+    pub(crate) challenge: [u8; 16],
```
dignifiedquire, line 75: encode as hex in debug?

iroh-relay/src/protos/handshake.rs
```diff
@@ -0,0 +1,750 @@
...
+/// Used when authentication via [`KeyMaterialClientAuth`] didn't work.
+#[derive(derive_more::Debug, serde::Serialize, serde::Deserialize)]
+pub(crate) struct ClientAuth {
+    /// The client's public key, a.k.a. the `NodeId`
+    pub(crate) public_key: PublicKey,
+    /// A signature of (a hash of) the [`ServerChallenge`].
+    ///
+    /// This is what provides the authentication.
+    #[serde(with = "serde_bytes")]
+    pub(crate) signature: [u8; 64],
```
dignifiedquire, line 90: encode as hex in debug?

iroh-relay/src/protos/handshake.rs
```diff
@@ -0,0 +1,750 @@
...
+}
+
+/// Confirmation of successful connection.
+#[derive(derive_more::Debug, serde::Serialize, serde::Deserialize)]
+pub(crate) struct ServerConfirmsAuth;
+
+/// Denial of connection. The client couldn't be verified as authentic.
+#[derive(derive_more::Debug, Clone, serde::Serialize, serde::Deserialize)]
```
dignifiedquire: why use `derive_more::Debug`?

iroh-relay/src/protos/handshake.rs
```diff
@@ -0,0 +1,750 @@
...
+    pub(crate) fn new(mut rng: impl RngCore + CryptoRng) -> Self {
+        let mut challenge = [0u8; 16];
+        rng.fill_bytes(&mut challenge);
+        Self { challenge }
+    }
+
+    /// The actual message bytes to sign (and verify against) for this challenge.
+    fn message_to_sign(&self) -> [u8; 32] {
+        blake3::derive_key(
+            "iroh-relay handshake v1 challenge signature",
```
dignifiedquire: would be nicer for this to be a constant

iroh-relay/src/protos/handshake.rs
```diff
@@ -0,0 +1,750 @@
...
+            &self.challenge,
+        )
+    }
+}
+
+impl ClientAuth {
+    /// Generates a signature for given challenge from the server.
```
dignifiedquire: missing a `the`

iroh-relay/src/protos/handshake.rs
```diff
@@ -0,0 +1,750 @@
...
+    #[cfg(feature = "server")]
+    pub(crate) fn new(mut rng: impl RngCore + CryptoRng) -> Self {
+        let mut challenge = [0u8; 16];
+        rng.fill_bytes(&mut challenge);
+        Self { challenge }
+    }
+
+    /// The actual message bytes to sign (and verify against) for this challenge.
+    fn message_to_sign(&self) -> [u8; 32] {
+        blake3::derive_key(
```
dignifiedquire: maybe add a note why using `derive_key` gives the properties we want to sign?

iroh-relay/src/protos/handshake.rs
```diff
@@ -0,0 +1,750 @@
...
+    pub(crate) fn new(secret_key: &SecretKey, challenge: &ServerChallenge) -> Self {
+        Self {
+            public_key: secret_key.public(),
+            signature: secret_key.sign(&challenge.message_to_sign()).to_bytes(),
+        }
+    }
+
+    /// Verifies this client's authentication given the challenge this was sent in response to.
+    #[cfg(feature = "server")]
+    pub(crate) fn verify(&self, challenge: &ServerChallenge) -> bool {
```
dignifiedquire: I think using a `Result` is a bit cleaner, and matches the style of verification we have already

iroh-relay/src/protos/handshake.rs
```diff
@@ -0,0 +1,750 @@
...
+        self.public_key
+            .verify(
+                &challenge.message_to_sign(),
+                &Signature::from_bytes(&self.signature),
+            )
+            .is_ok()
+    }
+}
+
+#[cfg_attr(wasm_browser, allow(unused))]
```
dignifiedquire: why allow unused, instead of disabling it?

iroh-relay/src/protos/handshake.rs
```diff
@@ -0,0 +1,750 @@
...
+
+    /// Verifies this client auth on the server side using the same key material.
+    ///
+    /// This might return false for a couple of reasons:
+    /// 1. The exported keying material might not be the same between both ends of the TLS session
+    ///    (e.g. there's an HTTPS proxy in between that doesn't think/care about the TLS keying material exporter).
+    ///    This situation is detected when the key material suffix mismatches.
+    /// 2. The signature itself doesn't verify.
+    #[cfg(feature = "server")]
+    pub(crate) fn verify(&self, io: &impl ExportKeyingMaterial) -> bool {
```
dignifiedquire: same as above, `Result<(), VerificationError>` is a bit nicer imo

iroh-relay/src/protos/handshake.rs
```diff
@@ -0,0 +1,750 @@
...
+        let Some(key_material) = io.export_keying_material(
+            [0u8; 32],
+            b"iroh-relay handshake v1",
```
dignifiedquire: this really should be a constant

iroh-relay/src/protos/handshake.rs
```diff
@@ -0,0 +1,750 @@
...
+    pub(crate) fn new(secret_key: &SecretKey, io: &impl ExportKeyingMaterial) -> Option<Self> {
+        let public_key = secret_key.public();
+        let key_material = io.export_keying_material(
+            [0u8; 32],
+            b"iroh-relay handshake v1",
+            Some(secret_key.public().as_bytes()),
+        )?;
+        Some(Self {
+            public_key,
+            signature: secret_key.sign(&key_material[..16]).to_bytes(),
```
dignifiedquire: can you add a note somewhere, why exactly the splitting of the key material is done like this?

iroh-relay/src/protos/relay.rs
```diff
@@ -171,236 +57,144 @@ pub enum RecvError {
...
-    #[cfg(wasm_browser)]
-    pub(crate) fn to_vec(&self) -> Vec<u8> {
-        self.write_to(Vec::with_capacity(self.encoded_len()))
-    }
-
-    /// Writes it self to the given buffer.
+    /// Encodes this frame for sending over websockets.
+    ///
+    /// Specifically meant for being put into a binary websocket message frame.
+    #[cfg(feature = "server")]
```
dignifiedquire, line 153: why not use postcard for this encoding?

iroh-relay/src/protos/common.rs
matheus23: This is left over from the old `FrameType` on main. It's used in the error display strings, but yeah these could just use the `Debug` impl instead. Would you prefer if I removed it?

iroh-relay/src/protos/relay.rs
matheus23: Because it doesn't have a way of decoding `Bytes` efficiently, AFAIK.

iroh-relay/src/protos/handshake.rs
matheus23: Copy/paste. No particular reason. FWIW, this doesn't add more to the compiled output than `derive(Debug)`, and I don't think it has a particularly big impact on compile times.
Removed it anyways though.

iroh-relay/src/protos/handshake.rs
matheus23: Right! They really should be, I wanted to convert them. Thanks for catching these.

iroh-relay/src/client.rs
```diff
@@ -240,17 +242,30 @@ impl ClientBuilder {
...
             })?
+            .add_header(
+                SEC_WEBSOCKET_PROTOCOL,
+                http::HeaderValue::from_static(RELAY_PROTOCOL_VERSION),
+            )
+            .expect("valid header name and value")
             .limits(tokio_websockets::Limits::default().max_payload_len(Some(MAX_FRAME_SIZE)))
-            .connect_on(stream)
-            .await?;
+            .config(tokio_websockets::Config::default().flush_threshold(usize::MAX));
```
flub, line 262: Does this effectively requires us to always flush manually?

iroh-relay/src/protos/handshake.rs
matheus23: Added these. TLDR: You can distinguish if it fails because of mismatching TLS keying material exporters or if the client failed to produce a valid signature.

iroh-relay/src/protos/handshake.rs
matheus23: Agreed! Fixed.

iroh-relay/src/protos/handshake.rs
```diff
@@ -0,0 +1,814 @@
...
+    ///
+    /// This is what provides the authentication.
+    #[serde(with = "serde_bytes")]
+    #[debug("{}", HEX.encode(signature))]
+    pub(crate) signature: [u8; 64],
+}
+
+/// Confirmation of successful connection.
+#[derive(derive_more::Debug, serde::Serialize, serde::Deserialize)]
+pub(crate) struct ServerConfirmsAuth;
```
flub, line 95: Does this frame take any space on the wire or is this essentially just the FrameType VarInt on the wire?

iroh-relay/src/protos/handshake.rs
```diff
@@ -0,0 +1,814 @@
...
+///
+/// This requires access to the client's secret key to sign a challenge.
+pub(crate) async fn clientside(
+    io: &mut (impl BytesStreamSink + ExportKeyingMaterial),
+    secret_key: &SecretKey,
+) -> Result<ServerConfirmsAuth, Error> {
+    let (tag, frame) = read_frame(
+        io,
+        &[ServerChallenge::TAG, ServerConfirmsAuth::TAG],
+        time::Duration::from_secs(30),
```
flub: Did you experience the need for timeouts on this level?  I would hope that if things are this slow a timeout on a higher level gives up on this.  On the client side we have https://github.com/n0-computer/iroh/blob/0776687e982050c7a7c46e8d430b350a24481718/iroh/src/magicsock/transports/relay/actor.rs#L503 which has a 10s timeout.  But IIUC that is only on the raw TLS connection.  Once connected it seems like in the ActiveRelayActor there's a 15s ping interval and pings have a 5s timeout.  So if the whole handshake exchange takes more than 20s the first ping will fail and the whole client will disconnect.

Anyway, 30s for individual frames is an eternity, is it not?  I'd be tempted to not have timeouts on this level and let something higher up the stack handle those.  From my above spelunking I think you can just remove the timeouts and it will already work.

iroh-relay/src/protos/handshake.rs
```diff
@@ -0,0 +1,814 @@
...
+            });
+        }
+        // Verification not succeeding is part of normal operation: The TLS exporter isn't required to match.
+        // We'll fall back to verification that takes another round trip more time.
+    }
+
+    let challenge = ServerChallenge::new(rng);
+    write_frame(io, &challenge).await?;
+
+    let (_, frame) = read_frame(io, &[ClientAuth::TAG], time::Duration::from_secs(10)).await?;
```
flub: I think we need to check out the use on the server-side as well like I did above for the client (btw, 10s here is inconsistent with the client side?  Maybe use a constant if this is still needed).  I think again ideally it's not this level that needs to provide timeouts.

FWIW you also don't have write timeouts.

iroh-relay/src/protos/handshake.rs
flub, line 1: I am impressed withe the level of testing in this file btw!

iroh-relay/src/protos/relay.rs
```diff
@@ -51,104 +36,7 @@ pub(crate) const PING_INTERVAL: Duration = Duration::from_secs(15);
...
-    #[snafu(transparent)]
-    Io { source: std::io::Error },
-    #[snafu(transparent)]
-    Timeout { source: time::Elapsed },
-    #[snafu(transparent)]
-    ConnSend { source: ConnSendError },
-    #[snafu(transparent)]
-    SerDe { source: postcard::Error },
-}
+pub(crate) const PER_CLIENT_SEND_QUEUE_DEPTH: usize = 512;
```
flub, line 39: This comment isn't actionable.  But :exploding_head:
Used to be 32 in the tailscale code by the looks of the old comment.  I suspect this was turned up at some point because it make a backpressure problem go away for a particular example.  But we've actually improved backpressure since.  Could we create an issue to investigate this and see if this number can be turned down?

iroh-relay/src/protos/relay.rs
```diff
@@ -411,108 +205,55 @@ impl Frame {
...
-                let dst_key = cache.key_from_slice(&content[..PublicKey::LENGTH])?;
-                let packet = content.slice(PublicKey::LENGTH..);
-                Self::SendPacket { dst_key, packet }
-            }
+        let res = match frame_type {
             FrameType::RecvPacket => {
-                if content.len() < PublicKey::LENGTH {
-                    return Err(InvalidFrameSnafu.build());
-                }
+                snafu::ensure!(content.len() >= NodeId::LENGTH, InvalidFrameSnafu);
```
flub: I think you also want to be careful to reject empty payloads, because that would allow sending 0-sized datagrams to other clients and I have no idea how all things would cope with that.  I'm assuming on the internet this isn't allowed.

So maybe change this to `>`?

iroh-relay/src/protos/relay.rs
```diff
@@ -536,161 +279,198 @@ impl Frame {
...
+
+    pub(crate) fn encoded_len(&self) -> usize {
+        let payload_len = match self {
+            Self::Ping(_) | Self::Pong(_) => 8,
+            Self::SendPacket { packet, .. } => {
+                32 // node id
+                + packet.len()
+            }
+        };
+        1 // frame type (all frame types currently encode as 1 byte varint)
```
flub: Can we `debug_assert!` this somehow?  I guess it's hard to know what the largest FrameType is going to be?

iroh-relay/src/client.rs
matheus23: Yes. Every call to `Sink::send` will flush, and every call to `Sink::send_all` will do so too at the end.

I think this version uses `Sink::send_all`, and in the other PR, I *think* I'm changing that into a bunch of `feed`s and a `flush`? But I'm not sure.

But yeah - I don't want to flush every couple of KBs automatically, there's no need for us to do that and throughput improves if we don't do so.

iroh-relay/src/protos/handshake.rs
matheus23: Nope, `postcard::to_allocvec(&ServerConfirmsAuth).unwrap().len() == 0`.

iroh-relay/src/protos/handshake.rs
matheus23: Fair. (The timeouts are inherited from the old code.)

iroh-relay/src/protos/relay.rs
matheus23: Let me improve code cohesion by adding `FrameType::encoded_len`, so it's right next to the frame type encoding code.

iroh-relay/src/client.rs
flub: Maybe add a comment to explain this?  I think it's good to draw attention to that here.

iroh-relay/src/protos/handshake.rs
flub: great!

iroh-relay/src/protos/relay.rs
matheus23: https://github.com/n0-computer/iroh/issues/3391

iroh-relay/src/protos/relay.rs
matheus23: Hmm. Yeah. I went back and forth on this, but I guess as long as there's no need for empty payloads, I shouldn't have them.

iroh-relay/src/protos/relay.rs
matheus23: Apparently it *is* possible: https://gist.github.com/dignifiedquire/7c64b868f71688d05249db00362eb70c

But I still don't want to do this:
1. I'd have to split the enum into structs for each case, since I have my own `FrameType` and don't want postcard to handle the frame type.
2. I'd need to duplicate each of these structs then, too.

matheus23: FWIW, we can still talk about this of course. But I'd say that would be a bigger change and should probably be another PR :S

iroh-relay/src/protos/common.rs
matheus23: I removed the impl

PR #3338, opened 2025-06-03
[WIP] refactor: remove add_node_addr API

### LINKED ISSUE #3329: remove endpoint::Builder::known_nodes
This API is essentially undocumented.  It seems to be a old way of adding a bunch of NodeAddrs at startup.  But given that `Endpoint::add_node_addr` is also deprecated in favour of the `StaticProvider` maybe it makes sense to remove this entirely?

dignifiedquire:
## Description

Starting to fix #3329, and trying to cleanup discovery interactions while at it

## Breaking Changes

**TODO**

### DIFF
--- a/iroh/src/discovery.rs
+++ b/iroh/src/discovery.rs
@@ -114,15 +114,17 @@ use n0_future::{
     stream::StreamExt,
     task::{self, AbortOnDropHandle},
     time::{self, Duration},
-    Stream, TryStreamExt,
+    Stream,
 };
 use nested_enum_utils::common_fields;
 use snafu::{ensure, IntoError, Snafu};
-use tokio::sync::oneshot;
 use tracing::{debug, error_span, warn, Instrument};
 
 pub use crate::node_info::{NodeData, NodeInfo, ParseError, UserData};
-use crate::Endpoint;
+use crate::{
+    watcher::{self, Watcher},
+    Endpoint,
+};
 
 #[cfg(not(wasm_browser))]
 pub mod dns;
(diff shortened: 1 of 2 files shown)


Frando:
Currently `iroh-gossip` uses `Endpoint::add_node_addr` to inject node info received via gossip into the endpoint. There's a PR in iroh-gossip that uses a gossip-specific discovery service instead: https://github.com/n0-computer/iroh-gossip/pull/51

iroh/src/discovery.rs
```diff
@@ -580,30 +583,22 @@ pub struct Lagged {
...
+            inner: Default::default(),
         }
     }
 
-    pub(crate) fn subscribe(&self) -> impl Stream<Item = Result<DiscoveryItem, Lagged>> {
-        use tokio_stream::wrappers::{errors::BroadcastStreamRecvError, BroadcastStream};
-        let recv = self.inner.subscribe();
-        BroadcastStream::new(recv).map_err(|BroadcastStreamRecvError::Lagged(n)| Lagged { val: n })
+    pub(crate) fn subscribe(&self) -> impl Stream<Item = Option<DiscoveryItem>> {
+        self.inner.watch().stream()
```
flub, line 597: IIUC this is a quite big behaviour change: previously you'd have a reasonable chance of seeing all discovered items, now you only get a single one.  Especially the ConcurrentDiscovery this could be an issue if multiple discovery services find something at the same time you'll miss one address.

iroh/src/endpoint.rs
```diff
@@ -1169,7 +1091,7 @@ impl Endpoint {
     ///
     /// [`MdnsDiscovery`]: crate::discovery::mdns::MdnsDiscovery
     /// [`StaticProvider`]: crate::discovery::static_provider::StaticProvider
-    pub fn discovery_stream(&self) -> impl Stream<Item = Result<DiscoveryItem, Lagged>> {
+    pub fn discovery_stream(&self) -> impl Stream<Item = Option<DiscoveryItem>> {
```
flub, line 1094: Just from looking at the signature change I think the old API was more correct here as well?

dignifiedquire:
this doesn't work as I would like it to..

PR #3339, opened 2025-06-03
refactor(iroh)!: rename ProtocolError to AcceptError

Frando:
## Description

To me, this reads much nicer in the examples etc. 

## Breaking Changes

- `iroh::endpoint::ProtocolError` -> `AcceptError`

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/examples/echo.rs
+++ b/iroh/examples/echo.rs
@@ -8,7 +8,7 @@
 
 use iroh::{
     endpoint::Connection,
-    protocol::{ProtocolError, ProtocolHandler, Router},
+    protocol::{AcceptError, ProtocolHandler, Router},
     watcher::Watcher as _,
     Endpoint, NodeAddr,
 };
@@ -83,7 +83,7 @@ impl ProtocolHandler for Echo {
     ///
     /// The returned future runs on a newly spawned tokio task, so it can run as long as
     /// the connection lasts.
-    async fn accept(&self, connection: Connection) -> Result<(), ProtocolError> {
+    async fn accept(&self, connection: Connection) -> Result<(), AcceptError> {
         // We can get the remote's node id from the connection.
         let node_id = connection.remote_node_id()?;
         println!("accepted connection from {node_id}");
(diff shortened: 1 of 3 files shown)


matheus23 (approved):
Yeah I like the renames.

The breaking changes list is... weird.
It's both not quite complete (you also renamed fields of the (now) `AcceptError` enum) as well as not necessary I think, since we haven't published a version of iroh with the error enums yet.

Frando:
> The breaking changes list is... weird.
It's both not quite complete (you also renamed fields of the (now) AcceptError enum) as well as not necessary I think, since we haven't published a version of iroh with the error enums yet.

True. I'll just remove it, the new errors weren't mentioned in #3161 at all.

dignifiedquire:
this is breaking, please mark the PR name and document it

matheus23:
> this is breaking, please mark the PR name and document it

Breaking relative to what? Relative to `main` or relative to the latest release?

Putting in "renamed `ProtocolError` to `AcceptError`" and similar notes in the 0.36 release won't make any sense, since there hasn't been any `ProtocolError` enum before.

And the error PR didn't really go into detail on which enums got introduced, so there's nothing to fix up in terms of breaking changes relative to that PR's breaking change list.

See also [the discussion above](https://github.com/n0-computer/iroh/pull/3339#pullrequestreview-2891940198)

EDIT: In any case - I don't want to argue for something here. I'm not the person cobbling together the changelog :sweat_smile: so anything that makes that person's life better I'm happy to argue for. But in this case I don't understand why adding some breaking changes list to this PR isn't only noise.

dignifiedquire:
> EDIT: In any case - I don't want to argue for something here. I'm not the person cobbling together the changelog 😅 so anything that makes that person's life better I'm happy to argue for. But in this case I don't understand why adding some breaking changes list to this PR isn't only noise.

The problem is that if you leave it out, we will end up with the wrong changelog potentially, as we will only have "add ProtocolError" in the original

PR #3340, opened 2025-06-03
refactor(iroh)!: simplify discovery errors

Frando:
## Description

A couple of variants where only used internally, and the the `NodeId` variant had the same semantics as `NoResults`, so they can be merged.

## Breaking Changes

- Some variants of the `iroh::discovery::DiscoveryError` have been changed

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/discovery/pkarr.rs
+++ b/iroh/src/discovery/pkarr.rs
@@ -47,7 +47,7 @@
 use std::sync::Arc;
 
 use iroh_base::{NodeId, RelayUrl, SecretKey};
-use iroh_relay::node_info::NodeInfo;
+use iroh_relay::node_info::{EncodingError, NodeInfo};
 use n0_future::{
     boxed::BoxStream,
     task::{self, AbortOnDropHandle},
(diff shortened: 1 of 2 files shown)


iroh/src/discovery/pkarr.rs
```diff
@@ -360,8 +362,8 @@ impl Discovery for PkarrResolver {
         let pkarr_client = self.pkarr_client.clone();
         let fut = async move {
             let signed_packet = pkarr_client.resolve(node_id).await?;
-            let info =
-                NodeInfo::from_pkarr_signed_packet(&signed_packet).context(ParsePacketSnafu)?;
+            let info = NodeInfo::from_pkarr_signed_packet(&signed_packet)
```
Frando, line 365: This is the only "real" change in the PR. This, so far, was the single discovery error that had the privilege of having their own variant on the `DiscoveryError`, all other errors from discovery services were shoveled into the `User` variant. I think we should either add more variants for other errors (eg the HTTP errors for the pkarr client), or not having any variants specific to discovery services. This PR opts for the latter for now.

PR #3355, opened 2025-06-23
feat(iroh): add new protocols to router at runtime

Frando:
## Description

This allows adding new protocols and removing protocols to an iroh `Router` after the router is built.
Also adds a timeout to shutdown for a protocol handler of 30s.

Also wrote a test for this.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/protocol.rs
+++ b/iroh/src/protocol.rs
@@ -388,7 +388,7 @@ impl Router {
     ///
     /// If a protocol handler was already registered for `alpn`, the previous handler will be
     /// shutdown. Existing connections will not be aborted by the router, but some protocol
-    /// handlers may abort existing connnections in their [`Router::shutdown`] implementation.
+    /// handlers may abort existing connections in their [`Router::shutdown`] implementation.
     /// Consult the documentation of the protocol handler to see if that is the case.
     pub async fn accept(
         &self,
@@ -413,7 +413,7 @@ impl Router {
     ///
     /// If a protocol handler was registered for `alpn`, the handler will be
     /// shutdown. Existing connections will not be aborted by the router, but some protocol
-    /// handlers may abort existing connnections in their [`Router::shutdown`] implementation.
+    /// handlers may abort existing connections in their [`Router::shutdown`] implementation.
     /// Consult the documentation of the protocol handler to see if that is the case.
     ///
     /// Returns an error if the router has been shutdown or no protocol is registered for `alpn`.


iroh/src/protocol.rs
```diff
@@ -137,6 +151,37 @@ impl From<quinn::ClosedStream> for AcceptError {
     }
 }
 
+#[allow(missing_docs)]
+#[derive(Debug, Snafu)]
+#[non_exhaustive]
+pub enum RouterError {
+    #[snafu(display("The router actor closed"))]
```
flub: "router actor" is something rather internal, no?  Can we not say "the router closed" or maybe even "the endpoint closed"?

iroh/src/protocol.rs
```diff
@@ -311,6 +381,49 @@ impl Router {
         self.cancel_token.is_cancelled()
     }
 
+    /// Adds a protocol to the list of accepted protocols.
+    ///
+    /// Configures the router to accept the [`ProtocolHandler`] when receiving a connection
+    /// with this `alpn`.
+    ///
+    /// Once the function yields, new connections with this `alpn` will be handled.
```
flub: "yields" or returns?  As far as I read the code it should be "returns"?

iroh/src/protocol.rs
```diff
@@ -311,6 +381,49 @@ impl Router {
...
+    ///
+    /// If a protocol handler was already registered for `alpn`, the previous handler will be shutdown.
```
flub: Can you elaborate more on what happens to existing connections?  Will they all be closed by the old handler?  How does that work?  If you're repeating how handlers shut down connections that's fine, docs should repeat things.

iroh/src/protocol.rs
```diff
@@ -311,6 +381,49 @@ impl Router {
...
+            })
+            .await
+            .map_err(|_| RouterError::Closed {})?;
+        reply_rx.await.map_err(|_| RouterError::Closed {})
+    }
+
+    /// Stops accepting a protocol.
+    ///
+    /// Note that this has only an effect on new connections. Existing connections that were
+    /// accepted with `alpn` won't be closed when calling [`Router::stop_accepting`].
```
flub: Can you call out the differences in behaviour with calling `accept` where a previous handler was already handling this protocol?

Frando:
@flub Thanks for the review, clarified the docs. LMK if there's still something unclear from reading the docs.

TheButlah:
unsure if I should post here or open an issue. is there some way that we can allow end-users to have a `Box<dyn DynProtocolHandler>`? Or at least, introduce a BoxedProtocolHandler type? Those types appear private rn.

Use case: facilitating third-party routers, or facilitating storing protocol handlers in structs, such as for the purposes of configuration:
```rust
#[derive(Debug, bon::Builder)]
pub struct Agent {
    #[builder(field)]
    handlers: HashMap<Alpn, Box<dyn ProtocolHandler>>,
    /// Provide a specific secret key for mTLS, instead of dynamically generating one
    secret_key: Option<iroh::SecretKey>,
    #[builder(default)]
    discovery: EnabledDiscoveryServices,
    /// in QUIC, the ALPN allows for multiple connections on same port, and it is ideal
    /// to circumvent firewalls to always use port 443 (same as https). You can also set
    /// this to 0 to dynamically choose a port.
    #[builder(default = 443)]
    port: u16,
}

impl<S: agent_builder::State> AgentBuilder<S> {
    fn router(
        mut self,
        alpn: &'static str,
        router: impl Into<Box<dyn ProtocolHandler>>,
    ) -> Self {
        self.handlers.insert(Alpn(alpn), router.into());
        self
    }
}

//...

        let mut router = iroh::protocol::Router::builder(endpoint);
        for (alpn, handler) in self.handlers {
            router = router.accept(alpn.0, handler);
        }

```

Frando:
> unsure if I should post here or open an issue. is there some way that we can allow end-users to have a `Box<dyn DynProtocolHandler>`? Or at least, introduce a BoxedProtocolHandler type? Those types appear private rn.

That's valid, IMO. I opened https://github.com/n0-computer/iroh/pull/3366.

Frando:
This is not planned at the moment, see https://github.com/n0-computer/iroh-examples/tree/main/custom-router for a user space implementation.

PR #3356, opened 2025-06-23
feat(iroh): re-export `n0_watcher::Watcher` trait

ramfox:
## Description

Re-export the `n0_watcher::Watcher` so it is easier for our users to use heavily relied on methods like `Endpoint::node_addr`.

This becomes especially painful when you are working on any protocols using iroh.

## Notes & open questions

Should this be exported from a different module? `iroh::endpoint::Watcher`?

## Change checklist
- [x] Self-review.


Frando (approved):

flub (approved):

flub:
> Should this be exported from a different module? `iroh::endpoing::Watcher`?

I think the toplevel is fine.  Though I love I'm not the only one making this typo all the time :grin:

PR #3362, opened 2025-06-24
feat: add methods to create variants of the `iroh-base::ticket::ParseError` enum.

ramfox:
## Description

Adds a `From` impl for `postcard::Error`, and adds `ParseError::verification_failed` and `ParseError::wrong_prefix` methods.

## Open Questions

Should we add a variant to `ParseError` for a more "open ended" error, like we do for `protocol::AcceptError`? Something like `ParseError::Custom`?

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh-base/src/ticket.rs
+++ b/iroh-base/src/ticket.rs
@@ -7,7 +7,7 @@ use std::{collections::BTreeSet, net::SocketAddr};
 
 use nested_enum_utils::common_fields;
 use serde::{Deserialize, Serialize};
-use snafu::{Backtrace, Snafu};
+use snafu::{Backtrace, IntoError, Snafu};
 
 use crate::{key::NodeId, relay_url::RelayUrl};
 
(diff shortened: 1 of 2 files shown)


iroh-base/src/ticket.rs
```diff
@@ -86,6 +86,28 @@ pub enum ParseError {
...
+    /// Return a `ParseError` variant that indicates verification of the
+    /// deserialized bytes failed.
+    pub fn verification_failed(message: &'static str) -> Self {
+        VerifySnafu { message }.build()
+    }
+}
+
+impl From<postcard::Error> for ParseError {
+    fn from(source: postcard::Error) -> Self {
+        PostcardSnafu.into_error(source)
```
dignifiedquire: you don't need to do this manually: https://docs.rs/snafu/latest/snafu/derive.Snafu.html#disabling-the-context-selector

ramfox: should this really just be `transparent` then? If we don't care about adding the "deserialization failed" display context?

dignifiedquire: yeah probably

dignifiedquire (approved):

PR #3350, opened 2025-06-16
A test for weird long connection behavior from same Node IDs

RingOfStorms:
## Description

[Regarding this Discord halp thread.](https://discordapp.com/channels/1161119546170687619/1303353870529069147/1383829507193966632)

Trying to figure out how to prove there is weird connection bugs in iroh.

Specifically:
* Graceful close is not graceful, note that when logging is on you will see WARN failed to clean up tasks on any of the graceful examples
    * run `cargo test -- --nocapture --test graceful_close` to see these logs, this is seen on the main branch of iroh as of writing this.
    
Sample output
```log
2025-06-16T21:05:34.991992Z DEBUG graceful_close:ep{me=ac83794a87}:magicsock:actor: iroh::magicsock: shutdown complete
2025-06-16T21:05:35.093639Z  WARN graceful_close:close{me=ac83794a87}: iroh::magicsock: tasks didn't finish in time, aborting remaining 1/3 tasks
2025-06-16T21:05:35.093786Z TRACE graceful_close:close{me=ac83794a87}: iroh::magicsock: magicsock closed
test endpoint::tests::graceful_close ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 78 filtered out; finished in 4.75s
```

* Connecting from the same node ID (using same secret) can take extremely long amounts of time. First run usually takes < 3 seconds and a subsequent run will take > 10 seconds
    * With this branch run: `cargo test --features="discovery-local-network" -- --nocapture --test can_connect_from_same_id`

Sample output
```log
2025-06-16T22:02:49.372917Z TRACE can_connect_from_same_id:connect{me="3b6a27bcce" alpn="n0/iroh/test" remote="378fc07264"}:drive{id=0}: iroh_quinn_proto::connection: timeout timer=Close
2025-06-16T22:02:49.373051Z TRACE can_connect_from_same_id:ep{me=3b6a27bcce}:magicsock:actor: iroh::magicsock: tick: msg msg=Shutdown
2025-06-16T22:02:49.373071Z DEBUG can_connect_from_same_id:ep{me=3b6a27bcce}:magicsock:actor: iroh::magicsock: shutting down
2025-06-16T22:02:49.373112Z  INFO can_connect_from_same_id:ep{me=3b6a27bcce}:magicsock:actor:reset{node=378fc07264}: iroh::magicsock::node_map::best_addr: clearing best_addr reason=Reset has_relay=false old_addr=[omitted]:34489
[...]

PR #3372, opened 2025-06-27
tests(iroh): add test to reconnect after forceful abort

Frando:
## Description

The test is adopted from `iroh-gossip` test `can_die_and_reconnect`.

The test fails. If I didn't mess up the port of the test from iroh-gossip, it should pass. Suspecting a regression here.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/tests/reconnect.rs
+++ b/iroh/tests/reconnect.rs
@@ -1,0 +1,176 @@
+use std::time::Duration;
+
+use iroh::{
+    endpoint::{BindError, Connection},
+    protocol::{AcceptError, ProtocolHandler, Router},
+    Endpoint, NodeAddr, NodeId, RelayMap, RelayMode, SecretKey, Watcher,
+};
+use n0_future::time::timeout;
+use n0_snafu::{Result, ResultExt};
+use rand::SeedableRng;
+use tokio::sync::mpsc;
+use tokio_util::sync::CancellationToken;
+use tracing::info;
+
+#[cfg(feature = "test-utils")]
+#[tokio::test]
+// #[traced_test]
+async fn can_die_and_reconnect() -> Result {
+    tracing_subscriber::fmt::init();
+
+    let max_wait = Duration::from_secs(5);
+
+    #[derive(Debug, Clone)]
+    struct TestProtocol(mpsc::Sender<(NodeId, String)>);
+
+    const TEST_ALPN: &[u8] = b"/iroh/test/1";
+
+    impl ProtocolHandler for TestProtocol {
+        async fn accept(&self, connection: Connection) -> Result<(), AcceptError> {
+            let remote_node_id = connection.remote_node_id()?;
+            let mut stream = connection.accept_uni().await?;
+            let data = stream
+                .read_to_end(64)
+                .await
+                .map_err(AcceptError::from_err)?;
+            let s = String::from_utf8(data).map_err(AcceptError::from_err)?;
+            self.0
+                .send((remote_node_id, s))
+                .await
+                .map_err(AcceptError::from_err)?;
+            Ok(())
+        }
+    }
+
+    /// Runs a future in a separate runtime on a separate thread, cancelling everything
+    /// abruptly once `cancel` is invoked.
+    fn run_in_thread<T: Send + 'static>(
+        cancel: CancellationToken,
+        fut: impl std::future::Future<Output = T> + Send + 'static,
+    ) -> std::thread::JoinHandle<Option<T>> {
+        std::thread::spawn(move || {
+            let rt = tokio::runtime::Builder::new_current_thread()
+                .enable_all()
+                .build()
+                .unwrap();
+            rt.block_on(async move { cancel.run_until_cancelled(fut).await })
+        })
+    }
+
+    /// Spawns a new client endpoint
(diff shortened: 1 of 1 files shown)


iroh/src/endpoint.rs
```diff
@@ -3216,4 +3216,85 @@ mod tests {
 
         Ok(())
     }
+
+    /// Test that we can immediately reconnect after respawning an endpoint with the same node id.
```
flub, line 3220: The important bit here is that there is no relay.  If you write a comment you should probably mention that too.

iroh/src/endpoint.rs
```diff
@@ -3216,4 +3216,85 @@ mod tests {
...
+        // First connection
+        let ep = n0_future::time::timeout(
+            TIMEOUT,
+            connect(client_secret_key.clone(), server_addr.clone(), 23),
+        )
+        .await
+        .e()??;
+        assert_eq!(rx.recv().await.unwrap().unwrap(), 23);
+        // close the endpoint in a separate task, to not lose time for our immediate respawn testing
+        let close1 = tokio::task::spawn(async move { ep.close().await });
```
flub: How long does this take?  Since both endpoints are still there this should be pretty fast, no?  Feels a bit odd to spawn this.

flub:
> My only concern is the 5s is asking for flakyness, but I can't really think of something better right away.

Oh, lol.  Windows CI already fails because of this...  Guess we'll have to figure this out right away.

flub:
I think you might be able to set this to a 30s timeout because without a relay that would still fail after 30s in the old version?  You can verify that.

flub:
Alternatively you can mimic what https://github.com/n0-computer/iroh/pull/3350 does and measure the time it takes to connect initially.  Then require the 2nd connection to succeed in 3 times that base time.  Could be more robust?

iroh/src/endpoint.rs
Frando: Yea, I'll remove it again. It was a trial for the windows failure, but that's not it, see comment below

Frando:
The flakyness is not only due to slow CI I think, I got it to fail locally too from time to time. My current reasoning is that the fix applied here is not enough (pasting from discord):

so with the change from the PR, a new path is pinged. once we get the pong, we check if this path should be the new best_addr. which takes us to here:
https://github.com/n0-computer/iroh/blob/main/iroh/src/magicsock/node_map/best_addr.rs#L111
the new path will only become the best_addr (and thus the send addr for the next quic packet) if it has a lower latency than the current best_addr, or if the current best_addr exceeds its trust interval 
in our test case, the previous addr is still trusted. the path is also physically identical, so it is purely random which of the two pingpongs had a lower latency 
if the pingpong from the new client endpoint has a lower latency, the addr from the new client endpoint is used from then on. if the pingpong from the old client endpoint happened to have a lower latency, the best_addr is kept and future udp packets are sent into nirvana

iroh/src/endpoint.rs
Frando: :+1: Expanded the comment

Frando:
I pushed a commit that makes the test pass reliably (locally for me at least):

https://github.com/n0-computer/iroh/pull/3372/commits/a8ee257e2f94c7b4ae0badd67b25513b98c3d46f

It adds the following case to `BestAddr::insert_if_better_or_reconfirm`, which is called when a pong is received on a UDP path:
```rust
// If we receive a pong on a different port but the same IP address as the current best addr,
// we assume that the endpoint has rebound, and thus use the new port.
} else if state.addr.addr.ip() == addr.ip() {
    self.insert(addr, latency, source, confirmed_at)
}
```
So when we receive a pong from the same IP address as our current best_addr, but with a different port, we unconditionally switch our best addr to that new port, even if the latency is lower than the best addr's latency from the last pingpong.

So far this was the simplest I could come up with. Can you think of any scenarios where that would lead to undesired outcomes?

iroh/src/magicsock/node_map/best_addr.rs
```diff
@@ -121,11 +121,19 @@ impl BestAddr {
...
+                // use the new address.
                 if !state.is_trusted(confirmed_at) || candidate.is_better_than(&state.addr) {
                     self.insert(addr, latency, source, confirmed_at);
+                // If the new address is equal to the current address, mark it as reconfirmed now and expand its trust
+                // interval.
                 } else if state.addr.addr == addr {
                     state.confirmed_at = confirmed_at;
                     state.trust_until = Some(source.trust_until(confirmed_at));
+                // If we receive a pong on a different port but the same IP address as the current best addr,
+                // we assume that the endpoint has rebound, and thus use the new port.
```
flub, line 134: My main issue with this logic is that I find it fairly arbitrary.  It fixes an artificial test case, but it seems like a sub-set of the cases that need to be fixed.  We're also not really sure we can send on this path at this point, but I'm not even sure if that's something we're always sure of anyway.

flub:
I wonder if the original version should still be merged here?

flub:
I believe the core issue of not being able to reconnect is solved in the multipath branch:

- If there is no connection the selected_path is forgotten.
- On a new connection all remote addresses are sent to for the Initial packet.

PR #3374, opened 2025-06-28
fix(iroh): Use std Mutex instead of tokio Mutex

juchiast:
## Description

Per tokio docs, if a mutex guard is not held across await points, it is possible and preferable to use std mutex. 

## Change checklist

- [x] Self-review.


flub (approved):
thanks!

flub:
Let's try once more.  The base branch shouldn't be that flaky though, so I'm a little suspicious.

PR #3381, opened 2025-07-07
[WIP] feat: use quinn multipath

dignifiedquire:
Work on integrating https://github.com/n0-computer/quinn/pull/28 into the iroh magic

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,3 +40,8 @@ unexpected_cfgs = { level = "warn", check-cfg = ["cfg(iroh_docsrs)", "cfg(iroh_l
 
 [workspace.lints.clippy]
 unused-async = "warn"
+
+
+[patch.crates-io]
+rustls = { git = "https://github.com/n0-computer/rustls", rev = "be02113e7837df60953d02c2bdd0f4634fef3a80" }
+netwatch = { git = "https://github.com/n0-computer/net-tools", branch = "feat-multipath" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -479,6 +479,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1628fb46dfa0b37568d12e5edd512553eccf6a22a78e8bde00bb4aed84d5bdbf"
 
 [[package]]
+name = "bytemuck"
+version = "1.23.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5c76a5792e44e4abe34d3abf15636779261d45a7450612059293d1d2cfc63422"
+
+[[package]]
 name = "byteorder"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1181,6 +1187,18 @@ name = "fallible-iterator"
 version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2acce4a10f12dc2fb14a218589d4f1f62ef011b2d0cc4b3cb1bba8e94da14649"
+
+[[package]]
+name = "fastbloom"
+version = "0.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "27cea6e7f512d43b098939ff4d5a5d6fe3db07971e1d05176fe26c642d33f5b8"
+dependencies = [
+ "getrandom 0.3.2",
+ "rand 0.9.1",
+ "siphasher",
+ "wide",
+]
 
 [[package]]
 name = "fastrand"
@@ -2425,9 +2443,8 @@ dependencies = [
 
 [[package]]
 name = "iroh-quinn"
-version = "0.14.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0cde160ebee7aabede6ae887460cd303c8b809054224815addf1469d54a6fcf7"
+version = "0.13.0"
+source = "git+https://github.com/n0-computer/quinn?branch=multipath-quinn-0.11.x#773fceabb27f1e56132198dd960d4bd1493e0ed0"
 dependencies = [
  "bytes",
  "cfg_aliases",
@@ -2446,12 +2463,13 @@ dependencies = [
 [[package]]
 name = "iroh-quinn-proto"
 version = "0.13.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "929d5d8fa77d5c304d3ee7cae9aede31f13908bd049f9de8c7c0094ad6f7c535"
+source = "git+https://github.com/n0-computer/quinn?branch=multipath-quinn-0.11.x#773fceabb27f1e56132198dd960d4bd1493e0ed0"
 dependencies = [
  "bytes",
- "getrandom 0.2.16",
- "rand 0.8.5",
+ "fastbloom",
+ "getrandom 0.3.2",
+ "lru-slab",
+ "rand 0.9.1",
  "ring",
  "rustc-hash",
  "rustls",
@@ -2466,9 +2484,8 @@ dependencies = [
 
 [[package]]
 name = "iroh-quinn-udp"
-version = "0.5.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c53afaa1049f7c83ea1331f5ebb9e6ebc5fdd69c468b7a22dd598b02c9bcc973"
+version = "0.5.12"
+source = "git+https://github.com/n0-computer/quinn?branch=multipath-quinn-0.11.x#773fceabb27f1e56132198dd960d4bd1493e0ed0"
 dependencies = [
  "cfg_aliases",
  "libc",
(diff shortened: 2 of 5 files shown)


iroh/src/magicsock/node_map.rs
```diff
@@ -166,15 +166,15 @@ impl NodeMap {
             .receive_relay(relay_url, src)
     }
 
-    pub(super) fn get_quic_mapped_addr_for_node_key(
+    pub(super) fn get_all_paths_add_for_node(
```
matheus23: Typo? Instead `get_all_paths_addr_for_node`?

iroh-base/src/node_addr.rs
```diff
@@ -36,9 +36,12 @@ use crate::{NodeId, PublicKey, RelayUrl};
 /// [discovery]: https://docs.rs/iroh/*/iroh/index.html#node-discovery
 /// [home relay]: https://docs.rs/iroh/*/iroh/relay/index.html
 /// [Relay server]: https://docs.rs/iroh/*/iroh/index.html#relay-servers
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(
+    derive_more::Debug, Clone, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash,
+)]
 pub struct NodeAddr {
     /// The node's identifier.
+    #[debug("{}", node_id.fmt_short())]
```
dignifiedquire: this should be reverted before merging

iroh-relay/src/client.rs
```diff
@@ -281,7 +281,7 @@ impl ClientBuilder {
         let conn = Conn::new(conn, self.key_cache.clone(), &self.secret_key).await?;
 
         event!(
-            target: "events.net.relay.connected",
+            target: "iroh::_events::net::relay::connected",
```
dignifiedquire, line 291: did you change the convention for events?

iroh/src/magicsock/transports.rs
```diff
@@ -376,6 +406,7 @@ impl UdpSender {
                         any_match = true;
                         match sender.send(*addr, src, transmit).await {
                             Ok(()) => {
+                                trace!("sent");
```
dignifiedquire: needs delete before merge

iroh/src/magicsock/transports.rs
```diff
@@ -391,6 +422,7 @@ impl UdpSender {
                         any_match = true;
                         match sender.send(url.clone(), *node_id, transmit).await {
                             Ok(()) => {
+                                trace!("sent");
```
dignifiedquire: needs delete before merge

iroh/src/magicsock/transports.rs
```diff
@@ -438,157 +475,232 @@ impl UdpSender {
...
+    /// Extracts the right [`Addr`] from the [`quinn_udp::Transmit`].
+    ///
+    /// Because Quinn does only know about IP transports we map other transports to private
+    /// IPv6 Unique Local Address ranges.  This extracts the transport addresses out of the
+    /// transmit's destination.
+    fn mapped_addr(&self, transmit: &quinn_udp::Transmit) -> io::Result<MultipathMappedAddr> {
+        self.msock
+            .metrics
+            .magicsock
+            .send_data
```
dignifiedquire: why is `send_data` increased if we call `mapped_addr`?

iroh/src/magicsock/transports.rs
```diff
@@ -438,157 +475,232 @@ impl UdpSender {
...
+                // need to respond from.
+                if let Some(src_ip) = quinn_transmit.src_ip {
+                    warn!(dst = ?mapped_addr, ?src_ip, dst_node = %node_id.fmt_short(),
+                        "oops, flub didn't think this would happen");
                 }
-                Err(ref err) => {
-                    warn!(dst = ?destination, "failed to send: {err:#}");
+
+                let sender = self.msock.node_map.node_state_actor(node_id);
+                let transmit = OwnedTransmit::from(quinn_transmit);
```
dignifiedquire, line 699: this is very painful, it means we have to allocate each packet that we send one more time 😭

dignifiedquire: I guess this only happens on the mixed path? which seems fine then

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -1,1417 +1,1005 @@
...
-    /// What this ping did to the [`NodeState`].
-    pub role: PingRole,
-    /// Whether the sender path should also be pinged.
-    ///
-    /// This is the case if an [`NodeState`] does not yet have a direct path, i.e. it has no
-    /// best_addr.  In this case we want to ping right back to open the direct path in this
-    /// direction as well.
-    pub needs_ping_back: Option<SendPing>,
-}
+// TODO: use this
```
dignifiedquire, line 69: lots of TODOs in this file

iroh/src/magicsock/endpoint_map/endpoint_state.rs
```diff
@@ -1,1417 +1,1005 @@
...
+    //
+    /// All connections we have to this remote node.
     ///
-    /// Note that sending datagrams to a node does not mean the node receives them.
-    last_used: Option<Instant>,
-    /// Last time we sent a call-me-maybe.
+    /// The key is the [`quinn::Connection::stable_id`].
+    connections: FxHashMap<usize, ConnectionState>,
+    /// Events emitted by Quinn about path changes.
+    #[allow(clippy::type_complexity)]
```
dignifiedquire: type alias?

iroh-base/src/node_addr.rs
flub: Oh, do you think so?  I found it really annoying to have to look at those full public keys when we debug-log a NodeAddr.  I considered using the alternative-debug format, but unfortunately tracing only has `?` and `%` support in their macros, for Debug and Display respectively.  I didn't thyink there's any strict rule about how a Debug should look?  It's not like you can `eval()` this back into a thing like e.g. Python.

iroh-relay/src/client.rs
flub: A long time ago, yes.  These ones were overlooked in that conversion.

The issue was that folks would use `RUST_LOG=iroh=debug` and the `events.abc` convention would mean we lost those.  Rather than tell folks to configure their logging config we changed to this convention instead.

iroh/src/magicsock/transports.rs
flub: Yeah, the tradeoff here is that this only happens for the Intial packet of each connection.  Additionally we are not copying the data itself, because that is already a `Bytes` instance.  So yeah, this was a consious tradeoff made with this design choice.

I guess this design choice isn't documented clearly enough yet.

iroh/src/endpoint.rs
```diff
@@ -150,12 +148,23 @@ impl Builder {
...
-    pub async fn bind(self) -> Result<Endpoint, BindError> {
+    pub async fn bind(mut self) -> Result<Endpoint, BindError> {
         let mut rng = rand::rng();
         let relay_map = self.relay_mode.relay_map();
         let secret_key = self
             .secret_key
             .unwrap_or_else(move || SecretKey::generate(&mut rng));
+
+        // Override some transport config settings.
+        self.transport_config
```
dignifiedquire: I wonder if we should check the config and print a warning if we change them, because this does remove the ability for users to configure some of these

iroh-base/src/node_addr.rs
dignifiedquire: it reduces the ability to reproduce information, that's why I am concerend about doing it by default. you can copy and paste the nodeid, relayurl and socket addrs currently, but suddenly you can't do it for the nodeid anymore

Frando: A sigil for alternate display in tracing would be the best solution indeed. Looked again and found https://github.com/tokio-rs/tracing/pull/3258 - I now pinged the tokio discord about how to move it forward (pr is open for 6 months without comments)

iroh/src/endpoint.rs
flub: Maybe.  Though the upstream API does not allow us to inspect the values :(

Cargo.toml
```diff
@@ -40,3 +40,20 @@ unexpected_cfgs = { level = "warn", check-cfg = ["cfg(iroh_docsrs)", "cfg(iroh_l
...
 [workspace.lints.clippy]
 unused-async = "warn"
+
+
+[patch.crates-io]
+iroh-quinn = { git = "https://github.com/n0-computer/quinn", branch = "main" }
+iroh-quinn-proto = { git = "https://github.com/n0-computer/quinn", branch = "main" }
+iroh-quinn-udp = { git = "https://github.com/n0-computer/quinn", branch = "main" }
+
+netwatch = { git = "https://github.com/n0-computer/net-tools", branch = "main" }
```
matheus23, line 50: Before we merge this, should we cut netwatch and quinn releases?
I guess there's just too much in-flux for us right now?

dignifiedquire: yeah, we should wait for a release until we are ready for an iroh release

Arqu (approved):
LGTM :D

PR #3384, opened 2025-07-10
refactor(iroh,iroh-relay): Remove legacy relay path, make websocket connections default

matheus23:
## Description

Split out from #3331. 

Switch to websockets and remove the old code path.

Also:
- I've removed `governor` for rate limiting
  - We used to guess the amount of bytes a frame was encoded as after we've decoded it, because that's how governor works (you do it on `Stream`s, not the underlying `AsyncRead`).
  - Instead we have our own couple of lines of rate-limiting logic (with tests) on the `AsyncRead` level.
  - The reason is we can't guess the size of websocket frames as easily anymore.

## Breaking Changes

In iroh:
- Removed `iroh::endpoint::Builder::relay_conn_protocol`. It's now always websockets
- Removed the `iroh::RelayProtocol` re-export (the type was removed in `iroh-relay`)

In iroh-relay:
- Removed `ClientBuilder::protocol`
- Removed `http::Protocol` type
- Renamed `frames_rx_ratelimited_total` metric to `bytes_rx_ratelimited_total`, which now tracks bytes not frames
- Removed `relay_accepts` and `websocket_accepts` metrics

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh-relay/src/server/streams.rs
+++ b/iroh-relay/src/server/streams.rs
@@ -281,8 +281,9 @@ impl Bucket {
         refill_period: time::Duration,
     ) -> Result<Self, InvalidBucketConfig> {
         // milliseconds is the tokio timer resolution
+        let refill = bytes_per_second.saturating_mul(refill_period.as_millis() as i64) / 1000;
         snafu::ensure!(
-            max > 0 && bytes_per_second > 0 && refill_period.as_millis() != 0,
+            max > 0 && bytes_per_second > 0 && refill_period.as_millis() as u32 > 0 && refill > 0,
             InvalidBucketConfigSnafu {
                 max,
                 bytes_per_second,
@@ -294,12 +295,13 @@ impl Bucket {
             max,
             last_fill: time::Instant::now(),
             refill_period,
-            refill: bytes_per_second * refill_period.as_millis() as i64 / 1000,
+            refill,
         })
     }
 
     fn update_state(&mut self) {
         let now = time::Instant::now();
+        // div safety: self.refill_period.as_millis() is checked to be non-null in constructor
         let refill_periods = now.saturating_duration_since(self.last_fill).as_millis() as u32
             / self.refill_period.as_millis() as u32;
         if refill_periods == 0 {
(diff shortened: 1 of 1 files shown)


iroh-relay/src/server/http_server.rs
```diff
@@ -1108,10 +1090,8 @@ mod tests {
         let public_key_a = key_a.public();
         let (client_a, rw_a) = tokio::io::duplex(10);
         let s = service.clone();
-        let handler_task = tokio::spawn(async move {
-            s.0.accept(Protocol::Relay, MaybeTlsStream::Test(rw_a))
-                .await
-        });
+        let handler_task =
+            tokio::spawn(async move { s.0.accept(MaybeTlsStream::Test(rw_a)).await });
```
dignifiedquire, line 1101: shouldn't we use `n0_future::task::spawn` for these? just wondering

iroh-relay/src/server/client.rs
```diff
@@ -883,33 +679,26 @@ mod tests {
...
-            RelayCodec::test(),
-        ));
-        let mut stream = RateLimitedRelayedStream::new(stream, limiter, Default::default());
+        let mut frame_writer = RelayedStream::test_client(io_write);
+        // Rate limiter allowing LIMIT bytes/s
+        let mut stream = RelayedStream::test_server_limited(io_read, LIMIT / 10, LIMIT);
 
         // Prepare a frame to send, assert its size.
-        let data = Bytes::from_static(b"hello world!!");
+        let data = Bytes::from_static(b"hello world!!1elf");
```
dignifiedquire, line 695: why this addition?

matheus23: :sweat_smile: The `LIMIT` is configured to be 50 bytes above. Either I reduce the limit or add some stuff to the data to make it bigger (The encoding got a little more efficient).

iroh-relay/src/server/http_server.rs
matheus23: Well, this is in the relay server code, so I figured it doesn't need to be, since we don't need to be able to compile this to Wasm.

iroh-relay/src/client/conn.rs
```diff
@@ -181,56 +140,34 @@ impl Stream for Conn {
...
+                Poll::Ready(Some(message))
+            }
+            #[cfg(wasm_browser)]
+            Some(ws_stream_wasm::WsMessage::Binary(vec)) => {
+                let frame = Frame::from_bytes(Bytes::from(vec), &self.key_cache)?;
+                Poll::Ready(Some(ReceivedMessage::try_from(frame)))
+            }
+            #[cfg(wasm_browser)]
+            Some(msg) => {
+                tracing::warn!(?msg, "Got websocket message of unsupported type, skipping.");
```
dignifiedquire: please use `warn!` directly

iroh-relay/src/server/client.rs
dignifiedquire: lol

dignifiedquire:
ci is still very sad

Issue #3198, opened 2025-02-24
deps: update to rand@0.9 and getrandom@0.3


dignifiedquire:
Blocked on at least

- [ ] https://github.com/dalek-cryptography/curve25519-dalek/pull/729
- [ ] merging https://github.com/quinn-rs/quinn/pull/2144 into `iroh-quinn`
- [x] ~~https://github.com/ihrwein/backoff/issues/74~~
- [x] https://github.com/boinkor-net/governor/issues/259
- [x] https://github.com/dariusc93/rust-igd/issues/12
- [x] https://github.com/sancane/rustun/issues/45
- [x] https://github.com/kolapapa/surge-ping/issues/40
- [ ] https://github.com/pubky/pkarr/pull/186

## Deps that already updated

- `tungstenite` has updated on `0.26`
- `pkarr` has updated on version 3

maackle:
iroh 0.34.1 and rand 0.9.1 (latest versions as of writing this comment) are incompatible. This won't compile:

    SecretKey::generate(rand::rngs::OsRng);

```
the trait bound `OsRng: ssh_key::rand_core::CryptoRngCore` is not satisfied
two types coming from two different versions of the same crate are different types even if they look the same
you can use `cargo tree` to explore your dependency tree
required for `OsRng` to implement `ssh_key::rand_core::CryptoRngCore`
```

Is there a simple workaround?

dignifiedquire:
you will have to use rand@0.8 to pass to iroh until all the referenced deps are updated

Dietr1ch:
It seems that this isn't blocked any more.

The last outstanding, unmerged issue got closed in favour of https://github.com/dalek-cryptography/curve25519-dalek/pull/777, which is already merged.

dignifiedquire:
This is great, but need to be careful though, this is currently only released as a pre release, so I am not sure we will want to use that

PR #3386, opened 2025-07-12
feat!: update to edition 2024 and update deps to latest

### LINKED ISSUE #3385: deps(netdev): build failure on FreeBSD/aarch64
Netdev merged a change that fixes builds on FreeBSD/aarch64, so bumping netdev to version 0.35.3 or 0.36.0 will then fix build failures for all downstream consumers of Iroh (dumbpipe, sendme, external projects like DeltaChat)

dignifiedquire:
## Description

Updates dependencies where possible.

Closes #3385 

Depends on 
- [x] https://github.com/n0-computer/net-tools/pull/30

## Breaking Changes

All crates are now `edition2024`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,3 +40,8 @@ unexpected_cfgs = { level = "warn", check-cfg = ["cfg(iroh_docsrs)", "cfg(iroh_l
 
 [workspace.lints.clippy]
 unused-async = "warn"
+
+
+[patch.crates-io]
+portmapper = { git = "https://github.com/n0-computer/net-tools", branch = "deps-update" }
+netwatch = { git = "https://github.com/n0-computer/net-tools", branch = "deps-update" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,5 +43,5 @@ unused-async = "warn"
 
 
 [patch.crates-io]
-portmapper = { git = "https://github.com/n0-computer/net-tools", branch = "deps-update" }
-netwatch = { git = "https://github.com/n0-computer/net-tools", branch = "deps-update" }
+portmapper = { git = "https://github.com/n0-computer/net-tools", branch = "main" }
+netwatch = { git = "https://github.com/n0-computer/net-tools", branch = "main" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,8 +40,3 @@ unexpected_cfgs = { level = "warn", check-cfg = ["cfg(iroh_docsrs)", "cfg(iroh_l
 
 [workspace.lints.clippy]
 unused-async = "warn"
-
-
-[patch.crates-io]
-portmapper = { git = "https://github.com/n0-computer/net-tools", branch = "main" }
-netwatch = { git = "https://github.com/n0-computer/net-tools", branch = "main" }
--- a/iroh-relay/src/server/http_server.rs
+++ b/iroh-relay/src/server/http_server.rs
@@ -9,11 +9,11 @@ use http::{
     response::Builder as ResponseBuilder,
 };
 use hyper::{
+    HeaderMap, Method, Request, Response, StatusCode,
     body::Incoming,
     header::{HeaderValue, SEC_WEBSOCKET_ACCEPT, UPGRADE},
     service::Service,
     upgrade::Upgraded,
-    HeaderMap, Method, Request, Response, StatusCode,
 };
 use n0_future::time::Elapsed;
 use nested_enum_utils::common_fields;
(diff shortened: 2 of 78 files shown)


iroh-relay/src/server/http_server.rs
```diff
@@ -518,10 +518,13 @@ impl RelayService {
                     async move {
                         match hyper::upgrade::on(&mut req).await {
                             Ok(upgraded) => {
-                                if let Err(err) = this.0.relay_connection_handler(upgraded).await {
-                                    warn!("error accepting upgraded connection: {err:#}",);
-                                } else {
-                                    debug!("upgraded connection completed");
+                                match this.0.relay_connection_handler(upgraded).await {
```
rklaehn: What was wrong with before? Clippy being bipolar again? I would have sworn that the if let is the favoured way to write this for clippy...

iroh/src/discovery.rs
```diff
@@ -699,7 +699,7 @@ impl DiscoverySubscribers {
         }
     }
 
-    pub(crate) fn subscribe(&self) -> impl Stream<Item = Result<DiscoveryItem, Lagged>> {
+    pub(crate) fn subscribe(&self) -> impl Stream<Item = Result<DiscoveryItem, Lagged>> + use<> {
```
rklaehn, line 702: What even is this?

iroh/src/test_utils.rs
```diff
@@ -483,7 +483,7 @@ pub(crate) mod pkarr_dns_state {
         node_info: &NodeInfo,
         origin: &str,
         ttl: u32,
-    ) -> impl Iterator<Item = hickory_resolver::proto::rr::Record> + 'static {
+    ) -> impl Iterator<Item = hickory_resolver::proto::rr::Record> + 'static + use<> {
```
rklaehn: Isn't this redundant?

iroh/src/discovery.rs
flub: https://blog.rust-lang.org/2024/10/17/Rust-1.82.0/#precise-capturing-use-syntax

rklaehn: Never mind. https://doc.rust-lang.org/edition-guide/rust-2024/rpit-lifetime-capture.html

iroh-relay/src/server/http_server.rs
dignifiedquire: 🤷 some tool was unhappy

iroh/src/test_utils.rs
dignifiedquire: fixed

Cargo.toml
```diff
@@ -40,3 +40,8 @@ unexpected_cfgs = { level = "warn", check-cfg = ["cfg(iroh_docsrs)", "cfg(iroh_l
 
 [workspace.lints.clippy]
 unused-async = "warn"
+
+
+[patch.crates-io]
+portmapper = { git = "https://github.com/n0-computer/net-tools", branch = "main" }
+netwatch = { git = "https://github.com/n0-computer/net-tools", branch = "main" }
```
matheus23: Let's not forget to remove the patch :v:

rklaehn:
There are lots of places where we have fn foo(&self) -> impl Bar + use<>.

As far as I can see all these places can be just replaced with -> impl Bar + 'static, which for me reads much more natural than the weird `use<>` syntax.

dignifiedquire:
> As far as I can see all these places can be just replaced with -> impl Bar + 'static, which for me reads much more natural than the weird `use<>` syntax.

But asfaiu that does not mean the same thing though

rklaehn:
> > As far as I can see all these places can be just replaced with -> impl Bar + 'static, which for me reads much more natural than the weird `use<>` syntax.
> 
> But asfaiu that does not mean the same thing though

Not in general, but in this case? The returned thing is supposed to have no lifetime dependencies on any of the parameters that are passed by reference (which is just &self). So how is that not the same as 'static? What other source of lifetimes other than the args is there?

Cargo.toml
dignifiedquire: done

dignifiedquire:
`+ 'static` definitely does sth different, the compiler is quite upset if I try to use it

rklaehn:
> `+ 'static` definitely does sth different, the compiler is quite upset if I try to use it

If you just use 'static everywhere instead of `use<>` everything still compiles. https://github.com/n0-computer/iroh/pull/3392

I will have to try out in play.rust-lang.org what the rules are from calling use<> from 'static or otherwise.

dignifiedquire:
> I will have to try out in play.rust-lang.org what the rules are from calling use<> from 'static or otherwise.

asfaict `use<>` avoids capturing the lifetime of `&self`, which I think `'static` does not do

rklaehn:
> > I will have to try out in play.rust-lang.org what the rules are from calling use<> from 'static or otherwise.
> 
> asfaict `use<>` avoids capturing the lifetime of `&self`, which I think `'static` does not do

`'static` does not capture `&self`. Static means fully self-contained, so you can use the future even after self goes out of scope. That is why tokio::task::spawn requires 'static. I really don't get what the difference is between use<> and 'static in that sense.

matheus23:
We could merge this as-is probably, but we could follow this up with #3198 as well.

But we'd need another iroh-quinn release where we merge upstream's https://github.com/quinn-rs/quinn/pull/2144

PR #3389, opened 2025-07-14
feat(iroh,iroh-relay)!: Use stride instead of custom split protocol, send ECN bits

matheus23:
## Description

Based on #3331 

- Enable doing active queue management in relays in the future (we now send ECN bits with messages!)
- Avoid doing our own packet merging & splitting, instead just forward quinn's `Transmit` format (bytes + optional stride)

## Breaking Changes

The relay wire protocol changed: All relayed messages now contain at least an additional ECN byte.
They *might* be accidentally compatible when GSO is not enabled, but they're likely not.
This means this version of iroh can't connect to older relays or older clients on newer relays.

- `ClientToRelayMsg::SendPacket` was removed in favor of `ClientToRelayMsg::Datagrams`
- `RelayToClientMsg::ReceivedPacket` was removed in favor of `RelayToClientMsg::Datagrams`
- `FrameType` has changed variants:
  - `SendPacket` and `RecvPacket` were removed
  - `ClientToRelayDatagram` and `ClientToRelayDatagramBatch` were added
  - `RelayToClientDatagram` and `RelayToClientDatagramBatch` were added

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-relay/src/protos/relay.rs
+++ b/iroh-relay/src/protos/relay.rs
@@ -411,8 +411,7 @@ impl ClientToRelayMsg {
                 + datagrams.encoded_len()
             }
         };
-        1 // frame type (all frame types currently encode as 1 byte varint)
-        + payload_len
+        self.typ().encoded_len() + payload_len
     }
 
     /// Tries to decode a frame received over websockets.


iroh-relay/src/protos/common.rs
```diff
@@ -23,10 +23,14 @@ pub enum FrameType {
     ServerConfirmsAuth = 2,
     /// The server frame type for authentication denial
     ServerDeniesAuth = 3,
+    /// 32B dest pub key + ECN bytes + one datagram's content
+    ClientToRelayDatagram = 4,
     /// 32B dest pub key + ECN byte + segment size u16 + datagrams contents
-    SendPacket = 4,
+    ClientToRelayDatagrams = 5,
```
flub: `ClientToRelayDatagramBatch`?  The `s` thing is somewhat subtle...

iroh/src/magicsock/transports/relay.rs
```diff
@@ -304,26 +307,19 @@ impl RelaySender {
...
-/// Split a transmit containing a GSO payload into individual packets.
-///
-/// This allocates the data.
-///
-/// If the transmit has a segment size it contains multiple GSO packets.  It will be split
-/// into multiple packets according to that segment size.  If it does not have a segment
-/// size, the contents will be sent as a single packet.
+/// Translate a UDP transmit to the `Datagrams` type for sending over the relay.
 // TODO: If quinn stayed on bytes this would probably be much cheaper, probably.  Need to
 // figure out where they allocate the Vec.
```
flub: Feel free to remove this comment...  I remember writing this.  At some point Quinn was using bytes in a Transmit I think.  But this is anyway a tricky thing because we cross the `IoSlice` boundary as well.  I don't think it deserves being so prominent in the code.

iroh/src/magicsock/transports/relay/actor.rs
```diff
@@ -680,11 +666,11 @@ impl ActiveRelayActor {
...
         match msg {
-            RelayToClientMsg::ReceivedPacket {
-                src_key: remote_node_id,
-                content,
+            RelayToClientMsg::Datagrams {
+                remote_node_id,
+                datagrams,
             } => {
-                trace!(len = %content.len(), "received msg");
+                trace!(len = %datagrams.contents.len(), "received msg");
```
flub: Totally a nit, but I don't think you need `%` for this, tracing can record `usize` directly without having to go via `Display`.

flub (approved):

flub (approved):
LGTM

PR #3392, opened 2025-07-16
refactor: try to use 'static to denote impls that don't reference argument lifetimes

rklaehn:
## Description

Try to use 'static instead of the new and somewhat weird use<> syntax to express that a function return argument does not reference any of the lifetimes of the function arguments or &self.

## Breaking Changes

'static seems to be subtly different than use<> even though I don't really understand the difference.

## Notes & open questions

I find 'static more clear to express that the result does not capture any lifetimes... YMMV.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2976,8 +2976,9 @@ dependencies = [
 
 [[package]]
 name = "netwatch"
-version = "0.6.0"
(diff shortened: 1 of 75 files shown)


dignifiedquire:
done manually

Issue #3395, opened 2025-07-17
bug: DHT record with no relay gets published


rklaehn:
**Describe the bug**

When writing a simple echo example with iroh 0.90, the published DHT pkarr record is empty. It does not contain information about the relay.

**Relevant Logs**

N/A

**Expected behavior**
Published DHT records should contain the relay url.

**Platform(s)**
OSX

**Additional Context / Screenshots / GIFs**

Record is **empty**, not missing. DHT is working as expected.

I think what is happening is that we are publishing too early, before the relay is known, and then for some reason the record does not get overwritten.

DNS discovery does not have the problem, presumably because writing to the dns is more deterministic than writing to the DHT.

<img width="598" height="364" alt="Image" src="https://github.com/user-attachments/assets/0da5d3b2-e48b-405c-873b-fcc68e7305e0" />

rklaehn:
Ok, nailed this down.

We are publishing all the info every time we get new info. We get the info one by one in the course of milliseconds. Apparently pkarr does not allow concurrent publish, which makes sense since it will make a mess.

```
2025-07-21T09:39:33.456793Z DEBUG iroh::discovery::pkarr: creating pkarr publisher that publishes to https://dns.iroh.link/pkarr
2025-07-21T09:39:33.534817Z DEBUG iroh::discovery::pkarr::dht: publishing NodeData { relay_url: None, direct_addresses: {192.168.0.242:64092}, user_data: None }
2025-07-21T09:39:33.535469Z DEBUG pkarr_publish{me=6384147509}: iroh::discovery::pkarr: Publish node info to pkarr (info changed)
2025-07-21T09:39:34.168614Z DEBUG iroh::discovery::pkarr::dht: publishing NodeData { relay_url: None, direct_addresses: {82.76.153.174:64092, 192.168.0.242:64092}, user_data: None }
Node ID: 63841475093fdfb9d6d5bb4f770fa19dcf07fc1ee87ee3f005a80f1902d0b5b3
Full address: NodeAddr { node_id: PublicKey(63841475093fdfb9d6d5bb4f770fa19dcf07fc1ee87ee3f005a80f1902d0b5b3), relay_url: Some(RelayUrl("https://euc1-1.relay.n0.iroh-canary.iroh.link./")), direct_addresses: {82.76.153.174:64092, 192.168.0.242:64092} }
[...]

PR #3397, opened 2025-07-22
fix: fix dht publishing at startup

rklaehn:
## Description

Before, you would publish two times in very close succession, which caused an error during the second publish. pkarr will notice if you try to publish while the previous publish is still ongoing, and abort.

## Breaking Changes

fn discovery::pkarr::dht::Builder::initial_publish_delay is removed, since this does not exist anymore.

## Notes & open questions

Note: I think the whole dht publishing code would benefit from some cleanup since a few things have changed. But this is just a minimal fix to get things to work again.

fixes https://github.com/n0-computer/iroh/issues/3395

### DIFF
--- a/iroh/src/discovery/pkarr/dht.rs
+++ b/iroh/src/discovery/pkarr/dht.rs
@@ -258,12 +258,17 @@ impl DhtDiscovery {
     /// Periodically publishes the node address to the DHT and/or relay.
     async fn publish_loop(self, keypair: SecretKey, signed_packet: SignedPacket) {
         let this = self;
-        let public_key = pkarr::PublicKey::try_from(keypair.public().as_bytes())
-            .expect("valid public key");
+        let public_key =
+            pkarr::PublicKey::try_from(keypair.public().as_bytes()).expect("valid public key");
         let z32 = public_key.to_z32();
         loop {
             // If the task gets aborted while doing this lookup, we have not published yet.
-            let prev_timestamp = this.0.pkarr.resolve_most_recent(&public_key).await.map(|p| p.timestamp());
+            let prev_timestamp = this
+                .0
+                .pkarr
+                .resolve_most_recent(&public_key)
+                .await
+                .map(|p| p.timestamp());
             let res = this.0.pkarr.publish(&signed_packet, prev_timestamp).await;
             match res {
                 Ok(()) => {


iroh/src/discovery/pkarr/dht.rs
```diff
@@ -273,15 +258,18 @@ impl DhtDiscovery {
...
-            // TODO: publish takes a compare-and-swap timestamp, make use of that.
-            let res = this.0.pkarr.publish(&signed_packet, None).await;
+            // If the task gets aborted while doing this lookup, we have not published yet.
+            let prev_timestamp = this
+                .0
+                .pkarr
+                .resolve_most_recent(&public_key)
+                .await
+                .map(|p| p.timestamp());
+            let res = this.0.pkarr.publish(&signed_packet, prev_timestamp).await;
```
matheus23, line 272: Is the fix to solve the TODO, i.e. give it a compare-and-swap timestamp? This way the newer timestamps will win?

matheus23 (approved):
Looks good. If it's possible to write a test for this, that might be kinda nice though.

iroh/src/discovery/pkarr/dht.rs
Nuhvi: CAS is one of the concurrency issues, but in this specific case, the error is `ConflictRisk` which is basically the DHT node telling you: "I am actually publishing another MutableItem, and you haven't set that in your CAS, so you might not be aware of it, and might lose data because of that".

When you `rseolve_most_recent` you basically become aware of that packet, and tell the DHT (and relays): "I know this packet was published, and I still want to publish this one'.

All of this is to protect you from losing data, but maybe we should add a `publish_force()` for cases where the user doesn't need this protection, as in the case of Iroh where the node info doesn't coexist with other application's information.

Another relevant thing here, is that cancelling the previous publishing tasks, doesn't really do anything in the DHT, if it did, maybe the `ConflictRisk` wouldn't be emitted, as the DHT realises that you are no longer interested in the data in that update. That is one feature I intend to check if I can add to the dht this week.

rklaehn:
> Looks good. If it's possible to write a test for this, that might be kinda nice though.

We do have DHT tests, but they are disabled as flaky. I guess it would make sense to test this against the actual global DHT instead of a test DHT, but that is a bit of a weird test.

I suspect that the issue would not happen with a locally running test DHT since publishing would be super fast.

PR #3396, opened 2025-07-18
fix(iroh): Keep track of non-best-addr path validation and switch to them if best addr is soon outdated

matheus23:
## Description

I'd consider this mostly a workaround.

See also the humongous comment above `TRUST_UDP_ADDR_SOON_OUTDATED`.

## Change checklist

- [ ] Self-review.
- [ ] Documentation updates following the [style guide](https://rust-lang.github.io/rfcs/1574-more-api-documentation-conventions.html#appendix-a-full-conventions-text), if relevant.
- [ ] Tests if relevant.

PR #3398, opened 2025-07-22
fix(iroh): Also keep track of non-best path's validity (v0.35)

matheus23:
## Description

This is an alternative to #3396 but also refactors the code a bit more.

Instead of trying to switch to a different address when the best address becomes outdated *soon*, we instead track the path validity for all paths we have the same way.

This means the `BestAddr` struct and module completely disappears. Instead we replace it with a `PathValidity` struct and module and track it inside `PathState`.

Then we implement the "stickyness" that the best address had (i.e. not recalculating the best address all the time and sticking to a path as long as it's valid) by tracking a `best: UdpSendAddr` and a `best_ipv4: UdpSendAddr` in `NodeUdpPaths`.

This also tries to generally update the `NodeUdpPaths` when we *receive* data instead of when we try to send.
This means that `NodeUdpPaths::get_send_addr` *might* be slightly outdated, but it'll snap back to the "correct" thing once we receive any pings or data or call me maybes etc.

I think this is a much cleaner approach, albeit with a bigger diff.

## Breaking Changes

There shouldn't be any breaking changes.

## Notes & open questions

There's some stuff like clearing the best address when `NodeMap::reset` is called that I didn't implement. I'm not sure if there's any point to that code path, it seems to be only called on `Endpoint::stop()`, so I don't know why we'd want to clear the state there.

Maybe more? I'm curious about what you think.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/magicsock/node_map/node_state.rs
+++ b/iroh/src/magicsock/node_map/node_state.rs
@@ -709,6 +709,7 @@ impl NodeState {
                     PathState::new(self.node_id, SendAddr::from(addr), source.clone(), now)
                 });
         }
+        self.udp_paths.update_to_best_addr(now);
         let paths = summarize_node_paths(&self.udp_paths.paths);
         debug!(new = ?new_addrs , %paths, "added new direct paths for endpoint");
     }


iroh/src/magicsock/node_map/path_state.rs
```diff
@@ -128,7 +128,16 @@ impl PathState {
...
             }
         }
-        self.recent_pong = Some(r);
+
+        self.validity = PathValidity::new(r);
+    }
+
+    pub(super) fn receive_payload(&mut self, now: Instant) {
+        self.last_payload_msg = Some(now);
+        // TODO(matheus23): It's not necessarily UDP. Kinda weird to have this thing
```
dignifiedquire: I think at least for tracking it is useful to have the different sources

dignifiedquire: but we could also simplify it

iroh/src/magicsock/node_map/path_validity.rs
```diff
@@ -0,0 +1,139 @@
...
+#[derive(Debug, Clone)]
+struct Inner {
+    recent_pong: PongReply,
+    confirmed_at: Instant,
+    trust_until: Instant,
+}
+
+#[derive(Debug)]
+pub(super) enum Source {
+    ReceivedPong,
```
dignifiedquire, line 32: wait, isn't this UDP based as well? I guess it really doesn't make sense to distinguish these anymore

iroh/src/magicsock/node_map/path_validity.rs
```diff
@@ -0,0 +1,139 @@
...
+    // PongTimeout,
+    MatchesOurLocalAddr,
+}
+
+impl Source {
+    fn trust_duration(&self) -> Duration {
+        match self {
+            Source::ReceivedPong => TRUST_UDP_ADDR_DURATION,
+            // // TODO: Fix time
+            // Source::BestCandidate => Duration::from_secs(60 * 60),
```
dignifiedquire: what about the `BestCandidate` situation, how is that handled now?

iroh/src/magicsock/node_map/path_state.rs
```diff
@@ -44,11 +47,15 @@ pub(super) struct PathState {
     /// The time this endpoint was last advertised via a call-me-maybe DISCO message.
     pub(super) call_me_maybe_time: Option<Instant>,
 
-    /// The most recent [`PongReply`].
+    // /// The most recent [`PongReply`].
+    // ///
+    // /// Previous replies are cleared when they are no longer relevant to determine whether
+    // /// this path can still be used to reach the remote node.
+    // pub(super) recent_pong: Option<PongReply>,
```
dignifiedquire: what about this? not needed anymore?

iroh/src/magicsock/node_map/path_validity.rs
matheus23: `BestCandidate` used to be set in `NodeUdpPaths::assign_best_addr_from_candidates_if_empty`, which in my experience was only called on the receive side the very first time you'd send data from the magic socket.

For some reason it doesn't have a best address set at that point, so it instead picks one from all the `PathState`s it has and assigns the best it can find with `Source::BestCandidate`.

I... am really unsure why in that special case it should trust this candidate for **one hour**. This seems ridiculously high and totally out of proportion with TRUST_UDP_ADDR_DURATION otherwise.

In practice I don't think this is all to bad (even if the best addr never expires for an hour). In theory that allows an attacker to waste lots of resources on our end if they manage to spoof an address from the other side.
Or it can mean we'd try sending on an address that isn't actually reachable anymore (doesn't pong) for quite a while before we fall back to the relay, but usually these cases would also coincide with the addrs in the call me maybe changing, thus that'd clear the best addr as well.

Idk. All in all, very weird code. But writing this comment makes me want to jump back into iroh 0.35 code and see why this `NodeUdpPaths::assign_best_addr_from_candidates_if_empty` function was needed at all.

iroh/src/magicsock/node_map/path_state.rs
matheus23: Yeah sorry, I'll remove it.

The recent pong is now tracked in `PathValidity`.
`PathValidity` is like a mix between `Option<PongReply>` and what used to be `BestAddr`.

iroh/src/magicsock/node_map/path_validity.rs
dignifiedquire: > For some reason it doesn't have a best address set at that point, so it instead picks one from all the `PathState`s it has and assigns the best it can find with `Source::BestCandidate`.

I think the idea was to avoid flip flopping, and keeping this stable, _unless_ new information comes in

iroh/src/magicsock/node_map/path_validity.rs
matheus23: Yeah pongs are also UDP.

Perhaps we should rename `Source::Udp` to something like `Source::QuicDatagrams` or so.

Basically: anyone on the internet can produce packets that will trigger path validity extension of type `Source::QuicDatagrams`, but only someone who knows the `SecretKey` to the other side can produce a path validity extension of type `Source::ReceivedPong`.

It would've made sense to have different trust timeouts for each case, but apparently we never had them.

With multipath, all of this will just be handled with path challenges, which is just so, *so* much better :sweat_smile:

iroh/src/magicsock/node_map/path_state.rs
matheus23: Yeah the naming is just weird.
`receive_payload` is also called for *relay* datagrams, not only UDP ones.

I've renamed `Source::Udp` to `Source::QuicPayload`.

iroh/src/magicsock/node_map/path_validity.rs
matheus23: I've renamed `Source::Udp` to `Source::QuicPayload`

iroh/src/magicsock/node_map/path_state.rs
matheus23: Removed the commented out code.

iroh/src/magicsock/node_map/path_validity.rs
matheus23: It shouldn't flip flop though, at least not more often than every `TRUST_UDP_ADDR_DURATION`, since once you validate a path for that duration, you don't pick another until it expires :/

iroh/src/magicsock/node_map/path_validity.rs
```diff
@@ -0,0 +1,128 @@
...
+/// - For [`Source::trust_duration`] longer starting at the most recent
+///   received application payload *while the path was valid*.
+#[derive(Debug, Clone, Default)]
+pub(super) struct PathValidity(Option<Inner>);
+
+#[derive(Debug, Clone)]
+struct Inner {
+    recent_pong: PongReply,
+    confirmed_at: Instant,
+    trust_until: Instant,
```
flub, line 27: It seems these always have the same relation.  Why store both?

This is mostly an observation, it doesn't bother met that much if you want to keep it.

iroh/src/magicsock/node_map/node_state.rs
```diff
@@ -713,13 +717,12 @@ impl NodeState {
...
         self.last_full_ping = None;
-        self.udp_paths
-            .best_addr
-            .clear(ClearReason::Reset, self.relay_url.is_some());
 
         for es in self.udp_paths.paths.values_mut() {
             es.last_ping = None;
         }
+
+        self.udp_paths.update_to_best_addr(Instant::now());
```
flub, line 726: Have you considered removing reset entirely if we never use it for a live endpoint anyway?

iroh/src/magicsock/node_map/udp_paths.rs
```diff
@@ -84,97 +97,75 @@ impl NodeUdpPaths {
     }
 
     #[cfg(test)]
-    pub(super) fn from_parts(paths: BTreeMap<IpPort, PathState>, best_addr: BestAddr) -> Self {
+    pub(super) fn from_parts(paths: BTreeMap<IpPort, PathState>, best: UdpSendAddr) -> Self {
         Self {
             paths,
-            best_addr,
-            chosen_candidate: None,
+            best_ipv4: best, // we only use ipv4 addrs in tests
```
flub, line 104: Could be worth a `debug_assert!` perhaps?  I wouldn't be surprised if I ended up writing tests with IPv6 at some point and it would catch me out.  That probably works entirely fine - Quinn's test suite is entirely IPv6 and doesn't seem to cause issues anywhere.

flub (approved):
Nice!  Contains some changes I've long wanted :smile:
The path validity seems to be a much better abstraction than best-addr was.  If this has seen enough testing I don't think there's anything blocking?

iroh/src/magicsock/node_map/node_state.rs
matheus23: It's removed in `main`.

iroh/src/magicsock/node_map/path_validity.rs
matheus23: Yeah I agree.

There's some more "duplicated state" stored here with some fields in `PongReply`.

I think refactoring this further, I'd clean up the `PathValidity` struct a little more. At this point it's literally a combination of the fields in `BestAddr` with another `PongReply`.

iroh/src/magicsock/node_map/path_validity.rs
matheus23: Actually that's incorrect (and we've observed more frequent flip-flopping), because we'll usually have multiple paths that hole-punch successfully, and they'll all turn `Valid` at around the same time (and we'll choose the lowest latency ones).

Then they all expire at roughly the same time, at which point we flip between them back and forth a bit.

It's... *really hard* to fix this flip-flopping. (1) because both ends of the pipe will usually choose to switch the best address at around the same time (after TRUST_UDP_ADDR_DURATION) and (2) it's very hard to agree on "the same" path on both ends of the connection (which would be the most stable configuration) and (3) the latency of a path goes up the more it's used.

Ideally all of this becomes irrelevant once we integrate path validation with our QUIC stack. At that point there's no reason for both ends to "agree" on the path to send and receive on and it's should be rare to "accidentally" lose path validation packets.

dignifiedquire:
closing as we don't intend to do a release of 0.35 with this

PR #3400, opened 2025-07-23
fix(iroh): Track path validity for all paths and replace `BestAddr` with `PathValidity`

matheus23:
## Description

#3398 but now rebased on `main`. See its description for more information.

There's only one small change: We don't need to think about `NodeMap::reset`, as that's removed in `main`.

## Breaking Changes

None.

## Change checklist

- [x] Self-review.


iroh/src/magicsock/node_map/udp_paths.rs
```diff
@@ -84,97 +97,75 @@ impl NodeUdpPaths {
...
+        } else if path.validity.is_outdated(now) {
+            UdpSendAddr::Outdated((*ipp).into())
+        } else {
+            UdpSendAddr::Unconfirmed((*ipp).into())
         }
     }
 }
+
+#[derive(PartialEq, Eq)]
+struct ReverseOrd<N: PartialOrd + Ord + PartialEq + Eq>(N);
```
dignifiedquire, line 174: maybe some docs?

iroh/src/magicsock/node_map/udp_paths.rs
```diff
@@ -84,97 +97,75 @@ impl NodeUdpPaths {
...
-                        || (pong.latency == best_latency && ipp.ip().is_ipv6()) =>
-                {
-                    Some(pong)
+    pub(super) fn best_addr(&self, have_ipv6: bool, now: Instant) -> UdpSendAddr {
+        let Some((ipp, path)) = self
+            .paths
+            .iter()
+            .filter(|(ipp, _)| have_ipv6 || ipp.ip.is_ipv4())
+            .max_by_key(|(ipp, path)| {
+                // We find the best by sorting on a key of type (Option<ReverseOrd<Duration>>, Option<ReverseOrd<Duration>>, bool)
```
dignifiedquire, line 140: maybe convert this into a doc comment on the function?

dignifiedquire (approved):

PR #3402, opened 2025-07-23
test(iroh): Make `endpoint_relay_connect_loop` not flaky

matheus23:
## Description

`endpoint_relay_connect_loop` was flaky before. This *should* fix this issue. It was last marked flaky (but for windows only) in #3354.

What seems to have happened:
- The test spawns a relay server
- The test spawns a server iroh Endpoint
- The server Endpoint does a net report with QAD probes
- For some reason the spawned relay server is very slow in responding to QAD requests (>3s)
- The QAD probes time out, the server Endpoint ends up without a home relay
- The test spawns a client iroh Endpoint
- The client Endpoint tries to connect for 30s
- The server Endpoint doesn't do another net report for 30s, though, so never ends up being reachable
- The client Endpoint times out.

To work around this, I'm starting the server endpoint and waiting for it to have a relay address.
IMO this is reasonable to do in tests.

I've also made the tests use `Connection::close` and `Connection::closed` properly and removed `SendStream::stopped` and `RecvStream::read_to_end(0)` calls.

## Notes

There's some other drive-by changes. Sorry about that, but IMO they're kinda too small for their own PRs:
- `net_report` thought that `Watchable::set` would return `Err` when there's no more watchers listening, but that's incorrect: It returns `Err` from `set` when the value set is the same as the currently stored value.
- I've also made some small cosmetic changes to `net_report`
- I've removed a 1600 bytes allocation from the hot path of receiving relay items in `ActiveRelayActor`.

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh/src/net_report.rs
+++ b/iroh/src/net_report.rs
@@ -450,7 +450,7 @@ impl Client {
                             PROBES_TIMEOUT,
                             run_probe_v4(ip_mapped_addrs, relay_node, quic_client, dns_resolver),
                         ))
-                        .instrument(info_span!("QAD IPv4", %relay_url)),
+                        .instrument(info_span!("QAD-IPv4", %relay_url)),
                 );
             }
 
@@ -468,7 +468,7 @@ impl Client {
                             PROBES_TIMEOUT,
                             run_probe_v6(ip_mapped_addrs, relay_node, quic_client, dns_resolver),
                         ))
-                        .instrument(info_span!("QAD IPv6", %relay_url)),
+                        .instrument(info_span!("QAD-IPv6", %relay_url)),
                 );
             }
         }


iroh/src/net_report.rs
```diff
@@ -700,30 +700,28 @@ async fn run_probe_v4(
     };
 
     let observer = Watchable::new(None);
-    let ob = observer.clone();
```
dignifiedquire, line 703: what about probev6?

matheus23: Right! Sorry will do the same for that fn.

matheus23: Fixed!

iroh/src/endpoint.rs
```diff
@@ -2452,65 +2452,67 @@ mod tests {
...
+                    info!(%i, peer = %node_id.fmt_short(), "accepted connection");
+                    let (mut send, mut recv) = conn.accept_bi().await.e()?;
+                    let mut buf = vec![0u8; chunk_size];
+                    for _i in 0..n_chunks_per_client {
+                        recv.read_exact(&mut buf).await.e()?;
+                        send.write_all(&buf).await.e()?;
                     }
-                    Ok::<_, Error>(())
+                    send.finish().e()?;
+                    conn.closed().await; // we're the last to receive data, so we wait for close
```
flub: Comment is wrong?  It's the last to **send** data.  Behaviour is right though.

iroh/src/magicsock/transports/relay/actor.rs
```diff
@@ -615,15 +615,9 @@ impl ActiveRelayActor {
...
-                    // TODO: This allocation is *very* unfortunate.  But so is the
-                    // allocation *inside* of PacketizeIter...
-                    let batch = std::mem::replace(
-                        &mut send_datagrams_buf,
-                        Vec::with_capacity(SEND_DATAGRAM_BATCH_SIZE),
-                    );
                     // TODO(frando): can we avoid the clone here?
                     let metrics = self.metrics.clone();
-                    let packet_iter = batch.into_iter().map(|item| {
+                    let packet_iter = send_datagrams_buf.drain(..).map(|item| {
```
flub, line 620: :facepalm: :heart: 

Though I wonder if this might not have been possible before the current shape of `Datagrams`, everything is now cheaply clonable or copy.

iroh/src/net_report.rs
```diff
@@ -450,7 +450,7 @@ impl Client {
                             PROBES_TIMEOUT,
                             run_probe_v4(ip_mapped_addrs, relay_node, quic_client, dns_resolver),
                         ))
-                        .instrument(info_span!("QAD IPv6", %relay_url)),
+                        .instrument(info_span!("QAD IPv4", %relay_url)),
```
flub:
```suggestion
                        .instrument(info_span!("QAD-IPv4", %relay_url)),
```
If we're nitpicking :grin: .  I think spaces in span names are weird.  Though I can never figure out if `-` or `_` or what makes most sense.  Prob best to keep consistent with the actual IPv6 span though.

matheus23: Will fix both spans

PR #3404, opened 2025-07-24
fix(iroh-relay): Fix proptests, make `Datagrams::segment_size` be an `Option<NonZeroU16>`

matheus23:
## Description

The proptests were failing in some rare case where `Datagrams::segment_size` was generated as `Some(0)`. (For some reason we haven't hit that before?).

I added the failing case hash to the proptest regressions file and fixed the problem.

I also changed `Datagrams::segment_size` to be an `Option<NonZeroU16>` instead of `Option<u16>`, which checks this invariant on the type level now.

## Breaking Changes

- `Datagrams::segment_size` is now an `Option<NonZeroU16>` instead of `Option<u16>`.

## Change checklist

- [x] Self-review.


matheus23:
*sigh*. Can't merge because `endpoint_relay_connect_loop` is flakier then ever. I need #3402 to go in first.

PR #3405, opened 2025-07-25
chore(iroh): Update `n0-watcher`

matheus23:
## Description

Updates the `n0-watcher` dependency to the new version (that has a lot fewer `Result`s everywhere).
- [x] Depends on https://github.com/n0-computer/net-tools/pull/31
- [x] Needs a release of `net-tools`

## Breaking Changes

- Updated `n0-watcher` from version `0.2` to `0.3`.
  Migration guide for users:
  - `endpoint.node_addr().initialized().await?` -> `endpoint.node_addr().initialized().await` (no `?` needed anymore) and similarly for `endpoint.home_relay()` and other uses of `Watcher`s.
  - `endpoint.node_addr().get()?` -> `endpoint.node_addr().get()` and similarly for `endpoint.home_relay()` and other uses of `Watcher`s.
  - If all you have is a `&impl Watcher` but you need the current value, then you can't call `Watcher::get` anymore, as that now takes a `&mut self` instead of `&self`. You can work around this by `.clone()`ing to an intermediate watcher:
    `watcher_ref.get()` -> `watcher_ref.clone().get()`

## Change checklist

- [x] Self-review.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2298,10 +2298,10 @@ dependencies = [
  "iroh-relay",
  "n0-future",
  "n0-snafu",
- "n0-watcher",
+ "n0-watcher 0.3.0",
  "nested_enum_utils",
  "netdev",
- "netwatch",
+ "netwatch 0.8.0",
  "parse-size",
  "pin-project",
  "pkarr",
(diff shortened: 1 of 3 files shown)


Frando (approved):
Diff looks good, nice to see this simplification. Migration guide also read very good and didn't spot anything missing.

I had a quick thought if we could have the Watcher be a future so that you could skip the `.initialized` but don't think this is possible due to same-crate rules for trait impls on generics. Well, people will get used to the `.initialized`.

Issue #3040, opened 2024-12-12
Expiry of discovered peers when they are no longer alive


andrewdavidmackenzie:
When a peer has been discovered and appears in the remotes of the "discoveree", it stays there for a long time (I can't tell if "forever", or just a very long time), well after the peer has closed/crashed/ended.

The suggested method of using last_used to filter out peers discovered some time ago doesn't work:

Docs
https://docs.rs/iroh-net/latest/iroh_net/endpoint/struct.RemoteInfo.html

It seems `last_used` field (not called `last_seen`) is only updated when data transfer between peers takes place, not when that peer is discovered, or "re-discovered". It returns `None` in discovered peers in my tests, and that seems to match the docs description also.

So, I think we'd either need some other method of expiring peers (mDNS has a TTL setting when registering a service), or this field would need to be redefined, and updated each time the peer is discovered/confirmed-alive. But I'd argue that of the TTL expires and we try to confirm-alive and the peer is not alive, then it should be removed from the remotes - and updating the `last_used` would not be useful.

ramfox:
Except for mDNS, all of our other discovery methods are "active", you have to request the dialing information each time to know the latest status of the nodeID. If discovery is enabled, we run discovery each time the node is dialed, to get the latest information.

We have added expiry for nodes discovered vis mDNS, which is exposed in the `Discovery::subscribe` method.

PR #3403, opened 2025-07-23
feat(iroh)!: add passive mode for mdns discovery (#3401)

Le-Maz:
## Description

This pull request adds support for passive mDNS discovery, enabling nodes to listen for peer advertisements without broadcasting their own presence on the local network.

### Summary of changes:
- Added `advertise(bool)` setter to `MdnsDiscoveryBuilder` (default `true`).
- Modified `MdnsDiscovery::new` and `spawn_discoverer` to respect the `advertise` flag.

## Breaking Changes

- `MdnsDiscovery::new(node_id: NodeId)` → `MdnsDiscovery::new(node_id: NodeId, advertise: bool)`  
  Code using `MdnsDiscovery::new` must now provide the `advertise` argument explicitly.

## Change Checklist

- [x] Self-review.
- [x] Documentation updates following the [style guide](https://rust-lang.github.io/rfcs/1574-more-api-documentation-conventions.html#appendix-a-full-conventions-text).
- [x] Tests if relevant.
- [x] All breaking changes documented.

PR #3409, opened 2025-07-28
feat(iroh)!: Emit mDNS expiry events

oscartbeaumont:
## Description

I'm working on a project where we want to have a list of peers that can be connected to without actually establishing a connection to each of them. The user can then later select a peer and we will establish a connection to the ones that are required.

For this to work we need to be able to tell our frontend about the available peers on the network. Right now Iroh emits events when peers are discovered but it doesn't provide a mechanism to detect when those peers are no longer online so they remain in the frontend UI until the application is restarted.

This PR implements a system for Iroh's discovery system to emit events when a peer is no-longer aviable. This is implemented into the core discovery system, however the only discovery mechanism that currently emits these are mDNS.

This PR solves #3040.

## Breaking Changes

All methods that previously retruned `DiscoveryItem` now return `DiscoveryEvent`. This includes the `Discovery` trait, `Endpoint::discovery_stream`, etc.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/discovery/static_provider.rs
+++ b/iroh/src/discovery/static_provider.rs
@@ -67,7 +67,7 @@ use super::{Discovery, DiscoveryError, DiscoveryItem, NodeData, NodeInfo};
 #[derive(Debug, Default, Clone)]
 #[repr(transparent)]
 pub struct StaticProvider {
-    nodes: Arc<RwLock<BTreeMap<NodeId, Option<StoredNodeInfo>>>>,
+    nodes: Arc<RwLock<BTreeMap<NodeId, StoredNodeInfo>>>,
 }
 
 #[derive(Debug)]
@@ -132,8 +132,8 @@ impl StaticProvider {
         let last_updated = SystemTime::now();
         let NodeInfo { node_id, data } = node_info.into();
         let mut guard = self.nodes.write().expect("poisoned");
-        let previous = guard.insert(node_id, Some(StoredNodeInfo { data, last_updated }));
-        previous.and_then(|x| x.map(|x| x.data))
+        let previous = guard.insert(node_id, StoredNodeInfo { data, last_updated });
+        previous.map(|x| x.data)
     }
 
     /// Augments node addressing information for the given node ID.
@@ -147,19 +147,16 @@ impl StaticProvider {
         let mut guard = self.nodes.write().expect("poisoned");
         match guard.entry(node_id) {
             Entry::Occupied(mut entry) => {
-                if let Some(existing) = entry.get_mut() {
-                    existing
-                        .data
-                        .add_direct_addresses(data.direct_addresses().iter().copied());
-                    existing.data.set_relay_url(data.relay_url().cloned());
-                    existing.data.set_user_data(data.user_data().cloned());
-                    existing.last_updated = last_updated;
-                } else {
-                    entry.insert(Some(StoredNodeInfo { data, last_updated }));
-                }
+                let existing = entry.get_mut();
+                existing
+                    .data
+                    .add_direct_addresses(data.direct_addresses().iter().copied());
+                existing.data.set_relay_url(data.relay_url().cloned());
+                existing.data.set_user_data(data.user_data().cloned());
+                existing.last_updated = last_updated;
             }
             Entry::Vacant(entry) => {
-                entry.insert(Some(StoredNodeInfo { data, last_updated }));
+                entry.insert(StoredNodeInfo { data, last_updated });
             }
         }
     }
@@ -167,7 +164,7 @@ impl StaticProvider {
     /// Returns node addressing information for the given node ID.
     pub fn get_node_info(&self, node_id: NodeId) -> Option<NodeInfo> {
         let guard = self.nodes.read().expect("poisoned");
-        let info = guard.get(&node_id)?.as_ref()?;
+        let info = guard.get(&node_id)?;
         Some(NodeInfo::from_parts(node_id, info.data.clone()))
     }
 
@@ -177,7 +174,7 @@ impl StaticProvider {
     pub fn remove_node_info(&self, node_id: NodeId) -> Option<NodeInfo> {
         let mut guard = self.nodes.write().expect("poisoned");
         let info = guard.remove(&node_id)?;
-        Some(NodeInfo::from_parts(node_id, info?.data))
+        Some(NodeInfo::from_parts(node_id, info.data))
     }
 }
 
@@ -191,7 +188,7 @@ impl Discovery for StaticProvider {
         let guard = self.nodes.read().expect("poisoned");
         let info = guard.get(&node_id);
         match info {
-            Some(Some(node_info)) => {
+            Some(node_info) => {
                 let last_updated = node_info
                     .last_updated
                     .duration_since(SystemTime::UNIX_EPOCH)
@@ -204,7 +201,6 @@ impl Discovery for StaticProvider {
                 );
                 Some(stream::iter(Some(Ok(item))).boxed())
             }
-            Some(None) => Some(stream::iter(None).boxed()),
             None => None,
         }
     }


flub:
Thanks for making this PR!  The code looks really solid.

Mind adding a description of how mDNS expriry works?  E.g. is it configurable, how is it triggered etc?  I only know the basics of mDNS and some pointers towards those details would help me understand how this fits into the iroh discovery mechanism.

dignifiedquire:
any specific reason this is still a draft?

oscartbeaumont:
I've still got to finish implementing a proper unit test. For some reason `discovery::mdns::tests::run_in_isolation::mdns_expiry` the test I wrote isn't passing. I am fairly sure the problem is in the test itself as this PR is working fine as a Cargo patch in one of my projects.

@flub will write up more tomorrow morning!

oscartbeaumont:
> Mind adding a description of how mDNS expriry works? E.g. is it configurable, how is it triggered etc? I only know the basics of mDNS and some pointers towards those details would help me understand how this fits into the iroh discovery mechanism.

It's all handled by the [`swarm-discovery`](https://docs.rs/swarm-discovery/latest/swarm_discovery/) so it's follows the logic defined by them. The expires events I have added are emitted when [`Peer::is_expiry`](https://docs.rs/swarm-discovery/latest/swarm_discovery/struct.Peer.html#method.is_expiry) returns true. The expiry logic is defined in `swarm-discovery` [here](https://github.com/rkuhn/swarm-discovery/blob/bfac489ba42524853e55fa57e9cc1c92d4cdcf4d/src/updater.rs#L59) but it's a bit convoluted due to it taking into account stuff like the amount of peers on the network to reduce traffic for large networks.

iroh/src/endpoint.rs
```diff
@@ -1117,7 +1117,7 @@ impl Endpoint {
     ///
     /// [`MdnsDiscovery`]: crate::discovery::mdns::MdnsDiscovery
     /// [`StaticProvider`]: crate::discovery::static_provider::StaticProvider
-    pub fn discovery_stream(&self) -> impl Stream<Item = Result<DiscoveryItem, Lagged>> + use<> {
+    pub fn discovery_stream(&self) -> impl Stream<Item = Result<DiscoveryEvent, Lagged>> + use<> {
```
matheus23, line 1133: I think the doc comment above needs to be adjusted to say `DiscoveryEvent` instead of `DiscoveryItem`.

flub:
> > Mind adding a description of how mDNS expriry works? E.g. is it configurable, how is it triggered etc? I only know the basics of mDNS and some pointers towards those details would help me understand how this fits into the iroh discovery mechanism.
> 
> It's all handled by the [`swarm-discovery`](https://docs.rs/swarm-discovery/latest/swarm_discovery/) so it's follows the logic defined by them. The expires events I have added are emitted when [`Peer::is_expiry`](https://docs.rs/swarm-discovery/latest/swarm_discovery/struct.Peer.html#method.is_expiry) returns true. The expiry logic is defined in `swarm-discovery` [here](https://github.com/rkuhn/swarm-discovery/blob/bfac489ba42524853e55fa57e9cc1c92d4cdcf4d/src/updater.rs#L59) but it's a bit convoluted due to it taking into account stuff like the amount of peers on the network to reduce traffic for large networks.

To build a mental picture of how these expiry events fit into iroh discovery I would like to have some mental picture of how the expiry events are emitted by swarm-discovery here.  Superficially it seems like it is just a timeout since they last announced themselves?  But I'm not sure if there's any more to it.  Searching the crate docs you linked for "expiry" doesn't give me anything clear.

I'd like to understand how this works because this is a breaking change to the discovery trait, and we need to be sure this is generic enough and will be future proof.

For example, if my "this is a simple timeout" reasoning above is correct, does it really make sense to let every discovery mechanism decide it's own timeout?  Maybe!  Does it make more sense to instead have an (maybe optional) "announced at" time?  Maybe!  Is this all a special case of the swarm-discovery and does it make more sense to add the expiry stuff to that itself without involving the discovery trait?  Maybe!  This is why I'd like to get a better picture of the entire system we're involving here.

oscartbeaumont:
If we think about this theoretically I think an expiry event should be emitted under two conditions:
 - A node doesn't annonces itself for an agreed on timeout (Could happen due to it being force shutdown, disconnected from the network, application crash, etc)
 - A node annonces it's going offline (The user gracefully shuts the application on a remote machine)

The first condition is just a timer that aligns with the TTL of the DNS record emitted by the mDNS system. DNS records are normally published with a TTL which is the amount of time they are valid to be cached for. This is a reasonable value to use for the purpose of knowing if the node is still available as any resonable implementation would update the network before the TTL elapses.

An argument could be made that condition 2 of a node going offline is the equilevent of it's record expiring (condition 1). The issue with this is if the node information is user-facing (as it is in my usecase) this is a horrible user expirence. When someone quits the application on a remote computer, as long as it gracefully shutdown, you would expect it to instantly disappear from the UI on all other machines on the network.

If we jump back into Iroh. The design i've implemented here with the discovery system emitting `DiscoveryEvent`'s is designed to support both of these usecases.

If we relied on userspace timers, I don't think we would be able to have the instant feedback when a node went offline as the timer isn't deeply tied into the information the discovery system has.

I personally think the individual discovery systems should implement the expiry themselves which means they can use whatever logic they want internally to determine if the node is still available. Although you could argue this is less flexible from a user perspective, I think there is generally only a single correct value for if a node is available or not.

For example in the case of mDNS DNS record, when the TTL expires the client should be treated as expired. If the default TTL was a problem I think this should be adjusted as the point of emitting the DNS record, not on a node which is receiving it.

I have only been using the mDNS discovery system so i'm not certain how this fits into the other systems but I assume all systems would have some form of TTL or expiry signal as you always need way of knowing if the discovery information is still going to be valid.

So tldr: this isn't just a simple timeout, it also takes into account a node going offline near-instantly which is crucial to the UX of the application i'm building.

Let me know if that makes a bit more sense, and anything if you would like me to clarify furthur!

flub:
@oscartbeaumont Thanks for the overview!  This helped me a lot.  Do you know if mDNS (and the swarm-discovery crate) do announce when they're going offline?

flub:
(Also, apologies for the delay!  I've been on holiday and am catching up on things now)

iroh/src/discovery.rs
```diff
@@ -352,13 +352,23 @@ pub trait Discovery: std::fmt::Debug + Send + Sync + 'static {
...
 impl<T: Discovery> Discovery for Arc<T> {}
 
+/// An event emitted from [`Discovery`] services.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum DiscoveryEvent {
+    /// A peer was discovered or it's information was updated.
+    Discovered(DiscoveryItem),
+    /// A peer was expired due to being inactive, unreachable, or otherwise
+    /// unavailable.
+    Expired(NodeId),
```
flub, line 369: From what I understand in your description this is strictly speaking not just expired.  This naming is probably one of the things that threw me off in understanding things.

Does this need to distinguish "expired" from "offline"?  Or should this have one item covering both?

iroh/src/discovery.rs
```diff
@@ -326,7 +326,7 @@ pub trait Discovery: std::fmt::Debug + Send + Sync + 'static {
     fn resolve(
         &self,
         _node_id: NodeId,
-    ) -> Option<BoxStream<Result<DiscoveryItem, DiscoveryError>>> {
+    ) -> Option<BoxStream<Result<DiscoveryEvent, DiscoveryError>>> {
```
flub: I think resolve is specifically to find a `DiscoveryItem` or not, would this conceivably ever need to return `DiscoveryEvent::Expired` (or whatever colour that bikesheds ends up having)?

iroh/src/discovery/static_provider.rs
dignifiedquire, line 1: I wonder if we should emit `Expired` for nodes that `remove_node_info` is called on

iroh/src/discovery.rs
oscartbeaumont: I personally think "offline" is not the correct terminology here and that "expired" converys the meaning much better.

I feel like offline conveys that a device is no longer contactable, while expired conveys that the node's discovery information is no longer valid. The key difference here is that connections and discovery are separate and I think the "offline" terminology should stick to only being used for connections.

If we think theorical for a second I think it would be possible to have a case where a node's discovery information expires while it's connections stay open and in this case using "offline" here would be extremely confuising. For example maybe an application has a discovery mode which can be toggled by the user. We don't necessarily want to kill all established connections when it's turned off but we do want to expire the discovery information.

The "expired" terminology is inspired by some previous P2P work i've done but I think it was originally inspired from libp2p's [`libp2p::mdns::Event`](https://docs.rs/libp2p/latest/libp2p/mdns/enum.Event.html#variant.Expired).

I could clarify the enum variant's comment but i'm personally not sure how I could make it clearer, would love some thoughts?

iroh/src/discovery.rs
oscartbeaumont: My understanding was `resolve` is scoped to a single node where `subscribe` is for getting the events for every node.

If that is correct then personally I think  `DiscoveryEvent` here makes sense, if not we could concider changing it back but personally I think that feels a little inconsistent.

I suppose the method's name being `resolve` doesn't fit the best with the new return type.

I'm happy to do whatever you think here, but personally I think we leave it or concider renaming `resolve` to something else.

iroh/src/discovery/static_provider.rs
oscartbeaumont: Done!

oscartbeaumont:
> Do you know if mDNS (and the swarm-discovery crate) do announce when they're going offline?

I have observed this in testing but I actually can't figure out what is causing it within `swarm-discovery`. I am fairly sure the [swarm discovery GC](https://github.com/rkuhn/swarm-discovery/blob/bfac489ba42524853e55fa57e9cc1c92d4cdcf4d/src/updater.rs#L47) is probally the cause which isn't exactly a push-based shutdown event like I thought but it's still detecting the device going offline very quickly so from a UX/usecase perspective I think the exact mechanism is irrelevant.

iroh/src/discovery.rs
ramfox: Hi hi! `resolve` should still return a `DiscoveryItem`. `subscribe` makes sense to return a stream of events, but `resolve` should never need to return an expired event. If a `NodeId` can no longer be resolved, it should return `None`.

I don't fully understand what would be inconsistent. If you mean, it seems cleaner to have the same return type in both `resolve` and `subscribe` methods, I think that it is not only fine to have different return types, but could be better in some ways. It would more clearly show, just by the function signature, that the methods have different purposes.

But let me know if you were referring to something else!

flub: `resolve` is meant to be a request for a single node.  The reason it returns a stream is because a discovery service might be using multiple queries behind the scenes and different results may come in at different times.  iroh wants to start using the first result as soon as possible, but if later results are better still have them available.  But this stream is not used for a "long" time.

Hence it makes more sense to leave this as a `DiscoveryItem`.

iroh/src/discovery.rs
flub: Ok, with the help of @ramfox I think I've finally wrapped my head around how things should work in this world.  IIUC essentially mDNS issues DNS records with relatively short TTLs.  When a TTL expires without being renewed people treat it as "no longer available" in some terms of availability.  So expired does make sense.

Having dug through our discovery services only mDNS has this notion really, it is the only one that implements `Discovery::subscribe`.  All other discovery services are much more request-response based.  To some degree `subscribe` being in the `Discovery` trait is a bit odd, we might fix this still.

iroh/src/discovery/static_provider.rs
flub: With current understanding this was a bad call.  The static provider does not, and should not, implement `Discovery::subscribe`.  Please revert this.

flub:
Hey, a friendly ping that we *would* appreciate adding this feature.  It does address a need for users and would be a great addition.  It took me a while to understand the design and implications of these pieces.  But a `DiscoveryEvent` on `subscribe` which includes an `Expired` event *does* make sense to me now.  So would be cool to see this adjusted to that.

If you're on holidays or are otherwise just busy and planning to get back to this later: apologies!  Don't mean to hurry you up.  Only worried that the discussions may have put you off a bit.

oscartbeaumont:
I will do my best to get to this in the next week or so. I have been meaning to get back to it but i've been incredibly busy.

Apologies for the delay!

flub:
@oscartbeaumont All good, take your time!  I was just worried I might have put you off with the back-and-forth a bit while trying to understand this.

iroh/src/discovery.rs
oscartbeaumont: Okay I totally understand now. I was under the impression this was a long-lived stream but scoped to a specific node when that isn't correct.

iroh/src/discovery/static_provider.rs
oscartbeaumont: Although it's possibly out of scope for this issue I would be interested what the justification is for this? With my understanding, the `Discovery` trait is a shared interface for a discovery system so I don't see why subscribing to changes on the static provider would not be desired when it can be done with other provides like mdns?

Although it's called the static provider it isn't really that static because you can call `add_node_info` and `remove_node_info` with an immutable reference, so as far as the usage of the `Discovery` trait is concerned it's peers can change whenever during the whole lifetime of the application similar to mDNS provider so allowing a consumer to subscribe to it would open up the ability to write generic abstractions around it.

flub: Our current thinking is that the `Discovery` trait is too confusing because it is trying to cover both of these use cases (hey, it took us a bunch of digging to understand how we currently use it in the various discovery methods in order to understand this PR!).  Only the mDNS discovery implements the `subscribe` method.  This is likely a historical artefact, because it was the easiest to add a the feature.  Instead we'd like to remove `Discovery::subscribe` and instead implement it directly on the `MdnsDiscovery`.  And once other discovery methods have a "live membership view" rather than just the request-response interface we might also encode it in a trait.

To answer the `StaticProvider` question directly, as in why it doesn't implement `Discovery::subscribe`: we have no expectation that the `StaticProvider` knows the current membership.  Say your app received a ticket-like thing via some side-channel it can add it to the `StaticProvider` but that does not mean it knows when that is back out of date - if it even ever worked.  Hence we don't consider the static provider to have this life notion of active nodes.

Thanks for asking all these questions!  It makes it clear we need to work a bit more on all the doc comments to make this clearer.  We have a lot of assumed knowledge and seeing how someone else interprets things based on the docs is really valuable.  And clearly we have some work to do.

iroh/src/discovery/mdns.rs
```diff
@@ -70,9 +70,8 @@ const DISCOVERY_DURATION: Duration = Duration::from_secs(10);
 /// Discovery using `swarm-discovery`, a variation on mdns
 #[derive(Debug)]
 pub struct MdnsDiscovery {
-    #[allow(dead_code)]
-    handle: AbortOnDropHandle<()>,
     sender: mpsc::Sender<Message>,
+    shutdown: mpsc::Sender<()>,
```
flub: Any reason you're changing the shutdown method from the `AbortOnDropHandle` to this shutdown channel?  We really like `AbortOnDropHandle` mostly because we really do not like tasks which are accidentally left running.  So in general we try to use tasks that abort when their handle is dropped in new code rather than spawning tasks without such a guard.  Just as a way to avoid bugs (yes, bugs with tasks not getting their shutdown signal have happened).

I'm guessing you did this to get a shutdown message in the logs?  If so a number of things:
- Probably should use a `tokio::sync::oneshot::channel`.
- The `AbortOnDropHandle` should stay, as a belt-and-braces safeguard.
- The actor shutdown should be logged on at most on info level, preferably even on debug.

But I'm not sure if it's worth the hassle, but I guess that depends on your motivation for adding it.

iroh/src/discovery/static_provider.rs
```diff
@@ -67,7 +67,7 @@ use super::{Discovery, DiscoveryError, DiscoveryItem, NodeData, NodeInfo};
 #[derive(Debug, Default, Clone)]
 #[repr(transparent)]
 pub struct StaticProvider {
-    nodes: Arc<RwLock<BTreeMap<NodeId, StoredNodeInfo>>>,
+    nodes: Arc<RwLock<BTreeMap<NodeId, Option<StoredNodeInfo>>>>,
```
flub: I don't think these changes are still needed are they?

ramfox:
Hey @oscartbeaumont, we are really excited to get these changes into our next release happening early next week. Please let me know if there is anything I can do to help get this merged. For example, if you are okay with me pushing to your branch I can fix the failing checks.

Thanks!

ramfox:
The wasm and cargo deny tests should pass after a rebase, but just a note: the last PR we merged (https://github.com/n0-computer/iroh/pull/3403) also contained mdns changes, so there will be merge conflicts.

Happy to resolve those and push if it's any help!

iroh/src/discovery/mdns.rs
oscartbeaumont: > Any reason you're changing the shutdown method from the AbortOnDropHandle to this shutdown channel?

I think this might be an artifact of something I removed earlier in this PR because I recall it being required but can't seem to see where that would be anymore.

I totally agree `AbortOnDrop` is the way to go when it's possible (Eg. you don't have to run custom potentially blocking code on exit) so going to move back!

iroh/src/discovery/static_provider.rs
oscartbeaumont: That makes sense! Thanks for the explanation.

oscartbeaumont:
Okay, this should be good to rerun CI and get a final review! I merged in `main` and fixed the final comments!

flub:
Thanks for seeing this through!

PR #3412, opened 2025-07-29
feat(iroh): update to new relay servers

dignifiedquire:
## Description

These are the ones with the new `0.91` relay code


Arqu (approved):
Correct

PR #3413, opened 2025-07-29
fix: wait for at least one ipv6 and ipv4 qad report

Frando:
## Description

If an endpoint only has a single relay URL configured in its relay map, we currently abort the QAD net reports after a single successful report. However, we also decide whether an endpoint supports IPv6 on the fact if a QAD IPv6 report completed successfully or not. Those are *two* reports though, and if we abort after the first one, we falsely assume that the endpoint has no IPv6 connectivity.

This PR changes this in a rather simple way: If we did start both IPv4 and IPv6 reports, we always wait at least until one of each completed.

Unrelated to this change, the PR also adds logging when the best address changes.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/net_report.rs
+++ b/iroh/src/net_report.rs
@@ -475,19 +475,17 @@ impl Client {
 
         let mut reports = Vec::new();
 
-        // We set _did_start to true if at least one report was started for each category..
-        let ipv4_started = !v4_buf.is_empty();
-        let ipv6_started = !v6_buf.is_empty();
-        // If we did not start any report for either category, _did_finish is set to false right away
+        // We set _pending to true if at least one report was started for each category.
+        // If we did not start any report for either category, _pending is set to false right away
         // (it "completed" in the sense that nothing will ever run). If we did start at least one report,
-        // _did_finish is set to false, and will be set to true further down once the first task
+        // _pending is set to true, and will be set to false further down once the first task
         // completed.
-        let mut ipv4_did_finish = !ipv4_started;
-        let mut ipv6_did_finish = !ipv6_started;
+        let mut ipv4_pending = !v4_buf.is_empty();
+        let mut ipv6_pending = !v6_buf.is_empty();
         loop {
             // We early-abort the tasks once we have at least `enough_relays` reports,
             // and at least one ipv4 and one ipv6 report completed (if they were started, see comment above).
-            if reports.len() >= enough_relays && ipv4_did_finish && ipv6_did_finish {
+            if reports.len() >= enough_relays && !ipv4_pending && !ipv6_pending {
                 debug!("enough probes: {}", reports.len());
                 cancel_v4.cancel();
                 cancel_v6.cancel();
@@ -498,7 +496,7 @@ impl Client {
                 biased;
 
                 val = v4_buf.join_next(), if !v4_buf.is_empty() => {
-                    ipv4_did_finish = true;
+                    ipv4_pending = false;
                     match val {
                         Some(Ok(Some(Ok(res)))) => {
                             match res {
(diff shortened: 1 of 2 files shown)


iroh/src/net_report.rs
```diff
@@ -475,18 +475,25 @@ impl Client {
...
+        let ipv6_did_start = !v6_buf.is_empty();
+        let mut ipv4_did_complete = !ipv4_did_start;
+        let mut ipv6_did_complete = !ipv6_did_start;
         loop {
             if reports.len() >= enough_relays {
-                debug!("enough probes: {}", reports.len());
-                cancel_v4.cancel();
-                cancel_v6.cancel();
-                break;
+                if ipv4_did_complete && ipv6_did_complete {
```
dignifiedquire: the ipv6 check should only be done if we have ipv6 support on the machine

dignifiedquire: ah nvm, I see the logic now

iroh/src/net_report.rs
```diff
@@ -475,18 +475,25 @@ impl Client {
 
         let mut reports = Vec::new();
 
+        let ipv4_did_start = !v4_buf.is_empty();
+        let ipv6_did_start = !v6_buf.is_empty();
+        let mut ipv4_did_complete = !ipv4_did_start;
```
dignifiedquire: maybe add a comment, on what happens here, this is a lot of negations

Frando: I now pushed comments.

matheus23:
> This PR changes this in a rather simple way: If we did start both IPv4 and IPv6 reports, we always wait at least until one of each completed.

We also have `mapping_varies_by_dest_ipv6` (same for IPv4).  Maybe we should wait for at least 2 probes of each *if* we have more than one relay.

Frando:
> We also have mapping_varies_by_dest_ipv6 (same for IPv4). Maybe we should wait for at least 2 probes of each if we have more than one relay.

This will increase the time the report takes, because now it might take a run across the ocean. 

I don't know enough how relevant that is, so not sure what the right call is here - will defer to others or need to dig in further first.

dignifiedquire:
> We also have mapping_varies_by_dest_ipv6 (same for IPv4). Maybe we should wait for at least 2 probes of each if we have more than one relay.

that’s not needed, we only need at least two ipv4 or two ipv6 and only if we have a least two relays

iroh/src/net_report.rs
```diff
@@ -475,8 +475,19 @@ impl Client {
 
         let mut reports = Vec::new();
 
+        // We set _did_start to true if at least on report was started for each category..
```
matheus23:
```suggestion
        // We set _did_start to true if at least one report was started for each category..
```

iroh/src/net_report.rs
```diff
@@ -475,8 +475,19 @@ impl Client {
...
+        let ipv4_started = !v4_buf.is_empty();
+        let ipv6_started = !v6_buf.is_empty();
+        // If we did not start any report for either category, _did_finish is set to false right away
+        // (it "completed" in the sense that nothing will ever run). If we did start at least one report,
```
matheus23: Nit: We could invert the boolean logic and call this `ipv4_pending` and `ipv6_pending` instead.

matheus23 (approved):

iroh/src/net_report.rs
Frando: oh yeah, that's simpler to follow indeed. Feel free to push or I can do tonight (or as a followup if this gets merged before)

flub:
> > We also have mapping_varies_by_dest_ipv6 (same for IPv4). Maybe we should wait for at least 2 probes of each if we have more than one relay.
> 
> that’s not needed, we only need at least two ipv4 or two ipv6 and only if we have a least two relays

So much confidence in that statement... :grin:

PR #3414, opened 2025-07-29
fix(iroh): Re-batch received relay datagram batches in case they exceed `max_receive_segments`

matheus23:
## Description

Fixes a panic in `RelayTransport::poll_recv` with the new relay code.

Since #3389 we forward the whole GSO batch via the relay. This then gets read on the other receiving endpoint, which tries to fill its `buf_out`, which might not be big enough, because its size is initialized with `max_receive_segments` taken into account, which might be different from the sending endpoint's `max_transmit_segments`.

So in case we get a batch that's too big, we use a `DatagramReBatcher` to split up our datagram batch into chunk sizes we can consume.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-relay/src/protos/relay.rs
+++ b/iroh-relay/src/protos/relay.rs
@@ -164,8 +164,8 @@ impl Datagrams {
     /// will result in returning essentially `Some(self.clone())`, while making `self` empty afterwards.
     ///
     /// Calling this on a datagram batch with e.g. 15 datagrams with `num_segments == 10` will
-    /// result in returning `Some(datagram_batch)` where that `datagram_batch` contains 10 datagrams and
-    /// `self` contains 5 datagrams.
+    /// result in returning `Some(datagram_batch)` where that `datagram_batch` contains the first
+    /// 10 datagrams and `self` contains the remaining 5 datagrams.
     ///
     /// Calling this on a datagram batch that doesn't contain `num_segments` datagrams, but less
     /// will result in making `self` empty and returning essentially a clone of `self`.


iroh/examples/transfer.rs
```diff
@@ -153,7 +153,9 @@ enum Commands {
 
 #[tokio::main]
 async fn main() -> Result<()> {
-    tracing_subscriber::fmt::init();
+    tracing_subscriber::fmt()
+        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+        .init();
```
matheus23, line 158: Eh - stray change, but IMO this is useful.

iroh-relay/src/protos/relay.rs
```diff
@@ -157,6 +157,47 @@ impl<T: AsRef<[u8]>> From<T> for Datagrams {
...
+
+        let usize_segment_size = usize::from(u16::from(segment_size));
+        let max_content_len = num_segments * usize_segment_size;
+        let contents = self
+            .contents
+            .split_to(std::cmp::min(max_content_len, self.contents.len()));
+
+        Some(Datagrams {
+            ecn: self.ecn,
+            segment_size: (num_segments > 1 && usize_segment_size < contents.len())
```
dignifiedquire: this is very hard to read

iroh-relay/src/protos/relay.rs
```diff
@@ -157,6 +157,47 @@ impl<T: AsRef<[u8]>> From<T> for Datagrams {
 }
 
 impl Datagrams {
+    /// Splits the current datagram into at maximum `num_segments` segments.
+    ///
+    /// Calling this on a datagram batch that only contains a single datagram (`segment_size == None`)
+    /// will result in returning essentially `Some(self.clone())`, while making `self` empty afterwards.
+    ///
+    /// Calling this on a datagram batch with e.g. 15 datagrams with `num_segments == 10` will
```
dignifiedquire, line 166: does it return the first or the last 10 datagrams?

iroh/src/magicsock/transports/relay.rs
```diff
@@ -100,6 +100,19 @@ impl RelayTransport {
                 }
             };
 
+            if buf_out.len() < dm.datagrams.contents.len() {
+                // Our receive buffer isn't big enough to process this datagram.
+                // Continuing would cause a panic.
+                break;
```
dignifiedquire, line 112: maybe we should log this? especially to make sure we appropriately split things now?

iroh-relay/src/protos/relay.rs
```diff
@@ -157,6 +157,47 @@ impl<T: AsRef<[u8]>> From<T> for Datagrams {
...
+    /// result in returning `Some(datagram_batch)` where that `datagram_batch` contains 10 datagrams and
+    /// `self` contains 5 datagrams.
+    ///
+    /// Calling this on a datagram batch that doesn't contain `num_segments` datagrams, but less
+    /// will result in making `self` empty and returning essentially a clone of `self`.
+    ///
+    /// Calling this on an empty datagram batch (i.e. one where `contents.is_empty()`) will return `None`.
+    pub fn take_segments(&mut self, num_segments: usize) -> Option<Datagrams> {
```
flub, line 174: Goodness I love `Bytes`.

iroh-relay/src/protos/relay.rs
matheus23: Would this be better?
```rs
let is_datagram_batch = num_segments > 1 && usize_segment_size < contents.len();

Some(Datagrams {
    ecn: self.ecn,
    segment_size: is_datagram_batch.then_some(segment_size),
    contents,
})
```

matheus23: Or would you prefer if I rewrote this using if-else?

iroh/src/magicsock/transports/relay.rs
matheus23: Right! we should.

iroh-relay/src/protos/relay.rs
dignifiedquire: the split out seems fine

iroh-relay/src/protos/relay.rs
matheus23: It returns the first 10, leaving the last 5 in `self`.
I adjusted the comment, is it better now?

dignifiedquire: yes

iroh/src/magicsock/transports/relay.rs
```diff
@@ -100,6 +100,25 @@ impl RelayTransport {
...
+            if buf_out.len() < dm.datagrams.contents.len() {
+                // Our receive buffer isn't big enough to process this datagram.
+                // Continuing would cause a panic.
+                warn!(
+                    quinn_buf_len = buf_out.len(),
+                    datagram_len = dm.datagrams.contents.len(),
+                    segment_size = ?dm.datagrams.segment_size,
+                    "dropping received datagram: quinn buffer too small"
+                );
+                break;
```
flub, line 112: One, in my opinion large, downside of your approach is that you have to plumb the signal of how large this an be all the way through to the ActiveRelayActor.  I find all that plumbing really unfortunate and complex.

Have you considered having an `Option<Datagrams>` on the `RelayTransport` itself?  I think that way you could store the split off ones here locally.  When you get called next you first consume that and then continue reading off the channel.  Would that not be a lot simpler?

matheus23: Didn't consider that.
But thinking about how I'd do this, I'm somewhat afraid of messing up the waking logic in that case.
I'd prefer not to have to mess with wakers :S

matheus23: FWIW I think this is very similar to the `ipv6_reported: Arc<AtomicBool>` we already use. I tried to move the `max_receive_segments` declarations close to that if possible.

flub: Do you really need to mess with wakers?  When you receive something that doesn't fit you will return a `Poll::Ready` and no waker is expected to be installed.  Next time it gets called it again returns `Poll::Ready` because there's something in the Option.  When the Option is empty you poll the channel and things behave as before.

matheus23: Yeah, I think you're right. I'll look into following up this PR with a little bit of a cleanup.

PR #3415, opened 2025-07-29
fix(iroh): Only clear `last_call_me_maybe` when the best addr became invalid

matheus23:
## Description

Fixes a bug introduced in #3400

We used to only reset `last_call_me_maybe` to `None` when the best address was cleared. The above PR accidentally changed that to always reset.

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh/src/magicsock/node_map/udp_paths.rs
+++ b/iroh/src/magicsock/node_map/udp_paths.rs
@@ -8,7 +8,7 @@
 use std::{collections::BTreeMap, net::SocketAddr};
 
 use n0_future::time::Instant;
-use tracing::debug;
+use tracing::{Level, event};
 
 use super::{IpPort, path_state::PathState};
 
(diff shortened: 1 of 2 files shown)


iroh/src/magicsock/node_map/udp_paths.rs
```diff
@@ -114,20 +114,34 @@ impl NodeUdpPaths {
...
         let best_ipv4 = self.best_addr(false, now);
         let best = self.best_addr(true, now);
+        let mut changed = false;
         if best_ipv4 != self.best_ipv4 {
-            debug!(?best_ipv4, "update best addr (IPv4)");
+            event!(
+                target: "iroh::_events::udp::best_ipv4",
+                Level::DEBUG,
+                ?best_ipv4,
+            );
```
matheus23, line 131: I thought it would be fitting to promote these to be events.

FWIW, we also have "new connection type" events that contain the send addr, but this additionally prints whether we think this is an outdated, valid or unconfirmed address.

PR #3419, opened 2025-07-30
fix(iroh): Use valid available IPv6 address, ignoring `have_ipv6`

matheus23:
## Description

Sometimes net_report says `udp_v6: false`, even though I can clearly send and receive to/from the other iroh endpoint I want to talk to.
Here are some logs from that situation:
```
2025-07-29T10:15:43.281088Z DEBUG ep{me=68b3f4f4cc}:actor: iroh::net_report: generated report in 728ms report=Report { udp_v4: true, udp_v6: false, mapping_varies_by_dest_ipv4: Some(false), mapping_varies_by_dest_ipv6: None, preferred_relay: Some(RelayUrl("[https://euc1-1.relay.n0.iroh.iroh.link./")](https://euc1-1.relay.n0.iroh.iroh.link./%22))), relay_latency: RelayLatencies { ipv4: {RelayUrl("[https://aps1-1.relay.n0.iroh.iroh.link./"](https://aps1-1.relay.n0.iroh.iroh.link./%22)): 182.188896ms, RelayUrl("[https://euc1-1.relay.n0.iroh.iroh.link./"](https://euc1-1.relay.n0.iroh.iroh.link./%22)): 21.317467ms, RelayUrl("[https://use1-1.relay.n0.iroh.iroh.link./"](https://use1-1.relay.n0.iroh.iroh.link./%22)): 117.29417ms}, ipv6: {}, https: {} }, global_v4: Some(176.199.xxx.xxx:26608), global_v6: None, captive_portal: None }
2025-07-29T10:15:55.933530Z DEBUG ep{me=68b3f4f4cc}:disco_in{node=e10b5cdea1 src=Ip([2a01:599:b2f:a1ad:495a:8b14:xxxx:xxxx]:49154)}:handle_disco{dm=Pong(Pong { tx_id: TransactionId(0xBC381DA98EFD2B20CAE0A96E), ping_observed_addr: Udp([2a02:8071:d85:19c0:9676:965:xxxx:xxxx]:38391) })}:handle_pong{m=Pong { tx_id: TransactionId(0xBC381DA98EFD2B20CAE0A96E), ping_observed_addr: Udp([2a02:8071:d85:19c0:9676:965:xxxx:xxxx]:38391) } src=Udp([2a01:599:b2f:a1ad:495a:8b14:xxxx:xxxx]:49154)}: iroh::_events::holepunched: remote_node=e10b5cdea1 path=[2a01:599:b2f:a1ad:495a:8b14:xxxx:xxxx]:49154 direction="outgoing"
```

Now, the `udp_v6: false` value originally comes from `netwatch`'s `InterfaceState`, as it sees my (what is now and probably was at the time fe80-prefixed) link-local address and characterizes it as "not reachable from the public internet".

But this PR instead ignores that value *iff* we have any `UdpSendAddr::Valid`, since that would mean it's an address that we've successfully sent and received from with the particular peer we care about.

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh/src/magicsock/node_map/node_state.rs
+++ b/iroh/src/magicsock/node_map/node_state.rs
@@ -294,22 +294,22 @@ impl NodeState {
         let (best_addr, relay_url) = match self.udp_paths.send_addr(have_ipv6) {
             UdpSendAddr::Valid(addr) => {
                 // If we have a valid address we use it.
-                trace!(%addr, "UdpSendAddr is valid, use it");
+                trace!(%addr, ?have_ipv6, "UdpSendAddr is valid, use it");
                 (Some(*addr), None)
             }
             UdpSendAddr::Outdated(addr) => {
                 // If the address is outdated we use it, but send via relay at the same time.
                 // We also send disco pings so that it will become valid again if it still
                 // works (i.e. we don't need to holepunch again).
-                trace!(%addr, "UdpSendAddr is outdated, use it together with relay");
+                trace!(%addr, ?have_ipv6, "UdpSendAddr is outdated, use it together with relay");
                 (Some(*addr), self.relay_url())
             }
             UdpSendAddr::Unconfirmed(addr) => {
-                trace!(%addr, "UdpSendAddr is unconfirmed, use it together with relay");
+                trace!(%addr, ?have_ipv6, "UdpSendAddr is unconfirmed, use it together with relay");
                 (Some(*addr), self.relay_url())
             }
             UdpSendAddr::None => {
-                trace!("No UdpSendAddr, use relay");
+                trace!(?have_ipv6, "No UdpSendAddr, use relay");
                 (None, self.relay_url())
             }
         };
@@ -334,7 +334,6 @@ impl NodeState {
                 Level::DEBUG,
                 remote_node = %self.node_id.fmt_short(),
                 conn_type = ?typ,
-                have_ipv6,
             );
             info!(%typ, "new connection type");
 


iroh/src/magicsock.rs
```diff
@@ -1905,11 +1910,11 @@ impl Actor {
     }
 
     async fn handle_network_change(&mut self, is_major: bool) {
-        debug!("link change detected: major? {}", is_major);
+        debug!("link change detected: major? {is_major}");
```
flub: Use a field please if updating this.

iroh/src/magicsock/node_map/node_state.rs
```diff
@@ -334,6 +334,7 @@ impl NodeState {
                 Level::DEBUG,
                 remote_node = %self.node_id.fmt_short(),
                 conn_type = ?typ,
+                have_ipv6,
```
flub: is this relevant here?

matheus23: I think so. Because if `have_ipv6` is false, this affects the choice of connection type.

I can see this being a little too noisy. Let me know if I should remove this.

matheus23:
Unfortunately further testing didn't have the intended result: I wasn't able to reproduce the case I had above where I was magically able to send/receive IPv6 traffic to someone else's home address, but *not* to the relays (which I agree is *very* unusual).

In any case, I still believe this is a valid change. If we end up for some reason having a *valid* IPv6 addr, we should probably use it.
If it's the case that it's only valid because we *had* IPv6 in the past, but not in the future, then it'll become invalid after 6.5 seconds and we'll use a valid/outdated/unconfirmed IPv4 address instead then.

iroh/src/magicsock/node_map/node_state.rs
matheus23: Removed it.

flub: You appear to have removed it from the trace logs but kept it in the event.  I questioned more whether it was useful in the event, trace logs are just trace logs.  I'm not sure it changes the meaning of the event whether that's there or not.  But in any case, logs and events are living things, we tweak them when we see fit.  So if you think this helps, then add it.  I was only checking you found this helpful on the event because I wasn't sure what extra info it gave.

flub (approved):

iroh/src/magicsock/node_map/node_state.rs
matheus23: :facepalm: Sorry I simply edited the wrong lines. Thanks for verifying!

EDIT: Should've made the right change now!

PR #3421, opened 2025-07-31
refactor(iroh): Re-batch datagrams inside `RelayTransport` instead of the `ActiveRelayActor`

matheus23:
## Description

This does the re-batching inside `RelayTransport` by storing a `pending_item: Option<RelayRecvDatagram>`.
When we `poll_recv`, we first try to use that pending item instead of polling a new one.

Once we've got a pending item, we try to split off as much from it as can possibly fit into our receive buffer and handle that.

## Breaking Changes

- `iroh_relay::protos::relay::Datagrams::take_segments` now return `Datagrams` instead of `Option<Datagrams>`, not distinguishing the case where `Datagrams` might be empty.

## Notes

Probably needs some test.
I'd love to test two `RelayTransport`s talking to each other with different `max_transmit_segments`/`max_receive_segments`, but I'm not sure I can make such a test setup happen easily.

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh-relay/src/protos/relay.rs
+++ b/iroh-relay/src/protos/relay.rs
@@ -161,14 +161,14 @@ impl Datagrams {
     /// the batch with at most `num_segments` and leaving only the rest in `self`.
     ///
     /// Calling this on a datagram batch that only contains a single datagram (`segment_size == None`)
-    /// will result in returning essentially `Some(self.clone())`, while making `self` empty afterwards.
+    /// will result in returning essentially a clone of `self`, while making `self` empty afterwards.
     ///
     /// Calling this on a datagram batch with e.g. 15 datagrams with `num_segments == 10` will
-    /// result in returning `Some(datagram_batch)` where that `datagram_batch` contains the first
-    /// 10 datagrams and `self` contains the remaining 5 datagrams.
+    /// result in returning a datagram batch taht contains the first 10 datagrams and leave `self`
+    /// containing the remaining 5 datagrams.
     ///
-    /// Calling this on a datagram batch that doesn't contain `num_segments` datagrams, but less
-    /// will result in making `self` empty and returning essentially a clone of `self`.
+    /// Calling this on a datagram batch with less than `num_segments` datagrams will result in
+    /// making `self` empty and returning essentially a clone of `self`.
     pub fn take_segments(&mut self, num_segments: usize) -> Datagrams {
         let Some(segment_size) = self.segment_size else {
             let contents = std::mem::take(&mut self.contents);
--- a/iroh-relay/src/protos/relay.rs
+++ b/iroh-relay/src/protos/relay.rs
@@ -157,7 +157,7 @@ impl<T: AsRef<[u8]>> From<T> for Datagrams {
 }
 
 impl Datagrams {
-    /// Splits the current datagram into at maximum `num_segments` segments, potentially returning
+    /// Splits the current datagram into at maximum `num_segments` segments, returning
     /// the batch with at most `num_segments` and leaving only the rest in `self`.
     ///
     /// Calling this on a datagram batch that only contains a single datagram (`segment_size == None`)
@@ -169,20 +169,14 @@ impl Datagrams {
     ///
     /// Calling this on a datagram batch that doesn't contain `num_segments` datagrams, but less
     /// will result in making `self` empty and returning essentially a clone of `self`.
-    ///
-    /// Calling this on an empty datagram batch (i.e. one where `contents.is_empty()`) will return `None`.
-    pub fn take_segments(&mut self, num_segments: usize) -> Option<Datagrams> {
-        if self.contents.is_empty() {
-            return None;
-        }
-
+    pub fn take_segments(&mut self, num_segments: usize) -> Datagrams {
         let Some(segment_size) = self.segment_size else {
             let contents = std::mem::take(&mut self.contents);
-            return Some(Datagrams {
+            return Datagrams {
                 ecn: self.ecn,
                 segment_size: None,
                 contents,
-            });
+            };
         };
 
         let usize_segment_size = usize::from(u16::from(segment_size));
@@ -193,11 +187,11 @@ impl Datagrams {
 
         let is_datagram_batch = num_segments > 1 && usize_segment_size < contents.len();
 
-        Some(Datagrams {
+        Datagrams {
             ecn: self.ecn,
             segment_size: is_datagram_batch.then_some(segment_size),
             contents,
-        })
+        }
     }
 
     fn write_to<O: BufMut>(&self, mut dst: O) -> O {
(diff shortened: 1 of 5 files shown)


iroh-relay/src/protos/relay.rs
```diff
@@ -157,32 +157,26 @@ impl<T: AsRef<[u8]>> From<T> for Datagrams {
 }
 
 impl Datagrams {
-    /// Splits the current datagram into at maximum `num_segments` segments, potentially returning
+    /// Splits the current datagram into at maximum `num_segments` segments, returning
     /// the batch with at most `num_segments` and leaving only the rest in `self`.
     ///
     /// Calling this on a datagram batch that only contains a single datagram (`segment_size == None`)
-    /// will result in returning essentially `Some(self.clone())`, while making `self` empty afterwards.
+    /// will result in returning essentially a clone of `self`, while making `self` empty afterwards.
```
flub, line 164: One problem I have with this behaviour is that a segment_size of something with a empty contents (or even with just one datagram) is supposed to be illegal according to the docs of the struct.

flub: Maybe this could be simplified by always having a `segement_size`?  It's only the serialisation that needs to be linux-compatible.  Likewise we can declare that an empty `Datagrams` is allows and provide an `is_empty` method.

iroh/src/magicsock/transports/relay.rs
```diff
@@ -100,6 +103,23 @@ impl RelayTransport {
...
+            let datagrams = dm.datagrams.take_segments(num_segments);
+            let empty_after = dm.datagrams.contents.is_empty();
+            let dm = RelayRecvDatagram {
+                datagrams,
+                src: dm.src,
+                url: dm.url.clone(),
+            };
+            // take_segments can leave `self.pending_item` empty, in that case we clear it
+            if empty_after {
+                self.pending_item = None;
```
flub, line 120: If `Datagrams` is allowed to be empty, you can also make `pending_item` not an Option and instead check in `poll_recv_queue` check if it's empty or if a new poll is needed.  I think that way there's less state coordination between the different places and thus easier to not get wrong.

flub (commented):
I'm not 100% sure about my comments.  But the `Datagram` semantics don't currently match it's docs.  And I think if my comments aren't completely wild it could end up with easier state in all places.

iroh/src/magicsock/transports/relay.rs
matheus23: Yeahh...
My wish would be that `Datagrams::take_segments` would take `self: &mut Option<Self>`, but that's not possible :/

Let me remove `Datagrams::take_segments` and instead move it into the `transports::relay` module and make it a top-level fn on `&mut Option<Datagrams>`

matheus23: Hrm my idea from the previous comment doesn't really work, because `take_segments` would be forced to return `Option<Datagrams>`.

And I also can't store a non-option `pending_item`, because it's more than just a `Datagrams`, it's actually a `RelayRecvDatagram`, which also stores a `NodeId` and `RelayUrl`...

iroh-relay/src/protos/relay.rs
matheus23: I think the only invariant that's invalidated here is that `self.contents` can be empty after this call.

matheus23: Actually that's not a claimed invariant. `Datagrams` doesn't have any restrictions on `contents`. We only check that on receive, but that's basically it.

Can you tell me which invariant from the `Datagams` documentation is actually invalidated?

flub: Line 141-143 in this file in this PR:
```rust
/// The segment size if this transmission contains multiple datagrams.
/// This is `None` if the transmit only contains a single datagram
pub segment_size: Option<NonZeroU16>,
```

We can change this to `segment_size: NonZeroU16` and add to the docs of the `contents` field that it might be empty, just should always be a multiple of `segment_size` if non-empty.  Then this becomes a little more consistent?

Apologies for my vague reviews, I should clearly give more context up front!

iroh/src/magicsock/transports/relay.rs
flub: Ok, sad, but fair enough.

iroh/src/magicsock/transports/relay.rs
```diff
@@ -156,6 +177,28 @@ impl RelayTransport {
...
+    ///
+    /// Returns a mutable reference to the stored pending item.
+    #[inline]
+    fn poll_recv_queue<'a>(
+        &'a mut self,
+        cx: &mut Context,
+    ) -> Poll<Option<&'a mut RelayRecvDatagram>> {
+        // Borrow checker doesn't quite understand an if let Some(_)... here
+        if self.pending_item.is_some() {
+            return Poll::Ready(self.pending_item.as_mut());
```
flub, line 191: Is a pending item never going to be too large?  Would you not need to again take the required number out of it and see if it still contains anything afterwards?

iroh-relay/src/protos/relay.rs
matheus23: In your initial review comment you write:

> One problem I have with this behaviour is that a segment_size of something with a empty contents (or even with just one datagram) is supposed to be illegal according to the docs of the struct.

I don't see where it says that?

This is a different invariant:

> The segment size if this transmission contains multiple datagrams.
> This is `None` if the transmit only contains a single datagram

The current code still respects that.

iroh/src/magicsock/transports/relay.rs
matheus23: It is going to be too large, yes.
The calling code will always call `.take_segments` on this item's contents (i.e. split out however much fits) and then leaves the rest in `self.pending_item`. It's finally set to `None` (allowing this function to actually fetch a new item), once `take_segments` left an empty item.

flub: oh right.

iroh-relay/src/protos/relay.rs
flub: I think after you call `Datagrams::take_segments` this guarantee is not upheld?  You can end up with a single item but `segment_size` being still `Some` or it can be empty, in which case `segment_size` can not be right.

Though I appreciate that the way you use it does at least solve the 2nd issue.

Anyway, I don't feel like my comments here are being helpful, or maybe I'm even misunderstanding things.  This conversation isn't really improving the quality of the code.  Apologies to hold things up.

flub (approved):
I might be getting something wrong, and this certainly isn't worth holding up the great improvement this is.  Apologies for dragging this out.

iroh-relay/src/protos/relay.rs
matheus23: Ugh, I'm sorry. You're right that I missed a case where the invariant is broken in `self`.
I kept double-checking the return value and what's stored in `self` after the early-return and missed what's stored in `self` after the late return.

Sorry this took so long :X

matheus23:
Sorry, you were completely right to "drag this out", at the end it was me dragging things out because I just kept getting confused about the invariant. I hope the last commit fixed this!

flub (approved):

PR #3422, opened 2025-07-31
fix(iroh): Always update the best addr after changes

matheus23:
## Description

There was a bug where in `NodeMap::add_node_addr` wouldn't cause an update to the `NodeUdpState::best` addr, because we wouldn't call `NodeUdpState::udpate_best_addr` after changes in `NodeState::update_from_node_addr`.

In an effort to fix this class of bugs once and for all, I've made `NodeUdpState::paths` private and added guarded access. The access guard has a `Drop` implementation that calls `update_best_addr`.

We originally saw this bug as an error message in iroh-blobs tests:
image

---

While making `NodeUdpState`'s fields private, I noticed that `NodeUdpState::best` was sometimes accessed instead of `NodeUdpState::send_addr`. This is now fixed, but required passing through the `have_ipv6` boolean quite a bit, so that's why the diff is slightly bigger.
Feel free to look at commits in isolation instead.

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh-relay/src/protos/handshake.rs
+++ b/iroh-relay/src/protos/handshake.rs
@@ -26,6 +26,7 @@
 //! [RFC 5705]: https://datatracker.ietf.org/doc/html/rfc5705
 use bytes::{BufMut, Bytes, BytesMut};
 use data_encoding::BASE32HEX_NOPAD as HEX;
+#[cfg(not(wasm_browser))]
 use http::HeaderValue;
 #[cfg(feature = "server")]
 use iroh_base::Signature;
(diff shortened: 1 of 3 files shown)


iroh/src/magicsock/node_map/udp_paths.rs
```diff
@@ -85,11 +85,34 @@ pub(super) struct NodeUdpPaths {
...
-    pub(super) best: UdpSendAddr,
+    best: UdpSendAddr,
     /// The current best address to send on from all IPv4 addresses we have available.
     ///
     /// Follows the same logic as `best` above, but doesn't include any IPv6 addresses.
-    pub(super) best_ipv4: UdpSendAddr,
+    best_ipv4: UdpSendAddr,
+}
+
+pub(super) struct MutAccess<'a> {
```
flub, line 95: This is cool.  Have you considered moving all the logic that needs to have access to the internals here to be methods on the `NodeUdpPaths` itself and protecting access and ensuring updating of best addr that way?  I think it's what I would have gone for, or at least tried.  I appreciate this is less changes.

I'm also aware that all these changes interact with the changes needed for multipath... and I haven't wrapped my head around what those are yet so don't know if more or less cleaning up here is helpful.

matheus23: Yeah I had considered that. Would've been substantially more changes I guess

PR #3424, opened 2025-08-01
fix(iroh): better waiting for the NodeAddr to be ready in the echo example

flub:
## Description

This is a major footgun right now, and even this basic example gets it wrong.

## Notes & open questions

There might be other examples affected, I didn't look.

My first version was very different, and absolutely horrible to write and still failed because of obscure borrowing errors.  This is probably rather difficult for people new to rust.  Maybe we need to provide a better helper function?

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh/examples/echo.rs
+++ b/iroh/examples/echo.rs
@@ -6,6 +6,8 @@
 //!
 //!     cargo run --example echo --features=examples
 
+use std::time::Duration;
+
 use iroh::{
     Endpoint, NodeAddr,
     endpoint::Connection,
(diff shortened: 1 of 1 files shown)


matheus23 (commented):
This complicates the example quite a bit.

If this is how we *should* wait for connectivity, should we update the way that `node_addr()` works?

> This is a major footgun right now, and even this basic example gets it wrong.

Can you explain what exactly the example gets wrong? What's the failure scenario that you're seeing?

flub:
> This complicates the example quite a bit.

Yes it does!

> If this is how we _should_ wait for connectivity, should we update the way that `node_addr()` works?

Maybe.
 
> > This is a major footgun right now, and even this basic example gets it wrong.
> 
> Can you explain what exactly the example gets wrong? What's the failure scenario that you're seeing?

Users keep running into the cases where they use `.node_addr().initialized()` and end up with an address that does not have a relay URL inside it.  And then run into connectivity problems.

iroh/examples/echo.rs
```diff
@@ -23,8 +25,24 @@ const ALPN: &[u8] = b"iroh-example/echo/0";
...
     let node_addr = router.endpoint().node_addr().initialized().await;
+    if node_addr.relay_url.is_none() {
+        println!("Did not connect to a home relay, connectivity may be degraded");
+    }
+    if node_addr.direct_addresses.is_empty() {
+        println!("Did not find local addresses, connectivity may be degraded");
+    }
 
+    // Connect from the client address.
+    let node_addr = router.endpoint().node_addr().initialized().await;
```
Frando, line 45: can be removed (equal to line 36)

dignifiedquire:
closing in favor of #3467

PR #3432, opened 2025-08-12
chore(iroh,iroh-relay): Bump iroh-metrics version to 0.35, bump portmapper to 0.6.1

matheus23:
## Description

Bumping these versions so that we can use iroh-metrics version 0.35 in iroh-n0des.

This ends up working end-to-end.

## Breaking Changes

- `EndpointMetrics` now doesn't implement iroh-metrics v0.34's `MetricsGroupSet`, but iroh-metrics v0.35 instead.

## Notes & open questions

The breaking change prevents this from being released as 0.35.1...

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1965,7 +1965,7 @@ dependencies = [
  "js-sys",
  "log",
  "wasm-bindgen",
- "windows-core 0.61.0",
+ "windows-core 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


matheus23:
The breaking change above probably makes it impossible to release this as version 0.35.1.

I think I'll only keep the branch for now, so closing.

PR #3434, opened 2025-08-14
test: add test to measure initial connection speed without relay.

rklaehn:
## Description

Adds a test for initial connection delay. It's not really a test yet since it doesn't test anything, but it is there to illustrate the issue.

Initial connection is *very* slow even with relays disabled and static discovery. There seems to be some fixed delay.

## Breaking Changes

None

## Notes & open questions

Question: not sure how we can turn this into a proper test without making it super flaky, since it is a timing test. Maybe measure the ratio between first and second round and have some very conservative bounds?

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


matheus23 (approved):
I'm not super happy with the code duplication, I think it's probably possible to factor out common code for constructing a `Transmit` from `send_ping_actions` and `try_send_ping_actions`, but it's *fine*, especially since this code won't live that long.

PR #3438, opened 2025-08-15
feat(iroh-base): impl Deref for PublicKey

### LINKED ISSUE #3437: NodeId / PublicKey should implement Deref<[u8; 32]>
Sometimes this is useful, and we already have .as_bytes(), so it's not like we hide the internals.

This might be a bad idea, but I currently don't see why...

The same argument could be made for SecretKey, but I really find it convenient mostly for PublicKey.

varun-doshi:
## Description
Fixes #3437 
Implement `Deref::[u8,32]` for `PublicKey` and ~`SecretKey`~

## Change checklist

- [x] Self-review.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -157,7 +157,7 @@ jobs:
         uses: actions/checkout@v5
 
       - name: Setup node.js version 22.5 # needed for browser-like websocket API support in node.js
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@v5
         with:
           node-version: 22.5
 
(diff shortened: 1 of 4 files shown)


dignifiedquire:
We shouldn't do this for `SecretKey`, it is too risky given it is secret material

varun-doshi:
> We shouldn't do this for `SecretKey`, it is too risky given it is secret material

Understood, changes made

rklaehn:
> We shouldn't do this for `SecretKey`, it is too risky given it is secret material

I am not so sure about this. SecretKey has to_bytes, so it's not like the bytes are hidden. I can see why you don't want to implement Display for SecretKey, but Deref seems pretty harmless.

Nevertheless the place where I find the Deref really useful is the PublicKey/NodeId, so I agree that doing it for SecretKey is not necessary. But it seems that if you have one way to get the bytes adding another slightly more convenient way isn't really bad for security.

rklaehn (approved):

Issue #3250, opened 2025-03-31
WebRTC datachannels


johnny-smitherson:
After #2799 there is no remaining ticket on webrtc. After https://github.com/n0-computer/iroh-examples/issues/113 I would like to publish our findings

# Motivation

WebRTC seems to be the only protocol to handle direct browser-to-browser networking, with wide adoption and support. Implementing it would remove the restriction that all browser traffic flow through iroh relays.

We have experimented with some real-world network configurations (went to people's offices and houses) and have found WebRTC works in the following circumstances:
- one peer is on open internet, no NAT, no firewall, no LAN (e.g. a VPS) -- success 100%
- peers are on same LAN, intranet, VPN, hotspot, host, browser: success 100%
- peers are on different residential/office ISP line: success rate 20%
- peers are on mobile network / cgnat / corporate firewall / "guest" type wifi: success 0%

So its actual use-cases would be closest to the use-cases of the of the "local network discovery with mDNS" feature - in that:
- it successfully detects and connects with all hosts that would have shown up on mDNS
- it works on browsers as well as native binaries
- it always works between all browser tabs on same pc/phone, including from different browsers
- it doesn't really work outside of the LAN space

We think it's interesting for Iroh to have this feature for the following use-cases:
- airdrop-style LAN file transfers, using only browsers instead of mDNS
- apps that combine a native client (e.g. video game client / files server) with a browser client (e.g. chatroom / files search webUI) on the same host
- real-time web app where user can open 10+ tabs in their browser
  - e.g. google-docs style app where each tab has a document, or chat app where each tab has a chat room
  - with WebRTC all tabs can automatically cross-connect without relay traffic
  - when using iroh-gossip, the local peers should naturally cluster together, greatly reducing relay traffic overall

## Disadvantages

- speed
  - ICE takes a few seconds for every peer
- complexity
  - each peer connection would be capable of "upgrading" to webrtc datachannel
  - WebRTC Offer/Answer messages would go over normal iroh connection, until a working datachannel is found
  - handle connection failures/restarts, datachannel backpressure, etc.
  - different native (webrtc.rs) & wasm (wasm-bindgen) implementations required
  - different reliable & unreliable datachannel configurations
- size https://github.com/n0-computer/iroh-examples/issues/113#issuecomment-2747279696
  > by shipping a whole WebRTC implementation in native nodes, we'll blow up code size for all users of iroh a little too much
[...]

PR #3440, opened 2025-08-18
feat(iroh): add initial scaffolding for WebRTC support

anchalshivank:
Hi team,

I’ve added the basic struct for WebRTC support to get things started.
Kindly guide me if this is a good direction or if there’s anything I should take care of early on.

Also, what signaling server setup would you recommend for Iroh’s WebRTC integration?

Thanks!


matheus23:
We're discussing this in the iroh discord: https://discord.com/channels/1161119546170687619/1406897550140772393

iroh/src/disco.rs
```diff
@@ -143,6 +144,8 @@ pub enum SendAddr {
     Udp(SocketAddr),
     /// Relay Url.
     Relay(RelayUrl),
+    /// Node Id
+    NodeId(NodeId)
```
matheus23: I think you'll want this to be sth like `WebRtc(SocketAddr)`. Not sure if there's any other info need to send webrtc data besides the socket addr. Perhaps a channel ID?

iroh/src/discovery/repomix-output.xml
flub, line 1: Erm, not sure this was meant to be here.  You might want to add this to your local .gitignore file.

anchalshivank: This file needs to be deleted

anchalshivank:
> I've had a quick scroll through and noticed a bunch of `println`s and commented out code still.  Could you add a note on what the stated of completion of the PR is, what the first round of review should pay attention to?

This PR is still in progress - I’ll clean up the println! statements and remove the commented-out code before finalizing. For the first round of review, I’d like you to mainly focus on whether I am correctly sending the WebRTC offer/ICE candidate from one node to another.

Current flow

Node A is spawned.

Node B connects.

Node B uses discovery to send a ping .
We do share different message types ping, pong , call me may be . I added another message type.
I also need to confirm when shall the webrtc offer be exchanged.

I refactored the message-sending logic so that send_ping now also carries WebRTC ICE candidate information.


Points where I need guidance

1. Virtual socket: Matheus mentioned we should bind a socket. I’m not fully sure if the current approach is sufficient. I could not get intuition.


2. Refactoring correctness: Did I refactor the send_ping (and related messages) correctly so that ICE candidates are shared properly?


3. Transport mode setup: The PR currently focuses only on native WebRTC with a transport mode of WebrtcRelay. Could you check if I set up the transport modes correctly?


4. Socket binding intuition: In handle::new(), IP transports generate a socket address. For WebRTC, should I also bind it to some sort of socket, or is the transport abstraction enough?


5. WASM case: In WASM, there’s no real socket. How should I be thinking about this when extending the transport to work in WASM?

For now you can ignore webrtc actor (peer connection state ) as it needs to be changed

iroh-base/src/node_addr.rs
```diff
@@ -42,8 +43,18 @@ pub struct NodeAddr {
     pub node_id: NodeId,
     /// The node's home relay url.
     pub relay_url: Option<RelayUrl>,
+    /// The node's channel_id port
+    pub channel_id: Option<ChannelId>,
     /// Socket addresses where the peer might be reached directly.
     pub direct_addresses: BTreeSet<SocketAddr>,
+    /// Static Webrtc connection information for the node
+    pub webrtc_info: Option<WebRtcInfo>,
+}
```
anchalshivank, line 52: **My understanding:**
NodeAddr is used by the discovery system to locate nodes in the network. For discovery purposes, we only need the node_id to find a node. Once found, WebRTC connections are established through the standard offer/answer exchange process.

**My concern:**
Since WebRTC connections require dynamic offer/answer negotiation anyway, is it necessary to store static WebRTC information like channel_id and webrtc_info (certificate fingerprints) here?

iroh-base/src/webrtc_port.rs
anchalshivank, line 1: I added this file to uniquely identify the webrtc connection on the node locally. We need to decide if this correct.

iroh-base/Cargo.toml
anchalshivank, line 1: this file can be ignored for now

iroh-relay/src/node_info.rs
```diff
@@ -224,7 +224,7 @@ impl From<NodeAddr> for NodeData {
 /// `UserData` implements [`FromStr`] and [`TryFrom<String>`], so you can
 /// convert `&str` and `String` into `UserData` easily.
 #[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
-pub struct UserData(String);
+pub struct UserData(pub String);
```
anchalshivank, line 227: Needs to revert it back. It was a quick fix for compilation erros. #3250

iroh/src/magicsock/node_map/node_state.rs
```diff
@@ -587,6 +620,7 @@ impl NodeState {
         // direct address paths to contact but no RelayUrl, we still need to send a DISCO
         // ping to the direct address paths so that the other node will learn about us and
         // accepts the connection.
+        println!("----------- 625");
```
anchalshivank: Needs to revert it back. #3250

iroh/src/magicsock/node_map/node_state.rs
```diff
@@ -115,6 +117,7 @@ pub(super) struct NodeState {
     ///
     /// The fallback/bootstrap path, if non-zero (non-zero for well-behaved clients).
     relay_url: Option<(RelayUrl, PathState)>,
+    webrtc_channel: Option<(ChannelId, PathState)>,
```
anchalshivank, line 141: I think this is incorrect. I need more intuition on what node state , what to put it here

iroh/src/magicsock/transports/webrtc/actor.rs
anchalshivank, line 1: We can ignore this file for now

iroh/src/magicsock/node_map.rs
```diff
@@ -55,6 +56,7 @@ pub(super) struct NodeMap {
 pub(super) struct NodeMapInner {
     by_node_key: HashMap<NodeId, usize>,
     by_ip_port: HashMap<IpPort, usize>,
+    by_webrtc_port: HashMap<WebRtcPort, usize>,
```
anchalshivank, line 65: I also need guidance on this part. If this is correct or needs changes

iroh/src/endpoint.rs
anchalshivank, line 1: We can ignore this file for now

iroh/src/magicsock.rs
```diff
@@ -825,9 +849,20 @@ impl MagicSock {
...
                             self.send_ping_queued(ping);
                         }
+                        PingAction::InitiateWebRtc(ping) => {
+                            println!("Received disco webrtc message! {:?}", ping);
+                        }
                     }
                 }
             }
+            disco::Message::WebRtcIceCandidate(ice) => {
+
```
anchalshivank, line 967: When we use [this file](https://github.com/anchalshivank/iroh-examples/blob/main/webrtc-chat/src/bin/cli2.rs), we can see the offer being received at this point.

iroh/src/magicsock.rs
```diff
@@ -1533,134 +1656,1195 @@ impl Handle {
...
+        let actor_task = Arc::new(Mutex::new(Some(AbortOnDropHandle::new(actor_task))));
+
+        Ok(Handle {
+            msock,
+            actor_task,
+            endpoint,
+            actor_token,
+        })
+    }
```
anchalshivank: We can ignore this part for now

iroh/src/magicsock.rs
```diff
@@ -1533,134 +1656,1195 @@ impl Handle {
...
+
+    /// The underlying [`quinn::Endpoint`]
+    pub fn endpoint(&self) -> &quinn::Endpoint {
+        &self.endpoint
+    }
```
anchalshivank: Also ignore this part

iroh/src/disco.rs
```diff
@@ -57,8 +56,11 @@ pub enum MessageType {
     Ping = 0x01,
     Pong = 0x02,
     CallMeMaybe = 0x03,
+    WebRtcIceCandidate = 0x06
```
anchalshivank: I think this will also need some refactoring

iroh/Cargo.toml
anchalshivank, line 1: Since I was running the complete stack locally, I had to modify the Cargo.toml. I think we should consider adding Cargo.toml to .gitignore, unless you suggest a better approach.

iroh-base/src/webrtc_port.rs
```diff
@@ -0,0 +1,189 @@
...
+
+/// A unique identifier for a WebRTC connection.
+///
+/// In the iroh network, WebRTC connections are established between nodes and need to be
+/// uniquely identified to handle multiple concurrent connections. A [`WebRtcPort`] combines
+/// a [`NodeId`] (which identifies the peer node) with a [`WebRtcPort`] (which identifies
+/// the specific channel/connection to that node).
+///
+/// This is particularly useful when:
+/// - A node needs to maintain multiple WebRTC connections to the same peer
```
flub, line 18: When would we need to maintain multiple WebRTC connections between the same two peers?  My mental model was that a single WebRTC connection would transport QUIC datagrams, which can be for any number of QUIC connections.  So I wasn't expecting to ever need more than one WebRTC connection between two nodes.

iroh-base/src/webrtc_port.rs
```diff
@@ -0,0 +1,189 @@
...
+    }
+}
+
+/// A unique identifier for a WebRTC channel.
+///
+/// [`WebRtcPort`] is used to distinguish between multiple WebRTC data channels or connections
+/// to the same peer node. It's a 16-bit unsigned integer, allowing for up to 65,536 unique
+/// channels per node pair.
+///
+/// The channel ID space is managed by the WebRTC implementation and should be:
```
flub, line 91: If managed by the WebRTC implementation why does the example show creating this from a static number?  That's confusing to me.  Or is it trying to show something else?

iroh-base/src/webrtc_port.rs
```diff
@@ -0,0 +1,189 @@
...
+    /// Creates a [`WebRtcPort`] from a `u16` value.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use iroh_base::WebRtcPort;
+    ///
+    /// let channel = WebRtcPort::from(42u16);
+    /// assert_eq!(channel.as_u16(), 42);
+    /// ```
```
flub, line 162: This level of comments is probably not very valuable.  Everything it says and shows is already said and shown in the type signature.  It doesn't tell me anything extra.  It's always worth looking at comments from the view of what information you are learning that there otherwise isn't yet, or is hard to extract.

iroh-base/src/node_addr.rs
```diff
@@ -52,7 +63,9 @@ impl NodeAddr {
         NodeAddr {
             node_id,
             relay_url: None,
+            channel_id: None,
             direct_addresses: Default::default(),
+            webrtc_info: None,
```
flub, line 68: Why are these two separate fields?

iroh/Cargo.toml
```diff
@@ -41,9 +41,10 @@ iroh-base = { version = "0.91.2", default-features = false, features = ["key", "
 iroh-relay = { version = "0.91", path = "../iroh-relay", default-features = false }
 n0-future = "0.1.2"
 n0-snafu = "0.2.1"
-n0-watcher = "0.3"
+#n0-watcher = "0.3"
+n0-watcher = {path = "../../n0-watcher"}
 nested_enum_utils = "0.2.1"
-netwatch = { version = "0.8" }
+netwatch = { path = "../../net-tools/netwatch" }
```
flub, line 47: Is there a corresponding net-tools PR?

iroh/Cargo.toml
```diff
@@ -101,14 +102,20 @@ tracing-subscriber = { version = "0.3", features = [
 ], optional = true }
 indicatif = { version = "0.18", features = ["tokio"], optional = true }
 parse-size = { version = "=1.0.0", optional = true, features = ['std'] } # pinned version to avoid bumping msrv to 1.81
+webrtc = "0.13.0"
+web-sys = "0.3.77"
+wasm-bindgen = "0.2.100"
+js-sys = "0.3.77"
+wasm-bindgen-futures = "0.4.50"
```
flub, line 109: Are these needed even for non-wasm?

iroh/src/disco.rs
```diff
@@ -57,6 +56,9 @@ pub enum MessageType {
     Ping = 0x01,
     Pong = 0x02,
     CallMeMaybe = 0x03,
+    WebRtcOffer = 0x06,
+    WebRtcAnwser = 0x07,
+    WebRtcIceCandidate = 0x08,
```
flub, line 61: Why choose to add new message types rather than extend the CallMeMaybe message?  Are there situations where you might only want to do UDP holepunching and where you only want to do WebRTC holepunching?  But even then you can populate different parts of the CallMeMaybe message.

I guess you have to do more of a handshake with the offer and answer.  Can you summarise the current flow of these exchanges over the relay and how it establishes a WebRTC connection?

iroh/src/disco.rs
```diff
@@ -111,6 +116,189 @@ pub enum Message {
...
+pub struct WebRtcOffer {
+    // Using a simple string for the candidate for now
+    pub offer: String,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct WebRtcAnswer {
+    // Using a simple string for the candidate for now
+    pub answer: String,
+    pub received_offer: String,
```
flub, line 231: Is this echoing back the offer to which this answer is a reply to?

iroh/src/endpoint.rs
```diff
@@ -701,6 +785,14 @@ impl Endpoint {
             .await?;
         let conn = connecting.await?;
 
+        //Now try to setup webrtc p2p connection
+
+        match self.msock.intiate_webrtc_offer(remote).await {
```
flub, line 790: I think this is not something that should be done here once on it's own?  I'd expect it should happen whenever the NodeMap/NodeState decides it needs to send a new call-me-maybe because it needs to find a better address.

iroh/src/magicsock/transports.rs
```diff
@@ -23,18 +35,33 @@ use self::ip::{IpNetworkChangeSender, IpSender};
...
     relay: Vec<RelayTransport>,
-
+    webrtc: Vec<WebRtcTransport>,
+    max_receive_segments: Arc<AtomicUsize>,
     poll_recv_counter: AtomicUsize,
 }
 
+///Transport Mode
+#[derive(Debug)]
+pub enum TransportMode {
```
flub, line 52: What's the motivation for controlling this?  So far we haven't given folks the option to select transports, so it seems like adding a new feature to me.

I do think there's a test-only feature to do relay-only or so, but we haven't decided if that should be exposed or not I think.

iroh/src/magicsock.rs
```diff
@@ -47,20 +48,41 @@ use quinn::{AsyncUdpSocket, ServerConfig};
...
+};
+use std::{
+    hash::{DefaultHasher, Hash, Hasher},
+    sync::atomic::AtomicUsize,
+};
+use tokio::{
+    net::unix::pipe::Sender,
+    sync::{Mutex as AsyncMutex, mpsc, oneshot},
+    task::block_in_place,
+};
```
flub, line 71: general import ordering according to the rust style guide is:

1. std
2. 3rd party
3. crate local

Let's keep to that, and certainly avoid multiple blocks for the same group, like you have std groups.

`cargo make format` in our repo (you need `cargo-make` installed) should hopefully do most of the formatting for you.

iroh/src/magicsock.rs
```diff
@@ -825,6 +872,131 @@ impl MagicSock {
...
+                for action in actions {
+                    match action {
+                        PingAction::SendWebRtcAnswer(offer) => {
+                            let actor_sender = self.actor_sender.clone();
+
+                            task::spawn(async move {
+                                if let Err(err) = actor_sender
+                                    .send(ActorMessage::SendWebRtcAnswer(offer))
+                                    .await
+                                {
```
flub, line 897: FWIW generally you can't do this: you're facing a bounded channels so sending can block, but you can't block here.  By solving this by spawning a task you're turning this bounded channel into an unbounded channel using the tokio runtime as your queue.  Might as well have used an unbounded channel, at least that's slightly less harmful and clearer in conveying how it works (but we also don't like those in iroh).

This will require more thought.  BUT!  Let's not yet worry about that for now, I'm focusing on understanding the required architecture for now, so let's not get distracted with trying to solve this yet.

iroh/src/magicsock.rs
```diff
@@ -825,6 +872,131 @@ impl MagicSock {
...
+                let value = answer.clone();
+
+                for action in actions {
+                    match action {
+                        PingAction::SetRemoteDescription => {
+                            let webrtc_sender = self.webrtc_actor_sender.clone();
+                            let value = value.clone();
+
+                            let _ = block_in_place(|| {
+                                tokio::runtime::Handle::current().block_on(async move {
```
flub, line 928: This is a creative attempt to solve the same problem.  But you're still blocking the current task and thus blocking driving receiving more datagrams from the wire.  So it makes the compiler happy, but the runtime behaviour still has the same problems.

I'd appreciate if for now you just used the same technique twice instead of obfuscating this problem in different ways.  And leave a clear todo note that this still need to be solved.

iroh/src/magicsock.rs
```diff
@@ -825,6 +872,131 @@ impl MagicSock {
...
+                    &self.metrics.magicsock,
+                );
+
+                for action in &actions {
+                    match action {
+                        PingAction::AddWebRtcIceCandidate => {
+                            let actor_sender = self.actor_sender.clone();
+
+                            block_in_place(|| {
+                                tokio::runtime::Handle::current().block_on(async {
```
flub, line 980: same

iroh/src/magicsock.rs
```diff
@@ -1532,50 +1932,265 @@ impl Handle {
             actor_token,
         })
     }
+    /// Creates a magic [`MagicSock`]
+    async fn new_webrtc_relay(opts: Options) -> Result<Self, CreateHandleError> {
```
flub, line 1936: What does the term "webrtc relay" mean in the WebRTC world?  I'm assuming this has little to do with the iroh relay?

iroh/src/magicsock.rs
```diff
@@ -1834,6 +2462,56 @@ fn bind_ip(
...
 
+#[cfg(not(wasm_browser))]
+fn bind_webrtc(
+    addr_v4: SocketAddrV4,
+    addr_v6: Option<SocketAddrV6>,
+    metrics: &EndpointMetrics,
+    secret_key: SecretKey,
+) -> io::Result<(Vec<WebRtcTransport>, portmapper::Client)> {
+    let port_mapper =
+        portmapper::Client::with_metrics(Default::default(), metrics.portmapper.clone());
```
flub, line 2473: What is the portmapper used for?

iroh/src/magicsock.rs
```diff
@@ -1834,6 +2462,56 @@ fn bind_ip(
...
+    let addr_v6 =
+        addr_v6.unwrap_or_else(|| SocketAddrV6::new(Ipv6Addr::UNSPECIFIED, ip6_port, 0, 0));
+
+    let v6 = match bind_with_fallback(SocketAddr::V6(addr_v6)) {
+        Ok(sock) => Some(Arc::new(sock)),
+        Err(err) => {
+            info!("bind ignoring IPv6 bind failure: {:?}", err);
+            None
+        }
+    };
```
flub, line 2488: is the v6 bind used at all?

iroh/src/magicsock.rs
```diff
@@ -1883,126 +2561,126 @@ impl Actor {
...
             tokio::select! {
-                _ = shutdown_token.cancelled() => {
-                    debug!("shutting down");
-                    return;
-                }
-                msg = self.msg_receiver.recv(), if !receiver_closed => {
-                    let Some(msg) = msg else {
-                        trace!("tick: magicsock receiver closed");
-                        self.msock.metrics.magicsock.actor_tick_other.inc();
+            _ = shutdown_token.cancelled() => {
```
flub, line 2564: Please leave the indentation in place.  Formatting in macro's is a bit pants, but let's keep it as close as we can to normal formatting.  And also try and keep this diff as readable as possible, so we can see just the new branches of the select.

iroh/src/magicsock/transports/webrtc.rs
```diff
@@ -0,0 +1,543 @@
...
+        dest_node: NodeId,
+        transmit: &Transmit,
+        channel_id: &ChannelId,
+    ) -> Poll<io::Result<()>> {
+        // Reserve space in the send queue
+        match ready!(self.sender.poll_reserve(cx)) {
+            Ok(()) => {
+                trace!(node = %dest_node, "WebRTC send: reserving channel space");
+
+                let payload = Bytes::copy_from_slice(transmit.contents);
```
flub, line 153: So everything we send via WebRTC pays the copy penalty?  That's not great.

iroh/src/magicsock/transports/webrtc.rs
```diff
@@ -0,0 +1,543 @@
...
+
+                let data = WebRtcData {
+                    channel_id: channel_id.clone(),
+                    delivery_mode: WebRtcDeliveryMode::Reliable, // TODO: Make configurable
```
flub, line 157: Datagrams are unreliable.  Should we not be sending these all as Unreliable?  By sending QUIC datagrams inside a reliable channel we have both QUIC and WebRTC doing buffering and re-ordering.  This is generally bad for congestion controllers who will start to misbehave.  Yes, the relay channel has this problem too - but here we have the option of doing something better it seems.

iroh/src/magicsock/transports/webrtc/actor.rs
```diff
@@ -0,0 +1,1108 @@
...
+    /// The destination for the WebRTC data
+    pub(crate) dest_node: NodeId,
+    /// WebRTC-specific data to send
+    pub(crate) data: WebRtcData,
+}
+
+pub(crate) enum WebRtcActorMessage {
+    CreateOffer {
+        local_node: PublicKey,
+        peer_node: PublicKey,
```
flub, line 134: Please use the `NodeId` type alias when appropriate everywhere.

The docs of `NodeId` should give you guidance: https://docs.rs/iroh/latest/iroh/type.NodeId.html

iroh/src/magicsock/transports/webrtc/actor.rs
```diff
@@ -0,0 +1,1108 @@
...
+        let peer_node = self.peer_node;
+        let sender = self.send_recv_datagram.clone();
+
+        dc.on_open(Box::new(move || {
+            Box::pin(async move {
+                println!("✅ Data channel OPENED for peer {}", peer_node);
+            })
+        }));
+
+        dc.on_message(Box::new(move |msg| {
```
flub, line 548: yikes, I guess the webrtc libs don't do poll interfaces?

iroh-base/src/webrtc_port.rs
anchalshivank: You are correct. I will fix this

iroh/src/magicsock/transports/webrtc.rs
flub: Having read more about how the new webrtc actor works and the libs it has to use... yeah, this seems unavoidable.  Sad though, that there's no better WebRTC implementation in rust.  OTOH it's probably nice to keep the wasm and native code somewhat close to each other.

iroh-base/src/webrtc_port.rs
anchalshivank: These comments need to be corrected. The main idea is that a port should be uniquely identified by both channel_id and node_id

iroh/Cargo.toml
anchalshivank: No, to run locally there were some dependency issue

iroh/src/magicsock.rs
anchalshivank: ok

iroh/src/magicsock/transports/webrtc/actor.rs
anchalshivank: ok

iroh/src/magicsock/transports/webrtc/actor.rs
anchalshivank: No it does not have poll interfaces

iroh/src/magicsock/transports.rs
anchalshivank: I added this code in a way that doesn’t change existing behavior.
My main confusion was around how a user should indicate which transport to use in iroh. As a quick solution, I introduced this controller so we can experiment with different modes without altering the current defaults.

Once the team confirms how users should select them, we can clean this up or remove the controller entirely.

iroh/src/magicsock/transports/webrtc.rs
anchalshivank: Yeah make sense. Hopefully at some point a more efficient WebRTC crate shows up, but this works for now.

iroh/src/magicsock/transports/webrtc.rs
anchalshivank: From what I understand, in WebRTC with a live P2P connection, poll_recv just returns whatever data arrives on the channel. In the actor, that channel forwards the data straight to the msock layer, and then it shows up in process_datagram.

I think there is no quic ordering happening

anchalshivank: I’m also a bit confused about how the WebRTC handle should be exposed to the user ,once I have a clearer picture of that, I think I’ll better understand where this QUIC reordering happens.

iroh/src/magicsock/transports.rs
flub: On first approximation I think the user should not get any say.  We should use:
1. Direct if possible.
2. WebRTC if direct is not possible.
3. Relay if neither of the above are possible.

That's also how we've done it up to now (without WebRTC of course)

iroh/src/magicsock/transports/webrtc.rs
flub: > I think there is no quic ordering happening

The general model of iroh (for the purpose of this discussion) is that it sends QUIC datagrams over several unreliable transports.  Primarily this is UDP, the transport QUIC was originally designed for.  So we want to keep our transports as close to the behaviour of UDP as possible.

The user gets a QUIC API to create streams and send user data.  The QUIC stack manages all things network when it builds QUIC packets and puts them in datagrams: it handles the rate at which to send them, re-sending if needed, buffering if out of order etc.  So all we want from WebRTC is to give us an unreliable datagram transport to put QUIC datagrams on.

So the user never gets to see or interact with WebRTC.  It's a transport iroh uses under the hood.  A user should be entirely unaware of WebRTC (ideally, they will probably be aware that on native we have to bind a socket for it as I understand).  The only thing it enables is that iroh can establish direct QUIC connections when at least one of the nodes is in a web browser.

iroh/src/magicsock/transports.rs
anchalshivank: I understand this part, we have to changes here - 
https://github.com/n0-computer/iroh/blob/4583b1280b8f5ab9d264940102382956ddbbb1a8/iroh/src/magicsock/transports.rs#L371

My current logic is as follows: first, iroh discovers another node. After that, I initiate a WebRTC offer. (I haven’t implemented a fallback mechanism if the offer fails , this definitely needs improvement.)

Once the WebRTC connection is established, my main concern is whether I’m mapping port addresses correctly. This is the most confusing part for me. I also need guidance on how `node_map` and `node_state` should be updated during this process.

1. The offer generation happens here: [[endpoint.rs#L790](https://github.com/anchalshivank/iroh/blob/4540a68b5660e03e5474934b6d50601f7c60cfa9/iroh/src/endpoint.rs#L790)](https://github.com/anchalshivank/iroh/blob/4540a68b5660e03e5474934b6d50601f7c60cfa9/iroh/src/endpoint.rs#L790).
2. The request is handled by the actor here: [[magicsock.rs#L1510](https://github.com/anchalshivank/iroh/blob/4540a68b5660e03e5474934b6d50601f7c60cfa9/iroh/src/magicsock.rs#L1510)](https://github.com/anchalshivank/iroh/blob/4540a68b5660e03e5474934b6d50601f7c60cfa9/iroh/src/magicsock.rs#L1510) and then here: [[magicsock.rs#L2778](https://github.com/anchalshivank/iroh/blob/4540a68b5660e03e5474934b6d50601f7c60cfa9/iroh/src/magicsock.rs#L2778)](https://github.com/anchalshivank/iroh/blob/4540a68b5660e03e5474934b6d50601f7c60cfa9/iroh/src/magicsock.rs#L2778).
3. I don’t fully understand this part: [[magicsock.rs#L2824](https://github.com/anchalshivank/iroh/blob/4540a68b5660e03e5474934b6d50601f7c60cfa9/iroh/src/magicsock.rs#L2824)](https://github.com/anchalshivank/iroh/blob/4540a68b5660e03e5474934b6d50601f7c60cfa9/iroh/src/magicsock.rs#L2824). I copied it from existing code to send data to another node, and it worked, but I’d like to understand it properly.
4. While creating the offer, I also start gathering [[ICE candidates](https://github.com/anchalshivank/iroh/blob/4540a68b5660e03e5474934b6d50601f7c60cfa9/iroh/src/magicsock.rs#L1212)](https://github.com/anchalshivank/iroh/blob/4540a68b5660e03e5474934b6d50601f7c60cfa9/iroh/src/magicsock.rs#L1212) (the receiver exists in the Actor). As soon as candidates are gathered, they’re sent to the peer here: [[magicsock.rs#L2768](https://github.com/anchalshivank/iroh/blob/4540a68b5660e03e5474934b6d50601f7c60cfa9/iroh/src/magicsock.rs#L2768)](https://github.com/anchalshivank/iroh/blob/4540a68b5660e03e5474934b6d50601f7c60cfa9/iroh/src/magicsock.rs#L2768).

I still don’t fully understand how iroh sends data to another peer. I copied some parts from disco to handle this, and it worked, but I need clarity. Once the data is sent, it’s received in `handle_disco_message()`. I noticed that disco messages only contain addresses derived from `Ip(node_id)`.

On the peer side, I repeat the process: create an answer, send ICE candidates, and follow the same flow.

At this point:

* Peer A sets `localDescription(offer)` and `remoteDescription(answer)` from Peer B, gathers ICE candidates, and adds Peer B’s candidates.
* Peer B sets `localDescription(answer)` and `remoteDescription(offer)`, gathers its ICE candidates, and adds Peer A’s.

The data channel then opens.

After this stage, my current implementation can send packets via `process_datagram` at [[magicsock.rs#L765](https://github.com/anchalshivank/iroh/blob/4540a68b5660e03e5474934b6d50601f7c60cfa9/iroh/src/magicsock.rs#L765)](https://github.com/anchalshivank/iroh/blob/4540a68b5660e03e5474934b6d50601f7c60cfa9/iroh/src/magicsock.rs#L765).

So far, it works - but I need to know if I made mistakes. In particular, I’d like confirmation on whether I correctly initialized `PingActions` and on what how to update node state.

iroh/src/magicsock.rs
anchalshivank: ok

iroh/src/magicsock.rs
anchalshivank: Thanks, I will do this

iroh/src/magicsock.rs
anchalshivank: This is just a temporary hack since the function isn’t async. I’ll add a TODO in the code and create an issue so we can revisit this properly

iroh/src/magicsock.rs
anchalshivank: The function name needs to be updated. Initially, when I wrote it, my assumption was that we’d only use the relay to discover iroh nodes, then exchange offer/answer via the relay, and finally set up the WebRTC connection.

In the end, I think there should only be a single function, async fn new().

As I mentioned earlier, I didn’t want to change code that already works. Once I get guidance on how to properly initialize the WebRTC transport, I’ll merge those changes into new().

iroh/src/magicsock.rs
anchalshivank: Honestly, I’m not sure. I copied this logic from [[here](https://github.com/anchalshivank/iroh/blob/4540a68b5660e03e5474934b6d50601f7c60cfa9/iroh/src/magicsock.rs#L2421)](https://github.com/anchalshivank/iroh/blob/4540a68b5660e03e5474934b6d50601f7c60cfa9/iroh/src/magicsock.rs#L2421).

The reason was that I was hitting errors like *[[could not bind the port](https://github.com/anchalshivank/iroh/blob/4540a68b5660e03e5474934b6d50601f7c60cfa9/iroh/src/magicsock.rs#L1715)](https://github.com/anchalshivank/iroh/blob/4540a68b5660e03e5474934b6d50601f7c60cfa9/iroh/src/magicsock.rs#L1715)*. To work around it, I copied that part of the code.

What I don’t fully understand is: which port is this referring to, and what exactly is being bound here? I’d really appreciate some guidance on this part.

iroh/src/magicsock.rs
anchalshivank: same https://github.com/n0-computer/iroh/pull/3440#discussion_r2355331849

iroh-base/src/node_addr.rs
anchalshivank: This needs to be fixed.
I think we do not need webrtc info and channel id here

iroh/Cargo.toml
anchalshivank: No

iroh/src/disco.rs
anchalshivank: Currently yes, this needs to be fixed

iroh/src/endpoint.rs
anchalshivank: Could you point me to the code that actually triggers the handshake? I wasn’t able to find it.

Also, I don’t fully understand the part you mentioned: *“whenever the NodeMap/NodeState decides it needs to send a new call-me-maybe”*. Could you please clarify and, if possible, provide a link to the relevant code?

iroh/src/disco.rs
anchalshivank: https://github.com/n0-computer/iroh/pull/3440#discussion_r2355278536

anchalshivank: I also noticed that `CallMeMaybe` messages are already being sent at regular intervals [[here](https://github.com/anchalshivank/iroh/blob/4540a68b5660e03e5474934b6d50601f7c60cfa9/iroh/src/magicsock.rs#L2658)](https://github.com/anchalshivank/iroh/blob/4540a68b5660e03e5474934b6d50601f7c60cfa9/iroh/src/magicsock.rs#L2658). Extending `CallMeMaybe` to carry WebRTC offers and answers would be too much, especially since offer/answer exchanges are triggered via a function call and only happen **once**.

That’s why I opted to add a new `MessageType` instead.

flub:
Hi, the conflicts on this branch are a bit too much now with mutlipath. The recommended approach is to do this probably using custom transports. See https://github.com/n0-computer/iroh/discussions/4024 where this conversation is continuing.

PR #3441, opened 2025-08-19
cargo: add aarch64-linux-gnu-gcc linker

### LINKED ISSUE #3439: bug: can't cross-compile for linux-arm64
**Describe the bug**
`cargo build --target=aarch64-unknown-linux-gnu` causes a bunch of errors that mostly look like this:
```
/usr/bin/ld: /home/iameli/code/iroh/target/aarch64-unknown-linux-gnu/debug/deps/rustls_platform_verifier-20fc7a067ba9027e.rustls_platform_verifier.913ca4cd1c10469e-cgu.0.rcgu.o: Relocations in generic ELF (EM: 183)
```

Interestingly, this doesn't stop me from cross-compiling from linux-amd64 to a variety of other targets, like darwin-arm64, darwin-amd64, and windows-amd64.

**Relevant Logs**
[Full logs here](https://gist.github.com/iameli/db128c2df4f3332d65f9edb697febc0d).

**Expected behavior**
Working cross-compilation

**Platform(s)**
Desktop:
I'm running the Streamplace builder image, which you can download from `docker pull us-central1-docker.pkg.dev/aquareum-tv/streamplace/build:builder-af492185d8fb948d84acace9029719df5d5dbf05` or build for yourself from https://github.com/streamplace/streamplace/blob/next/docker/build.Dockerfile

iameli:
Fixes https://github.com/n0-computer/iroh/issues/3439

How... is this not the default? Other platforms cross-compile just fine. Other Rust C code cross-compiles just fine. Cargo bug? IDK. Anyway.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,3 +1,6 @@
 [target.wasm32-unknown-unknown]
 runner = "wasm-bindgen-test-runner"
 rustflags = ['--cfg', 'getrandom_backend="wasm_js"']
+
+[target.aarch64-unknown-linux-gnu]
+linker = "aarch64-linux-gnu-gcc"


flub:
Is there also an LLVM linker normally?

iameli:
> Is there also an LLVM linker normally?

I tried that via `linker-flavor=lld` but it didn't help, just produced a different error message.

flub:
Fair.  Does this change affect compiling natively?  I notice locally that my `ld` is a symlink to `x86_64-linux-gnu-ld` so presumably that would also be the case on aarch64.  Though this might possibly be changing the default linker from llvm to gnu natively?  Is that a bad thing?

To be clear, I have no idea really.  Am only trying to think of how this change could go wrong.  I think in theory the linkers can all be mixed, and if someone cares deeply for perf or other reasons they already customise it probably.  So my guess is this is fine.

dignifiedquire:
is this actually distributed when publishing to crates.io? otherwise this should be documentation addition rather than a config change, as that is only for our CI asfaik

iameli:
> Though this might possibly be changing the default linker from llvm to gnu natively? Is that a bad thing?

I was under the impression that the gnu linker was the default on all Linux architectures; the log line from https://github.com/n0-computer/iroh/issues/3439 is definitely from `cc` and not `clang`.

Maybe this would affect Mac-to-Linux cross compilation? That might use clang as a default.

flub (approved):

flub:
I think @dignifiedquire is right that this probably will not affect use when iroh is used as a library though.

teohhanhui:
This breaks local dev of Iroh on Linux aarch64, e.g. Asahi Linux on Apple Silicon.

Hard-coding the cross-linker for an entire platform can't be right...

PR #3442, opened 2025-08-22
feat: Allow configuring the max number of TLS tickets

rklaehn:
## Description

Allows increasing the max number of TLS tickets without having to patch iroh. This is needed if you want to do a large number of 0rtt connections to different clients.

## Notes & open questions

Q: is this enough, or should we allow passing in an `Arc<dyn ClientSessionStore>`? The latter would mean exposing one more rustls type.

Note: this is mostly for playing around with it, not sure if we want to merge it as is. But then, it seems pretty straightforward. If we ever want the ability to provide an `Arc<dyn ClientSessionStore>` we can always make the `fn max_tls_tickets` just init an in-mem provider.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -45,7 +45,7 @@ use crate::{
     magicsock::{self, Handle, NodeIdMappedAddr, OwnAddressSnafu},
     metrics::EndpointMetrics,
     net_report::Report,
-    tls,
+    tls::{self, DEFAULT_MAX_TLS_TICKETS},
 };
 
 mod rtt_actor;
@@ -118,6 +118,7 @@ pub struct Builder {
     addr_v6: Option<SocketAddrV6>,
     #[cfg(any(test, feature = "test-utils"))]
     path_selection: PathSelection,
+    max_tls_tickets: usize,
 }
 
 impl Default for Builder {
@@ -142,6 +143,7 @@ impl Default for Builder {
             addr_v6: None,
             #[cfg(any(test, feature = "test-utils"))]
             path_selection: PathSelection::default(),
+            max_tls_tickets: DEFAULT_MAX_TLS_TICKETS,
         }
     }
 }
@@ -160,7 +162,7 @@ impl Builder {
             .unwrap_or_else(|| SecretKey::generate(rand::rngs::OsRng));
         let static_config = StaticConfig {
             transport_config: Arc::new(self.transport_config),
-            tls_config: tls::TlsConfig::new(secret_key.clone()),
+            tls_config: tls::TlsConfig::new(secret_key.clone(), self.max_tls_tickets),
             keylog: self.keylog,
         };
         let server_config = static_config.create_server_config(self.alpn_protocols);
@@ -472,6 +474,15 @@ impl Builder {
     #[cfg(any(test, feature = "test-utils"))]
     pub fn path_selection(mut self, path_selection: PathSelection) -> Self {
         self.path_selection = path_selection;
+        self
+    }
+
+    /// Set the maximum number of TLS tickets to cache.
+    ///
+    /// Set this to a larger value if you want to do 0rtt connections to a large
+    /// number of clients.
+    pub fn max_tls_tickets(mut self, n: usize) -> Self {
+        self.max_tls_tickets = n;
         self
     }
 }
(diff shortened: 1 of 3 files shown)


matheus23:
> Q: is this enough, or should we allow passing in an `Arc<dyn ClientSessionStore>`? The latter would mean exposing one more rustls type.
> 
> Note: this is mostly for playing around with it, not sure if we want to merge it as is. But then, it seems pretty straightforward. If we ever want the ability to provide an `Arc<dyn ClientSessionStore>` we can always make the `fn max_tls_tickets` just init an in-mem provider.

I went back and forth on this quite a bit.
Where I ended up was:
- I didn't want to add an option like `max_tls_tickets`, as I think that's too little control for people who use 0-RTT as a "power user". I think you quickly run into "how do I persist across restarts"
- I think we actually want to expose (something like) the whole `ClientSessionStore` trait. But I didn't want to introduce another non-1.0 dependency's (rustls) trait before iroh's 1.0.
- I wanted to avoid having both: It's weird that setting `max_tls_store` and then `client_session_store` one after another invalidates them. It's not immediately obvious that the first's effect gets overwritten by the second and vice versa.

rklaehn:
> * I wanted to avoid having both: It's weird that setting `max_tls_store` and then `client_session_store` one after another invalidates them. It's not immediately obvious that the first's effect gets overwritten by the second and vice versa.

What is better? A maybe slightly unintuitive API or being unable to use 0rtt at all for certain use cases? We got a similar situation with discovery. https://docs.rs/iroh/latest/iroh/endpoint/struct.Builder.html#method.discovery will *replace* discovery with the given discovery method. So if somebody does

```rust
.discovery_n0()
.discovery(my_custom_discovery)
```

They won't have discovery_n0, if I read the docs correctly. Being able to just adjust the size means that we don't need to expose another rustls type, but people can still use 0rtt for endpoints with many connections.

iroh/src/endpoint.rs
```diff
@@ -474,6 +476,15 @@ impl Builder {
         self.path_selection = path_selection;
         self
     }
+
+    /// Set the maximum number of TLS tickets to cache.
+    ///
+    /// Set this to a larger value if you want to do 0rtt connections to a large
+    /// number of clients.
```
flub, line 483: Could this mention the default?  Right now it is not clear that 0rtt is possible at all without setting this.  I think that should be made more clear.

flub:
> > Q: is this enough, or should we allow passing in an `Arc<dyn ClientSessionStore>`? The latter would mean exposing one more rustls type.
> > Note: this is mostly for playing around with it, not sure if we want to merge it as is. But then, it seems pretty straightforward. If we ever want the ability to provide an `Arc<dyn ClientSessionStore>` we can always make the `fn max_tls_tickets` just init an in-mem provider.
> 
> I went back and forth on this quite a bit. Where I ended up was:
> 
>     * I didn't want to add an option like `max_tls_tickets`, as I think that's too little control for people who use 0-RTT as a "power user". I think you quickly run into "how do I persist across restarts"
> 
>     * I think we actually want to expose (something like) the whole `ClientSessionStore` trait. But I didn't want to introduce another non-1.0 dependency's (rustls) trait before iroh's 1.0.
> 
>     * I wanted to avoid having both: It's weird that setting `max_tls_store` and then `client_session_store` one after another invalidates them. It's not immediately obvious that the first's effect gets overwritten by the second and vice versa.

I think I'd go with `max_tls_tickets` for 1.0 so we don't have to expose more foreign types.  And it allows us to add `client_session_store` later.  I know that means you have this situation where the advanced option overrides the simple options, though to me this seems reasonable.  You have that behaviour already, because there is an implicit max_tls_tickets right now that would get overridden by client_session_store.

iroh/src/endpoint.rs
rklaehn: Make the constant public or just duplicate the value in the docs? Maybe the latter, comments are not subject to semver... Just need to make sure that it remains consistent.

flub: Yeah, I was just thinking of duplicating it in the docs.  You can also be more vague in the docs, like "by default a small number, around 20, taking up about xxx KiB of memory".

rklaehn: Pushed an update.

flub (approved):

matheus23 (approved):

PR #3445, opened 2025-08-29
Add jitter to staggered DNS calls

julianpryde:
## Description

Adds a random jitter to DNS retries called with the `stagger_call` function.

## Breaking Changes

N/A

## Notes & open questions

Further research can determine if additional timeout jitters are needed when calling the other DnsResolver public functions.

Nominal 20% of default delay time as maximum deviation from original timeout delay.  Real world and lab testing can dermine if this is sufficient.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


julianpryde:
Closed because I was not on the most recent commit.  Will fetch from upstream and resubmit

Issue #3017, opened 2024-12-06
Add some jitter to dns retries


Arqu:
Currently when running a dns server it's possible to get in a degraded state after restarting the dns service as all the clients start pick up the dns server coming back at the same time and then the follow up re-publishing. Which results in a wave pattern of connections.

julianpryde:
I can write a PR for this.  I think adding the jitter to the timeout variables in the `lookup_ipv4()` and `lookup_ipv6()` functions in iroh/iroh-relay/src/dns.rs would be the smallest change necessary to make that happen.  Does that sound right?

flub:
Hum, I'm not sure it's that simple.  If we do that we're delaying every normal use of DNS.  I think we need to view this more from the callsites, list them all and then evaluate how each should behave.

What probably is safe thing to do generally is to add jitter to the `stagger_call` function.  That function exists for retries, so globally adding jitter to the retries is relatively safe I think.

julianpryde:
PR created.  I'll run some trials in a test environment to see how it changes performance.

PR #3447, opened 2025-08-30
fix(iroh): add jitter to dns retry calls

julianpryde:
## Description

Adds a random jitter to DNS retries called with the `stagger_call` function.

## Breaking Changes

N/A

## Notes & open questions

Further research can determine if additional timeout jitters are needed when calling the other DnsResolver public functions.

Nominal 20% of default delay time as maximum deviation from original timeout delay.  Real world and lab testing can dermine if this is sufficient.

Partially addresses #3017.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-relay/src/dns.rs
+++ b/iroh-relay/src/dns.rs
@@ -567,7 +567,7 @@ pub(crate) mod tests {
 
     #[test]
     #[traced_test]
-    async fn jitter_test_zero() {
+    fn jitter_test_zero() {
         let jittered_delay = add_jitter(&0);
         assert_eq!(jittered_delay, Duration::from_secs(0));
     }
--- a/iroh-relay/src/dns.rs
+++ b/iroh-relay/src/dns.rs
@@ -565,7 +565,7 @@ pub(crate) mod tests {
         assert_eq!(result, 5)
     }
 
-    #[tokio::test]
+    #[test]
     #[traced_test]
     async fn jitter_test_zero() {
         let jittered_delay = add_jitter(&0);
--- a/iroh-relay/src/dns.rs
+++ b/iroh-relay/src/dns.rs
@@ -28,9 +28,8 @@ pub const N0_DNS_NODE_ORIGIN_PROD: &str = "dns.iroh.link";
 /// The n0 testing DNS node origin, for testing.
 pub const N0_DNS_NODE_ORIGIN_STAGING: &str = "staging-dns.iroh.link";
 
-/// Maximum jitter factor for staggered calls. The actual jitter is calculated as
-/// a random number in the range: delay +/- (delay * (((max_jitter_factor) / 10) / 2))
-const MAX_JITTER_FACTOR: u64 = 4;
+/// Percent of total delay to jitter. 0.4 means +/- 20% of delay.
+const MAX_JITTER_PERCENT: u64 = 20;
 
 /// Potential errors related to dns.
 #[common_fields({
@@ -82,14 +81,6 @@ impl<E: std::fmt::Debug + std::fmt::Display> StaggeredError<E> {
             span_trace: n0_snafu::SpanTrace::generate(),
         }
     }
-}
-
-/// Error returned in the remote chance that the jitter calculation produces an int overflow.
-#[allow(missing_docs)]
-#[derive(Debug, Snafu)]
-enum JitterCalcError {
-    #[snafu(display("Either delay or MAX_JITTER_FACTOR is too large"))]
-    Overflow,
 }
 
 /// The DNS resolver used throughout `iroh`.
@@ -515,8 +506,7 @@ async fn stagger_call<
     // NOTE: we add the 0 delay here to have a uniform set of futures. This is more performant than
     // using alternatives that allow futures of different types.
     for delay in std::iter::once(&0u64).chain(delays_ms) {
-        let jittered_delay = add_jitter(delay).unwrap_or(*delay);
-        let delay = Duration::from_millis(jittered_delay);
+        let delay = add_jitter(delay);
         let fut = f();
         let staggered_fut = async move {
             time::sleep(delay).await;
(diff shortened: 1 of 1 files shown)


iroh-relay/src/dns.rs
```diff
@@ -548,4 +574,35 @@ pub(crate) mod tests {
...
+    #[tokio::test]
+    #[traced_test]
+    async fn jitter_test_nonzero_positive() {
+        let delay: u64 = 300;
+        for _ in 0..1000000{
+            assert!(jitter_test_backend(delay).await.unwrap() >= delay * 8 / 10);
+        }
+    }
+
+    #[tokio::test]
```
flub: You don't need an async test at all, use `#[test]` instead of `#[tokio::test]` and call `add_jitter()` directly in the tests and then you don't need the `jitter_test_backend` indirection.

iroh-relay/src/dns.rs
```diff
@@ -523,6 +536,19 @@ async fn stagger_call<
...
+        .checked_mul(MAX_JITTER_FACTOR).ok_or(JitterCalcError::Overflow)?
+        / 10;
+    let jitter = if max_jitter == 0 {
+        0
+    } else {
+        rand::random::<u64>() % max_jitter
+    };
+    Ok((delay.checked_mul(8).ok_or(JitterCalcError::Overflow)?
+        / 10).checked_add(jitter).ok_or(JitterCalcError::Overflow)?)
+}
```
flub, line 540: I must admit I don't follow the jitter calculation.  What bounds does the jitter have?  Is it a percentage of the total delay or does it operate differently?

I'm also wondering if it would not be possible to compute the jitter without having to error out.  Maybe you can do the computation on a `Duration` directly? (I have no idea why the original function doesn't take Duratoins)  Or maybe that doesn't help?

iroh-relay/src/dns.rs
julianpryde: Will fix in the edit

iroh-relay/src/dns.rs
julianpryde: The jitter is supposed to be a random number within +/- 20% of the original delay value.  I'll make that clearer in the edit.

The `Duration` struct also returns an `Option<>` on `checked_*` functions, so we would still have to handle the error. However, if it's safe to assume nobody will use a delay large enough to interact with `Duration::MAX`, it's probably safe to just use a `saturating_*`.

flub: I think `saturating_*` would be reasonable if calculated the right way.

julianpryde:
- [x] Updated PR title to match format
- [x] Made tests synchronous
- [x] Updated the jitter calculation logic
    - It seems like the _truly_ right way to calculate jitter and account for bounds errors is to create a `TimeoutDelay` struct in n0_computer/time.rs that has bounds and errors for if those bounds are exceeded, but that seems excessive at this point.
- [x] Ensured `clippy check` will pass

iroh-relay/src/dns.rs
```diff
@@ -28,6 +28,9 @@ pub const N0_DNS_NODE_ORIGIN_PROD: &str = "dns.iroh.link";
 /// The n0 testing DNS node origin, for testing.
 pub const N0_DNS_NODE_ORIGIN_STAGING: &str = "staging-dns.iroh.link";
 
+/// Percent of total delay to jitter. 0.4 means +/- 20% of delay.
+const MAX_JITTER_PERCENT: u64 = 20;
```
flub, line 32: According to the doc comment this is basically +/- 100% ?  Looking at the code I think you meant to say `20 means +/- 10% of delay`?

iroh-relay/src/dns.rs
```diff
@@ -548,4 +566,31 @@ pub(crate) mod tests {
         let result = stagger_call(f, &delays).await.unwrap();
         assert_eq!(result, 5)
     }
+
+    #[tokio::test]
```
flub: This also doesn't use any async?  Just `#[test]` should do.

iroh-relay/src/dns.rs
```diff
@@ -548,4 +566,31 @@ pub(crate) mod tests {
...
+        let jittered_delay = add_jitter(&0);
+        assert_eq!(jittered_delay, Duration::from_secs(0));
+    }
+
+    //Sanity checks that I did the math right
+    #[test]
+    #[traced_test]
+    fn jitter_test_nonzero_positive() {
+        let delay: u64 = 300;
+        for _ in 0..1000000{
```
flub: How long does this test take?  I understand that you needed some confidence, but to run each time this might be a bit much.  I guess this is a usecase for proptest, but also appreciate that is a lot for this simple change.  Maybe run it with something like 10 or 100?

iroh-relay/src/dns.rs
```diff
@@ -548,4 +566,31 @@ pub(crate) mod tests {
...
+    #[traced_test]
+    async fn jitter_test_zero() {
+        let jittered_delay = add_jitter(&0);
+        assert_eq!(jittered_delay, Duration::from_secs(0));
+    }
+
+    //Sanity checks that I did the math right
+    #[test]
+    #[traced_test]
+    fn jitter_test_nonzero_positive() {
```
flub: Probably nitpicking.  But the naming of `_negative` & `_positive` confused me.  Maybe `_lower_bound` and `_upper_bound` are better?

flub:
The new way of calculating is a lot easier to follow!  Thanks for these changes.

flub:
Could you please also fix the formatting issues?  The cargo deny and was32 CI failures are not due to this PR and we'll fix them outside of this.

iroh-relay/src/dns.rs
julianpryde: Yep, I definitely forgot to change the comment. Thanks for catching that!

iroh-relay/src/dns.rs
julianpryde: I added some zeros because I was paranoid I had done the math wrong and it didn't add any noticeable delay when I ran the tests.  I can lower it though to save CI/CD compute

julianpryde:
- [x] Updated comment on MAX_JITTER_PERCENT definition to make sense
- [x] Made `jitter_test_zero` a synchronous test
- [x] Reduced number of iterations of jitter bounds tests
- [x] Renamed jitter bounds tests
- [x] ran `cargo fmt`

iroh-relay/src/dns.rs
```diff
@@ -28,6 +28,9 @@ pub const N0_DNS_NODE_ORIGIN_PROD: &str = "dns.iroh.link";
 /// The n0 testing DNS node origin, for testing.
 pub const N0_DNS_NODE_ORIGIN_STAGING: &str = "staging-dns.iroh.link";
 
+/// Percent of total delay to jitter. 20 means +/- 20% of delay.
+const MAX_JITTER_PERCENT: u64 = 20;
```
flub, line 32: Last remaining comment, now I understand this value: Why did you choose +/- 20%?  Isn't that rather big?  I think I might have chosen +/- 10%.

julianpryde: I was actually thinking 20% was relatively conservative. Looking at [this](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/) article from aws, they found success with jittering a random value between 0 and the normal delay.

I know the assumptions are probably different between iroh's use case and theirs, so I cam run some tests in a lab environment and see what the impact is to dns server load and request response time.

flub: Cool, thanks for the context.  We do retries eagerly as well to deal with slow responses, so we don't quite want to make the jittering between [0..=delay].  I'm happy to give this 20% a try and see what happens.

I suspect in any case that this isn't going to solve the entire issue.  As I said in #3017 we need to collect all callsites and think about it with a global view.

PR #3448, opened 2025-08-31
feat(net_report)!: extend probes for NAT detection

Arqu:
## Description

Added 4 new fields to net report:
  - `mapping_varies_by_dest_port_ipv4/ipv6` - Detects if NAT assigns different public ports for different destination ports
  - ~~`hairpinning_ipv4/ipv6` - Detects if devices can connect to their own external address~~

`iroh-relay` QUIC server now automatically listens on both main port and port + 1

## Breaking Changes

Everything should be backwards compatible. We do extend the API surface a little bit on the net_report.

## Notes & open questions

Im unsure if the hairpinning test should run in a loop in case we need to keep ourselves covered on network change or does that re-trigger the probe set completely?

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-relay/src/quic.rs
+++ b/iroh-relay/src/quic.rs
@@ -17,7 +17,6 @@ pub const QUIC_ADDR_DISC_CLOSE_REASON: &[u8] = b"finished";
 
 #[cfg(feature = "server")]
 pub(crate) mod server {
-    use std::net::{SocketAddrV4, SocketAddrV6};
 
     use quinn::{
         ApplicationClose, ConnectionError,
@@ -120,28 +119,17 @@ pub(crate) mod server {
                 .context(EndpointServerSnafu)?;
             let bind_addr = endpoint.local_addr().context(LocalAddrSnafu)?;
 
-            // Spawn port variation server (port + 1) for NAT testing
-            let port_variation_addr = match quic_config.bind_addr {
-                SocketAddr::V4(addr) => {
-                    SocketAddr::V4(SocketAddrV4::new(*addr.ip(), addr.port() + 1))
-                }
-                SocketAddr::V6(addr) => {
-                    SocketAddr::V6(SocketAddrV6::new(*addr.ip(), addr.port() + 1, 0, 0))
-                }
-            };
-
+            // Spawn port variation server for NAT testing based on configuration
             let port_variation_addr = quic_config
                 .alternate_port
                 .get_bind_addr(quic_config.bind_addr);
             let port_variation_endpoint = match port_variation_addr {
                 None => None,
                 Some(addr) => {
-                    let endpoint = quinn::Endpoint::server(server_config.clone(), addr)
+                    let endpoint = quinn::Endpoint::server(server_config, addr)
                         .context(EndpointServerSnafu)?;
-                    info!(
-                        ?port_variation_addr,
-                        "QUIC port variation server listening on"
-                    );
+                    let actual_addr = endpoint.local_addr().context(LocalAddrSnafu)?;
+                    info!(?actual_addr, "QUIC port variation server listening on");
                     Some(endpoint)
                 }
             };
@@ -160,7 +148,7 @@ pub(crate) mod server {
                             accept_loop(endpoint, cancel).await
                         }
                     }
-                    .instrument(info_span!("quic-endpoint-port-var"));
+                    .instrument(info_span!("quic-endpoint-alt-port"));
                     let _ = tokio::join!(main_fut, port_variation_fut);
                 }
             });
@@ -186,7 +174,10 @@ pub(crate) mod server {
 
     async fn accept_loop(endpoint: quinn::Endpoint, cancel_accept_loop: CancellationToken) {
         let mut set = JoinSet::new();
-        debug!("waiting for connections on main port...");
+        let local_addr = endpoint
+            .local_addr()
+            .unwrap_or_else(|_| "unknown".parse().unwrap());
+        debug!("waiting for connections on {local_addr}...");
         loop {
             tokio::select! {
                 biased;
@@ -204,7 +195,7 @@ pub(crate) mod server {
                 }
                 res = endpoint.accept() => match res {
                     Some(conn) => {
-                         debug!("accepting connection on main port");
+                         debug!("accepting connection on {local_addr}");
                          let remote_addr = conn.remote_address();
                          set.spawn(async move {
                              if let Err(err) = handle_connection(conn).await {
@@ -213,7 +204,7 @@ pub(crate) mod server {
                          }.instrument(info_span!("qad-conn", %remote_addr)));
                     }
                     None => {
-                        debug!("main endpoint closed");
+                        debug!("endpoint closed");
                         break;
                     }
                 }
--- a/iroh/src/net_report.rs
+++ b/iroh/src/net_report.rs
@@ -461,11 +461,9 @@ impl Client {
                         .instrument(info_span!("QAD-IPv4", %relay_url)),
                 );
 
-                // Port variation probe (QUIC port + 1)
-                if let Ok(relay_addr) =
-                    reportgen::get_relay_addr_ipv4(&dns_resolver, &relay_node).await
-                {
-                    let port_variation = relay_addr.port() + 1;
+                // Port variation probe
+                if let Some(quic_config) = &relay_node.quic {
+                    let alternate_port = quic_config.port + 1;
                     v4_buf.spawn(
                         cancel_v4
                             .child_token()
@@ -476,11 +474,11 @@ impl Client {
                                     relay_node,
                                     quic_client,
                                     dns_resolver,
-                                    Some(port_variation),
+                                    Some(alternate_port),
                                 ),
                             ))
                             .instrument(
-                                info_span!("QAD-IPv4-PortVar", %relay_url, port = port_variation),
+                                info_span!("QAD-IPv4-PortVar", %relay_url, port = alternate_port),
                             ),
                     );
                 }
@@ -511,11 +509,9 @@ impl Client {
                         .instrument(info_span!("QAD-IPv6", %relay_url)),
                 );
 
-                // Port variation probe (QUIC port + 1)
-                if let Ok(relay_addr) =
-                    reportgen::get_relay_addr_ipv6(&dns_resolver, &relay_node).await
-                {
-                    let port_variation = relay_addr.port() + 1;
+                // Port variation probe
+                if let Some(quic_config) = &relay_node.quic {
+                    let alternate_port = quic_config.port + 1;
                     v6_buf.spawn(
                         cancel_v6
                             .child_token()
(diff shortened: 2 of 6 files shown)


iroh-relay/src/quic.rs
```diff
@@ -113,19 +113,57 @@ pub(crate) mod server {
...
                 .send_observed_address_reports(true);
 
-            let endpoint = quinn::Endpoint::server(server_config, quic_config.bind_addr)
+            // Spawn main QUIC server
+            let endpoint = quinn::Endpoint::server(server_config.clone(), quic_config.bind_addr)
                 .context(EndpointServerSnafu)?;
             let bind_addr = endpoint.local_addr().context(LocalAddrSnafu)?;
 
+            // Spawn port variation server (port + 1) for NAT testing
+            let port_variation_addr = match quic_config.bind_addr {
```
matheus23: Looks like this is `Some` in all cases, so I don't think it needs to be wrapped in `Option`.
Unless you wanted to make it fallible, e.g. when port + 1 fails to fall into the u16 range?

iroh/src/net_report.rs
```diff
@@ -728,6 +808,8 @@ async fn run_probe_v4(
                         node: node.clone(),
                         addr,
                         latency,
+                        dest_port: relay_addr.port(),
+                        hairpinning_works: None, // TODO: Implement hairpinning test
```
matheus23: Stray todo?

iroh/src/net_report.rs
```diff
@@ -795,6 +893,8 @@ async fn run_probe_v6(
...
```
matheus23: Stray todo?

flub:
What's the motivation for this?  Are their any relevant iroh bugs or discussions I've missed?

matheus23:
It seems related to https://github.com/n0-computer/iroh-n0des/pull/37, which in turn is related to hole-punching debugging tooling/features mentioned here: https://www.notion.so/number-zero/n0des-rolling-sync-up-2545df1306fb807eb36bda59dd971e92#2545df1306fb8064a19bc8870ec9b1fb

flub:
What I find odd (admittedly without looking into this) is that mapping-varies-by-dest should already exist in iroh?  And I have some recollection of hairpinning being recently removed (/cc @dignifiedquire ).  So I'm confused why these things are/need to be added here.  And hence I'm asking for some more context.

Arqu:
> What I find odd (admittedly without looking into this) is that mapping-varies-by-dest should already exist in iroh? And I have some recollection of hairpinning being recently removed (/cc @dignifiedquire ). So I'm confused why these things are/need to be added here. And hence I'm asking for some more context.

You're right. Mapping-varies-by-dest exists, but we also need mapping-varies-by-port to get the full spectrum. 
You're also right about us removing hairpinning which sent me into a nice rabbit hole of what is required or not for different use cases. Pretty sure we don't need hairpinning detection. It either works or you should be using local discovery anyways so that is no longer really important.

The reasoning behind this is to extend the doctor report with some more details and above all to provide us with some more details about our NAT. This feeds into the final result where we can use this info to generate a network fingerprint for users and also just nicely present their NAT state.

iroh-relay/src/quic.rs
Arqu: Just a leftover from some initial work here, definitely should be dropped.

flub:
> > What I find odd (admittedly without looking into this) is that mapping-varies-by-dest should already exist in iroh? And I have some recollection of hairpinning being recently removed (/cc @dignifiedquire ). So I'm confused why these things are/need to be added here. And hence I'm asking for some more context.
> 
> You're right. Mapping-varies-by-dest exists, but we also need mapping-varies-by-port to get the full spectrum.

I guess it's still not clear to me why we want to name a NAT type that we're not actually bothering identifying in iroh.  And if this code lives in iroh it's doubly weird because then it only exists for the public API, but we've been treating this API as internal.

Now in general it's probably not bad to have a tool that can detect more NAT types.  E.g. once we start trying to support some Destination Endpoint Dependent NAT types it will be useful to collect information about *how* the mapping varies.  But that will have to be even more custom because we'd want to find out if there are patterns in ports and IPs chosen by the mapping.  But that's like quite experimental, and I'd prefer if that code could live just inside of the doctor.

> The reasoning behind this is to extend the doctor report with some more details and above all to provide us with some more details about our NAT. This feeds into the final result where we can use this info to generate a network fingerprint for users and also just nicely present their NAT state.

The current design of the endpoint providing a report is with the explicit intention that you get *exactly* what iroh uses internally, so we can debug why iroh fails for some users.  Adding extra information into the report coming from iroh seems the wrong approach to providing something extra, because it breaks that 1:1 mapping.  So if we do want to provide extra information on top of what iroh itself uses, perhaps it should live outside of iroh?

Arqu:
> > > What I find odd (admittedly without looking into this) is that mapping-varies-by-dest should already exist in iroh? And I have some recollection of hairpinning being recently removed (/cc @dignifiedquire ). So I'm confused why these things are/need to be added here. And hence I'm asking for some more context.
> > 
> > 
> > You're right. Mapping-varies-by-dest exists, but we also need mapping-varies-by-port to get the full spectrum.
> 
> I guess it's still not clear to me why we want to name a NAT type that we're not actually bothering identifying in iroh. And if this code lives in iroh it's doubly weird because then it only exists for the public API, but we've been treating this API as internal.

This only extends the existing net report with 2 new fields. Nothing else really changes API wise. None of the NAT classification code is actually in here. 
The changes in the PR are: 
- added mapping_varies_by_port_[ipv4/ipv6] to the report
- extended probing so that we can actually capture that info
- extended relay to accept probes on another port for it.

It also is public API, at least the report, you want users to be able to introspect their network params, our most prominant use case for it currently is `iroh-doctor report`

> Now in general it's probably not bad to have a tool that can detect more NAT types. E.g. once we start trying to support some Destination Endpoint Dependent NAT types it will be useful to collect information about **_how_ the mapping varies**. But that will have to be even more custom because we'd want to find out if there are patterns in ports and IPs chosen by the mapping. But that's like quite experimental, and I'd prefer if that code could live just inside of the doctor.
> 

How the mapping varies is _exactly_ what we're capturing here. If we later need to extend with even more granular details we can follow up from here. Agree we can move some of the fancier exploration logic for port preferences etc up into doctor.

> > The reasoning behind this is to extend the doctor report with some more details and above all to provide us with some more details about our NAT. This feeds into the final result where we can use this info to generate a network fingerprint for users and also just nicely present their NAT state.
> 
> The current design of the endpoint providing a report is with the explicit intention that you get _exactly_ what iroh uses internally, so we can debug why iroh fails for some users. Adding extra information into the report coming from iroh seems the wrong approach to providing something extra, because it breaks that 1:1 mapping. So if we do want to provide extra information on top of what iroh itself uses, perhaps it should live outside of iroh?

IMHO this is not that of an extreme addition that's outside of the scope of iroh. It's also very specific to the given endpoint and definitely part of the same stack.

Cargo.lock
```diff
@@ -2766,11 +2766,11 @@ dependencies = [
 
 [[package]]
 name = "matchers"
```
dignifiedquire, line 2768: are these updates needed?

iroh-relay/src/quic.rs
```diff
@@ -113,19 +115,38 @@ pub(crate) mod server {
...
+                SocketAddr::V4(addr) => {
+                    SocketAddr::V4(SocketAddrV4::new(*addr.ip(), addr.port() + 1))
+                }
+                SocketAddr::V6(addr) => {
+                    SocketAddr::V6(SocketAddrV6::new(*addr.ip(), addr.port() + 1, 0, 0))
+                }
+            };
+
+            let port_variation_endpoint =
+                quinn::Endpoint::server(server_config.clone(), port_variation_addr)
```
dignifiedquire: no clone needed on the last usage

iroh-relay/src/quic.rs
```diff
@@ -113,19 +115,38 @@ pub(crate) mod server {
...
 
             let cancel = CancellationToken::new();
             let cancel_accept_loop = cancel.clone();
 
-            let task = tokio::task::spawn(
+            let main_task = tokio::task::spawn(
                 async move {
                     let mut set = JoinSet::new();
-                    debug!("waiting for connections...");
+                    debug!("waiting for connections on main port...");
```
dignifiedquire: maybe print the ports, instead of saying just main port?

iroh-relay/src/quic.rs
```diff
@@ -143,13 +164,16 @@ pub(crate) mod server {
...
-                                     );                                }
+                                     set.spawn(async move {
+                                         if let Err(err) = handle_connection(conn).await {
+                                             debug!("connection error: {err:#?}");
+                                         }
+                                     }.instrument(info_span!("qad-conn", %remote_addr)));
+                                }
                                 None => {
-                                    debug!("endpoint closed");
+                                    debug!("main endpoint closed");
```
dignifiedquire: all these `main` additions shouldn't be needed if you add an `instrument` on the task

iroh/src/net_report.rs
```diff
@@ -443,15 +443,47 @@ impl Client {
...
+                            ),
                         ))
                         .instrument(info_span!("QAD-IPv4", %relay_url)),
                 );
+
+                // Port variation probe (QUIC port + 1)
+                if let Ok(relay_addr) =
+                    reportgen::get_relay_addr_ipv4(&dns_resolver, &relay_node).await
+                {
+                    let port_variation = relay_addr.port() + 1;
```
dignifiedquire: I think assuming that this is always port + 1 is potentially problematic, I think we should make this somewhat configurable if we really need this

dignifiedquire:
I feel pretty strongly that these probes should not be run, unless explicitly triggered for the doctor, this increases the amount of traffic without any benefit on regular iroh operations

flub:
>     * extended relay to accept probes on another port for it.

So you are making relays listen on two ports?  Why is that needed?  To determine if you're behind a Destination Endpoint Dependent NAT you contact two different QAD servers and compare the results.  But each QAD server only needs to listen on one socket address itself.

iroh-relay/src/quic.rs
Arqu: they already are so I just dropped the labels.

Arqu:
> > ```
> > * extended relay to accept probes on another port for it.
> > ```
> 
> So you are making relays listen on two ports? Why is that needed? To determine if you're behind a Destination Endpoint Dependent NAT you contact two different QAD servers and compare the results. But each QAD server only needs to listen on one socket address itself.

Not really, this is to determine if you're behind a Destination Port Dependant NAT. Which is about the fact that your port changes depending on their ip _and_ port combination not just ip. Which makes holepunching much harder and is used as part of the classification logic for NATs downstream.

Arqu:
I've just pushed an update that cleans it up a bit and fixes tests. Still need to feature gate the port variation probes.

Frando:
> Which is about the fact that your port changes depending on their ip and port combination not just ip. Which makes holepunching much harder and is used as part of the classification logic for NATs downstream.

I'm still struggling to understand how this makes holepunching harder in iroh's case. If peer A is behind a destination dependent NAT mapping, it is already the case that the port returned from Stun/QAD cannot be used by other peers to connect. So if both peers are behind such NATs, holepunching fails (unless we would do birthday guessing stuff, which we aren't). Or am I missing a case where this would succeed, but would fail if one of the peers were behind a port dependent NAT?

Arqu:
All of this functionality is being moved into https://github.com/n0-computer/iroh-doctor/pull/48

PR #3458, opened 2025-09-17
chore: add `test-utils` cfg for `insecure_cert` call in the transfer example

ramfox:
also, upgrade netwatch, portmapper, and n0-snafu deps

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2271,7 +2271,6 @@ name = "iroh"
 version = "0.91.2"
 dependencies = [
  "aead",
- "anyhow",
  "axum",
  "backon",
  "bytes",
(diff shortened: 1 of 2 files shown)


iroh/examples/transfer.rs
```diff
@@ -192,6 +194,8 @@ impl EndpointArgs {
             }
         };
         builder = builder.secret_key(secret_key);
+
+        #[cfg(feature = "test-utils")]
```
flub: This is going to very quietly fail if you didn't have the feature enabled.  Does it make sense to have a `#[cfg(not(feature = "test-utils"))]` version that aborts if `self.env == Env::Dev`?

flub (approved):
I don't particularly want to block this, so approving.  Though I think you can make this a little bit more user friendly with very little change.

PR #3460, opened 2025-09-19
feat: Make fmt_short return an impl Display so we can avoid an allocation.

rklaehn:
## Description

Currently, fmt_short allocates a String. It is used quite a few times in trace! statements. This makes it not allocate.

Not sure if it is worth it.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-base/src/key.rs
+++ b/iroh-base/src/key.rs
@@ -134,12 +134,20 @@ impl PublicKey {
 
     /// Convert to a hex string limited to the first 5 bytes for a friendly string
     /// representation of the key.
-    pub fn fmt_short(&self) -> String {
-        data_encoding::HEXLOWER.encode(&self.as_bytes()[..5])
+    pub fn fmt_short(&self) -> impl Display + 'static {
+        PublicKeyShort(*self)
     }
 
     /// The length of an ed25519 `PublicKey`, in bytes.
     pub const LENGTH: usize = ed25519_dalek::PUBLIC_KEY_LENGTH;
+}
+
+struct PublicKeyShort(PublicKey);
+
+impl Display for PublicKeyShort {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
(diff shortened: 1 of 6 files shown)


iroh-base/src/key.rs
```diff
@@ -134,14 +134,22 @@ impl PublicKey {
...
-        data_encoding::HEXLOWER.encode(&self.as_bytes()[..5])
+    pub fn fmt_short(&self) -> impl Display + 'static {
+        PublicKeyShort(*self)
     }
 
     /// The length of an ed25519 `PublicKey`, in bytes.
     pub const LENGTH: usize = ed25519_dalek::PUBLIC_KEY_LENGTH;
 }
 
+struct PublicKeyShort(PublicKey);
```
flub: The old code used space for 10 bytes on the heap.  This version uses 32 bytes on the stack?

I think it's a great idea to try and get rid of the heap allocation.  But maybe you can give `PublicKeyShort` only the 5 bytes needed for the display?

rklaehn: Yes, you could do that easily. PublicKeyShort is not part of the public API, so you can optimize it.

iroh-base/src/key.rs
```diff
@@ -134,14 +134,22 @@ impl PublicKey {
 
     /// Convert to a hex string limited to the first 5 bytes for a friendly string
     /// representation of the key.
-    pub fn fmt_short(&self) -> String {
-        data_encoding::HEXLOWER.encode(&self.as_bytes()[..5])
+    pub fn fmt_short(&self) -> impl Display + 'static {
+        PublicKeyShort(self.0.as_bytes()[0..5].try_into().unwrap())
```
flub: Change this to an expect explaining why this is ok.  We tried to remove all unwrap calls at some point (I'm surprised there's no lint).

flub (approved):
Thanks!  I think the optimisation of only copying 5 bytes is worth it here.

PR #3466, opened 2025-09-26
chore: bump some spans up to warn, to ensure they are logged

dignifiedquire:
## Description

We were missing this information in some logs

## Breaking Changes

None

## Notes & open questions

Unclear if we should normalize differently, but this fixes the concrete issue for now.


iroh-relay/src/server/http_server.rs
```diff
@@ -539,7 +539,7 @@ impl RelayService {
                     Err(err) => warn!("upgrade error: {err:#}"),
                 }
             }
-            .instrument(debug_span!("handler"))
+            .instrument(warn_span!("handler"))
```
flub, line 542: Not that this does any harm, but a span without fields is perhaps not that interesting?

flub (approved):

iroh-relay/src/server/http_server.rs
dignifiedquire: potentially, but we have quite a few of these currently

PR #3467, opened 2025-09-26
feat(iroh)!: introduce `online` method

dignifiedquire:
## Description

A new attempt to start improving the address and boot up situation.

There is a new `online` method, which simply waits for the `NodeAddr` to be filled, usage is easiest to see by looking at how the examples have changed.

## Breaking Changes

- removed
  -  `iroh::Endpoint::direct_addresses`
  - `iroh::Endpoint::home_relay`
- changed
  - `iroh::Endpoint::node_addr` now returns synchronously the current addressing information
- added
  - `iroh::Endpoint::online`
  - `iroh::Endpoint::watch_node_addr`

### DIFF
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -13,7 +13,6 @@
 
 use std::{
     any::Any,
-    collections::BTreeSet,
     future::{Future, IntoFuture},
     net::{IpAddr, SocketAddr, SocketAddrV4, SocketAddrV6},
     pin::Pin,
@@ -22,7 +21,7 @@ use std::{
 };
 
 use ed25519_dalek::{VerifyingKey, pkcs8::DecodePublicKey};
-use iroh_base::{NodeAddr, NodeId, RelayUrl, SecretKey};
+use iroh_base::{NodeAddr, NodeId, SecretKey};
 use iroh_relay::RelayMap;
 use n0_future::{Stream, time::Duration};
 use n0_watcher::Watcher;
@@ -903,6 +902,26 @@ impl Endpoint {
         self.static_config.tls_config.secret_key.public()
     }
 
+    /// Returns the current [`NodeAddr`].
+    pub fn node_addr(&self) -> NodeAddr {
+        let node_id = self.node_id();
+        let relay_url = self.msock.home_relay().get().first().cloned();
+        let direct_addresses = self
+            .msock
+            .direct_addresses()
+            .get()
+            .unwrap_or_default()
+            .into_iter()
+            .map(|addr| addr.addr)
+            .collect();
+
+        NodeAddr {
+            node_id,
+            relay_url,
+            direct_addresses,
+        }
+    }
+
     /// Returns a [`Watcher`] for the current [`NodeAddr`] for this endpoint.
     ///
     /// The observed [`NodeAddr`] will have the current [`RelayUrl`] and direct addresses
@@ -918,15 +937,15 @@ impl Endpoint {
     ///     .alpns(vec![b"my-alpn".to_vec()])
     ///     .bind()
     ///     .await?;
-    /// let node_addr = endpoint.node_addr().initialized().await;
+    /// let node_addr = endpoint.watch_node_addr().initialized().await;
     /// # let _ = node_addr;
     /// # Ok(())
     /// # }
     /// ```
     #[cfg(not(wasm_browser))]
-    pub fn node_addr(&self) -> impl n0_watcher::Watcher<Value = Option<NodeAddr>> + use<> {
-        let watch_addrs = self.direct_addresses();
-        let watch_relay = self.home_relay();
+    pub fn watch_node_addr(&self) -> impl n0_watcher::Watcher<Value = Option<NodeAddr>> + use<> {
+        let watch_addrs = self.msock.direct_addresses();
+        let watch_relay = self.msock.home_relay();
         let node_id = self.node_id();
 
         watch_addrs
@@ -950,7 +969,7 @@ impl Endpoint {
     /// with a [`NodeAddr`] that only contains a relay URL, but no direct addresses,
     /// as there are no APIs for directly using sockets in browsers.
     #[cfg(wasm_browser)]
-    pub fn node_addr(&self) -> impl n0_watcher::Watcher<Value = Option<NodeAddr>> + use<> {
+    pub fn watch_node_addr(&self) -> impl n0_watcher::Watcher<Value = Option<NodeAddr>> + use<> {
         // In browsers, there will never be any direct addresses, so we wait
         // for the home relay instead. This makes the `NodeAddr` have *some* way
         // of connecting to us.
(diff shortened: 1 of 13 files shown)


iroh/src/endpoint.rs
```diff
@@ -903,6 +902,26 @@ impl Endpoint {
         self.static_config.tls_config.secret_key.public()
     }
 
+    /// Returns the current [`NodeAddr`].
```
flub, line 876: Needs more docs referring to how it can be empty and `.online()` and `.watch_node_addr()`.

iroh/src/endpoint.rs
```diff
@@ -965,69 +984,16 @@ impl Endpoint {
...
-    /// let ep = Endpoint::builder().bind().await.unwrap();
-    /// let _addrs = ep.direct_addresses().initialized().await;
-    /// # });
-    /// ```
-    ///
-    /// [QAD]: https://www.ietf.org/archive/id/draft-ietf-quic-address-discovery-00.html
-    pub fn direct_addresses(&self) -> n0_watcher::Direct<Option<BTreeSet<DirectAddr>>> {
-        self.msock.direct_addresses()
+    /// This has no timeout, so if that is needed, you need to wrap it in a timeout.
+    pub async fn online(&self) {
```
flub, line 1018: Should this handle going offline again?  Like when the home relay is no longer reachable it should be pending again until there's another home relay?

The docs should probably also elaborate on when and how you might want to use the endpoint without being online, like mDNS or more manual scenarios.

iroh/src/endpoint.rs
```diff
@@ -965,69 +984,16 @@ impl Endpoint {
...
+        let mut watch1 = self.msock.home_relay();
+        let mut watch2 = self.msock.direct_addresses();
+        tokio::join!(watch1.initialized(), watch2.initialized());
```
Frando, line 1002: I think this needs a cfg for wasm_browser to only wait on the relay (direct_addresses() never completes in the browser IIRC)

iroh/src/endpoint.rs
```diff
@@ -965,69 +984,16 @@ impl Endpoint {
...
+    /// Waits for the node to be considered online.
     ///
-    /// The direct addresses of the [`Endpoint`] are those that could be used by other
-    /// iroh nodes to establish direct connectivity, depending on the network
-    /// situation. The yielded lists of direct addresses contain both the locally-bound
-    /// addresses and the [`Endpoint`]'s publicly reachable addresses discovered through
-    /// mechanisms such as [QAD] and port mapping.  Hence usually only a subset of these
-    /// will be applicable to a certain remote iroh node.
+    /// This currently means at least one relay server was connected,
+    /// and at least one local IP address is available.
```
Frando, line 988: What about endpoints that don't have relays configured? Right now this would never resolve.
I think we should take the relay map into account, and if its empty (relays disabled) it should only wait for the local addrs.

iroh/src/endpoint.rs
dignifiedquire: > Should this handle going offline again?

No, that would be very complex to do, will make it clear in the docs.

iroh/src/endpoint.rs
dignifiedquire: Hmm, interesting idea, yes that does make sense

flub: Or perhaps it should be listed as something to watch out for?  Because you would no longer know you're online or not, so you're returning a bogus answer.  It could also log a warning if called with no relays.

iroh/src/endpoint.rs
flub: Is it that complex?  You only need a watcher on the home relay to implement this I think.  If it's None: block.  If it's Some(): return.

iroh/src/endpoint.rs
b5: one thought here: I like that `online` implies / means / requires a WAN. IMHO calling `online` when the endpoint has no relays configured that should be an error? 

that really muddies things with `.online()?` now needing to return a result. Maybe a compromise is to log a warning

iroh/src/endpoint.rs
dignifiedquire: I would need to store boxed futures, and make a new type, and stuff and manually poll things..

iroh/src/endpoint.rs
dignifiedquire: hmm, I like the current way, but logging something might make sense, not sure given I have now 3 different suggestions I think

flub: To be fair, if you don't have a relay connection you simply don't know if you have a WAN or not.  So you have no idea if you're online or not.

ramfox:
Personal opinion - I think the `online` function only makes sense in the context of knowing you are accessible to the outside internet, and so should only return when we have a home relay available.

So my preference would be that this hangs forever if there is no relay available, even if there is no relay in the relay map. This will also make sense when we allow users to allow adding a relay dynamically (after the endpoint is created). We hang until a relay is added and we make a connection to that relay, and then `online` would return.

The question of whether this should be more like a watchable, that indicates going back offline...I can see someone wanting this or expecting this in the future, but I don't want to get farther away from what this method is supposed to be, which is porcelain for waiting for both the `home_relay` and `direct_addresses`. I'd rather solve this by documentation now (eg, "watch the `watch_node_addr` stream for any changes in conductivity"), and wait for someone to ask us for this.

iroh/src/endpoint.rs
```diff
@@ -884,15 +925,25 @@ impl Endpoint {
...
+    /// of the private or local network, watch for changes in it's [`NodeAddr`].
+    /// If the `relay_url` is `None` or if there are no `direct_addresses` in
+    /// the [`NodeAddr`], you may not be dialable by other endpoints on the internet.
+    ///
+    /// [`RelayUrl`]: crate::RelayUrl
     #[cfg(not(wasm_browser))]
-    pub fn node_addr(&self) -> impl n0_watcher::Watcher<Value = Option<NodeAddr>> + use<> {
-        let watch_addrs = self.direct_addresses();
-        let watch_relay = self.home_relay();
+    pub fn watch_node_addr(&self) -> impl n0_watcher::Watcher<Value = Option<NodeAddr>> + use<> {
```
flub, line 944: How consistent is our API with this?  E.g. we have `home_relay` and `direct_addrs` which return watchers.  But for `node_addr` this is no longer the case.

I can think of two things:
1. revert the naming: e.g. `.node_addr() -> impl Watcher` and `.current_node_addr()`.
2. Change the `node_addr` watcher to **always** return a value: `.node_addr() -> impl Watcher<Value = NodeAddr>`.  That is, no longer wrap it in an `Option` and thus you no longer have `.initialized()`.

At the moment I think I prefer 2.

dignifiedquire: The other watchers have been removed though? this is an attempt to make the method names more descriptive

PR #3473, opened 2025-09-29
feat: add a DNS resolver trait

### LINKED ISSUE #3468: Decide about hickory in public API for DnsResolver
Related to #3177 

We need to implement a performant API for DNS resolution that we can own. We currently have a `From<hickory_resolver::TokioResolver>` for `iroh::DnsResolver`. This means hickory_resolver@0.25 would be in our public API, which means updating to a potential 0.26 would be a semver nogo during 1.0. 

We agree we need to address this, but don't know what the API needs to look like yet. There are also open concerns about perf due to Boxing.

We can close this when:
* [ ] We have an API for providing custom DNS resolution that uses only types compatible with a  1.0 stamp of approval
* [ ] That implementation is at least 95% as performant as the current implementation (doesn't have to be the case for custom DNS resolution)

Frando:
## Description

Replaces #2116 
Fixes #3468

This adds a `Resolver` trait to abstract over DNS resolution. It contains methods to resolve IPv4 or IPv6 addresses, and TXT records. Because the trait needs to be dyn-compatible for our usage (we don't want to have a generic for the DNS resolver on the Endpoint), all methods return boxed futures that contain boxed iterators. 

To not pay the cost of boxing when using the default hickory-based resolver, the iroh `DnsResolver` internally contains an enum which is either the default resolver or the boxed custom resolver.

Users can implement the `Resolver` trait on whatever struct to use a completely custom DNS resolver.

## Breaking Changes

Changes in `iroh_relay::dns` (reexported from `iroh::dns`):
*  `DnsResolver::lookup_txt` now returns `impl Iterator<Item = TxtRecord>`
* `TxtLookup` and `TXT` are removed.

Changes in `iroh_relay::node_info` (reexported from `iroh::discovery`):
* `NodeInfo::from_txt_lookup` now takes `domain_name: String, lookup: impl Iterator<Item = crate::dns::TxtRecordData>`

Moved items:
* `iroh_relay::node_info::LookupError` was moved to `iroh_relay::dns::LookupError`

## Notes & open questions

We likely want to still add a builder for our DnsResolver to allow setting a few more of the common options on the hickory resolver. But with this, people that want to use some completely custom way of resolving DNS records can do so.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-relay/src/dns.rs
+++ b/iroh-relay/src/dns.rs
@@ -142,11 +142,8 @@ impl DnsResolver {
     /// This does not work at least on some Androids, therefore we fallback
     /// to the default `ResolverConfig` which uses eg. to google's `8.8.8.8` or `8.8.4.4`.
     pub fn new() -> Self {
-        let resolver = Self::new_inner();
-        Self(Arc::new(RwLock::new(DnsResolverInner::Hickory {
-            resolver,
-            nameserver: None,
-        })))
+        let resolver = HickoryResolver::new(HickoryResolverOpts::SystemDefaults);
+        Self(Arc::new(RwLock::new(DnsResolverInner::Hickory(resolver))))
     }
 
     /// Creates a new [`DnsResolver`] from a struct that implements [`Resolver`].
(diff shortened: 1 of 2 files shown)


dignifiedquire (approved):
looks good to me

PR #3475, opened 2025-09-29
feat: add a builder for DnsResolver

Frando:
## Description

Based on #3473 

In addition to the trait added in #3473 ,this adds a builder to allow setting more than one custom nameserver on the default hickory resolver, and to allow using other protocols than UDP (namely TCP, TLS and HTTPS, which are the ones supported by hickory_proto with our current feature flags).

## Notes & open questions

There's two more protocols [supported by hickory](https://docs.rs/hickory-proto/latest/hickory_proto/xfer/enum.Protocol.html) though behind non-default feature flags: Quic and H3. If there's demand, we could enable those too.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


Frando:
I pushed a commit that has no functional changes, but makes it more obvious in the code that if we fail to load the host system's DNS config, we fallback to `ResolverConfig::google`. This was so far hidden in a `unwrap_or_default` (ResolverConfig::default calls ResolverConfig::google). I think we should have the fact that we fallback to google's DNS servers visible in the source code, so that if we want to change it to eg. cloudflare or something else, it's obvious to spot.

dignifiedquire (approved):

PR #3482, opened 2025-09-30
feat(iroh)!: add `MdnsDiscoveryBuilder::service_name` method

ramfox:
## Description
Changes the default service name to `irohv1` as well as changing the provenance name to simply `mdns`.

Adds an `MdnsDiscoveryBuilder::service_name` method to allow users to use their own service name for their mDNS records.

This commit also narrows down the options to create an `MdnsDiscovery` service. Now you must use `MdnsDiscovery::builder` to create an `MdnsDiscoveryBuilder`, then use the `build()` method to create the `MdnsDiscovery` service.

## Breaking Changes

  - `MdnsDiscovery::new` is now private
  - `MdnsDiscoveryBuilder::new` is now private
      - Use `MdnsDiscovery::builder()` to create an `MdnsDiscoveryBuilder`
  - Default service name changed from `iroh.local.swarm` to `irohv1`
  - Provenance field name in `DiscoveryItem` changed from `local.swarm.discovery` to `mdns`

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/discovery/mdns.rs
+++ b/iroh/src/discovery/mdns.rs
@@ -50,15 +50,15 @@ use tracing::{Instrument, debug, error, info_span, trace, warn};
 use super::{DiscoveryContext, DiscoveryError, IntoDiscovery, IntoDiscoveryError};
 use crate::discovery::{Discovery, DiscoveryEvent, DiscoveryItem, NodeData, NodeInfo};
 
-/// The n0 local swarm node discovery name
-const N0_LOCAL_SWARM: &str = "iroh.local.swarm";
+/// The n0 local service name
+const N0_SERVICE_NAME: &str = "irohv1";
 
 /// Name of this discovery service.
 ///
 /// Used as the `provenance` field in [`DiscoveryItem`]s.
 ///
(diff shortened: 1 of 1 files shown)


iroh/src/discovery/mdns.rs
```diff
@@ -659,7 +688,48 @@ mod tests {
...
+                node_id,
+                MdnsDiscovery::builder()
+                    .advertise(advertise)
+                    .build(node_id)?,
+            ))
+        }
+
+        #[tokio::test]
+        #[traced_test]
+        async fn test_service_names() -> Result {
```
Frando, line 701: Maybe add to the test that resolving on the correct swarm works OK, to be sure that the custom service name works fine that way. (Currently the test only asserts that nodes are not discovered on the other service name swarm)

Frando (approved):

dignifiedquire:
> * Are we down with `irohv1` or is `iroh.v1` prettier 😂 ?

no strong preference from my side

Frando:
I think `irohv1` is good, the domain separation with a dot is more noisy IMO. And `builder` only is fine IMO.

PR #3486, opened 2025-09-30
fix: force reqwest to always use rustls backend

Frando:
## Description

This changes all uses of `reqwest` in iroh to use `ClientBuilder::use_rustls_tls()` to force using the rustls TLS backend, even if the `default-tls` feature got enabled through feature unification (we don't enable it in iroh directly). 

Reasoning:

We have reports of HTTPS net report probes failing. They *do* fail under the following circumstances:

- `reqwest` uses the native-tls backend (which is platform dependent, and is openssl on linux) whenever the `default-tls` feature is enabled. it is on by default. if both `rustls-tls` and `default-tls` features are enabled, the native-tls backend wins.
- in iroh, we use `reqwest` with `default-features = false, features = ["rustls-tls"]`, so without other deps changing features, the rustls backend is used
- the rustls backend apparently cleans up the domain name, removing a final dot if present, before comparing it to the certificate's hostname
- openssl apparently does not do that, so if the passed-in hostname has a final dot, but the certificate has not, an SSL verification error is thrown
- now, if *any* other crate in a binary's deps enables the `default-tls` feature for reqwest, due to feature unification iroh now by default also uses the native tls backend and not rustls anymore

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-relay/src/main.rs
+++ b/iroh-relay/src/main.rs
@@ -237,7 +237,10 @@ impl From<AccessConfig> for iroh_relay::server::AccessConfig {
                 }))
             }
             AccessConfig::Http(mut config) => {
-                let client = reqwest::Client::default();
+                let client = reqwest::Client::builder()
+                    .use_rustls_tls()
+                    .build()
+                    .expect("request client builder");
                 // Allow to set bearer token via environment variable as well.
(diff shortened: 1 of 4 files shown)


dignifiedquire:
looks like this is not available in `wasm`

Frando:
Pushed a commit that cfg's it out for wasm

Issue #3377, opened 2025-07-01
Iroh printing a lot of "Failed to run HTTPS probe"


juchiast:
```
2025-07-01T07:40:10.141611Z  WARN reportgen.actor:run-probe{proto=Https delay=200ms relay_node=RelayNode { url: RelayUrl("https://aps1-1.relay.n0.iroh-canary.iroh.link./"), quic: Some(RelayQuicConfig { port: 7842 }) }}: iroh::net_report::reportgen: probe failed: Failed to run HTTPS probe
2025-07-01T07:40:10.239178Z  WARN reportgen.actor:run-probe{proto=Https delay=300ms relay_node=RelayNode { url: RelayUrl("https://aps1-1.relay.n0.iroh-canary.iroh.link./"), quic: Some(RelayQuicConfig { port: 7842 }) }}: iroh::net_report::reportgen: probe failed: Failed to run HTTPS probe
2025-07-01T07:40:10.339850Z  WARN reportgen.actor:run-probe{proto=Https delay=400ms relay_node=RelayNode { url: RelayUrl("https://aps1-1.relay.n0.iroh-canary.iroh.link./"), quic: Some(RelayQuicConfig { port: 7842 }) }}: iroh::net_report::reportgen: probe failed: Failed to run HTTPS probe
2025-07-01T07:40:10.414925Z  WARN reportgen.actor:run-probe{proto=Https delay=200ms relay_node=RelayNode { url: RelayUrl("https://euc1-1.relay.n0.iroh-canary.iroh.link./"), quic: Some(RelayQuicConfig { port: 7842 }) }}: iroh::net_report::reportgen: probe failed: Failed to run HTTPS probe
2025-07-01T07:40:10.510220Z  WARN reportgen.actor:run-probe{proto=Https delay=300ms relay_node=RelayNode { url: RelayUrl("https://euc1-1.relay.n0.iroh-canary.iroh.link./"), quic: Some(RelayQuicConfig { port: 7842 }) }}: iroh::net_report::reportgen: probe failed: Failed to run HTTPS probe
2025-07-01T07:40:10.592090Z  WARN reportgen.actor:run-probe{proto=Https delay=200ms relay_node=RelayNode { url: RelayUrl("https://use1-1.relay.n0.iroh-canary.iroh.link./"), quic: Some(RelayQuicConfig { port: 7842 }) }}: iroh::net_report::reportgen: probe failed: Failed to run HTTPS probe
[...]

PR #3487, opened 2025-09-30
fix: remove final dot in relay urls when doing https requests

Frando:
## Description

We append a final dot to all relay URL domains, to make sure that during DNS resolution the domain is treated as absolute, and no search suffix is appended. However, in the context of TLS name verification it is more correct to *not* have a final dot, so that the URL's hostname matches the name in the certificate. It seems this is not an issue with rustls, but when using openssl this is an issue (which we had reports for, but those are also fixed by #3486).

This PR adds methods to get the relay URL or hostname without the final dot, and uses those in the context of HTTPS requests.

## Notes & open questions

Should check if there's more uses of those URLs. Likely we want to do the same when actually *connecting* to the relay? But didn't yet find the perfect spot for that.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-base/src/relay_url.rs
+++ b/iroh-base/src/relay_url.rs
@@ -38,6 +38,44 @@ impl From<Url> for RelayUrl {
     }
 }
 
+impl RelayUrl {
+    /// Returns the URL while removing the final dot in the relay URL's domain name.
+    ///
+    /// By default, we add a final dot to relay URLs to make sure that DNS resolution always
+    /// considers them as top-level domains without appending a search suffix. When using
+    /// the URL for TLS hostname verification, usually a domain name without a final
+    /// dot is expected. So when using the URL in the context of HTTPS or TLS, use
+    /// this function to get the URL without the final dot.
+    pub fn without_final_dot(&self) -> Url {
+        let mut url = self.0.deref().clone();
+        if let Some(domain) = url.domain() {
+            if domain.ends_with('.') {
+                let domain_without_dot = domain[..(domain.len() - 1)].to_string();
+                url.set_host(Some(&domain_without_dot)).ok();
+            }
+        }
+        url
+    }
+
+    /// Return the string representation of the host (domain or IP address) for this URL, if any.
+    ///
+    /// If the host is a domain, and the domain ends with a final dot, the final dot is removed.
+    ///
+    /// See [`Self::without_final_dot`] for details on when you might want to use this.
+    pub fn host_str_without_final_dot(&self) -> Option<&str> {
+        if let Some(domain) = self.0.domain() {
+            if domain.ends_with('.') {
+                let domain_without_dot = &domain[..(domain.len() - 1)];
+                Some(domain_without_dot)
+            } else {
+                Some(domain)
+            }
+        } else {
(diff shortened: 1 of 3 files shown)


iroh-base/src/relay_url.rs
```diff
@@ -38,6 +38,44 @@ impl From<Url> for RelayUrl {
...
 
+impl RelayUrl {
+    /// Returns the URL while removing the final dot in the relay URL's domain name.
+    ///
+    /// By default, we add a final dot to relay URLs to make sure that DNS resolution always
+    /// considers them as top-level domains without appending a search suffix. When using
+    /// the URL for TLS hostname verification, usually a domain name without a final
+    /// dot is expected. So when using the URL in the context of HTTPS or TLS, use
+    /// this function to get the URL without the final dot.
+    pub fn without_final_dot(&self) -> Url {
```
dignifiedquire, line 49: can you add a test for this please?

iroh/src/net_report.rs
```diff
@@ -761,7 +764,10 @@ async fn run_probe_v6(
         reportgen::maybe_to_mapped_addr(ip_mapped_addrs.as_ref(), relay_addr_orig.into());
 
     debug!(?relay_addr_orig, ?relay_addr, "relay addr v6");
-    let host = relay_node.url.host_str().context("missing host url")?;
+    let host = relay_node
+        .url
+        .host_str_without_final_dot()
```
dignifiedquire, line 856: hmm, I am not sure, but I think this could keep the dot?

Frando: Why? This is doing a QUIC connection to the relay, doesn't this use the same cert as https?
(so this always worked because no reqwest involved and for quinn we always have the rustls backend, but the logic would be the same as for the other requests, no?)

iroh-base/src/relay_url.rs
Frando: Added a quick test.

iroh/src/net_report.rs
dignifiedquire: true

flub (commented):
I'm kind of -0.5 on this.

1. It's brittle because we need to remember where to use which representation of the URL.  Your comment already has questions about having found every spot.
2. #3486 already fixes this anyway.

dignifiedquire:
> 3. The original reason for the final dot still applies, why let reqwest do the extra DNS lookups?

From the investigation it sounded to melike that is not the case, the dot is removed by rustls at some point, so there seems to be no benefit in this case.

flub:
> > 3. The original reason for the final dot still applies, why let reqwest do the extra DNS lookups?
> 
> From the investigation it sounded to melike that is not the case, the dot is removed by rustls at some point, so there seems to be no benefit in this case.

@Frando correct me if I'm wrong, but I understood the dot is only removed to check whether the name on the certificate matches the dialed name.  So that dot is left in-place for the dialing itself?

ramfox:
closes https://github.com/n0-computer/iroh/issues/3377

flub:
Closing now #3623 is merged.

PR #3491, opened 2025-10-01
WIP: testing some congestion metrics and fixes

Arqu:
## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


dignifiedquire:
conflicts

iroh/src/magicsock/metrics.rs
```diff
@@ -77,4 +77,88 @@ pub struct Metrics {
...
+            nodes_contacted: Counter::default(),
+            nodes_contacted_directly: Counter::default(),
+            connection_handshake_success: Counter::default(),
+            connection_became_direct: Counter::default(),
+            path_marked_outdated: Counter::default(),
+            path_ping_failures: Counter::default(),
+            path_failure_resets: Counter::default(),
+            path_packet_loss_rate: packet_loss_buckets(),
+            path_rtt_variance_ms: rtt_variance_buckets(),
+            path_quality_score: quality_score_buckets(),
```
dignifiedquire: (not blocking) it's a bit sad we have to do this manually when including histograms now :( I wonder if we could adjust our macro or sth 

cc @Frando

Arqu: I was looking for some magic syntax sugar with a default annotation or sth but couldn't work it out.
The problem is for each histogram you want to provide a specific set of buckets that make sense for that metric specifically, so you need to "adjust" it here. However if there was some magic #[default...] annotation I couldn't find where I pass in those funcs, it would be great.

Frando: We can add such an annotation to the macro. Can do tomorrow.

iroh/src/magicsock/metrics.rs
```diff
@@ -77,4 +77,88 @@ pub struct Metrics {
...
+        }
+    }
+}
+
+fn packet_loss_buckets() -> Histogram {
+    Histogram::new(vec![0.0, 0.01, 0.05, 0.1, 0.2, 0.5, 1.0])
```
flub: I guess the buckets are:
- [0, 0.01)
- [0.01, 0.05)
- [0.1, 0.2)
- [0.2, 0.5)
- [0.5, 1.0)
- [1.0, infinity]

?

flub (approved):
Nothing really against it if it helps somewhere.  But be aware that so much of this will disappear soon.  E.g. I just deleted the entire path_validity module in the `feat-multipath` branch.

I'm expecting to not try and merge any of the metrics changes into there and instead we'll have to give it a separate metrics review to get back at least what may still be relevant once it's merged.

iroh/src/magicsock/metrics.rs
Frando: Here's the macro: https://github.com/n0-computer/iroh-metrics/pull/37

Arqu:
I think that's fine, this is mostly a short term exploration of the space.

iroh/src/magicsock/metrics.rs
Arqu: Yes

Arqu:
Currently failing because of the dependency on https://github.com/n0-computer/net-tools/pull/36

PR #3493, opened 2025-10-01
wip: some congestion trickery

Arqu:
## Description

Split off from the original PR as this is dubious whether it's useful but interesting to test a bit.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/lib.rs
+++ b/iroh/src/lib.rs
@@ -275,6 +275,7 @@ pub use iroh_base::{
     KeyParsingError, NodeAddr, NodeId, PublicKey, RelayUrl, RelayUrlParseError, SecretKey,
 };
 pub use iroh_relay::{RelayMap, RelayNode, node_info};
+pub use magicsock::InterfacePriority;
 pub use n0_watcher::Watcher;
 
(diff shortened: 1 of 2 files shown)


dignifiedquire:
the jitter part makes sense to me in general, @flub I am wondering is there sth in multipath/quic to deal with this issue, seems like problem that could happen there as well

flub:
> the jitter part makes sense to me in general, @flub I am wondering is there sth in multipath/quic to deal with this issue, seems like problem that could happen there as well

So the idea is to always add some jitter when scheduling a holepunching attempt in the future?

Arqu:
Closing this PR but keeping the branch as future reference. Unsure if we should file the jitter stuff as an issue to do at a later point?

flub:
> Closing this PR but keeping the branch as future reference. Unsure if we should file the jitter stuff as an issue to do at a later point?

Can you please file the jitter thing as an issue?  Because I already forgot the details.  Maybe even as a sub-issue of the multipath issue  (#3276 ), though definitely on one of the pre-1.0 milestones.

PR #3495, opened 2025-10-03
fix: `impl<T: Discovery> Discovery for Arc<T>`

### LINKED ISSUE #3489: bug: `impl Discover for Arc<T>` incorrect
https://github.com/n0-computer/iroh/blob/60d5310dfe42179f6b3a20e38da4e7144008e541/iroh/src/discovery.rs#L360

The following line:
```
impl<T: Discovery> Discovery for Arc<T> {}
```

makes it so that if someone changes/incorrectly writes:

```
builder.add_discovery(|_| Arc::new(Foo)));
```

instead of

```
builder.add_discovery(|_| Foo);
```

the code suddenly does nothing because [default impls for methods on Discovery will just ignore the `T`/`Foo`](https://github.com/n0-computer/iroh/blob/60d5310dfe42179f6b3a20e38da4e7144008e541/iroh/src/discovery.rs#L326). https://github.com/n0-computer/iroh/blob/60d5310dfe42179f6b3a20e38da4e7144008e541/iroh/src/discovery.rs#L32

For whatever reason  we did it in our code, and it was such a weird bug to debug. :D

ramfox:
## Description

Fixes the implementation of `Discovery` for an `Arc<impl Discovery>`

- Added a failing test illustrating that when we added an `Arc<Discovery>`, the endpoint was unable to actually discover anything
- Fixed the test by actually implementing the trait on an `Arc`

closes #3489 

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/discovery.rs
+++ b/iroh/src/discovery.rs
@@ -333,6 +333,7 @@ impl<T: Discovery> Discovery for Arc<T> {
     fn publish(&self, data: &NodeData) {
         self.as_ref().publish(data);
     }
+
     fn resolve(&self, node_id: NodeId) -> Option<BoxStream<Result<DiscoveryItem, DiscoveryError>>> {
         self.as_ref().resolve(node_id)
     }
--- a/iroh/src/discovery.rs
+++ b/iroh/src/discovery.rs
@@ -357,7 +357,14 @@ pub trait Discovery: std::fmt::Debug + Send + Sync + 'static {
     }
 }
 
-impl<T: Discovery> Discovery for Arc<T> {}
+impl<T: Discovery> Discovery for Arc<T> {
+    fn publish(&self, data: &NodeData) {
+        self.as_ref().publish(data);
+    }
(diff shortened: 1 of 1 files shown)


iroh/src/discovery.rs
```diff
@@ -329,7 +329,14 @@ pub trait Discovery: std::fmt::Debug + Send + Sync + 'static {
     }
 }
 
-impl<T: Discovery> Discovery for Arc<T> {}
+impl<T: Discovery> Discovery for Arc<T> {
+    fn publish(&self, data: &NodeData) {
+        self.as_ref().publish(data);
+    }
+    fn resolve(&self, node_id: NodeId) -> Option<BoxStream<Result<DiscoveryItem, DiscoveryError>>> {
```
flub, line 337:
```suggestion
    }

    fn resolve(&self, node_id: NodeId) -> Option<BoxStream<Result<DiscoveryItem, DiscoveryError>>> {
```
nitpicking (it surprises me rustfmt doesn't do this)

flub (approved):

PR #3496, opened 2025-10-03
fix: 0rtt flakes

Arqu:
## Description

So rough description of what was happening:
- the errors got swallowed but nuked the loop so it failed silently - fixed that
- the 0rtt tries connecting on the new port/addr but still prefers the old path so needs to wait out for the path invalidation to do it's job before it continues to work

Being slow enough made it sometimes pass. Locally now tests pass ~instantly vs 15 seconds or so after the fix to gracefully handle errors.

I'm unsure if this is safe at all, like handling path updates from other angles or partial updates etc.
If it turns out too risky, we can drop the path invalidation changes and just keep the graceful error handling. Tests will pass.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/magicsock/node_map/node_state.rs
+++ b/iroh/src/magicsock/node_map/node_state.rs
@@ -699,18 +699,6 @@ impl NodeState {
         }
 
         let mut access = self.udp_paths.access_mut(now);
-
-        // Invalidate paths that are not in the new address set.
-        // If the remote node sends us new addresses, old ones are likely stale (e.g., after restart).
-        // This forces immediate re-evaluation and prevents preferring "known bad" outdated paths
-        // over "unknown potentially good" new paths.
-        for (ip_port, path_state) in access.paths().iter_mut() {
-            if !new_addrs.contains(&(*ip_port).into()) && !path_state.validity.is_empty() {
-                debug!(path=?ip_port, "invalidating path not in new address set");
-                path_state.validity = PathValidity::empty();
-            }
-        }
-
         for &addr in new_addrs.iter() {
             access
                 .paths()
--- a/iroh/src/magicsock/node_map/node_state.rs
+++ b/iroh/src/magicsock/node_map/node_state.rs
@@ -699,6 +699,18 @@ impl NodeState {
         }
 
         let mut access = self.udp_paths.access_mut(now);
+
+        // Invalidate paths that are not in the new address set.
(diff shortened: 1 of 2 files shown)


iroh/src/magicsock/node_map/node_state.rs
```diff
@@ -699,6 +699,18 @@ impl NodeState {
         }
 
         let mut access = self.udp_paths.access_mut(now);
+
+        // Invalidate paths that are not in the new address set.
+        // If the remote node sends us new addresses, old ones are likely stale (e.g., after restart).
+        // This forces immediate re-evaluation and prevents preferring "known bad" outdated paths
+        // over "unknown potentially good" new paths.
```
dignifiedquire: @Frando @flub can you look at this please, I believe the two of you looked at this before

flub: I'm not sure.  On a cursory look I think we get `NodeAddr` for other reasons, e.g. when a user dials with a `NodeAddr`.  And that could be totally out of date and other info could be fresher.

In the multipath branch I currently send the initial packets to **all** known remotes.  So that would mean whatever the working path, it should find it.  I think it might need some kind of expiry of paths at some point, but wasn't going to do that immediately.

I've honestly been avoiding thinking too much about how to robustly solve these kind of issues in the current version, because it just slows down multipath work... but if others agree there's an improvement here that's fine.

Arqu:
So what's our path forward here, should I drop the path update and just make sure we survive errors in the loop? That's a "good enough" fix for the flaky tests for now. Just not a proper fix for the underlying issue.

dignifiedquire:
> So what's our path forward here, should I drop the path update and just make sure we survive errors in the loop?

Yeah lets do that, and file an issue, with the details that you discovered. And then we reexamine this once multipath has landed

iroh/src/magicsock/node_map/node_state.rs
Frando: I'm also unsure, I'll have to dig in a bit more to find out if this could have unintended side effects (ie falsely invalidating good paths).

Arqu:
Reverted the path invalidation logic and followed up with an issue https://github.com/n0-computer/iroh/issues/3504

dignifiedquire (approved):

PR #3416, opened 2025-07-30
chore: update dependencies

reneleonhardt:
## Description
- Update dependencies
- Let dependabot update dependencies

## Notes & open questions
- Waiting for a new ed25519-dalek release to allow more updates depending on rand 0.9
- renovate could update versions in workflows too, but for some like node and nextest it would be easier to just use the major 22 or minor 0.9 version respectively

PR #3497, opened 2025-10-06
ci: enable dependabot for crates and docker

dignifiedquire:
I would like to try this for a bit

Closes #3416


b5:
claude says we can configure when dependabot runs the check:

```yaml
version: 2
updates:
  - package-ecosystem: "cargo"
    directory: "/"
    schedule:
      interval: "weekly"
      day: "thursday"
      time: "08:00"
      timezone: "America/New_York"
```

PR #3505, opened 2025-10-07
feat(iroh)!: make direct_addresses always be initialised

dignifiedquire:
## Description

Turns out we just need to call the init early enough. 

Thanks to @flub for reminding me of this

## Breaking Changes

- changed
  - `iroh::Endpoint::watch_node_addr` now returns a watchable for `NodeAddr` not `Option<NodeAddr>`

## Notes & open questions

Currently this means that in wasm we have a different signature for `watch_node_addr()` because it can be empty, need some feedback on how to solve this

### DIFF
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -894,29 +894,12 @@ impl Endpoint {
     /// - the endpoint changes its preferred relay server
     /// - more addresses are discovered for this endpoint
     pub fn node_addr(&self) -> NodeAddr {
-        let node_id = self.node_id();
-        let relay_url = self.msock.home_relay().get().first().cloned();
-        let direct_addresses = self
-            .msock
-            .direct_addresses()
-            .get()
-            .unwrap_or_default()
-            .into_iter()
-            .map(|addr| addr.addr)
-            .collect();
-
-        NodeAddr {
-            node_id,
-            relay_url,
-            direct_addresses,
-        }
+        self.watch_node_addr().get()
     }
 
     /// Returns a [`Watcher`] for the current [`NodeAddr`] for this endpoint.
     ///
     /// The observed [`NodeAddr`] will have the current [`RelayUrl`] and direct addresses.
-    ///
-    /// Use [`Watcher::initialized`] to wait for a [`NodeAddr`] that is ready to be connected to:
     ///
     /// ```no_run
     /// # async fn wrapper() -> n0_snafu::Result {
@@ -926,7 +909,7 @@ impl Endpoint {
     ///     .alpns(vec![b"my-alpn".to_vec()])
     ///     .bind()
     ///     .await?;
-    /// let node_addr = endpoint.watch_node_addr().initialized().await;
+    /// let node_addr = endpoint.watch_node_addr().get();
     /// # let _ = node_addr;
     /// # Ok(())
     /// # }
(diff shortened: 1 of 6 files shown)


Frando:
> Currently this means that in wasm we have a different signature for watch_node_addr() because it can be empty, need some feedback on how to solve this

Just let it return an empty addr? (I.e. a NodeAddr with only the node id set). That's much preferable than a different signature IMO

iroh/src/endpoint.rs
```diff
@@ -894,30 +894,13 @@ impl Endpoint {
     /// - the endpoint changes its preferred relay server
     /// - more addresses are discovered for this endpoint
     pub fn node_addr(&self) -> NodeAddr {
-        let node_id = self.node_id();
```
flub, line 897: The first paragraph of the doc comment needs updating I think.  It could describe what you'd expect to be in the direct addresses immediately, but that they -and should- still change as discovered direct addresses become available.

iroh/src/magicsock.rs
```diff
@@ -478,6 +478,70 @@ impl MagicSock {
...
+    fn collect_local_addresses(
+        &self,
+        netmon_watcher: &mut n0_watcher::Direct<netmon::State>,
+        addrs: &mut BTreeMap<SocketAddr, DirectAddrType>,
+    ) {
+        let local_addrs: Vec<(SocketAddr, SocketAddr)> = self
+            .ip_bind_addrs()
+            .iter()
+            .copied()
+            .zip(self.ip_local_addrs())
```
flub: I don't follow why these two iters form pairs?  Are they even guaranteed to be the same length?

**update**: oh, this is moved code.  Probably fine then, but I wouldn't mind more comments because I don't understand what it's trying to do.

iroh/src/magicsock.rs
```diff
@@ -478,6 +478,70 @@ impl MagicSock {
         }
     }
 
+    #[cfg(not(wasm_browser))]
+    fn collect_local_addresses(
+        &self,
+        netmon_watcher: &mut n0_watcher::Direct<netmon::State>,
+        addrs: &mut BTreeMap<SocketAddr, DirectAddrType>,
```
flub: Whoa, I'm very not used to seeing an "out" parameter in rust code.

iroh/src/endpoint.rs
dignifiedquire: updated

iroh/src/magicsock.rs
dignifiedquire: added some more comments

iroh/src/magicsock.rs
dignifiedquire: it actually works out quite nice in rust if you need it, and makes your code go brr ;P

iroh/src/endpoint.rs
```diff
@@ -875,49 +875,25 @@ impl Endpoint {
...
-            .into_iter()
-            .map(|addr| addr.addr)
-            .collect();
-
-        NodeAddr {
-            node_id,
-            relay_url,
-            direct_addresses,
-        }
+        self.watch_node_addr().get()
```
flub, line 890: This kind of gets us back to the "why have two methods?" question.  This one is now pretty trivial.

flub: Though I do appreciate this way people don't have to learn about the `Watcher` trait.

iroh/src/magicsock.rs
flub: thanks, they help!  it makes sense now

PR #3512, opened 2025-10-08
fix(iroh): improve reusing of QAD reports

dignifiedquire:
## Description

This splits v4 vs v6 QAD report reuse between runs, to reduce traffic and overall work

### DIFF
--- a/iroh/src/net_report.rs
+++ b/iroh/src/net_report.rs
@@ -306,63 +306,69 @@ impl Client {
             report.update(&r);
         }
 
-        #[cfg(not(wasm_browser))]
-        let mut qad_v4_stream = self.qad_conns.watch_v4();
-        #[cfg(wasm_browser)]
-        let mut qad_v4_stream = n0_future::stream::empty::<Option<()>>();
-        #[cfg(not(wasm_browser))]
-        let mut qad_v6_stream = self.qad_conns.watch_v6();
-        #[cfg(wasm_browser)]
-        let mut qad_v6_stream = n0_future::stream::empty::<Option<()>>();
+        if self.have_enough_reports(&if_state, do_full, num_relays, &report) {
+            // check if we already have enough probes immediately after QAD returns
+            trace!("have enough probe reports, aborting further probes");
+            // shuts down the probes
+            drop(actor);
+        } else {
+            #[cfg(not(wasm_browser))]
+            let mut qad_v4_stream = self.qad_conns.watch_v4();
+            #[cfg(wasm_browser)]
+            let mut qad_v4_stream = n0_future::stream::empty::<Option<()>>();
+            #[cfg(not(wasm_browser))]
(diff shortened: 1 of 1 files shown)


iroh/src/net_report.rs
```diff
@@ -576,32 +571,99 @@ impl Client {
...
+                return true;
+            }
+            false
         } else {
-            None
+            // Incremental reports, here the requirements are reduced further
+            match (state.have_v4, state.have_v6) {
+                (true, true) => {
+                    // Both IPv4 and IPv6 are expected to be available
+                    if num_ipv4 > 1 && num_ipv6 > 1 {
```
flub: `>=` perhaps?  Or do you really want 2 each?  same off-by-one question in the full report above.

iroh/src/net_report.rs
```diff
@@ -576,32 +571,99 @@ impl Client {
...
+                }
+                (false, false) => {
+                    // Nothing works
+                    return false;
+                }
+            }
 
-            Some(timeout)
+            // If we have at least one probe per relay, we are happy
+            if num_total >= num_relays {
```
flub: does this case need to consider `state.have_v4` and `state.have_v6`?  E.g. if you have both and above doesn't yet return because its still waiting on ipv4 then this will return and no longer require ipv4?

iroh/src/net_report.rs
dignifiedquire: sigh, all of these should be `>=`....

dignifiedquire: fixed..

iroh/src/net_report.rs
dignifiedquire: this needs to only look at https probes, you are right

iroh/src/net_report.rs
```diff
@@ -330,12 +319,6 @@ impl Client {
             tokio::select! {
                 biased;
 
-                _ = &mut timeout_fut, if timeout_fut.is_some() => {
```
flub, line 333: Hum, I'm still not convinced.

Above happens:
```rust
let reports = self
            .spawn_qad_probes(&if_state, enough_relays, do_full)
            .await;
```
If you look at the implementation of `spawn_qad_reports` it does much more than just spawning them.  It waits up to 3s until it receives enough reports and then returns the reports.

**While** this happens the reportgen::Client is also running and will basically start https & captive portal probes after 200ms.

**Then** it goes to this select loop and waits for **new** qad reports, which will only yield something if the QAD address changes or if they still happen to arrive.  At the same time it reads reports from the reportgen client via `probe_rx`.

So if QAD returned a result in 150ms we'll *still* be waiting for at least an https or a captive portal probe result.

I think you need to do a `have_enough_reports` check before entering this loop?

dignifiedquire: added the suggested check, which I believe should solve this issue

PR #3522, opened 2025-10-09
feat(iroh)!: allow dynamic changing of the `RelayMap` of an endpoint

### LINKED ISSUE #3471: feat: ability to dynamically modify an endpoint RelayMap
There are situations where we'd like to be able to dynamically manipulate the set of Relays an endpoint uses as candidates for a home relay. Common use cases are scaling the relay network, or recovering from failure.

We'd likely rely on relays themselves as a source of relay network topology changes, but need to design & implement a new system for this.

dignifiedquire:
## Description

Closes #3471

## Breaking Changes

- added
  - `iroh::Endpoint::insert_relay`
  - `iroh::Endpoint::remove_relay`
  - `iroh::Endpoint::RelayMap::insert`
  - `iroh::Endpoint::RelayMap::remove`
- changed 
  - `iroh_relay::RelayMap::urls` 
  - `iroh_relay::RelayMap::nodes`
  - `iroh_relay::RelayMap::get_node`

## Notes & open questions

- Are the changes to the API of `RelayMap` acceptable?
- Is this enough?

### DIFF
--- a/iroh-relay/src/relay_map.rs
+++ b/iroh-relay/src/relay_map.rs
@@ -1,6 +1,10 @@
 //! based on tailscale/tailcfg/derpmap.go
 
-use std::{collections::BTreeMap, fmt, sync::Arc};
+use std::{
+    collections::BTreeMap,
+    fmt,
+    sync::{Arc, RwLock},
+};
 
 use iroh_base::RelayUrl;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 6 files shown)


iroh-relay/src/relay_map.rs
```diff
@@ -1,23 +1,45 @@
...
+
+impl Eq for RelayMap {}
+
 impl RelayMap {
     /// Returns the sorted relay URLs.
-    pub fn urls(&self) -> impl Iterator<Item = &RelayUrl> {
-        self.nodes.keys()
+    pub fn urls<T>(&self) -> T
+    where
+        T: FromIterator<RelayUrl>,
```
flub, line 35: that's... fancy

And maybe a little unfortunate as we're now forcing an allocation.  But I guess this is needed because the the inner lock.

iroh/src/magicsock.rs
```diff
@@ -2031,6 +2057,10 @@ impl Actor {
         }
     }
 
+    fn handle_relay_map_change(&mut self) {
+        self.re_stun(UpdateReason::RelayMapChange);
```
flub, line 2064: this is like the only thing that is needed, and the test just works?  I'm both impressed and surprised!

PR #3523, opened 2025-10-10
feat(iroh): introduce endpoint presets

### LINKED ISSUE #3470: replace `iroh::endpoint::Builder::discovery_n0` with a more declarative API
We like what the method does, but it has some inconsistency. n0 relays re configured by default, but `discovery_n0` wires up number 0 DNS. We'd like to harmonize the builder API around a uniform profile that uses easy defaults, and a blank "with opts" variant for full configuration.

dignifiedquire:
## Description

Closes #3470

## Breaking Changes

- removed
  - `iroh::endpoint::Builder::n0_discovery`
  - `iroh::endpoint::Builder::discovery_dht`
  - `iroh::endpoint::Builder::discovery_local_network`
  - `iroh::endpoint::Builder::discovery`
  - `iroh::discovery::DiscoveryContext`
  - `iroh::endpoint::Builder::default`
- added
  -  `iroh::Endpoint::bind` - this allows to immediately create an `Endpoint` with the defaults and binds it
  - `iroh::endpoint::Endpoint::empty_builder`
  - `iroh::endpoint::presets` - new module, defining a trait on how to preconfigure and `Endpoint`
  - `iroh::endpoint::Builder::new` - creates a new builder with the `N0` preset
  - `iroh::endpoint::Builder::preset`
  - `iroh::endpoint::Builder::empty` - creates a new builder, completely empty (no discovery, no relay map)
- changed
  - `iroh::'Endpoint::add_discovery` is now called just `discovery`
  - `iroh::discovery::IntoDiscovery` now takes an `Endpoint` instead of a `DiscoveryContext`

### DIFF
--- a/iroh/src/discovery.rs
+++ b/iroh/src/discovery.rs
@@ -448,7 +448,7 @@ impl ConcurrentDiscovery {
         self.services.read().expect("poisoned").is_empty()
     }
 
-    /// How many services are configured?
+    /// Returns the number of services configured.
     pub fn len(&self) -> usize {
         self.services.read().expect("poisoned").len()
     }
--- a/iroh/src/discovery.rs
+++ b/iroh/src/discovery.rs
@@ -62,12 +62,14 @@
 //! use iroh::{
 //!     Endpoint, SecretKey,
 //!     discovery::{dns::DnsDiscovery, pkarr::PkarrPublisher},
+//!     endpoint::RelayMode,
 //! };
 //!
 //! # async fn wrapper() -> n0_snafu::Result<()> {
 //! let ep = Endpoint::empty_builder()
 //!     .discovery(PkarrPublisher::n0_dns())
 //!     .discovery(DnsDiscovery::n0_dns())
+//!     .relay_mode(RelayMode::Default)
 //!     .bind()
 //!     .await?;
 //! # Ok(())
@@ -81,6 +83,7 @@
 //! # {
 //! # use iroh::{
 //! #    discovery::{dns::DnsDiscovery, pkarr::PkarrPublisher, mdns::MdnsDiscovery},
+//! #    endpoint::RelayMode,
 //! #    Endpoint, SecretKey,
 //! # };
 //! #
@@ -89,6 +92,7 @@
 //!     .discovery(PkarrPublisher::n0_dns())
 //!     .discovery(DnsDiscovery::n0_dns())
 //!     .discovery(MdnsDiscovery::builder())
+//!     .relay_mode(RelayMode::Default)
 //!     .bind()
 //!     .await?;
 //! # Ok(())
--- a/iroh/src/discovery.rs
+++ b/iroh/src/discovery.rs
@@ -66,10 +66,9 @@
 //! };
 //!
 //! # async fn wrapper() -> n0_snafu::Result<()> {
-//! let ep = Endpoint::empty_builder()
+//! let ep = Endpoint::empty_builder(RelayMode::Default)
 //!     .discovery(PkarrPublisher::n0_dns())
 //!     .discovery(DnsDiscovery::n0_dns())
-//!     .relay_mode(RelayMode::Default)
 //!     .bind()
 //!     .await?;
 //! # Ok(())
@@ -88,11 +87,10 @@
 //! # };
 //! #
 //! # async fn wrapper() -> n0_snafu::Result<()> {
-//! let ep = Endpoint::empty_builder()
+//! let ep = Endpoint::empty_builder(RelayMode::Default)
 //!     .discovery(PkarrPublisher::n0_dns())
 //!     .discovery(DnsDiscovery::n0_dns())
 //!     .discovery(MdnsDiscovery::builder())
-//!     .relay_mode(RelayMode::Default)
 //!     .bind()
 //!     .await?;
 //! # Ok(())
@@ -916,9 +914,8 @@ mod tests {
     ) -> (Endpoint, AbortOnDropHandle<Result<()>>) {
         let secret = SecretKey::generate(rng);
 
-        let ep = Endpoint::empty_builder()
+        let ep = Endpoint::empty_builder(RelayMode::Disabled)
             .secret_key(secret)
-            .relay_mode(RelayMode::Disabled)
             .alpns(vec![TEST_ALPN.to_vec()])
             .bind()
             .await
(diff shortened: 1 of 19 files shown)


iroh/src/endpoint/presets.rs
```diff
@@ -0,0 +1,62 @@
...
+//! # Ok(())
+//! # }
+//! ```
+use super::Builder;
+use crate::discovery::pkarr::PkarrPublisher;
+
+/// Defines a preset
+pub trait Preset {
+    /// Applies the configuration to the passed in [`Builder`].
+    fn apply(self, builder: Builder) -> Builder;
```
rklaehn, line 26: Any reason for preset to take self and not &self?

Imagine you had a complex preset, you would have to clone it to apply it.

dignifiedquire: Nothing specific, I was just expecting them to be use once only

iroh/src/endpoint.rs
```diff
@@ -117,13 +119,30 @@ pub struct Builder {
...
+        Self::empty().preset(preset)
+    }
+
+    /// Applies the given [`Preset`].
+    pub fn preset<P: Preset>(mut self, preset: P) -> Self {
+        self = preset.apply(self);
+        self
+    }
+
+    /// Creates an empty builder, including
```
Frando: Unfinished doc comment

Should note that this has neither discovery nor relays configured then!

iroh/src/endpoint/presets.rs
```diff
@@ -0,0 +1,70 @@
+//! Presets allow configuring an endpoint quickly with a chosen set of defaults.
+//!
+//! # Example
+//!
+//! ```no_run
+//! # async fn wrapper() -> n0_snafu::Result {
+//! use iroh::{Endpoint, Watcher, endpoint::presets};
+//!
+//! let endpoint = Endpoint::builder().preset(presets::N0).bind().await?;
```
Frando: This should use `Endpoint::empty_builder()` otherwise adding the preset is a no-op, no?

iroh/src/endpoint/presets.rs
```diff
@@ -0,0 +1,70 @@
...
+/// n0.computer dns server.
+///
+/// This will by default use [`N0_DNS_PKARR_RELAY_PROD`].
+/// When in tests, or when the `test-utils` feature is enabled, this will use the
+/// [`N0_DNS_PKARR_RELAY_STAGING`].
+///
+/// [`N0_DNS_PKARR_RELAY_PROD`]: crate::discovery::pkarr::N0_DNS_PKARR_RELAY_PROD
+/// [`N0_DNS_PKARR_RELAY_STAGING`]: crate::discovery::pkarr::N0_DNS_PKARR_RELAY_STAGING
+#[derive(Debug, Copy, Clone, Default)]
+pub struct N0;
```
Frando, line 49: I think we should have a preset for just adding the N0 relays but not the discovery, to make it easy to keep the previous default. So `Endpoint::empty_builder().preset(N0Relays).bind().await` would be equal to `Endpoint::builder().bind().await` in iroh 0.93.

It's a bit unfortunate that `N0Relays` reads like *no relays*. I think I'd prefer to call it `NumberZeroRelays` and `NumberZeroDiscovery` and `NumberZero` or similar. Now that the default for builder() is to include the full number zero preset, they won't have to be used in evrery example so it doesn't hurt if the names are a bit longer IMO.

`NumberZero` could then just to `NumberZeroRelays.apply(builder); NumberZeroDiscovery.apply(builder);` in its Preset impl.

dignifiedquire: I am not convinced a relay only preset makes sense, you can currently just write

```rust
Endpoint::empty_builder().relay_mode(RelayMode::Default).bind().await
```

which is not longer or more complicated than using a preset

iroh/src/endpoint/presets.rs
dignifiedquire: fixed

iroh/src/endpoint/presets.rs
ramfox: Agree that we don't need an empty relay mode preset.

iroh/src/discovery.rs
```diff
@@ -482,7 +444,7 @@ impl ConcurrentDiscovery {
         self.services.read().expect("poisoned").is_empty()
     }
 
-    /// How many services are configured
+    /// How many services are configured?
```
ramfox:
```suggestion
    /// Returns the number of services configured.
```

A nit, but perhaps a better way to phrase it.

ramfox:
Hey, so I pushed a commit that includes removing `.relay_mode(RelayMode::Disabled)` where we are now using an `empty_builder()`, since that already includes `RelayMode::Disabled` and added using an `empty_builder` in places that shouldn't have discovery enabled in the test (from what I can tell).

If I've misunderstood the purpose of a test feel free to revert or change.

ramfox:
Two concerns about this PR:
1) users are not going to be used to having discovery on by default
2) it's not immediately obvious that if you use an `empty_builder` and then add discovery services yourself, you may be foot-gunning yourself if you don't also add a specific relay mode to the endpoint builder as well, since most of our discovery really depends on being connected to a relay mode, and `empty_builder` sets the relay mode to `RelayMode::Disabled`

The first issue is just about communicate, and I think folks will just get used to it.

The second one I'm more concerned about because it's easy to just not realize what you are doing. This may also be solved a bit by documentation (for example, in the `empty_builder` method, mentioning that if you plan to add a discovery service you may also want to add a relay mode to the builder before binding.)

I think we need a bit of an answer around the second point before we can merge. We can decide it's enough to just communicate this through docs, or we can decide that's too big of a foot-gun and need to change some things, or we can decide it doesn't actually matter, or some other fourth option.

One general thought:
- for clarity, we may want to make some presets in our test utils, that makes it obvious what the expectations are for the test, even if the preset is really just equivalent to `Endpoint::empty_builder()` & feels redundant.
This may help if we:
a) change what the defaults are in APIs like `empty_builder` or `Endpoint::bind` & then need to search each test and make sure we update accordingly
b) just make it really clear when setting up the test, what the expectations are for endpoint at a glance.

Just a thought!

ramfox (changes requested):

dignifiedquire:
@ramfox I have pushed a version that enforces choosing a `RelayMode` when constructing an empty builder, in many use cases this is the same or less code than before, and only in the full empty version a little bit more verbose, which seems fine given the better understanding this lends to the reader and writer of the code

ramfox (approved):

Issue #3177, opened 2025-02-10
Review foreign types in public APIs


Frando:
We currently expose quite a lot of types from dependencies in the pubic API of `iroh`. We should review that list before 1.0.
For some types from widely used crates with a 1.0 release, e.g. `bytes::Bytes` , we're likely fine with exposing them.

For others, like our `From<hickory_resolver::TokioResolver> for DnsResolver` impl, we might want to remove them.

For types from pre-1.0 crates, leaving them in the public API of iroh means we cannot update the crate (apart from patch-level changes) without a semver breaking change in iroh.

Frando:
There's this tool [`cargo-check-external-types`](https://github.com/awslabs/cargo-check-external-types) which produces a list of foreign types that are used in public APIs.  (If you want to try it out: Note that it needs a specific nightly version, see commands below - add the toolchain with rustup install before).

Note that it does *not* follow re-exports.

`external-types.toml`
```
allowed_external_types = [
    "anyhow::*",
    "bytes::*",
    "iroh_base::*",
    "iroh_metrics::*",
    "iroh_relay::*",
    "serde::*",
    "url::*",
]

```


Here's the current list for `iroh`, with a few crates already excluded (in the `external-types.toml`)
```
cd iroh
cargo +nightly-2024-06-30 check-external-types --config ../external-types.toml --output-format markdown-table --all-features | tee table.md
```

| Crate | Type | Used In |
| ---   | ---  | ---     |
| futures_core | futures_core::stream::Stream | iroh/src/watchable.rs:261:0 |
| futures_lite | futures_lite::future::Boxed | iroh/src/protocol.rs:115:4 |
| futures_lite | futures_lite::future::Boxed | iroh/src/protocol.rs:118:4 |
| futures_lite | futures_lite::stream::Boxed | iroh/src/discovery.rs:158:4 |
| futures_lite | futures_lite::stream::Boxed | iroh/src/discovery.rs:188:4 |
| iroh_net_report | iroh_net_report::metrics::Metrics | iroh/src/metrics.rs:4:0 |
| iroh_quinn | iroh_quinn::connection::AcceptBi | iroh/src/endpoint.rs:1356:4 |
| iroh_quinn | iroh_quinn::connection::AcceptBi | iroh/src/endpoint.rs:46:4 |
| iroh_quinn | iroh_quinn::connection::AcceptUni | iroh/src/endpoint.rs:1343:4 |
[...]

PR #3529, opened 2025-10-13
feat(iroh-base)!: reduce external types in the iroh-base API for keys

dignifiedquire:
## Description

In preparation of stabilizing our APIs, this removes `ed25519_dalek` types from the public API of `iroh-base`, except for a few conversions, which are `doc(hidden)` to clearly mark them as not stable.

Ref #3177

## Breaking Changes

-  added `iroh_base::Signature` which replaces `ed25519_dalek::Signature` in the public API of `iroh_base`
- `iroh_base` error types have changed
- removed `Into` and `From` conversions for `PublicKey` - `ed25519_dalek::VerifyingKey`
- removed `Into` and `From` conversions for `SecretKey` - `ed25519_dalek::SigningKey`

### DIFF
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -1986,9 +1986,12 @@ impl Connection {
                         return Err(RemoteNodeIdSnafu.build());
                     }
 
-                    let peer_id = VerifyingKey::from_public_key_der(&certs[0])
-                        .map_err(|_| RemoteNodeIdSnafu.build())?
-                        .into();
+                    let peer_id = NodeId::from_bytes(
+                        VerifyingKey::from_public_key_der(&certs[0])
+                            .map_err(|_| RemoteNodeIdSnafu.build())?
+                            .as_bytes(),
+                    )
+                    .expect("valid key");
                     Ok(peer_id)
                 }
                 Err(err) => {
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -1986,12 +1986,11 @@ impl Connection {
                         return Err(RemoteNodeIdSnafu.build());
                     }
 
-                    let peer_id = NodeId::from_bytes(
+                    let peer_id = NodeId::from_verifying_key(
                         VerifyingKey::from_public_key_der(&certs[0])
-                            .map_err(|_| RemoteNodeIdSnafu.build())?
-                            .as_bytes(),
-                    )
-                    .expect("valid key");
+                            .map_err(|_| RemoteNodeIdSnafu.build())?,
+                    );
+
                     Ok(peer_id)
                 }
                 Err(err) => {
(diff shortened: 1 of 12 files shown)


Frando:
Not sure about `#[doc(hidden)]` on `From` impls. It is quite error-prone: Users may write `key.into()` just because you're used to it and - it works. But if we deem these impls fine to break in semver-compat releases, then people run into it nevertheless. So I'd suggest to remove the `From` impls and have private `into_verifying_key` etc. fns. And users would then have to go over the to_bytes/from_bytes route, right? Which is what we'd recommend anyway if we mark the From impls doc_hidden.

Edit: And if we need the conversions in downstream crates, maybe make them pub fns with doc_hidden and a doc comment that explains that these may break in semver-compat releases?

dignifiedquire:
> Not sure about `#[doc(hidden)]` on `From` impls. It is quite error-prone: Users may write `key.into()` just because you're used to it and - it works. But if we deem these impls fine to break in semver-compat releases, then people run into it nevertheless. So I'd suggest to remove the `From` impls and have private `into_verifying_key` etc. fns. And users would then have to go over the to_bytes/from_bytes route, right? Which is what we'd recommend anyway if we mark the From impls doc_hidden.
> 
> Edit: And if we need the conversions in downstream crates, maybe make them pub fns with doc_hidden and a doc comment that explains that these may break in semver-compat releases?

yeah we can do that as alternative

dignifiedquire:
@Frando changed based on your suggestions

iroh/src/endpoint.rs
```diff
@@ -1986,9 +1986,12 @@ impl Connection {
                         return Err(RemoteNodeIdSnafu.build());
                     }
 
-                    let peer_id = VerifyingKey::from_public_key_der(&certs[0])
-                        .map_err(|_| RemoteNodeIdSnafu.build())?
-                        .into();
+                    let peer_id = NodeId::from_bytes(
```
Frando: This will be validating the key twice, maybe add `NodeId::from_public_key_der(&[u8]) -> Result<Self>`? But can also be a separate PR, I think this is not changed from before.

We also really should cache the remote's node id on the `Connection` struct I think, it seems expensive to do this whenever it's accessed. But also separate PR ofc.

iroh/src/key.rs
```diff
@@ -3,18 +3,21 @@
...
 use nested_enum_utils::common_fields;
 use snafu::{ResultExt, Snafu, ensure};
 
 pub(crate) const NONCE_LEN: usize = 24;
 
 const AEAD_DATA: &[u8] = &[];
 
-pub(super) fn public_ed_box(key: &ed25519_dalek::VerifyingKey) -> crypto_box::PublicKey {
+pub(super) fn public_ed_box(key: &PublicKey) -> crypto_box::PublicKey {
+    let key = ed25519_dalek::VerifyingKey::from_bytes(key.as_bytes()).expect("valid key");
```
Frando: se could use `key.as_verifying_key()` here if we keep it pub. I think the only other usage atm  is in `iroh/src/tls/verifier.rs`. So Either let's use `as_verifying_key` here and there, or make it not pub and use from_bytes/as_bytes in both places. But I think using the pub+doc_hidden fn here is fine.

Frando (approved):
nice! some comments but no blockers.

PR #3537, opened 2025-10-14
test(iroh): fix flaky test_active_relay_inactive

dignifiedquire:


iroh/src/magicsock/transports/relay/actor.rs
```diff
@@ -1478,68 +1478,74 @@ mod tests {
...
-                let (tx, rx) = oneshot::channel();
-                inbox_tx.send(ActiveRelayMessage::PingServer(tx)).await.ok();
-                if tokio::time::timeout(Duration::from_millis(100), rx)
-                    .await
-                    .map(|resp| resp.is_ok())
-                    .unwrap_or_default()
-                {
-                    break;
+            // Wait until the actor is connected to the relay server.
+            tokio::time::timeout(Duration::from_millis(200), async {
```
flub: isn't this 200ms the likely cause of it being flaky?  You could bump this to say 5s?

dignifiedquire: no this always happens, the failing timeout is the shutdown below

PR #3543, opened 2025-10-15
chore: update copyright dates

dignifiedquire:


rklaehn (approved):
Is it enough to mention the last year, or should it be 2023-2025?

PR #3544, opened 2025-10-15
refactor(iroh-base)!: move tickets into their own crate

dignifiedquire:
## Description

These now live in https://github.com/n0-computer/iroh-tickets and will be published as `iroh-tickets`.

## Breaking Changes

- `iroh-base::ticket` has been removed

## Notes

The `0rtt` and `transfer` examples now use just `EndpointId`s to connect, and not tickets anymore.  Otherwise we end up in a problematic dependency tree. If necessary I can add direct addresses and relay urls to the `transfer` example again

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -27,6 +27,4 @@ ignore = [
 ]
 
 [sources]
-allow-git = [
-    "https://github.com/n0-computer/iroh-tickets.git",
-]
+allow-git = []


Arqu (changes requested):
I've made a PR https://github.com/n0-computer/chuck/pull/80 to update netsim to use the endpoint ID. However many of those are air gaped and don't have standard discovery mechanics available. The ticked contained direct addr info which was needed for it to work in this environment. Dunno if relay url is needed because they all are configured to the same test spawned one so that should be fine I think.

You can keep re-running this job till it passes with your changes to check:
https://github.com/n0-computer/iroh/actions/runs/18645720236/job/53152491327

Frando:
I'd propose to just use https://github.com/n0-computer/iroh-tickets in chuck then, no need to make it use discovery.

dignifiedquire:
I can also add manual passing of ip addrs to the example if that's what we need for chuck

ramfox:
added `--remote-relay-url` and `--remote-direct-address` as fields on the `transfer` example. Hopefully that might help alleviate pressure on `chuck` now that tickets have been removed? @Arqu

iroh/examples/transfer.rs
```diff
@@ -146,6 +147,10 @@ enum Commands {
     /// Fetch data.
     Fetch {
         remote_id: EndpointId,
+        #[clap(long)]
+        remote_relay_url: Option<RelayUrl>,
+        #[clap(long)]
+        remote_direct_address: Vec<SocketAddr>,
```
Arqu, line 153: remote_direct_addr :)

Arqu:
Netsim is now passing https://github.com/n0-computer/iroh/actions/runs/18645720236/job/53255016686
With this pr https://github.com/n0-computer/chuck/pull/80

We will need to coordinate the merges around these not to screw up other PRs

dpc:
At the time of writting https://docs.rs/iroh/latest/iroh/discovery/static_provider/struct.StaticProvider.html still mentions and links (broken) to EndpointTicket.

flub:
> At the time of writting https://docs.rs/iroh/latest/iroh/discovery/static_provider/struct.StaticProvider.html still mentions and links (broken) to EndpointTicket.

Thanks, fix here: https://github.com/n0-computer/iroh/pull/3814

PR #3554, opened 2025-10-16
feat!: transport generic EndpointAddr

### LINKED ISSUE #3417: feat: Future-proof `NodeAddr`
The current `NodeAddr` looks like this:
```rust
pub struct NodeAddr {
    pub node_id: NodeId,
    pub relay_url: Option<RelayUrl>,
    pub direct_addresses: BTreeSet<SocketAddr>,
}
```

However internally we have refactored things into `Transports` and conceptually the home relay is not that a special a transport.  In fact you could easily allow for an iroh endpoint to be reachable via multiple relay servers, perhaps a public one and one reachable only on some private network.

We should make sure that the `NodeAddr` is more agnostic towards transports.

dignifiedquire:
## Description

This marks the first public step towards generic transports. For now there are two types of `TransportAddr`s, `Relay` and `Ip` which, replace the previous explicit fields in `EndpointAddr` of `relay_url` and `direct_addrs`.

Closes https://github.com/n0-computer/iroh/issues/3417

## Breaking Changes

- `direct_addresses` are now called `ip_addresses`
- `EndpointAddr` type is completely changed
-  APIs with `addresse(s)` are now normalized to `addr(s)` to be consistent in the code base

### DIFF
--- a/iroh-base/src/endpoint_addr.rs
+++ b/iroh-base/src/endpoint_addr.rs
@@ -55,7 +55,7 @@ pub enum TransportAddr {
 }
 
 impl EndpointAddr {
-    /// Creates a new [`EndpointAddr`] with no `relay_url` and no `direct_addresses`.
+    /// Creates a new [`EndpointAddr`] with no addresses.
     pub fn new(id: PublicKey) -> Self {
         EndpointAddr {
             id,
@@ -69,11 +69,8 @@ impl EndpointAddr {
         self
     }
 
-    /// Adds the given direct addresses.
-    pub fn with_direct_addresses(
-        mut self,
-        addresses: impl IntoIterator<Item = SocketAddr>,
-    ) -> Self {
+    /// Adds the given IP addresses.
+    pub fn with_ip_addresses(mut self, addresses: impl IntoIterator<Item = SocketAddr>) -> Self {
         for addr in addresses.into_iter() {
             self.addrs.insert(TransportAddr::Ip(addr));
         }
@@ -93,7 +90,7 @@ impl EndpointAddr {
         self.addrs.is_empty()
     }
 
-    /// Returns the direct addresses of this peer.
+    /// Returns the IP addresses of this peer.
     pub fn ip_addresses(&self) -> impl Iterator<Item = &SocketAddr> {
         self.addrs.iter().filter_map(|addr| match addr {
             TransportAddr::Ip(addr) => Some(addr),
@@ -112,12 +109,12 @@ impl EndpointAddr {
 
 impl From<(PublicKey, Option<RelayUrl>, &[SocketAddr])> for EndpointAddr {
     fn from(value: (PublicKey, Option<RelayUrl>, &[SocketAddr])) -> Self {
-        let (id, relay_url, direct_addresses_iter) = value;
+        let (id, relay_url, ip_addrs_iter) = value;
         let mut addrs = BTreeSet::new();
         if let Some(url) = relay_url {
             addrs.insert(TransportAddr::Relay(url));
         }
-        for addr in direct_addresses_iter {
+        for addr in ip_addrs_iter {
             addrs.insert(TransportAddr::Ip(*addr));
         }
         EndpointAddr { id, addrs }
--- a/iroh-base/src/endpoint_addr.rs
+++ b/iroh-base/src/endpoint_addr.rs
@@ -41,13 +41,13 @@ pub struct EndpointAddr {
     /// The endpoint's identifier.
     pub id: EndpointId,
     /// The endpoint's addresses
-    pub addrs: BTreeSet<AddrType>,
+    pub addrs: BTreeSet<TransportAddr>,
 }
 
 /// Available address types.
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[non_exhaustive]
-pub enum AddrType {
+pub enum TransportAddr {
     /// IP based addresses
     Ip(SocketAddr),
     /// Relays
@@ -65,7 +65,7 @@ impl EndpointAddr {
 
     /// Adds a relay url.
     pub fn with_relay_url(mut self, relay_url: RelayUrl) -> Self {
-        self.addrs.insert(AddrType::Relay(relay_url));
+        self.addrs.insert(TransportAddr::Relay(relay_url));
         self
     }
 
@@ -75,13 +75,13 @@ impl EndpointAddr {
         addresses: impl IntoIterator<Item = SocketAddr>,
     ) -> Self {
         for addr in addresses.into_iter() {
-            self.addrs.insert(AddrType::Ip(addr));
+            self.addrs.insert(TransportAddr::Ip(addr));
         }
         self
     }
 
     /// Creates a new [`EndpointAddr`] from its parts.
-    pub fn from_parts(id: PublicKey, addrs: impl IntoIterator<Item = AddrType>) -> Self {
+    pub fn from_parts(id: PublicKey, addrs: impl IntoIterator<Item = TransportAddr>) -> Self {
         Self {
             id,
             addrs: addrs.into_iter().collect(),
@@ -96,7 +96,7 @@ impl EndpointAddr {
     /// Returns the direct addresses of this peer.
     pub fn ip_addresses(&self) -> impl Iterator<Item = &SocketAddr> {
         self.addrs.iter().filter_map(|addr| match addr {
-            AddrType::Ip(addr) => Some(addr),
+            TransportAddr::Ip(addr) => Some(addr),
             _ => None,
         })
     }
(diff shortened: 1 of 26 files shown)


iroh-base/src/endpoint_addr.rs
```diff
@@ -39,26 +39,33 @@ use crate::{EndpointId, PublicKey, RelayUrl};
...
-    pub direct_addresses: BTreeSet<SocketAddr>,
+    pub id: EndpointId,
+    /// The endpoint's addresses
+    pub addrs: BTreeSet<AddrType>,
+}
+
+/// Available address types.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[non_exhaustive]
+pub enum AddrType {
```
flub: Let's do some bikeshedding!  I don't expect a type having `Type` in it's name to contain a *value*.  So I don't like this name.

iroh-base/src/endpoint_addr.rs
```diff
@@ -39,26 +39,33 @@ use crate::{EndpointId, PublicKey, RelayUrl};
...
+    /// IP based addresses
+    Ip(SocketAddr),
```
flub: If I understand rust visibility correctly anyone can directly create this using `AddrType::Ip(any_crazy_socketaddr)`.  I guess there's little to do about it.  But I've found in `transports::Addr` that really I always want to have the **canonical** IP address used in the contained SocketAddr.  And I can't really enforce that because of this visibility.

OTOH creating our own version of `SocketAddr` is just going to be miserable for everyone.  I'm not sure how important being canonical here is.  I think you could make some construct where the enum itself is private and wrapped inside a struct, so that users always have to pass their SocketAddr through our constructor.  It would also mean new transports are basically new `from_something()` constructor functions.  I don't know how you solve accessing the addresses in this world.

Maybe there are also IP addresses we'd like to reject?  Like if someone is using an IPv6 address starting with [n0's ULA global ID](https://github.com/n0-computer/iroh/blob/4ed700e8da8daf6bb3d897ceadea52a4dad6952c/iroh/src/net_report/ip_mapped_addrs.rs#L30) that would not be a valid transport address.

iroh-base/src/endpoint_addr.rs
dignifiedquire: > Let's do some bikeshedding! I don't expect a type having `Type` in it's name to contain a _value_. So I don't like this name.

please suggest names, I haven't come up with a better one yet

iroh-base/src/endpoint_addr.rs
dignifiedquire: I think we will have to enforce this on an internal type, this also gets serialized and stored in things like DNS records, so I don't think it is realistic to enforce much about the canoncialization on this type in general

Frando: Normal users shouldn't ever have a need to create this type themselves, right?
It's always `endpoint.addr() -> serialize to ticket or discovery -> retrieve from ticket or discovery -> feed to endpoint via discovery or connect`

flub: Sure, but if it's possible users will do it.

iroh-base/src/endpoint_addr.rs
flub: `TransportAddr`?  I know internally that might be confusing with `transports::Addr`, but only when talking, in code it will be very clear.

dignifiedquire: works for me

dignifiedquire: done

iroh-base/src/endpoint_addr.rs
```diff
@@ -39,75 +39,85 @@ use crate::{EndpointId, PublicKey, RelayUrl};
...
+    /// Relays
+    Relay(RelayUrl),
+    /// IP based addresses
+    Ip(SocketAddr),
 }
 
 impl EndpointAddr {
-    /// Creates a new [`EndpointAddr`] with no `relay_url` and no `direct_addresses`.
-    pub fn new(endpoint_id: PublicKey) -> Self {
+    /// Creates a new [`EndpointAddr`] with no addresses.
```
flub: Be verbose!  Explain that you can still use this using discovery.  Users will read this without any context and never have heard of iroh.  They'll have skipped all the nice intro docs we have (we do have those, right?)

iroh-base/src/endpoint_addr.rs
```diff
@@ -39,75 +39,85 @@ use crate::{EndpointId, PublicKey, RelayUrl};
...
-            direct_addresses: direct_addresses_iter.iter().copied().collect(),
+        let (id, relay_url, ip_addrs_iter) = value;
+        let mut addrs = BTreeSet::new();
+        if let Some(url) = relay_url {
+            addrs.insert(TransportAddr::Relay(url));
+        }
+        for addr in ip_addrs_iter {
+            addrs.insert(TransportAddr::Ip(*addr));
         }
+        EndpointAddr { id, addrs }
```
flub: whoa, do we use this impl?  I didn't even know it existed.  It kind of has a similar problem in that it assumes only 2 transports and a single relay.  Wasn't that something we wanted to remove?  Seems like `from_parts` is more generic?

iroh-base/src/ticket/endpoint.rs
```diff
@@ -27,35 +26,34 @@ use crate::{
...
 pub struct EndpointTicket {
-    node: EndpointAddr,
+    addr: EndpointAddr,
 }
 
 /// Wire format for [`EndpointTicket`].
 #[derive(Serialize, Deserialize)]
 enum TicketWireFormat {
-    Variant0(Variant0NodeTicket),
+    Variant0(Variant1EndpointTicket),
```
flub: Why bother adding `Variant1` if you then break `Variant0`?

iroh/src/magicsock.rs
```diff
@@ -329,7 +329,7 @@ impl MagicSock {
     ///
     /// [`Watcher`]: n0_watcher::Watcher
     /// [`Watcher::initialized`]: n0_watcher::Watcher::initialized
-    pub(crate) fn direct_addresses(&self) -> n0_watcher::Direct<BTreeSet<DirectAddr>> {
+    pub(crate) fn ip_addresses(&self) -> n0_watcher::Direct<BTreeSet<DirectAddr>> {
```
flub: This is a curious and tricky rename.  The type still says `DirectAddr`.  But it's also an internal API so whatever, not important enough to hold this up.  We can figure out this naming iteratively.

flub (commented):
Generally LGTM.

I'd like to treat the doc comments of `EndpointAddr` and `TransportAddr` and their fields as if the user encounters all those concepts for the first time.  Like explain that `endpoint_id` is unique to this node.  `addr` is a list of transient addresses, they may change, or may never have worked.  Discovery is a thing.  `TransprotsAddr::Relay` should give a short intro to relays, `TransportsAddr::Ip` explain that they may need holepuching and all such things.

iroh-base/src/endpoint_addr.rs
dignifiedquire: yes, agreed, should be removed

iroh-base/src/ticket/endpoint.rs
dignifiedquire: I wanted to rename them all to variant1 as this is the variant for 1.0? 🤷

iroh/src/magicsock.rs
dignifiedquire: yeah..that's where I ended up with as well, wasn't sure how much to change the types, or leave it be for now

iroh-base/src/endpoint_addr.rs
dignifiedquire: fixed

dignifiedquire:
minimal versions failing because of https://github.com/rustls/rustls-platform-verifier/issues/164

iroh-base/src/ticket/endpoint.rs
dignifiedquire: updated to be called `Variant1`

iroh-base/src/endpoint_addr.rs
```diff
@@ -18,20 +18,20 @@ use crate::{EndpointId, PublicKey, RelayUrl};
...
 ///
-/// - One or more *direct addresses* on which the endpoint might be reachable.  Depending on the
+/// - One or more *IP based addresses* on which the endpoint might be reachable.  Depending on the
 ///   network location of both endpoints it might not be possible to establish a direct
 ///   connection without the help of a [Relay server].
 ///
 /// This structure will always contain the required [`EndpointId`] and will contain an optional
-/// number of network-level addressing information.  It is a generic addressing type used
-/// whenever a connection to other endpoints needs to be established.
+/// number of other addressing information.  It is a generic addressing type used whenever a connection
```
flub:
```suggestion
/// number of other transport addresses.  It is a generic addressing type used whenever a connection
```
perhaps?  Not sure, but the previous read weird.

flub (approved):
looking forward to use this on the multipath branch!

Issue #3336, opened 2025-06-03
docs: remote_node_id fails on the accept side for a 0-rtt connection


rklaehn:
If you use Connecting::into_0rtt on a Connecting that comes from Endpoint::accept and get an  `Ok(Connection)`, calling `remote_node_id` immediately afterwards will fail.

This is independent of if the other side has actually tried or succeeded to connect via 0-rtt. I get this even with a fresh client that obviously can't use 0-rtt yet, and got it even before I did the .into_0rtt dance on the client side at all.

I manually tried, and it turns out that the first location where you can call remote_node_id is after calling `Connection::accept_bi`.

We need to document this behaviour so people don't get confused, and also need to keep remote_node_id returning a Result.

b5:
related to worke happening in #3123

matheus23:
I can't actually reproduce this. Which is very confusing to me. If you have the original code around, that'd be helpful :v:

matheus23:
Current suspicion is that this happening is timing-dependent, packet loss dependent, or dependent on the size of the initial datagram.

ramfox:
closed by #3556

PR #3556, opened 2025-10-17
feat!: make `Connection::remote_id` and `Connection::alpn` infallible

ramfox:
## Description

This PR refactors the connection API to make `Connection::remote_id()` and `Connection::alpn()` infallible. Previously, these methods could fail if called before the handshake completed or if the handshake data was unavailable. Now, `Connection` guarantees that it represents a fully authenticated connection with verified remote identity and ALPN protocol, since it now can only be constructed after successful handshake completion and authentication, eliminating the need for fallible accessors.

### 0-RTT API Improvements

The 0-RTT API has been restructured with clearer types and semantics:

- Use `Incoming::accept` to return an `Accepting`. Use `Accepting::into_0rtt` to return an `IncomingZeroRttConnection`

- Use `Connecting::into_0rtt` to return a `OutgoingZeroRttConnection`

- **`OutgoingZeroRttConnection`**: Represents client-side 0-RTT connections created via `Connecting::into_0rtt()`. Allows sending 0-RTT data before the handshake completes. Call `handshake_completed()` to get a `ZeroRttStatus` indicating whether the 0-RTT data was accepted or rejected by the server.

- **`IncomingZeroRttConnection`**: Represents server-side 0-RTT/0.5-RTT connections created via `Accepting::into_0rtt()`. Allows receiving 0-RTT data from clients or sending 0.5-RTT data before the handshake completes. Call `handshake_completed()` to get a fully authenticated `Connection`.

- **`ZeroRttStatus` enum**: Returned by `OutgoingZeroRttConnection::handshake_completed()` to indicate whether the server accepted or rejected the 0-RTT data:
  - `ZeroRttStatus::Accepted(Connection)`: 0-RTT data was accepted, streams opened before handshake remain valid
  - `ZeroRttStatus::Rejected(Connection)`: 0-RTT data was rejected, pre-handshake streams will error and data must be resent

These types replace the previous the previous version of `Connection` & the `ZeroRttAccepted` type and provide a more explicit API for handling 0-RTT connection states and outcomes.

### Error Handling

- Connection authentication failures now return `AuthenticationError` during the conversion between a `quinn:Connection` (which can exist before the handshake completes) and an `iroh::Connection`. This error may occur if the remote endpoint is not using the expected TLS cryptography or if the remote endpoint is a plain quinn connection. You may also see these errors after calling the `handshake_completed` method on `IncomingZeroRttConnection` or `OutgoingZeroRttConnection`, if the connection was closed before the handshake completed.
- New `ConnectingError` type for errors during handshake completion

## Breaking Changes

- `iroh`
    - `changed`
        - `ConnectError::Connection` - fields changed
        - `AcceptError::Connection` - fields changed
        - `AcceptError::MissingRemoteEndpointId` - fields changed
        - `AcceptError::NotAllowed` - fields changed
        - `AcceptError::User` - fields changed
        - `Connecting::into_0rtt` -> returns `Result<OutgoingZeroRttConnection, Connecting>`
    - `removed`
        - `ProtocolHandler::on_connecting()` removed - implement `on_accepting()` instead, which takes `Accepting` rather than `Connecting`
        - `DynProtocolHandler::on_connecting()` removed - implement `on_accepting()` instead
        - `iroh::endpoint::IncomingFuture` - use `Accepting` instead
        - `iroh::endpoint::ZeroRttAccepted` - replaced by explicit 0-RTT connection types

### DIFF
--- a/iroh/src/magicsock.rs
+++ b/iroh/src/magicsock.rs
@@ -149,8 +149,8 @@ pub(crate) struct Handle {
     msock: Arc<MagicSock>,
     // empty when shutdown
     actor_task: Arc<Mutex<Option<AbortOnDropHandle<()>>>>,
-    /// Token to cancel the actor task.
-    actor_token: CancellationToken,
+    /// Token to cancel the actor task and shutdown the relay transport.
+    shutdown_token: CancellationToken,
     // quinn endpoint
     endpoint: quinn::Endpoint,
 }
@@ -1392,17 +1392,22 @@ impl Handle {
         let my_relay = Watchable::new(None);
         let ipv6_reported = Arc::new(AtomicBool::new(ipv6_reported));
 
-        let relay_transport = RelayTransport::new(RelayActorConfig {
-            my_relay: my_relay.clone(),
-            secret_key: secret_key.clone(),
-            #[cfg(not(wasm_browser))]
-            dns_resolver: dns_resolver.clone(),
-            proxy_url: proxy_url.clone(),
-            ipv6_reported: ipv6_reported.clone(),
-            #[cfg(any(test, feature = "test-utils"))]
-            insecure_skip_relay_cert_verify,
-            metrics: metrics.magicsock.clone(),
-        });
+        let shutdown_token = CancellationToken::new();
+
+        let relay_transport = RelayTransport::new(
+            RelayActorConfig {
+                my_relay: my_relay.clone(),
+                secret_key: secret_key.clone(),
+                #[cfg(not(wasm_browser))]
+                dns_resolver: dns_resolver.clone(),
+                proxy_url: proxy_url.clone(),
+                ipv6_reported: ipv6_reported.clone(),
+                #[cfg(any(test, feature = "test-utils"))]
+                insecure_skip_relay_cert_verify,
+                metrics: metrics.magicsock.clone(),
+            },
+            shutdown_token.child_token(),
+        );
         let relay_transports = vec![relay_transport];
 
         let secret_encryption_key = secret_ed_box(&secret_key);
@@ -1531,12 +1536,9 @@ impl Handle {
         #[cfg(not(wasm_browser))]
         actor.update_direct_addresses(None);
 
-        let actor_token = CancellationToken::new();
-        let token = actor_token.clone();
-
         let actor_task = task::spawn(
             actor
-                .run(token, local_addrs_watch, sender)
+                .run(shutdown_token.child_token(), local_addrs_watch, sender)
                 .instrument(info_span!("actor")),
         );
 
@@ -1546,7 +1548,7 @@ impl Handle {
             msock,
             actor_task,
             endpoint,
-            actor_token,
+            shutdown_token,
         })
     }
 
@@ -1587,7 +1589,7 @@ impl Handle {
             return;
         }
         self.msock.closing.store(true, Ordering::Relaxed);
-        self.actor_token.cancel();
+        self.shutdown_token.cancel();
 
         // MutexGuard is not held across await points
         let task = self.actor_task.lock().expect("poisoned").take();
--- a/iroh/src/magicsock/transports/relay.rs
+++ b/iroh/src/magicsock/transports/relay.rs
@@ -13,7 +13,7 @@ use n0_future::{
 };
 use n0_watcher::{Watchable, Watcher as _};
 use tokio::sync::mpsc;
-use tokio_util::sync::PollSender;
+use tokio_util::sync::{CancellationToken, PollSender};
 use tracing::{Instrument, error, info_span, trace, warn};
 
 use super::{Addr, Transmit};
@@ -38,7 +38,7 @@ pub(crate) struct RelayTransport {
 }
 
 impl RelayTransport {
-    pub(crate) fn new(config: RelayActorConfig) -> Self {
+    pub(crate) fn new(config: RelayActorConfig, cancel_token: CancellationToken) -> Self {
         let (relay_datagram_send_tx, relay_datagram_send_rx) = mpsc::channel(256);
 
         let (relay_datagram_recv_tx, relay_datagram_recv_rx) = mpsc::channel(512);
@@ -48,7 +48,7 @@ impl RelayTransport {
         let my_endpoint_id = config.secret_key.public();
         let my_relay = config.my_relay.clone();
 
-        let relay_actor = RelayActor::new(config, relay_datagram_recv_tx);
+        let relay_actor = RelayActor::new(config, relay_datagram_recv_tx, cancel_token);
 
         let actor_handle = AbortOnDropHandle::new(task::spawn(
             async move {
--- a/iroh/src/magicsock/transports/relay/actor.rs
+++ b/iroh/src/magicsock/transports/relay/actor.rs
@@ -855,8 +855,8 @@ impl RelayActor {
     pub(super) fn new(
         config: Config,
         relay_datagram_recv_queue: mpsc::Sender<RelayRecvDatagram>,
+        cancel_token: CancellationToken,
     ) -> Self {
-        let cancel_token = CancellationToken::new();
         Self {
             config,
             relay_datagram_recv_queue,


iroh/src/endpoint.rs
```diff
@@ -1925,12 +2273,8 @@ impl Connection {
     }
 
     /// Extracts the ALPN protocol from the peer's handshake data.
-    pub fn alpn(&self) -> Option<Vec<u8>> {
-        let data = self.handshake_data()?;
-        match data.downcast::<quinn::crypto::rustls::HandshakeData>() {
-            Ok(data) => data.protocol,
-            Err(_) => None,
-        }
+    pub fn alpn(&self) -> Vec<u8> {
```
dignifiedquire: would be nicer to return a slice

iroh/src/endpoint.rs
```diff
@@ -1682,14 +1688,80 @@ impl Future for Connecting {
...
+) -> Result<(EndpointId, Vec<u8>), ConnectionError> {
+    // todo: fix error
+    let remote_id = remote_id(inner).map_err(|_| ConnectionError::LocallyClosed)?;
+    let alpn = alpn(&inner).ok_or(ConnectionError::LocallyClosed)?;
+    Ok((remote_id, alpn))
+}
+
+/// Returns the [`EndpointId`] from the peer's TLS certificate.
+///
+/// The [`PublicKey`] of an endpoint is also known as a [`EndpointId`].  This [`PublicKey`] is
```
flub:
```suggestion
/// The [`PublicKey`] of an endpoint is also known as an [`EndpointId`].  This [`PublicKey`] is
```

iroh/src/endpoint.rs
```diff
@@ -1715,6 +1787,280 @@ impl Future for ZeroRttAccepted {
...
 }
 
+/// A QUIC connection that can possible be 0-RTT
+#[derive(Debug, Clone)]
+pub struct ZRTTConnection {
+    inner: quinn::Connection,
+}
+
+/// copy-pasted from `Connection`
+impl ZRTTConnection {
```
flub: Could this have an `async fn into_connection(self) -> Connection` method?

iroh/examples/0rtt.rs
```diff
@@ -74,16 +80,46 @@ async fn pingpong(
     Ok(())
 }
 
-async fn pingpong_0rtt(connecting: Connecting, i: u64) -> n0_snafu::Result<Connection> {
+enum Conn {
+    ZRTT(ZRTTConnection),
+    Full(Connection),
+}
```
flub: What if you took an entirely different approach.  This ~~made~~ sketched out with my proposed `ZrttConnection::into_connection`:
```rust
async fn pingpong(mut send: SendStream, mut recv: RecvStream, x: u64) -> n0_snafu::Result<()> {
    let data = x.to_be_bytes();
    send.write_all(&data).await.e()?;
    send.finish().e()?;
    let echo = recv.read_to_end(8).await.e()?;
    assert_eq!(echo, data);
    Ok(())
}

async fn connect(args: Args) -> n0_snafu::Result<()> {
    let endpoint_addr = args.endpoint.unwrap().endpoint_addr().clone();
    let endpoint = iroh::Endpoint::builder()
        .relay_mode(iroh::RelayMode::Disabled)
        .keylog(true)
        .bind()
        .await?;
    let t0 = Instant::now();
    for i in 0..args.rounds {
        let t0 = Instant::now();
        let connecting = endpoint
            .connect_with_opts(endpoint_addr.clone(), PINGPONG_ALPN, Default::default())
            .await?;
        let connection = if args.disable_0rtt {
            let connection = connecting.await.e()?;
            trace!("connecting without 0-RTT");
            let (send, recv) = connection.open_bi().await.e()?;
            pingpong(send, recv, i).await?;
            connection
        } else {
            match connecting.into_0rtt() {
                Ok((zrtt_connection, zrtt_accepted_fut)) => {
                    trace!("0-RTT possible from our side");
                    let (send, recv) = zrtt_connection.open_bi().await.e()?;
                    let zrtt_task = tokio::spawn(pingpong(send, recv, i));
                    match zrtt_accepted_fut.await {
                        true => {
                            zrtt_task.await;
                            zrtt_connection.into_connection().await
                        }
                        false => {
                            zrtt_task.abort();
                            let conn = zrtt_connection.into_connection.await;
                            let (send, recv) = conn.open_bi().await.e()?;
                            pingpong(send, recv, i).await?;
                            conn
                        }
                    }
                }
                Err(connecting) => {
                    trace!("0-RTT not possible from our side");
                    let conn = connecting.await.e()?;
                    let (send, recv) = conn.open_bi().await.e()?;
                    pingpong(send, recv, i).await?;
                    conn
                }
            }
        };
        tokio::spawn(async move {
            // wait for some time for the handshake to complete and the server
            // to send a NewSessionTicket. This is less than ideal, but we
            // don't have a better way to wait for the handshake to complete.
            tokio::time::sleep(connection.rtt() * 2).await;
            connection.close(0u8.into(), b"");
        });
        let elapsed = t0.elapsed();
        println!("round {i}: {} us", elapsed.as_micros());
    }
    let elapsed = t0.elapsed();
    println!("total time: {} us", elapsed.as_micros());
    println!(
        "time per round: {} us",
        elapsed.as_micros() / (args.rounds as u128)
    );
    Ok(())
}
```

It's kind of sad to use a task for this, but it was so much easier to express this way.  Maybe there's a nice readable way to express this without a task that I didn't think off.

flub: To explain what I'm trying here: remove the need for the `Conn` enum by passing streams directly to the 0-rtt using code-path.

ramfox: yes this seems like a good alternative

ramfox:
Still some open questions around the server side of the ZeroRttConnection!

iroh/examples/0rtt.rs
```diff
@@ -153,7 +142,7 @@ async fn accept(_args: Args) -> n0_snafu::Result<()> {
         while let Some(incoming) = endpoint.accept().await {
             tokio::spawn(async move {
                 let connecting = incoming.accept().e()?;
-                let (connection, _zero_rtt_accepted) = connecting
+                let connection = connecting
                     .into_0rtt()
                     .expect("accept into 0.5 RTT always succeeds");
```
matheus23: What do y'all think of having an `Incoming::accept_0rtt` that doesn't return a `Result`, thus wouldn't need this annoying `.expect("accept into 0.5 RTT always succeeds");`?

iroh/src/endpoint.rs
```diff
@@ -1505,21 +1505,74 @@ impl Future for IncomingFuture {
...
+}
+
 /// In-progress connection attempt future
 #[derive(derive_more::Debug)]
 #[pin_project]
 pub struct Connecting {
     #[pin]
     inner: quinn::Connecting,
     ep: Endpoint,
+    /// exists this is an outgoing connection, `None` if this is an incoming conn
```
matheus23:
```suggestion
    /// `Some(remote_id)` if this is an outgoing connection, `None` if this is an incoming conn
```

iroh/src/endpoint.rs
```diff
@@ -1505,21 +1505,74 @@ impl Future for IncomingFuture {
...
+///
+/// [`PublicKey`]: iroh_base::PublicKey
+pub fn remote_id_from_quinn_conn(
+    conn: &quinn::Connection,
+) -> Result<EndpointId, RemoteEndpointIdError> {
+    let data = conn.peer_identity();
+    match data {
+        None => {
+            warn!("no peer certificate found");
+            Err(RemoteEndpointIdSnafu.build())
```
matheus23: I actually think it's impossible for this to fail.
There's [this issue](https://github.com/n0-computer/iroh/issues/3336), but I can't reproduce that.

We should perhaps do a little more testing, but it might be valid to return `EndpointId` here instead of `Result`.

iroh/src/endpoint.rs
```diff
@@ -1689,9 +2034,24 @@ impl Future for ZeroRttAccepted {
...
+}
+
+impl TryFrom<quinn::Connection> for Connection {
+    type Error = ConnectionError;
+
+    fn try_from(value: quinn::Connection) -> Result<Self, Self::Error> {
+        Ok(Self {
+            remote_id: remote_id_from_quinn_conn(&value)
+                .map_err(|_| ConnectionError::LocallyClosed)?,
+            alpn: alpn_from_quinn_conn(&value).ok_or(ConnectionError::LocallyClosed)?,
```
matheus23: These errors seem a little arbitrary to me. `remote_id_from_quinn_conn` seems to work no matter if the underlying connection was closed or not.
If `alpn_from_quinn_conn` would fail, it should probably fail at a different point: On `Incoming::accept` or when `IncomingFuture` resolves, we should check if an ALPN is present and fail if not.

And then I think it *might* be nice to not store the `remote_id` and `alpn` twice (once in `quinn::Connection` and once in `iroh::Connection`), and instead always look it up via infallible `remote_id_from_quinn_conn` and `alpn_from_quinn_conn` versions. (Although this last comment might be a little bit controversial. If you disagree with me I wouldn't push back :sweat_smile: )

ramfox: Okay, thank you this answers one of the things that had been bothering me: is it even possible for these conversions to fail at this point. It made sense to me that if the crypto was invalid then how would we even get here?

If that's true, it simplifies a lot more.

Does the remote_id and alpn get stored on the `quinn::Connection` I thought the whole issue was that we had to do this conversion every time by calling `peer_identity` & `handshake_data`

iroh/examples/0rtt.rs
flub: Oh yes!  That'd be great

iroh/src/endpoint.rs
ramfox: Isn't the issue that it could possibly fail if you did 0-rtt and you haven't finished the handshake yet? Which we have engineered here to not be the case, true.

At this point in the creation of a connection, is it possible that the downcasts below fail? or that we don't have enough certs? Wouldn't we have already "caught" this if it didn't have the kinds of certificates we were expecting?

flub: Thank you so much to dig into this!  That's what we need to figure out here.

iroh/examples/0rtt.rs
ramfox: ooo yes.

iroh/src/endpoint.rs
matheus23: I mean, yeah. By "get stored" I mean that the bits of information required to derive the ALPN and remote ID are there.

iroh/src/endpoint.rs
matheus23: > Isn't the issue that it could possibly fail if you did 0-rtt and you haven't finished the handshake yet?

Hmmm. I think you might be right. `Incoming` can potentially only be the first datagram of a bunch, and you can get a `Connecting` from that synchronously, and from there get to `quinn::Connection` via `into_0rtt` synchronously as well.
I guess this doesn't really come up because our initial packets pretty much always fit into 1200 byte datagrams, but if for some reason that initial datagram doesn't contain all packets (e.g. because someone uses post-quantum keys), then we're toast.

iroh/src/endpoint.rs
ramfox: okie dokie. So first, I removed the `TryFrom` impl. I realized that trait impls are public and I don't want users to attempt to do this.

Second, I moved this into a private `conn_from_quinn_conn_dangerous` method that will panic (via `.expect("handshake has completed")`) if you attempt to run it before a connection has completed its handshake.

Third. I have a semi-strong opinion about just storing the field in the `Connection`.

ramfox:
Okay, this latest version has `ZeroRttClientConnection`, `ZeroRttServerConnection`, `ZeroRtt::Accepted` and `ZeroRtt::Rejected`.

Besides bikeshedding the names `ZeroRttClientConnection` and `ZeroRttServerConnection`, how do we feel about this?

I think we either go this route, or a route that combines both sides into `ZeroRttConnection` & the annoying bit is that the server side also returns a `ZeroRtt` rather than just a `Connection`.

iroh/src/endpoint.rs
```diff
@@ -1416,6 +1416,48 @@ impl Incoming {
         })
     }
 
+    /// Turns this into a connection that can accept 0-rtt data.
```
dignifiedquire: can we please split up this file at this point 💀

ramfox: done

ramfox:
Separated everything about `Connection` (`Connecting`, `Accept`, `Incoming`, any 0-rtt stuff) into `iroh/src/endpoint/connection.rs`

All structs are still exported from the `endpoint` mod so it's just an internal organizational change.

iroh/examples/0rtt.rs
```diff
@@ -106,10 +71,36 @@ async fn connect(args: Args) -> n0_snafu::Result<()> {
...
+                        ZeroRtt::Accepted(conn) => {
+                            let _ = zrtt_task.await.e()?;
+                            conn
+                        }
+                        ZeroRtt::Rejected(conn) => {
+                            zrtt_task.abort();
+                            let (send, recv) = conn.open_bi().await.e()?;
+                            pingpong(send, recv, i).await?;
+                            conn
+                        }
```
matheus23: Can we revert these changes to the example?
I personally find this version harder to understand:
1. We introduce more parallelism via a `tokio::spawn` that we didn't have before
2. Before, we checked whether our "ping sending" went through, and if it didn't we send it again after the handshake completed. But now, we also duplicate the `recv.read_to_end`. I don't know what that means for spawned task. Does that just hang another recv future?

I do really think I'd prefer the version where we check the "proceed" future in the `pingpong` function.

Perhaps I'm missing something though, I'm sure you've changed it because you found the old version confusing. Perhaps we can find a better third version if you explain what was confusing about the original example code?

iroh/examples/0rtt.rs
```diff
@@ -48,48 +46,15 @@ pub fn get_or_generate_secret_key() -> n0_snafu::Result<SecretKey> {
 /// We send the data on the connection. If `proceed` resolves to true,
 /// read the response immediately. Otherwise, the stream pair is bad and we need
 /// to open a new stream pair.
-async fn pingpong(
```
matheus23, line 51: At the moment the docs don't match the implementation. But IMO the implementation should be changed, not the docs (also see my other comment)

iroh/src/endpoint.rs
```diff
@@ -2739,56 +2072,26 @@ mod tests {
...
-
-                    // Handle stream errors gracefully
-                    let (mut send, mut recv) = match conn.accept_bi().await {
-                        Ok(s) => s,
-                        Err(e) => {
-                            tracing::warn!("Failed to accept bi stream: {e:?}");
-                            continue;
+                            };
+                            let stream_task = tokio::spawn(handle_stream(send, recv));
+                            (z_rtt.handshake_completed().await, stream_task)
```
matheus23: Can avoid spawning a task here, too?
I don't think it's necessary.
The handshake will continue in the background whether we drive the handshake completed future or not. It's mostly a notification.

Again, you might have a reason for doing this. Please let me know :)

iroh/src/endpoint/connection.rs
```diff
@@ -0,0 +1,1415 @@
...
+//!
+//! The [`Connection`] is how you send data to and receive data from the remote endpoint.
+//!
+//! There are many transitions states between attempting to start a connection and
+//! receiving a cryptographically secure connection.
+//!
+//! The main items in this module are:
+//!
+//! - [`Connection`] to create streams to talk to a remote endpoint.
+//! - [`Connecting`] to intercept a connection before it is fully formed.
```
matheus23:
```suggestion
//! - [`Connecting`] for operating on connections that haven't finished their handshake yet.
```

I feel like "intercept" might give the wrong impression (think intercepting "foreign" connections).

iroh/src/endpoint/connection.rs
```diff
@@ -0,0 +1,1415 @@
...
+//! - [`Incoming`] to accept or reject an incoming connection.
+//! - [`ZeroRttClientConnection`] to attempt to send 0-RTT data before the cryptographic
+//!   handshake has completed.
+//! - [`ZeroRttServerConnection`] to attempt to read 0-RTT data before the cryptographic
+//!   handshake has completed.
```
matheus23:
```suggestion
//! - [`ZeroRttServerConnection`] to attempt to read 0-RTT or send 0.5-RTT data before the cryptographic
//!   handshake has completed.
```

iroh/src/endpoint/connection.rs
```diff
@@ -0,0 +1,1415 @@
...
+///
+/// Only use this when the given `quinn::Connection` has definitely completed
+/// it's cryptographic handshake.
+fn conn_from_quinn_conn_dangerous(conn: quinn::Connection) -> Connection {
+    Connection {
+        remote_id: remote_id_from_quinn_conn(&conn).expect("handshake has completed"),
+        alpn: alpn_from_quinn_conn(&conn).expect("handshake has completed"),
+        inner: conn,
+    }
+}
```
matheus23: Small nit: It might be nicer to move the `.expect("handshake has completed")` up into the use sites of `conn_from_quinn_conn_dangerous`, as from the use site's context it's more obvious that that's the case.

iroh/src/endpoint/connection.rs
```diff
@@ -0,0 +1,1415 @@
...
+impl Future for IncomingFuture {
+    type Output = Result<Connection, ConnectionError>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut std::task::Context<'_>) -> Poll<Self::Output> {
+        let this = self.project();
+        match this.inner.poll(cx) {
+            Poll::Pending => Poll::Pending,
+            Poll::Ready(Err(err)) => Poll::Ready(Err(err)),
+            Poll::Ready(Ok(inner)) => {
+                let conn = conn_from_quinn_conn_dangerous(inner);
```
matheus23: Actually, I think this can still fail.
E.g. imagine someone running an iroh endpoint publicly, and someone connects to it via quinn without providing an ALPN.
I don't think we check whether an ALPN is present before this call (and I actually think right here is exactly the right point at which we should check this).

So basically, can we turn an error from `conn_from_quinn_conn` into a `ConnectionError`?
Basically just a protocol violation (protocol being "whatever is connecting doesn't respect the iroh connection protocol? requirements?" sth like that).

It's kind of important that we don't panic here. Otherwise anyone out there with a modified iroh will be able to DoS another node that `.accept`s their `Incoming`.
We still get the benefit of infallibility later in `Connection::remote_id` and `Connection::alpn`! :)

iroh/src/endpoint/connection.rs
```diff
@@ -0,0 +1,1415 @@
...
+    /// Converts this [`Connecting`] into a 0-RTT or 0.5-RTT connection at the cost of weakened
+    /// security.
+    ///
+    /// Returns `Ok` immediately if the local endpoint is able to attempt sending 0/0.5-RTT data.
+    /// If so, the returned [`ZeroRttClientConnection`] can be used to send application data without waiting for
+    /// the rest of the handshake to complete, at the cost of weakened cryptographic security
+    /// guarantees.
+    ///
+    /// This method will error if attempted on an incoming [`Connecting`]. To accept 0-RTT data,
+    /// use the [`Incoming::accept_0rtt`] method.
```
matheus23: Two things:
1. The docs here kind of contradict themselves. Above it still mentions 0.5-RTT which only makes sense on the server side, but later below we mention it not working on incoming connections.
2. Not allowing this on incoming connections actually reminds me that this won't work for our current `ProtocolHandler` trait, since that only gives you `Connecting`, not `Incoming`. But perhaps that's fine? Not sure. Since I guess ideally e.g. iroh-blobs can support 0-RTT without requiring the user to create their own router. (But also we can't know the ALPN from the `Incoming` :grimacing: )

About (2):
Sorry for the back and forth on this. It's a lot of moving things and I've just realized this.
I guess we have two options: Either we provide two fallible (possibly panicking) fns, `into_0rtt_server` and `into_0rtt_client`, or we duplicate the `Connecting` struct into `ConnectingClient` and `ConnectingServer` (or similar names).
I actually don't hate the latter, given `Connecting` doesn't have that many fns.

iroh/src/endpoint/connection.rs
ramfox: Yea, I went back and forth on this, which I why I added the `dangerous` ending to the name. Since I'm 50/50 and this is another vote in the "call expect at the call site" column, I'll change it.

iroh/src/endpoint/connection.rs
ramfox: Okay great point, I expected that if iroh expects an alpn and the other side does not provide one it would have errored already. Isn't this something that is taken care of by the handshake? Or does tls/quinn always allow a no-alpn fallback?

matheus23: Good Qs. I *think* quinn allows a bunch of things, including not having an ALPN. 
Quinn definitely allows the client to omit authentication. We return `true` in `ClientCertificateVerifier::offer_client_auth`, but I'm not sure that that means that we *require* client auth (I guess given the naming it really doesn't mean that).
So I think it's possible that after a handshake, the server might have a connection that doesn't have a remote endpoint ID for the client in its handshake data.

iroh/src/endpoint/connection.rs
ramfox: oh this is a great point.

let me try something and see if it sucks

iroh/src/endpoint/connection.rs
ramfox: wait do you mean endpoint ID or alpn? Because if it's endpoint ID then all of this is work is moot 😅

iroh/src/endpoint.rs
ramfox: yes thank you, i changed it back. I think it was a hold over from some weird point in the api changes.

iroh/examples/0rtt.rs
ramfox: Yea, the issue is we no longer have a `Connection` for both cases, we have a `Connection` for the rejected side and a `ZeroRttClientConnection` on the accepted side. 

see this comment: https://github.com/n0-computer/iroh/pull/3556#discussion_r2444622077

But this also may have gotten less gross with the recent changes. Let me see.

iroh/src/endpoint/connection.rs
```diff
@@ -356,6 +314,52 @@ pub enum AlpnError {
...
+    /// Returns a `ZeroRttServerConnection`.
+    ///
+    /// Panics if the `ZeroRttConnection` is not an incoming connection.
+    pub fn unwrap_incoming(self) -> ZeroRttServerConnection {
+        match self {
+            ZeroRttConnection::Outgoing(_) => unreachable!("expected Incoming Connection"),
+            ZeroRttConnection::Incoming(res) => res,
+        }
+    }
+}
```
matheus23: Hm. This is an interesting way to avoid doubling the `Connecting` type.
However... I'm not sure I really like it?
The problem is mainly that these `unreachable!`s aren't really... true. A user just has to call the wrong `unwrap_*` and they will reach the panic.

Perhaps another suggestion: We could add a type parameter to `Connecting` that's either `Out` or `In`.
Something like this:
```rs
pub struct Out;
pub struct In;

pub struct Connecting<Direction> { .. }

impl Connecting<Out> {
    pub fn into_0rtt_outgoing(self) -> Result<ZeroRttClientConnection, Connecting> { .. }
}

impl Connecting<In> {
    pub fn into_0rtt_incoming(self) -> ZeroRttServerConnection { .. }
}
```

wdyt?

iroh/src/endpoint/connection.rs
ramfox: Can you take a look at the latest changes.

Rather than separating out `Connecting` I instead made the result of `into_0rtt` more structured, but more complicated & removed `Incoming::accept_0rtt`.

We have `Connecting` in so many places in the code, including the protocol handler stuff, I think it would be better to keep the happy path as simple as possible and force the complication down to isolate it in the 0rtt codepath.

Anyways, I don't know how much better this is than just having one `ZeroRttConnection` for both incoming and outgoing BUT it is more structured 😂 and so forces folks to confront the different aspects of 0-rtt before using it

iroh/src/endpoint/connection.rs
ramfox: > The problem is mainly that these unreachable!s aren't really... true. A user just has to call the wrong unwrap_* and they will reach the panic.

So maybe "unreachable" isn't right, maybe it should just be a straight `panic`, but that's the point. If the user is calling the wrong thing, then they need to read the docs to make sure they understand how to actually use 0-rtt. Or we remove the `unwrap_*` and the user need to contend with matching on every `into_0rtt` response.

I have a hard time wanting to change `Connecting`, it would change some majorly used APIs that I think we need more buy in from the whole team on, Eg `connect_with_opts` and `on_connecting` for the protocol handlers.

matheus23: > I have a hard time wanting to change `Connecting`, it would change some majorly used APIs that I think we need more buy in from the whole team on, Eg `connect_with_opts` and `on_connecting` for the protocol handlers.

Is it that bad though? Basically `connect_with_opts` would return `Connecting<Out>` and `on_connecting` would give you a `Connecting<In>`. :point_right: :point_left:

matheus23: We changed this to use separated types on both ends: `Accepting` and `Connecting`

iroh/src/endpoint/connection.rs
```diff
@@ -0,0 +1,1721 @@
...
+
+#[allow(missing_docs, dead_code)]
+#[common_fields({
+    backtrace: Option<snafu::Backtrace>,
+})]
+#[derive(Debug, Snafu)]
+#[non_exhaustive]
+pub enum QuinnConnectionError {
```
matheus23: Would love if we could rename this. It has to be exposed to the user, but the user might not really know what "Quinn" is.

matheus23: I know we like to make all errors as specific as possible, but at this point we have `RemoteEndpointIdError`, `QuinnConnectionError` and `ConnectingError`.
Would it be insane to consolidate all of them into one `HandshakingError` or something similar? What do others think?

Frando: Or maybe just `ConnectError`?

matheus23: I think we already have a `ConnectError`, unfortunately.

ramfox: Changed to `AuthenticationError`

PR #3561, opened 2025-10-20
refactor!: convert from snafu to n0-error

Frando:
## Description

Changes the error handling across all crates in this repo from `snafu`/`n0-snafu` to **`n0-error`**. The latter is our new error library which features call-site location tracking and better ergonomics than snafu.

Repo: https://github.com/n0-computer/n0-error
Docs: https://docs.rs/n0-error/

*Note:*  The initial version of this PR was written by an AI (OpenAI codex) which I instructed in numerous prompts to convert iroh from n0-snafu/snafu to n0-error. It has since been cleaned up and reviewed line-by-line.

## Breaking Changes

*Changed*
* All error structs and enum variants now have a field `meta: n0_error::Meta`. Fields `backtrace` and `span_trace` are removed.
* All error structs and enums no longer implement `snafu::ErrorCompat`. They now implement `n0_error::StackError`. All our errors keep implementing `std::error::Error`. For downstream uses of these errors, not much changed.
* If you previously used `n0-snafu` to work with iroh errors, we recommend you switch to `n0-error`. See the [docs]( https://docs.rs/n0-error/) for how to work with n0-error.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh-base/src/key.rs
```diff
@@ -384,14 +380,14 @@ fn decode_base32_hex(s: &str) -> Result<[u8; 32], KeyParsingError> {
...
-            return Err(DecodeInvalidLengthSnafu.build());
+            return Err!(KeyParsingError::DecodeInvalidLength);
         }
         data_encoding::BASE32_NOPAD.decode_mut(input, &mut bytes)
     };
     match res {
         Ok(len) => {
             if len != PublicKey::LENGTH {
-                return Err(DecodeInvalidLengthSnafu.build());
+                return Err!(KeyParsingError::DecodeInvalidLength);
```
dignifiedquire: looking at this, I would much rather write

```rust
bail!(KeyParsingError::DecodeInvalidLength);
```

overloading `Err`  I find problematic personally

iroh-base/src/key.rs
```diff
@@ -128,7 +127,7 @@ impl PublicKey {
     pub fn verify(&self, message: &[u8], signature: &Signature) -> Result<(), SignatureError> {
         self.as_verifying_key()
             .verify_strict(message, &signature.0)
-            .map_err(|_| SignatureSnafu.build())
+            .map_err(|_| e!(SignatureError))
```
dignifiedquire: does this have to be a macro, or could this be a method on `SignatureError`?

iroh-base/src/ticket.rs
```diff
@@ -50,7 +49,7 @@ pub trait Ticket: Sized {
     fn deserialize(str: &str) -> Result<Self, ParseError> {
         let expected = Self::KIND;
         let Some(rest) = str.strip_prefix(expected) else {
-            return Err(KindSnafu { expected }.build());
+            return Err(ParseError::wrong_prefix(expected));
```
dignifiedquire: shouldn't this be called `ParseError::kind(expected)`?

dignifiedquire: nvm, found the impl below

dignifiedquire: could we have a macro so we can write

```rust
bail!(ParseError::Kind { expected });
```
without the  indirection method?

iroh-dns-server/src/config.rs
```diff
@@ -173,7 +173,7 @@ impl Config {
             PathBuf::from(val)
         } else {
             let path = dirs_next::data_dir()
-                .context("operating environment provides no directory for application data")?;
+                .std_context("operating environment provides no directory for application data")?;
```
dignifiedquire, line 176: I think this works reasonably well, a bit confusing at first, but easy to use at the end of the day

iroh-dns-server/src/http.rs
```diff
@@ -173,11 +172,11 @@ impl HttpServer {
                 Err(err) if err.is_cancelled() => {}
                 Ok(Err(err)) => {
                     warn!(?err, "task failed");
-                    final_res = Err(err).context("task");
+                    final_res = Err(err).std_context("task");
```
dignifiedquire: could we have something like 

```rust
final_res = whatever_std!(err, "task");
```

iroh-relay/src/client.rs
```diff
@@ -220,10 +236,9 @@ impl ClientBuilder {
                 _ => "wss",
             })
             .map_err(|_| {
-                InvalidWebsocketUrlSnafu {
-                    url: dial_url.clone(),
-                }
-                .build()
+                e!(ConnectError::InvalidWebsocketUrl {
```
dignifiedquire, line 232: this macro fills in the location details?

iroh-relay/src/client/conn.rs
```diff
@@ -146,9 +139,15 @@ impl Sink<ClientToRelayMsg> for Conn {
 
     fn start_send(mut self: Pin<&mut Self>, frame: ClientToRelayMsg) -> Result<(), Self::Error> {
         let size = frame.encoded_len();
-        snafu::ensure!(size <= MAX_PACKET_SIZE, ExceedsMaxPacketSizeSnafu { size });
+        n0_error::ensure!(
+            size <= MAX_PACKET_SIZE,
+            n0_error::e!(SendError::ExceedsMaxPacketSize { size })
```
dignifiedquire: could we make the macro such that we don't need to call `e!` inside?

iroh-relay/src/client/tls.rs
```diff
@@ -144,14 +144,15 @@ impl MaybeTlsStreamBuilder {
...
+        let port = url_port(&self.url).ok_or_else(|| e!(DialError::InvalidTargetPort))?;
         let addr = SocketAddr::new(dst_ip, port);
 
         debug!("connecting to {}", addr);
         let tcp_stream = time::timeout(DIAL_ENDPOINT_TIMEOUT, async move {
             TcpStream::connect(addr).await
         })
-        .await??;
+        .await
+        .map_err(|err| e!(DialError::Timeout, err))??;
```
dignifiedquire, line 155: what does this variant do?

iroh-relay/src/client/tls.rs
```diff
@@ -240,22 +245,26 @@ impl MaybeTlsStreamBuilder {
...
-            .build());
+            return Err!(DialError::ProxyConnectInvalidStatus {
+                status: res.status()
+            });
         }
 
-        let upgraded = hyper::upgrade::on(res).await.context(ProxyConnectSnafu)?;
+        let upgraded = hyper::upgrade::on(res)
+            .await
+            .map_err(|err| e!(DialError::ProxyConnect, err))?;
```
dignifiedquire, line 267: can't we do this with one of the `context` variants?

iroh-relay/src/dns.rs
```diff
@@ -62,80 +61,68 @@ pub trait Resolver: fmt::Debug + Send + Sync + 'static {
...
-    }
+/// Aggregates all the lookup errors when all staggered attempts fail.
+#[add_meta]
+#[derive(Error)]
+pub struct LookupStaggeredError {
+    #[display("no calls succeeded: [{}]", errors.iter().map(|e| e.to_string()).collect::<Vec<_>>().join(""))]
+    errors: Vec<LookupError>,
 }
 
+// no manual new; use e!(Type { errors })
```
dignifiedquire: random comment

iroh-base/src/key.rs
Frando: Yes, I can add such a `bail!` macro. Without `bail!` or the`Err!` macro it would be 
`Err(e!(KeyParsingError::DecodeInvalidLength))`

iroh-base/src/key.rs
Frando: Right, this is a struct so it can just be `SignatureError::new()` in this case. However, for enums I removed the constructors from the proc macro generated code (I had them before) so that jump-to-definition on variants works.

iroh-base/src/ticket.rs
Frando: Yeah, this was just kept from existing code. It would either be `bail` or `return Err!(ParseError::Kind { expected })` or `return Err(e!(ParseError::Kind { expected }))`. Which one we want is to be bikeshedded.

iroh-dns-server/src/config.rs
Frando: To me, this is the most annoying thing about this whole endevour, but unfortunately I don't think there's another solution. I made the `std_context` method the "longer" one (to type) to nudge towards using `context` wherever possible (ie. on our errors that implement StackError) because `std_context` will work on our errors too, but we'll lose the location info in the process. This is the much discussed issue that as long as we implement std::error::Error for our errors, we can't forbid calling the StdResultExt  methods on them, even though we *should*  always use the StackResultExt methods (which are only available on StackError results, not on std error results).

iroh-dns-server/src/http.rs
Frando: We can. Here it even can be the same macro for both StackError and std error, because we can use autoref specialization.

There is already a `anyerr!` macro that makes use of autoref specialization to not have to worry about whether the error is std or StackError. With this it would be `Err(anyerr!(err).context("task"))`. I can expand the macro to support context as second arg, so it could be `Err(anyerr!(err, "task"))`. Will do that.
Not sure if we need a variant that returns `Err(err)`, I think it's fine to write that by hand (like anyhow::anyhow!, it also returns the error not a result).

iroh-relay/src/client.rs
Frando: Yes. It expands to `ConnectError::InvalidWebsocketUrl {  url: dial_url.clone(), meta: Default::default() }`

iroh-relay/src/client/conn.rs
Frando: We can but then it wouldn't be usable with other error types anymore as it would add a `meta` field and if that doesn't exist it would give a compile error. We could add another macro `ensure_e!` or such, but not sure if that is really better than the current one. Will change to import `e` though, then it's much shorter already.

iroh-relay/src/client/tls.rs
Frando: Expand to `DialError::Timeout { source: err, meta: Meta::default() }`
See docs: https://n0-computer.github.io/n0-error/pr/4/docs/n0_error/macro.e.html

iroh-relay/src/client/tls.rs
Frando: `context` takes a string (or, impl Display) and adds this as context. We can't do without the `e!` macro if we want jump-to-usage on `ProxyConnect` to work and not type `meta: Default::default()` manually. So not sure what `context` should do here?

iroh-relay/src/dns.rs
Frando: removed

iroh-base/src/key.rs
dignifiedquire: I think that would be better to have than `Err!`

iroh-relay/src/client/tls.rs
dignifiedquire: yeah, the filling in the struct wouldn't be possible :(

Frando:
*Note: this is a previous version, see next comment instead*

Some examples when tests are failing:

```
Error: Unable to connect to remote (at iroh/src/endpoint.rs:633:26)
Caused by:
    0: Unable to connect to remote (at iroh/src/endpoint.rs:735:28)
    1: endpoint stopping
```
Here we get the same message twice, because the outer error delegates to an inner error with `#[transparent]`.

```
Failed to create bind sockets (at iroh/src/endpoint.rs:202:21)
Caused by:
    0: Failed to create bind sockets (at iroh/src/magicsock.rs:1389:31)
    1: failed to bind
```
Same: the outer error delegates to the next with `transparent`, so same message twice.

Another one:
```
Error: No addressing information available (at iroh/src/endpoint.rs:699:28)
Caused by:
    0: Discovery service failed (at iroh/src/endpoint.rs:1291:39)
    1: Discovery produced no results for b371e8d4c1 (at iroh/src/discovery.rs:645:21)
```

Frando:
with the recent changes we now have these messages

Regular error formatting, without `RUST_BACKTRACE=1`:
```
Error: No addressing information available
Caused by:
    Discovery service failed
    Discovery produced no results for b371e8d4c1
```
Changes to previous version: I removed the numbers in front of the causes. Do we want them? To me it looks a bit less cluttered without the numbers. But no strong opinion at all, happy to add them back if others prefer to have them.


Errors that involve a `#[error(transparent)]` in the chain:

Without `RUST_BACKTRACE=1`
```
Error: Unable to connect to remote
Caused by:
    endpoint stopping
```
Changes to previous version: Transparent errors are skipped, as intended.

With `RUST_BACKTRACE=1`:
```
Error: ConnectError::Connect (iroh/src/endpoint.rs:633:26)
Caused by:
    Unable to connect to remote (iroh/src/endpoint.rs:735:28)
    endpoint stopping
```
When `RUST_BACKTRACE=1` is set, we now *do* print the errors marked as transparent, with the full name of the enum variant and call-site location. We use the enum variant name, because for transparent error variants there is no display message for the error. Other than rust's default derive debug format, I included the enum name, because only the variant oftenly reads weird as an error message. Can change if preferred though.

iroh-dns-server/src/http.rs
Frando: Replaced with `Err(anyerr!(err, "context"))`

iroh-base/src/key.rs
Frando: Converted it to `Err(e!(..))` and removed the `Err!` macro. A `bail!` macro is now also available, but I don't really prefer using it all over the place. We can decide later if and when we want to use it.

iroh-relay/src/client/conn.rs
Frando: I changed the macro: `ensure!` now expects the syntax supported by `e!` as the second argument.

iroh-dns-server/src/store/signed_packets.rs
```diff
@@ -145,30 +145,32 @@ impl Actor {
...
+                                    Some(existing) => {
+                                        if existing.more_recent_than(&packet) {
+                                            res.send(false).ok();
+                                            continue;
+                                        } else {
+                                            // remove the old packet from the update time index
+                                            tables.update_time.remove(&existing.timestamp().to_bytes(), key.as_bytes()).anyerr()?;
+                                            true
+                                        }
+                                    } _ => {
```
dignifiedquire: odd formatting

iroh-relay/src/client.rs
```diff
@@ -272,10 +279,9 @@ impl ClientBuilder {
         let (conn, response) = builder.connect_on(stream).await?;
 
         if response.status() != hyper::StatusCode::SWITCHING_PROTOCOLS {
-            UnexpectedUpgradeStatusSnafu {
-                code: response.status(),
-            }
-            .fail()?;
+            return Err(e!(ConnectError::UnexpectedUpgradeStatus {
```
dignifiedquire: could use `ensure`

iroh-relay/src/client/conn.rs
```diff
@@ -146,9 +135,12 @@ impl Sink<ClientToRelayMsg> for Conn {
...
     fn start_send(mut self: Pin<&mut Self>, frame: ClientToRelayMsg) -> Result<(), Self::Error> {
         let size = frame.encoded_len();
-        snafu::ensure!(size <= MAX_PACKET_SIZE, ExceedsMaxPacketSizeSnafu { size });
+        n0_error::ensure!(
+            size <= MAX_PACKET_SIZE,
+            SendError::ExceedsMaxPacketSize { size }
+        );
         if let ClientToRelayMsg::Datagrams { datagrams, .. } = &frame {
-            snafu::ensure!(!datagrams.contents.is_empty(), EmptyPacketSnafu);
+            n0_error::ensure!(!datagrams.contents.is_empty(), SendError::EmptyPacket);
```
dignifiedquire: import `ensure`

Frando: fixed

iroh-relay/src/client.rs
Frando: changed

iroh-dns-server/src/store/signed_packets.rs
Frando: fixed

iroh/src/magicsock.rs
```diff
@@ -1687,8 +1690,9 @@ impl DiscoState {
     ) -> Result<disco::Message, DiscoBoxError> {
         let mut sealed_box = sealed_box.to_vec();
         self.get_secret(endpoint_id, |secret| secret.open(&mut sealed_box))
-            .context(OpenSnafu)?;
-        disco::Message::from_bytes(&sealed_box).context(ParseSnafu)
+            .map_err(|source| e!(DiscoBoxError::Open { source }))?;
+        disco::Message::from_bytes(&sealed_box)
+            .map_err(|source| e!(DiscoBoxError::Parse { source }))
```
dignifiedquire, line 1680: what's the difference between 

```rust
e!(DiscoBoxError::Parse { source })
```
and

```rust
e!(DiscoBoxError::Parse, source)
```

iroh/src/endpoint.rs
```diff
@@ -2992,19 +2968,26 @@ mod tests {
...
-            let mut uni = conn.accept_uni().await.e()?;
-            uni.read_to_end(10).await.e()?;
+            let conn = server
+                .accept()
+                .await
+                .anyerr()?
+                .accept()
+                .anyerr()?
+                .await
+                .anyerr()?;
```
dignifiedquire: 😭 😭 😭 😭 😭 😭

iroh/src/magicsock.rs
Frando: None, they're identical. Both expand to `DiscoBoxError::Parse { source, meta: Meta::default() }`

iroh/src/endpoint.rs
matheus23: Can this be simplified to `server.accept().await.anyerr()?.await.anyerr()?;`?

While `Incoming::accept` returns a `Connecting`, `Incoming::into_future` turns it into an `IncomingFuture` which outputs `Connection`.

Cargo.toml
```diff
@@ -40,3 +40,7 @@ unexpected_cfgs = { level = "warn", check-cfg = ["cfg(iroh_docsrs)", "cfg(iroh_l
 
 [workspace.lints.clippy]
 unused-async = "warn"
+
+[patch.crates-io]
+netwatch = { git = "https://github.com/n0-computer/net-tools", branch = "Frando/n0-snafu" }
+portmapper = { git = "https://github.com/n0-computer/net-tools", branch = "Frando/n0-snafu" }
```
matheus23: Some git deps remaining

iroh-base/src/relay_url.rs
```diff
@@ -54,7 +51,7 @@ impl FromStr for RelayUrl {
     type Err = RelayUrlParseError;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let inner = Url::from_str(s).context(RelayUrlParseSnafu)?;
+        let inner = Url::from_str(s).map_err(|err| RelayUrlParseError::new(err))?;
```
matheus23:
```suggestion
        let inner = Url::from_str(s).map_err(RelayUrlParseError::new)?;
```

iroh-relay/src/client.rs
```diff
@@ -47,75 +46,85 @@ mod util;
...
-    #[snafu(implicit)]
-    span_trace: n0_snafu::SpanTrace,
-})]
+#[stack_error(derive, add_meta, from_sources)]
 #[allow(missing_docs)]
-#[derive(Debug, Snafu)]
 #[non_exhaustive]
 pub enum DialError {
-    #[snafu(display("Invliad target port"))]
+    #[error("Invliad target port")]
```
matheus23: Sorry, ignore this if you're time-constrained. We should fix this sometime though.

```suggestion
    #[error("Invalid target port")]
```

iroh-relay/src/client.rs
```diff
@@ -271,12 +278,12 @@ impl ClientBuilder {
         }
         let (conn, response) = builder.connect_on(stream).await?;
 
-        if response.status() != hyper::StatusCode::SWITCHING_PROTOCOLS {
-            UnexpectedUpgradeStatusSnafu {
-                code: response.status(),
+        n0_error::ensure!(
```
matheus23, line 281: I think there's a place we missed importing the macro

```suggestion
        ensure!(
```

iroh-relay/src/server/client.rs
```diff
@@ -588,16 +553,16 @@ mod tests {
         match stream.next().await {
             Some(Ok(frame)) => {
                 if frame_type != frame.typ() {
-                    snafu::whatever!(
+                    n0_error::bail_any!(
```
matheus23: Nit: I think we wanted to import the macros fully.

iroh-relay/src/server/http_server.rs
```diff
@@ -474,41 +450,43 @@ impl RelayService {
         ) -> Result<&HeaderValue, RelayUpgradeReqError> {
             req.headers()
                 .get(&header)
-                .context(MissingHeaderSnafu { header })
+                .ok_or_else(|| e!(RelayUpgradeReqError::MissingHeader { header }))
         }
 
         let upgrade_header = expect_header(&req, UPGRADE)?;
-        snafu::ensure!(
+        n0_error::ensure!(
```
matheus23: Nit: A couple of qualified uses

iroh/examples/transfer.rs
```diff
@@ -212,7 +212,7 @@ impl EndpointArgs {
             }
             #[cfg(not(feature = "test-utils"))]
             {
-                snafu::whatever!(
+                n0_error::bail_any!(
```
matheus23, line 215: Nit: some more qualified uses

matheus23 (approved):
I only have small nits.

LGTM!

Cargo.toml
Frando: removed

iroh-relay/src/client.rs
Frando: This function is `#[cfg(not(wasm_browser))]` and this is the only occurence of `ensure` in this module, so I left it like this instead of doing a conditional import of ensure at the module level.

iroh/examples/transfer.rs
Frando: This is in a conditioanl section, not sure what's better, keep it like this or conditional import at module level?

iroh-relay/src/server/http_server.rs
Frando: fixed

iroh-relay/src/server/client.rs
Frando: fixed

iroh-relay/src/client.rs
Frando: fixed

iroh-base/src/relay_url.rs
Frando: fixed

PR #3562, opened 2025-10-21
chore!: update redb to version 3

dignifiedquire:
## Breaking Changes

This bumps redb to version 3, in `iroh-dns-server`, which means you need to run with `0.93` at least once to upgrade the database.


dignifiedquire:
needs thought regarding the msrv change..

PR #3567, opened 2025-10-22
docs(iroh): clarify ProtocolHandler::accept

eminence:
Clarify at the connection is always closed when `accept` returns, and that the AcceptError variant doesn't really matter


iroh/src/protocol.rs
```diff
@@ -175,7 +175,13 @@ pub trait ProtocolHandler: Send + Sync + std::fmt::Debug + 'static {
     ///
     /// Can be implemented as `async fn accept(&self, connection: Connection) -> Result<()>`.
     ///
-    /// The returned future runs on a freshly spawned tokio task so it can be long-running.
+    /// The returned future runs on a freshly spawned tokio task so it can be long-running. Once
+    /// `accept()` returns, the connection is closed, so the returned future is expected
+    /// to drive the protocol to completion.  If there in a protocol error, you can
```
dignifiedquire: typo `is a`

Frando: It should also clarify that the connection is only closed *if no other clones exist*, not alwayy when `accept` returns. I.e. if you send the connection or a clone to another task via a channel and return from `accept`, the connection will stay alive until closed explicitly by user code with `Connection::close` or until the last clone of `Connection` is dropped.

eminence: Good comment, thanks.  I just force-pushed a new version that does a better job with this

PR #3575, opened 2025-10-23
feat(iroh-relay): allow for extension of `RelayMap`s

dignifiedquire:


Frando (approved):

iroh-relay/src/relay_map.rs
```diff
@@ -111,14 +111,27 @@ impl RelayMap {
     pub fn remove(&self, url: &RelayUrl) -> Option<Arc<RelayConfig>> {
         self.relays.write().expect("poisoned").remove(url)
     }
+
+    /// Extends this `RelayMap` with another one.
+    pub fn extend(&self, other: &RelayMap) {
+        let mut a = self.relays.write().expect("poisoned");
+        let b = other.relays.read().expect("poisoned");
+        a.extend(b.iter().map(|(a, b)| (a.clone(), b.clone())));
+    }
```
matthiasbeyer, line 120: This should be `impl Extend for RelayMap` IMO.

Also: Why is there panicing `expect` inside of this function? This is not what a user expects from the documentation of the whole type!

matthiasbeyer: After some discussion on your discord tho, using the `Extend` trait does not seem possible because of how the types are designed here.
And the std RwLock cannot easily be replaced by a tokio one :cry:

PR #3577, opened 2025-10-25
ref(iroh): Relay map extend

matthiasbeyer:
## Description

Following https://github.com/n0-computer/iroh/pull/3575 (which was inspired by me complaining on your discord that this was missing from the interface) and the discussion on your discord, I figured I would fancy a slightly different approach to that PR and the interfaces introduced in your PR there.

This is twofold:

* Replace the `RelayMap::extend` with an actual `impl Extend ror RelayMap` (that can easily be used with an existing `RelayMap` by calling `rm.relays::<Vec<_>>()` on it
* Rewrite the existing `RelayMap::extend` with an `RelayMap::join`. Reason here is that my usecase (that, again, inspired this to be added) is to build a `RelayMap` object with defaults + custom entries _before_ using them, so there's no need to update (`.extend()`ing) an existing `RelayMap` during runtime. This is still preserved via the `impl Extend` though.

Tell me what you think! CCing people from discord:

* CC @flub
* CC @dignifiedquire 

## Breaking Changes

- `RelayMap::extend` (unreleased) -> `RelayMap::join`

### DIFF
--- a/iroh-relay/src/relay_map.rs
+++ b/iroh-relay/src/relay_map.rs
@@ -112,11 +112,14 @@ impl RelayMap {
         self.relays.write().expect("poisoned").remove(url)
     }
 
-    /// Extends this `RelayMap` with another one.
-    pub fn extend(&self, other: &RelayMap) {
-        let mut a = self.relays.write().expect("poisoned");
-        let b = other.relays.read().expect("poisoned");
-        a.extend(b.iter().map(|(a, b)| (a.clone(), b.clone())));
+    /// Joins this `RelayMap` with another one into a new one
+    pub fn join(self, other: RelayMap) -> RelayMap {
+        {
+            let mut a = self.relays.write().expect("poisoned");
+            let b = other.relays.read().expect("poisoned");
+            a.extend(b.iter().map(|(a, b)| (a.clone(), b.clone())));
+        }
+        self
     }
 }
 


iroh-relay/src/relay_map.rs
```diff
@@ -112,14 +112,37 @@ impl RelayMap {
         self.relays.write().expect("poisoned").remove(url)
     }
 
-    /// Extends this `RelayMap` with another one.
-    pub fn extend(&self, other: &RelayMap) {
+    /// Joins this `RelayMap` with another one into a new one
+    pub fn join(self, other: RelayMap) -> RelayMap {
```
flub, line 116: Isn't this called `chain` usually?  I think this is very similar to the `Iterator::chain` signature.

iroh-relay/src/relay_map.rs
```diff
@@ -112,14 +112,37 @@ impl RelayMap {
...
+        // SAFETY: We have exclusive access to this object
```
flub: I appreciate documenting this, though I think this is slightly incorrect:
- SAFETY is used to document `unsafe` code, panicking is not unsafe (tokio is full of panicking functions!)
- This is not about exclusive access, the assertion here is that the lock is not poisoned.  That is, there was no panic while holding the lock.  The reasoning is that if there was a panic while holding the lock the internal invariants that the lock protects may not be upheld.

But writing that reasoning out, you can probably avoid panicking here I think.  Because there are very simple internal variants you can probably just extract the internal state from the poisoned lock and put it back in there.  It does require you to audit all the write paths to make sure they don't contain panicking code themselves though, because otherwise you silently lose some state.  But all the write paths are internal via the functions so that's possible.  It's a bunch of work for marginal gain :smile: but you can get the ideal of no panics.  (Which has some small benefits to generated code size IIRC, but we've never really started optimising for this yet).

Anyway, the panics on poison are perfectly fine as well :)

iroh-relay/src/relay_map.rs
matthiasbeyer: Yes it is. I opted for `join` because a `Map` is not something that can (semantically) be put into a :chains: chain! :wink: I think of `chain()` something that can be applied to a _sequence_ of something (hence `Iterator::chain` but not `BTreeMap::chain` for example).

I have no strong feelings about the actual method name tho, so if you want me to change it, I'll happily do so.

iroh-relay/src/relay_map.rs
matthiasbeyer: > SAFETY is used to document unsafe code, panicking is not unsafe (tokio is full of panicking functions!)

True. Maybe `// This is safe because...` rather?

> This is not about exclusive access, the assertion here is that the lock is not poisoned. [...]

Also true. I need to change that.

About your last bit: Yes, but that's probably not worth the effort.

matthiasbeyer:
Added some fixups for the misleading comment.

ramfox:
@matthiasbeyer we are doing a release on monday! Would love to get this in if you could fix the failing CI.

matthiasbeyer:
This should suffice.

I can fix issues until tomorrow, Saturday, around 13:00 UTC. After that I am enjoying Hans Zimmer and on Sunday I will be traveling :wink: :+1:

iroh-relay/src/relay_map.rs
dignifiedquire: I am not convinced passing owned versions makes sense for a fully clonable thing. This API really doesn't communicate the fact that this doesn't actually create a new version, but rather modifies the underlying, likely shared instances

iroh-relay/src/relay_map.rs
```diff
@@ -112,11 +112,34 @@ impl RelayMap {
...
     /// Extends this `RelayMap` with another one.
-    pub fn extend(&self, other: &RelayMap) {
+    ///
+    /// You can use this like this:
+    ///
+    /// ```rust
+    /// # let relay_map_a: RelayMap = { unimplemented!() };
+    /// # let relay_map_b: RelayMap = { unimplemented!() };
+    ///
+    /// relay_map_a.extend(relay_map_b.relays::<Vec<_>>());
```
dignifiedquire, line 135: this is less efficient, and seems rather pointless to support, if you have to collect & allocate into a vector first before using the trait

flub: The inner lock seems to stop you from doing anything nice using `IntoIterator` or `Extend`. :disappointed:

flub (commented):
I think the main objection to the current API is that `extend` overlaps with `Extend::extend`?  But implementing the trait is really ugly due to the internal locks.

The existing API does guide you to do the right thing, which really is what matters.  I'm not sure supporting the Extend trait directly does improve ergonomics.  So overall I think I'm -1 on this so far.

iroh-relay/src/relay_map.rs
matthiasbeyer: Yep, that's the issue yes.

matthiasbeyer: If this would be a `DashMap`, I guess it would work.

That also would mean that not the whole map must be locked when adding/removing.

But it introduces other things that are dangerous, such as holding dashmap locks over `await` points (which can be linted against though).

flub: How do you lint against that?  I didn't know that was possible.

matthiasbeyer: Like that: https://git.hemera.systems/Hemera/dylint-various/src/branch/main/lints/dashmap-ref

:wink:

matthiasbeyer: But clippy can do that (by now) via the `await-holding-invalid-types` lint: https://rust-lang.github.io/rust-clippy/stable/index.html#await_holding_invalid_type

flub:
Closing for now, seems like the benefit is unclear and the existing methods are ergonomic enough. If there's a solution direction we haven't thought of yet in this PR feel free to open a new one or reopen with an updated proposal.

PR #3582, opened 2025-10-29
Improve wording for the Builder::discovery fn

matthiasbeyer:
## Description

After discussion on discord with @flub I tried to improve the wording of the function a bit here.

I am open to discussion of course.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1177,7 +1177,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


iroh/src/endpoint.rs
```diff
@@ -308,7 +308,8 @@ impl Builder {
     /// the endpoint. Discovery services that need to publish information need
     /// to use this secret key to sign the information.
     ///
-    /// If you add multiple discovery services, they will be combined using a
+    /// The internals of the [`Builder`] use the discovery objects passed to this function and
+    /// automatically combines them upon [`Builder::bind`] via a
```
flub, line 312: > If you call this method multiple times, all discovery services passed in will be combined via an internal [`crate::discovery::ConcurrentDiscovery`].  To remove all discovery services already added to the builder use [`Self::clear_discovery`].

Would be my attempt.  And that would remove the separate paragraph about clearing below.

Would be great if you could also fix the wrong function name in the paragraph above and link it to [`IntoDiscovery::into_discovery`]

matthiasbeyer:
Stop merging to other peoples PRs!

ramfox:
> Stop merging to other peoples PRs!

If this is directed at me, it's my job to ensure that our releases go out in a timely fashion and to help folks get their PRs merged ASAP for that.

I apologize for the frustration but:
a) this was not an appropriate way to communicate it. 
b) I reserve the right to try to keep PRs up to date in this way, when we are nearing our deadline and the PRs are doc adjustments & so merges will not interfere with code contributions.

matheus23:
> Stop merging to other peoples PRs!

This is a setting you have control over in GitHub:
image

But I'd kindly ask you to reserve us the right to modify your PRs, as that makes our lives easier.

flub:
Incorporated into #3801

PR #3592, opened 2025-10-31
docs(iroh): fix documentation on endpoint::Builder::empty

eminence:
# Description

Fixes a small documentation error.
See also the discussion on discord: https://discord.com/channels/1161119546170687619/1161119546644627528/1433606524336476182

> No, the docs are wrong —from a previous iteration that did not take a relay mode as an argument.
> We want folks to be forced to choose a relay mode, there are too many footguns that can happen if you don't think about your specific use case.

### DIFF
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -136,8 +136,7 @@ impl Builder {
         self
     }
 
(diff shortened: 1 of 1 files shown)


matheus23 (approved):
Thank you!

PR #3593, opened 2025-10-31
feat: remove Endpoint::latency

varun-doshi:
## Description

Removes `Endpoint::latency`
Also added `#[allow(dead_code)]` for the inner calls. I can remove the functions all together if they are not needed

## Change checklist

- [x] Self-review.


Frando:
Thanks! This will have to wait though I think until we have a replacement for this ready, which will happen after merging the multipath branch.

dignifiedquire:
Thanks, I replaced this with #3717, which targets the latest changes

PR #3601, opened 2025-11-04
fix: stop relay actor on endpoint close

Frando:
## Description

Since the transport abstraction was introduced, we no longer shut down the relay actor when calling `Endpoint::close`. The relay actor is currently only stopped when the last clone of the `iroh::Endpoint` is dropped. This is surprising and wasteful. This PR fixes this so that the relay actor is stopped when calling `Endpoint::close`.

Background: Since introducing the transport abstraction, the relay actor handle is now owned by `Transports` which is owned by `MagicUdpSocket` which is owned by the `quinn::Endpoint` as its `AsyncUdpSocket`. The `AsyncUdpSocket` trait has no notification when the endpoint is closing, thus currently the relay transport (and the relay actor, which it owns) is not notified at all when the endpoint is shutting down. Therefore, it keeps running (and potentially reconnecting to the relay or receiving messages) until it is stopped when the last clone of the endpoint is dropped and thus the `AbortOnDropHandle` owning the relay actor tasks is dropped, finally stopping the relay actor tasks.

This PR exposes an existing cancellation token within the relay actor to the magicsocket, allowing it to cancel the relay actors when calling `Endpoint::close`.

## Notes & open questions

I tested this by looking at logs after closing an endpoint. Not sure if there is a good way to test this. I'd vote to merge this even without a test because it fixes an issue, unless someone has a good idea on how to test this.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/magicsock.rs
+++ b/iroh/src/magicsock.rs
@@ -149,8 +149,8 @@ pub(crate) struct Handle {
     msock: Arc<MagicSock>,
     // empty when shutdown
     actor_task: Arc<Mutex<Option<AbortOnDropHandle<()>>>>,
-    /// Token to cancel the actor task.
-    actor_token: CancellationToken,
+    /// Token to cancel the actor task and shutdown the relay transport.
+    shutdown_token: CancellationToken,
     // quinn endpoint
     endpoint: quinn::Endpoint,
 }
@@ -210,7 +210,6 @@ pub(crate) struct MagicSock {
 
     /// Metrics
     pub(crate) metrics: EndpointMetrics,
-    cancel_transports: CancellationToken,
 }
 
 #[allow(missing_docs)]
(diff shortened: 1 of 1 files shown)


iroh/src/magicsock.rs
```diff
@@ -1587,6 +1594,7 @@ impl Handle {
             return;
         }
         self.msock.closing.store(true, Ordering::Relaxed);
+        self.msock.cancel_transports.cancel();
```
flub: Instead of a new token, would this be done better as a child token from the `actor_token` that already exists on the line below?

Frando:
Changed to reuse the existing cancel token.

iroh/src/magicsock.rs
Frando: Yeah, sounds good. Did that.

PR #3607, opened 2025-11-05
refactor!: make `OutgoingZeroRttConnection` implement `Clone`

Frando:
## Description

This allows to get hold of a future that resolves with `ZeroRttStatus` while doing work with the connection (e.g. opening streams) in parallel, without having to give up ownership of the `OutgoingZeroRttConnection`.

Used in https://github.com/n0-computer/irpc/pull/81

## Breaking Changes

* `OutgoingZeroRttConnection::handshake_completed()` now takes `&self` instead of `self`

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/protocol.rs
+++ b/iroh/src/protocol.rs
@@ -159,7 +159,21 @@ pub trait ProtocolHandler: Send + Sync + std::fmt::Debug + 'static {
     ///
     /// Can be implemented as `async fn on_accepting(&self, accepting: Accepting) -> Result<Connection>`.
     ///
-    /// This enables accepting 0-RTT data from clients, among other things.
+    /// Typically, this method is used as an early interception point to accept
+    /// or reject a connection.
+    ///
+    /// However, this method can also be used to implement the accept side of a
+    /// 0-RTT connection.
+    ///
+    /// ## 0-RTT
+    ///
+    /// `ProtocolHandler::on_accepting` allows you to take over the connection
+    /// state machine early in the handshake processes, by calling [`Accepting::into_0rtt`].
+    ///
+    /// When working with 0-RTT, you may want to implement all of your protocol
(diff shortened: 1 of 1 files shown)


iroh/src/endpoint/connection.rs
```diff
@@ -1517,7 +1520,7 @@ mod tests {
     use super::Endpoint;
     use crate::{
         RelayMode,
-        endpoint::{ConnectOptions, Incoming, ZeroRttStatus},
+        endpoint::{ConnectOptions, Incoming, connection::ZeroRttStatus},
```
matheus23: Nit: I think this change shouldn't be necessary.

matheus23 (approved):
IMO this is the way to go.

@Frando please fix the PR title before though :pray:

iroh/src/endpoint/connection.rs
```diff
@@ -396,7 +397,8 @@ impl Connecting {
                     accepted: ZeroRttAccepted {
                         inner: zrtt_accepted,
                         _discovery_drop_guard: self._discovery_drop_guard,
-                    },
+                    }
+                    .shared(),
```
rklaehn, line 401: If we use .shared here we should maybe expose it from n0_future, if it isn't already done. I have wanted it a few times.

rklaehn (approved):
As I mentioned on discord, I think we need to do this. Connections being Clone is baked in to quinn design. Having a special kind of connection that isn't clone would mean that we would have to use very different patterns to work with 0rtt connections and with normal connections, and it would mean that you can't abstract over the 2 kinds of connections as @frando does in his irpc PR.

Frando:
> @Frando please fix the PR title before though :pray: 

What needs to change?

PR #3608, opened 2025-11-05
feat: enhance address handling in EndpointState

### LINKED ISSUE #3602: Client sends to wrong address on server restart
When a client and server are connected using a direct address the client has the server's address stored in the "best address".  If you now restart the server and ask the client to create a new connection to the re-started server, which is using the same EndpointId but a different IP bind address, the connection is very slow.

The issue is that the "best address" is still known and is considered valid for 6.5s.  So even if the new connect call is given the new correct IP address of the server, it will keep sending to the previously chosen "best addr" until the 6.5s time out.  Only then will it fall back to the relay and manage to re-establish a connection.

virajbhartiya:
## Description

- Prefer newly provided app direct addresses immediately when updating an endpoint’s addresses.
- If `Source` is `App` or `NamedApp` and the current UDP send address isn’t part of the new set, clear validity for all existing direct paths and reset `last_call_me_maybe` to trigger immediate discovery instead of waiting the 6.5s trust window.
- Fixes slow reconnects when a server restarts with the same `EndpointId` but a new bind address.
Closes #3602
Implementation details:
- In `EndpointState::update_from_endpoint_addr`:
  - Capture the current send address before acquiring the mutable guard.
  - Collect `new_addrs` into a `Vec<SocketAddr>` and update paths via the guard.
  - If the current address isn’t in the new set, clear path validities and reset `last_call_me_maybe`.

Also:
- Resolve a borrow conflict (E0502) by avoiding immutable access to `udp_paths` while a `MutAccess` guard is held.

Touched files:
- `iroh/src/magicsock/endpoint_map/endpoint_state.rs`

## Breaking Changes

None.

## Notes & open questions

- Scope-limited to `Source::App` and `Source::NamedApp` so discovery-initiated updates keep current behavior.
- We may later consider similar invalidation for other sources if empirical data shows benefits.

## Change checklist

- [ ] Self-review.
- [ ] Documentation updates following the style guide, if relevant.
- [ ] Tests if relevant.
    - [ ] `quic-rpc`
    - [ ] `iroh-gossip`
    - [ ] `iroh-blobs`
    - [ ] `dumbpipe`
    - [ ] `sendme`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,13 @@
 
 All notable changes to iroh will be documented in this file.
 
-## [0.95.0](https://github.com/n0-computer/iroh/compare/v0.94.1..0.95.0) - 2025-11-04
+## [0.95.1](https://github.com/n0-computer/iroh/compare/v0.95.0..0.95.1) - 2025-11-05
+
+### 🚜 Refactor
+
+- [**breaking**] Make `OutgoingZeroRttConnection` implement `Clone` ([#3607](https://github.com/n0-computer/iroh/issues/3607)) - ([721bd2d](https://github.com/n0-computer/iroh/commit/721bd2d9cf386807a750a7d5abe42d8aed5aacc6))
+
+## [0.95.0](https://github.com/n0-computer/iroh/compare/v0.94.1..v0.95.0) - 2025-11-04
 
 ### ⛰️  Features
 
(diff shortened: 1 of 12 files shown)


flub:
Hi, I'm afraid we're not going to look into solving this on the main branch.  This code changes drastically on the feat-multipath branch where we'll have a look at what the best approach is to solving this.  Probably looking at QUIC PTO firing for this, which would apply in more situations.

PR #3609, opened 2025-11-05
feat(endpoint-id): add recognizable Base32 serialization

TotalKrill:
This implements a new Display and FromStr format for EndpointId, using Crockford Base32 encoding for a human-friendly and recognizable representation. As outlined in #3605 

Design goals:
  - URL-safe (no '+', '/', '=' or reserved symbols)
  - Case-insensitive (ABCD = abcd)
  - Easy to parse and display
  - DNS-label compatible (<64 characters total)
  - Visually distinctive, so it’s recognizable at a glance—similar in spirit to how IPv4, IPv6, MAC addresses, and UUIDs have familiar forms.

This format should make Iroh endpoint IDs easier to identify, share, and reason about in logs, APIs, and config files for years to come.# Please enter the commit message for your changes. Lines starting

## Description

The serialized form now consists of:
  - a fixed-length short prefix (first 5 bytes → 8 Base32 chars)
  - a dash separator ('-')
  - the full 32-byte Base32-encoded address (52 chars total)
  
### Example:
short: RICVBSGA
long: RICVBSGA-NBCGVQB7ZGU7LEKCZR4XT45I2JBIDX7ODWGEPE42QSLA

## Breaking Changes

It shortens the serialization for Endpoint Ids and introduces a custom format (short-rest), but it also maintains desserialization capabilities for older versions

### DIFF
--- a/iroh-base/src/key.rs
+++ b/iroh-base/src/key.rs
@@ -157,7 +157,7 @@ struct PublicKeyShort([u8; 5]);
 
 impl Display for PublicKeyShort {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        data_encoding::HEXLOWER.encode_write(&self.0, f)
+        data_encoding::BASE32_NOPAD_NOCASE.encode_write(&self.0, f)
     }
 }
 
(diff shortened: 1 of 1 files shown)


dignifiedquire:
Thanks for the suggestion, but we had a long discussion before and using base32 as the default display is not going to happen. Hex is still much more accepted and usable, and doesn't have the drawbacks of many different alphabets, like base32 has. 
We could add a helper method to make it easy to encode into base32 potentially if there is the need for it, but that's a different discussion.

PR #3614, opened 2025-11-06
feat(multipath): add RemoteInfo and ConnectionInfo

Frando:
## Description

* Adds `ConnectionInfo`, which is a weak info handle to a connection (does not keep the connection alive, cannot be upgraded to a full `Connection`). From a `ConnectionInfo` some static info is directly available, and as long as the connection is alive you can access the connection stats. You can also watch for path changes, and wait for the connection to be closed (without keeping it alive).
 * Adds functions `remote_info` and `remotes` to get a `RemoteInfo` or list of remote all remote infos. 
  * The `RemoteInfo` allows to get a list of `ConnectionInfo` that are currently active
  * It also allows to watch the currently selected path, replacing `Endpoint::conn_type`
      * We have to think through if this is hindering a potential future where we send on multiple paths in parallel
      * If we remove this, you would have to go through all connections, call `ConnectionInfo::paths_info`, and check if there is a ip path (-> direct) or only a relay path (-> relayed).
      * I'm torn a bit if we want this or not. It is a much simpler API, but maybe we shouldn't expose it?
  * It also allows to get the RTT of the currently selected path
* Removes `Endpoint::conn_type` and `Endpoint::latency`

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -26,7 +26,7 @@ use url::Url;
 
 pub use super::magicsock::{
     AddEndpointAddrError, ConnectionType, DirectAddr, DirectAddrType, PathInfo, PathsInfo,
-    endpoint_map::Source,
+    RemoteInfo, endpoint_map::Source,
 };
 #[cfg(wasm_browser)]
 use crate::discovery::pkarr::PkarrResolver;
@@ -55,8 +55,8 @@ pub use quinn::{
     AcceptBi, AcceptUni, AckFrequencyConfig, ApplicationClose, Chunk, ClosedStream,
     ConnectionClose, ConnectionError, ConnectionStats, MtuDiscoveryConfig, OpenBi, OpenUni,
     ReadDatagram, ReadError, ReadExactError, ReadToEndError, RecvStream, ResetError, RetryError,
-    SendDatagramError, SendStream, ServerConfig, StoppedError, StreamId, TransportConfig, VarInt,
-    WeakConnectionHandle, WriteError,
+    SendDatagramError, SendStream, ServerConfig, Side, StoppedError, StreamId, TransportConfig,
+    VarInt, WeakConnectionHandle, WriteError,
 };
 pub use quinn_proto::{
     FrameStats, PathStats, TransportError, TransportErrorCode, UdpStats, Written,
@@ -69,8 +69,8 @@ pub use quinn_proto::{
 
 pub use self::connection::{
     Accept, Accepting, AlpnError, AuthenticationError, Connecting, ConnectingError, Connection,
-    Incoming, IncomingZeroRttConnection, OutgoingZeroRttConnection, RemoteEndpointIdError,
-    ZeroRttStatus,
+    ConnectionInfo, Incoming, IncomingZeroRttConnection, OutgoingZeroRttConnection,
+    RemoteEndpointIdError, ZeroRttStatus,
 };
 
 /// The delay to fall back to discovery when direct addresses fail.
@@ -98,7 +98,7 @@ pub enum PathSelection {
 /// new [`EndpointId`].
 ///
 /// To create the [`Endpoint`] call [`Builder::bind`].
-#[derive(Debug)]
+#[derive(derive_more::Debug)]
 pub struct Builder {
     secret_key: Option<SecretKey>,
     relay_mode: RelayMode,
@@ -963,38 +963,24 @@ impl Endpoint {
     //
     // Partially they return things passed into the builder.
 
-    /// Returns a [`Watcher`] that reports the current connection type and any changes for
-    /// given remote endpoint.
+    /// Information about a remote endpoint.
     ///
-    /// This watcher allows observing a stream of [`ConnectionType`] items by calling
-    /// [`Watcher::stream()`]. If the underlying connection to a remote endpoint changes, it will
-    /// yield a new item.  These connection changes are when the connection switches between
-    /// using the Relay server and a direct connection.
+    /// From the [`RemoteInfo`] you can watch which path is selected, get the current
+    /// round-trip time (latency), and get a list of [`ConnectionInfo`].
     ///
-    /// Note that this does not guarantee each connection change is yielded in the stream.
-    /// If the connection type changes several times before this stream is polled, only the
-    /// last recorded state is returned.  This can be observed e.g. right at the start of a
-    /// connection when the switch from a relayed to a direct connection can be so fast that
-    /// the relayed state is never exposed.
-    ///
-    /// If there is currently a connection with the remote endpoint, then using [`Watcher::get`]
-    /// will immediately return either [`ConnectionType::Relay`], [`ConnectionType::Direct`]
-    /// or [`ConnectionType::Mixed`].
-    ///
-    /// It is possible for the connection type to be [`ConnectionType::None`] if you've
-    /// recently connected to this endpoint id but previous methods of reaching the endpoint have
-    /// become inaccessible.
-    ///
-    /// Will return `None` if we do not have any address information for the given `endpoint_id`.
-    pub fn conn_type(&self, endpoint_id: EndpointId) -> Option<n0_watcher::Direct<ConnectionType>> {
-        self.msock.conn_type(endpoint_id)
+    /// Returns `None` if we don't have any state for this remote.
+    pub fn remote_info(&self, endpoint_id: EndpointId) -> Option<RemoteInfo> {
+        self.msock.endpoint_map.remote_info(endpoint_id)
     }
 
-    /// Returns the currently lowest latency for this endpoint.
+    /// Returns a list of all remote endpoints that this endpoint is dealing with.
     ///
-    /// Will return `None` if we do not have any address information for the given `endpoint_id`.
-    pub async fn latency(&self, endpoint_id: EndpointId) -> Option<Duration> {
-        self.msock.latency(endpoint_id).await
+    /// This includes all endpoints to which we have active connections. It also may include endpoints
+    /// to which we are in the process of connecting, or have recently been connected to.
+    ///
+    /// TODO: Expand docs.
+    pub fn remotes(&self) -> Vec<RemoteInfo> {
+        self.msock.endpoint_map.remotes()
     }
 
     /// Returns the DNS resolver used in this [`Endpoint`].
--- a/iroh/src/endpoint/connection.rs
+++ b/iroh/src/endpoint/connection.rs
@@ -35,7 +35,7 @@ use n0_watcher::{Watchable, Watcher};
 use pin_project::pin_project;
 use quinn::{
     AcceptBi, AcceptUni, ConnectionError, ConnectionStats, OpenBi, OpenUni, ReadDatagram,
-    RetryError, SendDatagramError, ServerConfig, VarInt,
+    RetryError, SendDatagramError, ServerConfig, Side, VarInt, WeakConnectionHandle,
 };
 use quinn_proto::PathId;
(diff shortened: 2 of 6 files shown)


iroh/src/endpoint.rs
```diff
@@ -963,38 +963,24 @@ impl Endpoint {
...
+    /// Returns a list of all remote endpoints that this endpoint is dealing with.
+    ///
+    /// This includes all endpoints to which we have active connections. It also may include endpoints
+    /// to which we are in the process of connecting, or have recently been connected to.
     ///
-    /// Will return `None` if we do not have any address information for the given `endpoint_id`.
-    pub async fn latency(&self, endpoint_id: EndpointId) -> Option<Duration> {
-        self.msock.latency(endpoint_id).await
+    /// TODO: Expand docs.
+    pub fn remotes(&self) -> Vec<RemoteInfo> {
```
flub, line 982: I'd consider making this `impl Iterator`, it means we don't *have* to allocate a vec (even if we currently do internally).  E.g. if you have a busy server this could be thousands really.

iroh/src/endpoint/connection.rs
```diff
@@ -241,15 +241,16 @@ fn conn_from_quinn_conn(
...
-    Ok(Connection {
+    let conn = Connection {
         remote_id,
         alpn,
         inner: conn,
         paths_info,
-    })
+    };
+    let info = conn.to_info();
+    // Register this connection with the magicsock.
```
flub, line 251: I think we can remove this comment now, so we look a bit less like an AI :sweat_smile:

iroh/src/endpoint/connection.rs
```diff
@@ -1530,6 +1531,83 @@ impl Connection {
     pub fn set_max_concurrent_bi_streams(&self, count: VarInt) {
         self.inner.set_max_concurrent_bi_streams(count)
     }
+
+    /// Returns the side of the connection (client or server).
+    pub fn side(&self) -> Side {
+        self.inner.side()
+    }
```
flub, line 1538: Ah, thanks for adding this!

I'd write a lot more docs, like explaining why on the QUIC-level this is a thing and why you might want to know for a protocol despite that iroh always allows you to be both.  Basically assume you get a user landing from hn or reddit and exploring iroh for the first time without having a clue.

iroh/src/endpoint/connection.rs
```diff
@@ -1530,6 +1531,83 @@ impl Connection {
...
+
+    /// Returns a [`ConnectionInfo`], which is a weak handle to the connection
+    /// that does not keep the connection alive, but does allow to access some information
+    /// about the connection, and allows to wait for the connection to be closed.
```
flub, line 1542: I am a stickler for following the rust doc style guide strictly :)

So re-phrasing this so you have a 1-line summary followed by a blank line and longer explanation will focus you on writing meaningful words in that one line, because that's the line you'll see on rustdoc summary pages and will most often read from ide help.

(I'll only leave this comment once)

iroh/src/endpoint/connection.rs
```diff
@@ -1530,6 +1531,83 @@ impl Connection {
...
+    /// will be a path via the relay server and a holepunched path.  This returns all the
+    /// paths in use by this connection.
+    pub fn paths_info(&self) -> &impl Watcher<Value = PathsInfo> {
+        &self.paths_info
+    }
+
+    // We could add such util methods here, not sure.
+    pub fn has_direct_path(&mut self) -> bool {
+        self.paths_info.get().keys().any(|addr| addr.is_ip())
+    }
```
flub, line 1591: I think `selected_path` like `RemoteInfo` has is a better idea for now.

iroh/src/endpoint/connection.rs
```diff
@@ -1530,6 +1531,83 @@ impl Connection {
...
+
+    /// Current best estimate of this connection's latency (round-trip-time)
+    ///
+    /// Returns `None` if the connection has been dropped.
+    pub fn rtt(&self) -> Option<Duration> {
+        self.inner.upgrade().map(|conn| conn.rtt())
```
flub, line 1597: Where you are getting this value from is problematic currently.  That's a known issue on the multipath branch that this is kinda bogus.

Probably should take the rtt of the selected path.  If getting that isn't too expensive here.  It would be nice to expose probably.

iroh/src/endpoint/connection.rs
```diff
@@ -1530,6 +1531,83 @@ impl Connection {
...
+    }
+
+    /// Returns connection statistics.
+    ///
+    /// Returns `None` if the connection has been dropped.
+    pub fn stats(&self) -> Option<ConnectionStats> {
```
flub, line 1603: One issue I have with `ConnectionStats` is that this includes `PathStats` indexed by PathId.  I'm not really sure I want to expose PathIds to our users.  We also need to newtype this anyway... so perhaps the first version can just drop PathStats?  The more complete version would be to change that `ConnectionStats::paths` to be a map keyed by `TransportAddr`, but that might again be too much/expensive.  Maybe if you want those you should go paths themselves?  Not sure.

iroh/src/magicsock/endpoint_map.rs
```diff
@@ -122,17 +122,23 @@ impl EndpointMap {
...
+
+    pub(crate) fn remotes(&self) -> Vec<RemoteInfo> {
+        self.actor_handles
+            .lock()
+            .expect("poisoned")
+            .iter()
+            .map(|(eid, handle)| {
+                RemoteInfo::new(*eid, handle.sender.clone(), handle.selected_path.watch())
+            })
+            .collect()
```
flub, line 141: Ah right.  It's the lock inside again that forces you to collect.  This is kind of bad because of the many connections things.

I can think of wrapping the mutex in an Arc and that would allow you to clone it out and pass it to a custom iterator struct which could then take the lock on each `.next()` call...  Not sure if that's much better.  It was kind of nice not having an Arc here...

iroh/src/magicsock/endpoint_map/endpoint_state.rs
```diff
@@ -350,7 +359,10 @@ impl EndpointStateActor {
 
             // Store PathId(0), set path_status and select best path, check if holepunching
             // is needed.
-            if let Some(conn) = handle.upgrade() {
+            // TODO(frando): The code here upgraded the handle again (which we already upgraded a couple of lines above)
+            // Was there a reason for that?
```
flub, line 363: I think that was an outright bug.  No need to upgrade it twice.  A sign the function is too long perhaps...

iroh/src/magicsock/endpoint_map/endpoint_state.rs
```diff
@@ -515,19 +527,34 @@ impl EndpointStateActor {
...
-                    return Some(stats.rtt);
+                    return Some(rtt);
                 }
             }
             None
         });
         tx.send(rtt).ok();
     }
 
+    fn handle_msg_connection_infos(&self, tx: oneshot::Sender<Vec<ConnectionInfo>>) {
```
flub, line 543: Oof, this allocates a Vec for all connections again.  I do think that is problematic.

iroh/src/magicsock/endpoint_map/endpoint_state.rs
```diff
@@ -934,6 +961,7 @@ impl EndpointStateActor {
             }
             self.open_path(addr);
             self.close_redundant_paths(addr);
+            self.pub_selected_path.set(Some(addr.clone().into())).ok();
```
flub, line 964: I don't follow why we store selected path twice?  Can't we use the watchable directly?  Or is that too horrible?

iroh/src/magicsock/endpoint_map/endpoint_state.rs
```diff
@@ -1020,6 +1048,9 @@ pub(crate) enum EndpointStateMessage {
     /// TODO: This is more of a placeholder message currently.  Check MagicSock::latency.
     #[debug("Latency(..)")]
     Latency(oneshot::Sender<Option<Duration>>),
+    /// Returns info about currently open connections to the remote endpoint.
+    #[debug("Latency(..)")]
```
flub, line 1052: not latency :)

iroh/src/magicsock/endpoint_map/endpoint_state.rs
```diff
@@ -1085,6 +1117,14 @@ struct ConnectionState {
...
     /// The information we publish to users about the paths used in this connection.
     pub_path_info: Watchable<PathsInfo>,
+    /// The ALPN of the connection.
+    ///
+    /// We store this to be able to create a [`ConnectionInfo`] from the [`ConnectionState`].
+    alpn: Vec<u8>,
+    /// The side (client or server)
+    ///
+    /// We store this to be able to create a [`ConnectionInfo`] from the [`ConnectionState`].
+    side: Side,
```
flub, line 1127: So these exist purely for diagnostics right, nothing on the actor needs this info.

I'm not entirely happy about that, but can live with it. Would move them to their own section and clearly document them as such though.

iroh/src/endpoint/connection.rs
Frando: Fully. Haven't done docs properly at all yet, wanna settle the approach first.

iroh/src/magicsock/endpoint_map/endpoint_state.rs
Frando: I think this one here is much less problematic than the Vec for the remote infos. While busy machines will definitely have a lot of remotes, the number of connections to a single other endpoint will be smallish in the vast majority of uses. But yes, not allocating would definitely be better.

iroh/src/magicsock/endpoint_map/endpoint_state.rs
Frando: We can likely, yes.

Frando:
> So my biggest problem is that I don't think we can allocate a vector with an item (whatever the item is) for each connection. I guess that gives two main options:
> 
>     * Not provide any API that needs this.  I know you won't like this :)
> 
>     * Figure out how to make those this available as a liftime-disconnected iterator or something.  I know I won't like this :)
> 
>     * ???

First: Sure, not doing it is always an option. For me the monitor API in #3610 is more important, and that doesn't have any of these issues, because we simply hand out a `ConnectionInfo` at time of creation, and don't care any further. If users need a map of all connections or all remotes, they can build it up from there.

So we can, I think, reasonably decide that the monitor callback is the only introspection mechanism.

(#3610 and this PR share the `ConnectionInfo` struct, otherwise they're mostly independent, which one is based on which is rather arbitrary)

If we *do* want to stick to a way to iterate over active remotes and connections, well, I think we can

a) allocate the full list: fine for smaller nodes, not fine for those with many connected nodes.
b) lock on each iteration step: will have to evaluate closer what the effects would be
c) look into fancier concurrent datastructures like [dashmap](https://docs.rs/dashmap/) or [papaya](https://docs.rs/papaya/), or persistent datastructures with cheap clones like those from [rpds](https://docs.rs/rpds/latest/rpds/)

dashmap can deadlock so we'd have to be very careful how to expose it in such a way that it cannot deadlock by users holding it wrong. papaya I recently stumbled upon and could be made to work, I think. However, we'd have again to be very careful that users holding it slightly wrong don't interfere with interior performance of the endpoint. I'll do some more thinking and will have a closer look at some concurrent maps maybe.

flub:
So in my head we've passed this chapter and the current impl of things with hooks etc allows all functionality we want to provide for now. Is that correct? Do we still need more or can this be closed?

Frando:
Yes, let's close this.

PR #3615, opened 2025-11-06
fix: stop polling transports when the magicsock is closing

Frando:
## Description

Since we merged #3601 we sometimes get ugly (but harmless) error log messages when the endpoint is shutting down:
```
2025-11-06T11:28:26.721826Z ERROR ep-client:client: iroh::magicsock::transports::relay: relay_recv_channel closed
2025-11-06T11:28:26.721884Z ERROR ep-client:client: iroh_quinn::endpoint: I/O error: connection closed
```
This fixes it: We already have a check in the transports to return `Pending` without polling the actual transports once `Magicsock::is_closed` is `true`. This PR changes it to instead check `Magicsock::is_closing`. The latter is set **after** `Endpoint::wait_idle` completed, but before cancelling the actor and transport tasks. Whereas `is_closed` is set *after* the actor and transport tasks are fully shutdown. When `Endpoint::wait_idle` completes, we do not expect the quinn endpoint to do anything meaningful, so it is fine to stop polling the transports at this time, and thus not get errors from the transports into quinn if they fail (because they are shutting down).

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/magicsock/transports.rs
+++ b/iroh/src/magicsock/transports.rs
@@ -78,7 +78,7 @@ impl Transports {
         msock: &MagicSock,
     ) -> Poll<io::Result<usize>> {
         debug_assert_eq!(bufs.len(), metas.len(), "non matching bufs & metas");
-        if msock.is_closed() {
+        if msock.is_closing() {
             return Poll::Pending;
         }
 


flub (approved):
A separate question is why MagicSock has both `is_closing` and `is_closed`.  But this makes sense, thanks!

Frando:
> A separate question is why MagicSock has both `is_closing` and `is_closed`. But this makes sense, thanks!

Yeah, I was tempted to clean up the closing variables a bit further, but will leave that to a separate PR.

PR #3619, opened 2025-11-06
feat: Instead of duplicating the entire connection API, have a type parameter

rklaehn:
## Description

on Connection that describes the connection state, and implement the fns alpn and remote_id differently depending on what the state is.

The upside is that it should be easier to write code that works with both a 0rtt connection and a fully initialized connection without having to define a trait to abstract over both variants.

The downside is that it is a bit fancy with the types. But then for the non 0rtt case you would never see the fanciness, so maybe it is OK?

Todo:
- remove everything but connection and handshake_completed from Incoming/Outcoming...
- make the ConnectionState trait sealed
- document the ConnectionState instances

Note:

if you have a handler that needs to work with both handshake completed connections and non handshake completed connections, you could now just do:

```rust
async fn handler<T: ConnectionState>(conn: Connection<T>) { ... }
```

## Breaking Changes

- `iroh::endpoint::Connection` gets a type parameter to track connection state. Default is `Connection<HandshakeCompleted>`
- `iroh::endpoint::IncomingZeroRttConnection` is removed, replaced with type alias to `Connection<IncomingZeroRtt>` 
- `iroh::endpoint::OutgoingZeroRttConnection` is removed, replaced with type alias to `Connection<OutgoingZeroRtt>`

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/magicsock/transports/relay/actor.rs
+++ b/iroh/src/magicsock/transports/relay/actor.rs
@@ -1466,6 +1466,7 @@ mod tests {
 
     #[tokio::test]
     #[traced_test]
+    #[ignore = "flaky"]
     async fn test_active_relay_inactive() -> Result {
         let (_relay_map, relay_url, _server) = test_utils::run_relay_server().await?;
 


rklaehn:
There are 3 ways to do the two Incoming/Outgoing... structs:

1. have an explicit .connection() accessor on Incoming/Outgoing... (what I implemented)
2. Use Deref<Connection<IncomingZeroRtt>> (this seems to be non-idiomatic rust even though I use this quite a bit in blobs)
3. Eliminate IncomingZeroRttConnection and OutgoingZeroRttConnection and instead have a state enum inside Connection that contains either a future or an (EndpointID, Alpn) tuple. That way you could implement handshake_completed directly on Connection<IncomingZeroRtt> and Connection<OutgoingZeroRtt>

Maybe the last one is the way to go?

iroh/src/endpoint/connection.rs
```diff
@@ -581,34 +585,84 @@ pub enum ZeroRttStatus {
...
+    },
+    IncomingZeroRtt {
+        accepted: Shared<ZeroRttAccepted>,
+    },
+    OutgoingZeroRtt {
+        accepted: Shared<ZeroRttAccepted>,
+    },
+}
+
+pub trait ConnectionState {}
```
dignifiedquire: we should probably make this trait sealed if we go this route

matheus23:
Playing with this PR a bit, I think `trait ConnectionState`, and the structs `HandshakeCompleted`, `IncomingZeroRtt` and `OutgoingZeroRtt` need to be exposed in `iroh/endpoint.rs`. Although they're `pub` in `connection.rs`, the connection module isn't exposed itself.

matheus23:
Also, it would be nice if we'd be able to downcast a `Connection<T: ConnectionState>` into some kind of enum that is one of either of the connection states.

E.g. that'd be useful to implement getting the remote id on a function that's generic over `ConnectionState` in case that connection state happens to be `HandshakeCompleted` (irpc-iroh uses that to optionally add the remote ID to logs).

iroh/src/endpoint/connection.rs
```diff
@@ -581,34 +585,93 @@ pub enum ZeroRttStatus {
...
+/// application. [`Connection::close`] describes in more detail how to gracefully close a
+/// connection without losing application data.
+///
+/// May be cloned to obtain another handle to the same connection.
+#[derive(derive_more::Debug, Clone)]
+pub struct Connection<State: ConnectionState = HandshakeCompleted> {
+    inner: quinn::Connection,
+    /// State-specific information
+    info: State::Data,
+    _p: std::marker::PhantomData<State>,
```
Frando: I think this can be removed now that we have `info: State:::Data`

rklaehn:
> Also, it would be nice if we'd be able to downcast a `Connection<T: ConnectionState>` into some kind of enum that is one of either of the connection states.
> 
> E.g. that'd be useful to implement getting the remote id on a function that's generic over `ConnectionState` in case that connection state happens to be `HandshakeCompleted` (irpc-iroh uses that to optionally add the remote ID to logs).

We could have a fourth state `Generic` or `AnyState` where the state is an enum, but it would be some work.

Maybe instead add methods to the trait that work in any case and return Option/Result?

rklaehn:
> > Also, it would be nice if we'd be able to downcast a `Connection<T: ConnectionState>` into some kind of enum that is one of either of the connection states.
> > E.g. that'd be useful to implement getting the remote id on a function that's generic over `ConnectionState` in case that connection state happens to be `HandshakeCompleted` (irpc-iroh uses that to optionally add the remote ID to logs).
> 
> We could have a fourth state `Generic` or `AnyState` where the state is an enum, but it would be some work.
> 
> Maybe instead add methods to the trait that work in any case and return Option/Result?

@matheus23 so WDYT, should I just add endpoint_opt and alpn_opt that works for any state and be done with it, or do something more fancy to keep the API surface smaller?

matheus23:
I was thinking of something like this:

```patch
diff --git a/iroh/src/endpoint/connection.rs b/iroh/src/endpoint/connection.rs
index 6a4593adc..7f527708a 100644
--- a/iroh/src/endpoint/connection.rs
+++ b/iroh/src/endpoint/connection.rs
@@ -646,9 +646,20 @@ mod sealed {
 }
 
 /// Trait to track the state of a [`Connection`] at compile time.
-pub trait ConnectionState: sealed::Sealed {
+pub trait ConnectionState: sealed::Sealed + Sized {
     /// State-specific data stored in the [`Connection`].
     type Data: std::fmt::Debug + Clone;
+
+    /// Turns an arbitrary connection into an enum of possible typed connection states
+    fn into_state_enum(conn: Connection<Self>) -> StateEnum;
+}
+
+/// TODO docs
+#[derive(Debug)]
+pub enum StateEnum {
+    HandshakeCompleted(Connection<HandshakeCompleted>),
+    IncomingZeroRtt(Connection<IncomingZeroRtt>),
+    OutgoingZeroRtt(Connection<OutgoingZeroRtt>),
 }
 
 /// Marker type for a connection that has completed the handshake.
@@ -666,14 +677,26 @@ pub struct OutgoingZeroRtt;
 impl sealed::Sealed for HandshakeCompleted {}
 impl ConnectionState for HandshakeCompleted {
     type Data = HandshakeCompletedData;
+
+    fn into_state_enum(conn: Connection<Self>) -> StateEnum {
+        StateEnum::HandshakeCompleted(conn)
+    }
 }
 impl sealed::Sealed for IncomingZeroRtt {}
 impl ConnectionState for IncomingZeroRtt {
     type Data = IncomingZeroRttData;
+
+    fn into_state_enum(conn: Connection<Self>) -> StateEnum {
+        StateEnum::IncomingZeroRtt(conn)
+    }
 }
 impl sealed::Sealed for OutgoingZeroRtt {}
 impl ConnectionState for OutgoingZeroRtt {
     type Data = OutgoingZeroRttData;
+
+    fn into_state_enum(conn: Connection<Self>) -> StateEnum {
+        StateEnum::OutgoingZeroRtt(conn)
+    }
 }
 
 #[allow(missing_docs)]
@@ -686,6 +709,11 @@ impl<T: ConnectionState> Connection<T> {
         &self.inner
     }
 
+    /// TODO docs
+    pub fn into_state_enum(self) -> StateEnum {
+        T::into_state_enum(self)
+    }
+
     /// Initiates a new outgoing unidirectional stream.
     ///
     /// Streams are cheap and instantaneous to open unless blocked by flow control. As a
```

Where `into_state_enum` works like a sort of downcasting.

But that'll cost us another exposed enum that we need to name and another exposed function on the `ConnectionState` trait.

Perhaps adding a `try_remote_id` and `try_alpn` or sth like that on `Connection<T>`.

rklaehn:
> Where `into_state_enum` works like a sort of downcasting.
> 
> But that'll cost us another exposed enum that we need to name and another exposed function on the `ConnectionState` trait.
> 
> Perhaps adding a `try_remote_id` and `try_alpn` or sth like that on `Connection<T>`.

I think the state enum will only be used very rarely, so my current thinking is that we don't do anything for now. The state enum is something people can define for themselves without much friction, much less annoying than having to define a trait.

And even the try_remote_id and try_alpn are not strictly necessary. People can always do the match themselves before they go from concrete Connection<Whatever> to an abstract handler.

How about we merge this now, frando and me hack it into the places where it is needed, which are already quite esoteric, and then we decide if we need more simplification.

*If* we decide to do something about this, one option would also be to have a 4th state "generic". That would be the same number of additional public types, but would make usage nicer than an enum.

E.g. for open_bi with the enum you would have to do a match, open_bi with the 4th state you could just use as normal.

But again, I think there is not that much need to solve this right now. 1.0 means we can't modify, but we can still add syntax candy like this.

matheus23:
Tried out these changes here: https://github.com/n0-computer/irpc/pull/84

There's [one place](https://github.com/n0-computer/irpc/pull/84/files#diff-abb528a202202f13affaae3f03e2cb8e105a5dc82cacd8666bc7d388e46389f4R291-R294) where I'd like to have the remote endpoint ID from a generic `Connection<impl ConnectionState>`, but it's only for logging, and I think one could work around this by passing it from the outside in.

dignifiedquire:
Tried this out: not sure it would be worth it: https://github.com/n0-computer/iroh/tree/connection-state-try-remote

Frando:
Let's merge it as-is, and add more sugar later if we feel the need?

Frando (approved):

PR #3621, opened 2025-11-07
fix(iroh-bench): Gracefully close the quinn benchmark

matheus23:
## Description

Adds a call to `quinn::Endpoint::wait_idle` as the equivalent to `iroh::Endpoint::close`.

This fixes the quinn benchmark not timing out, but closing gracefully instead for me.
(Before, running `cargo run -p iroh-bench -- quinn` would take 18s to finish due to quinn waiting for a probe timeout, after this, this bench runs in 2s for me.)

## Change checklist

- [x] Self-review.


dignifiedquire (approved):
thanks, this was getting annyoing

PR #3622, opened 2025-11-07
refactor: improve path watching, add path stats

Frando:
## Description

This improves how we expose paths and path stats for connections, and also updates feat-multipath to use https://github.com/n0-computer/quinn/pull/168.

* The watcher for open paths internally uses a SmallVec to not allocate in the common case of not-too-many paths
* The path info for a path now includes a boolean whether this is the currently selected primary transmission path
* We no longer expose PathIds to users
* We expose stats for paths from `Connection::path_stats`

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/endpoint/connection.rs
+++ b/iroh/src/endpoint/connection.rs
@@ -1466,6 +1466,9 @@ impl Connection {
     ///
     /// The [`PathInfoList`] returned from the watcher contains a [`PathInfo`] for each
     /// transmission path.
+    ///
+    /// [`PathInfo::is_selected`]: crate::magicsock::PathInfo::is_selected
+    /// [`PathInfo`]: crate::magicsock::PathInfo
     pub fn paths(&self) -> impl Watcher<Value = PathInfoList> {
         self.paths.watch(self.inner.weak_handle())
     }
--- a/iroh/src/magicsock/endpoint_map/endpoint_state.rs
+++ b/iroh/src/magicsock/endpoint_map/endpoint_state.rs
@@ -1224,10 +1224,7 @@ impl PathInfoList {
 
 /// Information about a network path used by a [`Connection`].
 ///
-/// To get further information, like the latency, use [`Connection::path_stats`].
-///
 /// [`Connection`]: crate::endpoint::Connection
-/// [`Connection::path_stats`]: crate::endpoint::Connection::path_stats
 #[derive(derive_more::Debug, Clone)]
 pub struct PathInfo {
     path_id: PathId,


iroh/src/magicsock/transports.rs
```diff
@@ -387,6 +387,13 @@ impl From<Addr> for TransportAddr {
 }
 
 impl Addr {
+    pub(crate) fn from_transport_addr(endpoint_id: EndpointId, addr: TransportAddr) -> Self {
+        match addr {
+            TransportAddr::Relay(relay_url) => Self::Relay(relay_url, endpoint_id),
+            TransportAddr::Ip(socket_addr) => Self::Ip(socket_addr),
+            _ => unreachable!(),
+        }
+    }
```
flub: Missing blank line.

But why isn't this a `From` impl?  It's an internal type anyway so it's not leaking into the public API?

iroh/src/endpoint/connection.rs
```diff
@@ -1483,10 +1462,29 @@ impl Connection {
...
+            .iter()
+            .find_map(|(path_addr, path_id)| (path_addr == addr).then_some(*path_id))?;
+        self.inner.path_stats(path_id)
+    }
+
+    /// Returns an iterator over the statistics for all networks path used by this connection.
+    pub fn path_stats_iter(&self) -> impl Iterator<Item = (PathInfo, PathStats)> {
+        self.paths
+            .iter()
+            .flat_map(|(path_id, info)| self.inner.path_stats(path_id).map(|stats| (info, stats)))
```
flub: These two are kind of ugly.  I imagined `PathInfo::stats` to do this instead.  It does mean PathInfo needs to have a WeakRef to the connection, does it add extra complexity?

iroh/src/magicsock/endpoint_map/endpoint_state.rs
```diff
@@ -130,7 +131,7 @@ pub(super) struct EndpointStateActor {
     /// holepunching regularly.
     ///
     /// We only select a path once the path is functional in Quinn.
-    selected_path: Option<transports::Addr>,
+    selected_path: Watchable<Option<TransportAddr>>,
```
flub: Oh right, `transports::Addr` vs `TransportAddr` is the reason why you have two fields before?  Yeah, not sure if I want the code here to have to deal with this all the time.  The code in this actor has to be as simple and logical as possible.

Is the conversion from `transports::Addr` -> `TransportAddr` infallible?  Can the user-facing watcher not use a `Watcher::map` on this?  (I've not tried to use that feature myself before so could be off the mark)

iroh/src/magicsock/endpoint_map/endpoint_state.rs
```diff
@@ -1137,39 +1149,113 @@ impl ConnectionState {
...
+impl PathInfoList {
+    /// Returns an iterator over the path infos.
+    pub fn iter(&self) -> impl Iterator<Item = &PathInfo> {
+        self.0.iter()
+    }
+
+    /// Returns the [`PathInfo`] for a transport address.
+    pub fn get(&self, addr: &TransportAddr) -> Option<&PathInfo> {
+        self.0.iter().find(|info| info.remote_addr() == addr)
+    }
```
flub: I think in rust you don't provide this kind of helper because it misleads you about performance.  The user can do this themselves, no?

iroh/src/magicsock/transports.rs
Frando: Because it needs the `EndpointId`. Could be a `From<(EndpointId, TransportAddr)>` - not sure if that's worth it but can do.

iroh/src/endpoint/connection.rs
Frando: Yes, I thought about this. Passing around the `WeakConnection` is not an issue but `PathInfo::stats` would need to return an `Option` for when upgrading fails. For the iterator, we can `flat_map` this, which is nice. That why I opted for this. But I'm torn too.

iroh/src/magicsock/endpoint_map/endpoint_state.rs
Frando: It is infallible. Yes, I can try to swap this around.

iroh/src/magicsock/endpoint_map/endpoint_state.rs
Frando: Yes, I was unsure too. I can remove it again.

iroh/src/magicsock/transports.rs
Frando: Removed this because it became unused after swapping the selected_path watcher to use `transports::Addr`

iroh/src/magicsock/endpoint_map/endpoint_state.rs
Frando: I swapped this around, the selected_path watcher is now over `transports::Addr`.

iroh/src/magicsock/endpoint_map/endpoint_state.rs
Frando: Removed it.

iroh/src/magicsock/endpoint_map/endpoint_state.rs
```diff
@@ -1137,39 +1145,160 @@ impl ConnectionState {
...
+    pub fn is_relay(&self) -> bool {
+        self.remote.is_relay()
+    }
+
+    /// Returns stats for this transmission path.
+    pub fn stats(&self) -> PathStats {
+        self.handle
+            .upgrade()
+            .and_then(|conn| conn.path_stats(self.path_id))
+            .unwrap_or(self.stats)
```
flub, line 1292: Erm, you now just have these as a field?

iroh/src/magicsock.rs
```diff
@@ -290,12 +303,15 @@ impl MagicSock {
         // send this.  Before the handshake has completed we don't have anything useful to
         // do with this connection inside of the EndpointStateActor anyway.
         let weak_handle = conn.weak_handle();
-        let endpoint_state = self.endpoint_map.endpoint_state_actor(remote);
-        let msg = EndpointStateMessage::AddConnection(weak_handle, paths_info);
+        let (sender, selected_path) = self
```
flub, line 306: I'm going to try complaining about this bit once again :)

You're introducing a 2nd way of passing a watcher/watchable to/from the actor.  The selected path is passed from the actor to here while the open_paths is passed the other way round.

I'm not particularly fond of how open_paths is hooked up, I considered it a hack when I did it as well.  But it was an attempt to keep things in the sync world that could stay there.  I prefer it slightly over the way you just introduced because we don't need to push this selected_path watcher through the handle and have different constructor methods for the actor?  In either case, having two ways of doing this is a bit weird?

Frando: Difference is that the selected_path is per EndpointStateActor while open_paths is per Connection. I also didn't really like my solution, I'll see if I can come up with something better.

iroh/src/magicsock/endpoint_map/endpoint_state.rs
Frando: But they might be outdated. The field would only get refreshed when the open paths change.

PR #3624, opened 2025-11-07
perf: various improvements

dignifiedquire:


iroh/src/magicsock.rs
```diff
@@ -614,11 +617,9 @@ impl MagicSock {
                 // relies on quinn::EndpointConfig::grease_quic_bit being set to `false`,
                 // which we do in Endpoint::bind.
                 if let Some((sender, sealed_box)) = disco::source_and_box(datagram) {
-                    trace!(src = ?source_addr, len = datagram.len(), "UDP recv: DISCO packet");
```
matheus23, line 581: Do these show up in perf?

But yeah - perhaps it's fair to remove these.

dignifiedquire: they do 😭

flub: they're removed in the QNT branch too :)

But really, you should be benching with the tracing static verbosity levels off, no?  i.e. use `max_level_off` and/or `release_max_level_off`: https://docs.rs/tracing/latest/tracing/level_filters/index.html#compile-time-filters

dignifiedquire: cool, I didn't know about this, but in practice this will show up for folks though

iroh/src/magicsock/transports.rs
```diff
@@ -34,18 +34,19 @@ pub(crate) struct Transports {
     ip: Vec<IpTransport>,
     relay: Vec<RelayTransport>,
 
-    poll_recv_counter: AtomicUsize,
+    poll_recv_counter: usize,
+    source_addrs: tinyvec::TinyVec<[Addr; 4]>, // cache for source addrs
```
matheus23: 4 might be a little conservative, depending on if your system has GSO enabled or not. E.g. I consistently get `32` as the value for `metas.len()`.

matheus23: Actually, that's a `pub const` in `quinn-udp`: `BATCH_SIZE`. Perhaps this should just refer to that constant?

dignifiedquire: I don't want to fill the stack that much, we already fill it quite a bit. So I would only want to increase this if we have benchmarks showing this is clearly worth it

matheus23: 32 * `Addr` = 1280 bytes. The `Transport` struct is put on the stack in the `async fn Handle::new`. Those cases *can* cause ballooning stack sizes when the struct has to be held across an await point, which currently isn't the case.
In fact, it's moved into `MagicTransport::new` which immediately gets `Box::new`d.

I don't think that 1.3kB on the stack is too bad.

But also, given it's moved into a `Box`, doesn't that mean it'll live on the heap anyways?
Is there a performance difference to using a `Vec`?

dignifiedquire: > Is there a performance difference to using a `Vec`?

yeah, at least in my tests

flub: I think it would be good to leave a comment *why* this was put here.  Otherwise I can imagine it being removed easily in a future refactor.

iroh/src/magicsock/transports.rs
```diff
@@ -84,11 +86,11 @@ impl Transports {
             return Poll::Pending;
         }
 
-        let mut source_addrs = vec![Addr::default(); metas.len()];
-        match self.inner_poll_recv(cx, bufs, metas, &mut source_addrs)? {
+        self.source_addrs.resize_with(metas.len(), Addr::default);
```
flub: Would making this an array also have been an option?  You know the exact size... Though if it's too big you don't want it to be on the stack?

iroh/src/magicsock.rs
```diff
@@ -550,11 +552,12 @@ impl MagicSock {
 
         let mut quic_packets_total = 0;
 
-        for ((quinn_meta, buf), source_addr) in metas
-            .iter_mut()
-            .zip(bufs.iter_mut())
-            .zip(source_addrs.iter())
-        {
+        // zip is slow :(
+        for i in 0..metas.len() {
```
flub, line 556: I wonder if it is because of the double zip.  The QNT branch only has a single zip here I think.

flub (approved):

iroh/src/magicsock.rs
matheus23: It does seem like zip generally optimizes badly: https://github.com/rust-lang/rust/issues/143966#issuecomment-3075931838

iroh/src/magicsock/transports.rs
dignifiedquire: the exact size is runtime dependent, so I can not make it an array?

matheus23: Well, you know the maximum size will always be `quinn_udp::BATCH_SIZE`.

matheus23: I don't think that matters though because `Addr` isn't/cannot be `Copy`

iroh/src/magicsock/transports.rs
dignifiedquire: added a comment and switched to an array for now

PR #3625, opened 2025-11-07
use associated type to avoid using unreachable

dignifiedquire:


iroh/src/endpoint/connection.rs
```diff
@@ -622,25 +622,28 @@ pub type IncomingZeroRttConnection = Connection<IncomingZeroRtt>;
...
-    OutgoingZeroRtt {
-        accepted: Shared<ZeroRttAccepted>,
-    },
+pub struct HandshakeCompletedData {
+    endpoint_id: EndpointId,
+    alpn: Vec<u8>,
+}
+
+#[derive(Debug, Clone)]
+pub struct IncomingZeroRttData {
```
rklaehn, line 636: Does this approach require these to be pub? I don't think so.

rklaehn:
Hm, I tried a few things but still can't figure out how to avoid making the ...Data structs pub.

PR #3626, opened 2025-11-07
refactor: minor cleanups in endpoint state

Frando:
## Description

This has a few minor cleanups without any functional changes in the endpoint state actor:
* Remove double handle upgrade
* Add helper function `to_transport_addr` on the relay mapped addr map
* Use hash map `entry` API instead of `get` and `expect`
* Use `if let` chains to remove a level of indentation

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/magicsock/mapped_addrs.rs
+++ b/iroh/src/magicsock/mapped_addrs.rs
@@ -286,30 +286,37 @@ impl<K, V> Default for AddrMapInner<K, V> {
     }
 }
 
-/// Methods for the relay mapped address map.
+/// Functions for the relay mapped address map.
 impl AddrMap<(RelayUrl, EndpointId), RelayMappedAddr> {
     /// Converts a mapped socket address to a transport address.
     ///
-    /// To map to a [`transports::Addr`] only the [`RelayMappedAddr`] needs to be able to be
-    /// looked up.  So we can have this as a helper here.
+    /// This takes a socket address, converts it into a [`MultipathMappedAddr`] and then tries
+    /// to convert the mapped address into a [`transports::Addr`].
     ///
-    /// Returns `None` if the mapped address is a [`MultipathMappedAddr::Mixed`], or if the mapped
-    /// address is a [`MultipathMappedAddr::Relay`] and `self` does not contain the mapped address.
+    /// Returns `Some` with the transport address for IP mapped addresses and for relay mapped
+    /// addresses if an entry for the mapped address exists in `self`.
+    ///
+    /// Returns `None` and emits an error log if the mapped address is a [`MultipathMappedAddr::Mixed`],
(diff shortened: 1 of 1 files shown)


iroh/src/magicsock/endpoint_map/endpoint_state.rs
```diff
@@ -341,60 +341,53 @@ impl EndpointStateActor {
...
-                                .keys()
-                                .filter(|a| a.is_relay())
-                                .cloned()
-                                .collect::<Vec<_>>();
-                            for remote in relays {
-                                self.open_path(&remote);
-                            }
-                        }
+            if let Some(path) = conn.path(PathId::ZERO)
+                && let Some(path_remote) = path_remote(&self.relay_mapped_addrs, &path)
```
flub: TIL you can combine `let Some(x)` bindings like this.

iroh/src/magicsock/endpoint_map/endpoint_state.rs
```diff
@@ -1298,6 +1289,15 @@ impl PathInfo {
     }
 }
 
+fn path_remote(
```
flub: doc comment please.

I've gone back and forth on this.  I used to have a helper for this at some point :)  But I never like this kind of helper that just takes a bunch of random things by ref.

What about:
```rust
impl AddrMap<(RelayUrl, EndpointId), RelayMappedAddr> {
    /// Convert a mapped socket address to a transport addr.
    ///
    /// To map to a [`transports::Addr`] only the [`RelayMappedAddr`] needs to be able to be
    /// looked up.  So we can have this as a helper here.
    fn to_transport_addr(&self, sockaddr: SocketAddr) -> Option<transports::Addr> {
        MultipathMappedAddr::from(sockaddr).to_transport_addr(self)
    }
}
```

And probably `MultipathMappedAddr::to_transport_addr` should be removed and merged into this impl.  It's just going to sow confusion over there.

flub: I've also always been a little on the fence on whether creating the `AddrMap` was a useful abstraction level.  I'm not that sure.

I think now I'd try and make a single AddrMap struct (not trait) that contained both the mapped addrs in a single lock.  It would be a much more natural place for this conversion to live.  But anyway, we're here now and this isn't the most blocking thing.

iroh/src/magicsock/endpoint_map/endpoint_state.rs
Frando: Yes it's really nice. It's quite new, added in 2024 edition.

iroh/src/magicsock/endpoint_map/endpoint_state.rs
Frando: sgtm! pushed this.

flub (approved):

PR #3629, opened 2025-11-10
refactor(multipath): Make registering connections with the magicsock async

Frando:
## Description

Currently, `Magicsock::register_connection` is a sync function, but needs to send over an async channel to notify the endpoint state actor about the new connection. It currently employs a hack to achieve that: it spawns a tokio task for sending the message. 

This PR cleans this up by making `regsiter_connection` return a future, and awaits this future at the various sites where we go from quinn::Connection to iroh Connection. Luckily, all these call sites already are in async contexts.

* When going from `Connecting` or `Accepting` to `Connection`, we await the registration after having the `quinn::Connecting` completes. The future is stored in an option instead of using a state enum as you would usually, because we need unconditional access to the `quinn::Connecting` in the functions on  `Connecting`/`Accepting`.
* For the `(Incoming|Outgoing)ZeroRttConnection`, we store a future that first awaits the handshake and then registers the connection. So we need only a single future here.

With `register_connection` being async, we can also clean up some of the not-so-nice things introduced in #3622: Because we now have an async function, we can let the endpoint state actor return a reply. This makes it much more straightforward because we can have the endpoint state actor initialize a watcher for the paths and return it instead of having to do a weird dance with parts of the state being initialized or stored outside of the endpoint state actor to satisfy the sync function constraints. This is much nicer now IMO.

## Notes & open questions

This adds a boxed future into the process of going from a  `Connecting` to a `Connection`. If we really wanted, we could use a manually implemented future instead. However, I don't think one boxed future *per connection*  is an issue, so I'd prefer to leave it like this (implementing a manual future for `tokio::mpsc::Sender::send` is cumbersome).

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/endpoint/connection.rs
+++ b/iroh/src/endpoint/connection.rs
@@ -29,7 +29,7 @@ use ed25519_dalek::{VerifyingKey, pkcs8::DecodePublicKey};
 use futures_util::{FutureExt, future::Shared};
 use iroh_base::EndpointId;
 use n0_error::{e, stack_error};
-use n0_future::{boxed::BoxFuture, time::Duration};
+use n0_future::{TryFutureExt, boxed::BoxFuture, time::Duration};
 use n0_watcher::Watcher;
 use pin_project::pin_project;
 use quinn::{
@@ -41,7 +41,10 @@ use tracing::warn;
 use crate::{
     Endpoint,
     discovery::DiscoveryTask,
-    magicsock::endpoint_map::{PathInfoList, PathsWatchable},
+    magicsock::{
+        EndpointStateActorStoppedError,
+        endpoint_map::{PathInfoList, PathsWatchable},
+    },
 };
 
 /// Future produced by [`Endpoint::accept`].
@@ -153,14 +156,18 @@ impl IntoFuture for Incoming {
     type IntoFuture = IncomingFuture;
 
     fn into_future(self) -> Self::IntoFuture {
-        IncomingFuture(Box::pin(establish(self.ep, self.inner.into_future())))
+        IncomingFuture(Box::pin(async move {
+            let quinn_conn = self.inner.into_future().await?;
+            let conn = conn_from_quinn_conn(quinn_conn, &self.ep)?.await?;
+            Ok(conn)
+        }))
     }
 }
 
 /// Adaptor to let [`Incoming`] be `await`ed like a [`Connecting`].
 #[derive(derive_more::Debug)]
 #[debug("IncomingFuture")]
-pub struct IncomingFuture(EstablishFuture);
+pub struct IncomingFuture(BoxFuture<Result<Connection, ConnectingError>>);
 
 impl Future for IncomingFuture {
     type Output = Result<Connection, ConnectingError>;
@@ -192,6 +199,7 @@ async fn alpn_from_quinn_connecting(conn: &mut quinn::Connecting) -> Result<Vec<
 #[stack_error(add_meta, derive, from_sources)]
 #[allow(missing_docs)]
 #[non_exhaustive]
+#[derive(Clone)]
 pub enum AuthenticationError {
     #[error(transparent)]
     RemoteId { source: RemoteEndpointIdError },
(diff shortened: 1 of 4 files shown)


iroh/src/endpoint/connection.rs
```diff
@@ -220,30 +211,56 @@ pub enum AuthenticationError {
         #[error(std_err)]
         source: ConnectionError,
     },
+    #[error("endpoint state actor stopped")]
+    EndpointStateActorStopped,
```
flub: This is kind of ugly.  I certainly wouldn't call it `EndpointstateActorStopped` because this is user-facing and that name has nothing to do with anything the user knows about.

Maybe something like `InternalConsistencyError` and "internal consistency error: EndpointStateActor stopped"?  That way the user might have a clue that their only recourse is to file a bug.

iroh/src/endpoint/connection.rs
```diff
@@ -220,30 +211,56 @@ pub enum AuthenticationError {
...
+}
+
+impl From<EndpointStateActorStoppedError> for AuthenticationError {
+    #[track_caller]
+    fn from(_value: EndpointStateActorStoppedError) -> Self {
+        e!(Self::EndpointStateActorStopped)
```
flub: Is this losing the original location by doing this?

iroh/src/endpoint/connection.rs
```diff
@@ -220,30 +211,56 @@ pub enum AuthenticationError {
...
+///
+/// The returned future is `'static`, so it can be stored without being lifetime-bound on `&ep`.
 fn conn_from_quinn_conn(
     conn: quinn::Connection,
     ep: &Endpoint,
-) -> Result<Connection, AuthenticationError> {
+) -> Result<
+    impl Future<Output = Result<Connection, EndpointStateActorStoppedError>> + Send + 'static,
+    AuthenticationError,
+> {
```
flub, line 248: Just curious: is the reason you're not making this an `async fn` because you need to be able to specify the `'static`?  Having it spelled out does make things easier to debug when the 'static bound is lost at some point so I don't think you need to change this.  Just trying to see if I'm understanding things correctly.

iroh/src/endpoint/connection.rs
```diff
@@ -576,12 +577,11 @@ impl Future for ZeroRttAccepted {
...
 pub struct OutgoingZeroRttConnection {
     inner: quinn::Connection,
-    accepted: Shared<ZeroRttAccepted>,
-    ep: Endpoint,
+    handshake_completed_fut: Shared<BoxFuture<Result<ZeroRttStatus, AuthenticationError>>>,
 }
 
 /// Returned from [`OutgoingZeroRttConnection::handshake_completed`].
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
flub, line 584: Does this have any consequences?  What do the folks who've been working on 0rtt think of this?

/cc @ramfox @matheus23 @rklaehn @dignifiedquire

flub (approved):
Very nice!

iroh/src/endpoint/connection.rs
Frando: Yeah, sounds good.

I checked the cases when the endpoint state actor stop. Right now, it only if it fails to send to the `TransportActor`, which only fails if the `TransportsSenderActor` stops, which it never does (its `run` is infallible) unless it panics. So we might even `expect()` it?

It will change though because we gotta start killing unused endpoint state actors, so then `expect` might be too harsh? Not sure. Will rename the error for now.

iroh/src/endpoint/connection.rs
Frando: Yes, your understanding is correc.t I want a future without a lifetime bound on `&ep`, and there's no way to express that with `async fn`.

iroh/src/endpoint/connection.rs
flub: Yeah, stopping the EndpointStateActor is something I was thinking about how to do just when you distracted me with weird behaviour to debug :)

It will have to stop.  I was trying to see if it's possible to stop this without making anything racy: the EndpointMap would have to forget about the actor while it is guaranteed to have no connections and at that point if a new one is needed you can create a new one.  A long way to say that this invariant needs to be upheld anyway.

Though I still think I prefer this over panicking.  Quinn's approach of panicking is a bit of a pain.  I always feel like panicking is a poor way out, the types should have made panicking impossible.  Anyway, also a long way of saying that I'm in favour of keeping this an error for now.

iroh/src/endpoint/connection.rs
flub: Feel free to merge now and fix this later if there's outcry.

Frando:
As noted by the `cargo clippy` check this currently increases the size of the `Connection` struct significantly because it stores two mapped watchers with a rather large value directly on the Connection struct. I don't think this is ideal, and I'm working out a solution atm.

Edit: I reverted the change to the path watchers. I'll handle them in another PR.

iroh/src/endpoint/connection.rs
Frando: I didn't add tracking to `EndpointStateActorStoppedError`. If we don't expose it anywhere, it is of little use. We could add it as a source to this error, but then we'd have to make the type public.

iroh/src/endpoint/connection.rs
matheus23: Personally I think this is fine. I don't think there's potential for API misuse coming from this `Clone`.

PR #3630, opened 2025-11-10
fix(iroh): typo(s)

matthiasbeyer:
Nothing special


matheus23 (approved):

matheus23:
Stuff currently is very much in flux as we're working on a major transition to QUIC multipath (See #3276) and currently mostly work off of `feat-multipath` instead of `main`.
Just letting you know that we might be holding off on merging this after the multipath stuff resolves to `main`.

flub:
Don't see a reason to delay this :)

PR #3631, opened 2025-11-10
refactor(multipath): Store `Watcher`, not `Watchable` on `Connection`

Frando:
## Description

Currently the `Connection` stores a `Watchable` for the open and selected paths. This means, though, that a watcher stream for the open paths will only close once the last connection is dropped, not once the connection is closed. That is bad API and will lead to confusion.

In an initial version, I swapped the Watchables for a `n0_watcher::Map<(Direct, Direct)>` but this increases the size of the `Connection` struct from ~80 to ~700 bytes, because it would store both the two unmapped values and the mapped final value inline. It is also completely unneeded, because we'd clone out the watcher anyways on `Connection::watch`  and `get` or `stream` on the watcher would refetch the value.

Instead, this now uses https://github.com/n0-computer/n0-watcher/pull/31 to only store a weak pointer to the watchable's shared state without keeping it alive, and then creates the PathInfos on demand from that. The `stats` are now fetched while iterating, so if you're e.g. only checking the number of paths, the stats won't be fetched at all. 

We fetch the paths through upgrading the handle in the `next` method, so that we can return `None` if the connection was dropped in the meantime (the alternative would be a fallible `PathInfo::stats` method)

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1121,7 +1121,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


Frando:
Closing this - https://github.com/n0-computer/n0-watcher/pull/31 is likely not going to be merged and #3632 was merged for now.

PR #3632, opened 2025-11-10
refactor: use boxed watcher, not watchable, on connection

Frando:
## Description

Alternative to https://github.com/n0-computer/iroh/pull/3631

Replaces the `Watchable`s for path changes on the `Connection` with a boxed `Watcher`. The watcher is boxed because it would increase the `Connection` struct size significantly otherwise because the mapped-and-joined watcher with a `SmallVec`  of  `PathInfo`  inside is ~600 bytes atm.

The benefit of storing a `Watcher` and not a `Watchable` is that the watcher streams now close once the EndpointStateActor drops the state for the connection, which it does after the connection is closed.

Also adds a test for path watching, including testing that the streams now close when the connection closes.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/magicsock/endpoint_map/endpoint_state.rs
+++ b/iroh/src/magicsock/endpoint_map/endpoint_state.rs
@@ -1140,6 +1140,7 @@ impl ConnectionState {
 /// [`Connection`]: crate::endpoint::Connection
 #[derive(Clone, derive_more::Debug)]
 #[debug("PathsWatcher")]
+#[allow(clippy::type_complexity)]
 pub(crate) struct PathsWatcher(
     Box<
         n0_watcher::Map<
--- a/iroh/src/magicsock/endpoint_map.rs
+++ b/iroh/src/magicsock/endpoint_map.rs
@@ -11,21 +11,18 @@ use serde::{Deserialize, Serialize};
 use tokio::sync::mpsc;
 use tracing::warn;
 
-use crate::disco;
-
+// #[cfg(any(test, feature = "test-utils"))]
+// use crate::endpoint::PathSelection;
+pub(super) use self::endpoint_state::EndpointStateMessage;
+pub(crate) use self::endpoint_state::PathsWatcher;
+use self::endpoint_state::{EndpointStateActor, EndpointStateHandle};
+pub use self::endpoint_state::{PathInfo, PathInfoList};
 use super::{
     DirectAddr, DiscoState, MagicsockMetrics,
     mapped_addrs::{AddrMap, EndpointIdMappedAddr, RelayMappedAddr},
     transports::{self, TransportsSender},
 };
-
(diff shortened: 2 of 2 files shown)


iroh/src/magicsock/endpoint_map/endpoint_state.rs
```diff
@@ -1143,38 +1144,65 @@ impl ConnectionState {
...
+    Box<
+        n0_watcher::Map<
+            (
+                n0_watcher::Direct<SmallVec<[(TransportAddr, PathId); 4]>>,
+                n0_watcher::Direct<Option<transports::Addr>>,
+            ),
+            PathInfoList,
+        >,
+    >,
+);
```
matheus23, line 1154: Should we add a small internal comment as to why this has a `Box`?

Frando: Added to the doc comment (the type is `pub(crate)`).

iroh/src/magicsock/endpoint_map/endpoint_state.rs
```diff
@@ -1143,38 +1144,70 @@ impl ConnectionState {
...
-pub(crate) struct PathsWatchable {
-    /// Watchable for the open paths (in this connection).
-    open_paths: Watchable<PathAddrList>,
-    /// Watchable for the selected transmission path (global for this remote endpoint).
-    selected_path: Watchable<Option<transports::Addr>>,
+#[derive(Clone, derive_more::Debug)]
+#[debug("PathsWatcher")]
+#[allow(clippy::type_complexity)]
+pub(crate) struct PathsWatcher(
+    Box<
```
flub, line 1145: Looking at the docs this would invoke this clone:
```rust
impl<T, A> Clone for Box<T, A>
where
    T: Clone,
    A: Allocator + Clone,
fn clone (&self) -> Box<T, A>
Returns a new box with a clone on this box's contents.
```

So on clone this makes a new allocation.  Isn't the inside `Send + Sync`?  Wouldn't this be more efficient as an `Arc`?  IIRC you would then not have to allocate a new watcher on the heap each time.

Frando: Yes, this clones the full watcher. An `Arc` wouldn't work as `Watcher::get` needs `&mut self`.

Frando:
I'd be in favor of merging this, it is the cleaner design for now. I don't like storing the `Watchable` in the `Connection` because this keeps the path watcher alive after the connection closed if there is still a strong clone of `Connection` somewhere.

iroh/src/magicsock/endpoint_map/endpoint_state.rs
flub: Ah, didn't realise that. :(

PR #3643, opened 2025-11-11
refactor(multipath): Stop inactive endpoint actors

### LINKED ISSUE #3638: Stop inactive EndpointStateActors
If there are no active connections we do not want to keep the EndpointStateActor around forever.

The exact conditions to shut an actor down are tricky, and depend on what is possible for shutting them down.

This **needs** to make sure that there's no race-condition when stopping an actor.  The EndpointMap should be handing out a sender to an actor that's shutting down but should ensure a new actor is created in this case.

Depending on how that interaction between the actor and endpoint map happens the conditions can be different:

An actor can take its own decision of when to shutdown.

- Then it should probably look at how long ago it had a connection.  The PathState covers this info.
- The downside is that it doesn't know much about the whole NodeMap.  And it loses the PathState.
- Discovery is supposed to make up for losing the PathState though.  But it could have recent info from actual paths that established and this is valuable.
 

If the decision can be taken outside of an actor the following factors could be taken into account:
- endpoint high-water mark: above this stop all actors without connections
- endpoint low-water mark: below this don't stop any actor
- some kind of number of inactive actors to keep around
- some kind of minimal time to keep inactive actors around, because the PathState info of actual paths that worked is valuable for a little while.

Frando:
## Description

Fixes #3638 (partially)

This is a first, small solution to stop inactive endpoint actors after an idle timeout. I implemented it such that the *actor* decides once to stop, while making sure that we *never* create senders to actors that are shutting down.

Logic in the actor:
* The actor enters an idle timeout (set to 60 seconds) once it has no active connections, an empty inbox, and no inbox senders
* Once the timeout expires, it is rechecked that the idle conditions hold, and if so the actor exits
* Once any of the idle conditions don't hold anymore, the idle timeout is deactivated and restarted once the conditions are met again

The actor checks if the inbox's sender strong count equals 1, which means that no senders exist apart from the one held in the endpoint map. This check is protected with a mutex, to enter a critical section for closing the inbox while the lock is held in case the conditions are met. This is to ensure that there cannot be a race condition where a sender is cloned out right after the check in the actor returns true, but before the inbox is closed.

Logic in the endpoint map:
* When handing out senders, we acquire the shared lock, and check that the channel is not closed while the lock is held. This ensures that the actor never closes while a sender is alive. If the actor is closed, we remove the handle to the dead actor and create a new actor.
* On regular intervals (set to 60 seconds) the magicsock actor removes handles to dead actors.

## Notes & open questions

* I *think* my logic around the critical section and ensuring that we never close the actor while senders exist is sound. However, it needs careful review and tests. I'll do some thinking on how to best test this.
* Instead of employing an interval to remove dead actor handles, we could use a channel where the actor informs an outside-task which endpoint actors terminated, so that the outside-task can then lock the endpoint map and remove just those. Not sure if that's worth it.
    * Another solution here might be to spawn the actor tasks into a join set in the magicsock actor. However this would need further refactoring and would likely make spawning actors async. I think I'd prefer to keep that sync because it makes the surrounding code a lot simpler.
* This does not yet implement some of the more advanced reasoning that #3638 proposes. I think we should start with something simple that prevents memory exhaustion and tweak as needed. However, it could also be argued that we should start with a more featureful design right away.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/magicsock/endpoint_map.rs
+++ b/iroh/src/magicsock/endpoint_map.rs
@@ -12,22 +12,21 @@ use serde::{Deserialize, Serialize};
 use tokio::sync::mpsc;
 use tracing::warn;
 
+// #[cfg(any(test, feature = "test-utils"))]
+// use crate::endpoint::PathSelection;
+pub(super) use self::endpoint_state::EndpointStateMessage;
+pub(crate) use self::endpoint_state::PathsWatcher;
+use self::endpoint_state::{EndpointStateActor, EndpointStateHandle};
+pub use self::endpoint_state::{PathInfo, PathInfoList};
 use super::{
     DirectAddr, DiscoState, MagicsockMetrics,
     mapped_addrs::{AddrMap, EndpointIdMappedAddr, RelayMappedAddr},
     transports::{self, TransportsSender},
 };
-use crate::disco::{self};
-// #[cfg(any(test, feature = "test-utils"))]
-// use crate::endpoint::PathSelection;
+use crate::disco;
 
 mod endpoint_state;
 mod path_state;
-
-pub(super) use endpoint_state::EndpointStateMessage;
-pub(crate) use endpoint_state::PathsWatchable;
-use endpoint_state::{EndpointStateActor, EndpointStateHandle};
-pub use endpoint_state::{PathInfo, PathInfoList};
 
 /// Interval in which handles to closed [`EndpointStateActor`]s should be removed.
 pub(super) const ENDPOINT_MAP_GC_INTERVAL: Duration = Duration::from_secs(60);
--- a/iroh/src/magicsock/endpoint_map/endpoint_state.rs
+++ b/iroh/src/magicsock/endpoint_map/endpoint_state.rs
@@ -365,7 +365,7 @@ impl EndpointStateActor {
     async fn handle_msg_add_connection(
         &mut self,
         handle: WeakConnectionHandle,
-        tx: oneshot::Sender<PathsWatchable>,
+        tx: oneshot::Sender<PathsWatcher>,
     ) {
         let pub_open_paths = Watchable::default();
         if let Some(conn) = handle.upgrade() {
@@ -427,10 +427,11 @@ impl EndpointStateActor {
             }
             self.trigger_holepunching().await;
         }
-        tx.send(PathsWatchable {
-            open_paths: pub_open_paths,
-            selected_path: self.selected_path.clone(),
-        })
+        tx.send(PathsWatcher::new(
+            pub_open_paths.watch(),
+            self.selected_path.watch(),
+            handle,
+        ))
         .ok();
     }
 
@@ -1032,7 +1033,7 @@ pub(crate) enum EndpointStateMessage {
     /// needed, any new paths discovered via holepunching will be added.  And closed paths
     /// will be removed etc.
     #[debug("AddConnection(..)")]
-    AddConnection(WeakConnectionHandle, oneshot::Sender<PathsWatchable>),
+    AddConnection(WeakConnectionHandle, oneshot::Sender<PathsWatcher>),
     /// Adds a [`EndpointAddr`] with locations where the endpoint might be reachable.
     AddEndpointAddr(EndpointAddr, Source),
     /// Process a received DISCO CallMeMaybe message.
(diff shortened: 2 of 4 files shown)


iroh/src/magicsock/endpoint_map/endpoint_state.rs
```diff
@@ -143,6 +151,12 @@ pub(super) struct EndpointStateActor {
     ///
     /// They failed to open because we did not have enough CIDs issued by the remote.
     pending_open_paths: VecDeque<transports::Addr>,
+    /// A mutex to define a critical section around stopping the actor.
+    ///
+    /// The lock is held while checking the inbox's sender count. Likewise, the handle acquires the lock
+    /// before checking if the inbox is live and cloning out a new sender. This prevents a race condition
+    /// where the handle would clone out a new sender right after the actor decided to terminate.
+    close_lock: Arc<Mutex<()>>,
```
flub: Should be in the "hook into the rest of the magicsock" section I think.

It's kind of unpleasant that this is "by convention", I mean, it holds () inside.

I wonder if it is better to put the sender in this lock.  And pass a copy of the `Arc<Mutex<mpsc::Sender<EndpointStateMessage>>>` to the handle.  It means the actor would keep its own inbox sender around for nothing, but it does make this a bit harder to misuse.

Frando: I added a `guarded_channel` abstraction that properly hides the implementation. Much better IMO.

iroh/src/magicsock/endpoint_map/endpoint_state.rs
```diff
@@ -256,6 +267,23 @@ impl EndpointStateActor {
                     self.scheduled_holepunch = None;
                     self.trigger_holepunching().await;
                 }
+                _ = &mut idle_timeout => {
+                    // Check is_idle and if true close the inbox while the close_lock is held.
```
flub: obsolete comment

iroh/src/magicsock/endpoint_map/endpoint_state.rs
```diff
@@ -1021,14 +1049,34 @@ pub(crate) enum EndpointStateMessage {
...
+    /// is used to ensure that once we cloned the sender, the actor won't close until the sender is dropped.
+    pub(super) fn sender(&self) -> Option<mpsc::Sender<EndpointStateMessage>> {
+        self.sender.get()
+    }
+
+    /// Returns `true` if the endpoint actor channel is closed (i.e. if the actor has terminated).
+    pub(super) fn is_closed(&self) -> bool {
+        self.sender.is_closed()
+    }
+}
```
flub: are these functions still helpful  now?

iroh/src/magicsock/endpoint_map/endpoint_state.rs
```diff
@@ -1315,3 +1363,81 @@ impl Future for OnClosed {
         Poll::Ready(self.conn_id)
     }
 }
+
+mod util {
```
flub: this file is big enough, let's move this out to it's own file.

iroh/src/magicsock/endpoint_map.rs
```diff
@@ -131,30 +144,39 @@ impl EndpointMap {
...
-                    metrics,
-                );
-                let handle = actor.start();
-                let sender = handle.sender.clone();
-                handles.insert(eid, handle);
-
-                // Ensure there is a EndpointMappedAddr for this EndpointId.
-                self.endpoint_mapped_addrs.get(&eid);
-                sender
+        loop {
```
flub: I'm always nervous about these kinds of things...

Can you not avoid the repetition by moving creating the new actor to a function?  And then somehow losing the loop?  (I've not fully thought this through)

iroh/src/magicsock/endpoint_map/endpoint_state.rs
Frando: fixed

iroh/src/magicsock/endpoint_map/endpoint_state.rs
Frando: Yeah not really, removed them and made sender `pub(super)`

iroh/src/magicsock/endpoint_map/endpoint_state.rs
Frando: done

iroh/src/magicsock/endpoint_map.rs
Frando: Good idea. Did that. It is nicer without the loop indeed.

iroh/src/magicsock/endpoint_map/endpoint_state/guarded_channel.rs
flub, line 1: s/util.rs/guarded_channel.rs/  😄

iroh/src/magicsock/endpoint_map/endpoint_state/util.rs
```diff
@@ -0,0 +1,75 @@
+use std::sync::{Arc, Mutex};
+
+use tokio::sync::mpsc;
+
+/// Creates a new [`mpsc`] channel where the receiver can only close if there are no active senders.
+pub(crate) fn guarded_channel<T>(cap: usize) -> (GuardedSender<T>, GuardedReceiver<T>) {
```
flub: is `pub(super)` enough for everything in this module?  I'm always a fan of keeping visibility as tight as possible.

iroh/src/magicsock/endpoint_map/endpoint_state/guarded_channel.rs
Frando: done

iroh/src/magicsock/endpoint_map/endpoint_state/util.rs
Frando: the sender needs to be more pub than super because it is used in the endpoint_map. left that pub(crate) and changed the rest to be pub(super).

flub: Ah, i guess then it would be fine to leave everything pub(crate). But anyway, not terribly important.

Issue #3641, opened 2025-11-10
See if removing the TransportsSenderActor is feasible


flub:
transports.rs:
> TODO: Can I just send the TransportsSender along in the NodeStateMessage::SendDatagram message??  That way you don't have to hook up the sender into the NodeMap!

This is an interesting refactor that would change the `MagicSender::poll_send` to clone the `TransportsSender` and send it to the EndpointStateActor for MultipathMappedAddr::Mixed.  That way we could get rid of this actor in TransportsSenderActor.

Not 100% sure it's possible, but likely.

PR #3595, opened 2025-10-31
Switch to QUIC-NAT-Traversal instead of DISCO

flub:
## Description

This switches from the old DISCO to the so-new-it-doesnt-exit-yet QUIC
NAT Traversal.

## Breaking Changes

Nothing visible?  Maybe?

## Notes & open questions

The QUIC NAT Traversal API doesn't exist yet, so this won't even build
on any machine that's not mine.  I've locally patched in the dummy
methods that I use.

## Change checklist

- [ ] Self-review.
- [ ] Documentation updates following the [style guide](https://rust-lang.github.io/rfcs/1574-more-api-documentation-conventions.html#appendix-a-full-conventions-text), if relevant.
- [ ] Tests if relevant.
- [ ] All breaking changes documented.
  - [ ] List all breaking changes in the above "Breaking Changes" section.
  - [ ] Open an issue or PR on any number0 repos that are affected by this breaking change. Give guidance on how the updates should be handled or do the actual updates themselves. The major ones are:
    - [ ] [`quic-rpc`](https://github.com/n0-computer/quic-rpc)
    - [ ] [`iroh-gossip`](https://github.com/n0-computer/iroh-gossip)
    - [ ] [`iroh-blobs`](https://github.com/n0-computer/iroh-blobs)
    - [ ] [`dumbpipe`](https://github.com/n0-computer/dumbpipe)
    - [ ] [`sendme`](https://github.com/n0-computer/sendme)

PR #3644, opened 2025-11-11
refactor: remove the TransportsSenderActor

dignifiedquire:
First step for #3641, the rest can be done once quic holepunching has landed

Unfortunately `send_disco_message` also needs the sender, so it can't be fully removed from the `EndpointState`


flub:
> Unfortunately `send_disco_message` also needs the sender, so it can't be fully removed from the `EndpointState`

Right, but that's gone in #3595.  I wonder if we should keep the issue open until that is merged so we can re-visit it?  It would mean the TransportsSender disappears from a bunch of constructors to plug it through I think.

dignifiedquire:
sure

PR #3645, opened 2025-11-11
refactor(multipath): move discovery into `EndpointStateActor`

### LINKED ISSUE #3642: Improve discovery cooperation with EndpointStateActor
`Endpoint::get_mapping_addr_and_maybe_start_discovery` is a bit too simple right now.  Or perhaps this is sufficient to get us to 1.0, let's decide which it is.

It has no knowledge of what state the EndpointStateActor has on the remote node.  It uses `EndpointStateActor::CanSend` which is very primitive.  So it does not really know if discovery should be started.

Maybe `CanSend` is sufficient for now, but needs to use more information like if there's an active connection or not, how old the addresses are that it has.  Figure out how much smarter this needs to be.

The long-shot version of this is that the EndpointStateActor should be able to trigger discovery itself when it loses all paths to a remote.  But that's probably more than we need right now.

Frando:
## Description

Fixes #3642 

This moves discovery handling fully into the `EndpointStateActor`.
The pub(crate)  interface to trigger discovery and get a EndpointMappedAddr is now `Magicsock::resolve_remote`, which sends the provided addresses to the EndpointStateActor. The actor starts discovery if it does not have a selected path and if discovery is not running. It returns either immediately if there are any known paths, or waits for discovery to produce at least one result or an error. Once this returns, `resolve_remote` returns either with a EndpointMappedAddr or with the discovery error.

This means the current behavior is kept: We only start `quinn::Endpoint::connect` once we have at least one transport address for the remote. If not, we return the discovery error immediately from `iroh::Endpoint::connect`.

This opens the door for us to easily tune when to run discovery in other siutations, e.g. when all available paths to a remote are closed. However, for now this PR still only starts discovery when `Endpoint::connect` is called and no path is selected at the moment.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


iroh/src/magicsock.rs
```diff
@@ -306,14 +294,30 @@ impl MagicSock {
...
 
-    /// Returns `true` if we have at least one candidate address where we can send packets to.
-    pub(crate) async fn has_send_address(&self, eid: EndpointId) -> bool {
+    /// Returns `Ok` if we have at least one candidate address where we can send packets to.
+    pub(crate) async fn want_connect(
+        &self,
+        mut addr: EndpointAddr,
+    ) -> Result<EndpointIdMappedAddr, DiscoveryError> {
+        // Prune our own addreses from the endpoint address.
+        // TODO: Move this somewhere else?
```
flub: Yes please!

The EndpointStateActor should collect these addresses as usual but when choosing addresses to send to when handling `SendDatagram` it should filter out any that are itself.  That would mean the possible remote addresses are not lost when you move network.  The EndpointStateActor already has access to the local DirectAddrs so this is way more logical.

flub (commented):
I guess most tests needs to be moved the the StaticProvider?

I don't quite follow the WantConnect stuff.  Is this trying to make sure you can return an error straight from calling `Endpoint::connect` when you don't have any discovery results?  As otherwise you'd start connecting and then the connection would time out?

Frando:
> I don't quite follow the WantConnect stuff. Is this trying to make sure you can return an error straight from calling Endpoint::connect when you don't have any discovery results? 

Yes, exactly. It mimics the current behavior that `quinn::Endpoint::connect` is only started once we have at least one dialable remote address, and if discovery errors or returns no results, we return straight from `iroh::Endpoint::connect` without touching quinn at all. The alternative would be that all connections time out if discovery yields nothing.

Edit: I renamed it to `ResolveRemote` which is much clearer IMO.

Cargo.toml
```diff
@@ -52,10 +52,3 @@ netwatch = { git = "https://github.com/n0-computer/net-tools", branch = "main" }
 # iroh-quinn = { path = "../iroh-quinn/quinn" }
 # iroh-quinn-proto = { path = "../iroh-quinn/quinn-proto" }
 # iroh-quinn-udp = { path = "../iroh-quinn/quinn-udp" }
-
-
-[patch."https://github.com/n0-computer/quinn"]
-
-# iroh-quinn = { path = "../iroh-quinn/quinn" }
-# iroh-quinn-proto = { path = "../iroh-quinn/quinn-proto" }
-# iroh-quinn-udp = { path = "../iroh-quinn/quinn-udp" }
```
matheus23, line 61: Most dependencies in this repo in the `feat-multipath` branch still use the git dependency.

I'd say we keep this. Or we also remove the commented out patch lines above, as we need either both or neither: The patch lines above are so we can patch `quinn-udp` in `netwatch`, and these ones are for all the quinn dependencies in this monorepo.

Cargo.lock
```diff
@@ -2827,7 +2850,7 @@ dependencies = [
  "n0-error",
  "n0-future",
  "n0-watcher",
- "netdev",
+ "netdev 0.39.0",
```
matheus23: Somehow we have a new duplicated dependency now?

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -89,6 +93,18 @@ type PathEvents = MergeUnbounded<
...
+/// Either a stream of incoming results from [`ConcurrentDiscovery::resolve`] or infinitely pending.
+///
+/// Set to [`Either::Left`] with an always-pending stream while discovery is not running, and to
+/// [`Either::Right`] while discovery is running.
+///
+/// The stream returned from [`ConcurrentDiscovery::resolve`] is `!Sync`. We use the (safe) [`SyncStream`]
+/// wrapper to make it `Sync` so that the [`EndpointStateActor::run`] future stays `Send`.
+type DiscoveryStream = Either<
+    n0_future::stream::Pending<Result<DiscoveryItem, DiscoveryError>>,
+    SyncStream<BoxStream<Result<DiscoveryItem, DiscoveryError>>>,
```
matheus23, line 104: :| This is very annoying. Glad you figured that out.

iroh/src/magicsock/endpoint_map/endpoint_state.rs
```diff
@@ -218,7 +245,7 @@ impl EndpointStateActor {
     /// not processing items from the inbox while waiting on any async calls.  So some
     /// discipline is needed to not turn pending for a long time.
     async fn run(
-        &mut self,
+        mut self,
```
matheus23: Is there a particular reason for this change?

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -354,7 +381,15 @@ impl EndpointStateActor {
...
-                self.send_datagram(addr.clone(), transmit.clone()).await?;
+                // We never want to send to our local addresses.
+                // The local address set is updated in the main loop so we can use `peek` here.
+                if let transports::Addr::Ip(sockaddr) = addr
+                    && self.local_addrs.peek().iter().any(|a| a.addr == *sockaddr)
+                {
+                    warn!(%sockaddr, "not sending datagram to our own address");
+                } else {
+                    self.send_datagram(addr.clone(), transmit.clone()).await?;
+                }
```
matheus23, line 388: Wait. How would we even end up in this situation?

This would mean that `self.paths` - so the paths we know to the remote endpoint - would contain ... one/some of our own addresses (including ports)?

That seems like a bug? (I guess we have to guard against this, given that Discovery can essentially do whatever in user code...)
When have you seen this? Perhaps we should prevent these to become part of `paths` in the first place? Wouldn't we otherwise end up making quinn try to open paths to ourselves?

iroh/src/magicsock/endpoint_map/endpoint_state.rs
```diff
@@ -439,23 +470,19 @@ impl EndpointStateActor {
...
+        for addr in addrs {
+            let addr = match addr {
+                TransportAddr::Relay(relay_url) => Addr::from((relay_url, self.endpoint_id)),
+                TransportAddr::Ip(sockaddr) => Addr::from(sockaddr),
+                _ => {
+                    warn!(?addr, "Unsupported TransportAddr");
+                    continue;
+                }
+            };
+            self.add_path_entry(addr, source.clone());
```
matheus23: Should we make sure at this point to never add one of our own addresses to the paths?

matheus23 (commented):
I'm a little confused about the `local_addrs` filtering. Is there some situation in particular that you saw that meant you had to add filtering?
Ideally none of our discovery services will accidentally discover local addresses.
And perhaps ideally the endpoint doesn't break *too much* when these addresses *do* end up in our `EndpointStateActor::paths`, right?

Frando:
> I'm a little confused about the local_addrs filtering. Is there some situation in particular that you saw that meant you had to add filtering?
Ideally none of our discovery services will accidentally discover local addresses.
And perhaps ideally the endpoint doesn't break too much when these addresses do end up in our EndpointStateActor::paths, right?

I did not add it, only moved it around. It lives [here](https://github.com/n0-computer/iroh/blob/feat-multipath/iroh/src/magicsock.rs#L425) on `feat-multipath`. Because that whole function was removed, I moved the check over, as suggested by @flub.

Cargo.toml
Frando: Oh right that was an unintended change after merging `feat-multipath`. Will revert.

iroh/src/magicsock/endpoint_map/endpoint_state.rs
Frando: No. I think I myself changed this from `mut self` to `&mut self` in another PR, which first needed it but then not anymore after a refactor, and it slipped through. As long as we don't access `self` after `run`, `mut self` is cleaner IMO.

iroh/src/magicsock/remote_map/remote_state.rs
Frando: As outlined below, this check previously was in `add_endpoint_addr`, which was called both for addresses added via `Endpoint::connect` by the user, and for addresses retrieved from discovery.

Both may contain any valid address, so this guards against connecting to ourselves.

matheus23:
Given that check used to happen in `add_endpoint_addr`, should the check perhaps be moved to the equivalent of that at `EndpointStateAddr::add_addrs` instead of happening every time we `send_datagram`?

iroh/src/magicsock/endpoint_map/endpoint_state.rs
Frando: Maybe. @flub suggested to move the check against local_addrs to `handle_msg_send_datagram`, I did not think it through fully if another place would be better. Will give it another round of thought.

matheus23: Ah okay!
Well, I guess if we do the check in `add_addrs`, it's possible that the local addresses change *after* we've already added them.
I do think it'd be better to watch the changes though, and filter out our own addresses from `paths` every time they change, instead of filtering the paths every time we want to send datagrams!

FWIW, I do think `send_datagram` can become a *somewhat* hot function when e.g. someone uses 0-RTT data.

Cargo.lock
Frando: Inadvertent change due to cargo update somehow, reverted.

Frando: See https://github.com/n0-computer/iroh/pull/3667

Frando:
> Given that check used to happen in add_endpoint_addr, should the check perhaps be moved to the equivalent of that at EndpointStateAddr::add_addrs instead of happening every time we send_datagram?

This sounds good to me, but I'd like to hear @flub's opinion too.

Edit: I remember now why I  moved it to `handle_msg_send_datagram`. The local addrs might have changed between the call to `add_addr` and when these addrs are used in `handle_msg_send_datagram` so I thought that it would be best to check against the *current* set of local addrs when trying to use an address, instead of filtering when adding them (against an potentially always-changing set of local addrs).

iroh/Cargo.toml
```diff
@@ -76,6 +76,7 @@ futures-util = "0.3"
 
 # test_utils
 axum = { version = "0.8", optional = true }
+sync_wrapper = { version = "1.0.2", features = ["futures"] }
```
dignifiedquire, line 79: 😢 another dependency just for making things sync?

iroh/src/endpoint/connection.rs
```diff
@@ -370,23 +360,24 @@ pub enum ConnectingError {
     },
     #[error("Failure finalizing the handshake")]
     HandshakeFailure { source: AuthenticationError },
-    #[error("internal consistency error: EndpointStateActor stopped")]
-    InternalConsistencyError,
+    #[error("internal consistency error")]
+    InternalConsistencyError {
+        /// Private source type, cannot be accessed publicly.
```
dignifiedquire: maybe `cannot be crated publicly`?

iroh/src/magicsock.rs
```diff
@@ -305,14 +300,37 @@ impl MagicSock {
...
+    ///
+    /// Returns `Ok(Ok(EndpointIdMappedAddr))` if there is a known path or discovery produced
+    /// at least one result. This does not mean there is a working path, only that we have at least
+    /// one transport address we can try to connect to.
+    ///
+    /// Returns `Ok(Err(discovery_error))` if there are no known paths to the remote and discovery
+    /// failed or produced no results. This means that we don't have any transport address for
+    /// the remote, thus there is no point in trying to connect over the quinn endpoint.
+    ///
+    /// Returns `Err(EndpointStateActorStoppedError)` if the `EndpointStateActor` for the remote has stopped,
```
dignifiedquire: so we never stop this actor? this would mean we accumulate an unbounded number of actors over time?

iroh/src/magicsock/remote_map/remote_state.rs
dignifiedquire: hm, this is very late now, seems unfortunate that this error is not surfaced really to the user now

iroh/src/magicsock/endpoint_map/endpoint_state.rs
dignifiedquire: yeah `send_datagram` is pretty hot, we should do filtering before it lands there

matheus23:
> Edit: I remember now why I moved it to `handle_msg_send_datagram`. The local addrs might have changed between the call to `add_addr` and when these addrs are used in `handle_msg_send_datagram` so I thought that it would be best to check against the _current_ set of local addrs when trying to use an address, instead of filtering when adding them (against an potentially always-changing set of local addrs).

My suggestion would be to update `EndpointStateActor::paths` in right after the `self.local_addrs.updated()` hits in the `run` loop.

It's a little bit more bookkeeping, but avoids filtering in the `send_datagram` fn :)

iroh/src/magicsock/endpoint_map/endpoint_state.rs
Frando: Is it? This is the `handle_msg_send_datagram`  in the `EndpointStateActor`, which is only invoked for `MultipathMappedAddr::Mixed`, i.e. for the initial packet if there's no selected path yet.

iroh/Cargo.toml
Frando: another? which one is the other one?

Frando: We could add it to `n0-futures` though. The whole crate is <300 lines.
https://docs.rs/sync_wrapper/latest/src/sync_wrapper/lib.rs.html

dignifiedquire: another, to the many dependencies we already have 😓

iroh/src/magicsock.rs
Frando: We do stop them, but a) we never stop them if there are any senders to the actor, and b) we restart the actor on demand. So if the invariants hold this error never occurs. I.e. it is a bug in the implementation if this error ever happens, so I *think* it would be fine to `.expect()` instead, but @flub preferred to not do that and instead surface the error.

dignifiedquire: okay, got it 👍

iroh/src/magicsock/endpoint_map/endpoint_state.rs
```diff
@@ -151,9 +167,16 @@ pub(super) struct EndpointStateActor {
     ///
     /// They failed to open because we did not have enough CIDs issued by the remote.
     pending_open_paths: VecDeque<transports::Addr>,
+    /// Discovery service, cloned from the magicsock.
+    discovery: ConcurrentDiscovery,
+    /// Stream of discovery results, or always pending if discovery is not running.
+    discovery_stream: DiscoveryStream,
+    /// Pending requests from [`EndpointStateMessage::ResolveRemote`].
+    pending_resolve_requests: VecDeque<oneshot::Sender<Result<(), DiscoveryError>>>,
```
flub: I don't think these belong in the `Internal state - holepunching and path state` section of state.  Probably `discover` should be in the `Hooks into the rest of the MagicSocket.` section.  The `discovery_stream` and `pending_resolve_requests` could be a new `Internal state` section.

I'd be nice to keep things organised here.

iroh/src/magicsock/endpoint_map/endpoint_state.rs
```diff
@@ -354,7 +381,15 @@ impl EndpointStateActor {
                 "sending datagram to all known paths",
             );
             for addr in self.paths.keys() {
-                self.send_datagram(addr.clone(), transmit.clone()).await?;
+                // We never want to send to our local addresses.
+                // The local address set is updated in the main loop so we can use `peek` here.
+                if let transports::Addr::Ip(sockaddr) = addr
+                    && self.local_addrs.peek().iter().any(|a| a.addr == *sockaddr)
+                {
+                    warn!(%sockaddr, "not sending datagram to our own address");
```
flub: That's not warn, at most debug, but probably I'd try trace first.

iroh/src/magicsock/remote_map/remote_state.rs
flub: I've wanted this change for ages.  This is the right place to do it.  And there's nothing to surface to the user.

The user or a discovery service can totally reasonably add e.g. 192.168.2.5:123 to the addresses on which a remote is reachable.  It is totally reasonable that the local endpoint happens to be in another 192.168.2.0/24 network and has the same address.  In that case we just don't want to dial it.  But say I then walk over across the street to my neighbour and connect to their wifi.  We're now on the same 192.168.2.0/24 network and that address I already know could mean I manage to establish a new connection (I'll have a different address locally now, e.g. 192.168.2.6:123).

This is just things working smoothly as they should.

iroh/src/magicsock/endpoint_map/endpoint_state.rs
```diff
@@ -408,11 +443,7 @@ impl EndpointStateActor {
                 };
                 path.set_status(status).ok();
                 conn_state.add_open_path(path_remote.clone(), PathId::ZERO);
-                self.paths
-                    .entry(path_remote)
-                    .or_default()
-                    .sources
-                    .insert(Source::Connection { _0: Private }, Instant::now());
+                self.add_path_entry(path_remote, Source::Connection { _0: Private });
```
flub: I can't say I'm a fan of having a helper function for this.  It hides what state we're modifying.  A `self.paths.add_entry()` would be much more amenable.  I kind of think we should modify local state directly, otherwise we're building something too complex and need to tweak the abstractions.

As a similar example I had several iterations of pretty horrible state management issues before I ended up finding `ConnectionState` in its current form, which ended up being much more maintainable (and may need to evolve further at some point, this is by no ways some perfect abstraction).

iroh/src/magicsock/endpoint_map/endpoint_state.rs
```diff
@@ -439,23 +470,19 @@ impl EndpointStateActor {
...
-        for relay_url in addr.relay_urls() {
-            let addr = transports::Addr::from((relay_url.clone(), self.endpoint_id));
-            self.paths
-                .entry(addr)
-                .or_default()
-                .sources
-                .insert(source.clone(), Instant::now());
+    /// Adds new [`TransportAddr`] addresses to our list of potential paths.
+    fn add_addrs(&mut self, addrs: BTreeSet<TransportAddr>, source: Source) {
+        use transports::Addr;
```
flub: For consistency I'd prefer to not do this.  `transports::Addr` is used everywhere, it is clear.  Let's stick with that.

iroh/src/magicsock/endpoint_map/endpoint_state.rs
flub: `send_datagram` is not hot, it is only for new connections.  Here we are adding static information that may survive a longer than our own location on the network, so this is the wrong place to filter.

If `send_datagram` really does turn out to be hot we can do something like keep pre-filtered listed somehow and update that each time there's a new address or local addresses change.  But I honestly think that's over-architecting for now.  Let's keep thing simpler a little longer, there's already plenty of more important complexity here we need to tame.

iroh/src/magicsock/endpoint_map/endpoint_state.rs
```diff
@@ -439,23 +470,19 @@ impl EndpointStateActor {
...
-                .insert(source.clone(), Instant::now());
-        }
-        for relay_url in addr.relay_urls() {
-            let addr = transports::Addr::from((relay_url.clone(), self.endpoint_id));
-            self.paths
-                .entry(addr)
-                .or_default()
-                .sources
-                .insert(source.clone(), Instant::now());
+    /// Adds new [`TransportAddr`] addresses to our list of potential paths.
```
flub: Please also add what this really does.  Because appending to a list does not require a function.  What magic happens?

iroh/src/magicsock/endpoint_map/endpoint_state.rs
```diff
@@ -575,6 +610,64 @@ impl EndpointStateActor {
...
+            }
+            Some(Ok(item)) => {
+                if item.endpoint_id() != self.endpoint_id {
+                    warn!(?item, "Discovery emitted item for wrong remote endpoint");
+                } else {
+                    let source = Source::Discovery {
+                        name: item.provenance().to_string(),
+                    };
+                    let addr = item.into_endpoint_addr();
+                    self.add_addrs(addr.addrs, source);
```
flub: not calling `self.emit_pending_resolve_requests` here is a red flag for state management to me.  Because the other branches have this.  It makes things complex.

iroh/src/magicsock/endpoint_map/endpoint_state.rs
flub: Ok, I've made it all the way to the end of the endpoint_state.rs changes and finally have managed to come up with a suggestion.  You might encounter a few more comments on this topic.  Not 100% sure it's the right one as I haven't tried but anyway.  

I think the complexity I have issue with comes from that you need to respond to the `ResolveRemote` requests at two points:
- whenever `self.paths` changes
- whenever discovery found something
And then you end up with a brittle combination of mutating state and calling hooks at the right points all over the place.

Can this not be moved into `self.paths`?  Maybe path_state.rs can redeem itself for still being it's own module 😉 (currently a historical accident).  If we had an `EndpointPathState` (bikeshedding welcome) that you passed the oneshot senders into.  And that way you can have a clear abstraction boundary for paths being added and it can emit change responses whenever we get a new remote transport address for whatever reason.

Would that go a long way to reducing the brittleness and complexity that's being added here?

iroh/src/magicsock.rs
```diff
@@ -104,10 +104,18 @@ pub(crate) const PATH_MAX_IDLE_TIMEOUT: Duration = Duration::from_millis(6500);
 pub(crate) const MAX_MULTIPATH_PATHS: u32 = 16;
 
 /// Error returned when the endpoint state actor stopped while waiting for a reply.
-#[stack_error(derive)]
+#[stack_error(add_meta, derive)]
```
flub, line 103: ooh!  TIL why things where the shape they were before.

iroh/src/magicsock.rs
```diff
@@ -305,14 +300,37 @@ impl MagicSock {
             .filter_map(|addr| addr.into_socket_addr())
     }
 
-    /// Returns `true` if we have at least one candidate address where we can send packets to.
-    pub(crate) async fn has_send_address(&self, eid: EndpointId) -> bool {
-        let actor = self.endpoint_map.endpoint_state_actor(eid);
+    /// Resolves a [`EndpointAddr`] to an [`EndpointIdMappedAddr`] to connect to via [`Handle::endpoint`].
```
flub:
```suggestion
    /// Resolves an [`EndpointAddr`] to an [`EndpointIdMappedAddr`] to connect to via [`Handle::endpoint`].
```

iroh/src/magicsock/endpoint_map/endpoint_state.rs
dignifiedquire: > send_datagram is not hot, it is only for new connections.

that's pretty surprising 😅  especially given the name

Frando: The whole system was designed such that most datagrams don't ever pass through the EndpointStateActor, only the initial packets (that have a Mixed mapped addr) need to pass through there to be forwarded to all transport paths for this endpoint. Afterwards, the dest addrs are direct IP, or relay mapped addrs, and both don't need to go through the EndpointStateActor but go right to the UdpSender of the transport.

matheus23: > Is it? This is the `handle_msg_send_datagram` in the `EndpointStateActor`, which is only invoked for `MultipathMappedAddr::Mixed`, i.e. for the initial packet if there's no selected path yet.

This is why I was saying it can be somewhat hot for 0-RTT functions. IIUC, all packets that are sent before we get the first response will use `send_datagram`. So if you're sending a bunch of stuff in 0-RTT that's going to go through that mechanism.

---

With @flub's latest comments I can see how managing the state of filtered vs. unfiltered outside would be messy. And I agree, perhaps this is premature optimization. If we find the need, we can indeed keep a "cache" of the filtered list around.

Perhaps there's a way of communicating the outcome of this discussion in a good way in a comment, but even so, I'd consider my issue resolved :D

iroh/src/magicsock/endpoint_map/endpoint_state.rs
Frando: This would indeed be a better structure likely, yes.

I implemented this, see the latest commit on this branch.

iroh/src/magicsock/endpoint_map/endpoint_state.rs
Frando: This now goes over the `EndpointPathState`, lmk if you still think the interactions are not clear enough.

iroh/src/magicsock/remote_map/remote_state/path_state.rs
```diff
@@ -1,19 +1,147 @@
...
+            false
+        } else {
+            true
+        }
+    }
+
+    /// Notifies that a discovery run has finished.
+    ///
+    /// This will emit pending resolve requests.
+    pub(super) fn discovery_finished(&mut self, error: Option<DiscoveryError>) {
```
flub: minor nit, but doesn't `res: Result<(), DiscoveryError>` make more sense?

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -173,17 +197,19 @@ impl RemoteStateActor {
...
             path_events: Default::default(),
-            paths: FxHashMap::default(),
+            paths: Default::default(),
             last_holepunch: None,
             selected_path: Default::default(),
             scheduled_holepunch: None,
             scheduled_open_path: None,
             pending_open_paths: VecDeque::new(),
             sender,
+            discovery,
```
flub: So definitely nitpicking.  I alsoways liked: https://rust-lang.github.io/rust-clippy/stable/index.html#inconsistent_struct_constructor

But as the lint is not enabled I'm sure you get to choose what you do :)

flub (approved):

PR #3651, opened 2025-11-12
feat: relay only configuration

### LINKED ISSUE #2957: Add an option to always use relays
Together with https://github.com/n0-computer/iroh/issues/2901 it should be possible to establish all connections via proxy and not reveal the IP address even if communicating to unknown relays.

Some users may want to use P2P features of the app (calls, [webxdc realtime](https://delta.chat/en/2024-11-20-webxdc-realtime)) but never reveal the IP address.

dignifiedquire:
This is a next step into the world of configurable transports. We now allow disabling the IP based transports entirely. 
Internally this starts to prepare for a world where the user can configure multiple different transports, IP, relay and others in the future.

Closes #2957

### DIFF
--- a/iroh/src/magicsock/mapped_addrs.rs
+++ b/iroh/src/magicsock/mapped_addrs.rs
@@ -33,7 +33,7 @@ const RELAY_MAPPED_SUBNET: [u8; 2] = [0, 1];
 /// The Subnet ID for [`EndpointIdMappedAddr`].
 const ENDPOINT_ID_SUBNET: [u8; 2] = [0; 2];
 
-/// A default fake addr, using the maxium addr that the internal fake addrs could be using.
+/// A default fake addr, using the maximum addr that the internal fake addrs could be using.
 pub const DEFAULT_FAKE_ADDR: SocketAddrV6 = SocketAddrV6::new(
     Ipv6Addr::new(
         u16::from_be_bytes([ADDR_PREFIXL, 21]),


iroh/src/endpoint.rs
```diff
@@ -1837,6 +1889,104 @@ mod tests {
...
+                for info in infos {
+                    if info.is_ip() {
+                        panic!("should not happen: {:?}", info);
+                    }
+                    if info.is_relay() {
+                        break 'outer;
+                    }
+                }
+            }
+            info!("Have direct connection");
```
flub: nope

iroh/src/endpoint.rs
```diff
@@ -1837,6 +1889,104 @@ mod tests {
...
+                .secret_key(secret)
+                .alpns(vec![TEST_ALPN.to_vec()])
+                .insecure_skip_relay_cert_verify(true)
+                .relay_mode(RelayMode::Custom(relay_map))
+                .disable_ip()
+                .bind()
+                .await?;
+            ep.online().await;
+            let mut node_addr = ep.addr();
+            node_addr.addrs.retain(|addr| addr.is_relay());
```
flub: I think this shouldn't be needed anymore?  Maybe it should be an assert instead?

iroh/src/endpoint.rs
```diff
@@ -112,13 +112,27 @@ pub struct Builder {
...
+    transports: Vec<TransportConfig>,
     #[cfg(any(test, feature = "test-utils"))]
     path_selection: PathSelection,
     max_tls_tickets: usize,
 }
 
+impl From<RelayMode> for Option<TransportConfig> {
+    fn from(mode: RelayMode) -> Self {
+        match mode {
+            RelayMode::Disabled => None,
```
Frando, line 124: This is a change in behavior, isn't? Currently RelayMode::Disabled means the endpoint has no home relay, but it will still dial the home relays of other peers if you get addrs with relay urls. Now, it would mean the relay transport is entirely disabled, right?

But it makes sense, and the previous variant can still be done by setting RelayMode::Custom with an empty relay map I guess? Should be clearly documented though.

dignifiedquire: yeah, I think this is the more "correct" interpretation of `Disabled`

iroh/src/endpoint/connection.rs
```diff
@@ -732,7 +732,7 @@ impl OutgoingZeroRttConnection {
 
     // TODO: It seems `SendDatagram` is not yet exposed by quinn.  This has been fixed
     //       upstream and will be in the next release.
-    // /// Transmits `data` as an unreliable, unordered application datagram
+    // /// Transmits `data` as an unreliable, unordered application datagramb
```
ramfox:
```suggestion
    // /// Transmits `data` as an unreliable, unordered application datagram
```

iroh/src/endpoint.rs
```diff
@@ -1252,7 +1307,7 @@ impl Endpoint {
 /// Options for the [`Endpoint::connect_with_opts`] function.
 #[derive(Default, Debug, Clone)]
 pub struct ConnectOptions {
-    transport_config: Option<Arc<TransportConfig>>,
+    transport_config: Option<Arc<quinn::TransportConfig>>,
```
ramfox, line 1310: should we be wrapping this foreign type?

ramfox (approved):
looks good to me! two nits

Only question mark was around `disable_ip` and `disable_relay` methods. I think we are good to have them, especially since we have the expectation of only `ip` or `relay` in our other APIs. But curious what this will look like as we add transports or allow others to put in custom transports.

LGTM

ramfox:
oh also, lot's of `TODO: docs`, which should be addressed before merging, unless we are going to fix them before the multipath merge to main

iroh/src/magicsock/mapped_addrs.rs
```diff
@@ -33,6 +33,22 @@ const RELAY_MAPPED_SUBNET: [u8; 2] = [0, 1];
 /// The Subnet ID for [`EndpointIdMappedAddr`].
 const ENDPOINT_ID_SUBNET: [u8; 2] = [0; 2];
 
+pub const DEFAULT_FAKE_ADDR: SocketAddrV6 = SocketAddrV6::new(
+    Ipv6Addr::new(
+        u16::from_be_bytes([ADDR_PREFIXL, 21]),
+        u16::from_be_bytes([7, 10]),
```
Frando, line 40: Is there any meaning to these numbers? In either case, maybe a short comment on how they're chosen would be nice.

dignifiedquire: these are the same numbers we use for the other fake addrs

dignifiedquire: added a comment

iroh/src/endpoint.rs
dignifiedquire: added some docs

PR #3653, opened 2025-11-12
feat: WIP allow non-ed keys in EndpointID

rklaehn:
## Description

This is the latest attempt that leaves more internals alone.

Replaces https://github.com/n0-computer/iroh/pull/3576

The basic approach is to first replace the EndpointId type alias with PublicKey everywhere and then reintroduce EndpointId as a new type only in public APIs. The entire relay crate, despite being public, still works with PublicKey.

Every place where we have a PublicKey and need an EndpointId on the outside we can just use .into(). In tests it is Ok to go from EndpointId to PublicKey using .expect_ed().

Discovery works with EndpointId, but discovery services that fundamentally require ed keys, such as pkarr, will just do nothing for non ed keys.

Todo:

- [ ] tickets need a new version which has a variant identifier for the key
- [ ] tostring/parse roundtrip
- [ ] check public iroh api for remaining PublicKey
- [ ] bikeshed how the tostring for an EndpointId should look like
- [ ] remove or change Endpoint::secret_key

Todo (later?):

- [ ] change encoding for mdns to be able to encode an actual endpoint id, not just a public key

The relay protocol of course remains fully based on ed keys. Changing that to support different keys would be a major effort, but with multipath we might be able to just add a second one using whatever new keys we support.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/discovery/mdns.rs
+++ b/iroh/src/discovery/mdns.rs
@@ -352,8 +352,7 @@ impl MdnsDiscovery {
                         );
 
                         let mut resolved = false;
-                        let item =
-                            peer_to_discovery_item(&peer_info, &discovered_endpoint_id);
+                        let item = peer_to_discovery_item(&peer_info, &discovered_endpoint_id);
                         if let Some(senders) = senders.get(&discovered_endpoint_id) {
                             trace!(?item, senders = senders.len(), "sending DiscoveryItem");
                             resolved = true;


iroh-base/src/key.rs
```diff
@@ -23,6 +23,62 @@ use serde::{Deserialize, Serialize, de, ser};
 #[repr(transparent)]
 pub struct PublicKey(CompressedEdwardsY);
 
+/// The identifier for an endpoint in the (iroh) network.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
+pub enum EndpointId {
+    /// An Ed25519 public key.
+    Ed25519(PublicKey),
+    /// Other types of endpoint identifiers.
+    Other([u8; 8]),
```
flub: can this be made non-exhaustive instead of having a dummy type?  Or are you just trying out having two variants?

rklaehn: Yeah, this is just playing around until we decide to actually do this.

rklaehn:
We are not doing this before 1.0

PR #3663, opened 2025-11-13
fix(iroh)!: Correct the error structure

flub:
## Description

For some odd reason AuthenticationError was given a branch that had
quinn::ConnectionError inside it.  But logically the connection error
has nothing to do with authentication.  That should have been a red
flag.

AuthenticationError itself is almost always wrapped in
ConnectingError, which does correctly have a quinn::ConnectionError
branch.  And the few places where it was directly returned to the user
it arguably **should** have been wrapped in a ConnectingError.

The result of this is that before this fix you would get a very
confusing authentication error if the remote client closed the
connection right at the same time as the handshake completed for
it (yes, this is difficult to do at the right time, and it only
happens for the client since that completes the handshake one network
hop before the server).  But this was no authentication error, it is
simply a closed connection.  The new error structure captures this
correctly.

Similarly the InternalConsistencyError belongs on the ConnectingError.
Though that one should be impossible to produce since it's supposed to
be an invariant.

## Breaking Changes

- `AuthenticationError` loses the `ConnectionError` and
  `InternalConsistencyError` branches.  Both are on the
  `ConnectingError` instead.
- `OutgoingZeroRttConnection::handshake_completed` and
  `IncomingZeroRttConnection` now return a `ConnectingError` instead
  of `AuthenticationError`.

## Notes & open questions

While I've managed to trigger this error somewhat occasionally using a
program that races the closing with the completed handshake in a very
tight loop **before** applying this fix.  I'm completely failing to
trigger it since applying this fix, so I can admire the beautiful new
error reporting this fix should give.  It's a bit confusing.

**edit**: it **is** confusing.  But it is correct.  Because my flaky failure does always close the connection *after* it completes the handshake.  So ALPN and EndpointId are always available.  And then you yield a valid `Connection` when awaiting an `Incoming`, it just is already closed.

This fix could also be made against main.  I believe the same commit should be able to be cherry-picked and will probably apply fairly clean.  Do you think I should make it against main?

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh/src/endpoint/connection.rs
+++ b/iroh/src/endpoint/connection.rs
@@ -605,11 +605,11 @@ impl OutgoingZeroRttConnection {
     /// the handshake will error and any data sent should be re-sent on a
     /// new stream.
     ///
-    /// This may fail with [`AuthenticationError::ConnectionError`], if there was
+    /// This may fail with [`ConnectingError::ConnectionError`], if there was
     /// some general failure with the connection, such as a network timeout since
     /// we initiated the connection.
     ///
-    /// This may fail with other [`AuthenticationError`]s, if the other side
+    /// This may fail with [`ConnectingError::HandshakeFailure`], if the other side
     /// doesn't use the right TLS authentication, which usually every iroh endpoint
     /// uses and requires.
     ///
@@ -915,11 +915,11 @@ pub struct IncomingZeroRttConnection {
 impl IncomingZeroRttConnection {
     /// Waits until the full handshake occurs and then returns a [`Connection`].
     ///
-    /// This may fail with [`AuthenticationError::ConnectionError`], if there was
+    /// This may fail with [`ConnectingError::ConnectionError`], if there was
     /// some general failure with the connection, such as a network timeout since
     /// we accepted the connection.
     ///
-    /// This may fail with other [`AuthenticationError`]s, if the other side
+    /// This may fail with [`ConnectingError::HandshakeFailure`], if the other side
     /// doesn't use the right TLS authentication, which usually every iroh endpoint
     /// uses and requires.
     ///


iroh/src/endpoint/connection.rs
```diff
@@ -244,13 +230,21 @@ fn conn_from_quinn_conn(
...
-    }
-    let remote_id = remote_id_from_quinn_conn(&conn)?;
-    let alpn = alpn_from_quinn_conn(&conn).ok_or_else(|| e!(AuthenticationError::NoAlpn))?;
+    let (remote_id, alpn) = match static_info_from_conn(&conn) {
+        Ok(val) => val,
+        Err(auth_err) => {
+            // If the authentication error raced with a connection error, the connection
+            // error wins.
+            if let Some(conn_err) = conn.close_reason() {
+                return Err(e!(ConnectingError::ConnectionError { source: conn_err }));
```
flub, line 241: Ok, so this actually **fixes** the racy thing I had.  It took me a while to figure out why.  But basically what happens is that because the handshake **does** complete on the client before the connection is closed, we never get the `auth_err` here, because all the info is available.  So it never checks on whether the connection is already closes.  Which in turn apparently means you just get a closed `Connection`.  Which is pretty neat.

I don't think this means this branch can be removed though, because if the client closes *before* it completes the handshake you can get here again.

ramfox:
Ah, yes to answer your question: yes I think this should be made against main.

iroh/src/endpoint/connection.rs
```diff
@@ -471,7 +476,7 @@ impl Future for Connecting {
             } else {
                 let quinn_conn = std::task::ready!(self.inner.poll_unpin(cx)?);
                 match conn_from_quinn_conn(quinn_conn, &self.ep) {
-                    Err(err) => return Poll::Ready(Err(err.into())),
+                    Err(err) => return Poll::Ready(Err(err)),
                     Ok(fut) => self.register_with_magicsock = Some(Box::pin(fut.err_into())),
```
matheus23: Can't you use `let fut = conn_from_quinn_conn(quinn_conn, &self.ep)?;` here?

flub: Indeed, that's nicer.  pushed.

flub:
> Ah, yes to answer your question: yes I think this should be made against main.

I'm retracting this offer and will push it here only.  Turns out  main doesn't have the fallible EndpointStateActor::register_connection yet, and I don't really want to deal with losing that and then having to re-apply it when merging main into here.

PR #3664, opened 2025-11-14
fix(tests): Also run the tests in isolation in the default profile

flub:
## Description

This means these tests also work when nextest run locally.

## Notes & open questions

I'm not sure why this wasn't done when the ci profile override was
chosen.  What am I missing?

## Change checklist

- [x] Self-review.


flub:
Meh, merging early.  @Arqu and @divagant-martian please still review.  We can always fixup.

PR #3666, opened 2025-11-14
feat: prune old, inactive paths

ramfox:
## Description

Whenever we insert a new path, trigger pruning paths.

We currently only prune IP paths, and pruning paths only occurs if we have more than 30 IP paths.

We will prune any paths that did not successfully holepunch. 

If there are still over 30 IP paths left, then we order the "inactive" paths (paths that have been closed, but at one point holepunched), and prune the paths that were closed earliest.

## Notes and Questions

  - Added constants:
    - `MAX_IP_PATHS` = 30 - maximum IP paths per endpoint
    - `MAX_INACTIVE_IP_PATHS` = 10 - maximum inactive IP paths to keep
  - New `PathState` field:
    - `status` - tracks the `PathStatus` of the path
  - New `PathStatus` enum:
    -  `PathStatus::Open` - is an open path
    - `PathStatus::Inactive(Instant)` - was opened once, but currently inactive
    - `PathStatus::Unusable` - we attempted to use it, but it never connected
    - `PathStatus::Unknown` - we don't know the status yet
  - New methods on `RemotePathState`:
    - `abandoned_path` - marks a path as abandoned with timestamp, triggered when we get the `PathEvent::Abandoned` event
    - `prune_paths` - triggers path pruning, occurs whenever we insert a path to the `RemotePathState`
    - changed `insert` to `insert_open_path`
  - New `prune_ip_paths` function with sophisticated logic:
    - Only prunes if IP paths exceed `MAX_IP_PATHS`
    - Never prunes active paths or paths of unknown status
    - Always prunes failed holepunch attempts (PathStatus::Unusable)
    - Keeps 10 most recently inactive paths that were previously
  successful
    - Special case: if all paths failed, keeps `MAX_IP_PATHS` instead of pruning everything
- Added tests for edge cases and the typical case

### DIFF
--- a/iroh/src/magicsock/remote_map/remote_state/path_state.rs
+++ b/iroh/src/magicsock/remote_map/remote_state/path_state.rs
@@ -42,7 +42,7 @@ pub(super) enum PathStatus {
     Open,
     /// This path was once opened, but was abandoned at the given [`Instant`].
     Inactive(Instant),
-    /// This path was never usable and was abandoned at the given [`Instant`].
+    /// This path was never usable (we attempted holepunching and it didn't work).
     Unusable,
     /// We have not yet attempted holepunching, or holepunching is currently in
     /// progress, so we do not know the usability of this path.


iroh/src/magicsock/endpoint_map/path_state.rs
matheus23, line 1: Err.. I think this file should be removed probably. I think it's from a stray merge, right?

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -442,6 +432,8 @@ impl RemoteStateActor {
                 self.paths
                     .insert(path_remote, Source::Connection { _0: Private });
                 self.select_path();
+                // TODO(ramfox): do we need to prune paths here?
+                self.prune_ip_paths();
```
matheus23: Yes!
I think every time we call `paths.insert` we should also call `self.prune_ip_paths()`. Because the number of paths might get increased above the high watermark due to the insert.

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -525,6 +520,7 @@ impl RemoteStateActor {
 
         trace!("ping received, triggering holepunching");
         self.trigger_holepunching().await;
+        // TODO(ramfox): potentially prune addrs here?
```
matheus23: Yeah I think we should do so here, since we have `paths.insert`ed above.

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -744,6 +740,7 @@ impl RemoteStateActor {
             self.paths.disco_ping_sent(addr.clone(), msg.tx_id);
             self.send_disco_message(addr, disco::Message::Ping(msg))
                 .await;
+            // TODO(ramfox): potentially prune addrs here?
```
matheus23: No, I don't think so. We only need to prevent the paths from becoming excessively populated. We can prevent that by checking each time we `paths.insert`.

iroh/src/magicsock/remote_map/remote_state/path_state.rs
```diff
@@ -134,6 +145,26 @@ impl RemotePathState {
...
+        &mut self,
+        pending: &VecDeque<transports::Addr>,
+        selected_path: &Option<transports::Addr>,
+        open_paths: impl Iterator<Item = &'a transports::Addr>,
+    ) {
+        // if the total number of paths, relay or ip, is less
+        // than the max inactive ip addrs we allow, bail early
+        if self.paths.len() < MAX_INACTIVE_IP_ADDRESSES {
+            return;
+        }
```
matheus23: We've already done this check in `RemoteStateActor::prune_ip_paths`.
I think I'd be in favor of removing this check and inlining the `prune_ip_paths` call below.

ramfox:
@flub can I merge into `feat-multipath-quic-nat`? Looks like the same tests are failing.

.config/nextest.toml
flub, line 18: This seems like a local change you committed.  Please revert, unless this really was on purpose.

iroh/src/magicsock/remote_map/remote_state/path_state.rs
```diff
@@ -123,4 +159,319 @@ pub(super) struct PathState {
...
+
+    // if the total number of ip paths is less than the max, bail early
+    if ip_paths.len() < MAX_IP_PATHS {
+        return;
+    }
+
+    // paths that were opened at one point but have previously been closed
+    let mut inactive = Vec::new();
+    // paths where we attempted hole punching but it not successful
+    let mut failed = Vec::new();
```
flub: might be worth using `Vec::with_capacity(ip_paths.len())` for these?

iroh/src/magicsock/remote_map/remote_state/path_state.rs
```diff
@@ -123,4 +159,319 @@ pub(super) struct PathState {
...
+
+    for (addr, state) in ip_paths {
+        match state.abandoned {
```
flub: `if let Some(abandoned)`?  Where's clippy?

.config/nextest.toml
matheus23: Not sure what's up with the diff view. This was most recently changed to the same value in https://github.com/n0-computer/iroh/pull/3595/files

ramfox: not sure what to do about this, the diff view keeps telling me I have things in there that are changes, but when you look at the actual `feat-multipath` code...they aren't changes 😂

ramfox: okay it's fixed now!

ramfox:
@flub this is reviewable now w/ rebases

same tests failing here that fail on `feat-multipath`

iroh/src/magicsock/remote_map/remote_state/path_state.rs
```diff
@@ -123,4 +165,338 @@ pub(super) struct PathState {
...
+    /// to open this path, but it did not work.
+    ///
+    /// If this is `false` and the `abandoned` field is `None`, than we do not know
+    ///  yet if this path is usable.
+    pub(super) usable: bool,
+    /// The last time a path with this addr was abandoned.
+    ///
+    /// If this is `Some` and usable is `None`, than we attempted to use this path and it
+    /// did not work.
+    pub(super) abandoned: Option<Instant>,
```
dignifiedquire: It seems there is a state of `usable: true` and `abandoned: Some(..)` which we do not want to actually be able to have, so maybe we could use a state enum like this instead of these two fields

```rust
// TODO: bikeshed name
enum Usability { 
  Usable,
  Abandoned(Instant),
  Unknown,
}
```

iroh/src/magicsock/remote_map/remote_state/path_state.rs
```diff
@@ -123,4 +165,338 @@ pub(super) struct PathState {
...
+/// - We do not prune holepunched paths that are currently not in use too quickly. For example, if a large number of untested paths are added at once, we will not immediately prune all of the unused, but valid, paths at once.
+fn prune_ip_paths(paths: &mut FxHashMap<transports::Addr, PathState>) {
+    // if the total number of paths is less than the max, bail early
+    if paths.len() < MAX_IP_PATHS {
+        return;
+    }
+
+    let ip_paths: Vec<_> = paths.iter().filter(|(addr, _)| addr.is_ip()).collect();
+
+    // if the total number of ip paths is less than the max, bail early
```
dignifiedquire, line 207: We are trying to get away from making IP paths special, but rather referring to direct paths, so maybe use that naming?

iroh/src/magicsock/remote_map/remote_state/path_state.rs
```diff
@@ -29,19 +37,41 @@ pub(super) struct RemotePathState {
...
+    pub(super) fn abandoned_path(&mut self, addr: &transports::Addr) {
+        if let Some(state) = self.paths.get_mut(addr) {
+            state.abandoned = Some(Instant::now());
+        }
+    }
+
+    /// Mark a path as opened.
+    ///
+    /// If this path does not exist, it does nothing to the
+    /// `RemotePathState`
```
flub: That's a bit odd, isn't it? Wouldn't it be a bug that you mark a path as opened that hasn't been inserted?  I could be tempted with an insert and emit a trace log to signal the bug.

iroh/src/magicsock/remote_map/remote_state/path_state.rs
```diff
@@ -123,4 +165,338 @@ pub(super) struct PathState {
...
+    /// If this is `Some` and usable is `None`, than we attempted to use this path and it
+    /// did not work.
+    pub(super) abandoned: Option<Instant>,
+}
+
+/// Prunes the IP paths in the paths HashMap.
+///
+/// Only prunes if the number of IP paths is above [`MAX_IP_PATHS`].
+///
+/// Keeps paths that are active or have never been holepunched.
```
flub:
```suggestion
/// Keeps paths that are active or have never been used in holepunching attempts.
```

Otherwise it sounds like failed holepunches are kept forever to me, but the next line clarifies that's the wrong interpretation.

iroh/src/magicsock/remote_map/remote_state/path_state.rs
flub: Update: given that this is always used right after an `.insert()` I'd even be tempted to do the implicit insert as a feature, so there's only one call to self.paths at call sites.

flub (approved):

iroh/src/magicsock/remote_map/remote_state/path_state.rs
flub: Let's keep it like this for now. IP paths are still special because of holepunching.

ramfox: Ah, see, I was thinking the opposite, that this was literally only for IP addresses, not any direct address. I figured there might be different rules for some other direct address we collect, like maybe we always keep 5 bluetooth addrs separately from the 10 ip addresses.

iroh/src/magicsock/remote_map/remote_state/path_state.rs
ramfox: yea this is so much nicer & an easy refactor.

iroh/src/magicsock/remote_map/remote_state/path_state.rs
ramfox: hmmm also given that `insert` is only used when the path is considered open, I think it may be best to change `insert` to `insert_open_path` and ensure everything is correctly done in there. That may be what you were referring to.

flub: That's even better, indeed. I hadn't traced that side of the calls because I was too lazy to open my editor locally...

iroh/src/magicsock/remote_map/remote_state/path_state.rs
ramfox: changed the language

iroh/src/magicsock/remote_map/remote_state/path_state.rs
```diff
@@ -26,22 +34,54 @@ pub(super) struct RemotePathState {
...
+/// Describes the usability of this path, i.e. whether it has ever been opened,
+/// when it was closed, or if it has never been usable.
+#[derive(Debug, Default)]
+pub(super) enum PathStatus {
+    /// This path is open and active.
+    Open,
+    /// This path was once opened, but was abandoned at the given [`Instant`].
+    Inactive(Instant),
+    /// This path was never usable and was abandoned at the given [`Instant`].
+    Unusable,
```
dignifiedquire, line 46: there is no Instant?

ramfox: not being used, since if it's unusable, current logic dictates that we remove it regardless of the time.

Since this is just an internal API, I wasn't worried about changing the values of the enum if we find a need for it in the future.

flub: This state isn't persisted anywhere. We can add instants when we need them later, leaving it unused currently is a bit odd.

ramfox: oh I see, the issue is I never changed the docs to reflect me removing it!

dignifiedquire: > oh I see, the issue is I never changed the docs to reflect me removing it!

yeah, sorry should have been more clear, the docs said there is an instant but there was none

PR #3606, opened 2025-11-04
feat(metrics): add connection latency tracking

Arqu:
## Change checklist

- [ ] Self-review.
- [ ] Documentation updates following the [style guide](https://rust-lang.github.io/rfcs/1574-more-api-documentation-conventions.html#appendix-a-full-conventions-text), if relevant.
- [ ] Tests if relevant.
- [ ] All breaking changes documented.
  - [ ] List all breaking changes in the above "Breaking Changes" section.
  - [ ] Open an issue or PR on any number0 repos that are affected by this breaking change. Give guidance on how the updates should be handled or do the actual updates themselves. The major ones are:
    - [ ] [`quic-rpc`](https://github.com/n0-computer/quic-rpc)
    - [ ] [`iroh-gossip`](https://github.com/n0-computer/iroh-gossip)
    - [ ] [`iroh-blobs`](https://github.com/n0-computer/iroh-blobs)
    - [ ] [`dumbpipe`](https://github.com/n0-computer/dumbpipe)
    - [ ] [`sendme`](https://github.com/n0-computer/sendme)

PR #3669, opened 2025-11-17
feat(multipath): add back path congestion metrics

Frando:
## Description

Based on https://github.com/n0-computer/iroh/pull/3672

Add back some of the path congestion metrics added in #3491. @arqu please check my port of these metrics and if the calculations are correct. I commented-out the metrics added in #3491 that I didn't know how to port. But I think I got the important ones.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/magicsock/remote_map/remote_state/metrics.rs
+++ b/iroh/src/magicsock/remote_map/remote_state/metrics.rs
@@ -48,18 +48,17 @@ impl MetricsTracker {
                     .observe(stats.rtt.as_millis() as f64);
             }
 
-            let Some(rtt_variance) = self.path_rtt_variance.get_mut(path_id) else {
-                continue;
-            };
-            rtt_variance.add_rtt_sample(stats.rtt);
-            if let Some(variance) = rtt_variance.rtt_variance() {
-                metrics
-                    .path_rtt_variance_ms
-                    .observe(variance.as_millis() as f64);
-            }
+            if let Some(rtt_variance) = self.path_rtt_variance.get_mut(path_id) {
+                rtt_variance.add_rtt_sample(stats.rtt);
+                if let Some(variance) = rtt_variance.rtt_variance() {
+                    metrics
+                        .path_rtt_variance_ms
+                        .observe(variance.as_millis() as f64);
+                }
 
-            let quality = rtt_variance.quality_score(loss_rate);
-            metrics.path_quality_score.observe(quality);
+                let quality = rtt_variance.quality_score(loss_rate);
+                metrics.path_quality_score.observe(quality);
+            };
         }
     }
 


iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -75,6 +78,9 @@ const APPLICATION_ABANDON_PATH: u8 = 30;
 /// in a high frequency, and to keep data about previous path around for subsequent connections.
 const ACTOR_MAX_IDLE_TIMEOUT: Duration = Duration::from_secs(60);
 
+/// Interval in which connection and path metrics are emitted.
+const METRICS_INTERVAL: Duration = Duration::from_secs(10);
```
flub, line 90: Soo, I'm going to argue that this is fundamentally doing metrics wrong.  Metrics increment counters or record historgrams at discreet points.  If you are recording something on a timer you were supposed to provide the metrics that can compute the thing, and the metrics backend can compute it on the fly over the range it wants, using sampling intervals it wants etc etc.

iroh/src/magicsock/remote_map/remote_state/metrics.rs
flub, line 1: This is metrics backend code.  I'll have a much easier time with this PR if we don't try and do anything like that here.

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -220,6 +226,7 @@ impl EndpointStateActor {
         trace!("actor started");
         let idle_timeout = MaybeFuture::None;
         tokio::pin!(idle_timeout);
+        let mut metrics_interval = time::interval(METRICS_INTERVAL);
```
dignifiedquire, line 253: should this be cfged as well?

flub (commented):
I'd suggest to limit this to figure out what the easy and interesting counters to add.  Nothing more.  If we need congestion metrics we can do that in a later pass.

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -1012,6 +1034,13 @@ impl EndpointStateActor {
             }
         }
     }
+
+    fn record_metrics(&mut self) {
+        #[cfg(feature = "metrics")]
+        for state in self.connections.values_mut() {
+            state.record_metrics_periodic(&self.metrics, self.selected_path.get());
```
dignifiedquire, line 1089: sorry if I am missing sth, but should this be the path for the specific connection that we pass in here, this seems like a generic one?

iroh/src/magicsock/remote_map/remote_state/metrics.rs
```diff
@@ -0,0 +1,158 @@
...
+            };
+            metrics.path_packet_loss_rate.observe(loss_rate);
+
+            if Some(remote) == selected_path.as_ref() {
+                metrics
+                    .connection_latency_ms
+                    .observe(stats.rtt.as_millis() as f64);
+            }
+
+            let Some(rtt_variance) = self.path_rtt_variance.get_mut(path_id) else {
```
dignifiedquire: maybe do sth like `if let Some` instead of the continue, not that you accidentally skip metrics below that could have been added even though `path_rtt_variance` is not available

iroh/src/magicsock/endpoint_map/endpoint_state.rs
```diff
@@ -75,6 +78,9 @@ const APPLICATION_ABANDON_PATH: u8 = 30;
 /// in a high frequency, and to keep data about previous path around for subsequent connections.
 const ACTOR_MAX_IDLE_TIMEOUT: Duration = Duration::from_secs(60);
 
+/// Interval in which connection and path metrics are emitted.
+const METRICS_INTERVAL: Duration = Duration::from_secs(10);
```
Arqu: To be fair, it gets exposed like this and then the scraper picks it up every x seconds usually 10 or 15 so at worst it's slightly laggy. Though this does obliterate any resolution on the gauges.

iroh/src/magicsock/remote_map/remote_state.rs
Frando: Failed to get the `cfg` be happy with `tokio::select!`. Maybe should be forever-pending then if the cfg is not active?

iroh/src/magicsock/remote_map/remote_state/metrics.rs
Frando: This is present on main. It is not new, just ported from main to feat-multipath. See #3606 and #3491.

Frando: Happy to not do this here, but IIRC these are the metrics we are most interested in to compare feat-multipath to main via n0des? Correct me if I'm wrong please, @Arqu .

Edit: I've split the PR, the simple metrics are now here https://github.com/n0-computer/iroh/pull/3672 and this PR is about adding back the congestion metrics.

iroh/src/magicsock/remote_map/remote_state.rs
Frando: We have a single selected path per ~~connection ~~remote endpoint. This is passed through here, because the metrics tracking code further down uses the RTT of the selected path as the connection's RTT. (There is no connection-level RTT exposed anymore with multipath on the quinn level).

iroh/src/magicsock/remote_map/remote_state.rs
Frando: What would be discreet points to record connection latency at? Open/close path events?

Frando:
> I'd suggest to limit this to figure out what the easy and interesting counters to add. Nothing more. If we need congestion metrics we can do that in a later pass.

This is present on main. It is not new, just ported from main to feat-multipath. See https://github.com/n0-computer/iroh/pull/3606 and https://github.com/n0-computer/iroh/pull/3491.

Happy to not do this here and only the simple counters initially, but IIRC these are the metrics we are most interested in to compare feat-multipath to main via n0des? Correct me if I'm wrong please, @Arqu .

iroh/src/magicsock/remote_map/remote_state.rs
dignifiedquire: yeah, sth like that I have done in the past

iroh/src/magicsock/remote_map/remote_state.rs
dignifiedquire: but why is it the same for all connections?

Frando: Currently the selected path logic works on the level of the remote endpoint, and the result is then applied to all connections.

Frando:
I've split this PR: The basic, simple counter metrics are now in https://github.com/n0-computer/iroh/pull/3672. This PR is now based on 3672 and adds back the congestion metrics.

iroh/src/magicsock/remote_map/remote_state/metrics.rs
Frando: done

flub:
Going to close this one. We did talk about it briefly and we're not sure how applicable these metrics are. They were added for a specific case of the old code that doesn't necessarily translate to the new code. Though I'm not saying that we won't have to add more metrics, but let's find out from experience what we need to know.

PR #3672, opened 2025-11-17
feat(multipath): add back basic metrics

Frando:
## Description

This adds back basic metrics to `feat-multipath`.

* Refactor the `conn_` metrics (they were all currently unused on `feat-multipath`) into something meaningful:  Track the total number of opened and closed connections - their diff at any time is the number of currently-active conne
* Adds `transport_ip_paths_added`, `transport_ip_paths_removed`, `transport_relay_paths_added`, `transport_relay_paths_removed` counters
* Removes metrics that are unused and - IMO  - not going to come back
* Comments out metrics that are unused and likely to come back 

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/magicsock/metrics.rs
+++ b/iroh/src/magicsock/metrics.rs
@@ -45,13 +45,13 @@ pub struct Metrics {
     pub num_conns_closed: Counter,
 
     /// Number of IP transport paths opened.
-    pub transport_paths_ip_added: Counter,
+    pub transport_ip_paths_added: Counter,
     /// Number of IP transport paths closed.
-    pub transport_paths_ip_removed: Counter,
+    pub transport_ip_paths_removed: Counter,
     /// Number of relay transport paths opened.
-    pub transport_paths_relay_added: Counter,
+    pub transport_relay_paths_added: Counter,
     /// Number of relay transport paths closed.
-    pub transport_paths_relay_removed: Counter,
+    pub transport_relay_paths_removed: Counter,
 
     pub actor_tick_main: Counter,
     pub actor_tick_msg: Counter,
@@ -60,25 +60,4 @@ pub struct Metrics {
     pub actor_tick_direct_addr_heartbeat: Counter,
     pub actor_link_change: Counter,
     pub actor_tick_other: Counter,
-    // /// Histogram of connection latency in milliseconds across all endpoint connections.
-    // #[default(Histogram::new(vec![1.0, 5.0, 10.0, 25.0, 50.0, 100.0, 250.0, 500.0, 1000.0, f64::INFINITY]))]
-    // pub connection_latency_ms: Histogram,
-    // /*
-    // * Path Congestion Metrics
-    // */
-    // /// Number of times a path was marked as outdated due to consecutive ping failures.
-    // pub path_marked_outdated: Counter,
-    // /// Number of ping failures recorded across all paths.
-    // pub path_ping_failures: Counter,
-    // /// Number of consecutive failure resets (path recovered).
-    // pub path_failure_resets: Counter,
-    // /// Histogram of packet loss rates (0.0-1.0) observed on UDP paths.
-    // #[default(Histogram::new(vec![0.0, 0.01, 0.05, 0.1, 0.2, 0.5, 1.0]))]
-    // pub path_packet_loss_rate: Histogram,
-    // /// Histogram of RTT variance (in milliseconds) as a congestion indicator.
-    // #[default(Histogram::new(vec![0.0, 1.0, 5.0, 10.0, 20.0, 50.0, 100.0, 200.0]))]
-    // pub path_rtt_variance_ms: Histogram,
-    // /// Histogram of path quality scores (0.0-1.0).
-    // #[default(Histogram::new(vec![0.0, 0.3, 0.5, 0.7, 0.85, 0.95, 1.0]))]
-    // pub path_quality_score: Histogram,
 }
--- a/iroh/src/magicsock/remote_map/remote_state.rs
+++ b/iroh/src/magicsock/remote_map/remote_state.rs
@@ -192,7 +192,7 @@ impl RemoteStateActor {
         Self {
             endpoint_id,
             local_endpoint_id,
-            metrics,
+            metrics: metrics.clone(),
             local_direct_addrs,
             relay_mapped_addrs,
             discovery,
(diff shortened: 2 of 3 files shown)


iroh/src/magicsock/metrics.rs
```diff
@@ -14,11 +14,8 @@ pub struct Metrics {
     pub send_ipv4: Counter,
     pub send_ipv6: Counter,
     pub send_relay: Counter,
-    pub send_relay_error: Counter,
 
     // Data packets (non-disco)
-    pub send_data: Counter,
```
Frando, line 20: This was unused, and it would be just the sum of `send_ipv4`, `send_ipv6`, `send_relay`, so I removed it because the sum can be calculated client-side.

iroh/src/magicsock/metrics.rs
```diff
@@ -14,11 +14,8 @@ pub struct Metrics {
...
-    pub send_data_network_down: Counter,
```
Frando, line 21: This was unused, not sure if we'd want this back?

iroh/src/magicsock/metrics.rs
```diff
@@ -67,36 +69,25 @@ pub struct Metrics {
     pub actor_tick_direct_addr_heartbeat: Counter,
     pub actor_link_change: Counter,
     pub actor_tick_other: Counter,
-
-    /// Number of endpoints we have attempted to contact.
-    pub endpoints_contacted: Counter,
```
Frando, line 60: I don't think we ever could do this in a good way, because we don't keep lists of endpoint ids over restarts, so this was already a bad metric for a while - removed it.

iroh/src/magicsock/metrics.rs
```diff
@@ -67,36 +69,25 @@ pub struct Metrics {
...
-    /// Number of endpoints we have managed to contact directly.
-    pub endpoints_contacted_directly: Counter,
```
Frando, line 62: same as above, I don't think there's a good way to do it meaningfully.

iroh/src/magicsock/metrics.rs
```diff
@@ -67,36 +69,25 @@ pub struct Metrics {
...
-
-    /// Number of connections with a successful handshake.
-    pub connection_handshake_success: Counter,
-    /// Number of connections with a successful handshake that became direct.
-    pub connection_became_direct: Counter,
```
Frando, line 67: These two are better handled by the new `conn_` metrics above

dignifiedquire (approved):

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -1424,6 +1435,44 @@ impl Future for OnClosed {
...
+    fn record(&self, metrics: &MagicsockMetrics) {
+        match self {
+            TransportSummary::IpOnly => {
+                metrics.num_conns_transport_ip_only.inc();
+            }
+            TransportSummary::RelayOnly => {
+                metrics.num_conns_transport_relay_only.inc();
+            }
+            TransportSummary::IpAndRelay => {
+                metrics.num_conns_transport_ip_and_relay.inc();
```
flub: Talked with @Arqu and we agreed to remove these 3 metrics and instead have these 4:

- `ip_transport_paths_added`: counter
- `ip_transport_paths_removed`: counter
- `relay_transport_paths_added`: counter
- `relay_transport_paths_removed`: counter

That way we can increment counters when paths are opened and closed.  It doesn't give you full view into what connections have been doing, but neither does this, because typically nearly everything would have been `num_conns_transport_ip_and_relay`.

Frando:
I rebased this branch, removed the TransportSummary metrics and added the paths counters. I named the counters to `transport_paths_ip_added` etc, so that they share a prefix and are easily correlated when having a sorted list of metrics - lmk if you prefer your naming @flub.

iroh/src/magicsock/remote_map/remote_state/path_state.rs
```diff
@@ -54,7 +54,16 @@ impl RemotePathState {
     /// Insert a new address of an open path into our list of paths.
     ///
     /// This will emit pending resolve requests and trigger pruning paths.
-    pub(super) fn insert_open_path(&mut self, addr: transports::Addr, source: Source) {
+    pub(super) fn insert_open_path(
+        &mut self,
+        addr: transports::Addr,
+        source: Source,
+        metrics: &MagicsockMetrics,
```
flub: I'd suggest to clone MagicsockMetrics into here, it exists for the same duration as the actor anyway. But this is optional, as you wish.

iroh/src/magicsock/remote_map/remote_state/path_state.rs
```diff
@@ -54,7 +54,16 @@ impl RemotePathState {
...
+    ) {
+        match addr {
+            transports::Addr::Ip(_) => metrics.transport_paths_ip_added.inc(),
```
flub: Names are pretty subjective, I think I'll argue for `transport_{ip|relay}_paths_added`. Reason being that `paths_added` is the counter unit and `tranport_ip` belongs together as `ip` is the qualifier to transports. Good call to put transports at the front though.

Now what I *really* want is labels so we can do `paths_added{transport=ip}` but hey.

iroh/src/magicsock/metrics.rs
```diff
@@ -55,36 +60,25 @@ pub struct Metrics {
...
+    // pub path_failure_resets: Counter,
+    // /// Histogram of packet loss rates (0.0-1.0) observed on UDP paths.
+    // #[default(Histogram::new(vec![0.0, 0.01, 0.05, 0.1, 0.2, 0.5, 1.0]))]
+    // pub path_packet_loss_rate: Histogram,
+    // /// Histogram of RTT variance (in milliseconds) as a congestion indicator.
+    // #[default(Histogram::new(vec![0.0, 1.0, 5.0, 10.0, 20.0, 50.0, 100.0, 200.0]))]
+    // pub path_rtt_variance_ms: Histogram,
+    // /// Histogram of path quality scores (0.0-1.0).
+    // #[default(Histogram::new(vec![0.0, 0.3, 0.5, 0.7, 0.85, 0.95, 1.0]))]
+    // pub path_quality_score: Histogram,
```
flub: Shall we remove the commented out stuff? I think we can come up with this again if we do end up needing something like it (well, I'm hoping we'd come up with something better 😉 )

PR #3676, opened 2025-11-18
fix(multipath): fix remote state actor termination

Frando:
## Description

* fix idle timeout clear condition (previously it would hot loop)
* fix hot loop when local_addrs watchable becomes disconnected during shutdown
* when sending a datagram fails in the transports sender, include the dst address in the error message
* do not break the RemoteStateActor when sending a datagram fails

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/magicsock/remote_map/remote_state.rs
+++ b/iroh/src/magicsock/remote_map/remote_state.rs
@@ -126,7 +126,8 @@ pub(super) struct RemoteStateActor {
     /// Our local addresses.
     ///
     /// These are our local addresses and any reflexive transport addresses.
-    local_addrs: n0_watcher::Direct<BTreeSet<DirectAddr>>,
+    /// They are called "direct addresses" in the magic socket actor.
+    local_direct_addrs: n0_watcher::Direct<BTreeSet<DirectAddr>>,
     /// Shared state to allow to encrypt DISCO messages to peers.
     disco: DiscoState,
     /// The mapping between endpoints via a relay and their [`RelayMappedAddr`]s.
@@ -181,7 +182,7 @@ impl RemoteStateActor {
     pub(super) fn new(
         endpoint_id: EndpointId,
         local_endpoint_id: EndpointId,
-        local_addrs: n0_watcher::Direct<BTreeSet<DirectAddr>>,
+        local_direct_addrs: n0_watcher::Direct<BTreeSet<DirectAddr>>,
         disco: DiscoState,
         relay_mapped_addrs: AddrMap<(RelayUrl, EndpointId), RelayMappedAddr>,
         metrics: Arc<MagicsockMetrics>,
@@ -192,7 +193,7 @@ impl RemoteStateActor {
             endpoint_id,
             local_endpoint_id,
             metrics,
-            local_addrs,
+            local_direct_addrs,
             relay_mapped_addrs,
             discovery,
             disco,
@@ -275,7 +276,7 @@ impl RemoteStateActor {
                         self.selected_path.set(None).ok();
                     }
                 }
-                res = self.local_addrs.updated() => {
+                res = self.local_direct_addrs.updated() => {
                     if let Err(n0_watcher::Disconnected) = res {
                         trace!("direct address watcher disconnected, shutting down");
                         break;
@@ -387,7 +388,11 @@ impl RemoteStateActor {
                 // We never want to send to our local addresses.
                 // The local address set is updated in the main loop so we can use `peek` here.
                 if let transports::Addr::Ip(sockaddr) = addr
-                    && self.local_addrs.peek().iter().any(|a| a.addr == *sockaddr)
+                    && self
+                        .local_direct_addrs
+                        .peek()
+                        .iter()
+                        .any(|a| a.addr == *sockaddr)
                 {
                     trace!(%sockaddr, "not sending datagram to our own address");
                 } else if let Err(err) = self.send_datagram(addr.clone(), transmit.clone()).await {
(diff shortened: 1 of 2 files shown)


matheus23:
@Frando here's another version of fixing the idle timeout: https://github.com/n0-computer/iroh/pull/3678

I think I'd prefer that version :v:

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -253,7 +243,7 @@ impl RemoteStateActor {
                         self.selected_path.set(None).ok();
                     }
                 }
-                _ = self.local_addrs.updated() => {
+                _ = self.local_addrs.updated(), if self.local_addrs.is_connected() => {
```
matheus23: I mentioned this briefly while on the call, but I think we got distracted by something else before the answer:
Wouldn't it make sense to match on the `Result` you get back?
- In case it's `Err(Disconnected)` you know that you can probably close the actor. This should only happen when `magicsock::DiscoveredDirectAddrs` is dropped, which should only happen when `MagicSock` is dropped. Soo.. I think dropping out of the loop at this point is totally fine, even desired.
- IIUC, the `, if self.local_addrs.is_connected()` is a *pre*condition, so it's possible that you start polling `updated()` while it's connected, but then it disconnects and resolves to `Err(Disconnected)`. I don't think that was the intended change here. The fix for this would be to just match the `Result` from `updated`.

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -278,29 +268,31 @@ impl RemoteStateActor {
...
 
-            if self.connections.is_empty() && inbox.is_idle() && idle_timeout.is_none() {
+            let is_idle = self.connections.is_empty() && inbox.is_idle();
+            if idle_timeout.is_none() && is_idle {
                 trace!("start idle timeout");
                 idle_timeout
                     .as_mut()
                     .set_future(time::sleep(ACTOR_MAX_IDLE_TIMEOUT));
-            } else if idle_timeout.is_some() {
+            } else if idle_timeout.is_some() && !is_idle {
```
matheus23: As I commented, I think I'd prefer the other version from my PR to this fix. There's lots of states that `idle_timeout` can be in, and it's probably easier to default it to sth that continuously gets reset further and further into the future.

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -278,29 +268,31 @@ impl RemoteStateActor {
...
     #[instrument(skip(self))]
-    async fn handle_message(&mut self, msg: RemoteStateMessage) -> n0_error::Result<()> {
+    async fn handle_message(&mut self, msg: RemoteStateMessage) {
         // trace!("handling message");
         match msg {
             RemoteStateMessage::SendDatagram(transmit) => {
-                self.handle_msg_send_datagram(transmit).await?;
+                if let Err(err) = self.handle_msg_send_datagram(transmit).await {
+                    warn!("failed to send datagram: {err:#}");
+                }
```
matheus23: I think this is a bug:
We should avoid short-circuiting not on the `handle_msg_send_datagram` level, but on the `send_datagram` level. Otherwise a single path that fails will make us stop sending datagrams to any other paths that might still work.

So in essence that just means this `if let Err(err) = ...` should be moved inside the `handle_msg_send_datagram` call.

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -193,19 +193,12 @@ impl RemoteStateActor {
         // we don't explicitly set a span we get the spans from whatever call happens to
         // first create the actor, which is often very confusing as it then keeps those
         // spans for all logging of the actor.
-        let task = task::spawn(
-            async move {
-                if let Err(err) = self.run(rx).await {
-                    error!("actor failed: {err:#}");
```
flub, line 225: So this was a poor-man's attempt at being able to have "fatal" errors.  The right thing of course is to panic and spawn the tasks in a JoinSet and making sure the panic bubbles up to the endpoint.  But that's for another day I guess.  For now making this infallible is the safer approach I think.  While if forces you to handle the fatal error everywhere it will probably result in better code.

iroh/src/magicsock/remote_map/remote_state.rs
flub: I agree with this.  Though hope that one day we have structured concurrency and disconnected just can't happen.

This... might be another thing to look into as a fix for this.  I went with spawning these actors detached because it was the fasted thing to do and get on with thinking about the logic I had to solve.

But we should totally improve this.  They should be spawned in a JoinSet, but that might need an RemoteMap actor or something to manage these tasks and that itself be a child of the magicsock::Actor.  Or manage the JoinSet immediately in the magicsock::Actor, but I didn't know if that violated too many boundaries.

Though having said all that, I think we should handle this disconnect stuff here correctly and do the structured-concurrency fix some later time.

iroh/src/magicsock/transports.rs
```diff
@@ -539,9 +539,13 @@ impl TransportsSender {
             }
         }
         if any_match {
-            Err(io::Error::other("all available transports failed"))
+            Err(io::Error::other(format!(
+                "all available transports failed for destination {dst:?}"
```
flub: Hum, convention in the stdlib is that the caller should provide the context.  Don't we have somewhere else to add context?

Frando: sgtm. pushed a commit that does this.

iroh/src/magicsock/remote_map/remote_state.rs
matheus23: I have a draft PR for doing the JoinSet thing: https://github.com/n0-computer/iroh/pull/3681

Apart from that, I've [added a commit](https://github.com/n0-computer/iroh/pull/3676/commits/3f8d72a6996325bbfbf1b9c86d8a633fbb29e0e5) that removes the `if self.local_addrs.is_connected()` in favor of matching the result returned from `local_addrs.updated()´.

PR #3678, opened 2025-11-18
Simplify the idle timeout logic & don't short-circuit in `handle_msg_send_datagram`

matheus23:
Intended for merging into https://github.com/n0-computer/iroh/pull/3676

- Makes `idle_timeout` always be a `time::Sleep` that is `.reset` once there is any activity. I find this much easier to reason about.
- Fixes a bug where `handle_msg_send_datagram` short-circuits when sending on one path fails. Instead, it should *try* all paths, even if some of them fail.

### DIFF
--- a/iroh/src/magicsock/remote_map/remote_state.rs
+++ b/iroh/src/magicsock/remote_map/remote_state.rs
@@ -335,39 +335,28 @@ impl RemoteStateActor {
         // though we might not have a relay transport or ip-capable transport set up.
         // So these errors must not be fatal for this actor (or even this operation).
 
-        let mut any_succeeded = false;
-
         if let Some(addr) = self.selected_path.get() {
             trace!(?addr, "sending datagram to selected path");
 
-            if let Err(err) = self.send_datagram(addr, transmit).await {
-                trace!("failed to send datagram: {err:#}");
-            } else {
-                any_succeeded = true;
+            if let Err(err) = self.send_datagram(addr.clone(), transmit).await {
+                debug!(?addr, "failed to send datagram on selected_path: {err:#}");
             }
         } else {
             trace!(
                 paths = ?self.paths.keys().collect::<Vec<_>>(),
                 "sending datagram to all known paths",
             );
+            if self.paths.is_empty() {
+                warn!("Cannot send datagrams: No paths to remote endpoint known");
+            }
             for addr in self.paths.keys() {
                 if let Err(err) = self.send_datagram(addr.clone(), transmit.clone()).await {
-                    trace!("failed to send datagram: {err:#}");
-                } else {
-                    any_succeeded = true;
+                    debug!(?addr, "failed to send datagram: {err:#}");
                 }
             }
             // This message is received *before* a connection is added.  So we do
             // not yet have a connection to holepunch.  Instead we trigger
             // holepunching when AddConnection is received.
-        }
-
-        if !any_succeeded {
-            warn!(
-                selected_path = ?self.selected_path.get(),
-                paths = ?self.paths.keys().collect::<Vec<_>>(),
-                "failed to send datagram on any path"
-            );
         }
     }
 


iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -212,8 +212,8 @@ impl RemoteStateActor {
     /// discipline is needed to not turn pending for a long time.
     async fn run(mut self, mut inbox: GuardedReceiver<RemoteStateMessage>) {
         trace!("actor started");
-        let idle_timeout = MaybeFuture::None;
-        tokio::pin!(idle_timeout);
+        let idle_timeout = time::sleep(ACTOR_MAX_IDLE_TIMEOUT);
+        n0_future::pin!(idle_timeout);
```
dignifiedquire, line 216: how about using the std lib? https://doc.rust-lang.org/std/pin/macro.pin.html

matheus23: Fair. I wanted to match the rest of the code's style.

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -330,25 +328,33 @@ impl RemoteStateActor {
...
-                self.send_datagram(addr.clone(), transmit.clone()).await?;
+                if let Err(err) = self.send_datagram(addr.clone(), transmit.clone()).await {
+                    trace!("failed to send datagram: {err:#}");
+                }
             }
             // This message is received *before* a connection is added.  So we do
             // not yet have a connection to holepunch.  Instead we trigger
             // holepunching when AddConnection is received.
         }
-        Ok(())
```
dignifiedquire, line 354: we should log a warning though or even an error if all failed? as that means nothing is currently reachable

iroh/src/magicsock/remote_map/remote_state.rs
matheus23: Err.. Seems like that macro actually does something different. It doesn't seem to be a drop-in replacement for neither `tokio::pin!` nor `n0_future::pin!` in this situation.

iroh/src/magicsock/remote_map/remote_state.rs
matheus23: Logging a failure now.

I don't want to return an error and make the actor stop again. I think it's possible that we'll not have any paths or sth like that in rare cases and it recovers.

iroh/src/magicsock/remote_map/remote_state.rs
flub: `let mut idle_timeout = std::pin::pin!(time::sleep(...))` is the std version.  I don't really know which I prefer, this version existed in n0_future so I assumed we liked it for some reason?

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -264,20 +269,12 @@ impl RemoteStateActor {
                     if self.connections.is_empty() && inbox.close_if_idle() {
                         trace!("idle timeout expired and still idle: terminate actor");
                         break;
+                    } else {
+                        // Seems like we weren't really idle, so we reset
+                        idle_timeout.as_mut().reset(Instant::now() + ACTOR_MAX_IDLE_TIMEOUT);
```
flub, line 274: can this really happen?

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -333,25 +328,47 @@ impl RemoteStateActor {
...
+        // So these errors must not be fatal for this actor (or even this operation).
+
+        let mut any_succeeded = false;
+
         if let Some(addr) = self.selected_path.get() {
             trace!(?addr, "sending datagram to selected path");
-            self.send_datagram(addr, transmit).await?;
+
+            if let Err(err) = self.send_datagram(addr, transmit).await {
+                trace!("failed to send datagram: {err:#}");
```
flub: Maybe make this debug?

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -333,25 +328,47 @@ impl RemoteStateActor {
...
+            }
         } else {
             trace!(
                 paths = ?self.paths.keys().collect::<Vec<_>>(),
                 "sending datagram to all known paths",
             );
             for addr in self.paths.keys() {
-                self.send_datagram(addr.clone(), transmit.clone()).await?;
+                if let Err(err) = self.send_datagram(addr.clone(), transmit.clone()).await {
+                    trace!("failed to send datagram: {err:#}");
```
flub: I'd be tempted to put this on debug and don't do the any_succeeded dance with the warn.

iroh/src/magicsock/remote_map/remote_state.rs
matheus23: Yeah, I think in theory there could be a race condition between the inbox getting a message and the idle timeout firing.

flub: The `select!` call is `biased` and the timeout is at the end?  Or is there some other channel trickery here?  I'm not entirely happy with the complexity of this guarded_channel, it makes it hard to reason about this.

matheus23: You might be right that it's not possible.

Given we already have a stack of PRs (`feat-multipath` <- `Frando/mp-fix-remote-state-actor-loop` <- `matheus23/mp-discovery`) and getting rid of guarded_channel is probably too big for this PR, I think I'll keep those changes to another PR.

I have some ideas about how we might be able to get rid of `guarded_channel`. Will PR after merging this.

matheus23:
Not sure why netsim fails now :/ I've tried to spelunking in the logs, but couldn't find anything easily. I thought we'd fixed all netsim issues in this stack of PRs. :/ TBD.

Frando:
Weird, it seems to complete successfully yet the logs say the process completed with error code 1. @Arqu any ideas?

matheus23:
This log seems to come from chuck's `util.py` file, IIUC. And it seems like it defaults to exit code 1, if no exit code is provided?
Looking at the conditions when this function is called, it seems like it's called in `eject`, which is only called in `main.py` when `monitor_short_processes` returns any process errors, which can basically happen if the process returns a non-0 exit code or times out. However, these errors are printed, but I can't find a single "Process timeout" or "Process failed" in any log file. And as far as I can tell, there's no other place that `eject`, and thus transitively, `logs_on_error` are called.

help

PR #3681, opened 2025-11-20
refactor(iroh): Spawn `RemoteStateActor` in `JoinSet` and remove `guarded_channel.rs`

matheus23:
## Description

Depends on https://github.com/n0-computer/n0-future/pull/23 for `JoinSet::poll_join_next` and `JoinError::try_into_panic`.

- Removes `guarded_channel.rs`
- Instead, uses normal mpsc channels, and once it closes them, collects all messages that were missed before the channel was closed, sends them back to the `RemoteMap`, which will immediately spawn a new `RemoteStateActor` in case it missed some messages during shutdown.
- Spawns the `RemoteStateActor` in a `JoinSet` that's held in a `Mutex` inside the `RemoteMap` (effectively owned by the magic socket).

## Notes & open questions

I think it's good to look at this right now, but some considerations/TODOs:
- I don't like the mutex around the JoinSet in `RemoteMap`. Maybe only the `magicsock::Actor` should own the `JoinSet`, and the `RemoteMap` shouldn't actually store it itself.
- I'm not sure if draining all "initial messages" in `RemoteStateActor::run` is the right way to do this.

## Change checklist

- [x] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1160,7 +1160,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -1739,7 +1739,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.6.1",
+ "socket2 0.5.10",
  "tokio",
  "tower-service",
  "tracing",
@@ -2166,7 +2166,7 @@ dependencies = [
  "pin-project-lite",
  "rustc-hash",
  "rustls",
- "socket2 0.6.1",
+ "socket2 0.5.10",
  "thiserror 2.0.17",
  "tokio",
  "tokio-stream",
@@ -2208,9 +2208,9 @@ source = "git+https://github.com/n0-computer/quinn?branch=main#7ce0396198789ff7b
 dependencies = [
  "cfg_aliases",
  "libc",
- "socket2 0.6.1",
+ "socket2 0.5.10",
  "tracing",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -2780,7 +2780,7 @@ version = "0.50.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7957b9740744892f114936ab4a57b3f487491bbeafaf8083688b16841a4240e5"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -3315,7 +3315,7 @@ dependencies = [
  "quinn-udp",
  "rustc-hash",
  "rustls",
- "socket2 0.6.1",
+ "socket2 0.5.10",
  "thiserror 2.0.17",
  "tokio",
  "tracing",
@@ -3352,9 +3352,9 @@ dependencies = [
  "cfg_aliases",
  "libc",
  "once_cell",
- "socket2 0.6.1",
+ "socket2 0.5.10",
  "tracing",
- "windows-sys 0.60.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -3624,7 +3624,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


matheus23:
This PR has the problem that we only run the `remove_closed_remote_state_actors` fn every 60s.

Also: The `start_remote_actor` fn is called from two places and might become racy.

flub:
This would still be nice to have, please don't forget about it.

matheus23:
> Also: The `start_remote_actor` fn is called from two places and might become racy.

This is actually not possible. Calls to `start_remote_state_actor` are always behind an acquisition of the `actor_senders` lock in `remote_state.rs`.

matheus23:
> This PR has the problem that we only run the `remove_closed_remote_state_actors` fn every 60s.

Got around this problem by calling `JoinSet::poll_join_next` right after acquiring the lock for the join set, and calling that in the main magicsocket actor.

---

image

Yeah - fair point clippy.
If people have ideas for a struct or type alias encapsulating this, I'm all ears :sweat_smile:

iroh/src/magicsock/transports.rs
```diff
@@ -696,15 +696,14 @@ impl quinn::UdpSender for MagicSender {
                         "oops, flub didn't think this would happen");
                 }
 
-                let sender = self.msock.remote_map.remote_state_actor(node_id);
```
Frando, line 699: This was starting the RemoteStateActor if it's not running, whereas the new try_send only sends to running actors. Is this change intended?

Edit: Right, I think it's fine because sending to a just-spawned actor is a noop anyways.

iroh/src/magicsock/remote_map.rs
```diff
@@ -62,6 +61,23 @@ pub(crate) struct RemoteMap {
     local_direct_addrs: n0_watcher::Direct<BTreeSet<DirectAddr>>,
     discovery: ConcurrentDiscovery,
     shutdown_token: CancellationToken,
+    /// The state kept for spawning new tasks for remote state actors and cleaning them up.
+    ///
+    /// The lock for this needs to be acquired *after* the lock for the `actor_senders` is acquired.
+    actor_tasks: Mutex<Tasks>,
```
Frando: Why not put the senders and the tasks in the same mutex, to avoid the deadlock risk?

iroh/src/magicsock/transports.rs
matheus23: Yeah that was intended. I don't think we need this to start new relay actors.

To be fair, I could revert this change - I had some idea where I thought I could have only a single task own the `JoinSet` and it would do all the spawning - turns out that's not quite possible.

I think I'll revert.

iroh/src/magicsock/remote_map.rs
matheus23: Good question, I kept thinking about that as well. There's technically a place where you only acquire a lock to the senders map, but not the tasks. That might not be enough reason for it though.

I think this is just a footgun waiting to go off, let me put all the state into the same mutex. Makes me sleep better at night.

iroh/src/magicsock.rs
```diff
@@ -1207,9 +1205,8 @@ impl Actor {
                     self.msock.metrics.magicsock.actor_link_change.inc();
                     self.handle_network_change(is_major).await;
                 }
-                _ = remote_map_gc.tick() => {
-                    trace!("tick: gc map");
-                    self.msock.remote_map.remove_closed_remote_state_actors();
+                eid = poll_fn(|cx| self.msock.remote_map.poll_cleanup(cx)) => {
```
dignifiedquire, line 1208: this poll function takes a lock on the whole `remote_map`, isn't that racy against all other actions, if we continually do this?

Frando: Good point, I had a similar thought. We could just not recreate the future on each loop iteration and instead have a loop within the future, then we only lock once when a task completes or a new remote state actor is spawned.

Implemented it here: https://github.com/n0-computer/iroh/pull/3816

cc @matheus23

matheus23: It doesn't lock the *whole* `remote_map`, just the parts that access the remote state actors.
The lock is also only held very briefly during the call to `poll`, and dropped inside the future's poll before it continues. But that's by design - I don't want to block other reads or writes to the remote map just because of this poll. So I don't understand how that'd cause a race?

@Frando I don't think that's going to change much, besides not recreating the `poll_fn` future and never actually resolving the future, which I guess is fine? I don't understand what that's improving. Do we generally prefer looping this tokio::select less often? It's not that recreating the poll_fn is costly at all - you can throw away and recreate them as much as you like, I don't even think there's allocations involved with that?

Frando: > I don't understand what that's improving. 

The current version recreates the future (via poll_fn) on each magicsock actor loop tick.  This means it locks the remote map actor state once per magicsock actor loop tick. My version instead only locks when tasks are ready or new tasks are spawned.
I don't know how hot the magicsock actor is looping these days, but this might still save quite a few locks (ie. whenever another arm in the magicsock actor tokio::select! completes).

PR #3688, opened 2025-11-20
feat(iroh): introduce EndpointHooks

Frando:
## Description

This adds the conecpt of hooks to the iroh endpoint. `Hooks` are structs implementing the `EndpointHooks` trait and are used to intercept the establishment of connections. Multiple hooks can be added to the endpoint, and they will be invoked for each hook in the order they have been added to the endpoint.

Currently there's two methods on the `EndpointHooks` trait:

* `before_connect` is invoked before an outgoing connection is started.
* `after_handshake` is invoked for incoming and outgoing connections once the TLS handshake has completed

Both methods return an `Outcome`, which can either be `Reject` or `Accept`. If any hook returns `Reject`, the connection or connection attempt will be rejected.

The PR also adds `ConnectionInfo`, which is a struct that has information about a connection, but does not keep the connection itself alive. It allows to inspect stats and paths, and there's a `closed` method that returns a future which completes once the connection closes (without keeping the connection alive).

The PR includes two examples:
* `auth-hook` implements authentication for iroh protocols through a middleware and a separate authentication protocol. Individual protocols don't need to be aware of authentication at all.
* `monitor-connnections` monitors incoming and outgoing connections and prints connection stats once a connection closes.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/endpoint/connection.rs
+++ b/iroh/src/endpoint/connection.rs
@@ -244,8 +244,7 @@ fn conn_from_quinn_conn(
         .register_connection(info.endpoint_id, conn.weak_handle());
 
     // Check hooks
-    // TODO: avoid endpoint clone?
-    let ep = ep.clone();
+    let msock = ep.msock.clone();
     Ok(async move {
         let paths = fut.await?;
         let conn = Connection {
@@ -254,7 +253,7 @@ fn conn_from_quinn_conn(
         };
 
         if let AfterHandshakeOutcome::Reject { error_code, reason } =
-            ep.msock.hooks.after_handshake(&conn.to_info()).await
+            msock.hooks.after_handshake(&conn.to_info()).await
         {
             conn.close(error_code, &reason);
             return Err(e!(ConnectingError::LocallyRejected));


rklaehn:
I don't think the name middleware is appropriate here. I would associate middleware with something that I can use to actually transform incoming requests and outgoing responses, more in line with what I am doing for compression in iroh-blobs.

These are connection lifecycle hooks. Not sure what a short name is. Will this always be just about connections, or is it possible that we will add endpoint hooks to this later as well?

iroh/src/endpoint.rs
```diff
@@ -453,6 +460,21 @@ impl Builder {
...
+    /// Install hooks onto the endpoint.
+    ///
+    /// Endpoint hooks intercept the connection establishment process of an [`Endpoint`].
+    ///
+    /// You can install multiple hooks by calling this function multiple times.
+    /// Order matters: Endpoint hooks are invoked in the order they were installed onto the endpoint
+    /// builder. Once a hook returns reject, further processing
+    /// is aborted and other hooks won't be invoked.
+    ///
+    /// See [`EndpointHooks`] for details on the possible interception points in the connection lifecycle.
```
ramfox, line 475:
```suggestion

    /// Install hooks onto the endpoint.
    ///
    /// Endpoint hooks intercept the connection establishment process of an [`Endpoint`].
    ///
    /// You can install multiple [`EndpointHooks`] by calling this function multiple times.
    /// Order matters: hooks are invoked in the order they were installed onto the endpoint
    /// builder. Once a hook returns reject, further processing
    /// is aborted and other hooks won't be invoked.
    ///
    /// See [`EndpointHooks`] for details on the possible interception points in the connection lifecycle.
```

iroh/src/endpoint/connection.rs
```diff
@@ -241,12 +242,25 @@ fn conn_from_quinn_conn(
     let fut = ep
         .msock
         .register_connection(info.endpoint_id, conn.weak_handle());
+
+    // Check hooks
+    // TODO: avoid endpoint clone?
```
ramfox: Does this TODO need to be addressed?

dignifiedquire: resolved

rklaehn:
I like EndpointHooks much better!

PR #3691, opened 2025-11-21
feat(iroh): add ability to publish and resolve relay urls in mDNS

ramfox:
## Description

Adds the ability to resolve and publish relay URLs over mDNS. This is off by default.

RelayUrls are typically only used for hole punching, which likely is not needed when using mDNS. However, we have seen some test cases where mDNS packets get through local firewalls, but our iroh packets are blocked. In these cases, it may be prudent to also publish your relay url.

- Adds `MdnsDiscoveryBuilder::publish_relay_url` method

### DIFF
--- a/iroh/src/discovery/mdns.rs
+++ b/iroh/src/discovery/mdns.rs
@@ -77,7 +77,7 @@ use crate::{
     discovery::{Discovery, DiscoveryItem, EndpointData, EndpointInfo},
 };
 
-/// The n0 local service name
+/// The n0 local service name.
 const N0_SERVICE_NAME: &str = "irohv1";
 
 /// Name of this discovery service.
@@ -95,10 +95,10 @@ const USER_DATA_ATTRIBUTE: &str = "user-data";
 /// the TXT record supported by swarm-discovery.
 const RELAY_URL_ATTRIBUTE: &str = "relay";
 
-/// How long we will wait before we stop sending discovery items
+/// How long we will wait before we stop sending discovery items.
 const DISCOVERY_DURATION: Duration = Duration::from_secs(10);
 
-/// Discovery using `swarm-discovery`, a variation on mdns
+/// Discovery using `swarm-discovery`, a variation on mdns.
 #[derive(Debug, Clone)]
 pub struct MdnsDiscovery {
     #[allow(dead_code)]
@@ -129,7 +129,7 @@ impl Subscribers {
         Self(vec![])
     }
 
-    /// Add the subscriber to the list of subscribers
+    /// Add the subscriber to the list of subscribers.
     fn push(&mut self, subscriber: mpsc::Sender<DiscoveryEvent>) {
         self.0.push(subscriber);
     }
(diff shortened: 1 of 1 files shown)


ramfox:
## TODOS

- [ ] Add API to mDNS discovery to add a mapping function that let's users filter the kinds of addrs they want to publish, as well as order them. (takes a transport::Addr slice returns a transport::Addr vec?)
- [ ] adjust mDNS discovery publishing to map the addrs before publishing
- [ ] use match on addr to know how to format -> if IP do normal, if relay add to TXT attribute, only allow 4 txt attributes, any that don't get added log a warn with the addr
- [ ] add test that ensures we can get multiple relays
- [ ] add test that ensures we cannot add too many
- [ ] add tests that shows we respect the map filter

iroh/src/discovery/mdns.rs
```diff
@@ -169,6 +186,25 @@ impl MdnsDiscoveryBuilder {
         self
     }
 
+    /// Sets whether this endpoint should advertise its [`RelayUrl`].
+    ///
+    /// Overridden by the `MdnsDiscoveryBuilder::advertise` method. If this is
```
flub, line 191: "Overridden" threw me off, I think it's the wrong language.  Maybe something longer like "Only if mDNS advertising is enabled using [`Self::advertise`] will relay URLs be advertised."  But even that's not great.

Maybe the whole thing can be described differently:
```
/// Enables advertising this endpoint's [`RelayUrl`] in addition to IP addresses.
///
/// Enabling [`Self::advertise`] will only advertise IP addresses, enabling this will additionally advertise the endpoint's [`RelayUrl`]. Note that if [`Self::advertise`] is disabled no advertising happens at all, neither of IP addresses or relay URLs.
```

Or maybe something better even :)

Also, I think it's worth updating the doc comment of `advertise` itself as well to make clear it only advertises IP addresses but is also a master switch.

iroh/src/discovery/mdns.rs
```diff
@@ -114,7 +129,7 @@ impl Subscribers {
         Self(vec![])
     }
 
-    /// Add the subscriber to the list of subscribers
+    /// Add the subscriber to the list of subscribers.
```
flub, line 132: 😂

iroh/src/discovery/mdns.rs
```diff
@@ -283,9 +325,16 @@ impl MdnsDiscovery {
                         for addr in addrs {
                             discovery.add(addr.0, addr.1)
                         }
+                        if publish_relay_url {
+                            if let Some(relay) = data.relay_urls().next() {
+                                if let Err(err) = discovery.set_txt_attribute(RELAY_URL_ATTRIBUTE.to_string(), Some(relay.to_string())) {
+                                    warn!("Failed to set the relay url in mDNS: {err:?}");
```
flub, line 331: This can happen if the RelayUrl is too long right? I wonder if this could get very noisy. But maybe that's the point.

flub (approved):

ramfox:
A note on this PR.

In prep for the ability to add custom transports, we want to re-work this PR to be a model for how to add those in the future. 

We want to:
- add a map fn to the discovery service that lets you filter/order/map the addrs you want published for that particular service
- when you go to publish, run the map fn, add IP addrs as normal to mDNS
- any other addrs, add as a TXT attribute, going in the order specified by the user. Any that don't fit will get warn logged

ramfox:
closed in favor of #3960

PR #3692, opened 2025-11-22
feat(iroh)!: allow multiple IP transports, including filtering by interface

### LINKED ISSUE #3218: allow binding to an interface instead of an IP address
`Magicsock::listen` currently supports either listening on a specific ipv4 and/or ipv6 address.
However, binding to 0.0.0.0, either via `None` or passing that explicitly, will cause it to bind across multiple interfaces.

I have a use-case where I want to force Iroh connections through one or more specific network interface(s).
My only option right now is to first query those interface(s) for their IP address(es), and then pass *one* of them to Magicsock.

Therefore, I currently cannot bind on *two* interfaces (both v4 or v6) while avoiding a third.

I would love the ability to bind to an arbitrary list of interfaces, each on v4, v6, or both.

dignifiedquire:
## Description

This allows for multiple interfaces to be bound, and be actually used. 

You can now use this by passing

```rust

let endpoint = Endpoint::builder()
    .bind_addr("127.0.0.1:1234")?
    .bind_addr_with_opts("192.168.1.2:1234", BindOpts::default().set_prefix_len(24))?
    .bind()
    .await?
```

The selection of the interface is done internally by first looking at all specific bindings, and then fallbing back to the `default` version for this family.

Closes #3218 

## Breaking Changes

- `iroh`
  - removed
    - `Endpoint::bind_addr_v4`
    - `Endpoint::bind_addr_v6`
  - added
    - `Endpoint::bind_addr`
    - `Endpoint::bind_addr_with_opts`
    - `endpoint::BindOpts`
    - `endpoint::ToSocketAddr`

### DIFF
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -227,21 +227,46 @@ impl Builder {
 
     // # The very common methods everyone basically needs.
 
-    /// Adds a default IP transport, binding to the provided address.
+    /// Binds an IP socket at the provided socket address.
     ///
-    /// Setting the port to `0` will use a random port.
-    /// If the port specified is already in use, it will fallback to choosing a random port.
+    /// This is an advanced API to tightly control the sockets used by the endpoint. Most
+    /// uses do not need to explicitly bind sockets.
     ///
-    /// Only a single interface can be the default, per IP family, so this errors out if one is already set.
+    /// # Warning
     ///
-    /// If you want to remove the default transports, make sure to call `clear_ip` first.
+    /// - The builder always comes pre-configured with an IPv4 socket to be bound on the
+    ///   *unspecified* address: `0.0.0.0`. This is the equivalent of using `INADDR_ANY`
+    ///   special bind address and results in a socket listening on *all* interfaces
+    ///   available.
+    ///
+    /// - Likewise the builder always comes pre-configured with an IPv6 socket to be bound
+    ///   on the *unspecified* address: `[::]`. This bind is allowed to fail however.
+    ///
+    /// - Most likely when using this API those existing binds need to be removed explicitly
+    ///   using [`Self::clear_ip_transports`].
+    ///
+    /// - This should be called at most once for each address family: once for IPv4 and/or
+    ///   once for IPv6. Calling it multiple times for the same address family will result
+    ///   in undefined routing behaviour. To bind multiple sockets of the same address
+    ///   family, use [`Self::bind_addr_with_opts`].
+    ///
+    /// # Description
+    ///
+    /// Requests a socket to be bound on a specific address, with an implied netmask of
+    /// `/0`. This allows restricting binding to only one network interface for a given
+    /// address family.
+    ///
+    /// If the port specified is already in use, a random free port will be chosen. Using
+    /// port `0` in the socket address assigns a random free port.
     ///
     /// # Example
+    ///
     /// ```
     /// # #[tokio::main]
     /// # async fn main() -> n0_error::Result<()> {
     /// # use iroh::Endpoint;
     /// let endpoint = Endpoint::builder()
+    ///     .clear_ip_transports()
     ///     .bind_addr("127.0.0.1:1234")?
     ///     .bind_addr("[::1]:1235")?
     ///     .bind()
@@ -257,9 +282,56 @@ impl Builder {
         self.bind_addr_with_opts(addr, BindOpts::default())
     }
 
-    /// Binds an IP socket
+    /// Binds an IP socket at the provided socket address.
     ///
-    /// This is always in addition to any prior bindings, so if you want to remove other transports, make sure to call `clear_ip` first.
+    /// This is an advanced API to tightly control the sockets used by the endpoint. Most
+    /// uses do not need to explicitly bind sockets.
+    ///
+    /// # Warning
+    ///
+    /// - The builder always comes pre-configured with an IPv4 socket to be bound on the
+    ///   *unspecified* address: `0.0.0.0`. This is the equivalent of using `INADDR_ANY`
+    ///   special bind address and results in a socket listening on *all* interfaces
+    ///   available.
+    ///
+    /// - Likewise the builder always comes pre-configured with an IPv6 socket to be bound
+    ///   on the *unspecified* address: `[::]`. This bind is allowed to fail however.
+    ///
+    /// - Most likely when using this API those existing binds need to be removed explicitly
+    ///   using [`Self::clear_ip_transports`].
+    ///
+    /// # Description
+    ///
+    /// Requests a socket to be bound on a specific address. This allows restricting binding
+    /// to only one network interface for a given address family.
+    ///
+    /// [`BindOpts::set_prefix_len`] **should** be used to configure the netmask of the
+    /// network interface. This allows outgoing datagrams that start a new network flow to
+    /// be sent over the socket which is attached to the subnet of the destination
+    /// address. If multiple sockets are bound the standard routing-table semantics are
+    /// used: the socket attached to the subnet with the longest prefix matching the
+    /// destination is used. Practically this means the smallest subnets are at the top of
+    /// the routing table, and the first subnet containing the destination address is
+    /// chosen.
+    ///
+    /// If no socket is bound to a subnet that contains the destination address, the notion
+    /// of "default route" is used. At most one socket per address family may be marked as
+    /// the default route using [`BindOpts::set_is_default_route`], and this will be used
+    /// for destinations not contained by the subnets of the bound sockets. This network is
+    /// expected to have a default gateway configured.
+    ///
+    /// Be aware that using a subnet with a prefix length of `/0` will always contain all
+    /// destination addresses. It is valid to configure this, but no more than one such
+    /// socket should be bound or the routing will be non-deterministic. Prefer using
+    /// [`BindOpts::set_is_default_route`] on one of the sockets instead.
+    ///
+    /// Finally note that most outgoing datagrams are part of an existing network flow. That
+    /// is, they are in response to an incoming datagram. In this case the outgoing datagram
+    /// will be sent over the same socket as the incoming datagram was received on, and the
+    /// routing with the prefix length and default route as described above does not apply.
+    ///
+    /// If the port specified is already in use, a random free port will be chosen. Using
+    /// port `0` in the socket address assigns a random free port.
     ///
     /// # Example
     /// ```
@@ -267,6 +339,7 @@ impl Builder {
     /// # async fn main() -> n0_error::Result<()> {
     /// # use iroh::{Endpoint, endpoint::BindOpts};
     /// let endpoint = Endpoint::builder()
+    ///     .clear_ip_transports()
     ///     .bind_addr_with_opts("127.0.0.1:1234", BindOpts::default().set_prefix_len(24))?
     ///     .bind_addr_with_opts("[::1]:1235", BindOpts::default().set_prefix_len(48))?
     ///     .bind()
@@ -328,7 +401,7 @@ impl Builder {
         Ok(self)
     }
 
-    /// Removes all IP based transports
+    /// Removes all IP based transports.
     #[cfg(not(wasm_browser))]
     pub fn clear_ip_transports(mut self) -> Self {
         self.transports
@@ -336,7 +409,7 @@ impl Builder {
         self
     }
 
-    /// Removes all relay based transports
+    /// Removes all relay based transports.
     pub fn clear_relay_transports(mut self) -> Self {
         self.transports
             .retain(|t| !matches!(t, TransportConfig::Relay { .. }));
--- a/iroh/src/endpoint/bind.rs
+++ b/iroh/src/endpoint/bind.rs
@@ -10,11 +10,20 @@ use n0_error::stack_error;
 pub struct BindOpts {
     /// Sets the network prefix length of the subnet for this interface.
     ///
-    /// The prefix length is used in the routing table that is built to decide where datagrams without a specific source address are sent. If these number of leading bits in the destination IP address match the same number of leading bits in this bound socket address, then it matches the subnet and the datagram will be sent here. Otherwise the next bound sockets will be checked for a subnet match. Sockets are ordered from longest prefix to shortest prefix.
+    /// The prefix length is used in the routing table that is built to decide where
+    /// datagrams without a specific source address are sent. If these number of leading
+    /// bits in the destination IP address match the same number of leading bits in this
+    /// bound socket address, then it matches the subnet and the datagram will be sent
+    /// here. Otherwise the next bound sockets will be checked for a subnet match. Sockets
+    /// are ordered from longest prefix to shortest prefix.
     ///
-    /// If no bound address has a matching subnet, the bind marked with [`Self::is_default_route`] will be used.
+    /// If no bound address has a matching subnet, the bind marked with
+    /// [`Self::is_default_route`] will be used.
     ///
-    /// Note that most datagrams belonging to a traffic flow are in response to an incoming datagram. Those are usually sent on the same bound socket as they were received and will not consult the routing table derived from these bound sockets to select the socket on which they will be sent.
+    /// Note that most datagrams belonging to a traffic flow are in response to an incoming
+    /// datagram. Those are usually sent on the same bound socket as they were received and
+    /// will not consult the routing table derived from these bound sockets to select the
+    /// socket on which they will be sent.
     prefix_len: u8,
     /// If set, binding this interface is required and any errors will abort the
     /// initialization of the endpoint.
@@ -23,7 +32,9 @@ pub struct BindOpts {
     is_required: bool,
     /// Whether this socket should be used as default route.
     ///
-    /// The default route is used for outgoing datagrams not belonging to an existing traffic flow, which does not fit in any subnet of the bound sockets. It is assumed this subnet has a gateway router to route such packets.
+    /// The default route is used for outgoing datagrams not belonging to an existing
+    /// traffic flow, which does not fit in any subnet of the bound sockets. It is assumed
+    /// this subnet has a gateway router to route such packets.
     ///
     /// See [`Self::prefix_len`] for details of how such routing works.
     is_default_route: bool,
@@ -43,9 +54,17 @@ impl BindOpts {
     /// Sets the network prefix length of the subnet this interface is in.
     ///
     ///
-    /// The subnets of bound sockets are used to route outgoing datagrams not belonging to an existing traffic flow to the socket they should be sent on. Subnets are ordered from longest prefix length to shortest prefix length and the first subnet which contains the destination IP address will be chosen. If no subnet matches but there is a bound socket marked with [`Self::set_is_default_route`] then this socket will be used. In this case it is assumed the attached subnet has a gateway router to forward the datagram.
+    /// The subnets of bound sockets are used to route outgoing datagrams not belonging to
+    /// an existing traffic flow to the socket they should be sent on. Subnets are ordered
+    /// from longest prefix length to shortest prefix length and the first subnet which
+    /// contains the destination IP address will be chosen. If no subnet matches but there
+    /// is a bound socket marked with [`Self::set_is_default_route`] then this socket will
+    /// be used. In this case it is assumed the attached subnet has a gateway router to
+    /// forward the datagram.
     ///
-    /// Defaults to `0`, which means *all* IP addresses will belong to the subnet of this socket's address. If multiple sockets of the same address family (IPv4 or IPv6) are bound with such a `/0` prefix the socket which will be chosen is undefined.
+    /// Defaults to `0`, which means *all* IP addresses will belong to the subnet of this
+    /// socket's address. If multiple sockets of the same address family (IPv4 or IPv6) are
+    /// bound with such a `/0` prefix the socket which will be chosen is undefined.
     ///
     /// For IPv4 sockets the maximum prefix is `32`. For IPv6 the maximum prefix is `128`.
     pub fn set_prefix_len(mut self, prefix_len: u8) -> Self {
@@ -60,7 +79,8 @@ impl BindOpts {
 
     /// Sets whether bind errors are fatal for this socket.
     ///
-    /// If `false` and this socket fails to bind, the error will be silently ignored and the endpoint will still be created.
+    /// If `false` and this socket fails to bind, the error will be silently ignored and the
+    /// endpoint will still be created.
     ///
     /// Defaults to `true`.
     pub fn set_is_required(mut self, is_required: bool) -> Self {
@@ -75,7 +95,9 @@ impl BindOpts {
 
     /// Sets whether this is a default route.
     ///
-    /// The default route is used for outgoing datagrams not belonging to an existing traffic flow, which does not fit in any subnet of the bound sockets. It is assumed this subnet has a gateway router to route such packets.
+    /// The default route is used for outgoing datagrams not belonging to an existing
+    /// traffic flow, which does not fit in any subnet of the bound sockets. It is assumed
+    /// this subnet has a gateway router to route such packets.
     ///
     /// See [`Self::set_prefix_len`] for details on how this routing works.
     ///


Frando:
We could remove the `_default` API I think, if you set a netmask of 0 it will apply to everything. And instead return an error if you bind conflicting networks.

dignifiedquire:
> We could remove the `_default` API I think, if you set a netmask of 0 it will apply to everything. And instead return an error if you bind conflicting networks.

well currently they are treated differently internally as an explicit fallback, thats why the difference in configuration exists

iroh/src/magicsock/transports.rs
```diff
@@ -501,9 +485,9 @@ impl TransportsSender {
                 return Poll::Ready(Err(io::Error::other("IP is unsupported in browser")));
             }
             #[cfg(not(wasm_browser))]
-            Addr::Ip(addr) => {
-                for sender in &mut self.ip {
-                    if sender.is_valid_send_addr(addr) {
+            Addr::Ip(addr) => match addr {
+                SocketAddr::V4(_) => {
+                    for sender in self.ip.v4_iter_mut().filter(|s| s.is_valid_send_addr(addr)) {
```
matheus23: It probably makes sense to pass in `src: Option<IpAddr>` now that we populate that value.
After all, is the value that should first and foremost tell you which interface to use.

matheus23 (commented):
I'm not super convinced this is the right API :thinking: 

Looking at your example:
```rs

let endpoint = Endpoint::builder()
  .bind_addr_v4_default("127.0.0.1", 1234)
  .bind_addr_v4("192.168.1.2/24".parse()?, 1234) // bind to prefix_len of 24
  .bind()
  .await?
```

It looks like the configuration for an endpoint that will only send traffic in the local network and localhost.
So let's say quinn produces a datagram that's meant to be sent to `192.168.1.101:1234`.
Iroh will then first check this against your `192.168.1.2/24` and find a match. However, it turns out that interface is actually busy, so it continues.
Then it'll see the "default" `127.0.0.1` interface and try to send on that. But obviously it can't send outside localhost, so ideally that should just fail sending, but I think it won't? I'm not actually sure what will happen.

Also another thing about the default socket: What if someone wants to bind a dual-stack socket with `[::]`? I think that should match against an IPv4 address, but IIUC currently it doesn't?

Minor note: I'm not sure what should happen if you bind multiple sockets to the same port, but different IP addrs. What happens in these cases? I'm not sure.

iroh/src/magicsock/transports/ip.rs
```diff
@@ -1,30 +1,132 @@
...
+                            return true;
+                        }
+                    }
+                    false
+                }
+                SocketAddr::V4(_) => false,
+            },
+        }
+    }
+}
```
matheus23: Yeah I agree with Frando here I think: I'm not sure if it's that useful to have the default cases here.
What's the use case for having these be a special case? Make sure they're tried last?

dignifiedquire:
> What if someone wants to bind a dual-stack socket

We don't support dual socket, not currently, and not in this setup

Frando:
So, if we removed the default, we could have this:

```rust
let endpoint = Endpoint::builder()
  .bind_addr_v4("192.168.1.2/0".parse()?, 0) // bind to prefix_len of 0
  .bind_addr_v4("10.0.0.3/24".parse()?, 0) // bind to prefix_len of 24
  .bind_addr_v6("[::]/0".parse()?, 0)
  .bind()
  .await?
```
Which would mean:
- Send packets with destination `10.0.0.*` via interface `10.0.0.3`
- Send packets with destination `*.*.*.*` via interface `192.168.1.2`
- Send all IPv6 packets via the wildcard bind

Which is, I think, the common thing people would want from this API. I.e. have a default interface (netmask prefix len 0) and a more specific interface that only handles destinations in the interface's subnet.

But what would happen if a user sets up binds, but not one with prefix len 0? Then we would have to error for all sends that are not matched by any of the netmasks of the bound addrs.

Not sure if this is preferable to the the `_default` API, but it's one public API less.

iroh/src/magicsock/transports/ip.rs
```diff
@@ -77,32 +77,64 @@ impl Config {
...
+                    return false;
+                }
+                Self::V4 { ip_addr, .. } => {
+                    if let IpAddr::V4(src) = src {
+                        return ip_addr.addr().is_unspecified() || ip_addr.addr() == src;
+                    }
+                    return false;
+                }
+                Self::V6 {
+                    ip_addr, scope_id, ..
```
matheus23: I think this is just an unused variable warning that makes CI barf (`scope_id` is unused)

iroh/src/magicsock/transports/ip.rs
```diff
@@ -77,32 +77,64 @@ impl Config {
...
-            Self::V6 {
-                ip_addr, scope_id, ..
-            } => match dst {
-                SocketAddr::V6(dst_v6) => {
-                    if ip_addr.contains(dst_v6.ip()) {
-                        return true;
+    /// Does this configuration match to send to the given `src` and `dst` address.
+    pub(crate) fn is_valid_send_addr(&self, src: Option<IpAddr>, dst: SocketAddr) -> bool {
+        if let Some(src) = src {
+            match self {
```
matheus23: Thoughts on code style: Perhaps matching on `(src, self)` is potentially slightly nicer?

matheus23:
> Then it'll see the "default" `127.0.0.1` interface and try to send on that. But obviously it can't send outside localhost, so ideally that should just fail sending, but I think it won't? I'm not actually sure what will happen.

I think the example you give in the PR description is still weird. You wouldn't want to use a `127.0.0.1` address as the default address to bind to.
I'm not sure if you'd want to be able to have *no* default address? But I guess in that example, I'd say what you actually want is no default address and this configuration instead:
```rs
let endpoint = Endpoint::builder()
  .bind_addr_v4("127.0.0.1/8", 1234)
  .bind_addr_v4("192.168.1.2/24".parse()?, 1234) // bind to prefix_len of 24
  .bind()
  .await?
```
I.e. that'd be a configuration that only allows sending on localhost or LAN.

That said, one should be aware that this still *in theory* allows the endpoint to receive & process traffic from outside those netmasks.

iroh/src/magicsock/transports/ip.rs
matheus23: We figured out having a default is useful, this part of the code is essentially implementing a routing table - and those have a default route for good reason: You might have two interfaces bound that are /0, but you want to use one of those interfaces as the default choice for e.g. making new outgoing connections.

iroh/src/magicsock/transports/ip.rs
```diff
@@ -1,31 +1,162 @@
...
+
+    /// Does this configuration match to send to the given `src` and `dst` address.
+    pub(crate) fn is_valid_send_addr(&self, src: Option<IpAddr>, dst: SocketAddr) -> bool {
+        if let Some(src) = src {
+            match self {
+                Self::V4Default { ip_addr, .. } => {
+                    if let IpAddr::V4(src) = src {
+                        return ip_addr.is_unspecified() || *ip_addr == src;
+                    }
+                    false
```
matheus23: I think these comparisons can be written terser, and I think that might help readability:
```suggestion
                    (src.is_ipv4() && ip_addr.is_unspecified()) || IpAddr::V4(*ip_addr) == src
```

iroh/src/magicsock/transports/ip.rs
```diff
@@ -165,25 +296,15 @@ impl IpNetworkChangeSender {
...
     sender: UdpSender,
     metrics: Arc<MagicsockMetrics>,
 }
 
 impl IpSender {
-    pub(super) fn is_valid_send_addr(&self, dst: &SocketAddr) -> bool {
-        // Our net-tools crate binds sockets to their specific family.  This means an IPv6
-        // socket can not sent to IPv4, on any platform.  So we need to convert and
-        // IPv4-mapped IPv6 address back to it's canonical IPv4 address.
-        let dst_ip = dst.ip().to_canonical();
```
matheus23, line 179: Ah okay, I see this call moved to `poll_send` in transports.rs (for anyone else wondering).

iroh/src/magicsock/transports/ip.rs
```diff
@@ -232,3 +353,146 @@ impl IpSender {
...
+                            ip_v6.push(transport);
+                        }
+                    }
+                }
+                Err(err) => {
+                    if config.is_ipv6() {
+                        tracing::info!("bind ignoring IPv6 bind failure: {:?}", err);
+                    } else {
+                        return Err(err);
+                    }
```
matheus23: I'd appreciate if we stored something like a `config.is_mandatory()` boolean that is set to true for anything that's added by `bind_addr_with_mask` (or maybe specified as an option there).
And by default we would set the default IPv4 socket to mandatory, while the IPv6 socket is not set to mandatory.

iroh/src/magicsock/transports.rs
```diff
@@ -699,7 +708,7 @@ impl quinn::UdpSender for MagicSender {
                 let sender = self.msock.remote_map.remote_state_actor(node_id);
                 let transmit = OwnedTransmit::from(quinn_transmit);
                 return match sender.try_send(RemoteStateMessage::SendDatagram(
-                    self.sender.clone(),
+                    Box::new(self.sender.clone()),
```
matheus23, line 766: Boxed for size I suspect (for anyone else wondering). Without the box, `RemoteStateMessage` is at least 192 bytes (`TransportsSender`) + 56 bytes (`OwnedTransmit)`. Without, all of `RemoteStateMessage` is just 64 bytes.

iroh/src/endpoint.rs
```diff
@@ -219,26 +357,73 @@ impl Builder {
     ///
     /// Setting the port to `0` will use a random port.
     /// If the port specified is already in use, it will fallback to choosing a random port.
+    ///
+    /// Only a single interface can be the default, so this will replace the existing default
```
Frando: Is this true? If I read the code correctly, there is one default *per address family*, not a single global default, right?

So this is valid:
```rust
Endpoint::builder()
    .bind_addr("192.168.0.10:0")
    .bind_addr("[::]:0")
    .bind()
    .await
```
Whereas here, it will not bind to "192.168.0.10" as this will be overwritten by the "10.0.0.22" one.
```rust
Endpoint::builder()
    .bind_addr("192.168.0.10:0")
    .bind_addr("10.0.0.22:0")
    .bind_addr("[::]:0")
    .bind()
    .await
```

dignifiedquire: yeah, it is per IP family

Frando: This might be a little confusing for users so I think the docs need to be very upfront that this may only be called once per IP family. It was a bit clearer with `bind_addr_v4` and `bind_addr_v6`  because there we can just say "call this function only once".
With this API I can see users doing
```rust
Endpoint::builder()
    .bind_addr("192.168.0.10:0")
    .bind_addr("10.0.0.22:0")
```
and being confused why it doesn't work as they expected.

iroh/src/magicsock/transports/ip.rs
dignifiedquire: I think it is at harder to read tbh, so I would prefer to keep my version

dignifiedquire:
> maybe put the scope ID there,

`scope_id` can already be passed it is part of the `SocketAddrV6`

iroh/src/endpoint.rs
dignifiedquire: this errors out now

iroh/src/magicsock/transports/ip.rs
dignifiedquire: done

iroh/src/endpoint.rs
```diff
@@ -213,32 +222,93 @@ impl Builder {
...
+        let addr = addr.to_socket_addr().map_err(Into::into)?;
+
+        match addr {
+            SocketAddr::V4(addr) => {
+                if self
+                    .transports
+                    .iter()
+                    .any(|t| matches!(t, TransportConfig::Ip(IpConfig::V4Default { .. })))
+                {
+                    bail!(InvalidSocketAddr::DuplicateDefaultAddr);
```
Frando, line 367: Will this not always fail now unless `clear_ip()` is called first, because we always set a default in `Builder::empty`?

dignifiedquire: yes

Frando: We could add a flag `user_provided` to `IpConfig::V6Default` and `IpConfig::V4Default`, and only fail if this is true in the current default (and set it to true when creating in Builder::empty but not when creating in here)

iroh/src/endpoint/bind.rs
```diff
@@ -0,0 +1,193 @@
...
+            prefix_len: 0,
+            is_required: true,
+        }
+    }
+}
+
+impl BindOpts {
+    /// Sets the network prefix length which manages route.
+    ///
+    /// Defaults to `0`.
```
Frando: What happens if you add multiple interfaces in the same family that have a prefix length of 0? Which one is used to send? Should be documented.

iroh/src/endpoint/bind.rs
```diff
@@ -0,0 +1,193 @@
...
+use n0_error::stack_error;
+
+/// Options when configuring binding an IP socket.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct BindOpts {
+    /// Sets the network prefix length which manages route.
+    ///
+    /// Defaults to `0`.
+    ///
+    /// Must not be larger than `32`
```
flub: This isn't used for IPv6 addresses?  There they go up to /64 or /128 even.

Maybe you don't even really need a limit, if the netmask is larger then the address it is the equivalent of the netmask of all ones after all.

iroh/src/magicsock/transports/ip.rs
```diff
@@ -1,31 +1,176 @@
...
     local_addr: Watchable<SocketAddr>,
     metrics: Arc<MagicsockMetrics>,
 }
 
-fn bind_with_fallback(mut addr: SocketAddr) -> io::Result<netwatch::UdpSocket> {
-    debug!(%addr, "binding");
+/// IP transport configuration
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum Config {
+    /// Default IPv4 binding
```
flub: Needs to elaborate what "default" means here. I assume it means that this will be the default route for initiated outgoing packets that don't match the subnet of any other bind.

I do think splitting `V4` anv `V4Default` here is logically a bit wrong. Because each bound socket bound on a specific local interface **does** have the subnet of that interface, regardless of whether it is used as default if no subnets match or not. So I'd rather have an `is_default` in the fields of `V4` & `V6`.

iroh/src/endpoint/bind.rs
```diff
@@ -0,0 +1,193 @@
+use std::{
+    convert::Infallible,
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6},
+};
+
+use n0_error::stack_error;
+
+/// Options when configuring binding an IP socket.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct BindOpts {
```
flub, line 10: Is this actually used in a `pub` place?

iroh/src/magicsock/transports/ip.rs
```diff
@@ -1,31 +1,176 @@
...
-    net::{IpAddr, SocketAddr},
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6},
     num::NonZeroUsize,
     pin::Pin,
     sync::Arc,
     task::{Context, Poll},
 };
 
 use n0_watcher::Watchable;
+use netdev::ipnet::{Ipv4Net, Ipv6Net};
```
flub, line 11: Seems `ipnet` is its own crate that is at version 2.11.0. Maybe we should directly depend on that given that this type is being publicly exposed on our API and netdev itself is not 1.0?

iroh/src/magicsock/transports/ip.rs
```diff
@@ -1,31 +1,176 @@
...
+        scope_id: u32,
+        /// The port to bind on
+        port: u16,
+        /// Is binding mandatory?
+        is_required: bool,
+    },
+    /// General IPv4 binding
+    V4 {
+        /// The IP address to bind on
+        ip_addr: Ipv4Net,
```
flub: nit, call this field `ip_net` or even `net`. This also shows up in the functions where you use this where it's suddenly confusing to read `ip_addr.add()` or `ip_addr.contains(an_actual_ip_addr)`.

iroh/src/magicsock/transports/ip.rs
```diff
@@ -48,13 +193,13 @@ fn bind_with_fallback(mut addr: SocketAddr) -> io::Result<netwatch::UdpSocket> {
 }
 
 impl IpTransport {
-    pub(crate) fn bind(bind_addr: SocketAddr, metrics: Arc<MagicsockMetrics>) -> io::Result<Self> {
-        let socket = bind_with_fallback(bind_addr)?;
-        Ok(Self::new(bind_addr, Arc::new(socket), metrics.clone()))
+    pub(crate) fn bind(config: Config, metrics: Arc<MagicsockMetrics>) -> io::Result<Self> {
+        let socket = bind_with_fallback(config.into())?;
```
flub, line 188: I find this very surprising that someone passed in a Config with a specific port but it will now potentially ignore this port. For folks who configure things accurately this is bad behaviour.

iroh/src/magicsock/transports/ip.rs
```diff
@@ -232,3 +367,145 @@ impl IpSender {
...
+            ($socket:expr) => {
+                match $socket.poll_recv(cx, bufs, metas, source_addrs)? {
+                    Poll::Pending | Poll::Ready(0) => {}
+                    Poll::Ready(n) => {
+                        return Poll::Ready(Ok(n));
+                    }
+                }
+            };
+        }
+
```
flub, line 501: IIUC if we are CPU bound and can not keep up we'll prefer v4_default, then v4, then v6_default and then v6. I guess that's fine for now, but could we leave a TODO or some comment that this bias exists?

iroh/src/endpoint.rs
dignifiedquire: done

iroh/src/magicsock/transports/ip.rs
```diff
@@ -232,3 +367,145 @@ impl IpSender {
...
+    /// Default sender for v4
+    v4_default: Option<IpSender>,
+    v4: Vec<IpSender>,
+    /// Default sender for v6
+    v6_default: Option<IpSender>,
+    v6: Vec<IpSender>,
+}
+
+impl IpTransportsSender {
+    pub(super) fn v4_iter_mut(&mut self) -> impl Iterator<Item = &mut IpSender> {
```
flub, line 377: IIRC these two `iter_mut` functions are on the hot path.

But I do think this violates the normal routing semantics: I think the v4 (and v6) vec needs to be sorted from longest prefix to shorted prefix. That is from most specific subnet to most general subnet. And you need to iterate in order. Otherwise you might end up with a destination that matches two networks but is sent to the bigger one rather than the more specific one and I believe this is contrary to normal routing behaviour.

I'm also unsure about having `v4_default` chained at the end. I think this matches with my comment on the Config enum itself. I think from a routing view you treat all those the same, regardless of whether it is marked as default, and they are sorted on prefix-len. Then the default one is appended to the end with a prefix len of 0, which means it will match anything that has not yet been matched.

iroh/src/endpoint/bind.rs
dignifiedquire: fixed already

iroh/src/magicsock/transports/ip.rs
dignifiedquire: > So I'd rather have an is_default in the fields of V4 & V6.

they store different information at this point, which is why they are different enum variants

iroh/src/endpoint/bind.rs
dignifiedquire: yes, in `bind_addr_with_opts`?

iroh/src/magicsock/transports/ip.rs
dignifiedquire: it's not part of the public API anymore, so I'd rather not add another dependency if we don't have to?

iroh/src/magicsock/transports/ip.rs
dignifiedquire: this has been the behaviour for a long time, so changing this should be a separate discussion imho

iroh/src/magicsock/transports/ip.rs
dignifiedquire: I explicitly made this, so that the order of configuration that is passed in, is preserved. What you are suggesting is a different sorting algorithm, which we can do, but I am not fully clear why it would be preferrable.

iroh/src/magicsock/transports.rs
```diff
@@ -131,7 +151,35 @@ impl Transports {
         shutdown_token: CancellationToken,
     ) -> io::Result<Self> {
         #[cfg(not(wasm_browser))]
-        let ip = bind_ip(configs, metrics)?;
+        let ip_configs = {
+            let mut ip_configs = Vec::new();
+
+            // user defined overrides defaults
```
flub, line 166: I don't like this. I would prefer to get rid of `is_user_defined` (which I think is only used here?).

This makes it harder to understand what is happening. It encourages users to forget calling `Builder::clear_ip` if they are adding a default route. But now it behaves different then when they did add a non-default route. I think this should be solved by improving the builder instead of having the `is_user_defined` fields, which spreads complexity to many places.

iroh/src/endpoint.rs
```diff
@@ -213,32 +227,128 @@ impl Builder {
...
+    ///     .await?;
+    /// # Ok(()) }
+    /// ```
     #[cfg(not(wasm_browser))]
-    pub fn bind_addr_v4(mut self, bind_addr: SocketAddrV4) -> Self {
-        self.transports.push(TransportConfig::Ip {
-            bind_addr: bind_addr.into(),
-        });
-        self
+    pub fn bind_addr<A>(mut self, addr: A) -> Result<Self, InvalidSocketAddr>
```
flub: I do see the appeal of having a `bind_addr` variant that takes a simple `SocketAddr` and no other args. But I think that it would be much more consistent if that would be the equivalent of `bind_addr_with_opts(addr, BindOpts::default())`.

iroh/src/endpoint.rs
flub: Oh no! Just noticed this comment. I strongly disagree with this as I said in another comment. This pushes complexity to many places and makes it much harder to reason about as a user what addresses are being bound.

iroh/src/endpoint.rs
```diff
@@ -213,32 +227,128 @@ impl Builder {
...
     #[cfg(not(wasm_browser))]
-    pub fn bind_addr_v6(mut self, bind_addr: SocketAddrV6) -> Self {
-        self.transports.push(TransportConfig::Ip {
-            bind_addr: bind_addr.into(),
-        });
-        self
+    pub fn bind_addr_with_opts<A>(
+        mut self,
+        addr: A,
+        opts: BindOpts,
```
flub, line 353: I think `BindOpts::set_default_route` is needed so that you can do the equivalent bind that `Builder::bind_addr` would do.

iroh/src/endpoint.rs
```diff
@@ -213,32 +227,128 @@ impl Builder {
...
-    ///
-    /// If you want to remove the default transports, make sure to call `clear_ip` first.
+    /// Adds a default IP transport, binding to the provided address.
     ///
     /// Setting the port to `0` will use a random port.
     /// If the port specified is already in use, it will fallback to choosing a random port.
+    ///
+    /// Only a single interface can be the default, per IP family, so this errors out if one is already set.
+    ///
+    /// If you want to remove the default transports, make sure to call `clear_ip` first.
```
flub: How empty is a `Buidler::empty_builder`? I think right now it still includes these default binds, which honestly is a bit confusing.

Also, having to call `clear_ip` is quite the footgun. I think it would be more logical from the user's view if we always created a builder **without** any pre-configured binds. And only if the user did not call any `bind_addr*` method before calling `Builder::bind` **then** add the IPv4 UNSPECIFIED and optional IPv6 UNSPECIFIED binds.

dignifiedquire: > And only if the user did not call any bind_addr* method before calling Builder::bind then add the IPv4 UNSPECIFIED and optional IPv6 UNSPECIFIED binds.

I tried that, that doesn't work, as we have a test and use case for not having any IP transport already

dignifiedquire: > Also, having to call clear_ip is quite the footgun

we have it for other things like discovery, and it really is needed to make sure for example things from presets are removed

dignifiedquire: > How empty is a Buidler::empty_builder? I think right now it still includes these default binds, which honestly is a bit confusing.

it is as empty as we can make it, while still being able to do _something_ when starting it

iroh/src/magicsock/transports.rs
dignifiedquire: I don't disagree, but this was actually the simplest way I came up with so far, to get this behaviour

iroh/src/magicsock/transports/ip.rs
matheus23: I think what @flub is suggesting is removing `Config::V*Default` cases, and adding an `is_default: bool`.
And then representing e.g. `Config::V4Default { ip_addr: "0.0.0.0".parse()?, port: 1234, is_required: true }` as
`Config::V4 { ip_net: "0.0.0.0/0".parse()?, port: 1234, is_required: true, is_default: true }`.

So, in short, representing the old default case as having prefix len 0.

iroh/src/magicsock/transports/ip.rs
flub: It has been the default for a long time and when users notice it has been really surprising to them!

The API we're adding is explicitly to give more control over what is exactly bound. I think this should also be exposed in a `BindOpts` field so that you get control over it. And by default this should be set to not fallback as otherwise this will be very surprising to users.

If a user **does not** manually call any `Builder::bind_addr*` functions then setting that bind option to use the fallback on the two IP transports added in that case is reasonable.

iroh/src/endpoint.rs
dignifiedquire: we could make it fully empty, and add the default binds using the `n0` preset

iroh/src/endpoint.rs
dignifiedquire: they are doing different things, so maybe we just have 

- `bind_default_addr`
- `bind_addr_with_opts` 

to avoid confusion?

(note: that looks horrible, but I am running out of ideas)

iroh/src/endpoint.rs
flub: Consider that today you can call `clear_relay_transport` and then you **can not** add a relay transport back (you can only add a relay transport by using `Builder::empty(RelayMode)`) I would propose to rename `clear_relay_transport` and `clear_ip_transport` into `disable_*_transport` and then this is solved.

iroh/src/magicsock/transports.rs
```diff
@@ -89,11 +89,14 @@ impl TransportConfig {
...
+        use netdev::ipnet::Ipv4Net;
+
         Self::Ip {
-            config: ip::Config::V4Default {
-                ip_addr: Ipv4Addr::UNSPECIFIED,
+            config: ip::Config::V4 {
+                ip_net: Ipv4Net::new(Ipv4Addr::UNSPECIFIED, 0).expect("checked"),
                 port: 0,
                 is_required: true,
+                is_default: false,
```
matheus23, line 99: I mean, the prefix length for this is 0, so it probably doesn't matter much, but shouldn't this be set as default?

iroh/src/magicsock/transports.rs
```diff
@@ -102,12 +105,15 @@ impl TransportConfig {
...
+
         Self::Ip {
-            config: ip::Config::V6Default {
-                ip_addr: Ipv6Addr::UNSPECIFIED,
+            config: ip::Config::V6 {
+                ip_net: Ipv6Net::new(Ipv6Addr::UNSPECIFIED, 0).expect("checked"),
                 scope_id: 0,
                 port: 0,
                 is_required: false,
+                is_default: false,
```
matheus23, line 116: Same here?

iroh/src/endpoint.rs
dignifiedquire: done

iroh/src/magicsock/transports.rs
dignifiedquire: https://github.com/n0-computer/iroh/issues/3831 moved to an issue

iroh/src/magicsock/transports/ip.rs
dignifiedquire: changed to prefix len based sorting

iroh/src/magicsock/transports/ip.rs
dignifiedquire: will do this change in a follow up

matheus23 (approved):
Yay routing tables and interfaces.

The "default_ipv{4,6}" I commented on can be ignored. "Default" here is an overloaded term. It's fine to keep it as is (it's not public API, we can improve it later)

iroh/src/endpoint/bind.rs
```diff
@@ -0,0 +1,208 @@
...
+    /// Sets the network prefix length which manages route.
+    ///
```
flub:
```suggestion
    /// Sets the network prefix length of the subnet for this interface.
    ///
    /// The prefix length is used in the routing table that is built to decide where datagrams without a specific source address are sent. If these number of leading bits in the destination IP address match the same number of leading bits in this bound socket address, then it matches the subnet and the datagram will be sent here. Otherwise the next bound sockets will be checked for a subnet match. Sockets are ordered from longest prefix to shortest prefix.
    ///
    /// If no bound address has a matching subnet, the bind marked with [`Self::is_default`] will be used.
    ///
    /// Note that most datagrams belonging to a traffic flow are in response to an incoming datagram. Those are usually sent on the same bound socket as they were received and will not consult the routing table derived from these bound sockets to select the socket on which they will be sent.
    ///
```

iroh/src/endpoint/bind.rs
```diff
@@ -0,0 +1,208 @@
...
+    /// Defaults to `0`.
+    ///
+    /// Must not be larger than `32` for IPv4 and `128` for IPv6.
+    prefix_len: u8,
+    /// If set, binding this interface is required and any errors will abort the
+    /// initialization of the endpoint.
+    ///
+    /// Defaults to `true`.
+    is_required: bool,
+    is_default: bool,
```
flub:
```suggestion
    /// Whether this socket should be used as default route.
    ///
    /// The default route is used for outgoing datagrams not belonging to an existing traffic flow, which does not fit in any subnet of the bound sockets. It is assumed this subnet has a gateway router to route such packets.
    ///
    /// See [`Self::prefix_len`] for details of how such routing works.
    is_default: bool,
```
Consider renaming this to `is_default_route` to reduce "default" ambiguity.

iroh/src/endpoint/bind.rs
```diff
@@ -0,0 +1,208 @@
...
+        }
+    }
+}
+
+impl BindOpts {
+    /// Sets the network prefix length which manages route.
+    ///
+    /// Defaults to `0`.
+    ///
+    /// Must not be larger than `32`
```
flub:
```suggestion
    /// Sets the network prefix length of the subnet this interface is in.
    ///
    ///
    /// The subnets of bound sockets are used to route outgoing datagrams not belonging to an existing traffic flow to the socket they should be sent on. Subnets are ordered from longest prefix length to shortest prefix length and the first subnet which contains the destination IP address will be chosen. If no subnet matches but there is a bound socket marked with [`Self.set_is_default`] then this socket will be used. In this case it is assumed the attached subnet has a gateway router to forward the datagram.
    ///
    /// Defaults to `0`, which means *all* IP addresses will belong to the subnet of this socket's address. If multiple sockets of the same address family (IPv4 or IPv6) are bound with such a `/0` prefix the socket which will be chosen is undefined.
    ///
    /// For IPv4 sockets the maximum prefix is `32`. For IPv6 the maximum prefix is `128`.
```

iroh/src/endpoint/bind.rs
```diff
@@ -0,0 +1,208 @@
...
+    pub fn set_prefix_len(mut self, prefix_len: u8) -> Self {
+        self.prefix_len = prefix_len;
+        self
+    }
+
+    /// Returns the `prefix_len`.
```
flub:
```suggestion
    /// Returns the `prefix_len`, see [`Self::set_prefix_len`].
```

iroh/src/endpoint/bind.rs
```diff
@@ -0,0 +1,208 @@
...
+    pub fn prefix_len(&self) -> u8 {
+        self.prefix_len
+    }
+
+    /// If set, binding this interface is required and any errors will abort the
+    /// initialization of the endpoint.
+    ///
+    /// Defaults to `true`.
```
flub:
```suggestion
    /// Sets whether bind errors are fatal for this socket.
    ///
    /// If `false` and this socket fails to bind, the error will be silently ignored and the endpoint will still be created.
    ///
    /// Defaults to `true`.
```

iroh/src/endpoint/bind.rs
```diff
@@ -0,0 +1,208 @@
...
+    pub fn set_is_required(mut self, is_required: bool) -> Self {
+        self.is_required = is_required;
+        self
+    }
+
+    /// Returns the value for `is_required`.
```
flub:
```suggestion
    /// Returns the value set by [`Self::set_is_required`].
```

iroh/src/endpoint/bind.rs
```diff
@@ -0,0 +1,208 @@
...
+    pub fn is_required(&self) -> bool {
+        self.is_required
+    }
+
+    /// Set if this is a default route.
+    ///
+    /// Defaults to `false`.
```
flub:
```suggestion
    /// Sets whether this is a default route.
    ///
    /// The default route is used for outgoing datagrams not belonging to an existing traffic flow, which does not fit in any subnet of the bound sockets. It is assumed this subnet has a gateway router to route such packets.
    ///
    /// See [`Self::set_prefix_len`] for details on how this routing works.
    ///
    /// Defaults to `false`.
```

iroh/src/endpoint/bind.rs
```diff
@@ -0,0 +1,208 @@
...
+    pub fn set_is_default(mut self, is_default: bool) -> Self {
+        self.is_default = is_default;
+        self
+    }
+
+    /// Returns the value for `is_default`.
```
flub:
```suggestion
    /// Returns the current value set by [`Self::set_is_default`].
```

iroh/src/endpoint/bind.rs
```diff
@@ -0,0 +1,208 @@
...
+    pub fn is_default(&self) -> bool {
+        self.is_default
+    }
+}
+
+/// A simpler version of `ToSocketAddrs`, that does not do any DNS resolution.
```
flub: Please link `ToSocketAddrs`.

iroh/src/magicsock/transports/ip.rs
```diff
@@ -1,31 +1,178 @@
...
     socket: Arc<UdpSocket>,
     local_addr: Watchable<SocketAddr>,
     metrics: Arc<MagicsockMetrics>,
 }
 
-fn bind_with_fallback(mut addr: SocketAddr) -> io::Result<netwatch::UdpSocket> {
-    debug!(%addr, "binding");
+/// IP transport configuration
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum Config {
```
flub: Should this one not be `pub(crate)`? I don't think users can create this directly? Same for all the methods.

iroh/src/magicsock/transports/ip.rs
```diff
@@ -1,31 +1,178 @@
...
+        }
+    }
+
+    /// Does this configuration match to send to the given `src` and `dst` address.
+    pub(crate) fn is_valid_send_addr(&self, src: Option<IpAddr>, dst: SocketAddr) -> bool {
+        match src {
+            Some(src) => match (self, src) {
+                (
+                    Self::V4 {
+                        ip_net: ip_addr, ..
```
flub: hehe, nice rust-analyzer rename as asked by my previous nit 🫠 
This should also format shorter if you don't rename it I think.

iroh/src/magicsock/transports/ip.rs
```diff
@@ -232,3 +373,216 @@ impl IpSender {
...
+}
+
+impl IpTransportsSender {
+    pub(super) fn v4_iter_mut(&mut self) -> impl Iterator<Item = &mut IpSender> {
+        self.v4.iter_mut()
+    }
+
+    pub(super) fn v4_default_mut(&mut self) -> Option<&mut IpSender> {
+        if let Some(i) = self.default_v4_index {
+            return Some(&mut self.v4[i]);
```
flub, line 383: `return self.v4.get_mut(i)` is more panic safe? But would hide an error, so fine if you prefer to keep this.

iroh/src/magicsock/transports/ip.rs
dignifiedquire: done

iroh/src/magicsock/transports/ip.rs
dignifiedquire: yeah, I wanted to panic if there is a bug, and not silently fail

iroh/src/magicsock/transports/ip.rs
dignifiedquire: done

iroh/src/endpoint/bind.rs
dignifiedquire: done

PR #3708, opened 2025-11-28
refactor(iroh): simplify internal transports sending

dignifiedquire:
removes the need for two `send` impls

### DIFF
--- a/iroh/src/magicsock/remote_map/remote_state.rs
+++ b/iroh/src/magicsock/remote_map/remote_state.rs
@@ -357,22 +357,6 @@ impl RemoteStateActor {
         }
     }
 
-    async fn send_datagram(
-        &mut self,
-        dst: transports::Addr,
-        owned_transmit: OwnedTransmit,
-    ) -> n0_error::Result<()> {
-        let transmit = transports::Transmit {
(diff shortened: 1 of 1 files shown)


iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -396,15 +395,16 @@ impl RemoteStateActor {
             if self.paths.is_empty() {
                 warn!("Cannot send datagrams: No paths to remote endpoint known");
             }
-            for addr in self.paths.addrs() {
+            let addrs: Vec<_> = self.paths.addrs().cloned().collect();
+            for addr in addrs {
```
flub: Is this change required?  It's unfortunate that this allocates again.

flub (approved):

iroh/src/magicsock/remote_map/remote_state.rs
dignifiedquire: fixed, a bit complicated but worked

flub: oh, I missed that this was only on the RemoteStateActor so not that hot. But still nice to not have the allocations.

PR #3715, opened 2025-11-30
docs(iroh-relay): add section on how to use as a library

jost-s:
## Description

Add a section to the `iroh-relay` README that describes how to run an in-process relay server when using iroh/iroh-relay as a library.

## Notes & open questions

I was creating an instance of the relay until finding out about the function in `test-utils`, and skipping the certificate verification was hard to find as well.

## Change checklist

- [x] Self-review.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -102,22 +102,22 @@ dependencies = [
 
 [[package]]
 name = "anstyle-query"
-version = "1.1.4"
+version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9e231f6134f61b71076a3eab506c379d4f36122f2af15a9ff04415ea4c3339e2"
+checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
 name = "anstyle-wincon"
-version = "3.0.10"
+version = "3.0.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3e0633414522a32ffaac8ac6cc8f748e090c5717661fddeea04219e2344f5f2a"
+checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -240,7 +240,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "16e2cdb6d5ed835199484bb92bb8b3edd526effe995c61732580439c1a67e2e9"
 dependencies = [
  "base64",
- "http 1.3.1",
+ "http 1.4.0",
  "log",
  "url",
 ]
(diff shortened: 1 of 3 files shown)


iroh-relay/README.md
```diff
@@ -64,6 +64,18 @@ The relay server will run over http on port 3340, as it does using the `--dev` f
 
 The relay will use the configured TLS certificates for the QUIC connection, but use http (rather than https) for the server.
 
+### Using iroh as a library
```
matheus23: I think this heading could use some work. To me this section reads more like "running tests with local relay servers". Not sure if that's the best heading, but probably closer to the optimum.

jost-s: I agree that the heading could use some work.

All of these sections are under "Local testing", and the first two sub-headings refer to running a relay server from the command line as a binary. So the difference to me is that the new section is about running relay servers programmatically from within an a test file.

How about "Writing integration tests with local relay servers"? and making this a `##` heading?

jost-s: Actually that's pretty much what you suggested. 😁

matheus23: That sounds good!

matheus23: If you want to keep the ### heading, you could also do something like "Using iroh-relay as a library to run in-process relays"

jost-s: Cool, edited

iroh-relay/README.md
```diff
@@ -64,6 +64,18 @@ The relay server will run over http on port 3340, as it does using the `--dev` f
...
+### Using iroh-relay as a library to run in-process relays in integration tests
+
+When using iroh as a transport library in an application or other library, there is a simple way of running an in-process relay server:
+
+- Enable `test-utils` feature of the `iroh` crate
+```toml
+iroh = { version = "0.95", features = ["test-utils"] }
+```
+- Spawn a relay server by calling [`iroh::test_utils::run_relay_server().await`](https://docs.rs/iroh/latest/iroh/test_utils/fn.run_relay_server.html)
+This will start a relay server with self-signed TLS certificate, listening on a localhost port, and return the server's URL.
```
matheus23:
```suggestion
This will start a relay server with a self-signed TLS certificate, listening on a localhost port, and return the server's URL.
```

iroh-relay/README.md
```diff
@@ -64,6 +64,18 @@ The relay server will run over http on port 3340, as it does using the `--dev` f
...
+- For the clients (= endpoints) to successfully connect to the relay, disable TLS certificate verification on the client by calling [`Endpoint::insecure_skip_relay_cert_verify`](https://docs.rs/iroh/latest/iroh/endpoint/struct.Builder.html#method.insecure_skip_relay_cert_verify).
```
matheus23:
```suggestion
- For the iroh endpoints to successfully connect to the relay, disable TLS certificate verification by calling [`Endpoint::insecure_skip_relay_cert_verify`](https://docs.rs/iroh/latest/iroh/endpoint/struct.Builder.html#method.insecure_skip_relay_cert_verify).
```

matheus23 (approved):
Thank you :)

jost-s:
~~This needs a retry of the failed workflow.~~

Actually it doesn't seem to be a test flake, but a compilation problem. I can look into fixing that, but it's confusing that this docs change would have to do with it.

matheus23:
Yeah don't worry about it, we're handling it right now :)
wasm-bindgen released a new version while this PR was open, and unfortunately our CI isn't set up to handle that (for annoying reasons).

jost-s:
Alright, got it. Thanks for the response.

PR #3721, opened 2025-12-02
feat(iroh)!: encapsulate the `quinn::TransportConfig` to enforce certain minimums when used with multipath

### LINKED ISSUE #3635: Improve setting TransportConfig parameters
There's the problem we silently overwrite the some end-user settings:

https://github.com/n0-computer/iroh/pull/3381/files#r2428775236

Tricky to decide if this is a major problem.  But some resolution of that should happen.

We also set per-path keep-alive timeout and heartbeat interval twice.  Which might be confusing?

ramfox:
## Description

This PR encapsulates the `quinn::TransportConfig` in a new struct called `QuicTransportConfig`. It has all of the same methods as the `quinn::TransportConfig`, but the follow methods will log warnings if the user given values make iroh + multipath sub-optimal:
  - default_path_keep_alive_interval() - should be at most `HEARTBEAT_INTERVAL`ms
  - default_path_max_idle_timeout() - should be at most `PATH_MAX_IDLE_TIMEOUT`ms
  - max_concurrent_multipath_paths() - should be at least `MAX_MULTIPATH + 1`
  - set_max_remote_nat_traversal_addresses() - should be at least `MAX_MULTIPATH`

These values are also set properly by default when creating a `QuicTransportConfig`.

## quinn encapsulation/co-location

Created a new mod `quic` and co-located all the quinn exports there.

This would be the mod where we do any future encapsulation.

Note: the `StaticConfig` struct is a private struct that still uses `quinn::TransportConfig` directly.

~~Second note: the `ConnectConfig` was `Clone`, and no longer is. This is due to a weird tension in our encapsulation of `quinn::TransportConfig`. The quinn APIs typically expect an `Arc<quinn::TransportConfig>`, however if we store an `Arc` of the `quinn::TransportConfig` in `QuicTransportConfig`, then we can't actually mutate any fields inside it, unless we also put it in a `Mutex`. The "slickest" way to ensure that we did not need to store an `Arc<quinn::TransportConfig>`, was to ensure that we didn't have APIs that required cloning or Arc-ing the `QuicTransportConfig` itself (quinn::TransportConfig is also not clone).~~

~~It's possible this is terribly annoying for our users, so if someone else has an ergonomic way of handling this, please let me know.~~

~~waiting for PR n0-computer/quinn#213 to merge, then this can be merged~~ merged

~~third note: I adjusted the `HEARTBEAT_INTERVAL` and `PATH_MAX_IDLE_TIMEOUT` to be numerical values rather than `Duration`s, so they can be read-able in the documentation.~~

closes #3635 

## Breaking Changes
- `iroh`
    - changes
        - `QuinnTransportConfig` renamed to `QuicTransportConfig` & is now `Clone`
        - `ConnectOptions::transport_config: Option<Arc<quinn::TransportConfig>>` -> `ConnectOptions::transport_config: Option<QuicTransportConfig>
        - `ConnectOptions::with_transport_config(mut self, transport_config: Arc<quinn::TransportConfig>)` -> `ConnectOptions::with_transport_config(mut self, transport_config: QuicTransportConfig) 

## Open Questions
- [x] Need verification that my assumptions about "at least" and "at most" for the above are the correct assumptions about what we want. **done in #3635**

### DIFF
--- a/iroh/src/endpoint/quic.rs
+++ b/iroh/src/endpoint/quic.rs
@@ -369,14 +369,14 @@ impl QuicTransportConfig {
     /// concurrently when multipath is negotiated. For any path to be opened, the remote must
     /// enable multipath as well.
     ///
-    /// Note: setting this value below 13 is not recommended and will currently log a warning.
+    /// Note: this method will ignore values less than the recommended 13 and will log a warning.
     pub fn max_concurrent_multipath_paths(&mut self, max_concurrent: u32) -> &mut Self {
         if max_concurrent < MAX_MULTIPATH_PATHS + 1 {
             warn!(
-                "setting QuicTransportConfig::max_concurrent_multipath_paths to {}, below the recommended minimum {}",
-                max_concurrent,
+                "QuicTransportConfig::max_concurrent_multipath_paths must be at minimum {}, ignoring user supplied value",
                 MAX_MULTIPATH_PATHS + 1
             );
+            return self;
         }
         self.0.max_concurrent_multipath_paths(max_concurrent);
         self
@@ -388,14 +388,14 @@ impl QuicTransportConfig {
     /// interact with the [`QuicTransportConfig::max_idle_timeout`], if the last path is
     /// abandoned the entire connection will be closed.
     ///
-    /// Note: setting this timeout to higher than 6500ms is not currently recommended
-    /// and will log a warning.
+    /// Note: this method will ignore values higher than the recommended 6500 ms and will log a warning.
     pub fn default_path_max_idle_timeout(&mut self, timeout: Duration) -> &mut Self {
         if timeout > PATH_MAX_IDLE_TIMEOUT {
             warn!(
-                "setting QuicTransportConfig::default_path_max_idle to {:?}, above the recommended maximum {:?}",
-                timeout, PATH_MAX_IDLE_TIMEOUT
+                "QuicTransportConfig::default_path_max_idle must be at most {:?}, ignoring user supplied value",
+                PATH_MAX_IDLE_TIMEOUT
             );
+            return self;
         }
         self.0.default_path_max_idle_timeout(Some(timeout));
         self
@@ -408,13 +408,14 @@ impl QuicTransportConfig {
     /// [`QuicTransportConfig::keep_alive_interval`] will keep the connection active, with no
     /// control over which path is used for this.
     ///
-    /// Note: setting this timeout higher than 5 seconds is not recommended and will log a warning.
+    /// Note: this method will ignore values higher than the recommended 5 seconds and will log a warning.
     pub fn default_path_keep_alive_interval(&mut self, interval: Duration) -> &mut Self {
         if interval > HEARTBEAT_INTERVAL {
             warn!(
-                "setting QuicTransportConfig::default_path_keep_alive to {:?}, above the recommended maximum {:?}",
-                interval, HEARTBEAT_INTERVAL
+                "QuicTransportConfig::default_path_keep_alive must be at most {:?}, ignoring user supplied value",
+                HEARTBEAT_INTERVAL
             );
+            return self;
         }
         self.0.default_path_keep_alive_interval(Some(interval));
         self
@@ -431,13 +432,14 @@ impl QuicTransportConfig {
     /// enabled via [`Self::max_concurrent_multipath_paths`], a default value of
     /// 12 will be used.
     ///
-    /// Note: setting this value to lower than 12 is not recommended and will log a warning.
+    /// Note: this method will ignore values less than the recommended 12 and will log a warning.
     pub fn set_max_remote_nat_traversal_addresses(&mut self, max_addresses: u8) -> &mut Self {
         if max_addresses < MAX_MULTIPATH_PATHS as u8 {
             warn!(
-                "setting QuicTransportConfig::max_remote_nat_traversal_addresses to {}, below the recommended minimum {}",
-                max_addresses, MAX_MULTIPATH_PATHS
+                "QuicTransportConfig::max_remote_nat_traversal_addresses must be at least {}, ignoring user supplied value",
+                MAX_MULTIPATH_PATHS
             );
+            return self;
         }
         self.0.set_max_remote_nat_traversal_addresses(max_addresses);
         self
--- a/iroh/src/lib.rs
+++ b/iroh/src/lib.rs
@@ -275,8 +275,6 @@ pub use iroh_base::{
     Signature, SignatureError, TransportAddr,
 };
 pub use iroh_relay::{RelayConfig, RelayMap, endpoint_info};
-// these consts are public so they can be used for reference in documentation
-pub use magicsock::{HEARTBEAT_INTERVAL, MAX_MULTIPATH_PATHS, PATH_MAX_IDLE_TIMEOUT};
 pub use n0_watcher::Watcher;
 
 #[cfg(any(test, feature = "test-utils"))]
--- a/iroh/src/magicsock.rs
+++ b/iroh/src/magicsock.rs
@@ -84,18 +84,18 @@ pub use self::{metrics::Metrics, remote_map::PathInfo};
 ///
 /// If a path is idle for this long, a PING frame will be sent to keep the connection
 /// alive.
-pub const HEARTBEAT_INTERVAL: Duration = Duration::from_secs(5);
+pub(crate) const HEARTBEAT_INTERVAL: Duration = Duration::from_secs(5);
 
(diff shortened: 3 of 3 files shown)


iroh/src/endpoint/quic.rs
```diff
@@ -0,0 +1,491 @@
...
+    VarIntBoundsExceeded, WeakConnectionHandle, WriteError,
+};
+pub use quinn_proto::{
+    FrameStats, IdleTimeout, TransportError, TransportErrorCode, UdpStats, Written,
+    congestion::{Controller, ControllerFactory},
+    crypto::{
+        AeadKey, CryptoError, ExportKeyingMaterialError, HandshakeTokenKey,
+        ServerConfig as CryptoServerConfig, UnsupportedVersion,
+    },
+};
```
flub, line 26: I never know if quinn should be re-exporting all of quinn-proto types that need to be pub as well. I think that would make sense. But certainly for multipath we haven't tackled that yet, and they just end up being wherever they are.

If later we move these types to be re-exported by quinn, would that be a breaking change for iroh? I think not, so it's probably fine. But writing this down here to get sanity checked by someone else.

iroh/src/endpoint/quic.rs
```diff
@@ -0,0 +1,491 @@
...
+/// are out of scope with what is optimal for iroh.
+///
+/// Look at the following methods for more details:
+/// - [`QuicTransportConfig::keep_alive_interval`]
+/// - [`QuicTransportConfig::default_path_keep_alive_interval`]
+/// - [`QuicTransportConfig::default_path_max_idle_timeout`]
+/// - [`QuicTransportConfig::max_concurrent_multipath_paths`]
+/// - [`QuicTransportConfig::set_max_remote_nat_traversal_addresses`]
+#[derive(Debug)]
+pub struct QuicTransportConfig(pub(crate) TransportConfig);
```
flub: nit: I think this would be more readable:

```suggestion
pub struct QuicTransportConfig(pub(crate) quinn::TransportConfig);
```

iroh/src/endpoint/quic.rs
```diff
@@ -0,0 +1,491 @@
...
+    /// enabled for the connection to be preserved. Must be set lower than the idle_timeout of both
+    /// peers to be effective.
+    ///
+    /// # Errors
+    ///
+    /// Will return a [`QuicTransportConfigError::KeepAliveInterval`] if the given value is more than [`HEARTBEAT_INTERVAL`].
+    pub fn keep_alive_interval(
+        &mut self,
+        value: Duration,
+    ) -> Result<&mut Self, QuicTransportConfigError> {
```
flub: See also my comment on the errors here: #3635 (comment)
Having some methods here that return errors is a bit weird. It makes the API rather heterogenous and seem inconsistent for humans. I'm not a big fan of this approach.

As I say in the issue, I think I'd prefer to go with logging warnings. But am open to be convinced otherwise - I may not have thought of everything.

And finally, also as mentioned in the issue. This one can be freely chosen by the user, we only really care about per-path values.

iroh/src/endpoint.rs
```diff
@@ -48,30 +44,27 @@ use crate::{
 mod connection;
 pub(crate) mod hooks;
 pub mod presets;
+pub mod quic;
```
flub: Does this need to be pub? You're re-exporting all the types below, so I don't think it needs to be reachable from here? Feels more like exposing our internal structure.

iroh/src/lib.rs
```diff
@@ -275,6 +275,8 @@ pub use iroh_base::{
     Signature, SignatureError, TransportAddr,
 };
 pub use iroh_relay::{RelayConfig, RelayMap, endpoint_info};
+// these consts are public so they can be used for reference in documentation
+pub use magicsock::{HEARTBEAT_INTERVAL, MAX_MULTIPATH_PATHS, PATH_MAX_IDLE_TIMEOUT};
```
flub: My instincts remain to keep as much as private as possible 😁 

The names are pretty arbitrary (like why not PATH_HEARTBEAT_INTERVAL), the use of them might still change, they might become obsolete. tl;dr: can we keep this private please?

iroh/src/magicsock.rs
```diff
@@ -80,22 +80,22 @@ pub use self::{metrics::Metrics, remote_map::PathInfo};
 // /// expire at 30 seconds, so this is a few seconds shy of that.
 // const ENDPOINTS_FRESH_ENOUGH_DURATION: Duration = Duration::from_secs(27);
 
-/// The duration in which we send keep-alives.
+/// The number of milliseconds after which we send keep-alives.
 ///
 /// If a path is idle for this long, a PING frame will be sent to keep the connection
 /// alive.
-pub(crate) const HEARTBEAT_INTERVAL: Duration = Duration::from_secs(5);
+pub const HEARTBEAT_INTERVAL: u64 = 5000;
```
flub: I'd prefer to keep these private (see above comment), which also means we can keep it in the less error-prone and stronger-typed version.

flub:
> Second note: the `ConnectConfig` was `Clone`, and no longer is. This is due to a weird tension in our encapsulation of `quinn::TransportConfig`. The quinn APIs typically expect an `Arc<quinn::TransportConfig>`, however if we store an `Arc` of the `quinn::TransportConfig` in `QuicTransportConfig`, then we can't actually mutate any fields inside it, unless we also put it in a `Mutex`. The "slickest" way to ensure that we did not need to store an `Arc<quinn::TransportConfig>`, was to ensure that we didn't have APIs that required cloning or Arc-ing the `QuicTransportConfig` itself (quinn::TransportConfig is also not clone).
> 
> It's possible this is terribly annoying for our users, so if someone else has an ergonomic way of handling this, please let me know.

I don't really know why `quinn:TransportConfig` isn't `Clone`? Everything inside it is `Clone`. Maybe we even could make all the fields `pub` and mark it `#[non_exhaustive]`. It has a `Default` impl anyway so you can then still construct it. Somehow I wonder if the current layout pre-dates the existence of non_exhaustive?

iroh/src/endpoint/quic.rs
ramfox: shouldn't be a breaking change, since `use` just makes an alias to an underlying type, it doesn't re-create it in the mod it gets `use`-d in, so it has the same underlying source.

iroh/src/endpoint.rs
ramfox: oh yea thank you

iroh/src/endpoint/quic.rs
```diff
@@ -0,0 +1,445 @@
...
+    /// Note: setting this value below 13 is not recommended and will currently log a warning.
+    pub fn max_concurrent_multipath_paths(&mut self, max_concurrent: u32) -> &mut Self {
+        if max_concurrent < MAX_MULTIPATH_PATHS + 1 {
+            warn!(
+                "setting QuicTransportConfig::max_concurrent_multipath_paths to {}, below the recommended minimum {}",
+                max_concurrent,
+                MAX_MULTIPATH_PATHS + 1
+            );
+        }
+        self.0.max_concurrent_multipath_paths(max_concurrent);
```
flub, line 381: I think we should ignore the user's setting if they set it too small. if `max_concurrent < MAX_MULTIPATH_PATHS + 1` we print a warning saying we're ignoring their setting and actually set `self.0.max_concurrent_multipath_paths(MAX_MULTIPATH_PATHS + 1)`.

Otherwise it is too easy to get hard-to-debug issues, where we'll only figure things out much too late. The issues we will get from ignoring the user-provided value will be much easier to understand and then we can figure out what the usecases are and how to support them.

Same for the others.

iroh/src/magicsock.rs
```diff
@@ -84,18 +84,18 @@ pub use self::{metrics::Metrics, remote_map::PathInfo};
 ///
 /// If a path is idle for this long, a PING frame will be sent to keep the connection
 /// alive.
-pub(crate) const HEARTBEAT_INTERVAL: Duration = Duration::from_secs(5);
+pub const HEARTBEAT_INTERVAL: Duration = Duration::from_secs(5);
```
flub: Do these still need to be pub? Same comment as before applies, it means we can't remove them. Likely we'd stop using them if we face that problem and they'd be some random unused constants.

ramfox: hmmm i thought I made them `pub(crate)`, sorry I meant to take your initial comment, I didn't ignore it!

flub (approved):

iroh/src/magicsock.rs
flub: all good!

PR #3724, opened 2025-12-02
feat(iroh): add fast-apple-datapath feature and enable it by default

dignifiedquire:


iroh/bench/Cargo.toml
```diff
@@ -28,6 +28,8 @@ tracing-subscriber = { version = "0.3.0", default-features = false, features = [
 ] }
 
 [features]
-default = ["metrics"]
+default = ["metrics", "fast-apple-datapath"]
 metrics = ["iroh/metrics", "iroh-metrics"]
 local-relay = ["iroh/test-utils"]
+# Use private Apple APIs to send multiple packets in a single syscall.
+fast-apple-datapath = ["iroh/fast-apple-datapath", "quinn/fast-apple-datapath"]
```
flub, line 35: Not sure why you explicitly add `quinn/fast-apple-datapath` here as it should be pulled in vial the `iroh/fast-apple-datapath` feature. But it also doesn't do any harm.

flub (approved):

iroh/bench/Cargo.toml
dignifiedquire: mostly because it is a direct dependency, I am never fully sure what to do in these cases

PR #3734, opened 2025-12-04
feat: improve RelayMap and RelayMode configuration

dignifiedquire:


iroh/src/endpoint.rs
```diff
@@ -2471,4 +2480,30 @@ mod tests {
...
+        let _ep = Endpoint::empty_builder(RelayMode::custom([
+            "https://use1-1.relay.n0.iroh-canary.iroh.link.".parse()?,
+            "https://euw-1.relay.n0.iroh-canary.iroh.link.".parse()?,
+        ]))
+        .bind()
+        .await?;
+
+        let relays = RelayMap::try_from_iter([
+            "https://use1-1.relay.n0.iroh.iroh.link/",
+            "https://euc1-1.relay.n0.iroh.iroh.link/",
```
Arqu, line 2517: We are still inconsistent with our usage of `.` and `/` for the relay urls everywhere.

We should be sticking to `.`

flub: Yes, we should stick to `.` in all examples.

flub: And of course, defaults.

dignifiedquire: this is not an example but a test, I a actually want to make sure that it also works without  `.` tbh

iroh-relay/src/relay_map.rs
```diff
@@ -50,6 +50,29 @@ impl RelayMap {
...
     }
 
+    /// Creates a [`RelayMap`] from an iterator.
+    ///
+    /// # Example
+    /// ```rust
+    /// # use iroh_relay::RelayMap;
+    /// let map =
+    ///     RelayMap::try_from_iter(["https://relay_0.cool.com", "https://relay_1.cool.com"]).unwrap();
+    /// ```
```
flub, line 63: doc comment should probably point out that a default RelayConfig is used and that this means we'll assume the relay URL has QAD running on port 7842 or whatever relevant.

flub (approved):

iroh/src/endpoint.rs
Arqu: Ah, wasn't super clear from that. I thought you wanted to do the same thing in 2 ways ie empty builder and pass it in from an iter.

iroh-relay/src/relay_map.rs
dignifiedquire: done

flub:
@dignifiedquire please update and merge

PR #3742, opened 2025-12-07
fix(iroh): only switch paths if the new path is actually better

dignifiedquire:
For now the restriction is that for direct to direct path switching latency must be at least 10ms better.

### DIFF
--- a/iroh/src/magicsock/remote_map/remote_state.rs
+++ b/iroh/src/magicsock/remote_map/remote_state.rs
@@ -77,9 +77,6 @@ const ACTOR_MAX_IDLE_TIMEOUT: Duration = Duration::from_secs(60);
 /// The minimum RTT difference to make it worth switching IP paths
 const RTT_SWITCHING_MIN_IP: Duration = Duration::from_millis(5);
 
-/// The minimum RTT difference to make it worth switching relay paths
-const RTT_SWITCHING_MIN_RELAY: Duration = Duration::from_millis(3);
-
 /// How much do we prefer IPv6 over IPv4?
 const IPV6_RTT_ADVANTAGE: Duration = Duration::from_millis(3);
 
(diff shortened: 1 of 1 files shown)


matheus23:
This is the logic we had for "sticky" best path selection in iroh before:
```
    /// The current address we use to send on.
    ///
    /// This is *almost* the same as going through `paths` and finding
    /// the best one, except that this is
    /// 1. Not updated in `send_addr`, but instead when there's changes to `paths`, so that `send_addr` can take `&self`.
    /// 2. Slightly sticky: It only changes when
    ///   - the current send addr is not a validated path anymore or
    ///   - we received a pong with lower latency.
    best: UdpSendAddr,
```
At least, according to a comment I wrote a while ago.

Also, it's not the full story, since we only get pongs if we sent pings in the first place. Looking up when we do that:
- We send full pings if we send or receive a call-me-maybe
- We decide to send call-me-maybes if we don't have a confirmed UDP path (e.g. only relay, or we didn't receive pongs on the path for a while, thus it expired) or if the RTT is >5ms. We do this every stayin alive heartbeat, or every 60s if the RTT is <5ms.

This is a little different that what's happening with `select_path` in the current `RemoteStateActor`.
Perhaps we should build it similarly, where we commit to a choice once we make one?

dignifiedquire:
I think the setup was different enough before that we can't & don't want to replicate the exact same logic, especially as ping handling is done quite differently inside quinn now. 

We should figure out a better strategy than what I implemented here likely, but we need something better than what is on `main` currently to avoid the flip-floping and unnecessary  switching I am seeing.

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -862,14 +862,39 @@ impl RemoteStateActor {
...
+            .flatten();
+
+        let selected_path = match (self.selected_path.get(), direct_path) {
+            (Some(ref addr @ transports::Addr::Ip(_)), Some((new_rtt, new_direct_addr))) => {
+                let current_rtt = path_rtts.get(addr);
+                if let Some(&current_rtt) = current_rtt {
+                    if current_rtt > new_rtt + Duration::from_millis(10) {
+                        // Switch to a new path it is at least 10ms better
+                        // TODO: what number is right?
+                        // TODO: should we take IPv6 vs IPv4 into consideration here?
```
flub: 10ms is ~3000km at the speed of light. the ipv6 advantage is ~900km. I think 3000km is rather a lot. How about 1500km? That would be 5ms.

Not all the times/distances are RTT, so half the distance for physical visualisation really.

And yes, I'd also apply the IPv6 advantage here, otherwise it is inconsistent depending on what was selected before.

Can we move the IPV6_RTT_ADVANTAGE and this new constant to the top of the file like usual?

flub (approved):
Generally I agree this is needed, I'm only suggesting some tweaks to the values used.

iroh/src/magicsock/remote_map/remote_state.rs
dignifiedquire: writing a cleaner version of this

matheus23: What I worry is that the RTT might increase by more than 5ms if it has a lot of traffic put through it - thus reintroducing flip-flopping.

But my general read of this whole thing is rather - it seems like there doesn't exist a correct number?

flub: Yep, I'm also worried about that. In theory this is fixed by doing congestion control properly and not having buffer bloat and RTT not going crazy. In practice that's a lot of wishful thinking.

I'm also hoping that these values would not trigger that, but it is a guess and we'll have to see how it behaves really.

I think there is still more protection that I want to build in later. E.g. I want to somehow rank paths and not allow going to a "lower ranked" path if you still have a higher one.  E.g. if you have IPv6 localhost that'd have the highest rank, IPv4 localhost would have the next, then other addresses that are clearly local interfaces etc. That way you'd never switch from an IPv6 localhost to a docker-supplied address for example. Or you wouldn't switch from a link-local address to a public one.

But for this PR I think limiting to what is proposed here is a simple and manageable step in the right direction.

dignifiedquire: pushed

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -857,33 +866,55 @@ impl RemoteStateActor {
...
+            .filter_map(|(addr, rtt)| {
+                if let transports::Addr::Ip(SocketAddr::V6(addr)) = *addr {
+                    Some((addr, *rtt))
+                } else {
+                    None
+                }
+            })
+            .min_by_key(|(_addr, rtt)| *rtt);
+
+        // Find the fasted relay path.
```
matheus23:
```suggestion
        // Find the fastest relay path.
```

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -857,33 +866,55 @@ impl RemoteStateActor {
             .collect();
 
         // Find the fastest direct or relay path.
-        const IPV6_RTT_ADVANTAGE: Duration = Duration::from_millis(3);
-        let direct_path = path_rtts
+        let direct_path_ipv4 = path_rtts
```
matheus23, line 866: Shouldn't the comment above say "Find the fastest IP v4 direct path"?

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -857,33 +866,55 @@ impl RemoteStateActor {
...
+            let rtt = path_rtts.get(&addr).copied();
+            (addr, rtt)
         });
-        if let Some((rtt, addr)) = selected_path {
+        let selected_path = select_best_path(
+            current_path.clone(),
+            direct_path_ipv4,
+            direct_path_ipv6,
+            relay_path,
+        );
```
matheus23, line 905: :heart:

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -74,6 +74,15 @@ mod path_state;
 /// in a high frequency, and to keep data about previous path around for subsequent connections.
 const ACTOR_MAX_IDLE_TIMEOUT: Duration = Duration::from_secs(60);
 
+/// The minimum RTT difference to make it worth switching IP paths
```
flub, line 77: Maybe we could add the speed-of-light distance of these values to the doc comments here. It kind of helps frame how we chose those values and what effect we expect them to have.

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -887,7 +889,7 @@ impl RemoteStateActor {
             })
             .min_by_key(|(_addr, rtt)| *rtt);
 
-        // Find the fasted relay path.
+        // Find the fastests relay path.
```
matheus23: Hate to be the bearer of bad news :sweat_smile:

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -74,6 +74,15 @@ mod path_state;
 /// in a high frequency, and to keep data about previous path around for subsequent connections.
 const ACTOR_MAX_IDLE_TIMEOUT: Duration = Duration::from_secs(60);
 
+/// The minimum RTT difference to make it worth switching IP paths
+const RTT_SWITCHING_MIN_IP: Duration = Duration::from_millis(5);
+
+/// The minimum RTT difference to make it worth switching relay paths
+const RTT_SWITCHING_MIN_RELAY: Duration = Duration::from_millis(3);
```
flub: What's the reason to make this different for relays as for IP?

Multiple relays is also very theoretical currently. Our one relay address is currently sent to the relay actor with EndpointId destination and the actor doesn't necessarily send the datagrams via that relay but is free to send it via another relay that has that EndpointId connected. Something to keep in mind when working on this for now.

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -856,34 +865,58 @@ impl RemoteStateActor {
...
+        // Find the fastests relay path.
+        let relay_path = path_rtts
+            .iter()
+            .filter(|(addr, _rtt)| addr.is_relay())
+            .min_by_key(|(_addr, rtt)| *rtt)
+            .map(|(addr, rtt)| (addr.clone(), *rtt));
+
+        let current_path = self.selected_path.get().map(|addr| {
+            let rtt = path_rtts.get(&addr).copied();
+            (addr, rtt)
```
flub: I'm not sure about this. If the currently selected path is not available in `path_rtts` then it is not an option to keep it. It **must** not be selected because it no longer exists.

So the type of `current_path` should be `Option<(transport:Addr, Duration)>`. This has an impact on `select_best_path` as well.

flub (changes requested):
(need to dismiss my earlier review)

iroh/src/magicsock/remote_map/remote_state.rs
dignifiedquire: done

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -932,6 +965,231 @@ impl RemoteStateActor {
...
+fn select_best_path(
+    current_path: Option<(transports::Addr, Duration)>,
+    direct_path_ipv4: Option<(SocketAddrV4, Duration)>,
+    direct_path_ipv6: Option<(SocketAddrV6, Duration)>,
+    new_relay_path: Option<(transports::Addr, Duration)>,
+) -> Option<(transports::Addr, Duration)> {
+    let (current_path, current_rtt) = current_path.unzip();
+    match (current_path, direct_path_ipv4, direct_path_ipv6) {
+        (
+            Some(transports::Addr::Ip(SocketAddr::V4(addr_v4))),
```
flub: I think it might be better to not unzip `current_path` and instead match have `Some((transports:Addr::Ip(SocketAddr::V4(current_addr_v4)), current_rtt)` here?

iroh/src/magicsock/remote_map/remote_state.rs
dignifiedquire: I am fine making the numbers the same, but I wanted to explicitly make them different constants at least, to indicate that we are comparing different things

iroh/src/magicsock/remote_map/remote_state.rs
dignifiedquire: hopefully finally fixed..

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -932,6 +965,231 @@ impl RemoteStateActor {
...
+            Some((new_direct_addr, new_rtt)),
+            None,
+        ) => {
+            // IPv4 -> IPv4, no IPv6
+            if let Some(current_rtt) = current_rtt {
+                if current_rtt > new_rtt + RTT_SWITCHING_MIN_IP && addr_v4 != new_direct_addr {
```
flub: I find `addr_v4 != new_direct_addr` also weird because if that is true `current_rtt` has to be the same as `new_rtt` because they come from the same hashmap[1]. And this way of expressing doesn't really make that logic obvious. I think you can remove that comparison entirely?

[1] I guess a downside of this function signature for `select_best_path` is that this isn't really clear. And worse, a future refactor could break that.

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -932,6 +965,231 @@ impl RemoteStateActor {
...
+                }
+            } else {
+                // Use new path, we have no RTT information for the current path
+                Some((transports::Addr::Ip(new_direct_addr.into()), new_rtt))
+            }
+        }
+        (Some(transports::Addr::Ip(SocketAddr::V4(_))), None, Some((new_direct_addr, new_rtt))) => {
+            // IPv4 -> IPv6, no IPv4
+            if let Some(current_rtt) = current_rtt {
+                if current_rtt > new_rtt + IPV6_RTT_ADVANTAGE {
```
flub: The IPV6_RTT_ADVANTAGE is added to the IPv6 RTT rather than to the IPv4 RTT. I think that's wrong.

flub: You also want to check the switching threshhold I think:
`current_rtt + IPV6_RTT_ADVANTAGE > new_rtt + RTT_SWITCHING_MIN_IP`

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -932,6 +965,231 @@ impl RemoteStateActor {
...
+            }
+        }
+        (
+            Some(transports::Addr::Ip(SocketAddr::V6(addr_v6))),
+            None,
+            Some((new_direct_addr, new_rtt)),
+        ) => {
+            // IPv6 -> IPv6, no IPv4
+            if let Some(current_rtt) = current_rtt {
+                if current_rtt > new_rtt + RTT_SWITCHING_MIN_IP && addr_v6 != new_direct_addr {
```
flub: here the addr comparison must also be irrelevant for it to be correct.

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -932,6 +965,231 @@ impl RemoteStateActor {
...
+                }
+            } else {
+                // Use new path, we have no RTT information for the current path
+                Some((transports::Addr::Ip(new_direct_addr.into()), new_rtt))
+            }
+        }
+        (Some(transports::Addr::Ip(SocketAddr::V6(_))), Some((new_direct_addr, new_rtt)), None) => {
+            // IPv6 -> IPv4, no IPv6
+            if let Some(current_rtt) = current_rtt {
+                if current_rtt > new_rtt + RTT_SWITCHING_MIN_IP {
```
flub: `current_rtt > new_rtt + IPV6_RTT_ADVANTAGE + RTT_SWITCHING_MIN_IP` I think

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -932,6 +965,231 @@ impl RemoteStateActor {
...
+        }
+        (
+            Some(transports::Addr::Ip(_)),
+            Some((new_direct_addr_v4, new_rtt_v4)),
+            Some((new_direct_addr_v6, new_rtt_v6)),
+        ) => {
+            // v4/v6, new v4 and new v6
+            if let Some(current_rtt) = current_rtt {
+                if current_rtt >= new_rtt_v4 + RTT_SWITCHING_MIN_IP
+                    || current_rtt >= new_rtt_v6 + RTT_SWITCHING_MIN_RELAY
```
flub: this is missing the IPV6_RTT_ADVANTAGE that may need to be added to current_rtt depending on whether the current rtt is from a v4 or a v6 addr.

PR #3748, opened 2025-12-08
feat: basic holepunch metrics

### LINKED ISSUE #3695: Add holepunching metrics
It is probably possible for a client to add these metrics:

- holepunch_attempt: counter
- holepunch_success: counter

But it might takes some careful figuring out, and should be done on the QNT branch because it'd be wasted on the disco-hybrid

Frando:
## Description

This adds holepunching metrics, a few are because they are easy and give some insight. The main metrics are:

- `num_conns_opened`: an existing metric counting the number of connections opened.
- `num_conns_direct`: a new metric counting the number of connections which have ever been direct.

Together those do not give a direct count of the number of successful holepunches. But does give an accurate count of number of failed holepunches.

The `holepunch_attempts` metric is much more of an internal value that does not give direct information.

There are two more metrics:
- `paths_direct`: number of paths that are direct
- `paths_relay`: number of paths that are relayed

If `paths_direct_total < paths_relay_total` then you have some failed holepunching. The number will not be as accurate as using `num_conns_opened_total` and `num_conns_direct_total` however and is also more indicative of the total number of paths involved.

Fixes #3695

## Breaking Changes

Some metrics are removed, but these have not been released.

## Notes & open questions

One could also argue that you could count the "total" and "direct" **remotes**. This gives you a less-inflated number on how successful hoplepunching is. Though we really give an indication of the number of **failed** holepunches instead, so the over-inflation doesn't happen really.

The main reason I consider per-remote counting worse however is that we forget all state about a remote when a `RemoteStateActor` is stopped. This can then later come back and again result in wrong numbers.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh/src/magicsock/metrics.rs
```diff
@@ -30,8 +30,10 @@ pub struct Metrics {
     /*
      * Holepunching metrics
      */
-    /// The number of NAT traversal attempts initiated.
-    pub nat_traversal: Counter,
+    /// The number of remote endpoints for which of NAT traversal was initiated.
+    pub remote_holepunch_attempts: Counter,
```
flub: The "remote_" prefix makes me think this was initiated by the remote.

Essentially the metrics you added are:
- remotes connected to ("remotes_connection_established"?)
- remotes holepunched ("remotes_connection_direct"?)

IIUC. This could be useful I guess. Not sure, it's also a bit weird (see separate comment).

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -671,6 +675,9 @@ impl RemoteStateActor {
                     ?local_candidates,
                     ?remote_candidates,
                 );
+                if self.last_holepunch.is_none() {
+                    self.metrics.remote_holepunch_attempts.inc();
+                }
```
flub: I think the counter you removed is much more helpful. All the logic in trigger_holepunching should not call this function if it is not needed. So every time we call this function increment an attempt counter should be enough.

Though it is really hard to match up attempts with successes. I'm not sure it's possible right now even. Can't believe I naively wrote that down in the issue 😅

flub:
Help, I find this very hard to think about. What do we really want to measure here? I don't have a good answer to this so find commenting on the PR hard. @Arqu maybe has input on what we really need to have?

Things that are **easy** to measure:
- number of times we start holepuching
- number of times we explicitly open a path without holepunching (e.g. because it was holepunched in another connection).
- number of times a path was opened for any reason (e.g. on the server side or because of holepunching or because of explicit open_path).
- number of times we abandon a path.
- number of times a path is abandoned.


Things that are **hard**(er) to measure:
- number of times a path is opened from nat traversal (essentially shown in this PR currently though you also would have to be sure it wasn't from an explict open_path call which is not easy right now)
- number of successful holepunches
- number of failed holepunches

My inclination is to concentrate on the easy ones, but do you learn anything useful from those? You can't do any maths to tell you how good your holepunching is I think, so what's the point even? This goes back to, what are we trying to measure?

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -775,6 +782,16 @@ impl RemoteStateActor {
...
                         .insert_open_path(path_remote.clone(), Source::Connection { _0: Private });
+
+                    // Check if this path is the result of a holepunch attempt, and if so update metrics.
+                    if !self.has_holepunched
+                        && let transports::Addr::Ip(ip_addr) = &path_remote
+                        && let Some(hp) = self.last_holepunch.as_ref()
+                        && hp.remote_candidates.contains(ip_addr)
+                    {
+                        self.has_holepunched = true;
+                        self.metrics.remote_holepunch_success.inc();
```
flub: One issue with this impl is that this metric would only be incremented on the client, and not the server. But that's a bit counter-intuitive from the user point of view.

flub:
So I think the original intention was that you could have two metrics like:
- no of connections that were on relay and had holepunch attempts
- no of connections that moved from relay to direct after holepunching

And now you can do maths on them because you can compute the number of connections that did not holepunch and have a holepunching rate. But these metrics fall under the "hard" metrics :disappointed: I'm not sure how to do them.

Arqu:
I dropped the ball a bit here. Think @flub is on the right path here. 
I guess a number of things are here in play. Main question is what and for whom we're measuring things here. Iroh side metrics should generally serve the user and should be concerned with the performance of "their" node respectively.

That said I think it's fine that this is client centric and whether _my_ endpoint managed to holepunch. Still might be slightly counter intuitive as it basically tells you more about the state of the other side network config. Ideally we would be aware of a hole punching success towards ourselves too. 

The 2nd bit and more crucially, I don't think this should work to establish a network wide holepunch rate. You can fumble some maths and see some grouped averages or distributions between clients and their holepunch rates but nothing definitive. I intend to do a follow up PR that does a similar thing but from the Relay PoV which should help make more informed guesstimates on the network level. Still pretty hard though as we are operating with very little info.

For this PR I'd focus on getting the "easy" ones @flub listed as they give a lot of flexibility to do math and explore the state. However none of these counters I would limit to a 0-1 state but just let them increment naturally on what we deem the relevant event. It's easy enough to do the subtractions to get the actual state from it.

Frando:
LGTM, let's merge this!  (Can't GitHub approve because it's my PR)

We can iterate further on this until 1.0 but this gives us a baseline to try things out, so let's get this in.

PR #3752, opened 2025-12-09
feat: expose known remote addrs

### LINKED ISSUE #3521: Provide a way to get the latest dialing information for a node id
It should be possible to get the latest dialing info for a node id, to pass on to other nodes etc.

This used to be possible with [remote_info](https://docs.rs/iroh/0.92.0/iroh/endpoint/struct.Endpoint.html#method.remote_info), but that fn has been removed in iroh 0.93.

The concrete use case is in iroh-dht-experiment, where I want to enrich a set of node ids with current dialing info to reduce load on the discovery services.

https://github.com/n0-computer/iroh-dht-experiment/blob/561cad3740d3f13aa1b49f5d0b79c5198c1bd9a5/src/lib.rs#L1181

Frando:
## Description

Adds `Endpoint::remote_info` which returns `Option<RemoteInfo>`. The `RemoteInfo` exposes an iterator over all transport addrs for this remote that we know about. 

Fixes #3521

## Notes & open questions

This basically copies the `RemotePathInfo` from the RemoteStateActor into a new struct `RemoteInfo`. So, this an async fn over the channel to the RemoteStateActor, and allocates. As this is a rather niche API, I think that's fine.

I wasn't entirely sure how much info we want to expose. Right now, the PR only includes the latest `Source` for the address, as I wasn't sure if we really want or need to expose the full list of sources.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -23,31 +23,6 @@ use n0_watcher::Watcher;
 use tracing::{debug, instrument, trace, warn};
 use url::Url;
 
-use self::hooks::EndpointHooksList;
-pub use super::magicsock::{
-    DirectAddr, DirectAddrType, PathInfo,
-    remote_map::{PathInfoList, Source},
-};
-#[cfg(wasm_browser)]
-use crate::discovery::pkarr::PkarrResolver;
-#[cfg(not(wasm_browser))]
-use crate::dns::DnsResolver;
-use crate::{
-    discovery::{ConcurrentDiscovery, DiscoveryError, DynIntoDiscovery, IntoDiscovery, UserData},
-    endpoint::presets::Preset,
-    magicsock::{self, Handle, RemoteStateActorStoppedError, mapped_addrs::MappedAddr},
-    metrics::EndpointMetrics,
-    net_report::Report,
-    tls::{self, DEFAULT_MAX_TLS_TICKETS},
-};
-
-mod connection;
-pub(crate) mod hooks;
-pub mod presets;
-mod quic;
-
-pub use hooks::{AfterHandshakeOutcome, BeforeConnectOutcome, EndpointHooks};
-
 #[cfg(feature = "qlog")]
 pub use self::quic::{QlogConfig, QlogFactory, QlogFileFactory};
 pub use self::{
@@ -57,6 +32,7 @@ pub use self::{
         IncomingZeroRttConnection, OutgoingZeroRtt, OutgoingZeroRttConnection,
         RemoteEndpointIdError, ZeroRttStatus,
     },
+    hooks::{AfterHandshakeOutcome, BeforeConnectOutcome, EndpointHooks},
     quic::{
         AcceptBi, AcceptUni, AckFrequencyConfig, AeadKey, ApplicationClose, Chunk, ClosedStream,
         ConnectionClose, ConnectionError, ConnectionStats, Controller, ControllerFactory,
@@ -68,7 +44,28 @@ pub use self::{
         VarIntBoundsExceeded, WeakConnectionHandle, WriteError, Written,
     },
 };
-pub use crate::magicsock::transports::TransportConfig;
+use self::{hooks::EndpointHooksList, presets::Preset};
+#[cfg(wasm_browser)]
+use crate::discovery::pkarr::PkarrResolver;
+#[cfg(not(wasm_browser))]
+use crate::dns::DnsResolver;
+pub use crate::magicsock::{
+    DirectAddr, DirectAddrType, PathInfo,
+    remote_map::{AddrUsage, PathInfoList, RemoteAddr, RemoteInfo, Source},
+    transports::TransportConfig,
+};
+use crate::{
+    discovery::{ConcurrentDiscovery, DiscoveryError, DynIntoDiscovery, IntoDiscovery, UserData},
+    magicsock::{self, Handle, RemoteStateActorStoppedError, mapped_addrs::MappedAddr},
+    metrics::EndpointMetrics,
+    net_report::Report,
+    tls::{self, DEFAULT_MAX_TLS_TICKETS},
+};
+
+mod connection;
+pub(crate) mod hooks;
+pub mod presets;
+mod quic;
 
 /// Builder for [`Endpoint`].
 ///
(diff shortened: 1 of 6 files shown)


iroh/src/endpoint.rs
```diff
@@ -23,31 +23,23 @@ use n0_watcher::Watcher;
 use tracing::{debug, instrument, trace, warn};
 use url::Url;
 
-use self::hooks::EndpointHooksList;
```
Frando, line 26: The import ordering was off, and when I changed a line, `cargo make format` picked it up again. The only actual change here is that I added the re-exports from the new `remote_info` module.

dignifiedquire (approved):
given how easy it is to implement this, and how useful this is for users, I would say, lets do this

iroh/src/endpoint.rs
```diff
@@ -1060,6 +1057,18 @@ impl Endpoint {
         &self.msock.metrics
     }
 
+    /// Returns addressing information about a recently used remote endpoint.
```
dignifiedquire, line 1063: maybe mention that this is a snapshot in time, or sth, to indicate that this can have already changed by the moment it is used

iroh/src/magicsock/remote_map/remote_state/path_state.rs
```diff
@@ -62,6 +62,27 @@ impl RemotePathState {
             metrics,
         }
     }
+    pub(super) fn to_remote_addrs(&self) -> Vec<TransportAddrInfo> {
```
flub, line 66: missing blank line? (how does rustfmt not handle this?)

iroh/src/magicsock/remote_map/remote_state/remote_info.rs
```diff
@@ -0,0 +1,97 @@
+use std::collections::BTreeSet;
+
+use iroh_base::{EndpointAddr, EndpointId, TransportAddr};
+use n0_future::time::Instant;
+
+use crate::endpoint::Source;
+
+/// Information about a remote endpoint.
```
flub, line 3: I think this was already mentioned, but this should explain that this is a snapshot at the time it was created.

iroh/src/magicsock/remote_map/remote_state/remote_info.rs
```diff
@@ -0,0 +1,97 @@
...
+
+/// Information how a transport address is used.
+#[derive(Debug, Copy, Clone)]
+pub enum TransportAddrUsage {
+    /// The address is in active use.
+    Active,
+    /// The address was used, but is not currently.
+    Inactive {
+        /// Time when this address was last used.
+        last_used: Instant,
```
flub: I believe this is still problematic? I was hopeing to keep postponing having to figure this out to be fair.

iroh/src/magicsock/remote_map/remote_state/remote_info.rs
```diff
@@ -0,0 +1,97 @@
...
+    },
+    /// We tried to use this address, but failed.
+    Unusable,
```
flub: I am not comfortable exposing this to be fair. It forces us to keep this status around and I don't think I want to do that.

I think for now I'm at most comfortable with exposing two items on this enum:
- Active
- Inactive

No timestamps, nothing else.

iroh/src/magicsock/remote_map/remote_state/remote_info.rs
```diff
@@ -0,0 +1,97 @@
...
+        self.usage
+    }
+
+    /// Returns the most recent source of this address.
+    ///
+    /// We may learn about new addresses from multiple sources. This returns the most recent source
+    /// that told us about this address.
+    pub fn most_recent_source(&self) -> &Source {
+        &self.most_recent_source
+    }
```
flub: I'm also not comfortable exposing this yet. Source is a number of states that I don't know we'll always want to keep and want to keep around as such. Also, there is not a single source for an addr, certainly not the way it is used currently. Which makes this messy and indicates that it's not a great thing to expose.

Source **is** public right now, but only because of discovery methods. I had an issue originally to figure something out about that. This was solved in https://github.com/n0-computer/iroh/commit/50fdda393a9a58c352cec05c859d603edbbae030 in a way that seemed to solve the immediate issue. But I do now think it was probably the wrong approach, given the variants on that enum. I don't think this is suitable for a 1.0 API.

iroh/src/magicsock/remote_map.rs
```diff
@@ -118,6 +120,17 @@ impl RemoteMap {
...
+    pub(super) fn remote_state_actor_if_exists(
+        &self,
+        eid: EndpointId,
+    ) -> Option<mpsc::Sender<RemoteStateMessage>> {
+        self.actor_handles
+            .lock()
+            .expect("poisoned")
+            .get(&eid)
+            .and_then(|handle| handle.sender.get())
+    }
```
flub, line 132: Do you think this is wroth having? The actor would be started and shut down again. I guess that's not ideal. I'm just sad whenever there are two functions like this, one that creates-on-demand and one that returns an Option. I can live with it I guess.

flub (commented):
As usual I'm anxious about creating any API surface and am arguing to make it smaller 😁 😅 😰

iroh/src/magicsock/remote_map/remote_state/remote_info.rs
Frando: Alright. I changed it to only expose Active/Inactive. Though I do think that we might want to add "Unknown" or "Unusable" at one point.

iroh/src/magicsock/remote_map/remote_state/remote_info.rs
Frando: Yeah, I was quite unsure myself too. I removed it. We can always add something like it back later.

iroh/src/magicsock/remote_map.rs
Frando: I don't think we should start RemoteStateActors when someone calls Endpoint::remote_info. It is very valid to call remote_info with *any* endpoint id, so I think we should not do unneeded work in the case we don't know anything about the remote.

iroh/src/magicsock/remote_map/remote_state/remote_info.rs
Frando: Removed the instant.

iroh/src/magicsock/remote_map/remote_state/remote_info.rs
Frando: Added this.

flub (approved):

PR #3757, opened 2025-12-10
feat(iroh)!: newtype `ServerConfig` and `RetryError`, organize the quinn re-exports

ramfox:
## Description

`Incoming::retry` returns a `RetryError`, which you can turn back into a `quinn::Incoming`. We want that `quinn::Incoming` to instead be an `iroh::endpoint::Incoming`. This PR fixes that.

`Incoming::accept_with` takes a `quinn::ServerConfig`, however the `quinn::ServerConfig` allows you to override the `quinn::TransportConfig`, which we have new typed to `QuicTransportConfig`. So, this PR new types `quinn::ServerConfig` to `quic::ServerConfig`, and ensures that the `quinn::TransportConfig` is set appropriately for multipath use.

This PR also explicitly comments on each quinn export and labels what type uses that export and where it enters the public iroh API.

## Breaking Changes

`iroh`:
      - changed
          - `Incoming::retry` now returns an `iroh::endpoint::RetryError`, rather than a `quinn::RetryError`
          -  `Incoming::accept_with` now takes a `iroh::endpoint::ServerConfig`, rather than a `quinn::ServerConfig`
          -  `iroh::endpoint::QuicTransportConfig` has no more "set" methods, those have been moved to `iroh::endpoint::QuicTransportConfigBuilder`
          - `iroh:endpoint::QuicTransportConfig::builder()` returns a default `iroh::endpoint:: QuicTransportConfigBuilder`
     - removed  
          - removed `quinn::ServerConfig` (use `iroh::endpoint::ServerConfig`)
          - removed `quinn::RetryError` (use `iroh::endpoint::RetryError`)
    - added
          - added `iroh::endpoint::ServerConfigBuilder` 
          - added `iroh::endpoint::ServerConfig`
          - added `iroh::endpoint::RetryError`
          
     

## Notes & open questions

Okay. So. I impl'd all of the methods one-to-one from `quinn::ServerConfig`, but I think there are a bunch of fields we may want to restrict folks from messing with when they use `Incoming::accept_with`, as currently, they can use this API to replace our normal TlsConfig entirely.

I'm also unsure if there are any other bits in here (for example the `quinn::ServerConfig::migrate` field) that may interfere with expected multipath usage.

My proposal would be to either eliminate the `Incoming::accept_with` method entirely (therefore removing the need to publically export the `ServerConfig`), or reduce `ServerConfig` in scope so that only fields that will not interfere with "business as usual" for multipath/iroh remain.

## Change checklist

- [x] Self-review.


iroh/src/endpoint/quic.rs
```diff
@@ -484,3 +543,177 @@ impl QuicTransportConfig {
...
+/// Default values should be suitable for most internet applications.
+// Note: used in `iroh::endpoint::connection::Incoming::accept_with`
+// This is new-typed since `quinn::ServerConfig` takes a `TransportConfig`, which we new-type as a `QuicTransportConfig`
+#[derive(Debug, Clone)]
+pub struct ServerConfig {
+    inner: quinn::ServerConfig,
+    transport: QuicTransportConfig,
+}
+
+impl ServerConfig {
```
ramfox, line 602: @flub do you think you can look these methods over and see if there are any we want to restrict?

I added more context to the PR description, but in summary, `Incoming::accept_with` allows the user to pass in their own `ServerConfig`. However, many of these fields may really disrupt iroh/multipath usage (eg, you can replace our normal TlsConfig).

I'm wondering if we:
a) want to allow users to use `Incoming::accept_with` at all
b) or restrict what iroh allows users to mess with when using their own `ServerConfig`

iroh/src/endpoint/connection.rs
```diff
@@ -103,12 +105,9 @@ impl Incoming {
     /// See [`accept()`] for more details.
     ///
     /// [`accept()`]: Incoming::accept
-    pub fn accept_with(
-        self,
-        server_config: Arc<ServerConfig>,
-    ) -> Result<Accepting, ConnectionError> {
+    pub fn accept_with(self, server_config: ServerConfig) -> Result<Accepting, ConnectionError> {
```
flub: Curious why the signature change? It does make sense to put the Arc into the user's control. Usually you'd call this many times with the same config and thus would be forced to clone the ServerConfig a lot.

iroh/src/endpoint/connection.rs
```diff
@@ -522,7 +541,7 @@ impl Accepting {
     ///
     /// See also documentation for [`Connecting::into_0rtt`].
     ///
-    /// [`RecvStream::is_0rtt`]: quinn::RecvStream::is_0rtt
+    /// [`RecvStream::is_0rtt`]: crate::endpoint::RecvStream::is_0rtt
```
flub, line 554: ❤️

iroh/src/endpoint/quic.rs
```diff
@@ -8,25 +8,84 @@
...
-use std::{sync::Arc, time::Duration};
+use std::{
+    net::{SocketAddrV4, SocketAddrV6},
+    sync::Arc,
+    time::Duration,
+};
 
-// Missing still: SendDatagram and ConnectionClose::frame_type's Type.
+/// `quinn` types that are used in the public iroh API.
+// Each type is notated with the iroh type or quinn type that uses it.
```
flub, line 14: ❤️

iroh/src/endpoint/quic.rs
flub: I think we can't remove `Incoming::accept_with` because it is helpful to be able to set a custom TransportConfig, just like we allow when connecting. So I guess we go with b.

iroh/src/endpoint/quic.rs
```diff
@@ -484,3 +543,177 @@ impl QuicTransportConfig {
...
+    pub fn crypto(&self) -> Arc<dyn CryptoServerConfig> {
+        self.inner.crypto.clone()
+    }
+
+    /// Configuration for sending and handling validation tokens
+    pub fn validation_token(&self) -> ValidationTokenConfig {
+        self.inner.validation_token.clone()
+    }
+
+    pub(crate) fn to_arc(&self) -> Arc<quinn::ServerConfig> {
```
flub: This is internal so it doesn't matter a lot and is just a nit. But the way it is named I did not expect the type change. It seems like the type change is the important bit.

Maybe `to_inner_arc`?.

iroh/src/endpoint/quic.rs
```diff
@@ -484,3 +543,177 @@ impl QuicTransportConfig {
...
+// Note: used in `iroh::endpoint::connection::Incoming::accept_with`
+// This is new-typed since `quinn::ServerConfig` takes a `TransportConfig`, which we new-type as a `QuicTransportConfig`
+#[derive(Debug, Clone)]
+pub struct ServerConfig {
+    inner: quinn::ServerConfig,
+    transport: QuicTransportConfig,
+}
+
+impl ServerConfig {
+    /// Transport configuration to use for incoming connections
```
flub:
```suggestion
    /// Transport configuration used for incoming connections.
```
?

iroh/src/endpoint/quic.rs
```diff
@@ -484,3 +543,177 @@ impl QuicTransportConfig {
...
+    pub fn transport(&self) -> QuicTransportConfig {
+        self.transport.clone()
+    }
+
+    /// TLS configuration used for incoming connections
```
flub:
```suggestion
    /// TLS configuration used for incoming connections.
```
😄

iroh/src/endpoint/quic.rs
```diff
@@ -484,3 +543,177 @@ impl QuicTransportConfig {
...
+    ///
+    /// Must be set to use TLS 1.3 only.
```
flub: Since we don't allow users to modify this (see below) but only read it this bit of information is not helpful. Probably better to remove it.

iroh/src/endpoint/quic.rs
```diff
@@ -484,3 +543,177 @@ impl QuicTransportConfig {
...
+    pub fn validation_token(&self) -> ValidationTokenConfig {
+        self.inner.validation_token.clone()
+    }
+
+    pub(crate) fn to_arc(&self) -> Arc<quinn::ServerConfig> {
+        Arc::new(self.inner.clone())
+    }
+
+    /// Create a default config with a particular handshake token key
+    pub fn new(crypto: Arc<dyn CryptoServerConfig>, token_key: Arc<dyn HandshakeTokenKey>) -> Self {
```
flub: My inclination would be to forbid people from customising the `ServerConfig::crypto` entirely. Accessing it via the `pub fn crypto` function above is fine I think.

iroh/src/endpoint/quic.rs
```diff
@@ -484,3 +543,177 @@ impl QuicTransportConfig {
...
+        let mut inner = quinn::ServerConfig::new(crypto, token_key);
+        let transport = QuicTransportConfig::default();
+        inner.transport_config(transport.to_arc());
+        Self { inner, transport }
+    }
+
+    /// Set a custom [`QuicTransportConfig`]
```
flub:
```suggestion
    /// Sets a custom [`QuicTransportConfig`].
```
because I can't help the nitpicking

iroh/src/endpoint/quic.rs
```diff
@@ -484,3 +543,177 @@ impl QuicTransportConfig {
...
+    pub fn transport_config(&mut self, transport: QuicTransportConfig) -> &mut Self {
```
flub: Can we make these methods more symmetric?  e.g.
- `transport_config`
- `set_transport_config`

Currently it's `transport` and `transport_config` which is a bit odd and inconsistent. Similar for most other getter/setter pairs.

iroh/src/endpoint/quic.rs
flub: Looking at the default code that creates this it seems that the `token_key` is normally 64 bytes of randomness put into `ring::hkdf::Salt::new(hkdf::HKDF_SHA256, &[]).extract(&the_64_random_bytes)`. It feels like a tall order to ask folks to figure this out, though on the other hand it's a suitably high barrier to using this API which could be helpful.

Anyway, point is I think it's fine to allow folks to customise this value.

iroh/src/endpoint/quic.rs
```diff
@@ -484,3 +543,177 @@ impl QuicTransportConfig {
...
+        self
+    }
+
+    /// The preferred IPv6 address that will be communicated to clients during handshaking
+    ///
+    /// If the client is able to reach this address, it will switch to it.
+    pub fn preferred_address_v6(&mut self, address: Option<SocketAddrV6>) -> &mut Self {
+        self.inner.preferred_address_v6(address);
+        self
+    }
```
flub: let's remove these two and always have them as `None` for us. Maybe one day it's possible. But there are loads of things to figure out for that.

iroh/src/endpoint/quic.rs
```diff
@@ -484,3 +543,177 @@ impl QuicTransportConfig {
...
+
+    /// Create a server config with the given [`CryptoServerConfig`]
+    ///
+    /// Uses a randomized handshake token key.
+    pub fn with_crypto(crypto: Arc<dyn CryptoServerConfig>) -> Self {
+        let mut inner = quinn::ServerConfig::with_crypto(crypto);
+        let transport = QuicTransportConfig::default();
+        inner.transport_config(transport.to_arc());
+        Self { inner, transport }
+    }
```
flub: Lets remove this one for now, you can create something that isn't iroh anymore. Though you *can* also do interesting stuff, but let's handle those as feature requests.

iroh/src/endpoint/quic.rs
flub: Ah, seems like `with_crypto` was the API that allowed you to bypass constructing that `token_key`. We could make `new` without any arguments and have a `with_token_key` constructor that takes the `token_key` I guess.

flub (commented):
Really nice to get all the imports cleaned up, thank you for slogging through that!

iroh/src/endpoint/quic.rs
ramfox: Okay, so we can't just create a `ServerConfig` using a `new` without any arguments from some outside source. To allow users to create a `StaticConfig` that doesn't totally bork iroh, I added a method on the `Endpoint` called `create_server_config` (bikeshed welcome). It uses the internal `StaticConfig` that contains the TLS options and default transport config options, but expects the method caller to supply alpns.

That way a user can then replace the bits we want to allow them to replace, without replacing expected iroh crypto.

iroh/src/endpoint/quic.rs
ramfox: I went through and added periods to all of the copy-pasted methods 😂

iroh/src/endpoint/connection.rs
ramfox: yea that's a good point and I will add it back, since it does save some resources, however I removed it because our new-typing kind of made cloning the inner `quinn::ServerConfig` necessary.

We need to have a non-arc'ed `quinn::ServerConfig` inside our `ServerConfig` so that we can edit it, so under the hood we have to clone the `quinn::ServerConfig` anyway 🙃 

If I'm thinking about it the wrong way, I'm happy to be wrong.

iroh/src/endpoint/quic.rs
ramfox: > Ah, seems like with_crypto was the API that allowed you to bypass constructing that token_key. We could make new without any arguments and have a with_token_key constructor that takes the token_key I guess.

Oh, also the `token_key` (now `set_token_key`) method allows setting a different token key already, so I think we have that covered?

iroh/src/endpoint/quic.rs
```diff
@@ -484,3 +539,154 @@ impl QuicTransportConfig {
...
+
+    /// Whether to allow clients to migrate to new addresses.
+    ///
+    /// Improves behavior for clients that move between different internet connections or suffer NAT
+    /// rebinding. Enabled by default.
+    pub fn set_migration(&mut self, value: bool) -> &mut Self {
+        self.inner.migration(value);
+        self
+    }
```
flub: Hum, maybe we can remove this for now too? Maybe someone will find a good reason to do this, but it can be added in a point release if so. For now it's messing with iroh's internals really.

iroh/src/endpoint/quic.rs
```diff
@@ -484,3 +539,154 @@ impl QuicTransportConfig {
...
+
+    /// Configuration for sending and handling validation tokens.
+    pub fn validation_token(&self) -> ValidationTokenConfig {
+        self.inner.validation_token.clone()
+    }
+
+    /// Sets a custom [`QuicTransportConfig`].
+    pub fn set_transport_config(&mut self, transport: QuicTransportConfig) -> &mut Self {
+        self.inner.transport_config(transport.to_inner_arc());
```
flub, line 620: the amount of forced clones here are unfortunate. but i guess we can't do anything about that.

iroh/src/endpoint.rs
```diff
@@ -1137,6 +1144,11 @@ impl Endpoint {
         self.msock.is_closed()
     }
 
+    /// Create a [`ServerConfig`] for this endpoint that includes the given alpns.
```
flub: Maybe give some pointers to where and why you need to use this. I know you already link in the other direction, but it doesn't do any harm linking both ways.

flub (commented):
Upstream's ARC's around the transport config and server config were carefully constructed so that you could do a `accept_with` in a loop with customised config and all you ever had to do was increment the ref count on the Arc. And we now have a full clone of a struct every time. That's pretty unfortunate.

I've not tried to see if it can be solved or not... but it might be worth seeing if that could be possible?

The `Endpoint::create_server_config` method is fine I think, both in location and name.

ramfox:
@flub My only proposed "solution" to the `Arc` problem, would be to have a `QuicTransportConfigBuilder` and `QuicTransportConfig`/`ServerConfigBuilder` and `ServerConfig`.

The builder structs would contain un-arc-ed versions of the inner quinn types, and the built versions would contain arc-d versions that can no longer be edited.

ramfox:
@flub latest commit does the split between builder types for `QuicTransportConfig` and `ServerConfig` and the types that have inner arcs that cannot be edited.

I think it works out fine. What is interesting here is that in quinn those types are pretty opaque, only allowing visibility into very few fields after they've been edited, so it doesn't bloat the API tooooo much.

If you think this is the way to go, I'll add docs.

iroh/src/endpoint/quic.rs
```diff
@@ -104,14 +104,14 @@ use crate::magicsock::{HEARTBEAT_INTERVAL, MAX_MULTIPATH_PATHS, PATH_MAX_IDLE_TI
...
+/// - [`QuicTransportConfigBuilder::default_path_keep_alive_interval`]
+/// - [`QuicTransportConfigBuilder::default_path_max_idle_timeout`]
+/// - [`QuicTransportConfigBuilder::max_concurrent_multipath_paths`]
+/// - [`QuicTransportConfigBuilder::set_max_remote_nat_traversal_addresses`]
 #[derive(Debug, Clone)]
-pub struct QuicTransportConfig(pub(crate) quinn::TransportConfig);
+pub struct QuicTransportConfigBuilder(quinn::TransportConfig);
 
-impl Default for QuicTransportConfig {
+impl Default for QuicTransportConfigBuilder {
```
flub: Now it's a builder a `new` method that does exactly the same as `default` would make sense.

iroh/examples/transfer.rs
```diff
@@ -265,9 +265,9 @@ impl EndpointArgs {
 
         #[cfg(feature = "qlog")]
         {
-            let mut transport_config = iroh::endpoint::QuicTransportConfig::default();
-            transport_config.qlog_from_env("transfer");
-            builder = builder.transport_config(transport_config)
+            let mut cfg_builder = iroh::endpoint::QuicTransportConfigBuilder::default();
+            cfg_builder.qlog_from_env("transfer");
```
flub: I think you can now chain this easily:
```rust
let cfg_builder = iroh::endpoint::QuicTransportconfigBuilder::new()
    .qlog_from_env("transer");
```
(I'm assuming `new` exists there) This is totally nitpicking, but for examples trying to write the most idiomatic thing might be worth it.

flub (commented):
Agreed the builder works quite well for this! I'm fine with using this approach.

iroh/bench/src/iroh.rs
```diff
@@ -125,18 +128,18 @@ pub async fn connect_client(
 pub fn transport_config(max_streams: usize, initial_mtu: u16) -> QuicTransportConfig {
     // High stream windows are chosen because the amount of concurrent streams
     // is configurable as a parameter.
-    let mut config = QuicTransportConfig::default();
-    config.max_concurrent_uni_streams(max_streams.try_into().unwrap());
-    config.initial_mtu(initial_mtu);
+    let mut builder = QuicTransportConfigBuilder::default();
```
dignifiedquire: maybe have `QuicTransportConfig::builder()` be a thing, requires one less type to import for users

iroh/src/endpoint/quic.rs
```diff
@@ -484,3 +555,167 @@ impl QuicTransportConfig {
...
+        self.inner.crypto.clone()
+    }
+
+    /// Configuration for sending and handling validation tokens.
+    pub fn validation_token(&self) -> ValidationTokenConfig {
+        self.inner.validation_token.clone()
+    }
+
+    /// Sets a custom [`QuicTransportConfig`].
+    pub fn set_transport_config(&mut self, transport: QuicTransportConfig) -> &mut Self {
```
dignifiedquire: why doesn't this return `Self`?

dignifiedquire: and consume `self` like whave with other builders

iroh/src/endpoint/quic.rs
```diff
@@ -484,3 +555,167 @@ impl QuicTransportConfig {
...
+    pub fn build(self) -> ServerConfig {
+        ServerConfig(Arc::new(self.inner))
+    }
+
+    pub(crate) fn new(inner: quinn::ServerConfig, transport: QuicTransportConfig) -> Self {
+        Self { inner, transport }
+    }
+
+    /// Transport configuration used for incoming connections.
+    pub fn transport_config(&self) -> QuicTransportConfig {
```
dignifiedquire: are these accessor methods needed/used?

iroh/src/magicsock.rs
```diff
@@ -1571,12 +1571,13 @@ mod tests {
...
 
     /// Generate a server config with no ALPNS and a default transport configuration
-    fn make_default_server_config(secret_key: &SecretKey) -> ServerConfig {
+    fn make_default_server_config(secret_key: &SecretKey) -> quinn::ServerConfig {
         let quic_server_config =
             crate::tls::TlsConfig::new(secret_key.clone(), DEFAULT_MAX_TLS_TICKETS)
                 .make_server_config(vec![], false);
-        let mut server_config = ServerConfig::with_crypto(Arc::new(quic_server_config));
-        server_config.transport_config(Arc::new(quinn::TransportConfig::default()));
+        let mut server_config = quinn::ServerConfig::with_crypto(Arc::new(quic_server_config));
```
dignifiedquire, line 1590: do we still need to use `quinn::` here?

iroh/src/endpoint/quic.rs
ramfox: quinn used this pattern and I was keeping it closer to the original API, but agreed, now that it's explicitly a builder it should be changed to be more rust idiomatic.

iroh/src/endpoint/quic.rs
ramfox: tbh, unclear, but in quinn these fields are public and so can be accessed. Happy to take them out because they would reduce the surface area of exported quinn types quite a bit.

iroh/src/magicsock.rs
ramfox: Yes, `magicsock` uses the `quinn::ServerConfig`. Only devs will use our version, when using the `Incoming::accept_with` method.

I made the choice, right or wrong, to explicitly use `quinn::Foo` when we directly use the quinn API privately, so that it's clear it was intentional in future refactors.

iroh/src/endpoint/quic.rs
dignifiedquire: lets remove them for now, we can always add them if needed

iroh/src/magicsock.rs
dignifiedquire: sounds good

ramfox:
Okay so latest changes:
- `QuicTransportConfig` has a `builder` method that returns a `QuicTransportConfigBuilder`
- `QuicTransportConfigBuilder` expects you to use the above method, so it no longer has a public `new` nor a `Default` impl
- `ServerConfig` has a `builder` method that returns a `ServerConfigBuilder` (for completeness: you can only build a `ServerConfigBuilder` using the `Endpoint::create_server_config_builder` method)
- the accessor functions on `ServerConfig` and `ServerConfigBuilder` are removed, allowing us to remove the `crypto::ServerConfig` as `CryptoServerConfig` quinn exports, as well as `ConnectionId` and `Session`, that were in the public api of `crypto::ServerConfig`

dignifiedquire:
thanks for doing this, the builder code reads so much nicer now

dignifiedquire (approved):

PR #3758, opened 2025-12-10
refactor(iroh)!: make net-report private

ramfox:
## Description

We no longer use net-report in `iroh-doctor`, so we can make this private now and slim down the public iroh API.

## Breaking Changes

- `iroh`
    - removed
       - `net_report` module is now private
       - `iroh::net_report::Report` is now exported as `iroh::NetReport`
       - `iroh::net_report::Options` is no longer public (`pub(crate)` instead)
       - `iroh::net_report::QuicConfig` is no longer public (`pub(crate)` instead)

### DIFF
--- a/iroh/src/net_report/options.rs
+++ b/iroh/src/net_report/options.rs
@@ -1,6 +1,6 @@
 //! Options for creating a report gen client.
 
-pub use imp::Options;
+pub(crate) use imp::Options;
 
 #[cfg(not(wasm_browser))]
 mod imp {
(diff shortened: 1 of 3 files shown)


iroh/src/net_report/options.rs
```diff
@@ -41,6 +41,7 @@ mod imp {
 
     impl Options {
         /// Create an [`Options`] that disables all probes
+        #[allow(dead_code)]
         pub fn disabled() -> Self {
```
flub: Could you limit the visibility of all the things in the net-report module now? E.g. this Options is probably now `pub(crate)` or even `pub(super)`. It's always nice to explicitly have the visibility instead of having to mentally walk the module tree.

Also, if it's dead code let's remove it. These are easy enough to recreate.

flub (approved):
This was less exciting than the title made me believe. But any reduction in public API makes me happy!

LGTM though I'd adjust some more visibility qualifiers. But that's only my personal style really. The only clippy lint I found, redundant_pub_crate, is in fact entirely the opposite 😬

PR #3762, opened 2025-12-10
fix: shutdown block

Arqu:
## Description

Details bellow as PR comments.

Depends on https://github.com/n0-computer/quinn/pull/239

The quinn PR is not in an ideal state and needs more input.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1092,7 +1092,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -2276,7 +2276,7 @@ dependencies = [
 [[package]]
 name = "iroh-quinn"
 version = "0.14.0"
-source = "git+https://github.com/n0-computer/quinn?branch=arqu%2Ffix_cid_panic2#b4fd0a89741a33e6facea4254c661440a571ba84"
+source = "git+https://github.com/n0-computer/quinn?branch=arqu%2Ffix_cid_panic2#71c25ca266b69b79e86429f0dc5fceb4d6051b6f"
 dependencies = [
  "bytes",
  "cfg_aliases",
@@ -2285,7 +2285,7 @@ dependencies = [
  "pin-project-lite",
  "rustc-hash",
  "rustls",
- "socket2 0.5.10",
+ "socket2 0.6.1",
  "thiserror 2.0.17",
  "tokio",
  "tokio-stream",
@@ -2296,7 +2296,7 @@ dependencies = [
 [[package]]
 name = "iroh-quinn-proto"
 version = "0.13.0"
-source = "git+https://github.com/n0-computer/quinn?branch=arqu%2Ffix_cid_panic2#b4fd0a89741a33e6facea4254c661440a571ba84"
+source = "git+https://github.com/n0-computer/quinn?branch=arqu%2Ffix_cid_panic2#71c25ca266b69b79e86429f0dc5fceb4d6051b6f"
 dependencies = [
  "bytes",
  "fastbloom",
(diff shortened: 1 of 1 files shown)


iroh/examples/transfer.rs
```diff
@@ -368,7 +368,7 @@ async fn provide(endpoint: Endpoint, size: u64) -> Result<()> {
 
             // We sent the last message, so wait for the client to close the connection once
             // it received this message.
-            let res = tokio::time::timeout(Duration::from_secs(3), async move {
+            let res = tokio::time::timeout(Duration::from_secs(4), async move {
```
Arqu, line 371: These are needed because our default connection close time is a hair more than 3s (default PTO 333ms, with max calculating to 1.0x seconds and close time is 3x that)

iroh/src/magicsock.rs
```diff
@@ -979,6 +979,11 @@ impl Handle {
     #[instrument(skip_all)]
     pub(crate) async fn close(&self) {
         trace!(me = ?self.public_key, "magicsock closing...");
+
+        // Set closing flag BEFORE wait_idle() to prevent new net_report runs
+        // from creating QAD connections while we're draining existing ones.
+        self.msock.closing.store(true, Ordering::Relaxed);
+
```
Arqu, line 986: We have in flight conns if we close later which prevents closing at all and makes wait idle choke.

iroh/src/magicsock/transports.rs
```diff
@@ -533,7 +536,13 @@ impl TransportsSender {
                 }
             }
         }
-        Poll::Pending
+        if found_transport {
```
Arqu, line 539: Netsim continued to choke on this for the --relay-only tests because apparently when the other transport is disabled it gets stuck here pending forever.

Arqu:
Welp, less aggressive path filtering for PTO calcs seems to be not great.

iroh/src/magicsock/transports.rs
```diff
@@ -533,7 +536,13 @@ impl TransportsSender {
...
+            Poll::Pending
+        } else {
+            Poll::Ready(Err(io::Error::other(format!(
+                "no transport available for {dst:?}"
+            ))))
```
flub, line 544: Is this a permanent error? IIRC in the past we've always decided to return `Poll::Ready(Ok(()))` and let the the QUIC stack use loss detection to figure out this is lost.

The quinn-udp philosophy is that only permanent errors should be errors. That is: this socket will never ever be usable again and won't ever be able to send a datagram and needs to be re-bound to get any progress.

And I think that this is not the case here: if you have one transport you can still send on that transport. So I think this should probably not be an error but rather a `Poll::Ready(Ok(()))`.

It would still be helpful to log this though. But if we naively log this we'd end up with flooding the log. IIRC quinn-udp itself has a similar situation where they add a bit of state to the socket so that they only emit that log once every 60 seconds or something like that. Could we do the same? I'd vote debug-level logging for this.

iroh/src/magicsock/transports.rs
```diff
@@ -494,6 +494,7 @@ impl TransportsSender {
         src: Option<IpAddr>,
         transmit: &Transmit<'_>,
     ) -> Poll<io::Result<()>> {
+        let mut found_transport = false;
```
flub, line 497: If I were to nitpick I'd make this `let mut fallback_ret = Poll::Ready(Ok(()));` and let each transport set this to pending when they hit a pending. But there are many ways to do this and it's subjective and maybe even a nicer way exists.

iroh/src/magicsock/transports.rs
dignifiedquire: > Is this a permanent error? 

yes, transports are not configurable at runtime (for now) so if we have no transport available that works can error out

buut, I don't know why we would ever hit this in our current tests

flub: I think I disagree. You could be sending a datagram for a disabled transport which would result in an error. But later you could send a datagram to an enabled transport, and that should still succeed. So it's a transient error I think.

dignifiedquire: transports can't be enabled or disabled currently, this could be the case in the future, but right now that is not possible

PR #3763, opened 2025-12-11
fix(iroh): improve shutdown logic in the magicsock

dignifiedquire:
## Description

An attempt at improving the speed and reliability of the shutdown in magicsock

- set closing asap
- ensure remote actors are shut down using cancellation tokens
- shutdown actors before waiting for `wait_idle`
- use `compare_exchange` to improve `closing` logic

Ref #3762

### DIFF
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -1136,11 +1136,6 @@ impl Endpoint {
     /// Be aware however that the underlying UDP sockets are only closed once all clones of
     /// the the respective [`Endpoint`] are dropped.
     pub async fn close(&self) {
-        if self.is_closed() {
-            return;
-        }
-
-        tracing::debug!("Connections closed");
         self.msock.close().await;
     }
 
@@ -1320,7 +1315,7 @@ mod tests {
     use quinn::ConnectionError;
     use rand::SeedableRng;
     use tokio::sync::oneshot;
-    use tracing::{Instrument, error_span, info, info_span, instrument};
+    use tracing::{Instrument, debug_span, info, info_span, instrument};
     use tracing_test::traced_test;
 
     use super::Endpoint;
@@ -1473,7 +1468,10 @@ mod tests {
 
         info!(time = ?test_start.elapsed(), "test setup done");
 
+        let barrier = Arc::new(tokio::sync::Barrier::new(2));
+
         // The server accepts the connections of the clients sequentially.
+        let s_b = barrier.clone();
         let server = tokio::spawn(
             async move {
                 let eps = ep.bound_sockets();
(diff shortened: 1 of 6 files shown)


iroh/examples/transfer.rs
Frando, line 434: What about we move this log above the close right after `drain_stream`, and add another log "Closed cleanly" or such after the close timeout?
Then when we run it and it completes the transfer but only fails during  shutdown we can see that, and not like now where there's no visible difference between the cases (because if the shutdown errors the whole fn errors)

iroh/src/magicsock.rs
```diff
@@ -979,6 +999,23 @@ impl Handle {
...
+        }
+
+        // Only set closing if not already set
+        if self
+            .msock
+            .closing
+            .compare_exchange(false, true, Ordering::Acquire, Ordering::Relaxed)
+            .is_err()
+        {
+            return;
```
Frando, line 984: This means that if you call `Endpoint::close` twice on a clone of the endpoint, only the first call will wait for `quinn::Endpoint::wait_idle`, whereas the second call will complete immediately. I think we should always await `Endpoint::wait_idle`, even if  `closing` is already true.

iroh/src/magicsock.rs
```diff
@@ -979,6 +999,23 @@ impl Handle {
...
+        }
+
+        // Cancel any running netreports
+        self.shutdown_token_netreport.cancel();
```
Frando: We now have quite a few variables in play and I'm wondering if we can simplify it: I think we could remove `self.msock.closing` and instead use only two cancel tokens:
`self.shutdown_token_close_start` and `self.shutdown_token_close_endpoint_closed`
The former would be passed to netreport (i.e. replace `shutdown_token_netreport` and the `closing` AtomicBool), and the latter would be used for everything else.

(i.e. instead of having both a cancel token and an atomicbool, just use `cancel_token.is_cancelled()` for when the atomicbool was used)

iroh-relay/src/server/client.rs
```diff
@@ -258,8 +238,6 @@ struct Actor {
     timeout: Duration,
     /// Packets queued to send to the client
     send_queue: mpsc::Receiver<Packet>,
-    /// Important packets queued to send to the client
-    disco_send_queue: mpsc::Receiver<Packet>,
```
flub, line 262: I had no idea we were prioritising these! That was probably a good call :)

iroh/examples/transfer.rs
```diff
@@ -385,7 +385,7 @@ async fn provide(endpoint: Endpoint, size: u64) -> Result<()> {
                 HumanBytes((size as f64 / duration.as_secs_f64()) as u64)
             );
             if res.is_err() {
-                println!("[{remote}] Did not disconnect within 3 seconds");
+                println!("[{remote}] Error: Did not disconnect within 4 seconds");
```
flub: erm? uses 3s above?

iroh/src/endpoint.rs
```diff
@@ -1438,8 +1438,10 @@ mod tests {
     }
 
     #[tokio::test]
-    #[traced_test]
+    // #[traced_test]
```
flub: i assume this was debugging attempts that you forgot to clean up?

iroh/src/endpoint.rs
```diff
@@ -1461,7 +1463,10 @@ mod tests {
 
         info!(time = ?test_start.elapsed(), "test setup done");
 
+        let barrier = Arc::new(tokio::sync::Barrier::new(2));
+
         // The server accepts the connections of the clients sequentially.
+        let s_b = barrier.clone();
```
flub, line 1474: I'd add some comments as to why this is needed as otherwise the next reader will be puzzled.

iroh/src/magicsock.rs
```diff
@@ -233,7 +235,7 @@ impl MagicSock {
     }
 
     fn is_closing(&self) -> bool {
-        self.closing.load(Ordering::Relaxed)
+        self.closing.load(Ordering::SeqCst)
```
flub: Someone smarter than me says that using `SeqCst` means you haven't really thought about what you're doing...

iroh/src/magicsock.rs
```diff
@@ -979,6 +999,23 @@ impl Handle {
     #[instrument(skip_all)]
     pub(crate) async fn close(&self) {
         trace!(me = ?self.public_key, "magicsock closing...");
+        if self.msock.is_closed() {
+            return;
+        }
```
flub: Maybe even move that before the trace?

iroh/src/net_report.rs
```diff
@@ -553,7 +572,9 @@ impl Client {
                         Some(Ok(Some(Err(time::Elapsed { .. })))) => {
                             debug!("probe timed out");
                         }
-                        None => {}
+                        None => {
+                            debug!("report canceled");
```
flub: these should probably be trace since they're entirely normal behaviour?

iroh/src/magicsock.rs
dignifiedquire: fixed, though we never actually come here, because `Endpoint::close` shortcircuits way before this

iroh/src/magicsock.rs
dignifiedquire: resolved by merging @Frandos PR

iroh/src/endpoint.rs
dignifiedquire: yes 😅

iroh/src/endpoint.rs
dignifiedquire: yeah, I am not sure yet I will leave it in, but it makes it a lot saner to debug these tests

iroh/src/magicsock.rs
dignifiedquire: gone 🎉

iroh/examples/transfer.rs
dignifiedquire: fixed

iroh/examples/transfer.rs
flub: Meh, I thought this would be nice. But it really is a bug anyway if we don't get here so it's fine to leave as is.

iroh/src/magicsock.rs
flub: this would still be nice to move i think

iroh/src/magicsock.rs
```diff
@@ -150,12 +150,37 @@ pub(crate) struct Handle {
...
-    shutdown_token: CancellationToken,
     // quinn endpoint
     endpoint: quinn::Endpoint,
 }
 
+#[derive(Debug)]
+struct ShutdownState {
+    at_close_start: CancellationToken,
+    at_endpoint_closed: CancellationToken,
+    closed: AtomicBool,
```
matheus23, line 161: I don't like this `closed: AtomicBool`.

It's always really hard to reason about what these things do.

What I *want* is
- Have the body of `MagicSock::close` be a critical section. Only one task can enter it at a time.
- Once `MagicSock::close` finishes, the magic socket is *guaranteed* to be closed.
- If the magic socket is already closed `MagicSock::close` should return pretty much immediately.

I think we can achieve this with a very simple [async-once-cell](https://docs.rs/async-once-cell/latest/async_once_cell/).
It doesn't need to store an actual useful value. Maybe it stores a `Result<(), ErrorWhileClosingSocket>` or sth like that.

And then `MagicSock::close` calls `once_lock.get_or_init(/* closure that does the closing logic */).await`.

I think this gives us all of the above properties.

dignifiedquire: I think if we just check for `is_closing` as well, we already get the same behaviour?

dignifiedquire: added `is_closing` feel free to follow up with an improvement

PR #3770, opened 2025-12-12
fix(iroh): Accurately set `RecvMeta::dst_ip`, remove `normalized_local_addr`

### LINKED ISSUE #2491: No link-local IPv6 addresses used for call-me-maybe
The `DiscoveredEndpoints` struct (needs rename, no "endpoint") is responsible for collecting the direct addresses which are sent in a call-me-maybe message.  But they do not include link-local IPv6 addresses.  This means we will not discover the most direct route on an IPv6 local network.  There really is not much downside to including these addresses.

### LINKED ISSUE #2632: Specify `src_ip` correctly in `Transmit`s in `iroh`
Currently, the `src_ip` values for `quinn::Transmit`s are indirectly set by us via our rewriting of `dst_ip` (and choice of `dst_ip` in the relay case) in the `RecvMeta` of `poll_recv` (in `MagicSock`).

This can cause issues, e.g. when unspecified IP addrs (`0.0.0.0` or `::`) make it to `src_ip`. Windows seems more pedantic about this and errors out.

We return unspecified IP addrs, because we set `dst_ip` to `MagicSock::normalized_local_addr`.

As an interim solution, we've set it to `None` on windows.
I suspect this can cause issues when there's multiple IP interfaces connected to the device and the OS ends up choosing the wrong one.

Instead, we should probably figure out the interface we want to send from via netcheck.

More information in [this PR comment](https://github.com/n0-computer/iroh/pull/2595#issuecomment-2290947319).

matheus23:
## Description

Now that we have multipath and don't rely on quinn only seeing "EndpointMappedAddr"s, but instead have it see the *actual* addresses (minus the one MixedMappedAddr), we can actually address #2632 very easily. We just pass it through from how we get it from the socket/OS.

We don't have to worry about MixedMappedAddrs, because we only use them to send, they never end up at `dst_ip` in our transmits, because they get translated to actual IP adresses before hitting the network, and then get received as actual `dst_ip`s on the other side.

## Notes & open questions

Closes #2632

I think we can also close #2491 then? AFAICT, we removed the `is_unicast_link_local` IP check with QNT, so we actually send over link-local addresses now. This could mean that we fully support link-local IPs with this PR now :tada: 

These changes make sense to me, but I didn't test them super thoroughly. FWIW, all tests that fail in this PR already fail before the changes for me locally.

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh/src/magicsock.rs
+++ b/iroh/src/magicsock.rs
@@ -69,10 +69,7 @@ pub(crate) mod mapped_addrs;
 pub(crate) mod remote_map;
 pub(crate) mod transports;
 
-use self::{
-    mapped_addrs::{EndpointIdMappedAddr, MappedAddr},
-    transports::Addr,
-};
+use self::mapped_addrs::{EndpointIdMappedAddr, MappedAddr};
 pub use self::{metrics::Metrics, remote_map::PathInfo};
 
 // TODO: Use this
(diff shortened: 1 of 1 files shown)


flub (approved):
🎉

flub:
> I think we can also close #2491 then? AFAICT, we removed the `is_unicast_link_local` IP check with QNT, so we actually send over link-local addresses now. This could mean that we fully support link-local IPs with this PR now 🎉

Do we not filter out the link-local addresses from the local interfaces? And we also filter out localhost somewhere unless that's the only one found I think. But maybe that's fine because the OS will do hairpinning on the LAN interface address anyway.

matheus23:
> Do we not filter out the link-local addresses from the local interfaces? And we also filter out localhost somewhere unless that's the only one found I think. But maybe that's fine because the OS will do hairpinning on the LAN interface address anyway.

I can find the code that *used* to do that, but I can't find it in the current codebase. We directly use all addresses from our local interface.

dignifiedquire:
@matheus23 see https://github.com/n0-computer/net-tools/pull/50 for the important code

PR #3774, opened 2025-12-12
refactor(iroh): make some types more private and others more public

ramfox:
## Description

Builds on the work done in #3757 to make types more or less public.

- makes `CreateHandleError` public
- makes `IncomingFuture` public
- makes `net_report::GetRelayAddrError` and `net_report::MeasureHttpsLatencyError` more private
- removes unused `IpPorts` struct


flub (approved):
We should probably add the `#[warn(unnamable_types)]` to the lib.rs file now to keep think to this level?

PR #3775, opened 2025-12-13
fix(iroh): retry holepunching

dignifiedquire:
We need to retry holepunching on certain errors, otherwise we will never succeed. 
One major one is not having enough remote cids, as they have not been issued yet.

- recheck connections every 5 seconds if we need to trigger holepunching
- schedule holepunching to run if we have failed to holepunch

-[x] Depends on https://github.com/n0-computer/quinn/pull/254

### DIFF
--- a/iroh/src/magicsock/remote_map/remote_state.rs
+++ b/iroh/src/magicsock/remote_map/remote_state.rs
@@ -382,7 +382,7 @@ impl RemoteStateActor {
         }
 
         if !is_goodenough {
-            debug!("connections are not good enough, noting holepunch");
+            debug!("connections are not good enough, triggering holepunching");
             self.trigger_holepunching();
         }
     }


flub:
We used to holepunch every 5s if we didn't succeed. Forever. That's a very simple and brute-force-ish approach. But might be reasonable first cut?

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -349,6 +370,22 @@ impl RemoteStateActor {
...
+        let mut is_goodenough = true;
+        for conn in self.connections.values() {
+            // If we have at least 1 non relay path, we consider the connection good enough for now
+            // TODO: evaluate other criteria, like current RTT etc
+            let is_conn_goodenough = conn.open_paths.iter().any(|(_, addr)| !addr.is_relay());
+            is_goodenough &= is_conn_goodenough;
+        }
+
+        if !is_goodenough {
+            debug!("connections are not good enough, noting holepunch");
```
flub:
```suggestion
            debug!("connections are not good enough, triggering holepunching");
```
let's keep the log messages consistent :)

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -655,19 +692,20 @@ impl RemoteStateActor {
...
         }
 
-        self.do_holepunching(conn).await;
+        self.do_holepunching(conn);
     }
 
     /// Unconditionally perform holepunching.
     #[instrument(skip_all)]
-    async fn do_holepunching(&mut self, conn: quinn::Connection) {
+    fn do_holepunching(&mut self, conn: quinn::Connection) {
```
flub, line 701: huh, this all didn't need to be async anymore? good catch!

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -688,7 +726,9 @@ impl RemoteStateActor {
                 });
             }
             Err(err) => {
+                // TODO: distinguish between fatal and temporary errors
                 warn!("failed to initiate NAT traversal: {err:#}");
+                self.needs_holepunching.set(true).ok();
```
flub: Doesn't this mechanism already exist with `self.scheduled_holepunch`? Can we not re-use that?

dignifiedquire: done

Cargo.lock
flub, line 1: Are these changes intentional?

dignifiedquire: yes

dignifiedquire: well some of them.

PR #3783, opened 2025-12-17
feat(relay,dns): add healthz routes

Arqu:
## Description

We're missing some basic service utilities. I kept this simple and just set it all to the `/healthz` route.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-dns-server/build.rs
+++ b/iroh-dns-server/build.rs
@@ -1,22 +1,14 @@
-use vergen_gitcl::{BuildBuilder, CargoBuilder, Emitter, GitclBuilder, RustcBuilder};
+use vergen_gitcl::{Emitter, GitclBuilder};
 
 fn main() {
+    // Emit build-time environment variables
     if let Err(e) = emit_vergen() {
         eprintln!("vergen error: {e}");
     }
 }
 
 fn emit_vergen() -> Result<(), Box<dyn std::error::Error>> {
-    let build = BuildBuilder::all_build()?;
-    let cargo = CargoBuilder::all_cargo()?;
-    let gitcl = GitclBuilder::all_git()?;
-    let rustc = RustcBuilder::all_rustc()?;
-
-    Emitter::default()
-        .add_instructions(&build)?
-        .add_instructions(&cargo)?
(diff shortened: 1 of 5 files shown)


iroh-dns-server/src/http.rs
```diff
@@ -192,6 +192,34 @@ impl HttpServer {
...
+    Json(Health {
+        status: "ok",
+        version: env!("CARGO_PKG_VERSION"),
+        git_hash: option_env!("VERGEN_GIT_SHA").unwrap_or("unknown"),
+        http_requests: state.metrics.http_requests.get(),
+        http_requests_success: state.metrics.http_requests_success.get(),
+        http_requests_error: state.metrics.http_requests_error.get(),
+        dns_requests: state.metrics.dns_requests.get(),
+        dns_lookup_success: state.metrics.dns_lookup_success.get(),
+        dns_lookup_notfound: state.metrics.dns_lookup_notfound.get(),
```
dignifiedquire: that's a lot of details you are making publicly accessible, seems potentially risky

Arqu: hmmm fair, they are all available on the metrics side, maybe just nuke those and keep the version info

iroh-dns-server/build.rs
```diff
@@ -0,0 +1,22 @@
...
+fn main() {
+    if let Err(e) = emit_vergen() {
+        eprintln!("vergen error: {e}");
+    }
+}
+
+fn emit_vergen() -> Result<(), Box<dyn std::error::Error>> {
+    let build = BuildBuilder::all_build()?;
+    let cargo = CargoBuilder::all_cargo()?;
+    let gitcl = GitclBuilder::all_git()?;
```
flub: Given we only use this, why not disable all the features on the crate and only use this?

Additionally `all_git()` seems like it has to invoke a whole bunch of git commands, maybe just `sha()` would be lighter as well?

flub (commented):
Definitely in favour of doing this, but would prefer if we did as little as possible in the build.rs and remove all the unused features.

PR #3784, opened 2025-12-17
feat: implement rtt hints for nat traversal

dignifiedquire:
Depends on https://github.com/n0-computer/quinn/pull/278

### DIFF
--- a/iroh/src/magicsock/remote_map/remote_state.rs
+++ b/iroh/src/magicsock/remote_map/remote_state.rs
@@ -793,6 +793,21 @@ impl RemoteStateActor {
                 .private_socket_addr(),
         };
 
+        let rtt_hint = match &open_addr {
+            transports::Addr::Relay(url, _) => {
+                // Relay case, grab the latency from net_report
+                self.last_report
+                    .as_ref()
+                    .and_then(|r| r.relay_latency.get(url))
+            }
+            addr @ transports::Addr::Ip(_) => {
+                let all_rtts = self.calculate_path_rtts();
+                all_rtts
+                    .get(addr)
+                    .and_then(|rtts| rtts.iter().max().copied())
+            }
+        };
+
         for (conn_id, conn_state) in self.connections.iter_mut() {
             if conn_state.path_ids.contains_key(open_addr) {
                 continue;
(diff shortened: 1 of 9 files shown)


iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -707,7 +715,26 @@ impl RemoteStateActor {
...
+        let all_rtts = self.calculate_path_rtts();
+        let relay_mapped_addrs = self.relay_mapped_addrs.clone();
+        let relay_latencies = self.last_report.as_ref().map(|r| r.relay_latency.clone());
+        let rtt_hints = |addr: &SocketAddr| {
+            let addr = relay_mapped_addrs.to_transport_addr(*addr)?;
+            match addr {
+                transports::Addr::Relay(url, _id) => {
+                    // Relay case, grab the latency from net_report
+                    relay_latencies.as_ref().and_then(|l| l.get(&url))
+                }
```
matheus23, line 727: The relay latency is not the same thing as the RTT to the peer via the relay, because that depends on the (potentially two!) relays involved in going back and forth.

Besides, I don't think we need an RTT hint for any relay addresses, because they will/should never be used in holepunching.

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -707,7 +715,26 @@ impl RemoteStateActor {
...
+                addr @ transports::Addr::Ip(_) => {
+                    // IP case, grab the worst latency we currently have
+                    all_rtts
+                        .get(&addr)
+                        .and_then(|rtts| rtts.iter().max().copied())
```
matheus23, line 732: I think this should be `.min(Duration::from_millis(333))` additionally, just so we don't accidentally balloon the PTO in case there was some kind of weird bad RTT in our IP paths.

matheus23: Also, IIUC, this won't help broken paths, right? You only have a single connection in the RemoteState, and the remote address that doesn't work for holepunching won't have an RTT estimate. So you're back to having a 333ms initial RTT estimate for remote addresses that don't work. Am I missing something?

dignifiedquire: the probes will timeout faster for broken paths, that’s the whole point of this PR

matheus23: But... the way the code is written, I don't see how that's going to be the case?

dignifiedquire:
closing this for now, we might want to revisit this later but not right now

PR #3786, opened 2025-12-18
test: improve timeouts for shutdown and reduce some more test rounds

dignifiedquire:


iroh/examples/transfer.rs
```diff
@@ -385,7 +388,10 @@ async fn provide(endpoint: Endpoint, size: u64) -> Result<()> {
                 HumanBytes((size as f64 / duration.as_secs_f64()) as u64)
             );
             if res.is_err() {
-                println!("[{remote}] Error: Did not disconnect within 3 seconds");
+                println!(
+                    "[{remote}] Error: Did not disconnect within {} seconds",
+                    SHUTDOWN_TIME.as_secs()
+                );
```
matheus23:
```suggestion
                println!("[{remote}] Error: Did not disconnect within {SHUTDOWN_TIME:?}");
```

matheus23: Hm seeing that seconds are used below as well, perhaps I'm a tad too big of a fan of `Duration`'s debug impl...

iroh/src/endpoint.rs
```diff
@@ -1468,34 +1468,35 @@ mod tests {
...
-                        recv.read_exact(&mut buf).await.anyerr()?;
-                        send.write_all(&buf).await.anyerr()?;
-                    }
-                    info!(%i, peer = %endpoint_id.fmt_short(), "finishing");
-                    send.finish().anyerr()?;
-                    conn.closed().await; // we're the last to send data, so we wait for the other side to close
-                    info!(%i, peer = %endpoint_id.fmt_short(), "finished");
-                    info!("[server] round {i} done in {:?}", round_start.elapsed());
-                    s_b.wait().await;
+                    tokio::time::timeout(Duration::from_secs(4), async {
```
matheus23, line 1478: Can we use a constant for `Duration::from_secs(4)` with here and below as well?
Ideally with a decent comment about how RTT should usually be sub-100ms but shutdown can take 3s due to connection draining and being unsure about datagrams still flying around on holepunching attempts.

matheus23 (approved):

iroh/examples/transfer.rs
dignifiedquire: uhm how does that look?

flub: `Error: Did not disconnect within 3s`

`Duration`'s `Debug` impl is great. I use it all the time, I'm almost confused why they didn't make that the `Display` impl. It really is the odd one out for `Debug` impls in rust.

iroh/src/endpoint.rs
flub, line 1478: This could be tight on CI machines which are often unreasonably slow at random times. But I'm fine to find out if it works.

iroh/src/magicsock.rs
```diff
@@ -1702,38 +1702,43 @@ mod tests {
...
-                error!("[receiver] Error:\n{err:#?}");
-                true
-            }
-            Err(joinerr) => {
-                if joinerr.is_panic() {
-                    std::panic::resume_unwind(joinerr.into_panic());
-                } else {
-                    error!("[receiver] Error:\n{joinerr:#?}");
+    ) -> Result<()> {
+        tokio::time::timeout(Duration::from_secs(4), async move {
```
flub, line 1706: same reservation on the tight deadline

flub (approved):

iroh/src/endpoint.rs
Arqu: Unless we have >333ms rtt on any of the tested paths, this should always complete in <3.1s so 4s timeout should be plenty of spare room. If we cross that it's either a deliberate test or a bug.

PR #3794, opened 2025-12-19
fix(iroh): ensure IP paths have idle timeouts and keep alives

dignifiedquire:
We need to set idle timeouts and keep alives for IP paths in order to ensure that they are removed in case they get disconnected or broken.

TODO: what are the right magic numbers? 😭


dignifiedquire:
fixed by https://github.com/n0-computer/quinn/pull/286

PR #3796, opened 2025-12-19
feat(iroh): ping paths and trigger holepunching on networkchange

dignifiedquire:
Depends on https://github.com/n0-computer/quinn/pull/290

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -628,16 +628,6 @@ dependencies = [
 
 [[package]]
 name = "core-foundation"
-version = "0.9.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "91e195e091a93c46f7102ec7818a2aa394e1e1771c3ab4825963fa03e45afb8f"
-dependencies = [
- "core-foundation-sys",
- "libc",
-]
-
-[[package]]
-name = "core-foundation"
 version = "0.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b2a6cd9ae233e7f62ba4e9353e81a88df7fc8a5987b8d445b4d90c879bd156f6"
@@ -2009,7 +1999,7 @@ dependencies = [
  "n0-future",
  "n0-tracing-test",
  "n0-watcher",
- "netdev 0.40.0",
+ "netdev",
  "netwatch",
  "parse-size",
  "pin-project",
@@ -2178,7 +2168,7 @@ dependencies = [
 [[package]]
 name = "iroh-quinn"
 version = "0.14.0"
-source = "git+https://github.com/n0-computer/quinn?branch=main#124059612a8149fe973a3f8629ec84c4773af1de"
+source = "git+https://github.com/n0-computer/quinn?branch=main#7ce0396198789ff7b66b5d71bfc12fbd162589df"
 dependencies = [
  "bytes",
  "cfg_aliases",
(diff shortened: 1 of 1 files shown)


iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -368,6 +368,28 @@ impl RemoteStateActor {
                 };
                 tx.send(info).ok();
             }
+            RemoteStateMessage::NetworkChange { is_major } => {
```
flub, line 371: Can we not do this with the local addr watcher? I don't like having two mechanisms for this, and a watcher seemed nice.

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -368,6 +368,28 @@ impl RemoteStateActor {
...
+                        if let Err(err) = path.ping() {
+                            warn!(%err, ?path_id, ?addr, "failed to ping path");
+                        }
+                    }
+                }
+            }
+        }
+
+        if is_major {
+            self.trigger_holepunching();
```
flub, line 392: so this code-path trigges holepunching less often than the watcher for the local addrs? but otherwise does the same? it seems more like the is_major detection should be added to `trigger_holepunching`. And that should also send pings if that is helpful. but i don't think this is helpful yet, i think we need to be able to see the number of in-flight tail-loss probes in select path to benefit from this.

iroh/src/magicsock/remote_map/remote_state.rs
dignifiedquire: addr change update happens a lot later, this is why I added this mechanism, and it also happens for potentially other reasons. The addr change will happen eventually when net report finishes, but we don't need to wait for that, to already know we want to recheck our connections in here

flub: I see. could we more often update our local addrs? i don't think if the interfaces change that we need to wait until we've also run netcheck until we get updates. I'd like to receive them all as quickly as possible, but still only via one API.

dignifiedquire: one day, but that would require the much bigger refactor of netreport to be more stream based, which will not happen any time soon

iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -368,6 +368,28 @@ impl RemoteStateActor {
...
                 tx.send(info).ok();
             }
+            RemoteStateMessage::NetworkChange { is_major } => {
+                self.handle_network_change(is_major);
+            }
+        }
+    }
+
+    fn handle_network_change(&mut self, is_major: bool) {
+        for conn in self.connections.values() {
```
flub, line 378: I feel like I've forgotten half the things, I'm not really sure why we are sending pings today, because I don't think we can do anything with the resulting behaviour. Each path has it's keep-alive and idle-timeout already set. Sending an explicit ping will not change much: the idle timers are already set and will fire. Sending something now won't change that much I think?

iroh/src/magicsock/remote_map/remote_state.rs
flub: I'm going to accept this technical debt :)
It's annoyed me for a while though so hopefully this will motivate me to get round to improving this soon.

flub (approved):
Not sure the pings do anything, but otherwise I'll accept the technical debt being added here as a reasonable tradeoff for now. I'm just a bit sad it's found its way into the this new actor.

dignifiedquire:
> I'm going to accept this technical debt :)
> It's annoyed me for a while though so hopefully this will motivate me to get round to improving this soon.

very fair, I think we can clean this up in some form later

iroh/src/magicsock/remote_map/remote_state.rs
dignifiedquire: the idea is to make sure we have new information as early as possible

matheus23:
I think this PR needs at least two lines of PR description, otherwise it doesn't auto-merge IIUC?

PR #3797, opened 2025-12-19
feat(iroh): implement latency based connection updates

### LINKED ISSUE #3580: Consider re-holepunching if latency is "not good enough"
We used to have logic that would re-holepunch at regular intervals if the latency was "not good enough".  We have some commented out constants related to this still lying around to remind us:

- GOOD_ENOUGH_LATENCY
- UPGRADE_INTERVAL

Currently we re-holepunch whenever we know of new addressing changes.

We need to figure out what we want here.  Do we still need to re-holepunch on intervals if the latency is not good enough?  Or is the address-change logic sufficient?

I'm tempted to believe that the address-change logic is sufficient and that re-holepunching on interval is more likely to result in needless churn, e.g. if you have an active path a different inactive path that is possible might have a better RTT and you switch.  But once you send data the RTT goes to the same number anyway.  Re-holepunching if there's new information OTOH seems like it covers this already in a better way?

So this might just be a noop and deleting the commented out constants and todos.

dignifiedquire:
Closes https://github.com/n0-computer/iroh/issues/3580

### DIFF
--- a/iroh/src/magicsock/remote_map/remote_state.rs
+++ b/iroh/src/magicsock/remote_map/remote_state.rs
@@ -43,22 +43,18 @@ use crate::{
     util::MaybeFuture,
 };
 
+mod guarded_channel;
+mod path_state;
+mod remote_info;
+
 /// How often to attempt holepunching.
 ///
 /// If there have been no changes to the NAT address candidates, holepunching will not be
 /// attempted more frequently than at this interval.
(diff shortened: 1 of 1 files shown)


flub (approved):

matheus23:
I think this PR needs at least two lines of PR description, otherwise it doesn't auto-merge IIUC?

dignifiedquire:
cargo deny is very upset, I am working on it

PR #3800, opened 2025-12-22
feat(metrics): add cache metrics to dns-server

Arqu:
## Description

Two simple gauges to just have basic info about the cache sizes at runtime.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-dns-server/src/metrics.rs
+++ b/iroh-dns-server/src/metrics.rs
@@ -1,6 +1,6 @@
 //! Metrics support for the server
 
-use iroh_metrics::{Counter, MetricsGroup};
+use iroh_metrics::{Counter, Gauge, MetricsGroup};
 
 /// Metrics for iroh-dns-server
 #[derive(Debug, Default, MetricsGroup)]
@@ -38,4 +38,8 @@ pub struct Metrics {
     pub store_packets_updated: Counter,
     /// Number of expired packets
     pub store_packets_expired: Counter,
+    /// Current number of zones in the main cache
+    pub cache_zones: Gauge,
+    /// Current number of zones in the DHT cache
+    pub cache_zones_dht: Gauge,
 }
--- a/iroh-dns-server/src/metrics.rs
+++ b/iroh-dns-server/src/metrics.rs
@@ -1,6 +1,6 @@
 //! Metrics exposed by the server.
 
-use iroh_metrics::{Counter, MetricsGroup};
+use iroh_metrics::{Counter, Gauge, MetricsGroup};
 
 /// Counters exposed by iroh-dns-server.
 #[derive(Debug, Default, MetricsGroup)]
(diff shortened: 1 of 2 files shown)


iroh-dns-server/src/store.rs
```diff
@@ -103,7 +103,8 @@ impl ZoneStore {
         record_type: RecordType,
     ) -> Result<Option<Arc<RecordSet>>> {
         trace!("store resolve");
-        if let Some(rset) = self.cache.lock().await.resolve(pubkey, name, record_type) {
+        let mut cache = self.cache.lock().await;
```
Frando: This will now hold a lock on the cache until the end of the function, which can include a DHT resolve which can take long. Whereas previously the lock would only be held shortly. If you do want to keep a lock around the guaranteed-to-be-fast operations, then I propose to do a block scopes (`{ ... }`) that makes it clear for how long the lock is held.

iroh-dns-server/src/store.rs
```diff
@@ -146,11 +144,11 @@ impl ZoneStore {
             let packet_opt = pkarr.resolve(&key).await;
             if let Some(packet) = packet_opt {
                 debug!("DHT resolve successful {:?}", packet);
-                return self
-                    .cache
-                    .lock()
-                    .await
-                    .insert_and_resolve_dht(&packet, name, record_type);
+                let result = cache.insert_and_resolve_dht(&packet, name, record_type);
+                self.metrics
```
Frando: Maybe move the metrics tracking into the cache itself? Then you can't miss updating it. (i.e. clone the metrics into ZoneCache and update the metrics whenever something changes from there.

iroh-relay/src/protos/common.rs
```diff
@@ -7,7 +7,7 @@ use bytes::{Buf, BufMut};
 use n0_error::{e, stack_error};
 use quinn_proto::{
     VarInt,
-    coding::{Codec, UnexpectedEnd},
+    coding::{Decodable, Encodable, UnexpectedEnd},
```
Frando, line 10: Unrelated change?

Frando (dismissed):

flub:
does this still live?

Arqu:
It should, I'll get to it ASAP.

iroh-dns-server/src/metrics.rs
```diff
@@ -38,4 +38,8 @@ pub struct Metrics {
     pub store_packets_updated: Counter,
     /// Number of expired packets
     pub store_packets_expired: Counter,
+    /// Current number of zones in the main cache
+    pub cache_zones: Gauge,
+    /// Current number of zones in the DHT cache
+    pub cache_zones_dht: Gauge,
```
flub, line 44: Do you not want to do these as two counters instead of a gauge?

flub (approved):
No particular objection to this version. Though I am surprised you're not going for counters.

iroh-dns-server/src/metrics.rs
Arqu: these are just informative

PR #3803, opened 2025-12-23
refactor(iroh): do not export quinn-proto encoding traits

flub:
## Description

Some of the types exported are encodable and decodable, but we don't
need users to be able to encode and decode themselves. So let's hide
this functionality.

## Breaking Changes

Users can no longer encode and decode TransportErrorCode and StreamId
to/from the QUIC wire format.

## Notes & open questions

n/a

## Change checklist

- [x] Self-review.


matheus23:
I need to double-check this, but I might be using the encoding trait for encoding `VarInt`s in iroh-relay's protocol.

flub:
> I need to double-check this, but I might be using the encoding trait for encoding `VarInt`s in iroh-relay's protocol.

I would argue that should directly depend on iroh-quinn-proto if so?

matheus23:
Ah my bad - yeah it does depend on quinn-proto directly. I mixed up which repo I'm looking at :upside_down_face:

PR #3819, opened 2026-01-08
fix: update deny warnings

dignifiedquire:
- remove unused `instant` dep
- temporarily ignore the `lru@0.13` issue, blocked on https://github.com/pubky/pkarr/pull/197
- remove bincode dependency
- update lru to `0.16.3`

### DIFF
--- a/iroh/Cargo.toml
+++ b/iroh/Cargo.toml
@@ -98,6 +98,7 @@ tokio = { version = "1", features = [
 # wasm-in-browser dependencies
 [target.'cfg(all(target_family = "wasm", target_os = "unknown"))'.dependencies]
 wasm-bindgen-futures = "0.4"
+# we don't use time directly, but need to enable it because x509_parser uses these in browsers and we need to enable some feature flags for that to work
 time = { version = "0.3", features = ["wasm-bindgen"] }
 getrandom = { version = "0.3.2", features = ["wasm_js"] }
 


iroh/Cargo.toml
```diff
@@ -98,8 +98,6 @@ tokio = { version = "1", features = [
 # wasm-in-browser dependencies
 [target.'cfg(all(target_family = "wasm", target_os = "unknown"))'.dependencies]
 wasm-bindgen-futures = "0.4"
-# we don't use instant nor time directly, but need to enable it because backoff and x509_parser use these in browsers and we need to enable some feature flags for that to work
-instant = { version = "0.1", features = ["wasm-bindgen"] }
 time = { version = "0.3", features = ["wasm-bindgen"] }
```
matheus23, line 102: Can you preserve the comment? It's also about the time dependency, which we still keep here.

iroh/Cargo.toml
```diff
@@ -138,7 +136,7 @@ iroh-base = { version = "0.95.1", default-features = false, features = ["key", "
 
 # wasm-in-browser test/dev dependencies
 [target.'cfg(all(target_family = "wasm", target_os = "unknown"))'.dev-dependencies]
-tracing-subscriber-wasm = "0.1.0"
+wasm-tracing = "2.1.0"
```
matheus23, line 140: TIL

iroh/tests/integration.rs
```diff
@@ -137,20 +137,11 @@ async fn simple_endpoint_id_based_connection_transfer() -> Result {
...
-            // To avoide trace events in the browser from showing their JS backtrace
-            tracing_subscriber_wasm::MakeConsoleWriter::default()
-                .map_trace_level_to(tracing::Level::DEBUG),
-        )
-        // If we don't do this in the browser, we get a runtime error.
-        .without_time()
-        .with_ansi(false)
-        .init();
+    let mut config = wasm_tracing::WasmLayerConfig::new();
+    config.set_max_level(Level::TRACE);
```
matheus23, line 143: I'm scared if this means we'll start showing stacktraces in the browser for all trace logs :upside_down_face: 

Does `wasm_tracing` talk about not doing that, unlike `tracing_subscriber_wasm`?

matheus23:
I'm trying to see if these `wasm-tracing` changes work or not. For me locally, the wasm stuff seems to be broken? That's both main and this branch though. So perhaps this shouldn't stop this from merging...

iroh/tests/integration.rs
dignifiedquire: not sure, https://docs.rs/wasm-tracing/latest/wasm_tracing/fn.set_as_global_default_with_config.html

dignifiedquire:
> I'm trying to see if these `wasm-tracing` changes work or not. For me locally, the wasm stuff seems to be broken? That's both main and this branch though. So perhaps this shouldn't stop this from merging...

I tested it locally, and I get the logs as I expect them in my console when running the integration test

iroh/Cargo.toml
dignifiedquire: done

matheus23 (approved):
Heh okay, seems like `wasm-tracing` only stylizes the logs, but doesn't actually use `console.debug` or `console.log` depending on log level. This means you don't get the neat ability to select for only warn/info/etc filters in the DevTools.

Meh. It's fine though.
image

(But also, this means it doesn't ever call `console.trace`, thus it never ends up generating tons of callstacks)

PR #3832, opened 2026-01-08
feat(iroh-relay): Embeddable relay server

lucksus:
## Description
This PR renders the iroh-relay server embeddable as a library by making `RelayService` public ~and adding a native axum integration around it, enabling the relay service to be integrated directly into axum applications as a standard route handler without requiring connection-level routing, separate ports, or reverse proxies.~

Adds docs, examples and tests for all new and/or now public functions/structs.

## Key changes:
  - Refactored relay internals (RelayedStream, Client, Actor, Clients) to be generic over stream types while maintaining backward compatibility
  - ~Added new axum feature and server::axum_integration module~
  - ~Implemented AxumWebSocketAdapter that bridges axum's WebSocket to the relay's Stream/Sink traits~
  - ~Created RelayState and relay_handler for easy integration into axum routers~
  - ~Full relay protocol support including handshake, authentication, authorization, and client registration~
  - Made all types public that are needed to have the Axum integration be done in an external crate

## Breaking Changes

  None. All changes are internal refactoring or new additions. The generic implementations are backward compatible with existing relay server usage through type aliases and preserved public APIs.

## Notes & open questions

  - ~Rate limiting: Client-side rate limiting configuration is accepted but not fully enforced yet in the axum integration (similar limitation exists in the existing HTTP server integration)~
  - ~TLS key export: AxumWebSocketAdapter implements ExportKeyingMaterial but returns None since axum's WebSocket doesn't support TLS key material export. This matches the behavior of non-TLS connections in the existing relay server.~
  - ~Error handling: Axum WebSocket errors are wrapped in tokio_websockets::Error::Io variant for consistency with the relay's error handling~
  - ~Testing: Integration tests verify both non-TLS and TLS scenarios with full end-to-end relay communication~

### Could We Reduce This?

  We could potentially create a single public "escape hatch" function:
```Rust
  pub async fn handle_custom_websocket<S>(
      stream: S,
      access_config: &AccessConfig,
      clients: &Clients,
  ) -> Result<()>
  where S: Stream + Sink + ExportKeyingMaterial
```
  This would hide all the intermediate types and just expose one function. However, this would:
  - Remove flexibility (consuming crates couldn't customize the handshake or authorization logic)
  - Still require AccessConfig, Clients, and the trait bounds to be public
## Change checklist

  - [x] Self-review.
  - [x] AI review
  - [x] Documentation updates following the https://rust-lang.github.io/rfcs/1574-more-api-documentation-conventions.html#appendix-a-full-conventions-text, if relevant.
  - [x] Tests added.
  - All breaking changes documented.
    - N/A - No breaking changes to public API

### DIFF
--- a/iroh-relay/Cargo.toml
+++ b/iroh-relay/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "iroh-relay"
-version = "0.96.0"
+version = "0.96.1"
 edition = "2024"
 readme = "README.md"
 description = "Iroh's relay server and client"
@@ -19,7 +19,7 @@ crate-type = ["lib", "cdylib"]
 workspace = true
 
 [dependencies]
-bytes = "1.7"
+bytes = "1.11"
 derive_more = { version = "2.0.1", features = [
     "debug",
     "display",
@@ -31,7 +31,7 @@ http = "1"
 http-body-util = "0.1.0"
 hyper = { version = "1", features = ["server", "client", "http1"] }
 hyper-util = "0.1.1"
-iroh-base = { version = "0.96.0", path = "../iroh-base", default-features = false, features = ["key", "relay"] }
+iroh-base = { version = "0.96.1", path = "../iroh-base", default-features = false, features = ["key", "relay"] }
 iroh-metrics = { version = "0.38", default-features = false }
 n0-future = "0.3"
 num_enum = "0.7"
@@ -42,8 +42,8 @@ postcard = { version = "1", default-features = false, features = [
     "use-std",
     "experimental-derive",
 ] }
-quinn = { package = "iroh-quinn", version = "0.16", default-features = false, features = ["rustls-ring"] }
-quinn-proto = { package = "iroh-quinn-proto", version = "0.15" }
+quinn = { package = "iroh-quinn", version = "0.16.1", default-features = false, features = ["rustls-ring"] }
+quinn-proto = { package = "iroh-quinn-proto", version = "0.15.1" }
 rand = "0.9.2"
 reqwest = { version = "0.12", default-features = false, features = [
     "rustls-tls",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,19 @@
 
 All notable changes to iroh will be documented in this file.
 
-## [0.96.0](https://github.com/n0-computer/iroh/compare/v0.95.1..0.96.0) - 2026-01-28
+## [0.96.1](https://github.com/n0-computer/iroh/compare/v0.96.0..0.96.1) - 2026-02-06
+
+### 🐛 Bug Fixes
+
+- *(iroh)* Use latest portmapper to avoid duplicate dependencies ([#3903](https://github.com/n0-computer/iroh/issues/3903)) - ([5126857](https://github.com/n0-computer/iroh/commit/5126857f2dedd0fc38ba4a7f24f40fbb9c6d29bb))
+
+### ⚙️ Miscellaneous Tasks
+
+- Pin digest to version compatible with curve25519-dalek 5.0.0-pre.1 ([#3908](https://github.com/n0-computer/iroh/issues/3908)) - ([68243ed](https://github.com/n0-computer/iroh/commit/68243eda3caedd4e64c231a6553f4d5dc6c37aad))
+- Update bytes and time address advisories - ([f2eb1e3](https://github.com/n0-computer/iroh/commit/f2eb1e3c54a625117e53b9224a40c07c0ee52b3c))
+- Specify minimum iroh-quinn and iroh-quinn-proto patch versions ([#3918](https://github.com/n0-computer/iroh/issues/3918)) - ([db429c7](https://github.com/n0-computer/iroh/commit/db429c71497d6da28d310ecdb541c49ce0fce7c4))
+
+## [0.96.0](https://github.com/n0-computer/iroh/compare/v0.95.1..v0.96.0) - 2026-01-28
 
 ### ⛰️  Features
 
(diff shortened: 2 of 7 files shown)


iroh-relay/src/server/axum_integration.rs
matheus23, line 1: Ideally, we only expose what is necessary on the iroh side to make it possible to implement exactly this outside the library.

This code probably better belongs into the library or application that wants to integrate iroh-relay with axum.

iroh-relay/Cargo.toml
```diff
@@ -91,6 +91,7 @@ tracing-subscriber = { version = "0.3", features = [
 ], optional = true }
 blake3 = "1.8.2"
 serde_bytes = "0.11.17"
+axum = { version = "0.8.8", default-features = false, features = ["ws", "tokio", "http1"], optional = true }
```
matheus23: I think we'd prefer not introducing this dependency.
It's an optional feature, but will still prevent us from updating iroh-relay to newer versions of e.g. hyper as long as axum hasn't updated yet.
If the axum-specific code is moved to outside iroh-relay, this dependency won't be necessary here.

iroh-relay/src/server/axum_integration.rs
```diff
@@ -0,0 +1,432 @@
...
+    // Extract the client auth header if present
+    let client_auth_header = headers.get(crate::http::CLIENT_AUTH_HEADER).cloned();
+
+    debug!("Relay WebSocket upgrade request");
+
+    ws.on_upgrade(move |socket| async move {
+        if let Err(e) = handle_relay_websocket(socket, state, client_auth_header).await {
+            warn!("Error handling relay WebSocket: {:?}", e);
+        }
+    })
```
matheus23: I think this is one way of doing this.
But because axum uses tungstenite instead of tokio-websockets that we use, it requires `RelayedStream` to be generic over the websocket implementation.

An alternative might be to use axum's `ConnectionInfo` and extracting the underlying TLS stream and giving that to iroh-relay, and let it handle the whole HTTP upgrade and websocket framing itself.

iroh-relay/src/server/axum_integration.rs
lucksus: First of all, thanks for quick review! Yeah totally understand you don't want to add axum as a dep. Just though this might be of service for others wanting to integrate iroh-relay with an axum server. I will move all axum related code over to our end.

iroh-relay/src/server/axum_integration.rs
lucksus: I can't see how we could do the websocket upgrading (extracting underlying streams) from `ConnectInfo` or any handler or middleware **inside** Axum.

So I think we would have to keep the generics at least so we can do this here, but making those types public in iroh so we can move this code outside of iroh.

The only other alternative I see for use to have iroh and our axum server running together would be to not integrate in axum really but have custom hyper (i.e. connection based code) *before* axum and route relay connections around it.

matheus23: Right. Looks like I misunderstood `ConnectInfo` types - you can only get out the `SocketAddr`?

Let me check back with the team about this PR :)

jost-s: @matheus23 Can you give an estimate when this change is going to be reviewed or something similar released?

matheus23: Sorry for the delay here, I've only talked to the team yesterday. Getting back to you with something concrete within a week should be reasonable.
FWIW, so far the team agrees making the relay embeddable is something desired.

jost-s: That is great to hear, thanks for getting back. The reason I'm asking is that we want to add authentication to the server that would embed the iroh relay service by end of February at the latest. It would be fantastic if you could release this modification in time for that, but If you can't, we'll think of another way to accomplish that.

iroh-relay/src/server/clients.rs
```diff
@@ -45,7 +55,14 @@ impl Clients {
     }
 
     /// Builds the client handler and starts the read & write loops for the connection.
-    pub async fn register(&self, client_config: Config, metrics: Arc<Metrics>) {
+    pub async fn register<S>(&self, client_config: Config<S>, metrics: Arc<Metrics>)
+    where
+        S: Stream<Item = Result<bytes::Bytes, crate::protos::streams::StreamError>>
+            + Sink<bytes::Bytes, Error = crate::protos::streams::StreamError>
+            + Unpin
```
matheus23: We already have `BytesStreamSink` in `iroh-relay/src/proto/streams.rs`. We should re-use this here.

iroh-relay/src/server/client.rs
```diff
@@ -252,7 +265,12 @@ struct Actor {
     metrics: Arc<Metrics>,
 }
 
-impl Actor {
+impl<S> Actor<S>
+where
+    S: Stream<Item = Result<bytes::Bytes, crate::protos::streams::StreamError>>
+        + Sink<bytes::Bytes, Error = crate::protos::streams::StreamError>
+        + Unpin,
```
matheus23: We can use `BytesStreamSink` here, too.

matheus23:
There's also unfortunately more things that need to be made `pub` (and thus need comments). That's why the CI checks fail.

To fix the formatting CI check, run `cargo make format` with `cargo-make` installed.

matheus23:
The Wasm CI check should be fixed by updating to the `Cargo.lock` file used in `main`.

matheus23:
The netsim integration tests might currently fail, due to reasons unrelated to this PR IIRC. I'll keep an eye on this once you addressed the other CI issues.

lucksus:
Great 🙂 I'll try to address the failing tests shortly

lucksus:
Ok, everything is passing now except for the netsim integration tests.

matheus23:
Thank you for addressing CI. There are still two review comments from myself that you've missed @lucksus

lucksus:
Sorry I missed those comments, @matheus23. Commit pushed.

matheus23:
Thank you for this!

matheus23:
CI looks good from my side. The netsim check failing is a known issue we're working on.

jost-s:
Awesome you're integrating this!

When will this be released, do you know @matheus23?

matheus23:
It's not set in stone. It will be released with the 0.97 release, see the attached milestone. I don't have a precise estimate, but think in the order of weeks, not months.

PR #3835, opened 2026-01-09
fix(iroh): actually use user-provided bind addrs

Frando:
## Description

The new user-defined IP transports added #3692  don't actually work because they are filtered out in `Transports::bind`.
Here's a fix, and a test that tests the currently implemented behavior. The test fails without this fix.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -2755,15 +2755,13 @@ mod tests {
         assert!(
             bound_sockets
                 .iter()
-                .find(|x| x.ip() == IpAddr::V4(Ipv4Addr::LOCALHOST))
-                .is_some()
+                .any(|x| x.ip() == IpAddr::V4(Ipv4Addr::LOCALHOST))
         );
         // Test that the default wildcard socket is there
         assert!(
             bound_sockets
                 .iter()
-                .find(|x| x.ip() == IpAddr::V4(Ipv4Addr::UNSPECIFIED))
-                .is_some()
+                .any(|x| x.ip() == IpAddr::V4(Ipv4Addr::UNSPECIFIED))
         );
         ep.close().await;
         Ok(())
@@ -2790,8 +2788,7 @@ mod tests {
         assert!(
             bound_sockets
                 .iter()
-                .find(|x| x.ip() == IpAddr::V4(Ipv4Addr::LOCALHOST))
-                .is_some()
+                .any(|x| x.ip() == IpAddr::V4(Ipv4Addr::LOCALHOST))
         );
         ep.close().await;
         drop(ep);


iroh/src/endpoint.rs
```diff
@@ -2717,4 +2718,71 @@ mod tests {
...
+        let port: u16 = 12345;
+        let ep = Endpoint::empty_builder(RelayMode::Disabled)
+            .clear_ip_transports()
+            .bind_addr((Ipv4Addr::UNSPECIFIED, port))?
+            .bind()
+            .await?;
+        let bound_sockets = ep.bound_sockets();
+        assert_eq!(bound_sockets.len(), 1);
+        assert!(bound_sockets[0].is_ipv4());
+        assert_eq!(bound_sockets[0].port(), port);
```
matheus23: Unfortunately that isn't guaranteed I think. If the port is already in-use, we use a fallback.

iroh/src/endpoint.rs
```diff
@@ -2717,4 +2718,71 @@ mod tests {
...
         Ok(())
     }
+
+    #[tokio::test]
+    async fn test_bind_addr() -> Result {
+        // we use different port numbers for each test because it is not guaranteed that the sockets
+        // are free immediately after the endpoint is dropped.
+
+        // test 1: clear ip transports and add single IPv4 bind
+        let port: u16 = 12345;
```
flub: I'd strongly recommend to use `0` for port numbers in the test. I don't think your test cares about ports. And hardcoding ports **will** be flaky at some point.

iroh/src/endpoint.rs
```diff
@@ -2717,4 +2718,71 @@ mod tests {
...
+        let bound_sockets = ep.bound_sockets();
+        assert_eq!(bound_sockets.len(), 1);
+        assert!(bound_sockets[0].is_ipv4());
+        assert_eq!(bound_sockets[0].port(), port);
+        ep.close().await;
+        drop(ep);
+
+        // test 2: do not clear ip transports and add single non-default IPv4 bind
+        // this will bind three sockets: wildcard binds for IPv4 and IPv6, and our
+        // manually-added IPv4 bind.
```
matheus23: I'm personally a fan of separating the tests into individual test fns :eyes:

flub (changes requested):

iroh/src/endpoint.rs
Frando: Changed to port 0 and binding to LOCALHOST instead to test difference to the default unspecified bind.

iroh/src/endpoint.rs
Frando: Done

iroh/src/endpoint.rs
Frando: Changed to port 0 and binding to LOCALHOST instead to test difference to the default unspecified bind.

iroh/src/endpoint.rs
```diff
@@ -2717,4 +2718,84 @@ mod tests {
 
         Ok(())
     }
+
+    /// Testing bind_addr: Clear IP transports and add single IPv4 bind
+    #[tokio::test]
+    #[traced_test]
+    async fn test_bind_addr_no_clear() -> Result {
+        let ep = Endpoint::empty_builder(RelayMode::Disabled)
+            .clear_ip_transports()
```
flub, line 2727: erm, the test name is `_no_clear`?

iroh/src/endpoint.rs
```diff
@@ -2717,4 +2718,84 @@ mod tests {
...
+        Ok(())
+    }
+
+    /// Testing bind_addr: Do not clear IP transports and add single non-default IPv4 bind
+    ///
+    /// This will bind three sockets: wildcard binds for IPv4 and IPv6, and our
+    /// manually-added IPv4 bind.
+    #[tokio::test]
+    #[traced_test]
+    async fn test_bind_addr_clear() -> Result {
```
flub: and this one does not clear :)

iroh/src/endpoint.rs
Frando: fixed

iroh/src/endpoint.rs
Frando: fixed

PR #3836, opened 2026-01-09
refactor(iroh)!: no longer fall back to a random free port when using `endpoint::Builder::bind_addr` with an unusable port

Frando:
## Description

Currently, when setting a bind addr via `endpoint::Builder::bind_addr` we *always* fallback to a random free port. I think this is quite unexpected: When manually setting a bind addr with a port, I'd expect binding the endpoint would fail. 

This PR **changes the default for user-provided bind addresses to *not* fallback to a random free port**, instead returning an error from `Builder::bind`

## Breaking Changes

Not API breaking, but behavior breaking: When using `iroh::endpoint::Builder::bind_addr` with an address where the port is taken, iroh no longer will attempt to bind to the same address but with a random free port. Instead, binding the endpoint will fail.

## Notes & open questions

Initially this PR added a configuration option to BindOpts to optionally keep the previous fallback behavior, but we decided to remove it. Users may inspect the error from Endpoint::bind to find out that binding failed due to AddrInUse, and then try again with a different port themselves.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -2866,7 +2866,7 @@ mod tests {
 
         assert!(matches!(
             res,
-            Err(BindError::Sockets {
+            Err(BindError::CreateQuicEndpoint {
                 source: io_error,
                 ..
             })


flub (approved):
yay for tests!

iroh/src/endpoint.rs
```diff
@@ -2786,4 +2791,87 @@ mod tests {
...
+            .bind_addr_with_opts(
+                (Ipv4Addr::LOCALHOST, port),
+                BindOpts::default().set_is_required(false),
+            )?
+            .bind()
+            .await;
+
+        assert!(matches!(
+            res,
+            Err(BindError::Sockets {
```
```suggestion
            Err(BindError::CreateQuicEndpoint {
```

PR #3838, opened 2026-01-09
refactor(iroh)!: bind addrs with prefix len 0 are default routes

Frando:
## Description

Changes `bind_addr_with_opts` such that when binding addresses with a prefix length of 0, they are marked as default routes and thus replace the preconfigured unspecified address for the address family.

This means that now `bind_addr` *replaces* the preconfigured unspecified bind for that family, which is arguably what users expect.

Also changes `BindOpts::is_default_route` to be an `Option<bool>` internally. This allows to explicitly set `is_default_route(false)` even with a prefix_len of 0.

## Breaking Changes

* `endpoint::Builder::bind_addr()` now replaces the preconfigured unspecified bind addr for the address family. Previously, a new socket would be added in addition to the preconfigured wildcard ones.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -243,7 +243,8 @@ impl Builder {
     ///   on the *unspecified* address: `[::]`. This bind is allowed to fail however.
     ///
     /// - Adding a bind address removes the pre-configured unspecified bind address for this
-    ///   address family.
+    ///   address family. Use [`Self::bind_addr_with_opts`] to bind additional addresses without
+    ///   replacing the default bind address.
     ///
     /// - This should be called at most once for each address family: once for IPv4 and/or
     ///   once for IPv6. Calling it multiple times for the same address family will result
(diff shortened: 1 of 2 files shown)


iroh/src/endpoint/bind.rs
```diff
@@ -101,7 +101,8 @@ impl BindOpts {
     ///
     /// See [`Self::set_prefix_len`] for details on how this routing works.
     ///
-    /// Defaults to `false`.
+    /// Defaults to `false`. If the prefix length is set to `0` (the default) then this
+    /// will be set to `true` implicitly.
```
flub: - defaults to false
- if the default then it's true

This is very confusing...

iroh/src/endpoint.rs
```diff
@@ -318,12 +315,16 @@ impl Builder {
...
     ///
     /// Be aware that using a subnet with a prefix length of `/0` will always contain all
     /// destination addresses. It is valid to configure this, but no more than one such
-    /// socket should be bound or the routing will be non-deterministic. Prefer using
-    /// [`BindOpts::set_is_default_route`] on one of the sockets instead.
+    /// socket should be bound or the routing will be non-deterministic.
+    ///
+    /// To use a subnet with a non-zero prefix length as the default route, use
+    /// [`BindOpts::set_is_default_route]. Subnets with a prefix length of zero are always
+    /// marked as default routes.
```
matheus23:
```suggestion
    /// To use a subnet with a non-zero prefix length as the default route in addition to
    /// being routed when its prefix matches, use [`BindOpts::set_is_default_route].
    /// Subnets with a prefix length of zero are always marked as default routes.
```

matheus23: The diff is hard to read because of the changed newline flow, but essentially I inserted "in addition to being routed when its prefix matches, ".

matheus23:
There are conflicts with main (again?)

Frando:
I pushed a commit that makes `BindOpts::is_default_route` an `Option<bool>` internally. If `BindOpts::set_is_default_route`  is not called, then `is_default_route` returns `true`  if `prefix_len` is `0`, and `false` otherwise. If `set_is_default_route` is called, then that explicit value is always returned from `is_default_route`.

PR #3840, opened 2026-01-09
Revert "docs(endpoint::Builder): clarify that clearing relay transport doesn't alter holepunching"

flub:
Reverts n0-computer/iroh#3833

To holepunch you still need to have a QUIC connection to the peer via the relay transport. That same connection that carries application data. So this isn't true.


matheus23 (approved):
Ideally we have a replacement comment I guess?

I think the point is that clearing it doesn't mean that we don't make websocket and/or QAD connections to relay servers.

PR #3841, opened 2026-01-09
refactor(iroh): Remove mutexes from `RemoteMap`

matheus23:
## Description

This is a follow-up from #3681 

That PR introduced (roughly) another `tasks: Mutex<JoinSet<..>>` in `RemoteMap`.

This PR removes the `Mutex` from that.
It does so by making calls that worked by acquiring locks to `RemoteMap` into calls that use channels instead, except for the `.try_send(RemoteStateMessage::SendDatagram(...))` call.
We re-use the magic socket actor for that, since that's an actor at exactly the right level with very little to do at the moment.
The `SendDatagram` call is now implemented by sharing a `papaya::HashMap` of all `mpsc::Sender<RemoteStateMessage>` actor inboxes with the transports.
IIUC, it's possible that these are *very slightly* outdated, but that should be unlikely.
If the send doesn't make it through, then that error would reach the transport.

## Breaking Changes

There should be none. This is an internal refactor.

## Notes & open questions

In theory the papaya hashmap is not the ideal data structure. It allows both reading and writing from multiple threads.
What I really want is a map that gives me cheap clones with read-only access while having one "main" instance that can be modified with `&mut` references. I think an `ArcSwap`-based map would in theory get pretty close to the optimum here, but there are no good crates for this as far as I can tell.

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh/src/socket.rs
+++ b/iroh/src/socket.rs
@@ -373,6 +373,10 @@ impl Socket {
         res
     }
 
+    /// Tries to send a [`RemoteStateMessage`] to the `RemoteStateActor` for given [`EndpointId`].
+    ///
+    /// Returns an error if there currently is no remote state actor running for this, or when it
+    /// is currently shutting down.
     pub(crate) fn try_send_remote_state_msg(
         &self,
         endpoint_id: EndpointId,
--- a/iroh/src/socket/concurrent_read_map.rs
+++ b/iroh/src/socket/concurrent_read_map.rs
@@ -1,8 +1,10 @@
 //! This module implements a map that can be modified from only one task but read from many others.
 //!
-//! The combination of only being able to write using `&mut self` in [`ConcurrentReadMap`] and
-//! only allowing reading these from other places via read-only [`ReadOnlyMap`]s with `&self` is
-//! what ensures that this map avoids race conditions from multiple racing writers.
+//! We ensure this map avoids race conditions from multiple writers by doing these two things:
+//! - We only allow writing from one owner of the `&mut self` [`ConcurrentReadMap`].
+//!   It cannot be cloned.
+//! - The read-only replicas [`ReadOnlyMap`] can only read, not write, but allow reading from
+//!   concurrent tasks.
 
 use std::{hash::Hash, sync::Arc};
 
--- a/iroh/src/socket/remote_map.rs
+++ b/iroh/src/socket/remote_map.rs
@@ -51,7 +51,7 @@ pub(crate) struct RemoteMap {
 
     /// The senders for the inbox of each `RemoteStateActor` that runs.
     ///
-    /// This is separated out of `Tasks` to make keeping a mutable borrow if this possible
+    /// This is separated out of `Tasks` to make keeping a mutable borrow of the senders possible
     /// while we're spawning a task using another mutable borrow of `Tasks`.
     senders: ConcurrentReadMap<EndpointId, mpsc::Sender<RemoteStateMessage>>,
 


iroh/src/magicsock/transports.rs
```diff
@@ -759,32 +756,32 @@ impl quinn::UdpSender for MagicSender {
                         "oops, flub didn't think this would happen");
                 }
 
-                let sender = self.msock.remote_map.remote_state_actor(endpoint_id);
                 let transmit = OwnedTransmit::from(quinn_transmit);
-                match sender.try_send(RemoteStateMessage::SendDatagram(
+                match self.msock.try_send_datagram(
```
Frando: This now goes over the msock actor channel, whereas previously it would go directly to the remote state actor channel. Both are using `try_send`, so now we drop packets when the msock actor channel is full (i.e. all initial packets now go over the magicsock actor channel). Could this lead to more packets being dropped when many connections are openend simultaneously?

matheus23: Yeah. That's possible. Previously it would lead to lots of congestion on the remote map mutex(es).

I can't think of a way of fixing this right now, except for a mutex or an unbounded channel?

Open to ideas.

flub: Though previously those mutexes would not be blocking the magicsock actor, would they?

flub: is this a usecase for dashmap?

matheus23: Dashmap still uses mutexes, so it's ... also scary (deadlocks are possible!)

No, previously those mutexes wouldn't block the magicsock actor, but that actor was extremely dormant anyways (it would essentially only react to direct addr and netreport changes).

Frando: If we want to use a lock-free map, I found https://docs.rs/papaya/  to be far safer to use than dashmap. Performance characteristics seem to be [better too](https://github.com/ibraheemdev/papaya/blob/master/BENCHMARKS.md).

matheus23: Updated this to use papaya.

iroh/src/socket.rs
```diff
@@ -209,8 +219,8 @@ pub(crate) struct MagicSock {
     net_report: Watchable<(Option<Report>, UpdateReason)>,
     /// If the last net_report report, reports IPv6 to be available.
     ipv6_reported: Arc<AtomicBool>,
-    /// Tracks the networkmap endpoint entity for each address lookup key.
-    pub(crate) remote_map: RemoteMap,
```
Frando, line 211: Now with the papaya map, why not keep the RemoteMap on the MagicSock and send to there directly, why still send over to the MagicSock actor first?

matheus23: What do you mean? We do that in `MagicSock::try_send_remote_state_msg`.

We don't do this for all kinds of messages, because the papaya map *might* be ever so slightly outdated.

matheus23: If you want to reliably want to send a message, you'd ideally check if the `mpsc::Sender` you used was `closed`, and if so, restart the actor and send the message again. However, we can't do that unless we have the `&mut JoinSet`.

iroh/src/magicsock/remote_map.rs
```diff
@@ -159,86 +153,120 @@ impl RemoteMap {
...
+        addr: EndpointAddr,
+    ) -> Result<Result<EndpointIdMappedAddr, address_lookup::Error>, RemoteStateActorStoppedError>
+    {
+        let EndpointAddr { id, addrs } = addr;
+        let actor = self.remote_state_actor(id);
+        let (tx, rx) = oneshot::channel();
+        actor
+            .send(RemoteStateMessage::ResolveRemote(addrs, tx))
+            .await
+            .map(|_| RemoteStateActorStoppedError::new())?;
```
dignifiedquire: is this `map` correct? should this be map_err?

iroh/src/magicsock.rs
```diff
@@ -320,26 +331,23 @@ impl MagicSock {
...
-        match rx.await {
-            Ok(Ok(())) => Ok(Ok(self.remote_map.endpoint_mapped_addr(id))),
-            Ok(Err(err)) => Ok(Err(err)),
-            Err(_) => Err(RemoteStateActorStoppedError::new()),
-        }
+        self.actor_sender
+            .send(ActorMessage::ResolveRemote(addr, tx))
+            .await
+            .expect("TODO");
+        rx.await.expect("TODO")
```
dignifiedquire: what's our plan here? should we fix this?

iroh/src/magicsock/remote_map.rs
matheus23: Oof. Good catch. It can just be removed.

matheus23: I mean, at least this wouldn't have been a bug :sweat_smile:

iroh/src/socket/concurrent_read_map.rs
```diff
@@ -0,0 +1,75 @@
+//! This module implements a map that can be modified from only one task but read from many others.
+//!
+//! The combination of only being able to write using `&mut self` in [`ConcurrentReadMap`] and
+//! only allowing reading these from other places via read-only [`ReadOnlyMap`]s with `&self` is
+//! what ensures that this map avoids race conditions from multiple racing writers.
```
Frando: I found this sentence a little hard to parse (though not wrong), maybe split in two

iroh/src/socket/remote_map.rs
```diff
@@ -41,16 +46,34 @@ mod remote_state;
...
 pub(crate) struct RemoteMap {
-    //
-    // State we keep about remote endpoints.
-    //
+    /// Maps for converting between mapped and IP/relay addrs.
+    pub(crate) mapped_addrs: MappedAddrs,
+
+    /// The senders for the inbox of each `RemoteStateActor` that runs.
+    ///
+    /// This is separated out of `Tasks` to make keeping a mutable borrow if this possible
```
Frando:
```suggestion
    /// This is separated out of `Tasks` to make keeping a mutable borrow of the senders possible
```

iroh/src/endpoint.rs
```diff
@@ -2964,6 +2964,7 @@ mod tests {
         Ok(())
     }
 
+    #[ignore = "flaky"]
```
Frando, line 2967: Is this test known to be flaky? (Just making sure that it's not a leftover)

iroh/src/socket.rs
```diff
@@ -365,6 +375,17 @@ impl MagicSock {
         res
     }
 
+    pub(crate) fn try_send_remote_state_msg(
```
Frando, line 380: Maybe add a comment here for our future selves that this will fail to send if no RemoteStateActor is running.

iroh/src/endpoint.rs
matheus23: Yes. Unfortunately it is. I had filed https://github.com/n0-computer/iroh/issues/3878 when I added this line.

PR #3845, opened 2026-01-13
feat: Implement custom transports

rklaehn:
## Description

Adds user defined transports.

Transports have to impl a dynable trait, currently named CustomTransport. Implementing this requires implementing another dynable trait CustomSender. Creating a transport requires implementing yet another dynable trait CustomTransportFactory.

A custom addr is just an opaque blob with an u64 tranport id. Small blobs will be inlined, but that is an implementation detail.

Formatting currently is "<id hex>-<data hex>"

Replaces https://github.com/n0-computer/iroh/pull/3707

## Notes & open questions

- Question: encoding of custom addrs in DNS records.
   - Option 1: addr=1_a1b2c3 (tries to parse first as SocketAddr then as custom addr)
   - Option 2: custom=1_a1b2c3 (separate IrohAttr::Custom)

- ~~Question: Feature gate or not? User transports require exposing a bit more of the guts, e.g. transmit fields, so we might want to put this behind a feature flag that is documented to be not part of the stable API.~~

- ~~Question: Path selection.~~

    ~~Currently in select_best_path we never select an user path. For some reason it still works, but we probably should put in some logic there.~~

    ```rust
        Some((transports::Addr::User(_), _)) => {
            // todo: when should we select an user path?
            None
        }
    ```

- DRY the different mapped addrs?
    
    ~~We could DRY the 2 differnet mapped addrs a bit more. But once you start with this you get into a rabbit hole. Should relay transports just be a special kind of custom transport? If we could abstract the logic when to choose the relay, this might work. Maybe this is best left for later.~~

Update: path selection is now taken care of. Basically each path gets a 2-tuple consisting of primary/secondary and biased rtt, then sort and take the best. There is an API to add a bias for each addr type and also to configure for each transport if it is a transport of last resort.

Here is how to bias a transport:

```rust
Endpoint::builder()
    .transport_bias(
        AddrKind::Custom(TEST_TRANSPORT_ID),
        TransportBias::primary().with_rtt_advantage(CUSTOM_TRANSPORT_RTT_ADVANTAGE))
    .build().await?;
```

I also added some tests to make sure custom transports work if the ip transport is enabled.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1994,6 +1994,7 @@ dependencies = [
  "pin-project",
  "pkarr",
  "pkcs8",
+ "portable-atomic",
  "portmapper",
  "postcard",
  "pretty_assertions",
--- a/iroh/Cargo.toml
+++ b/iroh/Cargo.toml
@@ -35,6 +35,7 @@ n0-watcher = "0.6"
 netwatch = { version = "0.14" }
 papaya = { version = "0.2.3", default-features = false }
 pin-project = "1"
+portable-atomic = "1"
 pkarr = { version = "5", default-features = false, features = ["relays"] }
 quinn = { package = "noq", version = "0.16.1", default-features = false, features = ["rustls-ring"] }
 quinn-proto = { package = "noq-proto", version = "0.15.1" }
--- a/iroh/src/runtime.rs
+++ b/iroh/src/runtime.rs
@@ -1,9 +1,7 @@
-use std::{
-    pin::Pin,
-    sync::atomic::{AtomicU64, Ordering},
-};
+use std::pin::Pin;
 
 use iroh_base::EndpointId;
+use portable_atomic::{AtomicU64, Ordering};
 use tokio_util::sync::CancellationToken;
 #[cfg(not(wasm_browser))]
 use tokio_util::task::TaskTracker;
--- a/iroh/src/socket/mapped_addrs.rs
+++ b/iroh/src/socket/mapped_addrs.rs
@@ -8,13 +8,11 @@ use std::{
     fmt,
     hash::Hash,
     net::{IpAddr, Ipv6Addr, SocketAddr, SocketAddrV6},
-    sync::{
-        Arc,
-        atomic::{AtomicU64, Ordering},
-    },
+    sync::Arc,
 };
 
 use n0_error::{e, stack_error};
+use portable_atomic::{AtomicU64, Ordering};
 use rustc_hash::FxHashMap;
 use tracing::trace;
 


matheus23:
> * Question: Path selection.
>       Currently in select_best_path we never select an user path. For some reason it still works, but we probably should put in some logic there.
>       ```rust
>           Some((transports::Addr::User(_), _)) => {
>               // todo: when should we select an user path?
>               None
>           }
>       ```

The examples for user transports that I can come up with are a WebRTC transport, a WifiAware transport, and a BLE transport.
All of these should be preferred over relays, if possible.
If both IP transports and one of these custom transports end up working, then they should compete on RTT.

So that would mean they should basically be regarded as similar to the IP transports?



> * DRY the different mapped addrs?
>       We could DRY the 2 differnet mapped addrs a bit more. But once you start with this you get into a rabbit hole. Should relay transports just be a special kind of custom transport? If we could abstract the logic when to choose the relay, this might work. Maybe this is best left for later.

Relays are still too special. Even if we don't have any transports, we can still open relay connections (the relay connection management isn't self-contained in the relay transport). We also use relays for QAD.
I think we're far off from implementing relays as custom transports, and I don't think there's much reason for trying to do that.

iroh-base/src/endpoint_addr.rs
```diff
@@ -136,6 +139,103 @@ impl From<EndpointId> for EndpointAddr {
...
+    id: u64,
+    /// data
+    data: UserAddrBytes,
+}
+
+#[derive(Clone, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash)]
+enum UserAddrBytes {
+    Inline { size: u8, data: [u8; 30] },
+    Heap(Box<[u8]>),
+}
```
matheus23: I wonder if this should have special cases for IP addresses, given that many user transports are still going to work with IP addrs, e.g. a network-simulating test transport, WifiAware, and I think BLE, too.

rklaehn: It's meant to be opaque. Putting an ip addr or whatever in there is just work. Putting the ip addr into the inline version wouldn't even box I think.

rklaehn: I guess we could have some helpers. But the UserAddr itself should remain just a blob IMHO.

rklaehn:
> Relays are still too special. Even if we don't have any transports, we can still open relay connections (the relay connection management isn't self-contained in the relay transport). We also use relays for QAD.
> I think we're far off from implementing relays as custom transports, and I don't think there's much reason for trying to do that.

Fully agree. I just had to mention it, since you got a lot of places where relays are being treated in a very similar way to user addrs. But it is perfectly fine to keep them separate for now (or maybe forever... )

iroh/src/socket/remote_map/remote_state.rs
```diff
@@ -1152,6 +1159,10 @@ fn select_best_path(
                 }
             }
         }
+        Some((transports::Addr::User(_), _)) => {
+            // todo: when should we select an user path?
+            None
+        }
```
matheus23: This currently just says we never switch away from user transports, even if "better" ones are available.

We need to figure this out before merging. To reason about this, I'm thinking about example user transports:
- WifiAware
- WebRTC
- In-process simulated networks
- Bluetooth

I'd say that we should prefer IP paths if the work and have lower RTT.
So that'd mean that here we're checking if `current_rtt >= new_rtt + RTT_SWITCHING_MIN_IP` similar to above. And if that is true, we resolve to `best_ip_path`, if not, we keep the user path.

Additionally, we should check if we have a user transport available, and switch away from `None`, `Relay` or `Ip` to a user path in the match cases above, if the user transport ends up having a lower RTT.

---

Tbh, this could become quite complicated quickly. E.g. maybe some user transports actually want to stay in the background (with PATH_BACKUP), and only be used if (once we have an API for that) a stream is specifically attached to that path.
We should probably think about how much wiggle-room we have if we'd set the API in stone for 1.0.

TRANSPORTS.md
```diff
@@ -0,0 +1,10 @@
+This table contains a mapping from custom address id to the transport type.
+
+If you want to publish a globally available custom transport, choose an id and do a PR against this repo.
+
+
+
+| transport id | transport | address format | repo | status |
+|--------------|-----------|----------------|------|--------|
+| 0x544F52     | Tor       | Ed25519 public key (32 bytes) | [iroh-tor](https://github.com/n0-computer/iroh-tor) | experimental |
+| 0x424C45     | BLE       | Bluetooth MAC address (6 bytes) | | reserved |
```
laptou, line 10: Using Bluetooth MAC address is not really possible on Apple platforms. CoreBluetooth [doesn't provide any way](https://stackoverflow.com/questions/18973098/get-mac-address-of-bluetooth-low-energy-peripheral-in-ios) to identify devices by their MAC address, and instead uses UUIDs that are stable across sessions but different across hosts. We could work around this by requiring the `iroh` BLE GATT service to expose a characteristic which contains some sort of identifier and use this as the address instead.

laptou:
> We should probably think about how much wiggle-room we have if we'd set the API in stone for 1.0.

I think it might be helpful to have a system for [totally configurable RTT preference](https://github.com/laptou/iroh/commit/88bef30adb45345aee4b876c6d53f89032bd7a28). What do you think of my implementation? This doesn't do anything to keep streams/transports active in the background, but it does give the end user more fine-grained control over which transports are preferred.

laptou:
I implemented a BLE transport [here](https://github.com/laptou/iroh/blob/8d8a96f26c0bb620549c29f17024a5d03383dda2/iroh-ble/src/lib.rs). It's a little rough around the edges, but it totally works. I wonder if we can waive the 1200-byte initial packet requirement for QUIC when using this transport, since BLE MTUs are never that large.

matheus23:
Unfortunately the QUIC specification states:
> QUIC MUST NOT be used if the network path cannot support a maximum datagram size of at least 1200 bytes.

I'm not sure what the reasoning for that hard restriction is. It might just be easier to specify that way.

The way this requirement is checked is by padding the initial packets to 1200 bytes sizes.

TRANSPORTS.md
rklaehn: Yes, I had claude write me a minimal bluetooth device scanner, and it also used UUIDs. But this table is really very preliminary. If UUIDs or something else works better on mac, we can do that. Either use a different transport id or have multiple different encodings for an addr.

rklaehn:
> > We should probably think about how much wiggle-room we have if we'd set the API in stone for 1.0.
> 
> I think it might be helpful to have a system for [totally configurable RTT preference](https://github.com/laptou/iroh/commit/88bef30adb45345aee4b876c6d53f89032bd7a28). What do you think of my implementation? This doesn't do anything to keep streams/transports active in the background, but it does give the end user more fine-grained control over which transports are preferred.

There is going to be a blog post this week about custom transports. Have you seen it? We are reviewing it here: https://github.com/n0-computer/iroh.computer/pull/423

My thinking is that we need a way to bias each transport by providing a rtt bias similar to the existing bias for IPV6 over IPV4, and also be able to configure for each transport if it is a transport of last resort that you want to drop as soon as you got something better, or a transport that you want to compete with other good transports like IP.

Edit: I saw what you did with the TransportPreferences. Yes, I was thinking of something rather similar, but more generic so you can also configure biases for custom transports. Good work!

rklaehn:
> > QUIC MUST NOT be used if the network path cannot support a maximum datagram size of at least 1200 bytes.

Like @matheus23 said, being able to transport 1200 bytes without fragmentation is a hard requirement for QUIC. I was expecting this to be a bit of an issue for BLE.

flub:
> > > QUIC MUST NOT be used if the network path cannot support a maximum datagram size of at least 1200 bytes.
> 
> Like @matheus23 said, being able to transport 1200 bytes without fragmentation is a hard requirement for QUIC. I was expecting this to be a bit of an issue for BLE.

I think folks have run QUIC with smaller MTUs, it possible but less efficient (more overhead per packet) and not standards compliant.

Brainstorming about how to approach this: maybe we can create a new QUIC extension for BLE that is only negotiating a smaller minimum payload size in a transport parameter for BLE transports. If both peers support it you can set smaller min payloads for some paths. There are probably details that matter though, like you'd still want to have a min payload of 1200 for IP paths so how do you know when you open a new path what kind of path it is?

flub:
Other questions regarding QUIC and BLE come to mind as well. E.g. is there a need for address validation? Path validation?

laptou:
> > > QUIC MUST NOT be used if the network path cannot support a maximum datagram size of at least 1200 bytes.
> 
> Like @matheus23 said, being able to transport 1200 bytes without fragmentation is a hard requirement for QUIC. I was expecting this to be a bit of an issue for BLE.

[Apparently](https://claude.ai/share/6b0a698e-1f1b-4a98-ad77-ea4da49883c0), the reasons for this requirement are:
1. To prevent QUIC servers from being used for amplification attacks
2. To make sure the handshake with all of its key data can fit into a single packet, for simplicity's sake

The first reason doesn't really apply in the case of BLE; BLE devices can't really be used for amplification attacks. The second reason does make sense, but isn't exactly a blocker.

> I think folks have run QUIC with smaller MTUs, it possible but less efficient (more overhead per packet) and not standards compliant.

We could do this, but it doesn't seem like we really need to. It would make the implementation simpler in one way (BLE transport doesn't have to break up and reassemble datagrams) while making it more complex in another way (we need to design a way to activate nonstandard QUIC when using special transports).

> is there a need for address validation? Path validation?

Since we can't use the BLE MAC address directly as the device address, then it's worth thinking carefully about our design here. However, in the design I drafted, the public key is used as the identifier, which seems like good enough validation to me because the endpoint won't be able to decrypt the traffic if they don't have a matching private key.

>  I saw what you did with the TransportPreferences. Yes, I was thinking of something rather similar, but more generic so you can also configure biases for custom transports. Good work!

Thanks! As implemented on my branch, you can already configure biases for custom transports too.

rklaehn:
> > I think folks have run QUIC with smaller MTUs, it possible but less efficient (more overhead per packet) and not standards compliant.
> 
> We could do this, but it doesn't seem like we really need to. It would make the implementation simpler in one way (BLE transport doesn't have to break up and reassemble datagrams) while making it more complex in another way (we need to design a way to activate nonstandard QUIC when using special transports).

I would be strongly opposed to going with a non-standard minimum MTU for any transport. That would cause a lot of trouble if you combine transports. You configure BLE as an additional transport and suddenly all your other transports get much slower because the lowest common denominator MTU is now way too small for efficient IP transports. Also this would require a lot of new logic.

flub:
> I would be strongly opposed to going with a non-standard minimum MTU for any transport. That would cause a lot of trouble if you combine transports. You configure BLE as an additional transport and suddenly all your other transports get much slower because the lowest common denominator MTU is now way too small for efficient IP transports. Also this would require a lot of new logic.

The MTU is discovered for each path separately though. But we indeed have the challenge that until Quinn itself is aware of the transports it would not have separate TransportConfigs for each transport, so when on a path we'd have to figure out how to communicate to Quinn that this path has a different minimal MTU (which is something that would be in the TransportConfig).

Of course the longer-term solution, that we already want to do, is to make Quinn aware of the separate transports. And then each would have it's own TransportConfig and this would work easily.

iroh-base/src/endpoint_addr.rs
rklaehn: Thought about this some more, but I stay at the same conclusion: The string encoding of the CustomAddr should be generic. You could add an entire subsystem of a formatter registry to format them idiomatically, but I think this is not worth it.

iroh/src/socket/remote_map/remote_state.rs
rklaehn: See the new API. Basically this is solved by https://github.com/n0-computer/iroh/pull/3870. For each transport (=address type) you can have a rtt bias (positive or negative), and you can configure for each address type if it is a normal transport or a transport of last resort.

Then within each group transports compete according to their biased rtt, and there is stickiness. For switching the group there is no stickiness.

flub:
> [Apparently](https://claude.ai/share/6b0a698e-1f1b-4a98-ad77-ea4da49883c0), the reasons for this requirement are:
> 
>     1. To prevent QUIC servers from being used for amplification attacks
> 
>     2. To make sure the handshake with all of its key data can fit into a single packet, for simplicity's sake

I'm not going to argue with an LLM but take that with a serious grain of salt.


> > I think folks have run QUIC with smaller MTUs, it possible but less efficient (more overhead per packet) and not standards compliant.
> 
> We could do this, but it doesn't seem like we really need to. It would make the implementation simpler in one way (BLE transport doesn't have to break up and reassemble datagrams) while making it more complex in another way (we need to design a way to activate nonstandard QUIC when using special transports).

IIUC you are suggesting that the BLE transport would fragment the datagrams and re-assemble them before passing them back to the QUIC stack? This is generally frowned upon, the IETF has been moving away from layer 2 doing fragmentation. QUIC explicitly says that it's IP transports (it is only defined for IP transports) must not fragment. The reason for this is that in the development/evolution of congestion control algorithms this has been found to make matters worse by hiding behaviour from what the algorithm can observe. E.g. a single lost datagram is very easy to recover from and does not affect the observed latency of the link. A datagram that has part of it lost, is then delayed to wait for it's other half and then ends up being sent later is much, much worse because suddenly the latency jumped for one packet but the link is still didn't lose anything. Or maybe the link decided to delay the whole sequence instead of delivering that one datagram out of sequence as well. All of this affects how well the congestion controller can work.

For best results, and to be closest to the QUIC spec, the BLE transport should drop datagrams that need to be fragmented. Also bear in mind that QUIC will run Path MTU Discovery on the link. If you fragment this would get worse and worse since PMTUD would keep saying that larger packets get through and the MTU would keep being increased, making the BLE layer do more and more fragmentation which is bad.

If we do end up doing fragmentation in BLE as some sort of stop-gap measure until we manage to improve, it should not allow datagrams larger than the minimal MTU of 1200 (plus IP overhead, I think for layer 2 the observed packet size is actually 1280 or so, but have to double check that. The 1200 is the datagram payload size, i.e. the QUIC packet size).

Or, and I'm complicating things now and also making up numbers, if say for example the actual BLE MTU is 700 bytes so a 1280-byte IP packet would need two units, maybe it would make sense to allow filling up the space up to 2 BLE MTU units, i.e. a single IP packet of 1400 bytes. Because you already have 2 fragments. But not allow needing more fragments then strictly needed to meet the QUIC min MTU. If this last paragraph doesn't make sense, ignore it :)

rklaehn:
> > [Apparently](https://claude.ai/share/6b0a698e-1f1b-4a98-ad77-ea4da49883c0), the reasons for this requirement are:
> > ```
> > 1. To prevent QUIC servers from being used for amplification attacks
> > 
> > 2. To make sure the handshake with all of its key data can fit into a single packet, for simplicity's sake
> > ```
> 
> I'm not going to argue with an LLM but take that with a serious grain of salt.
> 
> > > I think folks have run QUIC with smaller MTUs, it possible but less efficient (more overhead per packet) and not standards compliant.
> > 
> > 
> > We could do this, but it doesn't seem like we really need to. It would make the implementation simpler in one way (BLE transport doesn't have to break up and reassemble datagrams) while making it more complex in another way (we need to design a way to activate nonstandard QUIC when using special transports).
> 
> IIUC you are suggesting that the BLE transport would fragment the datagrams and re-assemble them before passing them back to the QUIC stack? This is generally frowned upon, the IETF has been moving away from layer 2 doing fragmentation. 

If your underlying transport has a max guaranteed MTU smaller than what QUIC requires there is not much you can do. Either you allow MTU discovery to go below 1200, then you are no longer QUIC spec compliant, or you allow exactly the allowed minimum and fragment internally. This is not nice, but I don't see why it would not work. If the loss rate is low, the chances of losing a, say, 5 fragment packet is still acceptable. And if the loss is not completely random but in bursts, dropping a packet as soon as one of the 5 fragments is missing would not be a big deal even if the loss rate is higher.

I mean - it is not ideal but it would totally work, and would avoid the complex logic in the transport.

> For best results, and to be closest to the QUIC spec, the BLE transport should drop datagrams that need to be fragmented. Also bear in mind that QUIC will run Path MTU Discovery on the link. If you fragment this would get worse and worse since PMTUD would keep saying that larger packets get through and the MTU would keep being increased, making the BLE layer do more and more fragmentation which is bad.
> 
> If we do end up doing fragmentation in BLE as some sort of stop-gap measure until we manage to improve, it should not allow datagrams larger than the minimal MTU of 1200 (plus IP overhead, I think for layer 2 the observed packet size is actually 1280 or so, but have to double check that. The 1200 is the datagram payload size, i.e. the QUIC packet size).
> 
 Agreed. If we do this hack we would reject anything bigger than the minimum number of packets to send the minimal MTU.

E.g. 1 BLE packet is 254 bytes -> you need 5 packets to send 1200, 5 * 254 = 1270 is the max you don't drop.

> Or, and I'm complicating things now and also making up numbers, if say for example the actual BLE MTU is 700 bytes so a 1280-byte IP packet would need two units, maybe it would make sense to allow filling up the space up to 2 BLE MTU units, i.e. a single IP packet of 1400 bytes. Because you already have 2 fragments. But not allow needing more fragments then strictly needed to meet the QUIC min MTU. If this last paragraph doesn't make sense, ignore it :)

Totally makes sense.

(btw, I somehow had in my mind only the code from where the packets are built to where they get sent to the underlying socket, so I thought that it would cause a lot of trouble to go below the min MTU. But you are right that if this is per path it doesn't affect other paths, so going below the QUIC min MTU for just one path might not be so bad after all...)

flub:
> > IIUC you are suggesting that the BLE transport would fragment the datagrams and re-assemble them before passing them back to the QUIC stack? This is generally frowned upon, the IETF has been moving away from layer 2 doing fragmentation.
> 
> If your underlying transport has a max guaranteed MTU smaller than what QUIC requires there is not much you can do. Either you allow MTU discovery to go below 1200, then you are no longer QUIC spec compliant, or you allow exactly the allowed minimum and fragment internally.

QUIC is defined for transports that can carry IP packets of 1280 bytes without fragmenting. So whatever we do, there is no way to make a BLE transport spec compliant.

So it is all about how much you're not going to be spec compliant and which tradeoffs to make to get the best result.

> This is not nice, but I don't see why it would not work. If the loss rate is low, the chances of losing a, say, 5 fragment packet is still acceptable. And if the loss is not completely random but in bursts, dropping a packet as soon as one of the 5 fragments is missing would not be a big deal even if the loss rate is higher.

Dropping an entire packet as soon as one fragment is dropped is bad as well for the congestion controller. It will estimate a much lower throughput then there actually is.

> I mean - it is not ideal but it would totally work, and would avoid the complex logic in the transport.

I get that fragmentation is an easy solution. I'm not against doing it, but I'm pointing out why it is not great and why I'd consider it to be a temporary hack rather than it being a complete solution.

If a BLE MTU is really 254 bytes than that is very small, and maybe the right answer is actually a compromise between these two: allow fragmenting into 2 or 3 fragments but no more. We'd have to experiment really what performance looks like over BLE links with various MTUs and fragments per QUIC packet.

> (btw, I somehow had in my mind only the code from where the packets are built to where they get sent to the underlying socket, so I thought that it would cause a lot of trouble to go below the min MTU. But you are right that if this is per path it doesn't affect other paths, so going below the QUIC min MTU for just one path might not be so bad after all...)

Yes, it is per path. But the problem is that currently the way to control the MTU discovery, like at what size you start (1200 bytes if spec compliant) and up to what size you probe, is global per connection and not per path. So if you account for a small MTU on one path you start with that small MTU on all paths and you will get very bad startup performance for "normal" IP links as well.

Which is why I'm saying that once the QUIC stack is aware of the different transports it could behave much better.

rklaehn:
> Yes, it is per path. But the problem is that currently the way to control the MTU discovery, like at what size you start (1200 bytes if spec compliant) and up to what size you probe, is global per connection and not per path. So if you account for a small MTU on one path you start with that small MTU on all paths and you will get very bad startup performance for "normal" IP links as well.

Hm, this sounds like it just annoying work but nothing fundamentally difficult. So I am kinda coming around to your way to violate the QUIC spec...

This will come up again. There are a lot of transports with a very small MTU in the IOT space. E.g. zigbee.

iroh-base/src/endpoint_addr.rs
```diff
@@ -136,6 +155,141 @@ impl From<EndpointId> for EndpointAddr {
...
+
+    /// Data associated with this custom address
+    pub fn data(&self) -> &[u8] {
+        self.data.as_bytes()
+    }
+
+    /// Convert to byte representation
+    pub fn as_vec(&self) -> Vec<u8> {
+        let mut out = Vec::with_capacity(8 + self.data.len());
+        out[..8].copy_from_slice(&self.id().to_le_bytes());
```
dignifiedquire, line 306: this is broken, capacity doesn't allocate

iroh-base/src/endpoint_addr.rs
```diff
@@ -136,6 +155,141 @@ impl From<EndpointId> for EndpointAddr {
...
+        out[8..].copy_from_slice(self.data());
+        out
+    }
+
+    /// Parse from bytes
+    pub fn from_bytes(_data: &[u8]) -> Result<Self, &'static str> {
+        todo!()
```
dignifiedquire: what about this?

iroh/src/socket/remote_map/remote_state.rs
dignifiedquire, line 1052: should this account for custom transports in some form?

rklaehn: Not sure. A custom path is just another path, so I guess this does not have to be special cased for custom paths.

iroh-base/src/endpoint_addr.rs
rklaehn: Yeah, that part wasn't done yet. Hence the draft state. Also needs a roundtrip test.

iroh/src/socket/remote_map/remote_state.rs
dignifiedquire: we special case IP paths in here, that's why I am wondering

rklaehn: Oh god. I guess you are right.

We are using is_ip as a "this is a good path" stand-in for now. Not just here, but in a lot of places.

I guess we would have to change is_ip with checking if the path is a "primary". But you can't see that from an addr - custom transports can be either. So it would have to be a lookup in bias_map.

Something like this:
```
+        let is_primary = |addr: &transports::Addr| {
+            self.transport_bias
+                .get(addr)
+                .transport_type == transports::TransportType::Primary
+        };
+
         for (conn_id, conn_state) in self.connections.iter() {
             for (path_id, path_remote) in conn_state
                 .open_paths
                 .iter()
-                .filter(|(_, addr)| addr.is_ip())
+                .filter(|(_, addr)| is_primary(addr))
                 .filter(|(_, addr)| *addr != selected_path)
             {
-                if conn_state.open_paths.values().filter(|a| a.is_ip()).count() <= 1 {
+                if conn_state.open_paths.values().filter(|a| is_primary(a)).count() <= 1 {
                     continue; // Do not close the last direct path.
                 }
                 if let Some(path) = conn_state
```

rklaehn: Without custom transports is_primary returns true only for ip, so the behaviour would be unchanged.

rklaehn: There are only a few non-test places other than this where is_ip is used. BUT... I think custom transports as implemented now introduce the concept of a backup transport, that gets sorted together with the relay transports, but is not a relay.

I wonder if we should just drop this and say that all custom transports are direct. They would have to compete with Ip transports on the basis of their (biased) rtt. That would probably simplify not mixing up all the relay special cases with custom transports.

rklaehn: You could flip is_ip to just !is_relay and call it a day...

dignifiedquire: I am honestly not sure what to do here yet, lets make a todo/note in the comment and open an issue about it, so we don't forget about it

PR #3851, opened 2026-01-14
chore: add a github workflow that, when an issue is added to the `iroh` project board, it gains the `Ready` status

ramfox:
## Description

AI generated workflow, testing it before asking for review/merge

---

Okay, tested. When an issue is added to the `iroh` project board, it will be given the `👍 Ready` status.


ramfox:
Glad I did a run through of figuring out how workflows can interact with github actions, but I realized this particular action is NOT what we want—we add all issues to the github board and not all of them should be auto-marked as "ready". That should be scoped tothe current milestone or manually done. Since scoping it to work with the most "current" milestone would need to be a PR-d change each time the milestone changes, that doesn't really seem like a workable solution.

closing.

Issue #3846, opened 2026-01-13
bug: PathStats seem to be missing/outdated sometimes


matheus23:
This happens on iroh main, (not necessarily the latest release).

**Relevant Logs**
I'm comparing `PathStats` from the watcher you get on the iroh connection vs. iterating through a bunch of PathIds and fetching the stats for all of them that we actually get:
```
Path stats:
  Relay(RelayUrl("https://127.0.0.1:33909/")): RTT 66.865248ms, 7 packets sent
  Ip(127.0.0.1:56796): RTT 4.78022ms, 4 packets sent
  Ip(192.168.160.202:56796): RTT 19.12088ms, 8 packets sent
  Ip([3fff:172:20:20::1]:49108): RTT 6.737371ms, 2 packets sent
Path stats in conn:
  [0]: RTT 58.712621ms, 9 packets sent
  [1]: RTT 8.21914ms, 19 packets sent
  [2]: RTT 9.322508ms, 6 packets sent
  [3]: RTT 8.432184ms, 6 packets sent
  [4]: RTT 9.322508ms, 6 packets sent
  [5]: RTT 9.322508ms, 6 packets sent
  [6]: RTT 9.322508ms, 6 packets sent
```

The first set is the path stats from iroh, the second set is the path stats from quinn.

PathStats doesn't list paths that are abandoned, but not yet discarded, so it's possible that quinn shows more paths due to that.
But also, the data from the path stats seem to be outdated in general.

**Iroh**

Version: f2b6026d70fd5431054a1bfed410af2b8d863be1

**Additional Context / Screenshots / GIFs**
This is the code I use to print the stats:
```rs
async fn watch_paths(conn: Connection) -> Result<()> {
    let mut paths = conn.paths();
    let mut stats = BTreeMap::new();
    while conn.close_reason().is_none() {
        tokio::select! {
            _ = conn.closed() => break,
            _ = paths.updated() => {},
        }
        for path in paths.get() {
            stats.insert(path.id(), (path.remote_addr().clone(), path.stats()));
        }
    }
    let mut stats_by_remote = BTreeMap::<TransportAddr, PathStats>::new();
    for (_, (remote_addr, stats)) in stats {
        let value = stats_by_remote.entry(remote_addr).or_default();
        value.rtt += stats.rtt;
        value.sent_packets += stats.sent_packets;
    }
    println!("Path stats:");
    for (remote, stats) in stats_by_remote {
        println!(
[...]

PR #3860, opened 2026-01-15
feat(example): Improve path stat printing in transfer example

matheus23:
## Description

- Improve path stat printing in the transfer example.

Before:
```
Path stats:
  Relay(RelayUrl("https://127.0.0.1:43765/")): RTT 73.884361ms, 11 packets sent
  Ip([3fff:172:20:20::1]:47128): RTT 1.255437ms, 1 packets sent
```
(also missing a bunch of stats from closed paths)

Now:
```
Path stats:
  Relay(RelayUrl("https://staging-euw1-1.relay.iroh.network./")): RTT 31.046345ms, tx=42738 rx=12
  Ip(176.199.209.244:26617): RTT 25.767613ms, tx=212 rx=17189
  Ip([2a02:8071:d85:19c0::232c]:52316): RTT 264.134µs, tx=13 rx=691477
  Ip(172.18.0.1:45901): RTT 280.83µs, tx=6 rx=16759
  Ip([3fff:172:20:20::1]:52316): RTT 274.261µs, tx=6 rx=7516
  Ip(192.168.0.99:45901): RTT 280.83µs, tx=6 rx=8550
  Ip(172.17.0.1:45901): RTT 2.591812ms, tx=3 rx=2
  Ip([2a02:8071:d85:19c0:c2fb:b929:b323:b940]:52316): RTT 2.607457ms, tx=3 rx=1
  Ip([2a02:8071:d85:19c0:aecd:32d7:7c59:81a2]:52316): RTT 2.607457ms, tx=3 rx=1
  Ip(172.20.20.1:45901): RTT 2.62841ms, tx=3 rx=1
```

## Notes & open questions

There's also something else that's ... weird at least. When the connection is dropped, and we call `PathInfo::stats`, we get back the "fallback" value `self.stats`, but that's only set initially and never updated. So this will look like the stats at the end are suddenly drastically decreasing.

---

The transfer example now has two "side tasks": one for watching the path stats and one for watching the connection type.
I think I'd rather have them split than putting both of these tasks into one? Not sure if everyone might agree.

### DIFF
--- a/iroh/examples/transfer.rs
+++ b/iroh/examples/transfer.rs
@@ -22,7 +22,7 @@ use iroh::{
 use n0_error::{Result, StackResultExt, StdResultExt};
 use n0_future::task::AbortOnDropHandle;
 use netdev::ipnet::{Ipv4Net, Ipv6Net};
-use quinn::{PathId, PathStats};
+use quinn::PathStats;
 use tokio_stream::StreamExt;
 use tracing::{info, warn};
 use url::Url;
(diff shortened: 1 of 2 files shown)


matheus23:
Relevant issue: https://github.com/n0-computer/iroh/issues/3846

dignifiedquire:
ci sad

matheus23:
CI sad for reasons unrelated to this PR :/
- The Wasm check fails because there's a new wasm-bindgen version
- The cross check fails because it fails to bind IPv6 sockets

Frando (approved):
I manually tested and this now indeed prints all paths finally. Makes the path stats much more useful indeed.

It is a bit of a pity that this rather convoluted task is needed, but that's how it is for now. If we find better APIs to expose path stats in iroh, we can update. But for now let's merge this to at least have something that properly shows path stats for a connection in our codebase.

PR #3864, opened 2026-01-16
fix(iroh): Properly set quinn paths to available when holepunched

### LINKED ISSUE #3862: bug: Holepunched paths not used on client side
**Describe the bug**
Running the transfer example with the output modifications from #3860 and https://github.com/n0-computer/quinn/pull/332 I can see that the connection sends asymmetrically. The client side mostly sends via relay, while the server side responds via a holepunched IP path.

**Relevant Logs**

```
Path stats:
  [path_id=0] Relay(RelayUrl("https://staging-euw1-1.relay.iroh.network./")): RTT 27.353039ms, tx=2397 rx=12
  [path_id=1] Ip([3fff:172:20:20::1]:53418): RTT 174.748µs, tx=1 rx=3
  [path_id=2] Ip(172.17.0.1:52651): RTT 241.032µs, tx=66 rx=35406
  [path_id=3] Ip([2a02:8071:d85:19c0:c2fb:b929:b323:b940]:53418): RTT 463.451µs, tx=6 rx=1120
  [path_id=4] Ip(176.199.209.244:26739): RTT 13.089879ms, tx=188 rx=24
  [path_id=5] Ip([2a02:8071:d85:19c0:aecd:32d7:7c59:81a2]:53418): RTT 463.451µs, tx=6 rx=297
  [path_id=6] Ip(172.20.20.1:52651): RTT 463.451µs, tx=6 rx=84
  [path_id=7] Ip([2a02:8071:d85:19c0::232c]:53418): RTT 330.007µs, tx=6 rx=90
  [path_id=8] Ip(172.18.0.1:52651): RTT 330.007µs, tx=6 rx=74
  [path_id=9] Ip([3fff:172:20:20::1]:53418): RTT 330.007µs, tx=6 rx=85
  [path_id=10] Ip(192.168.0.99:52651): RTT 330.007µs, tx=6 rx=60
```
This is the client side after the transfer. Most importantly, we receive lots of datagrams over path 2 ("rx=35406"), which is successfully holepunched (it's also the "selected path"), but we send most data over path 0, the relay path ("tx=2397").

**Expected behavior**
Both sides should send via the IP path mostly.

**Iroh**

Version: Roughly commit 2698efced1a775130f5a3b8b0d1a150be589e322

matheus23:
## Description

Fixes #3862 

The problem was the following *on the client side*:
- quinn's QNT would establish a hole-punched path at path X, with the status set to *backup* initially (that's what QNT rightfully does)
- iroh gets the path open event, selects its best path, finds the IP path, and calls `RemoteStateActor::open_path`
- that function detects that this path is already tracked (?) and skips setting the path status to available
- quinn keeps sending via path 0, because all paths are backup and that's the lowest path ID.

The fix is to actually set the path to available in all cases.

Also in this PR:
- Change debug prints of `?path_id` to `%path_id` to use the display impl.
- Add another logging event: `iroh::_event::path::set_status`, which I personally found *really* useful when debugging.
- Updating to quinn `main`, because otherwise we don't have the required quinn fixes to make hole-punching work correctly, but we need it to work for the regression test I added.
- Adjustments from the quinn update.

## Notes & open questions

- Not quite sure why we get into this condition where we have the path tracked, but not set it to backup/available yet.

## Change checklist

- [x] Self-review.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2152,7 +2152,7 @@ dependencies = [
 [[package]]
 name = "iroh-quinn"
 version = "0.15.0"
-source = "git+https://github.com/n0-computer/quinn.git?branch=matheus23%2Ffix-iroh-unit-tests#47c60c457cc6a7c7a1c501a6c7827a5148b53759"
+source = "git+https://github.com/n0-computer/quinn?branch=main#332c8f7d7a37d14a19f73d4706603fa4fe44062b"
 dependencies = [
  "bytes",
  "cfg_aliases",
@@ -2172,7 +2172,7 @@ dependencies = [
 [[package]]
 name = "iroh-quinn-proto"
 version = "0.14.0"
-source = "git+https://github.com/n0-computer/quinn.git?branch=matheus23%2Ffix-iroh-unit-tests#47c60c457cc6a7c7a1c501a6c7827a5148b53759"
+source = "git+https://github.com/n0-computer/quinn?branch=main#332c8f7d7a37d14a19f73d4706603fa4fe44062b"
 dependencies = [
  "bytes",
  "derive_more",
@@ -2199,7 +2199,7 @@ dependencies = [
 [[package]]
 name = "iroh-quinn-udp"
 version = "0.7.0"
-source = "git+https://github.com/n0-computer/quinn.git?branch=matheus23%2Ffix-iroh-unit-tests#47c60c457cc6a7c7a1c501a6c7827a5148b53759"
+source = "git+https://github.com/n0-computer/quinn?branch=main#332c8f7d7a37d14a19f73d4706603fa4fe44062b"
 dependencies = [
  "cfg_aliases",
  "libc",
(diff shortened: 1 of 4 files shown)


iroh/src/magicsock/remote_map/remote_state.rs
```diff
@@ -818,26 +828,52 @@ impl RemoteStateActor {
...
+                    Level::DEBUG,
+                    remote = %self.endpoint_id.fmt_short(),
+                    ?open_addr,
+                    ?path_status,
+                    ?conn_id,
+                    %path_id,
+                    ?res,
+                );
+                continue;
+            }
```
matheus23, line 853: This is the critical bit.
Before we would just skip setting `path.set_status` with the `continue` that was there before, in case `conn_state.path_ids.contains_key(open_addr)`.

flub (approved):

matheus23:
I'm still trying to figure out how to write a good test for this. I should be pretty close though.

matheus23:
Okay this took me way too long. But I can only write a test that *really* works, if the quinn the test uses current quinn main.

PR #3865, opened 2026-01-16
chore(deps): Update lockfile, mark doctests that require ipv6 `no_run`

matheus23:
## Description

This should fix CI


matheus23:
Filed https://github.com/n0-computer/iroh/issues/3866 so we can reintroduce the tests I marked as `no_run` later.

PR #3870, opened 2026-01-19
refactor: Simplify path selection

rklaehn:
## Description

Fewer special cases in selection of the best path, in preparation to integrating user transports into path selection.

This should keep the current logic completey unchanged, but make it easier to add additional paths without a maze of special cases.

Implemented rules:
- bias for ipv6 (IPV6_RTT_ADVANTAGE)
- available (currently ipv4/ipv6) takes precendence over backup (currently relay)
- switching between different transports in the same category (available or backup) only happens if the advantage is bigger than RTT_SWITCHING_MIN_IP to avoid toggling

The purpose is to be able to integrate user transports more easily. Each user transport id would get a flag if it is considered available or backup (default backup?), and a rtt bias (default 0). Then you would add user transports to the selection process without any special cases.

I guess available or backup are not the right categories here. The meaning would be standard (a transport that has a theoretical chance to become *the best* transport) and fallback (a transport of last resort that you want to get rid of asap as soon as you have another option, regardless of rtt). E.g. a BLE transport would always be a fallback transport because available bandwidth is very limited regardless of rtt.

## Breaking Changes

None

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


iroh/src/socket/remote_map/remote_state.rs
```diff
@@ -1130,80 +1118,52 @@ impl RemoteStateActor {
...
-    addr_v6: SocketAddrV6,
-    rtt_v6: Duration,
-) -> (SocketAddr, Duration) {
-    if rtt_v6 <= rtt_v4 + IPV6_RTT_ADVANTAGE {
-        (addr_v6.into(), rtt_v6)
+    // Get current data. If we don't have data for the current path, don't switch.
+    let current_data = all_paths.get(addr)?;
+    if current_data.status != best_data.status {
+        // Always switch if the status is different (better).
+        Some((best_addr.clone(), best_data.rtt))
```
flub, line 1175: I don't really follow why path status is used here. Path status is only used as a way to instruct quinn which path to prefer. I does not play a role in selecting that path.

I think here you are using two values as a selector for "path of last resort" or not. But even for that 2 values seems a little odd, because it puts e.g. relay and bluetooth on equal footing. But should it be?

To me "biased_rtt" is starting to look like a "priority" value btw. Which admittedly given the rtt corrections we had, it already was.

rklaehn: This is just to 1:1 replicate the logic that was there before. Basically, if the old path was Relay and the new path is Ip, we want to switch regardless of rtt. Even if the relay path has a rtt < the ip path, we switch to ip.

Basically we got 2 groups. First groups is currently only relay, second groups is currently only ip. Then when we get from the bad group to the good group we switch regardless of rtt, but within each group we decide based on rtt (with bias).

dignifiedquire:
I like the structural idea, as well as the ability to expand the test coverage. What's missing to finish this @rklaehn ?

iroh/src/socket/remote_map/remote_state.rs
flub: Yeah, we have those two groups. But using path status for this is really confusing to me. I don't think the current logic uses path status for this, at least it doesn't in my head. Path status is an output of the selection algorithm to me: select this path and set this status.

Essentially this is a boolean or a 2-variant enum, signally if the path is "path of last resort".

The second part I'm struggling with - and this may be out of scope of this PR - is that having this path-of-last-resort being only two states means you would end up having to equal relay and bluetooth. While it is my understanding that you'd prefer to order them.

rklaehn: Yes, what I should do I guess is to provide another 2-case enum with more appropriate names that then possibly maps 1:1 to Available/Backup

You can use the rtt bias to bias in either the "normal transports" group and the "transport of last resort" group, so you could deprioritize BLE over relay by giving it a big positive rtt offset/bias.

rklaehn:
> I like the structural idea, as well as the ability to expand the test coverage. What's missing to finish this @rklaehn ?

Not much. I guess we can merge it as is, since it carefully tries to exactly replicate the existing logic while making things more generic, and then as a next step provide another 2 case enum in the public transport configuration API. That would happen in https://github.com/n0-computer/iroh/pull/3845 or in a subsequent PR to main. But it feels a bit silly to provide a bias API before adding custom transports.

iroh/src/socket/remote_map/remote_state.rs
```diff
@@ -1141,80 +1129,52 @@ impl RemoteStateActor {
     }
 }
 
+/// Data used during path selection.
+#[derive(Debug)]
+struct PathSelectionData {
+    /// Status of the path.
```
flub:
```suggestion
    /// Status of the path if it would be selected.
```

If you don't want to change this thing now, this also helps me understand what is going on.

iroh/src/socket/remote_map/remote_state.rs
```diff
@@ -1141,80 +1129,52 @@ impl RemoteStateActor {
...
+    /// Biased RTT for path selection.
+    ///
+    /// This is an i128 so we can subtract an advantage for e.g. IPv6 without underflowing.
+    biased_rtt: i128,
+}
+
+impl PathSelectionData {
+    /// Key for sorting paths. Lower is better.
+    ///
+    /// First part is the status, 0 for Available, 1 for Backup.
```
flub, line 1149: Oh, I never really realised that the order of fields of a pub type are part of the public API, in that you can cast them straight to their discriminant. I guess we can't stop you from doing this because we can't stop other folks from doing so. So I guess this isn't an actionable comment. But I still find this kind of unintended from my view. Oh well, better be careful.

iroh/src/socket/remote_map/remote_state.rs
```diff
@@ -1141,80 +1129,52 @@ impl RemoteStateActor {
...
-fn select_v4_v6(
-    addr_v4: SocketAddrV4,
-    rtt_v4: Duration,
-    addr_v6: SocketAddrV6,
-    rtt_v6: Duration,
-) -> (SocketAddr, Duration) {
-    if rtt_v6 <= rtt_v4 + IPV6_RTT_ADVANTAGE {
-        (addr_v6.into(), rtt_v6)
+    // Get current data. If we don't have data for the current path, don't switch.
+    let current_data = all_paths.get(addr)?;
```
flub: Looking at this function in isolation: this means that if we had a selected path which is no longer available, this would not switch away from it? That seems really surprising and wrong? If we don't have data for the current path should we not assume it no longer exists and select the best one out of the available paths?

iroh/src/socket/remote_map/remote_state.rs
rklaehn: Yeah, this is used all over the place for Ord instances. But it's the same for discriminants for serialization of enums, and for memory layout for structs.

I can make it more explicit once I add my own enum.

iroh/src/socket/remote_map/remote_state.rs
rklaehn: Ah, interesting. I was expecting selected_path to be None in this case. From the docs of selected_path, which is what is used to get addr:

```
    /// The path we currently consider the preferred path to the remote endpoint.
    ///
    /// **We expect this path to work.** If we become aware this path is broken then it is
    /// set back to `None`.
```

If selected_path is None, select_best_path will be called with current_path of None, and you will just return the best new path:

```rust
    let Some(addr) = current_path else {
        return Some((best_addr.clone(), best_data.rtt));
    };
```

rklaehn: Pushed an update.

iroh/src/socket/remote_map/remote_state.rs
```diff
@@ -1141,80 +1129,54 @@ impl RemoteStateActor {
...
-/// This prefers IPv6 paths over IPv4 paths.
-fn select_v4_v6(
-    addr_v4: SocketAddrV4,
-    rtt_v4: Duration,
-    addr_v6: SocketAddrV6,
-    rtt_v6: Duration,
-) -> (SocketAddr, Duration) {
-    if rtt_v6 <= rtt_v4 + IPV6_RTT_ADVANTAGE {
-        (addr_v6.into(), rtt_v6)
+    // Get current data. If we don't have data for the current path, don't switch.
```
flub: comment needs updating.

Issue #3871, opened 2026-01-19
transfer example feature request: instead of specifying a size, specify a duration for which to keep the connection open


b5:
and track how much data transfers in that time. This will help make our dogfooding tests less brittle

Frando:
Fixed by #3873

PR #3873, opened 2026-01-20
examples: improve transfer example

Frando:
## Description

Improves the transfer example we use for dogfooding:

* Change protocol such that the size is client- and not server-controlled
* Add `--mode` feature on `fetch` side to either download (like currently), upload, or transfer in both directions
* Add `--duration` feature on `fetch` side to limit by transfer duration instead of size
* Remove the unordered receive mode. It wasn't used (there was no cli flag so it was never enabled).
* Add `--output json` to print JSON instead of human-readable text.
* Improve docs

Because the protocol change is wire-breaking the ALPN is updated.

The JSON support was requested @Arqu a couple of time and should make our integration in netsim easier and less brittle. Output looks like this atm:

Provide side:
```
{"kind":"EndpointArgs","env":"Staging","relay_url":[],"no_relay":true,"no_discovery":false,"relay_only":false,"pkarr_relay_url":null,"no_pkarr_publish":false,"dns_origin_domain":null,"dns_server":null,"no_dns_resolve":false,"mdns":false,"bind_addr_v4":"127.0.0.1:1234","bind_addr_v4_additional":[],"bind_addr_v6":null,"bind_addr_v6_additional":[],"no_default_bind":false}
{"kind":"EndpointBound","endpoint_id":"5cc1982c0bc2612a2b62e7811242ec317576dace66688af743769e7eebdbb8c7","direct_addresses":["127.0.0.1:1234","[2003:f9:f724:8240:1566:102:9d0d:ea24]:33990","[2003:f9:f724:8240:a6e3:ea7a:0:edb]:33990"],"relay_url":null}
{"kind":"ConnectionAccepted","remote_id":"f2b872754fcaad0efac0bb57e3f19e38a8c4880f6c2789705f2b69b7a2211bed"}
{"kind":"ConnectionTypeChanged","status":"Selected","addr":{"Ip":"127.0.0.1:47519"},"rtt":2046,"remote_id":"f2b872754fcaad0efac0bb57e3f19e38a8c4880f6c2789705f2b69b7a2211bed"}
{"kind":"HandleRequest","request":{"Download":{"Duration":3000000}},"remote_id":"f2b872754fcaad0efac0bb57e3f19e38a8c4880f6c2789705f2b69b7a2211bed"}
{"kind":"UploadComplete","size":1216070367,"duration":3002590,"remote_id":"f2b872754fcaad0efac0bb57e3f19e38a8c4880f6c2789705f2b69b7a2211bed"}
{"kind":"ConnectionClosed","duration":3002772,"error":null,"remote_id":"f2b872754fcaad0efac0bb57e3f19e38a8c4880f6c2789705f2b69b7a2211bed"}
{"kind":"PathStats","paths":[{"remote_addr":{"Ip":"127.0.0.1:47519"},"rtt":95,"bytes_sent":813274,"bytes_recv":1244335203}],"remote_id":"f2b872754fcaad0efac0bb57e3f19e38a8c4880f6c2789705f2b69b7a2211bed"}
```

Fetch side:
```
{"kind":"EndpointArgs","env":"Staging","relay_url":[],"no_relay":true,"no_discovery":false,"relay_only":false,"pkarr_relay_url":null,"no_pkarr_publish":false,"dns_origin_domain":null,"dns_server":null,"no_dns_resolve":false,"mdns":false,"bind_addr_v4":null,"bind_addr_v4_additional":[],"bind_addr_v6":null,"bind_addr_v6_additional":[],"no_default_bind":false}
{"kind":"SecretGenerated","secret_key":"a0952bebe97106c30d6b22a50402d55d894b8e5e3d5b63d7e643bc5e6ad8f59a"}
{"kind":"EndpointBound","endpoint_id":"f2b872754fcaad0efac0bb57e3f19e38a8c4880f6c2789705f2b69b7a2211bed","direct_addresses":["172.16.10.33:47519","[2003:f9:f724:8240:1566:102:9d0d:ea24]:45524","[2003:f9:f724:8240:a6e3:ea7a:0:edb]:45524"],"relay_url":null}
{"kind":"Connected","remote_id":"5cc1982c0bc2612a2b62e7811242ec317576dace66688af743769e7eebdbb8c7","duration":3820}
{"kind":"StartRequest","mode":"Download","length":{"Duration":3000000}}
{"kind":"ConnectionTypeChanged","status":"Selected","addr":{"Ip":"127.0.0.1:1234"},"rtt":1162}
{"kind":"DownloadComplete","size":1216070367,"time_to_first_byte":5699,"num_chunks":31134,"duration":3004047}
{"kind":"ConnectionClosed","duration":3007952,"error":null}
{"kind":"EndpointClosed","duration":81556,"timed_out":false}
{"kind":"PathStats","paths":[{"remote_addr":{"Ip":"127.0.0.1:1234"},"rtt":84,"bytes_sent":1244335156,"bytes_recv":956724}]}
```

## Notes & open questions

This will likely break netsim.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/socket.rs
+++ b/iroh/src/socket.rs
@@ -157,6 +157,8 @@ pub(crate) struct Handle {
     sock: Arc<Socket>,
     // empty when shutdown
     actor_task: Arc<Mutex<Option<AbortOnDropHandle<()>>>>,
+    /// Channel to send to the internal actor.
+    actor_sender: mpsc::Sender<ActorMessage>,
     // quinn endpoint
     endpoint: quinn::Endpoint,
 }
@@ -200,12 +202,11 @@ impl ShutdownState {
 /// possible.
 #[derive(Debug)]
 pub(crate) struct Socket {
-    /// Channel to send to the internal actor.
-    actor_sender: mpsc::Sender<ActorMessage>,
     /// Channels for sending time-crucial messages to `RemoteStateActors`.
     ///
     /// Currently only exists to support sending `SendDatagram` messages.
     remote_actors: ReadOnlyMap<EndpointId, mpsc::Sender<RemoteStateMessage>>,
+
     /// EndpointId of this endpoint.
     public_key: PublicKey,
 
@@ -274,31 +275,6 @@ impl Socket {
         self.local_addrs_watch.clone().get()
     }
 
-    /// Registers the connection in the `RemoteStateActor`.
-    ///
-    /// The actor is responsible for holepunching and opening additional paths to this
-    /// connection.
-    ///
-    /// Returns a future that resolves to [`PathsWatcher`].
-    ///
-    /// The returned future is `'static`, so it can be stored without being liftetime-bound to `&self`.
-    pub(crate) fn register_connection(
-        &self,
-        remote: EndpointId,
-        conn: WeakConnectionHandle,
-    ) -> impl Future<Output = Result<PathsWatcher, RemoteStateActorStoppedError>> + Send + 'static
-    {
-        let (tx, rx) = oneshot::channel();
-        let sender = self.actor_sender.clone();
-        async move {
-            sender
-                .send(ActorMessage::AddConnection(remote, conn, tx))
-                .await
-                .map_err(|_| RemoteStateActorStoppedError::new())?;
-            rx.await.map_err(|_| RemoteStateActorStoppedError::new())
-        }
-    }
-
     #[cfg(not(wasm_browser))]
     fn ip_bind_addrs(&self) -> &[SocketAddr] {
         &self.ip_bind_addrs
@@ -308,69 +284,6 @@ impl Socket {
         self.local_addr()
             .into_iter()
             .filter_map(|addr| addr.into_socket_addr())
-    }
-
-    /// Resolves an [`EndpointAddr`] to an [`EndpointIdMappedAddr`] to connect to via [`Handle::endpoint`].
-    ///
-    /// This starts a `RemoteStateActor` for the remote if not running already, and then checks
-    /// if the actor has any known paths to the remote. If not, it starts address lookup and waits for
-    /// at least one result to arrive.
-    ///
-    /// Returns `Ok(Ok(EndpointIdMappedAddr))` if there is a known path or Address Lookup produced
-    /// at least one result. This does not mean there is a working path, only that we have at least
-    /// one transport address we can try to connect to.
-    ///
-    /// Returns `Ok(Err(address_lookup_error))` if there are no known paths to the remote and Address Lookup
-    /// failed or produced no results. This means that we don't have any transport address for
-    /// the remote, thus there is no point in trying to connect over the quinn endpoint.
-    ///
-    /// Returns `Err(RemoteStateActorStoppedError)` if the `RemoteStateActor` for the remote has stopped,
-    /// which may never happen and thus is a bug if it does.
-    pub(crate) async fn resolve_remote(
-        &self,
-        addr: EndpointAddr,
-    ) -> Result<Result<EndpointIdMappedAddr, AddressLookupError>, RemoteStateActorStoppedError>
-    {
-        let (tx, rx) = oneshot::channel();
-        self.actor_sender
-            .send(ActorMessage::ResolveRemote(addr, tx))
-            .await
-            .ok();
-        rx.await.map_err(|_| RemoteStateActorStoppedError::new())?
-    }
-
-    /// Fetches the [`RemoteInfo`] about a remote from the `RemoteStateActor`.
-    ///
-    /// Returns `None` if no actor is running for the remote.
-    pub(crate) async fn remote_info(&self, id: EndpointId) -> Option<RemoteInfo> {
-        let (tx, rx) = oneshot::channel();
-        self.actor_sender
-            .send(ActorMessage::RemoteInfo(id, tx))
-            .await
-            .ok()?;
-        rx.await.ok()
-    }
-
-    pub(crate) async fn insert_relay(
-        &self,
-        relay: RelayUrl,
-        endpoint: Arc<RelayConfig>,
-    ) -> Option<Arc<RelayConfig>> {
-        let res = self.relay_map.insert(relay, endpoint);
-        self.actor_sender
-            .send(ActorMessage::RelayMapChange)
-            .await
-            .ok();
-        res
-    }
-
-    pub(crate) async fn remove_relay(&self, relay: &RelayUrl) -> Option<Arc<RelayConfig>> {
-        let res = self.relay_map.remove(relay);
-        self.actor_sender
-            .send(ActorMessage::RelayMapChange)
-            .await
-            .ok();
-        res
     }
 
     /// Tries to send a [`RemoteStateMessage`] to the `RemoteStateActor` for given [`EndpointId`].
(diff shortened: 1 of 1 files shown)


b5:
I believe this closes #3871

iroh/examples/transfer.rs
```diff
@@ -1,11 +1,20 @@
+//! Transfer example: Transfer data between two endpoints and print stats.
+//!
+//! This example uses most of iroh's endpoint builder options and thus allows for flexible testing of connections
+//! with different discovery and transport configuration. Once a connection is established, it uses a simple protocol
+//! that transfers data between endpoints in both directions, measuring time and connection stats.
+//!
+//! The protocol is client-initiated and allows to set either a size or time limit on the transfers.
```
matheus23: Thank you for writing the example docs!

I think it'd be nice in here to say sth along the lines of "this example is mostly used for testing iroh connectivity and in iroh's CI jobs. It might not be the best example for learning about how to use iroh.".

iroh/examples/transfer.rs
```diff
@@ -167,15 +233,24 @@ struct EndpointArgs {
 enum Commands {
     /// Provide data.
     Provide {
-        #[clap(long, default_value = "100M", value_parser = parse_byte_size)]
-        size: u64,
```
matheus23, line 173: I think removing this might break netsim and a bunch of other things that use the transfer example.

Either we keep this backwards-compatible (in the CLI sense) or this PR needs to update all netsim things and maybe more.

iroh/examples/transfer.rs
```diff
@@ -366,19 +445,14 @@ impl EndpointArgs {
...
-async fn provide(endpoint: Endpoint, size: u64) -> Result<()> {
-    let endpoint_id = endpoint.id();
-    let endpoint_addr = endpoint.addr();
-
-    println!("Endpoint id:\n{endpoint_id}");
-    println!("Direct addresses:");
-    for addr in endpoint_addr.ip_addrs() {
-        println!("\t{addr}");
-    }
-    println!();
```
matheus23, line 378: I think we should still print these? Why did you remove them?

iroh/examples/transfer.rs
```diff
@@ -388,122 +462,148 @@ async fn provide(endpoint: Endpoint, size: u64) -> Result<()> {
...
-                    stats.rtt,
-                    stats.udp_tx.bytes,
-                    stats.udp_rx.bytes,
-                );
+        });
+    }
+
+    println!("Shutting down..");
+    endpoint.close().await;
+    println!("Endpoint closed");
```
matheus23: Thanks!

iroh/examples/transfer.rs
```diff
@@ -388,122 +462,148 @@ async fn provide(endpoint: Endpoint, size: u64) -> Result<()> {
...
+    Ok(())
+}
+
+#[tracing::instrument("conn", skip_all, fields(remote=%conn.remote_id().fmt_short()))]
+async fn handle_connection(conn: Connection) {
+    let endpoint_id = conn.remote_id();
+    info!(
+        "new connection from {endpoint_id} with ALPN {}",
+        String::from_utf8_lossy(TRANSFER_ALPN),
+    );
```
matheus23: I know this was in here before, but IMO we can just remove this.
Printing the ALPN doesn't give us any more info (it's hardcoded), and the fact that we get another connection is printed on stdout right below.

iroh/examples/transfer.rs
```diff
@@ -388,122 +462,148 @@ async fn provide(endpoint: Endpoint, size: u64) -> Result<()> {
...
+                [
+                    perform_request(&conn, RequestKind::Download, length, start),
+                    perform_request(&conn, RequestKind::Upload, length, start),
+                ]
+                .try_join_all()
+                .await?;
+            }
+        }
+        // We finished our requests. Close the connection with our graceful error code.
+        conn.close(GRACEFUL_CLOSE, b"done");
```
matheus23, line 611: This is incorrect: You can't close the connection when you just performed a `Request::Upload`.
This close will interrupt potentially queued stream frames.

I'd recommend changing the upload request protocol so that the provide side sends one final "ok I'm done", and *then* the fetch side can close the connection once it receives that.

iroh/examples/transfer.rs
```diff
@@ -388,122 +462,148 @@ async fn provide(endpoint: Endpoint, size: u64) -> Result<()> {
...
 
-    // We received the last message: close all connections and allow for the close
-    // message to be sent.
+    // Also close the endpoint, with a timeout.
     let shutdown_start = Instant::now();
-    let shutdown = if let Err(_err) = tokio::time::timeout(SHUTDOWN_TIME, endpoint.close()).await {
+    if let Err(_err) = tokio::time::timeout(SHUTDOWN_TIME, endpoint.close()).await {
         warn!(timeout = ?SHUTDOWN_TIME, "Endpoint closing timed out");
-        "Shutdown timed out".to_string()
+        println!("Shutdown timed out");
```
matheus23: I know this existed before, but can we remove the warning? Unless I'm missing something, I don't think it serves a purpose when there's also the println.

However, given how many duplicate tracing and printlns this file has, I suspect there must be a purpose behind them?

iroh/examples/transfer.rs
Frando: Fixed in latest commits

iroh/examples/transfer.rs
Frando: Removed in latest commits

iroh/examples/transfer.rs
Frando: Fixed

iroh/examples/transfer.rs
Frando: Yeah, it changes netsim. We can add it as `transfer2.rs` until netsim is updated?
I added JSON output so it will be easier for netsim now :)
But still it needs to be updated, which likely @arqu is best to do (I got quite lost in the chuck code last time I tried updating something).

iroh/examples/transfer.rs
matheus23: Okay, needed to dive into how you actually ended up doing this. You're basically sending a stream FIN from the receiving end on its `SendStream` *after* you received all data and make sure to wait for `recv.read_to_end(0).await` on the sending side to make sure that it's all there.

I... have never seen this before. But yeah I think that should work. Neat.

matheus23: Perhaps it's worth explaining this just a little bit more in the comments? :sweat_smile:

iroh/examples/transfer.rs
matheus23: Yeah maybe that's a way to go. We could call it `diagnostic-test` or something while we're at it. Although I'm not sure how fun it would be to try to maintain two similar-but-not-quite-the-same examples.
FWIW, Arqu is coming back tuesday next week.

Frando: Yeah, it should be a short-term thing for sure. One of these examples is enough.

Arqu: This is fine, I'll update netsim with the new ergonomics, no need to keep double examples or backwards compat here.

Arqu:
FWIW netsim is ready to merge: https://github.com/n0-computer/chuck/pull/88
Successful run against both branches here and on netsim: https://github.com/n0-computer/iroh/actions/runs/21290515507/job/61282577859

Issue #3867, opened 2026-01-16
bug: endpoint does not clean up resources swiftly when dropped ungracefully


ramfox:
- **We want users to call `endpoint::close` before the endpoint is dropped.** Calling `close` ensures that the endpoint is being a good/hygienic network citizen
- However, when an endpoint IS dropped, there is no communication that this breach in protocol happened.
- Furthermore, when it's dropped, it takes an unexpected amount of time to release its resources. (because quinn still monitors/babysits any open connections for PTO*3 amount of time before releases the connection resources.)
- This causes confusion for the user/dev. There is no indication for the dev that the endpoint was dropped un-gracefully BUT ALSO there are still resources that are being held onto.
- This can cause users issues when trying to recover. For example, the old endpoint may still attempt to connect to the relay using the given endpoint ID identity while the new endpoint tries to connect to it for the first time using that same endpoint ID, and they fight for contention until the old endpoint is finally cleaned up.

To fix both issues, we should:
1) communicating to the user that they are handling the endpoint improperly and
2) ensure that even when it handled improperly resources are released in an expected manner AKA we want to add the ability to drop everything in the endpoint on the floor when it is dropped un-gracefully.

b5:
this isn't release blocking, but @ramfox is going to try to get a PR up for this in the next day or two

PR #3879, opened 2026-01-22
feat: dropping the endpoint ungracefully should log and error, but still clean up resources immediately

ramfox:
## Description

closes https://github.com/n0-computer/iroh/issues/3867

relies on https://github.com/n0-computer/iroh/pull/3924

Alot of finagaling needed to happen to ensure that when the Endpoint is dropped, it's underlying resources are cleaned up ASAP.

First, added a custom `Runtime` that wraps spawned tasks in a `CancellationToken`. This allows for both graceful shutdown (which does not use the cancel token, but instead waits for tasks to finish naturally) and ungraceful abort (which uses the cancel token to stop all running tasks)

~~Second, the `quinn::Endpoint` inside of the `Handle` needed to be wrapped in an `Arc<RwLock<Option>>`, so we could take the underlying `quinn::Endpoint` and drop it on `Handle::close` and the new `Handle::abort` method. This value should only be written once (during the close or abort), and so the read lock should never be in contention.~~

**This was fixed in `iroh-quinn`, so no need to add a rw lock**

Third, wrote a new `Handle::abort` method that is sync and does not wait for graceful clean up after cancelling tokens, before dropping the underlying `quinn::Endpoint` or aborting any runtime tasks.

Fourth, in order to ensure that the `Endpoint` is only "aborted" on the last clone of the endpoint, the meat of the `Endpoint` has been moved into an `EndpointInner` that is wrapped in an `Arc`. When the `Arc` is finally dropped, the `Drop` impl on `EndpointInner` is called, which calls `Handle::abort`, if the `Endpoint` was not gracefully closed before dropping.

~~The last major change is the work that needed to be done on `Accept`. The call to `quinn::Endpoint::Accept` takes a reference to the `quinn::Endpoint`, but since it is now behind a `RwLock` it can not longer be borrowed. This refactor turned the `Accept` into a `BoxFuture`, whose closure owns a clone of the `quinn::Endpoint`.~~ 

Finally, let's make sure that we drop any references to address lookup services on close.

~~NOTE: While I used claude to code-review/double check assumptions while working on this PR in general, the `Accept` changes were where I relied on AI the most to give me a "sensible" work around to the self-reference problem I was facing when making the RwLock refactors. I would love some feedback here if this is an acceptable solution.~~
**No longer necessary since we do not wrap the quinn::Endpoint in a rwlock any more**

## Breaking Changes

New behavioral change! Then `Endpoint` does not do ANY best-effort to close connections gracefully before dropping. This makes it even more important to call `Endpoint::close` in all cases.

This change (that prevents the underlying `quinn::Endpoint` to do it's best to wait/send close frames before shutting down) has already highlighted in tests how much more brittle it is to drop the endpoint without calling `close` first.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/.github/sims/integration/iroh.json
+++ b/.github/sims/integration/iroh.json
@@ -1,0 +1,142 @@
+{
+    "name": "intg_iroh",
+    "cases": [
+        {
+            "name": "1_to_1",
+            "description": "",
+            "visualize": true,
+            "nodes": [
+                {
+                    "name": "i_srv",
+                    "count": 1,
+                    "cmd": "./bins/iroh-transfer --output json provide",
+                    "type": "public",
+                    "wait": 10,
+                    "connect": {
+                        "strategy": "none"
+                    },
+                    "param_parser": "iroh_endpoint_json"
+                },
+                {
+                    "name": "i_get",
+                    "count": 1,
+                    "cmd": "./bins/iroh-transfer --output json fetch --duration=10 --remote-direct-address=\"%s\" %s",
+                    "type": "public",
+                    "connect": {
+                        "strategy": "params_with_parsed_addrs",
+                        "node": "i_srv"
+                    },
+                    "process": "short",
+                    "parser": "iroh_json",
+                    "integration": "magic_iroh_client_json",
+                    "integration_require": {
+                        "transfer_success": "true",
+                        "final_conn_direct": "true"
+                    }
+                }
+            ]
+        },
+        {
+            "name": "1_to_3",
+            "description": "",
+            "nodes": [
+                {
+                    "name": "i_srv",
+                    "count": 1,
+                    "cmd": "./bins/iroh-transfer --output json provide",
+                    "type": "public",
+                    "wait": 10,
+                    "connect": {
+                        "strategy": "none"
+                    },
+                    "param_parser": "iroh_endpoint_json"
+                },
+                {
+                    "name": "i_get",
+                    "count": 3,
+                    "cmd": "./bins/iroh-transfer --output json fetch --duration=10 --remote-direct-address=\"%s\" %s",
+                    "type": "public",
+                    "connect": {
+                        "strategy": "params_with_parsed_addrs",
+                        "node": "i_srv"
+                    },
+                    "process": "short",
+                    "parser": "iroh_json",
+                    "integration": "magic_iroh_client_json",
+                    "integration_require": {
+                        "transfer_success": "true",
+                        "final_conn_direct": "true"
+                    }
+                }
+            ]
+        },
+        {
+            "name": "2_to_2",
+            "description": "",
+            "nodes": [
+                {
+                    "name": "i_srv",
+                    "count": 2,
+                    "cmd": "./bins/iroh-transfer --output json provide",
+                    "type": "public",
+                    "wait": 10,
+                    "connect": {
+                        "strategy": "none"
+                    },
+                    "param_parser": "iroh_endpoint_json"
+                },
+                {
+                    "name": "i_get",
+                    "count": 2,
+                    "cmd": "./bins/iroh-transfer --output json fetch --duration=10 --remote-direct-address=\"%s\" %s",
+                    "type": "public",
+                    "connect": {
+                        "strategy": "params_with_parsed_addrs",
+                        "node": "i_srv"
+                    },
+                    "process": "short",
+                    "parser": "iroh_json",
+                    "integration": "magic_iroh_client_json",
+                    "integration_require": {
+                        "transfer_success": "true",
+                        "final_conn_direct": "true"
+                    }
+                }
+            ]
+        },
+        {
+            "name": "2_to_4",
+            "description": "",
+            "nodes": [
+                {
+                    "name": "i_srv",
+                    "count": 2,
+                    "cmd": "./bins/iroh-transfer --output json provide",
+                    "type": "public",
+                    "wait": 10,
+                    "connect": {
+                        "strategy": "none"
+                    },
+                    "param_parser": "iroh_endpoint_json"
+                },
+                {
+                    "name": "i_get",
+                    "count": 4,
+                    "cmd": "./bins/iroh-transfer --output json fetch --duration=10 --remote-direct-address=\"%s\" %s",
+                    "type": "public",
+                    "connect": {
+                        "strategy": "params_with_parsed_addrs",
+                        "node": "i_srv"
+                    },
+                    "process": "short",
+                    "parser": "iroh_json",
+                    "integration": "magic_iroh_client_json",
+                    "integration_require": {
+                        "transfer_success": "true",
+                        "final_conn_direct": "true"
+                    }
+                }
+            ]
+        }
+    ]
+}
--- a/.github/sims/integration/iroh_full.json
+++ b/.github/sims/integration/iroh_full.json
@@ -1,0 +1,200 @@
+{
+    "name": "intg_iroh_full",
+    "cases": [
+        {
+            "name": "1_to_1",
+            "description": "",
+            "nodes": [
+                {
+                    "name": "1_r",
+                    "count": 1,
+                    "cmd": "./bins/iroh-relay --dev --config-path ./data/relay.cfg",
+                    "type": "public",
+                    "wait": 2,
+                    "connect": {
+                        "strategy": "none"
+                    }
+                },
+                {
+                    "name": "2_d",
+                    "count": 1,
+                    "cmd": "./bins/iroh-dns-server --config ./data/dns.test.cfg",
+                    "type": "public",
+                    "wait": 2,
+                    "connect": {
+                        "strategy": "none"
+                    }
+                },
+                {
+                    "name": "i_srv",
+                    "count": 1,
+                    "cmd": "./bins/iroh-transfer --output json provide --env dev --relay-url=\"http://10.0.0.1:3340\" --pkarr-relay-url=\"http://10.0.0.2:8080/pkarr\" --dns-origin-domain=\"10.0.0.2:5300\"",
+                    "type": "public",
+                    "wait": 10,
+                    "connect": {
+                        "strategy": "none"
+                    },
+                    "param_parser": "iroh_endpoint_json"
+                },
+                {
+                    "name": "i_get",
+                    "count": 1,
+                    "cmd": "./bins/iroh-transfer --output json fetch --size=1G --env dev --relay-url=\"http://10.0.0.1:3340\" --remote-relay-url=\"http://10.0.0.1:3340\" --pkarr-relay-url=\"http://10.0.0.2:8080/pkarr\" --dns-origin-domain=\"10.0.0.2:5300\" --remote-direct-address=\"%s\" %s",
+                    "type": "public",
+                    "connect": {
+                        "strategy": "params_with_parsed_addrs",
+                        "node": "i_srv"
+                    },
+                    "process": "short",
+                    "parser": "iroh_json"
+                }
+            ]
+        },
+        {
+            "name": "1_to_3",
+            "description": "",
+            "nodes": [
+                {
+                    "name": "1_r",
+                    "count": 1,
+                    "cmd": "./bins/iroh-relay --dev --config-path ./data/relay.cfg",
+                    "type": "public",
+                    "wait": 2,
+                    "connect": {
+                        "strategy": "none"
+                    }
+                },
+                {
+                    "name": "2_d",
+                    "count": 1,
+                    "cmd": "./bins/iroh-dns-server --config ./data/dns.test.cfg",
+                    "type": "public",
+                    "wait": 2,
+                    "connect": {
+                        "strategy": "none"
+                    }
+                },
+                {
+                    "name": "i_srv",
+                    "count": 1,
+                    "cmd": "./bins/iroh-transfer --output json provide --env dev --relay-url=\"http://10.0.0.1:3340\" --pkarr-relay-url=\"http://10.0.0.2:8080/pkarr\" --dns-origin-domain=\"10.0.0.2:5300\"",
+                    "type": "public",
+                    "wait": 10,
+                    "connect": {
+                        "strategy": "none"
+                    },
+                    "param_parser": "iroh_endpoint_json"
+                },
+                {
+                    "name": "i_get",
+                    "count": 3,
+                    "cmd": "./bins/iroh-transfer --output json fetch --size=1G --env dev --relay-url=\"http://10.0.0.1:3340\" --remote-relay-url=\"http://10.0.0.1:3340\" --pkarr-relay-url=\"http://10.0.0.2:8080/pkarr\" --dns-origin-domain=\"10.0.0.2:5300\" --remote-direct-address=\"%s\" %s",
+                    "type": "public",
+                    "connect": {
+                        "strategy": "params_with_parsed_addrs",
+                        "node": "i_srv"
+                    },
+                    "process": "short",
+                    "parser": "iroh_json"
+                }
+            ]
+        },
+        {
+            "name": "1_to_1ro",
+            "description": "",
+            "nodes": [
+                {
+                    "name": "1_r",
+                    "count": 1,
+                    "cmd": "./bins/iroh-relay --dev --config-path ./data/relay.cfg",
+                    "type": "public",
+                    "wait": 2,
+                    "connect": {
+                        "strategy": "none"
+                    }
+                },
+                {
+                    "name": "2_d",
+                    "count": 1,
+                    "cmd": "./bins/iroh-dns-server --config ./data/dns.test.cfg",
+                    "type": "public",
+                    "wait": 2,
+                    "connect": {
+                        "strategy": "none"
+                    }
+                },
+                {
+                    "name": "i_srv",
+                    "count": 1,
+                    "cmd": "./bins/iroh-transfer --output json provide --env dev --relay-only --relay-url=\"http://10.0.0.1:3340\" --pkarr-relay-url=\"http://10.0.0.2:8080/pkarr\" --dns-origin-domain=\"10.0.0.2:5300\"",
+                    "type": "public",
+                    "wait": 10,
+                    "connect": {
+                        "strategy": "none"
+                    },
+                    "param_parser": "iroh_endpoint_json"
+                },
+                {
+                    "name": "i_get",
+                    "count": 1,
+                    "cmd": "./bins/iroh-transfer --output json fetch --size=1G --env dev --relay-only --relay-url=\"http://10.0.0.1:3340\" --remote-relay-url=\"http://10.0.0.1:3340\" --pkarr-relay-url=\"http://10.0.0.2:8080/pkarr\" --dns-origin-domain=\"10.0.0.2:5300\" --remote-direct-address=\"%s\" %s",
+                    "type": "public",
+                    "connect": {
+                        "strategy": "params_with_parsed_addrs",
+                        "node": "i_srv"
+                    },
+                    "process": "short",
+                    "parser": "iroh_json"
+                }
+            ]
+        },
+        {
+            "name": "1_to_1x3",
+            "description": "",
+            "nodes": [
+                {
+                    "name": "1_r",
+                    "count": 1,
+                    "cmd": "./bins/iroh-relay --dev --config-path ./data/relay.cfg",
+                    "type": "public",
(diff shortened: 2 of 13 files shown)


matheus23:
> relies on #3879

This is a self-reference in the PR description?

iroh/src/runtime.rs
```diff
@@ -0,0 +1,142 @@
...
+        if self.tasks.is_closed() {
+            tracing::debug!(me = %self.id.fmt_short(), "runtime closed, dropping spawned task");
+            return;
+        }
+
+        use tracing::{Instrument, debug_span};
+
+        let task_id = self.task_counter.fetch_add(1, Ordering::Relaxed);
+        let cancel = self.cancel.clone();
+        let span = debug_span!("runtime", me = %self.id.fmt_short(), task_id);
```
matheus23: I think it's worth thinking about whether the added context is worth it or if just using `Span::current` is enough, since this will mean the log lines will use even more horizontal space and we generate even bigger log files.
I'd argue this might not be worth it over just using `Span::current`.

iroh/src/endpoint.rs
```diff
@@ -688,10 +690,28 @@ pub enum EndpointError {
...
+}
+
+/// Main purpose of this inner struct is to ensure that the `EndpointInner` is
+/// only dropped once, when the final reference to the `Endpoint` is dropped and
+/// the struct inside the `Arc` is freed.
+#[derive(Debug)]
+struct EndpointInner {
+    sock: Handle,
+    static_config: StaticConfig,
+}
```
matheus23: I think we're accumulating quite a few layers here.
`Handle`'s definition is this:
```rs
pub(crate) struct Handle {
    #[deref(forward)]
    sock: Arc<Socket>,
    // empty when shutdown
    actor_task: Arc<Mutex<Option<AbortOnDropHandle<()>>>>,
    /// Channel to send to the internal actor.
    actor_sender: mpsc::Sender<ActorMessage>,
    // quinn endpoint, set to `None` on close
    endpoint: Arc<RwLock<Option<quinn::Endpoint>>>,
    // Runtime used by quinn
    runtime: Arc<Runtime>,
}
```

So it effectively already `Arc`s all the things.

So when we add another `EndpointInner` which, again, is `Arc`ed, we're starting to repeat lots of this work.

I think it would be nice if we don't introduce yet another arcing layer. Not only that, this also means the `Drop` impl is more error prone: What if something else in the system has a reference to `Handle`, but not to `Endpoint`, and then that thing gets surprised when suddenly the `Handle`'s `Endpoint` gets removed?

Maybe the fields of `Handle` can become non-`Arc`, and it gains the `impl Drop` itself?
In that case, we could remove `EndpointInner`, and go back to just having an `Endpoint` with a `sock: Arc<Handle>`.

iroh/src/socket.rs
```diff
@@ -956,17 +994,60 @@ impl Handle {
...
+        trace!(me = ?self.public_key, "aborting socket...");
+
+        // Cancel at_close_start token, which cancels running netreports.
+        self.sock.shutdown.at_close_start.cancel();
+
+        // Cancel all actors.
+        self.sock.shutdown.at_endpoint_closed.cancel();
+
+        // Waits for the EndpointDriver and all ConnectionDrivers to shut down
+        self.runtime.abort();
```
matheus23: The comment is incorrect. This doesn't wait.

iroh/src/socket.rs
```diff
@@ -956,17 +994,60 @@ impl Handle {
...
+
+        // Drop our hold on the quinn endpoint.
+        self.endpoint.write().expect("poisoned").take();
+
+        // Drop the actor task.
+        self.actor_task.lock().expect("poisoned").take();
+
+        // Set socket to closed.
+        self.sock.shutdown.closed.store(true, Ordering::SeqCst);
```
matheus23: I don't think these comments are useful compared to reading the line of code itself and mostly pollute signal-to-noise ratio.

iroh/src/socket.rs
```diff
@@ -919,28 +952,33 @@ impl Handle {
...
+            // connection close codes, and close the endpoint properly.
+            // If this call is skipped, then connections that protocols close just shortly before the
+            // call to `Endpoint::close` will in most cases cause connection time-outs on remote ends.
+            debug!("wait_idle start");
+            ep.wait_idle().await;
+            debug!("wait_idle done");
+            // endpoint must be dropped here before we `take` it in the next lines.
+        }
+        // Drop our reference to the quinn endpoint.
+        self.endpoint.write().expect("poisoned").take();
```
matheus23: What's the reason for doing this?
Is there something in the `Drop` impl for `quinn::Endpoint` that requires this? Looking at the endpoint impl, the only thing this does is wake the `EndpointDriver` task.
If we forcefully abort all tasks from the outside anyways, I don't see what this accomplishes in addition.

Perhaps it would be useful to add a better comment here about what exactly this achieves, or even skip calling this. If we can manage to skip calling this, then that'd be amazing: We wouldn't need the `Option`, nor the `RwLock` around the endpoint, which means we can skip the changes to `Accept`, etc. So it might be worth trying that.
It might even be worth proposing changes to iroh-quinn that would make this easier, if this means we can skip the RwLock.

iroh/src/endpoint.rs
Frando: I agree with @matheus23  and also think we should look to have one layer less. 
If we'd move the `StaticConfig` into what currently is the `Handle`, we could have one layer less quite easily. We already store the `quinn::Endpoint` within the `Handle` for a while already, so the separation between `Handle`  and `Endpoint` isn't very important anymore IMO.

I think my suggestion would be something like this:

```rust
struct Endpoint(Arc<EndpointInner>);

// Replace `Handle` with this:
// Not `Clone`!
struct EndpointInner {
    // Still has to be `Arc`d separately because it is cloned to other places
    sock: Arc<Socket>,
    // No more Arc!
    // empty when shutdown
    actor_task: Mutex<Option<AbortOnDropHandle<()>>>,
    /// Channel to send to the internal actor.
    actor_sender: mpsc::Sender<ActorMessage>,
    // No more Arc!
    // quinn endpoint, set to `None` on close
    endpoint: RwLock<Option<quinn::Endpoint>>,
    // Runtime used by quinn
    runtime: Arc<Runtime>,
    // moved here from Endpoint
    static_config: StaticConfig
}
```

iroh/src/socket.rs
Frando: I think what dropping the `Endpoint` achieves is that the `AsyncUdpSocket` that it was given will be dropped, which frees the sockets. So without this the sockets would remain bound until the last `Endpoint` is dropped. Which - might be fine? Aborting the tasks is the far more important thing, and as noted here for that we don't need to drop the endpoint, we can abort the tasks via the runtime, which is great. Might be worth to check what happens when this line is removed. Because I agree that it would be much nicer to not have to wrap the `quinn::Endpoint` in an `RwLock<Option>>`.

iroh/src/socket.rs
```diff
@@ -956,17 +994,60 @@ impl Handle {
...
 
+        // Waits for the EndpointDriver and all ConnectionDrivers to shut down
+        self.runtime.shutdown().await;
+
         self.sock.shutdown.closed.store(true, Ordering::SeqCst);
 
         trace!("socket closed");
     }
 
+    /// Aborts the connection ungracefully:
```
Frando:
```suggestion
    /// Aborts the endpoint ungracefully:
```

iroh/src/socket.rs
Frando: If we want to free the sockets here, I think we could add this to our socket abstraction, it already has an `RwLock` around the actual socket to support rebinding, so adding support for `unbind` there might be the better way.

iroh/src/endpoint.rs
flub: I like @Frando 's suggestion here

iroh/src/endpoint/connection.rs
```diff
@@ -53,27 +52,25 @@ use crate::{
...
-#[pin_project]
-pub struct Accept<'a> {
-    #[pin]
-    #[debug("quinn::Accept")]
-    pub(crate) inner: quinn::Accept<'a>,
-    pub(crate) ep: Endpoint,
+pub struct Accept(#[debug("BoxFuture")] BoxFuture<Option<Incoming>>);
+
+impl Accept {
+    pub(crate) fn new(quinn_ep: Option<quinn::Endpoint>, ep: Endpoint) -> Self {
```
Frando: I guess this works, yes.

What will happend if I do:
```rust
let ep = iroh:::Endpoint::bind().await?;
let accept_fut = ep.accept();
ep.close().await;
let incoming = accept_fut.await;
```
will it return `None`?

Ideally, we can get rid of the `RwLock` around the `quinn::Endpoint` again, see below. I don't like the complexity much that this adds to these futures.

iroh/src/socket.rs
ramfox: 😂 😂 😂 ohhhhh man hold onto your butts about why this is necessary. Looking back at this PR, I didn't make it clear at all about why certain decisions that look batshit were made. Unfortunately, the only way that dropping both "immediately" AND "gracefully with a custom runtime" works at all is because of that stupid rwlock/option. There may be other versions of this that get it to work with much more disruptive refactors (including refactors to iroh-quinn), and I am v open to hearing about them, once you see the dumb headache we have.

Once thing to note here, is that for all decisions, we need to ensure it's works as expected for both the graceful shutdown and the ungraceful abort.

## The quinn::Endpoint is our problem child

The ultimate thing that screws us is that we need access to the `quinn::Endpoint` in the `Handle`.

Why? Because we need to call `connect_with`, `set_server_config`, `accept`, `close`, and `wait_idle` on the quinn::Endpoint.

However, **the `EndpointDriver` inside the `quinn::Endpoint` will only ever close if all `quinn::Endpoint`s have been dropped.** SO, part of the reason that we have so much trouble closing the iroh endpoint is because, until the `Handle` is dropped, we keep holding onto the `quinn::Endpoint` that we no longer want to be active...but it is active and continues to poll under the hood. So while dropping the `quinn::Endpoint` explicitly may also have the effect of unbinding sockets, it's not the actual issue here. Having the sockets open isn't a problem, it's that even when explicitly closed (`quinn::Endpoint::close` & `quinn::Endpoint::wait_idle`), the `EndpointDriver` is still polling until the last reference to that `quinn::Endpoint` is dropped.


## RwLock???

As for why the RwLock: We need to be able to drop the `quinn::Endpoint` explicitly, but because of our current API structure & expectations, we have to use interior mutability (both Endpoint::close and Handle::shutdown take &self). The `write` action only happens once ever: during graceful or ungraceful shutdown, so the lock never actually has any contention. One thing to be careful/aware of, is that now that we give out clones of `quinn::Endpoint` rather than `&` to it, a future dev could, in later refactors, clone the quinn endpoint and store it somewhere and not realize that they have made it impossible to shutdown cleanly. So this needs more docs, for sure.

Caveat:
We COULD move the `quinn::Endpoint` into it's own actor. Then the handle would not be holding onto a `quinn::Endpoint` itself - it would just hold onto the task, and when we close the `quinn::Endpoint` task, then the last `quinn::Endpoint` would be dropped. But that would then require sending messages back and forth between the actor and the handle & we would have to deal with channels and even more boxing of futures to ensure that we aren't blocking the channels with async work. The rwlock is annoying, but it is a lock with no contention, and this seems simpler/quicker/less disruptive that moving the `quinn::Endpoint` into an actor.

If there is another construction that I'm not seeing (or if you think the actor patterns is in fact better here) I would love suggestions. I hate being the person who is proposing to add a lock to our codebase 😂 

## runtime fun

So, adding the `Runtime` in this PR isn't actually necessary to get a clean drop and a clean graceful shutdown, and it may have contributed to some confusion here.

Long term, my thought was we may actually want to lean on the `Runtime` more, and it could be a better way for us to manage all our spawned tasks that we use in the iroh endpoint (the socket actor, the net-reports, eg). We could then get rid of the `ShutdownState` in the Handle, and move most of the clean up to the Runtime.

But back to why the runtime work may have added some unnecessary confusion here: In the graceful shutdown case, the `quinn::Endpoint`'s `EndpointDriver` will never close until all `quinn::Endpoint`s are dropped. So we need to `take` and drop the `quinn::Endpoint` before we `Runtime::shutdown` to close all of the runtime tasks gracefully, otherwise it hangs forever. This is actually how I initially caught our shutdown problem in the first place; it was what made it obvious that the `EndpointDriver` is never closed and continues to poll the quinn endpoint even after we have "closed" the iroh endpoint.

Funnily enough, this chicken & egg problem doesn't matter in the abort/drop case, because we just cancel all of the tasks, so the task powering the `quinn::Endpoint` is dropped on the floor. This chicken & egg issue only effects the graceful version of shutdown, when we also explicitly manage runtime tasks using a custom `Runtime`.

iroh/src/runtime.rs
ramfox: Having an endpoint id and a task id was the only way I could debug the runtime issues in tests.

But maybe that is just something one of us adds when they need to explicitly debug runtime issues? I understand why this would be noisy for most other cases. Is it reasonable to move to trace rather than remove?

iroh/src/socket.rs
ramfox: To philipp's point about wanting to skip this and the rwlock entirely...I would rather that too. And the best place may be to make refactors in `iroh-quinn`, but I don't have enough knowledge to know offhand how easy that would be/how disruptive to the current work that is happening on multipath.

iroh/src/socket.rs
ramfox: oh my bad, copy-pasted from graceful shutdown and never changed it...ty

iroh/src/endpoint/connection.rs
ramfox: Okay thank you for this feedback because this was the exact thing I was looking for about why boxing the future might cause issues, but didn't personally feel like I had enough prior knowledge about what to investigate to ensure I wasn't mucking everything up.

In the commit that you reviewed: no, this did not work (it actually hung during `ep.close().await`, because the boxed future held a clone of the `quinn::Endpoint`, so the endpoint never closed down). But I've added tests and fixed the issue, and now it does return `None`, as expected.

ramfox: But for the record, I also hate the rw lock

iroh/src/socket.rs
Frando: Thank you for all the details.

I think we should allow the EndpointDriver to stop after it has been closed and all connections are drained. I don't think we need to wait for all Endpoint structs to be dropped.

Here's a PR to quinn that makes the EndpointDriver stop once the endpoint is closed and all connections are drained: https://github.com/n0-computer/quinn/pull/426

Frando: Here's a PR on top of this PR that uses quinn#426 and removes the lock, seems to work! https://github.com/n0-computer/iroh/pull/3941

Frando: Ok, with #3941 we have the issue that `watch_addr` does not complete, because it references a watchable that is owned by Transports which is owned by the quinn::Endpoint which is no longer dropped on close (but stopped) with my branch. I've been discussing a bit with @matheus23 (see discord thread about `Watchable::close` in the iroh channel) and he argued that it could be fine if that just doesn't stop?

Also, even on your branch we'd have to carefully check if it is ensured that the watcher stream *always* stops on drop. It might not if there are still `Connections` which are not dropped, because they contain a `UdpSender` each which (might or might not) include the Watchable through shared Arcs (did not check this).

Alternatively, we could add some mechanism to the Watchables or Watchers to allow to force-close them if we don't want to rely on drop semantics. Or we document that it is not guaranteed for the watcher streams to stop on endpoint close (but only once everything is dropped), and add a `Endpoint::closed` function that returns a future that completes once the endpoint is closed, and advice people to `tokio::select!` on this if they want to terminate things together with the endpoint. Or even return a `CancellationToken` from `Endpoint::closed`, then users can do `run_until_cancelled`. Wdyt?

Apart from the issue about the watch_addr watchable being owned by `Transports` owned by `quinn::Endpoint` my branch with the quinn PR mentioned above passes all tests.

Frando: Here's a commit on top of #3941 that adds `Endpoint::closed` and documents that the watchers only become disconnected once the endpoint is dropped.
https://github.com/n0-computer/iroh/pull/3943/changes

iroh/src/runtime.rs
matheus23: Ah interesting. I wasn't even aware of how trace spans work until now. Yeah that sounds like a good idea :)

iroh/src/socket.rs
ramfox: both PRs have been merged into this one! thank you!

iroh/src/socket.rs
```diff
@@ -146,21 +148,63 @@ pub(crate) struct Options {
...
     #[deref(forward)]
     sock: Arc<Socket>,
     // empty when shutdown
-    actor_task: Arc<Mutex<Option<AbortOnDropHandle<()>>>>,
+    actor_task: Mutex<Option<AbortOnDropHandle<()>>>,
     /// Channel to send to the internal actor.
     actor_sender: mpsc::Sender<ActorMessage>,
-    // quinn endpoint
+    // quinn endpoint, set to `None` on close
     endpoint: quinn::Endpoint,
```
matheus23, line 168: This comment change is outdated now ✌️

iroh/src/endpoint.rs
```diff
@@ -1393,38 +1415,38 @@ impl Endpoint {
...
-        self.sock.is_closed()
+        self.inner.is_closed()
+    }
+
+    /// Returns a [`CancellationToken`] that is cancelled once the endpoint is shutting down.
+    ///
+    /// Cancelling the [`CancellationToken`] has no effect. Use [`Endpoint::close`] to close the endpoint.
+    pub fn closed(&self) -> CancellationToken {
+        self.inner.closed()
     }
```
matheus23, line 1448: I think I'd prefer this to work the same as `Connection::closed`: Not returning a `CancellationToken`, but instead maybe make this an `async fn closed(&self) { ... }` that effectively resolves to `()` when the endpoint is closed.

iroh/src/socket.rs
```diff
@@ -886,18 +934,25 @@ impl Handle {
...
 
-    /// The underlying [`quinn::Endpoint`]
-    pub fn endpoint(&self) -> &quinn::Endpoint {
+    /// Returns a clone of the underlying [`quinn::Endpoint`], if the handle hasn't been closed.
+    ///
+    /// We should be very careful about using this method, every clone of `quinn::Endpoint`
+    /// is another case we need to pay attention to to ensure it's dropped in a
+    /// timely manor. It can interfere with closing the `EndpointInner` cleanly (and
+    /// therefore the `iroh::Endpoint`).
+    pub(crate) fn quinn_endpoint(&self) -> &quinn::Endpoint {
```
matheus23: These docs changes need to be reverted again :v:

matheus23:
The PR description is also still weirdly self-referential with the "#3879". I worry you wanted to link something else there instead?

iroh/src/address_lookup.rs
```diff
@@ -460,6 +460,12 @@ impl ConcurrentAddressLookup {
     pub fn len(&self) -> usize {
         self.services.read().expect("poisoned").len()
     }
+
+    /// Removes all configured services.
+    pub fn clear(&self) {
+        let mut services = self.services.write().expect("poisoned");
+        *services = Vec::new();
```
dignifiedquire: nit: `services.clear()`?

iroh/src/socket.rs
```diff
@@ -947,11 +1001,11 @@ impl Handle {
         // connection close codes, and close the endpoint properly.
         // If this call is skipped, then connections that protocols close just shortly before the
         // call to `Endpoint::close` will in most cases cause connection time-outs on remote ends.
-        trace!("wait_idle start");
-        self.endpoint.wait_idle().await;
-        trace!("wait_idle done");
+        debug!("wait_idle start");
```
dignifiedquire: should this be debug?

iroh/src/address_lookup.rs
dignifiedquire: did you plan on changing or not?

matheus23: Please address dig's nit here. `self.services.write().expect("poisoned").clear()`.

ramfox: oh bother, I didn't push...sigh.

PR #3885, opened 2026-01-23
ci(dependabot): skip patch versions

flub:
## Description

dig: "I usually skip minor changes e.g. 1.3.1 -> 1.3.2"

Let's automate that. Or try at least.

## Breaking Changes

n/a

## Notes & open questions

n/a

## Change checklist

- [x] yolo test in CI


Frando:
lgtm :+1: 

If we still wanted to do patch updates, I think it is possible to configure dependabot to do something like a biweekly rollup of all patch updates together in a single PR

PR #3887, opened 2026-01-26
refactor!: goodbye magic

dignifiedquire:
Renames the remaining uses of magic to simpler versions. 

## Breaking Changes

- `Endpoint::magic_sock()` → `Endpoint::socket()`
-  `MagicsockMetrics` → `SocketMetrics`
- `EndpointMetrics::magicsock` → `EndpointMetrics::socket`
-  Metrics name `"magicsock"` → `"socket"`

### DIFF
--- a/iroh/src/socket.rs
+++ b/iroh/src/socket.rs
@@ -1932,15 +1932,15 @@ mod tests {
     #[traced_test]
     async fn test_direct_addresses() {
         let mut rng = rand_chacha::ChaCha8Rng::seed_from_u64(0u64);
-        let ms = Handle::new(default_options(&mut rng)).await.unwrap();
+        let sock = Handle::new(default_options(&mut rng)).await.unwrap();
 
         // See if we can get endpoints.
-        let eps0 = ms.ip_addrs().get();
+        let eps0 = sock.ip_addrs().get();
         info!("{eps0:?}");
         assert!(!eps0.is_empty());
 
         // Getting the endpoints again immediately should give the same results.
-        let eps1 = ms.ip_addrs().get();
+        let eps1 = sock.ip_addrs().get();
         info!("{eps1:?}");
         assert_eq!(eps0, eps1);
     }


iroh/src/endpoint/connection.rs
```diff
@@ -268,7 +268,7 @@ fn conn_from_quinn_conn(
         }
     };
 
-    // Register this connection with the magicsock.
+    // Register this connection with the socket.
     let fut = ep
         .msock
```
flub: Is this field name still apropriate? Should it be `sock` now?

iroh/src/socket.rs
```diff
@@ -1066,9 +1064,9 @@ enum ActorMessage {
 }
 
 struct Actor {
-    msock: Arc<MagicSock>,
+    msock: Arc<Socket>,
```
flub: `msock` here probably also the wrong field name.

iroh/src/socket/transports.rs
```diff
@@ -618,20 +618,20 @@ impl TransportsSender {
 /// knows about these and maps them back to the transport [`Addr`]s used by the wrapped
 /// [`Transports`].
 #[derive(Debug)]
-pub(crate) struct MagicTransport {
-    msock: Arc<MagicSock>,
+pub(crate) struct Transport {
```
flub, line 621: Can't say I'm wild about having `Transports` and `Transport`, in the same module even.

Given the first line of the doc comment, how about `MultipathMappedTransport`? Or something in that direction.

dignifiedquire: we already call them transports all the time, and have so for a while

iroh/src/socket.rs
dignifiedquire: done, `msock` -> `sock`

iroh/src/endpoint/connection.rs
dignifiedquire: done

iroh/src/socket/transports.rs
flub: transports are a thing that acts on a `transport::Addr`, this thing acts on a `SocketAddr`. I still think these names are way to close together, but whatever. We can rename this next time someone works on things around here.

PR #3890, opened 2026-01-27
refactor(iroh): move socket actor_sender out of shared socket state

Frando:
## Description

This is a tiny refactor without any functional changes. It moves the `actor_sender` to send messages to the socket actor out of the `Socket` struct and into the `Handle`. The `Socket` is shared between the actor task and the outside world. By moving the sender out of the socket, the actor can no longer send messages to itself. This is a good thing, because an actor sending to its own inbox is generally a footgun (can easily deadlock the channel). We don't do this, which is good and which this refactor reflects in the code.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


Frando:
Rebased after #3841 was merged

PR #3893, opened 2026-01-28
examples(iroh): add `--logs`  and `--mode ping` features to transfer example

Frando:
## Description

More improvements to our `transport` example to ease dogfooding:

* new `--logs` flag (for both provide and fetch) for simple file logging:
  - without further config, creates a directory `./logs/transfer-{cmd}-{date}-{endpoint_id}` and saves logs with `iroh=trace,transfer=trace` to a `logs` file in that directory.
  - different log level can be set with RUST_LOG
  - different path can be set with `--logs-path`
  - qlog files are also written to that directory if the `qlog` feature is enabled
* New `--mode ping` option for `fetch` command: instead of transferring as much data as possible, this just keeps the connection open until the `--duration` is expired (default: 10s). This gives much smaller log files (both tracing and qlog) so easier to debug when looking for connection changes only.
* Regular output now includes the time since start after each line. Useful to interpret the "switched to direct" etc lines after the fact (they didn't have any time indication so far)
* Better formatting for durations
* Better spans on accept side. Each connection gets an `id` now that easily allows to grep for that connection through both iroh and quinn logs
* Print path ids with the paths stats and changes. This allows to correlate with the tracing logs now.

With both logs and qlog in a single dir it's now really straightforward to share it eg with sendme.

## Notes & open questions

Should `--logs` include `iroh_quinn=trace` by default or not?

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/socket.rs
+++ b/iroh/src/socket.rs
@@ -753,10 +753,13 @@ impl Handle {
         let (actor_sender, actor_receiver) = mpsc::channel(256);
 
         #[cfg(not(wasm_browser))]
-        let ipv6 = transports
+        let has_ipv6_transport = transports
             .ip_bind_addrs()
             .into_iter()
             .any(|addr| addr.is_ipv6());
+
+        #[cfg(not(wasm_browser))]
+        let has_ip_transports = !transports.ip_bind_addrs().is_empty();
 
         let direct_addrs = DiscoveredDirectAddrs::default();
 
@@ -816,8 +819,6 @@ impl Handle {
             .await
             .map_err(|err| e!(BindError::CreateNetmonMonitor, err))?;
 
-        let qad_endpoint = endpoint.clone();
-
         #[cfg(any(test, feature = "test-utils"))]
         let client_config = if insecure_skip_relay_cert_verify {
             iroh_relay::client::make_dangerous_client_config()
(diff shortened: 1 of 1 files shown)


iroh-base/src/endpoint_addr.rs
```diff
@@ -45,10 +45,13 @@ pub struct EndpointAddr {
...
 
 /// Available address types.
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(
+    derive_more::Debug, Clone, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash,
+)]
 #[non_exhaustive]
 pub enum TransportAddr {
     /// Relays
+    #[debug("Relay({_0})")]
```
matheus23, line 54: Does that do anything? Isn't this the same output as before?

iroh/examples/transfer.rs
```diff
@@ -1169,3 +1287,29 @@ mod duration_micros {
...
+
+fn fmt_duration(d: Duration) -> impl fmt::Display {
+    if d > Duration::from_secs(1) {
+        format!("{:.2}s", d.as_secs_f32())
+    } else if d > Duration::from_millis(1) {
+        format!("{}ms", d.as_millis())
+    } else {
+        format!("{}µs", d.as_micros())
+    }
+}
```
matheus23, line 1315: Is this worth it compared to `Duration`'s `Debug` impl?

On use sites, we can instead use `{duration:?}` in all cases I think.
This just replicates the `impl Debug for Duration`, but as a more incomplete version.

iroh/src/socket/remote_map/remote_state.rs
```diff
@@ -1559,6 +1559,13 @@ impl PathInfo {
     pub fn rtt(&self) -> Duration {
         self.stats().rtt
     }
+
+    /// Returns the QUIC path id of this path.
+    ///
+    /// This can be used to uniquely identify a transmission path.
+    pub fn id(&self) -> PathId {
+        self.path_id
+    }
```
matheus23, line 1571: We should probably make sure that there's a `pub use quinn::PathId`, if we expose this.

iroh/examples/transfer.rs
Frando: Yeah, the difference is that the the secs are limited to two numbers after the comma. otherwise all are printed and it can get really long.

iroh/src/socket/remote_map/remote_state.rs
Frando: There is now on `main`.

iroh-base/src/endpoint_addr.rs
Frando: It uses the display impl not the debug impl, so it now is `Relay("https://foo.bar")` and before it was `Relay(RelayUrl("https://foo.bar"))`

matheus23 (approved):

PR #3895, opened 2026-01-28
refactor(transfer example): Remove `SelectedPath::Mixed` as an impossible state

matheus23:
## Description

There should never be a situation where multiple paths are selected, that's impossible from the way `RemoteStateActor` works.

This updates the example so that `SelectedPath::Mixed` is removed.


Frando (approved):
LGTM.

PR #3899, opened 2026-01-29
feat(iroh)!: Retain stats for closed and abandoned paths in the path watcher

Frando:
## Description

Improves the path watcher for connections to reliably allow accessing the stats of abandoned paths.

* Abandoned paths are not removed from the value of the watchable for a connection's paths, as long as there are active watchers (subscribers)
* Only once all watchers are dropped, we drop the abandoned paths from the watchable's value on the next update
* This means that if you keep a `PathWatcher` alive for the duration of a connection, it will contain *all* paths the connection ever used. 
* We use the changes from https://github.com/n0-computer/quinn/pull/386 and keep a `WeakPathHandle` for all paths in the watchable's value. This prevents the path stats to be dropped within a quinn connection even if the path is abandoned.
* We do no longer store path stats anywhere in iroh directly, but only ever access them from quinn. Due to keeping the `WeakPathHandle`, we can ensure that the stats are always available as long as the connection hasn't been dropped.

The combination of all this gives us a reliable way to access final path stats for all paths used in a connection, as long as you keep a reference to the connection around, which is quite straightforward to do and document.

## Breaking Changes

* `Connection::paths` and `ConnectionInfo::paths` now return  a `PathWatcher` (which still implements `n0_watcher::Watcher` but now also is a named struct)
* `PathInfo::stats` and `PathInfo::rtt`  now return `Option`. They return `None` if the underlying connection has been droped.

## Notes & open questions

I spend quite some time going back-and-forth over different approaches. Very open to other ideas, but I'm a bit out of further ideas currently and this is the best I could come up with so far.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/socket/remote_map/remote_state/path_watcher.rs
+++ b/iroh/src/socket/remote_map/remote_state/path_watcher.rs
@@ -261,7 +261,7 @@ impl PathInfo {
 
 /// Watchable for the network paths in a connection.
 ///
-/// This contains a watchable over a [`PathWatchValue`], and a watchable over the selected path for a remote.
+/// This contains a watchable over a [`CloseablePathList`], and a watchable over the selected path for a remote.
 ///
 /// This struct is owned by the [`super::ConnectionState`] and also cloned into the [`Connection`].
 /// Most methods are `pub(super)`. The only method that may be called from [`Connection`] is


iroh/src/socket/transports.rs
```diff
@@ -530,6 +530,17 @@ impl Addr {
             Self::Relay(..) => None,
         }
     }
+
+    pub(crate) fn is_transport_addr(&self, transport_addr: &TransportAddr) -> bool {
```
dignifiedquire: could this be an impl of PartialEq instead?

Frando: Yes, why not. Changed it.

iroh/src/endpoint/connection.rs
```diff
@@ -1005,21 +1004,35 @@ impl Connection<HandshakeCompleted> {
         self.data.info.endpoint_id
     }
 
-    /// Returns a [`Watcher`] for the network paths of this connection.
+    /// Returns a watcher for the network paths of this connection.
```
flub: Why not a link? Links are great! People won't know what a watcher is. And the fact you repeat the link below is just great, repetition is good.

iroh/src/socket/remote_map/remote_state/path_watcher.rs
```diff
@@ -0,0 +1,339 @@
...
+impl IntoIterator for PathWatcher {
+    type Item = PathInfo;
+    type IntoIter = PathInfoListIntoIter;
+
+    fn into_iter(mut self) -> Self::IntoIter {
+        self.get().into_iter()
+    }
+}
+
+impl Iterator for PathInfoListIntoIter {
```
flub: I'm not sure I `PathInfoListIntoIter` is the right name, this is the iterator itself is it not? `PathInfoListIter` would make more sense to me. It seems SmallVec uses the same naming though, so maybe I'm just wrong.

iroh/src/socket/remote_map/remote_state/path_watcher.rs
```diff
@@ -0,0 +1,339 @@
...
+            path: path.weak_handle(),
+            remote_addr,
+            is_closed: path.status().is_err(),
+            is_abandoned: false,
+            is_selected: false,
+        })
+    }
+
+    /// Returns the [`PathId`] of this path.
+    pub fn id(&self) -> PathId {
```
flub, line 211: Does iroh export (pub use) the `PathId` type?

iroh/src/socket/remote_map/remote_state/path_watcher.rs
```diff
@@ -0,0 +1,339 @@
...
+/// Information about a network transmission path used by a [`Connection`].
+///
+/// [`Connection`]: crate::endpoint::Connection
+#[derive(derive_more::Debug, Clone, Eq, PartialEq)]
+pub struct PathInfo {
+    #[debug("{}", path.id())]
+    path: WeakPathHandle,
+    remote_addr: TransportAddr,
+    is_closed: bool,
+    is_abandoned: bool,
```
flub: What do these values mean? `is_abandoned` is never even read and only set?

FWIW the great path state bikeshed resulted in having states:
- <some prior states>
- abandoned
- discarded

Going from any prior state to abandoned is "closing" (a verb), a rather loose term.

Looking at where you call `set_closed` and `set_abandoned`, I think that:
- `is_abandoned` -> `is_discarded`
- `is_closed` -> `is_abandoned`

iroh/src/socket/remote_map/remote_state/path_watcher.rs
```diff
@@ -0,0 +1,339 @@
...
+
+    /// Returns `true` if this path is currently the main transmission path for this [`Connection`].
+    ///
+    /// [`Connection`]: crate::endpoint::Connection
+    pub fn is_selected(&self) -> bool {
+        self.is_selected
+    }
+
+    /// Returns `true` if this path is closed.
+    pub fn is_closed(&self) -> bool {
```
flub, line 233: This public method name I find reasonable, despite my complaints about the internal naming. For someone without any multipath knowledge this can be helpful and the distinction between abandoned and discarded is not much help to the user.

I would however add a longer description saying that this really returns true as soon as a path has been abandoned by the local endpoint, and remains true forever after. So that folks who do know more about multipath can make sense of it.

iroh/src/socket/remote_map/remote_state/path_watcher.rs
```diff
@@ -0,0 +1,339 @@
...
+            if let Some(item) = list.iter_mut().find(|p| p.path.id() == id) {
+                item.is_closed = true;
+                item.is_abandoned = true;
+            }
+        });
+    }
+
+    /// Updates the watchable's value through a closure.
+    ///
+    /// After the update is performed, and if there are currently no watchers, data for abandoned paths
```
flub, line 325: ah, this is where is_abandoned is read. This reaffirms to me that is_discarded is the better name here.

flub (commented):
Generally LGTM (despite my sadness is discord), only the path state naming needs to be sorted. I know not everywhere uses the right terminology yet, e.g. the RemoteStateActor does also use some somewhat lose terms. We'll update things bit by bit to get to the consistent naming.

iroh/src/socket/remote_map/remote_state/path_watcher.rs
Frando: Yeah, I guess the `IntoIter` naming in e.g. SmallVec comes about when you have both an owned iterator (returned from `into_iter`) and a non-owned iterator (returned from `_iter`). The latter we also have but as unnames `impl Iterator` type. So for us there's no need to have the convoluted name, agreed. Renamed to `PathInfoListIter`.

iroh/src/socket/remote_map/remote_state/path_watcher.rs
Frando: It does

iroh/src/socket/remote_map/remote_state/path_watcher.rs
Frando: I checked this again, and I think a single flag `is_abandoned` is enough. The difference whether a path is discarded or not does not matter. It was a leftover, I think, from when the discarded state meant we would have to track stats differently, but this is no longer the case now that a `WeakPathHandle` keeps the path stats available in the quinn Connection.
So with the latest commit, there's now only a single `is_abandoned` state being tracked. Nice.

iroh/src/socket/remote_map/remote_state/path_watcher.rs
Frando: I added a comment. Thought about renaming to `is_abandoned` but maybe for the public API `is_closed` is better indeed, left it for now.

iroh/src/endpoint/connection.rs
Frando: Done

iroh/src/endpoint/connection.rs
```diff
@@ -1488,19 +1485,17 @@ mod tests {
...
-                .await
-                .unwrap(),
-            None
-        );
-        assert_eq!(
-            tokio::time::timeout(Duration::from_secs(1), paths_server.next())
-                .await
-                .unwrap(),
-            None
-        );
```
matheus23, line 1503: Can you try if adding a `tokio::time::pause()` ahead of this call makes it possible to reduce the timeout durations to 1 microsecond or sth like that?

iroh/src/socket/remote_map/remote_state/path_watcher.rs
```diff
@@ -0,0 +1,345 @@
...
+        cx: &mut std::task::Context<'_>,
+    ) -> Poll<Result<(), n0_watcher::Disconnected>> {
+        // When the `closed` flag is set on the watched value, we return `Disconnected`
+        // to end the watcher update stream. We can't rely on the watchable being dropped,
+        // because the watchable is cloned into the `Connection` and thus will stay alive
+        // until the last clone of a connection is dropped. However, we want the watcher
+        // to end once the connection closes. Therefore we use a manual signal here instead.
+        if self.paths_watcher.peek().closed {
+            return Poll::Ready(Err(n0_watcher::Disconnected));
+        }
```
matheus23, line 161: Oof. Okay.

I guess this is fully legal and works. FWIW, `fn is_connected` should probably be updated then, so you never have a watchable that returns `Err(Disconnected)`, but which has `is_connected` return true.

iroh/src/socket/remote_map/remote_state/path_watcher.rs
```diff
@@ -0,0 +1,345 @@
...
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+struct PathWatchValue {
+    /// The list of network transmission paths.
+    paths: SmallVec<[PathInfo; 4]>,
+    /// Set to `true` before the `RemoteStateActor` drops the `PathWatchable`.
+    ///
+    /// Afterwards, no further updates will be received.
+    closed: bool,
+}
```
matheus23, line 79: I personally found it pretty confusing why there is both a `PathWatchValue` and a `PathInfoList`, and why both refer to the same `paths: SmallVec<[PathInfo; 4]>`.
For some reason it clicked when I realized that the `paths` is just wrapped in `PathInfoList` every time we return it as a value from `PathWatcher`.

Wouldn't this become easier if `PathWatchValue` holds a `PathInfoList` instead, so you don't need to store another `current_paths: PathInfoList` in `PathWatcher`. Instead, you can re-use the one that's already in the `n0_watcher::Direct<PathWatchValue>`?

And yeah I know it's hard, but perhaps we can find a better name than `PathWatchValue`.
How about `ClosablePathList`?

iroh/examples/transfer.rs
```diff
@@ -866,7 +863,7 @@ fn parse_byte_size(s: &str) -> std::result::Result<u64, parse_size::Error> {
 }
 
 fn watch_conn_type(
-    paths_watcher: impl Watcher<Value = PathInfoList> + Send + Unpin + 'static,
+    conn: Connection,
```
flub: can this be `&Connection`?

iroh/src/endpoint/connection.rs
```diff
@@ -1010,16 +1009,30 @@ impl Connection<HandshakeCompleted> {
...
+    /// not be removed from the list of paths. This allows to reliably retrieve stats for
+    /// closed paths.
+    ///
+    /// A [`PathWatcher`] does not keep the [`Connection`] itself alive. If all references to
+    /// a connection are dropped, the [`PathWatcher`] will start to return an error when
+    /// updating. Its last value may still be used - note however that accessing
+    /// stats for a path via [`PathInfo::stats`] returns `None` if all references to a
+    /// [`Connection`] have been dropped. To reliably access path stats when a connection closes,
+    /// wait for [`Connection::closed`] and then call [`Connection::paths`] and directly
+    /// iterate over the path stats while the [`Connection`] struct is still in scope.
```
flub, line 1030: I'm confused with this. Didn't we also discuss making `Connection::closed` so that it would not keep the connection alive?

iroh/src/socket/remote_map/remote_state.rs
```diff
@@ -530,7 +531,7 @@ impl RemoteStateActor {
                 .entry(conn_id)
                 .insert_entry(ConnectionState {
                     handle: handle.clone(),
-                    pub_open_paths: pub_open_paths.clone(),
+                    path_watchable: path_watchable.clone(),
```
flub, line 534: yay for a better name!

iroh/src/socket/remote_map/remote_state.rs
```diff
@@ -497,9 +498,9 @@ impl RemoteStateActor {
     fn handle_msg_add_connection(
         &mut self,
         handle: WeakConnectionHandle,
-        tx: oneshot::Sender<PathsWatcher>,
+        tx: oneshot::Sender<PathWatchable>,
```
flub, line 501: I find it slightly odd to send away the watch**able** instead of the watch**er**, because that allows the remote to change the values. And design-wise that's not something we want to permit here I think.

iroh/src/socket/remote_map/remote_state.rs
```diff
@@ -1329,6 +1326,12 @@ struct ConnectionState {
     has_been_direct: bool,
 }
 
+impl Drop for ConnectionState {
+    fn drop(&mut self) {
+        self.path_watchable.close();
```
flub, line 1293: Is needing this explicit close a side-effect of cloning the entire watchable that is sent to the `AddConnection`'s oneshot sender? Nothing really wrong with it, but it feels a bit footgun-y

iroh/src/socket/remote_map/remote_state/path_watcher.rs
flub: ah, this matches what i commented on above. can this not be fixed by splitting the watcher and watchable?

iroh/src/socket/remote_map/remote_state/path_watcher.rs
```diff
@@ -0,0 +1,345 @@
...
+    /// Returns `true` if this path is currently the main transmission path for this [`Connection`].
+    ///
+    /// [`Connection`]: crate::endpoint::Connection
+    pub fn is_selected(&self) -> bool {
+        self.is_selected
+    }
+
+    /// Returns `true` if this path is closed.
+    ///
+    /// A path is considered closed as soon as the local endpoint has abandoned this path.
```
flub, line 229: This sentence is technically very correct. Though without knowing that a local endpoint must abandon the path as soon as it knows the remote has abandoned the path this leaves questions.

Given that we don't strictly use multipath terminology here maybe "as soon as the local or remote endpoint has abandoned this path" might be clearer?

iroh/src/socket/remote_map/remote_state/path_watcher.rs
```diff
@@ -0,0 +1,345 @@
...
+    /// A closed path will remain closed forever, so once this returns `true` it will never
+    /// return `false` afterwards. If the transmission path becomes available again in the future,
```
flub: Not sure I like introducing the term "transmission path". It probably is slightly a misnomer because it would also be used for receiving. I think we have used "network path" before.

iroh/src/socket/remote_map/remote_state/path_watcher.rs
```diff
@@ -0,0 +1,345 @@
...
+    pub fn is_ip(&self) -> bool {
+        self.remote_addr().is_ip()
+    }
+
+    /// Whether this is a relay transport path.
+    pub fn is_relay(&self) -> bool {
+        self.remote_addr().is_relay()
+    }
+
+    /// Returns stats for this transmission path.
```
flub: Let's avoid "transmission path" as a term. I think "network path" is reasonable.

iroh/src/socket/remote_map/remote_state.rs
Frando: The methods on `PathWatchable` to update the value are `pub(super)` so are not accessible from the connection, only from the `RemoteStateActor`.
The upside of doing it like this is that we can check if there's any subscribers, and not store abandoned paths in the watcher value if there are none. This is not easily possible, I think, if we add a watcher to the connection.

iroh/src/endpoint/connection.rs
Frando: Yes, we discussed that, but as of now `Connection::closed` is lifetime-bound onto `&Connection` so it keeps the connection alive. `Connection::on_closed` does not.

The final path stats are stored in the `quinn::Connection`, so if all `iroh::Connection` structs are dropped by the the time the PathWatcher processes the last event, then`PathInfo::stats` returns `None`. If a `Connection`  is still in scope, it is guaranteed to return `Some(PathStats)`.

iroh/src/socket/remote_map/remote_state.rs
Frando: Yes. See comment above for why I opted to do it like this.

iroh/src/socket/remote_map/remote_state/path_watcher.rs
Frando: Renamed all occurrences.

iroh/src/socket/remote_map/remote_state/path_watcher.rs
Frando: Renamed all occurences.

iroh/examples/transfer.rs
Frando: Changed

iroh/src/endpoint/connection.rs
flub: Do we have an issue that tracks removing all these duplicate on-closing things yet? Does making this change make it harder to de-duplicate these methods?

/cc @matheus23

iroh/src/socket/remote_map/remote_state.rs
flub: Indeed the update methods are not visible to the users, but I made this comment for our own code structure and maintainability. We make plenty of mistakes ourselves and it is easy to forget the implications when we later have a PR that does something somewhere just because it was possible.

> The upside of doing it like this is that we can check if there's any subscribers, and not store abandoned paths in the watcher value if there are none. This is not easily possible, I think, if we add a watcher to the connection.

Can you point to the function doing this optimisation right now so I can get an idea of the impact and the tradeoff being made here?

Frando:
On the remaining comments:

* The `PathWatchable`, which contains a `Watchable`, is cloned into the connection so that a `n0_watcher::Direct` is only created if the user actually creates a `PathWatcher` via `Connection::paths`. This is cheaper and uses less memory then in the case where a user does *not* create a `PathWatcher`. With the current architecture, the existence of a watcher for the watchable decides if all abandoned paths are kept or not (if a watcher exists they are kept, if not, then not). This is why I wanted a design where a watcher is only created if actually used, and not within each clone of a `Connection`. I think it would be wasteful to always collect all abandoned paths, but also it should be possible. I think the current design is a good compromise here.

* Note that even though the `PathWatchable` is cloned onto the `Connection`, the methods to set the value are not available, because they are `pub(super)`, thus only accessible from the `RemoteStateActor`. The `Connection` can only call the `pub(crate) fn watch` to create a new watcher.

* I don't think we can get rid of `PathInfo::stats` returning an `Option`, unless we make the `PathWatcher` contain a clone of the connection. We can do that, it would be a different tradeoff. If we did, the existence of a `PathWatcher` would prevent close-on-drop. Whereas in the current design, the existence of a `PathWatcher` does not guarantee that `PathInfo::stats` returns `Some`, unless a `Connection` is alive, then it is guaranteed. I tried to document this, if it isn't clear enough I can refine again.

* The last point is the reason why I wrote in the docs that if you want guaranteed path stats at the end of a connection, the recommended pattern is:
```rust
let paths = conn.paths();
// do whatever
conn.closed().await;
for path in paths.get() {
    let stats = path.stats().expect("safe because `conn` is still in scope");
}
```
I don't love it, but I think it is acceptable, and at least clearly documentable.

All that said, I would be in favor of merging this, as the current state on `main` is just broken.

iroh/src/socket/remote_map/remote_state.rs
Frando: That would be here: https://github.com/n0-computer/iroh/blob/bdfa85fd4bcad776be2d068e2a1c5f6ea532831d/iroh/src/socket/remote_map/remote_state/path_watcher.rs#L324

And a note on visibility - `PathWatchable` is never exposed to users, the `pub(super)` is to prevent ourselves from calling anything but `watch` from outside of the remote_state module. We could add an abstraction for this to `n0-watcher` potentially. If we'd keep the `n0_watcher::Direct` in the connection, we'd also needlessly store the watcher's value (which is an allocation if more than 4 paths) with every clone of the connection. The watcher *always* includes a clone of the value at the time of creation or last `get`. So even though the watcher might be entirely unused, it would store the path list with *every clone of the Connection*. And we can't `Arc` it either, because you need `mut` access to update the value, thus `Connection::paths` needs to returned an un-arced `Direct`.

This all led me to towards the design which stores a `Watchable` and not a `Direct` in the connection.

We could add a new abstraction for this in `n0-watcher`. I previously proposed a `WeakWatcher` (https://github.com/n0-computer/n0-watcher/pull/31) but since the `peek` method was added to the `Watcher` trait this is not easily possible anymore. We could add a `LazyWatcher` or such, that internally contains a `Watchable` but only exposes the `watch` method; basically moving the visibility restriction from the `remote_state::path_watcher::PathWatchable` into a new type from `n0-watcher`. Not sure if it's worth it though.

iroh/src/socket/remote_map/remote_state/path_watcher.rs
Frando: Thanks!

* Renamed `PathWatchValue` to `CloseablePathList`, lgtm!
* Reused `PathInfoList` so that there's only one place that defines the SmallVec

As to not storing `current_paths` and peeking on the value instead: No, that won't work, because `PathWatcher::current_path` has the `is_selected` field for each path set from the current value of the selected path watcher. I.e. the `PathWatcher` combines a watcher over the list of paths, and over which path is selected, and updates the selected flag on the locally-stored `PathInfoList` when either changes.
I had a revision that did not do that, but that meant that the selected path value has to be passed through to all the iterators etc, and it was more complicated. I will think a bit if I have another idea.

iroh/src/endpoint/connection.rs
matheus23: I filed https://github.com/n0-computer/iroh/issues/3959

I think the example works just as well with `on_closed`, since it doesn't consume the `conn` in scope and the connection is only dropped at the end of the scope.

matheus23 (approved):
Would still like that improvement to the test with `pause()`, but otherwise this looks good to me now 👍 

Of course there are still some open comments with flub.

matheus23:
@Frando this needs a doc reference fix still :v:

PR #3916, opened 2026-02-06
fix(iroh)!: reexport full dns module from iroh-relay

Frando:
## Description

We currently only export some types from `iroh_relay::dns` as `iroh::dns`. This means if you *actually* want to customize the DNS resolver you have to add a dependency on `iroh_relay` to your crate, because several of the types needed are currently not exported from `iroh`.

This fixes it by reexporting the whole `iroh_relay::dns` module as `iroh::dns`.

Also fixes a few doccomments.

## Breaking Changes

* `iroh::dns` now is a reexport of `iroh_relay::dns`. Everything public already was reexported from `iroh_relay::dns` so this should not actually be breaking.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


iroh/src/dns.rs
```diff
@@ -7,9 +7,7 @@
 //! See the [`endpoint_info`](crate::endpoint_info) module documentation for details on how
 //! iroh endpoint records are structured.
 
-pub use iroh_relay::dns::{
-    DnsResolver, N0_DNS_ENDPOINT_ORIGIN_PROD, N0_DNS_ENDPOINT_ORIGIN_STAGING,
-};
+pub use iroh_relay::dns::*;
```
dignifiedquire: I am sceptical of glob exports, how much would it be to make these explicit?

Frando: We can, it's just that we had this in the past and then forgot to update it when we added a new type in iroh_relay::dns that was used in a method of Resolver. So to prevent that I'd prefer  a glob export. Or even just reexport the module?

Frando: We could just reexport the module and move the tests somewhere else.

dignifiedquire: maybe cc @ramfox who thought about exports the most

Frando: I pushed a commit that fully reexports the `dns` module. I think this is much cleaner.

Frando:
@ramfox wdyt, should we merge this? Currently I know two downstream projects that need to depend on iroh_relay for the missing reexports.

dignifiedquire (approved):

PR #3918, opened 2026-02-06
chore: specify minimum iroh-quinn and iroh-quinn-proto patch versions

ramfox:
## Description

prepping for 0.96.1 patch release


ramfox:
sem-ver check is now only failing for 0.96.0, not for the current code in this PR, going to merge

PR #3921, opened 2026-02-09
refactor(iroh, iroh-relay)!: Relay should not kill old connections when same endpoint id connects

### LINKED ISSUE #3813: Detect shared secret key usage
Sometimes users create multiple endpoints with the same secret key due to misunderstandings. To help them catch mistakes we might be able to help recognise this situation:

The connection to the relay server will exhibit a particular disconnection/reconnection pattern. It might be possible to recognise this pattern and log a warning or info message with a hint of what might be going on.

Additionally if the same secret key is used in another endpoint in the same process it could be possible to error when creating the 2nd endpoint.

Frando:
## Description

When the iroh relay receives a connection from an endpoint to which it already has a connection, it currently drops the old one immediately.  The old endpoint did not receive any indication as to why its connection was dropped, so it would reconnect again right away. The connection would succeed: Now it was its turn to be accepted, and the other endpoint would be dropped from the relay. This back-and-forth loop would continue infinitely.

This PR changes this in the following way:

* When a relay server receives a connection from an endpoint it for which it already has a connection, the new connection becomes the target for all future messages.
* The old connection however is *not* dropped. Instead, it is continued until the client stops the connection. The client can still *send* messages, but it won't receive any further messages.
* When a connection is moved into this "inactive" state, the server sends a final `Health` message to the client, with a text string informing the client that it is now inactive because another endpoint connected. In iroh, we `warn!` log this message.
* When an active client disconnects, and there is an inactive client, the lastest inactive client will be promoted to active again

## Breaking Changes

* `iroh_relay::server::client::RunError` variants changed

## Notes & open questions

* I'd prefer to add a typed frame for the `Health` issue. However, adding new frames to the relay protocol currently will break all older clients. Therefore, this uses the stringly health frame for now.
* We might want to discuss further changes, i.e. if the "old" endpoint should maybe do something more heavy than just logging a warn message. It will be in a weird state onwards: It still can send messages to other endpoints over the relay, but will never receive responses. This means that all connection attempts that go over the relay will time out.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

Fixes #3813


iroh/src/endpoint.rs
```diff
@@ -3058,4 +3060,113 @@ mod tests {
...
+            .bind()
+            .instrument(error_span!("ep2"))
+            .await?;
+        info!("ep2 bound {:?}", ep2.id());
+        ep2.online().await;
+        println!("ep2 online");
+
+        // Wait a bit. `online` does not mean that the connection to the home relay was *established*,
+        // only that the home relay was *chosen* based on the net report probes.
+        tokio::time::sleep(Duration::from_millis(500)).await;
```
dignifiedquire: this screams flakiness, can we detect this differently?

iroh/src/endpoint.rs
```diff
@@ -3058,4 +3060,113 @@ mod tests {
...
+        )?;
+        println!("client connected to ep2");
+
+        // assert that ep1 did not receive a connection
+        assert!(now_or_never(ep1.accept()).is_none());
+
+        // We assert that we get the warn log once for endpoint 1, and not at all for endpoint 2.
+        logs_assert(|logs| {
+            let expected_line = |line: &str| {
+                line.contains("WARN") && line.contains("Another endpoint connected with the same endpoint id. No more messages will be received")
```
dignifiedquire, line 3336: can't we check this differently than log assertions? :/

Frando: Currently, no. There's nothing on the endpoint to check relay status. Maybe there should be.

iroh/src/endpoint.rs
Frando: Currently, no. Maybe `online` should only resolve once we're actually connected to the home relay?

dignifiedquire: yeah, potentially that would make sense

Frando: I pushed a change to the test that uses a log assertion instead of a timeout, so this can't flake anymore.

I'd like to leave adding a proper API for this to a followup though as it really is unrelated to the change in this PR.

ramfox (approved):
lgtm! especially with the proposed follow ups.

PR #3926, opened 2026-02-11
fix(iroh): Disable QAD if no IP transports are configured

Frando:
## Description

When no IP transports are configured, QAD will always fail, because it uses an endpoint without ip transports and thus can't send anything. Therefore, this PR changes this such that when IP transports are disabled, no QAD config is created for the relay client, thus QAD probes will be skipped.

### DIFF
--- a/iroh/src/socket.rs
+++ b/iroh/src/socket.rs
@@ -753,14 +753,15 @@ impl Handle {
         let (actor_sender, actor_receiver) = mpsc::channel(256);
 
         #[cfg(not(wasm_browser))]
-        let ipv6 = transports
+        let has_ipv6_transport = transports
             .ip_bind_addrs()
             .into_iter()
             .any(|addr| addr.is_ipv6());
 
-        let direct_addrs = DiscoveredDirectAddrs::default();
-
+        #[cfg(not(wasm_browser))]
         let has_ip_transports = !transports.ip_bind_addrs().is_empty();
+
+        let direct_addrs = DiscoveredDirectAddrs::default();
 
         let remote_map = {
             RemoteMap::new(
@@ -841,7 +842,7 @@ impl Handle {
                 ep: endpoint.clone(),
                 client_config,
                 ipv4: true,
-                ipv6,
+                ipv6: has_ipv6_transport,
             });
             net_report_config.quic_config(qad_config)
         };


iroh/src/endpoint.rs
```diff
@@ -3058,4 +3058,22 @@ mod tests {
...
+    #[tokio::test]
+    #[traced_test]
+    async fn connect_relay_online_without_ip() -> Result {
+        let (relay_map, _relay_url, _relay_server_guard) = run_relay_server().await?;
+        let client = Endpoint::empty_builder(RelayMode::Custom(relay_map.clone()))
+            .clear_ip_transports()
+            .insecure_skip_relay_cert_verify(true)
+            .bind()
+            .await?;
+        tokio::time::timeout(Duration::from_millis(500), client.online())
```
flub: This will be flaky on CI sadly.

iroh/src/socket.rs
```diff
@@ -1238,20 +1264,20 @@ impl Actor {
                     #[cfg(wasm_browser)]
                     let _unused_in_browsers = change;
                 },
-                state = self.netmon_watcher.updated() => {
+                state = &mut netmon_watcher_updated, if netmon_watcher_updated.is_some() => {
```
flub: Are you sure this is fine? Because `self.handlle_network_change` does go and tell the transports that they should go an rebind on a major network change. Which means the relay transport would close and open the connection, ensuring the relay tunnel has the best chance of being connected when an endpoint moves between networks. So I'm not sure you can just disable netmon like this.

Frando: Agreed. Let's leave this for a separate PR. I removed all netmon changes and strictly made this PR about disabling QAD if there are no IP transports.

iroh/src/endpoint.rs
Frando: True. Removed the test. There's not really a way to test this apart from timing because we don't expose further details AFAICS.

iroh/src/socket.rs
```diff
@@ -828,13 +828,17 @@ impl Handle {
...
+
         #[cfg(not(wasm_browser))]
-        let net_report_config = net_report_config.quic_config(Some(QuicConfig {
-            ep: qad_endpoint,
-            client_config,
-            ipv4: true,
-            ipv6,
-        }));
+        let net_report_config = {
+            let qad_config = has_ip_transports.then(|| QuicConfig {
```
dignifiedquire, line 841: please add a comment as to why this is happening

Frando: :+1: done

PR #3928, opened 2026-02-11
fix: improve holepunching after network changes

flub:
## Description

When the network changes we should recover our connection and
holepunch again. Because quinn now tracks 4-tuples a lot of network
paths would turn into black holes because the source address would no
longer match.

This hooks up network changes to quinn so that it will close and open
new paths correctly when this happens. This should result in most
paths not being validated after a network change which in turn means
any new ADD_ADDRESS and REACH_OUT frames will be sent on the validated
path instead of a path that is no longer functional.

## Breaking Changes

n/a

## Notes & open questions

The first few commits do some renaming and documenting of stuff that's not
strictly required. It was nicer and easier to build on top of that. If
you, the reviewer, would prefer I can split those off into a refactoring
PR and base this one on top of that.

## Change checklist

- [x] I still need to check if the server updates quinn so it can do
      the correct ADD_ADDRESS and REMOVE_ADDRESS stuff.
- [x] I still need to check the sequencing of the above so things
      happen in the correct order.
- [x] Investigate `endpoint_relay_connect_loop` test failure
- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/test_utils/test_transport.rs
+++ b/iroh/src/test_utils/test_transport.rs
@@ -540,7 +540,7 @@ mod tests {
         let t2 = network.create_transport(s2.public())?;
 
         // Strong RTT disadvantage for custom transport
-        let custom_bias = TransportBias::primary().with_rtt_disadvantage(Duration::from_secs(1));
+        let custom_bias = TransportBias::primary().with_rtt_disadvantage(Duration::from_secs(10));
         let config = EndpointConfig::default()
             .with_ip()
             .with_custom_bias(custom_bias);


iroh/src/socket/remote_map/remote_state.rs
```diff
@@ -957,7 +957,7 @@ impl RemoteStateActor {
                     self.paths.abandoned_path(&addr);
                 }
             }
-            PathEvent::Closed { id, .. } | PathEvent::LocallyClosed { id, .. } => {
+            PathEvent::LocallyClosed { id, .. } => {
```
Frando: If I'm reading things correctly, this event is *only* emitted when we opened a path, and the `PathOpen` timer fires without the path being established by then. So I think the below code should also run on `PathEvent::Abandoned` so that if one connection abandones a path, this is forwarded to all other connections?

flub: I'm postponing thinking about this for https://github.com/n0-computer/iroh/issues/3930

iroh/src/socket/remote_map/remote_state.rs
```diff
@@ -1126,6 +1102,43 @@ impl RemoteStateActor {
...
         }
     }
+
+    /// Handles regularly checking if any paths need hole punching currently
+    ///
+    /// Currently we need to have 1 IP path, with a good enough latency.
+    fn check_connections(&mut self) {
+        let mut is_goodenough = true;
+        for conn_state in self.connections.values() {
+            let mut is_conn_goodenough = false;
```
matheus23: I know you just moved the code around, but... Why do we have this variable? And why is this initialized to `false`?
This looks like you'll set `is_goodenough` to `false`, if you can't upgrade the connection handle? That seems weird to me.

iroh/src/socket.rs
```diff
@@ -1444,6 +1475,89 @@ impl Actor {
...
+            fn is_path_recoverable(
+                &self,
+                _path_id: noq::PathId,
+                network_path: noq_proto::FourTuple,
+            ) -> bool {
+                match MultipathMappedAddr::from(network_path.remote()) {
+                    MultipathMappedAddr::Mixed(_) => {
+                        // This address is only ever used to send an Initial packet to, it
+                        // should never appear as an established path.
+                        unreachable!();
```
flub: I'm kind of surprised no one complained about this...

flub: upon thinking about this, i think it should log an error and return it as not recoverable.

PR #3946, opened 2026-02-16
feat(netsim): filter tests

Arqu:
## Description

Improves on the current filtering logistics of netsim. Instead of just a skiplist, you now have a filter field. Depending on prefix you either run it:
- empty - no filtering
- prefix the list with `skip:` to skip specific tests in a run
- prefix the list with `only:` to only run specific tests
- defaults to `only:` if you just provide a list without a prefix

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -351,7 +351,7 @@ jobs:
       netsim_branch: "main"
       sim_paths: ${{ github.event_name == 'pull_request' && 'sims/integration' || 'sims/iroh/iroh.json,sims/integration' }}
       pr_number: ${{ github.event.pull_request.number || '' }}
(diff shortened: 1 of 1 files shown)


flub (approved):
does the skip and only filtering do submatches or only full matches?  that is is `only:nat` eqvallent to `only:*nat*` or would it not run anything.

I think it could make sense to do submatches, but also probably not too much trial-and-error to get the names correct otherwise.

Arqu:
I'll get that going on the netsim side

Arqu:
Done, matches on substring now
https://github.com/n0-computer/chuck/commit/975cafddab58370b0cbed4e53cd97eb341746385

Issue #3948, opened 2026-02-16
Make Incoming::remote_address either a TransportAddr or an IncomingAddr


rklaehn:
Incoming::remote_address for a connection initiated via a relay will currently show the fake IPV6 address.

One of the main purposes of remote_address is to implement rate limiting, and for that it would be very useful to know if the connection was initiated directly or via a relay. Also, the user should not see our fake IPV6 addrs if we can avoid it.

So make Incoming::remote_address either a TransportAddr or a custom enum IncomingAddr. The reason for the latter would be that we do know the EndpointId if we translate the IPV6 address, and this would be information that would be useful for a filter.

A separate but related issue: there isn't a remote_address in Accepting. But Accepting is the first state that a per-ALPN protocol handler gets to see. So if we wanted to allow protocol handlers to perform filtering or rate limiting by SocketAddr, we would have to add this. Also, it would be more symmetrical to the iroh-quinn API.

Q: should we add a new fn and deprecate remote_address, or change the signature of remote_address?
Q: TransportAddr or custom enum that contains EndpointId?

Frando:
Q1: I think we should do a breaking change to remote_addr
Q2: It's a bit unforunate API wise but I do think surfacing the endpoint id for relay connections is quite imporant for this usecase, so new `enum IncomingAddr { Ip(SocketAddr), Relay(EndpointId, RelayUrl) }` or similar it is i think.

rklaehn:
Agree with both. Will implement this.

rklaehn:
We now got a type for this: https://docs.rs/iroh/latest/iroh/endpoint/enum.IncomingAddr.html

PR #3949, opened 2026-02-16
feat(iroh): more precise information about incoming connections

rklaehn:
## Description

Provide more precise information about incoming connections. Instead of exposing the synthetic IPV6 addr for relay connections, we translate it to a special enum IncomingAddr that has all available info for a rate limiter or access control component to make a decision.

Implements https://github.com/n0-computer/iroh/issues/3948

## Breaking Changes

endpoint::Incoming::remote_address renamed to remote_addr, and returns IncomingAddr.
endpoint::Incoming::remote_address_validated renamed to remote_addr_validated.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/endpoint/connection.rs
+++ b/iroh/src/endpoint/connection.rs
@@ -178,7 +178,7 @@ impl Incoming {
     }
 
     /// Returns the remote address of this incoming connection.
-    pub fn remote_address(&self) -> IncomingAddr {
+    pub fn remote_addr(&self) -> IncomingAddr {
         self.ep
             .sock
             .to_transport_addr(self.inner.remote_address())
@@ -188,8 +188,8 @@ impl Incoming {
     /// Whether the socket address that is initiating this connection has been validated.
     ///
     /// This means that the sender of the initial packet has proved that they can receive
-    /// traffic sent to `self.remote_address()`.
-    pub fn remote_address_validated(&self) -> bool {
+    /// traffic sent to `self.remote_addr()`.
+    pub fn remote_addr_validated(&self) -> bool {
         self.inner.remote_address_validated()
     }
 }
--- a/iroh/src/endpoint/connection.rs
+++ b/iroh/src/endpoint/connection.rs
@@ -189,7 +189,15 @@ impl Incoming {
     /// This means that the sender of the initial packet has proved that they can receive
     /// traffic sent to `self.remote_address()`.
     pub fn remote_address_validated(&self) -> bool {
-        self.inner.remote_address_validated()
+        let addr = self.inner.remote_address();
+        match addr.ip() {
+            IpAddr::V6(ip)
+                if crate::socket::mapped_addrs::RelayMappedAddr::try_from(ip).is_ok() =>
+            {
+                true
+            }
+            _ => self.inner.remote_address_validated(),
+        }
     }
 }
 
--- a/iroh/src/endpoint/connection.rs
+++ b/iroh/src/endpoint/connection.rs
@@ -68,6 +68,15 @@ pub enum IncomingAddr {
     },
 }
 
+impl From<IncomingAddr> for iroh_base::TransportAddr {
+    fn from(addr: IncomingAddr) -> Self {
+        match addr {
+            IncomingAddr::Ip(addr) => Self::Ip(addr),
+            IncomingAddr::Relay { url, .. } => Self::Relay(url),
+        }
+    }
+}
+
 impl From<crate::socket::transports::Addr> for IncomingAddr {
     fn from(addr: crate::socket::transports::Addr) -> Self {
         match addr {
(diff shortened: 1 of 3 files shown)


iroh/src/endpoint/connection.rs
```diff
@@ -149,17 +177,28 @@ impl Incoming {
...
     /// This means that the sender of the initial packet has proved that they can receive
     /// traffic sent to `self.remote_address()`.
     pub fn remote_address_validated(&self) -> bool {
-        self.inner.remote_address_validated()
+        let addr = self.inner.remote_address();
+        match addr.ip() {
+            IpAddr::V6(ip)
+                if crate::socket::mapped_addrs::RelayMappedAddr::try_from(ip).is_ok() =>
+            {
+                true
```
flub: Is this true? While a well-behaving relay won't send you anything that doesn't really come from the src EndpointId I don't think this is guaranteed by anything in the protocols. So I think you still need to follow the inner logic.

iroh/src/endpoint/connection.rs
```diff
@@ -51,6 +51,34 @@ use crate::{
...
+    /// A direct connection from an IP address.
+    Ip(SocketAddr),
+    /// A connection via a relay.
+    Relay {
+        /// The URL of the relay.
+        url: RelayUrl,
+        /// The endpoint ID of the remote peer.
+        endpoint_id: EndpointId,
+    },
+}
```
flub, line 69: I still find it odd that this doesn't cater for anything about future transports, not even a non_exhaustive. But you're the custom transport expert so I'm fine with trusting your judgement here.

iroh/src/socket.rs
```diff
@@ -374,6 +374,18 @@ impl Socket {
...
+    /// Translates a raw [`SocketAddr`] (which may be a synthetic mapped address) into
+    /// a [`transports::Addr`].
+    ///
+    /// For regular IP addresses this returns `Addr::Ip`. For synthetic relay-mapped
+    /// IPv6 addresses this performs a reverse lookup and returns `Addr::Relay`.
+    pub(crate) fn to_transport_addr(&self, addr: SocketAddr) -> transports::Addr {
+        self.mapped_addrs
+            .relay_addrs
+            .to_transport_addr(addr)
+            .unwrap_or(transports::Addr::Ip(addr))
```
flub, line 465: This unwrap means you can still return the synthetic IPv6 addresses. The upstream API does not have this problem. I'm not sure what to do about that though. It really is an invariant of our system that should not be violated. I think I could live with this, given that an error is already logged. But maybe this API needs a big warning to warn of this tradeoff to make it infallible.

flub (changes requested):

iroh/src/endpoint/connection.rs
rklaehn: Ah, non-exhaustive is a must I guess...

iroh/src/endpoint/connection.rs
matheus23: Yeah I agree with flub here. Any malicious relay can fake this.

iroh/src/socket.rs
rklaehn: The way I understood this is that *if* addr is a relay mapped addr, the lookup is guaranteed to succeed.

But it's true that this will turn an invariant violation into just weird behaviour. Maybe it should panic instead?

flub: Yes, your understanding is correct. I just don't know how to best handle that, with a panic or with weird behaviour. I'm fin with this behaviour I think? Just would like it at least internally documented.

rklaehn: Added a big comment.

flub (approved):

matheus23:
Do we have an idea why netsim is not happy here?

Frando:
> Do we have an idea why netsim is not happy here?

unrelated runner issue, should be fixed now

matheus23:
After the main merge, this needs a bit of thought: Do we want to `pub(crate)` expose `Endpoint::inner`? We already have a `#[cfg(test)] fn inner() -> Result<EndpointInner, _>` though, so ...
Perhaps we can add a `to_transport_address` function to `Endpoint`?
FWIW, this needs fixing.

rklaehn:
> After the main merge, this needs a bit of thought: Do we want to `pub(crate)` expose `Endpoint::inner`? We already have a `#[cfg(test)] fn inner() -> Result<EndpointInner, _>` though, so ... Perhaps we can add a `to_transport_address` function to `Endpoint`? FWIW, this needs fixing.

I did try both. I ended up doing this because it is slightly smaller diff, but if you can prefer I can also do a forwarder.

rklaehn:
> After the main merge, this needs a bit of thought: Do we want to `pub(crate)` expose `Endpoint::inner`? We already have a `#[cfg(test)] fn inner() -> Result<EndpointInner, _>` though, so ... Perhaps we can add a `to_transport_address` function to `Endpoint`? FWIW, this needs fixing.

Is 8d254afeb7c8043ee7bc49c47c0eb9859d3f59fc what you had in mind?

PR #3951, opened 2026-02-17
feat: Rate limiting in router

rklaehn:
## Description

Add an optional mechanism for rate limiting and requesting requests in the iroh router.

The reason to have something like this is if we ever have a single iroh endpoint exposed to the world that can be overloaded or hit by a DOS attack. E.g. a public irpc service, n0des, ...

Unfortunately this has quite some API surface: a trait for the callbacks, and enums for the fn returns.

~~I also added an example that hammers a server with lots of clients and measures the CPU load. Here is the output of the latest version:~~

Simplified the sample. The hammering tool now lives in a separate repo, since it got a bit too big for an example.

```
Connection filter benchmark, 100 attempts per filter

Direct connections:

             no filter:   0 accepted,   0 rejected, 100 closed  (client: 40.63ms | server: 97.30ms cpu,     1028 ops/s, 41.00ms wall)
           ignore addr:   0 accepted, 100 rejected,   0 closed  (client:   5.01s | server:  3.46ms cpu,    28868 ops/s, 10.53ms wall)
           reject addr:   0 accepted, 100 rejected,   0 closed  (client:  8.04ms | server:  5.61ms cpu,    17835 ops/s, 10.08ms wall)
        retry + reject:   0 accepted, 100 rejected,   0 closed  (client: 37.18ms | server: 12.77ms cpu,     7833 ops/s, 39.96ms wall)
           reject alpn:   0 accepted,   0 rejected, 100 closed  (client: 31.34ms | server: 75.25ms cpu,     1329 ops/s, 33.84ms wall)

Relay connections (https://127.0.0.1:53395/):

             no filter:   0 accepted,   0 rejected, 100 closed  (client:   4.10s | server: 174.75ms cpu,      572 ops/s,   4.09s wall)
    reject endpoint id:   0 accepted, 100 rejected,   0 closed  (client:   4.04s | server: 21.50ms cpu,     4652 ops/s,   4.04s wall)
           reject alpn:   0 accepted,   0 rejected, 100 closed  (client:   4.08s | server: 88.07ms cpu,     1135 ops/s,   4.08s wall)
```

ignore is fastest, reject just a tad slower. reject alpn doesn't buy much over just closing the connection. It is just a convenient place to throttle or reject by alpn.

For relay connections, the cheapest is to reject by endpoint id. This is quite a bit faster than closing the connection since we get the endpoint id early. Rejecting by alpn in this case helps a bit more.

The example starts a server in a subprocess and then hammers it with n connections. The subprocess is measuring its own CPU time.

## Breaking Changes

None

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/protocol.rs
+++ b/iroh/src/protocol.rs
@@ -468,10 +468,6 @@ impl RouterBuilder {
     /// `decrypt()` is relatively expensive, so filters should reject based on
     /// cheaper signals (e.g. remote address) first.
     ///
-    /// For a higher-level filter API that splits the decision into named methods
-    /// (e.g. by socket address or endpoint id, or by proposed ALPNs from the
-    /// ClientHello), see the `incoming-filter` example.
-    ///
     /// [`Incoming`]: crate::endpoint::Incoming
     /// [`Incoming::decrypt`]: crate::endpoint::Incoming::decrypt
     pub fn incoming_filter(mut self, filter: IncomingFilter) -> Self {


iroh/src/protocol.rs
```diff
@@ -541,6 +698,23 @@ async fn handle_connection(incoming: crate::endpoint::Incoming, protocols: Arc<P
                 "remote",
                 tracing::field::display(connection.remote_id().fmt_short()),
             );
+
+            // ALPN filter: runs after the connection is established for both
+            // relay and direct connections.
```
Frando: Maybe the EndpointId filter should be run here for IncomingAddr::Ip now that the endpoint id is known for ip connections as well? If not, it should be clearly documented that the endpoint id filter is *not* run for ip connections. I think it would be simpler if we'd run it here, so that you can impl just that filter if all you want is to filter by endpoint id.

rklaehn: Maybe just naming? Call the other relay filter?

This is the one that is going to be called for every request, the one that takes just endpoint id is the fastest way to intercept a relay request.

You can of course use this one to filter or rate limit by endpoint id.

Frando: Also, it just now occurs to me that maybe all this filtering could be done through EndpointHooks with some new hooks instead of adding it as a new concept to the Router?

ansarizafar: Yes we need more hooks. I am using Irpc-iroh and I need a hook to validate JWT token on each request.

rklaehn:
https://github.com/n0-computer/quinn would add another place to filter incoming packets by ALPN

iroh/src/protocol.rs
rklaehn: > Also, it just now occurs to me that maybe all this filtering could be done through EndpointHooks with some new hooks instead of adding it as a new concept to the Router?

I will take a look, but I am not sure about this. We already have all the stages of the state machine exposed with the Incoming -> Accepting -> Connection, with various ways to retry and reject. A state machine is vastly superior to a hook if you want maximum flexibility. And if you want something predefined and more opinionated, that's the router.

iroh/src/protocol.rs
```diff
@@ -725,6 +899,72 @@ mod tests {
...
+        assert!(conn_err.is_err(), "expected connection to be refused");
+
+        r1.shutdown().await.anyerr()?;
+        e2.close().await;
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_connection_filter_reject_endpoint() -> Result {
```
Arqu: Can we extend this test or write another one which tests the relay path?

iroh/src/protocol.rs
rklaehn: > Yes we need more hooks. I am using Irpc-iroh and I need a hook to validate JWT token on each request.

Are you using irpc-iroh with the router or bare? I think it should be easy (or even the default) to use irpc-iroh with the router, then it would not matter for you if the hooks are at endpoint level or router level.

My current thinking is that the Endpoint API should have maximum flexibility, which means state machine instead of hooks. The router should provide hooks.

The reason is that hooks are limited. E.g. you want to handle a response to a request on a different task, or suspend it, or whatever, that is very easy with a state machine you drive yourself, but impossible without a rube goldberg machine if you have a hook callback. So callbacks in the router, FSM in Endpoint API.

ansarizafar: I am using irpc-iroh with router.  I want to verify/decode a JWT token on each method request (not connection request) and pass that data(user ID, endpoint Id, user role etc) to RPC method for request authorization. Is this somehow possible?

rklaehn: So the JWT token comes with each request, or on a separate APLN?

The latter is already possible. You could have an auth ALPN where you just send the JWT token, then have a map from endpoint id to data that you share between the auth ALPN and the service ALPN.

But I am not sure what exactly you want to do.

ansarizafar: I want to authorize each RPC method request and pass decoded user info(id, role) to rpc method, So the JWT token comes with each request. Is there any better method?

rklaehn: I think it is preferable to do this separately. Otherwise you will have to extend every single protocol with auth.

Basically, you have a server side state that keeps track of endpoints and their data. Then you do a request using ALPN "auth" or whatever that includes the JWT token. At this point you update the shared state and store the JWT data for this endpoint (indefinitely or for some time).

Then in the actual RPC handler you don't need any extra bytes per request. You just check if the request is allowed as soon as you have the endpoint id, otherwise fail.

This is safe because the endpoint id can not be forged unless you have exposed the private key or Ed25519 is broken.

If you store the credentials only for a limited time, you have to have logic on the client side to re-send a token either proactively or when a RPC request fails.


But this is a matter of taste really. I would find having a JWT on each possibly tiny RPC request not just cumbersome but also inefficient.

ansarizafar: I think separate ALPN solution will not work for my use case. My client app is a  full stack web app accessible on local LAN.  Users will login with their own user name/password by using the same client endpoint. Irpc-Iroh server has a login method for user auth. Login method sends a JWT token which will be included in every request. so I need a per request hook/middleware.

iroh/examples/connection-filter.rs
```diff
@@ -0,0 +1,319 @@
+//! Example demonstrating the [`ConnectionFilter`] hook with rate limiting.
```
dignifiedquire: I think this should live in iroh-examples ,it is very large

rklaehn: I stripped it down a lot.

iroh/src/endpoint/connection.rs
```diff
@@ -81,14 +82,12 @@ impl From<IncomingAddr> for iroh_base::TransportAddr {
     }
 }
 
-impl From<crate::socket::transports::Addr> for IncomingAddr {
-    fn from(addr: crate::socket::transports::Addr) -> Self {
+impl From<transports::Addr> for IncomingAddr {
```
rklaehn, line 85: Unrelated change, just some formatting.

iroh/examples/incoming-filter.rs
```diff
@@ -0,0 +1,70 @@
+//! Example demonstrating the [`IncomingFilter`] hook.
+//!
+//! This example requires all direct (UDP) connections to pass QUIC address
+//! validation via a retry token before being accepted. Relay connections are
```
Frando: Maybe we add a short explanation what "QUIC address validation" means here? e.g.
"This ensures that the server verifies that a client truly owns its claimed IP address before allocating resources, preventing denial-of-service attacks via spoofed source IPs."

iroh/src/protocol.rs
```diff
@@ -160,6 +162,73 @@ impl From<quic::ClosedStream> for AcceptError {
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum IncomingFilterOutcome {
+    /// Accept the connection.
+    Accept,
+    /// Tell the remote to retry with a token (a QUIC `RETRY` packet).
+    ///
+    /// What this does depends on the connection type:
+    ///
+    /// - **Direct (UDP) connections** — this is QUIC source address
```
Frando: nit which can just be ignored as well: To me these em dashes in source code makes me read this as AI content, because I don't think anyone types em dashes in source code. Maybe that's just me though and people actually do type them. I would prefer ASCII-only in source code, and replace them with a colon.

iroh/src/protocol.rs
```diff
@@ -160,6 +162,73 @@ impl From<quic::ClosedStream> for AcceptError {
...
+    ///   attempt with the token, and the next [`Incoming`] for that flow has
+    ///   [`Incoming::remote_addr_validated`] set to `true`. The token is
+    ///   bound to the source address.
+    ///
+    /// - **Relay connections** — there is no source address to validate
+    ///   (the relay already vouches for the packet origin), so the
+    ///   "validation" itself has no security meaning. However, the retry
+    ///   still imposes a real cost on the client: an extra round trip
+    ///   through the relay plus the work of sending a fresh ClientHello
+    ///   with the token. This filters out spray-and-pray clients that
```
Frando: what's a "spray and pray client"?

iroh/src/protocol.rs
```diff
@@ -398,9 +467,16 @@ impl RouterBuilder {
         Self {
             endpoint,
             protocols: ProtocolMap::default(),
+            incoming_filter: None,
         }
     }
 
+    /// Sets a [`IncomingFilter`] for the router.
```
Frando: nit: I think I'd move the longer doc block from `IncomingFilter` here and link to the fn from `IncomingFilter`, so that the docs are directly visible when browsing the RouterBuilder docs.

nit 2: Sets *an* `IncomingFilter`

iroh/src/protocol.rs
```diff
@@ -485,6 +562,30 @@ impl RouterBuilder {
                             break; // Endpoint is closed.
                         };
 
+                        if let Some(filter) = &incoming_filter {
+                            match filter(&incoming) {
+                                IncomingFilterOutcome::Accept => {}
+                                IncomingFilterOutcome::Retry => {
+                                    debug_assert!(
```
Frando: Do we really want this assert? If so, it should be documented in the InocmingFilterOutcome docs ("Returning `Retry` when `remote_addr_validated()` is `true` will cause a panic").
Or maybe just `warn` log it?

iroh/src/protocol.rs
```diff
@@ -730,6 +834,248 @@ mod tests {
...
+            protocol::{
+                IncomingFilterOutcome, Router,
+                tests::{ECHO_ALPN, Echo},
+            },
+        };
+
+        /// Two direct endpoints with a filtered router on the first.
+        ///
+        /// Binds to IPv4 loopback only so retry-token validation works on
+        /// multi-homed CI hosts (tokens are tied to the source address).
```
Frando, line 896: Does that mean retrying is broken when there's multipe interfaces?
If so that should be prominently documented, because this can not only happen in CI but also for regular endpoints, right?
If I have an `EndpointAddr` with two IPs that are both reachable, and the remote asks for address validation retry, and it doesn't work, this is bad. It basically would mean that you cannot use address validation reliably if your server is reachable over two IP addresses?
Maybe we should add a patchbay test to verify current behavior, and/or document the limitation clearly. I also wonder, though, if we shouldn't fix this?

iroh/src/protocol.rs
```diff
@@ -730,6 +834,248 @@ mod tests {
...
+                .bind()
+                .await?;
+            Ok((r1, e2, addr, guard))
+        }
+
+        #[tokio::test]
+        #[traced_test]
+        async fn addr_retry() -> Result {
+            let (r1, e2, addr) = direct_pair(|_| IncomingFilterOutcome::Retry).await?;
+            // Server sends retry (unvalidated), then refuses (validated).
```
Frando: "then refuses(validated)" where does this happen? Also, shouldn't this then cause the panic I commented on above (the debug_assert - the second try will be validated, but the filter still returns Retry)?

iroh/src/protocol.rs
```diff
@@ -730,6 +834,248 @@ mod tests {
...
+            Ok(())
+        }
+
+        #[tokio::test]
+        #[traced_test]
+        async fn addr_ignore() -> Result {
+            let (r1, e2, addr) = direct_pair(|_| IncomingFilterOutcome::Ignore).await?;
+            // No response at all — connect times out.
+            let result =
+                tokio::time::timeout(Duration::from_millis(500), e2.connect(addr, ECHO_ALPN)).await;
```
Frando, line 984: Instead of the tokio timeout might be cleaner to set idle_timeout in the QuicTransportConfig to 500ms and wait for that to expire

iroh/Cargo.toml
```diff
@@ -133,6 +133,7 @@ indicatif = { version = "0.18", features = ["tokio"] }
 parse-size = { version = "1.1.0", features = ['std'] }
 iroh-base = { version = "0.97.0", default-features = false, features = ["key", "relay"], path = "../iroh-base" }
 console = { version = "0.16" }
+governor = "0.9"
```
Frando: unused dep, remove

iroh/src/protocol.rs
rklaehn: Yeah, retrying under some circumstances is broken. I had this happen in CI and had to make sure I only subscribe to a single interface:

https://discord.com/channels/949724860232392765/950683937661935667/1493210239116251157

If we would fix it it should not hold up this PR, since it only exposes existing functionality in the iroh Endpoint at the ProtocolHandler level.

iroh/src/protocol.rs
rklaehn: Yeah, will warn log.

iroh/src/protocol.rs
rklaehn: Not sure. Everybody understands a timeout, QuicTransportConfig is an advanced API. And I would no longer be able to use the helper fn to set up a pair.

iroh/src/protocol.rs
rklaehn: Rewrote the test to do retry then accept.

rklaehn:
> I'm wondering a bit still if this should be an endpoint hook. Or, how we intend to evolve both APIs (endpoint hooks and router hooks) - we have a similar "can be done on both levels" with access control.
> 
It is already more hook like than before, but I think it is fine to have this only on the Router. If you use the endpoint state machine manually you can already do all of this without a hook.

> But this does not have to hold up the PR. We still have some time to settle until 1.0, and I also don't have a strong opinion in either direction.

iroh/src/protocol.rs
```diff
@@ -485,6 +564,31 @@ impl RouterBuilder {
...
+                            match filter(&incoming) {
+                                IncomingFilterOutcome::Accept => {}
+                                IncomingFilterOutcome::Retry => {
+                                    if !incoming.remote_addr_validated() {
+                                        warn!(
+                                            "filter returned Retry for an already validated connection",
+                                        );
+                                    }
+                                    if let Err(err) = incoming.retry() {
+                                        err.into_incoming().refuse();
```
Arqu, line 573: Should we log this err too?

iroh/src/protocol.rs
```diff
@@ -485,6 +564,31 @@ impl RouterBuilder {
                             break; // Endpoint is closed.
                         };
 
+                        if let Some(filter) = &incoming_filter {
+                            match filter(&incoming) {
```
Arqu, line 564: What if this errors, will it break the entire loop?

Edit: thinking more about it, given we're working against adversarial setups, somebody trying to fumble this can just fuzz against the filter until it finds a panic and just kill the entire accept loop, maybe turning the whole thing into not just a flood attack but outright silent nuke.

iroh/src/protocol.rs
Arqu: Actually we should definitely log this error, otherwise real users not following the protocol to the dot won't know why they just keep getting refused.

iroh/src/protocol.rs
```diff
@@ -485,6 +564,31 @@ impl RouterBuilder {
...
+                                        err.into_incoming().refuse();
+                                    }
+                                    continue;
+                                }
+                                IncomingFilterOutcome::Reject => {
+                                    incoming.refuse();
+                                    continue;
+                                }
+                                IncomingFilterOutcome::Ignore => {
+                                    incoming.ignore();
```
Arqu, line 582: Do we have any lingering state from these ignored ones on the noq side? Otherwise this might lead to piling up maliciously or not.

iroh/src/protocol.rs
```diff
@@ -398,9 +453,32 @@ impl RouterBuilder {
...
+    ///
+    /// The filter is called with the raw [`Incoming`] for each connection attempt
+    /// and returns an [`IncomingFilterOutcome`] that determines what happens next.
+    ///
+    /// Implementers have full access to the [`Incoming`] and can use any of its
+    /// methods (including [`Incoming::decrypt`]) to make their decision. Note that
+    /// `decrypt()` is relatively expensive, so filters should reject based on
+    /// cheaper signals (e.g. remote address) first.
+    ///
+    /// For a higher-level filter API that splits the decision into named methods
```
Frando: I think this is no longer the case, the example doesn't do that anymore

iroh/src/protocol.rs
Frando: *edit: this and a view other comments are replies to asmir's comments, which github weirdly doesnt show as such so my replies seem out of context... :cry:*
I'd say this is just a bug then, your filter shouldn't panic. We don't usually protect against panics in user-supplied hooks so far, don't think we need to start that here.

iroh/src/protocol.rs
Frando: Yeah, a log would be good. maybe even an `iroh::_events::conn::refused` (if you grep for `_events::conn` you can see the other events i added recently).

iroh/src/protocol.rs
Frando: if you have a rough idea of when it fails, would be great to add this to the docs somewhere

iroh/src/protocol.rs
rklaehn: I hope not. But if we do there is nothing in this PR we can do, the ignore mechanism is a core thing of noq that we just expose even before this PR.

rklaehn:
> Yeah, a log would be good. maybe even an iroh::_events::conn::refused (if you grep for _events::conn you can see the other events i added recently).

I added a log. Not sure about the event. Do we want an event every time we refuse, or just when we refuse because the filter incorrectly tells us to retry?

For the filter misbehaving, that isn't really fault of the endpoint, so the event would have to be something about the router. For an event for every reject, that would be somewhere different, no?

PR #3955, opened 2026-02-18
feat!: update relay protocol to iroh-relay-v2, add new Health frame, test protocol update mechanism

Frando:
## Description

This branch add a `iroh-relay-v2` protocol. The only changes to the `iroh-relay-v1` protocol are:
* removed `Health` frame (id 11) with string payload
* added `Status` frame (id 13) with binary-encoded enum payload that can be extended with further variants or payloads
* From now on, unknown frames will be ignored in iroh (instead of erroring out)

The actual change is thus quite minor, but serves as a real-world test for our protocol negotiation.

The PR also fixes a bug in how the relay server parsed the supported protocols; it is harmless because all existing deployments support a single version anyway, but good to fix now.

Everything is fully backwards-compatible on the wire (old clients can talk to new relays, and new clients can talk to old relays).

## Breaking Changes

* `iroh_relay::http::RELAY_PROTOCOL_VERSION` has been removed. The relay protocol now supports version negotiation between client and server. Use `iroh_relay::http::ProtocolVersion` instead. For the previous constant value, use `ProtocolVersion::V1.to_str()`.
* `iroh_relay::server::client::Config` has a new field `protocol_version: ProtocolVersion`.
* `iroh_relay::protos::relay::RelayToClientMsg` has a new `Status(Status)` variant and the existing `Health` variant is deprecated. Clients should handle `RelayToClientMsg::Status` for connection health information. The `Health` variant is still sent by V1 relay servers but should not be relied upon going forward.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.


iroh-relay/src/server/http_server.rs
```diff
@@ -568,16 +568,17 @@ impl RelayService {
                     details: "header value is not ascii".to_string()
                 })
             })?;
-        let supports_our_version = subprotocols
-            .split_whitespace()
```
Frando, line 600: The [spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-WebSocket-Protocol#sub-protocols) says that `Sec-Websocket-Protocol` is a **comma-separated** list of values. The current impl splits *by whitespace*, though. This is unfortunate because it would mean that if i connect with a header value of `iroh-relay-v2, iroh-relay-v1`, current relays would split it wrong. This is important though if we want to keep things working with old relays. Maybe we are lucky though: The client should specify protocols in order or preference, so `iroh-relay-v1` would come last if we add `iroh-relay-v2`. The split-by-whitecase will accept the requeset if (and only if) `iroh-relay-v1` is the *last* value, because all others would wrongly include the trailing comma. This means that relays deployed before this PR would still accept `Sec-Websocket-Protocol: iroh-relay-v1` and also `Sec-Websocket-Protocol: iroh-relay-v2, iroh-relay-v1`, but would (wrongly) deny `Sec-Websocket-Protocol: iroh-relay-v1, iroh-relay-v2`. Which .. might be fine to accept.

The alternative would be to skip standard adherence and keep splitting by whitespace during iroh 1.0. Would work, but it would be quite suboptimal IMO to not fix this oversight, if we can.

iroh-relay/src/server/client.rs
```diff
@@ -166,10 +173,19 @@ impl Client {
 
     pub(super) fn try_send_health(
         &self,
-        problem: String,
+        status: HealthStatus,
     ) -> Result<(), TrySendError<RelayToClientMsg>> {
-        self.message_queue
-            .try_send(RelayToClientMsg::Health { problem })
+        let message = match self.protocol_version {
```
Frando, line 178: This where the negotiated version is checked and behavior is adapted. Not sure how much I like it, it feels a bit manual. I first thought to move it a layer up, closer to the encoding, but this doesn't make it much easier. The `RelayToClientMsg` enum is also used when parsing messages on the client side, so it needs to have a variant for the stringly V1Health frame if we want to keep supporting older relays that might emit this frame (well, they don't because there's no released version that emits it, but still).

iroh-relay/src/server/http_server.rs
matheus23: Phew I'm happy this is in the relay server not in the iroh client :grimacing:

matheus23: Yes to fixing!

iroh-relay/src/server/http_server.rs
```diff
@@ -568,16 +568,17 @@ impl RelayService {
...
-            supports_our_version,
-            RelayUpgradeReqError::UnsupportedRelayVersion {
-                we_support: RELAY_PROTOCOL_VERSION,
-                you_support: subprotocols.to_string()
-            }
-        );
+        let protocol_version = subprotocols
+            .split(",")
+            .map(|s| s.trim())
+            .flat_map(|s| ProtocolVersion::try_from(s).ok())
```
matheus23: Prefer `.filter_map` over `flat_map`, generally. I find it clearer myself and it technically has an optimization where `size_hint` is better.

matheus23: Might even be nice to replace the TryFrom for ProtocolVersion with a dedicated `ProtocolVersion::match_from_string`. Then we could do `.filter_map(ProtocolVersion::match_from_string)`

iroh-relay/src/http.rs
```diff
@@ -13,10 +15,70 @@ pub const RELAY_PATH: &str = "/relay";
...
+///
+/// Sent as the websocket sub-protocol header `Sec-Websocket-Protocol` from
+/// the client. The server picks the best supported version and replies with it.
+///
+/// Variants are ordered by preference (highest first), so the [`Ord`] impl
+/// can be used during negotiation to pick the best version.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Default, VariantArray)]
+pub enum ProtocolVersion {
+    /// Added in iroh 0.97.0.
+    /// - Deprecated `Health` frame (id 11)
```
matheus23: There's no need to "deprecate" frames in newer versions. You just fully replace them/remove them.
We negotiated that in the version after all - we don't have to "wait" for some grace period to remove them.

iroh-relay/src/http.rs
```diff
@@ -13,10 +15,70 @@ pub const RELAY_PATH: &str = "/relay";
...
+    /// - Added new `Health` frame (id 13)
+    /// - Changed behavior such that unknown frames are allowed
+    #[default]
+    V2,
+    /// Deprecated version 1 (before iroh 0.97.0)
+    V1,
+}
```
matheus23: I think you can also use strum to generate the parsing code using [`derive(EnumString)`](https://docs.rs/strum/latest/strum/derive.EnumString.html) (and also `strum_macros::Display`)

iroh-relay/src/server/http_server.rs
ramfox: omg yes fix this. The fact that we might get lucky on the ordering for older versions is nice but kind of scary.

iroh-relay/src/http.rs
```diff
@@ -13,10 +15,70 @@ pub const RELAY_PATH: &str = "/relay";
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Default, VariantArray)]
+pub enum ProtocolVersion {
+    /// Added in iroh 0.97.0.
+    /// - Deprecated `Health` frame (id 11)
+    /// - Added new `Health` frame (id 13)
+    /// - Changed behavior such that unknown frames are allowed
+    #[default]
+    V2,
+    /// Deprecated version 1 (before iroh 0.97.0)
+    V1,
```
ramfox: doesn't declaring them in this order would mean calls to `max` would return V1 rather than V2 (which we call in `handle_relay_ws_upgrade`) (since we derive Ord)

ramfox:
only addition I might add here something on the client side that acknowledges we negotiated the protocol. Maybe just a debug line at this point, since it's not used anywhere on the client side yet?

matheus23:
I wrote this as feedback in discord, but that's quickly lost to the archives, so just repeating some higher-level feedback here:

My thinking was that we should separate our protocol versions by module. So e.g. now, we'd introduce two modules: `iroh-relay/src/protos/{v1,v2}/proto.rs`, etc. There's probably some thinking needed to figure out if we want to duplicate `common.rs` as well/if there's a way of extracting only the useful stuff from it, but otherwise that's roughly what I'm thinking.

If we do this, it is much easier to look at *one* protocol and see how it works rather than mixing multiple versions in the same code. I think that can become quite messy.

It's also nice to be able to have a bunch of `#[cfg(feature = "iroh-relay-v1")] pub mod v1;` and similar in `protos.rs`. This would then eventually make it possible to reduce code size if that becomes a concern.

iroh-relay/src/server/http_server.rs
Frando: done

iroh-relay/src/http.rs
Frando: done

iroh-relay/src/http.rs
Frando: done

iroh-relay/src/http.rs
Frando: good observation, thanks. changed and added a test for the protocol negotiation.

iroh-relay/src/protos/common.rs
```diff
@@ -45,6 +45,8 @@ pub enum FrameType {
     Ping = 9,
     /// 8 byte payload, the contents of ping being replied to
     Pong = 10,
+    /// REMOVED since relay-protocol-v2 (use [`Self::Health`] instead)
+    ///
```
matheus23: This comment doesn't make any sense to me anymore.

matheus23: Should this perhaps refer to `Self::HealthStatus`?

Frando: Right, thanks. Fixed.

iroh-relay/src/http.rs
```diff
@@ -13,10 +15,88 @@ pub const RELAY_PATH: &str = "/relay";
...
+}
+
+impl ProtocolVersion {
+    /// Returns an iterator of all supported protocol version identifiers, in order of preference.
+    pub fn all() -> impl Iterator<Item = &'static str> {
+        Self::VARIANTS
+            .iter()
+            .map(ProtocolVersion::to_str)
+            // Need to reverse order so that the latest version comes last.
+            .rev()
```
matheus23: That's somewhat unusual.

In TLS, versions are ordered from most preferred to least preferred. Can we do the same?

Frando: Oh yes, it was just the comment that was wrong. We need to call `rev` so that the lastest version comes *first*, because the variants are ordered with latest-as-last so that the derived `Ord`  impl puts the latest version as `max`. Fixed the comment.

matheus23 (approved):
Let's ship this - at the very least for the important bugfix.

Let's discuss changes to code structure later and open another PR for those if need be.

iroh-relay/src/http.rs
```diff
@@ -42,13 +42,14 @@ pub const CLIENT_AUTH_HEADER: HeaderName = HeaderName::from_static("x-iroh-relay
 #[strum(parse_err_ty = UnsupportedRelayProtocolVersion, parse_err_fn = strum_err_fn)]
 // Needs to be ordered with latest version last, so that the `Ord` impl orders by latest version as max.
 pub enum ProtocolVersion {
-    /// Version 1 (before iroh 0.97.0)
+    /// Version 1 (the only vesion supported until iroh 0.98.0)
```
matheus23: typo: "vesion" -> "version"

matheus23:
> From now on, unknown frames will be ignored in iroh (instead of erroring out)

Hmmm. I'm sorry I just reread the PR description again and that's why I'm coming back to this.

We should reconsider this (after merging!). We can always change this IMO, that's not a problem, as long as we change this before we do the next breaking change *without* adding another relay protocol version (and I think we'll always want to use a new protocol version even if we add optional frames).

This is contrary to what e.g. QUIC does: If you receive a frame you don't know - that's a protocol violation.

We should really think through what this means in various cases again before we do the next protocol change.

Frando:
> This is contrary to what e.g. QUIC does: If you receive a frame you don't know - that's a protocol violation.

Good point. I think you are right and we should fix that. I'll look into it.

PR #3957, opened 2026-02-19
chore(ci): colocate sims into iroh repo

Arqu:
## Description

We've done this in the quinn repo already. This should make life easier to both fiddle with sims while working but also to easily sync up changes when needed instead of having to sync merges across repos. Netsim itself changes less frequently and also is a lot more decoupled and tolerant of moving independently this way.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/.github/workflows/netsim_runner.yaml
+++ b/.github/workflows/netsim_runner.yaml
@@ -132,6 +132,12 @@ jobs:
         ./setup.sh
         ./cleanup.sh || true
 
+    - name: Copy sim configs
+      run: |
+        cp -r .github/sims/iroh/ ../chuck/netsim/sims/iroh/
+        cp -r .github/sims/integration/ ../chuck/netsim/sims/integration/
+        cp -r .github/sims/paused/ ../chuck/netsim/sims/paused/
+
     - name: Copy binaries to right location
       run: |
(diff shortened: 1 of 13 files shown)


.github/workflows/netsim_runner.yaml
```diff
@@ -132,6 +132,12 @@ jobs:
         ./setup.sh
         ./cleanup.sh || true
 
+    - name: Copy sim configs
+      run: |
+        cp -r .github/sims/iroh/ ../chuck/netsim/sims/iroh/
+        cp -r .github/sims/integration/ ../chuck/netsim/sims/integration/
+        cp -r .github/sims/paused/ ../chuck/netsim/sims/paused/
```
flub, line 139: So this is because netsim still checks in it's own hardcoded paths? I guess netsim could take the files to look at as an argument somehow, which might be a bit cleaner. But this will also work.

flub (approved):

.github/workflows/netsim_runner.yaml
Arqu: It does take paths as arguments. This was to preserve the old structure. I intend to do another cleanup pass on netsim in the near future now that it's getting some love. I want to move netsim up and rename chuck to netsim and do some other cleanup which would include reworking these flows a little too.

PR #3960, opened 2026-02-20
feat(iroh)!: add address filtering and reordering for Address Lookup Services

ramfox:
# Address Filtering for Address Lookup Services

### The Context

[PR #3691](https://github.com/n0-computer/iroh/pull/3691) surfaced an assumption mismatch between users and iroh's address lookup service. This user expected that relay URLs were being published over mDNS and could not understand why connections weren't being established. We never considered publishing relay URLs on mDNS internally and so our implementation only publishes ip addresses and user defined data.

It turned out there was a valid reason the user needed this: their environment had a firewall that permitted broadcast UDP traffic (which mDNS uses) but blocked other kinds of UDP traffic. So mDNS address lookup was working, but without a relay URL in the advertisement, the two endpoints had no way to reach each other — direct connections were blocked and there was no relay fallback advertised. The fix for their use case was to include the relay URL in mDNS advertisements.

This incident revealed that we had left the choice of what to and not to publish to the address lookup implementations entirely, and didn't leave room for the possibility/flexibility that users may have specific needs while advertising their addresses. 

### Custom Transports Make This Worse

The upcoming Custom Transports feature will introduce a `Custom(CustomAddr)` variant to `TransportAddr`, allowing users to plug in arbitrary transport layers (e.g., Bluetooth). This sharpens the problem:

- Users may want to advertise a custom transport address on one address lookup service but not another.
- Most of our existing address lookup services are built on top of DNS TXT records, which have hard byte-size limits. Custom transport addresses will likely need to be encoded into these records, and there may simply not be room for all of them, and we have no guidelines on what to do if this happens.
- There is no current mechanism for users to express preferences about which addresses get published where.

This is manageable — an address lookup service can document its constraints and make best-effort decisions about what it can fit — but we need to give users a way to participate in that decision. For example, in a byte-constrained mDNS advertisement, a user might want to ensure their preferred custom transport address is published first, rather than leaving the ordering up to the address lookup service itself.

### So What Do We Need?

We need a mechanism that:
1. Makes it clear to users that they can influence which addresses each address lookup service publishes.
2. Makes it clear to address lookup implementors that they need to reckon with publishing addresses that are not just IP addrs.
3. Still allows each address lookup service to make its own final decisions about what it can actually publish, given its own constraints.

The fourth and nice-to-have would be some way to force the address lookup implementors into ensuring they have good documentation about the preference of the address lookup service when it comes to illustrating what it can and cannot publish (in terms of kind of addresses as well as number of addresses). Haven't figured this out, so it may just be that we lead by example and have good documentation ourselves.

---

### Proposed Design

We introduce a concrete `AddrFilter` type and add required methods to the ~~`AddressLookup` and~~ `IntoAddressLookup` trait~~s~~ that ensures every address lookup implementation accepts a filter. Whether or not they use/respect the filter, is not something we can control in this setup, but it at least gives the expectation that we expect it.

#### `AddrFilter`

```rust
/// A filter and/or reordering function applied to transport addresses before publishing.
///
/// Takes the full set of transport addresses and returns them as an ordered `Vec`,
/// allowing both filtering (by omitting addresses) and reordering (by controlling
/// the output order). A `BTreeSet` cannot preserve a custom order, so the return
/// type is `Vec` to make reordering possible.
///
/// See the documentation for each address lookup implementation for details on
/// what additional filtering the implementation may perform on top.
// KASEY:
/// Our recommendation would be that the address lookup service will apply this
/// before its own internal filtering, before it publishes.
#[derive(Clone)]
pub struct AddrFilter(
    Arc<dyn Fn(BTreeSet<TransportAddr>) -> Vec<TransportAddr> + Send + Sync + 'static>
);

impl AddrFilter {
    // KASEY: I don't know if this is necessary, but would likely make it nicer to
    // work with
    pub fn new(
        f: impl Fn(BTreeSet<TransportAddr>) -> Vec<TransportAddr> + Send + Sync + 'static
    ) -> Self {
        Self(Arc::new(f))
    }

    // KASEY: I don't know if this is necessary, but would likely make it nicer to
    // work with
    pub fn apply(&self, addrs: BTreeSet<TransportAddr>) -> Vec<TransportAddr> {
        (self.0)(addrs)
    }
}
```

#### Changes to `IntoAddressLookup`

We need to be able to add the `AddrFilter` when building the discovery service. To do this, we can add a required `with_addr_filter` method is added to `IntoAddressLookup`

```rust
pub trait IntoAddressLookup: std::fmt::Debug + Send + Sync + 'static {
    /// Sets the address filter for this address lookup service.
    fn with_addr_filter(self, filter: AddrFilter) -> Self where Self: Sized;

    fn into_address_lookup(
        self,
        endpoint: &Endpoint,
    ) -> Result<impl AddressLookup, IntoAddressLookupError>;
}
```

#### Usage

What would this look like?

```rust
endpoint
    .address_lookup(PkarrPublisher::n0_dns()
        .with_addr_filter(AddrFilter::new(|addrs| {
            // only publish relay and IP addresses to pkarr, not custom transports,
            // returned as a Vec to allow reordering
            addrs.into_iter()
                .filter(|a| !matches!(a, TransportAddr::Custom(_)))
                .collect::<Vec<_>>()
        })))
    .address_lookup(MyBluetoothLookup::new()) // handles Custom addrs internally
```

---

### We also change the `AddressLookup` trait to require a `set_addr_filter`.

It's likely we only need to set this once, when the address lookup is built. BUT if we want to give users the ability to adjust this dynamically, it may be best to also add a `set_addr_filter` or `with_addr_filter` method on the `AddressLookup` trait itself. However, the user would need to hold onto a reference/clone of the address lookup service themselves, since calling this on `ep.address_lookup().with_addr_filter` would likely mean setting that specific filter to all the different lookup services that exists. But maybe we want that functionality too!

### Each address lookup system has its own builder methods, no trait change

Each address lookup service builder could independently add a filter method to its own builder. No changes to the traits would be required.

I rejected this because it meant no consistency and no way to enforce even limited consistency/expectations. Users have no reason to expect filtering to be available or to look for it and address lookup service implementators get no guidance.

### This isn't a concern for the address lookup services at all - but for the endpoint

The other option is that this isn't a setting on the address lookup service, but is instead a setting on the endpoint. There are two issues here, I believe.
First, this would likely mean having a single blanket filter/reduce, so the user can't customize it per-address lookup. Second, since `publish` expects an `EndpointAddr` with a `BTreeSet<TransportAddr>` we cannot preserve order once we turn the vec back into a btreeset. The other other option would be then to adjust the field in the `EndpointAddr` to be a vec, but I don't think we want that?

### Do something weird with `Ord` and `TransportAddr`

Tbh I'm not sure where you would start with this, since you would want it to be user supplied...

### Other options?

I am open to hearing them :)

---

### This gives tools and expectations, not guarantees

The `AddrFilter` mechanism gives users a way to express intent — "these are the addresses I want published to this service" — and creates the expectation that the address lookup service will use it. But the address lookup service is the final authority over what it actually publishes, so it could decide not to use the user supplied filter and can decide for itself what addresses it wants to publish. 

But likely, this is for the best. Address lookup services can have structural constraints the user cannot override. Eg, An address lookup service that encodes addresses as DNS TXT records cannot publish infinite `CustomAddr` variants it needs to be able to confidently only publish what it can. Other services may have reasons they can't publish them entirely.

### But this also means we should have a lot of documentation about our expectations.

The filter application should be consistent across all implementation:

1. The user-provided `AddrFilter` is applied first.
2. The address lookup service's own internal filtering is applied second.

Also, every address lookup implementation should document:

- Whether and how the user-provided `AddrFilter` is applied.
- What additional internal filtering the implementation applies (e.g., "only relay URLs are published; direct addresses are dropped regardless of the filter").
- Any capacity limits that may cause addresses to be dropped even after filtering (e.g., "only the first N addresses are published due to DNS TXT record size constraints").
- When capacity limits exist, whether addresses are dropped silently or with a warning log.

The goal is that a user reading the documentation for any address lookup service can predict exactly what will be published, without reading the source.

## Feedback

I'd love a way to have more guarantees here, but I haven't figured out the structure that would allow it. In fact, potentially the fact that this isn't a 100% rigid structure may actually be beneficial for some address lookup services.

## Behavior Changes

By default, no address lookup service applies any filtering — all addresses are passed through to publishing as-is. Filtering only occurs if an `AddrFilter` is explicitly set, either via a preset or a user-supplied closure. This is a breaking change for Pkarr publishing, which checked to see if a relay url existed, and if so, cleared all ip addresses, and Dht publishing, which had an explicit `include_direct_addresses` option.

For users using our defaults, this will not effect them. For users that have relays disabled, this will effect them if they use the default n0 preset, BUT, likely this will not effect most, because a user that has disabled relays they probably have not used the default n0 preset, and instead used the `empty_builder` and built whatever address lookup services they needed on their own.

#### Presets

Since we are leaving the choice for what to publish to the users, I've updated the Presets to include that the default explicitly filters address to only include the relay url. This is the same behaviour as before, but needed to be updated because the option to `include_direct_addresses` for, eg `DhtAddressLookup`, was removed.

**New preset: N0DisabledRelay**
I added a new preset that filters relays from published addresses and sets the relay mode to disabled. Not married to keeping it, but could be a good shorthand for folks. Did not add this as an option on the Endpoint::builder (eg Endpoint::disabled_relay_builder), however.

## Breaking Changes
- `iroh`
    - `removed`
        - `DhtAddressLookupBuilder::include_direct_addresses` removed. You can effect what addresses are published by using the new `DhtAddressLookupBuilder::set_addr_filter` method.
    - `changed`
        - trait `IntoAddressLookup` renamed to `AddressLookupBuilder` 

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -27,5 +27,6 @@ ignore = [
 
 [sources]
 allow-git = [
-  "https://github.com/n0-computer/quinn",
+  "https://github.com/n0-computer/noq",
+  "https://github.com/n0-computer/net-tools",
 ]


iroh-relay/src/endpoint_info.rs
```diff
@@ -210,6 +211,48 @@ impl EndpointData {
     pub fn has_addrs(&self) -> bool {
         !self.addrs.is_empty()
     }
+
+    /// Apply the given filter to the current addresses.
+    ///
+    /// Returns a vec to preserve re-ordering of addresses.
+    pub fn filtered_addrs(&self, filter: AddrFilter) -> Vec<TransportAddr> {
+        filter.apply(self.addrs.clone())
```
rklaehn: We are cloning the addrs and then making a vec out of them. This seems inefficient.

The order of elements in a BTreeSet is the natural order of the elements according to its Ord instance. And this takes an owned BTreeSet. So I see 2 options: 1. we filter the cloned BTreeSet in place or 2. we change this API to return an iterator and avoid the new allocation entirely.

Not sure if the latter will work with all the dyn, but currently I don't see why it would not work.

iroh/src/address_lookup/pkarr/dht.rs
```diff
@@ -199,6 +201,12 @@ impl Builder {
         self
     }
 
+    /// Sets a filter to control which addresses are published by this service.
+    pub fn set_addr_filter(mut self, filter: Option<AddrFilter>) -> Self {
```
rklaehn: So we have a filter per AddrLookup. Good. Also nice that we can get rid of the "include direct addrs" as a special case of a filter.

iroh/src/address_lookup/pkarr/dht.rs
```diff
@@ -296,15 +307,21 @@ impl AddressLookup for DhtAddressLookup {
             tracing::debug!("no keypair set, not publishing");
             return;
         };
-        if !data.has_addrs() {
+
+        // apply user-supplied filter
+        let addrs = match self.0.filter {
```
rklaehn: Not sure about the perf implications, but an option would be to have an AddrFilter identity and instead of having filter being an Option just set it to identity by default.

iroh/src/endpoint/presets.rs
```diff
@@ -43,29 +44,99 @@ pub trait Preset {
...
+///
 /// [`N0_DNS_PKARR_RELAY_PROD`]: crate::address_lookup::N0_DNS_PKARR_RELAY_PROD
 /// [`N0_DNS_PKARR_RELAY_STAGING`]: crate::address_lookup::N0_DNS_PKARR_RELAY_STAGING
 #[derive(Debug, Copy, Clone, Default)]
 pub struct N0;
 
 impl Preset for N0 {
     fn apply(self, mut builder: Builder) -> Builder {
-        builder = builder.address_lookup(PkarrPublisher::n0_dns());
+        let filter = AddrFilter::new(|addrs| {
```
rklaehn: It might be nice to have an AddrFilter::just_relays() instead of implementing this inline here. Could be useful for other people as well.

iroh/src/endpoint/presets.rs
```diff
@@ -43,29 +44,99 @@ pub trait Preset {
...
+/// the endpoint's IP addresses.
+///
+/// [`N0_DNS_PKARR_RELAY_PROD`]: crate::address_lookup::N0_DNS_PKARR_RELAY_PROD
+/// [`N0_DNS_PKARR_RELAY_STAGING`]: crate::address_lookup::N0_DNS_PKARR_RELAY_STAGING
+#[derive(Debug, Copy, Clone, Default)]
+pub struct N0DisableRelay;
+
+impl Preset for N0DisableRelay {
+    fn apply(self, mut builder: Builder) -> Builder {
+        let filter = AddrFilter::new(|addrs| {
```
rklaehn: Same here. Have a fn on AddrFilter that makes an AddrFilter (not sure if making this a constant is possible, but it doesn't matter that much... )

iroh-relay/src/endpoint_info.rs
rklaehn: Ah, never mind, you also want to allow reordering. Still, the two copies are unnecessary. I guess the comment should be "to allow reordering" then.

rklaehn:
Is there any downside in taking the BTreeSet by ref?

https://github.com/n0-computer/iroh/pull/3972

rklaehn:
Not quite sure if this is a net win, but it does reduce boilerplate for all AddrLookup implementers:

https://github.com/n0-computer/iroh/tree/rklaehn/identity-filter

ramfox:
@rklaehn merged your PR in and cherry picked the identity commit in too. Very in favor of all those changes.

Fixed the doc line for `EndpointInfo::filtered_addrs` as well.

rklaehn:
> Custom Transports Make This Worse

I don't really think custom transports are that different. You just need more filters. Custom transport addresses will have a default way to be stored, printed, parsed. So it's just another addr type. ip_only and relay_only will already get rid of them.

So, to clarify: the fn with_addr_filter on the IntoAddressLookup is just to set the expectation that a builder should have a way to set the filter. It is **not** called when creating the AddressLookup, otherwise the filter could just be an argument to into_address_lookup.

Also, for the default impl we can't implement it and just return self. If you already have an AddressLookup there is nothing you can do because as of now we don't implement dynamic filter change.

So yes, that makes sense. Make the builders consistent. But then the name IntoAddressLookup somehow seems weird. An `Into...` trait that has some weird extra method? Maybe we can just make this trait the trait you should implement for builders, so maybe call it just `AddressLookupBuilder`?

Other than that all of this seems fine.

rklaehn (commented):
See my note about the name of IntoAddressLookup, otherwise I think this is fine.

rklaehn (approved):
Approved if we rename IntoAddressLookup

Frando:
I'm having another look at this even though it's merged - and am wondering why you didn't go for a combinator style approach instead of having each address lookup impl deal with the address filter themselves?

i.e. have `AddresssLookupBuilder::with_address_filter` be a default impl on the trait that returns `Filtered<Box<dyn DynAddressLookupBuilder>>` and `build()` on that returns a `Filtered<AddressLookup>` which handles the filter-before-publish.

I can give it a try but before doing so wanted to check if this was considered already or not.

Edit: Gave it a try here: https://github.com/n0-computer/iroh/pull/3987

PR #3967, opened 2026-02-24
refactor: update quinn and simplify path handling in `RemoteStateActor`

Frando:
## Description

* Bumps iroh-quinn to `main` as of now.
* In the `RemoteStateActor`'s `ConnectionState` only track open paths. We never accessed the list of non-open paths, so it seems much simpler to me to only track what we need.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/socket/remote_map/remote_state.rs
+++ b/iroh/src/socket/remote_map/remote_state.rs
@@ -966,7 +966,7 @@ impl RemoteStateActor {
                     path_id = ?id,
                 );
 
-                // If one connection abandones a path, close it on all connections.
+                // If one connection abandons a path, close it on all connections.
                 for (conn_id, conn_state) in self.connections.iter_mut() {
                     let Some(path_id) = conn_state.paths_by_addr.get(&path_remote) else {
                         continue;


Cargo.lock
```diff
@@ -1157,7 +1157,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
```
flub, line 1160: lol, we're back to a linux-created lockfile.

iroh/src/socket/remote_map/remote_state.rs
```diff
@@ -946,18 +948,18 @@ impl RemoteStateActor {
...
-                    self.paths.abandoned_path(&addr);
-                }
+            PathEvent::Discarded { id, path_stats } => {
+                trace!(%id, ?path_stats, "path discarded");
             }
-            PathEvent::LocallyClosed { id, .. } => {
-                let Some(path_remote) = conn_state.paths.get(&id).cloned() else {
-                    debug!("path not in path_id_map");
+            PathEvent::Abandoned { id, .. } => {
+                // This is the last event for this path: remove from conn_state.
```
flub: erm, no, the discarded event is still emitted...

Maybe "From now on we no longer want to use this path" or something. Or... nothing?

iroh/src/socket/remote_map/remote_state.rs
```diff
@@ -946,18 +948,18 @@ impl RemoteStateActor {
...
+                let Some(path_remote) = conn_state.remove_path(&id) else {
+                    debug!(%id, "path not in path_id_map");
                     return;
                 };
+                // Also remove the path from the remote-global path tracking.
+                self.paths.abandoned_path(&path_remote);
+
                 event!(
                     target: "iroh::_events::path::closed",
```
flub: let's rename this to `abandoned` too. the events are internal debugging tools, not user-facing. so let's keep them as specific and accurate as possible.

Cargo.lock
Frando: hu, what's going on there? should I do sth about it?

iroh/src/socket/remote_map/remote_state.rs
Frando: fixed

iroh/src/socket/remote_map/remote_state.rs
Frando: fixed

Cargo.lock
matheus23: No, this is something that's been going on forever. We've never done anything about it but it also has never caused a problem.

PR #3972, opened 2026-02-25
feat(iroh): Let filtered_addrs take the BTreeSet by ref to avoid one copy.

rklaehn:
## Description

We still return a Vec so we can reorder.


ramfox:
Love it. Merging it in. Thank you.

PR #3973, opened 2026-02-25
feat: allow to customize TLS trust roots

Frando:
## Description

This adds customization to how certificates for non-iroh TLS connections are verified. It adds a new `CaRootsConfig` struct  which provides these options:

* Embedded WebPki roots (like currently), with optional extra roots
* Operating system's roots, with optional extra roots
* Custom list of roots only
* No verification (gated behind `test-utils` feature, like now)

This struct can be built into a `rustls::ClientConfig`, which is then used for:

* Connecting to relays, both HTTPS probes, QAD, and actual WebSocket connections
* Connecting to pkarr servers
* In the DNS resolver, if DNS-over-HTTPS is configured

I think that's all places where we use TLS connections in iroh. Will have another look if we missed any.

## Breaking Changes

Removed:
* `iroh::endpoint::Builder::insecure_skip_cert_verify`, use `Builder::ca_roots_config(CaRootsConfig::insecure_skip_verify())` instead
* `iroh_relay::client::ClientBuilder::insecure_skip_cert_verify`, remove `ClientBuilder::tls_client_config` instead
* `iroh::address_lookup::pkarr::PkarrResolverBuilder::build` and `iroh::address_lookup::pkarr::PkarrPublisherBuilder::build` now take a `tls_config: rustls::ClientConfig` arg. If building via `Endpoint::add_discovery` or `endpoint::Builder::discovery` no change is needed since the TLS  config is passed from the endpoint builder. If constructing manually for some reason, either pass `Endpoint::tls_config`  or construct a TLS config yourself, e.g. via `iroh::tls::CaRootsConfig::client_config`.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-relay/Cargo.toml
+++ b/iroh-relay/Cargo.toml
@@ -139,7 +139,7 @@ vergen-gitcl = "1.0.8"
 [features]
 default = ["metrics"]
 # Enables fetching TLS trust anchors from the operating system
-tls-system-certs = ["dep:rustls-platform-verifier"]
+platform-verifier = ["dep:rustls-platform-verifier"]
 server = [
     "metrics",
     "dep:clap",
--- a/iroh-relay/src/tls.rs
+++ b/iroh-relay/src/tls.rs
@@ -33,7 +33,7 @@ enum Mode {
     /// Use the operating system's certificate facilities for verifying the validity of TLS certificates.
     ///
     /// See [`rustls_platform_verifier`] for details how roots are retrieved on different platforms.
-    #[cfg(feature = "tls-system-certs")]
+    #[cfg(feature = "platform-verifier")]
     System,
     /// Only trust explicitly set root certificates.
     ExtraRootsOnly,
@@ -60,7 +60,7 @@ impl CaRootsConfig {
     ///
     /// Note: Additional certificates added via [`Self::with_extra_roots`] will be ignored on Android due to
     /// missing support in [`rustls`].
-    #[cfg(feature = "tls-system-certs")]
+    #[cfg(feature = "platform-verifier")]
     pub fn system() -> Self {
         Self {
             mode: Mode::System,
@@ -112,7 +112,7 @@ impl CaRootsConfig {
         crypto_provider: Arc<CryptoProvider>,
     ) -> io::Result<Arc<dyn ServerCertVerifier>> {
         Ok(match self.mode {
-            #[cfg(feature = "tls-system-certs")]
+            #[cfg(feature = "platform-verifier")]
             Mode::System => {
                 #[cfg(not(target_os = "android"))]
                 let verifier = rustls_platform_verifier::Verifier::new_with_extra_roots(
--- a/iroh/Cargo.toml
+++ b/iroh/Cargo.toml
@@ -151,7 +151,7 @@ test-utils = ["iroh-relay/test-utils", "iroh-relay/server", "dep:axum"]
 address-lookup-mdns = ["dep:swarm-discovery"]
 address-lookup-pkarr-dht = ["pkarr/dht"]
 # Enables fetching TLS trust anchors from the operating system
-tls-system-certs = ["iroh-relay/tls-system-certs"]
+platform-verifier = ["iroh-relay/platform-verifier"]
 qlog = ["quinn/qlog"]
 # Use private Apple APIs to send multiple packets in a single syscall.
 fast-apple-datapath = ["quinn/fast-apple-datapath"]


Frando:
Interesting: I switched the default to use the operating system's defaults, and tests fail on windows and android. Will have a closer look soon. But maybe we should stay with embedded WebPKI as default then.

Edit: Windows fails the integration test with
`
2026-02-25T13:27:49.7969117Z 2026-02-25T13:27:38.015343Z  WARN actor:reportgen-actor:run-probe{proto=Https delay=200ms relay=RelayConfig { url: RelayUrl("https://staging-euw1-1.relay.iroh.network./"), quic: Some(RelayQuicConfig { port: 7842 }) }}: iroh::net_report::reportgen: probe failed: Failed to run HTTPS probe: HTTP request failed: error sending request for url (https://staging-euw1-1.relay.iroh.network./ping): client error (Connect): invalid peer certificate: NotValidForName
`
i.e. it seems there's no host certs, or they are somehow not trusting our cert?

Frando:
I switched the default back to the embedded roots.

iroh/src/endpoint.rs
```diff
@@ -607,10 +612,14 @@ impl Builder {
     /// Skip verification of SSL certificates from relay servers
     ///
     /// May only be used in tests.
+    // TODO: Remove?
```
dignifiedquire: if we can do this entirely on the webtls config now, lets drop it

iroh/src/endpoint.rs
dignifiedquire, line 620: should this be an option on the webtlsconfig?

dignifiedquire, line 631: what about these other tls options?

Frando: No, this setting is about TLS config for iroh connections, not WebTls

Frando: We could add this to WebTlsConfig as well. Currently WebTls doesn't ever have resumption configured. Do we? Don't think there's any 0RTT though for our web uses

iroh-relay/src/tls.rs
```diff
@@ -0,0 +1,229 @@
...
+    client::{ClientConfig, WebPkiServerVerifier, danger::ServerCertVerifier},
+    crypto::CryptoProvider,
+};
+use webpki_types::CertificateDer;
+
+/// Configuration for verifying TLS certificates for HTTPS and other non-iroh TLS connections.
+#[derive(Debug, Clone)]
+pub struct WebTlsConfig {
+    inner: ClientConfig,
+}
```
matheus23: Two things:
1. I'm not sure if it's a good idea to store the final `rustls::ClientConfig` here. In my mind, this should only set the CA roots (And btw that's also what the PR title says).
   If we allow customizing the `CryptoProvider` in the future (I'm currently working on that), this will only change the crypto provider for all non-iroh traffic only, which I think is confusing!
   It's technically possible to use `ClientConfig::crypto_provider()` to use in iroh, but IMO that's confusing, too, given what the goal of this config is as expressed in the comment, its name and the PR title.
2. If we change this struct to only configure the CA roots, then I'd prefer renaming this to something like `CaRootConfig` or `PkiConfig`. `WebTls` is not a standard term. PKI or CA can be confusing, but I don't think it's less confusing than "Web TLS". We should of course expand on the PKI and CA acronyms and give users guidance on how to find out more about them in the documentation. "Web TLS" won't be something that's google-able, unlike "PKI roots" or "CA roots". (googling around a bit) Seems like "CaTrustRootsConfig" has the best response from the AI overlords.

Frando: I pushed a commit that changes it:
- Endpoint::builder().ca_root_config() now takes a `CaRootConfig`
- No way to set the CryptoProvider, this will come in a separate PR
- Instead of newtyping rustls::ClientConfig pass it around directly

iroh-relay/src/tls.rs
```diff
@@ -0,0 +1,163 @@
...
+};
+use webpki_types::CertificateDer;
+
+/// Configures the trust roots for verifying the validity of TLS certificates.
+///
+/// This is used throughout iroh whenever TLS connections are established that are not iroh connections.
+///
+/// This includes the connection to iroh relays, to pkarr servers, and DNS resolution over HTTPS.
+#[derive(Debug, Clone)]
+pub enum CaRootConfig {
```
matheus23: Small nit: Can we call it `CaRootsConfig`? 😅  (Sorry for the back and forth on this - rust analyzer renaming is your friend I hope)

---

Actual comment:
I don't think we should have this as an enum.
I think this should essentially be an opaque struct:
```rs
pub struct CaRootsConfig {
    verifier: Arc<dyn ServerCertVerifier>,
}
```

And then we provide methods such as `CaRootsConfig::platform_verifier()`, `CaRootsConfig::embedded_roots()`, etc.

With the way I defined the `struct` above, these have to be fallible. If we *want to*, we *could* also do sth like storing a `FnOnce() -> std::io::Result<Arc<dyn ServerCertVerifier>>` inside the struct, or maybe still store an enum, but it's not exposed, etc. It doesn't matter that much. What matters is that we avoid exposing an enum and only allow constructing the value, not destructuring it, the rest is an implementation detail I'd say.

But I'm not happy with the enum. It would need a `#[non_extensive]` at the least, but I think just having an opaque struct is better. I view it kinda like a builder (and it should be similarly extensible in the future).

Frando: I made it 
```rust
#[derive(Debug, Clone)]
pub struct CaRootsConfig {
    mode: Mode,
    extra_roots: Vec<CertificateDer<'static>>,
}
```
This keeps the API simpler IMO because the fallible step is at the time of constructing the client config, and we don't need an builder to add extra roots.

iroh/src/endpoint.rs
matheus23: I think this is out-of-scope for `CaRootsConfig` (now that we've renamed it that's maybe clearer).
Making keylog effect TLS connections other than the iroh ones is perhaps another issue/PR.

matheus23: IMO they're out-of-scope for the current `CaRootsConfig`.

Maybe in the future we group `CaRootsConfig` and friends into a `TlsConfig` down the line, but IMO that's another PR.

(Also TLS tickets are probably only relevant to iroh connections)

iroh-relay/src/tls.rs
matheus23: Right right. Yeah I missed that there's *two* fallible steps here: (1) is constructing the `dyn ServerCertVerifier` (might be infallible in e.g. the `NoCertVerifier` case) and (2) is constructing the `rustls::ClientConfig` itself.

One could argue that there might be value in moving the first error closer to where it's "decided" but I'm not sure.

So yeah - the current version looks good!

iroh-relay/src/tls.rs
```diff
@@ -0,0 +1,206 @@
...
+    client::{ClientConfig, WebPkiServerVerifier, danger::ServerCertVerifier},
+    crypto::CryptoProvider,
+};
+use webpki_types::CertificateDer;
+
+/// Configures the trust roots for verifying the validity of TLS certificates.
+///
+/// This is used throughout iroh whenever TLS connections are established that are not iroh connections.
+///
+/// This includes the connection to iroh relays, to pkarr servers, and DNS resolution over HTTPS.
```
matheus23: This should say "Certificate Authority" *somewhere*.

It would be nice to add that these CAs don't need to be trusted for the security (integrity/authenticity) of iroh connections.

iroh-relay/src/tls.rs
```diff
@@ -0,0 +1,206 @@
...
+#[derive(Debug, Clone)]
+pub struct CaRootsConfig {
+    mode: Mode,
+    extra_roots: Vec<CertificateDer<'static>>,
+}
+
+#[derive(Debug, Clone)]
+enum Mode {
```
matheus23, line 28: Generally not sure we need to duplicate the docs on this struct now.

iroh-relay/src/tls.rs
```diff
@@ -0,0 +1,206 @@
...
+    fn default() -> Self {
+        Self {
+            mode: Mode::EmbeddedWebPki,
+            extra_roots: vec![],
+        }
+    }
+}
+
+impl CaRootsConfig {
+    /// Use the operating system’s certificate facilities for verifying the validity of TLS certificates.
```
matheus23: What a curiously fancy `'` you have there :P

iroh-relay/src/tls.rs
```diff
@@ -0,0 +1,206 @@
...
+        Ok(config)
+    }
+}
+
+/// Returns iroh's default crypto provider.
+///
+/// Currently, this is [`rustls::crypto::ring`].
+pub fn default_provider() -> Arc<CryptoProvider> {
+    Arc::new(rustls::crypto::ring::default_provider())
+}
```
matheus23, line 166: I'd prefer to not have this public.

This will change in my own upcoming PR, but this would force us to always compile *some* instance of `CryptoProvider` into our libraries, which we don't want.

iroh-relay/Cargo.toml
```diff
@@ -91,6 +91,7 @@ tracing-subscriber = { version = "0.3", features = [
 ], optional = true }
 blake3 = "1.8.2"
 serde_bytes = "0.11.17"
+rustls-platform-verifier = "0.6.2"
```
matheus23: Can we make this an optional dependency?
`CaRootsConfig::system` should be gated on an optional `rustls-platform-verifier` feature being enabled.

Perhaps we can also name the feature sth else.

iroh/src/endpoint.rs
```diff
@@ -592,6 +598,14 @@ impl Builder {
         self
     }
 
+    /// Sets the TLS config for non-iroh TLS connections.
+    ///
+    /// This is used for HTTPS connections to iroh relay and pkarr address lookups servers.
```
matheus23: This doc comment needs updating. I don't think "TLS config" describes the `CaRootsConfig` precisely.

Also, it's additionally used for QAD connections to relays. Not sure if this should be mentioned here, perhaps instead we can make this more abstract, something along the lines of "used for TLS connections that utilize the existing web public key infastructure" and say that iroh connections themselves aren't protected by the web PKI.

iroh/src/endpoint.rs
```diff
@@ -1194,6 +1199,11 @@ impl Endpoint {
         Ok(self.inner.dns_resolver())
     }
 
+    /// Returns the TLS config used for connecting to iroh relays and other non-iroh TLS services.
+    pub fn tls_config(&self) -> &rustls::ClientConfig {
```
matheus23: Maybe add to the docs that this is useful for address-lookup services that want to use the same crypto provider and/or CA roots as what the user configured on the endpoint.

Also unsure about the `tls_config` name... But at the same time I can't really think of something better immediately, so feel free to keep it if you don't have a better name either.

iroh-relay/src/tls.rs
Frando: I think I copy-pasted this from `rustls-platform-verifier`;)

iroh-relay/src/tls.rs
Frando: Improved the docs, please re-review.

iroh-relay/src/tls.rs
Frando: Kept public for now after discussion in discord.

iroh-relay/Cargo.toml
Frando: Put it behind a feature flag `tls-system-certs`, open for other names though.

iroh/src/endpoint.rs
Frando: Improved the comment, please re-review.

iroh/src/endpoint.rs
Frando: Also improved the comment here.

PR #3986, opened 2026-02-27
tests: add initial patchbay tests

Frando:
## Description

Adds a first round of tests using our new network simulation framework, [patchbay](https://github.com/n0-computer/patchbay). Patchbay uses Linux network namespaces to create isolated network topologies with routers, NATs, and link impairment. This lets us test actual holepunching and NAT traversal in realistic conditions, without needing any external infrastructure.

The test suite covers establishing a direct path in different scenarios:

- Holepunching: two devices behind destination-independent NATs connect via relay, then upgrade to direct.
- Uplink switching: a device changes its network uplink (v4 and v6) mid-connection and recovers a direct path.
- Interface changes: network interfaces are added or removed while connected
- Link outage recovery: a device goes offline and comes back, direct connection resumes
- Degraded links: increasing levels of impairment (latency, jitter, packet loss) are applied to either side of a connection and holepunching still succeeds

Some tests that iroh doesn't pass yet are included but marked `#[ignore]`. They should be un-ignored as we improve things.

### Running the tests

On Linux, things just work with user namespace support:
```sh
cargo make patchbay
# expands to:
cargo nextest run -p iroh --features qlog --test patchbay --profile patchbay

# run a specific test with logs:
RUST_LOG=trace cargo make patchbay holepunch_simple --nocapture
```

Test output is saved in `./target/testdir-current/patchbay/<test-name>`.  Patchbay currently collects tracing logs for each device, qlog files (if the feature is enabled), and endpoint metrics.
There's also a browser UI for viewing timelines, topologies, and logs. The UI can also be used to compare different test runs.
```sh
cargo install --git https://github.com/n0-computer/patchbay patchbay-cli
patchbay serve --testdir --open
```

On macOS you'll need to run the patchbay tests in a VM or container. The `patchbay` CLI includes a tool to set up a container or QEMU VM for this. 
```sh
# both commands default to the container backend on macOS and native on linux
patchbay test -p iroh --test patchbay
patchbay serve --testdir --open
# use --vm to force VM mode even on linux
```

See the [patchbay docs](https://n0-computer.github.io/patchbay/) and [README](https://github.com/n0-computer/patchbay) for more details.

## Notes and open questions

There's a draft PR with some more tests: #4065, but they need more thought and work to really cover what we want to test.

## Breaking Changes

None.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh/tests/patchbay.rs
+++ b/iroh/tests/patchbay.rs
@@ -52,8 +52,11 @@ fn userns_ctor() {
 // Holepunch tests
 // ---
 
-/// Simple holepunch: Two devices behind destination-independent NATs,
-/// establish via relay, upgrade to direct.
+/// Two devices behind destination-independent NATs holepunch a direct connection.
+///
+/// Both devices connect through a relay first, then upgrade to a direct path.
+/// The client asserts that the connection starts as relayed, then waits for
+/// a direct (IP) path to be selected.
 #[tokio::test]
 #[traced_test]
 async fn holepunch_simple() -> Result {
@@ -81,15 +84,14 @@ async fn holepunch_simple() -> Result {
     Ok(())
 }
 
-/// Tests that changing the uplink of an interface works (i.e. switching wifis).
+/// Switches the client's IPv4 uplink to a different NAT mid-connection.
 ///
-/// In this test, the client device switches the uplink of its network interface.
-/// We then observe a change in the selected path's remote addr on the server side.
-/// How this change carries through in iroh is an implementation detail we don't care for here
-/// (whether it's a new path or the same path) but the server must observe the client's
-/// new address as the selected path.
+/// The client starts behind `nat2`, holepunches a direct path, then replugs
+/// its interface to `nat3`. The server waits until a direct path with a new
+/// remote address is selected. We verify with a ping that the new path works.
 ///
-/// The test currently fails, but should pass.
+/// Currently ignored because iroh does not yet recover reliably from an
+/// uplink switch.
 #[tokio::test]
 #[traced_test]
 #[ignore = "known to still fail"]
@@ -154,11 +156,13 @@ async fn switch_uplink_v4() -> Result {
     Ok(())
 }
 
-/// Tests that changing the uplink from IPv4 to IPv6 works.
+/// Switches the client's uplink from an IPv4 NAT to an IPv6-only ISP network.
 ///
-/// Similar to [`switch_uplink_v4`] but switches to an IPv6 only network.
+/// Similar to [`switch_uplink_v4`], but the client replugs from a Home NAT
+/// to an IPv6-only ISP router. The server waits for the selected path to
+/// switch from an IPv4 to an IPv6 remote address.
 ///
-/// The test currently fails in ~50% of runs, but should pass reliably.
+/// Currently ignored because this fails in roughly half of runs.
 #[tokio::test]
 #[traced_test]
 #[ignore = "known to still be flaky"]
@@ -197,6 +201,8 @@ async fn switch_uplink_v6() -> Result {
                 .context("did not become direct")?;
             info!(addr=?first.remote_addr(), "connection became direct, waiting for path change");
 
+            ping_accept(&conn, timeout).await.context("ping_accept 1")?;
+
             // Now wait until the direct path changes, which happens after the other endpoint
             // changes its uplink. We check is_ip() explicitly to avoid triggering on a
             // transient relay fallback during the network switch.
@@ -209,7 +215,7 @@ async fn switch_uplink_v6() -> Result {
                 .context("did not switch paths to v6")?;
             info!(addr=?second.remote_addr(), "connection changed path, wait for ping");
 
-            ping_accept(&conn, timeout).await?;
+            ping_accept(&conn, timeout).await.context("ping_accept 2")?;
             info!("ping done");
             conn.closed().await;
             Ok(())
@@ -221,19 +227,15 @@ async fn switch_uplink_v6() -> Result {
             // Wait for conn to become direct.
             paths.wait_ip(timeout).await.context("become direct")?;
 
-            // Wait a little more and then switch wifis.
-            tokio::time::sleep(Duration::from_secs(1)).await;
+            ping_open(&conn, timeout).await.context("ping_open 1")?;
+
             info!("switch IP uplink");
             dev.replug_iface("eth0", mobile.id()).await?;
 
             // We don't assert any path changes here, because the remote stays identical,
             // and PathInfo does not contain info on local addrs. Instead, the remote
             // only accepts our ping after the path changed.
-            info!("send ping");
-            ping_open(&conn, timeout)
-                .await
-                .context("failed at ping_open")?;
-            info!("ping done");
+            ping_open(&conn, timeout).await.context("ping_open 2")?;
             Ok(())
         })
         .run()
@@ -242,13 +244,13 @@ async fn switch_uplink_v6() -> Result {
     Ok(())
 }
 
-/// Test that switching to a faster link works.
+/// Adds a faster LAN interface and verifies the path becomes selected.
 ///
-/// Two devices, connected initially over holepunched NAT. Then mid connection
-/// device 2 plugs a cable into device 1's router, i.e. they now have a LAN
-/// connection.
-///
-/// Verify we switch to the LAN connection.
+/// The server sits on `nat1`. The client starts on `nat2` with a 4G-impaired
+/// link and has a second interface `eth1` connected to `nat1` (a LAN path),
+/// but `eth1` starts down. After holepunching over the impaired link, the test
+/// brings `eth1` up and waits for the selected path to change to the new,
+/// faster LAN address. A ping verifies the new path works.
 #[tokio::test]
 #[traced_test]
 async fn change_ifaces() -> Result {
@@ -256,7 +258,7 @@ async fn change_ifaces() -> Result {
     let nat1 = lab.add_router("nat1").nat(Nat::Home).build().await?;
     let nat2 = lab.add_router("nat2").nat(Nat::Home).build().await?;
 
-    // dev2 has two uplinks (wifi=Mobile3G on eth0, LAN on eth1). eth1 starts down.
+    // Client has two uplinks (eth0=4G via nat2, eth1=LAN via nat1). eth1 starts down.
     let server = lab
         .add_device("server")
         .iface("eth0", nat1.id())
@@ -276,9 +278,7 @@ async fn change_ifaces() -> Result {
     let timeout = Duration::from_secs(10);
     Pair::new(relay_map)
         .server(server, async move |_dev, _ep, conn| {
-            ping_accept(&conn, timeout)
-                .await
-                .context("failed at ping_accept")?;
+            ping_accept(&conn, timeout).await.context("ping_accept")?;
             conn.closed().await;
             Ok(())
         })
@@ -307,9 +307,7 @@ async fn change_ifaces() -> Result {
                 .context("did not switch paths")?;
             info!(addr=?next.remote_addr(), "new direct path established");
 
-            ping_open(&conn, timeout)
-                .await
-                .context("failed at ping_open")?;
+            ping_open(&conn, timeout).await.context("ping_open")?;
             Ok(())
         })
         .run()
@@ -318,10 +316,11 @@ async fn change_ifaces() -> Result {
     Ok(())
 }
 
-/// Brief link outage: after holepunching succeeds, the link goes down for 2
-/// seconds and comes back up. The connection should recover — either by
-/// falling back to relay during the outage or by re-establishing the direct
-/// path after recovery.
+/// Takes the client's link down for five seconds after holepunching, then brings it back.
+///
+/// After recovery, the test verifies that we can ping (via relay fallback or
+/// a re-established direct path), and then waits for a direct path to be
+/// selected again.
 #[tokio::test]
 #[traced_test]
 async fn link_outage_recovery() -> Result {
@@ -333,8 +332,8 @@ async fn link_outage_recovery() -> Result {
     let timeout = Duration::from_secs(15);
     Pair::new(relay_map)
         .server(server, async move |_dev, _ep, conn| {
-            ping_accept(&conn, timeout).await.context("ping 1")?;
-            ping_accept(&conn, timeout).await.context("ping 2")?;
+            ping_accept(&conn, timeout).await.context("ping_accept 1")?;
+            ping_accept(&conn, timeout).await.context("ping_accept 2")?;
             conn.closed().await;
             Ok(())
         })
@@ -349,11 +348,11 @@ async fn link_outage_recovery() -> Result {
             dev.link_up("eth0").await?;
             info!("link restored, waiting for recovery");
 
-            // After link recovery, we should be able to ping — via relay
+            // After link recovery, we should be able to ping, either via relay
             // fallback or re-established direct path.
             ping_open(&conn, Duration::from_secs(30))
                 .await
-                .context("ping after link recovery")?;
+                .context("ping_open after link_up")?;
             info!("connection recovered after link outage");
 
             // Eventually the direct path should come back.
@@ -361,7 +360,9 @@ async fn link_outage_recovery() -> Result {
                 .wait_ip(Duration::from_secs(30))
                 .await
                 .context("did not re-establish direct path")?;
-            ping_open(&conn, timeout).await.context("ping on direct")?;
+            ping_open(&conn, timeout)
+                .await
+                .context("ping_open after direct")?;
             Ok(())
         })
         .run()
@@ -374,11 +375,12 @@ async fn link_outage_recovery() -> Result {
 // Degradation ladder: find where holepunching breaks under worsening conditions
 // ---
 
-/// Increasingly degraded link on one side, clean link on the other.
-/// Each level adds more latency, loss, and reordering. The test runs each level
-/// twice: once with the impaired side accepting, once connecting.
+/// Increasingly degraded link conditions applied to one side of the connection.
+///
+/// Each level adds more latency, loss, and reordering. The levels are tested
+/// individually for both server-side and client-side impairment.
 const DEGRADE_LEVELS: &[LinkLimits] = &[
-    // 0: mild — good wifi
+    // 0: mild - good wifi
     LinkLimits {
         latency_ms: 10,
         jitter_ms: 5,
@@ -388,7 +390,7 @@ const DEGRADE_LEVELS: &[LinkLimits] = &[
         duplicate_pct: 0.0,
         corrupt_pct: 0.0,
     },
-    // 1: poor — bad wifi or 3G
+    // 1: poor - bad wifi or 3G
     LinkLimits {
         latency_ms: 100,
         jitter_ms: 30,
@@ -398,7 +400,7 @@ const DEGRADE_LEVELS: &[LinkLimits] = &[
         duplicate_pct: 0.0,
         corrupt_pct: 0.0,
     },
-    // 2: bad — congested 3G
+    // 2: bad - congested 3G
     LinkLimits {
         latency_ms: 200,
         jitter_ms: 60,
@@ -408,7 +410,7 @@ const DEGRADE_LEVELS: &[LinkLimits] = &[
         duplicate_pct: 0.0,
         corrupt_pct: 0.0,
     },
-    // 3: terrible — barely usable
+    // 3: terrible - barely usable
     LinkLimits {
         latency_ms: 300,
         jitter_ms: 80,
@@ -418,7 +420,7 @@ const DEGRADE_LEVELS: &[LinkLimits] = &[
         duplicate_pct: 0.0,
         corrupt_pct: 0.0,
     },
-    // 4: extreme — GEO satellite with heavy loss
+    // 4: extreme - GEO satellite with heavy loss
     LinkLimits {
         latency_ms: 500,
         jitter_ms: 100,
@@ -428,7 +430,7 @@ const DEGRADE_LEVELS: &[LinkLimits] = &[
         duplicate_pct: 0.0,
         corrupt_pct: 0.0,
     },
-    // 6: absurd — stress test
+    // 5: absurd - stress test
     LinkLimits {
         latency_ms: 800,
         jitter_ms: 200,
@@ -440,8 +442,11 @@ const DEGRADE_LEVELS: &[LinkLimits] = &[
     },
 ];
 
-/// Run a single degradation level: create devices with the given impairment,
-/// try to holepunch and ping, return Ok if successful.
+/// Runs a single degradation level.
+///
+/// Creates two devices behind Home NATs, applies the given [`LinkLimits`] to
+/// `impaired_side`, then attempts to holepunch and ping. Returns the
+/// [`TestGuard`] on success so the caller can mark it as passed.
 async fn run_degrade_level(impaired_side: Side, level: usize) -> Result<TestGuard> {
     let (lab, relay_map, _relay_guard, guard) = lab_with_relay(testdir!()).await?;
     let nat1 = lab.add_router("nat1").nat(Nat::Home).build().await?;
@@ -491,7 +496,7 @@ async fn run_degrade_level(impaired_side: Side, level: usize) -> Result<TestGuar
 
     match &result {
         Ok(()) => tracing::event!(
-            target: "iroh::_events::test_ladder_pass",
+            target: "test::_events::ladder_pass",
             tracing::Level::INFO,
             level,
             latency_ms = limits.latency_ms,
@@ -501,7 +506,7 @@ async fn run_degrade_level(impaired_side: Side, level: usize) -> Result<TestGuar
             "PASSED",
         ),
         Err(err) => tracing::event!(
-            target: "iroh::_events::test_ladder_fail",
+            target: "test::_events::ladder_fail",
             tracing::Level::WARN,
             level,
             latency_ms = limits.latency_ms,
--- a/iroh/tests/patchbay/util.rs
+++ b/iroh/tests/patchbay/util.rs
@@ -10,7 +10,7 @@ use n0_error::{Result, StackResultExt, StdResultExt, anyerr, ensure_any};
 use n0_future::{boxed::BoxFuture, task::AbortOnDropHandle};
 use patchbay::{Device, IpSupport, Lab, LabOpts, OutDir, TestGuard};
 use tokio::sync::{Barrier, oneshot};
-use tracing::{Instrument, debug, error, error_span, info};
+use tracing::{Instrument, debug, error, error_span, event, info};
 
 use self::relay::run_relay_server;
 
@@ -127,7 +127,7 @@ impl Pair {
     /// After completion, this will:
     /// - log the result of the run functions
     /// - record the endpoint metrics as a `patchbay::_metrics` tracing event
-    /// - emit an `iroh::_events::test::ok` or `::failed` event for each device
+    /// - emit a `test::_events::pass` or `test::_events::fail` event for each device
     ///
     /// Returns an error if any step or run function failed.
     pub async fn run(mut self) -> Result {
@@ -218,21 +218,17 @@ impl Pair {
 fn log_result_on_device<E: std::fmt::Display + Send + 'static>(dev: &Device, res: Result<(), E>) {
     let _ = dev.run_sync(move || {
         match res {
-            Ok(_) => {
-                tracing::event!(
-                    target: "iroh::_events::test::ok",
-                    tracing::Level::INFO,
-                    msg = %"device ok"
-                );
-            }
-            Err(error) => {
-                tracing::event!(
-                    target: "iroh::_events::test::failed",
-                    tracing::Level::ERROR,
-                    %error,
-                    msg = %"device failed"
-                );
-            }
+            Ok(_) => event!(
+                target: "test::_events::pass",
+                tracing::Level::INFO,
+                msg = %"device passed"
+            ),
+            Err(error) => event!(
+                target: "test::_events::fail",
+                tracing::Level::ERROR,
+                %error,
+                msg = %"device failed"
+            ),
         }
         Ok(())
     });
@@ -255,7 +251,7 @@ pub trait PathWatcherExt {
 
     /// Returns the currently selected path.
     ///
-    /// Panics if no patch is marked as selected.
+    /// Panics if no path is marked as selected.
     fn selected(&mut self) -> PathInfo;
 
     /// Wait until the selected path is a direct (IP) path.


iroh/tests/patchbay.rs
```diff
@@ -0,0 +1,1046 @@
...
+//! ```sh
+//! # On Linux (with user namespace support):
+//! RUSTFLAGS="--cfg patchbay_tests" cargo test --release -p iroh --test patchbay -- --test-threads=1
+//!
+//! # On macOS (via patchbay-vm):
+//! RUSTFLAGS="--cfg patchbay_tests" patchbay-vm test --release -p iroh --test patchbay -- --test-threads=1
+//! ```
+
+// patchbay only runs on linux
+#![cfg(target_os = "linux")]
```
dignifiedquire: this seems to indicate it won't compile on macos with the command above?

iroh/tests/patchbay.rs
```diff
@@ -0,0 +1,1046 @@
...
+                ping_open(&conn, timeout).await?;
+                Ok(())
+            },
+        )
+        .await?;
+    guard.ok();
+    Ok(())
+}
+
+// ---
```
dignifiedquire, line 374: this is becoming a long file, maybe split it up into multiple test files?

iroh/tests/patchbay.rs
```diff
@@ -0,0 +1,1046 @@
...
+        }
+    }
+    Ok((last_pass, guard))
+}
+
+/// Impaired side is the accepting (server) peer.
+#[tokio::test]
+#[traced_test]
+#[serial_test::serial]
+async fn degrade_ladder_impaired_server() -> Result {
```
dignifiedquire: instead of this loop, can we not make each level its own?

iroh/tests/patchbay.rs
Frando: good question. the command above will cross-compile to linux, so it should work as expected.

I just merged https://github.com/n0-computer/patchbay/pull/7 so patchbay now uses  `container` on macos which is much faster than qemu.

Would be great if some people on mac could test this out, and see if it works with the instructions in the test module here.

flub:
> cargo install --git https://github.com/n0-computer/patchbay -p patchbay-runner
> patchbay serve --testdir --open

Why the need to install a tool, can you not write a self-contained html file into the testdir?

iroh/Cargo.toml
```diff
@@ -138,6 +141,12 @@ console = { version = "0.16" }
 wasm-tracing = "2.1.0"
 wasm-bindgen-test = "0.3.62"
 
+# patchbay netsim test dependencies (linux only)
+[target.'cfg(target_os = "linux")'.dev-dependencies]
+ctor = "0.6"
+patchbay = { git = "https://github.com/n0-computer/patchbay.git", branch = "main" }
+testdir = "0.9"
```
flub:
```suggestion
testdir = "0.10"
```
for some nextest fixes

iroh/tests/patchbay.rs
```diff
@@ -0,0 +1,1046 @@
...
+use n0_error::{Result, StackResultExt, StdResultExt};
+use n0_tracing_test::traced_test;
+use patchbay::{Firewall, LinkCondition, LinkLimits, Nat, RouterPreset, TestGuard};
+use testdir::testdir;
+use tracing::info;
+
+use self::util::{Pair, PathWatcherExt, lab_with_relay, ping_accept, ping_open};
+
+#[path = "patchbay/util.rs"]
+mod util;
```
flub, line 41: This is a very strange construct. You can anchor it to `self::util` or so, no? But even without this should work?

iroh/tests/patchbay.rs
```diff
@@ -0,0 +1,1046 @@
...
+// NAT type matrix: verify holepunching across different NAT combinations
+// ---
+
+/// One peer behind Home NAT, the other on a public network.
+/// Holepunching should succeed: EIM mapping means the public peer can reach
+/// the NATted peer's mapped port once it learns the address via relay.
+#[tokio::test]
+#[traced_test]
+#[serial_test::serial]
+#[ignore = "stays relayed, holepunch times out (deadline elapsed)"]
```
flub: TIL that `cargo test` now prints the ignore reasons, that's nice! Can't wait for nextest to catch up.

iroh/tests/patchbay.rs
```diff
@@ -0,0 +1,1046 @@
...
+
+// ---
+// Holepunch tests
+// ---
+
+/// Simple holepunch: Two devices behind destination-independent NATs,
+/// establish via relay, upgrade to direct.
+#[tokio::test]
+#[traced_test]
+#[serial_test::serial]
```
flub: we also have a mechanism in `.config/nextest.toml` that makes tests run in serial on nextest. How does this interact? Running this stuff in nextest right now results in a lot of failures, while nextest is our default test runner currently.

I guess from this it means you can't run tests concurrently? This kind of surprises me. Why is it not possible to give each test it's own namespace and let them run concurrently?

iroh/tests/patchbay.rs
```diff
@@ -0,0 +1,1046 @@
...
+        .build()
+        .await?;
+    let dev1 = lab.add_device("dev1").uplink(public.id()).build().await?;
+    let dev2 = lab.add_device("dev2").uplink(home.id()).build().await?;
+    let timeout = Duration::from_secs(10);
+    Pair::new(dev1, dev2, relay_map)
+        .run(
+            async move |_dev, _ep, conn| {
+                let mut paths = conn.paths();
+                assert!(paths.is_relay(), "connection started relayed");
```
flub: Why is this not flaky? This checks to see whether the selected path is a relay path, but this test does not itself control when holepunching starts so I would expect this to be flaky. If the runtime manages to delay this task enough then iroh could already have holepunched.

.github/workflows/patchbay.yml
```diff
@@ -0,0 +1,120 @@
...
+        run: sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0
+        continue-on-error: true
+
+      - uses: actions/checkout@v6
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: mozilla-actions/sccache-action@v0.0.9
+
+      - name: Run patchbay tests
+        id: tests
+        run: cargo test --release -p iroh --test patchbay -- --test-threads=1 --nocapture
```
matheus23: Locally, I don't need `--test-threads=1` anymore.
Should we update this to `cargo nextest` instead?

```suggestion
        run: cargo nextest run --release -p iroh --test patchbay -P patchbay --no-fail-fast
```

Also needs an addition to `nextest.toml`:
```toml
[profile.patchbay]
slow-timeout = { period = "100s", terminate-after = 5 }
```
So it doesn't time out that quickly.

Not sure what to do about `--nocapture` in this case, though.

iroh/tests/patchbay.rs
Frando: I will remove the serial annotations and limit concurrency with nextest instead.

matheus23:
The command in the PR description needs to be updated: The tests are now called `patchbay`, not `netsim`

matheus23:
>     cargo install --git https://github.com/n0-computer/patchbay -p patchbay-runner
>     patchbay serve --testdir --open

This command from the PR description needs updating: It's called `patchbay-cli`, and you need to drop the `-p`, as that's not an allowed flag in `cargo install`.

iroh/tests/patchbay.rs
matheus23: I'd be in favor of doing this construction. Otherwise you need to move `util.rs` into the same directory as `patchbay.rs`, and it'll assume `util.rs` is another test binary, compiling it twice if you run `cargo test`. A little annoying.

iroh/tests/patchbay.rs
matheus23: `serial_test` is removed now, we don't need to run these tests serially. 
The dependency needs to removed from the `Cargo.toml` still, though.

iroh/tests/patchbay/util.rs
```diff
@@ -0,0 +1,363 @@
...
+        f(&self.selected())
+    }
+
+    fn is_ip(&mut self) -> bool {
+        self.match_selected(PathInfo::is_ip)
+    }
+
+    fn is_relay(&mut self) -> bool {
+        self.match_selected(PathInfo::is_relay)
+    }
```
matheus23: This is some very verbose code.
`is_ip` is the same as `self.selected().is_ip()`. No need to go via `match_selected`.
I'd say all of this can be removed.

The `wait_` stuff is useful though I can see that.

iroh/Cargo.toml
```diff
@@ -101,10 +101,12 @@ getrandom = { version = "0.3.2", features = ["wasm_js"] }
 # target-common test/dev dependencies
 [dev-dependencies]
 console_error_panic_hook = "0.1"
+n0-error = { version = "0.1", features = ["anyhow"] }
 postcard = { version = "1.1.1", features = ["use-std"] }
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 rand_chacha = "0.9"
 chrono = "0.4.43"
+serial_test = "3.4.0"
```
matheus23: We don't need this dependency anymore.

iroh/tests/patchbay.rs
```diff
@@ -0,0 +1,999 @@
...
+//! RUSTFLAGS="--cfg patchbay_tests" cargo test --release -p iroh --test patchbay -- --test-threads=1
+//!
+//! # On macOS (via patchbay-vm):
+//! RUSTFLAGS="--cfg patchbay_tests" patchbay-vm test --release -p iroh --test patchbay -- --test-threads=1
+//! ```
+
+// patchbay only runs on linux
+#![cfg(target_os = "linux")]
+// Only compile these tests when the patchbay_tests cfg is enabled.
+#![cfg(patchbay_tests)]
```
matheus23: RUSTFLAGS configs are really cumbersome to use.

I'd propose using nextest profiles instead. I've opened https://github.com/n0-computer/iroh/pull/4059 for this.

Frando: Thanks, I had the same idea just now ;) merged your changes into my changes and pushed.

iroh/Cargo.toml
Frando: Removed

iroh/tests/patchbay/util.rs
Frando: Yeah there's some leftovers from a previous iteration. Will clean up.

iroh/Cargo.toml
Frando: Thanks, this is much better with nextest now.

iroh/tests/patchbay.rs
Frando: Removed the depencency too.

iroh/tests/patchbay.rs
Frando: So the cfg linux is definitely correct, patchbay only compiles on linux. The usage via `patchbay` cli on macos is unaffacted.

iroh/tests/patchbay.rs
Frando: Did this.

.github/workflows/patchbay.yml
Frando: The patchbay profile was added.

.config/nextest.toml
```diff
@@ -15,4 +15,12 @@ test-group = 'run-in-isolation'
 threads-required = 32
 
 [profile.default]
+default-filter = 'not binary(patchbay)'
 slow-timeout = { period = "10s", terminate-after = 3 }
+
+[profile.patchbay]
+fail-fast = false
+retries = 1
```
matheus23, line 23: I'm not sure if we want `retries = 1`.
@dignifiedquire found that it's possible to get the tests to reliably never fail, even in "absurd" network conditions.
It might be a bad idea to enable this.

That said - perhaps we enable it for *this PR*, and once we've fixed issues, we remove the retries again.
Soooo let's file an issue, add that to a milestone and then make sure to remove it once stuff is stable?

Frando: sgtm!

Frando: Added an issue: https://github.com/n0-computer/iroh/issues/4061

Frando:
Updated this to no longer use a git dependency (I just released patchbay 0.2.0)

iroh/tests/patchbay.rs
Frando: Right, this could be flaky if the runtime stalls this longer than at least one RTT. I've never seen it flake though. Should we leave it in until it flakes? So far it served me as a useful assert that the connection didn't start on IP, because if it did, a check for holepunching would be worthless.

Frando:
> Why the need to install a tool, can you not write a self-contained html file into the testdir?

It would be possible, but when you double-click a html file it cannot `fetch` anything, so *everything* would have to be included in the HTML. Right now with `patchbay serve` logs and events are fetched on-demand.

I could build a version that allows to `http-serve ./taret/tesdir-current` and that would work (it would fetch things on demand) but then this would need more work and not sure if its any better than with `patchbay serve`.

iroh/tests/patchbay.rs
Arqu: +1

Frando:
> One thing I'd like to know more about is how exactly the link degradation applies.

I digged into this, and actually discovered a bug in patchbay. The link conditioning was only applied on the device-to-router egress path, not on the ingress path. This is likely why the numbers didn't match up.

I have a PR for patchbay do fix this: https://github.com/n0-computer/patchbay/pull/12

I'll update this PR to use it for a round of testing, then merge and release patchbay, then update this PR back to the released version.

matheus23:
> > One thing I'd like to know more about is how exactly the link degradation applies.
> 
> I digged into this, and actually discovered a bug in patchbay. The link conditioning was only applied on the device-to-router egress path, not on the ingress path. This is likely why the numbers didn't match up.

Yeah - this is what claude kept telling me, too, but I don't quite understand how that would explain the numbers. Any slowdown on the device-to-router egress path would put a delay on the whole round-trip, so how come a path can have an RTT lower than that?

Frando:
>  Any slowdown on the device-to-router egress path would put a delay on the whole round-trip, so how come a path can have an RTT lower than that?

I don't really know. There's tests in patchbay that confirm the numbers with some variations, they seem to pass reliably. I'll try to test more myself.

Frando:
I removed the "Publish to patchbay server" part for now. We don't want to hold this up on deploying the service, and the utility is debatable because we can just rerun locally if it fails.

So from my side this is ready for merge now. We can always iterate and add more tests later.

iroh/tests/patchbay.rs
```diff
@@ -0,0 +1,1075 @@
...
+                Ok(())
+            },
+        )
+        .await?;
+    guard.ok();
+    Ok(())
+}
+
+/// Both peers behind CGNAT (EIM+EIF). The most permissive real-world NAT.
+/// Holepunching should succeed easily since filtering is endpoint-independent.
```
matheus23: This claims that CGNAT is the most permissive NAT type?

iroh/tests/patchbay.rs
```diff
@@ -0,0 +1,1075 @@
...
+            },
+        )
+        .await?;
+    guard.ok();
+    Ok(())
+}
+
+/// Both peers behind FullCone NAT (EIM+EIF with hairpin). The most permissive
+/// NAT type — any external host can send to the mapped port. Holepunching
+/// always succeeds on the first try.
```
matheus23: And then this repeats the claim, but this time about FullCone NAT?

iroh/tests/patchbay.rs
```diff
@@ -0,0 +1,1075 @@
...
+            async move |dev, _ep, conn| {
+                let mut paths = conn.paths();
+                paths.wait_ip(timeout).await.context("initial holepunch")?;
+                info!("holepunched, now killing link for 2s");
+
+                // Take the link down.
+                dev.link_down("eth0").await?;
+                tokio::time::sleep(Duration::from_secs(2)).await;
+                dev.link_up("eth0").await?;
+                info!("link restored, waiting for recovery");
```
matheus23: Might be interesting to increase the time the link is down here - perhaps on the order of 15s or so - so that it's above the path idle timeout.

But I'm just musing. This is great already - we can add tests later.

iroh/tests/patchbay.rs
```diff
@@ -0,0 +1,1075 @@
...
+            },
+        )
+        .await?;
+    guard.ok();
+    Ok(())
+}
+
+/// Hotel WiFi: captive-portal firewall allows all outbound TCP but only UDP
+/// port 53 (DNS). Similar to corporate firewall but less restrictive on TCP.
+/// Relay via HTTPS should work, holepunching should not.
```
matheus23: Looking at the differences between `RouterPreset::Hotel` and `RouterPreset::Corporate`, the main difference seems to be that the hotel one is IPv4-only, whereas corporate is IPv4 + IPv6, and corporate only allows port 80/443, whereas the hotel one allows any port.

Tbqh, I don't think any of these things make much of a difference in terms of testing for iroh. In most cases I'd expect these tests to either both fail or none of them to fail, except maybe for some edge cases regarding IPv6.

Regardless, perhaps it's worth either adding more information to the test description, or linking the docs via something like "See also [`RouterPreset::Hotel`] and [`RouterPreset::Corporate`] for differences".

matheus23 (approved):
Just nits

.github/workflows/ci.yml
```diff
@@ -31,6 +31,8 @@ jobs:
     if: "github.event_name != 'pull_request' || ! contains(github.event.pull_request.labels.*.name, 'flaky-test')"
     timeout-minutes: 30
     runs-on: [self-hosted, linux, X64]
+    env:
+      RUSTFLAGS: "-Dwarnings --cfg skip_patchbay"
```
flub, line 35: is this still needed?

.github/workflows/patchbay.yml
```diff
@@ -0,0 +1,49 @@
...
+  SCCACHE_CACHE_SIZE: "10G"
+  IROH_FORCE_STAGING_RELAYS: "1"
+  NEXTEST_VERSION: "0.9.132"
+
+jobs:
+  patchbay_tests:
+    name: Patchbay Tests
+    permissions:
+      contents: read
+      pull-requests: write
```
flub: Does it currently need this? It seems weird to give permissions that are not needed.

.github/workflows/tests.yaml
```diff
@@ -25,7 +25,7 @@ env:
   SCCACHE_CACHE_SIZE: "10G"
   CRATES_LIST: "iroh,iroh-bench,iroh-dns-server,iroh-relay"
   IROH_FORCE_STAGING_RELAYS: "1"
-  NEXTEST_VERSION: "0.9.80"
+  NEXTEST_VERSION: "0.9.132"
```
flub: Are you bumping this for fun or for a good reason? We had flaky behaviour with nextest at some point and that is why we pinned it. It might be fixed now, in which case I'm all for bumping it. But it would be better to do this in a separate PR if possible.

.gitignore
```diff
@@ -1,3 +1,4 @@
 /target
 /logs
 iroh.config.toml
+/.patchbay
```
flub, line 4: what files do get stored here?

Cargo.toml
```diff
@@ -36,7 +36,7 @@ missing_debug_implementations = "warn"
 # do.  To enable for a crate set `#![cfg_attr(iroh_docsrs,
 # feature(doc_cfg))]` in the crate.
 # We also have our own `iroh_loom` cfg to enable tokio-rs/loom testing.
-unexpected_cfgs = { level = "warn", check-cfg = ["cfg(iroh_docsrs)", "cfg(iroh_loom)"] }
+unexpected_cfgs = { level = "warn", check-cfg = ["cfg(iroh_docsrs)", "cfg(iroh_loom)", "cfg(skip_patchbay)"] }
```
flub, line 39: this is also no longer used?

Makefile.toml
```diff
@@ -26,3 +26,8 @@ args = [
      "--config",
      "imports_granularity=Crate,group_imports=StdExternalCrate,reorder_imports=true,format_code_in_doc_comments=true",
 ]
+
+[tasks.patchbay]
+workspace = false
+command = "cargo"
+args = ["nextest", "run", "--profile", "patchbay"]
```
flub: :heart:

iroh/tests/patchbay/util.rs
```diff
@@ -0,0 +1,359 @@
...
+        let task1 = self.dev1.spawn(move |dev| {
+            async move {
+                let endpoint = endpoint_builder(&dev, relay_map2).bind().await?;
+                endpoint.online().await;
+                addr_tx.send(addr_relay_only(endpoint.addr())).unwrap();
+                let conn = endpoint.accept().await.unwrap().accept().anyerr()?.await?;
+                watch_selected_path(&conn);
+                peer1(dev.clone(), endpoint.clone(), conn.clone()).await?;
+                conn.closed().await;
+                endpoint.close().await;
```
flub: This (and the same for the peer) should have a few more `info!()` messages. I would add one least one before and after the callback.

iroh/tests/patchbay/util.rs
```diff
@@ -0,0 +1,359 @@
...
+                Err(err) => Err(anyerr!(err, "device {} panicked", dev.name())),
+                Ok(Err(err)) => Err(anyerr!(err, "device {} failed", dev.name())),
+                Ok(Ok(())) => Ok(()),
+            };
+            let res_str = res.as_ref().map_err(|err| format!("{err:#}")).cloned();
+            dev.run_sync(move || {
+                match res_str {
+                    Ok(()) => {
+                        tracing::event!(
+                            target: "iroh::_events::test_ok",
```
flub: Heh, I this is a bit abusing the events... it should be kind of reserved for iroh-the-library. I'm not yet sure how I feel about this.

 Can we make it all in the `iroh::_events::test::` namespace at least?

iroh/tests/patchbay/util.rs
```diff
@@ -0,0 +1,359 @@
...
+                }
+                self.updated().await?;
+            }
+        })
+        .await
+        .anyerr()?
+    }
+}
+
+pub async fn ping_open(conn: &Connection, timeout: Duration) -> Result {
```
flub, line 297: A succinct doc-comment would help reading tests with the help of rust-analyzer annotations. Much like you have them for wait_ip and wait_relay.

Also for accept.

iroh/tests/patchbay.rs
```diff
@@ -0,0 +1,1075 @@
...
+//!
+//! These tests require Linux with user namespace support. On non-Linux systems, you can use
+//! the `patchbay` CLI to get a Linux container or VM with the required capabilities.
+//! See patchbay docs for details.
+//!
+//! To run:
+//!
+//! ```sh
+//! # On Linux (with user namespace support):
+//! cargo nextest run -p iroh --test patchbay -P patchbay
```
flub: let's also mention cargo make here

iroh/tests/patchbay.rs
flub: huh, integration tests are build weirdly. Had to try this out to understand it.

iroh/tests/patchbay.rs
```diff
@@ -0,0 +1,1075 @@
...
+        .await?;
+    guard.ok();
+    Ok(())
+}
+
+/// Tests that changing the uplink of an interface works (i.e. switching wifis).
+///
+/// For this we observe a change in the selected path's remote addr on the *other* side.
+/// Whether the side that changes interfaces opens a new path or does an RFC9000-style migration
+/// is an implementation detail which we won't test for.
```
flub: Not sure this makes sense, because if it were to do an RFC9000-style migration the other side would not see a change in the selected path either I think.

iroh/tests/patchbay/util.rs
```diff
@@ -0,0 +1,359 @@
...
+    /// [`EndpointAddr`], and then the supplied functions are invoked, passing
+    /// the device, endpoint, and connection to user code.
+    ///
+    /// After a future complete, `peer1` awaits the connection to be closed,
+    /// whereas `peer2` closes the connection.
+    ///
+    /// Afterwards, both endpoints are closed and metrics are recorded through
+    /// [`Device::record_iroh_metrics`]. Will also emit a debug log with target
+    /// `patchbay::_events` with the result of the user-supplied work functions.
+    pub async fn run<F1, Fut1, F2, Fut2>(self, peer1: F1, peer2: F2) -> Result
```
flub: One issue that I have with this signature is that it is really hard in a test to tell which one of the two initiates the connection and which one accepts it (I'm in a review, don't have rust-analyzer here). It is kind of made worse by being disconnected from the new signature as well where the same order matters.

iroh might be p2p, but QUIC has definite server and client roles. And understanding those is important in tests.

I don't have a clever solution just yet. And also don't want to block on just this, but maybe someone else has a good idea.

iroh/tests/patchbay.rs
```diff
@@ -0,0 +1,1075 @@
...
+// NAT type matrix: verify holepunching across different NAT combinations
+// ---
+
+/// One peer behind Home NAT, the other on a public network.
+/// Holepunching should succeed: EIM mapping means the public peer can reach
+/// the NATted peer's mapped port once it learns the address via relay.
+#[tokio::test]
+#[traced_test]
+#[ignore = "stays relayed, holepunch times out (deadline elapsed)"]
+async fn holepunch_home_nat_one_side() -> Result {
```
flub: it is strange this doens't work. what mechanism did we figure to address these ignored tests?

flub:
Oh, another thing I found was that investigating any failure requires me to set RUST_LOG=trace. Why not enable that by default for the logfiles created? It uses some more disk space, but I don't think that's a big deal. Any other downside?

Cargo.toml
Frando: It is, see above. It's used in the cross tests to skip running the patchbay tests.

iroh/tests/patchbay/util.rs
Frando: Why abusing?
Can change the namespace, sure. It's really helpful to have this though because this makes this show up in the patchbay timeline view in the web ui. I wrote that one such that it shows all tracing lines with `_events`  in the target. So sure, I can change it to `iroh::_events::test`.

Frando:
> there are too many tests to review, i'm not sure how to tackle that. 

I could just remove some, and do another PR then to add more. Let me do that tomorrow.

iroh/tests/patchbay.rs
Frando: No clear mechanism yet. First step: Let's run them locally with `--ignored` and if someone has an ignored one passing, then do a PR to remove the ignore line and we see what CI says.

iroh/tests/patchbay/util.rs
Frando: `peer1` is server, `peer2` is client. I'll rename the args. And think a bit if I find a better way to express this. I do quite like how simple it is to use this in the tests though.

.github/workflows/ci.yml
matheus23: Yes, this skips patchbay tests for the cross-compilation job. This is needed. You might be mistaking this cfg with the previous iteration that had the `patchbay_test` cfg. This one is different.

iroh/tests/patchbay.rs
Frando: This test is among the ones I removed for the inital merge. I will review the tests and docs more and mark https://github.com/n0-computer/iroh/pull/4065 as ready for review once one.

iroh/tests/patchbay.rs
Frando: Changed to 5s. Can adapt further if needed, or we add more tests with that.

iroh/tests/patchbay.rs
Frando: This test is among the ones I removed for the inital merge. I will review the tests and docs more and mark https://github.com/n0-computer/iroh/pull/4065 as ready for review once one.

iroh/tests/patchbay.rs
Frando: :thumbsup: This test is among the ones I removed for the inital merge. I will review the tests and docs more and mark https://github.com/n0-computer/iroh/pull/4065 as ready for review once one.

.github/workflows/tests.yaml
Frando: At one point I used an option that was only supported in the later version. I don't think this is the case though, so will downgrade again.

.gitignore
Frando: If you use the `patchbay` cli this is where test results will be stored.

iroh/tests/patchbay/util.rs
Frando: I added more info messages.

iroh/tests/patchbay/util.rs
Frando: Added comments.

iroh/tests/patchbay.rs
Frando: Fixed

iroh/tests/patchbay.rs
Frando: Reworded the comment

iroh/tests/patchbay/util.rs
Frando: I rewrote this to have clearly marked `server` and `client` functions.

Frando:
I moved a number of tests to https://github.com/n0-computer/iroh/pull/4065 to make it easier to review this PR.
Also did a round of cleanups, I think I addressed all feedback.

.github/workflows/patchbay.yml
```diff
@@ -0,0 +1,55 @@
...
+
+concurrency:
+  group: patchbay-${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+env:
+  RUST_BACKTRACE: 1
+  SCCACHE_CACHE_SIZE: "10G"
+  IROH_FORCE_STAGING_RELAYS: "1"
+  NEXTEST_VERSION: "0.9.132"
```
matheus23: The nextest version we have here is still 0.9.132, perhaps we should align the versions to match, if possible (esp. since we haven't verified that 0.9.132 doesn't have the same bug that versions after 0.9.80 had).

matheus23 (approved):

iroh/tests/patchbay.rs
Frando: Removed several tests and moved them to https://github.com/n0-computer/iroh/pull/4065 (still a draft) so let's do this once we add those back.

.github/workflows/patchbay.yml
Frando: changed to 0.9.80

iroh/tests/patchbay/util.rs
flub: Abusing because these are not events emitted by iroh. They are emitted by tests. They are only emitted in the same way because patchbay already had code that was looking for these events. I don't think that's the cleanest design. It's a bit like emitting events in another libraries logging targets.

But also I'm fine with for now containing this into a specific namespace that hopefully won't naturally occur in iroh itself. So let's not spend too long on this.

iroh/tests/patchbay/util.rs
flub: oh, that's rather nice! :heart:

PR #3987, opened 2026-02-28
refactor(iroh): use combinator approach for address filtering

Frando:
## Description

This changes the address filtering API added in #3960: Instead of having each service deal with the filtering, we layer this as a combinator on top. `AddressLookupBuilder` now has a `with_addr_filter` method with a default impl, and returns a `FilteredBuilder` which again impls `AddressLookupBuilder` itself. Through this, this all works transparently as currently, but the individual services don't need to care about the filtering at all.

## Notes & open questions

Are there any downsides? This would save people who impl address lookup services from having to deal with the filtering themselves.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


ramfox (commented):
Yes, I think this is a better way to think about/handle adding a filter.

I had half a thought in my head that said we should allow services to disregard user provided filters if they needed to, but I actually think that's wrong. If the user supplied a filter that filters out all of the addresses that can actually be published to a service, I'd rather assume they did it on purpose and not publish things they explicitly asked to keep private.

dignifiedquire (approved):

PR #3988, opened 2026-03-02
feat(iroh): proper span for endpoints

Frando:
## Description

* Adds a `endpoint{id=id_short}` info-level span when binding the endpoint
* Use this span as parent span for RemoteStateActor tasks and the main socket actor
* Alter a few logging statements here and there that stood out while taking a good look at debug logs for the echo example. There's still more work here of course.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -26,7 +26,7 @@ use n0_watcher::Watcher;
 use netdev::ipnet::{Ipv4Net, Ipv6Net};
 use pin_project::pin_project;
 use tokio_util::sync::WaitForCancellationFutureOwned;
-use tracing::{debug, instrument, trace, warn};
+use tracing::{Instrument, Span, debug, info_span, instrument, warn};
 use url::Url;
 
 use self::hooks::EndpointHooksList;
@@ -190,6 +190,9 @@ impl Builder {
             .secret_key
             .unwrap_or_else(move || SecretKey::generate(&mut rng));
 
+        let span = info_span!("endpoint", id = %secret_key.public().fmt_short());
+        let _guard = span.enter();
+
         let static_config = StaticConfig {
             transport_config: self.transport_config.clone(),
             tls_config: tls::TlsConfig::new(secret_key.clone(), self.max_tls_tickets),
@@ -222,9 +225,14 @@ impl Builder {
             static_config,
         };
 
-        let inner = socket::Socket::spawn(sock_opts).await?;
-        trace!("created socket");
-        debug!(version = env!("CARGO_PKG_VERSION"), "iroh Endpoint created");
+        let inner = socket::EndpointInner::bind(sock_opts)
+            .instrument(Span::current())
+            .await?;
+        debug!(
+            id = %inner.static_config.tls_config.secret_key.public(),
+            iroh_version = %env!("CARGO_PKG_VERSION"),
+            "iroh endpoint created"
+        );
 
         let ep = Endpoint {
             inner: Arc::new(inner),
@@ -860,7 +868,7 @@ impl Endpoint {
     #[instrument(name = "connect", skip_all, fields(
         me = %self.id().fmt_short(),
         remote = tracing::field::Empty,
-        alpn = String::from_utf8_lossy(alpn).to_string(),
+        alpn = %String::from_utf8_lossy(alpn).to_string(),
     ))]
     pub async fn connect_with_opts(
         &self,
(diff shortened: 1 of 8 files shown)


iroh/src/net_report/reportgen.rs
```diff
@@ -528,7 +528,9 @@ impl Probe {
             }
             #[cfg(not(wasm_browser))]
             Probe::QadIpv4 | Probe::QadIpv6 => unreachable!("must not be used"),
-        }
+        };
+        debug!(?report, "probe finished");
```
Frando, line 532: we're logging "starting probe" above, so I think a finished log at same level makes sense, otherwise the span is kinda pointless.

iroh/src/socket/transports.rs
```diff
@@ -827,17 +836,25 @@ impl quinn::UdpSender for Sender {
             .sender
             .poll_send(cx, &transport_addr, quinn_transmit.src_ip, &transmit)
         {
-            Poll::Ready(Ok(())) => Poll::Ready(Ok(())),
+            Poll::Ready(Ok(())) => {
+                trace!(
```
Frando, line 840: we log received datagrams at trace level, but didn't log sent transmits at all. this adds a trace log to mirror the receive side.

iroh/src/socket/transports.rs
```diff
@@ -430,6 +430,15 @@ pub(crate) struct Transmit<'a> {
     pub(crate) segment_size: Option<usize>,
 }
 
+impl<'a> Transmit<'a> {
+    pub(crate) fn datagram_count(&self) -> usize {
```
Frando: used in the new trace log below

iroh/src/socket.rs
```diff
@@ -307,14 +309,11 @@ pub(crate) struct Socket {
     /// Metrics
     pub(crate) metrics: EndpointMetrics,
     pub(crate) hooks: EndpointHooksList,
+    /// Tracing span for this endpoint.
+    pub(crate) span: Span,
 }
 
 impl Socket {
-    /// Creates a [`Socket`] listening.
-    pub(crate) async fn spawn(opts: Options) -> Result<EndpointInner, BindError> {
```
Frando, line 314: not strictly related to this PR, but `Socket::spawn -> EndpointInner` is weird, isn't it. moved it to `impl EndpointInner`

iroh/src/endpoint.rs
```diff
@@ -222,9 +225,14 @@ impl Builder {
...
         };
 
-        let inner = socket::Socket::spawn(sock_opts).await?;
-        trace!("created socket");
-        debug!(version = env!("CARGO_PKG_VERSION"), "iroh Endpoint created");
+        let inner = socket::EndpointInner::bind(sock_opts)
+            .instrument(Span::current())
+            .await?;
+        debug!(
+            id = %inner.static_config.tls_config.secret_key.public(),
```
flub, line 232: How come `id` needs to be added here? Is this a different one than the one from the `endpoint` span you just added?

iroh/src/net_report/reportgen.rs
flub: I'm dubious on this one. Isn't the probe report already logged or something?

iroh/src/socket.rs
```diff
@@ -1001,7 +1004,7 @@ impl EndpointInner {
     /// after this call.
     ///
     /// [`Poll::Pending`]: std::task::Poll::Pending
-    #[instrument(skip_all)]
+    #[instrument(skip_all, parent = self.sock.span.clone())]
```
flub, line 1007: Oh, i wonder if this should be on more of the `Endpoint` public API. Anyway, that'd be fore another PR.

iroh/src/socket.rs
flub: Makes sense, I assume this code structure made more sense many refactors ago :)

iroh/src/socket/transports.rs
flub: Doesn't need to be `pub(crate)` though, does it?

flub (approved):
Nice, thanks for picking this up!

iroh/src/endpoint.rs
Frando: Here it's printed in full, not just the first 8 chars from `fmt_short`. I thought it would be good to *somewhere* log the full endpoint id, and this seemed like a fitting place.

iroh/src/socket.rs
Frando: Yeah that's the one that stood out when looking at the logs. We can add more once we spot them.

iroh/src/net_report/reportgen.rs
Frando: We print the final report from all probes, but not the results of individual probes, this we so far only do for QAD probes if I'm not missing something. Changed it so that the "probe started" are trace logs and the "probe finished" ones remain debug.

PR #3992, opened 2026-03-03
feat(iroh): Pluggable crypto backends

matheus23:
## Description

- Introduces a `ring` feature flag that enables turning off the ring dependency
- Introduces a `aws-lc-rs` feature flag that enables using aws-lc-rs as the crypto provider for rustls for everything TLS in iroh
- Adds `iroh::endpoint::Builder::crypto_provider` allowing anyone to configure custom crypto providers beyond ring and aws-lc-rs. This is a required call when neither `ring` nor `aws-lc-rs` are enabled, but will default to either (preferring ring) if the features are enabled.
- Adds `presets::Empty` and `presets::Minimal`. The latter sets the minimal set of required options on the builder, but is only available on the `ring` or `aws-lc-rs` feature flags.
- Removes `Endpoint::empty_builder` in favor of `Endpoint::builder` together with either the `presets::Empty` or `presets::Minimal`.

Depends on https://github.com/n0-computer/noq/pull/496

## Breaking Changes

- Added `iroh::endpoint::Builder::crypto_provider` for configuring the crypto provider. After having built the Endpoint, you can get back the crypto provider via `endpoint.tls_config().crypto_provider()`.
- Removed `Endpoint::empty_builder`. Use `Endpoint::builder(presets::Empty)` or `Endpoint::builder(presets::Minimal)` instead.
- **Behavioral**: If you don't call `iroh::endpoint::Builder::crypto_provider`, you will get an error when building in the call `Builder::bind()`.

## Notes & Open Questions

It's probably worth asking ourselves if we want to change the way we run CI now.
Most of our tests are cfg-ed out when you disable *both* the ring and aws-lc-rs features.
However, maybe we should add another test category that's like `--no-default-features --feature ring/aws-lc-rs`?
We could call it `minimal` (and I guess we'd have to decide to use either ring or aws-lc-rs).


iroh/src/tls/misc.rs
```diff
@@ -0,0 +1,82 @@
+use aes_gcm::{AeadInPlace, Aes256Gcm, KeyInit};
+
+use ctutils::CtEq;
+use quinn_proto::crypto;
+
+pub(crate) struct Blake3Prk(pub(crate) [u8; 32]);
+
+impl Blake3Prk {
+    pub fn new(rng: &mut impl rand::RngCore) -> Self {
```
dignifiedquire: nit: cryptorng? or is this not sensitive?

matheus23: updated

rklaehn:
Probably not for this PR, but we could have another feature where we enable a rust-only crypto provider. It is the only major non-feature change I have to make to get esp32 to work.

See https://github.com/n0-computer/iroh/pull/3999/changes#diff-07f36e33249ac4a7e40c2b4d58f304a80f63acccfc825f27451533a8cbdde72b

rklaehn:
We should also extend the deps test that I added in https://github.com/n0-computer/iroh/pull/4011 to make sure ring and aws-lc-rs are truly gone when running with default-features = false.

matheus23:
> We should also extend the deps test that I added in #4011 to make sure ring and aws-lc-rs are truly gone when running with default-features = false.

Cool check! I added `ring` and `aws-lc-rs` to the list.

iroh/src/address_lookup/memory.rs
```diff
@@ -117,31 +117,37 @@ impl MemoryLookup {
...
-    /// // create an endpoint with the memory lookup address_lookup
-    /// let endpoint = Endpoint::builder()
-    ///     .address_lookup(address_lookup)
-    ///     .bind()
-    ///     .await?;
-    /// # Ok(())
-    /// # }
-    /// ```
+    #[cfg_attr(
+        any(feature = "ring", feature = "aws-lc-rs"), // Endpoint::bind needs a crypto provider
```
Frando: The syntax switch to doc attribute is unfortunate.

Is it possible to put the cfg into the codeblock instead? Or could we enable the ring feature in dev deps by depending on iroh itself?

matheus23: > by depending on iroh itself

We did stuff like this in the past, it was painful. I don't remember the details, but I very much remember the pain :D

> Is it possible to put the cfg into the codeblock instead?

I did try essentially this:
```rs
# #[cfg(any(feature = "ring", feature = "aws-lc-rs"))]
async fn main() { ... }

# #[cfg(not(any(feature = "ring", feature = "aws-lc-rs")))]
# fn main() { }
```
And that didn't work. I have no idea why? I'd appreciate any guidance.

iroh-dns-server/src/http.rs
```diff
@@ -317,6 +317,10 @@ mod tests {
     #[tokio::test]
     #[traced_test]
     async fn test_doh() -> n0_error::Result {
+        rustls::crypto::ring::default_provider()
```
dignifiedquire: this seems problematic, as this installs it for all tests, introducing race conditions?

iroh-dns-server/Cargo.toml
```diff
@@ -67,16 +64,24 @@ n0-tracing-test = "0.3"
 pkarr = { version = "5", features = ["relays", "dht"], default-features = false }
 rand = "0.9.2"
 rand_chacha = "0.9"
-reqwest = { version = "0.12", default-features = false, features = [
-    "rustls-tls", "http2"
+reqwest = { version = "0.13", default-features = false, features = [
```
dignifiedquire, line 67: what about iroh-metrics, if we update reqwest?

iroh-relay/src/quic.rs
```diff
@@ -459,9 +459,12 @@ mod tests {
     /// In this case we don't simulate it via synthetically high RTT, but by dropping
     /// all packets on the server-side for 2 seconds.
     #[tokio::test]
-    // #[traced_test]
+    #[traced_test]
     async fn test_qad_connect_delayed() -> Result {
-        tracing_subscriber::fmt::try_init().ok();
+        rustls::crypto::ring::default_provider()
```
dignifiedquire: same concern as above

iroh/src/socket.rs
```diff
@@ -187,11 +193,16 @@ impl Drop for EndpointInner {
...
 pub(crate) struct StaticConfig {
     pub(crate) tls_config: tls::TlsConfig,
+    #[debug("QuicServerConifg")]
+    pub(crate) server_config: QuicServerConfig,
+    #[debug("QuicClientConfig")]
+    pub(crate) client_config: QuicClientConfig,
+    #[debug("Arc<RustlsTokenKey>")]
+    pub(crate) token_key: Arc<RustlsTokenKey>,
     pub(crate) transport_config: QuicTransportConfig,
-    pub(crate) keylog: bool,
```
dignifiedquire, line 194: why is this not needed anymore?

iroh/Cargo.toml
```diff
@@ -143,9 +142,9 @@ wasm-bindgen-test = "0.3.62"
 cfg_aliases = { version = "0.2.1" }
 
 [features]
-default = ["metrics", "fast-apple-datapath", "portmapper"]
+default = ["metrics", "fast-apple-datapath", "portmapper", "ring"]
 portmapper = ["dep:portmapper"]
-metrics = ["iroh-metrics/metrics", "iroh-relay/metrics", "portmapper?/metrics"]
+metrics = ["iroh-metrics/metrics", "iroh-relay/metrics", "portmapper/metrics"]
```
dignifiedquire: the ? shouldn't be removed

Cargo.toml
```diff
@@ -40,3 +40,8 @@ unexpected_cfgs = { level = "warn", check-cfg = ["cfg(iroh_docsrs)", "cfg(iroh_l
 
 [workspace.lints.clippy]
 unused-async = "warn"
+
+[patch.crates-io]
+noq = { git = "https://github.com/n0-computer/noq", branch = "main" }
```
dignifiedquire, line 45: is this still needed? which prs are missing from the release?

iroh/src/socket.rs
matheus23: We construct the `QuicClientConfig` and `QuicServerConfig` earlier - they already have `keylog` set.
Storing the keylog variable would just be unused code.

iroh-dns-server/Cargo.toml
matheus23: Right. I only tested this with `--no-default-features --features X Y`, but without `metrics` enabled.
So yeah - iroh-metrics would probably also need an update. Will look into this.

iroh-dns-server/src/http.rs
matheus23: Since writing this I've learned how to pass a provider to reqwest by value instead of via the process-default.

It's.. ugly but probably better. Will change.

matheus23: Updated.

Cargo.toml
matheus23: This: https://github.com/n0-computer/noq/pull/496

iroh-dns-server/Cargo.toml
matheus23: I mean, this PR works without upgrading iroh-metrics to reqwest 0.13. It'll just introduce a double-dependency.

I'd say this PR should go forth regardless of the timeline of that upgrade, but nonetheless I've opened this: https://github.com/n0-computer/iroh-metrics/pull/99

rklaehn:
I somehow don't like this whole "if neither feature flag x or y are on, you need to do foo in the builder, or else things will not work". So could we just hide the N0 preset when no default crypto provider is enabled? Then people would have to manually build with presets::Empty.

The downside of this is probably that a lot of tests won't even run with default-features=false, but I still like it more than rules about what you need to do depending on feature flags.

matheus23:
> The downside of this is probably that a lot of tests won't even run with default-features=false, but I still like it more than rules about what you need to do depending on feature flags.

This makes me realize, most test currently can't run without a crypto provider:
> $ cargo nextest run -p iroh --no-default-features --lib --no-fail-fast
> [...]
>      Summary [   3.154s] 93 tests run: 42 passed, 51 failed, 1 skipped
(It's all builder failures about not having a crypto provider)

It's interesting that our CI check doesn't catch this: I suspect workspace feature unification is to blame. *Something* ends up enabling either iroh's `ring` or `aws-lc-rs` feature, even in the "no default features" CI run.

This probably needs fixing.

And I don't think there's a way around this for our tests. I'll probably have to `#[cfg(any(feature = "ring", feature = "aws-lc-rs"))` the iroh tests that fail.

> I somehow don't like this whole "if neither feature flag x or y are on, you need to do foo in the builder, or else things will not work". So could we just hide the N0 preset when no default crypto provider is enabled? Then people would have to manually build with presets::Empty.

I agree that it's not great.
Hiding the preset sounds good. Perhaps we can also expose a `presets::n0(CryptoProvider)` so it's easy to enable a preset with a different crypto provider, which we won't have to hide?
Another thing I was thinking of for solving this was making the endpoint builder have a typestate pattern and failing compilation if `crypto_provider` is not set, but I don't particularly like this option due to its complexity.
Other than that I'm out of ideas.

rklaehn:
> Hiding the preset sounds good. Perhaps we can also expose a `presets::n0(CryptoProvider)` so it's easy to enable a preset with a different crypto provider, which we won't have to hide?

PresetsBuilderFactoryProvider

rklaehn:
> Another thing I was thinking of for solving this was making the endpoint builder have a typestate pattern and failing compilation if `crypto_provider` is not set, but I don't particularly like this option due to its complexity.
> Other than that I'm out of ideas.

Happened to me a few times. First try to "make invalid states unrepresentable" in builders, then scrap it all because the API surface area is just so large.

I think if you just don't have the preset at all if you don't have the features, starting from scratch with .builder(presets::Empty) and having to set everything manually is unsurprising. This is only an issue for a small fraction of our users, and most of them will just copy and paste an example. So I think it's fine.

By the way: isn't the crypto provider a global? So would it even work if you had 2 endpoints with different cryptoproviders? Is the setter on the builder just a fancy decoration for a global, or is it for real?

matheus23:
> By the way: isn't the crypto provider a global? So would it even work if you had 2 endpoints with different cryptoproviders? Is the setter on the builder just a fancy decoration for a global, or is it for real?

It... does have [this annoying API](https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html#method.install_default) to make it globally available, but it doesn't have to be.

For all of iroh's uses of CryptoProvider it is passed in - never using the global.

The only places where I use the global are some edge-cases where constructing an `axum-server` without using the global is a lot more cumbersome than just setting the global (and that's inside some doctests).

Other than that, it's **always passed around**.

So yes, you can have two iroh endpoints with different crypto providers in the same process.

iroh/src/address_lookup/memory.rs
matheus23: I've found other places that use `no_run` in combination with a `#[cfg(...)] { ... }`. I changed it to that.

iroh/Cargo.toml
```diff
@@ -154,6 +153,10 @@ platform-verifier = ["iroh-relay/platform-verifier"]
 qlog = ["noq/qlog"]
 # Use private Apple APIs to send multiple packets in a single syscall.
 fast-apple-datapath = ["noq/fast-apple-datapath"]
+# Use ring as the crypto backend.
+ring = ["noq/ring", "iroh-relay/ring", "webpki/ring"]
```
rklaehn: I am not sure I like the feature flag being just the name of the dep.

Would tls_ring make sense? I mean, what are we using ring for? See https://www.notion.so/number-zero/0ec342e4138c4624a98f19dc09a0df77?p=3225df1306fb80f988faee8d66287087&pm=c for some more thoughts about how to structure feature flags.

iroh/src/tls.rs
```diff
@@ -116,8 +116,25 @@ impl TlsConfig {
...
-            .expect("expected to have a TLS1.3-compatible crypto provider set (hardcoded)")
+        let quic = QuicServerConfig::try_from(crypto)?;
+        Ok(quic)
     }
 }
+
+#[allow(missing_docs)]
+#[n0_error::stack_error(derive, add_meta, from_sources)]
+#[non_exhaustive]
+pub enum TlsConfigError {
```
rklaehn, line 127: Would you also get an error here if the crypto provider misses support for rsa, for https probes etc.?

It should maybe not be overly strict. Anything that means that iroh just plain won't work should produce an error, but if people want to use relays with a simplified cert chain (X2 letsencrypt e.g.) we should probably not fail here, otherwise there would be another thing I would have to patch in the esp32 branch.

iroh/src/endpoint/presets.rs
```diff
@@ -3,32 +3,94 @@
...
+/// with the "ring" or "aws-lc-rs" feature flag.
+///
+/// It uses either [ring] or [aws-lc-rs], depending on which feature is enabled
+/// on iroh (preferring ring if both are enabled).
+///
+/// [ring]: rustls::crypto::ring::default_provider
+/// [aws-lc-rs]: rustls::crypto::aws_lc_rs::default_provider
+#[cfg(any(feature = "ring", feature = "aws-lc-rs"))]
+#[derive(Debug, Copy, Clone, Default)]
+pub struct Minimal;
```
rklaehn, line 59: Didn't we want to call this CryptoOnly or something?

iroh-relay/src/dns.rs
```diff
@@ -124,7 +123,8 @@ impl<E: StackError + 'static> StaggeredError<E> {
 pub struct Builder {
     use_system_defaults: bool,
     nameservers: Vec<(SocketAddr, DnsProtocol)>,
-    tls_client_config: Option<ClientConfig>,
+    #[cfg(any(feature = "ring", feature = "aws-lc-rs"))]
```
rklaehn: Would it be possible to use a convenience alias to DRY all those `#[cfg(any(feature = "ring", feature = "aws-lc-rs"))]`? Not sure, but it seems we already define one for `wasm_browser`, so it would not add an additional build step.

iroh/src/endpoint/presets.rs
rklaehn: My rationale for calling this CryptoOnly would be that I would like to reserve Minimal for a minimal config that does actually work out of the box, e.g. a relay network and a working set of address lookup services that don't bring huge deps, as well as a lightweight dns resolver.

iroh/src/tls.rs
matheus23: This fails if critical dependencies are missing. I don't know exactly what this entails right now, but I don't think RSA is included in this.

I think it's mainly the cipher suite for QUIC initial packets (AES-GCM-256) and *some* key exchange algorithm.

iroh/src/endpoint/presets.rs
matheus23: My preference would be to introduce `CryptoOnly` once it is different from `Minimal`.

Because what people want *today* is more commonly `Minmal`, not `CryptoOnly`. What they care about is a setup that can work but doesn't install any address lookup or relays, not that it only sets the `crypto_provider`.

> I would like to reserve Minimal for a minimal config that does actually work out of the box, e.g. a relay network and a working set of address lookup services that don't bring huge deps, as well as a lightweight dns resolver.

Hm. I think we have a different idea of what "minimal" should mean. IMO what you describe is not minimal, because it installs address lookups and a dns resolver.

Perhaps what you describe I'd rather call "Lightweight", not minimal.

iroh-relay/src/dns.rs
matheus23: Yep totally possible! And I think if I make the feature flags longer (by calling them `tls-ring` and `tls-aws-lc-rs`), then I really want to have this :)

rklaehn: I was also thinking that we might later add an (optional, of course) crypto provider rustls that is just the software only rustls, for out of the box embedded.

(you probably want a hardware specific embedded tls, like wolftls or whatever, but that should be part of an iroh embedded suite, not on main, since it would be tons of different configs1)

iroh/src/endpoint/presets.rs
rklaehn: OK, I am fine with Minimal. Not the end of the world!

iroh/Cargo.toml
matheus23: renamed

rklaehn (approved):

PR #4003, opened 2026-03-09
chore(iroh): Update noq commit, stop exposing `AeadKey`

matheus23:
## Description

Updates to the latest noq `main` branch.

`noq` recently stopped exposing `trait AeadKey`, because the `trait HandshakeTokenKey` doesn't have a function returning `Box<dyn AeadKey>` anymore (instead it is left up to the implementation to use appropriate AEAD constructions with or without an HKDF in front).

(Also fix the iroh unit test suite to allow compiling without the `metrics` feature.)

## Breaking Changes

- Stop exposing `trait AeadKey` and change `trait HandshakeTokenKey` to implement `seal` and `open` on its own, leaving choice of using an HKDF to the user.

### DIFF
--- a/iroh/src/endpoint.rs
+++ b/iroh/src/endpoint.rs
@@ -2182,9 +2182,12 @@ mod tests {
                 }
             }
             info!("Have direct connection");
-            // Validate holepunch metrics.
-            assert_eq!(ep.metrics().socket.num_conns_opened.get(), 1);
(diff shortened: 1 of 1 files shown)


dignifiedquire:
this is a breaking change, no? please document if so properly

matheus23:
> this is a breaking change, no? please document if so properly

There already is a section in the breaking changes list.

PR #4007, opened 2026-03-10
fix(example): always close the endpoint

flub:
## Description

Even if there's an error we want to gracefully close the endpoint.

I also added some extra logging.

## Breaking Changes

n/a

## Notes & open questions

Look, this is the code we're making everyone write now.

I appreciate this is just a quick fix, I did not check out the full structure of the example and try to organise it as nicely as possible. So this probably increases technical debt of this example a bit more.

## Change checklist

- [x] Self-review.

### DIFF
--- a/iroh/examples/transfer.rs
+++ b/iroh/examples/transfer.rs
@@ -744,8 +744,9 @@ async fn fetch(
         duration: start.elapsed(),
     });
 
-    // stats are collected by the paths watcher. this just works around our miserable api to
-    // give us a reliable way to tell us when the stats will stop changing or something.
+    // Stats are collected by the paths watcher, so we do not look at the stats returned by
+    // this call. It is however the only API we currently have to tell us when the
+    // connection is drained and the stats will no longer change.
     conn_info.closed().await;
     output.emit(PathStats::from_watcher(watcher));
 


iroh/examples/transfer.rs
```diff
@@ -733,7 +743,6 @@ async fn fetch(
         duration: start.elapsed(),
     });
 
-    close_endpoint_with_timeout(&endpoint, output).await;
     output.emit(PathStats::from_watcher(watcher));
```
matheus23, line 751: I think ideally you close the endpoint and only then emit path stats, otherwise you might be missing stats.

flub: ok, this makes me very sad. our stats APIs are **still** terrible. Did some kind of miserable fix. Probably.

matheus23: What?

It's a simple matter of fact that during `Endpoint::close` you generate traffic, thus path stats.

My proposal would be to just call `Endpoint::close` here. Calling `Endpoint::close` twice doesn't hurt anything.

flub: Calling `Endpoint::close` twice is indeed clever, but also an ugly hack. There really should just be a way to wait for the connection be drained. There is no more activity on the connection after that.

I was thinking of using `Connection::closed` first, but then realised it doesn't actually tell you the connection is drained. Because you close it locally it would just be completed right away.

Maybe the right solution is to have a `Connection::drained` method?

flub: Ok, so the reason I do things this way is because I want to emit the stats once the path is drained. And the only way to get that state is to go via the `ConnectionInfo::closed` thing. I think that is a shortcoming in our API design. But I'm not in a place to fix that here.

Is closing the endpoint twice better or worse than this hack? Both are hacks. To me this version expresses more the intent: emit the stats once the connection is drained.

iroh/examples/transfer.rs
```diff
@@ -744,8 +744,9 @@ async fn fetch(
         duration: start.elapsed(),
     });
 
-    // stats are collected by the paths watcher. this just works around our miserable api to
-    // give us a reliable way to tell us when the stats will stop changing or something.
+    // Stats are collected by the paths watcher, so we do not look at the stats returned by
+    // this call. It is however the only API we currently have to tell us when the
+    // connection is drained and the stats will no longer change.
```
matheus23, line 749: Thank you. I appreciate this change.

flub: Thanks for calling me out on my crappy late-night emergency fixes :smile:

PR #4008, opened 2026-03-11
fix: increase DNS timeout and address lookup stagger intervals

Frando:
## Description

* Increases DNS_TIMEOUT from 1s to 3s
* When looking up endpoint info via DnsAddressLookup, add more staggered calls up to 3s to not give up as quick as before
* Don't ever cache negative responses in our DNS resolver
* Slightly improve logging about DnsAddressLookup so we can debug more issues better.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-relay/src/defaults.rs
+++ b/iroh-relay/src/defaults.rs
@@ -30,8 +30,11 @@ pub mod timeouts {
     /// Timeout used by the relay client while connecting to the relay server,
     /// using `TcpStream::connect`
     pub(crate) const DIAL_ENDPOINT_TIMEOUT: Duration = Duration::from_millis(1500);
-    /// Timeout for our async dns resolver
-    pub(crate) const DNS_TIMEOUT: Duration = Duration::from_secs(1);
+
+    /// Default timeout for DNS queries issued by [`DnsResolver`].
+    ///
+    /// [`DnsResolver`]: crate::dns::DnsResolver
+    pub(crate) const DNS_TIMEOUT: Duration = Duration::from_secs(3);
 
     /// Maximum time the server will attempt to get a successful write to the connection.
     #[cfg(feature = "server")]
--- a/iroh-relay/src/defaults.rs
+++ b/iroh-relay/src/defaults.rs
@@ -34,7 +34,7 @@ pub mod timeouts {
     /// Default timeout for DNS queries issued by [`DnsResolver`].
     ///
     /// [`DnsResolver`]: crate::dns::DnsResolver
-    pub(crate) const DNS_TIMEOUT: Duration = Duration::from_secs(3);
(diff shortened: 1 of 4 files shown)


iroh/src/address_lookup/dns.rs
```diff
@@ -114,13 +121,18 @@ impl AddressLookup for DnsAddressLookup {
     ) -> Option<BoxStream<Result<AddressLookupItem, AddressLookupError>>> {
         let resolver = self.dns_resolver.clone();
         let origin_domain = self.origin_domain.clone();
+        let span = debug_span!("DnsAddressLookup", id=%endpoint_id.fmt_short(), %origin_domain);
```
flub: The `id` logged is the one of the endpoint being looked up? Can we make that a bit clearer? Maybe `lookup_id` or `query_id` or so? I'm not entirely sure.

flub (approved):

iroh-relay/src/defaults.rs
```diff
@@ -30,8 +30,11 @@ pub mod timeouts {
     /// Timeout used by the relay client while connecting to the relay server,
     /// using `TcpStream::connect`
     pub(crate) const DIAL_ENDPOINT_TIMEOUT: Duration = Duration::from_millis(1500);
-    /// Timeout for our async dns resolver
-    pub(crate) const DNS_TIMEOUT: Duration = Duration::from_secs(1);
+
+    /// Default timeout for DNS queries issued by [`DnsResolver`].
+    ///
+    /// [`DnsResolver`]: crate::dns::DnsResolver
+    pub const DNS_TIMEOUT: Duration = Duration::from_secs(3);
```
flub: Eh, not sure I agree with this. Just to make the docs happy is not enough of a reason to make something pub. I would rather just hardcode the value in the docs.

flub (changes requested):

iroh-relay/src/defaults.rs
Frando: Right, I can revert. I thought it might still be useful, because downstream code might want to do calculations with this timeout. I.e. apply their own timeouts by adding something to this value. But can also revert, dunno?

flub: Please revert. It is unrelated to the thing we're fixing here and this is kind of a hotfix. So it's always better to keep the changes minimal even if we eventually decide to make this pub for some good reason.

Frando: Reverted it.

iroh/src/address_lookup/dns.rs
Frando: Done

flub (approved):

PR #4010, opened 2026-03-11
feat(iroh): allow propagation of addr filters through the endpoint

### LINKED ISSUE #4009: Accidental publishing of IP addresses in DNS
On current main we are accidentally publishing IP addresses into the pkarr/DNS address lookup service. We should not publish those by default, it should be a very explicit user action to publish those. By default only after connecting to a peer should IP addresses be exchanged between two peers, over the encrypted QUIC connection.

dignifiedquire:
This allows propagating the filters to all added services, ensuring they are actually applied. 

The issue in #4009 happened because the preset filter wasn't applied.

Fixes #4009

### DIFF
--- a/iroh/examples/mdns_address_lookup.rs
+++ b/iroh/examples/mdns_address_lookup.rs
@@ -5,17 +5,32 @@
 //! This is an async, non-determinate process, so the number of EndpointIDs discovered each time may be different. If you have other iroh endpoints or iroh endpoints with [`address_lookup::MdnsAddressLookup`] enabled, it may discover those endpoints as well.
 use std::time::Duration;
 
-use iroh::{Endpoint, EndpointId, address_lookup, endpoint_info::UserData};
+use iroh::{
+    Endpoint, EndpointId, RelayMode,
+    address_lookup::{self, AddrFilter},
+    endpoint_info::UserData,
+};
 use n0_error::Result;
 use n0_future::StreamExt;
 use tokio::task::JoinSet;
+
+/// Creates an endpoint configured for local mDNS discovery only.
+///
+/// Uses `empty_builder` with relay disabled and an IP-only address filter,
+/// since mDNS discovers endpoints on the local network via IP addresses.
+async fn bind_mdns_endpoint() -> std::result::Result<Endpoint, iroh::endpoint::BindError> {
+    Endpoint::empty_builder(RelayMode::Disabled)
+        .addr_filter(AddrFilter::ip_only())
+        .bind()
+        .await
+}
 
 #[tokio::main]
 async fn main() -> Result<()> {
     tracing_subscriber::fmt::init();
     println!("Discovering Local Endpoints Example!");
 
-    let ep = Endpoint::bind().await?;
+    let ep = bind_mdns_endpoint().await?;
     let endpoint_id = ep.id();
 
     let mdns = address_lookup::MdnsAddressLookup::builder().build(endpoint_id)?;
@@ -64,7 +79,7 @@ async fn main() -> Result<()> {
     for _ in 0..endpoint_count {
         let ud = user_data.clone();
         set.spawn(async move {
-            let ep = Endpoint::bind().await?;
+            let ep = bind_mdns_endpoint().await?;
             ep.address_lookup()?
                 .add(address_lookup::MdnsAddressLookup::builder().build(ep.id())?);
             ep.set_user_data_for_address_lookup(Some(ud));


iroh/src/endpoint.rs
```diff
@@ -549,7 +553,26 @@ impl Builder {
     ///
     /// See the documentation of the [`crate::address_lookup::AddressLookup`] trait for details.
     pub fn address_lookup(mut self, address_lookup: impl AddressLookupBuilder) -> Self {
-        self.address_lookup.push(Box::new(address_lookup));
+        if let Some(filter) = self.addr_filter.clone() {
+            self.address_lookup
```
rklaehn: This means that if you add the address lookup before the filter, it won't apply. Not sure about this. We could also apply the filters in the final build step.

```rust
.address_lookup(foo)
.addr_filter(filter)
.address_lookup(bar)
.bind().await?
```

foo would be unfiltered, bar would be filtered, in the current impl, correct?

dignifiedquire: yeah, not great, agreed

iroh/examples/transfer.rs
```diff
@@ -465,6 +466,11 @@ impl EndpointArgs {
         };
         let mut builder = Endpoint::empty_builder(relay_mode);
         builder = builder.secret_key(secret_key);
+        if self.no_relay {
+            builder = builder.addr_filter(AddrFilter::ip_only());
+        } else {
+            builder = builder.addr_filter(AddrFilter::relay_only());
+        }
```
matheus23, line 473: I don't think we want to decide this on the endpoint level. E.g. mdns is likely to *want* to publish IP addrs, even when relays are enabled, while pkarr doesn't want to publish ip addrs.

IMO we should have `PkarrPublisherBuilder` have an addr filter on it by default (maybe implemented via the `AddrFilter` wrapper), and have that be overridable via a `PkarrPublisherBuilder::addr_filter` function.

And the same for `MdnsLookup`.

The only problem I don't really know how to solve is that the pkarr publisher needs to know if relays are enabled before it can decide whether to enable IP publishing or not. Perhaps we can delay the choice of default addr filter until the pkarr publisher is built?

iroh/src/endpoint.rs
dignifiedquire: fixed, now applied on the concurrent lookup

iroh/examples/transfer.rs
dignifiedquire: I think not doing it at the endpoint level optionally is a massive footgun, literally the authors of this new mechanism didn't see the issue in updating these examples

iroh/src/endpoint/presets.rs
```diff
@@ -54,23 +54,20 @@ pub struct N0;
 
 impl Preset for N0 {
     fn apply(self, mut builder: Builder) -> Builder {
-        let filter = AddrFilter::relay_only();
-        builder = builder.address_lookup(PkarrPublisher::n0_dns().with_addr_filter(filter.clone()));
+        builder = builder.addr_filter(AddrFilter::relay_only());
```
rklaehn, line 57: This means that if you use the n0 presets, all address lookups including mdns will only ever publish relay urls.

If we have an mdns example we would have to modify it, otherwise it won't work anymore. Also we should probably document this at the mdns address lookup.

dignifiedquire: yes we should, but that is also I think how it behaved before? I am not sure this is tricky

rklaehn:
Just looked at the mdns example.

It adds the mdns address lookup to the existing endpoint, after the builder is done.

    ep.address_lookup()?.add(mdns.clone());

So on the plus side, it still works. On the minus side, it is kind of confusing that for adding address lookups to an existing endpoint the filter does not apply...

dignifiedquire:
> So on the plus side, it still works. On the minus side, it is kind of confusing that for adding address lookups to an existing endpoint the filter does not apply...

it should still apply, because this inserts it into the concurrent setup, which now always filters

rklaehn:
> > So on the plus side, it still works. On the minus side, it is kind of confusing that for adding address lookups to an existing endpoint the filter does not apply...
> 
> it should still apply, because this inserts it into the concurrent setup, which now always filters

Ah, I thought you meant apply during building of the concurrent address lookup, but you actually integrated it into the concurrent lookup. THen it's fine.

iroh/src/endpoint/presets.rs
matheus23: It's not how 0.96 behaved and before behaved.
Originally we only published IP addrs in mdns, and later added relay publishing to mdns.

dignifiedquire:
Merging, but we will have more follow up work as indicated in the referenced PRs

Issue #3811, opened 2025-12-26
feature: Option to enable/disable UPnP/PCP


izderadicka:
Based on discussion in Discord UPnP/PCP  is always on -   it would be nice to have option to disable it.

rklaehn:
Portmapper can now be disabled, either completely via a feature flag, or dynamically using https://docs.rs/iroh/latest/iroh/endpoint/struct.Builder.html#method.portmapper_config

PR #4011, opened 2026-03-12
deps: Feature gate portmapper, and allow disabling portmapper by default.

rklaehn:
## Description

Iroh on esp32 is over 4 MiB out of the box, which means that it won't run on the majority of the esp32 devices out there.

There are a number of deps we can get rid of without affecting core functionality. Portmapper is the easiest to feature gate since it is already disabled for wasm. Portmapper and its dep idg-next are big enough that when disabling it it barely fits on esp32 4 MiB devices.

Implements https://github.com/n0-computer/iroh/issues/3811

## Breaking Changes

Builds that use no-default-features won't have the portmapper. This does not change APIs, but will change functionality.

## Notes & open questions

In general, if a feature is disabled, should we stub the public API to just report 0/noops, or hide it? E.g. if you were to stub portmapper metrics you would not need to enable portmapper for our bench project.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -308,6 +308,25 @@ jobs:
       - name: clippy check (default features)
         run: cargo clippy --workspace --all-targets --lib --bins --tests --benches --examples
 
+  check_optional_deps:
+    timeout-minutes: 10
+    name: Check optional deps are absent without features
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - uses: dtolnay/rust-toolchain@stable
+      - name: Verify optional deps are absent without features
+        run: |
+          tree=$(cargo tree -p iroh --no-default-features --edges normal)
+          fail=0
+          for dep in portmapper igd-next swarm-discovery; do
+            if echo "$tree" | grep -q "$dep"; then
+              echo "ERROR: $dep should not be in the dependency tree without its feature"
+              fail=1
+            fi
+          done
+          exit $fail
+
   msrv:
     if: "github.event_name != 'pull_request' || ! contains(github.event.pull_request.labels.*.name, 'flaky-test')"
     timeout-minutes: 30


iroh/src/portmapper.rs
```diff
@@ -0,0 +1,52 @@
...
+
+#[cfg(not(all(not(wasm_browser), feature = "portmapper")))]
+pub(crate) use stub::Client;
+
+#[cfg(not(all(not(wasm_browser), feature = "portmapper")))]
+pub(crate) fn create_client(_metrics: &crate::metrics::EndpointMetrics) -> Client {
+    Client::new()
+}
+
+#[cfg(not(all(not(wasm_browser), feature = "portmapper")))]
```
dignifiedquire: shouldn't we just make it optional instead of using a stub if it is disabled?

rklaehn: I am not sure. Using the stub means that you don't have to sprinkle cfgs all over the place. I would say in cases where it is easy and harmless to define a stub that is preferable over a cfg maze all over the place.

I had the cfg variant, and brought back the stub because I liked it more. I can revert if you want.

dignifiedquire: well, we also have the feature request to disable it dynamically so I was thinking we can combine that

rklaehn:
Added the ability to configure and disable the portmapper in the builder:

.portmapper_config(PortmapperConfig::Disabled).

PortmapperConfig is basically a bool newtype at this time, but the enabled case is open, so we could add config later.

We could be cute and change portmapper_config to take `impl Into<PortmapperConfig>` and then define `Into<PortmapperConfig>` for `bool`.

```rust
.portmapper(false) // no need to import PortmapperConfig
```

We might also return an error if the portmapper is disabled via the feature flag, but not sure about that. I guess it might make sense if we later want to add config options, then we could reuse that error if the options don't make sense.

PR #4012, opened 2026-03-12
fix(iroh)!: Don't publish IP addresses from `PkarrPublisher` by default

### LINKED ISSUE #4009: Accidental publishing of IP addresses in DNS

matheus23:
## Description

Partially reverts #3987 so we have more control over the default `AddrFilter` from within the `PkarrPublisherBuilder`, and because we need it for pkarr, we keep this consistent across publishing services.

As a bonus, now pure resolvers don't suddently have a builder method for address filtering that cannot apply to them (but they can still be wrapped in `FilteredAddressLookup`, even though that's less discoverable).

Fixes #4009 

## Breaking Changes

- **behavioral**: `PkarrPublisher` and `DhtPublisher` will now by default publish *only* the relay URL, instead of detecting if `RelayMode::Disabled` was set and in such cases publishing IP addresses. We recommend users to explicitly set `.addr_mode(AddrFilter::unfiltered())` if they set `relay_mode(RelayMode::Disabled)`, to re-enable publishing IP addresses publicly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -188,7 +188,7 @@ jobs:
         run: rustup target add wasm32-unknown-unknown
 
       - name: Install cargo-binstall
-        uses: cargo-bins/cargo-binstall@v1.17.6
+        uses: cargo-bins/cargo-binstall@v1.17.7
 
       - name: Install wasm-bindgen-test-runner
         run: cargo binstall wasm-bindgen-cli --locked --no-confirm
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -94,7 +94,7 @@ jobs:
             aws s3 cp s3://vorc/iroh-dns-server-linux-aarch64-${{ inputs.base_hash }} bins/linux/arm64/iroh-dns-server
         
         - name: Build Docker image (iroh-relay)
-          uses: docker/build-push-action@v6
+          uses: docker/build-push-action@v7
           with:
             context: .
             push: ${{ inputs.publish }}
@@ -104,7 +104,7 @@ jobs:
             file: docker/Dockerfile.ci
 
         - name: Build Docker image (iroh-dns-server)
-          uses: docker/build-push-action@v6
+          uses: docker/build-push-action@v7
           with:
             context: .
             push: ${{ inputs.publish }}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2135,7 +2135,7 @@ dependencies = [
  "serde",
  "serde_json",
  "smallvec",
- "strum",
+ "strum 0.28.0",
  "swarm-discovery",
  "sync_wrapper",
  "time",
@@ -2232,7 +2232,7 @@ dependencies = [
  "serde",
  "serde_json",
  "struct_iterable",
- "strum",
+ "strum 0.28.0",
  "tempfile",
  "tokio",
  "tokio-rustls",
@@ -2329,7 +2329,7 @@ dependencies = [
  "serde_json",
  "sha1",
  "simdutf8",
- "strum",
+ "strum 0.28.0",
  "time",
  "tokio",
  "tokio-rustls",
@@ -2650,7 +2650,7 @@ dependencies = [
  "serde_json",
  "serde_with",
  "smallvec",
- "strum",
+ "strum 0.27.2",
 ]
 
 [[package]]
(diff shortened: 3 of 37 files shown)


iroh/src/address_lookup/pkarr/dht.rs
```diff
@@ -198,8 +192,29 @@ impl Builder {
...
+    /// even bypass a relay connection entirely.
+    pub fn addr_filter(mut self, filter: AddrFilter) -> Self {
+        self.addr_filter = filter;
+        self
+    }
+
     /// Builds the address lookup mechanism.
-    pub fn build(self) -> Result<DhtAddressLookup, AddressLookupBuilderError> {
+    ///
+    /// `relays_enabled` determines the default address filter when none was set via
```
Frando: remove, outdated.

iroh-relay/src/endpoint_info.rs
```diff
@@ -254,6 +254,11 @@ impl AddrFilter {
         Self(Some(Arc::new(f)))
     }
 
+    /// Constructs a filter that doesn't filter addresses and passes all through.
+    pub fn unfiltered() -> Self {
+        Self::new(|addrs| addrs.iter().cloned().collect())
```
dignifiedquire, line 259: do all filters have to allocate? 😭 f

matheus23: I had *just* the branch sitting on my hard drive for you: https://github.com/n0-computer/iroh/pull/4018

matheus23: But do note the discussion section in that PR.

dignifiedquire:
the mdns example might need some updates

iroh/src/address_lookup/pkarr/dht.rs
```diff
@@ -42,25 +42,17 @@ const REPUBLISH_DELAY: Duration = Duration::from_secs(60 * 60);
 /// be used as both a publisher and resolver.  Calling [`DhtAddressLookup::publish`] will start
 /// a background task that periodically publishes the endpoint address.
 ///
-/// By default, [`DhtAddressLookup`] will publish all addresses it receives:
-/// direct IP addresses and relay URLs. No internal address limits or truncation
-/// are applied.
-///
-/// You can supply an [`AddrFilter`] via [`AddrLookupBuilder::with_addr_filter`] to
-/// control which addresses are published.
+/// [`DhtAddressLookup`] filters published addresses based on the endpoint's relay
```
Frando: comment is wrong now, it's always relay_only unless set differently by the user, right?

iroh/src/address_lookup/pkarr.rs
```diff
@@ -19,14 +19,11 @@
...
 //!
-//! [`PkarrPublisher`] publishes all addresses it receives by default, with no internal limiting.
-//! You can supply an [`AddrFilter`] via [`AddrLookupBuilder::with_addr_filter`] to limit the kinds
-//! and number of addresses that get published. When the [`PkarrPublisher`] is added via the
-//! [`N0` preset], a filter to publish only relay addresses is applied. If you add the publisher
-//! directly to an endpoint, you might want to do the same to not leak IP addresses publicly.
-//!
-//! Note that [`PkarrResolver`] and [`address_lookup::DnsAddressLookup`] only resolve and do not publish,
-//! so filtering does not apply to them.
+//! [`PkarrPublisher`] filters published addresses based on the endpoint's relay configuration
```
Frando: comment is wrong now, it's always relay_only unless set differently by the user, right?

iroh/src/address_lookup/pkarr.rs
```diff
@@ -214,6 +215,20 @@ impl PkarrPublisherBuilder {
...
     }
 
+    /// Sets the address filter to control which addresses are published to the pkarr server.
+    ///
+    /// By default [`AddrFilter::relay_only`] is used. This avoids leaking IP addresses to the
+    /// public pkarr server.
+    ///
+    /// However, enabling IP address publishing can be useful, e.g. when iroh runs on a machine
+    /// connected to the internet via public IP addresses without a firewall.
+    /// In such cases, publishing them can make dialing such endpoints via DNS or PKARR lookup
```
Frando: nit: I think we usually write `Pkarr` not `PKARR`

matheus23:
> the mdns example might need some updates

Does it? It seems to be working fine for me.

dignifiedquire:
> > the mdns example might need some updates
> 
> Does it? It seems to be working fine for me.

it does, but we don't need to remove filters anymore, we could keep using the preset now

matheus23:
Updated the example. This should be ready.

PR #4014, opened 2026-03-12
feat(iroh)!: cleanup preset API

dignifiedquire:
## Description

The current default builder approach hardcodes the number 0 servers directly, to make this more explicit, but still have a clean API for examples everywhere this refactors the builder and bind API of the endpoint.

## Breaking Changes

- `iroh::Endpoint::builder` now takes a preset
- `iroh::Endpoint::bind` now takes a preset
- `iroh::Endpoint::empty_builder` now assumes `RelayMode::Disabled` and takes no arguments
- `iroh::EndpointBuilder::empty` now assumes `RelayMode::Disabled` and takes no arguments

## Example

```rust
// Before
let endpoint = Endpoint::bind().await?;

// After
let endpoint = Endpoint::bind(presets::N0).await?;
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -188,7 +188,7 @@ jobs:
         run: rustup target add wasm32-unknown-unknown
 
       - name: Install cargo-binstall
-        uses: cargo-bins/cargo-binstall@v1.17.6
+        uses: cargo-bins/cargo-binstall@v1.17.7
 
       - name: Install wasm-bindgen-test-runner
         run: cargo binstall wasm-bindgen-cli --locked --no-confirm
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -94,7 +94,7 @@ jobs:
             aws s3 cp s3://vorc/iroh-dns-server-linux-aarch64-${{ inputs.base_hash }} bins/linux/arm64/iroh-dns-server
         
         - name: Build Docker image (iroh-relay)
-          uses: docker/build-push-action@v6
+          uses: docker/build-push-action@v7
           with:
             context: .
             push: ${{ inputs.publish }}
@@ -104,7 +104,7 @@ jobs:
             file: docker/Dockerfile.ci
 
         - name: Build Docker image (iroh-dns-server)
-          uses: docker/build-push-action@v6
+          uses: docker/build-push-action@v7
           with:
             context: .
             push: ${{ inputs.publish }}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2135,7 +2135,7 @@ dependencies = [
  "serde",
  "serde_json",
  "smallvec",
- "strum",
+ "strum 0.28.0",
  "swarm-discovery",
  "sync_wrapper",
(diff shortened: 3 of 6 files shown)


iroh/examples/connect.rs
```diff
@@ -37,7 +39,7 @@ async fn main() -> Result<()> {
     println!("public key: {}", secret_key.public());
 
     // Build a `Endpoint`, which uses PublicKeys as endpoint identifiers, uses QUIC for directly connecting to other endpoints, and uses the relay protocol and relay servers to holepunch direct connections between endpoints when there are NATs or firewalls preventing direct connections. If no direct connection can be made, packets are relayed over the relay servers.
-    let endpoint = Endpoint::builder()
+    let endpoint = Endpoint::builder(presets::N0)
```
Frando, line 42: Not super sure but I think I'd prefer `N0Preset` to `presets::N0`, i.e. rename `N0` to `N0Preset`. People use auto-imports and then you'd often have `N0` which reads weird. Also `N0Preset` is better discoverable with auto-complete. maye `PresetN0` would be even better for autocomplete, unsure.

This is not new to this PR, but this PR makes presets appear at many more places, so this warrants a new discussion on this I think.

iroh/src/endpoint.rs
```diff
@@ -153,27 +153,25 @@ impl Builder {
...
-    pub fn preset<P: Preset>(mut self, preset: P) -> Self {
+    pub fn preset(mut self, preset: impl Preset) -> Self {
         self = preset.apply(self);
         self
     }
 
-    /// Creates an empty builder with no address lookup  services.
-    pub fn empty(relay_mode: RelayMode) -> Self {
-        let mut transports = vec![
+    /// Creates an empty builder with no address lookup services, and `RelayMode::Disabled`.
```
Frando: add doclink for `RelayMode::Disabled`

iroh/src/endpoint.rs
```diff
@@ -788,25 +786,21 @@ impl Endpoint {
...
+    pub fn empty_builder() -> Builder {
+        Builder::empty()
     }
 
-    /// Constructs a default [`Endpoint`] and binds it immediately.
-    ///
-    /// Uses the [`presets::N0`] as configuration.
-    pub async fn bind() -> Result<Self, BindError> {
-        Self::builder().bind().await
+    /// Constructs a default [`Endpoint`] and binds it immediately using the provided [`Preset`].
```
```suggestion
    /// Constructs a default [`Endpoint`] using the provided [`Preset`] and binds it immediately.
```

Frando (commented):
I like it.

I'll miss `Endpoint::bind()` I think, but @okdistribute argued well that we should not ship n0 defaults silently. `Endpoint::bind(preset::N0)` (or maybe `Endpoint::bind(N0Preset)`, tbd) is still reasonably short for examples etc and makes things obvious. So yeah, let's do this.

rklaehn:
I am very much in favour of requiring presets everywhere, simply because having a preset going away due to a feature flag being disabled is much less weird than an essential fn like bind() or builder() going away or changing behaviour.

PR #4018, opened 2026-03-12
feat(iroh): Avoid allocations in `AddrFilter`

matheus23:
## Description

- Makes `AddrFilter` essentially a wrapper around `&Vec -> Cow<Vec>` instead of `&BTreeSet -> Vec`.
- Reworks `EndpointInfo` and `EndpointData` APIs. Some docs were outdated (e.g. claiming `with_relay_url(None)` would remove relays), etc. Generally also removed `Deref` and `DerefMut` implementations from `EndpointInfo -> EndpointData`.

## Breaking Changes

`iroh::address_lookup::EndpointInfo`:
- `EndpointInfo::with_relay_url` now takes `RelayUrl` instead of `Option<RelayUrl>`
- `EndpointInfo::with_ip_addrs` now takes `Vec<SocketAddr>` instead of `BTreeSet<SocketAddr>`.
- Removed `Deref` and `DerefMut` with `Target = EndpointData` from `EndpointInfo`.
- Added `EndpointData::{addrs, relay_urls, user_data, ip_addrs}`.

`iroh::address_lookup::EndpointData`:
- `EndpointData::new` now takes a `Vec<TransportAddr>` instead of `impl IntoIterator<Item = TransportAddr>` and keeps the ordering of addresses. Use `EndpointAddr::from_iter` instead (although less performant).
- `EndpointData::with_user_data` now takes `UserData` instead of `Option<UserData>`.
- `EndpointData::filtered_addrs` now returns `Cow<Vec<TransportAddr>>` instead of `Vec<TransportAddr>`.
- Added `EndpointData::{add_relay_url, add_ip_addrs, apply_filter}`.
- Added `impl From<BTreeSet<{TransportAddr, SocketAddr}>> for EndpointData`.
- Added `impl FromIterator<TransportAddr> for EndpointData`.

`iroh::address_lookup::AddrFilterFn`:
- Changed from being `BTreeSet`-based to `Vec`-based.
- Return type changed to `Cow<Vec<TransportAddr>>` from`Vec<TransportAddr>`.

## Notes

There's some work required to keep the `EndpointData` transport addresses deduplicated, and deduplicating a `Vec` is unfortunately somewhat expensive and requires allocations, but I think it's better to have this than not to.
I already saw some code in some publishing services that just keep adding new information to a single `EndpointData` and keep publishing. Without deduplication, this would go wrong. Maybe that should change nonetheless, but that'd be a PR for another day.

## Change checklist

- [x] Self-review.
- [x] Tests if relevant.

### DIFF
--- a/iroh-relay/src/endpoint_info.rs
+++ b/iroh-relay/src/endpoint_info.rs
@@ -34,7 +34,7 @@
 
 use std::{
     borrow::Cow,
-    collections::{BTreeMap, BTreeSet},
+    collections::{BTreeMap, BTreeSet, HashSet},
     fmt::{self, Display},
     hash::Hash,
     net::SocketAddr,
@@ -42,7 +42,6 @@ use std::{
     sync::Arc,
 };
 
-use ahash::AHashSet;
 use iroh_base::{EndpointAddr, EndpointId, KeyParsingError, RelayUrl, SecretKey, TransportAddr};
 use n0_error::{e, ensure, stack_error};
 use url::Url;
@@ -128,9 +127,9 @@ pub struct EndpointData {
     user_data: Option<UserData>,
 }
 
-fn dedup<T: Eq + Hash + Clone>(items: &mut Vec<T>) -> AHashSet<T> {
+fn dedup<T: Eq + Hash + Clone>(items: &mut Vec<T>) -> HashSet<T> {
     // Remove all duplicate entries, but keep the array order.
-    let mut seen = AHashSet::new();
+    let mut seen = HashSet::new();
     items.retain(|item| seen.insert(item.clone()));
     seen
 }


dignifiedquire (approved):

dignifiedquire:
yeah, having just a vec might make more sense

rklaehn:
I had something similar, but one of the goals of the filter functionality was to be able to not just filter but also reorder addrs. Hence the vec.

See the PR from ramfox: https://github.com/n0-computer/iroh/pull/3960

`/// A filter and/or reordering function applied to transport addresses before publishing.`

**and/or reordering**

Now, do we need reordering? I am not sure. But just wanted to mention that this design is making reordering impossible.

I would argue that ordering can be important in some cases. E.g. you can order from "good" to "bad" or "new" to "old", and dialing the addrs in a reasonable order will have much more benefit than saving an allocation.

matheus23:
Yeah - thanks for raising this @rklaehn 

I briefly discussed this with @Frando. Address reordering became impossible with #3987, but bringing it back makes sense.

My response in discord was: Let's switch everything to `Vec`, including the addresses in `EndpointData`. Does that make sense?

rklaehn:
> Yeah - thanks for raising this @rklaehn
> 
> I briefly discussed this with @Frando. Address reordering became impossible with #3987, but bringing it back makes sense.
> 
> My response in discord was: Let's switch everything to `Vec`, including the addresses in `EndpointData`. Does that make sense?

Yes, if the order is interesting we should also allow it there. However, I would prefer if we don't allow duplicates, since I really don't see a good reason to have those.

addrs are not pub in EndpointData, so I guess having this a vec but preventing dupes is viable.

matheus23:
> Yes, if the order is interesting we should also allow it there. However, I would prefer if we don't allow duplicates, since I really don't see a good reason to have those.
> 
> addrs are not pub in EndpointData, so I guess having this a vec but preventing dupes is viable.

The only problem is that this is somewhat ugly to enforce. You need to enforce everywhere you can now add ordered elements to `EndpointData`. But now that I looked into that struct and its methods, I realize there's a *lot* of weird APIs that need cleanup. I think I'll try to work on this.

iroh-relay/src/endpoint_info.rs
```diff
@@ -121,40 +122,88 @@ impl EndpointIdExt for EndpointId {
 #[derive(Debug, Clone, Default, Eq, PartialEq)]
 pub struct EndpointData {
     /// addresses where this endpoint can be reached.
-    addrs: BTreeSet<TransportAddr>,
+    addrs: Vec<TransportAddr>,
     /// Optional user-defined [`UserData`] for this endpoint.
     user_data: Option<UserData>,
 }
 
+fn dedup<T: PartialEq + Ord + Clone>(items: &mut Vec<T>) -> BTreeSet<T> {
```
rklaehn: It probably does not matter much for small n, but it seems that we don't care about order in this set anywhere. So I would use a HashSet or FxHashSet.

Now you could even use a sorted deduped vec / vecset and avoid allocations entirely for small n, but we can do that later.

iroh-relay/src/endpoint_info.rs
```diff
@@ -215,13 +240,52 @@ impl EndpointData {
...
+    pub fn filtered_addrs(&self, filter: &AddrFilter) -> Cow<'_, Vec<TransportAddr>> {
         filter.apply(&self.addrs)
     }
+
+    /// Returns the `EndpointData` with given filter applied.
+    pub fn apply_filter(&self, filter: &AddrFilter) -> Cow<'_, Self> {
+        match self.filtered_addrs(filter) {
+            Cow::Borrowed(_) => Cow::Borrowed(self),
+            Cow::Owned(addrs) => {
+                let mut data = EndpointData::new(addrs);
```
rklaehn, line 246: We can't really know if the filter will duplicate, so I think it is good to call new here even if it dedups again.

iroh-relay/src/endpoint_info.rs
matheus23: `TransportAddr` doesn't implement `Hash`.

Should I add that to `TransportAddr`, or just prefer using `BTreeSet`?

iroh-relay/src/endpoint_info.rs
```diff
@@ -121,40 +122,88 @@ impl EndpointIdExt for EndpointId {
...
-    addrs: BTreeSet<TransportAddr>,
+    addrs: Vec<TransportAddr>,
     /// Optional user-defined [`UserData`] for this endpoint.
     user_data: Option<UserData>,
 }
 
+fn dedup<T: PartialEq + Ord + Clone>(items: &mut Vec<T>) -> BTreeSet<T> {
+    // Remove all duplicate entries, but keep the array order.
+    let mut item_set = BTreeSet::new();
+    for (i, item) in items.clone().into_iter().enumerate() {
```
rklaehn: I think this clone can be avoided. Using ahash since it is already in the deps. Not sure about ahash vs fxhash.

```rust
fn dedup<T: Eq + Hash + Clone>(items: &mut Vec<T>) -> AHashSet<T> {
    let mut seen = AHashSet::new();
    items.retain(|item| seen.insert(item.clone()));
    seen
}
```

iroh-relay/src/endpoint_info.rs
rklaehn: It seems having Hash for TransportAddr is fine. Lots of people prefer the hash based collections due to minor speed advantage.

iroh-relay/src/endpoint_info.rs
```diff
@@ -127,9 +127,9 @@ pub struct EndpointData {
     user_data: Option<UserData>,
 }
 
-fn dedup<T: PartialEq + Ord + Clone>(items: &mut Vec<T>) -> BTreeSet<T> {
+fn dedup<T: Eq + Hash + Clone>(items: &mut Vec<T>) -> HashSet<T> {
     // Remove all duplicate entries, but keep the array order.
-    let mut item_set = BTreeSet::new();
+    let mut item_set = HashSet::new();
     for (i, item) in items.clone().into_iter().enumerate() {
```
rklaehn: I think this clone can be avoided. Using ahash since it is already in the deps. Not sure about ahash vs fxhash.

```rust
fn dedup<T: Eq + Hash + Clone>(items: &mut Vec<T>) -> AHashSet<T> {
    let mut seen = AHashSet::new();
    items.retain(|item| seen.insert(item.clone()));
    seen
}
```

iroh-relay/src/endpoint_info.rs
matheus23: Neat! Changed it to this.

iroh-relay/src/endpoint_info.rs
matheus23: Uh, I guess I was wrong. `TransportAddr` totally implements `Hash`... changed it!

iroh-relay/src/endpoint_info.rs
```diff
@@ -127,9 +127,9 @@ pub struct EndpointData {
     user_data: Option<UserData>,
 }
 
-fn dedup<T: PartialEq + Ord + Clone>(items: &mut Vec<T>) -> BTreeSet<T> {
+fn dedup<T: Eq + Hash + Clone>(items: &mut Vec<T>) -> HashSet<T> {
```
rklaehn, line 130: This is only ever used for TransportAddr, so maybe it does not have to be generic?

matheus23: There was a previous version where I also used it with `SocketAddr`.
IMO this helps make it clearer what it does (we know it can't "reach into" the `T` to decide deduplication besides `Eq`, `Hash` and `Clone`).

PR #4020, opened 2026-03-12
feat: Allow configuring NetReport

rklaehn:
## Description

This allows us to disable captive portal checks and https probes on esp32

## Breaking Changes

None

## Notes & open questions

Should we also allow for disabling QAD probes? I was thinking we leave it for later. The config object is non-exhaustive.

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.


iroh/src/net_report/options.rs
```diff
@@ -57,42 +68,47 @@ mod imp {
...
     /// Only HTTPS probes are supported in browsers.
     /// These are run by default.
     #[derive(Debug, Clone)]
     pub(crate) struct Options {
-        /// Enable https probes
-        ///
-        /// On by default
-        pub(crate) https: bool,
+        /// User-facing configuration.
+        pub(crate) config: Config,
```
dignifiedquire: nesting a `Config` inside `Options` makes me not very happy :/

rklaehn: Yeah, it looks weird. But this is only the wasm path.

In non wasm, the Options contains more than just the Config:

```rust
    pub(crate) struct Options {
        /// The configuration needed to launch QUIC address discovery probes.
        ///
        /// If not provided, will not run QUIC address discovery.
        pub(crate) quic_config: Option<QuicConfig>,
        /// TLS config for HTTPS probes.
        pub(crate) tls_config: rustls::ClientConfig,
        /// User-facing configuration.
        pub(crate) config: Config,
    }
```

dignifiedquire: maybe call it `user_config` or something like that?

iroh/src/endpoint.rs
```diff
@@ -98,10 +98,10 @@ pub use self::{
         VarIntBoundsExceeded, WriteError, Written,
     },
 };
-pub use crate::portmapper::PortmapperConfig;
 #[cfg(not(wasm_browser))]
 use crate::socket::transports::IpConfig;
 use crate::socket::transports::TransportConfig;
+pub use crate::{net_report::Config as NetReportConfig, portmapper::PortmapperConfig};
```
flub: fwiw the compiler errors users will face from this alias are super annoying since doc.rs will document the `NetReportConfig` as parameter type.

iroh/src/net_report.rs
```diff
@@ -85,6 +85,44 @@ pub use self::{
...
+///
+/// Controls which probes and checks are performed when generating network reports.
+/// All options default to `true`.
+#[derive(Debug, Clone)]
+#[non_exhaustive]
+pub struct Config {
+    /// Run HTTPS latency probes against relay servers.
+    ///
+    /// Disable this on constrained devices where the extra TCP connections are
+    /// too expensive — QAD (UDP) probes already measure relay latency.
```
flub: You really need to extensively describe the severe implications of doing this. Probably also add a big fat warning.

iroh/src/net_report.rs
```diff
@@ -85,6 +85,44 @@ pub use self::{
...
+    pub https_probes: bool,
+    /// Check for captive portals when generating the first report.
+    ///
+    /// Disable this on embedded devices or environments where captive portal
+    /// detection is not meaningful.
```
flub: Why is it not meaningful? If you plug your home-made chrome-cast device into your hotel tv then it might be relevant. I think describing the consequences so folks can understand the tradeoffs is better than giving a sweeping statement like this.

iroh/src/net_report.rs
```diff
@@ -85,6 +85,44 @@ pub use self::{
 };
 pub(crate) use self::{options::Options, reportgen::QuicConfig};
 
+/// Configuration for the net report service.
+///
+/// Controls which probes and checks are performed when generating network reports.
+/// All options default to `true`.
+#[derive(Debug, Clone)]
+#[non_exhaustive]
```
flub, line 91: :+1:

iroh/src/endpoint.rs
```diff
@@ -703,6 +706,15 @@ impl Builder {
         self
     }
 
+    /// Configures the net report service.
+    ///
+    /// Controls which probes (HTTPS latency, captive portal detection) are run.
+    /// Defaults to all probes enabled.
+    pub fn net_report_config(mut self, config: NetReportConfig) -> Self {
```
flub: This also needs a ton more description. People don't know what all these things are and what the implications are. This is right in the builder, the first thing they might be looking at.

Also, "net report service" is a pretty sad name that doesn't have any meaning (I still struggle with this!). I'm really not a fan of now using this in public APIs and docs.

iroh/src/endpoint.rs
rklaehn: I can just rename it as NetReportConfig directly. Just wanted to be consistent with the previous code:

```rust
pub use net_report::{
    Report as NetReport, TIMEOUT as NET_REPORT_TIMEOUT,
};
```

Should I just rename them all to what they are going to be exported as?

iroh/src/net_report.rs
rklaehn: What are the implications? I thought https probes were mostly a legacy thing?

QAD probes alone seem to be good enough to let my little esp32 figure out the best relay.

iroh/src/net_report.rs
rklaehn: I am not saying it isn't meaningful for embedded devices. Environments where captive portal detection is not meaningful would be if you *know* that you won't have a captive portal.

But you are right, I will change the wording.

iroh/src/endpoint.rs
rklaehn: Ok, I have now renamed the config. Let me know if I also should rename Report and TIMEOUT.

iroh/src/endpoint.rs
rklaehn: > Also, "net report service" is a pretty sad name that doesn't have any meaning (I still struggle with this!). I'm really not a fan of now using this in public APIs and docs.

net report component? 🤷

flub:
(/me deferring to @dignifiedquire for the rest of this review)

iroh/src/net_report.rs
```diff
@@ -83,6 +83,56 @@ pub(crate) use self::{
...
+pub struct NetReportConfig {
+    /// Run HTTPS latency probes against relay servers.
+    ///
+    /// HTTPS latency probes perform an empty HTTPS GET request to each configured
+    /// relay server and measure latency.
+    ///
+    /// They are performed in addition to the QUIC address discovery (QAD) probes.
+    /// They are the only way to detect relay latencies and thus the preferred relay
```
Frando: "In networks that do not allow QUIC traffic, they are the only way .."
to make it clearer that the "only way" only applies to some networks

Frando (approved):
LGTM

dignifiedquire (approved):

PR #4021, opened 2026-03-13
feat(iroh): add tracing events for connections

Frando:
## Description

This adds `iroh::_event::conn` events for incoming and outgoing connections. We already have path events, but no connection events, so this makes our event system more complete, and gives good context when reading `iroh::_events`.

### DIFF
--- a/iroh/src/socket/remote_map/remote_state.rs
+++ b/iroh/src/socket/remote_map/remote_state.rs
@@ -15,7 +15,7 @@ use n0_future::{
     time::{self, Duration, Instant},
 };
 use n0_watcher::{Watchable, Watcher};
-use noq::WeakConnectionHandle;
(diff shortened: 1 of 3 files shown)


iroh/src/socket/remote_map/remote_state.rs
```diff
@@ -568,7 +568,14 @@ impl RemoteStateActor {
             self.trigger_holepunching();
         }
     }
-    fn handle_connection_close(&mut self, conn_id: ConnId) {
+    fn handle_connection_close(&mut self, conn_id: ConnId, reason: ConnectionError) {
```
flub, line 572: could we add the random missing newline in front of this while we're here?

flub (approved):
yay for more people getting use out of the events!

PR #4022, opened 2026-03-13
refactor: Remove special fast path for hickory resolver

rklaehn:
## Description

Remove special fast path for hickory resolver

The downside is that we have 1 more box for each lookup_ipv6 or lookup_ipv4, but I doubt that it matters. These are network ops, maybe cache hits, but in any case much more heavyweight than a boxed future.

The upside is that hickory will be easier to feature gate later, which we need for embedded. And it's less code.

Rationale for this: hickory is really nasty on embedded. It requires giant (for embedded) stack sizes and also is a big dep. We need it for address_discovery_dns, but there is an alternative for that: just use pkarr for both put and get.

If we don't have a default dep on hickory we can implement a simple dns resolver that just uses getaddrinfo. It won't work for TXT records, but we don't need them for the core functionality of doing just plain A or AAAA requests to figure out relay ip addrs.

## Breaking Changes

None

## Change checklist

- [ ] Self-review.
- [ ] Tests if relevant.

### DIFF
--- a/iroh-relay/src/dns.rs
+++ b/iroh-relay/src/dns.rs
@@ -556,7 +556,6 @@ impl HickoryResolver {
         }
         Ok((config, options))
     }
-
 }
 
 impl Resolver for HickoryResolver {
@@ -582,8 +581,11 @@ impl Resolver for HickoryResolver {
         let resolver = self.resolver.clone();
         Box::pin(async move {
             let lookup = resolver.txt_lookup(host).await?;
-            let iter: BoxIter<TxtRecordData> =
-                Box::new(lookup.into_iter().map(|txt| TxtRecordData::from_iter(txt.iter().cloned())));
+            let iter: BoxIter<TxtRecordData> = Box::new(
+                lookup
+                    .into_iter()
+                    .map(|txt| TxtRecordData::from_iter(txt.iter().cloned())),
+            );
             Ok(iter)
         })
     }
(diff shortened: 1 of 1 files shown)


Frando:
When doing https://github.com/n0-computer/iroh/pull/3473 I had this first too, but @dignifiedquire had perf objections so we added the fast path. Not sure myself if its worth it. But featureflagging the enum variant shouldn't also be too hard. So, fine with me, but let's wait for feedback from dig?

rklaehn:
> When doing #3473 I had this first too, but @dignifiedquire had perf objections so we added the fast path. Not sure myself if its worth it. But featureflagging the enum variant shouldn't also be too hard. So, fine with me, but let's wait for feedback from dig?

Yeah, feature flagging the enum variant is of course possible.

But I somewhat dread that feature flagging too many enum variants will turn the codebase into an effectively readonly maze of feature flags.

And I would be seriously surprised if the boxed future makes a difference. I am the first person to be annoyed about boxed futures, but really only in super hot paths, like the middle of a hash function or a 64 byte write in iroh-blobs. For an entire DNS lookup I would bet that you can't even see the diff in a flamegraph, even if the lookup hits a cache.

dignifiedquire:
it’s fine, we‘ll fix perf if it becomes an issue

dignifiedquire (approved):

